PR #7585, opened 2022-04-12
Scalability data generation

lorerod:
# Description

This script is to generate Brac config test data.
Initially it only creates places, persons, users, pregnancy, assessment and assessment follow-up reports. Taking into account hierarchies.

Changes where made in `/test/factories/` and `/test/scalability/ `
Basically added new folders: `/test/factories/brac` and `/test/scalability/generate_brac_data`
The script has a readme on `/test/scalability/generate_brac_data`
I don't have an issue associated with this.

# Code review checklist

# License


tests/scalability/generate_brac_data/generate-brac-data.js
```diff
@@ -0,0 +1,217 @@
...
+  fs.writeFile(directory, content, err => {
+    if (err) {
+      console.error('CreateDataDoc ' + err);
+      return;
+    }
+  });
+};
+
+const createDataDirectory = (path, directoryName) => {
+  if (!fs.existsSync(path + directoryName)) {
```
dianabarsan, line 54: It's also best practice to not rely on `existsSync` before doing other operations on that path: 
> Using fs.exists() to check for the existence of a file before calling fs.open(), fs.readFile() or fs.writeFile() is not recommended. Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file does not exist.

(source: https://nodejs.org/api/fs.html#fsexistspath-callback)

So it's best if this becomes something like: 
```javascript
const createDataDirectory = async (directoryPath, directoryName) => {
  try {
    await fs.promises.mkdir(path.join(directoryPath, directoryName));
  } catch (err) {
    if (err.code === 'EEXIST') {
      return;
    }
    console.error('CreateDataDirectory ' + directoryName + ' failed - ' + err);
    throw err;
  }
};
```

tests/scalability/generate_brac_data/generate-brac-data.js
```diff
@@ -0,0 +1,217 @@
...
+    place.supervisor = managers[randomIndex]._id;
+  }
+  if (type === 'clinic') {
+    parentPersonFirstLevelId = place._id;
+    parentPersonSecondLevelId = parentFirstLevelId;
+    parentPersonThirdLevelId = parentSecondLevelId;
+  }
+
+  let isPrimaryContact = true;
+  for (let i = 0; i < numberOfPersons; i++) {
```
dianabarsan, line 121: This is a veeery big function, please break it appart into smaller chunks. 
For example, extract the contents of this for in a function: 
```javascript
const generatePerson = (type, place, isPrimaryContact) => { .... };
const generateUser = (person) => { ... };
const generateReports = (person) => { ... };
```

So this turns into: 
```javascript
const primaryContact = generatePerson(type, place, true);
if (needUser) {
  generateUser(primaryContact);
}
for (let i = 0; i < numberOfPersons - 1; i++) {
  generatePerson(type, place, false);
}
```

tests/scalability/generate_brac_data/generate-brac-data.js
```diff
@@ -0,0 +1,217 @@
...
+      createDataDoc(preconditionDataDirectory + place._id + dataExtension, JSON.stringify(place, {}, 2));
+      if (type === 'district_hospital') {
+        roles = ['national_admin', 'mm-online'];
+      }
+    }
+    createDataDoc(preconditionDataDirectory + person._id + dataExtension, JSON.stringify(person, {}, 2));
+    if (type === 'district_hospital') {
+      managers.push(person);
+    }
+    if (needUser) {
```
dianabarsan, line 140: Do you need a user for every person that is added, or just the primary contact?

tests/scalability/generate_brac_data/generate-brac-data.js
```diff
@@ -0,0 +1,217 @@
...
+    }
+    //For every even index number creates a report for the precondition script data
+    //For every odd index number creates a report for the main script data
+    if (person.family_member_type === 'member_eligible_woman' || person.family_member_type === 'member_child') {
+      let reportsDirectory = preconditionDataDirectory;
+      if (i % 2 === 0) {
+        reportsDirectory = preconditionDataDirectory;
+      } else {
+        createDataDirectory(mainDataDirectory, directParentPlace.contact._id);
+        reportsDirectory = mainDataDirectory + directParentPlace.contact._id + '/';
```
dianabarsan, line 162: Please use `path.join` to generate paths.

tests/scalability/generate_brac_data/generate-brac-data.js
```diff
@@ -0,0 +1,217 @@
...
+        place: personUser.facility_id
+      };
+      users.push(user);
+    }
+    //For every even index number creates a report for the precondition script data
+    //For every odd index number creates a report for the main script data
+    if (person.family_member_type === 'member_eligible_woman' || person.family_member_type === 'member_child') {
+      let reportsDirectory = preconditionDataDirectory;
+      if (i % 2 === 0) {
+        reportsDirectory = preconditionDataDirectory;
```
dianabarsan, line 159: This is a redundant assignment:
```suggestion
let reportsDirectory = preconditionDataDirectory;
if (i % 2 !== 0) {
  reportsDirectory = path.join(mainDataDirectory, directParentPlace.contact._id);
  await createDataDirectory(reportsDirectory);
}
```

tests/scalability/generate_brac_data/generate-brac-data.js
```diff
@@ -0,0 +1,217 @@
...
+      } else {
+        createDataDirectory(mainDataDirectory, directParentPlace.contact._id);
+        reportsDirectory = mainDataDirectory + directParentPlace.contact._id + '/';
+      }
+
+      if (person.family_member_type === 'member_eligible_woman') {
+        if (person.group_other_woman_pregnancy.other_woman_pregnant) {
```
dianabarsan, line 166: Please flatten this and only use one `if` condition. It would also be very useful if `member_eligible_woman` and `member_child` were not used here, and instead we would export something from the brac-person file like `shouldGeneratePregnancySurvey(person)` and `shouldGenerateAssessmentSurvey(person)`.

tests/scalability/generate_brac_data/generate-brac-data.js
```diff
@@ -0,0 +1,217 @@
...
+          createDataDoc(reportsDirectory + pregnancySurvey._id + dataExtension, JSON.stringify(pregnancySurvey, {}, 2));
+        }
+      }
+      if (person.family_member_type === 'member_child') {
+        const assesmentSurvey = bracSurvey.generateBracSurvey('assesment', directParentPlace, place, person);
+        createDataDoc(reportsDirectory + assesmentSurvey._id + dataExtension, JSON.stringify(assesmentSurvey, {}, 2));
+
+        const assesmentFollowUpSurvey = bracSurvey.generateBracSurvey(
+          'assesment_follow_up', directParentPlace, place, person);
+        createDataDoc(reportsDirectory + assesmentFollowUpSurvey._id + dataExtension,
```
dianabarsan, line 177: Please always use `path.join` or `path.resolve` to generate file system paths.

tests/scalability/generate_brac_data/README.md
```diff
@@ -0,0 +1,46 @@
...
+## Introduction
+
+This script is to generate Brac config test data.
+
+Initially it only creates places, persons, users and reports (pregnancy, assessment and assessment follow up). Taking into account hierarchies.
+
+Changes where made in /test/factories/ and /test/scalability/
+Basically added new folders:
+/test/factories/brac and /test/scalability/generate_brac_data
+
```
dianabarsan, line 12: I think the readme shouldn't mention changes, rather it should describe what is currently present and how it is used.
```markdown
Relevant factories exist in `test/factories/brac`. 
```

tests/scalability/generate_brac_data/README.md
```diff
@@ -0,0 +1,46 @@
...
+## Requirements
+
+- NodeJS
+- a target medic instance with brac config loaded
+
+
+## Steps to execute it
+1. Clone cht-core repo.
+2. Run npm ci, npm install and npm install @faker-js/faker --save-dev in cht-core root directory.
```
dianabarsan, line 21: These dependencies should be added in the root package, so that `npm ci` would suffice.
Alternatively, you could just install it in this folder, and require running `npm ci` here as well.

tests/scalability/generate_brac_data/README.md
```diff
@@ -0,0 +1,46 @@
...
+3. Go to `/tests/scalability/generate_brac_data` and edit the `data_directory` attribute in `data-config.json` doc.
```
dianabarsan, line 22: Can the data directory be passed as a parameter to the `generate-brac-data.js` script, and avoid needing to edit paths? This way we won't have people who generate this data accidentally commit the changes.

dianabarsan (changes requested):
Wow, awesome work! Given that this is the first time you've written in Javascript, this is very impressive! 

Most of my comments are about readability and consistency and duplication (and most repeat themselves :D). 
Please let me know if there are some things that you'd like to discuss or need more clarity.

tests/scalability/generate_brac_data/generate-brac-data.js
lorerod: For every person created at District Hospital and Health Center level.
```
const pairPlaceTypesNeedsUsers = {
  'district_hospital': true,
  'health_center': true,
  'clinic': false
};
```

tests/factories/brac/contacts/brac-place.js
```diff
@@ -0,0 +1,45 @@
...
+const Faker = require('@faker-js/faker');
+const householdSurveyFactory = require('../reports/create-family-household-survey');
+
+const bracPlace = () => {
+  return new Factory()
+    .sequence('_id', uuid.v4)
+    .attr('parent', '')
+    .attr('type', '')
+    .attr('is_name_generated', 'true')
+    .attr('name', '')
```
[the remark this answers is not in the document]
lorerod: I thought it would be a good idea if the process defines the place name based on the thread name to ensure that the place name is always unique (line 13 on generate-brac-data.js). What do you think?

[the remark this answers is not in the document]
dianabarsan: That sounds good :)

tests/factories/brac/reports/brac-assessment-follow-up.js
```diff
@@ -0,0 +1,140 @@
...
+      return groupBetter;
+    })
+  .attr('referral_follow_up_needed',
+    ['group_better', 'group_improved', 'group_danger_signs'],
+    (groupBetter, groupImproved, groupDangerSigns) => {
+      if ((groupBetter !== null && groupBetter.g_patient_better === 'no') ||
+        (groupImproved !== null && groupImproved.g_patient_treatment_outcome === 'bad_medicine_reaction') ||
+        (groupImproved !== null && groupImproved.g_patient_treatment_outcome === 'not_improving') ||
+        (groupDangerSigns !== null && groupDangerSigns.danger_signs !== null)) {
+        return true;
```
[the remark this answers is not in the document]
lorerod: You are right. Is a calculated field, and it can take string values of 'true' or 'false'

tests/factories/brac/contacts/brac-person.js
```diff
@@ -0,0 +1,214 @@
...
+    })
+    .attr('group_other_woman_pregnancy', ['pregnant_at_registration', 'subtype'], (pregnantAtRegistration, subtype) => {
+      if (subtype !== 'member_eligible_woman') {
+        return null;
+      } else {
+        return groupOtherWomanPregnancy.build({ other_woman_pregnant: pregnantAtRegistration });
+      }
+    })
+    .attr('c_name', ['subtype', 'name'], (subtype, name) => {
+      if (subtype === 'chw' || subtype === 'manager') {
```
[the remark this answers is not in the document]
lorerod: Yes! I like this. Will do something like this to make it clear.

tests/factories/brac/contacts/brac-person.js
```diff
@@ -0,0 +1,214 @@
...
+    .attr('reported_date', () => Date.now())
+    .attr('imported_date', () => moment().format('YYYY-MM-DD'))
+    .attr('has_disability', Faker.faker.datatype.boolean())
+    .attr('family_member_type', ['subtype'], (subtype) => {
+      if (subtype !== 'member_child' && subtype !== 'member_eligible_woman') {
+        return null;
+      } else {
+        return subtype;
+      }
+    })
```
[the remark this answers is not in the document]
dianabarsan: that sounds good!

tests/factories/brac/reports/brac-pregnancy.js
```diff
@@ -0,0 +1,268 @@
...
+    return groupAncVisit.g_anc_last_visit;
+  })
+  .attr('anc_visit_identifier', '')
+  .attr('anc_last_bp_reading', '')
+  .attr('patient_age_at_lmp', ['patient', 'group_lmp'], (patient, groupLmp) => {
+    const birthDate = moment(patient.date_of_birth);
+    const lmpDate = moment(groupLmp.g_lmp_date_8601);
+    const yearsDiff = birthDate.diff(lmpDate, 'years');
+    const ageAtLmp = moment(birthDate).subtract(yearsDiff);
+    return ageAtLmp;
```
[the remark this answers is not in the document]
lorerod: Yes it is exactly that. In the form definition the calculation is done like this: floor( difference-in-months( ${date_of_birth}, ${lmp_date_8601} ) div 12 )

tests/factories/brac/reports/brac-assessment.js
```diff
@@ -0,0 +1,438 @@
...
+    return ageInDays;
+  })
+  .attr('patient_age_display',
+    ['patient_age_in_years', 'patient_age_in_months'],
+    (patientAgeInYears, patientAgeInMonths) => {
+      const patientAgeDisplay = patientAgeInYears + ' years and ' + patientAgeInMonths % 12 + ' months';
+      return patientAgeDisplay;
+    })
+  .attr('group_assess', () => {
+    const isAlive = Faker.faker.random.arrayElement(['yes']);
```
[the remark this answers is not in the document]
lorerod: Yes, you are right. I will add 'no'. I remember thinking if 'no' is selected, then I have to generate a death report as well. But that report is not done yet.

tests/factories/brac/reports/brac-assessment.js
```diff
@@ -0,0 +1,438 @@
...
+            Faker.faker.datatype.number({ min: 1, max: 6 }));
+        }
+        groupImmLess2mo = {
+          vaccines_received_2mo: vaccinesReceived2mo,
+          imm_current_2mo: immCurrent2mo,
+          imm_given_2mo: immGiven2mo
+        };
+      }
+
+      if (patientAgeInMonths > 2 && patientAgeInMonths <= 9) {
```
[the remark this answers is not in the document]
lorerod: Ok. I think infant is until they can walk, and then toddler. Even if this doesn't correspond exactly I can use those. 
What do you think?

tests/factories/brac/reports/brac-assessment.js
```diff
@@ -0,0 +1,438 @@
...
+      if (!isAChildAndAlive(5, patientAgeInMonths, patientAgeInYears, groupAssess.is_alive)) {
+        return null;
+      }
+      let groupUnder2yr = null;
+      let groupFoodEaten = null;
+      let micronutrient = null;
+      let numSatchets = null;
+      let buyMpn = null;
+      let mpnNum = null;
+      if ((patientAgeInMonths > 6 && patientAgeInMonths <= 59)) {
```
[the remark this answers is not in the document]
lorerod: 😄 I will put < 60. Actually I just copy this from the xlsx: 
(${patient_age_in_months}   > 6 and ${patient_age_in_months} <= 59)

tests/factories/brac/reports/brac-assessment.js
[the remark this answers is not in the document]
lorerod: Or we can use the same terminology as brac config: 
less_2mo
2mo_9mo
9mo_18mo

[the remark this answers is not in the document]
dianabarsan: That works as well, it's definitely easier to read than the conditional statements.

lorerod:
@dianabarsan This is ready for another round of review. I can create a PR now if you prefer. Thanks!

tests/scalability/generate_brac_data/generate-brac-data.js
```diff
@@ -0,0 +1,221 @@
...
+    const districtHospital = generateHierarchy(
+      'district_hospital',
+      districtHospitalName + 'districthospital' + dh,
+      numberOfManagersPerDistrictHospital);
+
+    for (let hc = 0; hc < numberOfHealthCentersPerDistrictHospital; hc++) {
+      const healthCenter = generateHierarchy(
+        'health_center',
+        districtHospitalName + 'districthospital' + dh + 'healthcenter' + hc,
+        numberOfChwPerHealthCenter, districtHospital._id);
```
dianabarsan, line 209: This argument list is really long. Can you please define the name of the place separately?  

```suggestion
const healthCenterName = districtHospitalName + 'districthospital' + dh + 'healthcenter' + hc;
const healthCenter = await generateHierarchy('health_center', healthCenterName, numberOfChwPerHealthCenter, districtHospital._id);
```

tests/scalability/generate_brac_data/generate-brac-data.js
```diff
@@ -0,0 +1,221 @@
...
+
+      for (let c = 0; c < numberOfClinicsPerHealthCenter; c++) {
+        generateHierarchy('clinic',
```
dianabarsan, line 212: Please avoid inconsistent indentation, where some lines contain one argument, and other contain multiple arguments. It makes reading the code really difficult. Please, either inline all arguments, or one argument per line, like:

````javascript
function(arg1, arg2, arg3, arg4); 
function(   
  arg1, 
  arg2,
  arg3, 
  arg4
);
````

```suggestion
const clinicName = districtHospitalName + 'districthospital' + dh + 'healthcenter' + hc + 'clinic' + c;
await generateHierarchy('clinic', clinicName, numberOfFamilyMembers, healthCenter._id, districtHospital._id, healthCenter);

```

tests/scalability/generate_brac_data/generate-brac-data.js
```diff
@@ -0,0 +1,221 @@
...
+  } catch (err) {
+    if (err.code === 'EEXIST') {
+      return;
+    }
+    console.error('CreateDataDirectory ' + directoryName + ' failed - ' + err);
+    throw err;
+  }
+};
+
+const setParents = (firstLevelId, secondLevelId, thirdLevelId) => {
```
dianabarsan, line 69: We typically call this "lineage", because you define parent, grandparent and greatgrandparent. To make it consisteny with the rest of our codebase, can you rename this to: `getLineage` ?

tests/scalability/generate_brac_data/generate-brac-data.js
```diff
@@ -0,0 +1,221 @@
...
+
+const generateHierarchy = (type, placeName, numberOfPersons,
+  parentFirstLevelId, parentSecondLevelId, directParentPlace) => {
+  const place = bracPlaceFactory.generateBracPlace(placeName, type,
+    setParents(parentFirstLevelId, parentSecondLevelId));
+  let parentPersonThirdLevelId = null;
+  let parentPersonSecondLevelId = null;
+  let parentPersonFirstLevelId = null;
+  if (type === 'district_hospital') {
+    parentPersonFirstLevelId = place._id;
```
dianabarsan, line 161: It looks like `parentPersonFirstLevelId` is always set to `place._id`. Can we directly define it?

tests/scalability/generate_brac_data/generate-brac-data.js
```diff
@@ -0,0 +1,221 @@
...
+};
+
+const generateHierarchy = (type, placeName, numberOfPersons,
+  parentFirstLevelId, parentSecondLevelId, directParentPlace) => {
+  const place = bracPlaceFactory.generateBracPlace(placeName, type,
+    setParents(parentFirstLevelId, parentSecondLevelId));
+  let parentPersonThirdLevelId = null;
+  let parentPersonSecondLevelId = null;
+  let parentPersonFirstLevelId = null;
+  if (type === 'district_hospital') {
```
dianabarsan, line 160: I find following these 3 variables, and what they're supposed to represent, pretty hard. 
Could we define the lineage directly? 
```javascript
let personLineage = null;
if (type === 'district_hospital') {
  personLineage = getLineage(place._id);
}
if (type === 'health_center') {
  personLineage = getLineage(place._id, parentFirstLevelId);
}
if (type === 'clinic') {
  personLineage = getLineage(place._id, parentFirstLevelId, parentSecondLevelId);
}
```

tests/scalability/generate_brac_data/generate-brac-data.js
```diff
@@ -0,0 +1,221 @@
...
+    password: personUser.password,
+    roles: personUser.roles,
+    contact: person._id,
+    phone: person.phone,
+    place: personUser.facility_id
+  };
+  users.push(user);
+};
+
+const generateReports = (directParentPlace, place, person, isMainData) => {
```
dianabarsan, line 133: directParentPlace? Can this just be "parentPlace"?

tests/scalability/generate_brac_data/generate-brac-data.js
```diff
@@ -0,0 +1,221 @@
...
+    createDataDoc(reportsDirectory, assesmentSurvey._id + dataExtension, JSON.stringify(assesmentSurvey, {}, 2));
+    const assesmentFollowUpSurvey = bracSurvey.generateBracSurvey(
+      'assesment_follow_up', directParentPlace, place, person);
+    createDataDoc(reportsDirectory, assesmentFollowUpSurvey._id + dataExtension,
+      JSON.stringify(assesmentFollowUpSurvey, {}, 2));
+  }
+};
+
+const generateHierarchy = (type, placeName, numberOfPersons,
+  parentFirstLevelId, parentSecondLevelId, directParentPlace) => {
```
dianabarsan, line 154: I think that we can extract the lineage from the "directParentPlace" object itself, and no need to pass them as extra parameters. 
```javascript
const generateHierarchy = async (type, placeName, parentPlace, numberOfPersons) => {
....
  const placeLineage = { _id: parent._id, parent: parentPlace.parent };
  const place = bracPlaceFactory.generateBracPlace(placeName, type, placeLineage);
  const personLineage = { _id: place._id, parent: place.parent };
  .....
}
```

tests/scalability/generate_brac_data/generate-brac-data.js
```diff
@@ -0,0 +1,221 @@
...
+        numberOfChwPerHealthCenter, districtHospital._id);
+
+      for (let c = 0; c < numberOfClinicsPerHealthCenter; c++) {
+        generateHierarchy('clinic',
+          districtHospitalName + 'districthospital' + dh + 'healthcenter' + hc + 'clinic' + c,
+          numberOfFamilyMembers, healthCenter._id, districtHospital._id, healthCenter);
+      }
+    }
+  }
+  csvWriter.writeRecords(users);
```
dianabarsan, line 218: This is an async function that needs to be awaited.

dianabarsan (changes requested):
Awesome work! Thanks for implementing all the changes and adding explanatory comments! 

I've left some comments inline, mostly about readability. 
Also, I believe you've used npm v8+ when installing your dependencies. This ends changing the lock file version to 2 and changing the contents. We generally want to be consistent about lock files across the codebase. 
We are going to switch to only support npm v8+ with 4.0, but until then, I think we should stay on the old version of lock files. 

Could you, please, downgrade your npm to 6, rollback the package changes and install the two dependencies again? 
Thank you!

tests/factories/brac/reports/brac-assessment-follow-up.js
```diff
@@ -0,0 +1,137 @@
...
+      }
+      const groupBetter = {
+        g_patient_better: Faker.faker.random.arrayElement(['cured', 'still_recovering', 'still_in_facility', 'died'])
+      };
+      return groupBetter;
+    })
+  .attr('referral_follow_up_needed',
+    ['group_better', 'group_improved', 'group_danger_signs'],
+    (groupBetter, groupImproved, groupDangerSigns) => {
+      if ((groupBetter && groupBetter.g_patient_better === 'no') ||
```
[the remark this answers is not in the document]
lorerod: This is the value on calculation field in assessment_follow_up.xlsx for **referral_follow_up_needed**:
`if (${g_patient_better}='no' or ${g_patient_treatment_outcome}='bad_medicine_reaction' or ${g_patient_treatment_outcome}='not_improving' or count-selected(${g_danger_signs}) > 0, 'true', 'false')`

And the **type** of **g_patient_better** is **select_one patient_referral_outcome**, with the following values:
- cured
- still_recovering
- still_in_facility
- died

Can this be a bug in the form? I think the field to be used in this calculation instead of **g_patient_better** would be **patient_improved**.

What do you think? But this is not going to correspond to the form calculations.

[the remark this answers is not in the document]
dianabarsan: > Can this be a bug in the form

I think it's possible. I think we should signal this to the app services team, or who is managing the brac config. 

In the meantime, I'd rather our code made sense :) 
Either way, we should leave a comment about the discrepancy!

tests/factories/brac/reports/brac-pregnancy.js
```diff
@@ -0,0 +1,273 @@
...
+
+/**
+ * Calculate if women is pregnant.
+ * @param {string} edd the estimate delivery time
+ * @param {number} lmpApprox the approximate start date of last menstrual cycle
+ * @param {string} pregRes the results of the pregnancy test
+ * @param {string} pregResKit the pregnancy kit results.
+ */
+const isPregnant = (edd, lmpApprox, pregRes, pregResKit) => {
+  if (edd && lmpApprox !== UP_TO_2_MONTHS_AGO && lmpApprox !== UP_TO_3_MONTHS_AGO) {
```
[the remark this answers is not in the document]
lorerod: lmpApprox is one of those values. I will clarify in the comment.

lorerod:
@dianabarsan requested changes are ready.

dianabarsan:
I'll have a look next week @lorerod (I'm off the whole day today). 
Also, so I get properly notified and reminded about it, could you please re-request a review in the GitHub UI as well? 
image

Thanks!

tests/scalability/generate_brac_data/generate-brac-data.js
```diff
@@ -0,0 +1,203 @@
...
+const createCsvWriter = require('csv-writer').createObjectCsvWriter;
+const args = process.argv.slice(2);
+const districtHospitalName = args[0] + '-districthospital';
+const numberOfDistrictHospitals = sizeConfig.number_of_district_hospitals;
+const numberOfManagersPerDistrictHospital = sizeConfig.number_of_managers_per_district_hospitals;
+const numberOfHealthCentersPerDistrictHospital = sizeConfig.number_of_health_centers_per_district_hospital;
+const numberOfChwPerHealthCenter = sizeConfig.number_of_chw_per_health_center;
+const numberOfClinicsPerHealthCenter = sizeConfig.number_of_clinics_per_health_center;
+const numberOfFamilyMembers = sizeConfig.number_of_family_members;
+const dataDirectory = args[1];
```
dianabarsan, line 18: Something is amiss here. This file is called like: `node generate-brac-data.j [data-directory-path]` (As per the readme), so we never have two arguments.

Is the first argument supposed to be some name? 
Right now, if I follow the readme I get this error: 
```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
```

Also, I think we should add a nice check for the existence of the data directory passed to the script. Right now, if I pass a nonexistent path, I get weird errors about not being able to create other directories.

dianabarsan (commented):
Awesome work! 
All my comments are just tiny tiny improvements.

tests/scalability/generate_brac_data/generate-brac-data.js
lorerod: I actually added 2 more parameters: name and instance.
`generate-brac-data.js [district-hospital-base-name] [data-directory-path] [instance]`
**district-hospital-base-name:** is going to be used as a name base to generate all the hierarchy, this having in mind we can parallelize the execution (please read comment about the limitations of this feature right now).
**instance:** is going to be used to generate a config.json file for the next step script, with a url to the instance and a list of generated users info.

- [X] Added the check for the existence of the data directory.
- [X]  README.md updated

tests/scalability/replicate-brac-docs/add-docs-to-remote.js
```diff
@@ -0,0 +1,48 @@
+const args = process.argv.slice(2);
+const threadId = args[0];
+const dataDirectory = args[1];
+const config = require('./config.json');
+
+const dataConfig = require('../generate_brac_data/data-config.json');
```
dianabarsan, line 6: I'm uncertain where is file is used. Is this for your local testing?

lorerod: If I'm understanding your question correctly, dataDirectory has to be the base path where all the docs where generated in the generate-brac-data.js script.Maybe what I can do is in that script add that path to the config.json that is also generated.

dianabarsan: Oh, I'm sorry, I was referring to the whole `add-docs-to-remote.js` file. I don't see it used anywhere, is it?

lorerod: Yes, sorry about that. I didn't finish the jmeter script. Now I added the call there.

dianabarsan (approved):
Excellent work!

lorerod:
Thanks for your help @dianabarsan !

Issue #6282, opened 2020-03-09
Remove the Theme page


garethbowen:
Nobody knows it's there, and it's not very useful once we have a design system up and running. Removing it will slightly improve performance by reducing code size.

@n-orlowski Do you think it's useful for anyone?

n-orlowski:
@garethbowen currently useful for me to refer back to but you're right, ultimately this will be replaced

garethbowen:
@n-orlowski Would you prefer we held off removing it for a while?

n-orlowski:
@garethbowen took a closer look and am actually ok to remove as the theme page doesn't actually all align with the live design

<img width="885" alt="Screen Shot 2020-03-11 at 10 29 23 AM" src="https://user-images.githubusercontent.com/59581790/76445904-6e436e80-6383-11ea-96d2-178c2c1d76ad.png">

n-orlowski:
@garethbowen chatted with Marc about the future state with this page and it being used potentially for release testing; can we leave this up until we have the design system up and running and can confidently replace it?

garethbowen:
Absolutely. I've tentatively scheduled it for 4.0.0 which gives us some time to get the design system going.

njogz:
Ready for AT on this [PR](https://github.com/medic/cht-core/pull/7587)

ngaruko:
Page removed...
![image](https://user-images.githubusercontent.com/6979995/163939910-42550909-10a1-486e-a2b2-52a5a3b5e52f.png)

Other pages/links work as expected

ngaruko:
Ready to merge @njogz

PR #7587, opened 2022-04-13
remove theme page

njogz:
# Description

Remove theme page since it's no longer required

medic/cht-core#6282

# Code review checklist

# License


latin-panda (changes requested):
I think we also need to clean the css
https://github.com/medic/cht-core/blob/fa40f745064ef48c71537e08299e03e0635a63ae/webapp/src/css/inbox.less#L1036 
and 
https://github.com/medic/cht-core/blob/fa40f745064ef48c71537e08299e03e0635a63ae/webapp/src/css/inbox.less#L1233

latin-panda (approved):
Nice!

Issue #2967, opened 2016-11-30
Report API startup progress instead of 502


garethbowen:
As we don't start API until all the migrations have completed users are greeted with a 502 error after update which makes it unclear whether they should wait or try and work out what went wrong.

Find some way to server an "API is starting" page instead. Ideally this would have some indication of progress (eg: "currently running xyz migration").

SCdF:
So when I do this locally and I boot too fast I don't get a 502, I get a 404. Maybe my timing is just different though.

garethbowen:
It's probably different via nginx?

nice-snek:
hi frand @browndav, U is coot

Please close or schedule before the end of this sprint. See [triaging old issues](https://github.com/medic/medic-docs/blob/master/md/dev/workflow.md#triaging-old-issues).

Nice Snek

garethbowen:
I've spiked this to see if it would just work, and it does. Something like this...

```js
const status = {};
app.listen(apiPort);
app.get('/status', (req, res) => res.json(status));
migrationsAndChecksAndSoOn().then(err => {
  status.error = err;
  if (!err) {
    registerAllListeners();
  }
});
```

garethbowen:
Unassigning this so anyone can work on it.

garethbowen:
This would also be useful for horticulturist (#3582) in the updating UI.

garethbowen:
Working on this is a good chance to remove the `medic-api` branding from the `/setup/poll` API, just be sure to update it in a backwards compatible way so as not to break [cht-android AppUrlVerifier](https://github.com/medic/cht-android/blob/master/src/main/java/org/medicmobile/webapp/mobile/AppUrlVerifier.java#L25).

garethbowen:
Assigned to @dianabarsan as I think this is part of the archv3 she's done/doing.

dianabarsan:
Relevant express comment: https://github.com/expressjs/express/issues/2596#issuecomment-81353034

dianabarsan:
This was resolved by starting API server sooner, but with a temporary router that responds only in a couple of ways: 
- `/api/v1/startup-progress` returns startup progress
- any request that wants JSON will get a 503 status and a `{ error: 'Service unavailable' }` body
- every other request will be served with an HTML page that displays startup progress

[...]

PR #7588, opened 2022-04-13
API startup status page

dianabarsan:
# Description

Starts API server immediately, but with a different router that: 
- ends a 503 and error message if the request wants JSON 
- resolves with an html page otherwise
- the html page periodically queries an endpoint `/progress`, and updates the startup progress: 

image

The router replaced with the previous router at runtime, after all startup steps have been completed. 

**optionally** 
I've also added a warm views startup step that runs even when the installation is valid. This is so the startup page shows up even when, for example, a full install has been replicated to a new database and none of the views are warmed. 
This has a significant drawback, because any potential and insignificant API restart could mean downtime if certain views are not used. We could select a handful of "required" views that should always be up to date, or not warm views at startup at all. 

medic/cht-core#2967

# Code review checklist

# License


garethbowen (changes requested):
This makes me so happy!

Some suggestions inline.

With regards to view warming I think the ideal would be to block on the views that API needs (from memory, docs by type). Not sure how complicated this is.

That also highlights an issue with translating the setup page, because [translations currently require the docs by type view](https://github.com/medic/cht-core/blob/master/api/src/translations.js#L99). This should be changed to fetch by ID prefix, but whether this is part of this PR or an optional extra I leave up to you...

.github/workflows/build.yml
```diff
@@ -125,7 +125,7 @@ jobs:
       if: ${{ github.event_name != 'pull_request' }}
       run: |
         cd scripts/ci
-        npm ci
+        npm ci --legacy-peer-deps
```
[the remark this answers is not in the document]
dianabarsan: The npm version that gets installed on CI. :( 
The alternative is to fix the package details (I wanted to keep that change spam separate).

api/server.js
```diff
@@ -72,6 +92,7 @@ process
     process.exit(1);
   }
 
+  router = require('./src/routing');
```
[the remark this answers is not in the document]
dianabarsan: That the only way that it works. Hotswapping the reference to the router that the app uses.

api/src/public/setup/setup.html
```diff
@@ -0,0 +1,145 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>The CHT is starting up</title>
```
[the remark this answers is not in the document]
dianabarsan: If we want to display the page before server checks, we have absolutely no access to branding (because Couch might not even be responding).

api/src/public/setup/setup.html
```diff
@@ -0,0 +1,145 @@
...
+      height: 20px;
+    }
+
+    .loader.loading {
+      border-left-color: rgba(255, 255, 255, 0.8);
+    }
+
+    .loader.done {
+      border-color: #218E7F;
+    }
```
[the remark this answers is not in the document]
dianabarsan: Solid ball sounds nice! Thanks!

api/server.js
```diff
@@ -24,23 +29,36 @@ process
     process.exit(1);
   }
 
+  const app = express();
```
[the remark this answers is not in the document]
dianabarsan: If we include server checks here, then we can't rely on CouchDb to be accessible, so no branding :(

dianabarsan:
> That also highlights an issue with translating the setup page, because [translations currently require the docs by type view](https://github.com/medic/cht-core/blob/master/api/src/translations.js#L99).That also highlights an issue with translating the setup page, because [translations currently require the docs by type view](https://github.com/medic/cht-core/blob/master/api/src/translations.js#L99).

This has already been changed in the arch v3 branch, we do an _all_docs, however translations need to be extracted from local files, especially on a fresh install, and won't be available for translating. 
Same goes for branding. 
How do you think we should go about that?

dianabarsan:
Another issue that exists is that I can't require the `db` library before server checks have passed, specifically because it tries to create databases immediately (the vault, for example, with the recent changes to how we store credentials), and that complicates things even more.

api/src/public/setup/setup.html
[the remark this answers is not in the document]
dianabarsan: image

api/src/public/setup/setup.html
```diff
@@ -0,0 +1,145 @@
...
+      height: 100%;
+      padding: 0;
+      margin: 0;
+    }
+    body, input {
+      font-family: "Helvetica Neue",Helvetica,Arial,sans-serif;
+    }
+    .center {
+      padding-top: 2em;
+      width: 600px;
```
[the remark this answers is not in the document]
dianabarsan: Mobile view. Updated the css and added the missing viewport meta. 

image

dianabarsan:
> With regards to view warming I think the ideal would be to block on the views that API needs (from memory, docs by type)

> That also highlights an issue with translating the setup page, because [translations currently require the docs by type view](https://github.com/medic/cht-core/blob/master/api/src/translations.js#L99).
I know I previously said translations no longer use the view, they don't, but the [login controller](https://github.com/medic/cht-core/blob/master/api/src/controllers/login.js#L76) still does. 

Additionally, 
- [migrations main lib uses doc_by_type](https://github.com/medic/cht-core/blob/master/api/src/migrations.js#L23)
- some migrations use `contacts_by_reference`, `contacts_by_phone`, `registered_patients`, `contacts_by_type`
- config-watcher loads translations with `doc_by_type`

I think it's quite safe to assume we need `medic-client` views indexed, so I'll just change it so only it gets indexed on startup.

api/server.js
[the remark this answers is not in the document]
garethbowen: Fair point. The other idea I had is to move this to the Docker container startup checks, but that can be done in later change. Let's leave it as is for now.

api/src/public/setup/setup.html
```diff
@@ -0,0 +1,157 @@
...
+      font-size: 1em;
+      height: 100%;
+      padding: 0;
+      margin: 0;
+      overflow-y: hidden;
+
+    }
+    body, input {
+      font-family: "Noto Sans","Helvetica Neue",Helvetica,Arial,sans-serif;
+    }
```
garethbowen, line 89: Copying from webapp, this should be something like...

```suggestion
    @font-face {
      font-family: 'Noto';
      src: url(/fonts/NotoSans-Regular.ttf) format("truetype");
      font-weight: normal;
    }
    body, input {
      font-family: Noto, sans-serif;
    }
```

garethbowen: But.... I just realised that route ("/fonts/NotoSans-Regular.ttf") won't be available yet. Is it difficult to include it in the router?

dianabarsan:
> If we drop all those then I don't think we need to wait for view warming at all, and in fact I think it's better that we don't.
Completely agree. I'll make the changes. 

> Can't we do the extraction first, even before rendering the page?
I'll swap the steps.

api/src/templates/setup/setup.html
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        const listElement = createProgressListItem(action);
+        actionsElem.appendChild(listElement);
+      });
+    }
+
+    const getProgress = async () => {
+
+      const result = await fetch('/progress', { headers: { 'accept-language': locale } });
```
garethbowen, line 54: I think it would be cleaner to render all the HTML with all stages and translations using lodash templating, and then use this function to just toggle the CSS classes. That'll mean less login in JS and more handed off the the templating engine.

api/src/services/setup/router.js
```diff
@@ -0,0 +1,79 @@
...
+};
+
+const renderStartupPage = async (req) => {
+  const acceptLanguageHeader = req && req.headers && req.headers['accept-language'];
+  const enabledLocales = await getEnabledLocales();
+  const locale = getBestLocaleCode(acceptLanguageHeader, enabledLocales, config.get('locale'));
+
+  const templatePath = path.join(environment.templatePath, 'setup', 'setup.html');
+  const fileContents = await fs.promises.readFile(templatePath, 'utf8');
+  const template = _.template(fileContents);
```
garethbowen, line 54: Best practice is to cache the template, just in case all 100k users try to load the page at the same time. Something like...

```
let template;

const getTemplate = () => {
  if (template) {
    return template;
  }
  const templatePath = path.join(environment.templatePath, 'setup', 'setup.html');
  template = fs.promises.readFile(templatePath, 'utf8');
    .then(fileContents => _.template(fileContents));
  return template;
};
```

api/src/services/setup/router.js
```diff
@@ -0,0 +1,79 @@
...
+  return template({
+    title: config.translate('api.startup.title', locale, { branding: await getBranding() }),
+    locale: locale,
+  });
+};
+
+router.get('/progress', (req, res) => {
+  const locale = req && req.headers && req.headers['accept-language'];
+  res.json(startupLog.getProgress(locale));
+});
```
garethbowen, line 65: How do you feel about exposing this as a documented API so it's available for scripting against? We have other endpoints that do that too, but they're crude in comparison to this one, so we could deprecate them and make this the new hotness. If you agree with my other suggestion about doing localisation in the rendering step rather than in the result of this call, then I think it'd be a really clean API to script to.

That would mean changing it to `/api/v1/startup-progress` (or something), and exposing it from the other router too, so even when the app is up you get a response from this endpoint. Then it would just need a little documentation to be useful to others!

api/src/templates/setup/setup.html
```diff
@@ -0,0 +1,163 @@
...
+      }
+
+      if (!result.ok) {
+        console.error('Error when fetching progress', await result.text());
+        return;
+      }
+
+      const progress = await result.json();
+      clearList();
+      populateList(progress);
```
garethbowen, line 67: Have you thought about redirecting users to the login page once the setup is complete? For non-admin users trying to log in that's what I'd expect.

garethbowen (changes requested):
:smiling_face_with_three_hearts: 

This is so cool. A few minor (I think) suggestions inline then it'll be golden!

api/src/services/setup/router.js
garethbowen: This is becoming a repeating pattern - I was just writing this code up for a different issue I'm working on. Maybe it's worth having a library that abstracts the reading and caching for reuse?

api/src/templates/setup/setup.html
dianabarsan: > Have you thought about redirecting users to the login page once the setup is complete?

This already happens. The request to get progress fails and the page reloads, and loads the login page.

api/src/templates/setup/setup.html
dianabarsan: I would have liked to have the steps in a single place instead of two. But I'll give this a shot :)

api/src/services/setup/router.js
dianabarsan: I added a very simple template service.

api/src/middleware/locale.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  getLocale: (req, res, next) => {
+    req.locale = req && req.headers && req.headers['accept-language'];
```
garethbowen, line 3: If the user has explicitly chosen a locale it'll be stored in the cookie. I think it'd be better to use that value first and only fall back to the "accept-language" if needed.

api/src/templates/setup/setup.html
```diff
@@ -0,0 +1,185 @@
...
+    .action.hidden {
+      display: none;
+    }
+  </style>
+</head>
+<body>
+<div class="center">
+  <h1>{{ title }}</h1>
+  <ul id="actions">
+    <li class="action hidden" id="action_checks">
```
garethbowen, line 158: Let's loop over this instead of listing them all, eg: https://github.com/medic/cht-core/blob/master/api/src/templates/login/index.html#L26-L28

garethbowen (changes requested):
Two minor things.

Also I just thought, what happens with the service worker cache? Because the startup page is at the same URL as the login page will users who are offline cache the startup page replacing their cached login page? Another option would be to change the URL and not include it in the SW cache.

What happens for example if a user is actually logged in (has a userctx) and online but api is starting up? Would they be able to continue to fill in forms or would they just get stuck on the startup page? On the login page we check the userctx and redirect to the cached app, so we need to be sure we're not regressing here.

dianabarsan:
> Also I just thought, what happens with the service worker cache? Because the startup page is at the same URL as the login page will users who are offline cache the startup page replacing their cached login page? Another option would be to change the URL and not include it in the SW cache.

When API is not started up, the startup page shows up on every request that doesn't ask for json. At this time, users who have a service worker installed won't see it. The webapp will try to load as normal, and make a _session request, and since all requests ask for json, they'll get 502s. And they'll see a loading screen for ever. 

I understand the concern about the service worker. I'm not sure how to go about that. If I only display the startup page on a fixed URL, then I'll have to redirect users to that URL anyway, and the service worker will cache the redirect. 

> What happens for example if a user is actually logged in (has a userctx) and online but api is starting up? Would they be able to continue to fill in forms or would they just get stuck on the startup page? On the login page we check the userctx and redirect to the cached app, so we need to be sure we're not regressing here.

Offline users will work just the same, all the requests they make to the server will fail with 502s. 
Online users will see an infinite loading screen (this is a problem on webapp side which is not fixed in this PR).

api/src/templates/setup/setup.html
dianabarsan: Oh cool, I hadn't noticed that was possible.

api/src/middleware/locale.js
dianabarsan: Great suggestion, done!

dianabarsan:
> I understand the concern about the service worker. I'm not sure how to go about that.

What I could try is identify that the incoming request is a service worker cache request and never show the startup page when that happens. Every one of those requests has a `sw_precache` query parameter, for example.

garethbowen:
I think it's probably ok. The request for the service worker file will 502 which I'm assuming will mean the SW will not be registered or updated. It would be really unfortunate timing if the service worker request completed successfully and then before the browser could download all the resources the server restarted. Even if that did happen, the sw-precache library aborts if any of the resources return `!ok` so the status page will never end up in the cache.

> Online users will see an infinite loading screen (this is a problem on webapp side which is not fixed in this PR).

Which loading screen are you talking about? My read is, if they go to log in, they'll see the startup screen. If they're already logged in and request the webapp, the API will send the startup screen. If they're already on a page and go to a new page they'll make a bunch of JSON requests that will throw and they'll see an error on the screen. What am I missing?

garethbowen (approved):
Awesome work! Thanks for putting up with my questions.

I've replied about the SW issue, but I think we're good there. Ready for AT when you're happy with it.

dianabarsan:
> If they're already logged in and request the webapp,
that isn't what is happening right now. If you login as an online user, disconnect your internet and refresh the page, you'll see an infinite loading screen. You can test this on gamma.

garethbowen:
I guess it's fine as long as it's not a regression...

dianabarsan:
hey @garethbowen 
I pushed a few changes to fix the interaction between the admin app (during upgrade) and the startup router. I ended up adding a list of endpoints that are called during an upgrade and made special rules for them in the router, as they were not sending an `Accept` header (rather than editing every single place in the admin app or webapp where these requests are made). If you agree with this approach, we can think about adding more of these as required. 
Please let me know what you think. Thank you!

api/src/services/setup/router.js
```diff
@@ -0,0 +1,88 @@
...
+  '/_session',
+  '/api/v2/upgrade',
+  '/api/deploy-info',
+];
+paths.forEach(path => {
+  router.all(path, (req, res) => {
+    res.status(STATUS);
+    res.json({ error: 'Service unavailable' });
+  });
+});
```
garethbowen, line 74: I think I actually prefer the other approach of being explicit about requesting JSON. It seems more correct to call this out when making a scripting request. Other than "feeling" right there's not much of a difference so either is ok with me.

garethbowen (approved):
Nice one! Just one comment inline that you can take or leave...

api/src/services/setup/router.js
dianabarsan: that's what I did the first time around, and then started adding the paths to the router. 
A lot of the reason behind it was that, when I test, the admin app that that actually runs and needs the edit to is the "base" branch that I upgrade from. 

I pretty much agree that the request should explicitly ask for json.

dianabarsan: Changed!

dianabarsan:
I'm not sure what to do. Should I just merge this into archv3 and have it at-ed together?

garethbowen:
> Should I just merge this into archv3 and have it at-ed together?

Either is fine. I feel a little safer merging this now because with so much action going on in the deployment side it's good to see everything merged and integration tested together.

Issue #7601, opened 2022-04-27
Sync upgrade service api across repos


dianabarsan:
**Describe the issue**
Three services need to be in sync with the upgrade service API:
 - https://github.com/medic/upgrade-service-kubernetes
 - https://github.com/medic/cht-upgrade-service
 - https://github.com/medic/cht-core

The api was changed by `upgrade-service-kubernetes`, so the other two require a sync.

PR #7604, opened 2022-04-27
Upgrade endpoint payload change

dianabarsan:
# Description

Also set the upgrade as errored when the upgrade-service upgrade request fails. 
The alternative would have been to set an upgrade status that allows for a "complete" call retry (which is a bit more complicated). 

medic/cht-core#7601

# Code review checklist

# License

### DIFF
--- a/admin/src/js/controllers/upgrade-confirm.js
+++ b/admin/src/js/controllers/upgrade-confirm.js
@@ -26,7 +26,7 @@ angular.module('controllers').controller('UpgradeConfirmCtrl',
         .then(() => $uibModalInstance.close(true))
         .catch(err => {
           $log.error('Error when confirming', err);
-          const key = $scope.model.errorKey || 'instance.upgrade.error.confirm';
+          const key = $scope.model.errorKey || 'instance.upgrade.error.deploy';
           return $translate(key).then(msg => {
             $scope.status.error = msg;
             $scope.status.processing = false;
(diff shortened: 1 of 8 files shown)


admin/src/js/controllers/upgrade-confirm.js
```diff
@@ -26,7 +26,7 @@ angular.module('controllers').controller('UpgradeConfirmCtrl',
         .then(() => $uibModalInstance.close(true))
         .catch(err => {
           $log.error('Error when confirming', err);
-          const key = $scope.model.errorKey || 'instance.upgrade.error.confirm';
+          const key = $scope.model.errorKey || 'instance.upgrade.error.deploy';
```
dianabarsan, line 29: This error key didn't actually exist.

dianabarsan:
Hey @garethbowen . Please let me know what you think. Thanks for your time!

admin/src/js/controllers/upgrade.js
```diff
@@ -195,11 +195,12 @@ angular.module('controllers').controller('UpgradeCtrl',
         .post(url, { build })
         .catch(err => {
           // todo which status do we get with nginx???
```
garethbowen, line 197: Drop this todo now?

garethbowen (approved):
Nice.

dianabarsan:
I've pushed some changes: 
1. to change the payload to be snake case
2. to "finalize" the upgrade when the `upgrade-service/upgrade` call succeeds. (If the containers are up to date, docker doesn't actually restart the running container when doing an `up`).

garethbowen (approved):
That's awesome, thanks!

PR #7613, opened 2022-05-02
Add internal process type so validation passes

garethbowen:
# Description

This means validation passes - the type is ignored later in the process so it's not included in the release notes.

# Code review checklist

# License


mrjones-plip (approved):
lgtm 

minor point - should we call out in the technical issue template that any tickets won't be listed in release docs? Might sway the person opening the ticket to be more public if needed. approving PR as you'll know historically if this is relevant.

garethbowen:
Technical issues (updating dependencies, refactoring, etc) are still included in the release notes because they usually change some code, but hopefully in a way the user won't notice.

What's ignored are internal process and investigation issues because they won't change the actual code, as well as "won't fix" issues where work was done but nothing actually committed.

You could argue that these shouldn't be in the milestone at all but that's a separate discussion...

Issue #7503, opened 2022-02-07
User Creation: Allow user, contact & place in one step for multiple users


mrjones-plip:
**What feature do you want to improve?**
Using the [UI today](https://docs.communityhealthtoolkit.org/apps/tutorials/contact-and-users-1/#implementation-steps), to add a CHW, you first have to create their place and their contact in the CHT app, and then go over Admin app and create their user, selecting the contact and place you just created.  There is the additional overhead of delivering the credentials after the account has been created. 

**Describe the improvement you'd like**
It would be great if admins could have a way to do this in fewer steps to streamline the account creation. To make this really amazing, if the CHT could [automatically send a magic link](https://docs.communityhealthtoolkit.org/apps/concepts/access/#magic-links-for-logging-in) to the user.

Some features we should accommodate in the UX:
* error handling in tabular input (missing fields, unique phone numbers)
* allowing use to skip rows (fill out 2nd and 4th rows only) and still save data
* explanatory labels in admin form should be configurable, thus taken from translations

**Describe alternatives you've considered**
Use the existing manual process.

**Additional context**
A number of partners have expressed interested in having the ability of the respective Ministries of Health being able to do this with out the use of a command line in an efficient manner for a small number of users at a time. 

Per the continuous discover process: 
* [story map](https://docs.google.com/presentation/d/1XufQZEShpWXPel9azisVQlPPaoPpufpKijkRMhAw8K8/edit#slide=id.gfbfc44f19a_0_0) 
* [solution selection](https://docs.google.com/document/d/18MN-e2z-OwcHBE2EvfH5yIKHnVCL7hWvvncYtNeF81I/edit#heading=h.g1sek8njjyb7)
* [Interviews](https://drive.google.com/drive/folders/1x0ef7bXn2wytCCLFAmFw3tKovx2ZPoEi)

**Why Now?**
Partner(s): Guava Pineapple
Reason(s): A partner will be blocked by this for a major release or expansion

mrjones-plip:
Assigning to @n-orlowski to do UI/UX work on this and putting it into "Dev In progress".  After UI/UX work is done and approved, we'll move it over to @m5r  for coding.

mrjones-plip:
@abbyad  @michaelkohn @n-orlowski - added "Some features we should accommodate in the UX" in ticket body - feel free to add to this while we're UX prototype mode!

mrjones-plip:
[...]

PR #7614, opened 2022-05-02
7503 - Bulk user upload api

latin-panda:
# Description

**This is a work in progress!!!**

In this PR:
- Extends [/api/v1/users](https://docs.communityhealthtoolkit.org/apps/reference/api/#post-apiv1users) to accept csv and create users.
- **TODO...**

medic/cht-core#7503

# Code review checklist

# License


api/src/routing.js
```diff
@@ -56,6 +56,10 @@ const BUILDS_DB = 'https://staging.dev.medicmobile.org/_couch/builds/'; // jshin
 const cookie = require('./services/cookie');
 const app = express();
 
+// requires content-type application/x-www-form-urlencoded header
+const formParser = bodyParser.urlencoded({ limit: '32mb', extended: false });
+// requires content-type text/plain or application/xml header
+const textParser = bodyParser.text({ limit: '32mb', type: [ 'text/plain', 'application/xml', 'text/csv' ] });
```
latin-panda, line 62: I'm tempted to create a `csvParser`, unless we prefer to keep text types together for some reason.

mrjones-plip:
@latin-panda  - I'm working with @tatilepizs and we can't get this to work as of[5863cc3](https://github.com/medic/cht-core/pull/7614/commits/5863cc391ac723dafa8ed4f1602b59c38b2b9c04). Here's the steps we took to reproduce

1. check out `7503-bulk-user-creationg-api` branch and run `grunt` in main folder and `node server.js` once for `api` dir and once for `sentinel` dir
2. create a test CSV file called `7503.sample.csv` with this contents based on [your test lines](https://github.com/medic/cht-core/pull/7614/files#diff-29c8f8b69bb63bf0956dc153117fc26850a9c1b5118ae6cddd333d3ab6f763aeR2551):
   ```
   password,sername,type,place,contact.name,contact.phone,contact.address
   Secret1234,mary,person,498a394e-f98b-4e48-8c50-f12aeb018fcc,mary,2652527222,"1 King ST, Kent Town, 55555"
   Secret5678, peter ,person,498a394e-f98b-4e48-8c50-f12aeb018fcc,Peter, 2652279,"15 King ST, Kent Town, 55555 "
   ```
3. `POST` it with `curl` like so and get back an empty JSON response (`[]`):
   ```
   curl  -H "Content-Type: text/csv" --request POST -d @7503.sample.csv  http://medic:password@localhost:5988/api/v1/users
   []
   ```
   
Here's a video in case it's helpful:

https://user-images.githubusercontent.com/8253488/166338332-84b32f5b-ada4-41a8-836c-cc44f374d5b0.mp4

I'll try some JSON next instead of CSV maybe?

jkuester (approved):
ship

admin/src/js/controllers/users.js
```diff
@@ -52,6 +52,14 @@ angular.module('controllers').controller('UsersCtrl',
       });
     };
 
+    $scope.showAddMultipleUsersModal = function() {
+      Modal({
+        templateUrl: 'templates/multiple_user_modal.html',
+        controller: 'MultipleUserCtrl',
+        model: {},
+      });
+    };
```
craig-landry, line 61: Prefer the arrow function syntax when possible.

```suggestion
    $scope.showAddMultipleUsersModal = () => {
      Modal({
        templateUrl: 'templates/multiple_user_modal.html',
        controller: 'MultipleUserCtrl',
        model: {},
      });
    };
```

admin/src/js/services/create-user.js
```diff
@@ -1,38 +1,77 @@
...
+      /**
+         * Creates a user from a collection of updates
+         *
+         * Updates are in the style of the /api/v1/users/{username} service, see
+         * its documentation for more details.
+         *
+         * @param      {Object}  updates        Updates you wish to make
+         */
+      return function (updates) {
+        //const url = '/api/v1/users';
```
craig-landry, line 23: Please remove commented-out code.

```suggestion
```

admin/src/js/services/create-user.js
```diff
@@ -1,38 +1,77 @@
...
+
+      /**
+             * Creates a list of users from a formatted csv file
+             *
+             * data is the content of the csv file
+             *
+             * @param      {Object}  data        content of the csv file
+             */
+      return function (data) {
+        //const url = '/api/v1/users';
```
craig-landry, line 59: Please remove commented-out code.

```suggestion
```

admin/src/js/controllers/multiple-user.js
```diff
@@ -0,0 +1,151 @@
...
+      if (record.import) {
+        output += [
+          prepareStringForCSV(record.import.status),
+          prepareStringForCSV(record.import.message),
+          prepareStringForCSV(record.username),
+        ].join(delimiter) + eol;
+      }
+    });
+
+    const file = new Blob([output], { type: 'text/csv;charset=utf-8;' });
```
craig-landry, line 71: It is preferable to stream data in cases like this to avoid building it all up in memory. If it's data from the client though, the only option I know of is [Blob.stream](https://developer.mozilla.org/en-US/docs/Web/API/Blob/stream), but that needs newer browser support. If we can do that, it is preferred. If we cannot, the current in-memory implementation may still be fine.

admin/src/js/controllers/multiple-user.js
```diff
@@ -0,0 +1,151 @@
...
+  };
+
+  const prepareStringForCSV = (str) => {
+    if (!str) {
+      return str;
+    }
+    return '"' + str.replace(/"/g, '""') + '"';
+  };
+
+  const convertToCSV = doc => {
```
craig-landry, line 52: I'd like to stay out of the CSV business and instead use a library. Then we don't have to worry about things like quoting/escaping.

api/src/controllers/users.js
```diff
@@ -225,4 +225,34 @@ module.exports = {
...
+        let users;
+        let ignoredUsers;
+
+        if (typeof req.body === 'string') {
+          const parsedCsv = await usersService.parseCsv(req.body, logId);
+          users = parsedCsv.users;
+          ignoredUsers = parsedCsv.ignoredUsers;
+        } else {
+          users = req.body;
+        }
```
craig-landry, line 243: Please move this logic elsewhere. It'll leave you with something like this, which will make the larger function easier to reason about:

```
        const users = await loadUsers(req.body);
```

api/src/routing.js
```diff
@@ -78,17 +82,27 @@ const handleJsonRequest = (method, path, callback) => {
     }
   });
 };
+
+const handleJsonOrCsvRequest = (method, path, callback) => {
```
craig-landry, line 86: This branching seems like something the router could do for us, rather than us needed to have a route that does it. Can these (csv and json) be separated?

api/src/services/users.js
```diff
@@ -532,48 +551,128 @@ const validateUserContact = (data, user, userSettings) => {
...
+
+  const users = [];
+  const ignoredUsers = {};
+  for (let rowIdx = 0; rowIdx < allRows.length; rowIdx++) {
+    let ignoreUser = false;
+    let ignoreMessage = '';
+    const user = {};
+    try {
+      allRows[rowIdx]
+        .split(/,(?=(?:(?:[^"]*"){2})*[^"]*$)/)
```
craig-landry, line 623: At a minimum, this is probably a good case for a comment, to show what an expected string looks like. As with most comment scenarios though, some tests can tell the story better ;) 

A function that takes a single row and produces whatever split returns would help to contain the complexity with the regex here and then let you have a few test cases to make sure it stays working as expected over time.

craig-landry (commented):
I know this has been merged into the FR branch so we can have it functional for the demo. I was just reading along to understand the feature and have some comments from that. My comments are not intended to block anything now and are not meant to be comprehensive. They're just meant to be useful for when we go back to review this more fully.

Issue #7580, opened 2022-04-05
Flakey snackbar tests


garethbowen:
**Describe the issue**

The `snackbar.wdio-spec.js` tests fail quite frequently.

```
[Chrome Headless 100.0.4896.60 linux #0-28] Running: Chrome Headless (v100.0.4896.60) on linux
[Chrome Headless 100.0.4896.60 linux #0-28] Session ID: 403e3d1a-0e68-458e-b628-50702a27b7e0
[Chrome Headless 100.0.4896.60 linux #0-28]
[Chrome Headless 100.0.4896.60 linux #0-28] » /tests/e2e/snackbar.wdio-spec.js
[Chrome Headless 100.0.4896.60 linux #0-28] Snackbar
[Chrome Headless 100.0.4896.60 linux #0-28]    ✖ should display the snackbar with the syncing messages and then hide it
[Chrome Headless 100.0.4896.60 linux #0-28]    ✓ should display the snackbar with a clickable action
[Chrome Headless 100.0.4896.60 linux #0-28]
[Chrome Headless 100.0.4896.60 linux #0-28] 1 passing (23.4s)
[Chrome Headless 100.0.4896.60 linux #0-28] 1 failing
[Chrome Headless 100.0.4896.60 linux #0-28]
[Chrome Headless 100.0.4896.60 linux #0-28] 1) Snackbar should display the snackbar with the syncing messages and then hide it
[Chrome Headless 100.0.4896.60 linux #0-28] element ("#snackbar.active") still not displayed after 15000ms
[Chrome Headless 100.0.4896.60 linux #0-28] Error: element ("#snackbar.active") still not displayed after 15000ms
[Chrome Headless 100.0.4896.60 linux #0-28]     at /home/runner/work/cht-core/cht-core/node_modules/webdriverio/build/commands/browser/waitUntil.js:66:23
[Chrome Headless 100.0.4896.60 linux #0-28]     at async Element.wrapCommandFn (/home/runner/work/cht-core/cht-core/node_modules/@wdio/utils/build/shim.js:131:29)
[Chrome Headless 100.0.4896.60 linux #0-28]     at async Element.elementErrorHandlerCallbackFn (/home/runner/work/cht-core/cht-core/node_modules/webdriverio/build/middlewares.js:24:32)
[...]

PR #7615, opened 2022-05-02
Another attempt to make these tests reliable

garethbowen:
After many attempts to get these tests to pass reliably I believe it's now time to remove
them altogether. They can be reintroduced if and when we find a way to make them reliable.

medic/cht-core#7580

# Code review checklist

# License


tests/e2e/snackbar.wdio-spec.js
```diff
@@ -1,56 +0,0 @@
-const loginPage = require('../page-objects/login/login.wdio.page');
```
dianabarsan, line 1: I have to say this is extreme :))) 
Another option would be to disable the test with `xit`. I think that there are some floating that this test requires, but aren't used in other places, like all the `snackbar*` exports from the common elements page.

dianabarsan (approved):

tests/e2e/snackbar.wdio-spec.js
garethbowen: I thought about xit, but unless it's just a short term suspension I prefer it just living on in git history where it doesn't need to be maintained or refactored.

I've removed the orphaned code.

Issue #7617, opened 2022-05-03
Admin app is not localizable


michaelkohn:
**Describe the bug**
The Admin App is [supposed to be localizable ](https://forum.communityhealthtoolkit.org/t/admin-app-localization/1860/2?u=michael)but is [hard-coded to always use `en`](https://forum.communityhealthtoolkit.org/t/admin-app-localization/1860/3?u=michael). 

**To Reproduce**
1. Provide a French translation for the `instance.upgrade` key
3. From the login screen, select French as your language
4. Log in as a user who can access the admin app
5. Notice on the list of tabs on the left hand side that "Upgrade instance" is still in English 

**Expected behavior**
The admin app can be viewed in French

**Environment**
- Instance: https://demo-cht.dev.medicmobile.org/admin/#/upgrade
- Browser: Brave
- Client platform: MacOS
- App: Admin
- Version: 3.14.0

garethbowen:
I think what's missing is a call to the [SetLanguage service](https://github.com/medic/cht-core/blob/master/admin/src/js/services/language.js#L21). I think if this were called somewhere in the `main.js` it would just work.

garethbowen:
Ready for AT in `7617-translate-admin-app`.

AT steps in the issue description.

ngaruko:
Admin app translated
<img width="1041" alt="image" src="https://user-images.githubusercontent.com/6979995/169768037-b90ebea5-7155-4980-9519-f5da73941ad6.png">

<img width="1268" alt="image" src="https://user-images.githubusercontent.com/6979995/169768247-7602b610-3c62-4076-b2df-0b357626b366.png">

<img width="1280" alt="image" src="https://user-images.githubusercontent.com/6979995/169768389-c2924e6d-6b00-4306-aa60-f380cd908d46.png">

garethbowen:
Merged.

PR #7619, opened 2022-05-05
Use the locale cookie to set the language

garethbowen:
medic/cht-core#7617

# Code review checklist

# License

### DIFF
--- a/admin/src/js/controllers/main.js
+++ b/admin/src/js/controllers/main.js
@@ -3,11 +3,12 @@ angular.module('controllers').controller('MainCtrl',
     $log,
     $scope,
     $state,
-    $translate,
     $window,
     Auth,
+    Language,
     Location,
-    Session
+    Session,
+    SetLanguage
   ) {
     'ngInject';
 
@@ -23,7 +24,7 @@ angular.module('controllers').controller('MainCtrl',
         });
     };   
     
-    $translate.use('en');
+    Language().then(locale => SetLanguage(locale));
     $scope.authorized = false;
     $scope.navbarCollapsed = true;
     Auth.any([['can_configure'], ['can_view_outgoing_messages'], ['can_export_all']])
--- a/admin/src/js/services/language.js
+++ b/admin/src/js/services/language.js
@@ -24,16 +24,8 @@ const moment = require('moment');
       SetLanguageCookie
     ) {
       'ngInject';
-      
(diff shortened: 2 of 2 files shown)


garethbowen:
Review please @latin-panda 

We should fix this in 4.0.0 but it may also be useful to merge it in to your FR branch for your workshop next week.

mrjones-plip:
@dianabarsan  - @latin-panda requested [to be taken off this PR](https://medic.slack.com/archives/C02GME3BQE6/p1651780616318339?thread_ts=1651779163.329629&cid=C02GME3BQE6), so I removed her.  She's heads down on #7503 for the deadline at the end of this week. After that she's taking a vacation, so is doubly not good to be on this.

garethbowen:
@dianabarsan I choose you! No hurry - the FR work is done. This is just a small improvement to get in to 4.0.0 some time.

dianabarsan (approved):
Nice! My only concern is about not offering a way to change the display language in the admin app.

mrjones-plip:
I think we should block releasing this until we have a good default UX for this. As it stands, if you're in French and upgrade to 4.0, in which this feature is on with no way to turn it off, your admin UI changes in very unexpected ways:

image

We started to [track the translations needed to support French](https://docs.google.com/spreadsheets/d/1sWhyS54ILrEfjZimkmalq8TZZwrnkWG3awzJ6L1DI3o/edit#gid=0) (Medic Google Sheets link).

Deferring to @dianabarsan  and @garethbowen  if you agree though!

garethbowen:
@mrjones-plip I think the experience with incomplete translations for French is the same in the CHT webapp. For example, if a CHW picks French on the login page, and the project hasn't configured the missing translations then the experience is the same. This is by no means acceptable, but it's the current state of affairs - it is _possible_ to configure translations for any language you like, but only English is completely translated out of the box.

> this feature is on with no way to turn it off

The workaround is to disable the language so users can't pick it. The limitation to this approach is you cannot enable the language for the webapp and disable it for the admin app.

I think this change is still a net positive - with this projects can choose to either disable French, or complete the translation values themselves.

mrjones-plip:
Awesome - thanks for the feedback @garethbowen !

garethbowen:
> offering a way to change the display language in the admin app.

I was thinking about just doing this but it's a little fiddly and didn't want to block this improvement so I created a new issue for it: #7625

Issue #6598, opened 2020-08-18
In app "What's new" modal


n-orlowski:
We should let users know what has changed in their app since their last upgrade. This could include context on how to use a new feature, if UI elements have been updated, if there are new configuration updates or new workflows, etc. This would need to be configurable but we should provide a default for each release.

This ties into the remote onboarding work where:
a) A new user could see onboarding material
b) A returning user who has new training material would see the upgrade details
c) A returning user who has no new material will see no modal

MaxDiz:
Additional design work required so bumping to 3.13 based on roadmap meeting.

MaxDiz:
Adding @michaelkohn [brainstorm doc](https://docs.google.com/document/d/1gMkQQ4zbHBLWWw2NnHvITCkS-DVdoBs-sgCbdFtYAkQ/edit#) for reference

latin-panda:
This topic has been reevaluated and tested as part of the Care Team's continuous discovery work.

**The motivation**

It can be difficult for community health workers to understand UI/UX updates without direct training, but in-person training is costly, time-intensive, and now comes with additional COVID risk. CHWs should be able to have UI/UX changes delivered to them where they can continue to use the app without disruption and use new features effectively without the need for an in-person training event.

**The MVP to develop**
_See more details [here](https://docs.google.com/document/d/1pSDqgOqn_kGfLEd_Gi9rbctx_jfs0SPrvwHQX7xLHLU/edit#)_

- We (Medic) deal with getting that content into their app, we don’t need a configuration screen and/or cht-conf capability for the MVP, if it will slow us down
- We should have the ability to specify which users see cards and which do not, but that can be set manually for the MVP, if making an interface will slow us down
- It will be all or nothing…. any user that has access to see cards will see all cards, we do not need to separate content by any other attributes (for example, no need to differentiate content between supervisor, chw, nurse, etc….), if it will slow us down
- We do not need to account for cards in different languages right now, if it will slow us down
- If the app goes in the background or the user closes the app before reaching the end and tapping the “start (using app)” button, the next time they go into the app they should see cards again. We need to save where they left off (ie: it’s OK to always start at the beginning again).
- We do not need to allow users to access cards again after they have seen them, if it will slow us down
- We do need to track some [usage metrics / telemetry](https://docs.google.com/document/d/1WmWAK6i7W0ithN_WZYK-4ul3VpIHF9KJkkIjzlzozCY/edit#heading=h.4hwm8n9mib68)

[...]

PR #7628, opened 2022-05-23
6598 - Add training cards

latin-panda:
# Description

This PR:
- Extends Enketo service to work with modals and save completed trainings.
- Adds a new modal component, "training cards", to load Enketo training forms.
- Extends XML Forms service to subscribe to training forms and react when new trainings are added.
- Adds test coverage for all new code.

Ticket: https://github.com/medic/cht-core/issues/6598

# Code review checklist

# License


latin-panda:
Hi @garethbowen, I think it'd be good to have feedback at this point.
The only thing pending is telemetry (next thing I'll be working) and any feedback the team gives today's night.
Thanks before hand

garethbowen:
Sorry I'm not available to review this this week. I'll have a look next week if you haven't found another volunteer by then!

ddocs/medic-db/medic-client/views/trainings_by_contact/map.js
```diff
@@ -0,0 +1,5 @@
+function(doc) {
+  if (doc.type === 'data_training' && doc.form && doc.contact) {
+    emit([doc.contact._id], doc.form);
+  }
+}
```
garethbowen, line 5: 1. Do supervisors need to see this data? If not, it'd be better to put it in the user-meta db instead of medic so it doesn't impact other users.
2. For performance reasons it's better to avoid custom views if possible, and instead use CouchDBs built in doc ID view. In this case can we format the ID to make this view unnecessary, for example, if the doc ID was written like "training:<userid>:<formid>" then query using startswith and endswith to get completed trainings for this user?

ddocs/medic/views/docs_by_replication_key/map.js
```diff
@@ -48,6 +48,7 @@ function (doc) {
   };
   var value = { type: doc.type };
   switch (doc.type) {
+    case 'data_training':
```
garethbowen, line 51: Making any change to a view triggers a full rebuild of the view index which is expensive. Can we keep the `doc.type` as `data_record` or will that cause issues in other places?

webapp/src/ts/modals/training-cards/training-cards.component.html
```diff
@@ -0,0 +1,17 @@
...
+    <div>{{ errorTranslationKey | translate }}</div>
+  </div>
+
+  <div class="item-content" [hidden]="loadingContent || contentError">
+    <div class="body">
+      <mm-enketo [formId]="formWrapperId" [status]="enketoStatus" (onSubmit)="saveForm()" (onCancel)="cancel()"></mm-enketo>
+    </div>
+  </div>
+
+</mm-modal>
```
garethbowen, line 17: Does this have to be a modal? Consider having a modal which has a button that links to a page for training. That way enketo will just work and the user will have the full size to work in. Plus the page will have a URL so it can be refreshed, bookmarked, or sent to users via SMS.

webapp/src/ts/services/training-cards.service.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  public initTrainingCards() {
+    if (!this.canOpenTraining()) {
+      return;
+    }
+
+    this.xmlFormsService.subscribe(
+      'TrainingCards',
+      { trainingForms: true, contactForms: false },
+      async (error, xForms) => {
```
garethbowen, line 84: Pull this out as a separate named function rather than inlining it for readability.

webapp/src/ts/services/training-cards.service.ts
```diff
@@ -0,0 +1,114 @@
...
+    private userSettingsService: UserSettingsService,
+  ) {
+    this.globalActions = new GlobalActions(store);
+  }
+
+  private canOpenTraining() {
+    return !this.sessionService.isDbAdmin() && !document.getElementsByClassName('enketo').length;
+  }
+
+  private getActiveTrainingForms(xForms) {
```
garethbowen, line 33: I find "active" a bit confusing here. How about `getAvailableTrainingForms`?

garethbowen (changes requested):
Sorry for the delay here. I've given it a quick once over, but certainly not a full review. Happy to talk more about some of these concepts, particularly modal vs page, and db performance issues.

ddocs/medic/views/docs_by_replication_key/map.js
latin-panda: yes, I'm aware of the index rebuild :( 
I tried before but it'll mix with reports and sms reports, trainings are different type of thing.
If doing it, I would have to add conditionals across the code, and make sure the implementing project's config and integrations are okay with that and not mixing in the wrong places by mistake (lists, dashboards, exports, etc)
The `data_record` is too generic, being more specific adds clarity, also the code seems to use `data_record` as an indicator of "reports".
For this FR, making a new `doc.type` seemed less invasive and conceptually straight forward because we can address "trainings" in a contained/independent way. Is that okay? Any other idea?

webapp/src/ts/modals/training-cards/training-cards.component.html
latin-panda: Yes, Care Teams considered pages, but at the end it was designed with modals and tested with CHWs from i-tech and d3.

ddocs/medic-db/medic-client/views/trainings_by_contact/map.js
latin-panda: > Do supervisors need to see this data?
Someone from the partner will see it, but I'm not sure the role of this user, I'm going to ask for more details.

> For performance reasons it's better to avoid custom views if possible, and instead use CouchDBs built in doc ID view. In this case can we format the ID to make this view unnecessary, for example, if the doc ID was written like "training::" then query using startswith and endswith to get completed trainings for this user?

1. For my learning. What is a good scenario that justify a custom view?
2. To double check I understand. The ID will be something like this: "training:" + uuid  ---> "training:6b99de2a-dfc4-11ec-9d64-0242ac120002" Is that okay?

latin-panda: I remember reading somewhere long ago that using `startswith` and `endswith` to query docs is also expensive. But better than views, right?

webapp/src/ts/modals/training-cards/training-cards.component.html
latin-panda: A couple of reasons why using modals considering our users are:
- it helps them focus on a single self-contained task, we need them to complete the trainings for the success of upcoming new changes. They will have a cancel button in case they want to postpone the training, but they will not have other distractions available (tabs, hamburger menu) because the grey overlay. 
-  they remain in the same context they spect to be, and avoid getting them feeling lost because they got navigated to another page that will have different content every time. 

When using mobile, the space available to display form's elements in modal is very good, not too different when using page. But we can revisit this in case any element looks bad once we have the trainings designs.

I'm not sure if we want to bookmark trainings, specially if the training is retired because no longer relevant or it was completed before, or if we need them to focus in another training. Also trainings can have a strategic order, we can control with the start_date 

I'm tagging @n-orlowski and @michaelkohn in case we want to revisit some of these ideas or provide more info :)

ddocs/medic-db/medic-client/views/trainings_by_contact/map.js
garethbowen: CouchDB view performance is actually really solid, but it can be slow in creating, or reindexing views. That's why reusing existing views is so powerful - you get all the speed of the read, and none of the cost of the reindexing because it had to do that anyway.

> To double check I understand. The ID will be something like this: "training:" + uuid ---> "training:6b99de2a-dfc4-11ec-9d64-0242ac120002" Is that okay?

OMG! Markdown completely removed part of my comment, sorry! It's supposed to be `training:userid:uuid`. This way you can use startswith and endswith to filter down to just the users docs. Even better if the doc can be stored in the user meta db because then you don't need the `userid` bit.

> I remember reading somewhere long ago that using startswith and endswith to query docs is also expensive. But better than views, right?

startswith and endswith are params to the view query and perform well to get a subset of the view results.

ddocs/medic/views/docs_by_replication_key/map.js
garethbowen: I guess it depends how much these will be treated like reports, and how much they will be treated like a completely separate thing. It could be the same as reports for things like...
- marking a training Task as complete
- showing a supervisor Target of how many CHWs have completed the training
- showing trainings on the contact summary so the Supervisor can see which ones the CHW has done
- etc

If we ever do things like that it'll be much easier if it's just another report.

However you're right, if we want to _exclude_ the completed trainings from these features then we'll need to add a bunch of statements to filter them out.

My personal view is we should treat them as regular reports so app developers can build those features if they want to, while including exceptions to filter them out only where necessary.

Can you think of places we definitely don't want to return training reports?

webapp/src/ts/modals/training-cards/training-cards.component.html
garethbowen: All good points, and I realise this is late in the piece for a design change. I'm happy to follow up with a larger group if you think that'd help.

latin-panda: Looks like we're keeping modals, but please feel free to open a discussion in the care team channel! Thanks!

ddocs/medic-db/medic-client/views/trainings_by_contact/map.js
latin-panda: Looking across the code I noticed that we use more `userCtx.name`, using it can also improve readability, at the end it turned out: `training:userCtxName:uuid`. I hope this is okay :)

ddocs/medic/views/docs_by_replication_key/map.js
latin-panda: I met with the team today, we decided to just let training form behave as the system allows it (just like reports), we won't be filtering trainings from anywhere and they will be `data_record`. We'll check user feedback and reevaluate if we need to modify something in the training behaviour

ddocs/medic-db/medic-client/views/trainings_by_contact/map.js
garethbowen: Perfect! Good call.

webapp/src/ts/modals/training-cards/training-cards.component.ts
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  private async _loadForm() {
+    try {
+      this.loadingContent = true;
+      this.hideModalFooter = true;
+      this.geoHandle && this.geoHandle.cancel();
+      this.geoHandle = this.geolocationService.init();
+      const form = await this.xmlFormsService.get(this.trainingForm);
+      return this.ngZone.run(() => this.renderForm(form));
```
garethbowen, line 84: What is this returning? The `renderForm` function doesn't return anything, so presumably this doesn't either? I think you can just drop the `return` keyword.

webapp/src/ts/modals/training-cards/training-cards.component.ts
```diff
@@ -0,0 +1,217 @@
...
+    // see https://github.com/medic/cht-core/issues/2198#issuecomment-210202785 for AngularJS behavior
+    this.enketoService.unload(this.form);
+    this.globalActions.clearEnketoStatus();
+  }
+
+  private loadForm() {
+    return this.ngZone.runOutsideAngular(() => this._loadForm());
+  }
+
+  private async _loadForm() {
```
garethbowen, line 77: The `_` prefix was used in the old days when we didn't have `private` members. Now it's not needed. Rename this to `loadForm`

webapp/src/ts/modals/training-cards/training-cards.component.ts
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    this.recordTelemetryPreSave();
+    this.globalActions.setEnketoSavingStatus(true);
+    this.resetFormError();
+
+    try {
+      const docs = await this.enketoService.save(this.trainingForm, this.form, this.geoHandle);
+      console.debug('Saved form and associated docs', docs);
+      this.globalActions.setSnackbarContent(this.translateService.instant('training_cards.form.saved'));
```
garethbowen, line 162: Don't use `instant` if you can help it. It's better to use the async version and just show the snackbar once translation is returned. The reason is in case the translation service is not ready yet.

webapp/src/ts/modals/training-cards/training-cards.component.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  private recordTelemetryPostRender() {
+    this.telemetryData.postRender = Date.now();
+    this.telemetryData.action = 'add';
+    this.telemetryData.form = this.trainingForm.replace('training:', '');
+
+    this.telemetryService.record(
+      `enketo:training:${this.telemetryData.form}:${this.telemetryData.action}:render`,
+      this.telemetryData.postRender - this.telemetryData.preRender
+    );
```
garethbowen, line 184: I don't think this is a good idea. For one thing it'll make it hard to find the form because you will need to know it's a training form and add the prefix, but also it may clash with another form with the same name. The telemetry key is different but you're already adding `training:` back there so I think you can just remove it, eg:

```suggestion
    this.telemetryData.form = this.trainingForm;

    this.telemetryService.record(
      `enketo:${this.telemetryData.form}:${this.telemetryData.action}:render`,
      this.telemetryData.postRender - this.telemetryData.preRender
    );
```

This goes for the other telemetry items below too.

webapp/src/ts/services/enketo.service.ts
```diff
@@ -378,12 +387,24 @@ export class EnketoService {
...
   }
 
+  private isFormInModal($wrapper) {
+    if (!$wrapper || !$wrapper.length) {
+      return false;
+    }
+
+    return $wrapper.closest('.modal .modal-body').length > 0;
+  }
+
```
garethbowen, line 407: This seems fragile because it relies on the CSS classes in the DOM. If someone decided one day to change the names you wouldn't expect the training forms to suddenly stop working.. It would be better to pass in a boolean when rendering the form so internally this just checks the boolean.

webapp/src/ts/services/enketo.service.ts
```diff
@@ -686,16 +707,25 @@ export class EnketoService {
...
+          form: formInternalId,
+          type: 'data_record',
+          content_type: 'xml',
+          reported_date: Date.now(),
+          contact: this.extractLineageService.extract(contact),
+          from: contact && contact.phone
+        };
+
+        if (this.isTrainingForm(formInternalId)) {
+          doc._id = `${this.TRAINING_PREFIX}${this.sessionService.userCtx()?.name}:${uuid()}`;
```
garethbowen, line 723: It'd be easier to read if this were done with a join, eg:

```suggestion
          doc._id = [ this.TRAINING_PREFIX, this.sessionService.userCtx()?.name, uuid() ].join(':');
```

This would require also changing the TRAINING_PREFIX to not have the `:` included but I think that's an improvement.

I would probably also have a `const` for the username so it's not all in one line, eg:

```suggestion
          const username = this.sessionService.userCtx()?.name;
          doc._id = [ this.TRAINING_PREFIX, username, uuid() ].join(':');
```

webapp/src/ts/services/training-cards.service.ts
```diff
@@ -0,0 +1,124 @@
...
+import { ModalService } from '@mm-modals/mm-modal/mm-modal';
+import { SessionService } from '@mm-services/session.service';
+import { RouteSnapshotService } from '@mm-services/route-snapshot.service';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class TrainingCardsService {
+  private globalActions;
+  private readonly TRAINING_PREFIX = 'training:'
```
garethbowen, line 17: Because this is used in multiple classes it would be better to be exposed somewhere rather than redefined everywhere it's used. That will help find other uses of it if we ever want to refactor it.

webapp/src/ts/services/training-cards.service.ts
```diff
@@ -0,0 +1,124 @@
...
+    private sessionService: SessionService,
+    private routeSnapshotService: RouteSnapshotService,
+  ) {
+    this.globalActions = new GlobalActions(store);
+  }
+
+  private canOpenTraining() {
+    const routeSnapshot = this.routeSnapshotService.get();
+    return !this.sessionService.isDbAdmin() &&
+      (!document.getElementsByClassName('enketo').length ||
```
garethbowen, line 33: This is another situation where we really don't want to rely on CSS. Can't we ask the enketo service if it's currently got a form rendered?

webapp/src/ts/services/training-cards.service.ts
```diff
@@ -0,0 +1,124 @@
...
+        (routeSnapshot?.data?.tab === 'tasks' && !routeSnapshot?.params?.id));
```
garethbowen, line 34: What is this doing?

webapp/src/ts/services/training-cards.service.ts
```diff
@@ -0,0 +1,124 @@
...
+      .map(xForm => ({
+        id: xForm._id,
+        code: xForm.internalId,
+        startDate: xForm.start_date,
+        duration: xForm.duration,
+        userRoles: xForm.user_roles,
+      }))
+      .filter(form => {
+        const hasRole = form.userRoles?.find(role => userCtx.roles.includes(role));
+        if (form.userRoles?.length && !hasRole) {
```
garethbowen, line 50: You don't need to check the length - if it's undefined or empty then `hasRole` will be falsey, right?

```suggestion
        if (!hasRole) {
```

webapp/src/ts/services/training-cards.service.ts
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        const startDate = new Date(form.startDate);
+        if (!form.duration) {
+          return startDate < today;
+        }
+
+        const endDate = new Date(form.startDate);
+        endDate.setDate(endDate.getDate() + form.duration);
+        return startDate < today && endDate > today;
```
garethbowen, line 61: Let's turn this around and shortcircuit if the startDate > today...

```suggestion
        if (startDate > today) {
          return false; // training has not started yet
        }
        if (!form.duration) {
          return true; // training never ends
        }

        const endDate = new Date(form.startDate);
        endDate.setDate(endDate.getDate() + form.duration);
        return endDate > today;
```

This way you don't even look at the enddate unless you need to.

webapp/src/ts/services/training-cards.service.ts
```diff
@@ -0,0 +1,124 @@
...
+      if (completedTrainings) {
+        trainingForms = trainingForms.filter(form => !completedTrainings.has(form.code));
+      }
+
+      if (!trainingForms.length) {
+        return;
+      }
+
+      this.globalActions.setTrainingCard(trainingForms[0].code);
+      this.modalService.show(TrainingCardsComponent, { backdrop: 'static' });
```
garethbowen, line 105: If I'm understanding this correctly, then if a new card is found then it'll immediately pop up on top of everything, without warning. Is that right? Is that what we want?

garethbowen (changes requested):
Awesome!

I've put a few mostly minor suggestions inline.

This will also need a happy day end-to-end test in webdriverio so we have automatic regression testing.

webapp/src/ts/modals/training-cards/training-cards.component.ts
latin-panda: I asked the same when we did the angular migration, but at the end it turned out as a way to avoid name conflicts (still a thing in some languages/practices), between the function that contains ngZone ([here](https://github.com/medic/cht-core/pull/7628/files/4ac0f6a2a0f6224c57539aeeacf072f79481a41d#diff-94d07e089bc01aba387ccbb407e1b7ed08b021c77672da72e2511b1a92b4cef9R73)) and the function that runs outsize "angular ng zone". At the end we're doing this here to keep consistency with everywhere else in the code.

webapp/src/ts/modals/training-cards/training-cards.component.ts
latin-panda: I added the `enketo:training:` piece to keep consistency with what we do with [reports forms](https://github.com/medic/cht-core/blob/master/webapp/src/ts/modules/reports/reports-add.component.ts#L219) and [contact forms](https://github.com/medic/cht-core/blob/1f0f990150bf2e7df0242185f87365e96dbbecf9/webapp/src/ts/modules/contacts/contacts-report.component.ts#L154). 
Just to double check, do we want to keep consistency?

webapp/src/ts/services/training-cards.service.ts
latin-panda: currently we're not keeping an state of the form enketo is using. I can add something in the redux

webapp/src/ts/modals/training-cards/training-cards.component.ts
garethbowen: Keep the `taining:` bit in the ID, just don't remove it from the `form` property. My suggestion is consistent with those links you sent.

webapp/src/ts/services/training-cards.service.ts
garethbowen: My understanding of this code is it won't show up when the page loads, but will show up when the training form replicates down. In other words they will start the app, 10 seconds later the form will download, by then they're halfway through completing a task, and THEN we pop up and tell them to do the training.

Is that what the code is doing now? You can easily test this by logging in, uploading a training form, doing a sync, and see what happens.

I think a better user experience would be to show the training the next time the app is reloaded. For one thing at least they're not in the middle of a task and we interrupt them. But secondly, if there are app changes (like the icons or fabs) then the training will show _just_ before they see them which is perfect timing.

Basically I don't think you need to subscribe to `xmlFormsService` at all - just check for new forms at startup.

garethbowen:
I've come up with a couple of other questions overnight while thinking about this...

1. Have you tested what happens if the training form has a long page? Does the modal get a scrollbar, or does the modal overflow and the window gets a scrollbar?
2. Have you thought about purging at all? Will existing purging rules eventually purge the successful completion reports and prompt the user to train again?
3. If the user cancels the training will they be prompted to do it again next time the app loads?
4. If there are multiple trainings for the user to go through, will the next one show up when they finish the first one? Do they have to reload to get the next one?

latin-panda:
> Have you tested what happens if the training form has a long page? Does the modal get a scrollbar, or does the modal overflow and the window gets a scrollbar?

Yes, the modal has scroll, but the team will advice to keep pages short as part of the best practices when designing trainings.

> Have you thought about purging at all? Will existing purging rules eventually purge the successful completion reports and prompt the user to train again?

Yes, because we're letting training to behave as reports then the docs can be purged, the purge function will need to exclude trainings to prevent that. Also trainings can have a duration in days, so even if the completed training is deleted, it won't display to the user if the training has expired. 

> If the user cancels the training will they be prompted to do it again next time the app loads?

Yes, trainings will display when the app re/loads or if a new training is downloaded and it's set to display immediately. I haven't implemented an interval to show trainings again after X minutes, thinking to not disturb the user during other tasks, but this idea isn't completely discarded in the MVP yet, we'll observe user behaviours. 

> If there are multiple trainings for the user to go through, will the next one show up when they finish the first one? Do they have to reload to get the next one?

Trainings are display one at the time on re/load

Thanks for the questions, I'm going to link the team here in case there's a reconsideration or new ideas :)

webapp/src/ts/modals/training-cards/training-cards.component.ts
latin-panda: For now I've left the prefix to keep consistency in the code

webapp/src/ts/services/training-cards.service.ts
latin-panda: It's going to fix once we retrieve the enketo current form

webapp/src/ts/services/training-cards.service.ts
latin-panda: Once the training is replicated in the user's device, if they close the app and open again or just reload, the training is displayed. As well if it just replicated and it needs to show immediately because of how it was configured. 

>I think a better user experience would be to show the training the next time the app is reloaded. For one thing at least they're not in the middle of a task and we interrupt them. But secondly, if there are app changes (like the icons or fabs) then the training will show just before they see them which is perfect timing.

Unless the partner wants to train in workflow changes, that doesn't trigger the app update message for reloading, and let's say they want to train immediately as soon as the workflow changes are synced. Not sure if we want to cover this case now or just label it as edge case and just display it next time the app loads.
I'm going to link the team to this question for confirmation and get back with feedback.

webapp/src/ts/services/training-cards.service.ts
latin-panda: I tried a bunch of things (DOM query, CSS, state in reducers, etc) but the most stable and effective way so far has been to add a hideTraining property in the route’s data of pages where trainings shouldn’t display

webapp/src/ts/services/training-cards.service.ts
latin-panda: I took the topic to the team ([care team channel](https://medic.slack.com/archives/C02GME3BQE6/p1654668461417949)), the team believes that the scenario below: 

> Partner wants to train in workflow changes, that doesn't trigger the app update message for reloading. They want to train immediately as soon as the updated workflow is synced into the user's device

.. might happen more frequently than "training only in app load". The team would like to keep both and gather user feedback to further iterate based on user behaviour.

latin-panda:
I took the topic to the team ([care team channel](https://medic.slack.com/archives/C02GME3BQE6/p1654668461417949)) and these are the take aways for each topic: 

> Have you tested what happens if the training form has a long page? Does the modal get a scrollbar, or does the modal overflow and the window gets a scrollbar?

A/ Long content: We should always aim for concise content and for now we can control this. We are willing to iterate more to improve experience after the test proves this solution is effective.

> Have you thought about purging at all? Will existing purging rules eventually purge the successful completion reports and prompt the user to train again?

A/ Purging: Good one to consider later, but it shouldn’t block this test.

> If the user cancels the training will they be prompted to do it again next time the app loads?

A/ Canceling: If a user cancels the training it makes sense to show it to them again on reload (instead of after x minutes), we want to observe the user behaviour.

> If there are multiple trainings for the user to go through, will the next one show up when they finish the first one? Do they have to reload to get the next one?

A/ Multiple trainings: We control this for the test.

[Linking here](https://github.com/medic/cht-core/pull/7628#discussion_r896671198) the other topic about when to display trainings for traceability with team discussion.

webapp/src/ts/services/training-cards.service.ts
```diff
@@ -0,0 +1,129 @@
...
+      .map(xForm => ({
+        id: xForm._id,
+        code: xForm.internalId,
+        startDate: xForm.start_date,
+        duration: xForm.duration,
+        userRoles: xForm.user_roles,
+      }))
+      .filter(form => {
+        const hasRole = form.userRoles?.find(role => userCtx.roles.includes(role));
+        console.log('form.userRoles', form.userRoles, 'hasRole', hasRole);
```
garethbowen, line 44: Accidentally committed?

webapp/src/ts/services/training-cards.service.ts
```diff
@@ -0,0 +1,129 @@
...
+        if (!hasRole) {
+          return false;
+        }
+
+        const startDate = new Date(form.startDate);
+        console.log('startDate', startDate, 'today', today);
```
garethbowen, line 50: Ditto?

garethbowen (approved):
Awesome work! This is a really exciting feature.

Two suggestions inline, but approved pending those fixes.

Issue #7634, opened 2022-06-01
Allow login by SMS for bulk user import


m5r:
**What feature do you want to improve?**
Bulk user upload tracked by #7503

**Describe the improvement you'd like**
Allow admins to enable login by SMS for the users created in bulk

**Describe alternatives you've considered**
Fall back to manual user creation, creating them one by one with the login by SMS checkbox ticked

m5r:
Merged to `master` with bulk user upload in #7678

PR #7635, opened 2022-06-01
Support login by SMS in bulk user upload

m5r:
# Description

I've also made a few edits to the [spreadsheet](https://docs.google.com/spreadsheets/d/13f5Tj6to8FK2uy1bC9ibfAxzuYQHCtzG7FOxpr92NpE/edit?usp=sharing) that adds a `token_login` field, validates its value (`TRUE`/`FALSE` only) and make the password generation dependent on this field.

medic/cht-core#7634

# Code review checklist

# License

### DIFF
--- a/api/src/services/users.js
+++ b/api/src/services/users.js
@@ -663,7 +663,6 @@ const parseCsv = async (csv, logId) => {
 
   progress.status = BULK_UPLOAD_PROGRESS_STATUSES.PARSED;
   await bulkUploadLog.updateLog(logId, progress, logData);
-  console.log("users", users[0]);
   return { users, ignoredUsers };
 };
 
--- a/api/src/services/users.js
+++ b/api/src/services/users.js
@@ -569,7 +569,18 @@ const assignCsvCellValue = (data, attribute, value) => {
   if (!attribute.length || attribute.toLowerCase().indexOf(':excluded') > 0) {
     return;
   }
-  data[attribute] = typeof value === 'string' ? value.replace(/^"|"$/g, '').trim() : value;
+
+  if (typeof value !== 'string') {
+    data[attribute] = value;
+    return;
+  }
+
+  data[attribute] = value.replace(/^"|"$/g, '').trim();
+
+  if (['TRUE', 'FALSE'].includes(data[attribute])) {
+    // converts the "TRUE" or "FALSE" string to boolean
+    data[attribute] = eval(data[attribute].toLowerCase());
+  }
 };
 
 const parseCsvRow = (data, header, value, valueIdx) => {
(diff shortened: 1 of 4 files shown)


mrjones-plip:
While he's not formally available to start with Allies work just yet, @njogz  will be back soon, so removing @latin-panda and @elvisdorkenoo from the review in hopes of keeping them focused on Care Teams.

latin-panda:
I think making code reviews is fine and good practice anyways since we are all working in the same repository and maintaining the project regardless of our focus group. It also brings diversity of ideas and cross-learning.
I'm happy to take any code review as the team need it 🙂

api/src/services/users.js
```diff
@@ -569,7 +569,18 @@ const assignCsvCellValue = (data, attribute, value) => {
...
-  data[attribute] = typeof value === 'string' ? value.replace(/^"|"$/g, '').trim() : value;
+
+  if (typeof value !== 'string') {
+    data[attribute] = value;
+    return;
+  }
+
+  data[attribute] = value.replace(/^"|"$/g, '').trim();
+
+  if (['TRUE', 'FALSE'].includes(data[attribute])) {
```
latin-panda, line 580: Do you think we have to also consider the case of `TRUE` or `FALSE` coming as lower case?

latin-panda (commented):
Thank you for sending this!

api/src/services/users.js
m5r: I don't think so because spreadsheet editor automatically convert `true` to `TRUE` no matter the capitalization. The Google Sheets used to make the CSV also enforces this by narrowing down this cell value to `TRUE` or `FALSE` only

latin-panda: Nice thank you!

api/tests/mocha/services/users.spec.js
```diff
@@ -2604,6 +2604,55 @@ describe('Users service', () => {
...
+          contact: { name: 'mary', phone: '2652527222', address: '1 King ST, Kent Town, 55555' },
+          token_login: true,
+        },
+        {
+          password: 'Secret9876',
+          username: 'devi',
+          type: 'person',
+          place: '498a394e-f98b-4e48-8c50-f12aeb018fcc',
+          contact: { name: 'devi', phone: '265252', address: '12 King ST, Kent Town, 55555' },
+          token_login: 'truthy mistake',
```
latin-panda, line 2635: What do you think about whenever the attribute is `token_login`, then we prevent other values than boolean, so if a value is invalid then we just leave it undefined? 
`token_login` is kind of reserved, right?

m5r: This case could only happen if a user goes out of their way to edit the csv exported by the google sheets but I agree with you in that we should validate the input  
I'm not sure this kind of validation should happen in the csv parser though so I'm pushing some changes that will make any truthy value except `true` to be considered `false`. Let me know what you think about it

latin-panda: I think your idea makes it safer, thanks!

latin-panda (approved):
LGTM!

PR #7638, opened 2022-06-06
69 automate complete a delivery

lorerod:
# Description

Changes on e2e tests add the automated scenario of completing a pregnancy delivery form. 
/medic/cht-release-testing/issues/69

# Code review checklist

# License


tests/page-objects/contacts/contacts.wdio.page.js
```diff
@@ -52,6 +51,15 @@ const rightAddPerson = (create_key) => $(`span[test-id="rhs_add_contact"] p[test
...
 const childrenCards = () => $$('.right-pane .card.children');
+const newActionContactButton = () => $('.action-container .right-pane .actions .mm-icon .fa-stack');
+const forms = () => $$('/html/body/app-root/div/mm-actionbar/div/div/div/div[2]/div/span/ul/li');
+const formTitle = () => $('#form-title');
+const contactCardTitle = () =>
+  $('/html/body/app-root/div/div[2]/div/ng-component/div[2]/div/' +
+    'contacts-content/div/div/div/div[1]/div[2]/div/div[1]/h3');
+const contactInfoName = () =>
+  $('/html/body/app-root/div/div[2]/div/ng-component/div[2]/div/contacts-content/div/div/div/div[1]/div/div[1]/div');
+const contactInfoSex = () => $('//*[@id="contact_summary"]/div[3]/div/div/p');
```
craig-landry, line 62: Selectors like this that rely heavily on the structure of the DOM are fragile and prone to breaking with small changes on the page. 

Try to use ids and css classes to find elements by when you can, although be careful of being bound to stylistic attributes since those to may change easily. When no other stable option is available, it is preferred that we use a `test-id` attribute. This means you go into the code for the page and add your own element to that source code. For example, you might have html that looks like this:

```
<div>
  <span>Fred</span>
  <span>Rogers</span>
</div>
```
...the second span is showing the last name, but there's no clean way to select that "last name" field. You would do this:

```
<div>
  <span>Fred</span>
  <span test-id="last-name">Rogers</span>
</div>
```

Now you can make a selector like:
```
$('span[test-id="last-name"]');
```

This also means that if a developer comes along later and changes that html, they'll see the attribute and know it's used in a test so whatever change they make, they can retain the attribute and know the test will keep working as intended.

tests/page-objects/contacts/contacts.wdio.page.js
```diff
@@ -213,6 +221,47 @@ const editDistrict = async (districtName, editedName) => {
...
+      await (await formTitle()).waitForDisplayed();
+      return;
+    }
+  }
+  throw new Error(`Form with name: "${name}" not found`);
+};
+
+const openReport = async () => {
+  await (await rhsReportListElement()).waitForDisplayed();
+  await (await rhsReportListElement()).click();
```
craig-landry, line 247: This is already an async function, meaning it's going to return a promise automatically. If you `await` it here, that'll just wait on the promise to get a real value, and then this whole function will wrap that value in another promise for the return value. Better to just return the promise you have in the first place.

```suggestion
  return (await rhsReportListElement()).click();
```

Please make a similar change to all of the following functions that do similar.

tests/page-objects/reports/reports.wdio.page.js
```diff
@@ -42,6 +42,9 @@ const checkCss = 'input[type="checkbox"]';
 const sentTask = async () => (await reportBodyDetails()).$('ul .task-list .task-state .state');
 const reportByUUID = (uuid) => $(`li[data-record-id="${uuid}"]`);
 
+const patientName = () => $('.subject .name');
+const reportType = () => $('#reports-content > div.body.report-body > div.item-summary > div > div:nth-child(2)');
```
craig-landry, line 46: Similar to my note above about fragile css selectors, this one is kind of fragile. Also, for the parts like `.report-body` and `.item-summary' here, you don't need to require that the class is on any particular element (that .div prefix).

craig-landry (changes requested):

craig-landry (approved):

ngaruko (approved):

PR #7652, opened 2022-06-17
Add config linking to other reporting channels

garethbowen:
# Description

[description]

medic/cht-core#[number]

# Code review checklist

# License


garethbowen:
@craig-landry Can I grab a review? I'm not quite sure how this works, but I'm hoping it helps send support requests to the forum!

craig-landry:
I'm also not sure how it works. Ideally we'd require (or at least have it be a default) to ask for a forum post url before submitting such an issue. That's different though and doesn't need to block this, which seems like the recommended approach from GH for now.

craig-landry (approved):

Issue #7650, opened 2022-06-16
conflicts diff script crashes because of change in view code


dianabarsan:
**Describe the bug**
[The scripts/conflicts/diff](https://github.com/medic/cht-core/blob/41e6582b4c2ce9831428bcc30d05b422ad45cf1c/scripts/conflicts/diff.js) throws an error:
```
Generating diffs for conflicts on http://admin:pass@127.0.0.1:6500/medic
Found 1 conflicts
TypeError: Cannot read property 'length' of null
    at /home/diana/projects/medic/scripts/conflicts/diff.js:33:19
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

**To Reproduce**
Run the script against a 3.9+ CHT instance. 
 
**Expected behavior**
It should output something like: 
```
=====================
org.couchdb.user:aeroscope
=====================
 {
-  _rev: "8-de3fe258b9e30b7096b9125ce1f2276d"
+  _rev: "4-d50f73817a473255c40e6b06eebbe78b"
-  facility_id: "c5d73dde-c62b-4297-a78f-929e59438686"
+  facility_id: "dfe8e4d8-5338-4321-8751-e2ee64ec5bbf"
-  contact_id: "afa1cfabc855af4abd7a9721dd001624"
+  contact_id: "50472724-b83f-4ce8-b7f7-58f267a5442f"
 }
```

**Environment**
- Version: 3.9+

**Additional context**
This is caused by [this commit](https://github.com/medic/cht-core/commit/67bb273e2d20b775b27a6c0bdf20b169284aff1d) adding a reduce function to the conflicts view. 
Fixing this just requires passing `{ reduce: true }` to the query in the script: 
https://github.com/medic/cht-core/blob/41e6582b4c2ce9831428bcc30d05b422ad45cf1c/scripts/conflicts/diff.js#L25

dianabarsan:
This is ready to be AT ed on `PrjShrestha:Prajwol-7650-Diff-Script-crash-fix`

tatilepizs:
**Config:** Default
**Environment:** Local
**Platform:** WebApp
**Browser:** Chrome

---

### Reproducible on `Master`
After running the script `diff.js` this is the output.

<details>
<summary>Output</summary>

```
➜  conflicts git:(master) node diff.js
Generating diffs for conflicts on http://admin:password@localhost:5984/medic
Found 1 conflicts
TypeError: Cannot read property 'length' of null
[...]

PR #7668, opened 2022-07-05
Prajwol 7650 diff script crash fix and added code to compare and output each conflict and its diff to file.

PrjShrestha:
# Description

- Fixes diff script crash
- saves diff output to json files

medic/cht-core#7650

# Code review checklist

# License


scripts/conflicts/package-lock.json
```diff
@@ -1,8 +1,361 @@
 {
   "name": "conflicts",
   "version": "0.1.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
dianabarsan, line 4: I think you've inadvertently pushed changes to this file, because you'r using npm 7.x+ 
Can you please rollback these changes and use npm 6.x until we officially support newer versions?

scripts/conflicts/diff.js
```diff
@@ -10,8 +11,13 @@ const argv = parseArgs(process.argv);
 
 const server = argv._[2] || argv.server || process.env.COUCH_URL;
 
+// Example
+//server = 'https://user:pass@medic-bhaktapur.com/medic'
```
dianabarsan, line 15: Please change this example to show the way to actually pass the server url as an environment variable or argument, instead of instructing on how to hardcode it.

scripts/conflicts/diff.js
```diff
@@ -20,37 +26,90 @@ if (!server || argv.h || argv.help) {
 
 console.log(`Generating diffs for conflicts on ${server}`);
 
+
 const DB = new PouchDB(server);
 
-DB.query('medic-conflicts/conflicts')
+
+const writeToFile = function(path, content) {
+  const writeStream = fs.createWriteStream(path);
```
dianabarsan, line 34: You don't need to use `createWriteStream` here. It's simpler if you just use `writeFile` or `writeFileSync`: 
```suggestion
fs.writeFileSync(path, JSON.stringify(content, null, 2));
```

Another note: we use 2 spaces for indentation, so it's helpful to have the same format across your json as well.

scripts/conflicts/diff.js
```diff
@@ -20,37 +26,90 @@ if (!server || argv.h || argv.help) {
...
-              console.log('\n\n');
-            });
-        }
+        const docId = r.id;
+        const arrayRevsIdThatShouldNotBeThere = r.key;
+        return DB.get(docId, { open_revs: 'all' })
+          .then(results => {
+
+            // Get the main_doc. Maindoc is the doc that isn't in the conflict list
+            const main_doc = results.filter(r => !arrayRevsIdThatShouldNotBeThere.includes(r.ok._rev))[0].ok;
```
dianabarsan, line 60: Please always use camelCase for variables

dianabarsan:
```suggestion
const mainDoc = results.find(r => !conflictingRevs.includes(r.ok._rev)).ok;
```

scripts/conflicts/diff.js
```diff
@@ -20,37 +26,90 @@ if (!server || argv.h || argv.help) {
...
-
-              console.log('=====================');
-              console.log(r.id);
-              console.log('=====================');
-              console.log(jsondiff.diffString(mine, theirs));
-              console.log('\n\n');
-            });
-        }
+        const docId = r.id;
+        const arrayRevsIdThatShouldNotBeThere = r.key;
```
dianabarsan, line 55: `arrayRevsIdThatShouldNotBeThere` is pretty unclear to me. is `conflictingRevs` not a suitable name?

scripts/conflicts/diff.js
```diff
@@ -20,37 +26,90 @@ if (!server || argv.h || argv.help) {
...
+        return DB.get(docId, { open_revs: 'all' })
+          .then(results => {
+
+            // Get the main_doc. Maindoc is the doc that isn't in the conflict list
+            const main_doc = results.filter(r => !arrayRevsIdThatShouldNotBeThere.includes(r.ok._rev))[0].ok; 
+
+            //Now we are going to put each doc, its conflicts and thier diff
```
dianabarsan, line 62:
```suggestion
// Now we are going to write each doc, their conflicts and their diffs
```
Typos

scripts/conflicts/diff.js
```diff
@@ -20,37 +26,90 @@ if (!server || argv.h || argv.help) {
...
+
+            // Get the main_doc. Maindoc is the doc that isn't in the conflict list
+            const main_doc = results.filter(r => !arrayRevsIdThatShouldNotBeThere.includes(r.ok._rev))[0].ok; 
+
+            //Now we are going to put each doc, its conflicts and thier diff 
+            //in one file for manual inspection if required.
+            //The doc id might have characters that cannot be used to create folders.
+            let conflictDirectoryName = r.id;
+            if(r.id.includes('target')){
+              conflictDirectoryName = 'target_' + r.id.split(':')[0];
```
dianabarsan, line 67: Could you please explain the choice around assigning the directory name like this? 
The `id` string for a target is going to be something like: `target~2021-12~44219d91bffd45eca39942e3a1d288fb~org.couchdb.user:adventives`. This means that the folder name is going to be: `target_target~2021-12~44219d91bffd45eca39942e3a1d288fb~org.couchdb.user`. Is this what you were going for?

scripts/conflicts/diff.js
```diff
@@ -20,37 +26,90 @@ if (!server || argv.h || argv.help) {
...
 
+    const mainConflictFilesDirectory = './doc-conflicts';
     let ps = Promise.resolve();
+    if (!fs.existsSync(mainConflictFilesDirectory)){
+      try {
+        fs.mkdirSync(mainConflictFilesDirectory);
+      }catch(err){
+        console.log(`Error while creating main conflict folder: ${err}`);
+      }
+    }
```
dianabarsan, line 51: Please pull all of this in a separate function. 
Also, consider the scenario where the folder already exists and you have previous conflicts saved there. If some of those were already resolved and the user didn't clean them up, it might end up confusing. I would fail the script if the folder was not empty, just to make sure that the resulting files are always consistent and correct.

scripts/conflicts/diff.js
```diff
@@ -20,37 +26,90 @@ if (!server || argv.h || argv.help) {
...
+            //Now we are going to put each doc, its conflicts and thier diff 
+            //in one file for manual inspection if required.
+            //The doc id might have characters that cannot be used to create folders.
+            let conflictDirectoryName = r.id;
+            if(r.id.includes('target')){
+              conflictDirectoryName = 'target_' + r.id.split(':')[0];
+            } else if (r.id.includes('task')){ //If doc is task add ask and part of doc id
+              conflictDirectoryName = 'task_' + r.id.split(':')[1];
+            } else {  // if it is just a contact doc use the doct Id to create folder
+              conflictDirectoryName = r.id;
```
dianabarsan, line 71: you've already declared `conflictDirectoryName` as being `r.id`.

scripts/conflicts/diff.js
```diff
@@ -20,37 +26,90 @@ if (!server || argv.h || argv.help) {
...
+
+            //Now we are going to put each doc, its conflicts and thier diff 
+            //in one file for manual inspection if required.
+            //The doc id might have characters that cannot be used to create folders.
+            let conflictDirectoryName = r.id;
+            if(r.id.includes('target')){
+              conflictDirectoryName = 'target_' + r.id.split(':')[0];
+            } else if (r.id.includes('task')){ //If doc is task add ask and part of doc id
+              conflictDirectoryName = 'task_' + r.id.split(':')[1];
+            } else {  // if it is just a contact doc use the doct Id to create folder
```
dianabarsan, line 70: typo `doct`

scripts/conflicts/diff.js
```diff
@@ -20,37 +26,90 @@ if (!server || argv.h || argv.help) {
...
+            const main_doc = results.filter(r => !arrayRevsIdThatShouldNotBeThere.includes(r.ok._rev))[0].ok; 
+
+            //Now we are going to put each doc, its conflicts and thier diff 
+            //in one file for manual inspection if required.
+            //The doc id might have characters that cannot be used to create folders.
+            let conflictDirectoryName = r.id;
+            if(r.id.includes('target')){
+              conflictDirectoryName = 'target_' + r.id.split(':')[0];
+            } else if (r.id.includes('task')){ //If doc is task add ask and part of doc id
+              conflictDirectoryName = 'task_' + r.id.split(':')[1];
```
dianabarsan, line 69: As above, an example id for a task document is: `task~org.couchdb.user:adventives~4b710ca5-6c86-43a7-97ef-797ec5435dbc~pregnancy-danger-sign-follow-up~a random task4~1643343620261`. 
Is it super valuable to strip out `task~org.couchdb.user:` ?

scripts/conflicts/diff.js
```diff
@@ -20,37 +26,90 @@ if (!server || argv.h || argv.help) {
...
+        const arrayRevsIdThatShouldNotBeThere = r.key;
+        return DB.get(docId, { open_revs: 'all' })
+          .then(results => {
+
+            // Get the main_doc. Maindoc is the doc that isn't in the conflict list
+            const main_doc = results.filter(r => !arrayRevsIdThatShouldNotBeThere.includes(r.ok._rev))[0].ok; 
+
+            //Now we are going to put each doc, its conflicts and thier diff 
+            //in one file for manual inspection if required.
+            //The doc id might have characters that cannot be used to create folders.
```
dianabarsan, line 64: Which characters? 
If so, then it would be simpler to just replace those characters with acceptable ones, and apply the replacement for all ids, instead of having so many conditionals.

scripts/conflicts/diff.js
```diff
@@ -20,37 +26,90 @@ if (!server || argv.h || argv.help) {
...
+            
+            conflictDirectoryName.replace(/[^0-9a-z]/gi, '');
+            const conflictDirectoryPath = `${mainConflictFilesDirectory}/${conflictDirectoryName}`; 
+            if (!fs.existsSync(conflictDirectoryPath)){
+              try {
+                fs.mkdirSync(conflictDirectoryPath);
+              } catch(err){
+                console.log(err);
+              }
+            }
```
dianabarsan, line 82: Please reuse the function you've already added, as per a suggestion above, to create a folder with a given path.

scripts/conflicts/diff.js
```diff
@@ -20,37 +26,90 @@ if (!server || argv.h || argv.help) {
...
+
+            writeToFile(conflictDirectoryPath + '/' + 'mainDocument.json', main_doc);
```
dianabarsan, line 84: Please use `path.join` to generate paths instead of string concatenation: 
```suggestion
            writeToFile(path.join(conflictDirectoryPath, 'mainDocument.json'), main_doc);
```

scripts/conflicts/diff.js
```diff
@@ -20,37 +26,90 @@ if (!server || argv.h || argv.help) {
...
+            r.key.forEach( conflict_doc_id => {
+              const conflict_doc = results.find(r => r.ok._rev === conflict_doc_id).ok;
```
dianabarsan, line 86: Please always use camelCase for variables.

scripts/conflicts/diff.js
```diff
@@ -20,37 +26,90 @@ if (!server || argv.h || argv.help) {
...
+              console.log(stringJsonDiff);
+
+              //Write JSON diff between the conflict files
+              writeToFile(conflictDirectoryPath +'/diff_' +conflict_doc_id +'.json', jsonDiff);
+
+              //Write conflict doc to the drive
+              writeToFile(conflictDirectoryPath +'/'+ conflict_doc_id + '.json', conflict_doc);
+
+              //Write string diff between the conflict files
+              writeToFile(conflictDirectoryPath + '/string_diff_' + conflict_doc_id + '.txt', stringJsonDiff);
```
dianabarsan, line 100: The string diff looks super weird when written to the file: 
```
" {\n\u001b[31m-  _rev: \"8-de3fe258b9e30b7096b9125ce1f2276d\"\u001b[39m\n\u001b[32m+  _rev: \"4-d50f73817a473255c40e6b06eebbe78b\"\u001b[39m\n\u001b[31m-  facility_id: \"c5d73dde-c62b-4297-a78f-929e59438686\"\u001b[39m\n\u001b[32m+  facility_id: \"dfe8e4d8-5338-4321-8751-e2ee64ec5bbf\"\u001b[39m\n\u001b[31m-  contact_id: \"afa1cfabc855af4abd7a9721dd001624\"\u001b[39m\n\u001b[32m+  contact_id: \"50472724-b83f-4ce8-b7f7-58f267a5442f\"\u001b[39m\n }\n"
```

Is it an option to save the diff as json? Or is this text format more useful somehow?

dianabarsan (changes requested):
Nice work! 
I left a couple of suggestions inline.

scripts/conflicts/diff.js
PrjShrestha: Thank you for your feedback.

I have changed the replace regex to `conflictDirectoryName.replace(/[/\\?%*:|"<>]/g, '')` which should cover most if not all illegal characters.

Regarding multiple, if conditions.
The main goal was to make the folder name such that the folder name was part of the document ID in couch DB so that it is easily searchable if required.  I also wanted tasks , targets and contacts to be separated on the folders.

The patterns I saw in Target and task naming was that: 
Targets would require the **split to at the user colon** to have a unique searchable id like string.
`**target~2022-05~c7ddd59d-55f6-5c3c-832c-f13eb2869c83~org.couchdb.user**:bardiwas_chn4`

Whereas for **tasks we can't split at the colon** so only thing we can do is replace the colon with space.  
`task~org.couchdb.user**:**bardiwas_chn6~927796f2-d90b-44bc-a5aa-342326e5ddf7~pregnancy-home-visit-month-9~anc.pregnancy_home_visit.known_lmp~1652772300265`

Now the contact and targets docs are directly searchable, while for the task from we just need to replace the space after colon with space.

scripts/conflicts/diff.js
PrjShrestha: Replaced with conflictingRevs

scripts/conflicts/diff.js
dianabarsan: I think the splitting and ID manipulation could look pretty arbitrary to someone who has not seen the code, and just using the document ID as is would be best.

PrjShrestha: Document ID as is cannot be the folder name due to the ":" hence I had to split it.

scripts/conflicts/diff.js
PrjShrestha: Thanks. Changed.

scripts/conflicts/diff.js
PrjShrestha: I have put out both the string and JSON diff. The String diff was there from the beginning so I didn't change it.

PrjShrestha: Removed string diff. Json diff had more and clearer content. Thank you for suggestion.

scripts/conflicts/diff.js
PrjShrestha: Done.

scripts/conflicts/diff.js
PrjShrestha: Thank you for the suggestions. The program now exits if the directory is not empty. Also extracted the code to create directory name and directory to another separate function.

scripts/conflicts/diff.js
```diff
@@ -22,35 +28,94 @@ console.log(`Generating diffs for conflicts on ${server}`);
 
 const DB = new PouchDB(server);
 
-DB.query('medic-conflicts/conflicts')
+const writeToFile = function(path, content) {
+  content  = JSON.stringify(content, null, 3); 
+  fs.writeFile(path, content, (err) => {
```
dianabarsan, line 33: This is a callback function. Please either: 
- use `fs.promises.writeFile` and make the function async
- keep the callback but make everything callback based. 
- make everything synchronous `fs.writeFileSync`

```javascript
const writeToFile = async (path, content) => {
  try {
    const jsonContent = JSON.stringify(content, null, 3);
    await fs.promises.writeFile(path, content);
  } catch (err) {
    console.error('Error when writing file', err);
    throw err; 
  }
}
```

scripts/conflicts/diff.js
```diff
@@ -22,35 +28,94 @@ console.log(`Generating diffs for conflicts on ${server}`);
...
+    }
+  });
+};
+
+const createDirectory = function(conflictDirectoryPath) {
+  if (!fs.existsSync(conflictDirectoryPath)){
+    try {
+      fs.mkdirSync(conflictDirectoryPath);
+    } catch(err){
+      console.log(err);
```
dianabarsan, line 45: We should provide some context about where this error is being thrown. 
What would happen if you can;t create the folder (because of permissions for example)? Every next write is going to fail and the result will be a pagefull of hard to read errors. 
I think we should hard fail and stop execution if we can't create the folder or can't write a file. 

```suggestion
      console.error('Error when creating conflict folder', err);
      throw err;
```

scripts/conflicts/diff.js
```diff
@@ -22,35 +28,94 @@ console.log(`Generating diffs for conflicts on ${server}`);
...
+};
+
+const isDiectoryEmpty = function isEmpty(path) {
+  return fs.readdirSync(path).length === 0;
+};
+
+const getConflictDirectoryPath = function (r, mainConflictFilesDirectory) {
+  let conflictDirectoryName = r.id;
+  if (r.id.includes('target')) {
+    conflictDirectoryName = r.id.split(':')[0];
```
dianabarsan, line 57: Please add a comment that explains what is happening here. 
The comment should contain: 
1. An example document _id that matches `target`
2. The expected directory name based on that example

scripts/conflicts/diff.js
```diff
@@ -22,35 +28,94 @@ console.log(`Generating diffs for conflicts on ${server}`);
...
+  } else { // if it is just a contact doc use the doc Id to create folder
+    conflictDirectoryName = r.id;
```
dianabarsan, line 59: `conflictDirectoryName` is already `r.id`.

You should either: 
- drop the else 
- drop the initial declaration
- use a ternary

```javascript
const conflictDirectoryName = r.id.includes('target') ? r.id.split(':')[0] : r.id;
```

scripts/conflicts/diff.js
```diff
@@ -22,35 +28,94 @@ console.log(`Generating diffs for conflicts on ${server}`);
...
+  }
+  //The doc id that is used for foldername might have characters that cannot 
+  // be used to create folders. Hence we are replacing illegeal characters with space. 
+  // Reverse this process to search for doc in DB.
+  conflictDirectoryName = conflictDirectoryName.replace(/[/\\?%*:|"<>]/g, ' ');
+  const conflictDirectoryPath = `${mainConflictFilesDirectory}/${conflictDirectoryName}`;
```
dianabarsan, line 65: Please always use `path.resolve` or `path.join` instead of concatenating strings when creating paths. 
```suggestion
  const conflictDirectoryPath = path.join(mainConflictFilesDirectory, conflictDirectoryName);
```

scripts/conflicts/diff.js
```diff
@@ -22,35 +28,94 @@ console.log(`Generating diffs for conflicts on ${server}`);
...
+  return conflictDirectoryPath;
+};
+
+DB.query('medic-conflicts/conflicts', {reduce:false})
   .then(conflicts => {
     console.log(`Found ${conflicts.rows.length} conflicts`);
 
+    const mainConflictFilesDirectory = './doc-conflicts';
```
dianabarsan, line 73:
```suggestion
    const mainConflictFilesDirectory = path.join(__dirname, 'doc-conflicts');
```

scripts/conflicts/diff.js
```diff
@@ -22,35 +28,94 @@ console.log(`Generating diffs for conflicts on ${server}`);
...
+};
+
+DB.query('medic-conflicts/conflicts', {reduce:false})
   .then(conflicts => {
     console.log(`Found ${conflicts.rows.length} conflicts`);
 
+    const mainConflictFilesDirectory = './doc-conflicts';  
     let ps = Promise.resolve();
+    createDirectory(mainConflictFilesDirectory);
+    if(!isDiectoryEmpty(mainConflictFilesDirectory)){
```
dianabarsan, line 76: I think the empty check should be executed before we attempt to create the directory. 

How I see this is: 
- check if directory exists
- if exists, check if empty and check if writable
- if not empty, fail hard
- if can't write, fail hard
- if not exists, create
- if can't create, fail hard

An alternative is to:
- check if directory exists
- if exists, delete (without recursive)
- if delete fails, fail hard. Delete would fail because: directory not empty OR we don't have permissions
- create directory
- if create fails, fail hard

dianabarsan (commented):
Awesome work, a couple of nitpicks inline.

dianabarsan (commented):
Review submitted, adding this comment so I stop getting notifications. 

@PrjShrestha please re-request a review when ready. Thanks!

dianabarsan (commented):
Awesome! Thanks for all the changes. Just some very minor suggestions inline and apologies for sending you on a more work-involved path.

scripts/conflicts/diff.js
```diff
@@ -22,35 +28,96 @@ console.log(`Generating diffs for conflicts on ${server}`);
...
+    throw err; 
+  }
+};
+
+const createDirectoryIfValid = function(conflictDirectoryPath) {
+  if (!fs.existsSync(conflictDirectoryPath)){
+    try {
+      fs.mkdirSync(conflictDirectoryPath);
+    } catch(err){
+      console.error('Error when creating conflict folder', err);
```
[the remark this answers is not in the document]
dianabarsan: I think we should exit here as well. There's no point in continuing if we can't create the folder.
```suggestion
      console.error('Error when creating conflict folder', err);
      process.exit(1);
```

scripts/conflicts/diff.js
```diff
@@ -22,35 +28,91 @@ console.log(`Generating diffs for conflicts on ${server}`);
...
+    try {
+      fs.mkdirSync(conflictDirectoryPath);     
+      return;
+    } catch(err){
+      console.error('Error when creating conflict folder', err);
+    }   
+  }
+  if(!isDiectoryEmpty(conflictDirectoryPath)){
+    console.error('Conflict Directory is not empty, exiting');
+    process.exit();
```
dianabarsan, line 47:
```suggestion
    process.exit(1);
```

dianabarsan (approved):
Awesome work! Just a couple of nitpicks about exit codes inline.

PR #7669, opened 2022-07-06
Automated test guide

tatilepizs:
# Description

Document about a small guide that we can follow to create new automated test cases

# Code review checklist

# License

### DIFF
--- a/tests/AUTOMATE_TEST_GUIDE.md
+++ b/tests/AUTOMATE_TEST_GUIDE.md
@@ -1,7 +1,8 @@
 # Create new autoamted test case - Guide
 
-A structure that can be used to have an idea about the creation of a new automated test case.
+This style guide provides editorial guidelines for anyone creating new automated test cases for the CHT-Core.
 
+There are three files that are the base of every new automaated test case, the most importat one is the `<name>.wdio.spec.js` that it will contains the actual test that is going to be executed. It needs to have that name convention because the [config file](https://github.com/medic/cht-core/blob/master/tests/wdio.conf.js#L42-L44) has as an attribute to execute all the files that have `*.wdio-spec.js`.
 ## Base files to use
 
 * Page-object file. (`../tests/page-objects/forms/<name>.wdio.page.js`)
@@ -14,4 +15,7 @@ A structure that can be used to have an idea about the creation of a new automat
 
 * e2e testing file. (`../tests/e2e/forms/<name>.wdio.spec.js`)
 
-    This file should contain **only** the implementation of the test that is going to be executed. All the logging and contact creation should be only referenced in this file. For a better understanding follow the `pregnancy-visit.wdio-spec.js` file as an example.
+    This file should contain **only** the implementation of the test that is going to be executed. All the logging and contact creation should be only referenced in this file. For a better understanding follow the [`pregnancy-visit.wdio-spec.js`](https://github.com/medic/cht-core/blob/master/tests/e2e/forms/pregnancy-visit.wdio-spec.js) file as an example.
+
+### Note:
+We decided to separate every functionality in files/folders because we want to make sure that we can reuse as much code as we can. If something new is implemented and might be used for another test will be important to insolate it in a separate file so it can be used in future tests.
--- a/tests/AUTOMATE_TEST_GUIDE.md
+++ b/tests/AUTOMATE_TEST_GUIDE.md
@@ -2,7 +2,7 @@
 
 This style guide provides editorial guidelines for anyone creating new automated test cases for the CHT-Core.
 
-There are three files that are the base of every new automaated test case, the most importat one is the `<name>.wdio.spec.js` that it will contains the actual test that is going to be executed. It needs to have that name convention because the [config file](https://github.com/medic/cht-core/blob/master/tests/wdio.conf.js#L42-L44) has as an attribute to execute all the files that have `*.wdio-spec.js`.
+There are three files that are the base of every new automated test case, the most important one is the `<name>.wdio.spec.js`, which contains the actual test that will be executed. It needs to have that name convention because the [config file](https://github.com/medic/cht-core/blob/master/tests/wdio.conf.js#L42-L44) has as an attribute to execute all the files that have `*.wdio-spec.js`.
 ## Base files to use
 
 * Page-object file. (`../tests/page-objects/forms/<name>.wdio.page.js`)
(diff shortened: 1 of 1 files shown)


tests/AUTOMATE_TEST_GUIDE.md
```diff
@@ -0,0 +1,17 @@
...
+
+    This file should contain the `xml` access and details for the form that is going to be tested, and the methods to access every field that is available in the form.
+
+* Data file. (`../tests/page-objects/forms/data/user.po.data.js`)
+
+    This file contains all the information about a _"dummy"_ contact that can be used for testing. All the attributes from this contact can be updated as required in the execution code for the test accessing the object. If a specific contact is required for the test a new file can be created with that information.
+
+* e2e testing file. (`../tests/e2e/forms/<name>.wdio.spec.js`)
+
+    This file should contain **only** the implementation of the test that is going to be executed. All the logging and contact creation should be only referenced in this file. For a better understanding follow the `pregnancy-visit.wdio-spec.js` file as an example.
```
jonathanvq, line 17: Please add a [link](https://github.com/medic/cht-core/blob/master/tests/e2e/forms/pregnancy-visit.wdio-spec.js) to the 'pregnancy-visit.wdio-spec.js' file instead of the name of the file? That way, people can easily find the reference you suggest here and understand it quickly.

tests/AUTOMATE_TEST_GUIDE.md
```diff
@@ -0,0 +1,17 @@
+# Create new autoamted test case - Guide
+
+A structure that can be used to have an idea about the creation of a new automated test case.
```
jonathanvq, line 3: Although I understand what you are saying here, I think it could help if this introduction changes a bit. Perhaps, similar to what we have for the [documentation style guide](https://docs.communityhealthtoolkit.org/contribute/docs/style-guide/). So, I suggest something like:

"This style guide provides editorial guidelines for anyone creating new automated test cases for the CHT-Core."

That is just an example; I'm not suggesting that's what it should say exactly. But, the point is, when someone reads that, they can see it as "oh, this is what I need to do, this is my reference from now on".

tests/AUTOMATE_TEST_GUIDE.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Base files to use
```
jonathanvq, line 5: I like the concept of the base files to use, but I also think it could be beneficial if you explain why we prefer to separate that functionality into separate classes (and folders). Do you think it could help other engineers the separation of concerns you're suggesting to do here? So they can see any new functionality related with forms, should go into a separate file with X naming convention, in Y folder. Same for contacts, etc, and so at the end the *.wdio.spec.js file looks like this: (then you provide a code example in this guide, not just the link)

jonathanvq (changes requested):
Thanks, @tatilepizs, for creating this! I think it would help any engineer working on automated test cases. I have added some suggestions.

Thanks.

tests/AUTOMATE_TEST_GUIDE.md
tatilepizs: Thanks

tests/AUTOMATE_TEST_GUIDE.md
tatilepizs: Thanks! I added a note that I think explains a little better the purpose, let me know if you think that I should add something else.

jonathanvq (changes requested):
Added a couple of minor requests. Thanks.

tests/AUTOMATE_TEST_GUIDE.md
```diff
@@ -0,0 +1,21 @@
+# Create new autoamted test case - Guide
+
+This style guide provides editorial guidelines for anyone creating new automated test cases for the CHT-Core.
+
+There are three files that are the base of every new automaated test case, the most importat one is the `<name>.wdio.spec.js` that it will contains the actual test that is going to be executed. It needs to have that name convention because the [config file](https://github.com/medic/cht-core/blob/master/tests/wdio.conf.js#L42-L44) has as an attribute to execute all the files that have `*.wdio-spec.js`.
```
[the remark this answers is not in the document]
tatilepizs: Thank you for this Jonathan, I really appreciate that you have also taken the time on reviewing grammar.

jonathanvq (approved):
Thanks for all the work put into this. Looks good to me, and I'll approve it, but I won't merge it until we get @ngaruko's feedback as he may have other items to discuss.

ngaruko (changes requested):
Thanks @tatilepizs for this needed documentation. Now that we are on it, I wonder if it would be a great idea to also add some notes on the [factory model](https://github.com/medic/cht-core/tree/master/tests/factories) that helps generate hierarchies almost out of the box.

tests/AUTOMATE_TEST_GUIDE.md
```diff
@@ -0,0 +1,21 @@
...
+* Data file. (`../tests/page-objects/forms/data/user.po.data.js`)
+
+    This file contains all the information about a _"dummy"_ contact that can be used for testing. All the attributes from this contact can be updated as required in the execution code for the test accessing the object. If a specific contact is required for the test a new file can be created with that information.
+
+* e2e testing file. (`../tests/e2e/forms/<name>.wdio.spec.js`)
+
+    This file should contain **only** the implementation of the test that is going to be executed. All the logging and contact creation should be only referenced in this file. For a better understanding follow the [`pregnancy-visit.wdio-spec.js`](https://github.com/medic/cht-core/blob/master/tests/e2e/forms/pregnancy-visit.wdio-spec.js) file as an example.
+
+### Note:
+We decided to separate every functionality in files/folders because we want to make sure that we can reuse as much code as we can. If something new is implemented and might be used for another test will be important to insolate it in a separate file so it can be used in future tests.
```
ngaruko, line 21:
```suggestion
We decided to separate every functionality in files/folders because we want to make sure that we can reuse as much code as we can. If something new is implemented and might be used for another test will be important to isolate it in a separate file so it can be used in future tests.
```

ngaruko:
> Thanks @tatilepizs for this needed documentation. Now that we are on it, I wonder if it would be a great idea to also add some notes on the [factory model](https://github.com/medic/cht-core/tree/master/tests/factories) that helps generate hierarchies almost out of the box.

@tatilepizs - For this one, I think it is good enough to add a small paragraph to the last notes, along the lines of "For `brac` and `cht` configurations, there are [factory models](https://github.com/medic/cht-core/tree/master/tests/factories) that generate test data out of the box with the possibility to cusomize if needed".

tatilepizs:
Thank you @ngaruko for your suggestions

ngaruko (approved):
Looks good. Thanks again for documenting this.

jonathanvq:
@tatilepizs, I was about to merge this since it has been approved, but I see some build errors. Can you please check and fix it so I can merge it? I know you are not making source code changes here, which is weird, but I still think we should at least check what the problem is.

Issue #7611, opened 2022-05-02
Release 3.15.0-FR-bulk-user-upload


mrjones-plip:
When development is ready to begin on a [Feature Release](https://docs.communityhealthtoolkit.org/core/releases/feature_releases/#release-names), an engineer on the appropriate Care Team or Allies should be nominated as a Release Engineer. They will be responsible for making sure the following tasks are followed, though not necessarily doing the work themselves.

# Planning

- [x] Create a GH Milestone for the release.
- [x] Add all the issues to be worked on to the Milestone.
- [X] Have an actual named deployment and specific end user that will be testing this Feature Release. They need to test in production, on the latest version. No speculative Feature Releases.
- [X] Assign an engineer as Release Engineer for this release. - @latin-panda  \o/

# Development

- [x] Create a new release branch in `cht-core` from the most recent release and call it  `<major>.<minor>.<patch>-FR-<FEATURE-NAME>`. If latest is `3.15.0` and the feature is to "allow movies to be uploaded", call it `3.15.0-FR-movie-upload`. Done before the release so all PRs can be set to merge to this branch, and not to `master`.
- [x] Set the version number in `package.json` and `package-lock.json` and submit a PR. The easiest way to do this is to use `npm --no-git-tag-version version <feature-release>`.
- [x] Ensure QA is briefed and is partnering with the Trio to ensure early and often checks of the feature are on track to be of production quality from the start.

# Releasing

This is an iterative process and it's assumed there will be multiple numbered releases throughout development of the Feature Release.

- [x] Build a beta named `<major>.<minor>.<patch>-FR-<FEATURE-NAME>-1` by pushing a git tag and when CI completes successfully notify the QA team that it's ready for release testing. If an updated Feature Release is needed, increment the last `1` by calling it `<major>.<minor>.<patch>-FR-<FEATURE-NAME>-2` etc.

# Close-out

- [x] Validate with the actual end user that this Feature Release delivers a quantifiable improvement.  If yes, plan on adding the feature to the next minor release by creating a new ticket to merge the code to `master`.  If no, we leave the code dead in this branch, never to be merged to `master`, but still loved all the same.
- [x] Mark this issue "done" and close the Milestone.

m5r:
With `3.16.0` being released relatively soon, should we close this issue?

mrjones-plip:
Correct! There's just two items left on the check list above:

[...]

PR #7678, opened 2022-07-14
Bulk user upload

m5r:
# Description

medic/cht-core#7611  
Related docs PR: medic/cht-docs#727

This PR adds a button to the users admin page to create users in bulk from a CSV file.  
You can test it by:
 - making a copy of the [google sheets spreadsheet](https://docs.google.com/spreadsheets/d/1yUenFP-5deQ0I9c-OYDTpbKYrkl3juv9djXoLLPoQ7Y/copy)
 - filling it out according to the [current documentation](medic/cht-docs#727)
 - exporting it to CSV
 - uploading it in the admin app to create users

# Code review checklist

# License


tests/e2e/api/controllers/users.js
```diff
@@ -950,19 +950,24 @@ describe('Users API', () => {
...
-          const response = error.responseBody;
-          chai.expect(response.code).to.equal(400);
-          chai.expect(response.error).include('Missing fields password for user at index 1');
-          chai.expect(response.details).to.deep.equal({
-            failingIndexes: [
-              { fields: ['password'], index: 1 },
-            ],
-          });
-        }
+        const response = await utils.request({ path: '/api/v1/users', method: 'POST', body: users });
```
latin-panda, line 953:
```suggestion
        const response = await utils.request({ path: '/api/v2/users', method: 'POST', body: users });
```

I wonder if we should keep the test for v1 and then also test in v2

admin/src/js/controllers/multiple-user.js
```diff
@@ -0,0 +1,151 @@
+angular.module('controllers').controller('MultipleUserCtrl', function(
```
latin-panda, line 1: Please add unit test for the scope functions since those are public

latin-panda (changes requested):
Excited to see this moving forward!
I've left some comments inline

admin/src/js/controllers/multiple-user.js
njogz: Do you have any specific tests you think are necessary or simply making sure the scope functions are called and have the correct values is sufficient?

tests/e2e/api/controllers/users.js
njogz: Decided to keep it as it is.

admin/src/js/controllers/multiple-user.js
```diff
@@ -0,0 +1,151 @@
...
+
+  $scope.showDisplayUploadConfirm = function () {
+    $scope.clearScreen();
+    $scope.displayUploadConfirm = true;
+    $scope.$apply();
+  };
+
+  $scope.backToAppManagement = function () {
+    $window.location.href = '/admin/#/users';
+    $window.location.reload();
```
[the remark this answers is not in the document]
njogz: I couldn't think of any. We can circle back to it later if you are okay with that.

njogz:
@latin-panda I have addressed some of your comments and responded to the questions. Please have another look.

admin/src/css/theme.less
```diff
@@ -185,3 +185,33 @@ ul {
 .people-section {
   margin-bottom: 30px;
 }
+
+.modal {
```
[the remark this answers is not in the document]
njogz: Yes. I didn't notice any adverse effects on other modals.

admin/src/js/controllers/multiple-user.js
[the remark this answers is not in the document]
latin-panda: Okay, as long as the experience isn't too annoying for the user :)

admin/src/js/controllers/multiple-user.js
latin-panda: Yes, it's not that much actually:
- $scope.onCancel is cleaning, setting variables correctly and closing modal
- $scope.processUpload is setting the scope correctly based on a mock log, the outputFileUrl is correct and it's catching errors 
- $scope.backToAppManagement navigates and cleans correctly 
- $scope.showFinishSummary and  $scope.showDisplayUploadConfirm cleans and set just the necessary

ddocs/medic/_attachments/translations/messages-en.properties
```diff
@@ -1385,6 +1387,32 @@ user.place.help = All users must be associated with a place in the system to be
 user.username.help = This is what you will use to log in to the app.
 username.invalid = Invalid user name. Valid characters are lower case letters, numbers, underscore (_), and hyphen (-).
 username.taken = Username "{{username}}" already taken.
+users.manage.back = Back to user list
```
latin-panda, line 1390: - Please sort these keys alphabetically, we try to keep them that way, for example `users.manage.back` should go after `users.manage.add-single-user`
- Also please use underscore instead of dash to separate words. For example look at `instance.upgrade.no_new_releases`

This apply for the other lines below
Thanks!

ddocs/medic/_attachments/translations/messages-en.properties
```diff
@@ -1385,6 +1387,32 @@ user.place.help = All users must be associated with a place in the system to be
...
+users.import.unavailable.title = Currently unavailable
+users.import.unavailable.instructions = Someone else in the system is currently in the process of adding multiple users.
+users.import.unavailable.retry = Please try again later
+users.import.summary.added-users = users added
+users.import.summary.previously-added-users = previously imported
+users.import.summary.failed-users = contained errors
+users.import.summary.click = Click
+users.import.summary.here = here
+users.import.summary.download-status-file = to download your status file.
+users.import.summary.resubmit.instructions = To resubmit the users containing errors :
```
latin-panda, line 1411:
```suggestion
users.import.summary.resubmit.instructions = To resubmit the users containing errors:
```

I think it's like that

admin/src/js/services/create-user.js
```diff
@@ -1,38 +1,69 @@
...
-        method: 'POST',
-        url: url,
-        data: updates,
-        headers: {
-          'Content-Type': 'application/json',
-          'Accept': 'application/json'
+(function () {
+
+  'use strict';
+  const url = '/api/v2/users';
```
latin-panda, line 4:
```suggestion
  const URL = '/api/v2/users';
```

Since it's a global constant

admin/src/js/services/create-user.js
```diff
@@ -1,38 +1,69 @@
...
+      $http,
+      $log,
+      $q
+    ) {
+      'ngInject';
+
+      /**
+       * Creates a user from a collection of updates
+       *
+       * Updates are in the style of the /api/v1/users/{username} service, see
```
latin-panda, line 17:
```suggestion
       * Updates are in the style of the /api/v2/users/{username} service, see
```

admin/src/js/services/create-user.js
```diff
@@ -1,38 +1,69 @@
...
+            'Content-Type': 'application/json',
+            'Accept': 'application/json'
+          }
+        });
+      };
+
+      /**
+       * Creates a user from a collection of updates
+       * 
+       * data is the content of the csv file
```
latin-panda, line 43: You can delete this line, it's already explained in line 45 :)

latin-panda (changes requested):
I had a quick look at the new changes, looks good!

Additional to the open feedback from top, I added these minor ones :)

admin/src/js/services/language.js
```diff
@@ -24,16 +24,8 @@ const moment = require('moment');
       SetLanguageCookie
     ) {
       'ngInject';
-      
-      const setDatepickerLanguage = function(language) {
```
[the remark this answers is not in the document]
latin-panda: @njogz did you verify if datepickers are translated?

ddocs/medic/_attachments/translations/messages-en.properties
```diff
@@ -1385,6 +1387,32 @@ user.place.help = All users must be associated with a place in the system to be
...
+users.import.confirm_upload_text = Are you sure you want to upload
+users.import.instructions = To add multiple users, please upload the appropriate spreadsheet.
+users.import.processing.title = Processing users
+users.import.processing.instructions = This process can take a long time, if you're uploading a lot of users, please be patient.
+users.import.summary.added_users = users added
+users.import.summary.click = Click
+users.import.summary.download_status_file = to download your status file.
+users.import.summary.failed_users = contained errors
+users.import.summary.here = here
+users.import.summary.previously_added_users = previously imported
```
latin-panda, line 1402: This one after
`users.import.summary.resubmit.instructions`

ddocs/medic/_attachments/translations/messages-en.properties
```diff
@@ -1385,6 +1387,32 @@ user.place.help = All users must be associated with a place in the system to be
...
+users.import.summary.resubmit.instructions = To resubmit the users containing errors:
+users.import.summary.resubmit.step1 = 1. Open the downloaded status file
+users.import.summary.resubmit.step2 = 2. Cross-reference the errors against your formatted spreadsheet
+users.import.summary.resubmit.step3 = 3. Correct rows with errors and export to .csv
+users.import.summary.resubmit.step4 = 4. Reupload updated spreadsheet
+users.import.title = Import from file
+users.import.upload_button = Upload .csv
```
latin-panda, line 1409: This one after `users.import.unavailable.title`

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "medic",
-  "version": "3.15.0",
+  "version": "3.15.0-FR-bulk-user-upload",
```
latin-panda, line 3: Do we revert the FR version before merging to master? 
same with `package.json`

cc: @njogz @garethbowen

admin/src/js/services/language.js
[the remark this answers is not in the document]
njogz: @latin-panda I have not yet verified that. Where in the admin app would be the best place to test this?

admin/tests/unit/controllers/multiple-user.spec.js
```diff
@@ -0,0 +1,175 @@
...
+      };
+    });
+  });
+
+  afterEach(() => {
+    sinon.restore();
+  });
+
+  describe('$scope functions', () => {
+    it('$scope.processUpload sets othe $scope values correctly', async () => {
```
latin-panda, line 62:
```suggestion
    it('$scope.processUpload sets the $scope values correctly', async () => {
```

latin-panda (changes requested):
Nice! I just a couple of minor things and a question inline

ddocs/medic/_attachments/translations/messages-en.properties
njogz: On this I was sorting by the first letter after `summary` so it made sense to have it in this order

admin/src/js/services/language.js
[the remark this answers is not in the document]
latin-panda: I took a look an apparently `datepicker` function was only used in this place, so it was prob a old unused code.

package-lock.json
njogz: Good catch. It makes sense to do so so making the change

latin-panda (approved):
Good Job!! 
Btw, I remember @m5r was working on fixing a scenario:
`assign the newly created contact as primary contact of a place only if we're also creating the place`. 
If you haven't already, it's worth double checking with him if that should be in this PR :)

njogz:
Thanks @latin-panda. @m5r should the fix be in before I merge?

mrjones-plip:
With @m5r out today and wanting to keep momentum up,  let's merge this now to `master` .  Worst case the default contact will be in 4, but more likely we'll be able to add another quick feature to 3.16.  Either way, there's a lot of good effort that this will unlock on 3.16 so we can start cherry picking everything in.

m5r:
Thanks for bringing it up @latin-panda

@njogz @mrjones-plip here is the [draft PR](https://github.com/medic/cht-core/pull/7693)  
I agree with Ashley, let's merge this PR to master first and we'll merge the primary contact addition with a separate PR

PR #7684, opened 2022-07-20
Archv3 clustered

dianabarsan:
# Description

- adds haproxy config
- adds couchdb config
- builds haproxy and couchdb images at runtime and pushes them to AWS 
- updates the compose template to include 3x couchdb
- cherry picked cluster safe credentials
- cherry picked upgrade permission
- some e2e test stabilization (mostly retries and increased timeouts). 

# Code review checklist

# License


dianabarsan:
Hey @garethbowen 

This is part 1 of bringing archv3 closer to `master`. I pulled in the 2 changes that needed some adjustments: the `can_upgrade` permission and the cluster safe credentials work (I just cherry picked all of your commits from that branch) and added some retries in e2e tests and waits in e2e tests: 

1. when changing users passwords, it takes a while before all nodes accept the same cookie
2. I think the resource usage on the CI vm increased and it's making the browser slower. 

Could you please have a look? Thanks!!

api/src/controllers/settings.js
```diff
@@ -53,6 +53,8 @@ module.exports = {
       .then(() => {
         const replace = req.query && req.query.replace;
         const overwrite = req.query && req.query.overwrite;
+        // eslint-disable-next-line no-console
+        console.log(JSON.stringify(req.body, null, 2));
```
garethbowen, line 57: Mistake?

garethbowen (approved):
I've had a quick skim through and it looks fine. Was there something specific you wanted me to check?

api/src/controllers/settings.js
dianabarsan: oh yes! I'll remove, thanks for noticing!

dianabarsan:
Hey @garethbowen 

I pushed a couple of commits (the specific things to be looked at): 
1. to fix the logging bug
2. I re-merged the couchdb docker files because I realized I miss-merged before and found (and fixed) 2 small problems
 - failure to run apt update, details here: https://github.com/medic/cht-core/pull/7586#discussion_r926257077
 - not checking cluster node length and returning that cluster is ready before any nodes are added - I added an additional check to compare `all_nodes` length to the number of nodes in the cluster: https://github.com/medic/cht-core/blob/archv3-clustered-take-2/couchdb/set-up-cluster.sh#L8
3. I noticed API can still sometimes start too soon (before cluster is ready). I changed server-checks to execute 3 requests to _membership and compare results and fail if they are different. This is still not 100% because we don't have a healthcheck yet, and we can sometimes only hit one of the nodes with all 3 requests. So I've added the additional system databases check (because on single node we only create the system databases _after_ we set up the cluster: https://github.com/medic/cht-core/blob/archv3-clustered-take-2/couchdb/set-up-cluster.sh#L102). 

Thanks!

garethbowen (approved):
Awesome. Great work.

Issue #7690, opened 2022-07-25
Uploading a form through the app management GUI does not set the xml version


garethbowen:
**Describe the bug**
When uploading a form in the admin app the xml version should be set so that app builders can perform logic on a report based on which form created it.

**To Reproduce**
1. Go to App Management app
2. Select App Forms from the navigation menu
3. Upload a new form
4. Submit a report using the new form
5. Look up the report in CouchDB and notice it has no `form_version` property

**Expected behavior**
The report has a `form_version` property with the time that the form was uploaded and a hash of the form xml.

**Logs**
None.

**Screenshots**
None.

**Environment**
- Instance: localhost
- Browser: Firefox, Chrome
- Client platform: Linux
- App: admin
- Version: 3.15.0

**Additional context**
[Forum post](https://forum.communityhealthtoolkit.org/t/questions-about-form-versioning-introduced-in-cht-core-3-15-cht-conf-3-10/2032)

garethbowen:
Ready for AT in `7690-admin-app-xml-version`. Reproduction steps above. You may like to test #7689 at the same time...

ngaruko:
Looks good> Form uploaded using the app management GUI has `Xmlversion`
**_Master_**
![image](https://user-images.githubusercontent.com/6979995/182098617-6d434fd2-e41f-44a8-8a5e-6dd1837fd2cf.png)

_**Branch**_
![image](https://user-images.githubusercontent.com/6979995/182099138-d99f72ac-45d5-45eb-a885-995027e799a4.png)

garethbowen:
Merged.

PR #7691, opened 2022-07-25
Add xmlVersion when uploading to admin app

garethbowen:
# Description

medic/cht-core#7690

# Code review checklist

# License

### DIFF
--- a/admin/src/js/controllers/forms-xml.js
+++ b/admin/src/js/controllers/forms-xml.js
@@ -86,6 +86,51 @@ angular.module('controllers').controller('FormsXmlCtrl',
       return formId;
     };
 
+    const getXmlHash = (xml) => {
+      const utf8 = new TextEncoder().encode(xml);
+      return crypto.subtle.digest('SHA-256', utf8).then((hashBuffer) => {
+        const hashArray = Array.from(new Uint8Array(hashBuffer));
+        return hashArray
+          .map((bytes) => bytes.toString(16).padStart(2, '0'))
+          .join('');
+      });
+
+    };
+
+    const addXmlVersion = (doc, xml) => {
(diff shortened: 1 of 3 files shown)


garethbowen:
@m5r Can I get a review please?

admin/src/js/controllers/forms-xml.js
```diff
@@ -116,31 +161,8 @@ angular.module('controllers').controller('FormsXmlCtrl',
...
-                .then(doc => {
-                  doc.title = title;
-                  Object.assign(doc, meta);
-                  doc.type = 'form';
-                  doc.internalId = formId;
-                  AddAttachment(doc, 'xml', xml, 'application/xml');
-                  return doc;
-                });
-            });
+            .then(() => createDoc(xml, meta));
```
m5r, line 165: Splitting this chunk into 3 smaller functions makes it easier to read :+1:

admin/tests/unit/controllers/forms-xml.js
```diff
@@ -1,12 +1,15 @@
 describe('FormsXmlCtrl controller', () => {
   'use strict';
 
+  const NOW = moment('2000-01-15').valueOf();
```
m5r, line 4: This constant is used only by one test, please define it in the nearest scope instead  
I think the best place for it would be in `describe('saves the doc', () => { ... })`

m5r (approved):
Looks good to me. Only one small change required but I'll approve so you can merge immediately after the change is made

admin/tests/unit/controllers/forms-xml.js
garethbowen: Good call! Done.

PR #7693, opened 2022-07-25
Set contact as primary contact of a place with bulk user upload

m5r:
# Description

Assign the newly created contact as primary contact of a place only if we're also creating the place

You can test this PR by:
- creating a user, its contact and its place all at once with the bulk user upload feature ([docs PR](https://github.com/medic/cht-docs/pull/727))
- making sure the newly created place has the newly created user's contact as primary contact

# Code review checklist

# License

### DIFF
--- a/api/src/services/users.js
+++ b/api/src/services/users.js
@@ -551,8 +551,25 @@ const validateUserContact = (data, user, userSettings) => {
   }
 };
 
+/* eslint-disable max-len */
+/**
+ * @param {Object} data
+ * @param {string} data.username Identifier used for authentication
+ * @param {string[]} data.roles
+ * @param {(Object|string)=} data.place Place identifier string (UUID) or object this user resides in. Required if the roles contain an offline role.
+ * @param {(Object|string)=} data.contact A person identifier string (UUID) or object based on the form configured in the app. Required if the roles contain an offline role.
+ * @param {string=} data.password Password string used for authentication. Only allowed to be set, not retrieved. Required if token_login is not enabled for the user.
+ * @param {string=} data.phone Valid phone number. Required if token_login is enabled for the user.
+ * @param {Boolean=} data.token_login A boolean representing whether or not the Login by SMS should be enabled for this user.
+ * @param {string=} data.fullname Full name
+ * @param {string=} data.email Email address
+ * @param {Boolean=} data.known Boolean to define if the user has logged in before. Used mainly to determine whether or not to start a tour on first login.
+ * @param {string=} data.type Deprecated. Used to infer user's roles
+ * @param {string} appUrl request protocol://hostname
+ */
+/* eslint-enable max-len */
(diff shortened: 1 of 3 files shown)


m5r:
@latin-panda could you review this please?

api/src/services/users.js
```diff
@@ -551,16 +551,21 @@ const validateUserContact = (data, user, userSettings) => {
   }
 };
 
-const createUserEntities = async (user, appUrl, preservePrimaryContact) => {
+const createUserEntities = async (data, appUrl) => {
```
latin-panda, line 554: It'd be helpful to add a comment above the function describing what is `data` parameter. Sometimes it's hard to know what to expect without debugging 

```
   * @param {Object} data User information containing: username .........
```

api/src/services/users.js
```diff
@@ -551,16 +551,21 @@ const validateUserContact = (data, user, userSettings) => {
...
+  // preserve the place's primary contact only if it's an existing one
```
latin-panda, line 555:
```suggestion
  // Preserve the place's primary contact, if data.place already exists in the DB. 
``` 
BTW. What happens if the people forget to add the place id? How often would an error like this happen?

api/src/services/users.js
```diff
@@ -551,16 +551,21 @@ const validateUserContact = (data, user, userSettings) => {
...
+  // => if we're creating the place alongside the user, set the contact as the place's primary contact
+  const isCreatingNewPlace = !_.isUndefined(data.place) && !_.isNull(data.place) && _.isUndefined(data.place._id);
```
latin-panda, line 557: We're not using this variable much right? And since we already have a comment explaining the business logic.
What about if we just keep the `preservePrimaryContact`? 

```
preservePrimaryContact = !!(data && data.place && data.place._id)
```

latin-panda (changes requested):
Nice job!! I left some question below :)

api/src/services/users.js
m5r: If the place id is not set it will create a new place and move the user's contact there which is the expected behavior, this is tested by unit tests in `api/tests/mocha/services/users.spec.js:1381`

This mistake can only happen when hitting the API directly with a script. The only place in the code base that calls this endpoint is in the admin app when creating or editing a user and the place property is always a string representing the place's id.

But re-reading some of the places-related code in the api, I will change the condition to check on the `_rev` property instead of `_id`, this is how `getOrCreatePlace()` decides if it should create a place or not

api/src/services/users.js
m5r: It's the same `data` that's documented by the JSDoc for `createUser` and `createUsers` but I'll add it `createUserEntities` to make it clearer

m5r:
Thanks @latin-panda, could you give it another look please?

latin-panda (approved):
Thanks! 🤩 Looks good

Issue #6382, opened 2020-04-29
Use Noto font on the login page


garethbowen:
**What feature do you want to improve?**
Currently we use `"Helvetica Neue",Helvetica,Arial,sans-serif` as the font on the login page. This is inconsistent with the Noto San used in the add, and in some languages and devices could end up being unable to render the characters.

**Describe the improvement you'd like**
Switch to Noto Sans. It is already cached in the service worker so should just work with no additional overheads. eg:

```css
@font-face {
  font-family: 'Noto';
  src: url(/fonts/NotoSans-Regular.ttf) format("truetype");
}
body, input {
  font-family: Noto, sans-serif;
}
```

**Describe alternatives you've considered**
Leaving it as is - it works ok and I'm probably the only one who's noticed...

zannain:
I can update the font here. First time contributing-does the implementation you added above still hold up?

garethbowen:
@zannain Welcome! Yes I think that should work. You'll need to test it to make sure.

garethbowen:
Since we added translation links on the login page I've noticed we show tofu on the login page for target languages so this is a more serious issue than just inconsistent fonts.

![Screenshot from 2021-09-29 15-14-52](https://user-images.githubusercontent.com/2064938/135191645-b8f79521-74fc-481d-ad77-612f29e9af3f.png)

ngaruko:
Looks good:
On master 
![image](https://user-images.githubusercontent.com/6979995/180954259-39d068a4-cb95-4167-9286-8e5ae37786ed.png)

on branch 
![image](https://user-images.githubusercontent.com/6979995/180954397-71a28cdf-4ca3-4171-9493-c6813da90c50.png)

elvisdorkenoo:
Feature has been merged.

PR #7695, opened 2022-07-25
change the login font family to Noto sans serif

elvisdorkenoo:
# Description

Switch the font family of the login page from "Helvetica Neue",Helvetica,Arial,sans-serif" to "Noto, sans-serif".

medic/cht-core#6382

# To test this PR
- [ ] Go to the login page
- [ ] Make sure the font-family "Noto, sans serif" is correctly loaded and is the font used on the page.

# Code review checklist

# License

### DIFF
--- a/api/src/public/login/style.css
+++ b/api/src/public/login/style.css
@@ -1,3 +1,11 @@
+@font-face {
+  font-family: 'Noto';
+  src: url(/fonts/NotoSans-Regular.ttf) format("truetype");
+  font-weight: normal;
+}
+body, input {
+  font-family: Noto, sans-serif;
+}
 body {
   background-color: #323232;
   color: #FFFFFF;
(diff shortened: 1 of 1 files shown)


api/src/public/login/style.css
```diff
@@ -7,7 +7,7 @@ body {
   margin: 0;
 }
 body, input {
-  font-family: "Helvetica Neue",Helvetica,Arial,sans-serif;
+  font-family: Noto, sans-serif;
```
garethbowen, line 10: This won't work if the device doesn't already have Noto installed. For everyone else you'll need to link to the Noto source, eg: https://github.com/medic/cht-core/blob/master/webapp/src/css/font.less#L3-L7

elvisdorkenoo: ok @garethbowen, I linked it first, then I tried to remove it and it worked too, so I thought it not to be needed, now I get it, thank you.

garethbowen (approved):
LGTM!

Issue #7653, opened 2022-06-20
Report's filter redesign 


latin-panda:
**Description**

[Taken from the forum:](https://forum.communityhealthtoolkit.org/t/feature-testing-filters/1917)
As we are working on incremental updates to the UI/UX of the CHT to be more scalable, intuitive, mobile-first and consistent with other Android experiences, we are looking to test a new design of filters.

![image](https://user-images.githubusercontent.com/66472237/174545990-39d95647-5aa9-46fe-9357-fee14277602f.png)

The current design allows users to select the type of filter from a row of drop downs with icons. Each filter type expands and users are able to filter content by selecting the appropriate checkboxes. The search icon expands to show a search bar.

![image](https://user-images.githubusercontent.com/66472237/174546374-e0677694-fdf2-4ce7-9c71-26a4c82ebc6b.png)

The proposed design decouples the search and filter functionalities. In this proposal, the user is able to view filters by tapping on the filter icon. This opens a drawer with a list of filter types. By tapping on the dropdown arrows, the user is able to view detailed filter options and tap on the appropriate checkboxes.

Our intention is to simplify the filtering experience by providing a cleaner interface and clearer labels that allows for better scalability.

**Additional context**
This work relates with this item from the [Road Map](https://github.com/medic/cht-roadmap/issues/61)
[Figma](https://www.figma.com/file/3wFzMcNEGxV6se1QWftckk/Filter?node-id=0%3A1)

latin-panda:
[Taken from forum](https://forum.communityhealthtoolkit.org/t/feature-testing-filters/1917/2)
Open questions from @michaelkohn  are:

**Active Filter**
It’s nice to know if a filter is currently applied (5th screenshot on the new design). There is a visual cue for this in the current app (though you can’t see it in mobile view). Is there a design pattern for doing that with the new filter icon (ie the funnel)?
![image](https://user-images.githubusercontent.com/66472237/174546846-3a89e720-3d99-43ff-8f8d-7320b8ec94a8.png)

**Place filter**
“Place” filters are tricky because there is a hierarchy element. It’s currently conveyed with indentation and can be confusing if you have many levels in your hierarch. I’m curious how it might look in the new design?
![image](https://user-images.githubusercontent.com/66472237/174546937-1bf532a5-8dd8-47ed-9033-24403125b1b8.png)

cc: @n-orlowski

latin-panda:
@n-orlowski how does it look in desktop?

[...]

Issue #1000, opened 2015-06-12
Bulk delete reports


abbyad:
During testing and training, numerous messages and reports are submitted that need to be deleted once the instance goes to production, especially since it affects usability and analytics.

Currently the only way for a project manager to delete messages/reports is to go through them individually, which is extremely time consuming. We should look at options for bulk delete first, and perhaps consider a "test mode" for the webapp.

Note that some CHW training may overlap with production use in other districts, so it is not sufficient to delete all messages/reports.

mandric:
We might also consider bulk updates, in the past you could delete/validate/invalidate/modify facility for a selected group of records.

abbyad:
This feature request continues to be requested by many users, including our own PMs and tech leads.

Here is a design that should allow for delete, as well as modify facility and validity:

![image](https://cloud.githubusercontent.com/assets/789512/9694527/75acfe6e-5327-11e5-975f-77a6ca2a5461.png)

This should be compatible with the desktop mode as well. The only significant difference would be the Edit page, which could be either a modal or in the right pane. If it is in the right pane the number of reports selected could be modified before saving the edits. This could be a nice touch, but could lead to confusion as well. It is likely simpler to keep the Edit in a modal for now.

bishwas-medic:
Just been inquired about this feature by one of our project officer. The overlapping schedules for training and reporting for different areas and the slow unreliable internet makes it very very hard for them to delete the training reports and monitor the real reports. Would love to see this feature present in the next release.

mandric:
@bishwas-medic this is next on my list after https://github.com/medic/medic-webapp/issues/1764. If you have any input on usability or UX let us know.  As soon as I have something in the works I will post here for feedback.

bishwas-medic:
@mandric beside the UI that Marc has suggested on one of the comments above, it would be great if we could also have an option for "Select All" feature. That would help in deleting large number of reports in very short time. The feature should be:
* Able to select all reports that fits the filter that is applied.
* Usable only after any filter is applied. Shouldn't be usable when no filter is applied to the reports displayed.
* User restricted. Only admin should be able to select all an delete.
* Have "Deselect All" option.

diannakane:
Does deleting a report delete the schedules created by that report? 

And if a person is created in the test process, does a separate deletion process that needs to happen to remove test data?

diannakane:
Include a stop gap "are you sure?" warning dialogue box before bulk deleting, listing how many records are about to be deleted.

[...]

PR #7696, opened 2022-07-26
7653 - Reports filter redesign

latin-panda:
# Description

This PR:
- Adds 2 new components: a filter that displays on the right side of the report's screen, and an actions bar that contains a search input, a button to open the filter and a button to clear all filters.
- Updates the layout and style classes of: reports, reports detail, reports edit and add, actions-bar component. This is to insert the filter as a sidebar.
- Updates html and styles of the search input, form filter, place filter, date range filter and status filter components.
- Refactored existing filter components to support both old and new designs
- Fixes a current issue affecting place filter: if a child place was already selected, and the user selects the parent place, it was unselecting the child. The expected behaviour is to select the parent and all its children. 
- Fixes a current issue when logging telemetry: when a filter is applied, then cleared individually by clicking "clear", telemetry will keep logging the filter as if it was still in use. The expected behaviour is to not log filters with no selected criteria. 
- Fixes a current issue affecting both old and new filter, when select mode is on and some reports are selected, [it shouldn’t let the user filter or search](https://github.com/medic/cht-core/issues/1000#issuecomment-170813237)
- Adds a new permission (`can_view_old_reports_filter `) to turn on the old filter design, admin user can always see the new filter.
- Refactors existing related e2e tests.
- Refactors tour to skip the new filter design only (a decision taken in Care Team)
- Adds a validation message when the "To" date field is before than "From" date field in date range filter.

medic/cht-core#7653

[CHT Docs PR](https://github.com/medic/cht-docs/pull/746)

# Code review checklist

# License


webapp/src/css/dropdown.less
```diff
@@ -109,19 +112,6 @@
       > a[role="menuitem"] {
         padding: 5px 20px 3px 35px;
         text-indent: -20px;
-
-        &.indent-1 {
```
latin-panda, line 113: This isn't scalable, there are projects with more than 6 levels, the [above code](https://github.com/medic/cht-core/pull/7696/files#diff-0b56f98a5ceb46296cd8b10467780cfa4349edf0fdf2171bd6b4e9f2841a45cdR100) fixes that issue.

webapp/src/ts/reducers/global.ts
```diff
@@ -23,6 +23,7 @@ const initialState = {
   },
   facilities: [],
   filters: {},
+  sidebarFilter: {},
```
latin-panda, line 26: The `sidebarFilter` keeps the component state that helps interoperability with other components outside the scope.
The `filter` keeps the actual selected criteria to filter data.

latin-panda:
@jkuester can you please review this PR? I will be checking performance (old vs new filter) and add some more unit test, however I think it's stable enough for a first review.

@tatilepizs can you please have a look at my e2e? I'd like to know if my code is aligned with wdio best practices

Thank you both!

webapp/src/ts/components/filters/facility-filter/facility-filter.component.ts
```diff
@@ -149,7 +160,13 @@ export class FacilityFilterComponent implements OnInit, AbstractFilter, AfterVie
 
   applyFilter(facilities) {
     const facilityIds = facilities.map(facility => facility.doc?._id);
-    this.globalActions.setFilter({ facilities: { selected: facilityIds } });
+    let selectedFacilities;
+
+    if (facilityIds.length) {
```
jkuester, line 165: What is the need for this if-check?  Is there a difference between `facilities` being `null` vs being an object where `selected` is an empty array?

webapp/src/ts/components/filters/facility-filter/facility-filter.component.ts
```diff
@@ -165,9 +182,13 @@ export class FacilityFilterComponent implements OnInit, AbstractFilter, AfterVie
...
 
-  toggle(facility) {
+  private toggle(facility, filter) {
     const recursiveFacilities = this.getFacilitiesRecursive(facility);
-    recursiveFacilities.forEach(facility => this.dropdownFilter.toggle(facility));
+    const newToggleValue = !filter.selected.has(facility);
+    // Exclude places with already correct toggle state, then toggle the rest.
+    recursiveFacilities
+      .filter(place => filter.selected.has(place) !== newToggleValue)
+      .forEach(place => filter.toggle(place));
```
jkuester, line 191: :heart:

webapp/src/ts/components/filters/facility-filter/facility-filter.component.ts
jkuester: Same question applies to the `applyFilter` function for the other filters too.

jkuester (changes requested):
I have only made it though the implementation code so far (tests still pending) but so far things are super great!  I learned a ton about how to implement great responsive UI components by reading through this code!  

Couple suggestions and questions inline.

webapp/src/ts/services/tour.service.ts
```diff
@@ -509,20 +514,34 @@ export class TourService {
...
+      }
+    };
+  }
 
+  private getSettings(name) {
     const settings = this.getTour(name);
+    const skipSettings = this.getSkipSettings()[settings.name];
     settings.autoscroll = false;
+    settings.steps = settings.steps.filter(step => {
+      return !(skipSettings?.condition() && skipSettings.steps.includes(step.element));
```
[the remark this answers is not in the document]
latin-panda: Good points. I think the idea is to someday use another training mechanism and retired the tour, for new development we decided to not use it and keep it just for the old stuff.

I'm choosing a middle point solution: 
1. I'm adding a skip function to the steps we need to skip, so code isn't convoluted
2. I'm going to remove these steps when retrieving the tour in `getTour(name)`. At that point we still have some control over the tour before passing it over to the next process. I'd like to avoid refactoring [bootstrap-tour-standalone.js](https://github.com/medic/cht-core/blob/master/webapp/src/js/bootstrap-tour-standalone.js)

webapp/src/css/sidebar-filter.less
```diff
@@ -0,0 +1,327 @@
...
+          .mm-dropdown-submenu {
+            display: none;
+          }
+
+          &.accordion-caret {
+            &:after {
+              content: '\f0d7'; // https://fontawesome.com/v4/icon/caret-down
+              font-family: FontAwesome;
+              font-size: @font-large;
+              color: @gray-medium;
```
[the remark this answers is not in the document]
latin-panda: yeah, it's already the same shade as in the design, but I'm going to ask Nicole once she's back next Monday. Leaving this comment open for now

webapp/src/ts/modules/reports/reports-sidebar-filter.component.ts
```diff
@@ -0,0 +1,107 @@
...
+    this.filters = [
+      this.formTypeFilter,
+      this.facilityFilter,
+      this.fromDateFilter,
+      this.toDateFilter,
+      this.statusFilter,
+    ];
+  }
+
+  applyFilters(force?) {
```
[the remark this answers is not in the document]
latin-panda: Nice catch

webapp/src/ts/modules/reports/reports-sidebar-filter.component.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  clearFilters(fieldIds?) {
+    const filters = fieldIds ? this.filters.filter(filter => fieldIds.includes(filter.fieldId)) : this.filters;
+    filters.forEach(filter => filter?.clear());
+  }
+
+  countSelected() {
+    this.filterCount.total = 0;
+    this.filters.forEach(filter => {
+      if (!filter?.countSelected) {
```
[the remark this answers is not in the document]
latin-panda: Good catch, we actually don't need to check that function anymore. Before I used to have the search filter inside the filter bar but then we took it out, search filter doesn't have some of these functions

webapp/src/ts/components/filters/facility-filter/facility-filter.component.ts
latin-panda: yes, there is an issue in master. Whenever the user searches, the system is recording [telemetry metrics](https://github.com/medic/cht-core/blob/5de528f9cb74dfff2522831d476081139bd0a09a/webapp/src/ts/services/search.service.ts#L146) about the filters (date, facility, etc) that have selected criteria.

If you select some places and forms, telemetry records `search:reports:places:forms`, then if you go and click "clear" from places. Then telemetry records `search:reports:places:forms`.
The expected record is `search:reports:forms`, because you're not using places filter anymore. 

We need these metrics to know what filters are being actively used by users.

Thinking that we are persisting the selection in rxjs store and that we aren't storing filters the same way (date is object, statuses is array, while forms and places are instance of Set, not sure about the original reason), then I thought it'd be better to store unused filters as `undefined`, to keep a standard filter state for empty, also to fix this issue.... and hopefully avoid other futures issues.

latin-panda: It feels like a filter's responsibility to store it's data okay 🤔
Let me know your thoughts

webapp/src/ts/modules/reports/reports-actions-bar.component.ts
```diff
@@ -0,0 +1,43 @@
...
+
+@Component({
+  selector: 'mm-reports-actions-bar',
+  templateUrl: './reports-actions-bar.component.html'
+})
+export class ReportsActionsBarComponent implements OnInit, OnDestroy {
+  @Output() search: EventEmitter<any> = new EventEmitter();
+  @Output() openFilter: EventEmitter<any> = new EventEmitter();
+  @Output() resetFilter: EventEmitter<any> = new EventEmitter();
+  @Input() disabled;
```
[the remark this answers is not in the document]
latin-panda: It disables the filter and search when the reports select mode is on (`Select` button at the bottom of the screen). Which I just found a bug, I'm going to fix it now.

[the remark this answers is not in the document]
latin-panda: It turns out the issue is also happening in the old filter, so I'm going to fix both.

webapp/src/ts/services/tour.service.ts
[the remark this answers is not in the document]
jkuester: Sounds good!  There is no point in completely refactoring something that we plan to phase out.

webapp/src/ts/components/filters/facility-filter/facility-filter.component.ts
jkuester: Not sure we are talking about the same thing.  Basically, was just wondering why this function needed to be refactored?  Since we are calling `setFilter` either way, it seems like we do not really gain anything from the extra logic.  But, it is not a big deal and should work fine how you have it.  :+1:

webapp/src/ts/modules/reports/reports-sidebar-filter.component.html
```diff
@@ -0,0 +1,119 @@
...
+            #fromDate
+            class="filter"
+            fieldId="fromDateFilter"
+            (search)="applyFilters()"
+            [disabled]="disabled"
+            [isStartDate]="true">
+          </mm-date-filter>
+
+          <label>{{ 'reports.sidebar.filter.to_date' | translate }}</label>
+          <mm-date-filter
```
jkuester, line 74: I expect there is an easy CSS fix for this, but I did notice that the date widget for the `to` date gets cutoff if the window is not tall enough.  (Cannot see the bottom row)

image

webapp/src/ts/components/filters/facility-filter/facility-filter.component.ts
jkuester: Ahh yes, thank you for the detailed explanation. This makes sense now!

webapp/src/ts/modules/reports/reports-sidebar-filter.component.html
latin-panda: This dropdown calendar component is a floating element outside the scroll scope of the container (it won't scroll with the rest of things), its position is calculated with javascript. Other good libraries automatically recalculate and re-position the calendar when it can't fit, but the library that we're using is limited and doesn't do that.

I've added a workaround but it's not perfect. Tiny devices like probably some very old Tecno phone, won't fit when opening on top or below the date input.

I'm recommending to the team to move the "date filter" section on top of the "form filter" that way we ensure the calendar will fit in the screen, even if it's not the most used filter by the user.

The date component has been a challenge while working on this assignment, it has limited features for what we need in the new design, [apparently not supported anymore](https://github.com/dangrossman/daterangepicker/discussions/2266#discussioncomment-2594296), not interest of merging contributions ([55 PR](https://github.com/dangrossman/daterangepicker/pulls)), last release was 2 years ago but we can't upgrade ([Dont upgrade bootstrap-daterangepicker](https://docs.communityhealthtoolkit.org/core/guides/update-dependencies/#problems)) anyways. I quick look for a better option but the ones I found didn't supported Nepali out of the box, we need to look more and change it for a smarter modern option that supports all languages (I'll open a ticket for this)

webapp/src/ts/modules/reports/reports-actions-bar.component.ts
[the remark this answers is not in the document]
latin-panda: Fixed in both

tests/e2e/reports/report-date-filter.wdio-spec.js
```diff
@@ -1,108 +1,63 @@
...
 
-    await reportsTab.filterByDate(moment('05/16/2016', 'MM/DD/YYYY'), moment('05/17/2016', 'MM/DD/YYYY'));
+  it('should filter by date', async () => {
     await (await reportsTab.firstReport()).waitForDisplayed();
 
-    expect(await (await reportsTab.allReports()).length).to.equal(2);
-    expect(await (await reportsTab.reportsByUUID(savedUuids[1])).length).to.equal(1);
-    expect(await (await reportsTab.reportsByUUID(savedUuids[3])).length).to.equal(1);
+    await reportsTab.filterByDate(moment('05/16/2016', 'MM/DD/YYYY'), moment('05/17/2016', 'MM/DD/YYYY'));
+    await browser.pause(300); // Wait for filter to finish
```
tatilepizs, line 55: I think that we should avoid explicit waits as much as we can. If you replace the  `await browser.pause(300)` for `await commonElements.waitForPageLoaded()` it will wait until the filtered reports are displayed.

tests/e2e/reports/reports-sidebar-filter.wdio-spec.js
```diff
@@ -0,0 +1,70 @@
...
+const userFactory = require('../../factories/cht/users/users');
+const placeFactory = require('../../factories/cht/contacts/place');
+const personFactory = require('../../factories/cht/contacts/person');
+const reportFactory = require('../../factories/cht/reports/generic-report');
+
+describe('Reports Sidebar Filter', () => {
+  let savedReports;
+  const today = moment();
+  const parent = placeFactory.place().build({ _id: 'dist1', type: 'district_hospital' });
+  const user = userFactory.build({ username: 'john', roles: [ 'chw' ] });
```
tatilepizs, line 16: Same in here, you could remove the parameter sent to the `userFactory.build()`, it will create by default a user named "user1" with the role "chw"

tests/e2e/reports/reports-sidebar-filter.wdio-spec.js
```diff
@@ -0,0 +1,70 @@
...
+
+  it('should filter by date', async () => {
+    await (await reportsPage.firstReport()).waitForDisplayed();
+    await reportsPage.openSidebarFilter();
+    await reportsPage.openSidebarFilterAccordion('#date-filter-accordion');
+
+    await reportsPage.setSidebarFilterDate('#fromDateFilter');
+    await reportsPage.setSidebarFilterDate('#toDateFilter', 2, 'r3c5');
+
+    await browser.pause(300); // Wait for filter to finish
```
tatilepizs, line 62: Same here, it will be better if we avoid explicit waits as much as we can. If you replace the `await browser.pause(300)` for `await commonElements.waitForPageLoaded()` it will wait until the filtered reports are displayed.

tests/page-objects/reports/reports.wdio.page.js
```diff
@@ -196,6 +199,33 @@ const filterByDate = async (startDate, endDate) => {
...
+
+const openSidebarFilterAccordion = async (selector) => {
+  await (await $(`${selector} .panel-heading`)).click();
+  return (await $(`${selector} .panel-collapse.show`)).waitForDisplayed();
+};
+
+const setSidebarFilterDate = async (fieldId, calendarIdx = 1, day = 'r1c2') => {
+  await (await $(fieldId)).waitForDisplayed();
+  await (await $(fieldId)).click();
+  await browser.pause(300); // Wait for animation to finish
```
tatilepizs, line 215: After I added the `commonElements.waitForPageLoaded()` in the `reports-sidebar-filter.wdio-spec.js` file I removed this `browser.pause(300)` and the test passed successfully.
Could you please re-check if you really need it?

tests/page-objects/reports/reports.wdio.page.js
```diff
@@ -196,6 +199,33 @@ const filterByDate = async (startDate, endDate) => {
...
+  const dateRangePicker = `.daterangepicker:nth-of-type(${calendarIdx})`;
+  await (await $(dateRangePicker)).waitForDisplayed();
+
+  const leftArrow = $(`${dateRangePicker} .table-condensed th>.fa-chevron-left`);
+  await (await leftArrow).click();
+
+  const date = $(`${dateRangePicker} .table-condensed tr td[data-title="${day}"]`);
+  await (await date).click();
+
+  await browser.pause(300); // Wait for animation to finish
```
tatilepizs, line 226: Same for this `pause`. Could you please re-check if you really need it?

tests/e2e/reports/reports-sidebar-filter.wdio-spec.js
```diff
@@ -0,0 +1,70 @@
...
+    await commonElements.goToReports();
+  });
+
+  it('should filter by date', async () => {
+    await (await reportsPage.firstReport()).waitForDisplayed();
+    await reportsPage.openSidebarFilter();
+    await reportsPage.openSidebarFilterAccordion('#date-filter-accordion');
+
+    await reportsPage.setSidebarFilterDate('#fromDateFilter');
+    await reportsPage.setSidebarFilterDate('#toDateFilter', 2, 'r3c5');
```
tatilepizs, line 60: I think it will be better if we keep the "selectors" in the `*.page` file. Could you please create constants for those selectors in the `reports.wdio.page.js` file.

tatilepizs (changes requested):
Thank you for creating this e2e test Jenny, it looks really good!

I just have some suggestions.

tests/e2e/reports/report-date-filter.wdio-spec.js
latin-panda: Quite interesting function with nested loops (do-while, for), assuming that we have few elements with `.container-fluid .loader` and that hopefully wdio query aren't expensive then it should be okay. Done!

tests/page-objects/reports/reports.wdio.page.js
latin-panda: Oh wooooow, I ran it few times and haven't failed, Thanks!

webapp/src/css/sidebar-filter.less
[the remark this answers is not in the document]
latin-panda: I've sent the question

tests/page-objects/reports/reports.wdio.page.js
```diff
@@ -257,6 +288,11 @@ module.exports = {
   getTaskState,
   openForm,
   formTitle,
+  sidebarFilterToDate,
+  sidebarFilterFromDate,
+  openSidebarFilter,
+  openSidebarFilterDateAccordion,
+  setSidebarFilterDate,
```
jkuester, line 295: Minor, but it would encapsulate some complexity within this util (instead of exposing it to the tests) if you were to export separate `setSidebarFilterFromDate` and `setSidebarFilterToDate` functions instead of exporting `sidebarFilterToDate`, `sidebarFilterFromDate`, and `setSidebarFilterDate`.  The new functions could still use those three internally, but the consuming test code does not need to know about them.

jkuester (approved):
I am very excited about this functionality!  Great work!

webapp/src/css/sidebar-filter.less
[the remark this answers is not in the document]
latin-panda: It's going to stay like this for now and get user feedback

latin-panda:
Hi @jkuester I pushed 3 small changes, in different commits hoping that's easier to review:
1. Now the new filter is the default, but partners can use the `can_view_old_reports_filter` to display the old filter if they want.
2. Nicole send me some feedback about the date filter error. The error message will be display after the 2 fields (From and To) and the field border is going to be red when there's error.
3. A tiny performance improvement when selecting a parent place, once all children places are selected then we just send to search for reports once. 

Wondering if you would like to take a look before I send it to AT? :)

latin-panda:
I pushed small changes addressing feedback from this morning.

tatilepizs (approved):
Great!

jkuester (approved):
This looks good to go!

latin-panda:
Thank you both! 🎉

PR #7710, opened 2022-08-02
Fix e2e test - delivery form

latin-panda:
# Description

This PR fixes flaky e2e test:
- Updates the target in default config to not consider contacts without date of birth
- Updates test data, so new born's date of birth is "today", this will make the target count the new born even when running the test on the 1st day of the month.
- Updates "Live births" expected count to 1.
- Updates rules integration since expected targets for the pregnancy is 0, previously the new born target was generated. 
- Fixes analytics e2e that had the same problem with new born target, it was counting the user's contact without date of birth.

# Code review checklist

# License


shared-libs/rules-engine/test/rules-emitter.spec.js
```diff
@@ -214,7 +214,8 @@ describe('rules-emitter', () => {
 
       const { tasks, targets } = await rulesEmitter.getEmissionsFor([chtDocs.contact], [chtDocs.pregnancyReport]);
       expect(tasks.length).to.eq(1);
-      expect(targets.length).to.eq(1);
```
latin-panda, line 217: This target isn't related to pregnancies.
Screen Shot 2022-08-02 at 1 09 21 pm

latin-panda:
Hi @dianabarsan when you have time, can you please have a look? It's a small one
Thanks!

dianabarsan (approved):
Awesome! Thanks a lot for looking into this.

PR #7715, opened 2022-08-04
Automated test for immunization workflow on standard config

tatilepizs:
# Description

Add the e2e automated test for the immunization workflow on Standard config. To run it use the command `npm run standard-wdio-local`

This test:
- Create a new child under 2 years using SMS (CW form).
- Verify the record in reports for the child.
- Create an immunization visit (webapp)
- Verify the record in reports for the visit.
- Create an immunization visit using SMS (IMM form).
- Verify the record in reports for the visit.
- Verify that the targets were updated.

[medic/cht-core #5](https://github.com/medic/cht-release-testing/issues/5)

# Code review checklist

# License


tatilepizs:
@latin-panda, @ngaruko could you please help me to review this PR? 

@jonathanvq, I also made changes to the automated guide that we created. I added information related to the standard config, could you please help me to review it?

Thanks!

tests/page-objects/contacts/contacts.wdio.page.js
```diff
@@ -286,5 +344,8 @@ module.exports = {
   createNewAction,
   openReport,
   getContactCardTitle,
-  getContactInfoName
+  getContactInfoName,
+  getContactMedicID,
+  addHealthPrograms,
+  getImmCardVaccinesValues
```
latin-panda, line 350:
```suggestion
  getImmCardVaccinesValues,
```
Next time someones add a new function, git will show 1 change in the diff.

tests/page-objects/contacts/contacts.wdio.page.js
```diff
@@ -253,6 +276,40 @@ const getContactInfoName = async () => {
...
+ * Add health programs to a health facility, standard config
+ * 
+ * @param {string} program 
+ * Refers to the different Health Programs, "anc" for Antenatal care,  "pnc" for Postnatal care,  "imm" for Immunizations and "gpm" for Growth monitoring (nutrition)
+ */
+const addHealthPrograms = async (program = 'anc') => {
+  await (await editContactButton()).waitForDisplayed();
+  await (await editContactButton()).click();
+  await healthProgram(program).click();
+  if(program === 'imm'){
```
latin-panda, line 294:
```suggestion
  if (program === 'imm') {
```

latin-panda: You could have a constant for these programs. Usually constants are uppercase on the top of the file.
```
const HEALTH_PROGRAMS = { IMM: 'imm', ANC: 'anc' };
```
If you need the label as well, then:
```
const HEALTH_PROGRAMS = { 
   IMM: {
      code: 'imm',
      label: 'Immunizations'
   },
};
```

tests/page-objects/forms/generic-form.wdio.page.js
```diff
@@ -63,6 +63,10 @@ const verifyReport = async () => {
   expect(validatedReport.patient).to.be.undefined;
 };
 
+const submitForm = async () => {
```
latin-panda, line 66: Since no more instructions after click you can do this:
```
const submitForm = () => submitButton().click();
```

tests/page-objects/forms/generic-form.wdio.page.js
```diff
@@ -72,5 +76,6 @@ module.exports = {
   nameField,
   selectContact,
   verifyReport,
-  editForm
+  editForm,
+  submitForm
```
latin-panda, line 80:
```suggestion
  submitForm,
```

tests/page-objects/contacts/contacts.wdio.page.js
```diff
@@ -56,6 +56,13 @@ const forms = () => $$('.action-container .detail-actions .actions.dropup .open
 const formTitle = () => $('#form-title');
 const contactCardTitle = () => $('.inbox .content-pane .material .body .action-header');
 const contactInfoName = () => $('.content-pane .material .body .card .row .heading-content');
+const contactMedicID = () => $('div[class = "cell patient_id"] > div > p');
+
+const healthProgram = (program) => $(`input[name="/data/health_center/use_cases"][value="${program}"]`);
+const vaccines = () => $$('input[name="/data/health_center/vaccines"]');
+const immCardVaccinesSelector = 'div[class="col col-sm-6 col-xs-12"] > div > div > p';
```
latin-panda, line 63:
```suggestion
const IMM_CARD_VACCINES_SELECTOR = 'div[class="col col-sm-6 col-xs-12"] > div > div > p';
```

tests/page-objects/forms/immunization-visit-form.wdio.page.js
```diff
@@ -0,0 +1,93 @@
+let countAppliedVaccines = 0;
```
latin-panda, line 1: I'm not 100% convinced of keeping a data state here, can you move it to `immunization-visit-form.standard-wdio-spec.js` just after `let babyMedicID = '';`? 

I think page objects are helpers to get and set DOM elements. Once the vaccine is applied then the page object tells the spec that it was applied and the spec keeps record or it.

tests/page-objects/forms/immunization-visit-form.wdio.page.js
```diff
@@ -0,0 +1,93 @@
...
+const YELLOW_FEVER_VACCINE = 'input[name="/immunization_visit/group_yellow_fever/g_yellow_fever"]';
+
+const notes = () => $('textarea[name="/immunization_visit/group_note/g_chw_sms"]');
+
+const selectAppliedVaccines = async (selector, option = "no") => {
+    const vaccinesSelector = await $$(`${selector}[value*="${option}"]`);
+    for (const vaccine of vaccinesSelector) {
+        await vaccine.click();
+    }
+    countAppliedVaccines += !option.includes('no') ? vaccinesSelector.length: 0;
```
latin-panda, line 28:
```suggestion
    countAppliedVaccines += option.includes('no') ? 0 : vaccinesSelector.length;
```
It's faster to read and understand positive statement.

tests/page-objects/forms/immunization-visit-form.wdio.page.js
```diff
@@ -0,0 +1,93 @@
...
+};
+
+const selectAllVaccines = async () => {
+    const cbVaccines = await $$('input[name="/immunization_visit/group_select_vaccines/g_vaccines"]');
```
latin-panda, line 32: Does it make sense that this is a selector on top of the page?

tests/page-objects/forms/immunization-visit-form.wdio.page.js
```diff
@@ -0,0 +1,93 @@
...
+    for (const vaccine of cbVaccines) {
+        await vaccine.click();
+    }
+};
+
+const addNotes = async (note = 'Test notes') => {
```
latin-panda, line 38:
```
const addNotes = (note = 'Test notes') => notes().setValue(note);
```
Same with getNotes, getAppliedVaccines

tests/page-objects/forms/immunization-visit-form.wdio.page.js
latin-panda: same with getPatientNameSummaryPage, getAppliedVaccinesSummary, getAppliedVaccinesSummary

tests/page-objects/forms/immunization-visit-form.wdio.page.js
```diff
@@ -0,0 +1,93 @@
...
+  VITAMIN_A_VACCINE,
+  YELLOW_FEVER_VACCINE,
+  selectAllVaccines,
+  selectAppliedVaccines,
+  addNotes,
+  getNotes,
+  getPatientNameSummaryPage,
+  getAppliedVaccines,
+  getAppliedVaccinesSummary,
+  getFollowUpSMS
```
latin-panda, line 92:
```suggestion
  getFollowUpSMS,
```
I like that you are exporting all constants first, then functions, it's so organised

latin-panda (changes requested):
This is great! I left some comments, suggestions, questions, etc 
:)

tatilepizs:
@latin-panda could you please help me to review the target for `Vaccines given`? it is counting 1 when I selected all the vaccines, I am not sure if it counts the visits or the number of vaccines received in the visit.

tests/AUTOMATE_TEST_GUIDE.md
```diff
@@ -2,7 +2,19 @@
...
+There are three files that are the base of every new automated test case, the most important one is the `spec.js`, which contains the actual test that will be executed. 
+
+Right now we have two different configs set up:
+* **default**
+    * Config file: [`wdio.conf.js`](https://github.com/medic/cht-core/blob/master/tests/wdio-standard.conf.js)
+    * Name convention for the `spec` file: `<name>.wdio-spec.js`
+    * Command to execute the tests that belong to this config:  `npm run wdio-local`
+* **standard**
+    * Config file: [`wdio-standard.conf.js`](https://github.com/medic/cht-core/blob/master/tests/wdio.conf.js)
+    * Name convention for the `spec` file: `<name>.standard-wdio-spec.js`
```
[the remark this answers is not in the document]
latin-panda: > Create the standard folder inside e2e. I removed the standard part from the name of the js files because as you said, they are already in the standard folder. This is what it contains

This is great! We'll have 1 pattern, easy to understand, easy to replicate if we automate more configs like BRAC. Thank you!

> Update the command to execute the standard test: npm run standard-wdio-local because it makes more sense that the word standard is at the beginning

Sounds good 👍

tests/e2e/forms/immunization-visit-form.standard-wdio-spec.js
```diff
@@ -0,0 +1,166 @@
...
+const healthCenter = places.find(place => place.type === 'health_center');
+const user = userFactory.build({ place: healthCenter._id, roles: ['district_admin'] });
+const babyName = 'Baby1';
+let babyMedicID = '';
+
+describe('Immunization Visit', async() => {
+    before(async() => {
+        await utils.saveDocs([...places]);
+        await utils.createUsers([user]);
+        await loginPage.cookieLogin();
```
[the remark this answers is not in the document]
latin-panda: > Do you think that I should create a new user for the admin part and use only the loginPage.login(user) with two different users?

It's okay like this. I just wanted to learn if one was better option than the other one.

tests/e2e/forms/immunization-visit-form.standard-wdio-spec.js
```diff
@@ -0,0 +1,166 @@
...
+        await immunizationVisitForm.selectAppliedVaccines(immunizationVisitForm.VITAMIN_A_VACCINE, 'yes');
+        await genericForm.nextPage();
+        await immunizationVisitForm.selectAppliedVaccines(immunizationVisitForm.YELLOW_FEVER_VACCINE, 'yes');
+        await genericForm.nextPage();
+        await immunizationVisitForm.addNotes();
+        await genericForm.nextPage();
+
+        expect(await immunizationVisitForm.getPatientNameSummaryPage()).to.equal(babyName);
+        expect(await immunizationVisitForm.getAppliedVaccines()).to.equal(await immunizationVisitForm.getAppliedVaccinesSummary() - 1);
+        expect(await immunizationVisitForm.getFollowUpSMS()).to.include(babyName && await immunizationVisitForm.getNotes());
```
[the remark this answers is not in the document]
latin-panda: The baby name is `Baby1`, so I modified the assertion to look for `raul`, but the test still passed. It should have failed because the name isn't the expected. 
```
expect(await immunizationVisitForm.getFollowUpSMS()).to.include('raul' && babyMedicID && await immunizationVisitForm.getNotes());
```

The reason is because in Javascript this statement:
```
babyName && babyMedicID && await immunizationVisitForm.getNotes()
```
It's the equivalent of:
```
if (babyName !== undefined && babyMedicID !== undefined) {
   await immunizationVisitForm.getNotes();
}
```
An easy test in the browser's console would be:
```
"Hello miss, can I have mac & cheese please?".includes("miss" && "mac & cheese"); // returns true
"Hello miss, can I have mac & cheese please?".includes("Cleo" && "mac & cheese"); // returns true
```
I way to fix it is checking if the message include each of these separately:
```
expect(message).to.include(babyName);
expect(message).to.include(babyID);
... etc
```

tests/e2e/forms/immunization-visit-form.standard-wdio-spec.js
```diff
@@ -0,0 +1,166 @@
...
+        await genericForm.nextPage();
+        await immunizationVisitForm.selectAppliedVaccines(immunizationVisitForm.VITAMIN_A_VACCINE, 'yes');
+        await genericForm.nextPage();
+        await immunizationVisitForm.selectAppliedVaccines(immunizationVisitForm.YELLOW_FEVER_VACCINE, 'yes');
+        await genericForm.nextPage();
+        await immunizationVisitForm.addNotes();
+        await genericForm.nextPage();
+
+        expect(await immunizationVisitForm.getPatientNameSummaryPage()).to.equal(babyName);
+        expect(await immunizationVisitForm.getAppliedVaccines()).to.equal(await immunizationVisitForm.getAppliedVaccinesSummary() - 1);
```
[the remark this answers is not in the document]
latin-panda: Yeah, these enketo pages are harder. I think it's better in the `wdio.page file`.

Adding `:not(:last-child)` worked for me: 
`$('label.question.note.or-branch.non-select.or-appearance-li:not(:last-child)')`
Screen Shot 2022-08-05 at 12 30 31 pm

However you will need to add a comment why excluding the last child since it's not straight forward to understand.

tests/page-objects/contacts/contacts.wdio.page.js
```diff
@@ -56,6 +56,13 @@ const forms = () => $$('.action-container .detail-actions .actions.dropup .open
 const formTitle = () => $('#form-title');
 const contactCardTitle = () => $('.inbox .content-pane .material .body .action-header');
 const contactInfoName = () => $('.content-pane .material .body .card .row .heading-content');
+const contactMedicID = () => $('div[class = "cell patient_id"] > div > p');
```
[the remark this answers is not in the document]
latin-panda: > If I add .app-root.contact it is not finding anything,

Ups sorry I missed a `s`, this one works: `$('.app-root.contacts #contact_summary .cell.patient_id > div > p')`
Screen Shot 2022-08-05 at 11 55 29 am

As you did it with `#contact_summary` it's already good.

The high level classes are here:
Screen Shot 2022-08-05 at 11 56 46 am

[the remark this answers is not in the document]
latin-panda: > const immCardVaccinesSelector = 'div[class="col col-sm-6 col-xs-12"] > div > div > p';

The thing is, contact summary cards doesn't have an ID, or at least I can't see it.
So , to find values in the immunisation card, you can use the card's label and the field's label, like this in [contacts-content.component.html](https://github.com/medic/cht-core/blob/master/webapp/src/ts/modules/contacts/contacts-content.component.html#L52)

Screen Shot 2022-08-05 at 12 41 54 pm
Screen Shot 2022-08-05 at 12 40 53 pm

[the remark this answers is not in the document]
latin-panda: > const healthProgram = (program) => $(`input[name="/data/health_center/use_cases"][value="${program}"]`);
> const vaccines = () => $$('input[name="/data/health_center/vaccines"]');

Forms have id (`immunization_visit`),  what do you think? 🤔
Screen Shot 2022-08-05 at 12 51 53 pm

latin-panda:
> could you please help me to review the target for Vaccines given? 

Sure! I'll have a close look on Monday. 
I've replied the other comments but I still need to do a proper review later :) Great stuff!

tests/e2e/forms/immunization-visit-form.standard-wdio-spec.js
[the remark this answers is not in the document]
tatilepizs: That worked perfectly, thank you!

tests/e2e/forms/immunization-visit-form.standard-wdio-spec.js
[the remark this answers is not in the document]
tatilepizs: Oh wow! I didn't know that and I didn't test it. Thank you!!

tests/page-objects/contacts/contacts.wdio.page.js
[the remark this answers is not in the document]
tatilepizs: That is great, thanks Jenny. Done!

tests/page-objects/contacts/contacts.wdio.page.js
```diff
@@ -124,6 +132,21 @@ const addPlace = async (type, placeName, contactName) => {
...
+// Default config
+const addPlace = async (type, placeName, contactName) => {
+  await addPlacePrimaryContact(type, contactName);  
+  await (await dateOfBirthField()).addValue('2000-01-01');
+  await (await contactSexField()).click();
+  await genericForm.nextPage();
+  await addPlaceDetails(type, placeName);
+};
+
+const addPlaceStandardConfig = async (type, placeName, contactName) => {
```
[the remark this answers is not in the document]
tatilepizs: Thank you for the help Jenny. Done!

latin-panda:
@tatilepizs The ["Vaccines given" target](https://github.com/medic/cht-core/blob/c7a5c03b18497b8bea8324b74d8070f0ce57a37b/config/standard/targets.js#L367) was configured to count 1 per every different instance (`inst` in the code). It won't count the number of vaccines the patient got in one visit and it won't count the immunisation visits.

In my tests the instance was `37bfee59-c6b7-44f1-9742-9f539587299f~imm-vaccines-given-this-month01234567891011121314151617181920212223242526272829303132333435363738`
Where `37bfee59-c6b7-44f1-9742-9f539587299f` is my patient id 
While `01234567891011121314151617181920212223242526272829303132333435363738` is a concatenation from 0 to the quantity of vaccines applied (not very strong logic 👀)

So if you submit 2 immunisation reports in the same date with the same quantity of vaccines, it will count as 1. 

In my opinion:
- If we agree that this target is about the number of vaccines given, then we need to change the target to count 5 vaccines when a patient got 5 different vaccines. 
- If we agree that this target is about visits, then the target title is confusing ("Vaccines given") and it should be changed. Then fix the instance id to count the visits even if they occurred in the same date and same quantity of vaccines where applied.

@garethbowen do you remember the intention of this target? Maybe we don't fully understand it, otherwise we can fix it 🤔 as per above.

garethbowen:
Oof, that looks messy. I think the change was introduced in [this commit](https://github.com/medic/cht-core/commit/84cc7caf936f829be831801502c29477bdcefb4d). The change is instead of creating a brand new target in the loop, we just modify the ID and emit the same object over and over.

The target _should_ still work because it's emitting a different ID for every dose given, but it's not very clean code.

Firstly, double check that my assertion is correct and the target is counting every dose for every visit.
Secondly, it'd be worth cleaning this up so it doesn't confuse the next reader, so the ID is more readable.

latin-panda (approved):
Great job ✨
I left one question below

latin-panda:
Ticket to fix targets [is here](https://github.com/medic/cht-core/issues/7719)

tests/e2e/standard/forms/immunization-visit.standard-wdio-spec.js
```diff
@@ -0,0 +1,170 @@
...
+const loginPage = require('../../../page-objects/login/login.wdio.page');
+const commonPage = require('../../../page-objects/common/common.wdio.page');
+const contactPage = require('../../../page-objects/standard/contacts/contacts.wdio.page');
+const reportsPage = require('../../../page-objects/reports/reports.wdio.page');
+const analyticsPage = require('../../../page-objects/analytics/analytics.wdio.page');
+const genericForm = require('../../../page-objects/forms/generic-form.wdio.page');
+const placeFactory = require('../../../factories/cht/contacts/place');
+const userFactory = require('../../../factories/cht/users/users');
+const gatewayApiUtils = require('../../../gateway-api.utils');
+const immVisitForm = require('../../../page-objects/forms/immunization-visit.wdio.page');
```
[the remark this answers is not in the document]
tatilepizs: I thought about that, the thing is that this immunization page will not be used only for `standard` because there is a test in `brac` for the immunization and I think that the page can be used also in that test.
I think that we will need to find a different way to organize the `page-object` files.

webapp/src/ts/modules/contacts/contacts-content.component.html
```diff
@@ -49,19 +49,19 @@ <h2>{{selectedContact?.doc?.name}}</h2>
 
         <ng-container *ngIf="selectedContact?.summary">
           <div *ngFor="let card of selectedContact?.summary?.cards">
-            <div class="card">
+            <div class="card" [attr.test-id]="card.label">
```
ngaruko, line 52: Good job adding these test attributes

ngaruko (commented):
Great job. That was one big spec. Also, good you added a new test config file and updated the test guide 👍 .

ngaruko (approved):

webapp/src/ts/modules/contacts/contacts-content.component.html
tatilepizs: Thanks Bede! I received help from Jenny 😊

Issue #5697, opened 2019-05-29
Update the levels displayed in breadcrumbs across the app


amandacilek-zz:
We use breadcrumbs to display contextual information about where a person/place belongs in the app hierarchy. Depending on the user, sometimes the breadcrumbs can contain redundant and unnecessary information. In this example, I am Amanda Cilek, and "Amanda Cilek's Area" is the place I belong to. I do not need to see "Amanda Cilek's Area" in the breadcrumbs of every single report that I have submitted.
![Screen Shot 2019-05-28 at 4 52 21 PM](https://user-images.githubusercontent.com/26235957/58519816-03190a80-816a-11e9-9255-5acbc8f7763d.png)

This request is to update how we handle breadcrumbs so that we show all levels **except for** the level the user themselves belongs to, the top-most level. The data underneath will still have the entire hierarchy, we'll just choose how much to show based on the logged in user. For example, when I look at my reports tab in the above instance, I should no longer see "Amanda Cilek's Area" at the end of every breadcrumb. However if my manager logged in and looked at the same reports, my manager _would_ still see "Amanda Cilek's Area".

This update should apply to wherever breadcrumbs show on the **LEFT HAND SIDE** of the app (currently Messages and Reports, but also Tasks once [#5710](https://github.com/medic/cht-core/issues/5710) is done). 

Wherever breadcrumbs are shown on the **RIGHT HAND SIDE**, we should continue to show all breadcrumbs as users often use these as a shortcut to navigate back to the top level.

MaxDiz:
Hi @garethbowen This issue should be completed alongside #5710 (currently scheduled in v3.8) to make it functional. I also have a call with @amandacilek and @yembrick this week to review any outstanding design work on these issues and will report back

garethbowen:
Ok. Adding to 3.8.0.

maremegaye:
From the above Amanda’s post to now few breadcrumbs have been deleted. Hereunder a table that describes the breadcrumbs deleted and those not deleted across different users and different projects. 
- For the CHW, we will need to delete the breadcrumbs in the report tab
- For the Supervisor, we will delete the breadcrumbs in the report and message tabs
- For the Admin view, we will need to delete the breadcrumbs of the message tab

https://docs.google.com/document/d/1t9wSTpNHnxyprVy-EX5MxGD8_YMTZ1qIZrg9gVb93LQ/edit

garethbowen:
[...]

Issue #5710, opened 2019-06-04
Add "belongs to" breadcrumbs to tasks in the LHS list


amandacilek-zz:
#5697 is related and either needs to be completed before, or at the same time as, this request.

**User stories**
- As a CHW, when I look at my tasks list, I would like to know the family or primary contact that an individual person belongs to, so that I can identify patients easier. (For example, tasks about children currently only display the child’s name. But children often have similar names and some parents wait months to give their baby a name. The child’s name is not always enough information.)

- As a CHW, when I look at my tasks list, I would like to know the family or primary contact that an individual person belongs to, so that I can better plan my visits and can complete all tasks for the same family at the same time. (Currently a CHW may have more than one open task for the same family, but because the tasks list is sorted by due date and doesn’t list the family name, they may miss other tasks for the same family.)

- As a CHW, I would like tasks for the same family to be grouped together so that I can be more efficient with my visit time and planning. (This spec/feature doesn’t address grouping, but just noting that displaying the family name is probably a desired prerequisite to sorting by family.)

**Summary**
- After discussing many possibilities, we decided that we would like to add a third line of text to tasks to display the additional contextual information.
- This new line of text should display breadcrumbs showing where the person/place the task is about belongs to. 
- The functionality should match how we display breadcrumbs elsewhere on the app. Note that we are updating breadcrumbs across the app so that they do not list out the level the user themselves belongs to repetitively (i.e. CHW Janet will not see “CHW Janet’s Area” repeated all over the place, ticket #5697).
- The text should be styled just like breadcrumbs elsewhere on the app (Noto Sans 14px).

**Examples**
- For CHWs on the webapp looking at a task about a particular person, the breadcrumbs should most often be just the family name (e.g. “Beatrice Bass Family”). Only a single level is listed because the next level, CHW Janet’s Area, is the CHW’s own level. 
![Screen Shot 2019-05-22 at 12 16 03 PM](https://user-images.githubusercontent.com/26235957/58914178-b8037800-86d2-11e9-95f9-309408a24c34.png)

- Some tasks are for families instead of individual people, like family surveys or bed net distribution. In these cases, the breadcrumbs may not display at all, since there are no levels between the family and the CHW area.
![Screen Shot 2019-05-22 at 12 13 35 PM](https://user-images.githubusercontent.com/26235957/58914201-c2be0d00-86d2-11e9-84a6-8811ee497dae.png)

[...]

PR #7732, opened 2022-08-16
#5697 - Update the levels displayed in breadcrumbs across the app

elvisdorkenoo:
# Update the levels displayed in breadcrumbs across the app

This request is to update how we handle breadcrumbs so that we show all levels except for the level the user themselves belongs to, the top-most level.
Summary of expected behavior [here](https://docs.google.com/presentation/d/1yHRZc3B6VIhABeVOrKJ-BRRmY_X2zr1cVFAKOvHwbw4/edit#slide=id.g14477939221_0_218) : 

medic/cht-core#5697

# Code review checklist

# License


webapp/src/ts/modules/reports/reports.component.ts
```diff
@@ -191,8 +200,13 @@ export class ReportsComponent implements OnInit, AfterViewInit, OnDestroy {
       report.heading = this.getReportHeading(form, report);
       report.summary = form ? form.title : report.form;
       report.lineage = report.subject && report.subject.lineage || report.lineage;
+      // filter out the lineage level that belongs to the online logged in user
+      if(!this.sessionService.isOnlineOnly()) {
+        report.lineage = report.lineage.filter((level) => {
+          return (level !== this.currentLevel);
+        });
```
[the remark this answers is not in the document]
jkuester: That being said, will the user's currentLevel always be the last element in the lineage?  I think so, or the existing behavior of this would be kind of weird.

If that is the case, then it would be cheaper to just focus on the last element instead of looping through all of them...

```js
        if(report.lineage.length && report.lineage[length - 1] === this.currentLevel) {
          report.lineage.pop();
        }
```

webapp/tests/karma/ts/modules/reports/reports.component.spec.ts
```diff
@@ -66,7 +78,10 @@ describe('Reports Component', () => {
       isOnlineOnly: sinon.stub()
     };
     datePipe = { transform: sinon.stub() };
-
+    userContactService = {
+      get: sinon.stub().resolves(userContactDoc),
+      getCurrentLineageLevel : sinon.stub().resolves('parent')
```
jkuester, line 83: Should include tests to cover the various new logic.  Make sure that the current lineage level for the user is filtered out from the displayed lineage (but only when the user is offline).

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
```diff
@@ -181,4 +202,11 @@ describe('Messages Component', () => {
 
     expect(spySubscriptionsUnsubscribe.callCount).to.equal(1);
   });
+
+  it('it should retrieve the hierarchy level of the connected user', () => {
```
jkuester, line 206: Should include tests to cover the various new logic.  Make sure that the current lineage level for the user is filtered out from the displayed lineage (but only when the user is offline).

jkuester (changes requested):
This is a solid start!  I have added some questions/suggestions inline.

webapp/src/ts/modules/messages/messages.component.ts
```diff
@@ -152,6 +160,15 @@ export class MessagesComponent implements OnInit, OnDestroy {
     return this.messageContactService
       .getList()
       .then((conversations = []) => {
+        // filter out the lineage level that belongs to the online logged in user
+        if(!this.sessionService.isOnlineOnly()) {
+          conversations.map((conversation) => {
```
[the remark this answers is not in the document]
latin-panda: +1 
If the structure of the objects contained in `conversations` is the same, then the second option sounds better.

webapp/src/ts/modules/messages/messages.component.ts
```diff
@@ -152,6 +160,15 @@ export class MessagesComponent implements OnInit, OnDestroy {
     return this.messageContactService
       .getList()
       .then((conversations = []) => {
+        // filter out the lineage level that belongs to the online logged in user
+        if(!this.sessionService.isOnlineOnly()) {
```
latin-panda, line 164:
```suggestion
        if (!this.sessionService.isOnlineOnly()) {
```
As per the code style, we keep spaces round the brackets for `if` statements

webapp/src/ts/modules/messages/messages.component.ts
```diff
@@ -48,6 +53,9 @@ export class MessagesComponent implements OnInit, OnDestroy {
   ngOnInit(): void {
     this.subscribeToStore();
     this.tourService.startIfNeeded(this.route.snapshot);
+    this.userContactService.getCurrentLineageLevel().then((currentLevel) => {
```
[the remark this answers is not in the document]
latin-panda: When having more than one function in chain. Please put each function in a separate line like this:
```
this.fruitService
   .harvestBananas()
   .collectBananas();
```
This is a common practice in functional programming 

Also you can simplify assignments like this: 
```
this.fruitService.then(bananas => this.basket = bananas);
```

webapp/src/ts/modules/messages/messages.component.ts
```diff
@@ -40,6 +43,8 @@ export class MessagesComponent implements OnInit, OnDestroy {
     private modalService: ModalService,
     private tourService: TourService,
     private responsiveService: ResponsiveService,
+    private userContactService:UserContactService,
```
latin-panda, line 46:
```suggestion
    private userContactService: UserContactService,
```
Looks like all the other parameters are adding this space, we don't have a rule for this in specific but I think it improves readability.

webapp/src/ts/modules/messages/messages.component.ts
```diff
@@ -40,6 +43,8 @@ export class MessagesComponent implements OnInit, OnDestroy {
...
+    private sessionService:SessionService,
```
latin-panda, line 47:
```suggestion
    private sessionService: SessionService,
```

webapp/src/ts/modules/messages/messages.component.ts
```diff
@@ -152,6 +160,15 @@ export class MessagesComponent implements OnInit, OnDestroy {
     return this.messageContactService
       .getList()
       .then((conversations = []) => {
+        // filter out the lineage level that belongs to the online logged in user
+        if(!this.sessionService.isOnlineOnly()) {
+          conversations.map((conversation) => {
+            conversation.lineage = conversation.lineage?.filter((level) => {
```
latin-panda, line 166: You can make this filter function 1 line, as suggested before.

webapp/src/ts/modules/reports/reports.component.ts
```diff
@@ -26,6 +26,7 @@ import { TranslateService } from '@mm-services/translate.service';
 import { ReportsSidebarFilterComponent } from '@mm-modules/reports/reports-sidebar-filter.component';
 import { AuthService } from '@mm-services/auth.service';
 import { OLD_REPORTS_FILTER_PERMISSION } from '@mm-modules/reports/reports-filters.component';
+import {UserContactService} from '@mm-services/user-contact.service';
```
latin-panda, line 29:
```suggestion
import { UserContactService } from '@mm-services/user-contact.service';
```
I think the IDE should have settings for all this. If you're using WebStorm then: Preferences > code style > javascript or typescript > Spaces. I'm using Webstorm and it adds the imports with the space around it

webapp/src/ts/modules/reports/reports.component.ts
```diff
@@ -143,6 +147,11 @@ export class ReportsComponent implements OnInit, AfterViewInit, OnDestroy {
     this.globalActions.setFilter({ search: this.route.snapshot.queryParams.query || '' });
     this.tourService.startIfNeeded(this.route.snapshot);
     this.setActionBarData();
+
+    this.userContactService.getCurrentLineageLevel().then((currentLevel) => {
```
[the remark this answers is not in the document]
latin-panda: same feedback about chaining function - one function per line please.

webapp/src/ts/modules/reports/reports.component.ts
```diff
@@ -191,8 +200,13 @@ export class ReportsComponent implements OnInit, AfterViewInit, OnDestroy {
       report.heading = this.getReportHeading(form, report);
       report.summary = form ? form.title : report.form;
       report.lineage = report.subject && report.subject.lineage || report.lineage;
+      // filter out the lineage level that belongs to the online logged in user
+      if(!this.sessionService.isOnlineOnly()) {
```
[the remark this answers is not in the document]
latin-panda: The IDE settings for code style can help you with that

webapp/src/ts/services/user-contact.service.ts
```diff
@@ -33,4 +33,17 @@ export class UserContactService {
         throw err;
       });
   }
+
+  async getCurrentLineageLevel(){
```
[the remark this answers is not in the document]
latin-panda: +1

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
latin-panda: +1
every object in `conversations` to have set the right lineage value in different possible scenarios

latin-panda (changes requested):
Nice! 
I agree with Josh feedback and added a couple more things.

latin-panda:
I think you need to fix the failing e2e... it's probably not a flaky one 🤔

webapp/src/ts/modules/reports/reports.component.ts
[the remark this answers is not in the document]
elvisdorkenoo: Yeah I felt the filter a bit expensive, thanks for the suggestion, then I guess it would make this line finally looking like 
```
      if (this.currentLevel && report.lineage.length && report.lineage[length - 1] === this.currentLevel) {
        report.lineage.pop();
      }
```

webapp/src/ts/services/user-contact.service.ts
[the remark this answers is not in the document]
elvisdorkenoo: Yeah I agree, I unsuccessfully tried to find a better place for it. I would be fine duplicating the function, but as I will need to ask breadcrumbs to Tasks tab as well when working on #5710 , it would make it a tri-plication 🤔

webapp/src/ts/modules/reports/reports.component.ts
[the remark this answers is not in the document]
elvisdorkenoo: Yeah you're right, I also felt using the filter a bit expensive but during my tests the size of the lineage array is not always what we would expected : some lineages look like `["CHW Betty's Area", null, null, null, null, null]`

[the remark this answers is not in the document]
jkuester: Oh weird!  I spend a bunch of time digging and trying to understand all the code that loaded the lineage data and still I do not feel like I have a solid grasp on how that data is produced.  I do not anticipate these lineage arrays being very large one way or the other, so there is really no performance concern here. I am fine with using `filter` here if that seems safer than `pop`.

latin-panda:
@elvisdorkenoo I provided feedback for this code in [this other PR](https://github.com/medic/cht-core/pull/7757#pullrequestreview-1095913023)

latin-panda:
I think the PR for the [changes in tasks tab](https://github.com/medic/cht-core/pull/7757#pullrequestreview-1095913023) should only show changes related to tasks, but for some reason it shows changes related to this PR (messages tab, reports tab). 

Somehow it's mixed and that's why I got confused when reviewing. It's better to work 1 ticket at the time and getting it passed all the way to AT, it would help to focus better, avoid getting branches and conversations mixed.

In the current situation, you can submit one PR with everything 1. if it's confusing you too much, or 2. if it's too hard to keep them separate, that means changes for task only (https://github.com/medic/cht-core/issues/5710) in one PR, then changes of messages and reports in one PR (https://github.com/medic/cht-core/issues/5697).

What do you think?

elvisdorkenoo:
> I think the PR for the [changes in tasks tab](https://github.com/medic/cht-core/pull/7757#pullrequestreview-1095913023) should only show changes related to tasks, but for some reason it shows changes related to this PR (messages tab, reports tab).
> 
> Somehow it's mixed and that's why I got confused when reviewing. It's better to work 1 ticket at the time and getting it passed all the way to AT, it would help to focus better, avoid getting branches and conversations mixed.
> 
> In the current situation, you can submit one PR with everything if:
> 
> 1. if it's confusing you too much, or
> 2. If it's too hard to keep them separate, that means changes for task only ([Add "belongs to" breadcrumbs to tasks in the LHS list #5710](https://github.com/medic/cht-core/issues/5710)) in one PR, then changes of messages and reports in one PR ([Update the levels displayed in breadcrumbs across the app #5697](https://github.com/medic/cht-core/issues/5697)).
> 
> What do you think?

Yeah you are right it's a bit confusing... 
We are trying to deliver the two of them at the same time, and that's the source of the confusion.
What I did is the following : 
- apply your recommandations related to messages and reports in the branch of the current PR and push
- update task #7757 branch with this branch last changes
- apply your recommandations related to tasks in the branch updated branch of #7757 
- so you can see thoses changes when reviewing the last commit [here](https://github.com/medic/cht-core/pull/7757/commits/ad8a21eb50ccef20425be83a8155bb2da3b71313).
From there we can find a way to solve the confusion, let me know if you suggest anything else.
The final goal would be to merge #5710 into #5697

latin-panda (changes requested):
Thanks @elvisdorkenoo it's clearer now. 

I've added some feedback and some questions inline.

webapp/src/ts/modules/reports/reports.component.ts
```diff
@@ -178,21 +179,24 @@ export class ReportsComponent implements OnInit, AfterViewInit, OnDestroy {
...
+  prepareReports(reports) {
     return reports.map(report => {
       const form = _find(this.forms, { code: report.form });
       report.icon = form && form.icon;
       report.heading = this.getReportHeading(form, report);
       report.summary = form ? form.title : report.form;
       report.lineage = report.subject && report.subject.lineage || report.lineage;
+      // remove the lineage level that belongs to the offline logged-in user, normally the last one
+      if (this.currentLevel) {
+        report.lineage = report.lineage.filter(level => level !== this.currentLevel);
```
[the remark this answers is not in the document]
elvisdorkenoo: From my investigation it can only be in the middle of lineage, when the user is online only, and this case, we decide to show the whole lineage without altering it

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
```diff
@@ -172,6 +198,126 @@ describe('Messages Component', () => {
...
+          lineage : [ 'Amy Johnsons Household', 'St Elmos Concession', 'Chattanooga Village', 'CHW Bettys Area']
+        },
+        { key: 'b', message: { inAllMessages: true },
+          lineage : [ 'Amy Johnsons Household', 'St Elmos Concession', 'Chattanooga Village']
+        },
+        { key: 'c', message: { inAllMessages: true },
+          lineage : [ 'Amy Johnsons Household', 'St Elmos Concession', 'Chattanooga Village', 'Ramdom Place']
+        },
+        { key: 'd', message: { inAllMessages: true },
+          lineage : [ 'Amy Johnsons Household', 'St Elmos Concession', 'Chattanooga Village', 'CHW Bettys Area']
```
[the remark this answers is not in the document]
elvisdorkenoo: As I understood, lineage calculation stops at your place level, and if you are online only i.e. you don't belong to any place, 'CHW Bettys Area' can be in the middle, but in that case, it's irrelevant and we don't update the lineage.

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
```diff
@@ -15,6 +15,8 @@ import { ModalService } from '@mm-modals/mm-modal/mm-modal';
 import { NavigationComponent } from '@mm-components/navigation/navigation.component';
 import { TourService } from '@mm-services/tour.service';
 import { NavigationService } from '@mm-services/navigation.service';
+import { UserContactService } from '@mm-services/user-contact.service';
+import { SessionService } from '@mm-services/session.service';
```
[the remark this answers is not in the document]
elvisdorkenoo: Yeah I was not done differently in there, so I felt that I should follow, but you are right, I'll update.

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
```diff
@@ -172,6 +198,126 @@ describe('Messages Component', () => {
...
+      'if the connected user is offline and belongs to a place of the conversation lineage', fakeAsync( () => {
+      const offlineUserContactDoc2 = {
+        _id: 'user',
+        parent: {
+          _id: 'parent',
+          name: 'CHW Bettys Area',
+          parent: userContactGrandparent,
+        },
+      };
+      const conversations = [
```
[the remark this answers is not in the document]
elvisdorkenoo: got it

webapp/tests/karma/ts/modules/reports/reports.component.spec.ts
```diff
@@ -263,4 +278,141 @@ describe('Reports Component', () => {
     });
   });
 
+  describe('lineage updates', () => {
```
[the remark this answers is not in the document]
elvisdorkenoo: sure

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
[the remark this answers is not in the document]
elvisdorkenoo: got it

webapp/tests/karma/ts/modules/reports/reports.component.spec.ts
```diff
@@ -263,4 +278,156 @@ describe('Reports Component', () => {
...
+          _id: 'ee21ea15-1ebb-4d6d-95ea-7073ba965525',
+          lineage: undefined,
+          heading: 'report.subject.unknown',
+          icon: undefined,
+          summary: undefined,
+          unread: true,
+
+        },
+      ];
+      userContactService.get.resolves(userContactDoc);
```
jkuester, line 359: Maybe I am missing something here, but don't we need to use a user whose parent is actually contained in the lineage?  (So that we can verify the lineage was not actually edited because the user is online?)

jkuester (changes requested):

webapp/src/ts/modules/messages/messages.component.ts
```diff
@@ -48,6 +53,11 @@ export class MessagesComponent implements OnInit, OnDestroy {
   ngOnInit(): void {
     this.subscribeToStore();
     this.tourService.startIfNeeded(this.route.snapshot);
+    if (!this.sessionService.isOnlineOnly()) {
+      this
+        .getCurrentLineageLevel()
+        .then(currentLevel => this.currentLevel = currentLevel);
```
jkuester, line 59: @latin-panda you have way more front-end experience then I do, so you might know if this is worth worrying about!  I got to thinking about how loading the `currentLevel` here is async and so technically there is nothing stopping the `updateConversations` code below running before we actually get the `currentLevel` set.  

Would it be better to just assign `currentLevel` to a promise that we have to wait for down in `updateConversations`?  e.g this code would become something like `this.currentLevel = this.getCurrentLineageLevel();`  Then down in  `updateConversations` we do something like `Promise.all([this.messageContactService.getList(), this.currentLevel]).then(....`

webapp/src/ts/modules/reports/reports.component.ts
```diff
@@ -178,21 +179,24 @@ export class ReportsComponent implements OnInit, AfterViewInit, OnDestroy {
...
+  prepareReports(reports) {
     return reports.map(report => {
       const form = _find(this.forms, { code: report.form });
       report.icon = form && form.icon;
       report.heading = this.getReportHeading(form, report);
       report.summary = form ? form.title : report.form;
       report.lineage = report.subject && report.subject.lineage || report.lineage;
+      // remove the lineage level that belongs to the offline logged-in user
+      if (this.currentLevel && report.lineage && report.lineage.length) {
+        report.lineage = report.lineage?.filter(level => level !== this.currentLevel);
```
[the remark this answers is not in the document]
elvisdorkenoo: Yeah you are right it seems unnecessary especially the `report.lineage?.filter`.
I added the `&& report.lineage.length` because I didn't want to through some useless filter if the lineage comes out as empty array as supposed in my tests data.

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
```diff
@@ -33,12 +47,25 @@ describe('Messages Component', () => {
...
+      {
+        selector: 'getConversations',
+        value: [
+          { 
+            key: 'a', 
+            message: { inAllMessages: true },
+            lineage: [ 'Amy Johnsons Household', 'St Elmos Concession', 'Chattanooga Village', 'CHW Bettys Area' ]
+          },
+        ],
+      },
```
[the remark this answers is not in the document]
elvisdorkenoo: Yeah you are right, sorry

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
```diff
@@ -172,13 +201,128 @@ describe('Messages Component', () => {
...
+    };
+    const offlineUserContactDoc2 = {
+      _id: 'user',
+      parent: {
+        _id: 'parent',
+        name: 'CHW Bettys Area',
+        parent: userContactGrandparent,
+      },
+    };
+    const conversations =  [
```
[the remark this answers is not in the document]
elvisdorkenoo: yeah thanks to @latin-panda to suggest adding theres scenarios 🙂

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
```diff
@@ -172,13 +201,128 @@ describe('Messages Component', () => {
...
+        expect(messageContactService.getList.callCount).to.equal(1);
+        expect(component.currentLevel).to.equal('parent');
+        expect(component.conversations).to.deep.equal(conversations);
+      }));
+
+    it('should not change the conversations lineage if the connected user is online only', fakeAsync( () => {
+      sinon.resetHistory();
+
+      messageContactService.getList.resolves(conversations);
+      userContactService.get.resolves(userContactDoc);
```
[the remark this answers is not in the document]
elvisdorkenoo: Yeah we can use that scenario as well but with `isOnlineOnly.returns(true)`  it's not  necessary to call `userContactService.get.resolves(userContactDoc)` and  `expect(component.currentLevel).to.equal('CHW Bettys Area')` because it's irrelevant as `component.currentLevel` is only set when user is offline. 
The current value of `component.currentLevel ` only comes from the constructor in ngOnInit and not from `userContactService.get.resolves(userContactDoc)` consequences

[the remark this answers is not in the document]
elvisdorkenoo: Yeah we can use that scenario as well but with `isOnlineOnly.returns(true)`  it's not  necessary to call `userContactService.get.resolves(userContactDoc)` and  `expect(component.currentLevel).to.equal('CHW Bettys Area')` because it's irrelevant as `component.currentLevel` is only set when user is offline. 
The current value of `component.currentLevel ` only comes from the constructor in ngOnInit and not from `userContactService.get.resolves(userContactDoc)` consequences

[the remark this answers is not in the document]
elvisdorkenoo: so if you are confortable, we can do something like this :

```
    it('should not change the conversations lineage if the connected user is online only', fakeAsync( () => {
      sinon.resetHistory();

      messageContactService.getList.resolves(conversations);
      userContactService.get.resolves(bettyOfflineUserContactDoc);
      sessionService.isOnlineOnly.returns(true);

      component.ngOnInit();
      tick();
      component.updateConversations({merge : true});
      tick();

      expect(component.conversations).to.deep.equal(conversations);
    }));
```

webapp/tests/karma/ts/modules/reports/reports.component.spec.ts
elvisdorkenoo: As I was explaining [here](https://github.com/medic/cht-core/pull/7732#discussion_r964392778). 
When the user is online only, component.currentLevel is not even set in the component, so calling component.currentLevel = await component.getCurrentLineageLevel() is not necessary.

Would you be confortable with something like this ? :
    it('should not change the reports lineage if user is online only', async () => {
      const expectedReports = [
        {
          _id: '88b0dfff-4a82-4202-abea-d0cabe5aa9bd',
          lineage: [ 'St Elmos Concession', 'Chattanooga Village', 'CHW Bettys Area' ],
          heading: 'report.subject.unknown',
          icon: undefined,
          summary: undefined,
          unread: true,
        },
        {
          _id: 'a86f238a-ad81-4780-9552-c7248864d1b2',
          lineage:  [ 'Chattanooga Village', 'CHW Bettys Area' ],
          heading: 'report.subject.unknown',
          icon: undefined,
          summary: undefined,
          unread: true,
        },
        {
          _id: 'd2da792d-e7f1-48b3-8e53-61d331d7e899',
          lineage: [ 'Chattanooga Village' ],
          heading: 'report.subject.unknown',
          icon: undefined,
          summary: undefined,
          unread: true,
        },
        {
          _id: 'ee21ea15-1ebb-4d6d-95ea-7073ba357229',
          lineage: [ 'CHW Bettys Area' ],
          heading: 'report.subject.unknown',
          icon: undefined,
          summary: undefined,
          unread: true,
        },
        {
          _id: 'ee21ea15-1ebb-4d6d-95ea-7073ba357229',
          lineage: [],
          heading: 'report.subject.unknown',
          icon: undefined,
          summary: undefined,
          unread: true,
        },
        {
          _id: 'ee21ea15-1ebb-4d6d-95ea-7073ba965525',
          lineage: undefined,
          heading: 'report.subject.unknown',
          icon: undefined,
          summary: undefined,
          unread: true,

        },
      ];
      // userContactService.get.resolves(userContactDoc);
      // sessionService.isOnlineOnly.resolves(true);

      component.currentLevel = undefined;    //await component.getCurrentLineageLevel();

      const updatedReports = component.prepareReports(reports);

      // expect(component.currentLevel).to.equal('parent');

      expect(updatedReports).to.deep.equal(expectedReports);
    });

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
```diff
@@ -23,6 +25,18 @@ describe('Messages Component', () => {
   let changesService;
   let exportService;
   let modalService;
+  let userContactService;
+  let sessionService;
+
+  const userContactGrandparent = { _id: 'grandparent' };
```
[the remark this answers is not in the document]
latin-panda: Ohh thanks! I thought it was written as 2 words

webapp/src/ts/modules/messages/messages.component.ts
```diff
@@ -152,6 +162,14 @@ export class MessagesComponent implements OnInit, OnDestroy {
     return this.messageContactService
       .getList()
       .then((conversations = []) => {
+        // remove the lineage level that belongs to the offline logged in user
+        if (this.currentLevel) {
+          conversations.forEach(conversation => {
+            if (conversation.lineage) {
+              conversation.lineage = conversation.lineage.filter(level => level !== this.currentLevel);
```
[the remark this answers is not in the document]
latin-panda: CHT allows creating multiple places with the exact same name. It's probably not common but it can happen. 
@elvisdorkenoo do we have access to the place's `_id`?

webapp/src/ts/modules/messages/messages.component.ts
latin-panda: Yes thank you, that's possible, I was considering something similar yesterday, but your suggestion about assigning `currentLevel` to a promise is simpler and what we usually do. 

@elvisdorkenoo is it making sense?

webapp/src/ts/modules/messages/messages.component.ts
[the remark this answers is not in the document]
elvisdorkenoo: When calculating the lineage for `tasks` whose breadcrumbs are pretty news, yes. For `messages and reports`, the lineage is just an array containing the name of the places, but there's always a way to get the `_id` . 
We can also come back to the solution to **simply remove the last element**, the only problem is that the lineage array sometimes come out with some` _null items as last elements,_` so we can **filter them out, and then remove the last element of the lineage array**.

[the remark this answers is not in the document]
jkuester: IMHO the safest/simplest thing to do would be to filter the null values like you suggested and then remove the last element _if it matches the `currentLevel`_.  If it does not match, I think we should do nothing.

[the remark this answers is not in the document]
elvisdorkenoo: yeah agree

webapp/src/ts/modules/messages/messages.component.ts
elvisdorkenoo: yeah totally makes sense, thanks.

webapp/src/css/common.less
```diff
@@ -76,6 +76,10 @@ ul {
 
 .lineage {
   margin-bottom: 0;
+  white-space: normal !important;
```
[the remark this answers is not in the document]
elvisdorkenoo: yeah done  !

latin-panda:
@elvisdorkenoo I've fixed the unit tests please have a look at [the commit](https://github.com/medic/cht-core/pull/7732/commits/03bd4166f1dd44fbab254d24606c16d130f3556d) to see how it was fixed. 

TL;DR: wait for all promises to resolve before asserting and make sure the stub is configured according to the function's return type. 
- returns promise? -> then use `sinon.stub().resolves(...)` 
- returns primitive value? --> then use `sinon.stub().returns(...)`)

webapp/src/ts/modules/reports/reports.component.ts
```diff
@@ -178,23 +177,34 @@ export class ReportsComponent implements OnInit, AfterViewInit, OnDestroy {
...
+            report.lineage = report.lineage.filter(level => level);
+            if(report.lineage[report.lineage.length-1] === currentLevel){
+              report.lineage.pop();
+            }
+          }
+          report.unread = !report.read;
+          return report;
+        });
+      }
+    );
```
latin-panda, line 206: it's always always better to keep code flat and check for null, this version adds validation when getting the last element of the array

```suggestion
  const userLineageLevel = await this.currentLevel;
  
  return reports.map(report => {
  const form = _find(this.forms, { code: report.form });
  report.icon = form && form.icon;
  report.heading = this.getReportHeading(form, report);
  report.summary = form ? form.title : report.form;
  report.lineage = report.subject && report.subject.lineage || report.lineage;
  report.unread = !report.read;
          
  // remove the lineage level that belongs to the offline logged-in user
  if (currentLevel && report?.lineage?.length) {
    report.lineage = report.lineage.filter(level => level);
    const item = report.lineage?.length && report.lineage[report.lineage.length -1]
    if (item === userLineageLevel) {
      report.lineage.pop();
    }
  }
  
  return report;
 );
```

webapp/src/ts/modules/reports/reports.component.ts
```diff
@@ -222,10 +232,10 @@ export class ReportsComponent implements OnInit, AfterViewInit, OnDestroy {
     return this.searchService
       .search('reports', this.filters, options)
       .then((reports) => this.addReadStatusService.updateReports(reports))
-      .then((updatedReports) => {
-        updatedReports = this.prepareReports(updatedReports);
-
-        this.reportsActions.updateReportsList(updatedReports);
+      .then(updatedReports => {
```
latin-panda, line 235: Look how we're chaining promises in this function, and please think about the advantages of doing this, this is to run asynchronous code in a sequence (sync way).
You need to do the same with `prepareReports(updatedReports)` to keep data safe. I highly recommend to read and more importantly experiment with promises and asynchronous code because JavaScript in that regards won't warn you about undesired behaviours in code.

```
.then(reports => this.addReadStatusService.updateReports(reports))
.then(reports => this.prepareReports(reports))
.then(reports => {
  const updatedReports = this.reportsActions.updateReportsList(updatedReports);
  this.moreItems = updatedReports.length >= options.limit;
  ....
  ...
  ..
```

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
```diff
@@ -172,13 +191,117 @@ describe('Messages Component', () => {
       expect(component.loading).to.be.false;
       expect(component.conversations).to.eql( newConversations);
     });
+
+    it('ngOnDestroy() should unsubscribe from observables', () => {
```
latin-panda, line 195: This test isn't related to `updateConversations()`, it stays in "root" level

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
```diff
@@ -172,13 +191,117 @@ describe('Messages Component', () => {
...
+      { key: 'e',
+        message: { inAllMessages: true },
+      },
+    ];
+
+    it('it should retrieve the hierarchy level of the connected user', async () => {
+      expect(await component.currentLevel).to.equal('parent');
+    });
+
+    it('should not change the conversations lineage if the connected user is online only', fakeAsync( () => {
```
latin-panda, line 239:
```suggestion
    it('should not change the conversations lineage if the connected user is online only', fakeAsync(() => {
```
Same with the other occurrences of `fakeAsync`

webapp/src/ts/modules/messages/messages.component.ts
```diff
@@ -148,10 +156,21 @@ export class MessagesComponent implements OnInit, OnDestroy {
...
+        if (currentLevel) {
+          conversations?.forEach(conversation => {
+            if (conversation.lineage && conversation.lineage.length) {
+              conversation.lineage = conversation.lineage.filter(level => level);
+              if(conversation.lineage[conversation.lineage.length-1] === currentLevel){
+                conversation.lineage.pop();
+              }
+            }
+          });
+        }
```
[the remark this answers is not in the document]
elvisdorkenoo: the issue here is that I try to avoid to have `conversation.lineage = undefined ` in the output, when lineage data is missing from conversation because it's messing previous unit tests up.

webapp/src/ts/modules/reports/reports.component.ts
elvisdorkenoo: well noted.

webapp/src/ts/modules/messages/messages.component.ts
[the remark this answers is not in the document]
latin-panda: @elvisdorkenoo  you can exit early
```
if (conversation.lineage?.length) {
  return;
}
conversation.lineage = conversation.lineage?.filter(level => level);
.... continue with the code ....
```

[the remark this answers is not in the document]
elvisdorkenoo: yeah tha's flatter, thanks.

elvisdorkenoo:
One question, do we change the target branch of the PR from master to 3.17.x  ? @jkuester @latin-panda

webapp/src/ts/modules/reports/reports.component.ts
```diff
@@ -193,6 +194,15 @@ export class ReportsComponent implements OnInit, AfterViewInit, OnDestroy {
       report.lineage = report.subject && report.subject.lineage || report.lineage;
       report.unread = !report.read;
 
+      // remove the lineage level that belongs to the offline logged-in user
+      if (userLineageLevel && report?.lineage?.length) {
+        report.lineage = report.lineage.filter(level => level);
+        const item = report.lineage?.length && report.lineage[report.lineage.length -1];
```
jkuester, line 200:
```suggestion
        const item = report.lineage[report.lineage.length -1];
```

Minor, but don't need to check the lineage length here again.

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
```diff
@@ -174,11 +201,107 @@ describe('Messages Component', () => {
...
+
+      component.ngOnInit();
+      tick();
+      component.updateConversations({merge : true});
+      tick();
+
+      expect(component.conversations).to.deep.equal(conversations);
+    }));
+
+    it('should not change the conversations lineage if the connected user is online only', fakeAsync(() => {
```
jkuester, line 254: Is this test just exactly the same as the one before?

jkuester (approved):
Couple minor suggestions, but otherwise I think this is pretty much ready to go!

jkuester:
@elvisdorkenoo regarding the target branch for this PR, I think we should leave it as `master`.  Once this per is Squashed and merged into `master`, we can cherrypick the commit into `3.17.x`.

jkuester (commented):
Also, wanted to bring this up for the sake of completeness (before we send this over to QA).  @elvisdorkenoo @latin-panda do you think we need any kind of e2e test for this?  

I have gone back and forth on this, but I am kinda leaning towards adding a test. This functionality depends on interacting with multiple services. The current unit tests have properly stubbed out the behavior of those services, but it seems like a good idea to have at least one e2e test that verifies the breadcrumbs are actually displayed as expected in the app when logged in as an offline user....

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
elvisdorkenoo: yeah thanks for noticing it, it's fixed.

elvisdorkenoo:
> Also, wanted to bring this up for the sake of completeness (before we send this over to QA). @elvisdorkenoo @latin-panda do you think we need any kind of e2e test for this?
> 
> I have gone back and forth on this, but I am kinda leaning towards adding a test. This functionality depends on interacting with multiple services. The current unit tests have properly stubbed out the behavior of those services, but it seems like a good idea to have at least one e2e test that verifies the breadcrumbs are actually displayed as expected in the app when logged in as an offline user....

I tried to take a look to the existing e2e tests to see if there is anything about breadcrumbs as in this PR we are only updating  the displayed levels, and for now, there are no e2e for breadcrumbs that I can find.

latin-panda:
@elvisdorkenoo we have been building the test coverage recently, so I think we still don't have an e2e test for breadcrumbs in specific. As per Gareth comments before, we are supposed to ship PRs with at least 1 e2e test, which I agree :) 
[Take a look at this file](https://github.com/medic/cht-core/blob/master/tests/e2e/default/reports/sidebar-filter.wdio-spec.js), it's a good sample of how to do an e2e for Reports tab, you just need a selector to check breadcrumbs texts.

Do you think you can complete 1 e2e today? You will have to pay extra details in how these are written to get less feedback as possible.

elvisdorkenoo:
> @elvisdorkenoo we have been building the test coverage recently, so I think we still don't have an e2e test for breadcrumbs in specific. As per Gareth comments before, we are supposed to ship PRs with at least 1 e2e test, which I agree :) [Take a look at this file](https://github.com/medic/cht-core/blob/master/tests/e2e/default/reports/sidebar-filter.wdio-spec.js), it's a good sample of how to do an e2e for Reports tab, you just need a selector to check breadcrumbs texts.
> 
> Do you think you can complete 1 e2e today? You will have to pay extra attention in how these are written to get less feedback as possible.

Yes, I will add the e2e test, thanks for the sample.

latin-panda:
@elvisdorkenoo how is the e2e going? Do you think we review today?

elvisdorkenoo:
> @elvisdorkenoo how is the e2e going? Do you think we review today?

Yes thanks for the help, I did a push with en e2e test for report, you can take a look @latin-panda and @jkuester

webapp/tests/karma/ts/modules/reports/reports.component.spec.ts
```diff
@@ -263,4 +279,168 @@ describe('Reports Component', () => {
...
+        },
+        {
+          _id: 'ee21ea15-1ebb-4d6d-95ea-7073ba357229',
+          lineage: [],
+          heading: 'report.subject.unknown',
+          icon: undefined,
+          summary: undefined,
+          unread: true,
+        },
+        {
```
latin-panda, line 418: You can delete this object since is the identical to the previous one.

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
```diff
@@ -174,11 +201,108 @@ describe('Messages Component', () => {
...
+      fakeAsync( () => {
+        sinon.resetHistory();
+
+        messageContactService.getList.resolves(conversations);
+        userContactService.get.resolves(userContactDoc);
+        sessionService.isOnlineOnly.returns(false);
+
+        component.ngOnInit();
+        tick();
+        component.updateConversations({merge : true});
```
latin-panda, line 249:
```suggestion
        component.updateConversations({ merge: true });
```
I think I made a comment about this before 🤔

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
```diff
@@ -174,11 +201,108 @@ describe('Messages Component', () => {
...
+    it('should not change the conversations lineage if the connected user is online only', fakeAsync(() => {
+      sinon.resetHistory();
+
+      messageContactService.getList.resolves(conversations);
+      userContactService.get.resolves(bettyOfflineUserContactDoc);
+      sessionService.isOnlineOnly.returns(true);
+
+      component.ngOnInit();
+      tick();
+      component.updateConversations({merge : true});
```
latin-panda, line 264:
```suggestion
      component.updateConversations({ merge: true });
```

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
```diff
@@ -174,11 +201,108 @@ describe('Messages Component', () => {
...
+    ];
 
-    component.ngOnDestroy();
+    it('it should retrieve the hierarchy level of the connected user', async () => {
+      expect(await component.currentLevel).to.equal('parent');
+    });
 
-    expect(spySubscriptionsUnsubscribe.callCount).to.equal(1);
+    it('should not alter conversations when user is offline and parent place is not relevant to the conversation',
+      fakeAsync( () => {
```
latin-panda, line 240:
```suggestion
      fakeAsync(() => {
```

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
```diff
@@ -174,11 +201,108 @@ describe('Messages Component', () => {
...
+          },
+        ];
+
+        messageContactService.getList.resolves(conversations);
+        userContactService.get.resolves(bettyOfflineUserContactDoc);
+        sessionService.isOnlineOnly.returns(false);
+
+        component.ngOnInit();
+        tick();
+        component.updateConversations({ merge : true });
```
latin-panda, line 301:
```suggestion
        component.updateConversations({ merge: true });
```

webapp/src/ts/modules/messages/messages.component.ts
```diff
@@ -148,10 +156,22 @@ export class MessagesComponent implements OnInit, OnDestroy {
...
+      .all([ this.messageContactService.getList(), this.currentLevel ])
+      .then(([ conversations, currentLevel ]) => {
+        // Remove the lineage level that belongs to the offline logged-in user.
+        if (currentLevel) {
+          conversations?.forEach(conversation => {
+            if (!conversation.lineage?.length) {
+              return;
+            }
+            conversation.lineage = conversation.lineage.filter(level => level);
+            if(conversation.lineage[conversation.lineage.length -1] === currentLevel){
```
latin-panda, line 170:
```suggestion
            if (conversation.lineage[conversation.lineage.length -1] === currentLevel) {
```

tests/e2e/default/reports/breadcrumbs.wdio-spec.js
```diff
@@ -0,0 +1,98 @@
...
+  });
+  const onlineUser = userFactory.build({
+    username: 'onlineuser',
+    roles: [ 'program_officer' ],
+    place: district_hospital._id,
+    contact: contact2._id,
+  });
+  const patient = personFactory.build({
+    _id: 'patient1',
+    parent: { _id: clinic._id, parent: { _id: health_center._id, parent: { _id: district_hospital._id }} }  });
```
latin-panda, line 53:
```suggestion
    parent: { _id: clinic._id, parent: { _id: health_center._id, parent: { _id: district_hospital._id }}}  
  });
```
This should close in the next line, because you broke the line in the opening bracket.

tests/e2e/default/reports/breadcrumbs.wdio-spec.js
```diff
@@ -0,0 +1,98 @@
...
+  const reports = [
+    reportFactory.build(
+      {
+        form: 'P',
+        reported_date: moment([ today.year(), today.month() - 4, 1, 23, 30 ]).valueOf(),
+        patient_id: 'patient1',
+      },
+      {
+        patient, submitter: offlineUser.contact, fields: { lmp_date: 'Feb 3, 2022', patient_id: 'patient1' },
+      }),
```
latin-panda, line 63: This should close in the next line, because you broke the line in the opening bracket. 
```suggestion
      },
    ),
```

latin-panda (approved):
Nice! I don't have further suggestions regarding logic. However I left some comments regarding format which should be quick to fix, please make sure that CI is fully green after you push the changes. 

@jkuester can you please have a look? Thanks for your time.

webapp/tests/karma/ts/modules/reports/reports.component.spec.ts
elvisdorkenoo: it's identical to the previous but I think it's normal, that is the `expectedReports` ie the output after the processing, the input is different.
```
          _id: 'ee21ea15-1ebb-4d6d-95ea-7073ba357229',
          lineage: [ 'CHW Bettys Area' ],
          heading: 'report.subject.unknown',
          icon: undefined,
          summary: undefined,
          unread: true,
        },
        {
          _id: 'ee21ea15-1ebb-4d6d-95ea-7073ba357229',
          lineage: [],
          heading: 'report.subject.unknown',
          icon: undefined,
          summary: undefined,
          unread: true,
        },
```
should become :
```
        {
          _id: 'ee21ea15-1ebb-4d6d-95ea-7073ba357229',
          lineage: [],
          heading: 'report.subject.unknown',
          icon: undefined,
          summary: undefined,
          unread: true,
        },
        {
          _id: 'ee21ea15-1ebb-4d6d-95ea-7073ba357229',
          heading: 'report.subject.unknown',
          icon: undefined,
          lineage: [],
          summary: undefined,
          unread: true,
        },
```
after the content of lineage is updated

webapp/tests/karma/ts/modules/messages/messages.component.spec.ts
elvisdorkenoo: yeah sorry did include that into my IDE and it reformatted these as it wants, fixed now. Thanks.

webapp/tests/karma/ts/modules/reports/reports.component.spec.ts
latin-panda: Ohhh got it, thanks!

latin-panda (approved):
LGTM! Thank you!

jkuester (approved):
Everything here looks good!  

Did we decide we did not need an e2e test for the message's breadcrumbs?

elvisdorkenoo:
> Everything here looks good!
> 
> Did we decide we did not need an e2e test for the message's breadcrumbs?

Thank you @jkuester, it's the same for the tasks in #7757 as well. I’ve created [this](https://github.com/medic/cht-core/issues/7806) ticket and I will be doing this next as a separate PR

Issue #6215, opened 2020-01-22
The `can_edit` permission doesn't stop you from editing documents (only uploading those edits)


SCdF:
The `can_edit` permission is used in two places in our code base: whether or not someone has permission on arbitrary writes (decided in api's routing), and whether or not upward sync should be enabled (in client side db-sync code).

This creates a confusing situation if `can_edit` is disabled for offline users. The user is still completely capable of creating new reports, editing existing reports, etc. However, *upwards replication simply never occurs*, with no indication that anything is wrong.

A `can_edit` permission check should wrap all write attempts (e.g. creating forms, editing forms etc) so disable the ability to perform those actions, and present a proper "read only" view.
 
Or, if that is too hard, we should not allow offline users to be configured without this permission.

Or, if _that_ is too hard, we should warn in the console on boot (ie in the db sync code that checks this permission) very loadly that this user is basically misconfigured, and it's going to look like they are successfully making changes when they are not.

SCdF:
It's also worth mentioning that there may be a perception that this is determining **just** edit capability (ie you can create but not edit). We would want to make sure our documentation is clear that edit means editing *and* creating.

dianabarsan:
I don't think there is a perception of `create` vs `edit` in this case. 
We do a similar check in `api` (here: https://github.com/medic/cht-core/blob/master/api/src/routing.js#L662) and filter _every_ put/post/delete through this. 
Historically, this was part of the auditing process: https://github.com/medic/cht-core/blob/2.18.x/api/audit-proxy.js#L37

latin-panda:
Similarly with the verification feature in Reports, there are `can_verify_reports ` and `can_edit_verification`. 

If `can_edit` is false but `can_verify_reports` and `can_edit_verification` are true, the online user can still verify and edit a verification of a report. 

I brought up the topic in today's stand up, we think for simplicity `can_edit` enable `create`, `update` and `delete` actions of any document in `medic` data base, when user is online or offline. 
That means the user can't verify or delete reports. The `can_verify_reports `, `can_edit_verification`, `can_delete_reports` and `can_bulk_delete_reports` won't be effective, if `can_edit` is **false**.

I'm going to fix the verify reports and the bulk delete reports to behave like this. 
cc: @garethbowen

garethbowen:
[...]

PR #7733, opened 2022-08-17
6215 can edit permission

latin-panda:
# Description

This PR:
- Adds checks to `can_edit` permission in every endpoint that writes in `medic` db. 
- Adds `mmAuth` directory in action buttons across the app to check `can_edit` permission and prevent creating docs in local since the user won't be able to sync those in `medic` db.
- Adds `AppRouteGuardProvider` in contacts, reports and tasks routings to check for the same permission and prevent navigation to pages that adds documents to `medic` db.

https://github.com/medic/cht-core/issues/6215

[CHT-Docs PR](https://github.com/medic/cht-docs/pull/752)

# Code review checklist

# License


webapp/src/ts/components/actionbar/actionbar.component.html
```diff
@@ -31,8 +31,13 @@
         class="general-actions left-pane"
         *ngIf="currentTab === 'reports' && !selectMode"
         [ngClass]="{ 'col-sm-3': sidebarFilter?.isOpen, 'col-sm-4': !sidebarFilter?.isOpen }">
-        <div class="actions dropup" dropdown [dropup]="true" [isDisabled]="!nonContactForms?.length">
-          <a class="mm-icon mm-icon-inverse mm-icon-caption dropdown-toggle" dropdownToggle dropdata-toggle="dropdown" [ngClass]="{'mm-icon-disabled': !nonContactForms?.length}">
+        <div
+          class="actions dropup"
+          dropdown [dropup]="true"
+          [isDisabled]="!nonContactForms?.length"
+          mmAuth [mmAuthAny]="['can_edit', 'can_export_all', 'can_export_messages']"
```
latin-panda, line 38: Once we implement the FAB, we'll have a better solution to declare and render action buttons, we'll have a better way to know if the container has any button visible and the html will be cleaner.

api/src/controllers/users.js
```diff
@@ -9,7 +9,7 @@ const bulkUploadLog = require('../services/bulk-upload-log');
 
 const hasFullPermission = req => {
   return auth
-    .check(req, 'can_update_users')
+    .check(req, ['can_edit', 'can_update_users'])
```
garethbowen, line 12: I think this should use `hasAllPermissions` not `check`. The difference is that `check` also does some work to see if you're allowed to work in a given part of the hierarchy, but you're not passing in the third parameter so that work is wasted.

```suggestion
    .getUserCtx(req)
    .then(userCtx => auth.hasAllPermissions(userCtx, ['can_edit', 'can_update_users']))
```

api/src/controllers/users.js
```diff
@@ -127,7 +127,7 @@ module.exports = {
   },
   create: (req, res) => {
     return auth
-      .check(req, 'can_create_users')
+      .check(req, ['can_edit', 'can_create_users'])
```
garethbowen, line 130: Change to `hasAllPermissions` (as above).

api/src/controllers/users.js
```diff
@@ -204,7 +204,7 @@ module.exports = {
   },
   delete: (req, res) => {
     auth
-      .check(req, 'can_delete_users')
+      .check(req, ['can_edit', 'can_delete_users'])
```
garethbowen, line 207: Change to `hasAllPermissions` (as above).

api/src/controllers/users.js
```diff
@@ -229,7 +229,7 @@ module.exports = {
   v2: {
     create: async (req, res) => {
       try {
-        await auth.check(req, 'can_create_users');
+        await auth.check(req, ['can_edit', 'can_create_users']);
```
garethbowen, line 232: Change to `hasAllPermissions` (as above).

api/src/routing.js
```diff
@@ -415,7 +414,7 @@ app.get('/api/v1/users-info', authorization.handleAuthErrors, authorization.getU
 
 app.postJson('/api/v1/places', function(req, res) {
   auth
-    .check(req, 'can_create_places')
+    .check(req, ['can_edit', 'can_create_places'])
```
garethbowen, line 417: Change to `hasAllPermissions` (as above).

api/src/routing.js
```diff
@@ -427,7 +426,7 @@ app.postJson('/api/v1/places', function(req, res) {
 
 app.postJson('/api/v1/places/:id', function(req, res) {
   auth
-    .check(req, 'can_update_places')
+    .check(req, ['can_edit', 'can_update_places'])
```
garethbowen, line 429: Change to `hasAllPermissions` (as above).

api/src/routing.js
```diff
@@ -441,7 +440,7 @@ app.postJson('/api/v1/places/:id', function(req, res) {
 
 app.postJson('/api/v1/people', function(req, res) {
   auth
-    .check(req, 'can_create_people')
+    .check(req, ['can_edit', 'can_create_people'])
```
garethbowen, line 443: Change to `hasAllPermissions` (as above).

webapp/src/ts/components/actionbar/actionbar.component.html
garethbowen: I don't think this behaves quite like you want, because the mmAuthAny directive will return true if the user has any of the permissions, so adding `can_edit` doesn't hide it.

I _think_ this may work..

```suggestion
          mmAuth="can_edit" [mmAuthAny]="['can_export_all', 'can_export_messages']"
```

webapp/src/ts/components/actionbar/actionbar.component.html
```diff
@@ -61,58 +66,68 @@
...
-        <div test-id='create_places' class="actions dropup" mmAuth [mmAuthAny]="[ actionBar?.left?.childPlaces && 'can_create_places' ]">
-          <a *ngIf="actionBar?.left?.childPlaces?.length > 1"
-             mmAuth="can_create_places"
-             class="mm-icon mm-icon-inverse mm-icon-caption dropdown-toggle"
-             data-toggle="dropdown"
-          >
-            <span class="fa fa-plus"></span>
-            <p>{{'Add place' | translate}}</p>
-          </a>
+        <div mmAuth [mmAuthAny]="[ 'can_edit', 'can_export_all', 'can_export_contacts' ]">
```
garethbowen, line 69: Same here.

webapp/src/ts/components/actionbar/actionbar.component.html
```diff
@@ -61,58 +66,68 @@
...
+              <p>{{'Export' | translate}}</p>
+            </a>
+          </div>
         </div>
       </div>
 
       <div class="col-sm-4 general-actions left-pane" *ngIf="currentTab === 'messages'">
-        <div class="actions dropup">
-          <a class="mm-icon mm-icon-inverse mm-icon-caption send-message" *ngIf="actionBar?.left?.openSendMessageModal" (click)="actionBar?.left?.openSendMessageModal($event)">
+        <div class="actions dropup" mmAuth [mmAuthAny]="['can_edit', 'can_export_all', 'can_export_messages']">
```
garethbowen, line 124: Same here.

webapp/src/ts/components/actionbar/actionbar.component.html
```diff
@@ -187,8 +202,12 @@
       </div>
 
       <div class="col-sm-8 detail-actions right-pane" *ngIf="currentTab === 'contacts'">
-        <div class="actions dropup" *ngIf="showActionBar && selectedContactDoc && !loadingContent && snapshotData?.name === 'contacts.detail'">
-          <span>
+        <div
+          mmAuth [mmAuthAny]="['can_edit', 'can_view_call_action', 'can_view_message_action']"
```
garethbowen, line 206: Same here.

webapp/src/ts/components/actionbar/actionbar.component.html
```diff
@@ -237,46 +256,52 @@
...
             <p>{{'call' | translate}}</p>
           </a>
-          <a *ngIf="actionBar?.right?.sendTo"
-             mmAuth="can_view_message_action"
-             class="mm-icon mm-icon-inverse mm-icon-caption mobile-only"
-             [attr.href]="'sms:' + actionBar?.right?.sendTo?.phone | safeHtml:'url'"
-             [ngClass]="{ 'mm-icon-disabled': !actionBar?.right?.sendTo?.phone }"
+          <a
+            *ngIf="actionBar?.right?.sendTo"
+            mmAuth="can_view_message_action"
```
garethbowen, line 271: I think this should be "can_edit" too?

```suggestion
            mmAuth="can_edit,can_view_message_action"
```

garethbowen (changes requested):
Very nice! Thanks for going the extra mile on this.

I've requested a couple of changes in a handful of places, but overall this is excellent.

webapp/src/ts/components/actionbar/actionbar.component.html
latin-panda: The export permissions aren't modifying the `medic` db, I've tested with online/offline users and they can export successfully. 
I think export permissions aren't depending on `can_edit`

webapp/src/ts/components/actionbar/actionbar.component.html
latin-panda: Phone calls and sms messages in mobile open the phone and sms apps in the phone. It's not modifying the `medic` db. I think these 2 aren't strictly depending on `can_edit`

webapp/src/ts/components/actionbar/actionbar.component.html
latin-panda: This action is for mobile device, it opens a sms app using the link `sms:+12345667854`, it's not writing the data base. We discussed this in the stand up last week, even if we block it, ppl can always sms someone directly using their default sms app.

webapp/src/ts/components/actionbar/actionbar.component.html
latin-panda: Export might not be dependent on `can_edit`

webapp/src/ts/components/actionbar/actionbar.component.html
latin-panda: Thanks. Export might not be dependent on `can_edit`

api/src/controllers/users.js
latin-panda: Yeah the name is a bit generic and looks like this is the preferred function around for some reason.

I checked the usages of `check` function and we're not using the third parameter, maybe it was used in an old code that doesn't exists anymore and we forgot to clean this function. What I like about this function is that it's already extracting the `userCtx` before checking for permissions. 

I've refactored the function to:
- Keep extracting the `userCtx`
- Keep checking for permissions
- Stop checking if user is admin, because this is already done by `hasAllPermissions`
- Stop checking for the facility id, because it's not currently used and we can always add it later if we need it again.
- Return the  `userCtx` if user has all permissions, basically we know what to expect since we talk in `userCtx` schema/definition "terms". Then `userCtx.name` and `userCtx.roles` is available.

webapp/src/ts/components/actionbar/actionbar.component.html
latin-panda: There are other conditions we should consider when hiding the containers of action buttons, if we want to make it perfect for every corner case. 
For example: if it's mobile, if it's online user, if it has the data the action needs (phone number), etc...

We can do this a lot easier once we do the FAB work with a better design to declare and render action buttons, the container should easily know if it's displaying actions buttons. 

So we can:
1. Leave it like this a bit imperfect
2. Let the container display always, regardless if it has buttons or not. 

Screen Shot 2022-08-24 at 3 50 59 pm

api/src/controllers/users.js
latin-panda: `check` was refactored

api/src/controllers/users.js
latin-panda: `check` was refactored

api/src/controllers/users.js
latin-panda: `check` was refactored

api/src/routing.js
latin-panda: `check` was refactored

api/src/routing.js
latin-panda: `check` was refactored

api/src/routing.js
latin-panda: `check` was refactored

api/src/controllers/export-data.js
```diff
@@ -93,29 +93,24 @@ module.exports = {
     //    by the following auth check in ctx.district (maybe?)
     //  - Still don't let offline users use this API, and instead refactor the
     //    export logic so it can be used in webapp, and have exports works offline
-    return auth.getUserCtx(req)
+    return auth
+      .check(req, ['can_export_all', ...service.permission(type)])
```
garethbowen, line 97: You've changed permissions - it used to be `can_export_all` OR can export the specific type, and you've made it AND.

Either create a new `hasAnyPermission` function, or go back to the previous way of checking for `_all` and then if that fails, checking for the specific permission.

webapp/src/ts/components/actionbar/actionbar.component.html
garethbowen: > Let the container display always, regardless if it has buttons or not.

I think that's fine for now. You're right, it's complicated to work out whether the bar should show or not. It's a bit of an edge case to have no buttons at all, so let's leave this until we do the FAB redesign.

webapp/src/ts/components/actionbar/actionbar.component.html
garethbowen: Ahh right, sorry! I thought this was a CHT message, not an SMS message button!

webapp/src/ts/components/actionbar/actionbar.component.html
garethbowen: Ahh I see. It's hard to see what's going on in this file. Needs a serious refactor, probably around the same time as FABs.

garethbowen (changes requested):
Nice one! Great work on the refactoring and removal of dead code.

I think you've changed permissions in the export API but other than that looks good.

Sorry about the confusion about the UI. It's very difficult to read with the deeply nested templates. I think you've done enough refactoring for one PR so leave it for now, but we should definitely simplify the templates when we do the FABs.

api/src/controllers/export-data.js
latin-panda: 😩 I my eyes definitely skipped the exclamation mark in this `if`. Thanks, it's fixed

garethbowen (approved):
Nice one. Thanks for this!

PR #7739, opened 2022-08-19
Update default/standard config forms to pass validation

### LINKED ISSUE #7738: Update default/standard config forms so they pass validation
After the [recent updates](https://github.com/medic/cht-conf/issues/462) to cht-conf to validate forms, some of the sample forms provided with cht-core are now failing validation. They should be updated to conform to the latest standards.

jkuester:
# Description

Updates the default/standard config forms so that they pass the new cht-conf validation checks.  The two main issues that were fixed are:

- Invalid XPaths. The explicit XPaths in form expressions are now checked to be sure they point to an actual field in the form. There were a number of forms that had invalid XPaths caused by things like:
    - Typos in the field name
    - References to input fields that were not actually listed in the form
    - References to fields at the wrong point in the hierarchy
- Using the, now deprecated, `horizontal` appearance. This now triggers a warning that you should use `columns`. There are drop-in `column` replacement values for the `horizontal` appearances we were using, so I have just updated the forms to use those.

Where it was possible, I have tried to include tests that covered the fields with updated expressions. Existing tests for the `standard` config tasks/targets depend on a crazy old version of `medic-conf-test-harness` and there does not appear to be a clean upgrade path without significantly refactoring these tests.  I have logged https://github.com/medic/cht-core/issues/7744 to solve this later.  In the meantime, I did not want to contribute to the problem by adding more tests dependent on the old util.  So, I have also included the latest version of `cht-conf-test-harness` and written my new tests against that util.

Closes #7738

# Code review checklist

# License


config/default/test/contacts.js
```diff
@@ -24,6 +24,7 @@ module.exports = {
     },
 
     'type': 'person',
+    'created_by_doc': 'fake_delivery_report_uuid',
```
jkuester, line 27: Using this in the [`pnc_danger_sign_baby` task tests](https://github.com/medic/cht-core/pull/7739/files#diff-0931f83e17f49e76f2ced633d5bdcf8fc8b63f4cf98c7c2da900706e86d7c609) to verify that `__delivery_uuid` field is properly set from the task inputs.

jkuester:
@latin-panda could you have a look at this when you get the chance?  Do not be too worried by the large file count!  Most of the changes are minor and I have made an effort to make sure the `xml` file changes only include the _actual_ changes (hopefully that makes it easier to see what happened).  I have also done my best to limit the added tests to just the tests necessary to cover the affected functionality.

latin-panda (approved):
Sorry this review took so long, somehow it escaped from my attention. I try to review PRs within 1 day, please feel free to buzz me if I'm taking too long to comeback with a review. 

It looks good to me!

PR #7745, opened 2022-08-24
Architecture V3 - Add Nginx

henokgetachew:
# Description

Adds Nginx to Architecture V3 along with unit tests.

medic/cht-core#7686

Items for a future PR: 
 * Running the tests using GitHub Actions
 * Adding integration/e2e tests

# Code review checklist

# License


nginx/nginx.conf
```diff
@@ -0,0 +1,80 @@
...
+            proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
+            proxy_set_header        X-Forwarded-Proto https;
+            proxy_set_header        X-Forwarded-Host $server_name;
+        }
+
+        error_page   500 502 503 504  /50x.html;
+
+        location = /50x.html {
+            root html;
+        }
```
garethbowen, line 75: As far as I can see this file does not exist. I think this will cause an additional error to be thrown. Might be better to just remove this until we want to provide custom error pages.

henokgetachew: Removed.

garethbowen (approved):
Looks good to me!

scripts/build/cht-core.yml.template
```diff
@@ -41,6 +41,35 @@ services:
...
       - {{ network }}
 
+  cht-nginx:
+    image: {{{ repo }}}/cht-nginx:{{ tag }}
+    container_name: {{ nginx_container_name }}
+    depends_on:
+      - cht-api
+      - haproxy
+    ports:
+      - "${NGINX_HTTP_PORT:-80}:80"
```
dianabarsan, line 51: I think that, right now, API also exposes its port (5988) and all of our e2e tests pass requests directly to API instead of nginx. Can we run e2e tests and only connect to nginx?

garethbowen: I think so. If we do that we should also remove direct access to API (`expose` not `ports`). Do you think it's worth doing in this PR?

dianabarsan: Not necessarily in this PR, but I think in the final version, we should run all tests through nginx.

garethbowen: Ok. If you're happy merging this as is, I'll get started on the branch to run e2e tests through nginx and block direct API access.

PR #7748, opened 2022-08-25
Update all image versions to ubuntu-22.04

jkuester:
# Description

[description]

medic/cht-core#[number]

# Code review checklist

# License

### DIFF
--- a/admin/tests/unit/controllers/forms-xml.js
+++ b/admin/tests/unit/controllers/forms-xml.js
@@ -177,7 +177,9 @@ describe('FormsXmlCtrl controller', () => {
         .upload()
         .then(() => digest())
         .then(() => {
-          expectStatusError('Upload failed: Unexpected token o in JSON at position 1');
+          expectStatusError(
+            'Upload failed: Unexpected token &#39;o&#39;, &quot;not a valid&quot;... is not valid JSON'
+          );
         });
     });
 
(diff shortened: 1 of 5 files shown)


admin/tests/unit/controllers/forms-xml.js
```diff
@@ -177,7 +177,9 @@ describe('FormsXmlCtrl controller', () => {
         .upload()
         .then(() => digest())
         .then(() => {
-          expectStatusError('Upload failed: Unexpected token o in JSON at position 1');
+          expectStatusError(
+            'Upload failed: Unexpected token &#39;o&#39;, &quot;not a valid&quot;... is not valid JSON'
```
jkuester, line 181: This error message is bubbling up from the system level.  With the old message, this test actually failed for me locally (running 20.04).  The new version of this message causes the test to pass both locally and on the CI run (with 22.04).

config/default/package.json
```diff
@@ -21,7 +21,7 @@
     "chai-things": "^0.2.0",
     "eslint-plugin-json": "^2.0.1",
     "cht-conf": "^3.7.0",
-    "medic-conf-test-harness": "^2.1.3",
+    "cht-conf-test-harness": "^2.4.1",
```
jkuester, line 24: For some reason, the old version of the test harness refused to start on 22.04.  Not sure if there was something weird with Chrome/Puppeteer or what.  Updating to the newer version of cht-conf-test-harness needed to be done anyway and it solved the issue so the tests run fine on 22.04.

admin/tests/unit/controllers/forms-xml.js
m5r: I strangely had the same behavior locally, nice catch!

m5r (approved):
LGTM

PR #7751, opened 2022-08-29
Automate pregnancy workflow on standard config

tatilepizs:
Add the e2e automated test for the pregnancy workflow on Standard config. To run it use the command npm run standard-wdio-local

This test is divided into three different files:

#### Pregnancy
* Before the tests are executed:
  *  Creates two women, one using the webapp and one using SMS.
* Then this test:
   * Submit a new pregnancy using webapp
   * Verify the report created.
   * Submit a new pregnancy using SMS.
   * Verify the report created.
   * Verify the targets page.

#### Pregnancy Visit
* Before the tests are executed:
  * Submit a new pregnancy, for a woman generated using the personFactory file, using SMS.
* Then this test:
   * Submit a new pregnancy visit using webapp.
   * Verify the report created.
   * Submit a new pregnancy visit using SMS.
   * Verify the report created.

#### Delivery
* Before the tests are executed:
  * Creates two women using SMS.
  * Creates new pregnancies for the women created using SMS.
  * Creates pregnancy visits for the two pregnant women (four for the first woman and one for the second woman).
* Then this test:
   * Submit a delivery using webapp.
   * Verify the report created.
   * Submit a delivery using SMS.
   * Verify the report created.
   * Verify the targets page.

cc: @jonathanvq 

Ticket: medic/cht-release-testing#97

# Code review checklist

# License


tatilepizs:
@jkuester, @ngaruko could you please help me with the code review?

tests/page-objects/standard/forms/pregnancy-visit.wdio.page.js
```diff
@@ -0,0 +1,33 @@
...
+  await note.waitForDisplayed();
+  await note.setValue(text);
+};
+
+const getFollowUpSMS = async () => {
+  const sms = await followUpSMS();
+  await sms.waitForDisplayed();
+  return sms.getText();
+};
+
```
jkuester, line 26:
```suggestion
```

tests/page-objects/standard/forms/delivery.wdio.page.js
```diff
@@ -0,0 +1,74 @@
+const OUTCOME = {liveBirth: 'healthy', stillBirth: 'still_birth', miscarriage: 'miscarriage'};
+const LOCATION = {facility: 'f', homeAttendant: 's', homeNoAttendant: 'ns'};
```
jkuester, line 2: +1 for this pattern!  It really helps clarify what is actually going on.

tests/page-objects/standard/forms/delivery.wdio.page.js
```diff
@@ -0,0 +1,74 @@
...
+const outcomeSummary = () => $(`${FORM} ` + 
+  `span[data-value=" /delivery/group_delivery_summary/display_delivery_outcome "]`);
+const locationSummary = () => $(`${FORM} span[data-value=" /delivery/group_summary/r_delivery_location "]`);
+const followUpSMS = () => $(`${FORM} span[data-value=" /delivery/chw_sms "]`);
+
+const selectPregnancyOutcome = async (value = OUTCOME.liveBirth) => {
+  const outcome = await pregnancyOutcome(value);
+  await outcome.waitForDisplayed();
+  await outcome.click();
+  return await (await pregnancyOutcomeLabel(value)).getText();
```
jkuester, line 24:
```suggestion
return (await pregnancyOutcomeLabel(value)).getText();
```

Minor, but I don't think you actually need this outer `await`.  Since you are returning a value from an `async` method, I think it is okay to just return a promise (without awaiting it).  

Might be worth trying it.  If it does not work, just ignore this comment. If it does work, you could simplify this code (and the other similar functions below).

tests/e2e/standard/enketo/pregnancy.wdio-spec.js
```diff
@@ -0,0 +1,143 @@
...
+
+describe('New pregnancy', () => {
+  const places = placeFactory.generateHierarchy();
+  const healthCenter = places.find(place => place.type === 'health_center');
+  const user = userFactory.build({ place: healthCenter._id, roles: ['district_admin'] });
+  const pregnantWoman1 = 'Woman1';
+  const pregnantWoman2 = 'Woman2';
+
+  before(async () => { 
+    await utils.saveDocs([...places]);
```
jkuester, line 22:
```suggestion
    await utils.saveDocs(places);
```

Maybe I am missing something here, but is there a reason to create a new array before passing it to `saveDocs`?

tests/e2e/standard/enketo/pregnancy.wdio-spec.js
```diff
@@ -0,0 +1,143 @@
...
+    await loginPage.cookieLogin();
+    await commonPage.goToPeople(healthCenter._id);
+    
+    // Create Woman1 - webapp
+    await contactPage.contactPageDefault.createNewPerson();
+    await newPersonPage.fillDefaultContactForm();
+    await genericForm.submitForm();
+
+    // Createn Woman2 - SMS N form
+    const messageNewWoman = `N ${pregnantWoman2}`;
```
jkuester, line 33:
```suggestion
    // Create Woman2 - SMS N form
```

tests/e2e/standard/enketo/pregnancy.wdio-spec.js
```diff
@@ -0,0 +1,143 @@
...
+    expect(followUpSMS).to.include(note);
+
+    await genericForm.submitForm();
+    await commonPage.waitForPageLoaded();
+    
+    expect(await (await contactPage.pregnancyCard()).isDisplayed()).to.be.true;
+    expect(await contactPage.getPregnancyCardRisk()).to.equal('High risk');
+  });
+
+  it('Verify pregnancy report - Woman1 - webapp', async () => {
```
jkuester, line 74: This is not really a problem, but more of an observation.  Typically, with automated tests, it is recommended that each test be self-contained and atomic (and not rely on any other test to have run first).  In the case of this test, I think it needs the previous test to run first.

I don't think this is a huge problem if this is part of the general pattern we want to follow for these e2e tests.  However, if we are not already committed to this pattern, I would recommend we avoid dependencies between tests since it can make debugging issues a lot more tricky (and makes the idea of eventually running these test in parallel impossible...).

tests/e2e/standard/enketo/pregnancy-visit.wdio-spec.js
```diff
@@ -0,0 +1,87 @@
...
+const genericForm = require('../../../page-objects/forms/generic-form.wdio.page');
+const pregnancyVisitForm = require('../../../page-objects/standard/forms/pregnancy-visit.wdio.page');
+
+describe('Pregnancy Visit', () => {
+  const places = placeFactory.generateHierarchy();
+  const healthCenter = places.find(place => place.type === 'health_center');
+  const user = userFactory.build({ place: healthCenter._id, roles: ['district_admin'] });
+  const pregnantWoman = personFactory.build({ patient_id: '12345', 
+    date_of_birth: moment().subtract(25, 'years').format('YYYY-MM-DD'), 
+    parent: {_id: healthCenter._id, parent: healthCenter.parent} });
```
jkuester, line 20:
```suggestion
  const pregnantWoman = personFactory.build({
    patient_id: '12345',
    date_of_birth: moment().subtract(25, 'years').format('YYYY-MM-DD'), 
    parent: { _id: healthCenter._id, parent: healthCenter.parent }
  });
```
Minor, but is a lot easier to read if you format it a bit.

tests/e2e/standard/enketo/delivery.wdio-spec.js
```diff
@@ -0,0 +1,176 @@
...
+  const places = placeFactory.generateHierarchy();
+  const healthCenter = places.find(place => place.type === 'health_center');
+  const user = userFactory.build({ place: healthCenter._id, roles: ['district_admin'] });
+  const pregnantWoman1 = 'Woman1';
+  const pregnantWoman2 = 'Woman2';
+  let medicIDW1 = '';
+  let medicIDW2 = '';
+
+  before(async () => {
+    await utils.saveDocs([...places]);
```
jkuester, line 24:
```suggestion
    await utils.saveDocs(places);
```

Maybe I am missing something here, but is there a reason to create a new array before passing it to saveDocs?

jkuester (changes requested):
Great tests!  Super clean!  Just added some minor suggestions in-line (and one more theoretical question abour our general e2e testing patterns...)

tests/utils.js
```diff
@@ -139,7 +139,7 @@ const deleteAll = (except) => {
     'branding',
     'partners',
     'settings',
-    /^form:contact:/,
+    /^form:/,
```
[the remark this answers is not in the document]
tatilepizs: Yeap, it was a little hard for me to find it 😅

tests/page-objects/contacts/contacts.wdio.page.js
```diff
@@ -259,6 +260,11 @@ const getContactMedicID = async () => {
   return (await contactMedicID()).getText();
 };
 
+const createNewPerson = async () => {
+  await newPersonContactButton().waitForDisplayed();
+  await newPersonContactButton().click();
```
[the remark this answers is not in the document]
tatilepizs: It is the same thing, I don't know why I didn't see that was already implemented.
I removed that and use the method `addPerson` instead and removed the file `new-person.wdio.page.js` because it was not needed anymore.
Thank you for seeing this! 😄

tests/page-objects/contacts/contacts.wdio.page.js
```diff
@@ -259,6 +260,11 @@ const getContactMedicID = async () => {
   return (await contactMedicID()).getText();
 };
 
+const createNewPerson = async () => {
```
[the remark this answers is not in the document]
tatilepizs: This method was removed because the functionality was already there.
Thanks again

tests/page-objects/standard/forms/delivery.wdio.page.js
tatilepizs: You are completely right, it worked!
Thanks

tests/e2e/standard/enketo/pregnancy.wdio-spec.js
tatilepizs: You are right, there is no need to create a new array. I changed it in the three files.
Thanks!

tests/e2e/standard/enketo/pregnancy.wdio-spec.js
tatilepizs: That is a really good point, I haven't thought about it.

In the last individual test of this group I check the `targets` and those values include the pregnancy registered by `webapp` and the pregnancy registered by `SMS`, so I will not be able to separate anything and the five `it` will need to be all together in a single one.

Do you think that it will be a better idea to have it all in just one individual test?

I agree with you that no other test should depend on the result of another test, but I also like the idea to know what is being tested in every step of the process, and at the end every file has the `describe` that includes all the individual tests. 
What do you think?

jkuester: > Do you think that it will be a better idea to have it all in just one individual test?

I am interested to hear what @ngaruko thinks, but as nasty as it would look, I think it would be better to have one big test, then to have inter-dependencies between tests.

One interesting alternative would be to centralize the part that is needed for each test (e.g. submitting the pregnancy form) into a `before` method that runs once for the `describe`.  Then you can still break out the various assertions into separate `it` functions to give it some better structure. Then you could also use nested `describe` blocks to layer in some more complexity (e.g. submit the regular pregnancy form in the outer block and then have an inner `describe` with a `before` that submits the sms form.  Then `it` functions in the inner `describe` can include assertions that require both forms to have been submitted...)

ngaruko: I think it makes sense @jkuester @tatilepizs to keep each test independent and use the `before hooks` for pre-requisites. If the pre-requisite is not part of the actual test, the most efficient/stable way I have seen is to maybe try and use a non-UI way to achieve that , when possible, same we do when we create users or seed other docs needed for a test.

tests/utils.js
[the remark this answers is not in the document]
jkuester: So, this is random, @tatilepizs, but I just want to make sure you know your debugging work here was not in vain!

I was just seeing weird issues with my new tests working when I ran them on their own, but failing when I ran the whole suite.  Eventually I realized it was because the "New Action" button was disabled.  I was very confused by what was going on, but after awhile I remembered that this sounded very similar to the issue you had seen.  Looked up this code change again and dropped it in my branch.  Sure enough!  It seems to have solved my issue!  (Without me having to debug for hours to find a fix :sweat: )  :partying_face:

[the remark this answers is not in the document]
tatilepizs: Thank you for sharing it! 😄 It makes me really happy that this change was able to help you! 🎉

tests/e2e/standard/enketo/pregnancy.wdio-spec.js
tatilepizs: Thank you both for the feedback!
I moved to just one individual test (`it`) the creation of the report, the verification of the contact cards, and the verification of the reports, which actually makes more sense.

tatilepizs:
@jkuester the build finally passed successfully, and I made some changes in the process.
Could you please take a look?

tatilepizs:
Ok, the build finally passed after the merge with `master` 🎉 
@jkuester could you please review the changes one more time?

tests/e2e/default/wdio.conf.js
```diff
@@ -334,7 +334,7 @@ const baseConfig = {
     if (users.length) {
       await utils.deleteUsers(users);
     }
-    await utils.revertDb([], true);
+    await utils.revertDb([/^form:/], true);
```
jkuester, line 337: :heart:

jkuester (approved):
Very nice!  My last question is if there are any existing test classes that can now be removed since we have replaced them with these new tests?  (Thinking specifically of the [submit-delivery-form.specs.js](https://github.com/medic/cht-core/blob/master/tests/e2e/protractor/forms/submit-delivery-form.specs.js)...)

tatilepizs:
Unfortunately no, because the [submit-delivery-form.specs.js](https://github.com/medic/cht-core/blob/master/tests/e2e/protractor/forms/submit-delivery-form.specs.js) is for the `default` config, and these tests are for the `standard` those are tests that we need to migrate 😄

PR #7752, opened 2022-08-30
Remove existent backend servers before adding more.

dianabarsan:
# Description

Removes existent backend servers from haproxy config before adding the current ones. 

# Code review checklist

# License


garethbowen (approved):
Makes sense.

A slightly cleaner solution would be to use multiple haproxy config files - one that's static and one that's entirely generated. That way we could just drop the previous one and create a brand new one rather than having to sed and append.

Still, approved! We can improve things as we go...

dianabarsan:
> A slightly cleaner solution would be to use multiple haproxy config files

I really like this idea. And I don't think it's hard to implement!

dianabarsan:
I've split the config in two files. I am of two minds about splitting them further. Let me know what you think, @garethbowen . Thank you!

garethbowen (approved):
Awesome!

PR #7754, opened 2022-08-30
Use phone from user doc and not contact when creating new user

jkuester:


m5r (approved):
Looking good! We will have some conflicts with the username generation fix so let's get your branch merged first and I'll deal with the conflicts on my end

jkuester:
@m5r I ended up needing to rework the logic here.  We wanted to be able to possibly edit the phone number in the form and then use that for the new _user's_ phone (and not just the new contact.  So, I updated the form to load the old _user's_ phone as the default phone number.  

If the old user has a form, that is used as the new phone number for the new user (and the associated contact).  If the old user does not have a phone, one must be provided on the form to use for the new user/contact.  (The old contact's phone number is ignored.)

I did go ahead and merge your PR and then adjust this code to work with it.  Once I confirm this builds, I will merge it as well...

PR #7757, opened 2022-09-01
5710 add breadcrumbs to tasks lists

elvisdorkenoo:
# Description

This PR adds a third line of text to tasks to display the additional contextual information.
This new line of text should display breadcrumbs showing where the person/place the task is about belongs to.
The functionality should match how we display breadcrumbs elsewhere on the app. Note that we are updating breadcrumbs across the app so that they do not list out the level the user themselves belongs to repetitively (i.e. CHW Janet will not see “CHW Janet’s Area” repeated all over the place, ticket https://github.com/medic/cht-core/issues/5697).
The text should be styled just like breadcrumbs elsewhere on the app (Noto Sans 14px).

medic/cht-core#5710

# Code review checklist

# License


elvisdorkenoo:
> (Just not exactly sure what do to when assigned to a _draft_ PR....) 😄
leave draft reviews 😅.
Sorry you are right, I should the assignment when Draft PR -> PR.
I would happily welcome some feedback, but : 
it's not yet PR because I still have the fix broken unit tests and add new ones ; and the truth is I'm struggling at it.

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -97,6 +104,13 @@ export class TasksComponent implements OnInit, OnDestroy {
     this.hasTasks = false;
     this.loading = true;
     this.debouncedReload = _debounce(this.refreshTasks.bind(this), 1000, { maxWait: 10 * 1000 });
+    if (!this.sessionService.isOnlineOnly()) {
+      this
+        .getCurrentLineageLevel()
+        .then((currentLevel) => {
```
latin-panda, line 110:
```suggestion
        .then(currentLevel => this.currentLevel = currentLevel);
```

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -116,14 +130,13 @@ export class TasksComponent implements OnInit, OnDestroy {
     window.location.reload();
   }
 
-  private hydrateEmissions(taskDocs) {
+  private async hydrateEmissions(taskDocs) {
```
latin-panda, line 133: No `await`, no need of `async`
```suggestion
  private hydrateEmissions(taskDocs) {
```

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -142,7 +155,29 @@ export class TasksComponent implements OnInit, OnDestroy {
       .then(taskDocs => {
         this.hasTasks = taskDocs.length > 0;
         this.loading = false;
-        this.tasksActions.setTasksList(this.hydrateEmissions(taskDocs));
+        this.hydrateEmissions(taskDocs)
```
latin-panda, line 158:
```suggestion
        this
          .hydrateEmissions(taskDocs)
```

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -142,7 +155,29 @@ export class TasksComponent implements OnInit, OnDestroy {
...
+          .then((tasks) => {
```
latin-panda, line 159:
```suggestion
          .then(tasks => {
```

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -142,7 +155,29 @@ export class TasksComponent implements OnInit, OnDestroy {
...
+            // get lineages for all tasks
+            this.getLineagesFromTaskDocs(tasks)
```
latin-panda, line 161:
```suggestion
            this
              .getLineagesFromTaskDocs(tasks)
```

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -142,7 +155,29 @@ export class TasksComponent implements OnInit, OnDestroy {
...
+              .then((subjects) => {
```
latin-panda, line 162:
```suggestion
              .then(subjects => {
```

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -164,4 +198,14 @@ export class TasksComponent implements OnInit, OnDestroy {
   listTrackBy(index, task) {
     return task?._id;
   }
+
+  async getCurrentLineageLevel(){
```
latin-panda, line 202: No `await`, no need of `async`
```suggestion
  getCurrentLineageLevel(){
```

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -164,4 +198,14 @@ export class TasksComponent implements OnInit, OnDestroy {
...
+    return this.userContactService.get().then(user => user?.parent?.name);
+  }
+
+  async getLineagesFromTaskDocs(taskDocs){
+    const ids = _map(taskDocs, 'forId');
+    return await this.lineageModelGeneratorService.reportSubjects(ids);
```
latin-panda, line 208:
```suggestion
    return this.lineageModelGeneratorService.reportSubjects(ids);
```
It returns a promise already, the `await` won't have effect. Remember to remove the `async`

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -164,4 +198,14 @@ export class TasksComponent implements OnInit, OnDestroy {
   listTrackBy(index, task) {
     return task?._id;
   }
+
+  async getCurrentLineageLevel(){
+    return this.userContactService.get().then(user => user?.parent?.name);
+  }
+
+  async getLineagesFromTaskDocs(taskDocs){
+    const ids = _map(taskDocs, 'forId');
```
latin-panda, line 207: Can you use native javascript `map` function?
`taskDocs.map(task => task.forId)`

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -142,7 +155,29 @@ export class TasksComponent implements OnInit, OnDestroy {
       .then(taskDocs => {
         this.hasTasks = taskDocs.length > 0;
         this.loading = false;
-        this.tasksActions.setTasksList(this.hydrateEmissions(taskDocs));
+        this.hydrateEmissions(taskDocs)
+          .then((tasks) => {
+            // get lineages for all tasks
+            this.getLineagesFromTaskDocs(tasks)
+              .then((subjects) => {
+                const deepCopy = obj => JSON.parse(JSON.stringify(obj));
```
latin-panda, line 163: What's the issue you're trying to resolve with a deepCopy? 

What is the structure of this `obj`?

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -142,7 +155,29 @@ export class TasksComponent implements OnInit, OnDestroy {
...
+                const lineagedTasks = deepCopy(tasks);
+                lineagedTasks?.forEach((task) => {
```
latin-panda, line 165: This logic is hard to read and it's adding a lot of nested code blocks. This can be an indicator of code smell and that it can be structured better. We try to keep code as flat as possible.

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -142,7 +155,29 @@ export class TasksComponent implements OnInit, OnDestroy {
...
+                  // map tasks with lineages
+                  let lineage = _map(
```
latin-panda, line 167: can you use native javascript functions for `map` and `find`?

latin-panda: This can be a separate function, then be a step by step logic, that would make it easier to read. For example:
```
private getTaskLineage(subjects, task) {
  if (!subjects?.lenght || !this.currentLevel) { 
     return; // Exit early whenever possible!! Keeps code flat and understandable 
  }

 // Step by step logic to help readability & maintenance 
  const subject = subject.find( ...... find leanage .... );
  return lineage.map(...... extract what you need ....); 
}

private removeCurrentLineage(lineage) {
  if (!this.currentLevel || !lineage?.length) {
    return; // Exit early whenever possible!! Keeps code flat and understandable 
  }

  return lineage.filter(level => level && level !== this.currentLevel);
}
```

latin-panda: Then in the `refreshTasks()` will look cleaner and flatter because we took out some code, we can restructure something like this, but test it!:

```
async refreshTasks() {
try {
   ...  Original telemetry code here ....

  const isEnabled = await this.rulesEngineService.isEnabled();
  this.tasksDisabled = !isEnabled;
  const taskDocs = isEnabled ? this.rulesEngineService.fetchTaskDocsForAllContacts() : [];

  this.hasTasks = taskDocs.length > 0;
  this.loading = false;

  const hydratedTasks = await this.hydrateEmissions(taskDocs) || [];
  const subjects = await this.getLineagesFromTaskDocs(hydratedTasks);

  hydratedTasks.forEach(task => {
    const lineage = getTaskLineage(task); // the new function
    task.lineage = removeCurrentLineage(lineage); // the other new function
  });

  this.tasksActions.setTasksList(hydratedTasks);

  if (!this.tasksLoaded) {
     this.tasksActions.setTasksLoaded(true);
  }

  ... telemetry code ...
  } catch (exception) {
 .... exception code....
 } 
}
```

latin-panda (changes requested):
@elvisdorkenoo I see this PR is not pointing to master, is that intentional?

I didn't have time to review the unit tests but at least I'm sending you some feedback from the other files. Let me know your thoughts or questions if something isn't clear :)

webapp/src/ts/modules/tasks/tasks.component.ts
elvisdorkenoo: Yeah great suggestion,  thank you mvery much, I refactored the whole refreshTasks() function following your guidelines, you can take a look to the updated PR.

elvisdorkenoo:
> @elvisdorkenoo I see this PR is not pointing to master, is that intentional?
> 
> I didn't have time to review the unit tests but at least I'm sending you some feedback from the other files. Let me know your thoughts or questions if something isn't clear :)

Thank you everything is clear.
Yeah the PR is pointing #5697 working branch, because I think after merging #7757 into #7732 it would be done in one one delivery and also make AT work easier and faster, please let me know if we can make it confusing.

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -116,52 +128,86 @@ export class TasksComponent implements OnInit, OnDestroy {
     window.location.reload();
   }
 
-  private hydrateEmissions(taskDocs) {
+  private async hydrateEmissions(taskDocs) {
```
latin-panda, line 131: No `await`, then no need of `async`
```suggestion
  private hydrateEmissions(taskDocs) {
```

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -116,52 +128,86 @@ export class TasksComponent implements OnInit, OnDestroy {
...
-      })
-      .catch(err => {
-        console.error('Error getting tasks for all contacts', err);
-
-        this.error = true;
-        this.loading = false;
-        this.hasTasks = false;
-        this.tasksActions.setTasksList([]);
-      });
+  async refreshTasks() {
```
latin-panda, line 142: Since this is not called in the DOM, then it can remain private as before.
```suggestion
  private async refreshTasks() {
```

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -116,52 +128,86 @@ export class TasksComponent implements OnInit, OnDestroy {
...
+      this.tasksDisabled = !isEnabled;
+      const taskDocs = isEnabled ? await this.rulesEngineService.fetchTaskDocsForAllContacts() : [];
+
+      this.hasTasks = taskDocs.length > 0;
+      this.loading = false;
+
+      const hydratedTasks = await this.hydrateEmissions(taskDocs) || [];
+      const subjects = await this.getLineagesFromTaskDocs(hydratedTasks);
+
+      if(subjects) {
```
latin-panda, line 158:
```suggestion
      if (subjects) {
```
This one has been recurrent. The IDE can do this for you :) look into the IDE settings

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -116,52 +128,86 @@ export class TasksComponent implements OnInit, OnDestroy {
...
+      this.hasTasks = false;
+      this.tasksActions.setTasksList([]);
+    }
   }
 
   listTrackBy(index, task) {
     return task?._id;
   }
+
+  getCurrentLineageLevel(){
```
latin-panda, line 188: Since it's not used in the DOM
```suggestion
  private getCurrentLineageLevel() {
```

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -116,52 +128,86 @@ export class TasksComponent implements OnInit, OnDestroy {
...
+    return this.userContactService.get().then(user => user?.parent?.name);
+  }
+
+  getLineagesFromTaskDocs(taskDocs){
```
latin-panda, line 192:
```suggestion
  private getLineagesFromTaskDocs(taskDocs) {
```

webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
```diff
@@ -16,6 +16,9 @@ import { TasksComponent } from '@mm-modules/tasks/tasks.component';
 import { NavigationComponent } from '@mm-components/navigation/navigation.component';
 import { Selectors } from '@mm-selectors/index';
 import { NavigationService } from '@mm-services/navigation.service';
+import {UserContactService} from '@mm-services/user-contact.service';
```
latin-panda, line 19:
```suggestion
import { UserContactService } from '@mm-services/user-contact.service';
```
Add these spaces pls

webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
```diff
@@ -16,6 +16,9 @@ import { TasksComponent } from '@mm-modules/tasks/tasks.component';
...
+import {SessionService} from '@mm-services/session.service';
```
latin-panda, line 20:
```suggestion
import { SessionService } from '@mm-services/session.service';
```
Add these spaces pls, the IDE can add these automatically.

webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
```diff
@@ -140,7 +166,7 @@ describe('TasksComponent', () => {
     expect(consoleErrorMock.args[0][0]).to.equal('Error getting tasks for all contacts');
   });
 
-  it('tasks render', async () => {
+  it('tasks render', async() => {
```
latin-panda, line 169:
```suggestion
  it('tasks render', async () => {
```
Before it was okay, the reason is that it's a reserved word of js

webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
```diff
@@ -278,4 +303,161 @@ describe('TasksComponent', () => {
...
+        _id: 'parent',
+        name: 'CHW Bettys Area',
+      },
+    };
+    const taskDocs = [
+      { _id: '1', emission: { _id: 'e1', dueDate: '2020-10-20' }, forId: 'a', owner: 'a' },
+      { _id: '2', emission: { _id: 'e2', dueDate: '2020-10-20' }, forId: 'b', owner: 'b' },
+    ];
+    const taskLineages = [
+      {
```
latin-panda, line 320: This is breaking in a weird place. When breaking object please put every property in separate lines, like this:
```
{
  _id: ...
  lineage: [ ... ]
}
```
The idea is to have different data possibilities that can break your code. When writhing tests, think about what can break your code. Because you know users not always follow the happy path.

You can test this with 3 objects.
1. obj has lineage
2. obj lineage property is missing
3. obj lineage property is an empty array

webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
```diff
@@ -278,4 +303,161 @@ describe('TasksComponent', () => {
...
+        _id: 'b', lineage: [
+          {name: 'Amy Johnsons Household'},
+          {name: 'St Elmos Concession'},
+          {name: 'Chattanooga Village'},
+        ]
+      }
+    ];
+
+    it('it should not change the reports lineages if the connected user is online only', async() => {
```
latin-panda, line 337: Targets! no reports :)

And space after `async` as it's reserved word. 
```suggestion
    it('should not alter target lineages if user is online only', async () => {
```

webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
```diff
@@ -278,4 +303,161 @@ describe('TasksComponent', () => {
...
+        [
+          {
+            _id: 'e1',
+            date: new Date('2020-10-20'),
+            dueDate: '2020-10-20',
+            lineage: [
+              'Amy Johnsons Household',
+              'St Elmos Concession',
+              'Chattanooga Village',
+              'CHW Bettys Area'
```
latin-panda, line 348:
```suggestion
              'CHW Bettys Area',
```

webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
```diff
@@ -278,4 +303,161 @@ describe('TasksComponent', () => {
...
+            date: new Date('2020-10-20'),
+            dueDate: '2020-10-20',
+            lineage: [
+              'Amy Johnsons Household',
+              'St Elmos Concession',
+              'Chattanooga Village',
+            ],
+            overdue: true,
+            owner: 'b',
+          }
```
latin-panda, line 364:
```suggestion
          },
```

webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
```diff
@@ -278,4 +303,161 @@ describe('TasksComponent', () => {
...
+        ];
+      userContactService.get.resolves(bettysContactDoc);
```
latin-panda, line 366: Unit tests usually have 3 main sections:

1. Test settings and test data definition
2. Invocation of code unit that is being tested
3. Assertions

Some people likes to add comments to flag the sections for example: // GIVEN, // WHEN, // THEN. You will see that in some old android code that we have. 
But here we usually tend to add just simple spaces between to organise it and make it easier for the reader. See exampl:
```
it('.......', () => {
  const testData = [ ..... data ....];

  userContactService.get.resolves(bettysContactDoc);
  sessionService.isOnlineOnly.returns(true);
  ... other test settings ...

  // space to separate test settings and test data from unit code execution
  await new Promise(resolve => { 
      sinon.stub(TasksActions.prototype, 'setTasksList').callsFake(resolve);
      getComponent(); // <--- This is the unit code execution that we're testing
  });

  // space to separate the assertions
  expect(component.currentLevel).to.be.undefined;
  expect((<any>TasksActions.prototype.setTasksList).args).to.deep.equal([[expectedTasks]]);
});
```

webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
```diff
@@ -278,4 +303,161 @@ describe('TasksComponent', () => {
...
+      lineageModelGeneratorService.reportSubjects.resolves(taskLineages);
+      await new Promise(resolve => {
+        sinon.stub(TasksActions.prototype, 'setTasksList').callsFake(resolve);
+        getComponent();
+      });
+      expect(component.currentLevel).to.be.undefined;
+      expect((<any>TasksActions.prototype.setTasksList).args).to.deep.equal([[expectedTasks]]);
+    });
+
+    it('it should not change the reports lineages if the connected user is offline with unrelated lineage', async() => {
```
latin-panda, line 378: Targets! no reports :)

And space after async as it's reserved word.
```suggestion
    it('should not alter tasks lineage if user is offline with unrelated lineage', async () => {
```

latin-panda: These things are usually caught in a self-code review and that would have saved time

webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
```diff
@@ -278,4 +303,161 @@ describe('TasksComponent', () => {
...
+      await new Promise(resolve => {
+        sinon.stub(TasksActions.prototype, 'setTasksList').callsFake(resolve);
+        getComponent();
+      });
+      const tasks = await (<any>TasksActions.prototype.setTasksList).args;
+      expect(component.currentLevel).to.equal('parent');
+      expect(tasks).to.deep.equal([[expectedTasks]]);
+    });
+
+    it('it should update the reports lineages if the connected user is offline with related place', async() => {
```
latin-panda, line 420: Targets! no reports :)

And space after async as it's reserved word.
```suggestion
    it('should update the tasks lineage if user is offline with related place to lineage', async () => {
```

latin-panda (changes requested):
@elvisdorkenoo thanks for fixing this PR, it's a lot easier to review just the tasks changes!

I left a couple of feedbacks in line.

webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
elvisdorkenoo: well noted

webapp/src/ts/modules/tasks/tasks.component.html
```diff
@@ -32,6 +32,7 @@ <h4>
               </div>
 
               <div *ngIf="task.priorityLabel" class="warning">{{task.priorityLabel}}</div>
+              <div *ngIf="task.lineage?.length" class="detail" [innerHTML]="task.lineage | lineage"></div>
```
latin-panda, line 35: @elvisdorkenoo this line is breaking the e2e, not sure why.

webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
```diff
@@ -43,6 +57,15 @@ describe('TasksComponent', () => {
     contactTypesService = {
       includes: sinon.stub(),
     };
+    sessionService = {
+      isDbAdmin: sinon.stub().returns(false),
+      isOnlineOnly: sinon.stub().returns(false),
+      userCtx: sinon.stub().returns({ name: 'Sarah' })
```
jkuester, line 63: Do we actually need to stub `isDbAdmin` and `userCtx` here?  I do not remember seeing these functions being used....

webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
```diff
@@ -150,12 +176,11 @@ describe('TasksComponent', () => {
       { _id: '2', emission: { _id: 'e2', dueDate: pastDate.format('YYYY-MM-DD') }, owner: 'b' },
     ];
     rulesEngineService.fetchTaskDocsForAllContacts.resolves(taskDocs);
-
     await new Promise(resolve => {
       sinon.stub(TasksActions.prototype, 'setTasksList').callsFake(resolve);
       getComponent();
     });
-
+    clock.tick();
```
jkuester, line 183: Why was this added here?

jkuester (changes requested):

latin-panda:
@elvisdorkenoo I think the selector was applied incorrectly, the e2e should be okay now.

webapp/src/ts/modules/messages/messages.component.ts
```diff
@@ -160,12 +164,12 @@ export class MessagesComponent implements OnInit, OnDestroy {
...
-        if(!this.sessionService.isOnlineOnly()) {
-          conversations.map((conversation) => {
-            conversation.lineage = conversation.lineage?.filter((level) => {
-              return (level !== this.currentLevel);
-            });
+        // remove the lineage level that belongs to the offline logged-in user, normally the last one
+        if (this.currentLevel) {
+          conversations.forEach((conversation) => {
+            if (conversation.lineage) {
+              conversation.lineage = conversation.lineage.filter(level => level !== this.currentLevel);
```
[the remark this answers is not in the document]
elvisdorkenoo: From my understanding, the lineage calculation for offline user belonging to a place, always stops at the level of current user.
So : 
```
Level A
Level B <--- my offline user is here
Lebel C
```
is unlikely to happen, instead we would have  
```
Level A
Level B <--- my offline user is here
null
```
and the result should be 
`Level A`

webapp/src/ts/modules/reports/reports.component.ts
```diff
@@ -187,24 +180,22 @@ export class ReportsComponent implements OnInit, AfterViewInit, OnDestroy {
...
       report.summary = form ? form.title : report.form;
       report.lineage = report.subject && report.subject.lineage || report.lineage;
-      // filter out the lineage level that belongs to the online logged in user
-      if(!this.sessionService.isOnlineOnly()) {
-        report.lineage = report.lineage.filter((level) => {
-          return (level !== this.currentLevel);
-        });
+      // remove the lineage level that belongs to the offline logged-in user, normally the last one
+      if (this.currentLevel) {
+        report.lineage = report.lineage.filter(level => level !== this.currentLevel);
```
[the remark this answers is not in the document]
elvisdorkenoo: I changed the code to this :
```
      if (this.currentLevel && report.lineage && report.lineage.length) {
        report.lineage = report.lineage.filter(level => level);
        if(report.lineage[report.lineage.length-1] === this.currentLevel){
          report.lineage.pop();
        }
      }
```

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -118,50 +130,84 @@ export class TasksComponent implements OnInit, OnDestroy {
...
+      this.hasTasks = taskDocs.length > 0;
+      this.loading = false;
+
+      const hydratedTasks = await this.hydrateEmissions(taskDocs) || [];
+      const subjects = await this.getLineagesFromTaskDocs(hydratedTasks);
+
+      if (subjects) {
+        hydratedTasks.forEach(task => {
+          const lineage = this.getTaskLineage(subjects, task);
+          task.lineage = this.currentLevel ? this.removeCurrentLineage(lineage) : lineage;
```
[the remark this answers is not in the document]
elvisdorkenoo: True !!

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -118,50 +130,84 @@ export class TasksComponent implements OnInit, OnDestroy {
...
   listTrackBy(index, task) {
     return task?._id;
   }
+
+  private getCurrentLineageLevel() {
+    return this.userContactService.get().then(user => user?.parent?.name);
+  }
+
+  private getLineagesFromTaskDocs(taskDocs) {
+    const ids = taskDocs.map(task => task.forId);
```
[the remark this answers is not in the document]
elvisdorkenoo: Honestly, I forgot about that possibility, and yup turning that ids array into a Set, would save us some time ! Thanks !!

webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
```diff
@@ -278,4 +303,136 @@ describe('TasksComponent', () => {
...
+          { name: 'Chattanooga Village' },
+        ],
+      },
+    ];
+
+    it('should not alter tasks lineage if user is online only', async () => {
+      const expectedTasks = [
+        {
+          _id: 'e1',
+          date: new Date('2020-10-20'),
```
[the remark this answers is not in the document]
elvisdorkenoo: I struggled to get the main code and the test date formats to match, and apparently what did work for me, did not for me, you for the universal solution

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -118,50 +130,84 @@ export class TasksComponent implements OnInit, OnDestroy {
...
+      this.tasksDisabled = !isEnabled;
+      const taskDocs = isEnabled ? await this.rulesEngineService.fetchTaskDocsForAllContacts() : [];
+
+      this.hasTasks = taskDocs.length > 0;
+      this.loading = false;
+
+      const hydratedTasks = await this.hydrateEmissions(taskDocs) || [];
+      const subjects = await this.getLineagesFromTaskDocs(hydratedTasks);
+
+      if (subjects) {
```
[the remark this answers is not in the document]
elvisdorkenoo: True !!

webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
elvisdorkenoo: I think it's the equivalent of` tick() `when using  `clock = sinon.useFakeTimers(now.valueOf())` but you are right, it's not needed anymore, the test works also without it.

webapp/src/ts/modules/tasks/tasks.component.ts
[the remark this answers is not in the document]
elvisdorkenoo: Yeah, I agree, done !

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -118,50 +130,84 @@ export class TasksComponent implements OnInit, OnDestroy {
...
     return task?._id;
   }
+
+  private getCurrentLineageLevel() {
+    return this.userContactService.get().then(user => user?.parent?.name);
+  }
+
+  private getLineagesFromTaskDocs(taskDocs) {
+    const ids = taskDocs.map(task => task.forId);
+    return this.lineageModelGeneratorService.reportSubjects(ids);
```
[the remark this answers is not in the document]
elvisdorkenoo: OK ! applied !

webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
```diff
@@ -278,4 +304,145 @@ describe('TasksComponent', () => {
...
+
+      expect(await component.currentLevel).to.equal('parent');
+      expect((<any>TasksActions.prototype.setTasksList).args).to.deep.equal([[expectedTasks]]);
+    });
+
+    it('should update the tasks lineage if user is offline with related place to lineage', async () => {
+      const expectedTasks = [
+        {
+          _id: 'e1',
+          date: new Date('2020-10-20'),
```
latin-panda, line 417: Apply @jkuester about dates in this test too. 
If you still have the TZ issue, take a look at how we use `sinon.useFakeTimers`

latin-panda (changes requested):
Nice! 🤩🤩 Almost there! Please have a look at the suggestion in line. Thanks!

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -118,50 +128,92 @@ export class TasksComponent implements OnInit, OnDestroy {
...
+
+  private getCurrentLineageLevel() {
+    return this.userContactService.get().then(user => user?.parent?.name);
+  }
+
+  private getLineagesFromTaskDocs(taskDocs) {
+    const ids = [...new Set(taskDocs.map(task => task.forId))];
+    return this
+      .lineageModelGeneratorService.reportSubjects(ids)
+      ?.then(subjects => new Map(subjects.map(subject => [subject._id, subject.lineage])));
```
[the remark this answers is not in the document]
elvisdorkenoo: >  is it possible that reportSubjects return undefined? If not, then the ? check isn't needed.
Yes it is possible, some other unit test data sets do not include the needed informations, so in these cases `reportSubjects` return `undefined`.

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -118,50 +128,92 @@ export class TasksComponent implements OnInit, OnDestroy {
...
+  private cleanAndRemoveCurrentLineage(lineage) {
+    if(!lineage?.length){
+      return;
+    }
+    lineage = lineage.filter(level => level);
+    this.currentLevel.then(currentLevel => {
+      if(currentLevel === lineage[lineage.length-1]){
+        lineage.pop();
+      }
+    });
```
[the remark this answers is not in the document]
latin-panda: I was suggesting it because you're doing a filter after that validation and the resulting array can be empty, therefore the `lineage.length - 1` results in negative index. However it looks like JavaScript just returns `undefined` for negative indexes. 
Screen Shot 2022-09-14 at 9 22 06 am

webapp/src/ts/modules/tasks/tasks.component.ts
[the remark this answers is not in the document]
latin-panda: @elvisdorkenoo, we try not to modify code base to fix unit tests. 
In this case please remove the `?`  from `this.lineageModelGeneratorService.reportSubjects(ids)` as suggested.
Then fix the unit tests by resolving empty array like this around line 66 of the test file:
` lineageModelGeneratorService = { reportSubjects: sinon.stub().resolves([]) };` 
This worked for me.

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -118,50 +128,91 @@ export class TasksComponent implements OnInit, OnDestroy {
...
+      this.tasksDisabled = !isEnabled;
+      const taskDocs = isEnabled ? await this.rulesEngineService.fetchTaskDocsForAllContacts() : [];
+
+      this.hasTasks = taskDocs.length > 0;
+      this.loading = false;
+
+      const hydratedTasks = await this.hydrateEmissions(taskDocs) || [];
+      const subjects = await this.getLineagesFromTaskDocs(hydratedTasks);
+      if (subjects?.size) {
+        for (const task of hydratedTasks) {
```
latin-panda, line 157: In theory the `this.currentLevel` promise should resolve immediately the following iterations and the performance impact of awaiting for every task shouldn't be big.... 

Another option is to await 1 time first and send over the currentLevel to the `getTaskLineage`, something like this:
```
... other code...

const subjects = await this.getLineagesFromTaskDocs(hydratedTasks);

if (subjects?.size) {
  const userLineageLevel = await this.currentLevel;
  hydratedTasks.forEach(task => task.lineage = this.getTaskLineage(subjects, task, userLineageLevel));
}
...continue code...
``` 
And the `cleanAndRemoveCurrentLineage` wont' be `async` anymore.

@jkuester @elvisdorkenoo what do you think?

webapp/src/ts/modules/tasks/tasks.component.ts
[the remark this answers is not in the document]
elvisdorkenoo: yeah worked for me as well, thanks

webapp/src/ts/modules/tasks/tasks.component.ts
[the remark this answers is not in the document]
elvisdorkenoo: yeah good thing we don't have to deal with `NullPointerException` 🙂

webapp/src/ts/modules/tasks/tasks.component.ts
elvisdorkenoo: Yeah I like the second option, it would make me passing over `userLineageLevel `  parameter, but it's fine.

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -118,50 +128,91 @@ export class TasksComponent implements OnInit, OnDestroy {
...
+    const ids = [...new Set(taskDocs.map(task => task.forId))];
+    return this.lineageModelGeneratorService
+      .reportSubjects(ids)
+      .then(subjects => new Map(subjects.map(subject => [subject._id, subject.lineage])));
+  }
+
+  private getTaskLineage(subjects, task, userLineageLevel) {
+    if (!subjects?.size) {
+      return;
+    }
```
jkuester, line 200: Minor, but this check is unnecessary since you have done it already above.

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -118,50 +128,91 @@ export class TasksComponent implements OnInit, OnDestroy {
...
+
+      this.hasTasks = taskDocs.length > 0;
+      this.loading = false;
+
+      const hydratedTasks = await this.hydrateEmissions(taskDocs) || [];
+      const subjects = await this.getLineagesFromTaskDocs(hydratedTasks);
+      if (subjects?.size) {
+        const userLineageLevel = await this.currentLevel;
+        for (const task of hydratedTasks) {
+          task.lineage = await this.getTaskLineage(subjects, task, userLineageLevel);
```
jkuester, line 159: Pretty sure we don't need to await this call.

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -118,50 +128,91 @@ export class TasksComponent implements OnInit, OnDestroy {
...
   private hydrateEmissions(taskDocs) {
     return taskDocs.map(taskDoc => {
-      const emission = { ...taskDoc.emission };
+      const emission = {...taskDoc.emission};
       const dueDate = moment(emission.dueDate, 'YYYY-MM-DD');
       emission.date = new Date(dueDate.valueOf());
       emission.overdue = dueDate.isBefore(moment());
       emission.owner = taskDoc.owner;
-
+      emission.forId = taskDoc.forId ? taskDoc.forId : taskDoc.owner;
```
jkuester, line 136: Are we sure that we are not going to inadvertently trigger some additional logic by setting the `forId` to `owner`?

I am not sure exactly what the difference is between the two fields anyway, but I am a little worried that this could have unintended consequences...

jkuester (commented):
Looking good!  Couple minor comments.  

The big question I have left is the same one as the [other PR](https://github.com/medic/cht-core/pull/7732#pullrequestreview-1108138360) regarding whether or not we need an e2e test for these breadcrumbs...

webapp/src/ts/modules/tasks/tasks.component.ts
latin-panda: @elvisdorkenoo - To play safe, you can:
- Line 135 has `emission.owner` already defined and available. 
- Line 136 - Add `emission.forId` and don't default it to `task.owner` because it's already expose in `emission.owner`.
- In your `getLineagesFromTaskDocs` check for both properties: `const ids = [...new Set(taskDocs.map(task => task.forId || task.owner))];`

webapp/src/ts/modules/tasks/tasks.component.ts
latin-panda: and you can use for each as in [the sample here](https://github.com/medic/cht-core/pull/7757/files/8a60b20e18ca8d13daacb0c53c6fea6bb860fced#r970280412)

webapp/src/ts/modules/tasks/tasks.component.ts
elvisdorkenoo: `task => task.forId || task.owner ` is actually what I've before changing it this way when I got the confirmation that when it's set, task.forId always equals task.owner. But yeah let's play it safe.

latin-panda:
@elvisdorkenoo is it ready for review again?

latin-panda:
@elvisdorkenoo Looks like the e2e is failing consistently in the same place, can you please have a look?

elvisdorkenoo:
> @elvisdorkenoo Looks like the e2e is failing consistently in the same place, can you please have a look?

I tried to take a look, even reverted my changes, it is also failing when I run that e2e locally, it looks mostly a timeout issue.

jkuester (approved):
Just want to say that this looks good (pending the build issues being fixed).  

New e2e tests will be added for this functionality as a part of https://github.com/medic/cht-core/issues/7806

latin-panda (approved):
LGTM!

Issue #7746, opened 2022-08-25
Search bar redesign


latin-panda:
**Description**
Taken from the [forum post](https://forum.communityhealthtoolkit.org/t/updates-to-the-search-feature/2107): 

As a follow up to our work on [filters](https://forum.communityhealthtoolkit.org/t/feature-testing-filters/1917) ([ticket](https://github.com/medic/cht-core/issues/7653)), the Search feature is next on our [UI evolution](https://forum.communityhealthtoolkit.org/t/updates-to-the-cht-ui-ux/1998).

We are moving to an expandable search to free up real estate in the top bar, collapsing the current search bar into an icon that opens when tapped. Current functionality will remain the same, with users being able to both search and filter results simultaneously.

<details>
<summary> The old version can be seen here </summary>

<img width="700" src=https://user-images.githubusercontent.com/66472237/186577484-500b9909-6801-4bdb-a25a-64d10a83b459.png></img>
<br>
<img width="300" src=https://user-images.githubusercontent.com/66472237/186577524-d46c02bd-99ed-44f8-a0e1-1d95d36a453e.png></img>

</details>

![Screen Shot 2022-08-25 at 11 48 05 am](https://user-images.githubusercontent.com/66472237/186577270-987cbccc-7c7c-41ab-9cee-f067097446be.png)
![Screen Shot 2022-08-25 at 11 48 50 am](https://user-images.githubusercontent.com/66472237/186577385-d9bbde9f-c516-4aaa-941a-4d1e62267065.png)

**Implementation**
The Old and New version of the search and filters interface will be available for a period of time. By default, users will see the new version. You can have users continue to see the old version by giving them the `can_view_old_filter_and_search` [permission](https://docs.communityhealthtoolkit.org/apps/reference/app-settings/user-permissions/) but... the old version will be removed in a future release so it is important to have a plan to get users onto the new version ASAP.

latin-panda:
This is ready for AT in [this PR](https://github.com/medic/cht-core/pull/7758) 
Please have a look at the PR description for details.

tatilepizs:
### Test

**Config:** Default
**Environment:** Local with docker helper script
**Platform:** WebApp
**Browser:** Chrome
**cht-core version:** `7746-search-bar-redesign`
**Platform - mobiles:** Mobile App. Downloaded from play store **version 1.0.1**

### Mobiles

**Phone:** Samsung Galaxy A01
**Android System WebView Version:** 103.0.5060.129
[...]

PR #7758, opened 2022-09-01
7746 - Search bar redesign

latin-panda:
# Description

This PR apply the new design to Contacts and Reports' search bar. One thing to noticed is that the old design should be working fine as before, it can be enabled by adding `can_view_old_filter_and_search` [permission](https://docs.communityhealthtoolkit.org/apps/reference/app-settings/user-permissions/) to the config. 
- Creates a new Search Bar Component what is used in Contacts tab and Reports tab.
- Skips tour's step that explain about search, we did the same for the new filter in reports tab.
- Resets button is now only for filter, no for search, and it's located inside the filter side bar. 
- Adds a new button to clean search, a left chevron.
- Uses a new LESS variable `@media-small-mobile: 400px;` in all media query for small screens.
- The Sort stays pretty much the same as originally designed. The placement was fixed to match the new position of the tigger icon and the dropdown's text is darker to help readability. 
- Adds skip parameter to Global Actions `clearFilters`, this allows us to keep the value of a filter criteria, originally this was clearing search and filters, but now we just want to clear filters so we can skip "search" when clicking in "Reset". Any other situation (like ngDestroy of a tab page or the old filter) can clear both search and filters.

https://github.com/medic/cht-core/issues/7746
[CHT-Docs PR](https://github.com/medic/cht-docs/pull/776)

# Code review checklist

# License


latin-panda:
Hi @m5r can you please review? Thanks for your time!

config/default/app_settings.json
```diff
@@ -264,7 +264,8 @@
     ],
     "can_upgrade": [
       "program_officer"
-    ]
+    ],
+    "can_view_old_filter_and_search": []
```
m5r, line 268: Nice! That makes it easy to play with this feature flag locally

m5r (approved):
LGTM, the new design looks good and it's nice to have a way to revert to the old one for partners who need more time to transition

webapp/src/css/variables.less
```diff
@@ -86,6 +86,7 @@
 @content-margin: calc((100vw - @content-width) / 2);
 @admin-content-width: 1440px;
 @media-mobile: 767px;
+@media-small-mobile: 400px;
```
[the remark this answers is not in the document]
latin-panda: Thank you!

config/default/app_settings.json
latin-panda: Thanks!

Issue #7759, opened 2022-09-01
Timeouts for view requests when syncing can leave changes requests hanging forever


dianabarsan:
**Describe the bug**
When an offline user requests `_changes`, two view requests are required to determine which documents the user has access to. 
If one of the view requests times out because the view is not indexed, the initiating `_changes` request hangs on forever. 

**To Reproduce**
1. Start with a hefty database, indexing the `medic` ddoc view should take over 5 minutes. 
2. Login as an offline user. 
3. Edit `docs_by_replication_key` view in the `medic` ddoc. The change should be inconsequential, even editing or adding a comment.
4. Start indexing. 
5. Sync with your offline user. 
6. Watching the api logs, you should see something like: 
```
2022-09-01 07:54:58 ERROR: Server error: {
  error: 'timeout',
  reason: 'The request could not be processed in a reasonable amount of time.',
  status: 500,
  name: 'timeout',
  message: 'The request could not be processed in a reasonable amount of time.'
} 
```
This is where your request should have been interrupted. However your client will still show that sync-ing is ongoing.

**Expected behavior**
You should get an error that syncing was not possible. Or, at least, the request should not hang forever. 

**Environment**
- App: api
- Version: 3.10+

**Additional information:**
This is a tricky situation, because of our heartbeat. 
We start sending a heartbeat immediately after we receive the request (before we start setting up required view requests). Sending a heartbeat through means that we've already sent headers and a 200 status. So if the error occurs after the first heartbeat, we can no longer properly communicate the error to the client, through a status. 

Regression after this was fixed in: https://github.com/medic/cht-core/issues/6084

dianabarsan:
I think the easy solution, for now, is to make sure we always end errored requests. 
But from the client's perspective, that 200 means that pouchdb considers that the request succeeded and reports that replication is successful, even though it was not.

dianabarsan:
I think the easy solution, for now, is to make sure we always end errored requests. 
But from the client's perspective, that 200 means that pouchdb considers that the request succeeded and reports that replication is successful, even though it was not.

dianabarsan:
This is ready for AT on `7759-always-end-errored-requests`

latin-panda:
@tatilepizs @dianabarsan just having a look at 3.17 progress.. how is this going? :)

ngaruko:
[...]

PR #7762, opened 2022-09-02
Always end errored requests

dianabarsan:
# Description

Always ends errored requests. Updates client fetch to treat case where unsuccessful changes request has 200 status. 

medic/cht-core#7759
# Code review checklist

# License

### DIFF
--- a/api/src/server-utils.js
+++ b/api/src/server-utils.js
@@ -6,14 +6,12 @@ const logger = require('./logger');
 const MEDIC_BASIC_AUTH = 'Basic realm="Medic Web Services"';
 const cookie = require('./services/cookie');
 
-const APP_JSON = 'application/json';
-
-const wantsJSON = req => req.get('Accept') === APP_JSON;
+const wantsJSON = req => req.accepts('json');
 
 const writeJSON = (res, code, error, details) => {
   if (!res.headersSent) {
     res.status(code);
-    res.setHeader('Content-Type', APP_JSON);
+    res.type('json');
   }
   // using res.json would also automatically try to set the Content-Type header, which fails if headers are sent
   res.end(JSON.stringify({ code, error, details }));
--- a/api/src/server-utils.js
+++ b/api/src/server-utils.js
@@ -6,13 +6,17 @@ const logger = require('./logger');
 const MEDIC_BASIC_AUTH = 'Basic realm="Medic Web Services"';
 const cookie = require('./services/cookie');
 
-const wantsJSON = req => req.get('Accept') === 'application/json';
+const APP_JSON = 'application/json';
+
+const wantsJSON = req => req.get('Accept') === APP_JSON;
 
 const writeJSON = (res, code, error, details) => {
-  if (!res.headersSent && !res.ended) {
+  if (!res.headersSent) {
     res.status(code);
-    res.json({ code, error, details });
+    res.setHeader('Content-Type', APP_JSON);
   }
+  // using res.json would also automatically try to set the Content-Type header, which fails if headers are sent
+  res.end(JSON.stringify({ code, error, details }));
 };
 
 const respond = (req, res, code, message, details) => {
--- a/api/src/server-utils.js
+++ b/api/src/server-utils.js
@@ -6,7 +6,7 @@ const logger = require('./logger');
 const MEDIC_BASIC_AUTH = 'Basic realm="Medic Web Services"';
 const cookie = require('./services/cookie');
 
-const wantsJSON = req => req.accepts('json');
+const wantsJSON = req => req.get('Accept') === 'application/json';
 
 const writeJSON = (res, code, error, details) => {
   if (!res.headersSent) {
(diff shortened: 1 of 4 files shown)


dianabarsan:
Checking PouchDb code, I didn't find a clean way into tricking it to dump this changes request and try another. Everywhere, they nullcheck `response.results` and consider empty results if the property is not present - which means that the "replication" chain is interrupted - because it thinks there are no more changes to get - which in turn poses a problem for us, since the app will report successful replication and mislead users. 

Instead, I added a changes response validation check in the `fetch` "middleware" in webapp. 

Please let me know what you think!

api/tests/mocha/controllers/login.spec.js
```diff
@@ -55,6 +56,7 @@ describe('login controller', () => {
     sinon.stub(environment, 'isTesting').get(() => false);
 
     sinon.stub(auth, 'isOnlineOnly').returns(false);
+    sinon.stub(privacyPolicy, 'exists').resolves(false);
```
dianabarsan, line 59: These tests failed for me locally.

webapp/src/ts/constants.ts
```diff
@@ -6,6 +6,25 @@
 // go and if it's still too long then API will respond with a 413.
 const BODY_LENGTH_LIMIT = 32000000; // 32 million
 
+// To keep long _changes requests open, API sends a heartbeat (a new line) every 10 seconds, starting on the 10sh
+// second. Sending data through automatically adds a success 200 status, even though the results of the request is
+// not known, and it can still fail.
+// PouchDb only checks the request status to determine whether the request is successful or not. \
+// In order to enable a retry, and not incorrectly communicate success to the client, validate _changes request
+// responses and update the respons object accordingly before passing it to PouchDb for processing.
```
dianabarsan, line 14: I noticed the typos in this comment in a re-read. To optimize on CI time, I'm going to push the fixes to typos along with the review feedback.

api/src/server-utils.js
```diff
@@ -6,13 +6,17 @@ const logger = require('./logger');
 const MEDIC_BASIC_AUTH = 'Basic realm="Medic Web Services"';
 const cookie = require('./services/cookie');
 
-const wantsJSON = req => req.get('Accept') === 'application/json';
+const APP_JSON = 'application/json';
+
+const wantsJSON = req => req.get('Accept') === APP_JSON;
```
garethbowen, line 11: It'd be slightly cleaner to use [`req.accepts`](https://expressjs.com/en/api.html#req.accepts) and [`res.type`](https://expressjs.com/en/api.html#res.type), or even [`res.format`](https://expressjs.com/en/api.html#res.format) if you're feeling ambitious!

webapp/src/ts/constants.ts
```diff
@@ -26,7 +45,10 @@ export const POUCHDB_OPTIONS = {
         opts.headers.set(header, POUCHDB_OPTIONS.remote_headers[header]);
       });
       opts.credentials = 'same-origin';
-      return window.PouchDB.fetch(url, opts);
+      const isChangesRequest = parsedUrl.pathname.startsWith('/_changes');
+      return window.PouchDB
+        .fetch(url, opts)
+        .then(response => isChangesRequest ? processChangesResponse(response) : response);
```
garethbowen, line 51: I'd feel a little more comfortable not adding the `.then` if we're not going to do anything, eg:

```suggestion
      const promise = window.PouchDB.fetch(url, opts);
      if (parsedUrl.pathname.startsWith('/_changes')) {
        promise.then(response => processChangesResponse(response));
      }
      return promise;
```

garethbowen (changes requested):
Nice approach. Shame we had to do it!

A couple of comments inline but no blockers.

api/src/server-utils.js
dianabarsan: There's one issue with `req.accepts`. In the absence of an `Accepts` header, `req.accepts` will always return truthy, as anything passes as accepted, which is a bit different from what we're doing, where we specifically require the Accepts header to include 'application/json'. 

This changes behavior a little, including one e2e test that test for responses when json headers are not sent. 

Please let me know if you want to relax the accepts behavior for returning errors to requests.

garethbowen: Let's not change the default behaviour, just in case...

res.format solves this by calling the first option. I haven't tested this but it might work..

```js
const writeHtml = (res, code, message, details) => {
  if (!res.headersSent) {
    res.writeHead(code, {
      'Content-Type': 'text/plain',
    });
  }
  if (message.message) {
    message = message.message;
  }
  if (typeof message !== 'string') {
    message = JSON.stringify(message);
  }
  if (details) {
    message += ': ' + JSON.stringify(details);
  }
  res.end(message);
};

const respond = (req, res, code, message, details) => {
  res.format({
    'text/html': () => {
      writeHtml(res, code, message, details);
    },
    'application/json': () => {
      writeJSON(res, code, message, details);
    }
  });
};
```

I'll leave it up to you whether that's better or not...

dianabarsan: From the documentation https://expressjs.com/en/api.html#res.format : 
```
The Content-Type response header is set when a callback is selected. However, you may alter this within the callback using methods such as res.set() or res.type().
```

garethbowen: D'oh! Ok. What you had is perfect...

garethbowen (approved):
Awesome. Thanks for indulging me!

PR #7766, opened 2022-09-06
Run e2e tests through nginx

garethbowen:
# Description

# Code review checklist

# License

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,8 +72,105 @@ jobs:
         path: images/
       if: ${{ !env.INTERNAL_CONTRIBUTOR }}
 
-  publish:
+  config-tests:
     needs: build
+    name: ${{ matrix.grunt-cmd }}
+    runs-on: ubuntu-18.04
+    strategy:
+      fail-fast: false
+      matrix:
+        grunt-cmd: [ 'exec:test-config-standard', 'exec:test-config-default']
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js 16.x
+      uses: actions/setup-node@v1
+      with:
+        node-version: 16.x
+    - name: Grunt Install
+      run: npm install -g grunt-cli
+    - run: npm ci --legacy-peer-deps
+    - name: Run Tests
+      run: node --stack_size=10000 `which grunt` ${{ matrix.grunt-cmd }}
+
+  tests:
+    needs: build
+    name: ${{ matrix.grunt-cmd }}
+    runs-on: ubuntu-18.04
+    env:
+      NODE: 16.x
+
+    strategy:
+      fail-fast: false
+      matrix:
+        grunt-cmd: ['ci-webdriver-default', 'ci-webdriver-standard', 'ci-e2e-integration', 'ci-e2e', 'ci-e2e-mobile', 'ci-e2e-cht']
+
+    steps:
+    - name: Login to Docker Hub
+      uses: docker/login-action@v1
+      with:
+        username: ${{ env.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+      if: ${{ env.INTERNAL_CONTRIBUTOR }}
+
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v1
+      with:
+        aws-access-key-id: ${{ secrets.ECR_ACCESS_KEY_ID }}
+        aws-secret-access-key: ${{ secrets.ECR_SECRET_ACCESS_KEY }}
+        aws-region: eu-west-2
+      if: ${{ env.INTERNAL_CONTRIBUTOR }}
+
+    - name: Login to Amazon ECR
+      id: login-ecr
+      uses: aws-actions/amazon-ecr-login@v1
+      if: ${{ env.INTERNAL_CONTRIBUTOR }}
+
+    - name: Get branch name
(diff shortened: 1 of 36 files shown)


.github/workflows/build.yml
```diff
@@ -92,54 +93,16 @@ jobs:
...
-      run: |
-        echo "BUILD_NUMBER=$GITHUB_RUN_ID" >> $GITHUB_ENV
-        echo "BRANCH=$BRANCH_NAME" >> $GITHUB_ENV
-    - name: Publish
-      run: |
-        cd scripts/ci
-        npm ci --legacy-peer-deps
-        node ./publish.js
-        node ./tag-docker-images.js
-
```
garethbowen, line 132: Moved this block down, to where it runs chronologically.

api/src/controllers/forms.js
```diff
@@ -27,11 +28,15 @@ const isCollectForm = doc => doc && doc.context && doc.context.collect;
...
+  const xmls = forms
     .filter(form => isCollectForm(form))
-    .map(form => `${req.protocol}://${req.headers.host}/api/v1/forms/${form.internalId}.xml`);
+    .map(form => formsService.getXFormAttachment(form))
+    .filter(attachment => !!attachment)
+    .map(attachment => Readable.from(attachment.data));
+
   return new Promise((resolve, reject) => {
-    openrosaFormList(urls, (err, xml) => {
+    openrosaFormList(xmls, { downloadUrl }, (err, xml) => {
```
garethbowen, line 39: Passing the xmls means openrosaFormList doesn't have to request them which is a problem now that it's inside cht-net.

api/src/controllers/forms.js
```diff
@@ -51,6 +56,9 @@ const listFormsJSON = forms => {
 
 module.exports = {
   list: (req, res) => {
+    if (!req.userCtx) {
+      return serverUtils.notLoggedIn(req, res);
+    }
```
garethbowen, line 61: I've made auth required. This feels like a regression but a test asserts that no auth fails: https://github.com/medic/cht-core/blob/master/tests/integration/medic-collect/medic-collect.spec.js#L41

tests/e2e/api/controllers/_changes.spec.js
```diff
@@ -389,33 +386,10 @@ describe('changes handler', () => {
...
-        res.on('end', () => {
-          body = JSON.parse(body);
-          shouldBatchChangesRequests = semver.lte(couchVersionForBatching, body.version);
-        });
-      });
-
-      req.on('error', e => e);
-      req.end();
-    });
-
```
garethbowen, line 418: Dropped due to direct Couch access being blocked. This could be refactored to use another method but I don't think we need to support couch 2.3.0 any more do we?

garethbowen:
Review please @dianabarsan 

It's timing out on the upgrade test but I think that's because it's running the _old_ API on 5988 so the check of port 443 fails. I think once we merge this we should change that test so it upgrades a version that uses nginx to another version that uses nginx.

dianabarsan (approved):
Great stuff! Thanks for cleaning things up. All my comments are just really small nits.

api/src/controllers/forms.js
```diff
@@ -27,11 +28,15 @@ const isCollectForm = doc => doc && doc.context && doc.context.collect;
 // @param {Object} data - couchdb view data
 // @param {Object} req - the request object
 const listFormsXML = (forms, req) => {
-  const urls = forms
+  const downloadUrl = `${req.protocol}://${req.headers.host}/api/v1/forms/` + '${formId}.xml'; // formId is replaced by openrosaFormList
```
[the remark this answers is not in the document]
garethbowen: I know :(

tests/apply-config.js
```diff
@@ -17,13 +16,13 @@ const defaultActions = [
 ];
 
 const applyConfig = async path => {
-  const COUCH_URL = `http://${auth.username}:${auth.password}@${constants.API_HOST}:${constants.API_PORT}/${constants.DB_NAME}`;
+  const couchUrl = `${constants.BASE_URL_AUTH}/${constants.DB_NAME}`;
```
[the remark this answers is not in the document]
garethbowen: Done.

tests/e2e/api/controllers/_changes.spec.js
garethbowen: Ok. I've removed the associated production code too!

garethbowen:
@dianabarsan Dropping support for CouchDB 2.2 meant I could rip out a bunch of production code, so if you wouldn't mind reviewing the change please?

Also, did you see this comment:

> [The build is] timing out on the upgrade test but I think that's because it's running the old API on 5988 so the check of port 443 fails. I think once we merge this we should change that test so it upgrades a version that uses nginx to another version that uses nginx. Suggestions welcome!

dianabarsan:
> Also, did you see this comment:

Yes, I did. The way I've been fixing that was by duping the current branch and changing the source branch for the upgrade for the duplicate branch.

dianabarsan:
Re: upgrade, it just occurred to me that the default certificate mode that nginx uses is `own_cert`, which means that if cht-upgrade-service doesn't pass the environment variable to change to a self signed certificate, then nginx won't start. I'm going to push the change to `cht-upgrade-service` to forward nginx env, but we also need to either: change the default value of the env variable or pass the correct value in the upgrade e2e test.

dianabarsan (approved):
Awesome! Thanks for taking the time.

Issue #7768, opened 2022-09-06
Fix webapp enketo multimedia unit tests 


dianabarsan:
**Describe the issue**
[These tests](https://github.com/medic/cht-core/blob/master/webapp/tests/karma/ts/services/enketo.service.spec.ts#L1840) have been disabled because they were [blocking the following unit tests](https://github.com/medic/cht-core/issues/7765).

**Describe the improvement you'd like**
Fix the unit tests in such a way that they don't cause page reloads and stop karma flow.

PR #7774, opened 2022-09-09
7768 - Fix unit tests for multimedia forms

latin-panda:
# Description

This PR fixes unit tests for multimedia forms.
- Adds validation to prevent navigation when using `windows.history` and `form.pages._getCurrentIndex()` returns undefined. This mainly because we want to test the feature and karma doesn't do well on navigation. Also, we can't mock `windows.history` because it's ready only. 
- Enables unit tests that were breaking karma

https://github.com/medic/cht-core/issues/7768

Screen Shot 2022-09-12 at 3 12 19 pm

# Code review checklist

# License

### DIFF
--- a/webapp/src/ts/services/enketo.service.ts
+++ b/webapp/src/ts/services/enketo.service.ts
@@ -340,8 +340,10 @@ export class EnketoService {
           .then(valid => {
             if (valid) {
               const currentIndex = form.pages._getCurrentIndex();
-              window.history.pushState({ enketo_page_number: currentIndex }, '');
-              this.setupNavButtons($wrapper, currentIndex);
+              if (currentIndex !== undefined) {
+                window.history.pushState({ enketo_page_number: currentIndex }, '');
+                this.setupNavButtons($wrapper, currentIndex);
+              }
               this.pauseMultimedia($wrapper);
             }
             this.forceRecalculate(form);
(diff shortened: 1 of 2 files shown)


latin-panda:
@dianabarsan Can you please have a look? I will update the branch once the e2e fix is in master. 
Thanks!

webapp/src/ts/services/enketo.service.ts
```diff
@@ -340,8 +340,10 @@ export class EnketoService {
           .then(valid => {
             if (valid) {
               const currentIndex = form.pages._getCurrentIndex();
-              window.history.pushState({ enketo_page_number: currentIndex }, '');
-              this.setupNavButtons($wrapper, currentIndex);
+              if (currentIndex !== undefined) {
```
dianabarsan, line 343: I'm usually reluctant to change app code just to overcome test limitations. Is there a way mock `_getCurrentIndex` so this change isn;t necessary?

latin-panda: even if I make it return something, the `window.history` executes and karma doesn't like it

latin-panda: actually it's just the `window.history.back`, 1 sec. I'm going to push a second version

latin-panda: Okay, I've reverted the validation and now it just updates the test file. Pushed!
Screen Shot 2022-09-12 at 6 30 22 pm

dianabarsan (approved):
Awesome! Thanks for the fix!

PR #7776, opened 2022-09-09
Add e2e and integration tests for user_replace

jkuester:
## transitions
- Updates the error handling for the transition so that the transitions framework will save any errors to the `replace_user` report, itself, if they occur during the transition (by setting `err.changed = true` before throwing it).
- Adds init logic to fail during Sentinel initialization if the `user_replace` transition is enabled, but token logic is not (since without token login, the new user would not be able to login...)
- Adds unit tests

## api
- Adds some additional error checking to validate the new contact has `name` and `phone` fields
- Updates the re-parenting logic to only change forms with `reported_date` that is `>` the `replace_user` form (instead of `>=`).  This will prevent the `replace_user` form, itself, from getting re-parented, which causes it to re-trigger the transition again

## config/default
- Make the `user_replace` transition `false` by default in the default app_settings.  With the new init validation in the transition, we need to either enable `token_login` by default or we need the `user_replace`transition disabled by default.  

## e2e test
- Adds new e2e test (in `tests/e2e/forms`) that runs through a complete workflow by submitting a new `replace_user` form and validating that the transition happens as expected

## integration tests

- Adds new integration tests (in `tests/e2e/sentinel/transitions`) that thoroughly test the logic of the Sentinel transition


tests/utils.js
```diff
@@ -139,6 +139,7 @@ const deleteAll = (except) => {
     'branding',
     'partners',
     'settings',
+    'form:replace_user',
```
jkuester, line 142: FYI, this is kind of a hack that we need for now.  Without this, the form will be deleted by some of the existing tests.  I know that QA ran into the same problem in some of their tests and are working on a better solution, but for now this is the easiest way to  keep everything working.  We can change this to leverage QA's solution when we merge to `master`.

Issue #7731, opened 2022-08-16
Update `today` XPath function to not return current time


jkuester:
**What feature do you want to improve?**
Currently we have a [custom implementation](https://github.com/medic/cht-core/blob/3d00f5ffc34510ce43dae26190ef41f57ea01154/webapp/src/js/enketo/medic-xpath-extensions.js#L86) of the `today` functionality that is identical to what is returned by the `now` function (the current date and time in the current timezone).  

**Describe the improvement you'd like**
As noted in this OpenRosa issue (https://github.com/enketo/openrosa-xpath-evaluator/issues/27), this behavior does not seem to follow the [ODK Spec](https://getodk.github.io/xforms-spec/#fn:today).  Unfortunately, the wording in the spec is a little confusing (to me at least) with regards to how it is actually supposed to behave.  However, both [openrosa](https://github.com/enketo/openrosa-xpath-evaluator/blob/01648be20c638f773467e299a9b65a050a409a97/src/openrosa-extensions.js#L302) and [javarosa](https://github.com/getodk/javarosa/blob/b72e0d48947c7f2b5c2d05ab9783ea9b8a259b04/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java#L333) have coalesced on a single interpretation of the spec where `now` returns the current date and time while `today` returns the current date with the time set to midnight of the current timezone.  I think we should remove our custom implementation of `today` and just use the one provided by OpenRosa.

As things stand today, we have a bunch of forms that are doing things like `date(floor(decimal-date-time(today())))` to get the start of today.  All this extra logic could be eliminated with the new implementation of `today`.  Any logic that relies on the current time being included in `today` can just update to the spec-compliant `now` function.  

As discussed in places like [here](https://github.com/medic/cht-core/issues/6345#issuecomment-1152800834), [here](https://github.com/medic/cht-core/issues/6345#issuecomment-1155647161), and [here](https://github.com/medic/cht-core/issues/6345#issuecomment-1201489355) the [Enketo uplift](https://github.com/medic/cht-core/pull/7256) is already going to cause subtle behavior changes in the behavior of `format-date` and `decimal-date` (both frequently used along with `today`). Changing `today` at the same time as these other changes would allow projects to implement proper fixes to their date handling once and for all instead of breaking some things now with the Enketo uplift and again later if we decide to switch to a spec-compliant version of `today`.  

**Describe alternatives you've considered**
[...]

PR #7821, opened 2022-09-23
Remove custom today XPath function

jkuester:
# Description

This PR removes our custom `today` XPath function (allowing us to inherit the default functionality from the OpenRosa `today` function.  The difference is that our custom implementation of `today` would include the _current time value_ while OpenRosa's default implementation (which matches JavaRosa) is that `today` only returns today's date at _midnight in the current timezone_.  

## Form changes: 

In places where we were using `today`, If we do not actually care about the time value, I just let it be. If it seemed like the time value _might_ be relevant, I switched to use `now`.  When today was being wrapped with a `floor` call, I just removed the `floor` since it should no longer be necessary (see the discussion on the issue for more details about this).  Finally, I also tried to remove unnecessary calls to `decimal-date-time` as I went. We were using these a lot when comparing date values, but this is unnecessary. (Enketo supports comparing date values with `<>=`. They do not have to be converted to numbers.)

Closes https://github.com/medic/cht-core/issues/7731

# Code review checklist

# License


jkuester:
@njogz good call. I don't think there is an issue, but I will be sure to include this in the documentation changes we are adding in https://github.com/medic/cht-docs/pull/802

Issue #7832, opened 2022-10-04
CERTIFICATE_MODE=OWN_CERT fails with "cannot load certificate"


mrjones-plip:
**Describe the bug**
In Arch v3 you can specify you're bringing your own TLS cert by setting `CERTIFICATE_MODE` to `OWN_CERT`.  This will pass our test for [files being present](https://github.com/medic/cht-core/blob/69b478371091c0fa3cac00003e04af052050346d/nginx/ssl-install.sh#L97), but because we [hard code the path to the certs](https://github.com/medic/cht-core/blob/69b478371091c0fa3cac00003e04af052050346d/nginx/nginx.conf#L42), nginx fails to start.

**To Reproduce**
Steps to reproduce the behavior assuming no existing docker containers, volumes or networks:
1. Download the current `master` [docker compose file for core](https://staging.dev.medicmobile.org/_couch/builds/medic%3Amedic%3Amaster/docker-compose%2Fcht-core.yml): `curl -o docker-compose_cht-core.yml https://staging.dev.medicmobile.org/_couch/builds/medic%3Amedic%3Amaster/docker-compose%2Fcht-core.yml`
2. Download the current `master` [docker compose file for couch](https://staging.dev.medicmobile.org/_couch/builds/medic%3Amedic%3Amaster/docker-compose%2Fcht-couchdb.yml): `curl -o docker-compose_cht-couchdb.yml https://staging.dev.medicmobile.org/_couch/builds/medic%3Amedic%3Amaster/docker-compose%2Fcht-couchdb.yml`
3. Create an empty `cht-ssl` volume `docker volume create cht-ssl`
4. Populate the volume with valid local-host.ip certs:
   	```
	docker run --rm -d -v cht-ssl:/root --name deleteme alpine sleep 3600
	docker exec -it deleteme sh -c "wget -P /root/  http://local-ip.co/cert/server.pem"
	docker exec -it deleteme sh -c "wget -P /root/ http://local-ip.co/cert/chain.pem"
	docker exec -it deleteme sh -c "cat  /root/server.pem /root/chain.pem > /root/cert.pem"
	docker exec -it deleteme sh -c "wget -P /root/ http://local-ip.co/cert/server.key"
	docker exec -it deleteme sh -c "mv /root/server.key /root/key.pem"
	docker stop deleteme
   	```
[...]

PR #7834, opened 2022-10-05
Fix bringing own certs 

mrjones-plip:
# Description

1 line fix to allow passing in an env var to set the path for the mount point for `cht-nginx` service

medic/cht-core#7832

# Code review checklist

# License

### DIFF
--- a/scripts/build/cht-core.yml.template
+++ b/scripts/build/cht-core.yml.template
@@ -81,7 +81,7 @@ services:
       - "${NGINX_HTTP_PORT:-80}:80"
       - "${NGINX_HTTPS_PORT:-443}:443"
     volumes:
-      - cht-ssl:/root/.acme.sh/
+      - cht-ssl:"${SSL_VOLUME_MOUNT_PATH:-/root/.acme.sh/}"
     environment:
       - "CERTIFICATE_MODE=${CERTIFICATE_MODE:-SELF_SIGNED}"
       - "SSL_CERT_FILE_PATH=${SSL_CERT_FILE_PATH:-/etc/nginx/private/cert.pem}"
--- a/scripts/build/cht-core.yml.template
+++ b/scripts/build/cht-core.yml.template
@@ -81,7 +81,7 @@ services:
       - "${NGINX_HTTP_PORT:-80}:80"
       - "${NGINX_HTTPS_PORT:-443}:443"
     volumes:
-      - cht-ssl:"${SSL_VOLUME_MOUNT_PATH:-/root/.acme.sh/}"
+      - cht-ssl:${SSL_VOLUME_MOUNT_PATH:-/root/.acme.sh/}
     environment:
       - "CERTIFICATE_MODE=${CERTIFICATE_MODE:-SELF_SIGNED}"
       - "SSL_CERT_FILE_PATH=${SSL_CERT_FILE_PATH:-/etc/nginx/private/cert.pem}"


mrjones-plip:
Steps to test:

Shared steps for "Self cert still works" and "own  cert  works"

1. Download the current `master` [docker compose file for core](https://staging.dev.medicmobile.org/_couch/builds/medic%3Amedic%3Amaster/docker-compose%2Fcht-core.yml): `curl -o docker-compose_cht-core.yml https://staging.dev.medicmobile.org/_couch/builds/medic%3Amedic%3Amaster/docker-compose%2Fcht-core.yml`
2. Download the current `master` [docker compose file for couch](https://staging.dev.medicmobile.org/_couch/builds/medic%3Amedic%3Amaster/docker-compose%2Fcht-couchdb.yml): `curl -o docker-compose_cht-couchdb.yml https://staging.dev.medicmobile.org/_couch/builds/medic%3Amedic%3Amaster/docker-compose%2Fcht-couchdb.yml`
3. Create an empty `cht-ssl` volume `docker volume create cht-ssl`
4. Populate the volume with valid local-host.ip certs:
   	```
	docker run --rm -d -v cht-ssl:/root --name deleteme alpine sleep 3600
	docker exec -it deleteme sh -c "wget -P /root/  http://local-ip.co/cert/server.pem"
	docker exec -it deleteme sh -c "wget -P /root/ http://local-ip.co/cert/chain.pem"
	docker exec -it deleteme sh -c "cat  /root/server.pem /root/chain.pem > /root/cert.pem"
	docker exec -it deleteme sh -c "wget -P /root/ http://local-ip.co/cert/server.key"
	docker exec -it deleteme sh -c "mv /root/server.key /root/key.pem"
	docker stop deleteme
   	```

### Tests pass

This branch's build should pass CI

### Self cert still works

After following 4 steps above, run: `COUCHDB_PASSWORD=password COUCHDB_USER=medic docker-compose -f docker-compose_cht-core.yml -f docker-compose_cht-couchdb.yml up` (this defaults to `SELF_SIGNED` for `CERTIFICATE_MODE`)

Validate port `443` is using self signed cert

run this command to destroy all your environment, including  networks and volumes:  `COUCHDB_PASSWORD=password COUCHDB_USER=medic docker-compose -f docker-compose_cht-core.yml -f docker-compose_cht-couchdb.yml down -v`

### own  cert  works

After following 4 steps above, run: `SSL_VOLUME_MOUNT_PATH=/etc/nginx/private/ CERTIFICATE_MODE=OWN_CERT COUCHDB_PASSWORD=password COUCHDB_USER=medic docker-compose -f docker-compose_cht-core.yml -f docker-compose_cht-couchdb.yml up` (this explicitly sets to `OWN_CERT` for `CERTIFICATE_MODE` and tests setting the new env var `SSL_VOLUME_MOUNT_PATH` to match what `nginx` is expecting)

Validate port `443` is using locl-ip.co's cert

run this command to destroy all your environment, including  networks and volumes:  `COUCHDB_PASSWORD=password COUCHDB_USER=medic docker-compose -f docker-compose_cht-core.yml -f docker-compose_cht-couchdb.yml down -v`

scripts/build/cht-core.yml.template
```diff
@@ -81,7 +81,7 @@ services:
       - "${NGINX_HTTP_PORT:-80}:80"
       - "${NGINX_HTTPS_PORT:-443}:443"
     volumes:
-      - cht-ssl:/root/.acme.sh/
+      - cht-ssl:"${SSL_VOLUME_MOUNT_PATH:-/root/.acme.sh/}"
```
dianabarsan, line 84: I had to remove the quotes around the value, I got this error: 
```
invalid mount config for type "volume": invalid mount path: '"/etc/nginx/private/"' mount path must be absolute
```

dianabarsan (commented):
Nice quick fix! 

How does this work with the change where we use compose project name and expect the volume name to be dynamic?

scripts/build/cht-core.yml.template
mrjones-plip: yike! good catch - thanks. Fixed in 90bcda4ed

I'd caught this on the compose file I'd mocked up, but not on the template I'd committed. Is there an easy way to render the `*.yml.template` -> `*.yml` to test the templates being converted to real compose files?

mrjones-plip:
> How does this work with the change where we use compose project name and expect the volume name to be dynamic?

While getting your cert into the volume will be trickier - it still works as expected!  We can also smooth this over with a compose file that will respond to both the `SSL_VOLUME_MOUNT_PATH` and `COMPOSE_PROJECT_NAME` env vars.

To test this, I wrote up this compose file and updated it in [step 3 above](https://github.com/medic/cht-core/pull/7834#issuecomment-1268710481).   I was able to run these both and it worked exactly as expected:

Bare:
1.  `SSL_VOLUME_MOUNT_PATH=/etc/nginx/private/ docker-compose -f docker-compose_load-certs.yml up`
2. `SSL_VOLUME_MOUNT_PATH=/etc/nginx/private/ CERTIFICATE_MODE=OWN_CERT COUCHDB_PASSWORD=password COUCHDB_USER=medic docker-compose -f docker-compose_cht-core.yml -f docker-compose_cht-couchdb.yml up`

With Project:
1.  `COMPOSE_PROJECT_NAME=testeee SSL_VOLUME_MOUNT_PATH=/etc/nginx/private/ docker-compose -f docker-compose_load-certs.yml up`
2. `COMPOSE_PROJECT_NAME=testeee SSL_VOLUME_MOUNT_PATH=/etc/nginx/private/ CERTIFICATE_MODE=OWN_CERT COUCHDB_PASSWORD=password COUCHDB_USER=medic docker-compose -f docker-compose_cht-core.yml -f docker-compose_cht-couchdb.yml up`

TIL that if the volume doesn't exist, both `docker` and `docker-compose` (and thus `docker compose`) will create it for you on first run!

dianabarsan:
So when you created the volume manually, you named it `testeee-cht-ssl-1` ?

mrjones-plip:
> So when you created the volume manually, you named it testeee-cht-ssl-1 ?

I didn't create it manually.  Docker will implicitly create a volume for you if it it doesn't exist.  If you don't explicitly define the name and `COMPOSE_PROJECT_NAME` isn't set, it will prepend the directory name.  If you do set `COMPOSE_PROJECT_NAME`, it will prepend the project name instead of the directory name.  There isn't a suffix added, unlike containers which will get an incrementing number (eg `_1`).

Here's an example video showing the the "Bare" and "With Project" [step 1 above](https://github.com/medic/cht-core/pull/7834#issuecomment-1270743324) running the `docker-compose_load-certs.yml` file :

https://user-images.githubusercontent.com/8253488/194603389-dbda6247-6a6a-455e-8685-a65776fb79c0.mp4

mrjones-plip:
Oh yes! If I had declared the `volume` stanza like this:

```yaml
volumes:
    cht-ssl:
      name: cht-ssl
```

_Any_ service that mounted the volume, no matter the parent directory or the project name, would always mount literally the same volume with the name `cht-ssl`

dianabarsan:
So, to recap: 
- to use OWN_CERTS you most likely need to create the volume manually and add the certs
- to use COMPOSE_PROJECT_NAME and not set a global volume name (and have multiple installs of cht work at the same time on the same machine), you need to create this volume in such a way that it will match the name that the future containers will use. 

Am I covering everything?

mrjones-plip:
> to use OWN_CERTS you most likely need to create the volume manually and add the certs

Correct - since `cht-nginx` fails if the certs aren't there, we need to bootstrap it before hand.  You could run with self signed certs, copy your own certs in, and then restart it in own certs mode I guess :shrug: 

> to use COMPOSE_PROJECT_NAME and not set a global volume name (and have multiple installs of cht work at the same time on the same machine), you need to create this volume in such a way that it will match the name that the future containers will use.

precisely! 

I think the easiest way to do both is the `docker-compose_load-certs.yml` which uses the same env vars as core and docker compose files, but don't feel strongly about this!

dianabarsan:
Is it an option to just mount an existent folder where users can place the own certs?

mrjones-plip:
> Is it an option to just mount an existent folder where users can place the own certs?

Oh - likely! I didn't want to impose too much on the design decisions made already. I'll look into this next week - I'm out Monday, so Tuesday!

dianabarsan:
Yes, discussing options about how to handle this was on our agenda, I would lean towards mounting an existent volume for simplicity, but @Hareet  had some concerns. 

I also quite like the compose file that loads certs, but I think it writing it will require some more advanced docker skills from self hosted deployments, unless we can provide some templates for them. Also depends, I suppose, on the feasibility of starting all containers under the same project name. 

I'm approving this to unblock you from using nginx local ip.

dianabarsan (approved):

mrjones-plip:
> Yes, discussing options about how to handle this was on our agenda, I would lean towards mounting an existent volume for simplicity, but @Hareet had some concerns.

Using a `bind` mount (eg mounting and existent volume on the host) isn't a best practice for docker.  I don't feel too strongly about this and would defer to Hareet et SRE al. for best practice here.

> I'm approving this to unblock you from using nginx local ip.

thank you! Really appreciate all the thought and constructive criticism on this PR. I'm open to coming back to how we do this if we have any concerns later on.

Issue #7840, opened 2022-10-10
API credentials endpoint fails CouchDb secret is too short


dianabarsan:
**Important**: This is a public repository. Anyone in the world can see what's posted here. If you are posting screenshots or log files, please **carefully examine them for** the presence of any kind of **protected health information** (PHI). Images or logs containing PHI _must_ be posted in fully-redacted form, with no visible PHI.

**Describe the bug**
The `/api/v1/credentials` fails with 500 when trying to set any credentials, when CouchDb secret is too short. 

**To Reproduce**
1. Set your CouchDb secret to be very short (mine was "monkey")
2. Try to PUT some credentials through API (/api/v1/credentials)
3. Response is 500

**Expected behavior**
Credentials should be stored even if the "secret" is short. 
We had the same issue before, when CouchDb secret was too long, but that was fixed before merging the PR. (https://github.com/medic/cht-core/pull/7577#discussion_r894495591)

**Logs**
```
2022-10-10 10:06:53 [33mWARN[39m: Non-numeric error code: ERR_CRYPTO_INVALID_KEYLEN 
2022-10-10 10:06:53 [31mERROR[39m: Server error: RangeError: Invalid key length
    at Cipheriv.createCipherBase (node:internal/crypto/cipher:116:19)
    at Cipheriv.createCipherWithIV (node:internal/crypto/cipher:135:3)
    at new Cipheriv (node:internal/crypto/cipher:243:3)
    at Object.createCipheriv (node:crypto:138:10)
    at /shared-libs/settings/src/index.js:51:29
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Promise.all (index 1) {
  [stack]: 'RangeError: Invalid key length\n' +
    '    at Cipheriv.createCipherBase (node:internal/crypto/cipher:116:19)\n' +
    '    at Cipheriv.createCipherWithIV (node:internal/crypto/cipher:135:3)\n' +
    '    at new Cipheriv (node:internal/crypto/cipher:243:3)\n' +
    '    at Object.createCipheriv (node:crypto:138:10)\n' +
    '    at /shared-libs/settings/src/index.js:51:29\n' +
    '    at runMicrotasks (<anonymous>)\n' +
    '    at processTicksAndRejections (node:internal/process/task_queues:96:5)\n' +
    '    at async Promise.all (index 1)',
  [message]: 'Invalid key length',
[...]

PR #7842, opened 2022-10-11
Allow for short CouchDb Secret

dianabarsan:
# Description

When secret is too short, concatenate with itself until we have desired string length. 

medic/cht-core#7840

# Code review checklist

# License

### DIFF
--- a/shared-libs/settings/src/index.js
+++ b/shared-libs/settings/src/index.js
@@ -41,7 +41,16 @@ const getKey = () => {
   const url = `${getCouchConfigUrl()}/couch_httpd_auth/secret`;
   return request
     .get(url, { json: true })
-    .then(key => Buffer.from(key).slice(0, KEY_LENGTH));
+    .then(key => {
+      let buffer = Buffer.from(key);
+      if (buffer.length < KEY_LENGTH) {
+        const duplicateCount = Math.ceil(KEY_LENGTH / buffer.length);
+        const bufferList = Array.from({ length: duplicateCount }).map(() => buffer);
+        buffer = Buffer.concat(bufferList);
+      }
+
+      return buffer.slice(0, KEY_LENGTH);
+    });
 };
 
 const encrypt = (text) => {
--- a/shared-libs/settings/src/index.js
+++ b/shared-libs/settings/src/index.js
@@ -42,6 +42,10 @@ const getKey = () => {
   return request
     .get(url, { json: true })
     .then(key => {
+      if (!key.length) {
+        throw new Error('Invalid cypher. CouchDB Secret needs to be set in order to use secure credentials.');
+      }
+
       let buffer = Buffer.from(key);
       if (buffer.length < KEY_LENGTH) {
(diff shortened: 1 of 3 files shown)


tests/utils.js
```diff
@@ -651,6 +651,7 @@ const dockerComposeCmd = (...params) => {
     DB1_DATA: db1Data,
     DB2_DATA: db2Data,
     DB3_DATA: db3Data,
+    COUCHDB_SECRET: 'monkey',
```
jkuester, line 654: I am trying to understand what setting this does.  Am I correct in thinking that this is overwriting the default envar set in the [docker-compose file](https://github.com/medic/cht-core/blob/bc54098b0aad7397142cf762b7b1a865700c6fdb/couchdb/test.couchdb-cluster.yml#L14)?  

Do we want to leave this set as the short `monkey` value or do you think it would be better to set a more proper example value (or even just remove this at let it continue to use the default from the compose file)?

dianabarsan: The compose file doesn't have a default setting, it was removed a few commits ago.

jkuester: > The compose file doesn't have a default setting, it was removed a few commits ago.

:confused: I see that the default password was removed, but it looks like there is still a [default `COUCHDB_SECRET`](https://github.com/medic/cht-core/blob/6e77c909bc72d4e2f11aadbf9d0cea4593f6c95e/couchdb/test.couchdb-cluster.yml#L14).... (But maybe I am misunderstanding something here!)

dianabarsan:
The length is invalid for our purposes, in this particular case. 
I suppose we could impose a minimum length and fail to start CouchDb if the length is too low. A problem that can arise for existing installations, that already have a short secret, where they could end up not being able to upgrade to 4.0 without losing all active sessions (since they'll be forced to use a different secret). 

Tagging @garethbowen and @Hareet to discuss whether we should impose a minimum couch secret length.

shared-libs/settings/src/index.js
```diff
@@ -41,7 +41,16 @@ const getKey = () => {
   const url = `${getCouchConfigUrl()}/couch_httpd_auth/secret`;
   return request
     .get(url, { json: true })
-    .then(key => Buffer.from(key).slice(0, KEY_LENGTH));
+    .then(key => {
+      let buffer = Buffer.from(key);
+      if (buffer.length < KEY_LENGTH) {
+        const duplicateCount = Math.ceil(KEY_LENGTH / buffer.length);
```
garethbowen, line 47: Is it possible for buffer.length to be 0, causing a divide by zero error?

garethbowen (dismissed):
One potential bug but otherwise all good!

garethbowen:
Let's raise an issue to impose a minimum at startup and we can look at it in a later release.

shared-libs/settings/src/index.js
dianabarsan: I pushed a commit to have the code throw an explicit error when the secret is empty.

garethbowen (approved):
Awesome

PR #7847, opened 2022-10-17
Handle many users linked to same contact when replacing a user

m5r:


jkuester (commented):
Looking good!  I have left a couple comments that I will fix before merging this PR (and added some additional tests). 

The big thing is I think we need a bit more tweaking on top of these changes to handle the fact that multiple users _with the same contact_ might try to be replaced. The current implementation will not support that, we we can address that in a new PR.

PR #7853, opened 2022-10-19
Fix unit tests so they also run in TZs that are < UTC

jkuester:
# Description

Some of the unit test fail for me when I run them locally due to timezone issues. (The TZ of my local system is CST (UTC-6)).  It appears that all the unit test (`grunt unit`) run fine in UTC or in TZs that are > UTC (such as EAT), but they fail in TZ that are < UTC.  

The goal of this PR is to correct those issues so that the unit tests work regardless of the system TZ.

# Code review checklist

# License

### DIFF
--- a/shared-libs/rules-engine/package.json
+++ b/shared-libs/rules-engine/package.json
@@ -4,7 +4,7 @@
   "description": "",
   "main": "src/index.js",
   "scripts": {
-    "test": "mocha test/*.spec.js test/**/*.spec.js",
+    "test": "TZ=UTC mocha test/*.spec.js test/**/*.spec.js",
     "tz": "TZ=Canada/Pacific npm test && TZ=Africa/Monrovia npm test && TZ=Pacific/Auckland npm test"
   },
   "repository": {
--- a/shared-libs/rules-engine/package.json
+++ b/shared-libs/rules-engine/package.json
@@ -4,7 +4,7 @@
   "description": "",
   "main": "src/index.js",
   "scripts": {
-    "test": "TZ=UTC mocha test/*.spec.js test/**/*.spec.js",
+    "test": "mocha test/*.spec.js test/**/*.spec.js",
     "tz": "TZ=Canada/Pacific npm test && TZ=Africa/Monrovia npm test && TZ=Pacific/Auckland npm test"
   },
   "repository": {
--- a/shared-libs/rules-engine/test/integration.spec.js
+++ b/shared-libs/rules-engine/test/integration.spec.js
@@ -146,6 +146,8 @@ describe('Rules Engine Integration Tests', () => {
     // the task is 5 days old when it is discovered
     const NOW = MS_IN_DAY * 5;
     sinon.useFakeTimers(NOW);
+    // Reset the "calculatedDate"
+    await rulesEngine.updateEmissionsFor(['patient']);
 
     await triggerFacilityReminderInReadyState(['patient']);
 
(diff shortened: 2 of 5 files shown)


shared-libs/rules-engine/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "",
   "main": "src/index.js",
   "scripts": {
-    "test": "mocha test/*.spec.js test/**/*.spec.js",
```
jkuester, line 7: So, I have not actually been able to figure this one out yet (hence the hack here).  I got so deep into debugging this without actually being able to figure out what was going on that I wanted to just throw it out here and see if anyone else had ideas.

Basically some of the rules-engine _integration_ tests fail when the TZ is `<UTC`.  So, `UTC` and `EAT` work just fine, but `CDT`, `EST` etc fail (including the built-in `npm run tz` script here...).  For example, [this test](https://github.com/medic/cht-core/blob/0d4c275b088a92edb5b971d4f40aafcde3f93d42/shared-libs/rules-engine/test/integration.spec.js#L150) fails here [in the `triggerFacilityReminderInReadyState` function](https://github.com/medic/cht-core/blob/0d4c275b088a92edb5b971d4f40aafcde3f93d42/shared-libs/rules-engine/test/integration.spec.js#L662) because `db.bulkDocs.callCount` is `2` instead of `1`.  The extra call to `bulkDocs` is coming from a _target_ that is getting triggered in addition to the expected _task_.  

After a bunch of debugging, I have tracked the issue down to [here in the `provider-wireup`](https://github.com/medic/cht-core/blob/6514ce2fc8f5d19b2a680a1aa53355ee737129a8/shared-libs/rules-engine/src/provider-wireup.js#L299) where we get the current interval.  When running with `UTC` I get:

```json
{
  "start": 0,
  "end": 2678399999 '
}
```

but when running with `CDT` I get:

```json
{
"start": 21600000,
"end": 2699999999
}
```

This difference is caused because [the `moment()` call here](https://github.com/medic/cht-core/blob/bc97124153bf4533939281d7dd3bb050187de57c/shared-libs/calendar-interval/src/index.js#L31) is returning the moment relative to the current system TZ (`21600000` is midnight, Jan 1st 1970 CDT).

jkuester:
@dianabarsan would you be able to have a quick look at this whenever you get the chance?  (No rush.)  The behavior involved in the `rules-engine` tests is pretty complex. My changes to those tests make it so that they pass, but I am not 100% confident they are the proper changes that we need...  (All the other test failures were pretty straightforward.)

shared-libs/rules-engine/test/integration.spec.js
```diff
@@ -146,6 +146,8 @@ describe('Rules Engine Integration Tests', () => {
     // the task is 5 days old when it is discovered
     const NOW = MS_IN_DAY * 5;
     sinon.useFakeTimers(NOW);
+    // Reset the "calculatedDate"
+    await rulesEngine.updateEmissionsFor(['patient']);
```
dianabarsan, line 150: I think this might actually nullify the test. The patient should get marked as "dirty" by time passing (the useFakeTimers call simulates time passing).

dianabarsan: I changed my timezone to Phoenix Arizona and checked the exact error here. 
The errors that I got were because I got extra documents written in this test, namely targets. This means that, in the setup (beforeEach) where3 we actually do an `updateEmissionsFor`, we were on a different reporting date than in the test. Which makes sense, because the ts is 1 and your TS is with -. 

I fixed the tests locally by changing the starting date to be `MS_IN_DAY`, so we make sure that _everyone_ is in the same reporting period and targets don't get recalculated. 
As a result, I had to change more timestamps across the file, but no logic-altering changes were needed. 

To be very clear, this is the line that I changed: 
https://github.com/medic/cht-core/blob/0d4c275b088a92edb5b971d4f40aafcde3f93d42/shared-libs/rules-engine/test/integration.spec.js#L110

dianabarsan (changes requested):
Thanks a lot for picking this! The sms tests were failing for me as well. 
I left one comment about the rules engine tests

shared-libs/rules-engine/test/integration.spec.js
jkuester: Ah, yes this is exactly the kind of thing I was worried about.  Your `MS_IN_DAY` fix seems to work perfectly!  Thanks!

dianabarsan (approved):
Awesome!

Issue #7156, opened 2021-06-19
Specialized supervisor roles: add ability to filter replicated needs_signoff reports by form type


dianabarsan:
**Is your feature request related to a problem? Please describe.**
With our current supervisor support, we don't have the ability to create specialized supervisors, who should just oversee one or part of the workflows. There's no way to make a specialized overseer supervisor role. 

**Describe the solution you'd like**
Add a new config option to `replication_depth` to contain a list of form internal ids. When this new config exists, for affected role, limit replication of needs_signoff reports to just ones that match the listed form internal ids. 
Additionally, we could also allow this list to contain "negated" form internal ids, to stop a role from replicating _just some_ needs signoff reports. 

**Describe alternatives you've considered**
None.

**Additional context**
Recently, one large deployment started to implement a new workflow, which involves a change in hierarchy to add an additional layer near the top and a new supervisor role that should just oversee this new workflow, sitting at this new top layer. This deployment already has supervisors on lower layers and already uses needs_signoff reports for existent workflows.

garethbowen:
There is a related use case about restricting the types of reports that CHWs can replicate, for example, HIV workflows may be restricted to specialists.

Ideally the replication_depth configuration would use permissions rather than form IDs to make them easier to group together.

dianabarsan:
Using permissions to define replicated docs is going to create an indirect, hard to grasp relation between the two, and I'm afraid it will create situations where replication "rules" for a role are changed without people being aware that users need to resync to get the correct docs. 
Unless we create some mechanism where replication knows how to handle this (by resetting the checkpointer and resyncing), I'd be reluctant to create this indirect relation.

garethbowen:
Yeah for sure. Let's do that too!

jonathanbataire:
@dianabarsan' s suggestion to improve supervisor contact rep_depth as well "The way I was thinking to have this work is to have an additional property to the replication_depth, where you would list types of reports or types of contacts that would be replicated from every level.
Please let me know if this makes sense to you. The change would be more extensive, but it won't be breaking and can be added to a minor release."

@dianabarsan moving convo here cause i closed PR

PR #7858, opened 2022-10-21
Replication depth to get only person contacts at each level

jonathanbataire:
# Description
For a hierarchy branch -> chw_area -> HHs
In order for a supervisor a the branch level to view target aggregates they would need to have a replication depth of 2 (in order to replicate chw contacts in the chw_area). 
This has a down side where the supervisor replicates even the HH contacts which maybe unnecessary especially for aggregating targets in large deployments. The doc count of the supervisor can grow rapidly.
This has caused challenges in https://github.com/medic/config-brac/issues/458 where supervisors are failing to log in. 

## Changes
- Introduce a mid level for replication depth at levels 1 and above. 
This can be used to replicate person contacts in a place without getting the places under it.
This way, using the supervisor example above, a 1.5 replication depth can get the chw_area person contacts necessary for target aggregates without fetching HH places.    

# Code review checklist

# License


ddocs/medic-db/medic/views/contacts_by_depth/map.js
```diff
@@ -10,6 +10,10 @@ function(doc) {
       if (parent._id) {
         emit([parent._id], value);
         emit([parent._id, depth], value);
+        if(doc.type === 'person' && depth > 1){
```
dianabarsan, line 13: I don't think this approach will work. With configurable hierarchies, the `type` won't be `person` even if this is a person contact, and there will be no way to know whether this is a person or not. I think your feature request will need some additional design decisions.

dianabarsan (commented):
I think we're going to need to rethink this. The view will not know whether a document is a person type contact or not. 
With Brac, would a change in hierarchy fix this issue?

ddocs/medic-db/medic/views/contacts_by_depth/map.js
jonathanbataire: Okay thanks @dianabarsan

jonathanbataire:
> I think we're going to need to rethink this. The view will now know whether a document is a person type contact or not. With Brac, would a change in hierarchy fix this issue?

Brac might change hierarchy but it still have 3 levels just different grouping they are trying to match MoH's hierarchy

dianabarsan:
I think this feature request kind of matches this existing issue, except applying just to contacts: https://github.com/medic/cht-core/issues/7156 .
The way I was thinking to have this work is to have an additional property to the replication_depth, where you would list types of reports or types of contacts that would be replicated from every level. 
Please let me know if this makes sense to you. The change would be more extensive, but it won't be breaking and can be added to a minor release.

jonathanbataire:
> I think this feature request kind of matches this existing issue, except applying just to contacts: #7156 . The way I was thinking to have this work is to have an additional property to the replication_depth, where you would list types of reports or types of contacts that would be replicated from every level. Please let me know if this makes sense to you. The change would be more extensive, but it won't be breaking and can be added to a minor release.

@dianabarsan this sounds great  when can we get this in we would really use this feature for BRAC

Issue #7859, opened 2022-10-21
Improve admin upgrade page UX when upgrade service fails to update containers


dianabarsan:
**What feature do you want to improve?**
If an upgrade goes well, we expect API to be down for a couple of seconds, while the new container boots up. This means that the admin upgrade page expects to receive http errors while requesting upgrade progress. 

At the same time, when we call the upgrade endpoint to finalize the upgrade, and the request succeeds, we check whether our currently installed version is the version that we requested via the upgrade. If it is not, we display an error: 

![image](https://user-images.githubusercontent.com/35681649/197211475-37c29a49-38a9-42c5-8469-560ab3976b99.png)

The browser logs will only display an error without any details: 

![image](https://user-images.githubusercontent.com/35681649/197211727-0dbe66e2-f1f2-4b80-8139-05dc9e3a0ef4.png)

API logs will not report an error: 
```
2022-10-21 13:38:06 INFO: Saving ddocs for medic-users-meta 
2022-10-21 13:38:06 INFO: Staging complete 
2022-10-21 13:38:06 INFO: Indexing staged views 
2022-10-21 13:38:06 INFO: Indexing views complete 
2022-10-21 13:38:06 INFO: Completing install 
REQ 56bb7cba-9d82-46c0-aa1c-4c4e492ab491 192.168.176.1 - GET /api/v2/upgrade HTTP/1.0
RES 56bb7cba-9d82-46c0-aa1c-4c4e492ab491 192.168.176.1 - GET /api/v2/upgrade HTTP/1.0 200 906 10.413 ms
2022-10-21 13:38:08 INFO: Install complete 
2022-10-21 13:38:08 INFO: Finalizing install 
2022-10-21 13:38:08 INFO: Detected ddoc change - reloading 
2022-10-21 13:38:09 INFO: Deleting existent staged ddocs 
2022-10-21 13:38:09 INFO: Install finalized 
2022-10-21 13:38:09 INFO: Running DB compact and view cleanup for medic 
2022-10-21 13:38:09 INFO: Running DB compact and view cleanup for medic-sentinel 
2022-10-21 13:38:09 INFO: Running DB compact and view cleanup for medic-logs 
2022-10-21 13:38:09 INFO: Running DB compact and view cleanup for medic-users-meta 
REQ 1c127a1b-b657-4a7c-adbd-bb5c85b39d62 192.168.176.1 - GET /api/v2/upgrade HTTP/1.0
[...]

PR #7863, opened 2022-10-24
Log error when no compose files were overwritten on upgrade complete

dianabarsan:
# Description

medic/cht-core#7859

# Code review checklist

# Compose URLs:

* CHT Core: https://staging.dev.medicmobile.org/_couch/builds/medic%3Amedic%3A7859-no-doc-error/docker-compose%2Fcht-core.yml
* CouchDB Single: https://staging.dev.medicmobile.org/_couch/builds/medic%3Amedic%3A7859-no-doc-error/docker-compose%2Fcht-couchdb.yml
* CouchDB Cluster: https://staging.dev.medicmobile.org/_couch/builds/medic%3Amedic%3A7859-no-doc-error/docker-compose%2Fcht-couchdb-clustered.yml
 
# License

### DIFF
--- a/api/src/services/setup/upgrade-steps.js
+++ b/api/src/services/setup/upgrade-steps.js
@@ -116,12 +116,9 @@ const complete = async (buildInfo) => {
   const stagingDoc = await upgradeUtils.getStagingDoc(buildInfo);
   const payload = upgradeUtils.getUpgradeServicePayload(stagingDoc);
 
+  // The upgrade request will initiate container upgrades. If successful, the API container will stop and the execution
+  // of this code will be interrupted. 
   const response = await upgradeUtils.makeUpgradeRequest(payload);
-
-  if (!upgradeUtils.upgradeResponseSuccess(payload, response)) {
-    logger.error('No compose files were updated: %o', response);
-    throw new Error('No compose files were updated');
-  }
 
   logger.warn(
     'upgrade-service upgrade request succeeded, without CHT-API restarting. ' +
--- a/api/src/services/setup/upgrade-steps.js
+++ b/api/src/services/setup/upgrade-steps.js
@@ -116,7 +116,20 @@ const complete = async (buildInfo) => {
   const stagingDoc = await upgradeUtils.getStagingDoc(buildInfo);
   const payload = upgradeUtils.getUpgradeServicePayload(stagingDoc);
 
-  return await upgradeUtils.makeUpgradeRequest(payload);
+  const response = await upgradeUtils.makeUpgradeRequest(payload);
+
+  if (!upgradeUtils.upgradeResponseSuccess(payload, response)) {
+    logger.error('No compose files were updated: %o', response);
+    throw new Error('No compose files were updated');
+  }
+
+  logger.warn(
+    'upgrade-service upgrade request succeeded, without CHT-API restarting. ' +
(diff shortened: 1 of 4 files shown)


dianabarsan:
Penny for your thoughts, please, @garethbowen . Thanks!

api/src/services/setup/upgrade-steps.js
```diff
@@ -116,7 +116,20 @@ const complete = async (buildInfo) => {
   const stagingDoc = await upgradeUtils.getStagingDoc(buildInfo);
   const payload = upgradeUtils.getUpgradeServicePayload(stagingDoc);
 
-  return await upgradeUtils.makeUpgradeRequest(payload);
+  const response = await upgradeUtils.makeUpgradeRequest(payload);
+
+  if (!upgradeUtils.upgradeResponseSuccess(payload, response)) {
+    logger.error('No compose files were updated: %o', response);
+    throw new Error('No compose files were updated');
```
garethbowen, line 123: I presume this shows the generic error message in the admin UI? If so, we should improve this error message to have more information about what went wrong and how to fix it.

api/src/services/setup/upgrade-steps.js
```diff
@@ -116,7 +116,20 @@ const complete = async (buildInfo) => {
   const stagingDoc = await upgradeUtils.getStagingDoc(buildInfo);
   const payload = upgradeUtils.getUpgradeServicePayload(stagingDoc);
 
-  return await upgradeUtils.makeUpgradeRequest(payload);
+  const response = await upgradeUtils.makeUpgradeRequest(payload);
+
+  if (!upgradeUtils.upgradeResponseSuccess(payload, response)) {
```
garethbowen, line 121: From a design perspective it would be better to do this work in the `makeUpgradeRequest` function rather than having to make a second call to work out if the first call was successful or not.

api/src/services/setup/upgrade-steps.js
```diff
@@ -116,7 +116,20 @@ const complete = async (buildInfo) => {
   const stagingDoc = await upgradeUtils.getStagingDoc(buildInfo);
   const payload = upgradeUtils.getUpgradeServicePayload(stagingDoc);
 
-  return await upgradeUtils.makeUpgradeRequest(payload);
+  const response = await upgradeUtils.makeUpgradeRequest(payload);
+
```
garethbowen, line 120: If successful then API will be restarted before execution makes it past this line, right? It's probably worth a comment or similar in the code, because it's not an obvious code exit point...

garethbowen (changes requested):
Good idea! A few suggestions for readability.

api/src/services/setup/upgrade-steps.js
dianabarsan: > If successful then API will be restarted before execution makes it past this line

This is correct. Adding the comment.

garethbowen (approved):
Perfect, thanks!

dianabarsan:
Hey @garethbowen

I pushed a change which will also treat responses from `upgrade-service-kubernetes`, but that one still requires changing the endpoint output in the service itself. I also changed the message to make it clearer, as you requested above and I failed to notice before.

garethbowen (approved):
Even better now  :)

Issue #7869, opened 2022-10-27
Migrate contact edit protactor test to wdio.


lorerod:
**What feature do you want to improve?**
Migrate contact edit protactor test to wdio (tests/e2e/protractor/contacts/contact-edit.specs.js)

**Describe the improvement you'd like**
Write the same test scenario in wdio and delete the protactor spec.
**Feature Tested:**

Edit contacts with the default config.
**Scenarios:**

Remove the primary contact from the health center.

Issue #5704, opened 2019-06-03
Update webapp e2e tests to be executed as an offline user


dianabarsan:
Most of our webapp workflows target offline users, while all of our webapp e2e tests, except one, run as an online user account. 
In order to have our e2e test coverage be closer aligned with actual usage, we should update webapp e2e tests that cover workflows usually performed by offline users to be tested using offline user accounts.

Given that offline users require syncing (likely manual, in this case) to get updates from the server, we might consider grouping all these e2e tests, so they run using one set of settings/forms, if flakyness and complications around needing to sync and refresh become problematic. 

Some webapp e2e tests might make sense to be run as online users, candidates could be tests around Messages tab and everything that requires SMS (for example registration_by_sms, sms-gateway).

tatilepizs:
Closing this ticket because we are doing this work incrementally every time that we need to modify an e2e test. There is not a specific PR related to this ticket

PR #7870, opened 2022-10-27
Migrate contact edit test

lorerod:
# Description

Migrate the contact edit test from protractor to wdio.

This test:
- Remove the primary contact from the health center.
- The test use the default config

Ticket: [#7869](https://github.com/medic/cht-core/issues/7869)

# Code review checklist

# License


tests/e2e/default/contacts/edit.wdio-spec.js
```diff
@@ -0,0 +1,30 @@
...
+
+describe('Editing contacts with the default config. ', () => {
+  const places = placeFactory.generateHierarchy();
+  const healthCenter = places.find(place => place.type === 'health_center');
+  const healthCenterPrimaryContact = userFactory.build({ place: healthCenter._id });
+  const docs = [...places, healthCenterPrimaryContact];
+
+  before(async () => {
+    await utils.saveDocs(docs);
+    await utils.createUsers([healthCenterPrimaryContact]);
```
tatilepizs, line 17: The user `healthCenterPrimaryContact` is created but is never used as a user. 
Do you really need to create a user?

tests/e2e/default/contacts/edit.wdio-spec.js
```diff
@@ -0,0 +1,30 @@
+const loginPage = require('../../../page-objects/default/login/login.wdio.page');
+const commonPage = require('../../../page-objects/default/common/common.wdio.page');
+const placeFactory = require('../../../factories/cht/contacts/place');
+const userFactory = require('../../../factories/cht/users/users');
+const contactPage = require('../../../page-objects/default/contacts/contacts.wdio.page');
+const utils = require('../../../utils');
+
+
+describe('Editing contacts with the default config. ', () => {
```
tatilepizs, line 9: The name of the test is `Editing contacts...` but checking the `it` i don't see that the contact name or age or note are edited. 
Should we consider changing the name of the test to be something like `Remove primary contact`? or maybe add another `it` that edit some information about the contact and verify that it was saved successfully?

tatilepizs (changes requested):
Thank you @lorerod , the test looks good.
I just added a couple of questions.

tests/e2e/default/contacts/edit.wdio-spec.js
lorerod: You are right @tatilepizs. Maybe to follow [this](https://github.com/medic/cht-core/issues/5704) issue, I can change the test to log in as an offline user and delete the primary contact. I may need to create another place with the primary contact to be deleted, because the contact deleted right now is related to the offline user. What do you think?

tatilepizs: That is a really good idea, and it is align with what we want in the future.
thanks @lorerod ! 😄

tests/e2e/default/contacts/edit.wdio-spec.js
lorerod: Great idea @tatilepizs! I will complete this spec with another `it` with the actual edition of a contact. Thanks!

tatilepizs (approved):
It looks really good. 
Thank you for adding the scenario for the edit contact name!🌟

Issue #7862, opened 2022-10-24
Use the right palette variables across webapp


latin-panda:
**Describe the issue**
There is a [comment](https://github.com/medic/cht-core/blob/master/webapp/src/css/variables.less#L1) that says:

```
Basic colors. WARNING: Don't use these outside this file as the variable names are not useful
```

But it's easily ignorable, also, in many places we're using the wrong variables for colors, we should use the [palette](https://github.com/medic/cht-core/blob/master/webapp/src/css/variables.less#L26). 

**Describe the improvement you'd like**
Make more difficult to use the wrong color variables.

**Describe alternatives you've considered**
LESS doesn't have private variables, a workaround is to move the ones that we can’t use but that are needed for `variable.less` to a file named `private-vars.less`, so `variable.less` is the only one to import it.

latin-panda:
The variables in `private-vars.less` are still accessible anywhere that imports `variable.less`, so this idea doesn't work. However having them in 2 files might help devs to understand the right way of using them

latin-panda:
This is ready for AT [here](https://github.com/medic/cht-core/pull/7876) I feel very confident about the change, the app's color should still be the same as in master.

lorerod:
**Environment**
**Client platform:** MacOS
**Browser:** Chrome
**App:** Webapp
**Versions 1:** 4.0.0
**Instance:** Local using [cht-core.yml](https://staging.dev.medicmobile.org/_couch/builds_4/medic%3Amedic%3A4.0.0/docker-compose%2Fcht-core.yml) and [cht-couchdb.yml ](https://staging.dev.medicmobile.org/_couch/builds_4/medic%3Amedic%3A4.0.0/docker-compose%2Fcht-couchdb.yml) 

**Versions 2:** Branch`7862-private-color-variables`
**Instance:** Local using Developer setup 

The app's colors look the same in both versions.

latin-panda:
Thanks @lorerod!

This has been merged to our main branch.

PR #7876, opened 2022-10-31
Extracting non-palette color vars to a new file

latin-panda:
# Description
Extracting color variables to another file, the ones that are not part of the palette.
The variables in private-vars.less are still accessible anywhere that imports variable.less, so the original idea doesn't work. However having them in 2 files might help devs to understand the right way of using them

https://github.com/medic/cht-core/issues/7862

# Code review checklist

# Compose URLs:

# License

### DIFF
--- a/webapp/src/css/private-vars.less
+++ b/webapp/src/css/private-vars.less
@@ -1,0 +1,29 @@
+/**
+* WARNING:
+* Don't import this file. This is only for "variable.less" usage.
+* Don't use these variables. The variable names aren't descriptive enough.
+* Instead, use the variables from "variable.less".
+*/
+@white: #FFFFFF;
+@gray-ultra-light: #F2F2F2;
+@gray-light: #E0E0E0;
+@gray-medium: #A0A0A0;
+@gray-dark: #777777;
+@gray-ultra-dark: #333333;
+@black: #000000; ///aaaaaaa
+@blue: #63A2C6;
+@periwinkle: #7193EE;
+@yellow: #E9AA22;
+@pink: #F47B63;
+@teal: #76B0B0;
+@blue-highlight: #EFF5F9;
+@periwinkle-highlight: #F0F4FD;
+@yellow-highlight: #FCF6E7;
+@pink-highlight: #FDF1EF;
+@teal-highlight: #e4efef;
+@red: #E33030;
+@yellow-dark: #C78330;
+@teal-dark: #218E7F;
+@blue-dark: #007AC0;
+@blue-ultra-dark: #23527C;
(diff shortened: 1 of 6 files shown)


latin-panda:
Hi @m5r, can you please review? 
Perhaps you know a better alternative to this solution 
 :) Thanks!

webapp/src/css/private-vars.less
```diff
@@ -0,0 +1,29 @@
+/**
+* WARNING:
+* Don't import this file. This is only for "variable.less" usage.
+* Don't use these variables. The variable names aren't descriptive enough.
+* Instead, use the variables from "variable.less".
+*/
```
m5r, line 6: I couldn't find a solution to reduce Less variables' scope so this is the best we can do for now :+1:

webapp/src/css/private-vars.less
```diff
@@ -0,0 +1,29 @@
...
+@white: #FFFFFF;
```
m5r, line 7: This variable is still used in https://github.com/medic/cht-core/blob/7862-private-color-variables/webapp/src/css/content-list.less#L225 but I don't know if that's intended

webapp/src/css/private-vars.less
```diff
@@ -0,0 +1,29 @@
...
+@gray-ultra-light: #F2F2F2;
+@gray-light: #E0E0E0;
+@gray-medium: #A0A0A0;
+@gray-dark: #777777;
+@gray-ultra-dark: #333333;
+@black: #000000; ///aaaaaaa
```
m5r, line 13: Is the comment still needed?

m5r (approved):
I'm sorry I don't have an alternative to propose here but it's still an improvement over what we had before, thank you Jenni!  
I've left a few small comments but I'll approve so as to not block you until I come back on monday

webapp/src/css/private-vars.less
latin-panda: Cool, thanks for checking!

webapp/src/css/private-vars.less
latin-panda: Yes, this is intended, that piece changed with the [bulk delete work](https://github.com/medic/cht-core/pull/7846/files#diff-7531bbe609e35f2936eac69d48cd591e18105b79dce25dc64de3c5c4833c1418L224) it was my attempt to reduce code conflicts :)

webapp/src/css/private-vars.less
latin-panda: hahaha sorry for the weird comment, I forgot to remove it. Thanks for catching it

PR #7877, opened 2022-10-31
Automate death and undo death report

tatilepizs:
# Description

Add the e2e automated test for `death report` and `undo death report` enketo forms.

This PR is divided in two different tests:

#### Death report
- Logs in with an offline user.
- Submit a death report.
- Verify that the `death card` was created successfully.
- Verify that the report created.
- Verify that the targets were updated accordingly.

#### Undo death report
- Logs in with an offline user.
- Submit a death report.
- Submit an undo death report.
- Verify that the `death card` is not present.
- Verify that the report created.
- Verify that the target related to `deaths` is set to `0`

cc: @jonathanvq 

# Code review checklist

# Compose URLs:

# License


tests/page-objects/default/enketo/undo-death-report.page.js
```diff
@@ -0,0 +1,14 @@
+const FORM ='form[data-form-id="undo_death_report"]';
+
+const submitReporConfirmation = (value) => $(`${FORM} ` +
```
latin-panda, line 3: Typo `Repor`
The "submit" word got me confused when reading the code below, we're not submitting the report yet, right? 
What about naming it `confirmUndoDeath` or `confirmUndoDeathOption` since it retrieves one radio element?

If it doesn't fit in the line, there are these options:

OP 1: selector in one line - allows you to colapse the block in the IDE, if you need. 
```
const confirmUndoDeathOption = (value) => {
  return $(`${FORM} input[name="/undo_death_report/undo/undo_information"][value="${value}"]`);
};
```

OP 2: unwrap form constant
```
const confirmUndoDeathOption = (value) => $(FORM +
  ` input[name="/undo_death_report/undo/undo_information"][value="${value}"]`);
```

OP 3: block style
```
const confirmUndoDeathOption = (value) => $(
  FORM +
  ` input[name="/undo_death_report/undo/undo_information"][value="${value}"]`
);
```

tests/page-objects/default/enketo/undo-death-report.page.js
```diff
@@ -0,0 +1,14 @@
...
+  `input[name="/undo_death_report/undo/undo_information"][value="${value}"]`);
+
+const selectSubmitReportConfirmation = async (value = 'yes') => {
```
latin-panda, line 6: If you decide to do the prev comment then you can do this here too. 
```suggestion
const setConfirmUndoDeathOption = async (value = 'yes') => {
```

tests/page-objects/default/enketo/undo-death-report.page.js
latin-panda: you can apply the one you chose in `death-report.page.js` selectors too :)

tests/page-objects/default/enketo/death-report.page.js
```diff
@@ -0,0 +1,66 @@
...
+const PLACE_OF_DEATH = {healthFacility: 'health_facility', home: 'home', other: 'other'};
+
+const FORM = 'form[data-form-id="death_report"]';
+const SUMMARY_SECTION = `${FORM} section[name="/death_report/group_review"]`;
+
+const deathDate = () => $(`${FORM} section[name="/death_report/death_details"] input.ignore.input-small`);
+const deathPlace = (value) => $(`${FORM} input[name="/death_report/death_details/place_of_death"][value="${value}"]`);
+const deathInformation = () => $(`${FORM} textarea[name="/death_report/death_details/death_information"]`);
+const patientNameSummary = () => $(`${SUMMARY_SECTION} ` +
+  `span[data-itext-id="/death_report/group_review/r_patient_details:label"] ` +
```
latin-panda, line 13: If you end up removing the template strings (${..}) thing, then you can use regular single quotes.

tests/page-objects/default/enketo/death-report.page.js
```diff
@@ -0,0 +1,66 @@
+const moment = require('moment');
+const genericForm = require('./generic-form.wdio.page');
+
+const PLACE_OF_DEATH = {healthFacility: 'health_facility', home: 'home', other: 'other'};
```
latin-panda, line 4: should the properties be uppercase since refer to constants? What are your thoughts? I'm not 100% strong about it

tests/page-objects/default/enketo/death-report.page.js
```diff
@@ -0,0 +1,66 @@
...
+};
+
+const setDeathInformation = async (text = 'Test note') => {
+  const note  = await deathInformation();
+  await note.waitForDisplayed();
+  await note.setValue(text);
+};
+
+const getSummaryDetails = async () => {
+  return {
```
latin-panda, line 38: If this function will return an object with only summary things inside, then you can short the names like: `patientNameSummary` to `patientName`

tests/page-objects/default/enketo/death-report.page.js
```diff
@@ -0,0 +1,66 @@
...
+    patientNameSummary: await patientNameSummary().getText(),
+    deathDateSummary: await deathDateSummary().getText(),
+    deathInformationSummary: await deathInformationSummary().getText(),
+  };
+};
+
+const submitDeathReport = async ({
+  deathDate: deathDateValue = moment().subtract(1, 'day').format('YYYY-MM-DD'),
```
latin-panda, line 46: What is the benefit of adding the defaults values here again? Clarity 🤔 
For example: `setDeathDate` function already has the default value.

tests/page-objects/default/enketo/death-report.page.js
```diff
@@ -0,0 +1,66 @@
...
+  deathPlace: deathPlaceValue = PLACE_OF_DEATH.healtFacility,
+  deathInformation: deathInformationValue = 'Test note',
+} = {}) => {
+  await selectDeathPlace(deathPlaceValue);
+  await setDeathInformation(deathInformationValue);
+  await browser.pause(5000);
```
latin-panda, line 52: What are we waiting for during this pause?

tests/page-objects/default/contacts/contacts.wdio.page.js
```diff
@@ -56,6 +56,7 @@ const formTitle = () => $('#form-title');
 const contactCardTitle = () => $('.inbox .content-pane .material .body .action-header');
 const contactInfoName = () => $('.content-pane .material .body .card .row .heading-content');
 const contactMedicID = () => $('#contact_summary .cell.patient_id > div > p');
+const contactStatus = () => $('.content-pane .material .body .card .row .heading-content > div');
```
latin-panda, line 59: This looks fragile and very similar to `contactInfoName`, do you think we can add a `test-id` like we did before? 
I can help if you need :)

tests/e2e/default/enketo/undo-death-report.wdio-spec.js
```diff
@@ -0,0 +1,64 @@
...
+const contactPage = require('../../../page-objects/default/contacts/contacts.wdio.page');
+const genericForm = require('../../../page-objects/default/enketo/generic-form.wdio.page');
+const reportsPage = require('../../../page-objects/default/reports/reports.wdio.page');
+const analyticsPage = require('../../../page-objects/default/analytics/analytics.wdio.page');
+const deathReportForm = require('../../../page-objects/default/enketo/death-report.page');
+const undoDeathReportForm = require('../../../page-objects/default/enketo/undo-death-report.page');
+
+describe('Submit an undo death report', () => {
+  const places = placeFactory.generateHierarchy();
+  const healthCenter = places.find(place => place.type === 'health_center');
```
latin-panda, line 16: We're doing this more often. I think `placeFactory.generateHierarchy()` can return a map then we just access the object like this:
```
const places = placeFactory.generateHierarchy();
const healthCenter = places['health_center'];  // Regular js object
// Or 
const healthCenter = places.get('health_center'); // Actual Map type object.
```

tests/e2e/default/enketo/undo-death-report.wdio-spec.js
```diff
@@ -0,0 +1,64 @@
...
+    await utils.createUsers([offlineUser]);
+    await loginPage.login(offlineUser);
+
+    await commonPage.goToPeople(person._id);
+    await contactPage.createNewAction('Death report');
+    await deathReportForm.submitDeathReport();
+    await commonPage.sync(true);
+  });
+
+  it('Should submit ans undo death report', async () => {
```
latin-panda, line 31: `and` or `a`
```suggestion
  it('Should submit and undo death report', async () => {
```

tests/e2e/default/enketo/death-report.wdio-spec.js
```diff
@@ -0,0 +1,78 @@
...
+  const person = personFactory.build({ parent: {_id: healthCenter._id, parent: healthCenter.parent} });
+
+  before(async () => {
+    await utils.saveDocs([...places, person]);
+    await utils.createUsers([offlineUser]);
+    await loginPage.login(offlineUser);
+  });
+
+  it('Should submit a death report', async () => {
+    const deathDate = moment().subtract(1, 'day');
```
latin-panda, line 27: if the computer current date is end of the month or first day of the month, will the target still count correctly? Because I think targets consider records in the current month 🤔
But I don't remember if this target is configured to use the report's submit date or the death date.

latin-panda (changes requested):
Very nice job ✨
I like how this automation isn't only testing the form submission but also checking the effects in other parts of the system: form summary, reports lists, contact cards and even targets. Thank you
I left some minor comments and questions below.

tests/page-objects/default/enketo/undo-death-report.page.js
tatilepizs: Thank you for these suggestions Jenny!

tests/page-objects/default/enketo/death-report.page.js
tatilepizs: After reading your comment I started looking to see if there was like a rule or some standards related to this, I found a comment in a forum that talks a little bit about that:

> ... Uppercase indicates that the variable is a constant and they are immutable.
But the properties inside a constant are mutable and hence they CAN be declared as lowercase.

And checking the examples in [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const#const_in_objects_and_arrays) it looks like it is ok to leave the properties in lower case 🙂

tests/page-objects/default/enketo/death-report.page.js
tatilepizs: 😅  That was part of the test that I was doing and forgot to delete it, thank you!

tests/page-objects/default/enketo/death-report.page.js
tatilepizs: You are completely right, thank you!

tests/page-objects/default/enketo/death-report.page.js
tatilepizs: There is no benefit at all. Sometimes I forget that in javascript there are a lot of things that you can do that you can't in other languages.
Being completely honest I thought that if didn't put a default value in that function it will fail if I try to call it without parameters 😅 
My mistake, thank you for point it out!

tests/page-objects/default/enketo/death-report.page.js
latin-panda: Nice, thanks for checking, let's leaving like this

tests/e2e/default/enketo/death-report.wdio-spec.js
```diff
@@ -0,0 +1,78 @@
...
+    const deathNote = 'Test note';
+
+    await commonPage.goToPeople(person._id);
+    await contactPage.createNewAction('Death report');
+    await deathReportForm.selectDeathPlace();
```
lorerod, line 32: Please pass a value to setDeathPlace() that you can compare later. This change will help to see what is expected and why.

lorerod (changes requested):
Thanks, @tatilepizs. It is excellent how you covered these two scenarios completely. I added a couple of suggestions that may be helpful. Let me know what you think.

tests/page-objects/default/enketo/death-report.page.js
```diff
@@ -0,0 +1,66 @@
...
+    patientNameSummary: await patientNameSummary().getText(),
+    deathDateSummary: await deathDateSummary().getText(),
+    deathInformationSummary: await deathInformationSummary().getText(),
+  };
+};
+
+const submitDeathReport = async ({
+  deathDate: deathDateValue = moment().subtract(1, 'day').format('YYYY-MM-DD'),
+  deathPlace: deathPlaceValue = PLACE_OF_DEATH.healtFacility,
+  deathInformation: deathInformationValue = 'Test note',
```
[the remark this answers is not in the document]
tatilepizs: Thanks Lore, let me try to explain a little bit better the function.

Those are the parameters of the function, the values that I am asigning there are the default values, that means that if no value is set when the function is call it will use the default value, in this case it will be `PLACE_OF_DEATH.healtFacility` for example.

The reason why I am sending the parameters as a dictionary and not as a regular variable is to give the possibility to send just one parameter when the function is called, using just the name (`deathPlace` for example). I leaned this in a exercise from `exercism` 😄  -> [here](https://exercism.org/tracks/javascript/exercises/queen-attack/solutions/tlepiz) it is in case you want to take a look.

tests/e2e/default/enketo/undo-death-report.wdio-spec.js
tatilepizs: That is a really good idea. I implemented the changes, but I am not sure if I did it in the correct way, or if something should be different/better. Could you please check?
Thanks!

tests/e2e/default/enketo/death-report.wdio-spec.js
tatilepizs: You are right again, and today was the perfect day to test that, I didn't have to modify anything, and it failed, so I changed it to the current day.
Do you think that it could be another scenario where it can failed?

tests/page-objects/default/contacts/contacts.wdio.page.js
tatilepizs: Let me try it, I want to try without any help. 
I will let you know, thank Jenny!

tests/e2e/default/enketo/death-report.wdio-spec.js
latin-panda: No, at the moment 🤔. The safest is to set a date in the middle of the month

tests/e2e/default/enketo/undo-death-report.wdio-spec.js
latin-panda: Looks good, I took a quick look :)

tests/page-objects/default/enketo/death-report.page.js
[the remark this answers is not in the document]
lorerod: So sorry for the confusion. I get it now! Thanks.

latin-panda (approved):
Great job 🙌 Thanks for sending this!

lorerod (approved):
Thank you, @tatilepizs, for all the improvements you made.

tests/page-objects/default/enketo/undo-death-report.page.js
```diff
@@ -0,0 +1,14 @@
+const FORM = 'form[data-form-id="undo_death_report"]';
+
+const confirmUndoDeathOption = (value) => $(FORM +
+  ` input[name="/undo_death_report/undo/undo_information"][value="${value}"]`);
```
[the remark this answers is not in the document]
tatilepizs: I can't change it to `'` because the variable `${value}` would no work 😅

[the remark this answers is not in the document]
latin-panda: ups sorry, somehow I didn't see it

PR #7881, opened 2022-11-02
add missing opening quote on couch readme

mrjones-plip:
# Description

~~correct env var in couchdb readme as `CLUSTER_PEER_IPS` is deprecated in favor of `COUCHDB_SERVERS`~~ Saved for later - instead just fix missing opening quote.

# Code review checklist

# Compose URLs:

# License

### DIFF
--- a/couchdb/README.md
+++ b/couchdb/README.md
@@ -35,12 +35,12 @@ This field identifies each CouchDB instance. If you are adding a new node to a c
 
 This is an identifier by which other nodes can access this node. This value should contain either the node's IP address or the [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) of the node. 
 
-#### COUCHDB_SERVERS
+#### CLUSTER_PEER_IPS
 
 This field should only be set on the cluster set up coordination node aka `setup-coordination-node` . In other clustered database systems this is known as the master/main node, however, CouchDB does not have the master/slave concept as any node can be master. This field should contain a comma-separated list of IP addresses or FQDNs. If you are deploying a clustered setup on a basic docker network, please ensure there is a (.) in your service names, to validate the FQDN requirements for Erlang. 
 
 ```yaml
-COUCHDB_SERVERS="couchdb.1,couchdb.2,couchdb.3"
+CLUSTER_PEER_IPS="couchdb.2,couchdb.3"
 ```
 
 #### COUCHDB_SYNC_ADMINS_NODE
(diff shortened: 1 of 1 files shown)


couchdb/README.md
```diff
@@ -35,12 +35,12 @@ This field identifies each CouchDB instance. If you are adding a new node to a c
 
 This is an identifier by which other nodes can access this node. This value should contain either the node's IP address or the [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) of the node. 
 
-#### CLUSTER_PEER_IPS
+#### COUCHDB_SERVERS
 
 This field should only be set on the cluster set up coordination node aka `setup-coordination-node` . In other clustered database systems this is known as the master/main node, however, CouchDB does not have the master/slave concept as any node can be master. This field should contain a comma-separated list of IP addresses or FQDNs. If you are deploying a clustered setup on a basic docker network, please ensure there is a (.) in your service names, to validate the FQDN requirements for Erlang. 
 
 ```yaml
-CLUSTER_PEER_IPS=couchdb.2,couchdb.3"
```
dianabarsan, line 43: Both values are used. 
`CLUSTER_PEER_IPS` are values that get passed to the main node, `COUCHDB_SERVERS` is passed to haproxy.

dianabarsan: CouchDb does not use `COUCHDB_SERVERS` at all, actually. `COUCHDB_SERVERS` is used by haproxy and the healthcheck.

mrjones-plip: Oh - wow - thanks so much for course correcting me!  Let's revert with a minor correction then: fixed in 97faba6e2 (adds an opening `"`)

But the problem I now have is that I know `COUCHDB_SERVERS` is the env var I need [per my testing](https://github.com/medic/cht-release-testing/issues/208#issue-1433868373), but it's not [in HAProxy readme](https://github.com/medic/cht-core/tree/master/haproxy) either, which is why I thought it was needed here.  I think we should add  `COUCHDB_SERVERS` to a readme, but I'm not sure which one.  Possibly to HAProxy's and to the yet-to-exist CHT Core compose readme based on [this search](https://github.com/medic/cht-core/search?q=COUCHDB_SERVERS).

Can you advise?

dianabarsan: I think we're going to end up needing a centralized place for the environment variables explanation, given their reuse across services.

mrjones-plip: Agreed!  I think it likely makes sense to move all the `readme.md` files to the docs site and spell out the env var use - we can use [the work you and I did already](https://github.com/medic/cht-upgrade-service#export-the-environment-variables) as a jumping off point.  That page has a lot of good stuff already!

For now, i'll rename this PR to be apt and we can quickly close it out.

dianabarsan (approved):

Issue #7806, opened 2022-09-20
Add e2e to verify breadcrumbs in Message tab and Tasks tab


elvisdorkenoo:
**What feature do you want to improve?**
End to End tests for breadcrumbs in  Messages tab and Tasks tab

**Describe the improvement you'd like**
As we added and/or updated the displayed breadcrumbs, we should add some e2e tests to make sure everything is ok

**Describe alternatives you've considered**
Add the e2e test to a previous #7732 pull request addressing the original issue

**Additional context**
_We are adding the tests in a new issue because #7732 and #7757 PRs need to go to AT asap_

PR #7889, opened 2022-11-04
add e2e tests to breadcrumbs

elvisdorkenoo:
# Description

This code adds e2e tests to the messages and tasks tabs. It is a code improvement completing the already merged
[#7732](https://github.com/medic/cht-core/pull/7732) and [#7757](https://github.com/medic/cht-core/pull/7757) pull Requests for [#5697](https://github.com/medic/cht-core/pull/7732) [#5710](https://github.com/medic/cht-core/pull/7757) issues.
medic/cht-core#7806

# Code review checklist

# Compose URLs:

# License


tests/e2e/default/tasks/tasks-breadcrumbs.wdio-spec.js
```diff
@@ -0,0 +1,156 @@
...
+        },
+        {
+          contactName: 'patient2',
+          formTitle: 'person_create',
+          lineage: health_center.name,
+          dueDateText: 'Due today',
+          overdue: true
+        },
+      ]);
+    });
```
lorerod, line 152: I would love to see one more scenario testing with an online user `describe('for online user', () => {...` `it('should display correct tasks with breadcrumbs for online user', async ()...`

tests/e2e/default/tasks/tasks-breadcrumbs-config.js
```diff
@@ -0,0 +1,50 @@
...
+  return count ? reportCount >= count : found;
+};
+
+module.exports = [
+  {
+    name: 'person_create',
+    icon: 'icon-person',
+    title: 'person_create',
+    appliesTo: 'contacts',
+    appliesToType: ['person'],
```
lorerod, line 26: It would be exciting to have a general task config file where we can create different types of tasks. For example, it would be helpful to create a task that applies to a place and other that applies to a report, to enhance the scenarios in tasks-breadcrumbs. What do you think?

tests/e2e/default/tasks/tasks-breadcrumbs-config.js
```diff
@@ -0,0 +1,50 @@
+const oneDay = 24 * 60 * 60 * 1000;
```
lorerod, line 1: This line is also in the `due-dates-config.js` file. Can we have just one task config file? Or would it be mixing scenarios?

tests/e2e/default/sms/messages-breadcrumbs.wdio-spec.js
```diff
@@ -0,0 +1,98 @@
+const utils = require('../../../utils');
+const commonElements = require('../../../page-objects/default/common/common.wdio.page');
+const loginPage = require('../../../page-objects/default/login/login.wdio.page');
+const userFactory = require('../../../factories/cht/users/users');
+const placeFactory = require('../../../factories/cht/contacts/place');
+const personFactory = require('../../../factories/cht/contacts/person');
+const messagesPo = require('../../../page-objects/default/sms/messages.wdio.page');
```
lorerod, line 7: Could you please rename this constant to `messagesPage` or `messagesPageObject`?

tests/e2e/default/sms/messages-breadcrumbs.wdio-spec.js
```diff
@@ -0,0 +1,98 @@
...
+    place: district_hospital._id,
+    contact: contact2._id,
+  });
+  const patient = personFactory.build({
+    _id: 'patient1',
+    phone: '+14152223344',
+    name: 'patient1',
+    parent: { _id: clinic._id, parent: { _id: health_center._id, parent: { _id: district_hospital._id }}}
+  });
+  const contactNameSelector = ' .sender .name';
```
lorerod, line 53: This selector could be better in the `page-object/default/sms/messages.wdio.page.js` file. Could you please move it there?

tests/e2e/default/sms/messages-breadcrumbs.wdio-spec.js
```diff
@@ -0,0 +1,98 @@
...
+  });
+
+  it('should display messages with breadcrumbs for online user', async () => {
+    await loginPage.login(onlineUser);
+    await commonElements.waitForPageLoaded();
+    await commonElements.goToMessages();
+    await messagesPo.openSendMessageModal();
+    await messagesPo.messageText('Contact');
+    await messagesPo.searchSelect(patient.phone, contactNameSelector, patient.phone);
+    await messagesPo.submitMessage();
```
lorerod, line 72: I think it will be better to create a method `sendMessage(phone, entrySelector)`. It can be in this file, or optionally you can create it in `page-objects/default/sms/messages.wdio.page.js`

tests/e2e/default/sms/messages-breadcrumbs.wdio-spec.js
```diff
@@ -0,0 +1,98 @@
...
+    const message = await messagesPo.messageByIndex(1);
+    const messagesLineages =  await (await messagesPo.listMessageLineage(message)).getText();
+
+    const expectedLineage = clinic.name.concat(health_center.name, district_hospital.name);
+
+    expect(messagesLineages).to.equal(expectedLineage);
+  });
+
+
+  it('should display reports with updated breadcrumbs for offline user', async () => {
```
lorerod, line 84: This has to be `it('should display messages with updated breadcrumbs for offline user', async () => {`

lorerod, line 84: I would love to see a scenario to check breadcrumbs on messages for a supervisor.

lorerod (changes requested):
Thanks, @elvisdorkenoo, for adding these test scenarios! I added a couple of suggestions that may be helpful. Let me know what you think.

tests/e2e/default/tasks/tasks-breadcrumbs-config.js
```diff
@@ -0,0 +1,50 @@
...
+        id: 'person-creation-follow-up',
+        start: 3,
+        end: 7,
+        dueDate: function (event, contact) {
+          return contact.contact.reported_date;
+        }
+      }
+    ]
+  },
+];
```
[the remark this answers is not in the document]
elvisdorkenoo: you got it. Config directory containing `tasks-breadcrumbs-config`, `tasks-group-config` and `due-dates-config`

tests/e2e/default/tasks/tasks-breadcrumbs.wdio-spec.js
```diff
@@ -0,0 +1,156 @@
...
+  reported_date: new Date().getTime(),
+});
+
+const getTasksInfos = async (tasks) => {
+  const infos = [];
+  for (const task of tasks) {
+    infos.push(await tasksPage.getTaskInfo(task));
+  }
+  return infos;
+};
```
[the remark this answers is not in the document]
elvisdorkenoo: done 👍🏽

tests/e2e/default/tasks/tasks-breadcrumbs.wdio-spec.js
elvisdorkenoo: I tried, but with the default config as I understand it, and as it usually the case the  `"can_view_tasks"` is only activated for chw and supervisor. Online users as program officers don't usually do tasks. Usually for breadcrumbs I test offline and online user cases, but here I added supervisor even though the logic is pretty much the same for both of them, when it comes to how we display breadcrumbs.

tests/e2e/default/tasks/tasks-breadcrumbs-config.js
elvisdorkenoo: I'm not sure that I understand the question very well but I would say, as explained in the issues https://github.com/medic/cht-core/pull/7732 https://github.com/medic/cht-core/pull/7757, it should not matter in this case if the task applies to a place or to a person, the most important is their linege, and the place of the connected user, to display the breadcrumbs as requested.

tests/e2e/default/tasks/tasks-breadcrumbs-config.js
elvisdorkenoo: Yeah, indeed, and maybe that line is not the only one ! The thing is `due-dat`e test has its own config, presets and test scenarios, and we also have t`asks-group-config` ; I was wondering if the fusion does worth the confusion, especially if I break some working tests on the way 😕

tests/e2e/default/sms/messages-breadcrumbs.wdio-spec.js
elvisdorkenoo: Sure, done !

tests/e2e/default/tasks/tasks-breadcrumbs.wdio-spec.js
lorerod: Perfect! thanks for clarifying.

tests/e2e/default/sms/messages-breadcrumbs.wdio-spec.js
elvisdorkenoo: I moved the `selector` to `messagePage`, then created a `sendMessageToPhone(message, phone`)  🙂

tests/e2e/default/sms/messages-breadcrumbs.wdio-spec.js
elvisdorkenoo: yup

tests/e2e/default/sms/messages-breadcrumbs.wdio-spec.js
elvisdorkenoo: good catch, thanks.

tests/e2e/default/tasks/tasks-breadcrumbs-config.js
lorerod: Ok, so I understand those different scenarios will be redundant. Perfect.

tests/e2e/default/sms/messages-breadcrumbs.wdio-spec.js
elvisdorkenoo: Yeah, I understand. As explained in [this](https://github.com/medic/cht-core/pull/7889#discussion_r1015641012) comment, the breadcrumbs displaying logic is the same for chw and supervisor, the difference depends on if they are online/offline, the place of the subject, and the place of the connected user, as this is the additional e2e test, many other scenarios are covered  in the unit tests for [messages](https://github.com/medic/cht-core/blob/master/webapp/tests/karma/ts/modules/messages/messages.component.spec.ts), for [reports](https://github.com/medic/cht-core/blob/master/webapp/tests/karma/ts/modules/reports/reports.component.spec.ts) and for [tasks](https://github.com/medic/cht-core/blob/master/webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts)

lorerod: Got it! Thanks

lorerod (approved):

lorerod (approved):

tests/page-objects/default/sms/messages.wdio.page.js
```diff
@@ -53,7 +53,7 @@ const sendMessageToPhone = async (message, phone) => {
 const searchSelect = async (searchText, entrySelector) => {
   await messageRecipientSelect().setValue(searchText);
   const elm = await $(entrySelector);
-  await elm.waitForDisplayed(4000);
+  await elm.waitForDisplayed(5000);
```
lorerod, line 56: I change this line of code locally to `await elm.waitForDisplayed();` run the test, and it passes. These time-based errors always happen, but when retrying, it works. I know is a problem but I don't think the solution is to increase the timeout. Can you try assigning more resources to your docker and rerun it?

lorerod: We took this to slack and the problem is not locally, is when running github-ci

Issue #7525, opened 2022-02-16
Register feedback doc when forms fail to load. 


dianabarsan:
**What feature do you want to improve?**
We catch all errors that happen when forms fail to load ([an example](https://github.com/medic/cht-core/blob/e8406d3e384cdbc48faa87e7134162b94a0bca33/webapp/src/ts/modules/contacts/contacts-edit.component.ts#L172)). However we don't register feedback docs when these crashes happen. 

**Describe the improvement you'd like**
Create a feedback doc when a form fails to load. 

**Describe alternatives you've considered**
Find some way to trickle down the error further so it gets in the automatic feedback doc capturing mechanism. 

**Additional context**
https://forum.communityhealthtoolkit.org/t/help-understand-and-fix-error-loading-form-please-try-again-or-check-with-an-administrator/1681/9

latin-panda:
@elvisdorkenoo looks like this was merged but the ticket is open, if nothing is left to be done, please close the ticket and add the milestone so it's included in the release notes :)

elvisdorkenoo:
Yeah it's done !

PR #7894, opened 2022-11-07
add more log to XMLFormService to add precision in case form load fails

elvisdorkenoo:
# Description

We catch all errors that happen when forms fail to load. However we don't register feedback docs when these crashes happen.

medic/cht-core#7525

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds/medic:medic:7525-register-failing-forms-feedback-doc/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds/medic:medic:7525-register-failing-forms-feedback-doc/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds/medic:medic:7525-register-failing-forms-feedback-doc/docker-compose/cht-couchdb-clustered.yml)
 
# License

### DIFF
--- a/.github/ISSUE_TEMPLATE/z_release_major.md
+++ b/.github/ISSUE_TEMPLATE/z_release_major.md
@@ -52,7 +52,7 @@ Keep an eye on the forum for the release announcement in the next couple of week
     - An SRE to review and confirm the documentation on upgrade instructions and breaking changes is sufficient
 - [ ] Until release testing passes, make sure regressions are fixed in `master`, cherry-pick them into the release branch, and release another beta.
 - [ ] Create a release in GitHub from the release branch so it shows up under the [Releases tab](https://github.com/medic/cht-core/releases) with the naming convention `<major>.<minor>.<patch>`. This will create the git tag automatically. Link to the release notes in the description of the release.
-- [ ] Confirm the release build completes successfully and the new release is available on the [market](https://staging.dev.medicmobile.org/builds/releases). Make sure that the document has new entry with `id: medic:medic:<major>.<minor>.<patch>`
+- [ ] Confirm the release build completes successfully and the new release is available on the [market](https://staging.dev.medicmobile.org/builds_4/releases). Make sure that the document has new entry with `id: medic:medic:<major>.<minor>.<patch>`
 - [ ] Execute the scalability testing suite on the final build and download the scalability results on S3 at medic-e2e/scalability/$TAG_NAME. Add the release `.jtl` file to `cht-core/tests/scalability/previous_results`. More info in the  [scalability documentation](https://github.com/medic/cht-core/blob/master/tests/scalability/README.md).
 - [ ] Upgrade the `demo-cht.dev` instance to this version.
 - [ ] Announce the release on the [CHT forum](https://forum.communityhealthtoolkit.org/c/product/releases/26), under the "Product - Releases" category using this template:
--- a/webapp/src/ts/services/enketo.service.ts
+++ b/webapp/src/ts/services/enketo.service.ts
@@ -28,6 +28,7 @@ import { ServicesActions } from '@mm-actions/services';
 import { ContactSummaryService } from '@mm-services/contact-summary.service';
 import { TranslateService } from '@mm-services/translate.service';
 import { TransitionsService } from '@mm-services/transitions.service';
+import { FeedbackService } from '@mm-services/feedback.service';
 import { GlobalActions } from '@mm-actions/global';
 
 @Injectable({
@@ -54,6 +55,7 @@ export class EnketoService {
     private zScoreService: ZScoreService,
     private transitionsService: TransitionsService,
     private translateService: TranslateService,
+    private feedbackService:FeedbackService,
     private ngZone: NgZone,
   ) {
     this.inited = this.init();
@@ -441,6 +443,11 @@ export class EnketoService {
 
         window.CHTCore.debugFormModel = () => form.model.getStr();
         return form;
+      }).catch(err => {
+        const errorMessage = `Failed during the form "${formDoc.internalId}" rendering : `;
+        console.error(errorMessage, err.message);
+        this.feedbackService.submit(errorMessage + err.message, false);
+        return Promise.reject(new Error(errorMessage + err.message));
       });
   }
 
--- a/webapp/src/ts/services/xml-forms.service.ts
+++ b/webapp/src/ts/services/xml-forms.service.ts
@@ -85,10 +85,9 @@ export class XmlFormsService {
         return docs[0];
       })
       .catch(err => {
-        const errorMessage = 'Error in XMLFormService getByView ';
-        console.error(errorMessage, err.message);
-        this.feedbackService.submit(errorMessage + err.message, false);
-        return Promise.reject(new Error(errorMessage));
+        const errorTitle = 'Error in XMLFormService : getByView : ';
+        console.error(errorTitle, err.message);
+        return Promise.reject(new Error(errorTitle + err.message));
       });
   }
 
@@ -270,7 +269,7 @@ export class XmlFormsService {
     return this
       .getById(internalId)
       .catch(err => {
-        console.warn('Error in XMLFormService getById : ', err?.message, err?.status, err);
+        console.warn('Error in XMLFormService : getById : ', err?.message, err?.status, err);
         if (err.status === 404) {
           // fallback for backwards compatibility
           return this.getByView(internalId);
@@ -279,13 +278,17 @@ export class XmlFormsService {
       })
       .then(doc => {
         if (!this.findXFormAttachmentName(doc)) {
+          const errorTitle = 'Error in XMLFormService : findXFormAttachmentName : ';
           const errorMessage = `The form "${internalId}" doesn't have an xform attachment`;
-          this.feedbackService.submit('Error in XMLFormService findXFormAttachmentName : ' + errorMessage, false);
-          console.error('Error in XMLFormService findXFormAttachmentName : ', errorMessage);
-          return Promise.reject(new Error(errorMessage));
+          console.error(errorTitle, errorMessage);
+          return Promise.reject(new Error(errorTitle + errorMessage));
         }
         return doc;
+      }).catch(err => {
+        this.feedbackService.submit(err.message, false);
+        throw err;
       });
+
   }
 
   getDocAndFormAttachment(internalId) {
@@ -296,13 +299,14 @@ export class XmlFormsService {
           .then(blob => this.fileReaderService.utf8(blob))
           .then(xml => ({ doc, xml }))
           .catch(err => {
+            const errorTitle = 'Error in XMLFormService : getDocAndFormAttachment : ';
+            let errorMessage = `Failed to get the form "${internalId}" xform attachment`;
             if (err.status === 404) {
-              const errorMessage = `The form "${internalId}" doesn't have an xform attachment`;
-              console.error('Error in XMLFormService getDocAndFormAttachment : ', errorMessage);
-              this.feedbackService.submit('Error in XMLFormService getDocAndFormAttachment : ' + errorMessage, false);
-              return Promise.reject(new Error(errorMessage));
+              errorMessage = `The form "${internalId}" doesn't have an xform attachment`;
             }
-            throw err;
+            console.error(errorTitle, errorMessage);
+            this.feedbackService.submit(errorTitle + errorMessage, false);
+            return Promise.reject(new Error(errorTitle + errorMessage));
           });
       });
   }
--- a/webapp/src/ts/services/xml-forms.service.ts
+++ b/webapp/src/ts/services/xml-forms.service.ts
@@ -68,22 +68,18 @@ export class XmlFormsService {
   }
 
   private getByView(internalId) {
-    let errorMsg = '';
     return this
       .init
       .then(docs => docs?.filter(doc => doc.internalId === internalId))
       .then(docs => {
         if (!docs.length) {
-          errorMsg = `No form found for internalId "${internalId}"`;
-          return Promise.reject(new Error(errorMsg));
+          return Promise.reject(new Error(`No form found for internalId "${internalId}"`));
         }
         if (docs.length > 1) {
-          errorMsg = `Multiple forms found for internalId: "${internalId}"`;
-          return Promise.reject(new Error(errorMsg));
+          return Promise.reject(new Error(`Multiple forms found for internalId: "${internalId}"`));
         }
         return docs[0];
-      })
-      .catch(err => console.error('Error in XMLFormService getByView : ', errorMsg, err));
+      });
   }
 
   private evaluateExpression(expression, doc, user, contactSummary) {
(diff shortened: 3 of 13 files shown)


webapp/src/ts/services/xml-forms.service.ts
```diff
@@ -68,18 +68,22 @@ export class XmlFormsService {
...
         }
         if (docs.length > 1) {
-          return Promise.reject(new Error(`Multiple forms found for internalId: "${internalId}"`));
+          errorMsg = `Multiple forms found for internalId: "${internalId}"`;
+          return Promise.reject(new Error(errorMsg));
         }
         return docs[0];
-      });
+      })
+      .catch(err => console.error('Error in XMLFormService getByView : ', errorMsg, err));
```
dianabarsan, line 86: I think that using `err.message` will achieve the same effect _and_ it will not require defining the additional variable. 

Also, is there a reason we don't propagate this error any more? 
 
```suggestion
      .catch(err => {
        console.error('Error in XMLFormService getByView : ', err?.message, err)
        throw err;
      });
```

webapp/src/ts/services/xml-forms.service.ts
```diff
@@ -264,11 +268,14 @@ export class XmlFormsService {
           // fallback for backwards compatibility
           return this.getByView(internalId);
         }
+        console.error(`Error in XMLFormService getByView : the form "${internalId}" was not found`);
```
dianabarsan, line 271: This error is triggered in the `get` function, not `getByView` like the message suggests: 
```suggestion
        console.error(`Error in XMLFormService.get : the form "${internalId}" was not found`);
```

webapp/src/ts/services/xml-forms.service.ts
```diff
@@ -264,11 +268,14 @@ export class XmlFormsService {
...
         throw err;
       })
       .then(doc => {
         if (!this.findXFormAttachmentName(doc)) {
-          return Promise.reject(new Error(`The form "${internalId}" doesn't have an xform attachment`));
+          const errorMsg = `The form "${internalId}" doesn't have an xform attachment`;
+          console.log('Error in XMLFormService getByView : '+errorMsg);
```
dianabarsan, line 277: I believe this is supposed to be a `console.error`? Also, the error is in `XMLFormService.get`. 
```suggestion
          console.log('Error in XMLFormService.get: ' + errorMsg);
```

dianabarsan (changes requested):

webapp/src/ts/services/xml-forms.service.ts
elvisdorkenoo: yeah we should propagate and using `err.message` made it simpler

.github/ISSUE_TEMPLATE/z_release_major.md
```diff
@@ -52,7 +52,7 @@ Keep an eye on the forum for the release announcement in the next couple of week
     - An SRE to review and confirm the documentation on upgrade instructions and breaking changes is sufficient
 - [ ] Until release testing passes, make sure regressions are fixed in `master`, cherry-pick them into the release branch, and release another beta.
 - [ ] Create a release in GitHub from the release branch so it shows up under the [Releases tab](https://github.com/medic/cht-core/releases) with the naming convention `<major>.<minor>.<patch>`. This will create the git tag automatically. Link to the release notes in the description of the release.
-- [ ] Confirm the release build completes successfully and the new release is available on the [market](https://staging.dev.medicmobile.org/builds/releases). Make sure that the document has new entry with `id: medic:medic:<major>.<minor>.<patch>`
+- [ ] Confirm the release build completes successfully and the new release is available on the [market](https://staging.dev.medicmobile.org/builds_4/releases). Make sure that the document has new entry with `id: medic:medic:<major>.<minor>.<patch>`
```
dianabarsan, line 55: The result of cherry-picking the change into the branch is that the diff shows up in the changed files in GitHub. This can be overwhelming if the cherry picked changes cover much volume. 
To avoid this, there are two options: 
1. Merge the target branch in yours.
2. Rebase your branch over the target branch

In my experience, merging is quicker and less error prone, because you only need to resolve conflicts once.

webapp/src/ts/services/xml-forms.service.ts
```diff
@@ -264,11 +264,14 @@ export class XmlFormsService {
           // fallback for backwards compatibility
           return this.getByView(internalId);
         }
+        console.error('Error in XMLFormService getByView : ', err?.message, err);
```
dianabarsan, line 267: I think this error is a result of the `getById` call, not the `getByView` one.

webapp/src/ts/services/xml-forms.service.ts
```diff
@@ -264,11 +264,14 @@ export class XmlFormsService {
...
         throw err;
       })
```
dianabarsan, line 269: I think we also need to catch errors and log errors from that are thrown by `getByView`, right?
It would be super helpful if you would add unit tests that trigger each of these errors and assert that `console.error` is called with the expected message.

dianabarsan (commented):
Nice! I added a couple of questions below.

.github/ISSUE_TEMPLATE/z_release_major.md
elvisdorkenoo: Ok, I understand, thanks for the remark  and the advice.

webapp/src/ts/services/xml-forms.service.ts
elvisdorkenoo: yeah, you are right, I tried to add some more, please check it out

webapp/tests/karma/ts/services/xml-forms.service.spec.ts
```diff
@@ -1112,6 +1127,9 @@ describe('XmlForms service', () => {
         })
         .catch(err => {
           expect(err.message).to.equal(`The form "${internalId}" doesn't have an xform attachment`);
+          expect(error.callCount).to.equal(1);
+          expect(error.args[0][0]).to.equal('Error in XMLFormService getDocAndFormAttachment : ');
+          expect(error.args[0][1]).to.equal(`The form "${internalId}" doesn't have an xform attachment`);
```
dianabarsan, line 1132: Could you please add a test where the `getByView` function throws an error and assert over what is thrown and what is logged?

dianabarsan (changes requested):
Thanks for the changes! 
I checked out the branch and I have a form that throws a loading error, but I'm not seeing a feedback being created: 
image

I think that this is because [our exception handler](https://github.com/medic/cht-core/blob/0649e9d392342d28f6646113b024f7375a9505a6/webapp/src/ts/providers/exception-handler.provider.ts#L38) is only triggered for uncaught exceptions.

webapp/tests/karma/ts/services/xml-forms.service.spec.ts
```diff
@@ -1069,6 +1076,9 @@ describe('XmlForms service', () => {
         })
         .catch(err => {
           expect(err.message).to.equal(`Multiple forms found for internalId: "${internalId}"`);
+          expect(error.callCount).to.equal(1);
+          expect(error.args[0][0]).to.equal('Error in XMLFormService getByView : ');
+          expect(error.args[0][1]).to.equal(`Multiple forms found for internalId: "${internalId}"`);
         });
```
elvisdorkenoo, line 1082: getView failing test added here

webapp/tests/karma/ts/services/xml-forms.service.spec.ts
```diff
@@ -1088,6 +1098,11 @@ describe('XmlForms service', () => {
         })
         .catch(err => {
           expect(err.message).to.equal(`No form found for internalId "${internalId}"`);
+          expect(warn.callCount).to.equal(1);
+          expect(warn.args[0][0]).to.equal('Error in XMLFormService getById, falls back to using view : ');
+          expect(error.callCount).to.equal(1);
+          expect(error.args[0][0]).to.equal('Error in XMLFormService getByView : ');
+          expect(error.args[0][1]).to.equal(`No form found for internalId "${internalId}"`);
         });
```
elvisdorkenoo, line 1106: getView failing test added here

webapp/tests/karma/ts/services/xml-forms.service.spec.ts
elvisdorkenoo: I think I did, please the comments below

dianabarsan: So, a test where the stub to get the view is something like: 
```
dbQuery.rejects({ an: 'error' });
```

elvisdorkenoo: ok, will do for `getById` as well.

webapp/src/ts/services/xml-forms.service.ts
```diff
@@ -268,7 +279,10 @@ export class XmlFormsService {
...
       .then(doc => {
         if (!this.findXFormAttachmentName(doc)) {
-          return Promise.reject(new Error(`The form "${internalId}" doesn't have an xform attachment`));
+          const errorMessage = `The form "${internalId}" doesn't have an xform attachment`;
+          this.feedbackService.submit('Error in XMLFormService findXFormAttachmentName : ' + errorMessage, false);
+          console.error('Error in XMLFormService findXFormAttachmentName : ', errorMessage);
+          return Promise.reject(new Error(errorMessage));
         }
         return doc;
       });
```
dianabarsan, line 288: So we don't need multiple calls to `feedbackService.submit`, can you please add a catch all here that will record feedback, and remove the 3 other ones: 
```
.catch(err => {
  this.feedbackService.submit('Error in XMLFormService : ' + err.message, false);
  throw err;
});
```

webapp/src/ts/services/xml-forms.service.ts
```diff
@@ -283,7 +297,10 @@ export class XmlFormsService {
           .then(xml => ({ doc, xml }))
           .catch(err => {
             if (err.status === 404) {
-              return Promise.reject(new Error(`The form "${internalId}" doesn't have an xform attachment`));
+              const errorMessage = `The form "${internalId}" doesn't have an xform attachment`;
+              console.error('Error in XMLFormService getDocAndFormAttachment : ', errorMessage);
+              this.feedbackService.submit('Error in XMLFormService getDocAndFormAttachment : ' + errorMessage, false);
+              return Promise.reject(new Error(errorMessage));
             }
             throw err;
```
dianabarsan, line 305: I think a feedback doc will need to be created when any error is thrown, not just 404s.

webapp/tests/karma/ts/services/xml-forms.service.spec.ts
```diff
@@ -1021,6 +1063,12 @@ describe('XmlForms service', () => {
         })
         .catch(err => {
           expect(err.message).to.equal(`The form "${internalId}" doesn't have an xform attachment`);
+          expect(error.callCount).to.equal(1);
+          expect(error.args[0][0]).to.equal('Error in XMLFormService findXFormAttachmentName : ');
+          expect(feedbackService.submit.callCount).to.equal(1);
+          expect(feedbackService.submit.args[0][0]
+            .startsWith('Error in XMLFormService findXFormAttachmentName : '))
+            .to.be.true;
```
dianabarsan, line 1071: I think you can use `match` here, and avoid using the "matched" text in the call:
```suggestion
          expect(feedbackService.submit.args[0][0]).to.match(/Error in XMLFormService findXFormAttachmentName/);
```
This applies to the other tests as well.

webapp/tests/karma/ts/services/xml-forms.service.spec.ts
```diff
@@ -1112,6 +1186,9 @@ describe('XmlForms service', () => {
         })
         .catch(err => {
           expect(err.message).to.equal(`The form "${internalId}" doesn't have an xform attachment`);
+          expect(error.callCount).to.equal(1);
+          expect(error.args[0][0]).to.equal('Error in XMLFormService getDocAndFormAttachment : ');
+          expect(error.args[0][1]).to.equal(`The form "${internalId}" doesn't have an xform attachment`);
```
dianabarsan, line 1191: Please add the assertions about the feedback doc being added.

webapp/tests/karma/ts/services/xml-forms.service.spec.ts
```diff
@@ -1112,6 +1186,9 @@ describe('XmlForms service', () => {
...
         });
```
dianabarsan, line 1192: Please add a test where `dbGetAttachment` rejects with an error that is not a 404 and assert that a feedback doc is added.

dianabarsan (changes requested):
Cool stuff, just a couple of suggestions inline.

webapp/tests/karma/ts/services/xml-forms.service.spec.ts
elvisdorkenoo: ok

webapp/tests/karma/ts/services/xml-forms.service.spec.ts
elvisdorkenoo: ok

webapp/tests/karma/ts/services/xml-forms.service.spec.ts
elvisdorkenoo: ok

webapp/src/ts/services/xml-forms.service.ts
elvisdorkenoo: ok

dianabarsan (commented):
Thanks so much! Changes look great and are effective in capturing errors when form documents don't exist. Error messages are also clear at pinpointing what caused the actual error. 
 
However I think errors _in_ forms are not covered yet - I tested this by loading a form that contains a countdown widget, which is now broken. 

I think that we also have to catch, register feedback and then throw the error further when forms rendering fails. I think a good place to do that is here: https://github.com/medic/cht-core/blob/58ea7489dd89a114e3c4170db9da5fa6845612b8/webapp/src/ts/services/enketo.service.ts#L410, but leaving that up to you.

webapp/src/ts/services/enketo.service.ts
```diff
@@ -441,6 +443,11 @@ export class EnketoService {
 
         window.CHTCore.debugFormModel = () => form.model.getStr();
         return form;
+      }).catch(err => {
+        const errorMessage = `Failed during the form "${formDoc.internalId}" rendering : `;
+        console.error(errorMessage, err.message);
+        this.feedbackService.submit(errorMessage + err.message, false);
```
dianabarsan, line 449: Awesome!! 
image

webapp/src/ts/services/xml-forms.service.ts
```diff
@@ -268,10 +278,17 @@ export class XmlFormsService {
...
         if (!this.findXFormAttachmentName(doc)) {
-          return Promise.reject(new Error(`The form "${internalId}" doesn't have an xform attachment`));
+          const errorTitle = 'Error in XMLFormService : findXFormAttachmentName : ';
+          const errorMessage = `The form "${internalId}" doesn't have an xform attachment`;
+          console.error(errorTitle, errorMessage);
+          return Promise.reject(new Error(errorTitle + errorMessage));
         }
         return doc;
+      }).catch(err => {
+        this.feedbackService.submit(err.message, false);
```
dianabarsan, line 288: Also awesome!
image

dianabarsan (approved):
🌟

webapp/src/ts/services/enketo.service.ts
elvisdorkenoo: 🎉

dianabarsan:
Hi @elvisdorkenoo 
Was this AT-ed?

elvisdorkenoo:
> Hi @elvisdorkenoo
> Was this AT-ed?

Not yet, sorry... As discussed, I re-opened the issue, and will ask QA if they are ok AT-ing on master.
Thanks.
cc @ngaruko

Issue #6994, opened 2021-03-11
Reports with empty patient_id fields don't link to contact in the UI


dianabarsan:
**Describe the bug**
When a report has an empty value for their `patient_id` field, or lacks the field entirely, will not link to the patient contact in the UI, despite having a valid `patient_uuid` field. 

**To Reproduce**
1. Disable `generate_patient_id_on_people` and `generate_shortcode_on_contacts` transitions.
2. Create a contact. 
3. Submit a report for this contact.
4. Open the report and see that there is no link to your contact.
![image](https://user-images.githubusercontent.com/35681649/110736805-f5e0e900-8234-11eb-8956-4e023e69a703.png)
5. Enable `generate_patient_id_on_people` or `generate_shortcode_on_contacts` and edit the contact (to trigger sentinel to process the doc). 
6. Submit another report for the contact and notice that the new report does link to the contact. 

**Expected behavior**
When a report has a valid and existent subject, the reports detail page should link to this contact. 

**Environment**
- App: webapp
- Version: 3.10, 3.11

PR #7899, opened 2022-11-08
add contact link to report with  #6994

elvisdorkenoo:
# Description

When a report has an empty value for their patient_id field, or lacks the field entirely, will not link to the patient contact in the UI, despite having a valid patient_uuid field.

Instructions for AT :
- before fix
Fill and submit a report form missing or empty patient_id field, but having valid patient_uuid field, you would notice there is no link to patient contact on the report detail ui.
- After fix
The report detail ui now have a link to the patient contact.

medic/cht-core#6994

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds/medic:medic:6994-add-contact-link-to-report-ui/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds/medic:medic:6994-add-contact-link-to-report-ui/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds/medic:medic:6994-add-contact-link-to-report-ui/docker-compose/cht-couchdb-clustered.yml)
 
# License

### DIFF
--- a/ddocs/medic-db/medic/views/doc_summaries_by_id/map.js
+++ b/ddocs/medic-db/medic/views/doc_summaries_by_id/map.js
@@ -34,7 +34,6 @@ function(doc) {
     var subject = {};
     var reference = doc.patient_id ||
                     (doc.fields && doc.fields.patient_id) ||
-                    doc.patient_uuid ||
                     (doc.fields && doc.fields.patient_uuid) ||
                     doc.place_id ||
                     (doc.fields && doc.fields.place_id);
--- a/ddocs/medic-db/medic/views/doc_summaries_by_id/map.js
+++ b/ddocs/medic-db/medic/views/doc_summaries_by_id/map.js
@@ -21,6 +21,7 @@ function(doc) {
     }
 
     if (error.fields.indexOf('patient_id') !== -1 ||
+      error.fields.indexOf('patient_uuid') !== -1 ||
       error.fields.indexOf('patient_name') !== -1 ||
       error.fields.indexOf('place_id') !== -1) {
       return true;
@@ -33,6 +34,8 @@ function(doc) {
     var subject = {};
     var reference = doc.patient_id ||
                     (doc.fields && doc.fields.patient_id) ||
+                    doc.patient_uuid ||
+                    (doc.fields && doc.fields.patient_uuid) ||
                     doc.place_id ||
                     (doc.fields && doc.fields.place_id);
 
--- a/webapp/src/ts/modules/reports/reports-content.component.html
+++ b/webapp/src/ts/modules/reports/reports-content.component.html
@@ -35,7 +35,7 @@
           </div>
 
           <div *ngIf="summaries[$index]?.subject?.type" class="subject">
-            <a class="name" *ngIf="summaries[$index]?.subject._id || summaries[$index].patient._id"
+            <a class="name" *ngIf="summaries[$index]?.subject._id || summaries[$index]?.patient?._id"
               [routerLink]="['/contacts', summaries[$index].subject._id || summaries[$index].patient._id]"
               [innerHTML]="(summaries[$index].subject.type === 'name' && summaries[$index].subject.value) || summaries[$index].subject.name || ('report.subject.unknown' | translate)">
             </a>
--- a/webapp/src/ts/modules/reports/reports-content.component.html
+++ b/webapp/src/ts/modules/reports/reports-content.component.html
@@ -35,11 +35,11 @@
           </div>
 
           <div *ngIf="summaries[$index]?.subject?.type" class="subject">
-            <a class="name" *ngIf="summaries[$index]?.subject._id || summaries[$index].fields?.patient_uuid"
-              [routerLink]="['/contacts', summaries[$index].subject._id || summaries[$index].fields.patient_uuid]"
+            <a class="name" *ngIf="summaries[$index]?.subject._id || summaries[$index].patient._id"
+              [routerLink]="['/contacts', summaries[$index].subject._id || summaries[$index].patient._id]"
               [innerHTML]="(summaries[$index].subject.type === 'name' && summaries[$index].subject.value) || summaries[$index].subject.name || ('report.subject.unknown' | translate)">
             </a>
-            <span class="name" *ngIf="!summaries[$index]?.subject._id && !summaries[$index].fields?.patient_uuid"
+            <span class="name" *ngIf="!summaries[$index]?.subject._id && !summaries[$index].patient._id"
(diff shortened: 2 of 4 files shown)


elvisdorkenoo:
@dianabarsan I tried to reproduce the issue following the [steps](https://github.com/medic/cht-core/issues/6994#issue-828724220). 
image

Turn out the pages is looking for a `fields?.patient_uuid` but fields is an array instead of object and does lack that values. However, even with the transitions `generate_patient_id_on_people` and `generate_shortcode_on_contacts`, we still have access to `patient._id` so I used that value instead as condition and source of information to get contact link.

webapp/src/ts/modules/reports/reports-content.component.html
```diff
@@ -35,11 +35,11 @@
           </div>
 
           <div *ngIf="summaries[$index]?.subject?.type" class="subject">
-            <a class="name" *ngIf="summaries[$index]?.subject._id || summaries[$index].fields?.patient_uuid"
-              [routerLink]="['/contacts', summaries[$index].subject._id || summaries[$index].fields.patient_uuid]"
+            <a class="name" *ngIf="summaries[$index]?.subject._id || summaries[$index].patient._id"
```
dianabarsan, line 38: Cool. 
I think we need to nullcheck patient here, in case it's not set: 
```suggestion
            <a class="name" *ngIf="summaries[$index]?.subject._id || summaries[$index]?.patient?._id"
```
And same goes for the other span.

dianabarsan (changes requested):
Nice, thanks for this change. 

I think the best option here would be to edit [the view](https://github.com/medic/cht-core/blob/c1bc3638b76e4a313beedaec3356cc08c9bf372f/ddocs/medic-db/medic/views/doc_summaries_by_id/map.js#L32) and the [cloned-view](https://github.com/medic/cht-core/blob/52605ac752bb40ab61ee1198fd7fe1bc0fbef982/webapp/src/ts/services/get-summaries.service.ts#L37) in webapp to support patient_uuid subjects as well. That would require a view update, and would likely disqualify this from making it to a patch release, and would have to be released in a minor. 

Additionally, I think it would be very helpful if you would add an e2e test, or an addition to an existent e2e test to check that the view code works (since views don't have unit tests). 

Thanks!

elvisdorkenoo:
@dianabarsan please can you ta

> Nice, thanks for this change.
> 
> I think the best option here would be to edit [the view](https://github.com/medic/cht-core/blob/c1bc3638b76e4a313beedaec3356cc08c9bf372f/ddocs/medic-db/medic/views/doc_summaries_by_id/map.js#L32) and the [cloned-view](https://github.com/medic/cht-core/blob/52605ac752bb40ab61ee1198fd7fe1bc0fbef982/webapp/src/ts/services/get-summaries.service.ts#L37) in webapp to support patient_uuid subjects as well. That would require a view update, and would likely disqualify this from making it to a patch release, and would have to be released in a minor.
> 
> Additionally, I think it would be very helpful if you would add an e2e test, or an addition to an existent e2e test to check that the view code works (since views don't have unit tests).
> 
> Thanks!

Thank you for the feedback, please @dianabarsan can you take a look before I get started with the e2e test.

ddocs/medic-db/medic/views/doc_summaries_by_id/map.js
```diff
@@ -33,6 +34,8 @@ function(doc) {
     var subject = {};
     var reference = doc.patient_id ||
                     (doc.fields && doc.fields.patient_id) ||
+                    doc.patient_uuid ||
```
dianabarsan, line 37: `doc.patient_uuid` is not a property that is used by the CHT. `patient_uuid` should only be part of `fields`. If we add this to the view, we would validate the support for `doc.patient_uuid`, while we do not actually support it.

webapp/src/ts/modules/reports/reports-content.component.html
```diff
@@ -35,11 +35,11 @@
           </div>
 
           <div *ngIf="summaries[$index]?.subject?.type" class="subject">
-            <a class="name" *ngIf="summaries[$index]?.subject._id || summaries[$index].fields?.patient_uuid"
-              [routerLink]="['/contacts', summaries[$index].subject._id || summaries[$index].fields.patient_uuid]"
+            <a class="name" *ngIf="summaries[$index]?.subject._id || summaries[$index]?.patient?._id"
```
dianabarsan, line 38: With the view change, I think this is not needed anymore, right? `summaries[$index]?.subject._id` should be sufficient?

webapp/src/ts/services/get-summaries.service.ts
```diff
@@ -39,6 +39,8 @@ export class GetSummariesService {
     const reference =
       doc.patient_id ||
       (doc.fields && doc.fields.patient_id) ||
+      doc.patient_uuid ||
```
dianabarsan, line 42: As above, `doc.patient_uuid` should not be used. Only `doc.fields.patient_uuid` should be used.

webapp/tests/mocha/unit/views/doc_summaries_by_id.spec.js
```diff
@@ -96,7 +96,8 @@ const postNatalVisit = {
 const postNatalVisitBis = Object.assign({}, postNatalVisit, {
   _id: '4971a859-bde7-4ff0-a0ed-326925b83038-bis',
   fields: Object.assign({}, postNatalVisit.fields, {
-    patient_id: null
+    patient_id: null,
```
dianabarsan, line 99: Please add a couple more tests here where: 

- we have patient_id and no patient_uuid
- we don't have patient_id, but we have patient_uuid
- we have both patient_id and patient_uuid

dianabarsan (changes requested):
Great, thanks for the changes! I left a couple of requests inline.

webapp/tests/mocha/unit/views/doc_summaries_by_id.spec.js
elvisdorkenoo: Got it ! add, add the first two, I think the last one is already covered.

elvisdorkenoo:
> Awesome!
> I think this could also use an e2e test, we can add that in a separate PR.

Great, thanks @dianabarsan, if you agree, I will go ahead and merge, then create a new issue for the e2e tests.

lorerod (approved):
Look great @elvisdorkenoo !

elvisdorkenoo:
> Look great @elvisdorkenoo !

thanks @lorerod

Issue #7912, opened 2022-11-17
Calling `/api/v1/settings/deprecated-transitions` prevents any future transitions to run over new records in API


dianabarsan:
**Describe the bug**
The call to `/api/v1/settings/deprecated-transitions` will cause the settings object in the transitions lib to be unset, preventing all calls that pass documents to transitions to resolve. 

**To Reproduce**
1. Enable `death_reporting`. This doesn't just affect this transition, but this transition's code will be effective in triggering the error. Also make sure that you have no other transitions config errors, transitions must be working in order to trigger the bug. 
2. Create an SMS form. Any form will do, I just copied the ["DR" form from standard config](https://github.com/medic/cht-core/blob/master/config/standard/app_settings.json#L3916). 
3. Submit a DR form through `api/v2/records` to check that the call works. 
4. Call `/api/v1/settings/deprecated-transitions` . You can even do it from your browser. 
5. Submit a DR form through `api/v2/records`. Notice that the request hangs. Also if you check api logs you'll see something like 
```
2022-11-17 09:31:19 ERROR: UNHANDLED REJECTION! 
2022-11-17 09:31:19 ERROR:   Reason: TypeError: Cannot read properties of undefined (reading 'death_reporting')
    at Object.get (/home/diana/projects/medic/shared-libs/transitions/src/config.js:10:21)
    at getConfig (/home/diana/projects/medic/shared-libs/transitions/src/transitions/death_reporting.js:13:32)
    at getConfirmFormCodes (/home/diana/projects/medic/shared-libs/transitions/src/transitions/death_reporting.js:14:35)
    at isConfirmForm (/home/diana/projects/medic/shared-libs/transitions/src/transitions/death_reporting.js:17:31)
    at Object.filter (/home/diana/projects/medic/shared-libs/transitions/src/transitions/death_reporting.js:52:8)
    at canRun (/home/diana/projects/medic/shared-libs/transitions/src/transitions/index.js:220:16)
    at applyTransition (/home/diana/projects/medic/shared-libs/transitions/src/transitions/index.js:270:18)
    at wrapper (/home/diana/projects/medic/shared-libs/transitions/node_modules/lodash/lodash.js:5213:19)
    at /home/diana/projects/medic/shared-libs/transitions/node_modules/async/dist/async.js:3674:28
    at replenish (/home/diana/projects/medic/shared-libs/transitions/node_modules/async/dist/async.js:446:21) {
[...]

PR #7914, opened 2022-11-18
Fix deprecated-transitions api 

dianabarsan:
# Description

- fixes transition filter errors causing processing transitions to hang (forever)
- `deprecated-transitions` not gets `transitionsLib` from config
- adds unit test coverage for `deprecated-transitions`
- adds e2e test

medic/cht-core#7912

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds/medic:medic:7912-fix-deprecated-transitions-api/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds/medic:medic:7912-fix-deprecated-transitions-api/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds/medic:medic:7912-fix-deprecated-transitions-api/docker-compose/cht-couchdb-clustered.yml)
 
# License

### DIFF
--- a/shared-libs/transitions/test/unit/transitions.js
+++ b/shared-libs/transitions/test/unit/transitions.js
@@ -1,8 +1,11 @@
 const sinon = require('sinon');
-const assert = require('chai').assert;
+const { assert, expect, ...chai } = require('chai');
+const chaiAsPromised = require('chai-as-promised');
+chai.use(chaiAsPromised);
 const transitions = require('../../src/transitions');
 const config = require('../../src/config');
 const _ = require('lodash');
+const infodoc = require('@medic/infodoc');
 
 const requiredFunctions = {
   onMatch: 1,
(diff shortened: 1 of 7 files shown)


dianabarsan:
Hi @jkuester . Would you mind a review? Thanks!

shared-libs/transitions/test/unit/transitions.js
```diff
@@ -333,5 +336,186 @@ describe('transitions', () => {
     assert.isDefined(deprecatedTransitions);
     assert.equal(deprecatedTransitions.length, 2);
   });
+
+  describe('applyTransition',  () => {
+    it('should apply transition with change', async () => {
+      const transition = {
+        key: 'mytransition',
+        filter: sinon.stub().returns(true),
+        onMatch: sinon.stub().resolves(true),
```
jkuester, line 345: Serious question (not trying to be "smart" here, but actually wanting to understand the criteria): is this a case where we should follow the [method chaining formatting guidelines](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/#method-chaining) and put these calls on separate lines?

shared-libs/transitions/test/unit/transitions.js
```diff
@@ -333,5 +336,186 @@ describe('transitions', () => {
...
+    });
+
+    it('should force running a transition', async () => {
+      const transition = {
+        key: 'mytransition',
+        filter: sinon.stub(),
+        onMatch: sinon.stub().resolves(true),
+      };
+
+      sinon.stub(transitions, 'canRun').returns(false);
```
jkuester, line 437:
```suggestion
      sinon.stub(transitions, 'canRun');
```

Unless I am missing something, we don't need to set any behavior here since this should never be called.

shared-libs/transitions/test/unit/transitions.js
```diff
@@ -333,5 +336,186 @@ describe('transitions', () => {
...
+        transitions.applyTransition(
+          { key: transition.key, change, transition },
+          (err, result) => err ? reject(err) : resolve(result)
+        );
+      });
+      expect(result).to.equal(false);
+      expect(transition.filter.args).to.deep.equal([[change.doc, change.info]]);
+      expect(transition.onMatch.called).to.equal(false);
+      expect(infodoc.updateTransition.called).to.equal(false);
+    });
```
jkuester, line 473: Do we typically try to include assertions for the error logging?

tests/integration/transitions/sentinel-api-transitions.spec.js
```diff
@@ -1055,4 +1055,73 @@ describe('transitions', () => {
         chai.expect(updatedPerson.patient_id).to.be.ok;
       });
   });
+
+  it('calling getDeprecatedTransitions endpoint does not break API transitions', () => {
```
jkuester, line 1059: Someone in the future is going to laugh when they read this test and think "I wonder what horrible bug lead to this test case...."  😂

jkuester (approved):
LGTM!  I am very relieved to see that your solution was the same as what we did in #7709!  😂

shared-libs/transitions/test/unit/transitions.js
dianabarsan: Method chaining on multiple lines is strictly for readability and to be consistent: 
So it's: 
- super short functions that fit in one line (eg `sinon,stub().withArgs(3).returns(6);`) and refer to one thing (semantically)
- two functions that fit well in one line (eg `array.map(value => value * 2).filter(value => value)`)
- every call on a separate line

Idea is that there is no inconsistency, with some functions inlined and some not:
```
thing.call()
  .call()
  .call()
```

shared-libs/transitions/test/unit/transitions.js
dianabarsan: I think it depends on the case, and so far it comes down to preference, and whether the logged errors are really imporant: do they make it to a feedback doc?, does the client see them?, etc. 
In this case, for a patch release that I might need to backport 3 versions back, I wanted to reduce the risk of a more complex cherry-pick.

tests/integration/transitions/sentinel-api-transitions.spec.js
dianabarsan: You're right. I left comments with the issue number in the past, but I found I just `git blamed` to get info about the code and seldom copied the issue number (or link) and pasted it in the browser. 

What do you prefer? 
- a comment mentioning the issue number 
- renaming the test (the bloody endpoint needs at least one e2e test!)
- no change

jkuester: IMHO git blame is sufficient for doing historical research into a code change.  (That is kinda the whole point of commit history...)  

Sorry, I should have been more clear that I do not think anything actually needs to change here.  I was just amused at how the most nonsensical test cases always result from weird bugs (and anytime you see a test case that seems really random, it is probably the result of an issue like this...).

Issue #7916, opened 2022-11-18
Submit an undo death report "before all" hook for Submit an undo death report frequently fails


dianabarsan:
**Describe the issue**
Very frequently, `ci-webdriver-default` CI test fails with:
```
[Chrome Headless 107.0.5304.87 linux #0-23] 1) Submit an undo death report "before all" hook for Submit an undo death report
[Chrome Headless 107.0.5304.87 linux #0-23] element (".header-logo") still not displayed after 15000ms
[Chrome Headless 107.0.5304.87 linux #0-23] Error: element (".header-logo") still not displayed after 15000ms
[Chrome Headless 107.0.5304.87 linux #0-23]     at /home/runner/work/cht-core/cht-core/node_modules/webdriverio/build/commands/browser/waitUntil.js:66:23
[Chrome Headless 107.0.5304.87 linux #0-23]     at async Element.wrapCommandFn (/home/runner/work/cht-core/cht-core/node_modules/@wdio/utils/build/shim.js:131:29)
[Chrome Headless 107.0.5304.87 linux #0-23]     at async Element.elementErrorHandlerCallbackFn (/home/runner/work/cht-core/cht-core/node_modules/webdriverio/build/middlewares.js:24:32)
[Chrome Headless 107.0.5304.87 linux #0-23]     at async Element.wrapCommandFn (/home/runner/work/cht-core/cht-core/node_modules/@wdio/utils/build/shim.js:131:29)
[Chrome Headless 107.0.5304.87 linux #0-23]     at async Element.wrapCommandFn (/home/runner/work/cht-core/cht-core/node_modules/@wdio/utils/build/shim.js:131:29)
[Chrome Headless 107.0.5304.87 linux #0-23]     at async Element.elementErrorHandlerCallbackFn (/home/runner/work/cht-core/cht-core/node_modules/webdriverio/build/middlewares.js:24:32)
[Chrome Headless 107.0.5304.87 linux #0-23]     at async Element.wrapCommandFn (/home/runner/work/cht-core/cht-core/node_modules/@wdio/utils/build/shim.js:131:29)
[Chrome Headless 107.0.5304.87 linux #0-23]     at async waitForPageLoaded (/home/runner/work/cht-core/cht-core/tests/page-objects/default/common/common.wdio.page.js:177:3)
[...]

Issue #7915, opened 2022-11-18
Incorrect container names are used when getting e2e logs


dianabarsan:
**Describe the issue**
Typos in some container names mapping prevent from getting failed test logs. 

**Describe the improvement you'd like**
Fix the names

PR #7917, opened 2022-11-18
Fix container names when getting broken test logs

dianabarsan:
# Description

- change what what we wait for when logging in with an offline user
- wait until sentinel has updated the person document before proceeding (avoid webapp race conditions when opening forms)

medic/cht-core#7916
medic/cht-core#7915

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7915-fix-container-names/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7915-fix-container-names/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7915-fix-container-names/docker-compose/cht-couchdb-clustered.yml)
 
# License


dianabarsan:
Hey @latin-panda 
Wondering if you could review this tiny PR for me?

tests/page-objects/default/login/login.wdio.page.js
```diff
@@ -24,7 +24,8 @@ const login = async ({ username, password, createUser = false, locale, loadPage
   }
 
   if (loadPage) {
-    await commonPage.waitForLoaders();
+    const waitForPartialLoad = privacyPolicy || adminApp;
```
latin-panda, line 27: What is the reason for making these specific (privacyPolicy, adminApp) instead of generic (isPartialLoad or waitForPartialLoad)? 
If there's another app or page, then we'll need to add another property.

Also, when reading the usage of this function:
 `await loginPage.login({ username: constants.USERNAME, password: constants.PASSWORD, adminApp: true });` 
The meaning of the `adminApp` flag wasn't straightforward and I thought that maybe it was force redirection to the admin page after login 😅

latin-panda (commented):
LGTM!! Just a question inline before approving please :)

tests/page-objects/default/login/login.wdio.page.js
dianabarsan: The redirection to the admin app is in API. I understand the source of confusion. 
I didn't like something like partial load because we already had a 'loadPage' parameter that was used throughout. having two parameters 'loadpage' and 'partialload' seemed more confusing. are we loading or not? rather than being explicit about how it loads (partially), i was explicit about what i expected it to load. 

i don't expect the list of special page's to grow soon.

latin-panda: > The redirection to the admin app is in API. I understand the source of confusion.

yes, my initial thought was a regular browser navigation

latin-panda (approved):
Looks good!

Issue #7642, opened 2022-06-10
Successfull users-info request reported as errored


dianabarsan:
**Important**: This is a public repository. Anyone in the world can see what's posted here. If you are posting screenshots or log files, please **carefully examine them for** the presence of any kind of **protected health information** (PHI). Images or logs containing PHI _must_ be posted in fully-redacted form, with no visible PHI.

**Describe the bug**
When logging in as an offline user and watching the console, you'll see a warning: 
```
Error fetching users-info - ignoring {total_docs: 11455, warn_docs: 11455, warn: true, limit: 10000}
```
Even though the request was successful. 

**To Reproduce**
1. Observe browser console. 
2. Login as an offline user.
4. See warning

**Expected behavior**
There should be no warning. 

**Logs**
```
Error fetching users-info - ignoring {total_docs: 11455, warn_docs: 11455, warn: true, limit: 10000}
```

**Environment**
- Browser: Chrome
- App: webapp
- Version: 4.0

**Additional context**
Tagging this as low priority because the warning doesn't block execution and doesn't affect the rest of the workflow. https://github.com/medic/cht-core/blob/1f0f990150bf2e7df0242185f87365e96dbbecf9/webapp/src/js/bootstrapper/index.js#L69-L71

dianabarsan:
@alium0978 Can we close this off if the fix was merged?

dianabarsan:
It's not obvious if the fix was AT-ed.

andrablaj:
No, I don't think it was AT-ed. That was my bad as I asked Samuel today if he merged the PR, rather than asking him if the PR was AT-ed. Sorry for that.
@tatilepizs @lorerod @ngaruko putting this on your radar when you perform release testing. Not sure what is the process when a PR gets merged without AT-ing, can you please advise on that? Thanks

samuelimoisili:
Sorry, it wasn't AT-ed, how to I resolve it? Thanks for calling it out.

tatilepizs:
Thank you @samuelimoisili and @andrablaj for the heads up.

---

### Test details

**Config:** Default
**Environment:** Local with docker helper script
**Platform:** WebApp
**Browser:** Chrome

### Reproducible on _cht old version_
I looked at the PRs that are still open to find one that was old enough so I can reproduce the error.
When I logged in with an offline user the warning **was shown.**

<details>
<summary>Image attached.</summary>

[...]

PR #7946, opened 2022-11-30
Returns `code` as part for the response object from the API

alium0978:
# Description
Updated the API to return `code` as part of the response object.

medic/cht-core#7642 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7642-offline-user-login-warning/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7642-offline-user-login-warning/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7642-offline-user-login-warning/docker-compose/cht-couchdb-clustered.yml)
 
# License


api/src/controllers/users.js
```diff
@@ -222,6 +222,7 @@ module.exports = {
         warn_docs: warn,
         warn: warn >= usersService.DOC_IDS_WARN_LIMIT,
         limit: usersService.DOC_IDS_WARN_LIMIT,
+        code: res.statusCode,
```
dianabarsan, line 225: I think adding the statusCode as part of the body is redundant and not common practice. So I don't think we should fall back to this.

I think the fix should be on the client-side, where we assume that `utils.fetchJSON` will actually return the statusCode, when it does not, it only returns the body of the response if the response is successful: 
https://github.com/medic/cht-core/blob/1f0f990150bf2e7df0242185f87365e96dbbecf9/webapp/src/js/bootstrapper/utils.js#L35-L36

To simulate how this works, run this code replacing the url with your CouchDb local install:
```
const fetch = require('node-fetch');
const url = 'http://admin:pass@127.0.0.1:6984';

(async () => {
  const response = await fetch(url);
  console.log(response);
  const data = await response.json();
  console.log(data);

  const willError = await fetch(`${url}/thisdbdoesnotexist`);
  console.log(willError);
  const errorData = await willError.json();
  console.log(errorData);
})();
```

dianabarsan (commented):
Wow, quick PR! I love the testing! 
I left one comment about how we should keep the changes on the client-side only.

api/src/controllers/users.js
alium0978: Should `fetchJSON` add the status code to the data?

```json
{
   ...data,
  code: res.statusCode
}
```

dianabarsan: I think that the most intuitive way of doing it is to have `fetchJson` throw an error when the response is not ok. 
We can leverage [`response.ok`](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok), and throw an error that includes the error code (optionally also the response body) if the response is not ok. 
There's an example included here: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#checking_that_the_fetch_was_successful 
Here's another example of how to handle fetch errors from the PouchDb source code: https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-adapter-http/src/index.js#L217

alium0978: Does this work? The response from `fetchJSON` will return an interface similar to this:

```ts
interface FetchJSONResponse {
  code: number;
  ok: Boolean;
  data: any;
}
```

The new `fetchJSON` implementation will be similar to this:
```js
const fetchJSON = (path) => {
  const baseUrl = getBaseUrl();
  const options = {
    credentials: 'same-origin',
    headers: opts.remote_headers
  };

  const result = {};

  return fetch(`${baseUrl}${path}`, options)
    .then(res => {
      result.code = res.status;
      result.ok = res.ok;
      return res.json();
    })
    .then(data => {
      result.data = data;
      return result;
    });
};
```

The success check in the bootstrapper will check if the `response` is `ok` alongside the `code`.

dianabarsan: I think this will work well. 

We don't yet have a standard of handing fetch requests. We use `request` (via `request-promise-native`) server-side to make http requests, and that uses the pattern of throwing an error when the status is not in the success range. 
Because `request` is no longer supported, and frontend already uses `fetch`, we will probably migrate to `node-fetch` for server-side gradually soon. 
All of our code relies on the http request to throw an error when an non-successful status code is received, so we will probably continue with this pattern even when using `node-fetch`, to reduce the volume of refactors necessary. 

This is why I suggested you should throw an error on a non-successful status in this case as well, because we will most likely use this pattern on fetch responses server-side, and it's generally best to be consistent.

dianabarsan (approved):
Nice work! I left one tiny comment inline.

Issue #7753, opened 2022-08-30
Supervisor CHW Create


mrjones-plip:
**What feature do you want to improve?**
Right now there is no way to for an offline user, like a supervisor, to create another user, like a CHW.  Instead they need to reach out to an CHT admin, an online user, to create the account for them.  This can often take a while and leaves the supervisor with a device and a CHW that are prevented from giving care to the area's patients. 

**Describe the improvement you'd like**
It would be great if supervisors, using an authentication code, were allowed to create a CHW user in an area.  This would cut out a lot of the delay in the current flow. 

**Describe alternatives you've considered**
NA

**Additional context**
This is similar to the offline replace feature #7703, but differs in that the supervisor would use their own device and the new CHW would get a new device provisioned for them.

Private docs:
* [Figma](https://www.figma.com/proto/igJBl04RYEsHiUMEehNyUy/User-management?node-id=1095%3A1876&scaling=scale-down&page-id=1026%3A961&starting-point-node-id=1031%3A1938)
* [Story Map](https://docs.google.com/presentation/d/1SsfNbuqPEVJhN-HGDEWOpA90yTPjYVA62q7vbVxr8MI/edit#slide=id.g1451dc5f7a9_0_0)
* [Discussion](https://docs.google.com/document/d/18l9175kuNlFsCiIhg2Rz7AMHrixfs6SBZWxkTut-ZGE/edit)

## Usage

Enabling this feature is similar to the existing [user-replace](https://docs.communityhealthtoolkit.org/apps/reference/app-settings/transitions/#create_user_for_contacts) functionality.  

- Load the updated code by either:
  - Checking out the [`7753-supervisor-chw-create`](https://github.com/medic/cht-core/tree/7753-supervisor-chw-create) branch and building it locally
  - Upgrading your instance to the latest feature release  `4.1.0-4.1.0-FR-supervisor-chw-create-beta.4` ([see below](https://github.com/medic/cht-core/issues/7753#issuecomment-1373024792))
- Update app_settings:
  - Enable token login:   `"token_login": {  "enabled": true }`
  - Set app url to your host (`192.168.1.249` in this case): `"app_url": "https://192-168-1-249.my.local-ip.co/",`
  - Enable create user for contacts transition: `"transitions": { "create_user_for_contacts": true }`
- Upload `default` config forms from the [`7753-supervisor-chw-create`](https://github.com/medic/cht-core/tree/7753-supervisor-chw-create) branch 
- Open the webapp and add a new person contact
[...]

PR #7988, opened 2022-12-15
feat: add support for auto-creating users for new contacts

m5r:
# Description

This is a new feature to give offline/online users the ability to create another user.  For most instances, this would be limited to supervisor and their ability to create a new CHW under them, however it can be configured however you want.  After the form is submitted locally, it will be pushed up to the server. The CHT server will process the request, create a user and then the new CHW will get an SMS with a [magic link](https://docs.communityhealthtoolkit.org/apps/concepts/access/#magic-links-for-logging-in-token-login) they can use to login.

Closes medic/cht-core#7753

[Docs PR](https://github.com/medic/cht-docs/pull/949)

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7753-supervisor-chw-create/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7753-supervisor-chw-create/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7753-supervisor-chw-create/docker-compose/cht-couchdb-clustered.yml)
 
# License


shared-libs/transitions/src/transitions/create_user_for_contacts.js
```diff
@@ -154,30 +154,49 @@ module.exports = {
...
-        changed: true,
-        message: errors
-          .map(error => error.message || JSON.stringify(error))
-          .join(', '),
-      };
+    const isReplacingUser = Boolean(change.doc.user_for_contact.replace);
+    if (isReplacingUser) {
+      const usersToReplace = getUsersToReplace(change.doc);
+      const replaceUsers = usersToReplace.map(user => replaceUser(change.doc, user));
+      const replacementResults = await Promise.allSettled(replaceUsers);
```
jkuester, line 178: I think we should work the add-user processing into this same `Promise.allSettled` (along with the error handling) so that a failure in the user-replace does not prevent the user-add from running.

shared-libs/transitions/src/transitions/create_user_for_contacts.js
```diff
@@ -154,30 +154,49 @@ module.exports = {
...
+          message: errors
+            .map(error => error.message || JSON.stringify(error))
+            .join(', '),
+        };
+      }
+      return true;
+    }
+
+    try {
+      const contact = await people.getOrCreatePerson(change.doc._id);
```
jkuester, line 194: I am pretty sure that `change.doc` already contains the contact doc.  Is there a reason we need to get it again here?

shared-libs/transitions/src/transitions/create_user_for_contacts.js
```diff
@@ -154,30 +154,49 @@ module.exports = {
...
+      await createNewUser({ roles: [change.doc.role] }, contact);
```
jkuester, line 195: Since we officially support multiple roles for a user, we should probably try loading from `change.doc.roles` first and only fall back to `role` if there is no `roles` array.

webapp/src/ts/services/contact-save.service.ts
```diff
@@ -55,6 +56,10 @@ export class ContactSaveService {
     }
 
     const doc = this.prepare(submitted.doc);
+    if (typeof original === 'undefined' && doc.should_create_user === 'true') {
+      doc.user_for_contact = {};
+      doc.user_for_contact.add = { status: UserCreationStatus.READY, roles: [doc.role] };
```
jkuester, line 61: Would be nice to move this code into the `create-user-for-contacts.service` (as a `setAddUser` function or something).

webapp/src/ts/services/contact-save.service.ts
```diff
@@ -55,6 +56,10 @@ export class ContactSaveService {
     }
 
     const doc = this.prepare(submitted.doc);
+    if (typeof original === 'undefined' && doc.should_create_user === 'true') {
```
jkuester, line 59: Not a big deal, but from the perspective of naming consistency, would it make sense to change `should_create_user` to `create_user_for_contact`?

jkuester, line 59:
```
typeof original === 'undefined'
```

This means we only perform this operation on new contacts.  Do we want that or should we support the possibility of adding a user when editing a contact? 

Also, I think we need to move this logic to the `prepare` function so that it gets run in any of the ways that a person is added via a contact form (e.g. if the person is created as a sibling doc at the same time the place is created or even if it is created in a repeat on the form).

jkuester: If we do support adding users while editing contacts, we need to make sure we _remove_ the `should_create_user` field here so that we do not keep creating users every time the contact is edited.

config/default/forms/contact/person-create.xml
```diff
@@ -821,6 +893,7 @@ DOB ISO: <output value=" /data/person/ephemeral_dob/dob_iso "/></value></text>
               <contact_id/>
               <facility_id/>
               <name/>
+              <ddd/>
```
jkuester, line 896: We don't need this field anymore right?

config/default/forms/contact/person-create.xml
```diff
@@ -893,12 +968,13 @@ DOB ISO: <output value=" /data/person/ephemeral_dob/dob_iso "/></value></text>
...
-      <bind nodeset="/data/person/phone" type="tel" jr:constraintMsg="jr:itext('/data/person/phone:jr:constraintMsg')" constraint="true()"/>
-      <bind nodeset="/data/person/phone_alternate" type="tel" jr:constraintMsg="jr:itext('/data/person/phone_alternate:jr:constraintMsg')" constraint="true()"/>
+      <bind nodeset="/data/person/phone" type="tel" constraint="true()" jr:constraintMsg="jr:itext('/data/person/phone:jr:constraintMsg')"/>
+      <bind nodeset="/data/person/phone_alternate" type="tel" constraint="true()" jr:constraintMsg="jr:itext('/data/person/phone_alternate:jr:constraintMsg')"/>
       <bind nodeset="/data/person/role" type="select1"/>
       <bind nodeset="/data/person/role_other" type="string" relevant="selected(  /data/person/role ,'other')" required="true()"/>
       <bind nodeset="/data/person/external_id" type="string"/>
       <bind nodeset="/data/person/notes" type="string"/>
+      <bind nodeset="/data/person/should_create_user" type="select1" required="true()" calculate="if(selected(  /data/person/role ,'chw') or selected(  /data/person/role ,'chw_supervisor'), 'true', 'false')" readonly="true()"/>
```
jkuester, line 977: You have this field setup with `type = select_one yes_no` and with the `hidden` appearance.  I think it would be more simple to just change the type of the field to `calculate` and then clear out the `appearance` (and the `required` and `read_only`).  That should achieve the same behavior, but in a more straightforward manner.

jkuester, line 977: Also, for the `calculate` here, I think you could simplify it to just say: `selected( ${role},'chw') or selected( ${role},'chw_supervisor')`.  That should give you a boolean value in the `should_create_user` field on the resulting contact doc....

jkuester (changes requested):
Looking good!  Left some initial thoughts below....

shared-libs/transitions/src/transitions/create_user_for_contacts.js
m5r: You're right I double checked and they're the same doc

shared-libs/transitions/src/transitions/create_user_for_contacts.js
```diff
@@ -154,30 +154,49 @@ module.exports = {
...
-      .map(({ reason }) => reason);
-    if (errors.length) {
-      throw {
-        changed: true,
-        message: errors
-          .map(error => error.message || JSON.stringify(error))
-          .join(', '),
-      };
+    const isReplacingUser = Boolean(change.doc.user_for_contact.replace);
+    if (isReplacingUser) {
```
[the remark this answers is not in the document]
m5r: But the transition's `filter` function already covers this part from line 161 to line 169  
Unless there are cases where transitions are ran despite the `filter` returning false, I think just checking for the presence of `change.doc.user_for_contact.replace` in `onMatch` should be sufficient

shared-libs/transitions/src/transitions/create_user_for_contacts.js
```diff
@@ -154,30 +154,49 @@ module.exports = {
...
       || !doc.user_for_contact
-      || !doc.user_for_contact.replace
     ) {
       return false;
     }
 
-    return !!Object
-      .values(doc.user_for_contact.replace)
-      .find(({ status }) => status === USER_CREATION_STATUS.READY);
+    if (doc.user_for_contact.replace) {
```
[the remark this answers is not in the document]
m5r: Good catch. It should still allow to add a user for the contact if there is no replacement going on.

```ts
if (doc.user_for_contact.replace && doc.user_for_contact.add) {
  const hasFinishedReplacing = Object
    .values(doc.user_for_contact.replace)
    .every(({ status }) => [USER_CREATION_STATUS.COMPLETE, USER_CREATION_STATUS.ERROR].includes(status));
  const isAddingUser = doc.user_for_contact.add.status === USER_CREATION_STATUS.READY;

  return hasFinishedReplacing && isAddingUser;
}
```

shared-libs/transitions/src/transitions/create_user_for_contacts.js
m5r: The `Promise.allSettled` makes sense for the replacement because we got an array of promises creating potentially many users but in this case a simple `try`/`catch` should be enough, no need to go through the gymnastics of creating an array of promises and looping through an array of errors when we know we'll always have just one element.

I agree with you we should be consistent with error handling though and I made a few changes to keep the behavior consistent with the replace code and extracted the couple of lines there into a `addUser` function

shared-libs/transitions/src/transitions/create_user_for_contacts.js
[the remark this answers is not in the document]
jkuester: Is there a particular reason we should prevent adding a new user _at the same time_ as we are replacing one?  I am not sure it would ever be possible to get into this situation, but it seems unnecessary to have extra logic to prevent that from happening....

shared-libs/transitions/src/transitions/create_user_for_contacts.js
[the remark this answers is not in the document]
jkuester: I think we should support adding a new users for a contact even if the contact already had a user that was replaced.  In that case, the filter would return `true` (when `doc.user_for_contact.add.status === USER_CREATION_STATUS.READY`), but the `change.doc.user_for_contact.replace` would also exist for the previously replaced user....

shared-libs/transitions/src/transitions/create_user_for_contacts.js
jkuester: I un-resolved this conversation since I think it depends on the outcome of the earlier thread regarding support for  adding and replacing users for the same contact.  If we decide we do not want to support that, we can re-resolve this conversation.  If we end up wanting to support adding and replacing at the same time, we should revisit this logic....

shared-libs/transitions/src/transitions/create_user_for_contacts.js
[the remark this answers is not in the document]
jkuester: One thing to keep in mind here is the behavior of Sentinel when syncing multiple changes to a doc from an offline user.  Imagine you are offline and make changes A, B, and C to a doc. Then I come online and sync back to the server.  Sentinel will only submit the doc once (with the C revision) to the transitions.  So if a user is offline and triggers and _add_ and a _replace_ and then comes online and syncs, then I think we want the transition to be able to handle both these actions at the same time....

webapp/src/ts/services/transitions/test-user-for-contact.transition.ts
```diff
@@ -0,0 +1,41 @@
...
+  init() {
+    return true;
+  }
+
+  filter(docs: Doc[]) {
+    if (!docs) {
+      return false;
+    }
+
+    return !!docs.filter(doc => doc?.create_user_for_contact === 'true').length;
```
jkuester, line 26: I think it is probably worth checking to see if the doc is actually a person contact: `this.contactTypesService.isPersonType(doc);`

webapp/src/ts/services/transitions/test-user-for-contact.transition.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  async run(docs: Doc[]) {
+    if (!docs) {
+      return [];
+    }
+
+    docs = docs.filter(Boolean).map(doc => {
```
jkuester, line 34: Since multiple docs can be added/updated here in this list we need to find the ones that actually need the user created.  I think I would extract the logic from the `filter` method above and reuse it down here to filter down to just the docs we need to update.

jkuester, line 34: Can this just be a `forEach` and not a `map`?

jkuester (changes requested):

shared-libs/transitions/src/transitions/create_user_for_contacts.js
```diff
@@ -127,6 +127,11 @@ const getUsersToReplace = (originalContact) => {
     .map(([username, { replacement_contact_id }]) => ({ username, replacementContactId: replacement_contact_id }));
 };
 
+const addUser = async (contact) => {
+  const roles = Array.isArray(contact.roles) && contact.roles.length > 0 ? contact.roles : [contact.role];
```
jkuester, line 131: Unexpectedly (to me) if the contact does not have `roles` or `role`, then `[contact.role]` is technically `[ undefined ]` (and had a length of 1) instead of just being an empty array.  This causes a kind of weird sounding error when trying to create the user.  Instead of something normal like `Missing required fields: roles` you get the message: `doc.roles can only contain strings`.

This could be confusing to debug.  I think maybe we want to actually just bake the role requirement into the logic here and throw an error from the transition if there is no `role` or `roles` (and not call `createNewUser` at all in that case....

jkuester (changes requested):

shared-libs/transitions/src/transitions/create_user_for_contacts.js
```diff
@@ -148,36 +153,68 @@ module.exports = {
...
-      || !doc.user_for_contact.replace
     ) {
       return false;
     }
 
-    return !!Object
-      .values(doc.user_for_contact.replace)
-      .find(({ status }) => status === USER_CREATION_STATUS.READY);
+    const isReplacingUser = !!doc.user_for_contact.replace;
+    const isAddingUser = doc.user_for_contact.create && doc.user_for_contact.create.add === 'true';
```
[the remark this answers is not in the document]
jkuester: > I'm not 100% convinced about the truthy check. Right now, the form sets the property to 'true' (string) and I haven't seen it behave any differently. Do we expect projects to configure forms that set the value to something different?

Ah, I forgot that for some reason Enketo (or maybe it is our JSON conversion?) will still end up setting a string value for a boolean field....  In that case I think it makes sense to leave the `=== 'true'` check.  (Also worth noting that the Enketo service does a [similar check](https://github.com/medic/cht-core/blob/7e28a83a1164b8dfa04eae26fdc9f4af356fe448/webapp/src/ts/services/enketo.service.ts#L567) for the string value of a boolean field...)

shared-libs/transitions/src/transitions/create_user_for_contacts.js
```diff
@@ -148,36 +153,68 @@ module.exports = {
...
   onMatch: async change => {
-    const usersToReplace = getUsersToReplace(change.doc);
-    const replaceUsers = usersToReplace.map(user => replaceUser(change.doc, user));
-    const replacementResults = await Promise.allSettled(replaceUsers);
-    const errors = replacementResults
-      .filter(({ status }) => status === 'rejected')
-      .map(({ reason }) => reason);
-    if (errors.length) {
+    const isReplacingUser = Boolean(change.doc.user_for_contact.replace);
+    if (isReplacingUser) {
```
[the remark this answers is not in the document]
m5r: I've been mulling this over and I don't think we should run those two actions in parallel. We could run into a race condition where the user wasn't created yet and the replace fails to process because of the not-yet-existing user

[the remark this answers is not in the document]
m5r: I'm coming up with something that keeps the syntax as nice and readable as your suggestion but runs each action sequentially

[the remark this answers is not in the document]
jkuester: > We could run into a race condition where the user wasn't created yet and the replace fails to process because of the not-yet-existing user

It if fine with me to not run these processes in parallel, but I did want to note that I don't think we have to worry about this scenario where we need to replace the newly created user.  This is because the data structure in `user_for_contact.replace` requires the actual _username_ of the user being replaced (which will not exist for the new user before we randomly generate it when creating the user).  I think this means it is not possible to create and replace the _same user_ at the _same time_.

m5r:
@jkuester I think we're good for rebasing off master to resolve those conflicts and get someone else to review :eyes:

config/default/forms/contact/person-create.xml
```diff
@@ -899,6 +902,7 @@ DOB ISO: <output value=" /data/person/ephemeral_dob/dob_iso "/></value></text>
       <bind nodeset="/data/person/role_other" type="string" relevant="selected(  /data/person/role ,'other')" required="true()"/>
       <bind nodeset="/data/person/external_id" type="string"/>
       <bind nodeset="/data/person/notes" type="string"/>
+      <bind nodeset="/data/person/user_for_contact/create" type="string" calculate="selected(  /data/person/role ,'chw') or selected(  /data/person/role ,'chw_supervisor')"/>
```
jkuester, line 905: Should we make `phone` required if `create` is true? (Since the user creation will fail without a phone)?  Kinda tricky question since the `create_user_for_contacts` transition is not enabled by default (but there is no way to check that from within the form....)

jkuester (changes requested):

shared-libs/transitions/src/transitions/create_user_for_contacts.js
```diff
@@ -148,26 +165,31 @@ module.exports = {
...
       || !doc.user_for_contact
-      || !doc.user_for_contact.replace
     ) {
       return false;
     }
 
-    return !!Object
-      .values(doc.user_for_contact.replace)
-      .find(({ status }) => status === USER_CREATION_STATUS.READY);
+    return Boolean(isCreatingUser(doc, info) || isReplacingUser(doc));
```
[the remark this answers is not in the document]
m5r: `isReplacingUser` can return `undefined` which is falsey but not `false` and I wanted to make sure the function returns a boolean mainly because one of the tests was complaining it was receiving `undefined` instead of `false`

shared-libs/transitions/src/transitions/create_user_for_contacts.js
```diff
@@ -127,6 +128,22 @@ const getUsersToReplace = (originalContact) => {
     .map(([username, { replacement_contact_id }]) => ({ username, replacementContactId: replacement_contact_id }));
 };
 
+const addUser = async (contact) => {
+  if (!contact.roles && !contact.role) {
+    throw new Error(`Contact [${contact._id}] must have a "role" or "roles" property.`);
+  }
+
+  const roles = Array.isArray(contact.roles) && contact.roles.length > 0 ? contact.roles : [contact.role];
```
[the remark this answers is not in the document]
m5r: `Array.isArray()` also makes sure we don't try to access `.length` if `contact.roles` is undefined. Plus if somehow someone passes a string, the `length > 0` check alone would pass even though we're expecting an array

config/default/forms/contact/person-create.xml
jkuester: Discussed with the @m5r and @tatilepizs.  Decided that the best thing to do is actually just make `user_for_contact/create` dependent on the `phone`.  So, we will only mark `create = 'true'` if a phone number is provided for the contact.

tests/integration/sentinel/transitions/create-user-for-contacts.spec.js
```diff
@@ -183,68 +174,78 @@ describe('create_user_for_contacts', () => {
...
-    // Can still log in as other user
-    assert.include(await loginAsUser(otherUser), { statusCode: 302 });
+    // Login tokens sent
+    const queuedMsgs = await messagesUtils.getQueuedMessages();
+    assert.lengthOf(queuedMsgs, 2);
+    queuedMsgs.forEach(msg => assert.equal(msg.type, 'token_login'));
+    const replacedUserMsg = queuedMsgs.find(msg => msg.user === newUserSettings._id);
+    assert.equal(replacedUserMsg.tasks[0].messages[0].to, NEW_PERSON.phone);
+    const additionalUserMsg = queuedMsgs.find(msg => msg.user === additionalUserSettings._id);
+    assert.equal(additionalUserMsg.tasks[0].messages[0].to, additionalPerson.phone);
```
dianabarsan, line 200: I think this needs a little bit of attention. 

As I said in a comment above, checking infodoc initial_replication_date vs last_replication_date will probably skip processing a lot of docs if sentinel is slower than edits come it. I changed the code to check if any transitions had run instead. 
So now I have this scenario: 

- Enable the transition and turn off sentinel. 
- I create a contact with phone number "1234"
- From the admin app I create a user for this contact, and set their phone number to "1234" as well. 
- Login with this user. 
- Submit a user replacement form. The new contact will also have phone number "1234". 
- Sync. 
- Start sentinel. 

What I'm seeing as a result is that two token login links were sent to "1234", one for the create and one for the replace. 
What is the user supposed to do in this situation? Which of the SMS-s is going to be correct?

tests/utils.js
```diff
@@ -977,6 +977,14 @@ module.exports = {
     return results;
   },
 
+  saveDocIfNotExists: async doc => {
+    try {
+      await module.exports.getDoc(doc._id);
+    } catch (_) {
+      await module.exports.saveDoc(doc);
+    }
+  },
```
dianabarsan, line 986: Nit. It's cheaper to try and save the doc omitting the _rev field. If the doc exists, it'll throw a conflict that you would ignore, and you end up excluding the get request.

dianabarsan (changes requested):
Very nice work! 

I found two things that I'm worried about and have detailed below.

shared-libs/transitions/src/transitions/create_user_for_contacts.js
```diff
@@ -148,26 +169,31 @@ module.exports = {
...
-    const replacementResults = await Promise.allSettled(replaceUsers);
-    const errors = replacementResults
+  onMatch: async ({ doc, info }) => {
+    const promises = [];
+
+    if (isCreatingUser(doc, info)) {
+      promises.push(addUser(doc));
+    }
+
+    if (isReplacingUser(doc)) {
```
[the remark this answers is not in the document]
dianabarsan: I've just realised I'm touching this issue in more detail in another comment. Please feel free to ignore this one.

shared-libs/transitions/src/transitions/create_user_for_contacts.js
```diff
@@ -127,6 +128,26 @@ const getUsersToReplace = (originalContact) => {
...
+  }
+
+  const roles = Array.isArray(contact.roles) && contact.roles.length > 0 ? contact.roles : [contact.role];
+  await createNewUser({ roles }, contact);
+  delete contact.user_for_contact.create;
+};
+
+const isCreatingUser = (contact, info) => contact.user_for_contact.create === 'true'
+  && !transitionUtils.hasRun(info, NAME)
+  && info.initial_replication_date === info.latest_replication_date;
```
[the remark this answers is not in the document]
jkuester: Well, shoot!  For some reason I thought that `latest_replication_date` was set in Sentinel...

Unfortunately, I don't think we can just check for the existence of `transitions` since that value seems to be [set automatically](https://github.com/medic/cht-core/blob/a43f4fb3c32fb4286d4a939c01fa3457516824ae/shared-libs/infodoc/src/infodoc.js#L97) when you get the info doc...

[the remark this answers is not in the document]
jkuester: This is insane.  As far as I can tell there is no way to reliably check if a doc has been seen by Sentinel before!  I think we are either going to have to build that functionality out or we just need to give up on the requirement that you can only add users for new contacts (and just let you add users for existing contacts too).  I am going to take this question back to the issue....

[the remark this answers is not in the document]
dianabarsan: That code does not run in API, fortunately. 
These are the only two functions that are called by API: https://github.com/medic/cht-core/blob/a43f4fb3c32fb4286d4a939c01fa3457516824ae/shared-libs/infodoc/src/infodoc.js#L290-L291

I think it's possible to add something that will indicate that this is the first time sentinel has ever processed the document. If you can't leverage something that already exists, just add something new. Luckily this doesn't have to be backwards compatible with anything, since you don't care about docs created prior to this change to be evaluated as new.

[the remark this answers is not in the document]
jkuester: > That code does not run in API, fortunately.

I am not sure why API matters though, since it [gets run by Sentinel](https://github.com/medic/cht-core/blob/d3ee100dc0fb27347c48b8126ed78c788ae234cf/shared-libs/transitions/src/transitions/index.js#L53) before executing the transitions.  Either way, we `transitions` is not going to be a viable check and Rube Goldberg is proud....

I guess I can "solve" my problem by making another random edit to the info doc in Sentinel to set some magic flag.  :facepalm:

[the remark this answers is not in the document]
dianabarsan: > random edit to the info doc in Sentinel to set some magic flag.

I think that if you add the "flag" here, it won't even ever get saved: https://github.com/medic/cht-core/blob/a43f4fb3c32fb4286d4a939c01fa3457516824ae/shared-libs/infodoc/src/infodoc.js#L137

because we only ever update specific properties of the infodocs.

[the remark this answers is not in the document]
dianabarsan: Either way this a useful addition, it even seems silly you _can't_ tell whether this is the first time Sentinel has processed a certain document. 
You could even consider adding a specific date field to the infodoc with `processed_by_sentinel`.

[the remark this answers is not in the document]
jkuester: Okay, so here is the approach that I am trying:
- Update `infodoc` to not auto-populate `info.transitions` when calling `infodoc.get`
- That means that when the transitions lib _starts_ processing a doc change it can know if it has processed the doc before based on the presence of `info.transitions`
    - Based on that check, the transitions lib can set a `change.initialProcessing` flag which will allow the individual transitions to know if this is the first time the doc is being processed by Sentinel
- Once the transitions lib finishes processing a change it will save the `infodoc.transitions` value to the DB so that subsequent runs can properly set the `initialProcessing` flag to false

I like this approach because the behavior of `infodoc.get` is less surprising. We don't end up adding any extra flags to the info doc in the DB (beyond the existing `transitions`). And the `initialProcessing` flag logic can be localized to the transitions util.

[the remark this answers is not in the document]
dianabarsan: I think this is good. I don't think we actually need the transitions property before writing the first transition result, and all the rest of the code should already account for a missing property, since it initially does not exist in the infodoc.

shared-libs/transitions/test/unit/transitions/create_user_for_contacts.js
```diff
@@ -588,5 +727,98 @@ describe('create_user_for_contacts', () => {
         });
       });
     });
+
+    it('creates user and replaces user for the same contact at the same time', async () => {
```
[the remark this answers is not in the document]
jkuester: @dianabarsan your scenario here seems more simple (since there is no `replace` happening).  Even still, I agree that this seems like a plausible scenario that could happen.  Unfortunately, I don't think there is an easy fix to this behavior.  We would need to be able to check and see if there is already an existing user for that contact. Based on similar discussions we had during the replace project, that might mean a new Couch view to query on.  Seems like quite a bit of work just to avoid extra users created in an edge case.  

That being said, my main goal is to have things behave in the _least surprising way._  So, let me know which behavior seems best here!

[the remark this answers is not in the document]
dianabarsan: > able to check and see if there is already an existing user for that contact

Off the top of my head, an alternative would be for the users api, that will be used for creating any user, regardless, checks whether the contact has the create new user flag and remove it (the flag)? How messy would the code get?

tests/integration/sentinel/transitions/create-user-for-contacts.spec.js
jkuester: I am not sure if in your mind this situation is different now that we are not using initial/last_replication_date, but I think it is worth noting that the new user from the `replace` flow will only have the _same phone_ as the `create` user if the same phone number is used when creating the _new replacement contact_.  An alternative scenario would be that the new replacement contact is given a different phone and the token login links go to different phone numbers.  

We could add logic to the transition to only run both `create` and `replace` at the same time if they are going to different phone numbers, but that feels overly complex.  I would rather we just decide on an agreed upon default behavior and go with it:

1. Run both `create` and `replace`
2. Only run `create`
3. Only run `replace`
4. Run neither

None of these should be hard to implement.  Mokthar and I were originally thinking that `1.` was going to be least surprising, but if you prefer one of the other options, I am happy to go that direction.

tests/utils.js
jkuester: I think in this case it is better to just leave this how it is.  The problem is that `saveDoc` has a bunch of extra `formDocProcessing` logic.  If you call it with a form doc that does not exist, it will not quickly fail, but will eventually timeout after 6000ms with the error `Timed out looking for details in logs.` because `formDocProcessing` is looking for stuff about the doc in the API logs....

We could roll-our-own `saveDoc` for this function, but that seems unnecessary....

tests/integration/sentinel/transitions/create-user-for-contacts.spec.js
dianabarsan: It's a difficult problem. 

I think that 1) produces the undesirable scenario that I described above, where a single phone number will get two SMSs at the same time, after they had been logged out, with only one of them being correct (to the replaced user) and the other one being to the created user.  
In this scenario, the user getting the SMS from the replace (the last action that was added) would have been correct. 

I would go for the last action that was added to the contact, if possible. This would mean that: replace removes create and create removes replace, so that only one of them (the last one) gets run. 

If that's not possible, I would just run the replace - especially since we're kind of designing this to only work on new contacts.

dianabarsan: > worth noting that the new user from the replace flow will only have the same phone as the create user if the same phone number is used when creating the new replacement contact. 

I used the forms that were provided in the tests/default config, the ones that create a new contact within the replace form. That form automatically copies the phone number from the replaced contact onto the new contact, does it not?

shared-libs/transitions/test/unit/transitions/create_user_for_contacts.js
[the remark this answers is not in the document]
jkuester: I like this idea!  I have pushed the changes to the branch. Does not seem too messy!

tests/utils.js
dianabarsan: I had forgotten about adding the form update watcher in saveDoc. Please disregard :D

dianabarsan (commented):
This is coming along beautifully! I'm very happy with this. 
All of my comments are about cleanup or just nits and reflections.

shared-libs/transitions/src/transitions/create_user_for_contacts.js
```diff
@@ -127,6 +131,26 @@ const getUsersToReplace = (originalContact) => {
...
+  ) {
+    throw new Error(`Contact [${contact._id}] must have a "role" or "roles" property.`);
+  }
+
+  const roles = Array.isArray(contact.roles) && contact.roles.length > 0 ? contact.roles : [contact.role];
+  await createNewUser({ roles }, contact);
+};
+
+const isCreatingUser = ({ doc, info, initialProcessing }) =>
+  doc.user_for_contact.create === 'true'
```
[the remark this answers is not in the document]
jkuester: Just got to get that optional chaining!  :laughing:

tests/integration/sentinel/transitions/create-user-for-contacts.spec.js
jkuester: Technically, the `replace` should always be the _last_ one that is set (since `create` should only be set when actually first writing the contact).  For the sake of simplicity, I have just updated the transition to only run `replace` if both are set.

dianabarsan: Excellent! 
Thanks a lot!

dianabarsan (approved):
Awesome work! I think this is a really nice feature and hope it gets put to good use! 
Thanks a lot for your patience and incorporating all feedback. 🌟

PR #7987, opened 2022-12-15
Set default contact script: use v1 API for CHT 3.x compatabaility

mrjones-plip:
# Description

Prior work in this branch used `/api/v2/users` which didn't exist in 3.x.  By using `/api/v1/users` I see it's backwards compatible which opens up more possible users of this script. 

# Code review checklist
- [X] Tested: Unit and/or e2e where appropriate

# Compose URLs

If Build CI hasn't passed, these may 404:

* __CHT_CORE_COMPOSE_URL__
* __COUCH_SINGLE_COMPOSE_URL__
* __COUCH_CLUSTER_COMPOSE_URL__
 
# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

PR #7989, opened 2022-12-15
Default contact script:  api v2 -> v1 change, work with sparser place objects

mrjones-plip:
# Description

Fixes  to #7987

* fixes readme to reference `index.js` vs `index.cjs`
* checks for  `place.contact` not being set
* downgrades use of User API from `/v2` to `/v1` so it works with earlier 3.x instances.

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:set-default-contact-script/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:set-default-contact-script/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:set-default-contact-script/docker-compose/cht-couchdb-clustered.yml)
 
# License

### DIFF
--- a/scripts/set-default-contact/index.js
+++ b/scripts/set-default-contact/index.js
@@ -50,7 +50,7 @@ const getObjectFromMedicDb = async id => {
 const hasDefaultContact = async user => {
   if(typeof user.place === 'object' && typeof user.contact === 'object' &&  user.place._id  ){
     const place = await getObjectFromMedicDb(user.place._id);
-    return !((typeof place.contact === 'object' && !place.contact._id) || typeof place.contact !== 'object');
+    return place.contact && place.contact._id;
   }
   // return true for invalid users like admin or medic so we don't process them
   return true;
--- a/scripts/set-default-contact/index.js
+++ b/scripts/set-default-contact/index.js
@@ -15,7 +15,7 @@ const compileUrl = path => {
 };
 
 const getChtUsers = async () => {
-  console.log('   Using URL taken from COUCH_URL env var: ', process.env.COUCH_URL);
+  console.log('   Using COUCH_URL env var:', process.env.COUCH_URL);
   console.log('');
   //  pin  to v1 of API so it is backwards compatible with CHT 3.x
   const url = compileUrl('/api/v1/users');
--- a/scripts/set-default-contact/index.js
+++ b/scripts/set-default-contact/index.js
@@ -17,7 +17,7 @@ const compileUrl = path => {
 const getChtUsers = async () => {
   console.log('   Using URL taken from COUCH_URL env var: ', process.env.COUCH_URL);
   console.log('');
-  const url = compileUrl('/api/v2/users');
+  const url = compileUrl('/api/v1/users');
   const options = {
     uri: url.href,
     json: true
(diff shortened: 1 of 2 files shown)


scripts/set-default-contact/index.js
```diff
@@ -49,7 +49,7 @@ const getObjectFromMedicDb = async id => {
 const hasDefaultContact = async user => {
   if(typeof user.place === 'object' && typeof user.contact === 'object' &&  user.place._id  ){
     const place = await getObjectFromMedicDb(user.place._id);
-    return !(typeof place.contact === 'object' && !place.contact._id);
+    return !((typeof place.contact === 'object' && !place.contact._id) || typeof place.contact !== 'object');
```
jkuester, line 52: TLDR: I think we can simplify this logic:
```js
    return typeof place.contact === 'object' && place.contact._id;
```

Steps:
- We start with:
  - `!((typeof place.contact === 'object' && !place.contact._id) || typeof place.contact !== 'object')`
- Use DeMorgan's law to distribute the outer `!`:
  - `!(typeof place.contact === 'object' && !place.contact._id) && typeof place.contact === 'object'`
- Use DeMorgan's law again on the inner portion:
  - `(typeof place.contact !== 'object' || place.contact._id) && typeof place.contact === 'object'`
- Reduce since we cannot have both !x AND x:
  - `place.contact._id && typeof place.contact === 'object'`

jkuester, line 52: That being said, is there a special reason we are checking `place.contact === 'object'`?  Is there ever a valid case where `contact` would be a primitive value (like a UUID string)?  If so, should we consider a place with such a `contact` value to have a default contact or not?  

If we have no reason to expect that `place.contact` would ever be primitive (or if a primitive value for `place.contact` should never be considered a default contact), then I feel like this check can simply be:

```suggestion
    return place.contact && place.contact._id;
```

jkuester (changes requested):
Couple minor thoughts inline!  Good call on making this script backwards compatible!

scripts/set-default-contact/index.js
```diff
@@ -17,7 +17,7 @@ const compileUrl = path => {
 const getChtUsers = async () => {
   console.log('   Using URL taken from COUCH_URL env var: ', process.env.COUCH_URL);
   console.log('');
-  const url = compileUrl('/api/v2/users');
+  const url = compileUrl('/api/v1/users');
```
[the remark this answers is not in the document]
mrjones-plip: yes! good call ceaa19327

scripts/set-default-contact/index.js
mrjones-plip: Love this!  That line made me feel a little nauseous - this approach is wonderfully clarifying. Thanks for taking the time to walk me through it.

Follow up below

mrjones-plip: With the version before this PR:

```javascript
!((typeof place.contact === 'object' && !place.contact._id)
```

I was trying to say, "If the place has a `contact` object, but it's `_id` is not set".  This is because I checked and  the CHT does this in the `default` config when it creates a new place:

```json
{
  _id: '80b7dc4a-434e-470d-b49f-a39a96b66c1b',
  _rev: '6-78f0646dc4df774c4498b874e0afd2e7',
  contact: { _id: '' },
}
```

However, then I found out that we can also NO `contact` object at all, say when you use `cht-conf` to import JSON docs:


```json
{
  _id: '80b7dc4a-434e-470d-b49f-a39a96b66c1b',
  _rev: '6-78f0646dc4df774c4498b874e0afd2e7'
}
```

So then I was like, "If the place has a `contact` object, but it's `_id` is not set OR `contact` isn't set at all' (aka not an object)" . 

And here we are!  

Let's check in real-time tomorrow (if you're free) to make sure we're both on the same page! :paperclip:

jkuester (approved):
:rocket:

craig-landry:
The failing tests have nothing to do with the script that was changed here. I'm going to force merge this now.

Issue #7992, opened 2022-12-20
Upgrade webapp's angular to v15.x.x


latin-panda:
**Describe the issue**
The current version is v12.x.x, we need to keep it updated. Latest version is v15

latin-panda:
- Running time of unit tests
  - Angular 12:  
    - `grunt unit-webapp` total time: 3m 31.3s 
    - `grunt unit-webapp-no-dependencies` total time: 1m 35.5s
  - Angular 15: 
    - `grunt unit-webapp` total time: **2m 47.6s**  
    - `grunt unit-webapp-no-dependencies` total time: **56.8s**
- Build time
  - Angular 12: 
    - `grunt dev-webapp` total time: 1m 80s 
    - `grunt dev-webapp-no-dependencies` total time: 49.46s
  - Angular 15: 
    - `grunt dev-webapp` total time: **40.46s**  
    - `grunt dev-webapp-no-dependencies` total time: **26.82s**
- Size
  - Angular 12: 
    - Initial total: 5.53 MB
    - `main.js` 13.97 MB
  - Angular 15: 
    - Initial total: **14.00 MB**
    - `main.js` **12.24 MB**


**Common error while upgrading Angular**
Whenever we have the error below, it's likely that `@angular/compiler`, `@angular-devkit/build-angular`  or `@angular-builders/custom-webpack` aren't resolved properly in the `package-lock.json`. So, uninstall these 3, then install them again in that order.

![Screen Shot 2022-12-22 at 12 36 21 pm](https://user-images.githubusercontent.com/66472237/209088049-aaf33272-dd0c-4640-9917-a309c5782eea.png)

jkuester:
> Common error while upgrading Angular

Should we add this note to the [Updating Dependencies docs](https://docs.communityhealthtoolkit.org/contribute/code/core/update-dependencies/)?

latin-panda:
Good idea, I'm going to do that tomorrow.

This is ready for AT in [this PR](https://github.com/medic/cht-core/pull/7999), besides testing general functionality, please also:
- test on mobile
- and also performance 
- 
More details in the PR's description.

latin-panda:
Following up on @jkuester suggestion, I've [opened a PR](https://github.com/medic/cht-docs/pull/942) to include details about angular's exception when upgrading and a tip to fix it.

tatilepizs:
### Test details

Everything is working as usual, nothing different 🎉 

- Default config (webapp and phone, android 10)
[...]

PR #7999, opened 2022-12-22
Upgrade webapp's angular to v15

latin-panda:
# Description

Upgrading from v12 to v13:
- All @angular/XX libraries were upgraded to v13
- [Typescript was upgraded to v4.4.2](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html) which is the minimum required by Angular 13
- Angular 13 now creates a `.angular/cache` directory [to improve build speed](https://blog.angular.io/angular-v13-is-now-available-cce66f7bc296#05df). This was added to the gitignore file.
- The ` MethodCall` from angular compiler [was deprecated here](https://github.com/angular/angular/pull/42882/files#diff-1002bdc266028e1c303492a63ed5990124b8fd29a8f87098d07f86c0519d47f0R319), I've replaced it with `Call`
- Now there's an error when trying to build the app and when there's a `window.location.reload(true|false)` [this parameter isn't standard](https://developer.mozilla.org/en-US/docs/Web/API/Location/reload), I've removed it.

Upgrading from v13 to v14:
- [Typescript was upgraded to 4.6](https://devblogs.microsoft.com/typescript/announcing-typescript-4-6/#breaking-changes) which is the minimum required for Angular 14.
- The typescript eslint libs were upgraded to support the new version of typescript.
- I fixed the unit tests, and now we need to reset the store in the `beforeEach`.
- Also, the new version of ngrx changed a bit how [tests are done](https://ngrx.io/guide/store/testing#testing-selectors) in selectors, that’s something I fixed.
- The [parser provider](https://github.com/medic/cht-core/blob/master/webapp/src/ts/providers/parse.provider.ts) was fixed, something changed in the [angular compiler classes](https://github.com/angular/angular/blob/50f95f831e7da84e2b879dd4b9b7764ca07ef0e7/packages/compiler/src/expression_parser/parser.ts).
- At this point I can see an improvement in speed, when running `grunt unit-webapp` on `master branch` (angular v12) it takes 3m 31.3s to complete, while on my branch (angular v14) it takes 2m 7s
- Removed `polyfills-es5.js`, since angular **drop support for IE11**, see [End of IE11 support](https://blog.angular.io/angular-v13-is-now-available-cce66f7bc296)

Upgrading from v14 to v15:
- Removed `relativeLinkResolution` from app router module, [depracation](https://angular.io/guide/update-to-version-15#relativelinkresolution-is-not-configurable-in-the-router).
- Upgraded typescript to 4.8 min supported by Angular v15
- Upgraded typescript eslint libs to match the new version of typescript. (@typescript-eslint/eslint-plugin: 5.43, @typescript-eslint/parser: 5.43)
- Upgraded NgRx libraries and the Ngx Translate Core lib to support Angular v15 (@ngrx/effects, @ngrx/store, @ngx-translate/core)
- Set the `tsconfig.base.json`'s target to ES2020 to support latest typescript, and fix the error: `Class constructor cannot be invoked without 'new'`
- Fixed unit test configuration to match how angular finds specs files. That's in angular.json and test.ts
- Fixed angular material styles and classes selectors, more in [mdc v15](https://material.angular.io/guide/mdc-migration).

**NOTE:**
Whenever we have the error below, it's likely that `@angular/compiler`, `@angular-devkit/build-angular`  or `@angular-builders/custom-webpack` aren't resolved properly in the `package-lock.json`. So, uninstall these 3, then install them again in that order.

Screen Shot 2022-12-22 at 12 36 21 pm

medic/cht-core#7992

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7992_upgrade_webapp_angular_to_v15/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7992_upgrade_webapp_angular_to_v15/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7992_upgrade_webapp_angular_to_v15/docker-compose/cht-couchdb-clustered.yml)
 
# License


latin-panda:
Hi @dianabarsan and @jkuester can you please review? Let me know any questions 
You can find here a time diff between v12 and v15 [in this comment.](https://github.com/medic/cht-core/issues/7992#issuecomment-1378422496)

webapp/tests/karma/ts/selectors/index.spec.ts
```diff
@@ -125,315 +125,336 @@ describe('Selectors', () => {
...
 
     it('should getEnketoStatus', () => {
-      expect(Selectors.getEnketoStatus(state)).to.deep.equal(clonedState.global.enketoStatus);
+      expect(Selectors.getEnketoStatus.projector(state.global.enketoStatus))
+        .to.deep.equal(clonedState.global.enketoStatus);
     });
 
     it('should getEnketoEditedStatus', () => {
-      expect(Selectors.getEnketoEditedStatus(state)).to.equal(clonedState.global.enketoStatus.edited);
+      expect(Selectors.getEnketoEditedStatus.projector(state.global.enketoStatus))
```
dianabarsan, line 228: I'm wondering why it's necessary to directly serve the direct parent of the state property we want to get. Is there no way we can test these selectors by passing the whole state and making sure they're indeed picking the correct value? These tests seem devalued now that we spoon-feed the value they're supposed to return.

webapp/tests/karma/ts/services/changes.service.spec.ts
```diff
@@ -59,7 +59,10 @@ describe('Changes service', () => {
     store = TestBed.inject(MockStore);
   }));
 
-  afterEach(() => sinon.restore());
+  afterEach(() => {
+    store.resetSelectors();
```
dianabarsan, line 63: This almost seems like the default. What happens if we don't reset selectors?

webapp/tests/karma/ts/providers/parse.provider.spec.ts
```diff
@@ -142,19 +143,36 @@ describe('Parse provider', () => {
...
+        show_delivery_form: true,
+      };
+      const summaryDeceased = {
+        alive: false,
+        muted: false,
+        show_delivery_form: true,
+      };
+
+      const trueResult = parse(expression, context, { user, contact, summary: summaryAlive });
+      const falseResult = parse(expression, context, { user, contact, summary: summaryDeceased });
```
dianabarsan, line 172: Thanks for updating this test!

webapp/src/ts/providers/parse.provider.ts
```diff
@@ -272,12 +282,16 @@ class ASTCompiler {
...
     const stmts = this.cStmts;
     const _args = [];
+
     for (const arg of ast.args) {
       _args.push(this.build(arg));
     }
-    const fn = this.build(ast.receiver);
+
+    const functionName = ast.receiver.name;
+    const parent = this.build(ast.receiver.receiver);
```
dianabarsan, line 291: `parent` is already used in browsers, as a global object in the execution environment. I'd prefer if we don't reuse this variable name.

dianabarsan (commented):
Great work! 

I'm testing this out and everything seems to work as expected. 
I am seeing some `expressionChangedAfterItWasChecked` errors though. 

For example, I got this one when I used browser.history.back() (via backspace) and loaded reports list: 

image

I gpt this one also when going back from a contact to editing a contact (edit contact, cancel, and then browser history back to the edit form). It also happened when I reloaded the page while having a contact form open. 

image

webapp/tests/karma/ts/selectors/index.spec.ts
latin-panda: Yes, agree, [this is](https://ngrx.io/guide/store/testing#testing-selectors) how they suggest to do it :(( 
I tried to find a way without changing too much...

It depends on how we define the selector, for example, the [getEnketoEditedStatus](https://github.com/medic/cht-core/blob/8e96ab03d51cf0b5c64632137ff0b6c417f1ec38/webapp/src/ts/selectors/index.ts#L41) can be created to always receive the full state: 
` getEnketoEditedStatus: createSelector((state) => state.global?.enketoStatus?.edited),`

Selectors will get more verbose

webapp/tests/karma/ts/services/changes.service.spec.ts
latin-panda: Other spec files will get contaminated stores and have side effects. It won't reset the store before every test like before

webapp/tests/karma/ts/selectors/index.spec.ts
dianabarsan: How would selectors need to change in order for tests to stay as they are? An example would help, please don't go on to change _every_ selector :D

webapp/tests/karma/ts/services/changes.service.spec.ts
latin-panda: The good side is that we can now make separate test cases using the store as it changes over actions.

dianabarsan: So sad. I think we should maybe consider adding a global store property and adding a global afterEach to reset it, at some point.

webapp/src/ts/providers/parse.provider.ts
latin-panda: 🤦🏽‍♀️ I missed that, I'll rename the variable. Thanks

dianabarsan: Internet Explorer used to throw errors 15 years ago if you overwrote `parent` :D

dianabarsan: No worries. I bet we overwrite it in other places in webapp code. I have flashbacks from the old days when I see it assigned to something else.

webapp/tests/karma/ts/selectors/index.spec.ts
latin-panda: I made a quick sample of selectors with unit tests. 

Screen Shot 2023-01-12 at 9 12 48 pm

What do you think?

dianabarsan: Option 2 seems to have best test coverage and is pretty close to what we already have, right?

dianabarsan: Option 1 is also really good, a little more verbose.

latin-panda: Okay, I'm going to update the selectors with Options # 2 - pass first level object of state. It's a good balance.

package.json
```diff
@@ -147,10 +147,6 @@
       "path": "./api/build/static/webapp/main.js",
       "maxSize": "950 kB"
     },
-    {
-      "path": "./api/build/static/webapp/polyfills-es5.js",
-      "maxSize": "55 kB"
```
jkuester, line 152: I am a little nervous about this.  I read the [blogpost](https://blog.angular.io/angular-v13-is-now-available-cce66f7bc296) you linked to above and agree that it only describes this as necessary for supporting IE11.  However, if it was only needed for IE, I would not have expected it to be present before this upgrade since our [`.browserlistrc`](https://github.com/medic/cht-core/blob/392ce01762dc90958e38e8e61c05f57550d2f819/.browserslistrc) is only targeting Chome/FF (no IE).  

What is more, I double-checked the [Browser Support](https://www.w3schools.com/js/js_es6.asp) listed for `es6` and it specified Chrome 58 as the first version with support for `es6`.  This seems to mean that Chrome 53-57 would still require `es5`.  IDK, maybe I am totally missing something here, but at least we need to make sure we test this carefully on some devices with really old webview versions....

webapp/src/ts/services/enketo.service.ts
```diff
@@ -825,7 +825,7 @@ export class EnketoService {
 
 interface ContactSummary {
   id: string;
-  xmlStr: string;
+  xmlStr?: string;
```
jkuester, line 828:
```suggestion
  xml: Document;
```

I think we just need to fix the name of this field.  AFICT, `xmlStr` is not ever used....

package.json
latin-panda: > This seems to mean that Chrome 53-57 would still require es5

I think we're assuming that core v4.x.x supports [Chrome and Firefox latest](https://docs.communityhealthtoolkit.org/core/releases/#dependencies) with the intention to start using modern js features. I'd expect partners to be at least 80+

webapp/src/ts/services/enketo.service.ts
latin-panda: That's correct! Done

webapp/src/ts/providers/parse.provider.ts
latin-panda: Done

webapp/tests/karma/ts/selectors/index.spec.ts
latin-panda: Done

latin-panda:
@dianabarsan about the `expressionChangedAfterItWasChecked`, I can reproduce it in v4.1.0 and master. [I've created an issue](https://github.com/medic/cht-core/issues/8022) with some steps and screenshots of the logs. Thanks

webapp/tests/karma/ts/services/changes.service.spec.ts
latin-panda: That's a good idea and we can include the `sinon.restore()` as it's very much in every spec file. We can do this in another PR.

latin-panda:
@dianabarsan and @jkuester thanks for the feedback, it's ready for another review :)

dianabarsan (approved):
Great work!

webapp/tsconfig.base.json
```diff
@@ -9,7 +9,7 @@
     "experimentalDecorators": true,
     "moduleResolution": "node",
     "importHelpers": true,
-    "target": "es5",
+    "target": "es2020",
```
craig-landry, line 12: This gets further transpiled later in our build chain, right? I'm just checking since es2020 won't work with the Chrome/WebView versions some of our programs have deployed.

jkuester: Based on [this conversation](https://github.com/medic/cht-core/pull/7999#discussion_r1068613363) I think we _might_ not need to worry about this any more....

Certainly the linked dependency guidelines specify that we only support `Chrome and Firefox latest` on `4.x`.  I think the idea is that the system webview/browser needs to be kept up to date (even on older devices).

package.json
jkuester: Related Warning I see when building the webapp:

> One or more browsers which are configured in the project's Browserslist configuration will be ignored as ES5 output is not supported by the Angular CLI. Ignored browsers: chrome 60, chrome 59, chrome 58, chrome 57, chrome 56, chrome 55, chrome 54, chrome 53

webapp/src/ts/providers/parse.provider.ts
```diff
@@ -156,11 +157,20 @@ class ASTCompiler {
...
-    const ast = this.cAst;
+  processLiteralPrimitive(ast) {
     return isString(ast.value) ? `"${ast.value}"` : ast.value;
   }
 
+  processUnaryLiteralPrimitive(ast) {
+    const literalValue = this.processLiteralPrimitive(ast.expr);
+
+    if (ast.operator === '-') {
+      return isString(literalValue) ? '-' + literalValue : (literalValue * -1);
```
jkuester, line 168: Do we need a unit test to cover the `isString... === true` side of this ternary?  The only time I see this line hit in the current parse tests, `literalValue` is a number...

jkuester (approved):
:rocket:

webapp/tsconfig.base.json
latin-panda: If we keep `es5` and build, it throws several errors with typescript v4.8 (min required for Angular 15).

package.json
latin-panda: Good catch, I've updated the `.browserlistrc` and the `isUsingSupportedBrowser` functions to support:
```
chrome: '>=80', // Chrome 80 was released on February 4, 2020; for desktop and Android.
firefox: '>=98', // Firefox 93 was released on October 5, 2021; for desktop and Android.
```
Chrome v80 seems the safest after comparing the es2020 features with the browser support. Firefox stays the same as before.  That gives 3 years of Chrome versions to today.

latin-panda: Angular actually [suggests](https://angular.io/guide/build#configuring-browser-compatibility) 
```
last 1 Chrome version
last 1 Firefox version
```

But it's probably too much :)

latin-panda: Angular detected the new Chrome version in `.browserlistrc` and the following warning is gone.
> One or more browsers which are configured in the project's Browserslist configuration will be ignored as ES5 output is not supported by the Angular CLI. Ignored browsers: chrome 60, chrome 59, chrome 58, chrome 57, chrome 56, chrome 55, chrome 54, chrome 53


However, this other warning is still showing up:
> TypeScript compiler options "target" and "useDefineForClassFields" are set to "ES2022" and "false" respectively by the Angular CLI. To control ECMA version and features use the Browerslist configuration. For more information, see https://angular.io/guide/build#configuring-browser-compatibility

Even if we set the `.browserlistrc` as Angular does, or move that file inside webapp, the warning keeps coming up. The app still working fine so far.

webapp/src/ts/providers/parse.provider.ts
latin-panda: That scenario results in `NaN`, I've added unit tests to cover it.

latin-panda:
Thanks @dianabarsan and @jkuester! I've rebased with the main branch and pushed [the latest feedback](https://github.com/medic/cht-core/pull/7999/commits/157135f2107823b2e40bcf7a9b39fe7a07545717):
- I've updated the `.browserlistrc` and the` isUsingSupportedBrowser` function from the browser detection services to support chrome v80 which includes the features of es2020.
- and added unit tests to cover a negative literal in the parser service.

Once the CI is green, I will merge. If there is any other feedback, I can submit another PR right after this one. I would like the team to start using this version and report any problems before v4.2.0 release  :)

dianabarsan:
> Once the CI is green, I will merge. 

Does this not get AT-ed?

PR #8000, opened 2022-12-22
Update webapp logic to use client-side transitions

m5r:
Quick experiment to see if online client-side transitions could be a thing


jkuester:
@m5r I am going to YOLO this PR and merge it.  Based on @mrjones-plip's :+1: in https://github.com/medic/cht-core/issues/7753#issuecomment-1372919930 for this approach I think we move forward with it!  

I want to get it merged back into `7753-supervisor-chw-create` so that can remain the stable base branch for Tatiana and I start working on test changes.

Issue #7925, opened 2022-11-24
Issue in Bikram Sambat Date Picker


yrimal:
**Describe the bug**
There is an issue in Bikram Sambat date picker when I am put the valid value first and invalid value later on. 

**To Reproduce**
1. Create any form with date input and open it in CHT in Nepali locale. 
2. When you are in date field, enter valid Nepali Date. For e.g: (30, Kartik (or any month), 2079) and try validating form (going to next page or submitting).
4. If you've gone to next page, come back to date field. 
5. If you submitted the report, click on edit report and come back to date field. 
6. Try putting in any invalid dates (99 Kartik, 2079) 
7. Then submit form, it gets submitted without error. 

**Expected behavior**
Date should be validated if users are editing report and enter invalid date. 

**Screenshot**
Please watch this recording where I am editing a report with valid date and putting in invalid date later on. 

https://user-images.githubusercontent.com/6102813/203792958-56628198-d296-487f-bea8-f6866448b02a.mov

**Environment**
- Instance: Any
- Browser: Chrome
- App: webapp
- Version: 3.14.2

dianabarsan:
@yrimal 
To make sure I understand this, editing the date doesn't trigger validation? But adding the date initially does? 
Thanks

yrimal:
That's correct @dianabarsan. Additionally, If invalid date is submitted, the field keeps the original date which was in correct format.

dianabarsan:
Cool, thanks for clarifying @yrimal . 
I am trying this in 4.x and I'm seeing some inconsistent behavior where: 

- If I edit the year, then I get a constraint error
![image](https://user-images.githubusercontent.com/35681649/203813118-61cca695-2b2f-4708-a226-2dee8572d753.png)

- if I edit the day, I don't
![image](https://user-images.githubusercontent.com/35681649/203813254-af1267de-e53a-4aaf-9339-92900a312d61.png)

dianabarsan:
Just an update:
- I can add the "invalid" date even on add 
- I can't find any "validation" in the widget code.

@yrimal 
Can you please try to create a report with an invalid date?

yrimal:
Hi @dianabarsan , I've updated the form to replicate this issue. 

[deliver.xlsx](https://github.com/medic/cht-core/files/10089027/deliver.xlsx)

yrimal:
[...]

PR #8012, opened 2023-01-10
Fix for #7925 by using updated bs-bootstrap

yrimal:
# Description

Fixes issue related to bikram sambat where when invalid date was inputed, cht doesn't throw an error. 

medic/cht-core#7925

Fixes for updated package is on medic/bikram-sambat#18

# Code review checklist

# Compose URLs

* __CHT_CORE_COMPOSE_URL__
 
# License

### DIFF
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -45,7 +45,7 @@
     "@ngx-translate/core": "^13.0.0",
     "@webcomponents/webcomponentsjs": "^2.7.0",
     "bikram-sambat": "^1.7.0",
-    "bikram-sambat-bootstrap": "^1.5.0",
+    "bikram-sambat-bootstrap": "https://gitpkg.now.sh/medic/bikram-sambat/bootstrap",
     "bootstrap": "^3.4.1",
     "bootstrap-daterangepicker": "^2.1.30",
     "bowser": "^2.11.0",
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -45,7 +45,7 @@
     "@ngx-translate/core": "^13.0.0",
     "@webcomponents/webcomponentsjs": "^2.7.0",
     "bikram-sambat": "^1.7.0",
-    "bikram-sambat-bootstrap": "https://gitpkg.now.sh/medic/bikram-sambat/bootstrap",
+    "bikram-sambat-bootstrap": "^1.6.0",
     "bootstrap": "^3.4.1",
     "bootstrap-daterangepicker": "^2.1.30",
     "bowser": "^2.11.0",
--- a/.github/workflows/replace-staging-urls.yml
+++ b/.github/workflows/replace-staging-urls.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     types: [opened, reopened, synchronize]
 
+env:
+  INTERNAL_CONTRIBUTOR: ${{ secrets.AUTH_MARKET_URL && 'true' }}
+
 jobs:
 
   replace-pr-tokens-with-staging-URLs:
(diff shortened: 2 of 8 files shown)


jkuester (approved):
I can confirm that this change worked as expected for me!  When editing a doc using the bikram-sambat-datepicker widget and changing a date field to an invalid value, the date value recorded by the form will be blanked out (instead of silently remaining the previously entered value).  If the form is saved without entering a valid date value, then the value for the date field will be empty.  If the date field is `required`, then the form will not allow continuing until a valid date is entered.

webapp/package.json
```diff
@@ -45,7 +45,7 @@
     "@ngx-translate/core": "^13.0.0",
     "@webcomponents/webcomponentsjs": "^2.7.0",
     "bikram-sambat": "^1.7.0",
-    "bikram-sambat-bootstrap": "^1.5.0",
+    "bikram-sambat-bootstrap": "https://gitpkg.now.sh/medic/bikram-sambat/bootstrap",
```
craig-landry, line 48: I'm not familiar with `gitpkg.now.sh` can we use another option like a git url here instead to avoid relying on this service?

craig-landry (changes requested):

webapp/package.json
yrimal: I totally agree with this @craig-landry. I initially tried referencing github url, but since this needs to be referenced from subfolder of our repo, referencing that with github url didn't work properly. So, I found Gitpkg, the [open source](https://github.com/EqualMa/gitpkg) tool to reference any github urls, branch, subfolder as a reference. 

I haven't referenced github subfolders previously in npm, if you could provide some working reference or previous examples, then I will update this.

dianabarsan: This change is temporary, just while this issue is AT-ed. 
After AT, we will publish a new version of `bikram-sambat-bootstrap` on npm, that includes the fix, and PR will be updated to just bump the version to `1.6.0` (instead of directly installing from a branch).

craig-landry: > This change is temporary, just while this issue is AT-ed.

Ah, cool. Thanks 👍

dianabarsan (commented):
Thanks for adding this! 

Just to note, this PR is to be updated once the code is AT-ed, as it installs a recently updated dependency (the code where the issue was actually fixed). 
Once the code is AT-ed, we will publish a new version of `bikram-sambat-bootstrap` and then update this PR to install that version instead. 
At that point, we can merge this and mark the issue as resolve.

dianabarsan:
For the failing tests, this is because publishing docker images for external contributors is broken. I've logged this here: https://github.com/medic/cht-core/issues/8014 and I'm working on fixing this very soon.

dianabarsan:
Hi @yrimal 

I've merged a fix for failing builds into `master`. Would you mind rebasing your branch to get the tests to run? Thanks!

yrimal:
Hi @dianabarsan , I rebased it and this [integration test](https://github.com/medic/cht-core/actions/runs/3918717612/jobs/6700063890) is failing. Can you help with this ?

dianabarsan:
Really glad to see the other tests be green, it means my fix worked. 
I've restarted your integration test, it seems like it failed to install dependencies due to a timeout.

craig-landry:
I've removed myself from the reviewers of this PR. I'd commented previously, but don't need to review further as the others on this list will have much better feedback than myself :slightly_smiling_face:

craig-landry (approved):
I had removed myself, but as @yrimal reminded me, I had requested changes so am still required. My only note was that dep line and I've since been educated on that. From my perspective this stuff is good and I'll leave the rest of the review to the qualified people :slightly_smiling_face:

yrimal:
I've nothing to commit, this is ready to go @dianabarsan .

Issue #8009, opened 2023-01-10
Display version and build number separately in the admin upgrade page


dianabarsan:
**What feature do you want to improve?**
Right now the version that is displayed in the admin upgrade page is the full version, which includes the build number. 
This is useful for branches, where there might be multiple builds for the same branch. It's not useful for releases, where there should only be one build for each release. 

![image](https://user-images.githubusercontent.com/35681649/211512168-feb007b6-3530-46ed-8e94-3f2b6d4dd75f.png)

**Describe the improvement you'd like**
Display full build version separately, while also displaying clean semver only version. 

**Describe alternatives you've considered**
- Don't display build at all (then there is no way of knowing if you're on the latest version of a branch)
- No changes

**Additional context**
This came up on the forum: https://forum.communityhealthtoolkit.org/t/docker-helper-cht-version/2356

dianabarsan:
This is ready for AT on https://github.com/medic/cht-core/pull/8013 . 

I've added an extra field for `build`, and left the version be just the base_version. 
![image](https://user-images.githubusercontent.com/35681649/212609947-9b3c69ec-8c28-4471-afdc-f44072766c6c.png)

This change affects both admin and api, so after you upgrade to the branch, please refresh the admin upgrade page to see the changes.

tatilepizs:
### Test details

**Config:** Default
**Environment:** Local with docker helper script
**Platform:** WebApp
**Browser:** Chrome

---

### Reproducible on `master`
Both, version and build number are shown in the same field.

<details>
<summary>Image attached</summary>

![image](https://user-images.githubusercontent.com/94494491/212773620-ea606bea-ee2c-4d57-84b5-8775461618ea.png)

</details>

### Fixed on `8009-separate-admin-build-and-version`
Version and build details are shown in different fields.

<details>
<summary>Image attached</summary>

![image](https://user-images.githubusercontent.com/94494491/212773514-1a4c92c4-f11f-45f2-ae2c-815180d2d2ef.png)

</details>

---

Test passed successfully. :tada:

dianabarsan:
Merged to `master`.

PR #8013, opened 2023-01-11
Separate base version from build in admin app upgrade page

dianabarsan:
# Description

Changes `version` field to display `base_version` (like `4.2.0`) and adds a separate field that displays the full build version (which helps differentiate builds for the same branch from one another). 

image

medic/cht-core#8009

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8009-separate-admin-build-and-version/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8009-separate-admin-build-and-version/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8009-separate-admin-build-and-version/docker-compose/cht-couchdb-clustered.yml)
 
# License

### DIFF
--- a/api/resources/translations/messages-en.properties
+++ b/api/resources/translations/messages-en.properties
@@ -753,6 +753,7 @@ instance.upgrade = Upgrades
 instance.upgrade.at = On
 instance.upgrade.betas = Betas
 instance.upgrade.branches = Branches
+instance.upgrade.build.version = Build
 instance.upgrade.cancel = Cancel upgrade
 instance.upgrade.cancelling = Cancelling upgrade
 instance.upgrade.cancel.note = Cancelling will stop the current upgrade and reset any staging progress made so far.
--- a/tests/e2e/upgrade/upgrade.wdio-spec.js
+++ b/tests/e2e/upgrade/upgrade.wdio-spec.js
@@ -43,11 +43,15 @@ describe('Performing an upgrade', () => {
     const currentVersion = await upgradePage.getCurrentVersion();
     expect(currentVersion).to.include(version.getVersion());
 
+    await browser.refresh(); // load updated code of admin app
+    await upgradePage.goToUpgradePage();
+    const currentBuild = await upgradePage.getBuild();
+
     // there should be no staged ddocs
     const ddocs = await getDdocs();
     const staged = ddocs.filter(ddoc => ddoc._id.includes('staged'));
     expect(staged.length).to.equal(0);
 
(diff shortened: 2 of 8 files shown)


dianabarsan:
Hi @jkuester . Would you mind taking a look? Thanks!

api/resources/translations/messages-en.properties
```diff
@@ -753,6 +753,7 @@ instance.upgrade = Upgrades
 instance.upgrade.at = On
 instance.upgrade.betas = Betas
 instance.upgrade.branches = Branches
+instance.upgrade.build.version = Build
```
dianabarsan, line 756: Maybe `Build number` would be more obvious?

jkuester: "Build" seems fine to me (especially since we say "Version" and not "Version number")

tests/e2e/upgrade/upgrade.wdio-spec.js
```diff
@@ -43,11 +43,15 @@ describe('Performing an upgrade', () => {
     const currentVersion = await upgradePage.getCurrentVersion();
     expect(currentVersion).to.include(version.getVersion());
 
+    await browser.refresh(); // load updated code of admin app
```
jkuester, line 46: It is interesting to me that we need to refresh the  page here (I assume to get the proper build number) when previously the same value was coming back on line 44 as the old version number without having to refresh the page... :thinking:

Am I missing something here?

jkuester (approved):
:rocket:

tests/e2e/upgrade/upgrade.wdio-spec.js
dianabarsan: Yea :) 

When you upgrade, the admin page is still rendered by the old version if we don't refresh (it's an angular app). And the "old version" can't display the new build field that I've just added in this PR (I tried lots of options, but this just seems the easiest one to just remove after this code is merged).

dianabarsan: This is safe to remove after this code is merged (in a new PR). 
But maybe it's ok if it stays?

jkuester: Ahhh, hah!  I was stuck thinking it was the _value_ of the field that was not getting updated. Did not consider that the field, itself, does not exist until the page is reloaded...

:+1:

Issue #8103, opened 2023-02-25
Protractor test failures are not retried and failed retries clear results from previous failure


dianabarsan:
**Describe the issue**
We added retries to protractor e2e test failures. 
However, they do not seem to work: the retry will fail immediately with an error:
```
 Re-running tests , attempt :  1
 Re-running the following test files : /home/runner/work/cht-core/cht-core/tests/e2e/protractor/client-side-transitions/forms/index.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/client-side-transitions/muting.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/contacts/contact-summary.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/contacts/reports-subject.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/sms/africas-talking.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/sms/rapidpro.spec.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/sms/registration-by-sms.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/sms/send-message.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/targets/target-aggregates.spec.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/translations/incorrect-locale.spec.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/client-side-transitions/forms/index.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/client-side-transitions/muting.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/contacts/contact-summary.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/contacts/reports-subject.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/sms/africas-talking.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/sms/rapidpro.spec.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/sms/registration-by-sms.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/sms/send-message.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/targets/target-aggregates.spec.js,/home/runner/work/cht-core/cht-core/tests/e2e/protractor/translations/incorrect-locale.spec.js


Closing report
[...]

PR #8105, opened 2023-02-27
Fix protractor retries and screenshots

garethbowen:
# Description

medic/cht-core#8103

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8103-protractor-retries/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8103-protractor-retries/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8103-protractor-retries/docker-compose/cht-couchdb-clustered.yml)
 
# License


tests/utils.js
```diff
@@ -1338,7 +1340,7 @@ module.exports = {
   tearDownServices: stopServices,
   endSession: async (exitCode) => {
     await module.exports.tearDownServices();
-    return module.exports.reporter.afterLaunch(exitCode);
+    await new Promise((resolve) => module.exports.reporter.afterLaunch(resolve.bind(this, exitCode)));
```
garethbowen, line 1343: This was the root issue, predating but perhaps exacerbated by, the retries change - the screenshot reporter afterLaunch function expects a callback to be passed.

dianabarsan (commented):
Looking at the failure captured here: https://github.com/medic/cht-core/issues/8103#issuecomment-1445553380, it seems like only the _last_ retry snapshot is kept, which makes me think that all the previous snapshots were still deleted. 
I also gave it a try, intentionally failing one e2e test, and got the same. 

I know we don't intend to keep protractor around for long, so I don't think this is worth a lot of effort. 
I think a lot of test retries will fail because no cleanup is made after previous test (like in the case of my test: https://github.com/medic/cht-core/actions/runs/4282260079), so the last screenshot that is saved might have little value compared to the first. 
Is it possible to make sure we don't clean up the results folder at all?

garethbowen:
Ooof. I've spent some time looking into this.

So... retries (especially in protractor) didn't seem to solve any real flaky tests because the setup and teardown didn't happen between retries in a way that would allow the test to re-run. That, combined with the fact we're dropping protractor anyway, makes me reluctant to spend too much time on it - better to keep improving WDIO instead.

If you agree, I'll revert all the retry code in Protractor to get us back to where we were before, with the same level of flake but screenshots so we can debug it.

@dianabarsan What do you think?

dianabarsan:
> If you agree, I'll revert all the retry code in Protractor to get us back to where we were before, with the same level of flake but screenshots so we can debug it.

Yes, please!

garethbowen:
@dianabarsan Retries removal review please!

dianabarsan (approved):
Thanks for looking into this! I'm sorry we only ended up back where it started.

Issue #6289, opened 2020-03-12
Upgrade to CouchDB v3


garethbowen:
CouchDB 3.0 has been released and eventually we'll want to migrate to it.

- Investigate technical compatibility with the CHT and raise issues to fix the bugs, tests, documentation, etc in a backwards compatible way.
- Run performance and scalability tests to determine the extent of any improvements or regressions. Wherever possible raise issues against CouchDB where performance has regressed to give their team plenty of time to investigate before we're ready to upgrade.

This issue can be closed once we have confidence that we have identified and documented all issues that are blocking us from upgrading today.

garethbowen:
I upgraded to Couch 3 locally and the `medic` db became available to all. Fortunately our server checks found it and refused to start API. The problem was that I had no roles configured for the medic database so it was considered public. The workaround was to add all my roles to the `medic` db.

dianabarsan:
I can confirm. 
It seems like the `require_valid_user` doesn't do anything anymore and all database security comes from it's own `_security`. Additionally, when creating new databases, I've noticed that they have a default `_security` that only includes the `_admin` role. 

Additionally, the `_security` object of the `_users` database cannot be changed and only admins can access it. 
Non-admin users can't create new databases.
When an admin creates a new database, it has a default `_security` object that only allows `_admin` role to access the DB (`_admin` is a member and an admin).

Ths2-9Y-LqJt6:
Hey all! I'm a developer looking to possibly help with CHT.  My first step was to see how good your [developer set up docs](https://github.com/medic/cht-core/blob/master/DEVELOPMENT.md) were. In the happy path, they're good! But, being on Ubuntu 18.04, the docs are now a sad path because of CouchDB 3.0 being the default.  Since this ticket is about CouchDB 3.0, commenting here, but happy to open a "improve the dev docs" ticket/PR if that's more helpful.

**tl;dr** - While it looks like this is a CouchDB docs/install issue, maybe we should call out that for Ubuntu >18 you need to be extra careful to install CouchDB 2 like we do for MacOS and homebrew?  I know that we very clearly say `CouchDB 2.x` in our [dependencies](https://github.com/medic/cht-core/blob/master/DEVELOPMENT.md#dependencies), but as non-CouchDB dev who needed to install everything from scratch, this bit me.

[...]

PR #8107, opened 2023-02-27
feat(#6289): Bumps CouchDb version to 3.3.2

dianabarsan:
# Description

User roles are added to db security on settings change. 
Config remains unchanged with a couple of exceptions: 
- `request_timeout` no longer `infinity` because of [this error](https://github.com/apache/couchdb/issues/3789)
- added `[attachments]` section back, as cht-conf depends on those settings to exist and it isn't part of default config anymore

medic/cht-core#6289

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7964-couch3/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7964-couch3/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7964-couch3/docker-compose/cht-couchdb-clustered.yml)
 
# License


couchdb/10-docker-default.ini
```diff
@@ -1,7 +1,7 @@
 ; couchdb/local.d/package.ini
 
 [fabric]
-request_timeout = infinity
+request_timeout = 31536000 ; 1 year
```
dianabarsan, line 4: When using `infinity`, I got an error similar to this one: https://github.com/apache/couchdb/issues/3789
```
[error] 2023-02-27T12:10:48.341702Z couchdb@couchdb-1.local <0.754.0> 08bd1a6cec req_err(2329098955) unknown_error : timeout_value
    [<<"fabric_util:get_shard/4 L149">>,<<"fabric:get_security/2 L220">>,<<"chttpd_auth_request:db_authorization_check/1 L109">>,<<"chttpd_auth_request:authorize_request/1 L19">>,<<"chttpd:handle_req_after_auth/2 L406">>,<<"chttpd:process_request/1 L388">>,<<"chttpd:handle_request_int/1 L325">>,<<"mochiweb_http:headers/6 L140">>]
[notice] 2023-02-27T12:10:48.341813Z couchdb@couchdb-1.local <0.754.0> 08bd1a6cec haproxy:5984 192.168.240.8 admin GET /_users 500 ok 1
```

api/src/db.js
```diff
@@ -163,4 +164,37 @@ if (UNIT_TEST_ENV) {
 
     throw new Error(`Error while saving docs: ${errors.join(', ')}`);
   };
+
+  const DEFAULT_SECURITY_STRUCTURE = {
+    names: [],
+    roles: [],
+  };
```
garethbowen, line 171: I think if you call `addRoleToSecurity` multiple times this default object will mutate because it's never cloned. You could either clone this on access on line 184, or I think a better solution would be to change this constant a getter...

```suggestion
  const getDefaultSecurityStructure = () => {
    return { names: [], roles: [] };
  };
```

api/src/db.js
```diff
@@ -163,4 +164,37 @@ if (UNIT_TEST_ENV) {
...
+
+  module.exports.addRoleToSecurity = async (dbname, role, addAsAdmin) => {
+    if (!dbname || !role) {
+      return;
```
garethbowen, line 175: I think it'd be better to throw here to make it clear that it didn't succeed.

api/src/db.js
```diff
@@ -163,4 +164,37 @@ if (UNIT_TEST_ENV) {
...
+    securityUrl.pathname = `${dbname}/_security`;
+
+    const securityObject = await rpn.get({ url: securityUrl.toString(), json: true });
+    const property = addAsAdmin ? 'admins' : 'members';
+    if (!securityObject[property]) {
+      securityObject[property] = DEFAULT_SECURITY_STRUCTURE;
+    }
+
+    if (!securityObject[property].roles) {
+      throw new Error('Invalid database security %o', securityObject);
```
garethbowen, line 188: The second parameter of the Error constructor shouldn't be a random object. I suspect you switched from `console.error` to `throw new Error` at some point?

api/src/db.js
```diff
@@ -163,4 +164,37 @@ if (UNIT_TEST_ENV) {
...
+    const securityUrl = new URL(environment.serverUrl);
+    securityUrl.pathname = `${dbname}/_security`;
+
+    const securityObject = await rpn.get({ url: securityUrl.toString(), json: true });
+    const property = addAsAdmin ? 'admins' : 'members';
+    if (!securityObject[property]) {
+      securityObject[property] = DEFAULT_SECURITY_STRUCTURE;
+    }
+
+    if (!securityObject[property].roles) {
```
garethbowen, line 187: This is weird right, because we've found the security object in the db and now we're throwing, which will be a bad time if this is run during migration. Can't we just add the `roles` array to proactively "fix" the security object?

api/src/db.js
```diff
@@ -163,4 +164,37 @@ if (UNIT_TEST_ENV) {
 
     throw new Error(`Error while saving docs: ${errors.join(', ')}`);
   };
+
+  const DEFAULT_SECURITY_STRUCTURE = {
+    names: [],
+    roles: [],
+  };
+
+  module.exports.addRoleToSecurity = async (dbname, role, addAsAdmin) => {
```
garethbowen, line 173: Taking `addAsAdmin` as a boolean makes it hard to read when invoking it... Consider adding two functions: `addRoleToAdmins` and `addRoleToMembers` to make it clear.

couchdb/10-docker-default.ini
garethbowen: I wonder if it should be 1 hour to match our load balancer timeouts?

couchdb/10-docker-default.ini
```diff
@@ -10,6 +10,8 @@ os_process_limit = 1000
 os_process_timeout = 60000
 max_dbs_open = 5000
 attachment_stream_buffer_size = 16384
+users_db_security_editable = true
```
garethbowen, line 13: Why do we need this?

couchdb/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM couchdb:2.3.1 as base_couchdb_build
+FROM couchdb:3.3.1 as base_couchdb_build
```
garethbowen, line 1:
```suggestion
FROM couchdb:3.3.2 as base_couchdb_build
```

garethbowen (changes requested):
Minor suggestions inline.

Exciting stuff!

couchdb/10-docker-default.ini
dianabarsan: We have this ancient migration that adds `national_admin` to `_users`: https://github.com/medic/cht-core/blob/master/api/src/migrations/add-national_admin-role.js#L58
We have historically allowed `national_admins` to make changes to users, we would be breaking backwards compatibility.

couchdb/10-docker-default.ini
dianabarsan: I think that not all requests get to CouchDb through load balancer. On the other hand, I have no strong feelings about this timeout.

couchdb/10-docker-default.ini
garethbowen: I think it's probably time to break that backwards compatibility. I'm worried that changing this setting puts us at a security risk, and the national_admin role has been deprecated for years.

What do you think about a migration to add the `_admin` role to all `national_admin`s?

This could be done as a separate PR.

couchdb/10-docker-default.ini
garethbowen: I agree... I suspect this was set for things like the sentinel db watcher which could theoretically run for a year. But it's probably a bad idea for it to run for a year, and having a timeout is probably a good idea. Happy to leave this as is and discuss separately.

couchdb/10-docker-default.ini
dianabarsan: Hm, ok. I'm thinking about other ways we would break functionality by making all `national_admin`s actual admins. 
This issue comes to mind: https://github.com/medic/cht-core/issues/8243 where we would have previously allowed to turn off permissions for `national_admins`.

garethbowen (approved):
:tada:

couchdb/10-docker-default.ini
```diff
@@ -19,14 +21,12 @@ socket_options = [{sndbuf, 262144}, {nodelay, true}]
...
-port = 5986
-bind_address = 0.0.0.0
 secure_rewrites = false
 max_http_request_size = 134217728 ; 128 MiB
 WWW-Authenticate = Basic realm="Medic Mobile Web Services"
 
 [couch_httpd_auth]
-# timeout is set to 1 year in seconds, 31536000
+; timeout is set to 1 year in seconds, 31536000
 timeout = 31536000
```
kennsippell, line 30: would be great to expose prometheus monitoring via an unauthenticated endpoint https://github.com/medic/cht-watchdog/issues/76.  happy to do this in a second PR after this merges

fardarter: Can I suggest you develop a user configurable config layer? We're developing against this and having to modify directly means it's difficult to stay conformant with upstream. 

See for example:

```suggestion
echo "
[daemons]
httpsd = {couch_httpd, start_link, [https]}
[admins]
admin = ${COUCHDB_PASSWORD}
[ssl]
port = 6984
enable = true
cert_file = /opt/couchdb/etc/certs/cert.pem
key_file = /opt/couchdb/etc/certs/privkey.pem
cacert_file = /opt/couchdb/etc/certs/chain.pem" |

sudo crudini --merge /opt/couchdb/etc/local.ini
```

fardarter: Can I suggest you develop a user configurable config layer? We're developing against this and having to modify directly means it's difficult to stay conformant with upstream. 

See for example:

```suggestion
echo "
[daemons]
httpsd = {couch_httpd, start_link, [https]}
[admins]
admin = ${COUCHDB_PASSWORD}
[ssl]
port = 6984
enable = true
cert_file = /opt/couchdb/etc/certs/cert.pem
key_file = /opt/couchdb/etc/certs/privkey.pem
cacert_file = /opt/couchdb/etc/certs/chain.pem" |

sudo crudini --merge /opt/couchdb/etc/local.ini
```

dianabarsan: Hi @fardarter 

Thanks a lot for this great suggestion! 
I think this is out of scope of this PR, and this has been discussed while we originally developed 4.x. We have not reached a conclusion, and we decided this would be a goal for later. 

Would you mind creating an issue that details exactly which configs you need and how you're thinking about persisting them in the container? 
Thanks a lot!

fardarter: Our current plan is to build our own version of the images with patches as needed, so this would be a pre-image build step. Would argue that supporting user patching behaviour would help with keeping people contributing to upstream.

I've set up a system where we're doing our best to only patch what we need. Could be improved; happy to demo it.

I'm actually trying to find your contact details at the moment (which is why I'm on this PR). Don't suppose I could give you mine (and then we could delete the comment)?

My senior dev on this project has fallen ill and he was using this PR as a way to let us use 3.3.2 (we'd like to use the bitnami image on azure, but I can script those differences).

I'm supposed to be doing ops so I'm a little lost as to whether we can even use 3.3.2 at this juncture, or whether it would be foolish. I'd prefer not having to build out a 2.2.1 on a VM where there's a preconfigured, hardened bitnami image. 

I'm also a little lost as to what the right command is to test this locally, but my tests seem to be breaking.

dianabarsan: Open discussion is welcome and desirable on this PR or on the [CHT Forum](https://forum.communityhealthtoolkit.org/). 
If you have specific questions, or have a technical challenge, the forum would be the best place to express these, so that other community members would benefit. 
I believe there already is a thread about this PR: https://forum.communityhealthtoolkit.org/t/upgrade-of-couchdb-from-version-2-3-1-to-3-3-1/2928 , I'm not sure if this is at all related to your issue. 

Please feel free to reach out on either platform! 
I'm really keen to hear more about your desired outcome. 

Thanks!

dianabarsan:
@garethbowen are we ready for this? 
I had to include a commit that stabilizes an e2e test (same code in this PR: https://github.com/medic/cht-core/pull/8536) which was failing consistently on this branch.

api/tests/mocha/db.spec.js
```diff
@@ -3,6 +3,7 @@ require('chai').use(require('chai-as-promised'));
 const { expect } = require('chai');
 const rewire = require('rewire');
 const rpn = require('request-promise-native');
+process.env.UNIT_TEST_ENV = 1;
```
garethbowen, line 6: Isn't this already set in the env?

dianabarsan: I think it should be. Removed.

tests/integration/cht-conf/cht-conf-actions.spec.js
```diff
@@ -51,7 +51,7 @@ describe('cht-conf actions tests', () => {
   it('should upload branding', async () => {
     const branding = await utils.getDoc('branding').catch(error => {
       if (error){
-        console.log(error);
+        console.err(error);
```
garethbowen, line 54: Is this a thing??

```suggestion
        console.error(error);
```

tests/integration/api/server.spec.js
```diff
@@ -211,12 +211,15 @@ describe('server', () => {
       await utils.startApi();
       await utils.delayPromise(1000);
 
+      await utils.delayPromise(1000);
```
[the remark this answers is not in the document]
dianabarsan: I think this is a bad merge.

tests/integration/cht-conf/cht-conf-actions.spec.js
dianabarsan: wtb compiler. :(

garethbowen:
@dianabarsan I can't think of any reason to hold off and nothing else to test. Let's push the button and get 4.4 out.

The main concern is going to be upgrading multi-node instances, right? We don't have a good solution for that right now, but we're not likely to have one next week either so I feel like it's better to get this done now before more installations are running multi-node solutions.

dianabarsan:
I think my main concern is not hitting the right config for us - which is the biggest unknown here. I'm merging.

Issue #8093, opened 2023-02-21
Prevent data loss through re-sync when restoring data from backup


dianabarsan:
**What feature do you want to improve?**
Data loss when restoring from backups is most likely unavoidable. 
However there is one mitigation we can deploy, with relatively little cost. 

First, follow these steps to reproduce how data loss happens now: 
1. Launch CHT (4.x is easiest, but this applies to every release since 2017) and create an offline user
2. Login as the offline user. 
3. Create a data backup. 
4. As the offline user, create some documents and sync. 
5. As the admin, check that the documents created on step 4 exist on the server. 
6. Overwrite live data with the backup (simulate a restoration from backup). 
7. Check that the documents created on step 4 don't exist on the server (this is expected, those docs were created AFTER the backup). 
8. Sync as the offline user. 
9. Check that the documents created on step 4 don't exist on the server. 

What happens on step 9 is caused by the way we save the replication checkpointer in Pouchdb: [at replication source](https://github.com/medic/cht-core/blob/master/webapp/src/ts/services/db-sync.service.ts#L88). This means that the document that holds the replication sequence for client uploads exists only on the client. The client doesn't know the server was restored from a backup, so it will keep uploading docs based on the locally stored sequence. This sequence is _after_ the docs created on step 4. 

**Describe the improvement you'd like**
PouchDb supports multiple settings for the `checkpoint`, and by default will save the checkpointer on both `source` and `target`. 
If the checkpointer is also saved on the target, when reaching step 9, PouchDb will check replication history on both checkpointers and choose the latest common seq. Because the server will have an older seq, the documents created on 4 will be re-uploaded. No data loss. 

With this approach, in case the server checkpointer becomes corrupted, is deleted, etc, the client will start replication from 0. This means that the local database is iterated over in batches and, for each batch, it will query a server endpoint (_revs_diff) to check whether the docs already exist. This endpoint is unfiltered, and requests are forwarded directly to CouchDb.  Only missing docs will be uploaded.
We should assess the performance impact of a client resetting replication, both for the client and for the server. 

When implementing this, we should also make sure that we somehow "start" by writing the server-side checkpointer, if possible, so we avoid all clients going through the process described above. 

**Describe alternatives you've considered**
Don't change anything. 

**Additional context**
We recently restored a production  deployment from a backup due to server-side data corruption.

dianabarsan:
This is ready for AT on https://github.com/medic/cht-core/pull/8108

lorerod:
[...]

Issue #8097, opened 2023-02-23
Flaky test:  Edit contacts with the default config should remove the primary contact from the clinic when the contact is deleted


dianabarsan:
**Describe the issue**
Test is failing frequently and repeatedly on retries: 
```
------------------------------------------------------------------
[chrome 110.0.5481.100 linux #0-7] Running: chrome (v110.0.5481.100) on linux
[chrome 110.0.5481.100 linux #0-7] Session ID: 339f9ca4-2d38-4792-8357-0dfc4b70efe3
[chrome 110.0.5481.100 linux #0-7]
[chrome 110.0.5481.100 linux #0-7] » /tests/e2e/default/contacts/edit.wdio-spec.js
[chrome 110.0.5481.100 linux #0-7] Edit contacts with the default config. 
[chrome 110.0.5481.100 linux #0-7]    ✓ should edit contact name
[chrome 110.0.5481.100 linux #0-7]    ✖ should remove the primary contact from the clinic when the contact is deleted
[chrome 110.0.5481.100 linux #0-7]    ✖ should remove the primary contact from the clinic when the contact is deleted
[chrome 110.0.5481.100 linux #0-7]    ✖ should remove the primary contact from the clinic when the contact is deleted
[chrome 110.0.5481.100 linux #0-7]    ✖ should remove the primary contact from the clinic when the contact is deleted
[chrome 110.0.5481.100 linux #0-7]    ✖ should remove the primary contact from the clinic when the contact is deleted
[chrome 110.0.5481.100 linux #0-7]    ✖ should remove the primary contact from the clinic when the contact is deleted
[chrome 110.0.5481.100 linux #0-7]
[chrome 110.0.5481.100 linux #0-7] 1 passing (58.1s)
[chrome 110.0.5481.100 linux #0-7] 1 failing
[chrome 110.0.5481.100 linux #0-7]
[chrome 110.0.5481.100 linux #0-7] 1) Edit contacts with the default config.  should remove the primary contact from the clinic when the contact is deleted
[chrome 110.0.5481.100 linux #0-7] element ("#contact_summary") still not displayed after 15000ms
[chrome 110.0.5481.100 linux #0-7] Error: element ("#contact_summary") still not displayed after 15000ms
[chrome 110.0.5481.100 linux #0-7]     at async Object.waitForContactLoaded (/home/runner/work/cht-core/cht-core/tests/page-objects/default/contacts/contacts.wdio.page.js:126:3)
[...]

PR #8108, opened 2023-02-27
Store server-side checkpointer

dianabarsan:
# Description

medic/cht-core#8093

# Code review checklist

# Compose URLs

* __CHT_CORE_COMPOSE_URL__
 
# License

### DIFF
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -71,6 +71,7 @@
     "pojo2xml": "^1.7.2",
     "pouchdb-browser": "^7.3.1",
     "pouchdb-debug": "^7.2.1",
+    "pouchdb-generate-replication-id": "^8.0.1",
     "rxjs": "^7.8.0",
     "select2": "4.0.3",
     "simple-password-tester": "^1.0.0",
--- a/webapp/src/ts/services/db-sync.service.ts
+++ b/webapp/src/ts/services/db-sync.service.ts
@@ -13,6 +13,7 @@ import { CheckDateService } from '@mm-services/check-date.service';
 import { TelemetryService } from '@mm-services/telemetry.service';
 import { GlobalActions } from '@mm-actions/global';
 import { TranslateService } from '@mm-services/translate.service';
+import { MigrationsService } from '@mm-services/migrations.service';
 
 const READ_ONLY_TYPES = ['form', 'translations'];
 const READ_ONLY_IDS = ['resources', 'branding', 'service-worker-meta', 'zscore-charts', 'settings', 'partners'];
@@ -76,6 +77,7 @@ export class DBSyncService {
     private store:Store,
     private translateService:TranslateService,
     private purgeService:PurgeService,
+    private migrationsService:MigrationsService,
   ) {
     this.globalActions = new GlobalActions(store);
   }
@@ -85,7 +87,6 @@ export class DBSyncService {
       name: 'to',
       options: {
         filter: readOnlyFilter,
-        checkpoint: 'source',
       },
       allowed: () => this.authService.has('can_edit'),
       onDenied: (err?) => this.dbSyncRetryService.retryForbiddenFailure(err),
@@ -105,6 +106,7 @@ export class DBSyncService {
   private inProgressSync;
   private knownOnlineState = window.navigator.onLine;
   private syncIsRecent = false; // true when a replication has succeeded within one interval
+  private migrationsExecuted = false;
   private readonly intervalPromises = {
     sync: undefined,
     meta: undefined,
@@ -317,16 +319,30 @@ export class DBSyncService {
     }
   }
 
+  private async migrateDb() {
+    if (!this.migrationsExecuted) {
+      try {
+        await this.migrationsService.runMigrations();
+        this.migrationsExecuted = true;
+      } catch (err) {
+        console.error('Error while running DB migrations', err);
+        throw err;
+      }
+    }
+  }
+
   /**
   * Synchronize the local database with the remote database.
   *
   * @returns Promise which resolves when both directions of the replication complete.
   */
-  sync(force?) {
+  async sync(force?) {
     if (!this.isEnabled()) {
       this.sendUpdate({ state: SyncStatus.Disabled });
       return Promise.resolve();
     }
+
+    await this.migrateDb();
 
     if (force) {
       this.globalActions.setSnackbarContent(this.translateService.instant('sync.status.in_progress'));
--- a/webapp/src/ts/services/migrations.service.ts
+++ b/webapp/src/ts/services/migrations.service.ts
@@ -21,11 +21,11 @@ export class MigrationsService {
 
   async runMigrations () {
     for (const migration of this.getMigrations()) {
-      if (await migration.hasRun()) {
+      if (await migration.hasRun(this.dbService)) {
         continue;
       }
-
       await migration.run();
+      await migration.setHasRun(this.dbService);
     }
   }
 }
--- a/webapp/src/ts/services/migrations.service.ts
+++ b/webapp/src/ts/services/migrations.service.ts
@@ -1,0 +1,31 @@
+import { Injectable } from '@angular/core';
+
+import { DbService } from '@mm-services/db.service';
+import { TargetCheckpointerMigration } from '@mm-services/migrations/target-checkpointer.migration';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class MigrationsService {
+  constructor(
+    private dbService:DbService,
+    private targetCheckpointerMigration:TargetCheckpointerMigration,
+  ) {
+  }
+
+  private getMigrations() {
+    return [
+      this.targetCheckpointerMigration,
+    ];
+  }
+
+  async runMigrations () {
+    for (const migration of this.getMigrations()) {
+      if (await migration.hasRun()) {
+        continue;
+      }
+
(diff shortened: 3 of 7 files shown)


dianabarsan:
hey @garethbowen 
Would you mind a review? I feel like I've ruminated on this enough, I'd love to hear your thoughts.

tests/e2e/default/contacts/edit.wdio-spec.js
```diff
@@ -40,6 +40,7 @@ describe('Edit contacts with the default config. ', () => {
   });
 
   it('should remove the primary contact from the clinic when the contact is deleted', async () => {
+    await commonPage.refresh();
```
garethbowen, line 43: Is this to fix this issue? https://github.com/medic/cht-core/issues/8097

Consider a separate PR?

tests/integration/cht-conf/cht-conf-actions.spec.js
```diff
@@ -57,7 +57,7 @@ describe('cht-conf actions tests', () => {
     expect(branding.title).to.equal('Medic');
   });
 
-  it('should uploadload  forms', async () => {
+  it('should upload  forms', async () => {
```
garethbowen, line 60:
```suggestion
  it('should upload forms', async () => {
```

:D

tests/page-objects/default/common/common.wdio.page.js
```diff
@@ -214,6 +220,13 @@ const syncWithoutWaitForSuccess = async () => {
   await (await syncButton()).click();
 };
 
+const syncAndWaitForFailure = async () => {
+  await openHamburgerMenu();
+  await (await syncButton()).click();
```
garethbowen, line 225:
```suggestion
  await syncWithoutWaitForSuccess();
```

Also could update line 197 in the same way.

tests/utils.js
```diff
@@ -197,7 +197,7 @@ const deleteAll = (except) => {
     'branding',
     'partners',
     'settings',
-    /^form:contact:/,
+    /^form:/,
```
garethbowen, line 200: This feels like a big change. Did all the tests just work?

tests/utils/browser.js
```diff
@@ -32,7 +35,121 @@ const getCookies = (...cookieNameList) => {
...
+      .then(result => callback({ result }))
+      .catch(err => callback({ err }));
+  }, doc);
+
+  if (err) {
+    throw err;
+  }
+
+  return result;
+};
```
garethbowen, line 52: Can we come up with a way to do this without the boilerplate? Ideally we'd have a private utility function which means this function can be simplified down to...

```js
const createDoc = async (doc) => {
  return await fancyWrapperWithGoodName((doc) => {
    const db = window.CHTCore.DB.get();
    return db.put(doc);
  }, doc);
});
```

Needs a little varargs work, but would simplify each of these functions so you can actually see what they're doing.

tests/utils/browser.js
```diff
@@ -32,7 +35,121 @@ const getCookies = (...cookieNameList) => {
...
+
+const info = async () => {
+  const { err, result } = await browser.executeAsync((callback) => {
+    const db = window.CHTCore.DB.get();
+    return db
+      .info()
+      .then(response => callback({ result: response }))
```
garethbowen, line 135:
```suggestion
      .then(result => callback({ result }))
```

webapp/package.json
```diff
@@ -71,6 +71,7 @@
     "pojo2xml": "^1.7.2",
     "pouchdb-browser": "^7.3.1",
     "pouchdb-debug": "^7.2.1",
+    "pouchdb-generate-replication-id": "^8.0.1",
```
garethbowen, line 74: Should be the same version as pouchdb-browser, right?

webapp/src/ts/services/db-sync.service.ts
```diff
@@ -317,17 +318,29 @@ export class DBSyncService {
...
+    }
+  }
+
   /**
   * Synchronize the local database with the remote database.
   *
   * @returns Promise which resolves when both directions of the replication complete.
   */
-  sync(force?) {
+  async sync(force?) {
```
garethbowen, line 336: "async sync"... LOL

webapp/src/ts/services/migrations.service.ts
```diff
@@ -0,0 +1,41 @@
...
+    this.runningMigrations = false;
+  }
+
+  async runMigrations () {
+    if (this.runningMigrations) {
+      return await this.runningMigrationsPromise;
+    }
+
+    this.runningMigrations = true;
+    this.runningMigrationsPromise = this.run();
```
garethbowen, line 37: Let's combine these two... if `runningMigrationsPromise` is set, then `runningMigrations` is true, right?

webapp/src/ts/services/migrations/target-checkpointer.migration.ts
```diff
@@ -0,0 +1,58 @@
...
+})
+export class TargetCheckpointerMigration extends Migration {
+  public name = 'checkpointer';
+  constructor(
+    private dbService:DbService,
+  ) {
+    super();
+  }
+
+  private async getCheckpointerId () {
```
garethbowen, line 18:
```suggestion
  private async getCheckpointerId() {
```

webapp/src/ts/services/migrations/target-checkpointer.migration.ts
```diff
@@ -0,0 +1,58 @@
...
+      }
+      throw err;
+    }
+  }
+
+  async run() {
+    console.info('Running target checkpointer migration');
+    const localDoc = await this.getLocalCheckpointerDoc();
+    if (!localDoc) {
+      return false;
```
garethbowen, line 44: When is it possible that there is no local checkpointer? Only for new devices, right? If that's the case, then we should just mark it as run (`return true`) and never actually do the migration.

webapp/src/ts/services/db-sync.service.ts
```diff
@@ -317,17 +318,29 @@ export class DBSyncService {
     }
   }
 
+  private async migrateDb() {
+    try {
+      await this.migrationsService.runMigrations();
+    } catch (err) {
+      console.error('Error while running DB migrations', err);
+      this.sendUpdate({ state: SyncStatus.Unknown });
+      throw err;
```
garethbowen, line 327: This may be a bit aggressive if someone is offline. Can we check the status and silently fail (or console.info) so they don't get spammed with errors and feedback docs just for being offline?

tests/e2e/default/db/checkpointer-migration.wdio-spec.js
```diff
@@ -0,0 +1,211 @@
...
+    await expect(browserUtils.getDoc(MIGRATION_FLAG_ID)).to.be.rejectedWith({ status: 404 });
+
+    await browser.throttle('online');
+    await commonPage.refresh();
+    await commonPage.sync();
+
+    await browserUtils.getDoc(MIGRATION_FLAG_ID);
+  });
+
+  it('should re-upload docs when the server rolls back', async () => {
```
garethbowen, line 167: Nice

tests/e2e/default/db/db-sync.wdio-spec.js
```diff
@@ -147,46 +148,6 @@ describe('db-sync', () => {
...
-        .catch(err => callback({ err }));
-    }, doc);
-
-    if (err) {
-      throw err;
-    }
-
-    return result;
-  };
-
```
garethbowen, line 189: Thanks for cleaning these up!

garethbowen (changes requested):
Great work as always. Comments inline.

webapp/src/ts/services/migrations.service.ts
dianabarsan: It felt a little weird that resolving a promise would delete the reference to that promise, this is why I went with two variables. 
I'll change it and see how we feel.

webapp/src/ts/services/db-sync.service.ts
dianabarsan: I think that the effect is the same as if replication is attempted when the device is offline (except now replication will be blocked if it is). 
I definitely think that we should handle offline cases more gracefully, right now (without this migration code) the console log is riddled with "failed to fetch" errors.

tests/page-objects/default/common/common.wdio.page.js
dianabarsan: Unfortunately, the code on line 197 is used in an e2e test where the user gets logged out immediately after pressing the sync button, to the failed message is never displayed.

garethbowen: Right. I'm saying we can reuse the `syncWithoutWaitForSuccess` function instead of duplicating it.

webapp/src/ts/services/migrations/target-checkpointer.migration.ts
garethbowen: Perfect. Thanks for the explanation!

webapp/src/ts/services/db-sync.service.ts
garethbowen: Fair point.

garethbowen (approved):
Amazing. Thank you for taking the time to clean up other code as you went!

Looks like there's a conflict but other than that you're good to go.

Issue #8134, opened 2023-03-20
Refactor initial replication to not use PouchDb replicate


dianabarsan:
**Describe the performance issue**
When performing initial replication, the user needs to download and save all documents from the server. 
Right now, initial replication uses the standard Pouch/Couch APIs and sequence of calls: 
- call _changes
- check local revs for existence of docs
- call _bulk_get or _all_docs to download missing docs
- repeat until _changes returns no results

The _changes feed request is expensive, because it requires calls to: 
- get all allowed doc ids, which is an expensive multi-key view request 
- get a batch of changes for these doc ids

**Describe the improvement you'd like**
I'm proposing we don't use PouchDb replicate at all, and instead:
- create an API endpoint that returns the list of doc_ids and their latest revs, along with the updated_seq of the database
- the client iterates over these docs_ids, checks local revs and calls a batched _bulk_get to get all missing docs
- call .replicate, passing the since equal to the updated_seq that the server sent to create the up to date checkpointer

**To Reproduce**
Steps to record the performance metrics:
TBD

**Measurements**
TBD

**Environment**
- App: webapp, api
- Version: eg: 4.3.0

dianabarsan:
AT notes: 

- the client gets all the docs they need, and no extra ones
- the docs count warning still appears and shows the same values as before (for the same user)
- the download counter works as before
- once the app is loaded and you sync for the first time, the "fully synced" should appear quickly and you shouldn't see a flood of requests in the browser network tab
- further syncing works as expected 
- if new docs are created on the server, that the client can see, after the initial replication had been already started, the client should get those documents
- refreshing the browser page (or disconnecting from the internet) should restart replication - but only docs that don't yet exist on the client should be downloaded - requests to download docs can be seen on the network tab

I'll add more items to this list as time goes on.

dianabarsan:
This is now ready for AT on https://github.com/medic/cht-core/pull/8141.

dianabarsan:
It belongs to the infrastructure focus group. 
And thank you, you're a treasure!

dianabarsan:
Hi @bede. 
What you're seeing is not what I would expect, it's as if your initial replication did not finish and then you end up downloading all docs after the app was loaded. Can you detail the steps that you were following up to that point? 
Maybe share the instance where this can be replicated? 

Thanks!

dianabarsan:
Ok, mystery solved :)

[...]

PR #8141, opened 2023-03-23
perf(#8134): Refactor initial replication

dianabarsan:
# Description

- adds new api endpoint that will return all necessary bits for initial replication (doc-ids-revs, warn count, whether to warn and update_seq of the medic db)
- client iterates over doc-ids-revs in batches of 100, checking if existing locally. if not, downloads with attachments with _bulk_get
- calls to .replicate.from and .replicate.to with `since` parameters to generate up-to-date checkpointers to be used by subsequent syncs

medic/cht-core#8134

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:bulk-initial-replication/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:bulk-initial-replication/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:bulk-initial-replication/docker-compose/cht-couchdb-clustered.yml)
 
# License


dianabarsan:
Initially replicating a user with 6k docs on desktop: 

https://user-images.githubusercontent.com/35681649/227721143-71de3f12-e4a1-48a2-9e32-a49888806d26.mp4

Initially replicating a user with 6k docs on Android Phone (Pixel 6a):

https://user-images.githubusercontent.com/35681649/227721188-e9e3eef6-d965-4c5a-91eb-48643ce12ccc.mp4

dianabarsan:
Hello @garethbowen 

I believe this is ready for reviewing. Would you mind having a look?

api/src/routing.js
```diff
@@ -655,6 +655,14 @@ app.all(
   dbDocHandler.request,
   authorization.setAuthorized // adds the `authorized` flag to the `req` object, so it passes the firewall
 );
+const initialReplication = require('./controllers/initial-replication');
```
garethbowen, line 658: Put the `require` at the top of the doc so it fails fast if the file is missing.

api/src/routing.js
```diff
@@ -655,6 +655,14 @@ app.all(
...
+app.get(
+  '/initial-replication/get-ids',
```
garethbowen, line 660: Let's version this to future proof it a little.

```suggestion
  '/api/v1/initial-replication/get-ids',
```

api/src/routing.js
```diff
@@ -655,6 +655,14 @@ app.all(
...
+  authorization.handleAuthErrors,
+  authorization.onlineUserPassThrough,
+  jsonQueryParser,
```
garethbowen, line 663: What are we parsing here?

api/src/services/initial-replication.js
```diff
@@ -0,0 +1,39 @@
...
+    warn: unpurgedWarnIds.length >= DOC_IDS_WARN_LIMIT,
+    limit: DOC_IDS_WARN_LIMIT,
+    lastSeq: info.update_seq,
+  };
+};
+
+const getDocIdsRevPairs = async (docIds) => {
+  const result = await db.medic.allDocs({ keys: docIds });
+  return result.rows
+    .filter(row => row.value && row.value.rev)
```
garethbowen, line 32: Maybe also filter out deleted docs? Or will they already be filtered out of the `docIds`?

tests/e2e/default/enketo/add-family.wdio-spec.js
```diff
@@ -31,7 +31,7 @@ describe('Family form', () => {
     await cookieLogin();
   });
 
-  it('Submit Add Family form', async () => {
+  xit('Submit Add Family form', async () => {
```
garethbowen, line 34: Flake?

tests/factories/cht/reports/pregnancy.js
```diff
@@ -233,82 +224,90 @@ const defaultFields = {
...
+
+const geo = {
+  'latitude':  0.999151,
+  'longitude': 35.150476,
+  'altitude': null,
+  'accuracy': 1518,
+  'altitudeAccuracy': null,
+  'heading': null,
+  'speed': null
+};
```
garethbowen, line 296: Thanks for cleaning these up - what you've done is a definite improvement so if you want to go ahead without any further changes that's fine, but...

There's a whole lot of repetition here, right? Can we make new factories for the geo, geolog, and hidden fields objects so we can reuse those too?

tests/utils.js
```diff
@@ -226,9 +226,12 @@ const deleteAll = (except) => {
...
         .map(({ doc }) => {
-          doc._deleted = true;
-          doc.type = 'tombstone'; // circumvent tombstones being created when DB is cleaned up
-          return doc;
+          return {
+            _id: doc._id,
+            _rev: doc._rev,
+            _deleted: true,
+            type: 'tombstone' // circumvent tombstones being created when DB is cleaned up
+          };
```
garethbowen, line 234: Nice one.

tests/utils.js
```diff
@@ -319,6 +322,17 @@ const setUserContactDoc = (attempt=0) => {
...
+const deleteLocalDocs = async () => {
+  const localDocs = await module.exports.requestOnTestDb({ path: '/_local_docs?include_docs=true' });
+
+  for (const row of localDocs.rows) {
+    if (row && row.doc && row.doc.replicator === 'pouchdb') {
+      row.doc._deleted = true;
+      await module.exports.saveDoc(row.doc);
+    }
+  }
+};
```
garethbowen, line 334: Let's bulk docs this for performance reasons.

webapp/src/js/bootstrapper/initial-replication.js
```diff
@@ -0,0 +1,190 @@
...
+  const results = await localDb.allDocs({ keys: requiredDocs });
+  const errors = results.rows.filter(row => row.error);
+  const hasRequiredDocs = !errors.length;
+
+  const replicationLog = await getReplicationLog(localDb);
+
+  if (!hasRequiredDocs) {
+    return true;
+  }
+
```
garethbowen, line 178: Rearrange this for performance reasons...

```suggestion
  if (!hasRequiredDocs) {
    return true;
  }
  const replicationLog = await getReplicationLog(localDb);
```

webapp/src/js/bootstrapper/initial-replication.js
```diff
@@ -0,0 +1,190 @@
...
+};
+const isReplicationNeeded = async (localDb, userCtx) => {
+  const requiredDocs = [
+    '_design/medic-client',
+    'settings',
+    `org.couchdb.user:${userCtx.name}`,
+  ];
+  const results = await localDb.allDocs({ keys: requiredDocs });
+  const errors = results.rows.filter(row => row.error);
+  const hasRequiredDocs = !errors.length;
```
garethbowen, line 171: Use `Array.some` ?

```suggestion
  const missingDocs = results.rows.some(row => row.error);
```

webapp/tests/mocha/unit/initial-replication.spec.js
```diff
@@ -0,0 +1,968 @@
...
+
+let localDb;
+let remoteDb;
+let userCtx;
+let clock;
+let onPromise;
+
+const FLAG_ID = '_local/initial-replication';
+
+/** global window **/
```
garethbowen, line 16: I see this showing up more and more... It's an eslint flag, right? Why not update the eslint config instead?

garethbowen (changes requested):
Awesome work. Thanks for taking the time to refactor and clean things up, particularly the bootstrapper!

tests/e2e/default/enketo/add-family.wdio-spec.js
dianabarsan: Yea, repeated flake in many different ways.

garethbowen:
Also a nit pick about the commit message... I think it should be `perf(...)` not `feat(...)`?

api/src/services/initial-replication.js
dianabarsan: Deleted docs would be excluded automatically, since the ids are view query results. The only way a deleted doc can make it here is if it is deleted between the view request and this one.

api/src/routing.js
dianabarsan: Oh, thanks for noticing. Initially I was passing some info, but that's now been removed.

garethbowen (approved):
Great work! This will make a huge difference to new users.

webapp/src/js/bootstrapper/initial-replication.js
```diff
@@ -0,0 +1,188 @@
...
+  await localDb.put(replicationLog);
+};
+
+const getDataUsage = () => {
+  if (window.medicmobile_android && typeof window.medicmobile_android.getDataUsage === 'function') {
+    return JSON.parse(window.medicmobile_android.getDataUsage());
+  }
+};
+
+const displayTooManyDocsWarning = ({ warn_docs, limit }) => {
```
kennsippell, line 53: good fit for ui-status.js?

webapp/src/js/bootstrapper/initial-replication.js
```diff
@@ -0,0 +1,188 @@
...
+    return await localDb.get(INITIAL_REPLICATION_LOG);
+  } catch (err) {
+    return null;
+  }
+};
+const isReplicationNeeded = async (localDb, userCtx) => {
+  const requiredDocs = [
+    '_design/medic-client',
+    'settings',
+    `org.couchdb.user:${userCtx.name}`,
```
kennsippell, line 167: this doc wasn't previously required right? seems like a good requirement
i'm curious if you're being defensive here or if you are fixing any known problem

webapp/src/js/bootstrapper/initial-replication.js
```diff
@@ -0,0 +1,188 @@
...
+  const results = await localDb.allDocs({ keys: requiredDocs });
+  const missingDocs = results.rows.some(row => row.error);
+
+  if (missingDocs) {
+    return true;
+  }
+
+  const replicationLog = await getReplicationLog(localDb);
+  // new user who has started replicating, but did not complete
+  if (replicationLog && !replicationLog.complete) {
```
kennsippell, line 178: nit: could just `return replicationLog && !replicationLog.complete` here

api/src/services/initial-replication.js
```diff
@@ -0,0 +1,39 @@
...
+const getContext = async (userCtx) => {
+  const info = await db.medic.info();
+  const authContext = await authorization.getAuthorizationContext(userCtx);
+  const docsByReplicationKey = await authorization.getDocsByReplicationKey(authContext);
+
+  const excludeTombstones = { includeTombstones: false };
+  const allowedIds = authorization.filterAllowedDocIds(authContext, docsByReplicationKey, excludeTombstones);
+  const unpurgedIds = await purgedDocs.getUnPurgedIds(userCtx.roles, allowedIds);
+
+  const excludeTombstonesAndTasks = { includeTombstones: false, includeTasks: false };
```
kennsippell, line 16: does this mean we will not replicate tasks up during initial replication? is that a change in behaviour from today?

tests/e2e/default/db/initial-replication.wdio-spec.js
```diff
@@ -0,0 +1,237 @@
...
+    // it should not restart initial replication if the local doc is missing on refresh
+    await browserUtils.deleteDoc(LOCAL_LOG);
+    await refreshAndWaitForAngular();
+
+    // it should support reloading the page while offline
+    await browser.throttle('offline');
+    await refreshAndWaitForAngular();
+  });
+
+  it('should support "disconnects"', async () => {
```
kennsippell, line 228: did you consider a test which confirms that this initial replication stuff yields the same documents as the standard replication path?  a "parity test"? 

i'm not sure how easy/hard that is but i like that more holistic pattern of test for cases like this -- where we will maintain two codepaths and expect them to achieve the same result

kennsippell (commented):
what an exciting change!

i'm just trying to read and learn. left a few comments while i'm here

webapp/src/js/bootstrapper/initial-replication.js
kennsippell: i also wonder what happens if initial replication fails (for whatever reason). 
when the user retries they will get the slow experience if these few docs were successful
why don't we put them into this 10x faster experience until initial replication actually completes?

webapp/src/js/bootstrapper/initial-replication.js
```diff
@@ -0,0 +1,188 @@
...
+};
+
+const getReplicationLog = async (localDb) => {
+  try {
+    return await localDb.get(INITIAL_REPLICATION_LOG);
+  } catch (err) {
+    return null;
+  }
+};
+const isReplicationNeeded = async (localDb, userCtx) => {
```
kennsippell, line 163: so if initial replication fails for whatever reason: when the user retries, they will get the slow/expensive replication experience if these few docs were successful...

why don't we put them into this 10x faster experience until initial replication fully completes?

garethbowen: I think that's what the `replicationLog.complete` check does on line 178, so you have to have the required docs AND have a complete replication log before you go to slow mode.

webapp/src/js/bootstrapper/initial-replication.js
```diff
@@ -0,0 +1,188 @@
...
+    `org.couchdb.user:${userCtx.name}`,
+  ];
+  const results = await localDb.allDocs({ keys: requiredDocs });
+  const missingDocs = results.rows.some(row => row.error);
+
+  if (missingDocs) {
+    return true;
+  }
+
+  const replicationLog = await getReplicationLog(localDb);
```
garethbowen, line 176: I'm wondering if we should change the logic here a little... It would be more efficient to check for `replicationLog.complete` and return false before checking for missing docs, and that will be far more common use case as it'll be hit every time you start up rather than just the first few times.

The downside would be if there's any sort of data loss where the required docs are deleted somehow, then the app would load without the required docs.

webapp/src/js/bootstrapper/initial-replication.js
dianabarsan: @garethbowen is correct. I create a flag doc before starting the quick initial replication. The flag is created without a `complete` property. The quick initial replication keep restarting on every refresh until it completes.

webapp/src/js/bootstrapper/initial-replication.js
dianabarsan: > The downside would be if there's any sort of data loss where the required docs are deleted somehow, then the app would load without the required docs.

I'm happy to sacrifice the 100ms it takes to query for 3 docs. I think we would be regressing if we made this change.

webapp/src/js/bootstrapper/initial-replication.js
dianabarsan: > if you are fixing any known problem

I would not call it known - because there is no associated issue afaik, but it is a problem old initial replication had, where a user's app could start up without the user-settings doc and have all sorts of issues - like failing to load tasks for example. 

I'll make some time this week to simulate and create the issue, just so it's logged somewhere.

webapp/src/js/bootstrapper/initial-replication.js
dianabarsan: Yea, I agree! I'll move it! Thanks for noticing.

Issue #6281, opened 2020-03-09
Allow app builders to specify disabled languages in config


garethbowen:
**What feature do you want to improve?**
Whether or not a language is disabled is stored in generated docs in the database and can only be set via the app management console. This makes it less likely that app builders will define the list of languages that are actually supported and users are confronted with a long list of broken languages to choose from. Most apps probably support only one or two languages fully so this can be simplified significantly.

**Describe the improvement you'd like**
Instead of defining whether or not the language is enabled in the messages-* doc, it should be in the appsettings doc so it can be hardcoded in the projects configuration. Make sure to update the default configuration to only include fully supported languages to improve the out-of-the box experience. Also include a migration so as not to break existing projects that _have_ configured it correctly.

**Describe alternatives you've considered**
We could leave the enabled flag where it is and provide a medic-conf hook to disable languages but this is technically more complex.

Alternatively we could disable any language not explicitly provided for in the config but that is an unintended consequence of uploading new languages.

**Additional context**
This will make the language selector in #6280 much cleaner.

m5r:
Is it a good time to get rid of the `locales` property in the default config's `app_settings.json`? There is no documentation about it in [our docs](https://docs.communityhealthtoolkit.org/apps/reference/app-settings/), it's [not supported since version 2.7.1](https://github.com/medic/cht-core/pull/2441), but it's still present in every config I've checked so far, including the default config in cht-core and the config for one of [the most recent project](https://github.com/medic/config-moh-mali-chw/blob/master/app_settings/base_settings.json#L13-L22) that went live.


> Also include a migration so as not to break existing projects that *have* configured it correctly.

I'm thinking that the migration should look like this:
- get the translations docs
- save the translations' `code` to a `enabledTranslations` (better name pending) property in app settings
- remove the `enabled` property from each translation doc

Then it's going to be mostly code cleanup to make sure every code branch relying on `locale.doc.enabled` uses something like `settings.enabledTranslations.includes(locale.doc.code)`

garethbowen:
> Is it a good time to get rid of the locales property in the default config's app_settings.json?

Definitely. Good spotting!

> save the translations' code to a enabledTranslations (better name pending) property in app settings

The downside of this approach is if the app developer then uploads the settings again it'll overwrite the `enabledTranslations` property. We might have to implement it in a backwards compatible way where you can either specify it in settings _or_ in the file as it is now, and then deprecate the old way and remove it in the next major.

m5r:
[...]

PR #8142, opened 2023-03-23
Configure enabled languages

m5r:
# Description

medic/cht-core#6281

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6281-configure-disabled-languages/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6281-configure-disabled-languages/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6281-configure-disabled-languages/docker-compose/cht-couchdb-clustered.yml)
 
# License


api/src/migrations/extract-enabled-locales.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  if (_.isObject(doc.generic) || _.isObject(doc.values)) {
+    return true;
+  }
+
+  logger.warn(`Failed to load translations for "${doc.code}"("${doc.name}"). Translations document malformed.`);
+  return false;
+};
+
```
garethbowen, line 47: As [mentioned in this comment](https://github.com/medic/cht-core/issues/6281#issuecomment-1480622151) I'm not convinced that a migration will work here.

An alternative approach would be to change all the code that...
1. Looks in the settings for `enabledLocales` and if it's found, use it
2. Falls back to the current code when it's not found.

Another option would be to find all languages that are enabled in settings _or_ in the translation doc but that makes it harder to deprecate and remove the old way.

garethbowen: > a first migration that sets the enabledLocales in settings that we release now~ish -- with the fallback logic

I don't think the migration adds anything. What I expect will happen is...

1. During the upgrade the migration will run and update the `settings` doc.
2. Now anyone looking for enabled locales will not need the fallback logic - it'll just work.
3. Soon after the app dev uploads the settings from CI or CLI and overwrites the `settings` doc which removes the `enabledLocales` property.
4. Now we're back to where we started and the fallback logic will run every time

I think it's less risky to just write the fallback logic and not bother with the migration.

m5r: Fair point, that would render the migration useless...

I'll move forward with just the fallback logic and open another issue to deprecate the old way for 5.x

m5r: Opened up the follow up issue medic/cht-core#8157

admin/src/js/controllers/edit-user.js
```diff
@@ -37,8 +36,6 @@ angular
 
     $scope.cancel = () => $uibModalInstance.dismiss();
 
-    Languages().then(languages => $scope.enabledLocales = languages);
```
garethbowen, line 40: Nice catch!

api/src/translations.js
```diff
@@ -100,7 +102,18 @@ const getTranslationDocs = () => {
 };
 
 const getEnabledLocales = () => {
-  return getTranslationDocs().then(docs => docs.filter(doc => doc.enabled));
+  return Promise.all([settingsService.get(), getTranslationDocs()])
+    .then(([settings, docs]) => {
```
garethbowen, line 106: `docs` is a bit generic... What about `translations` or `translationDocs`?

admin/src/js/controllers/display-languages.js
```diff
@@ -21,11 +21,21 @@ angular.module('controllers').controller('DisplayLanguagesCtrl',
     'use strict';
     'ngInject';
 
-    const createLocaleModel = function(doc, totalTranslations) {
+    const createLocaleModel = function(doc, totalTranslations, enabledLocales) {
```
garethbowen, line 24: TODO... settings structure - right now it's an array which means we can never extend it with any other info, eg: https://github.com/medic/cht-core/issues/8158

garethbowen (changes requested):
Great work, thanks!

A few comments inline, and I have some more general comments...

1. We should provide default `enabledLocales` configuration, so that new projects and demos work out of the box. At the moment it should probably be just English (because we're missing translations in other languages) but hopefully we can build out a small set of fully supported languages and demos will be slick! Just double check that this doesn't update existing settings when upgrading - I'm 99% sure it doesn't but worth a look.
2. You've set `enabledLocales` as an array, which is fine for this work, but if we ever want to store more information about locales then we'll need a separate settings property which is messy (for example, #8158 ). To be future proof it would be better to have it as an object with an `enabled` boolean property. If you agree so far, then we may need to change the name of the property to be more generic (eg: `locales`) but would have to be careful not to conflict with the old property. If we do decide to call it `locales` then it may be backwards compatible if we check for the `enabled` property, and default to `true`.

This is quite complex, so reach out if anything needs a better explanation!

m5r:
@garethbowen 
> To be future proof it would be better to have it as an object with an enabled boolean property.	

That's a great suggestion and considering the issue you linked to, it would be best to anticipate this change now.

> If we do decide to call it locales then it may be backwards compatible if we check for the enabled property, and default to true.

The issue I see here is that we would end up never using the fallback code branch because I haven't seen a deployment without the `locales` property in their settings despite it being deprecated for a while. In this situation, I suggest we either choose a different property name or we introduce the migration now.  
If we decide to introduce the migration now and an app dev decides to upload the settings from CI or CLI and overwrites the `settings` doc, it wouldn't be an issue **if** `settings.locales` only lists languages supported by the project.

garethbowen:
> I suggest we either choose a different property name or we introduce the migration now.

I agree. While I think `locales` is a really good name, having the old deprecated config and the new config with the same name will only lead to confusion and complicated bugs. I think we just go for a new name so we don't get any collisions and it's clear what's going on. Maybe `languages`? Maybe you can come up with something better...

jkuester (changes requested):
Couple things in no particular order:

1. Don't forget tests!
2. Just now reading the latest dicussion out on the PR, so some of my in-line suggestions are a bit out-dated (assuming `enabledLocales` is not going to be an array....)
3. Also, I agree that re-using `locales` seems like a bad idea.  `langauges` is an okay name, but I would suggest `translations` (unless I missed somthing and that is already taken....)
4.  I get that we cannot have a server-side migration to move the config out of the messages docs and into the settings.  However, the current both/and approach seems prone to confusing behavior (see below) and it forces every place that wants to check the enabled translations (admin/api/webapp) to read _both_ settings and message docs.  That is a bunch of repeated steps.
    - The confusing scenario goes like this:
        - Running old code, admin enables EN and ES
        - CHT upgraded to new version
        - Admin sets `enabledLocales = ['EN', 'ES']` in app settings
        - Admin uses App Management UI to _disable_ `ES` - this updates the settings doc to have `enabledLocales = ['EN']`
        - Admin accidentally pushes settings without `enabledLocales` - this will cause `EN` _and_ `ES` to be enabled since that is what was still stored in the docs `enabled` field.
    - My suggestion here is, at this point, to avoid referencing the `app_settings` at all when _reading_ the enabled translations
        - When reading the translations, all the places (api/webapp/admin) will just read the `medic-client/doc_by_type` view.
        - In a change listener for api or sentinel (not sure which one is a better place) we can listen for `settings` changes and add logic to update the `enabled` field on the message docs whenever the contents of `enabledLocales` does not match what is set on the messages docs 
        - In admin, when enabling/disabling the translation, it should update the `settings` doc if the `enabledLocales` structure already exists else it should update the `enabled` field on the message docs.

admin/src/js/controllers/edit-user.js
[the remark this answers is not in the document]
garethbowen: I think this used to be used back when you could change another user's locale through the admin app. Back then we had a dropdown of `enabledLocales` so you could pick one. When the ability to change another user's locale was removed we must have forgotten to remove this, so we still got the data, just never used it.

m5r:
@jkuester thanks a lot for your input! I agree that the current approach comes with some overhead but it allows deployments to upgrade without having to change their settings yet, they decide when they want to opt-in this new way of configuring their enabled languages.  
Each project has its own repo tracking `app_settings` changes so once they decide to enable this setting, it's explicitly defined in the file and won't go anywhere unless manually deleted.  
On a side note, we planned to streamline the logic and get rid of the overhead in medic/cht-core#8157

Potentially overriding enabled languages set from the UI is a real risk but I wonder how frequently that happens (vs configuring languages once at the beginning of the project and forgetting about it) and if it's worth the extra complexity? cc @garethbowen

m5r:
I initially went with `translations` but this property was already used in the past and there is the `extract-translations` migration that prevents us from reusing that property so I went with `languages`

garethbowen:
> Potentially overriding enabled languages set from the UI is a real risk but I wonder how frequently that happens (vs configuring languages once at the beginning of the project and forgetting about it) and if it's worth the extra complexity?

Basically everything in the app management app will be overwritten if the project then uses cht-conf so it's acceptable. Essentially you should use one or the other but not both.

jkuester (changes requested):
Still seems really unfortunate that you can have conflicting `enabled` values set in the `messages-xx` doc vs the `settings` doc, but I suppose that is the price of shipping this feature now and not later as a breaking change....

Everything works well! Just need tests!  (Would be nice to have a dedicated e2e test that enables/disables languages via settings without touching the `messages-xx` docs.)

jkuester (approved):
Tres bien!

tests/e2e/default/translations/enabled-languages.wdio-spec.js
```diff
@@ -0,0 +1,110 @@
...
+        enabled: true,
+      },
+    ]);
+
+    // assert:
+    //   - Spanish is not available on the login page
+    //   - Swahili is available on the login page
+    const logoutButton = await $('span=Log out');
+    await logoutButton.click();
+    await loginPage.loginButton().waitForDisplayed();
```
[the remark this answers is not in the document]
m5r: I thought so too but we can't because it's meant to log the user out from the webapp, not the admin app

tests/e2e/default/translations/enabled-languages.wdio-spec.js
```diff
@@ -0,0 +1,110 @@
...
+          enabled: true,
+        },
+        {
+          locale: 'fr',
+          enabled: true,
+        },
+      ],
+    };
+    await utils.updateSettings(settings, true);
+    await browser.waitUntil(async () => {
```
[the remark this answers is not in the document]
m5r: I was seeing some runs failing because the login page didn't have the expected languages where this bit of code helped stabilize the test but I can't reproduce it anymore. I'm getting rid of it!

tests/e2e/default/translations/enabled-languages.wdio-spec.js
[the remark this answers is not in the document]
jkuester: Ah, yes!  :facepalm: Not a big deal, but this bit of functionality could also go in `admin.wdio.page.js` since it is specific to the admin console....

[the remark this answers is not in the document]
m5r: Ah, nice catch!

tatilepizs (approved):
Thank you Mokhtar for adding and updating the e2e tests 🎉

tests/page-objects/default/admin/admin.wdio.page.js
```diff
@@ -1,5 +1,35 @@
...
+
+const toggleLanguage = async (locale, shouldEnable) => {
+  await languagesPanel().waitForDisplayed();
+  await (await localePanelHeader(locale)).click();
+  const languageAccordion = await localePanelBody(locale);
+  await languageAccordion.waitForDisplayed();
+  const buttonLabel = shouldEnable ? 'Enable' : 'Disable';
+  const button = languageAccordion.$(`span=${buttonLabel}`);
+  await button.waitForDisplayed();
+  await (await button).click();
```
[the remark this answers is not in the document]
m5r: Ooh thanks for catching this!

m5r:
@garethbowen looks like I can't merge until you approve the changes :sweat_smile:

admin/tests/unit/controllers/display-languages.js
```diff
@@ -116,4 +118,132 @@ describe('Display Languages controller', function() {
...
+      ]
+    });
+
+    createController();
+    setTimeout(() => {
+      rootScope.$digest();
+      const englishLanguage = scope.languagesModel.locales.find(locale => locale.doc.code === 'en');
+      chai.expect(englishLanguage.doc.code).to.equal('en');
+      chai.expect(englishLanguage.enabled).to.eq(true);
+      scope.disableLanguage(englishLanguage.doc);
```
garethbowen, line 164: This test would be a lot cleaner if you changed `disableLanguage` to return the promise. That way instead of setting a timeout and hoping everything happens in the next tick you could just await the response.

garethbowen (changes requested):
Awesome work. Sorry for coming late to the party.

Mostly these suggestions are nitpicks but there's one bug that we should fix.

admin/src/js/controllers/display-languages.js
```diff
@@ -40,9 +49,25 @@ angular.module('controllers').controller('DisplayLanguagesCtrl',
...
+            .catch(err => {
+              $log.error('Error updating settings', err);
+            });
+        }
+
+        doc.enabled = enabled;
+        return DB().put(doc).catch(err => {
+          $log.error('Error updating translation doc', err);
+        });
       });
```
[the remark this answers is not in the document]
m5r: Nice catch!

admin/tests/unit/controllers/display-languages.js
m5r: That's much cleaner indeed, thanks!

garethbowen (approved):
Great work on those tests! They look so good now.

Issue #8089, opened 2023-02-20
Caught errors do not generate feedback documents.


dianabarsan:
**Describe the bug**
Feedback docs get generated automatically on unhandled exceptions. Webapp has been getting better and better at catching errors and displaying useful/actionable data to the user, and also missing the trigger to create a feedback doc. 
Other examples of issues where we added feedback docs deliberately for specific errors: 
- https://github.com/medic/cht-core/issues/7363 - errors in task targets
- https://github.com/medic/cht-core/issues/6629 - errors in contact-summary
- https://github.com/medic/cht-core/issues/7525 - errors in loading forms

Feedback docs are essential for assessing app health, and are the easiest way to discover most common problems in deployments. 

This bug is about errors when saving reports (submitted through reports, contacts, tasks, training forms). 

**To Reproduce**
Steps to reproduce the behavior on saving a report:
1. Upload a form that should throw an error when saving. I picked one that I was already using in another issue: https://github.com/medic/cht-core/issues/8072 . If that gets fixed first, another bugged form is required. 
2. Submit this form. 
3. After submission fails, check generated feedback docs. 
4. See that no feedback doc was generated. 

**Expected behavior**
A feedback doc should be generated when a report cannot be saved. 

**Logs**
The error gets logged correctly. 

**Environment**
- App: webapp
- Version: 4.2.0 (has been the case for a long time, version TBD)

**Additional context**
There are probably more cases where we avoid feedback doc creation on potentially critical errors. 
**We should reconsider when automatic feedback doc generation is triggered.**

kennsippell:
Example of another scenario not caught from @m5r 
https://github.com/medic/cht-core/pull/8164#pullrequestreview-1377421173

dianabarsan:
Merged to `master`

Issue #7363, opened 2021-10-05
Crash in Tasks/Targets system not creating feedback documents


kennsippell:
**Describe the bug**
Production instance was encountering the error "Error fetch tasks" for multiple users.

I'm attempting to use feedback documents to understand when this issue occurred and how many users were impacted.
https://github.com/medic/config-moh-kenya-app/issues/75

**To Reproduce**
Query postgres for feedback documents in the range that the incident occurred

```
select * from useview_feedback
where period_start > '2021-08-27T0' and period_start < '2021-08-30T0'
```

**Expected behavior**
We have observed this crash for multiple users in this date range. But there are no feedback documents indicating the crash has occurred.

**Environment**
- Instance: https://safaridoctors-ke.app.medicmobile.org/#/about
- App: safari-doctors
- Version: 3.10.0

kennsippell:
Ready for AT on `7363-rules-feedback`

To test upload invalid "stuff" within `app_settings.json` "tasks.rules". Upload to CHT and load page. You should see "error loading tasks" on tasks tab.  Look at feedback documents and you should see one with "Failed to initialize RulesEngineService"

garethbowen:
@kennsippell With 4.2.x branched I think you can go ahead and merge this.

kennsippell:
There is [this pending PR feedback](https://github.com/medic/cht-core/pull/8164#issuecomment-1503172574) which I'm struggling to find time to address. Hoping to get there in the next few weeks.

garethbowen:
Merged!

PR #8164, opened 2023-03-31
fix(#7363): Create feedback document when RulesEngineService fails to initialize

kennsippell:
# Description

This change generates a feedback document if the RulesEngineService fails to initialise.

Testing this, I was surprised by how many serious client-side issues log to console without creating a feedback document. Made me realise the importance of https://github.com/medic/cht-core/issues/8089.

#7363 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7363-rules-feedback/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7363-rules-feedback/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7363-rules-feedback/docker-compose/cht-couchdb-clustered.yml)
 
# License


dianabarsan:
Since @jkuester is out, I'll assign myself to review this PR.

webapp/src/ts/services/rules-engine.service.ts
```diff
@@ -154,6 +155,11 @@ export class RulesEngineService implements OnDestroy {
                 initializeTelemetryData.record();
               });
           });
+      })
+      .catch(error => {
+        const errorMessage = `Failed to initialize RulesEngineService`;
+        this.feedbackService.submit(errorMessage);
```
dianabarsan, line 161: The way the feedback service works is that it will include the last 10 entries that have been logged. If you catch the error here, before it gets logged, you will end up with a feedback document that is actually missing the actual error: 
image

To test this, I manually added some gibberish to the "settings.tasks.rules" property. The error that actually got logged in the console later is: 

image

I think it would be super useful to somehow include the error in the feedback doc as well.

dianabarsan (changes requested):
Thanks for this change, I added one improvement suggestion inline.

webapp/src/ts/services/rules-engine.service.ts
kennsippell: Thinking about this more - that log line contains the full compiled config. What if it is like 1MB? What if the log line happens many times (like each time the user navigates to tasks/targets tab). 

I agree it should be in the feedback doc but I'm also think we might want to truncate log lines

dianabarsan: We probably want to truncate all log lines to some decent value.

kennsippell: Looking at prod data there are many arguments with >20000 chars today. I'm thinking to truncate at 1000 characters. It's kinda arbitrary.

dianabarsan: I think you can get a fairly good idea of what the error is in 1000 chars.

kennsippell:
`JSON.stringify(new Error('msg'))` yields `"{}"` so to get the actual error details into the feedback doc, we actually need [a little voodoo](https://github.com/medic/cht-core/pull/8164/commits/967c5fcd832cffd5fbe1302a849d679967b07fcc#diff-99fd64f3964eb102f6f2d886d5673e68eb83b6802dd481aca4893ce6eb9b21b3R58)

dianabarsan (approved):
Beautiful

webapp/src/ts/services/feedback.service.ts
```diff
@@ -69,8 +73,13 @@ export class FeedbackService {
     this.LEVELS.forEach(level => {
       const original = this.options.console[level];
       this.options.console[level] = (...args) => {
+        const logEvent = {
+          level,
+          arguments: JSON.stringify(args, getCircularReplacer()).substring(0, 1000),
```
m5r, line 78: Why this particular hard limit?

m5r (changes requested):
Great work Kenn! I just left a question inline

One suggestion I'll leave here is to throw the caught error in `TasksComponent` because for now it still leaves problems like https://github.com/medic/config-moh-kenya-app/pull/76 with no feedback doc. [This](https://github.com/medic/cht-core/pull/8174/files) is what I had in mind

webapp/src/ts/services/feedback.service.ts
kennsippell: Hey @m5r !  The 1000 char limit is [discussed here](https://github.com/medic/cht-core/pull/8164#discussion_r1160168017) and I welcome your input on it!

kennsippell:
Regarding that code path in `TasksComponent`: I totally agree. 

Although I hunted this specific feedback scenario down and patched it here - I don't necessarily think we should hunt them all down one-by-one.  

Looking through the codebase there appear to be many places where errors are logged but feedback documents are not created.  In my overly-idealized fantasy - system administrators want to know about all errors happening on phones and if we don't want to know about them then they shouldn't be logged as errors. Issue https://github.com/medic/cht-core/issues/8089 suggests that we tackle this in a holistic manner and my biggest take-away from this work + your feedback is that this issue is more important than I realized.

dianabarsan:
Re: feedback documents for caught errors

I created this a while ago: https://github.com/medic/cht-core/issues/8089

m5r:
> How does that sounds to you?

I agree with the more holistic approach to this issue. I think the foundations have already been laid out with the [custom ExceptionHandler](https://github.com/medic/cht-core/blob/master/webapp/src/ts/providers/exception-handler.provider.ts) and it's now a matter of choosing what error we want to bubble up to feedback docs

> Do you have specific actions you'd like to see taken in this PR?

If you think it's relevant, I'd like to see the error in `TasksComponent` be re-thrown so that a feedback doc gets generated in case that component fails to load tasks

garethbowen:
@kennsippell Have you got time to refresh this PR and get it back for review?

dianabarsan:
@kennsippell Have you got time to refresh this PR and get it back for review?

kennsippell:
Sorry for the terrible followthrough here. I'll resolve conflicts and retest this today prior to the 4.3 release.

@m5r would it be alright if we merge this for the 4.3 release (finalizing now) and I'll handle your feedback scenario via https://github.com/medic/cht-core/issues/8089 in 4.4?

jkuester:
@kennsippell since @m5r is out until the end of the week I just wanted to jump in and say that addressing his comment as a part of #8089 seems like a good idea given the time constraints for this PR.  LGTM!  

(And if @m5r disagrees with me, he can fight me in Nairobi in a few weeks....  :laughing: )

garethbowen:
It looks like this is all good to merge, but because @m5r is out it needs an admin to override, so I'm going to do that now to unblock 4.3.0.

Thanks everyone!

Issue #8166, opened 2023-04-04
Haproxy crashes instance during scalability test due to high memory usage


dianabarsan:
**Describe the bug**
I started a scalability test, with 100 concurrent users, and haproxy crashed while the test was running. The behavior was that: 
- api returns 500 errors for every request 
- haproxy docker container was still running (and not restarting)
- last haproxy log reported all workers had been killed 

**To Reproduce**
1. Start scalability test against current `master`
2. Watch as scalability test ends with errors

**Expected behavior**
Haproxy should not crash

**Logs**
Beginning of haproxy log - to check that we're indeed on the latest config: 
```
Starting enhanced syslogd: rsyslogd.
global
  maxconn 150000
  spread-checks 5
  lua-load-per-thread /usr/local/etc/haproxy/parse_basic.lua
  lua-load-per-thread /usr/local/etc/haproxy/parse_cookie.lua
  lua-load-per-thread /usr/local/etc/haproxy/replace_password.lua
  log stdout len 65535 local2 debug
  tune.bufsize 1638400
  tune.http.maxhdr 1010
```

API Error: 
```
REQ 20d65ffb-3c6a-4290-af76-5bd28d9a5cf3 35.178.234.114 ac2 GET /medic/_changes?style=all_docs&heartbeat=10000&since=0&limit=100 HTTP/1.0
2023-04-04 14:18:54 ERROR: Server error: RequestError: Error: connect ECONNREFUSED 10.0.1.11:5984
    at new RequestError (/api/node_modules/request-promise-core/lib/errors.js:14:15)
    at Request.plumbing.callback (/api/node_modules/request-promise-core/lib/plumbing.js:87:29)
    at Request.RP$callback [as _callback] (/api/node_modules/request-promise-core/lib/plumbing.js:46:31)
    at self.callback (/api/node_modules/request/request.js:185:22)
    at Request.emit (node:events:513:28)
    at Request.onRequestError (/api/node_modules/request/request.js:877:8)
    at ClientRequest.emit (node:events:513:28)
    at Socket.socketErrorListener (node:_http_client:494:9)
    at Socket.emit (node:events:513:28)
    at emitErrorNT (node:internal/streams/destroy:157:8) {
  name: 'RequestError',
  message: 'Error: connect ECONNREFUSED 10.0.1.11:5984',
  cause: Error: connect ECONNREFUSED 10.0.1.11:5984
      at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1278:16) {
[...]

PR #8170, opened 2023-04-07
fix(#8166): Update tune.bufsize and remove rsyslog from haproxy image

dianabarsan:
# Description

According to documentation, when increasing `tune.bufsize`, `maxconn` should be decreased by the same factor, otherwise there is a risk of OOM. 
> values larger than default size will increase memory usage, possibly causing the system to run out of memory. 

https://www.haproxy.com/documentation/hapee/latest/onepage/#3.2-tune.bufsize

In our config, we both increased maxconn and tune.bufsize by x 100. This caused the haproxy memory footprint to increase to up to 18GB. This memory is not released when connections are dropped and crashed instances that had a lower amount of available memory than this. 

Additionally, this PR removes `rsyslog` from haproxy image. It was no longer needed, since docker logs are already enough for funnelling haproxy logs, and having the service running prevented the container from restarting when the main haproxy process ran out of memory. 
The container automatically restarts now, so setting a default resource limit to the container becomes an option. 

medic/cht-core#8166

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8166-new-haproxy/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8166-new-haproxy/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8166-new-haproxy/docker-compose/cht-couchdb-clustered.yml)
 
# License


dianabarsan:
I've ran some scalability tests over this new config: https://docs.google.com/spreadsheets/d/1tWlECCmB9k7yXQ4cdS0wKHSQYDAsp0emFjRT-BiiCiY/edit?usp=sharing

I am getting some errors - where haproxy sends 502s to api. 
After 4 scalability runs, haproxy uses 3GB of memory. This memory _is not released_ when the process is idle:
image

dianabarsan:
I'd love some input on this.

dianabarsan:
I'm experimenting with buffer limit right now, I can try http-server-close as well. 
Thanks a lot for the input @henokgetachew !

dianabarsan:
Ok, it turns out we already use `option http-server-close`

Below are the results of running repeated scalability tests over two variants of this PR: 
- for the first group, I completely removed the `tune.bufsize`
- for the second group, I set the `tune.bufsize` value to 2x documented default value and added ` tune.buffers.limit` as the same value as `maxconn`. 

<google-sheets-html-origin><style type="text/css"></style>

Nbr clients | Nbr docs | Mean (min) | Mean (ms) | Median (min) | Median (ms) | Errors | Bufsize
-- | -- | -- | -- | -- | -- | -- | --
  |   |   |   |   |   |   |  
300 | 1800 | 157.7068943 | 9462413.66 | 162.93505 | 9776103 | 16 | default
300 | 1800 | 146.9544652 | 8817267.91 | 157.2243583 | 9433461.5 | 5 | default
300 | 1800 | 165.3795982 | 9922775.89 | 165.3177083 | 9919062.5 | 200 | default
300 | 1800 | 160.7876768 | 9647260.61 | 163.1117167 | 9786703 | 3 | default
  |   |   |   |   |   |   |  
300 | 1800 | 160.7876768 | 9944662.7 | 163.1117167 | 9953405 | 25 | 2x default
300 | 1800 | 157.1082967 | 9426497.8 | 158.6040333 | 9516242 | 10 | 2x default
300 | 1800 | 165.7443783 | 8669043.58 | 165.8900833 | 8672980 | 1 | 2x default
300 | 1800 | 157.1082967 | 8130034.95 | 158.6040333 | 8910909.5 | 33 | 2x default

The "errors" are all identical: 
```
2023-04-17 05:33:24 ERROR: Server error: FetchError: invalid json response body at http://haproxy:5984/medicss/_design/medic/_view/docs_by_replication_key reason: Unexpected token < in JSON at position 0
```

dianabarsan:
I think I'm done with this. I've had this live on the scalability setup for a while (weeks?) and have been running scalability tests against it and it's been stable. 

Feedback / reviews please!

haproxy/entrypoint.sh
```diff
@@ -24,4 +22,4 @@ echo $COUCHDB_USER > /srv/storage/haproxy/passwd/username
 echo $COUCHDB_PASSWORD > /srv/storage/haproxy/passwd/admin
 
 # Start haproxy
-exec /usr/local/bin/docker-entrypoint.sh "$@"
+exec /usr/local/bin/docker-entrypoint.sh -f $DEFAULT -f $BACKEND
```
garethbowen, line 25: Oh this is much nicer, thanks!

haproxy/default_frontend.cfg
```diff
@@ -1,28 +1,12 @@
...
-# parameter should be decreased by the same factor as this one is increased. If an
-# HTTP request is larger than (tune.bufsize - tune.maxrewrite), HAProxy will
-# return HTTP 400 (Bad Request) error. Similarly if an HTTP response is larger
-# than this size, HAProxy will return HTTP 502 (Bad Gateway).
-
-# https://www.haproxy.com/documentation/hapee/latest/onepage/#3.2-tune.http.maxhdr
-# Similarly, too large responses
-# are blocked with "502 Bad Gateway".
+  tune.bufsize 32768
+  tune.buffers.limit 60000
```
garethbowen, line 9: The [documentation](https://www.haproxy.com/documentation/hapee/2-4r1/onepage/#tune.buffers.limit) suggests a much lower number here. My interpretation is they think it should be about 4 so we're about 15000x over. Also, this:

> It is best not to touch this value unless advised to do so by an HAProxy core developer.

Why this number? Please add a code comment.

garethbowen (commented):
I know you've extensively tested this and don't want to rock the boat, but I can't understand why our app would require these numbers, particularly the `tune.buffers.limit` number. Have I missed something?

haproxy/default_frontend.cfg
dianabarsan: From tests,  I'm seeing a lot more 502 errors with 1/10 tune.buffers.limit. 
I'll add a detailed statistic tomorrow.

garethbowen (approved):
Approved. I wish there was a way to work out what the numbers should be but if this works in testing lets run with it, and look at removing haproxy altogether later.

haproxy/default_frontend.cfg
dianabarsan: I couldn't get a single scalability workflow to work using `tune.buffers.limit 6000` :( I've reverted the change.

garethbowen: Thanks for trying! Let's run with what we've got and revisit when we evaluate auditing solutions later...

Issue #8189, opened 2023-04-19
Convert app management add/edit user modal to a page


garethbowen:
**What feature do you want to improve?**
If you go to App Management > Users > Add (or click on a row to Edit) a modal dialog is shown. Modals aren't good for forms because it restricts usability, eg: you can't hit back, you can't refresh, you can't open in a new tab, screen real estate is restricted unnecessarily, scrolling doesn't behave quite right, etc.

**Describe the improvement you'd like**
Convert the modal into a page with a unique URL.

**Describe alternatives you've considered**
- Fake it by making the modal bigger and looking like a page. This doesn't solve all the accessibility issues above.
- Leave it as is.

**Additional context**

![Screenshot from 2023-04-19 13-52-00](https://user-images.githubusercontent.com/2064938/232946073-c204d836-3efc-4119-b2a3-c05e18ace2ac.png)

isis-meng:
Hi, I attempted to push some changes to a new branch but I don't appear to have access. Is there something I should do to correct my setup? Thanks!

![image](https://user-images.githubusercontent.com/59855227/233191725-1518d8c1-19c7-4132-ab73-eeb072628519.png)

garethbowen:
@isis-meng I suggest you make a git fork of the cht-core repo into your own user, then you'll be able to create your own branches. Then when it's ready you can PR from your fork to the `medic` fork.

isis-meng:
Thanks! I've opened a PR but it's still a work-in-progress -- I've left a comment with some questions I have.

AbeebProPlus:
Hi. I would love resolve this issue if it has not been resolved

garethbowen:
Hi @AbeebProPlus ! Thanks for reaching out. @isis-meng has already started working in the above PR - how are you getting on with this issue?

AbeebProPlus:
It's just a dent on the UI. I hope it's fixed soon enough.

garethbowen:
No response on the PR review so removing from the milestone.

@AbeebProPlus If you felt like picking up where @isis-meng left off that'd be great!

Horlawale-ui:
Hello @garethbowen This PR seems to have been stale
If the PR still needs resolving, can I pick it up??

mrjones-plip:
[...]

PR #8192, opened 2023-04-19
feat(#8189): Convert Add/Edit User modal to page

isis-meng:
# Description

Convert the Add/Edit User modal into a page with a unique URL.

medic/cht-core#8189

# Code review checklist

# Compose URLs

* __CHT_CORE_COMPOSE_URL__

# License

### DIFF
--- a/admin/src/js/controllers/edit-user.js
+++ b/admin/src/js/controllers/edit-user.js
@@ -1,3 +1,5 @@
+const _ = require('lodash/core');
+
 const passwordTester = require('simple-password-tester');
 const phoneNumber = require('@medic/phone-number');
 const PASSWORD_MINIMUM_LENGTH = 8;
@@ -20,8 +22,9 @@ angular
     $q,
     $rootScope,
     $scope,
+    $state,
+    $stateParams,
     $translate,
-    $uibModalInstance,
     ContactTypes,
     CreateUser,
     DB,
@@ -35,7 +38,11 @@ angular
     'use strict';
     'ngInject';
 
-    $scope.cancel = () => $uibModalInstance.dismiss();
+    $scope.errors = {};
+    $scope.status = null;
+    $scope.saving = false;
+    $scope.user = $stateParams.user;
+    $scope.name = $stateParams.name;
 
     Languages().then(languages => $scope.enabledLocales = languages);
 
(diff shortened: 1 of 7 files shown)


isis-meng:
Some questions I have:
- What is the most appropriate way to convert error messages from modal-specific formats to be displayed for pages? (eg., lines 459-460, 474-475 in `edit-user.js`)
- What is a good way to allow information to persist in the form between refreshes?

Any advice would be really appreciated, thank you!

admin/src/js/controllers/edit-user.js
```diff
@@ -408,22 +415,22 @@ angular
...
             $scope.editUserModel.id
           );
-          $uibModalInstance.close();
+          $state.go('users');
         })
         .catch(err => {
           if (err && err.data && err.data.error && err.data.error.translationKey) {
             $translate(err.data.error.translationKey, err.data.error.translationParams).then(function(value) {
-              $scope.setError(err, value);
+              $log.error(value, err);
```
garethbowen, line 430: `$scope.setError` used to show the error in the footer of the modal so the user could see what went wrong. It looks like this is just putting it in the console which most users won't check. Let's make sure the user gets some feedback.

admin/src/js/main.js
```diff
@@ -246,6 +246,15 @@ angular.module('adminApp').config(function(
...
     })
+    .state('edit-user', {
+      url: '/users/edit/:name',
+      controller: 'EditUserCtrl',
+      templateUrl: 'templates/edit_user.html',
+      params: {
+        user: null,
+        name: null
+      },
+    })
```
garethbowen, line 257: AngularJS does some magic with these state names, so if you have them as dot separated then it understands that the two pages are related. I think it'll be simpler like this...

```suggestion
    .state('user.edit', {
      url: '/edit/:name', // now that it's a child the "/users" part is included automatically.
      controller: 'EditUserCtrl',
      templateUrl: 'templates/edit_user.html',
      params: {
        user: null,
        name: null
      },
    })
```

NB: I haven't tested this at all so you may need to tweak it a bit!

admin/src/templates/index.html
```diff
@@ -15,7 +15,7 @@
         <div class="navigation">
           <ul>
             <li ng-class="{'active-nav': checkActive('display')}" mm-auth="can_configure"><a ui-sref="display.date-time" translate>admin.display</a></li>
-            <li ui-sref-active="active-nav" mm-auth="can_configure"><a ui-sref="users" translate>Users</a></li>
+            <li ng-class="{'active-nav': checkActive('users')}" mm-auth="can_configure"><a ui-sref="users" translate>Users</a></li>
```
garethbowen, line 18: I don't think this change is needed if the state name is updated as per my previous suggestion.

admin/src/js/controllers/edit-user.js
```diff
@@ -464,17 +471,20 @@ angular
       return asynchronousValidations
         .then(valid => {
           if (!valid) {
-            $scope.setError();
+            // $scope.setError();
+            $log.info("FIX ME ASYNC VALID")
```
garethbowen, line 475: If I recall correctly the `setError` function clears the "processing" state so the loading spinner goes away, the buttons are enabled again, and maybe sets some styling on the form. We'll want to include that styling here too.

garethbowen (changes requested):
Thanks! This is looking great.

I've left a few tips inline.

> What is a good way to allow information to persist in the form between refreshes?

What information do you want to persist? Usually on web forms I would expect information to be lost.

garethbowen:
@isis-meng How are you getting on here? Is there anything I can do to help?

garethbowen:
Closing as ancient.

Issue #6505, opened 2020-06-19
Prevent users from opening and filling forms they're not authorized to see/fill


dianabarsan:
**What feature do you want to improve?**
We currently filter the report forms we display as available actions in the actionbar to the ones that the user is authorized to see (via the `context.permissions` property in the forms document).
With https://github.com/medic/cht-core/issues/6401, we also filter the create contact forms in a similar manner.
However, if a user accesses the form (be it contact or report form) by typing a valid URL that opens said form, we have no mechanism in place to authorize this action.

**Describe the improvement you'd like**
We could display an error screen, similar to the ones where the form fails to load, informing the user they're not authorized to complete the action.

**Describe alternatives you've considered**
Alternatively we could redirect the user to a "valid" page directly, and show a popup.

**Additional context**
We could, potentially, also apply this to edits (both of contacts AND reports).

garethbowen:
> We could, potentially, also apply this to edits (both of contacts AND reports).

I think it should also apply to edits. Additionally, we should display an error if the user doesn't have the `can_edit` permission.

latin-panda:
Documentation about form's [context.permission](https://docs.communityhealthtoolkit.org/apps/reference/forms/app/#properties)

PR #8200, opened 2023-04-21
chore(#6505): prevent opening unauthorized forms by url

latin-panda:
# Description

Adds checks to verify that reports and contact forms can't be access by URL if the `context.permission` and `context.expression` doesn't match.

[medic/cht-core#[number]](https://github.com/medic/cht-core/issues/6505)

# Code review checklist

# Compose URLs

* __CHT_CORE_COMPOSE_URL__

# License

### DIFF
--- a/webapp/src/ts/services/xml-forms.service.ts
+++ b/webapp/src/ts/services/xml-forms.service.ts
@@ -152,7 +152,7 @@ export class XmlFormsService {
   }
 
   private checkFormExpression(form, doc, user, contactSummary) {
-    if (!form.context.expression) {
+    if (!form.context?.expression) {
       return true;
     }
 
@@ -170,7 +170,7 @@ export class XmlFormsService {
   }
 
   private checkFormPermissions(form) {
-    if (!form.context.permission) {
+    if (!form.context?.permission) {
       return true;
     }
 
--- a/webapp/src/ts/services/xml-forms.service.ts
+++ b/webapp/src/ts/services/xml-forms.service.ts
@@ -236,8 +236,16 @@ export class XmlFormsService {
 
     return this
       .filterContactTypes(form.context, options.doc)
-      .then(valid => valid && this.checkFormPermissions(form))
-      .then(valid => valid && this.checkFormExpression(form, options.doc, user, options.contactSummary));
+      .then(valid => valid && this.canAccessForm(form, user, options));
+  }
+
+  async canAccessForm(form, userContact?, options?) {
+    if (!await this.checkFormPermissions(form)) {
+      return false;
+    }
+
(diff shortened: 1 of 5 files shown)


latin-panda:
@jkuester can you please review it? ✌️

latin-panda:
@jkuester please hold the review, I'm going to change the solution significantly

latin-panda:
@jkuester I've changed my original solution, can you please review it? Thanks!

webapp/src/ts/services/enketo.service.ts
```diff
@@ -435,49 +434,66 @@ export class EnketoService {
     });
   }
 
-  private renderForm(formContext: EnketoFormContext) {
+  private async canAccessForm(formDoc, userContact, instanceData, contactSummary) {
+    const user = userContact || await this.getUserContact();
```
jkuester, line 438: I don't think we can call `getUserContact` here since this would prevent admin users (without associated contacts) from opening _contact_ forms.

IMHO, down in the `renderContactForm` we should just add another call to `userContactService.get()` and set the result on the `formContext`. That way, here downstream in the render logic we never have to try getting the user contact again....

webapp/src/ts/services/xml-forms.service.ts
```diff
@@ -236,8 +236,16 @@ export class XmlFormsService {
...
-      .then(valid => valid && this.checkFormExpression(form, options.doc, user, options.contactSummary));
+      .then(valid => valid && this.canAccessForm(form, user, options));
+  }
+
+  async canAccessForm(form, userContact?, options?) {
+    if (!await this.checkFormPermissions(form)) {
+      return false;
+    }
+
+    const user = userContact || await this.userContactService.get();
```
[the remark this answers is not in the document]
jkuester: Ohh, I think I see. If it is an admin user without a contact, this might be null.  Still, in that case we are never going to get a value back from `userContactService.get()`, so I still don't think we need this check.

jkuester (changes requested):
Very nice!  Couple thoughts in line.  Also I wonder if this might be a good e2e test case as well (since it involves form context > webapp logic > error displayed on front-end...)

webapp/src/ts/services/enketo.service.ts
latin-panda: Looking at the code, this is for any user that doesn't have a contact associated.
Applied suggestion

latin-panda:
@ngaruko can you please have a look at this work and give it a try?

latin-panda:
@jkuester @ngaruko Just a reminder to review this PR, please :)

jkuester (approved):
:rocket:

Issue #8131, opened 2023-03-16
Training cards are appearing on top of privacy policies


dianabarsan:
**Describe the bug**
I experienced this on first login with a user, where the instance had both an active training and a privacy policy enabled. 

**To Reproduce**
1. Enable privacy policies.
2. Add a training.
3. Login with a new user.
4. See training on top of privacy policy. 

**Expected behavior**
My guess is that trainings should appear _after_ the user has accepted the privacy policy. 

**Screenshots**
![image](https://user-images.githubusercontent.com/35681649/225690319-fb7ade52-6a72-46cf-9abf-c6e176047f10.png)

**Environment**
- Instance: local
- App: webapp
- Version: 4.2.0

Elsbokk:
Hi, could I pick this up? Thanks!

dianabarsan:
Hi @Elsbokk 
You are very welcome to pick this up! Please let us know if we can assist in any capacity. 
Many thanks for taking an interest!

latin-panda:
Hi @Elsbokk, I found out yesterday that the Privacy Policy component uses the [StartupModalsService](https://github.com/medic/cht-core/blob/master/webapp/src/ts/services/startup-modals.service.ts) to display modals after the policies are accepted. Just in case you haven't seen it before and that this service can help with the task 🙂

Elsbokk:
Hi @latin-panda, I actually saw that last night! Unfortunately, I had to step away from the task for a bit due to an urgent personal matter, but now that I'm back I'm working to resolve it quickly. I've tried the simplest solution of simply treating the training cards in the same way that the startup modal service is being treated, and calling their initialization from the privacy policy component/check. This is actually working just fine, but I think that a more elegant solution would be to specifically use the startup modals service to make that call, as their behavior seems to be very similar.

Elsbokk:
Hi @dianabarsan, I have opened [this pr](https://github.com/medic/cht-core/pull/8201), please take a look whenever it's convenient.

One of the only things that I would like to note is that because I'm using the startup modals service to handle training cards, I've made some small changes to the training cards initialization process so they'll fit in with the rest of the startup modals a little cleaner, visually. 

For reference, this would happen if a training was added before setup is completed:

![training-over-welcome](https://user-images.githubusercontent.com/25468153/233226507-f82c0f8a-47b7-4a13-94e6-991a9e764a53.png)

[...]

PR #8201, opened 2023-04-21
fix(#8131): Display training after privacy policy accepted

Elsbokk:
# Description

In versions prior to 4.3, training cards would appear on top of privacy policies for instances with an active training an a privacy policy enabled. This was partially due to training cards being displayed via the `initForms` method in `app.component.ts`, independent of the status of other elements of the user's session.

This PR presents a solution in which training cards are initialized and displayed alongside the startup modals in `startup-modals.service.ts` (as per @latin-panda's suggestion). In order to display training cards in sequence with the other startup modals, rather than directly above them. 

Like the startup modals, training cards are displayed by `ModalService.show`, and returning this promise to be used in `startup-modals.service.ts` allows for trainings to be displayed in sequence with the other startup modals.

To achieve this, the method `TrainingCardsService.showTrainingCards` was added, which calls `TrainingCardsService.initTrainingCards` and returns a promise that resolves to the "shown" training cards once they are available.

The resulting behavior can be seen below:

https://user-images.githubusercontent.com/25468153/233744078-97ee10b8-ace4-423a-8169-3fa24b48deaa.mp4

medic/cht-core#8131

# Code review checklist

# Compose URLs

* __CHT_CORE_COMPOSE_URL__

# License


webapp/src/ts/app.component.ts
```diff
@@ -277,10 +275,10 @@ export class AppComponent implements OnInit, AfterViewInit {
     // initialisation tasks that can occur after the UI has been rendered
     this.setupPromise = Promise.resolve()
       .then(() => this.chtScriptApiService.isInitialized())
-      .then(() => this.checkPrivacyPolicy())
       .then(() => this.initRulesEngine())
       .then(() => this.initTransitions())
       .then(() => this.initForms())
+      .then(() => this.checkPrivacyPolicy())
```
garethbowen, line 281: I'm curious why this was moved?

garethbowen:
@Elsbokk Very nice, thank you!

It looks like there's some eslint failures on the build (try running `grunt eslint` on your command line).

webapp/src/ts/app.component.ts
Elsbokk: I actually don't remember why that was changed, it must have slipped by me. Doesn't seem like that would really do anything either, reverted to avoid unnecessary confusion.

Elsbokk:
Alright, eslint errors are fixed!

dianabarsan:
Closing in favor of https://github.com/medic/cht-core/pull/8721

Issue #8205, opened 2023-04-28
Nginx can't connect to API after container restarts because of dynamic IP allocation


dianabarsan:
**Describe the bug**
Docker allocates dynamic IPs within a network. When a container restarts, there's a chance the IP changes. However nginx caches DNS queries to backend servers and never tries them again. 
As a result, after an API restart and IP change, nginx fails to connect. 

This is a known issue: 
https://github.com/docker/compose/issues/3314
https://stackoverflow.com/questions/27912917/how-to-configure-docker-port-mapping-to-use-nginx-as-an-upstream-proxy
https://serverfault.com/questions/722438/nginx-refresh-upstream-server-ip
https://stackoverflow.com/questions/63154521/connect-failed-113-host-is-unreachable-while-connecting-to-upstream-nginx-f

**To Reproduce**
1. Start CHT 4.
1. Run `docker network inspect <your network name>` and check API's IP
1. Restart some containers to get IPs to swap. I used: 
```docker stop <haproxy> -t 0 && docker stop <api> -t 0 && docker start <haproxy> && docker start <api>```. 
For some reason, when restarting a single container, I never got an IP change, but did get it quite consistently when restarting two. 
1. Wait for one minute - or make sure everything is up and running by checking container logs. Make a request to the server. 
1. Get 502. 

**Expected behavior**
nginx should be able to remap backends ips. 

**Logs**
```
2023/04/27 21:19:33 [error] 41#41: *8603 connect() failed (111: Connection refused) while connecting to upstream, 
client: 172.30.0.1, server: , request: "GET /api/info HTTP/1.1", upstream: "http://172.30.0.7:5988/api/info", 
host: "localhost"
172.30.0.1 - admin [27/Apr/2023:21:19:33 +0000] "GET /api/info HTTP/1.1" 502 157 "-" "-"
```

The IP of api was `172.30.0.6` at this point. 

**Environment**
- App: nginx
- Version: 4+

**Additional context**
I discovered this when working on https://github.com/medic/cht-core/issues/8173, after adding an e2e test that restarted haproxy, which resulted in an expected API restart as well. Once API was up, nginx could not connect. 
Restarting nginx fixes this issue.

dianabarsan:
This is ready for AT on https://github.com/medic/cht-core/pull/8206. 
The solution is to force nginx to re-resolve all DNS entries every 10 seconds. What you should see is that 10s after API is up again, with a new IP, nginx should successfully connect and proxy requests.

[...]

PR #8025, opened 2023-01-16
Stabilize bulk delete e2e test

dianabarsan:
# Description

This test frequently failed with 
```
element ("#reports-content .report-body .details") still displayed after 15000ms
```

Example: https://github.com/medic/cht-core/actions/runs/3925903933/jobs/6711271409#step:15:1113

This is because we were opening and closing report details when multiple reports were selected, and the open/close didn't target the same report. 
I've moved the open/close test when we only select a single report. 

# Code review checklist

- [ ] Readable: Concise, well named, follows the [style guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/), documented if necessary.
- [ ] Documented: Configuration and user documentation on [cht-docs](https://github.com/medic/cht-docs/)
- [ ] Tested: Unit and/or e2e where appropriate
- [ ] Internationalised: All user facing text
- [ ] Backwards compatible: Works with existing data and configuration or includes a migration. Any breaking changes documented in the release notes.

# Compose URLs

If Build CI hasn't passed, these may 404:

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:fix-bulk-delete-e2e/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:fix-bulk-delete-e2e/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:fix-bulk-delete-e2e/docker-compose/cht-couchdb-clustered.yml)
 
# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

PR #8206, opened 2023-04-28
fix(#8205): Adds nginx resolver

dianabarsan:
# Description

medic/cht-core#8025

# Code review checklist

# Compose URLs

* __CHT_CORE_COMPOSE_URL__

# License

### DIFF
--- a/nginx/templates/server.conf.template
+++ b/nginx/templates/server.conf.template
@@ -1,5 +1,8 @@
+resolver 127.0.0.11 valid=2s; # 127.0.0.11 is from default /etc/resolv.conf
+set $upstream_endpoint http://${API_HOST}:${API_PORT};
+
 location / {
-  proxy_pass              http://${API_HOST}:${API_PORT};
+  proxy_pass              $upstream_endpoint;
   proxy_set_header        Host            $host;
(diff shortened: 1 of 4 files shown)


nginx/templates/server.conf.template
```diff
@@ -1,5 +1,8 @@
+resolver 127.0.0.11 valid=2s; # 127.0.0.11 is from default /etc/resolv.conf
```
garethbowen, line 1: 2s is quite aggressive when everything is running. I think we can increase this to get more benefit from caching. Obviously this may make outages slightly longer but I don't think it will, because my assumption is...

1. API goes down
2. up to 2 seconds later the resolver cache expires
3. some time later API comes back up
4. the DNS resolves correctly

If API takes longer than 10s to start then there's no point having the resolve timeout less than that.

I'd go go for something between 10 and 30 seconds myself.

garethbowen (approved):
Approved pending a suggestion to tweak the timeout length.

nginx/templates/server.conf.template
dianabarsan: Ok, cool, I'll bump it to 10.

Issue #7250, opened 2021-08-11
Race condition in loading contacts can cause previously selected contact to load 


dianabarsan:
**Describe the bug**
Contacts page could end up loading the previously loaded contact if a contact change is received just prior to the contact being de-selected. 

**To Reproduce**
This is very hard to reproduce with actual usage, but it has been causing a wdio e2e test to fail intermittently. 
Example of failure is: https://github.com/medic/cht-core/runs/3297387483?check_suite_focus=true#step:18:282
This failure is quite common (once every 10 suites?). 

**Expected behavior**
The selected contact shouldn't be changed the the previously selected contact.

**Environment**
- Instance: GHA Virtual machine
- App: webapp
- Version: 3.12

**Additional context**
I added helper logs in the code, so I better understand what's going on, and reran the test locally until I got a failure. This is what is happening: 

```
routing to contact "health_center_uuid"
selecting contact "health_center_uuid"
set selected contact "health_center_uuid"
contact receives external change "health_center_uuid"
selecting contact "health_center_uuid"
set selected contact null <- navigating away from the contact content page, in the case of the test, we route to contact-edit to add a new contact
set selected contact "health_center_uuid" <- the callback from the external change contact load updates the state with the old contact, so the next time we load the contacts content component, the state is actually dirty!
routing to contact "clinic_uuid" <- loading another contact 
selecting contact "clinic_uuid"
contact receives external change "health_center_uuid" <- because the state is dirty when we load the component, the code reacts to a change of the previously selected contact
selecting contact "health_center_uuid"
set selected contact "clinic_uuid"
set selected contact "health_center_uuid"
```

dianabarsan:
I just realized I had already created an issue for the flaky e2e test:https://github.com/medic/cht-core/issues/7244

latin-panda:
This is how I was able to see the problem:
- login as an online user
- set the network slow
- watch the log of actions to see the value of `selectedContact` in Store
- move from "contact view" to "contact edit", back and forth many times and quickly, but at the end land in "contact edit"
- make sure you've waited for all the logs to finish printing.
- observe when in "contact edit page" the `selectedContact` from the store is set to the last `contact view data`.  The expected is to be `null`.

I've provided a fix [here](https://github.com/medic/cht-core/pull/8234), see the logs after fix: 
<img src="https://github.com/medic/cht-core/assets/66472237/73eb77b5-ce0f-4580-b287-22b93a7421c9" width="300">

[...]

PR #8234, opened 2023-05-11
fix(#7250): race condition in loading contacts

latin-panda:
# Description

Stores the contact ID we want to fetch. Then as promises resolve, it checks that the contact we want to load is the same as from the data we had retrieved. 

medic/cht-core#7250

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7250-Race-condition-in-loading-contacts/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7250-Race-condition-in-loading-contacts/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7250-Race-condition-in-loading-contacts/docker-compose/cht-couchdb-clustered.yml)

# License


latin-panda:
@ngaruko and @dianabarsan 

There's a [job before on this topic](https://github.com/medic/cht-core/commit/bf36dd4cb543ee7d183dc2f7dcb4287144c880ba). The only part where we still need to check for the ID that it's loading is after the [first retrieve of data](https://github.com/medic/cht-core/blob/45eb3e23abe57c945656f4f6226054d583c5e4bd/webapp/src/ts/effects/contacts.effects.ts#L97).

Please have a look and let me know your thoughts.

webapp/src/ts/actions/contacts.ts
```diff
@@ -9,6 +9,7 @@ export const Actions = {
   selectContact: createMultiValueAction('SELECT_CONTACT'),
   setSelectedContact: createSingleValueAction('SET_SELECTED_CONTACT', 'selected'),
   setContactsLoadingSummary: createSingleValueAction('SET_CONTACT_LOADING_SUMMARY', 'value'),
+  setContactIdToFetch: createSingleValueAction('SET_CONTACT_ID_TO_FETCH', 'id'),
```
dianabarsan, line 12: For consistency (we don't use the term "fetch" anywhere else), could you please rename this to something like `setContactIdToSelect` ?

webapp/src/ts/effects/contacts.effects.ts
```diff
@@ -82,26 +95,34 @@ export class ContactsEffects {
...
 
-  private setTitle(selected) {
-    const routeSnapshot = this.routeSnapshotService.get();
-    const deceasedTitle = routeSnapshot?.data?.name === 'contacts.deceased'
-      ? this.translateService.instant('contact.deceased.title') : null;
-    const title = deceasedTitle || selected.type?.name_key || 'contact.profile';
-    this.globalActions.setTitle(this.translateService.instant(title));
+  private setTitle(contactId) {
+    return this
+      .verifySelectedContactNotChanged(contactId)
```
dianabarsan, line 100: I wonder if we can move this to the check to the main promise chain (to line 67) and keep this function cleaner and doing what it's called (sets the title).

dianabarsan (changes requested):
oops I accidentally hit enter without filling the textarea! 

Thanks a lot for this change, @latin-panda ! It's great! I added a small request to keep code cleaner and a renaming suggestion.

webapp/src/ts/actions/contacts.ts
latin-panda: Yeah, the name is weird. I want to make a difference when it looks for contact data and when it sets the contact as "selected" once we have the data so it's more straightforward for devs when it's used. 

Maybe `setContactIdToFind`? or `setContactIdToSummon`? 😅

dianabarsan: `setContactIdToLoad` ?

webapp/src/ts/effects/contacts.effects.ts
latin-panda: Done

webapp/src/ts/actions/contacts.ts
latin-panda: Done

latin-panda:
@ngaruko I found the steps to reproduce the issue reliably; it's in [the ticket](https://github.com/medic/cht-core/issues/7250#issuecomment-1543777876).

dianabarsan (approved):
Nice, these things are such a pain. i think we should definitely normalize this kind of flow in all other areas where you could get subsequent select events, before initial ones finish executing.

Issue #8100, opened 2023-02-23
DNS queries for *.my.local-ip.com fail causing development instances to be unreachable


mrjones-plip:
**Describe the issue**
We have 3 projects that heavily rely on the *.my.local-ip.com TLS certs:
* [NGINX Local IP](https://github.com/medic/nginx-local-ip/)
* [CHT Docker Helper for 3.x](https://docs.communityhealthtoolkit.org/apps/guides/hosting/3.x/app-developer/#cht-docker-helper)
* [CHT Docker Helper for 4.x](https://docs.communityhealthtoolkit.org/apps/guides/hosting/4.x/app-developer/#cht-docker-helper-for-4x)

Because TLS requires DNS and because the DNS server for my.local-ip.com often fails, it feels like a developers instances is broken, when in fact DNS is failing to resolve for their my.local-ip.com IP.  

**Describe the improvement you'd like**
Provide a new solution that offers the same conveniences as my.local-ip.com but has stable DNS. Either a commercial offering or a self hosted option based on open source

**Describe alternatives you've considered**
* Hope that my.local-ip.com resolves their DNS issue
* Stop offering a valid TLS cert from a real CA and revert back to only https://localhost self signed certs

mrjones-plip:
related forum/slack posts around local-ip issues:
* [ENTOFOUND error on cht-conf call with CHT Docker Helper 4.x](https://forum.communityhealthtoolkit.org/t/entofound-error-on-cht-conf-call-with-cht-docker-helper-4-x/2475/1)
* [Local-ip.co certificate expired](https://forum.communityhealthtoolkit.org/t/local-ip-co-certificate-expired/1377/3)
* ["Once the DNS issue gets resolved i'll definitely switch [to CHT Docker Helper] for the TLS certs" (private slack)](https://medic.slack.com/archives/C024KTGRW/p1677179740470009?thread_ts=1674816373.076689&cid=C024KTGRW)
*[ "I face a lot of problems, the docker containers were looking fine, but I was “disconnected” so many times, especially when I changed something in the config and tried to upload it again, I saw a lot of “fetching errors” in the browser console, and I was not able to sync, after a while it started working again, but this behavior happened to me over and over again."](https://medic.slack.com/archives/CBDHSGQ8L/p1670538521137339)

mrjones-plip:
## Research findings

### Commercial 
I did not find any commercial offerings :(

### OSS/Free services
[...]

PR #8263, opened 2023-05-22
Kubernetes configuration, Helm Chart templates and a One-Shot Script that Deploys the CHT Locally on a k3d Cluster

henokgetachew:
# Description

This Draft PR contains: 
* Kubernetes templates for the CHT. 
* Helm charts/templates
* One shot script that sets up a Distributed/Clustered CHT locally using these templates on k3d

## See it in Action

`cd`into the kubernetes directory run the setup scripts:

`./cht-deploy` if you are on Linux

`./deploy-cht-os-x` if you are on OS-X

The scripts will take a while to complete. Once that's done, a stripped down kubernetes cluster should be running locally. The CHT resources should also have been created for you. To confirm, run this:

`kubectl -n cht-dev-namespace get all`

This will give you all the resources deployed under the test namespace.

### Draft PR?
This is not yet fully complete. I see some bugs/race conditions with the code right now. I'll move it to out of draft after fleshing those out.

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:cht-k8s-configuration/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:cht-k8s-configuration/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:cht-k8s-configuration/docker-compose/cht-couchdb-clustered.yml)

# License


Hareet:
A few notes:
- If we want to use one helm chart for eks + k3d, we'll need to template the storage driver (ebs), as well as ingress, external-dns
- Do we really need services for inter-pod communication right now? It's probably making it a little harder to debug currently. Using localhost would be okay here. You can view deployment and service yamls from eks gamma-dev namespace.
- Same with the configmaps. Let's first get a deploy to work, and then template it out later.

We'll be adding some more controllers and operators here (external-dns, fluent-bit), which will require new charts so I'm of the idea to simplify the helm chart as much as possible and focus on dev 4.x charts for EKS. We'll handle k3s soon.

deploy/helm/cht-chart/.helmignore
```diff
@@ -0,0 +1,23 @@
+# Patterns to ignore when building packages.
```
garethbowen, line 1: I think I'd prefer this whole `deploy` directory to be moved to `scripts` as a subdir.

deploy/helm/cht-chart/templates/couchdb-3-deployment.yaml
```diff
@@ -0,0 +1,83 @@
...
+          hostPath:
+            path: {{ .Values.local.diskPath }}
+      {{- else }}
+      volumes:
+        - name: couchdb-3-claim0
+          persistentVolumeClaim:
+            claimName: couchdb-3-claim0
+      {{- end }}
+status: {}
+{{- end }}
```
garethbowen, line 83: This is very repetitive - is there no way to script 3 nodes so you don't have to duplicate the yaml?

garethbowen (commented):
I'm by no means an expert but I had a scan through and left some comments.

Eventually are we planning to drop docker compose altogether or maintain it in parallel? It seems like a waste of effort to keep both going... That would mean updating the docker helper tool and our CI pipeline.

Does this just work with the upgrade service as it is?

deploy/helm/cht-chart/templates/couchdb-3-deployment.yaml
henokgetachew: I have made the number of nodes configurable :). Defaulting to 3.

deploy/helm/cht-chart/.helmignore
henokgetachew: The idea was to have a directory only dealing with deployment and put different kinds of deployment options in here. Still want to move it?

garethbowen: Yes I think so. Otherwise the top level ends up too wide. But I'm open to other suggestions!

deploy/helm/cht-chart/templates/api-deployment.yaml
```diff
@@ -0,0 +1,39 @@
...
+      maxUnavailable: 1
+  template:
+    metadata:
+      labels:
+        cht.service: api
+    spec:
+      containers:
+        - env:
+            - name: BUILDS_URL
+              value: https://staging.dev.medicmobile.org/_couch/builds_4
```
Reagan, line 25: MIght it be worth externalizing this to `Values.yml`, particularly if production builds pull their build config from a different URL?

deploy/helm/cht-chart/templates/couchdb-n-deployment.yaml
```diff
@@ -0,0 +1,101 @@
...
+    matchLabels:
+      cht.service: couchdb-{{ $nodeNumber }}
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      labels:
+        cht.service: couchdb-{{ $nodeNumber }}
+    spec:
+      tolerations:
```
Reagan, line 23: I am curious here. There are a bunch of tolerations configured, particularly for couch db deployments, but there isn't documentation for applying the taints that make them applicable. Would you consider adding this documentation to a README (unless I missed the docs somewhere :) )?

deploy/helm/cht-chart/templates/eks-api-ingress.yaml
```diff
@@ -0,0 +1,29 @@
+{{- if eq .Values.environment "remote" }}
+
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: api-ingress
+  annotations:
```
Reagan, line 7: While it's likely that that this PR will be run against EKS for the foreseeable future, there may be folks who may want to run it on GKS or other k8s providers. And the annotations here, if I recall correctly, would only be relevant for ALBs. That would include generation of certs. Would it be worth:

1. Externalizing these annotations and the desired ingress controller type into `Values.yaml` and have them pulled in here with an {{ if eq .Values.ingress.type }} or something similar.
2. Support other ways of auto-generating certs and renewing them. I think this PR uses a pre-generated cert. LE certs, because of their relatively short expiry windows, for instance, may not be too amenable to this. Here, I am thinking of adding the capability to support something like `cert-manager` and it's `cluster-issuer`.

deploy/helm/cht-chart/templates/eks-api-ingress.yaml
```diff
@@ -0,0 +1,29 @@
...
+    alb.ingress.kubernetes.io/target-type: ip
+    alb.ingress.kubernetes.io/healthcheck-port: traffic-port
+    alb.ingress.kubernetes.io/certificate-arn: {{ .Values.ingress.annotations.certificate }}
+    alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS": 443}]'
+spec:
+  ingressClassName: alb
+  rules:
+    - host: {{ .Values.ingress.host }}
+      http:
+        paths:
```
Reagan, line 21: With this one path configured, since access to the couchDB web UI will not be public, do you think it will be worth adding documentation on port-forwarding access to it's service+port? Just because access to that web UI may be useful in troubleshooting.

deploy/helm/cht-chart/values.yaml
```diff
@@ -0,0 +1,34 @@
+project_name: <your-project-name>
```
Reagan, line 1: I know I have found it particularly useful to be able to run this PR with a default set of values, as defined here. But helm execution breaks because of these params with angled brackets. To help future users of this script, would it be worth changing these default params in angled params to valid helm params (e.g. `<your-project-name>` to a valid release name like `test-project`) so that future SREs get stuff running, hopefully at the first go and refactor these params to their desire later.

deploy/helm/cht-chart/templates/eks-api-ingress.yaml
henokgetachew: Good point. I'll look into that.

Reagan:
Looks really great @henokgetachew . I see that the content in the `kubernetes` folder that was probably used in the helm charts is part of the PR. And `deploy-cht` seems to really run the helm charts in `helm`. Do you still want to keep the `kubernetes` folder as part of this PR?

deploy/helm/cht-chart/templates/api-deployment.yaml
henokgetachew: I'd been moving back and forth on this. i.e. considered suggesting retiring staging.dev and using a more modern solution. However it would take a non-trivial amount of work. As it stands, that value can't be configurable because it's pretty much hard coded in the builds. @dianabarsan might have more to say here.

dianabarsan (commented):
I had a partial review pending since Nairobi and now I can't add comments :D 
I'll remove irrelevant comments after the fact.

deploy/helm/cht-chart/templates/api-deployment.yaml
```diff
@@ -0,0 +1,39 @@
...
+            - name: COUCH_URL
+              valueFrom:
+                secretKeyRef:
+                  name: cht-couchdb-credentials
+                  key: COUCH_URL
+            - name: UPGRADE_SERVICE_URL
+              value: http://upgrade-service.{{ .Values.namespace }}.svc.cluster.local:5008
+          image: public.ecr.aws/medic/cht-api:{{ .Values.cht_image_tag }}
```
dianabarsan, line 33: The upgrade service will receive the whole image tag when upgrading, like: public.ecr.aws/medic/cht-api:4.2.0. Can we use that and not hardcode the whole URL.

deploy/helm/cht-chart/templates/couchdb-n-deployment.yaml
```diff
@@ -0,0 +1,101 @@
...
+      - key: {{ $root.Values.toleration.key }}
+        operator: {{ $root.Values.toleration.operator }}
+        value: {{ $root.Values.toleration.value | quote }}
+        effect: {{ $root.Values.toleration.effect }}
+      containers:
+      - name: cht-couchdb-{{ $nodeNumber }}
+        image: public.ecr.aws/medic/cht-couchdb:{{ $root.Values.cht_image_tag }}
```
dianabarsan, line 30: Same, I think the upgrade service should send the whole image tag.

Reagan:
@henokgetachew When running the `deploy-cht` script locally, the `upgrade-service` pod never comes up and enters a CrashLoopBackOff with the following error

```
(venv) (base) reagan@reagan-MACH-WX9:~/Projects/medic/cht-core$ k logs upgrade-service-5c988d75d8-4gndb -n cht-dev-namespace

> medic-upgrade-service@1.0.0 start
> npx ts-node src/main/index.ts

Listening on port tcp://10.43.249.27:5008
Error: listen EACCES: permission denied tcp://10.43.249.27:5008
    at Server.setupListenHandle [as _listen2] (node:net:1355:21)
    at listenInCluster (node:net:1420:12)
    at Server.listen (node:net:1519:5)
    at Function.listen (/usr/src/app/node_modules/express/lib/application.js:618:24)
    at Object.<anonymous> (/usr/src/app/src/main/index.ts:63:5)
    at Module._compile (node:internal/modules/cjs/loader:1099:14)
    at Module.m._compile (/usr/src/app/node_modules/ts-node/src/index.ts:1455:23)
    at Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Object.require.extensions.<computed> [as .ts] (/usr/src/app/node_modules/ts-node/src/index.ts:1458:12)
    at Module.load (node:internal/modules/cjs/loader:975:32) {
  code: 'EACCES',
  errno: -13,
  syscall: 'listen',
  address: 'tcp://10.43.249.27:5008',
  port: -1
}
npm notice 
npm notice New major version of npm available! 8.5.2 -> 9.8.0
npm notice Changelog: <https://github.com/npm/cli/releases/tag/v9.8.0>
npm notice Run `npm install -g npm@9.8.0` to update!
npm notice 
npm notice 
npm notice New major version of npm available! 8.5.2 -> 9.8.0
npm notice Changelog: <https://github.com/npm/cli/releases/tag/v9.8.0>
npm notice Run `npm install -g npm@9.8.0` to update!
npm notice 
```
Here's the state of the other pods on the cluster

```
(venv) (base) reagan@reagan-MACH-WX9:~/Projects/medic/cht-core$ k get po -n cht-dev-namespace
NAME                                       READY   STATUS             RESTARTS      AGE
cht-haproxy-healthcheck-667dfdf47d-zpdgh   1/1     Running            0             22m
cht-couchdb-3-7d89fbf7dc-9zrkl             1/1     Running            0             22m
cht-couchdb-2-5d6bcb94-7llz5               1/1     Running            0             22m
cht-couchdb-1-7977d9d5c-2gw2k              1/1     Running            0             22m
cht-sentinel-6bd849674d-wpz7h              1/1     Running            0             22m
cht-haproxy-bb58b7b4f-8r6cv                1/1     Running            0             22m
cht-api-676dc5d479-79rw9                   1/1     Running            0             22m
upgrade-service-5c988d75d8-4gndb           0/1     CrashLoopBackOff   8 (63s ago)   22m
```
Could it be that node needs to be started with `sudo` for the `upgrade-service` as defined here `https://github.com/medic/upgrade-service-kubernetes/blob/bfca759f7c9d36ed43cd59c7e8a3ccb4b9f5bdc3/Dockerfile#L18` cc @dianabarsan

henokgetachew:
@Reagan, that's the race condition I talk about [here](https://github.com/medic/cht-core/tree/cht-k8s-configuration/deploy#known-issue) in the known issues. Use the fix script included in the code and it does work without `sudo`.

Without that script, it eventually starts working after sometime of backing off and restarting. However, we want to completely get rid of that condition. I'll look into it further this week.

Reagan:
Thanks, for the clarification @henokgetachew! I thought the race condition would only occur if the upgrade was triggered via the UI. So I've ran `deploy-race-condition-fix cht-dev-namespace` and the `upgrade-service` has restarted a number of times as you mentioned. I'll wait for it to run successfully

```
(venv) (base) reagan@reagan-MACH-WX9:~/Projects/medic/cht-core/deploy$ k get po -n cht-dev-namespace
NAME                                       READY   STATUS             RESTARTS      AGE
cht-haproxy-healthcheck-667dfdf47d-zpdgh   1/1     Running            0             78m
cht-couchdb-3-7d89fbf7dc-9zrkl             1/1     Running            0             78m
cht-couchdb-2-5d6bcb94-7llz5               1/1     Running            0             78m
cht-couchdb-1-7977d9d5c-2gw2k              1/1     Running            0             78m
cht-sentinel-6bd849674d-wpz7h              1/1     Running            0             78m
cht-haproxy-bb58b7b4f-8r6cv                1/1     Running            0             78m
cht-api-5cd8cff984-5tbwf                   1/1     Running            0             7m26s
upgrade-service-f468c4bf9-chzgt            0/1     CrashLoopBackOff   6 (42s ago)   7m26s
```

deploy/helm/cht-chart/templates/couchdb-claim0-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  labels:
+    cht.service: couchdb-claim0
+  name: couchdb-claim0
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: {{ .Values.couchdb.couchdb_node_storage_size }}
```
Reagan, line 13: I think this should be `.Values.couchdb.single_node_storageSize`

deploy/helm/cht-chart/templates/couchdb-claim0-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if and (not .Values.couchdb.clusteredCouch_enabled) (eq .Values.environment "remote") }}
```
Reagan, line 1: @henokgetachew What about `local` environment installations? They would also need a PV and PVC for database files to be written & read.

henokgetachew:
Consider race conditions solved :). It's such an interesting issue - a k8s behavior that the docs can't account for. I had to specify the port specifically for the pods to avoid trying to bind to the service's IP address with `tcp` - which meant it had to finish creation before the service got created. 🤷‍♂️

henokgetachew:
FYI: [This](https://github.com/medic/cht-core/pull/8263/commits/341d923cdc222e729d64b1c19ffb884f8940488b) commit makes sure that two cht deployments can't be deployed to the same namespace - addressing the risk of the upgrade service upgrading multiple instances at the same time.

deploy/helm/cht-chart/templates/couchdb-claim0-persistentvolumeclaim.yaml
henokgetachew: that variable has been removed in favor of this one variable that sits under the `couchdb` property.

deploy/kubernetes/ingress/k3d-api-ingress.yaml
```diff
@@ -0,0 +1,20 @@
...
+kind: Ingress
+metadata:
+  name: api-ingress
+spec:
+  tls:
+  - hosts:
+    - cht-dev.my.local-ip.co
+    secretName: api-tls-secret
+  rules:
+  - host: cht-dev.my.local-ip.co
```
mrjones-plip, line 11: I assume k3d is handling DNS such that these resolve without an IP in the URL (eg `192-168-1-20.local-ip.medicmobile.org`).  Otherwise, updating to use `local-ip.medicmobile.org` per the rest of my feedback:

```suggestion
    - cht-dev.local-ip.medicmobile.org
    secretName: api-tls-secret
  rules:
  - host: cht-dev.local-ip.medicmobile.org
```

deploy/tasks.py
```diff
@@ -0,0 +1,171 @@
...
+        if crt_file_path and key_file_path:
+            subprocess.run(["cp", crt_file_path, "certificate.crt"], check=True)
+            subprocess.run(["cp", key_file_path, "private.key"], check=True)
+        else:
+            raise Exception("certificate_crt_file_path and certificate_key_file_path must be set in values when cert_source is 'specify-file-path'")
+    elif values.get('cert_source', '') == 'my-ip-co':
+        subprocess.run(["curl", "http://local-ip.co/cert/server.pem", "-o", "certificate.crt"], check=True)
+        subprocess.run(["curl", "http://local-ip.co/cert/server.key", "-o", "private.key"], check=True)
+    elif values.get('cert_source', '') == 'eks-medic':
+        print("Moving on. Certificate provided by the eks cluster.")
```
mrjones-plip, line 76:
```suggestion
        subprocess.run(["curl", "https://local-ip.medicmobile.org/fullchain", "-o", "certificate.crt"], check=True)
        subprocess.run(["curl", "https://local-ip.medicmobile.org/key", "-o", "private.key"], check=True)
    elif values.get('cert_source', '') == 'eks-medic':
        print("Moving on. Certificate provided by the eks cluster.")
```

mrjones-plip (dismissed):
With all the trouble we had in https://github.com/medic/cht-core/issues/8100, we ended up deploying a new service at https://local-ip.medicmobile.org/ that's based on the open source project [LocalTLS](https://github.com/Corollarium/localtls).  we should use this instead of the now undependable and deprecated `local-ip.co`.

henokgetachew:
We were having a discussion about this PR with @jonathanvq.  Given that this PR is getting used in production with bugs sorted out and it has become much larger than originally intended. Remaining comments are by and large not very serious. Does anyone mind if we merged it so that it's in main and in the docs? If we feel strongly about a certain aspect, we could file a ticket and follow up with another smaller PR later on. Any objections? @Reagan @dianabarsan @garethbowen @Hareet @mrjones-plip? If no objections, can someone approve/merge it please? Thanks.

garethbowen:
@henokgetachew Can you please rebase this PR? It shouldn't conflict but there are lots of build improvements I'd like to run this code through.

henokgetachew:
@garethbowen I have rebased the PR onto `master`

garethbowen:
Green build!

@dianabarsan It looks like most of the comments left are yours - would you mind doing another pass on the review?

dianabarsan:
I think there are a couple of comments that were not addressed. @henokgetachew would you mind going over those and sharing some thoughts?

deploy/helm/cht-chart/templates/api-deployment.yaml
```diff
@@ -0,0 +1,35 @@
...
+  strategy: {}
+  template:
+    metadata:
+      labels:
+        cht.service: api
+    spec:
+      containers:
+        - env:
+            - name: BUILDS_URL
+              value: https://staging.dev.medicmobile.org/_couch/builds_4
```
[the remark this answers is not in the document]
henokgetachew: Addressed.

deploy/helm/cht-chart/templates/couchdb-1-deployment.yaml
```diff
@@ -0,0 +1,65 @@
...
+    metadata:
+      labels:
+        cht.service: couchdb-1
+    spec:
+      containers:
+        - env:
+            - name: CLUSTER_PEER_IPS
+              value: couchdb-2.{{ .Values.namespace }}.svc.cluster.local,couchdb-3-service.{{ .Values.namespace }}.svc.cluster.local
+            - name: COUCHDB_LOG_LEVEL
+              value: error
```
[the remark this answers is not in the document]
henokgetachew: Addressed.

deploy/helm/cht-chart/templates/api-deployment.yaml
henokgetachew: Addressed.

deploy/helm/cht-chart/templates/couchdb-n-deployment.yaml
henokgetachew: Addressed.

deploy/helm/cht-chart/templates/couchdb-claim0-persistentvolumeclaim.yaml
henokgetachew: This change has been performed for the clustered setup already. It will be performed for the single-node setup next.

henokgetachew:
@dianabarsan I have addressed the comments.

deploy/tasks.py
```diff
@@ -0,0 +1,171 @@
+from invoke import task
```
dianabarsan, line 1: In light of what we've discussed yesterday during the infra FG call, please create an issue to migrate this file to JavaScript.

henokgetachew: Done. Ticket filed [here](https://github.com/medic/cht-core/issues/8551).

deploy/helm/cht-chart/templates/couchdb-n-deployment.yaml
henokgetachew: Good point. This will be added in the formal documentation on the docs site.

deploy/helm/cht-chart/templates/eks-api-ingress.yaml
henokgetachew: The first comment is addressed. Check out the `.Values.cluster_type` Different ingress types will be added and executed based on that parameter. Right now we support traefik (for k3s-k3d) and alb type for eks.
Re: the second comment: there have been issues trying to manage certs on our own in the past and decided to not do them. However, I think with k8s it's worth trying. I'd say for another PR though :)

deploy/helm/cht-chart/.helmignore
henokgetachew: Addressed.

mrjones-plip:
moving to new PR here:
*  https://github.com/medic/cht-core/pull/8693

garethbowen (approved):
Approved to unblock. Please merge when ready and we can do any further improvements in future PRs.

Issue #8260, opened 2023-05-19
Migrate rapidpro e2e test from protractor to wdio 


tatilepizs:
**What feature do you want to improve?**
Migrate [rapidpro](https://github.com/medic/cht-core/blob/master/tests/e2e/protractor/sms/rapidpro.spec.js) e2e test from protractor to wdio.

**Describe the improvement you'd like**
Write the same test scenario in wdio and delete the protactor spec.

Test scenarios:

-  RapidPro SMS Gateway - Webapp Terminating messages
  ✓ should fail with no incoming key configured
  ✓ should fail with no authentication provided
  ✓ should fail with incorrect authentication
  ✓ should fail with malformed authentication
  ✓ should fail when message was not created
  ✓ should create messages
  ✓ should create reports

- RapidPro SMS Gateway - Webapp originating messages and state updates
  ✓ should not call RapidPro endpoint when credentials are not set
  ✓ should call endpoint with set credentials
  ✓ should set correct states from broadcast api
  ✓ should update the states correctly from messages api
  ✓ should not poll messages for docs that are in pending or scheduled states or lack gateway ref 
  ✓ should keep updating until record is in final state
  ✓ should poll for state updates and consume queue
  ✓ should handle rate limiting when polling

PR #8267, opened 2023-05-23
chore(#8260): migrate rapidpro e2e test

tatilepizs:
# Description

Migrate the rapidpro e2e test from protractor to wdio.

This test covers the following scenarios:

- RapidPro SMS Gateway - Webapp Terminating messages
  ✓ should fail with no incoming key configured
  ✓ should fail with no authentication provided
  ✓ should fail with incorrect authentication
  ✓ should fail with malformed authentication
  ✓ should fail when message was not created
  ✓ should create messages
  ✓ should create reports

- RapidPro SMS Gateway - Webapp originating messages and state updates
  ✓ should not call RapidPro endpoint when credentials are not set
  ✓ should call endpoint with set credentials
  ✓ should set correct states from broadcast api
  ✓ should update the states correctly from messages api
  ✓ should not poll messages for docs that are in pending or scheduled states or lack gateway ref
  ✓ should keep updating until record is in final state
  ✓ should poll for state updates and consume queue
  ✓ should handle rate limiting when polling

medic/cht-core#8260

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:migrate-rapidpro/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:migrate-rapidpro/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:migrate-rapidpro/docker-compose/cht-couchdb-clustered.yml)

# License


tests/e2e/default-mobile/reports/bulk-delete.wdio-spec.js
```diff
@@ -13,15 +13,15 @@ describe('Bulk delete reports', () => {
   const offlineUser = userFactory.build({ username: 'offline_chw_bulk_delete', place: healthCenter._id });
   const patient = personFactory.build({ parent: { _id: healthCenter._id } });
   const reports = [
-    reportFactory.build(
+    reportFactory.report().build(
```
jkuester, line 16: Please don't go back and change all of these, but I did want to note that another way you could have done this that would not have required changing every line would be to change the `require` up above to be:

```js
const { report: reportFactory } = require('@factories/cht/reports/generic-report');
```

tests/e2e/default/more-options-menu/offline-user/all-permissions.wdio-spec.js
```diff
@@ -39,15 +39,14 @@ const patient = personFactory.build({
   _id: uuid(),
   parent: { _id: clinic._id, parent: { _id: health_center._id, parent: { _id: district_hospital._id }}}
 });
-const xmlReport = reportFactory.build({ form: 'home_visit', content_type: 'xml' }, { patient, submitter: contact });
+const xmlReport = reportFactory.report().build(
+  { form: 'home_visit', content_type: 'xml' },
+  { patient, submitter: contact }
+);
```
jkuester, line 45: This is another thing that is really minor, but has tripped me up a lot in the past (and if we ever get some linting around this could cause issues).  Technically our coding style for [method chaining](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/#method-chaining) says that when chaining methods, they all need to go on their own line.  So, something like this:

```suggestion
const xmlReport = reportFactory
  .report()
  .build(
    { form: 'home_visit', content_type: 'xml' },
    { patient, submitter: contact }
  );
```

tests/e2e/default/sms/rapidpro.wdio-spec.js
```diff
@@ -0,0 +1,593 @@
...
+      try {
+        await utils.request({
+          path: endpoint,
+          method: 'POST',
+          headers: { authorization: `Token ${INCOMING_KEY}` },
+          noAuth: true,
+        });
+        throw new Error('should have thrown');
+      } catch (err) {
+        //expect(err.responseBody).to.contain({ code: 403, error: 'No incoming key configured' });
```
jkuester, line 78:
```suggestion
```

tests/e2e/default/sms/rapidpro.wdio-spec.js
```diff
@@ -0,0 +1,593 @@
...
+
+      const { heading, summary} = await messagesPage.getMessageInListDetails(rawPhone);
+      expect(heading).to.equal(rawPhone);
+      expect(summary).to.equal(smsContent);
+
+      await messagesPage.openMessage(rawPhone);
+      const { name } = await messagesPage.getMessageHeader();
+      expect(name).to.equal(rawPhone);
+
+      const { content, state, dataId } = await messagesPage.getMessageContent(1);
```
jkuester, line 185: Nice job working the `dataId` into the existing utils instead of adding more custom code for it!  :star_struck:

tests/e2e/default/sms/rapidpro.wdio-spec.js
```diff
@@ -0,0 +1,593 @@
...
+          type: 'ANC Reminders LMP',
+        },
+      ]
+    });
+
+    beforeEach(async () => {
+      const settings = {
+        sms: {
+          outgoing_service: 'rapidpro',
+          rapidpro: {url: utils.hostURL(server.address().port)},
```
jkuester, line 298:
```suggestion
          rapidpro: { url: utils.hostURL(server.address().port) },
```

tests/e2e/default/sms/rapidpro.wdio-spec.js
```diff
@@ -0,0 +1,593 @@
...
+      await utils.saveDoc(pregnancyReportWithTasks);
+      await browser.pause(1200); // interval to check the queue should run every second
+      expect(broadcastsEndpointRequests.length).to.equal(0);
+    });
+
+    it('should call endpoint with set credentials', async () => {
+      await setOutgoingKey();
+
+      await utils.saveDoc(pregnancyReportWithTasks);
+      await browser.waitUntil(() => broadcastsEndpointRequests.length === 4, 1200);
```
jkuester, line 317: triggered

tests/e2e/default/sms/rapidpro.wdio-spec.js
```diff
@@ -0,0 +1,593 @@
...
+        .map(item => item[0])
+        .sort((a, b) => a.text.localeCompare(b.text));
+
+      verifyPhoneAndMessage(bodies[0], 1);
+      verifyPhoneAndMessage(bodies[1], 3);
+      verifyPhoneAndMessage(bodies[2], 4);
+      verifyPhoneAndMessage(bodies[3], 5);
+
+      const headers = broadcastsEndpointRequests.map(item => item[1]);
+      headers.forEach(header => expect(header.authorization).to.equal(`Token ${OUTGOING_KEY}`));
```
jkuester, line 329: I like how you combined all these duplicate assertions into one here with the `forEach`, but you should probably also assert the `headers.length` too (because this line will not fail if `headers` is empty...)

tests/e2e/default/sms/rapidpro.wdio-spec.js
```diff
@@ -0,0 +1,593 @@
...
+      await browser.waitUntil(() => {
+        const requestedBroadcastIds = messagesEndpointRequests.map(([query]) => query.broadcast);
+        return _.uniq(requestedBroadcastIds).length === 7;
+      }, 4000);
+      await browser.pause(1000); // wait for the docs to actually be updated
+
+      const requestedBroadcastIds = [];
+      const expectedBroadcastIds = docs.map(doc => doc.tasks[0].gateway_ref).sort();
+      messagesEndpointRequests.forEach(([ query, headers ]) => {
+        expect(query.broadcast).to.not.be.undefined;
```
jkuester, line 395:
```suggestion
        expect(query.broadcast).to.exist;
```

This should mean the same thing right?

jkuester (approved):
10/10!  :clap:

tests/e2e/default-mobile/reports/bulk-delete.wdio-spec.js
tatilepizs: 🤦‍♀️  I didn't think about that, thank you, I will have it in mind for next time 🙂

tests/e2e/default/sms/rapidpro.wdio-spec.js
tatilepizs: That is a really good point. thank you! 👍

Issue #8280, opened 2023-05-30
Set default CouchDb log level to `info` or `debug`


dianabarsan:
**What feature do you want to improve?**
Currently, our default log level for CouchDb set in our docker-compose files is `error`. 
When debugging issues in production, especially related to performance, or load, having more logs is instrumental in understanding server activity. 

**Describe the improvement you'd like**
Make default log level more inclusive, to get more information. If this becomes too noisy for some, they can opt off. 

**Additional context**
In recent outage log analysis, we ended up requiring CouchDB access logs to work out the duration of certain requests.

PR #8282, opened 2023-05-30
chore(#8280): Changes default CouchDB log level to info

dianabarsan:
# Description

Updates CouchDb entrypoint script to always overwrite log level with provided environment variable 

medic/cht-core#8280

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8280-couch-log-level/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8280-couch-log-level/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8280-couch-log-level/docker-compose/cht-couchdb-clustered.yml)

# License


dianabarsan:
Hey @mrjones-plip mind giving this a go? thanks a bunch!

couchdb/docker-entrypoint.sh
```diff
@@ -96,6 +96,8 @@ if [ "$1" = '/opt/couchdb/bin/couchdb' ]; then
     if [ "$COUCHDB_LOG_LEVEL" ]; then
         if ! grep -Pzoqr "\[log\]\nlevel =" /opt/couchdb/etc/local.d/*.ini; then
             printf "\n[log]\nlevel = %s\n" "$COUCHDB_LOG_LEVEL" >> $CLUSTER_CREDENTIALS
+        else
+            sed -i "s/level = .*/level = $COUCHDB_LOG_LEVEL/g" $CLUSTER_CREDENTIALS
```
mrjones-plip, line 100: In a happy path search, this would find `\[log\]\nlevel =` in `CLUSTER_CREDENTIALS` which we define [line 15](https://github.com/medic/cht-core/blob/27d3446fc631fb4e111d2ebab12f52e635c32ebb/couchdb/docker-entrypoint.sh#LL15C22-L15C70) as `/opt/couchdb/etc/local.d/cluster-credentials.ini`. In this case it is fine to naively replace the `level` value in  `cluster-credentials.ini`.

However, is there a slightly unhappy path where `\[log\]\nlevel =` is found in `/opt/couchdb/etc/local.d/deployment-specific-credentials.ini`?  We would then attempt, and fail, to replace in `cluster-credentials.ini`.  Thus line 100's `sed`  would have no affect.

Another way of looking this might be: if we only ever set `level` in  `cluster-credentials.ini`, we could remove the wildcard search in line 97 and tightly bind it to `cluster-credentials.ini`.

mrjones-plip (changes requested):
Yay!  Love that you thought to make this backwards compatible so that it checks the log `level` on every run of the entry point for the service.

I'm not sure of the clustered couchdb *.ini file setup, so I asked for clarification (or possible code change) on how `level` gets set in the nested `if`.  

looking it over, I see we do deep, recursive checks already (see [here](https://github.com/medic/cht-core/blob/27d3446fc631fb4e111d2ebab12f52e635c32ebb/couchdb/docker-entrypoint.sh#L38) and [here](https://github.com/medic/cht-core/blob/27d3446fc631fb4e111d2ebab12f52e635c32ebb/couchdb/docker-entrypoint.sh#L83), for example), so I might be overthinking this!

couchdb/docker-entrypoint.sh
dianabarsan: Such an eagle eye, @mrjones-plip. Love it. 

I'm tempted to keep it simple and change the `if` to only search in `cluster-credentials.ini`. Deferring to @Hareet since he is most familiar with this code.

dianabarsan: @Hareet could you please provide some insights about this, please?

Hareet: Cool finds. I like: 
`if to only search in cluster-credentials.ini`, no need to use wildcard search

dianabarsan: Thanks for the input, @Hareet ! I've updated all the ini search functions to only target `cluster-credentials.ini` for consistency. Would you mind a proper review? Thanks!

dianabarsan:
@mrjones-plip @Hareet Can I please get some feedback on this? Thanks!

mrjones-plip:
hey @dianabarsan !  I'm digging myself out after being off for a month.  I can get to it by the end of the week though!

couchdb/docker-entrypoint.sh
mrjones-plip: much cleaner - thanks!

mrjones-plip (approved):
lgtm!

PR #8291, opened 2023-06-01
feat(#8329): API route to measure the effect of purging rules changes

m5r:
# Description

medic/cht-core#8329

- [x] copy purging behavior code from `sentinel` over to `api`
- [ ] write tests to cover the "dry run" code -- in progress
- [ ] merge both `sentinel` and `api` purging code to a shared library -- this includes changes to accept a function to run the queries unique to each project, `sentinel` uses batches of paginated queries because it needs to actually purge documents after querying whereas `api` queries can get away with a single query per type of docs because it's only reading from couchdb views

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:purging-dry-run-api/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:purging-dry-run-api/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:purging-dry-run-api/docker-compose/cht-couchdb-clustered.yml)

# License


api/src/services/purging-config.js
```diff
@@ -0,0 +1,638 @@
...
+  const willPurgeCount = sumBy(results, 'willPurgeCount');
+  const willUnpurgeCount = sumBy(results, 'willUnpurgeCount');
+
+  const nextRun = getNextRun(appSettingsPurge);
+
+  return { wontChangeCount, willPurgeCount, willUnpurgeCount, nextRun };
+};
+
+module.exports = {
+  dryRun,
```
kennsippell, line 636: curious about the expected performance for this endpoint? if purging takes several hours in production, how long would this endpoint take to execute? what implications will this have on replication performance?

mrjones-plip:
Since I didn't see it anywhere, adding that Allies is tabling this effort because of the performance concerns both @kennsippell  and @dianabarsan have expressed, as well as the test results @m5r found:

>  I got some disappointing news about our purging dry run solution :disappointed:
> 
> I've started a dry run of a purge in my morning on a clone of Muso-Mali with a beefy machine with similar specs: Xeon E5-2686 v4 @ 2.30GHz, 256 GB of RAM, ~650GB of data stored on a 1.5 TB disk. I'm using a fork of CHT 3.13.0 with the purging dry run API living on the temporary branch 3.13.0-FR-dry-run-purging.
>
>It's the beginning of the night over here and the dry run is still going. It took nearly 5 hours to simulate purging contacts, processing ~10k records with each batched request. Our assumption was that queries were cheap and mutating the data was the expensive part of purging that makes the process so slow but it turns out the queries are expensive as we're seeing roughly the same performances as actual purging despite using couchdb views.
>
>It averages 35% of CPU usage with spikes to 80% and any loss of connection between `cht-conf` and the API during the dry run results in wasted CPU usage as `cht-conf` can't reconnect to the API to wait for the results while the API keeps running the dry run.
>
> With all this, it's safe to say we cannot move forward with this solution and we should go back to the design step for this feature.

dianabarsan:
Thanks a lot for following up on this, @mrjones-plip . Can this PR be closed then?

mrjones-plip:
ah - yeah!  I'll close it for @m5r who is on leave

Issue #8296, opened 2023-06-05
Refactor downwards continuous replication following the Nairobi protocol


dianabarsan:
**What feature do you want to improve?**
The continuous replication algorithm currently uses a mix between a custom filtered replication (via two view requests + one changes request) and standard CouchDb endpoints. 

This means that: 
- we can't use multiple replicas, because we can't reliably use sticky sessions for the view requests + the changes request
- we require 2 potentially very expensive queries (one view request + changes request with many doc ids)
- the need for tombstones

**Describe the improvement you'd like**
The Nairobi protocol replaces downwards continuous replication, and instead uses an API endpoint that was added as part of the [initial replication improvement](https://github.com/medic/cht-core/issues/8134).  

Rough steps:
- Client uploads all their locally changed docs; this will still use the existing PouchDb replication. 
- Client makes a request for all doc ids and revs they should have
- Client compares with ids and revs in the pouchdb
 - Download and store all docs that don't exist in pouch or have been updated
- For docs that exist in pouch but not couch, query the server to find out if they have been deleted or purged, and if so remove from pouch

The advantages are:
- potentially no longer requires tombstones to exist
- allows for shard replicas because no more changes feed requests
- lots of complex code removed

**Describe alternatives you've considered**
[Alternatives have been described here](https://docs.google.com/document/d/1Jz6QDYDJiICz5lLEeFB3odh-YoI_0c_AKHJvoVE-Exo/edit#).

PR #8314, opened 2023-06-09
feat(#8296): the Nairobi replication protocol

dianabarsan:
# Description

Refactors continuous downwards replication to: 
- not use PouchDb replication
- use same design as initial replication, downloading a list of all doc ids, and comparing them with existent data set
- if a doc is updated or new, it gets downloaded
- if a doc is missing, we check whether it should be deleted. This leverages a new endpoint that, provided a list of docs ids, returns a list of ids that are deleted or purged. The client deletes the documents that this endpoint returns.
- purging for the client is no longer necessary, deleting purged docs occurs when calling the endpoint above, this PR removes all purging endpoints
- tombstones are no longer necessary, Sentinel no longer generates tombstones. Existent tombstones will not be removed. 
- when contacts are deleted, their associated reports are orphaned
- restricted _changes requests and _all_docs requests always return "default" changes and docs (ddoc, settings, user-settings and serviceworker) to allow for restricted users to receive upgrades after this change

medic/cht-core#8296

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8296-the-nairobi-protocol/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8296-the-nairobi-protocol/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8296-the-nairobi-protocol/docker-compose/cht-couchdb-clustered.yml)

# License


webapp/src/ts/services/db-sync.service.ts
```diff
@@ -117,58 +99,118 @@ export class DBSyncService {
...
+  private async downloadDocsBatch (batch) {
+    if (!batch.length) {
+      return;
+    }
+
+    const res = await this.dbService.get({ remote: true }).bulkGet({ docs: batch, attachments: true });
+    const docs = res.results
+      .map(result => result.docs && result.docs[0] && result.docs[0].ok)
+      .filter(doc => doc);
+    await this.dbService.get().bulkDocs(docs, { new_edits: false });
```
garethbowen, line 166: I still have concerns about partial downloads, for example, if the user has 200 docs to download, gets the first batch, then disconnects. This will leave the db in an inconsistent state. The chances of having to download > 100 is probably quite low, and it was possible to happen even with the seq ordered approach but I think we've increased the chance with this approach.

We could...

1. sort the doc ids so we get system docs, then contacts, then reports, then tasks, but that's crude at best
2. store the docs in a temporary location (maybe memory but an on disk db would probably be required) and only copy them over once all batches complete, but that's complicated.
3. just roll with it and it'll be eventually consistent (the approach right now).

What do you think? Can you convince me I'm over-engineering it?

dianabarsan: I believe the argument I used in Nairobi to convince you that you're over engineering it with transactional approach is that the changes feed, if we were to follow vanilla replication protocol, is not ordered. It will pick an amount of docs from each shard, starting from the seq provided, and then return the new seq. The amount of docs picked from each shard every time we request is not deterministic, and you will get different results every time you request a low number of changes starting from a specific seq. 
Vanilla replication is eventually consistent, and we were already in this situation. 

Additionally, the order in which the changes feed is sorted is not necessarily consistent with CHT workflows, because you can have a new contact, a new report for this contact, and then have the contact edited later on. The report will be received much earlier because the contact was pushed to the end of the changes feed. 
So another argument for: we were already inconsistent and relying on eventual consistency. 

Convinced?

garethbowen: Yeah I think so. That and the fact that most users won't exceed the BATCH_SIZE means they'll get everything in one go anyway. I still think we've made it worse, but not by enough. Let's go with it as is.

api/src/controllers/changes.js
```diff
@@ -1,316 +1,22 @@
...
-};
+const db = require('../db');
 
-const initCurrentSeq = () => db.medic.info().then(info => currentSeq = info.update_seq);
+const processRequest = async (req, res) => {
+  const docIds = [
+    'service-worker-meta',
+    '_design/medic-client',
+    'settings'
+  ];
```
garethbowen, line 9: Let's pull this out as a constant - it's sort of important.

api/src/services/authorization.js
```diff
@@ -296,7 +275,7 @@ const findContactsByReplicationKeys = (replicationKeys) => {
 
   replicationKeys = _.uniq(replicationKeys);
   const keys = [];
-  replicationKeys.forEach(id => keys.push(['shortcode', id], ['tombstone-shortcode', id]));
+  replicationKeys.forEach(id => keys.push(['shortcode', id]));
```
garethbowen, line 278: What about...

```suggestion
  const keys = replicationKeys.map(id => ['shortcode', id]);
```

api/src/services/authorization.js
```diff
@@ -312,22 +291,9 @@ const findContactsByReplicationKeys = (replicationKeys) => {
...
-        if (!result || !result.rows || !result.rows.length) {
-          return acc;
-        }
-
-        acc.push(...result.rows.map(row => row.doc).filter(doc => doc));
-        return acc;
-      }, []);
-      return contacts;
-    });
+    .then(results => results.rows.map(row => row.doc).filter(doc => doc));
```
garethbowen, line 296: As this is a hot part of the code I wonder if it's worth doing some micro-optimisations, for example, this line iterates over the results twice but it could be done in one loop but it would be less readable. This is similar in the first few lines of the `findContactsByReplicationKeys` function where we do a without, a uniq, and a map (ie: 3 iterations).

Not necessary for this iteration but it'd be cool to do a run over the hot code sometime and see if we can get some real benefits.

shared-libs/tombstone-utils/src/tombstone-utils.js
```diff
@@ -1,141 +1,11 @@
...
+  isTombstoneId: tombstoneId => module.exports.regex.test(tombstoneId),
   generateTombstoneId: generateTombstoneId,
-
   getTombstonePrefix: id => `${id}${TOMBSTONE_ID_SEPARATOR}`,
-
 };
-
-// exposed for testing
-module.exports._isDeleteStub = isDeleteStub;
-module.exports._getPreviousRev = getPreviousRev;
```
garethbowen, line 141: Is it even worth having the shared-lib any more?

tests/e2e/default/db/ongoing-replication.wdio-spec.js
```diff
@@ -0,0 +1,176 @@
...
+    await sentinelUtils.waitForSentinel();
+
+    await browser.throttle('online');
+    await commonPage.sync();
+    const localDocsPostSync = await browserUtils.getDocs();
+    const localDocIds = data.ids(localDocsPostSync);
+
+    expect(_.intersection(localDocIds, docIdsToDelete)).to.deep.equal([]);
+  });
+
```
garethbowen, line 161: - should handle conflicts (somehow)

webapp/src/ts/services/db-sync.service.ts
```diff
@@ -76,32 +77,12 @@ export class DBSyncService {
...
-      onDenied: (err?) => this.dbSyncRetryService.retryForbiddenFailure(err),
-    },
-    {
-      name: 'from',
-      options: {
-        heartbeat: 10000, // 10 seconds
-        timeout: 1000 * 60 * 10, // 10 minutes
-      },
-      allowed: () => Promise.resolve(true),
-      onChange: (replicationResult?) => this.rulesEngineService.monitorExternalChanges(replicationResult),
```
garethbowen, line 101: Do we no longer need this?

webapp/src/js/bootstrapper/initial-replication.js
```diff
@@ -50,40 +48,15 @@ const getDataUsage = () => {
...
-};
-
 const getMissingDocIdsRevsPairs = async (localDb, remoteDocIdsRevs) => {
   const localDocs = await getLocalDocList(localDb);
   return remoteDocIdsRevs.filter(({ id, rev }) => !localDocs[id] || localDocs[id] !== rev);
 };
 
-const getDownloadList = async (localDb) => {
+const getDownloadList = async (localDb = true) => {
   const response = await utils.fetchJSON('/api/v1/initial-replication/get-ids');
```
garethbowen, line 57: This is the old URL, right? I take it there's some more work to do here to migrate over. There also might be enough common code to reuse between here and the replication service.

webapp/src/ts/services/db-sync.service.ts
dianabarsan: I think we do, Thanks for pointing out this todo I forgot about!

shared-libs/tombstone-utils/src/tombstone-utils.js
dianabarsan: I thought about removing it entirely. Let me see how much it's actually used once I remove code in _all_docs and db-doc.

api/src/services/authorization.js
dianabarsan: 👍🏻 
I think I went on a preg replace spree here

shared-libs/tombstone-utils/src/tombstone-utils.js
dianabarsan: It turns out we ended up only using one function from the lib, that determined whether a doc is a tombstone depending on the format of the doc _id. 
This is still used by the API and Sentinel changes watchers to skip processing tombstones. Tombstones will continue to exist in old deployments and there are chances they get edited (or deleted) at some point. I'm partially tempted to keep this one function _somewhere_. What do you think?

api/src/services/authorization.js
dianabarsan: I agree. This is already a pretty big PR so I'm reluctant to do refactors here. 

`findContactsByReplicationKeys` is expected to iterate over a small number of elements - because it's a bottom-up search - where we're only getting the relevant contacts from the total pool of contacts that the user can see.

garethbowen: Fair enough. Let's leave optimisations till later.

shared-libs/tombstone-utils/src/tombstone-utils.js
garethbowen: Let's leave it for now. I can't think of anywhere else to put it...

garethbowen (approved):
Amazing work.

ngaruko:
@dianabarsan I wonder if there is a way to add a sort of performance test, with a little bit larger dataset?

dianabarsan:
> add a sort of performance test,

I think adding performance tests to CI suites is generally a bad idea, if that is what you are referring to. 
I have this PR that updates the performance suite to use the new endpoints: https://github.com/medic/cht-core/pull/8398
You could just run this perf suite over as large a dataset as you wish.

ngaruko (approved):
Good. Thanks for adding the PR that updates the performance suite to use the new endpoints

dianabarsan:
I kinda wanna merge this. Anyone has any objections?

garethbowen:
Do it!

Issue #7356, opened 2021-10-02
Online user stuck in loading screen when offline


dianabarsan:
**Describe the bug**
Instead of clearly communicating that there is a connection problem, an online user is stuck on a loading screen when they can't make a connection to the server (regardless of the network issue being client-side or server-side). 

**To Reproduce**
1. Login as an offline user on an instance. Check that your service worker is installed. 
2. Either: a) throttle your network to offline or b) kill your API service
3. Refresh the page
4. Look at the spinner for all eternity. 

**Expected behavior**
We should indicate that there is a connection problem instead. 

**Logs**
The browser logs are riddled with "Failed to fetch" errors, as expected. 

**Screenshots**
![image](https://user-images.githubusercontent.com/35681649/135705165-80a4e859-7468-4948-ac86-11f00ba13c6c.png)

**Environment**
- Instance: gamma.dev
- App: webapp
- Version: 3.13

garethbowen:
@dianabarsan Is this a duplicate of #7330 ?

dianabarsan:
I _think_ they can be merged together. 
The difference is that, in this issue we have app code running (so the failure is in Angular side, past bootstrap) and in the issue you linked, it was in the bootstrap code not being able to handle a disconnect.

garethbowen:
@dianabarsan The title of this issue says "Online user" but the first step of the reproduction says "Login as an offline user". That should say "Login as an online user" right?

As well as showing a connection error, we should have a retry button and/or an automatic retry.

dianabarsan:
Thanks for noticing the discrepancy, @garethbowen . Indeed this is about online users, I've updated the issue description. 

Showing an error and reload button, similarly to the screen that we display during bootstrap for offline users, would work well.

garethbowen:
@njogz Can this be closed now?

PR #8320, opened 2023-06-13
feat(#7356): check if API is accessible and set an error message if it is not

njogz:
# Description

Check if API is accessible and set an error message if it is not

medic/cht-core#7356

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7356-online-user-stuck-loading-screen/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7356-online-user-stuck-loading-screen/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7356-online-user-stuck-loading-screen/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/webapp/src/ts/modules/error/error.component.ts
+++ b/webapp/src/ts/modules/error/error.component.ts
@@ -2,6 +2,7 @@ import { Component, OnInit, OnDestroy } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
 import { Store } from '@ngrx/store';
 import { Subscription, timer } from 'rxjs';
+
 import { Selectors } from '@mm-selectors/index';
 import { TranslateService } from '@mm-services/translate.service';
 
@@ -46,6 +47,7 @@ export class ErrorComponent implements OnInit, OnDestroy {
         if (isLoaded) {
           this.timeoutElapsed = false;
         } else {
+          // Translations are not yet loaded, start timeout to give more time before displaying the default message
           this.startTimeout();
         }
         this.translationsLoaded = !!isLoaded;


webapp/src/js/bootstrapper/index.js
```diff
@@ -87,6 +87,17 @@
 
   /* pouch db set up function */
   module.exports = (POUCHDB_OPTIONS) => {
+    utils.checkApiAccessible()
```
garethbowen, line 90: It would be better to limit this to just online users so it doesn't impact offline users at all.

garethbowen, line 90: Also, this isn't blocking on the check so I'm not sure how it would work. I think you need to return or await the promise resolution.

garethbowen, line 90: Another issue with this approach is if API becomes unavailable between this request and the next one then you'll still get the same infinite spinner.

The issue talks about the console getting loads of errors - can we catch those errors and update the UI based on them rather than creating a new request? That way there's no additional network traffic and it handles a range of errors and there's no race condition where this could still be an issue.

garethbowen (changes requested):
I still have reservations about the `checkApiAccessible` call for a few reasons...

1. It adds a network request which is unnecessary. While you've improved this by only requesting for admins, it's still another delay to loading the app, because every time an admin logs in it will wait for the server to respond before actually finishing the bootstrapping. On a slow connection this will be annoying.
2. There's still a race condition - the call might succeed and then your internet drops and you still get stuck in the spinner.
3. It's very specific - if the request fails for any other reason (timeout, or authentication required, or something) this `checkApiAccessible` call will resolve because API is up, but you'll still get stuck in the spinner.

I think the proper solution is to catch the request failures in the Angular code and handle it there. I spent an hour this morning digging through the errors, and lots of them are uncaught promise rejections which is really bad. It's difficult to track them down but I think we can simply add a catch all in `app.component.ts` in the `this.setupPromise` so if any of the chained promises reject we handle it.

By adding this handler I got the spinner to go away and the app to redirect to an error page.

```
      .catch(err => {
        this.globalActions.setPrivacyPolicyAccepted(true);
      })
```

I'm sure with a little more work you can show a useful error, and ideally refresh the page periodically to check if API is back up.

I think by catching errors in the Angular init we'll handle a range of different errors not just this specific one.

What do you think?

webapp/src/ts/app.component.ts
```diff
@@ -278,7 +279,12 @@ export class AppComponent implements OnInit, AfterViewInit {
       .then(() => this.initForms())
       .then(() => this.initUnreadCount())
       .then(() => this.checkDateService.check(true))
-      .then(() => this.startRecurringProcesses());
+      .then(() => this.startRecurringProcesses())
+      .catch(err => {
+        console.error('Error during initialisation', err);
+        this.showContent = true;
+        this.router.navigate(['/error', '418' ]);
```
garethbowen, line 286: I'm a teapot???

I feel like a 503 "Service Unavailable" might be more appropriate  :P

webapp/src/ts/app.component.ts
```diff
@@ -563,6 +569,7 @@ export class AppComponent implements OnInit, AfterViewInit {
       .then(({ privacyPolicy, accepted }: any = {}) => {
         this.globalActions.setPrivacyPolicyAccepted(accepted);
         this.globalActions.setShowPrivacyPolicy(privacyPolicy);
+        this.showContent = true;
```
garethbowen, line 572: I'm not a fan of setting the value here because it's a bit unexpected when calling a function called `checkPrivacyPolicy` that it would suddenly show content. I'd add a new `then` just below line 281, eg:

```
.then(() => this.startRecurringProcesses())
.then(() => this.showContent = true)
.catch(err ...)

File: webapp/src/ts/app.component.ts [hunk 4]
```diff
@@ -92,6 +92,7 @@ export class AppComponent implements OnInit, AfterViewInit {
   reportForms;
   unreadCount = {};
   useOldActionBar = false;
+  showContent = false;
```
garethbowen, line 95: Would this be better in the Store rather than directly in the component?

Also, let's call it something that explains what it represents rather than what it does, eg:

```suggestion
  initalisationComplete = false;
```

api/resources/translations/messages-en.properties
```diff
@@ -658,6 +658,8 @@ error.403.description = You have insufficient privileges to view this page. Talk
 error.403.title = Access denied
 error.404.description = The page you were looking for was not found.
 error.404.title = Not found
+error.418.description = Error loading. Check your internet connection and try again or check with an administrator.
+error.418.title =Error
```
garethbowen, line 662:
```suggestion
error.418.title = Error
```

webapp/src/js/bootstrapper/index.js
```diff
@@ -3,7 +3,7 @@
   'use strict';
 
   const registerServiceWorker = require('./swRegister');
-  const { setUiStatus, setUiError, setLocale } = require('./ui-status');
+  const uiStatus = require('./ui-status');
```
garethbowen, line 6: I'm still not sure about this change - maybe just revert this whole file?

webapp/src/js/bootstrapper/utils.js
```diff
@@ -40,11 +40,24 @@ const fetchJSON = async (path) => {
...
+  const options = {
+    credentials: 'same-origin',
+  };
+  const response = await fetch(`${baseUrl}/api/info`, options);
+  if (response.ok) {
+    return true;
+  }
+  return false;
+};
+
```
garethbowen, line 54: I don't think this is called from anywhere any more - maybe revert it?

webapp/src/ts/modules/error/error.component.html
```diff
@@ -1,8 +1,12 @@
...
 <div class="inner">
-  <div class="col-sm-12 page help">
+  <div *ngIf="translationsLoaded" class="col-sm-12 page help">
     <h3>{{error.title | translate}}</h3>
     <p>{{error.description | translate}}</p>
   </div>
+  <div *ngIf="!translationsLoaded" class="col-sm-12 page help">
+    <h3>Error</h3>
+    <p>Error loading. Check your internet connection and try again or check with an administrator.</p>
+  </div>
```
garethbowen, line 11: Nice one!

webapp/src/ts/modules/error/error.component.ts
```diff
@@ -9,18 +10,24 @@ const errors = {
...
 export class ErrorComponent implements OnInit {
   error;
+  translationsLoaded;
 
-  constructor(private route: ActivatedRoute) {}
+  constructor(private route: ActivatedRoute, private translateService:TranslateService,) {}
 
   ngOnInit() {
     this.error = errors[this.route.snapshot.params.code] || errors['404'];
+    this.translationsLoaded = this.translateService.instant('error.403.description') !== 'error.403.description';
```
garethbowen, line 31: Does this return true if the translation isn't found, eg: a new language?

webapp/tests/mocha/unit/utils.spec.js
```diff
@@ -53,4 +53,38 @@ describe('utils', () => {
...
+
+      const result = await utils.checkApiAccessible();
+
+      expect(result).to.equal(false);
+      expect(fetch.args).to.deep.equal([[
+        'http://localhost:5984/api/info',
+        { credentials: 'same-origin' },
+      ]]);
+    });
+  });
```
garethbowen, line 89: I think you can revert this addition too.

garethbowen (changes requested):
Thanks! I like this approach. It feels like it'll catch other errors and handle them gracefully too.

webapp/src/ts/modules/error/error.component.ts
latin-panda: The translation can eventually load on a slow connection. What we usually do is check the `getTranslationsLoaded` from the store. This value is set when the app.component.ts load the translations. 

```
      .select(Selectors.getTranslationsLoaded)
      .subscribe(isLoaded => this.translationsLoaded = !!isLoaded);

// unsubscribe subscriptions in the destroy function
```

If the connection is slow, there are a few options:
- show the generic message until the translations are loaded, if ever. 
- show a spinner and timeout after a while, then show the generic message. 

IMO The second option creates less confusion for the user.

latin-panda (changes requested):
Nice Job! I added some comments

garethbowen (approved):
Awesome!

webapp/src/ts/modules/error/error.component.ts
```diff
@@ -9,18 +13,55 @@ const errors = {
...
+
+    const subscription = this.store
+      .select(Selectors.getTranslationsLoaded)
+      .subscribe(isLoaded => {
+        if (isLoaded) {
+          this.timeoutElapsed = false;
+        } else {
+          this.startTimeout();
```
latin-panda, line 49: @njogz I don't fully understand. Why we are giving extra time for setting `this.timeoutElapsed = true`? Maybe a more descriptive function name can help and/or comment.

latin-panda: If you add a commit for this, can you please add a space between third-party imports and Medic's imports, that's something we're trying to keep :)

latin-panda: I think `this.startTimeout();` should happen in line 43, before the subscription, so it shows the default message if the translations don't resolve after a long time. Because I don't see in the code that `getTranslationsLoaded` will resolve `false`.
You can explain that in the comment in the `startTimeout(){` definition.

PR #8341, opened 2023-06-25
add init / avoid removing .env

delcroip:
add init to start from env file, this is usefull for duplicating a project for instance to create a dev version

remove just rename .env to avoid loosing the db key

Project name added on the compose to be able to still use docker-compose command

correct URL if COMMOM_NAME used

### DIFF
--- a/scripts/docker-helper-4.x/cht-docker-compose.sh
+++ b/scripts/docker-helper-4.x/cht-docker-compose.sh
@@ -43,6 +43,7 @@ COUCHDB_UUID=$(openssl rand -hex 16)
 COUCHDB_DATA=$homeDir/couch
 CHT_COMPOSE_PATH=$homeDir/compose
 CHT_NETWORK=$projectName-cht-net
+PROJECT_NAME=$projectName
 EOL
 }
 
@@ -176,6 +177,10 @@ fi
 if [[ -n "${2-}" && -n $projectName ]]; then
 	containerIds=$(docker ps --all --filter "name=${projectName}" --quiet)
 	case $2 in
+	"init")
+		echo "initiating project \"${projectName}\"..."
+		create_compose_files
+		;;
 	"stop")
 		echo "Stopping project \"${projectName}\"..." | tr -d '\n'
 		docker kill $containerIds 1>/dev/null
@@ -221,8 +226,8 @@ if [[ -n "${2-}" && -n $projectName ]]; then
 		fi
 
 		if [[ -f "${projectName}.env" ]]; then
-			echo "Removing .env file in this directory..." | tr -d '\n'
-			rm ${projectName}.env
+			echo "Backing up .env file in this directory..." | tr -d '\n'
+			mv ${projectName}.env ${projectName}.env.bak
 			echo -e "${green} done${noColor} "
 		else
 			echo "No .env file found, skipping."
@@ -288,8 +293,14 @@ fi
 # shellcheck disable=SC1090
 source "./$projectFile"
 
-projectURL=$(get_local_ip_url "$(get_lan_ip)")
 
+
+
+if [[ -z "$COMMON_NAME" ]]; then
+	projectURL=$(get_local_ip_url "$(get_lan_ip)")
+else 
+	projectURL="https://$COMMON_NAME:$NGINX_HTTPS_PORT"
+fi
 echo ""
 docker-compose --env-file "./$projectFile" --file "$homeDir/upgrade-service.yml" up --detach
 
@@ -342,3 +353,4 @@ echo -e "${green} Have a great day!${noColor} "
 echo ""
 
 set -e
+
--- a/scripts/docker-helper-4.x/cht-docker-compose.sh
+++ b/scripts/docker-helper-4.x/cht-docker-compose.sh
@@ -88,9 +88,15 @@ show_help_existing_stop_and_destroy() {
     echo "Stop and destroy all project data:"
     echo "    ./cht-docker-compose.sh ENV-FILE.env destroy"
     echo ""
-    echo "https://docs.communityhealthtoolkit.org/apps/guides/hosting/4.x/app-developer/"
+    echo "Init an an ENV-FILE.env to allow customisation before server creation"
+    echo "    ./cht-docker-compose.sh ENV-FILE.env init"
     echo ""
+    echo "recreate server with an existing ENV-FILE.env"
+    echo "    ./cht-docker-compose.sh ENV-FILE.env init"
+    echo ""
+    echo "https://docs.communityhealthtoolkit.org/apps/guides/hosting/4.x/app-developer/"
 }
+
 
 get_lan_ip() {
   # init empty lan address
@@ -177,7 +183,11 @@ fi
 if [[ -n "${2-}" && -n $projectName ]]; then
 	containerIds=$(docker ps --all --filter "name=${projectName}" --quiet)
 	case $2 in
-	"init")
+ 	"init")
+		echo "initiating project \"${projectName}\"..."
+		init_env_file
+		;;
+	"recreate")
 		echo "initiating project \"${projectName}\"..."
 		create_compose_files
 		;;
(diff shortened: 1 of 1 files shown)


scripts/docker-helper-4.x/cht-docker-compose.sh
```diff
@@ -176,6 +177,10 @@ fi
 if [[ -n "${2-}" && -n $projectName ]]; then
 	containerIds=$(docker ps --all --filter "name=${projectName}" --quiet)
 	case $2 in
+	"init")
+		echo "initiating project \"${projectName}\"..."
+		create_compose_files
+		;;
```
mrjones-plip, line 183: I couldn't get this to work - can you check the logic to activate it?  I was expecting this to work: `./cht-docker-compose.sh new-project.env init`

However I got this:

```
./cht-docker-compose.sh new-project.env init

File "new-project.env" doesnt exist - be sure to include ".env" at the end!
```

If you're gong to keep this logic, please add matching output for `show_help_existing_stop_and_destroy()` [helper function](https://github.com/medic/cht-core/blob/7e148852ea4b2a7a0c36438d789c3bff7a675d09/scripts/docker-helper-4.x/cht-docker-compose.sh#L80).

scripts/docker-helper-4.x/cht-docker-compose.sh
```diff
@@ -221,8 +226,8 @@ if [[ -n "${2-}" && -n $projectName ]]; then
 		fi
 
 		if [[ -f "${projectName}.env" ]]; then
-			echo "Removing .env file in this directory..." | tr -d '\n'
-			rm ${projectName}.env
+			echo "Backing up .env file in this directory..." | tr -d '\n'
+			mv ${projectName}.env ${projectName}.env.bak
```
mrjones-plip, line 230: Neat idea!  you said:

> remove just rename .env to avoid loosing the db key

Can you explain what use case for keeping the key if you've deleted the CouchDB container and all the data in it?

scripts/docker-helper-4.x/cht-docker-compose.sh
```diff
@@ -288,8 +293,14 @@ fi
...
 
-projectURL=$(get_local_ip_url "$(get_lan_ip)")
 
+
+
+if [[ -z "$COMMON_NAME" ]]; then
+	projectURL=$(get_local_ip_url "$(get_lan_ip)")
+else 
+	projectURL="https://$COMMON_NAME:$NGINX_HTTPS_PORT"
+fi
```
mrjones-plip, line 303: I don't see any reference to `COMMON_NAME` anywhere and generating the URL with port, IP and protocol by hand is brittle and error prone.  I think we should remove this section.  

If you feel different, lemme know!

mrjones-plip (changes requested):
@delcroip - wonderful - Thanks so much for taking the time to offer some improvements to the CHT Docker Helper!  

I have some questions about the PR before we can move to merge.

scripts/docker-helper-4.x/cht-docker-compose.sh
delcroip: the idea it to have the new-project.env file already created (copy from prod or a destroyed cht) then run 

`./cht-docker-compose.sh new-project.env  init`

scripts/docker-helper-4.x/cht-docker-compose.sh
delcroip: you need to migrate the server or archive it (project ends) but you want to be able to restart it with `init`

scripts/docker-helper-4.x/cht-docker-compose.sh
delcroip: COMMON_NAME is use in case of AUTOI_GENERATE tls certificate, DOMAIN might be a better choice here

delcroip:
i see your point with init, so I create `recreate` to start for an existing env file and `init` just create an env file

delcroip:
here the doc for PROJECT_NAME https://docs.docker.com/compose/reference/#use--p-to-specify-a-project-name

mrjones-plip:
Hi @delcroip  - I was on holiday for the month of July - sorry for the delay in getting back to you.  I'll jump into this PR in the next week or so.

scripts/docker-helper-4.x/cht-docker-compose.sh
mrjones-plip: Ah - I think i see a bit better.

Is your thinking that you don't want to use `127-0-0-1.local-ip.medicmobile.org` which `get_lan_ip()` will fall back to and want to explicitly specify it when you make the CLI call?  Given we [only work with `*.local-ip.medicmobile.org` certificates](https://github.com/medic/cht-core/blob/4a51fd6cf92002361d79008f7ff5ac2385a0d58f/scripts/docker-helper-4.x/cht-docker-compose.sh#L322), maybe we could go with something like this:

```suggestion
if [[ -z "$OVERRIDE_IP" ]]; then
	projectURL=$(get_local_ip_url "$(get_lan_ip)")
else 
	projectURL=$(get_local_ip_url ${OVERRIDE_IP})
fi
```

scripts/docker-helper-4.x/cht-docker-compose.sh
```diff
@@ -87,10 +88,16 @@ show_help_existing_stop_and_destroy() {
     echo "Stop and destroy all project data:"
     echo "    ./cht-docker-compose.sh ENV-FILE.env destroy"
     echo ""
-    echo "https://docs.communityhealthtoolkit.org/apps/guides/hosting/4.x/app-developer/"
+    echo "Init an an ENV-FILE.env to allow customisation before server creation"
+    echo "    ./cht-docker-compose.sh ENV-FILE.env init"
+    echo ""
+    echo "recreate server with an existing ENV-FILE.env"
```
mrjones-plip, line 94:
```suggestion
    echo "Recreate server with an existing ENV-FILE.env:"
```

scripts/docker-helper-4.x/cht-docker-compose.sh
```diff
@@ -87,10 +88,16 @@ show_help_existing_stop_and_destroy() {
     echo "Stop and destroy all project data:"
     echo "    ./cht-docker-compose.sh ENV-FILE.env destroy"
     echo ""
-    echo "https://docs.communityhealthtoolkit.org/apps/guides/hosting/4.x/app-developer/"
+    echo "Init an an ENV-FILE.env to allow customisation before server creation"
```
mrjones-plip, line 91:
```suggestion
    echo "Initialize an an ENV-FILE.env to allow customization before server creation:"
```

scripts/docker-helper-4.x/cht-docker-compose.sh
```diff
@@ -176,6 +183,15 @@ fi
...
 	containerIds=$(docker ps --all --filter "name=${projectName}" --quiet)
 	case $2 in
+ 	"init")
+		echo "initiating project \"${projectName}\"..."
+		init_env_file
+  		exit 0
+		;;
+	"recreate")
+		echo "initiating project \"${projectName}\"..."
+		create_compose_files
```
mrjones-plip, line 193:
```suggestion
		echo "Initializing project \"${projectName}\"..."
		init_env_file
		exit 0
		;;
	"recreate")
		echo "Recreating project \"${projectName}\"..."
		create_compose_files
```

mrjones-plip:
Thanks for the continued work on this!

After jumping back into the code, I think I'm lacking a larger picture of what the use case is here for `init` and `recreate`.  Can you do a brief write up of the real world problem you're trying to solve?

scripts/docker-helper-4.x/cht-docker-compose.sh
```diff
@@ -176,6 +183,15 @@ fi
 if [[ -n "${2-}" && -n $projectName ]]; then
 	containerIds=$(docker ps --all --filter "name=${projectName}" --quiet)
 	case $2 in
+ 	"init")
```
mrjones-plip, line 186: `init` doesn't seem to work :

https://github.com/medic/cht-core/assets/8253488/fa4392e9-87cc-4d03-b565-c9bd681777d8

mrjones-plip (changes requested):

garethbowen:
@delcroip Thanks for this! Do you have time to dust it off and work on the feedback that @mrjones-plip shared?

delcroip:
Thx I will check, i am currently checking using container_name directive In the compose to allow both the scripts and plain compose usage

mrjones-plip:
@delcroip - there hasn't been any activity on this PR so I'm going to close it.  If you're still interested in working on this, please drop us a note on the PR!

PR #8345, opened 2023-06-26
swap in matomo

craig-landry:
Use Matomo here instead of other options.


craig-landry:
Instead of changing this branch I'm going to make a new parallel fr branch to better follow the convention

PR #8354, opened 2023-06-30
feat(#8426): API express internals + endpoint monitoring via Prometheus

kennsippell:
# Description

Adds new endpoint `/api_prometheus_metrics` exposing details of what's happening within CHT API Express Server via Prometheus.

Conversation in medic/cht-watchdog#60, but I'm installing from a private branch here to work around https://github.com/PayU/prometheus-api-metrics/issues/112 and https://github.com/PayU/prometheus-api-metrics/issues/114.

This exposes:
* Prom-client default metrics for Node.js
* Prom-client default metrics for Node.js Garbage collection
* prometheus-api-metrics metrics for Express endpoints

Works with [these dashboards](https://grafana.com/grafana/dashboards/12230-node-js-dashboard/)

medic/cht-watchdog#60
#8426

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:dog60-endpoint-monitoring/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:dog60-endpoint-monitoring/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:dog60-endpoint-monitoring/docker-compose/cht-couchdb-clustered.yml)

# License


api/src/routing.js
```diff
@@ -111,6 +112,16 @@ app.postJsonOrCsv = (path, callback) => handleJsonOrCsvRequest('post', path, cal
 app.postJson = (path, callback) => handleJsonRequest('post', path, callback);
 app.putJson = (path, callback) => handleJsonRequest('put', path, callback);
 
+app.use(prometheusMiddleware({
+  metricsPath: '/api_prometheus_metrics',
+  // based on one-month analysed period of production traffic
+  // each durationBuckets captures ~5% of traffic unless otherwise indicated
+  durationBuckets: [
```
kennsippell, line 119: Seeking a thought partner here

Prometheus stores values in histograms. All values are approximated into these buckets and so we need to understand the distribution of our data to get meaningful useful output.

I analysed the total_time spent in requests for 23 days of production traffic for Muso-Mali (1.2M requests). This is generally regarded as a "healthy instance" which is well used and loved. I grouped and counted them all by the closest second. [Result is here](https://docs.google.com/spreadsheets/d/1tpennYg1H5vGXK_hFTf8mfAjB2U-m3Lr8jf69PrTaJU/edit#gid=2042819910).

I suggest here to make 20 buckets with data evenly distributed across the buckets. To cater to newer projects, I've split the first bucket to be more granular.

Yes: this is saying that 5% of production traffic currently spends >8 hours at our API tier.

api/src/routing.js
```diff
@@ -111,6 +112,16 @@ app.postJsonOrCsv = (path, callback) => handleJsonOrCsvRequest('post', path, cal
 app.postJson = (path, callback) => handleJsonRequest('post', path, callback);
 app.putJson = (path, callback) => handleJsonRequest('put', path, callback);
 
+app.use(prometheusMiddleware({
+  metricsPath: '/api_prometheus_metrics',
```
kennsippell, line 116: not sure where to put this. somewhere under `/api/`?

api/src/routing.js
dianabarsan: I think this is a good starting point, especially to get some data and reassess later. 
Having >8h requests seems very strange, since our ALB timeout is set to 1h. These requests won't serve anyone. 

The other option I'm thinking is to have less buckets and think Gaussian curve, with ~6 buckets: 
image

kennsippell: I like the idea to fit our data to a distribution curve. [Sum of erlang distributions](https://newrelic.com/blog/best-practices/expected-distributions-website-response-times) would be a good option. Another [good read here](https://newrelic.com/blog/best-practices/performance-metrics-in-time-series-data) if we ever want to model this instead of basing off production data.

I suspect six buckets is too few. The default for the library is to have 9 buckets just monitoring values from 0 to 0.5 seconds.   We are interested in a range of response times which are 10000x larger so I suspect ~25 buckets is not inappropriate.  Part of me is actually inclined toward more buckets since the cost is not high and our understanding of this is still quite low.

api/src/routing.js
kennsippell: I have moved to `/api/vi/prometheus_metrics`.

Can you elaborate on the risk of exposing PHI? I want to make sure I'm clear about your concerns.

api/src/routing.js
kennsippell: Opened https://github.com/medic/cht-core/issues/8372 for requests > 8hr

api/src/routing.js
dianabarsan: Cool, you're correct, there no immediate concern for PHI being exposed through this API at this moment. However, we should still be vigilant whenever we update it.

api/src/routing.js
dianabarsan: I think we should go with your suggestion of 38 buckets. 
> our understanding of response durations is still quite low.

I agree.

api/src/routing.js
```diff
@@ -111,6 +112,19 @@ app.postJsonOrCsv = (path, callback) => handleJsonOrCsvRequest('post', path, cal
 app.postJson = (path, callback) => handleJsonRequest('post', path, callback);
 app.putJson = (path, callback) => handleJsonRequest('put', path, callback);
 
+app.use(prometheusMiddleware({
+  metricsPath: '/api/v1/prometheus_metrics',
```
garethbowen, line 116: Can we change the name to be more specific? The monitoring endpoint is already consumed by prometheus so it'd be good to be clear on the distinction. Also other endpoints are using hyphens not underscores. 

Is it possible to combine the two together or does this middleware require its own endpoint? If it must be distinct, then let's use something like `/api/v1/express-metrics` or `/api/v1/endpoint-performance`?

kennsippell: I like express-metrics. Will update

api/src/routing.js
kennsippell: Apdex monitoring [here](https://github.com/medic/cht-watchdog/pull/75#issue-1809492389) is one more factor to consider when setting these buckets. Currently using 90 and 180 for it.

api/src/routing.js
```diff
@@ -111,6 +112,18 @@ app.postJsonOrCsv = (path, callback) => handleJsonOrCsvRequest('post', path, cal
 app.postJson = (path, callback) => handleJsonRequest('post', path, callback);
 app.putJson = (path, callback) => handleJsonRequest('put', path, callback);
 
+app.use(prometheusMiddleware({
+  metricsPath: '/api/v1/express-metrics',
```
kennsippell, line 116: @jkuester There is an option for a [metricsPrefix](https://github.com/PayU/prometheus-api-metrics#options). Should I set it to `cht_`? `cht_api_`? Do you care?

jkuester: My vote is for `cht_api` as the best way to avoid collisions with names and keep everything straight.

api/src/routing.js
```diff
@@ -110,6 +111,19 @@ app.postJsonOrCsv = (path, callback) => handleJsonOrCsvRequest('post', path, cal
...
+  metricsPrefix: 'cht_api',
+  // based on one-month analysed period of production traffic
+  durationBuckets: [
+    0.004, 0.007, 0.013, 0.027, 0.05,
```
garethbowen, line 119: These buckets are weird at first glance. All the others double in size but these don't, even when you take rounding into account. Rounding to millis (3dp) yields:

```
    0.003, 0.006, 0.013, 0.025, 0.05,
```

Having said that, I'm not sure if there's much value in worrying about queries under 0.1s is there? Is there any downside to having such micro sized buckets?

garethbowen (approved):
Looks good to me.

One nitpick inline.

api/src/routing.js
kennsippell: logic for the buckets is here: https://github.com/medic/cht-core/pull/8354#discussion_r1248249460. i'm happy to make adjustments, but that was thinking. give that a read and then let me know if your thoughts change?  i'm fine to eliminate micro-buckets if that seems valuable

kennsippell:
> Has someone actually hooked this up to watchdog yet?

Yep over here: 
https://github.com/medic/cht-watchdog/pull/74
https://github.com/medic/cht-watchdog/pull/75

api/src/routing.js
garethbowen: I read all of that but it doesn't explain where the numbers in the "proposed buckets" column in the spreadsheet came from (they're not calculations, just numbers). Am I missing something?

kennsippell: ic. it sounds like you understand the "All Request Percentiles" column and the "Offline Replication Percentiles" column but not how they are combined.  that was kinda random and imprecise ... i interleaved and smoothed without keeping the work. 

so you likely aren't missing anything...  i'll change 0.027 to 0.025. if you have other suggestions can you propose as code changes?

~45% of requests are <0.1s so that is quite a big flattening

kennsippell:
```suggestion
    0.004, 0.007, 0.013, 0.025, 0.05,
```

dianabarsan (approved):
Smooth

api/src/routing.js
garethbowen: Ok that makes sense, and the code comment makes it clear. Thanks!

kennsippell:
Tested with local CHT, grafana, prometheus. API-side looks good.

mrjones-plip:
So awesome to see this merged! Nice work all.

Issue #8317, opened 2023-06-12
Split up the ci-webdriver-default action


garethbowen:
**Describe the issue**
Now that we've converted e2e tests from protractor to wdio the run takes about 50 minutes to complete. This is too long.

**Describe the improvement you'd like**
Break up the suite into smaller test suites that can run in parallel. These suites should be broken up into functional groups, eg: all the enketo tests, all the contacts tests, all the admin tests, etc.

It doesn't matter terribly much how many suites we end up with but there will be a sweet (no pun intended) spot - if broken down too far then the build will take a long time to complete because there's a limit on how many runners we can have in parallel and there is a one-off cost for starting the suite. Use your judgement about where this limit should be but somewhere between 4-10 suites should work.

**Describe alternatives you've considered**
Reducing test coverage (obviously bad).
Finding other ways to speed tests up (bigger effort).

PR #8380, opened 2023-07-09
chore(#8317): Split up the ci-webdriver-default action 

ngaruko:
# Description

Organise tests into suite and run them separately.

medic/cht-core#8317 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8317-split-upci-webdriver-default-action/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8317-split-upci-webdriver-default-action/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8317-split-upci-webdriver-default-action/docker-compose/cht-couchdb-clustered.yml)

# License


tests/e2e/default/wdio.conf.js
```diff
@@ -44,9 +43,44 @@ const baseConfig = {
...
+    ],
+    workflows: [
+      './analytics/**/*.wdio-spec.js',
+      './contacts/**/*.wdio-spec.js',
+      './reports/**/*.wdio-spec.js',
+      './targets/**/*.wdio-spec.js',
+      './tasks/**/*.wdio-spec.js',
+      './sms/**/*.wdio-spec.js',
+    ],
+    technical:[
```
dianabarsan, line 71: I wanted to suggest another name for this suite, because all suites are technical. 
I believe your choice of name wants to suggest that this is the low-level internal suite. 
What do you think about these classifications? "low-level" and "high-level" instead of `basic` + `passive` + `technical` ?

dianabarsan (changes requested):
Excellent stuff. I left some suggestions inline.

.github/workflows/build.yml
```diff
@@ -93,15 +93,27 @@ jobs:
 
   tests:
     needs: build
-    name: ${{ matrix.grunt-cmd }}
+    name: ${{ matrix.grunt-cmd }}_${{ matrix.suite ||'' }}
```
dianabarsan, line 96: Can you please change the separator from `_` for `-` (across all the PR)?
I'd like to have some consistency in the names of the jobs, because now they look like this: `ci-webdriver-default_basic`

.github/workflows/build.yml
```diff
@@ -93,15 +93,27 @@ jobs:
...
     runs-on: ubuntu-22.04
     env:
       NODE: 16.x
 
     strategy:
       fail-fast: false
       matrix:
-        grunt-cmd: ['ci-webdriver-default', 'ci-webdriver-standard', 'ci-e2e-integration', 'ci-webdriver-default-mobile']
+        grunt-cmd: ['ci-webdriver-standard', 'ci-e2e-integration', 'ci-webdriver-default-mobile']
+        suite: [all]
```
tatilepizs, line 105: Correct me if I am wrong, but this means that it will run all the e2e tests everything time that we create a build, and I think that the main idea of this change was to define a set of e2e tests that are "mandatory" or "critical" for each build to be created.

I think that we need to create a suite with a combination of tests that we want to be run every single time that a new build is created, for example:

- new-lineage-structure.wdio-spec
- search-contacts.wdio-spec
- pregnancy.wdio-spec
- login-logout.wdio-spec
- navigation.wdio-spec
- enabled-languages.wdio-spec
- add-user.wdio-spec
- ... any other test that the team thinks is valuable for each time that we create a build

tests/e2e/standard/wdio.conf.js
```diff
@@ -9,9 +9,11 @@ chai.use(require('chai-exclude'));
...
 // Override specific properties from wdio base config
 const standardConfig = Object.assign(wdioBaseConfig.config, {
-  specs: [
-    '**/*.wdio-spec.js'
-  ],
+  suites: {
+    all: [
+      './**/*.wdio-spec.js'
+    ]
+  },
```
tatilepizs, line 16: Maybe we can start defining suites also for the standard config, thinking that we probably will add more tests in the feature.

I think that the suites that we create for `default` should be also for `standard`

tatilepizs (changes requested):
Thank you for starting these changes Bede ⭐ 

I added some comments/questions.

.github/workflows/build.yml
dianabarsan: I think we want to run all tests for every build.

tatilepizs: Perfect, thank you!

webapp/tests/mocha/unit/testingtests/e2e/utils.spec.js
```diff
@@ -132,4 +134,24 @@ describe('Test utils', () => {
         });
     });
   });
+  
+  it('Check that all test specs belong to a test suites', () => {
+    const testFolders = [];
+    for (const [, value] of Object.entries(constants.SUITES)) {
```
dianabarsan, line 140: Is it necessary to pull out suites into `constants`? Is importing directly from `wdio.conf.js` not possible? 
If yes, I think it would be better to store then separately from `constants`, since it's only used by this config file. Generally we could consider breaking down `wdio.conf.js` to be more modular. It's a big file at the moment.

webapp/tests/mocha/unit/testingtests/e2e/utils.spec.js
```diff
@@ -132,4 +134,24 @@ describe('Test utils', () => {
...
+      value.forEach(path => testFolders.push(path.split('/')[1]));
+    }
+    const getDirectories =   (src, callback) => {
```
dianabarsan, line 143: We should promisify this: 
```suggestion
const getDirectories = (src) =>
      new Promise((resolve, reject) => glob(src, (err, res) => err ? reject(err) : resolve(res)));
```

webapp/tests/mocha/unit/testingtests/e2e/utils.spec.js
```diff
@@ -132,4 +134,24 @@ describe('Test utils', () => {
...
+      glob(src, callback);
+    };
+    //get all spec files in tests/e2e/default
+    getDirectories('../../../../../../tests/e2e/default/**/*.wdio-spec.js', (err, res)  => {
```
dianabarsan, line 147: This function doesn't return anything for me, I think the path doesn't get resolved properly. I think we have to resolve it:
```suggestion
    return getDirectories(path.join(__dirname, '../../../../../../tests/e2e/default/**/*.wdio-spec.js'))
```

webapp/tests/mocha/unit/testingtests/e2e/utils.spec.js
```diff
@@ -132,4 +134,24 @@ describe('Test utils', () => {
...
+      if (err) {
+        console.log('Error', err);
```
dianabarsan, line 149: The test should fail in this case. After you promisified then this test is no longer necessary.

dianabarsan (changes requested):
Nice! I left some comments inline.

webapp/tests/mocha/unit/testingtests/e2e/utils.spec.js
```diff
@@ -132,4 +135,21 @@ describe('Test utils', () => {
...
+      value.forEach(path => testFolders.push(path.split('/')[1]));
+    }
+    
+    const getDirectories = (src) => 
+      new Promise((resolve, reject) => glob(src, (err, res) => err ? reject(err) : resolve(res)));
+    return getDirectories(path.join(__dirname, '../../../../../../tests/e2e/default/**/*.wdio-spec.js'))
+      .then(res => {
+        res.forEach( spec => {
+          //assert that file belongs to a folder in 'default' directory
+          assert(testFolders.includes(spec.split('default/')[1].split('/')[0]), spec + ' does not belong to a folder' );
```
dianabarsan, line 151: I think this is pretty hard to read, from a single glance I'm not sure what output to expect and what we are comparing. It seems we are comparing folders, but that could also mean that we could skip a spec file. 
I would suggest we just create a list of all specs files and compare. I'm thinking something like: 
```
it('Check that all test specs belong to a test suites', async () => {
    const pathToDefaultTesting = path.resolve(__dirname, '../../../../../../tests/e2e/default');
    const suiteSpecs = [];
    const getDirectories = (src) =>
      new Promise((resolve, reject) => glob(src, (err, res) => err ? reject(err) : resolve(res)));

    for (const relativePaths of Object.values(wdiUtils.suites)) {
      for (const relativePath of relativePaths) {
        const resolvedPath = path.resolve(pathToDefaultTesting, relativePath);
        suiteSpecs.push(...await getDirectories(resolvedPath));
      }
    }

    const allSpecs = await getDirectories(path.join(pathToDefaultTesting, '/**/*.wdio-spec.js'));
    assert.sameMembers(allSpecs, suiteSpecs);
  });
```

tests/utils/wdio-utils.js
```diff
@@ -0,0 +1,38 @@
+const SUITES = {
```
dianabarsan, line 1: I'm not so sure about the location of this file, why is it in /utils when it's only useful for `default`. I think it would make more sense to rename this to `suites.js` or something and save it in the `/default` folder.

dianabarsan (changes requested):
This is so great! Just some readability suggestions for the test and a request to move the suites file to a more scoped location.

tests/e2e/default/wdio.conf.js
```diff
@@ -44,9 +43,11 @@ const baseConfig = {
...
   // will be called from there.
   //
-  specs: [
-    '**/*.wdio-spec.js',
-  ],
+  
+  suites: {
+    all: ['**/*.wdio-spec.js'],
+    ...wdiUtils.suites,
+  },
```
garethbowen, line 50: I think we should move this conf file up one level as it's the base config for all. Then we can make a new `default/wdio.conf.js` which extends the base and just defines the specs/suites.

Doesn't need to be done in this PR but felt like it fitted.

ngaruko: Good point @garethbowen . I have pushed it up

tests/e2e/standard/wdio.conf.js
ngaruko: Definitely, if other tests grow to be a kind of bottleneck as the default tests were becoming, we should think of splitting them too.

dianabarsan (approved):
I think this is very cool! 
Please let me know if you are ready to be merged. I'll merge it and update the branch protection rules afterwards.

garethbowen (approved):
Awesome!

The "lowlevel" suite is still too long to run IMO but this is a definite improvement - it'll be easy to split suites up further in a future PR. Thanks!

ngaruko:
@dianabarsan I think it is time to merge and update the branch protection rules

dianabarsan:
I've been restarting this one job that fails. I'd be really worried to force merge without all tests passing. I'll keep restarting until green and then merge and update branch protection rules.

.github/workflows/build.yml
```diff
@@ -93,15 +93,27 @@ jobs:
...
+        include:
+          - grunt-cmd: ci-webdriver-default
+            suite: core
+          - grunt-cmd: ci-webdriver-default
+            suite: data
+          - grunt-cmd: ci-webdriver-default
+            suite: data
```
dianabarsan, line 112: Hi @ngaruko 

Am I seeing this wrong or are we running the same suite twice, `data`, and never running `enketo`?

dianabarsan (changes requested):

.github/workflows/build.yml
ngaruko: Good catch @dianabarsan . Definitely some copy/paste gone wrong 😮

dianabarsan:
Hey @ngaruko 
I wanted to merge this, but it had a conflict, so I pushed a merge to resolve the conflict. I hope you don't mind.

dianabarsan (approved):
I think this improves ci builds by a lot! Not just that the builds take less time, but if one test is unstable, it won't take an hour to get one retry. Thanks for getting this done!

Issue #8287, opened 2023-05-30
Usage Analytics to understand how health workers are using the CHT


michaelkohn:
Initial goal is to understand how users (CHWs and Supervisors in particular) are interacting with the app today so that we can establish a baseline understanding of how certain actions are being completed. The goal of this initiative is not about identifying performance issues, errors, etc… like our [telemetry data provides](https://docs.communityhealthtoolkit.org/apps/guides/performance/telemetry/#performance-data). We want to understand how users are navigating while they are using the app, if/when they are using certain features.

_Here's the initial [Brainstorm doc](https://docs.google.com/document/d/1xlV2b-y3_9WLxkhiys5ItWq7AkHB4Y2R2WfPrLGRDso/edit#), [Related Forum Post](https://forum.communityhealthtoolkit.org/t/what-would-you-like-to-know-about-how-health-workers-use-the-cht/2775), and a review of [platforms](https://docs.google.com/spreadsheets/d/1EUvVCAGowNwmjGSc--J6jb1nPuo026QtV_PLjbzefpM/edit#gid=0)._

### Types of insights to uncover

1. What path did they take (ie what screens did they visit) to get to a specific screen/action?
2. How long did it take from start to finish?
3. How much time elapsed between each action?
4. What actions did they take on each screen (scrolling, tapping, etc…)?

### Types of questions we seek to answer

1. Are tasks being completed from the contacts tab or the tasks tab?
2. When completing a task from the tasks tab, where was the task in the list? Are CHWs coming here to find out what they should be doing or simply coming here to complete something they knew they had to do?
3. What are they doing after they complete a task?
4. How many key actions are happening within a single session?
5. Which actions are being “abandoned” often (ie opening a screen but not completing any actions)?
6. Are users accessing the targets tab throughout the month or only during meetings/at the end of the month? 
7. Do users take a direct route to specific pages or is there a lot of back and forth? How long do they spend looking for information on the screen? What are their scroll activities?
8. Do paths differ based on the type of form or task?
9. What % of time is spent on each tab?
10. Where and how often are [rage clicks](https://www.fullstory.com/blog/rage-clicks/) happening?
11. How many times per day/week/month are users even opening the app?
12. Which validation errors are they seeing most often and when?
13. How often CHWs edit reports/contacts and which are the most edited fields.
14. Etc...

### Other considerations

1. Segmentation of users
2. Who will be using this data
3. How will it be accessed
4. Turning recording on/off
[...]

PR #8393, opened 2023-07-13
feat(#8287): matomo angular integration

latin-panda:
# Description

**Do not merge!**

This pull request intends to visualize the code we need to support Matomo in CHT, and _this is not the final work_. 
The team will look for partners using 4.x that want to try this feature. Then we will make a feature release branch based on the partner's project version. 

The feature can be enabled with `can_track_usage_analytics` permission and can be configured by adding the following in `app_settings.js`
```
"usage_analytics": {
    "server_url": "<intent analytics server URL>",
    "site_id": "<insert site id>",
    "site_sha": "<insert sha>"
}
```

https://github.com/medic/cht-core/issues/8287

**Data usage**
* Matomo default features for free self-hosted, with the following premium modules: User Flows; and Heatmap and session recording.  
    * 9 min usage - 21.95 MiB (23019010 bytes)
    * 20 min usage - 35.53 MiB (37257708 bytes)
* Matomo default features for free self-hosted, with the following premium modules: User Flows only.
    * 9 minutes: 167.81 KiB (171836 bytes)
    * 20 min: 203.87 KiB (208758 bytes)

**Supports**
* PWA
* CHT-Android
* Firefox (desktop and mobile)
* Chrome (desktop and mobile)

The recommendation is to use PWA or CHT-Android because Matomo doesn't work well in the following cases:
* Incognito mode
* Add blocker is enabled
* Cookies or javascript is disabled
* (Chrome) signed-in user in the browser has settings in its account that can affect Matomo functionality. 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:matomo_angular_integration/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:matomo_angular_integration/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:matomo_angular_integration/docker-compose/cht-couchdb-clustered.yml)

# License


api/src/routing.js
```diff
@@ -149,52 +150,11 @@ app.use(
...
-        ],
-      },
-      browserSniff: false,
-    },
-  })
-);
+app.use((req, res, next) => {
+  contentSecurityPolicy
+    .get(req, res)
+    .then(policy => helmet(policy)(req, res, next));
```
latin-panda, line 156: I didn't find a way to mock `helmet`, so using `helmet` here instead of in contentSecurityPolicy file to cover the CSP config with unit tests

tests/e2e/default/login/login-logout.wdio-spec.js
```diff
@@ -81,8 +81,6 @@ describe('Login page funcionality tests', () => {
       await (await commonPage.analyticsTab()).waitForDisplayed();
 
       const cookies = await browser.getCookies();
-      expect(cookies.length).to.equal(3);
```
latin-panda, line 84: Matomo adds cookies depending on the configuration in the Matomo server.

webapp/src/ts/services/usage-analytics.service.ts
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  private replaceUuid(url) {
+    if (!this.UUID_REGEX.test(url)) {
+      return url;
+    }
+
+    const newURL = url.replace(this.UUID_REGEX, 'UUID');
+    const navigation = this.router.getCurrentNavigation();
+    const usageAnalyticsValue = navigation?.extras?.state?.usageAnalyticsValue;
```
latin-panda, line 130: We don't need the UUIDs, instead to understand the reports, we need the contact_type and the form name

latin-panda:
@garethbowen This is what it takes to use Matomo in CHT. The most significant change is in the CSP since Matomo's project SHA can change because of an update in the Matomo config. The CSP needs to handle values coming from somewhere that App Builders can work easily; for now, this is in app-settings.json

If that's considered an improvement to how CSP is done today, I can submit that change in a separate PR, because IMO this POC will take a long time to merge.

garethbowen:
@latin-panda I'm a bit worried that it'll be too easy to forget to update the sha when the config is changed. Is there another way to do this so the sha is updated automatically? Are we self hosting Matomo? If so, can we cache the matomo config and serve it from API to make CSP easier?

latin-panda:
>Are we self-hosting Matomo?

@garethbowen Yes, we are

>Is there another way to do this so the sha is updated automatically? 

As of today, I am still looking for a way. They intend that the system admins use their UI to configure Matomo. Some things can be configured in JavaScript (webapp only), mainly related to tracking data and connecting with Matomo, which we already do in the `usage-analytics.service.ts` Their segmentation is sites like stores or blogs, and so are their assumptions on how each site is managed independently. 

In our experience using the free-tier package, every time we changed the site's ID or the Matomo server's URL (from cloud to self-hosted), the SHA changed, and then we had to update the CSP because it was blocking the script (worked by Craig, Michael and me 😅). This still makes sense to me; these two don't often change over time.

Last week, we got a discount code and used the Admin UI to enable some premium features (heatmaps, session recording, user flows), which can't be done programmatically; we changed how Matomo tracks and handles URLs to avoid cropping information. Not sure if any of that influenced it, but after testing again today, I don't see the CSP blocking the Matomo script anymore because of a missing SHA.

I made the SHA setting optional in the CSP, and I'll upload the changes to our test instance and monitor.

garethbowen (commented):
Comment provided...

latin-panda:
Closing for now, we can reopen the PR later if we need

Issue #8396, opened 2023-07-14
Next page not loading properly after clicking on the back button


latin-panda:
**Describe the bug**
The next page not loading properly after clicking on the back button. [This line](https://github.com/medic/cht-core/blob/47c72bba9ce3efc4115f4ce22cc39c3a5a118e4f/webapp/src/ts/modules/reports/reports.component.ts#L336) is preventing from retrieving the list of reports when the selected report is loaded.

https://github.com/medic/cht-core/assets/66472237/4b5e8980-53ef-48ba-b8e5-dbc6de671965

**To Reproduce**
1. Load mobile view
2. Open contact with reports
3. Click on one report
4. Click on the back button from the title bar (not the browser back button)
5. See how the Report tab is stuck on loading 

**Expected behavior**
Should load list of reports 

**Environment**
- Instance: any
- Browser: Chrome
- Client platform: MacOS
- App: webapp
- Version: I checked 3.17 to current (4.2.x)

latin-panda:
For the record, doing the same but with tasks, it's working fine.

PR #8400, opened 2023-07-20
fix(#8396): fixes back navigation in reports

latin-panda:
# Description

Fixes back navigation in reports by always loading the list of reports

medic/cht-core#8396

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8396-fixing-nav-in-reports/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8396-fixing-nav-in-reports/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8396-fixing-nav-in-reports/docker-compose/cht-couchdb-clustered.yml)

# License


webapp/src/ts/modules/reports/reports.component.ts
```diff
@@ -324,20 +324,16 @@ export class ReportsComponent implements OnInit, AfterViewInit, OnDestroy {
     this.scrollLoaderProvider.init(scrollCallback);
   }
 
-  search(force = false) {
+  search() {
```
latin-panda, line 327: The way out from the reports content page when on mobile is to navigate to the reports list; I change the function to load the list just like [Contacts do](https://github.com/medic/cht-core/blob/47c72bba9ce3efc4115f4ce22cc39c3a5a118e4f/webapp/src/ts/modules/contacts/contacts.component.ts#L408)

latin-panda:
Hi @dianabarsan can you please have a look when you have time?

webapp/src/ts/components/filters/freetext-filter/freetext-filter.component.ts
```diff
@@ -45,9 +45,7 @@ export class FreetextFilterComponent implements OnDestroy, OnInit, AbstractFilte
     }
 
     this.globalActions.setFilter({ search: this.inputText });
-    // always force the search, so the user is taken from the report detail page to the list page on mobile,
-    // when clicking on a case_id link
-    this.search.emit(true);
```
dianabarsan, line 50: I checked this interaction, clicking on a report with a `case_id` and it's broken right now. Removing this further breaks that interaction.

dianabarsan (changes requested):
Hi @latin-panda 

I checked out your branch and checked the interaction. However, I'm worried that we're further regressing on an issue: https://github.com/medic/cht-core/issues/8401
I think we should take this regression into account when fixing the search.

webapp/src/ts/components/filters/freetext-filter/freetext-filter.component.ts
latin-panda: Since it's two different problems and I'm going to work on them in separate PRs. Once the other one is merged, I'll update this PR.

dianabarsan: Awesome, thanks a lot!

latin-panda: I've sent a fix for [that issue](https://github.com/medic/cht-core/pull/8403), which doesn't conflict with this other solution so far.

Issue #8401, opened 2023-07-20
Clicking on report `case_id` does not apply search filter


dianabarsan:
**Describe the bug**
When a report has a `case_id` field, this field is highlighted as a link in the report details page, and the link is meant to apply a filter that shows all other reports that have the same `case_id`. The interaction should be described here: https://github.com/medic/cht-core/issues/6291

This interaction is broken, clicking on the `case_id` does nothing. 

**To Reproduce**
1. Add a `case_id` field to a report. 
2. Open the report detail page. 
3. Notice that the `case_id` field is a link. Click the link. 
4. Nothing happens. 

**Expected behavior**
The reports list should be filtered to only display reports that have the same case_id. 

**Logs**
No errors reported. 

**Environment**
- App: webapp
- Version: master, suspecting regressions happened when filters were reworked.

latin-panda:
FYI @ngaruko

dianabarsan:
@latin-panda would you mind adding the appropriate Affects-* tags? I haven't looked closely at what point in development this was introduced.

PR #8403, opened 2023-07-21
fix(#8401): search filter by case_id

latin-panda:
# Description

Adds support to search by case_id

medic/cht-core#8401

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8401_case_id_search_filter/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8401_case_id_search_filter/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8401_case_id_search_filter/docker-compose/cht-couchdb-clustered.yml)

# License


webapp/src/ts/components/search-bar/search-bar.component.ts
```diff
@@ -31,12 +41,17 @@ export class SearchBarComponent implements AfterContentInit, OnDestroy {
...
     private responsiveService: ResponsiveService,
+    private searchFiltersService: SearchFiltersService,
   ) { }
 
   ngAfterContentInit() {
     this.subscribeToStore();
   }
 
+  ngAfterViewInit() {
+    this.searchFiltersService.init(this.freetextFilter);
```
latin-panda, line 52: The element is ready in the DOM in `ngAfterViewInit` and not before.
While we should subscribe to the store after the component is inited (`ngAfterContentInit`), this avoids the Angular's `expression has changed exception` that we resolved a while ago as part of another ticket.

latin-panda:
@dianabarsan @ngaruko Please review, thanks!

ngaruko (approved):
Looks good. Thanks for updating the e2e tests

tests/e2e/default/reports/search-reports.wdio-spec.js
```diff
@@ -1,114 +1,73 @@
...
-    await commonPage.goToReports();
   });
 
-  it('search by NON empty string should display results with contains match and then clears', async () => {
-    // Waiting for initial load
-    await reportsPage.getAllReportsText();
+  it('should return results matching the search term and then return all data when clearing search', async () => {
+    await commonPage.goToReports();
+    // Asserting first load reports
+    expect((await reportsPage.reportsListDetails()).length).to.equal(4);
```
dianabarsan, line 45:
```suggestion
expect((await reportsPage.reportsListDetails()).length).to.equal(reportDocs.length);
```

tests/e2e/default/reports/search-reports.wdio-spec.js
```diff
@@ -1,114 +1,73 @@
...
 
-    // Searching by keyword
     await searchPage.performSearch('sittu');
-    let allLHSContacts = await reportsPage.getAllReportsText();
-    expect(allLHSContacts.sort()).to.deep.equal([sittuPerson.name]);
+    await commonPage.waitForLoaders();
+    expect((await reportsPage.reportsListDetails()).length).to.equal(2);
+    expect(await (await reportsPage.reportByUUID(reportDocs[0].id)).isDisplayed()).to.be.true;
```
dianabarsan, line 50: I think we should denominate these reports somehow, instead of referring to them by index. 
Something like: 
```
const [smsSittu, smsPottu, pregnancySittu, pregnancyPottu] = await utils.saveDocs(reports);
...
expect(await (await reportsPage.reportByUUID(smsSittu.id)).isDisplayed()).to.be.true;
```

tests/e2e/default/reports/search-reports.wdio-spec.js
```diff
@@ -1,114 +1,73 @@
...
+    await commonPage.goToReports();
+    await reportsPage.openReport(reportDocs[2].id);
+    // Asserting first load reports
+    expect((await reportsPage.reportsListDetails()).length).to.equal(4);
+
+    await reportsPage.clickOnCaseId();
+    await commonPage.waitForLoaders();
+    expect((await reportsPage.reportsListDetails()).length).to.equal(2);
+    expect(await (await reportsPage.reportByUUID(reportDocs[2].id)).isDisplayed()).to.be.true;
+    expect(await (await reportsPage.reportByUUID(reportDocs[3].id)).isDisplayed()).to.be.true;
```
dianabarsan, line 71: I think we should add a test for mobile as well, because we expect the user to be taken from the report detail page to the report list when clicking on the `case_id` field on mobile.

tests/page-objects/default/reports/reports.wdio.page.js
```diff
@@ -4,22 +4,23 @@ const utils = require('@utils');
 
 const REPORTS_LIST_ID = '#reports-list';
 const SELECT_ALL_CHECKBOX = `${REPORTS_LIST_ID} .select-all input[type="checkbox"]`;
-const reportBodyDetailsSelector = '#reports-content .report-body .details';
-const reportBodyDetails = () => $(reportBodyDetailsSelector);
-const reportTasks = () =>  $(`${reportBodyDetailsSelector} .scheduled-tasks`);
+const REPORT_BODY_DETAILS_SELECTOR = '#reports-content .report-body .details';
+const reportBodyDetails = () => $(REPORT_BODY_DETAILS_SELECTOR);
+const reportTasks = () =>  $(`${REPORT_BODY_DETAILS_SELECTOR} .scheduled-tasks`);
+const reportTargetFilter = () =>  $(`${REPORT_BODY_DETAILS_SELECTOR} [test-id="field-target-filter"]`);
```
dianabarsan, line 10: Are we sure there is always just one target filter?

webapp/src/ts/modules/reports/reports-content.component.ts
```diff
@@ -172,6 +172,9 @@ export class ReportsContentComponent implements OnInit, OnDestroy {
   }
 
   search(query) {
+    if (this.selectMode) {
```
dianabarsan, line 175: This is kinda weird in mobile. 
There is no visual difference between being in select mode or not when seeing report detail, but if I click on the case_id while in select mode, then the link doesn't do anything. 
I think we should be consistent somehow?

dianabarsan (changes requested):
Nice work, thanks so much for picking this. I left a few minor comments inline and some questions.

webapp/src/ts/modules/reports/reports-content.component.ts
latin-panda: I brought up the topic to last week's Care Teams meeting. We're going to review the experience when everyone from the trio is back from holidays. There's a [ticket](https://github.com/medic/cht-core/issues/8419) for tracking. 
For now, we're just going to make the filter work, i.e. fix the bug. We'll fix the experience after evaluation.
For now, if the select mode is on, the link will be gray. 
Screenshot 2023-07-31 at 11 23 28 am

tests/page-objects/default/reports/reports.wdio.page.js
latin-panda: using field label

tests/e2e/default/reports/search-reports.wdio-spec.js
```diff
@@ -1,114 +1,76 @@
...
+    expect((await reportsPage.reportsListDetails()).length).to.equal(reportDocs.length);
+    expect(await (await reportsPage.reportByUUID(sittuSMSPregnancy.id)).isDisplayed()).to.be.true;
+    expect(await (await reportsPage.reportByUUID(potuSMSPregnancy.id)).isDisplayed()).to.be.true;
+    expect(await (await reportsPage.reportByUUID(sittuPregnancy.id)).isDisplayed()).to.be.true;
+    expect(await (await reportsPage.reportByUUID(potuPregnancy.id)).isDisplayed()).to.be.true;
+  });
+
+  it('should return results when searching by case_id', async () => {
+    const sittuPregnancy = reportDocs[2];
+    const potuPregnancy = reportDocs[3];
```
dianabarsan, line 64:
```suggestion
const [, sittuPregnancy, potuPregnancy] = reportDocs;
```

tests/page-objects/default-mobile/search/search.wdio.page.js
```diff
@@ -0,0 +1,14 @@
...
+const openSearchBox = () => $('.mm-search-bar-container .search-bar-left-icon .fa-search');
+
+const performSearch = async (term) => {
+  await (await openSearchBox()).waitForClickable();
+  await (await openSearchBox()).click();
+  await searchPageDefault.performSearch(term);
+};
+
+module.exports = {
+  searchPageDefault,
```
dianabarsan, line 12: I wonder if this separation is necessary. Why not just do:
```suggestion
...searchPageDefault
```

dianabarsan (approved):
I have two small nits inline. Looking great, thanks for this!

tests/page-objects/default-mobile/search/search.wdio.page.js
latin-panda: Yes, I was following the style from this other [page object](https://github.com/medic/cht-core/blob/b98c1ff90370c9257639becb9ee3668d6415957b/tests/page-objects/default-mobile/reports/reports.wdio.page.js#L74). I remember that QA was thinking on restructuring the mobile e2e, I prefer to follow the style and wait to see how they will restructure page objects, before introducing a new way of doing this.
Since there aren't that many, it can be done in a different PR. 

cc: @tatilepizs @ngaruko

tests/e2e/default/reports/search-reports.wdio-spec.js
latin-panda: I thought about this, but I don't like that much to have "empties". In this case, we need two: 
`const [ , , sittuPregnancy, potuPregnancy ]`
Currently, it's equally as readable, just more lines. But if you feel strongly about this, I can change it :)

Issue #8306, opened 2023-06-08
Replace grunt with npm scripts


garethbowen:
**Describe the issue**
Grunt is effectively unmaintained which means we get stuck on legacy versions of build dependencies which will eventually bite us. It and the ecosystem also has some bugs or performance issues we have to work around.

**Describe the improvement you'd like**
Convert grunt steps into npm scripts with JS utilities to execute commands directly.

**Describe alternatives you've considered**
Instead of npm scripts there are other build options but they don't provide much in the way of benefits over the pure JS and may also be deprecated in the future.

garethbowen:
Replacing grunt-contrib-watch is not trivial. The native solution in Node is lacking essential features until Node 20 (not yet LTS). the grunt lib uses [gaze](https://github.com/shama/gaze) which looks good but hasn't been updated in a while. Other options are [chokidar](https://github.com/paulmillr/chokidar), [watch](https://github.com/mikeal/watch), or [watchr](https://github.com/bevry/watchr).

fardarter:
Have you considered adopting something like esbuild incrementally? https://esbuild.github.io/api/#build

Plugin format is very vanilla js: https://bilalbudhani.com/chokidar-esbuild/

Vite uses it under the hood and that's going to have a long tail of support based on trends: https://vitejs.dev/guide/why.html

garethbowen:
esbuild for Angular is apparently [still experimental](https://angular.io/guide/esbuild) and we're not quite on a compatible version of Angular yet anyway, but it looks like a [nice win](https://dev.to/danywalls/maximizing-your-angular-build-performance-with-esbuild-1ea9) when it's ready.

However we use grunt and grunt-contrib-watch for more than just the webapp - for example, if you make a change to a map/reduce view we build a new ddoc and upload it to your couchdb using a custom watcher. So we still need to solve for the problem outside of webapp.

I have a proof of concept going which uses Gaze directly which should be fine for now (most important is getting rid of grunt). We can iteratively improve in another round!  :)

garethbowen:
Done.

PR #8408, opened 2023-07-26
chore(#8306): remove grunt-contrib-watch

garethbowen:
# Description

Removes grunt-contrib-watch
Also replaces our custom patch scripts with `patch-package` npm module

medic/cht-core#8306

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8306-remove-grunt-6/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8306-remove-grunt-6/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8306-remove-grunt-6/docker-compose/cht-couchdb-clustered.yml)

# License


scripts/build/watch.js
```diff
@@ -0,0 +1,92 @@
...
+
+  // api-public-files
+  {
+    files: [ 'api/src/public/**/*' ],
+    task: 'exec:copy-api-resources',
+  },
+
+  // ddocs
+  {
+    files: [ 'ddocs/*-db/**/*' ],
```
dianabarsan, line 45: I'm changing a line in one secondary ddoc and I'm getting this error:
```
/home/diana/projects/medic-second/ddocs/logs-db/logs/views/connected_users/map.js updated...
Running "exec:copy-ddocs" (exec) task
Running "set-ddocs-version" task
Running "set-build-info" task
Warning: ENOTDIR: not a directory, scandir '/home/diana/projects/medic-second/build/ddocs/logs.json' Use --force to continue.
Aborted due to warnings.
```

scripts/build/watch.js
```diff
@@ -0,0 +1,92 @@
...
+  // webapp-js
+  // instead of watching the source files which are watched separately, watch the build folder and upload on rebuild
+  {
+    files: [ 'api/build/static/webapp/**/*', '!api/build/static/webapp/service-worker.js' ],
+    task: 'update-service-worker',
+  },
+
+  // api-public-files
+  {
+    files: [ 'api/src/public/**/*' ],
```
dianabarsan, line 39: I'm making a single change in a file in this folder, which ends up copying all files, and then subsequently triggers another task 4 times. Is there a way we can debounce them? 
```
/home/diana/projects/medic-second/api/src/public/privacy-policy/script.js updated...
Running "exec:copy-api-resources" (exec) task
Done.

/home/diana/projects/medic-second/api/build/static/webapp/robots.txt updated...
/home/diana/projects/medic-second/api/build/static/webapp/manifest.json updated...
/home/diana/projects/medic-second/api/build/static/webapp/manifest.appcache updated...
/home/diana/projects/medic-second/api/build/static/webapp/appcache-upgrade.html updated...
Running "update-service-worker" task
Running "update-service-worker" task
Running "update-service-worker" task
Running "update-service-worker" task
Done.
Done.
Done.
Done.
```

dianabarsan (changes requested):
Great progress! 
I have some questions inline, don't know how achievable.

webapp/package.json
```diff
@@ -16,6 +16,9 @@
     "node": ">=16.12.0",
     "npm": ">=8.3.1"
   },
+  "scripts": {
+    "postinstall": "patch-package"
```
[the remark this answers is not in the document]
garethbowen: How about making sure the comment is in the patch itself, for example this patch has an example (which is out of date but you get the idea): https://github.com/medic/cht-core/pull/8408/files#diff-35348b3d34728f9deb1deedb464fecb1d843f263b735aaf45fafa45f9e22e723

scripts/build/watch.js
garethbowen: Yes there is... but it looks like the default is 500 so it's not working: https://github.com/shama/gaze/blob/19b87085606e80e02e5ca05ea942cf8a34732b59/lib/gaze.js#L37

I'll dig in to this and see if I can work out why...

garethbowen: I think we'd need this PR: https://github.com/shama/gaze/pull/258/files

garethbowen: Ultimately the problem is... you modify 1 file, which calls a brute force grunt task (yet to be converted) which copies all files from one folder to another (even though you only modified one). This then triggers the watcher for each file copied (I'm not sure why it only fires 4 times but maybe it maxes out somewhere?) which builds the service worker each time.

Options...

1. Use the patch linked above. This introduces a lag (but we could make it like 10ms or something tiny), but means we're now on a fork of gaze which sucks (even though it's not being updated anyway...).
2. Add the equivalent delay in our watch.js which means we're main of gaze.
3. Make the copy commands smarter - when the gaze event fires it tells you which file was updated. If we copy just this file then updating one file will only copy one file and only trigger one "update-service-worker" task. This is good because there's no point copying files that aren't updated anyway, but it's a lot more maintenance because we need to maintain the brute force version (for fresh builds) as well as the targeted version (for incremental builds).

I'm leaning towards option 2 because then I make the change in one place and everything just works. What do you think?

scripts/build/watch.js
garethbowen: Ok... the problem here was a slight difference in the grunt tasks. Previously the clean build called `set-build-info` but the watch build did not. This was the only difference between the two and I must have for simplicity to make both the same. Unfortunately if the build dir isn't clean then the `json` ddocs are sitting there and our code was trying to readDir which failed.

I couldn't see a clean way to have a single task handle both cases, and still don't like the idea of having two nearly identical tasks, so I've modified the `set-build-info` to ignore files.

Let me know what you think!

scripts/build/watch.js
dianabarsan: I think option 2 sounds good. it's cheapest to try out anyway :)

webapp/package.json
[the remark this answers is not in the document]
dianabarsan: Having the comment in the file works really well, thank you!

scripts/build/watch.js
dianabarsan: Cool, I see you added a check before doing scan, this works well. I agree with not wanting two nearly identical tasks.

webapp/package.json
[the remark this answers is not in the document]
garethbowen: Done.

scripts/build/watch.js
garethbowen: Done.

garethbowen:
@dianabarsan Review again please. The build needs a re-run but I'm off to bed  :)

Issue #8238, opened 2023-05-12
Figure out a different way to use purged docs cache


dianabarsan:
**What feature do you want to improve?**
Right now, purged docs are cached for a very short interval (5 minutes) - for considerations of memory management. 
Recent research has shown that: 
- users end up having a very large number of documents before purging - with 40k being common, growing up to 100k docs
- cache is used in 40% of requests

**Describe the improvement you'd like**
Experiment with increasing TTL, and only bust cache when purging has run again. 

**Describe alternatives you've considered**
If experiments show that the memory footprint of the cache for a large project (suppose 4000 users each with 60k pre-purge docs, cached), investigate whether cache can be saved on disk.

**Additional context**
@kennsippell has been analyzing production logs and observing request behavior. Querying for purged docs in this project, which has  very aggressive purging and high number of pre-purge docs, the purging requests take up to 40% of server time.

garethbowen:
Bumping to 4.4.0 as we're almost ready to release 4.3.0 and this has more work needed yet.

dianabarsan:
Merged to `master`

PR #8410, opened 2023-07-26
feat(#8238): store purged docs cache in couchdb

dianabarsan:
# Description

- each user gets their own doc in a new database `medic-cache` to store purged document ids
- the database is destroyed (deleted and recreated) every time purging runs
- the cache lifetime is infinite

medic/cht-core#8238

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8238-purged-docs-cache/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8238-purged-docs-cache/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8238-purged-docs-cache/docker-compose/cht-couchdb-clustered.yml)

# License


api/src/db.js
```diff
@@ -97,6 +102,11 @@ if (UNIT_TEST_ENV) {
     }
   };
 
+  module.exports.wipeCacheDb = async () => {
+    await module.exports.cache.destroy();
```
dianabarsan, line 106: Tested this locally, and it deletes the database and also frees us used disk space.

dianabarsan:
I believe this is ready for a first round of reviewing. Would you mind having a look, @garethbowen ? Thanks a lot for your time!

api/src/services/db-watcher.js
```diff
@@ -0,0 +1,68 @@
...
+  watchSentinelChanges();
+  watchUsersChanges();
+};
+
+module.exports = {
+  listen,
+  medic: subscribeToMedic,
+  sentinel: subscribeToSentinel,
+  users: subscribeToUsers,
+};
```
garethbowen, line 68: Let's reduce the duplication in this file - with an array of dbs to watch we can loop to fill in the exports, the listen function, and the watch*changes functions.

api/src/services/purged-docs.js
```diff
@@ -1,13 +1,14 @@
 const db = require('../db');
+const dbWatcher = require('./db-watcher');
 const environment = require('../environment');
 const purgingUtils = require('@medic/purging-utils');
-const cacheService = require('./cache');
```
garethbowen, line 4: I think this is the only use of the cacheService so we can drop it and the node-cache package.

api/src/services/purged-docs.js
```diff
@@ -51,59 +65,109 @@ const getPurgedIdsFromChanges = result => {
...
+};
+
+const watchMedic = () => {
+  dbWatcher.medic(({ id }) => {
+    if (id.startsWith(USER_DOC_PREFIX)) {
+      const name = id.replace(USER_DOC_PREFIX, '');
+      return clearCache(name);
+    }
+  });
+};
```
garethbowen, line 155: Why do we need this as well as watching _users?

api/src/services/purged-docs.js
```diff
@@ -51,59 +65,109 @@ const getPurgedIdsFromChanges = result => {
...
+const watchSettings = () => {
+  const key = 'district_admins_access_unallocated_messages';
+  let district_admins_access_unallocated_messages = config.get(key);
+  configWatcher.watch(() => {
+    const newConfigValue = config.get(key);
+    if (newConfigValue !== district_admins_access_unallocated_messages) {
+      district_admins_access_unallocated_messages = newConfigValue;
+      clearCache();
+    }
+  });
```
garethbowen, line 166: This feels like something the service should provide, like...

```
configWatcher.watch('district_admins_access_unallocated_messages', () => {...});
```

api/src/services/purged-docs.js
```diff
@@ -51,59 +65,109 @@ const getPurgedIdsFromChanges = result => {
...
+const getCachedIds = async (cacheDocId) => {
+  const cachedDoc = await getPurgeCacheDoc(cacheDocId);
+  return cachedDoc?.doc_ids;
+};
+
+const setCachedIds = async (cacheDocId, ids) => {
+  ids.sort();
+  const cachedDoc = await getPurgeCacheDoc(cacheDocId) || { _id: cacheDocId };
+  cachedDoc.doc_ids = ids;
+  await db.cache.put(cachedDoc);
```
garethbowen, line 85: Might need to catch 409s.

garethbowen (changes requested):

api/src/services/db-watcher.js
```diff
@@ -0,0 +1,68 @@
...
+
+const watchMedicChanges = () => {
+  db.medic
+    .changes(OPTS)
+    .on('change', (change) => {
+      medicEmitter.emit('change', change);
+    })
+    .on('error', (err) => {
+      logger.error('Error watching medic changes, restarting: %o', err);
+      process.exit(1);
```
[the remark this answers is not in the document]
dianabarsan: This is how the old listener was structured. It had no `complete` event and rage quit on errors. 
I doubt `complete` ever fires on `continuous`.

api/src/services/purged-docs.js
dianabarsan: Cool, I considered dropping the package too.

api/src/services/purged-docs.js
dianabarsan: Because we store `contact_id` in the `user-settings` document and `facility_id` in the `user`  document :( 
https://github.com/medic/cht-core/blob/a686679a6a4d197c40c0036c25733b4ff96716aa/shared-libs/user-management/src/users.js#L761

api/src/services/purged-docs.js
dianabarsan: This is a cool idea! But it would add some complexity - each time settings are updated, we'd need to iterate over all properties and emit for every change. Do you think it's worth to add this now?

api/src/services/db-watcher.js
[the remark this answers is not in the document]
dianabarsan: I don't recall ever seeing a timeout from this continuous changes request. I think the parameters that are sent to CouchDb will prevent the timeout: 
```
if (opts.continuous && !('heartbeat' in opts)) {
  opts.heartbeat = DEFAULT_HEARTBEAT;
}
```

So if there is indeed a timeout, it means CouchDb is in real trouble.

api/src/db.js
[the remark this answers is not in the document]
dianabarsan: Uh ... not paranoid. I'll implement the cache service that can queue things.

api/src/services/db-watcher.js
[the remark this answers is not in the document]
garethbowen: It would be possible in the future when we have replicas for one couch node to go down but the cluster itself still be available. I assume if one node goes down and it happens to be the one serving this request, then the request will drop but API shouldn't crash.

As this was already an issue, feel free to add this to the replicas issue and we can solve it later.

api/src/services/purged-docs.js
garethbowen: Ahh ok! Thanks.

api/src/services/purged-docs.js
garethbowen: Happy to leave it for now.

garethbowen (changes requested):
Great progress, thanks!

api/src/services/purged-docs-cache.js
```diff
@@ -0,0 +1,37 @@
...
+  destroyPromise = purgeDb.destroy();
+  await destroyPromise;
+  purgeDb = null;
+  destroyPromise = null;
+};
+
+module.exports = {
+  get,
+  wipe,
+};
```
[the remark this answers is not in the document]
dianabarsan: I like this very much! Done.

garethbowen (approved):
This looks amazing. Thanks!

PR #8415, opened 2023-07-29
chore: enable strictNullChecks for webapp

jkuester:
# Description

Basically all this PR does is set the [`strictNullChecks` TS config option](https://www.typescriptlang.org/tsconfig#strictNullChecks) in our webapp TS config (and also updates the types in the existing TS code to allow the app to still build...).

(Note that `strictNullChecks` is a _subset_ of what you get by enabling the [`strict` flag](https://www.typescriptlang.org/tsconfig#strict). So, this is a good step in that direction.)

Implementation code changes have been kept to an absolute minimum. Besides updates to _typing_ (which technically have no effect on the actual behavior of the code) there were a few places where I had to change a value from `undefined` to `null` (or vice versa) to remain compatible with inherited type definitions.  Other than these changes, the actual logic of the implementation code should be completely unchanged!

## Benefits of `strictNullChecks`

- Prevents the need for unnecessary and redundant null-checking throughout the code since types can be used to guarantee something is valued
- TS can indicate where optional chaining (or other null checking) is required since it knows what _could be_ `null`

## Methodology for this PR

- The majority of the changes involved code that was creating empty Array literals (`[]`) that were populated later.  With `strictNullChecks`, TS was interpreting the type of those arrays as `never[]` if they were created with no type specified.  The fix here is just to specify the type of thing the array will hold when instantiating the variable.
- Another common problem was when default values were provided for function parameters. Often if `null` was provided (and no type was specified for the parameter, then TS was concluding that the _type_ for those variables was just `null`.  The fix here was just to specify the type for parameters that had default values.
- I have put a modest amount of effort into using _more specific_ types and not just slapping `any` on everything.  Where I was absolutely certain of a type (typically local variables), I used that specific type.  Otherwise, I did not dig too deep. If I did not know the exact type, but I did know that it was an _actual object_ (and not just primitive/null/undefined), I would use `Record<string, any>`.
- The main goal was to change as little as possible while still being able to successfully build the project. This is why I have not tried to consistently apply types across the changed files or anything. I have only updated the minimum places necessary to allow for building the project.
- There are a ton of places where minor logic changes could have "improved" the typing for a particular function/variable. I have resisted the urge to make these logic changes here and instead focused on updating the types to reflect _the current state of the code_.
    - The result of this is some objectively bad types being used (e.g. `object | null`) but I think it is preferable to leave these bad types as beacons for future refactors instead of trying to fix everything now and run the risk of causing a subtle defect that is missed because of the scope of these changes...

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:strict_null_checks/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:strict_null_checks/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:strict_null_checks/docker-compose/cht-couchdb-clustered.yml)

# License


webapp/src/ts/components/filters/date-filter/date-filter.component.ts
```diff
@@ -93,7 +93,7 @@ export class DateFilterComponent implements OnInit, OnDestroy, AbstractFilter, A
       if (this.isRange) {
         return;
       }
-      let date = this.isStartDate ? this.dateRange.from : this.dateRange.to;
+      let date: moment.Moment | undefined = this.isStartDate ? this.dateRange.from : this.dateRange.to;
       if (!date) {
         date = moment();
       }
```
garethbowen, line 99: To me this is potentially worse than it was, and it was pretty bad before! :laughing: 

With the typing now you're letting the IDE know it might be undefined, but the next line ensures that it's NOT undefined. I'm not sure my IDE is smart enough to figure that out.

I think a better solution would be to extract a function which returns the date which does the ternary logic and the null check and returns a moment.Moment which can never be undefined. This also means it can be a `const` (I think).

webapp/src/ts/services/fast-action-button.service.ts
```diff
@@ -148,7 +155,7 @@ export class FastActionButtonService {
 
   getReportRightSideActions(context: ReportActionsContext): Promise<FastAction[]> {
     const actions = [
-      this.getSendMessageAction(context.communicationContext, { isPhoneRequired: true, useMailtoInMobile: true }),
+      this.getSendMessageAction(context.communicationContext!!, { isPhoneRequired: true, useMailtoInMobile: true }),
```
garethbowen, line 158: I've never seen the `!!` after the expression (and Firefox does not like it). It's devilishly difficult to Google for (all examples I found are for _before_ the expression. Is this just casting it to a boolean? Why put it at the end rather than at the start? Are you trying to teach old dogs like me a new trick???

webapp/src/ts/services/header-tabs.service.ts
```diff
@@ -120,7 +120,7 @@ export class HeaderTabsService {
   }
 }
 
-interface HeaderTab {
+export interface HeaderTab {
```
garethbowen, line 123: :heart:

webapp/tests/karma/ts/modules/tasks/tasks-group.component.spec.ts
```diff
@@ -602,9 +602,6 @@ describe('TasksGroupComponent', () => {
       store.overrideSelector(Selectors.getPreventNavigation, true);
       store.refreshState();
 
-      component.tasks = null;
-      expect(component.canDeactivate('')).to.equal(true);
-
```
garethbowen, line 607: But how can you be sure if you don't check it twice?!

webapp/src/ts/components/filters/date-filter/date-filter.component.ts
jkuester: Ah I forgot to come and roll back these changes after updating the `dateRange` types above.  AFICT this `moment.Moment | undefined` typing is wrong since these can actually have number values as well...

Either way, I don't actually need to specify the type here for `date` at all.  I 100% agree that this code is not great and could be refactored to be much better, but when possible for this PR, I think it will be best to avoid making non-type changes.

webapp/src/ts/modals/edit-user/edit-user-settings.component.html
```diff
@@ -37,7 +37,7 @@
 
     <div class="form-group">
       <label for="language">{{'Language' | translate}}</label>
-      <select id="language" class="form-control" [(ngModel)]="editUserModel.language.code" name="language">
+      <select id="language" class="form-control" [ngModel]="editUserModel.language?.code" (ngModelChange)="editUserModel.language?.code" name="language">
```
jkuester, line 40: This was a strange one.  For some reason, Angular did not like it when I used the `?` with the combined `[(ngModel)]`, but it is fine as long as you break it out.  (Got the idea from SO, where this seemed to be just a known limitation of this combined expression...)

webapp/src/ts/modules/messages/messages.routes.ts
```diff
@@ -31,6 +31,7 @@ export const routes: Routes = [
               }
             };
           }
+          return null;
```
jkuester, line 34: This change was required because of how this code needs to implement `Route` and that specifies the return values for `Route.matcher` (which includes `null` but not `undefined`)

webapp/src/ts/providers/translation-loader.provider.ts
```diff
@@ -14,7 +14,7 @@ export class TranslationLoaderProvider implements TranslateLoader {
 
   getTranslation(locale) {
     if (!locale) {
-      return;
+      return undefined as any;
```
jkuester, line 17: This is probably the worst change I had to make since it is allowing the function to return a value (`undefined`) that is explicitly against the return type for the interface (`Observable<any>`).  This function is never supposed to return null/undefined.  That being said, the `locale` parameter is also not technically allowed to be null/undefined so I don't think we could ever hit this code in prod....

webapp/src/ts/services/fast-action-button.service.ts
garethbowen: Maybe we should use three just to be absolutely sure it's not null?

Thanks for the update! I thought I was losing my Googling skills...

garethbowen (approved):
Nice one! Thanks for taking the time to clean this up.

latin-panda (approved):
Thanks for taking this! ✨✨

Issue #8416, opened 2023-07-29
Include formatted date in API morgan logging


dianabarsan:
**What feature do you want to improve?**
Our API settings for `winston` logging includes the current date, however most of API logs are access logs, which are handled by `morgan` middleware. The format for the `morgan` logging output doesn't include a date. 

**Describe the improvement you'd like**
We are about to start doing log analysis, so precise logging is a milestone in achieving this. Standardizing logging format across types of logs of the same service, and in the end all services, contributes to easy to parse, easy to understand logging.

Issue #4470, opened 2018-04-25
medic-api logs are inconsistently prefixed


alxndrsn:
# Observed

Calls to `console.log()` are _sometimes_ prefixed with timestamp and log level:

```
2018-04-24T17:38:03.746Z - info: TIMED SECTION COMPLETE bindValidatedDocIds().docs_by_replication_key 4013 ms
 TIMED SECTION COMPLETE bindValidatedDocIds().before-callback 4013 ms
 [feed:5480149124257525] prepareResponse() feed.res.finished false

 2018-04-24T17:38:03.747Z - info: [feed:5480149124257525] getChanges() bindServerIds().callback calling feed.res.end()...
 TIMED SECTION COMPLETE getChanges().end 5186 ms

…

 2018-04-24T17:38:04.081Z - info: TIMED SECTION COMPLETE bindValidatedDocIds().docs_by_replication_key 3134 ms
 TIMED SECTION COMPLETE bindValidatedDocIds().before-callback 3134 ms
 [feed:8659737961919713] prepareResponse() feed.res.finished false

 2018-04-24T17:38:04.082Z - info: [feed:8659737961919713] getChanges() bindServerIds().callback calling feed.res.end()...
 TIMED SECTION COMPLETE getChanges().end 4030 ms

…

 2018-04-24T17:38:04.397Z - info: TIMED SECTION COMPLETE bindSubjectIds().before-callback 1249 ms

…

 2018-04-24T17:38:07.685Z - info: TIMED SECTION COMPLETE bindValidatedDocIds().docs_by_replication_key 4473 ms

 2018-04-24T17:38:07.738Z - info: TIMED SECTION COMPLETE bindValidatedDocIds().before-callback 4490 ms
 [feed:3665822082676205] prepareResponse() feed.res.finished true
 [feed:3665822082676205] getChanges() bindServerIds().callback calling feed.res.end()...
 TIMED SECTION COMPLETE getChanges().end 6535 ms
```

# Expected

I'd expect _all_ calls to have this prefix

browndav:
This has been removed in https://github.com/medic/medic-os/pull/32.

garethbowen:
@browndav I think this is unrelated. The script that was removed was about bundling the logs, this is about what gets written to the logs.

garethbowen:
This appears to be fixed in our Docker environment, but now we get two dates for all the request logging, eg:

```
[...]

Issue #3002, opened 2016-12-08
sentinel is logging the date twice


garethbowen:
```
2016-12-08T17:19:35.630Z - info: 2016-12-08T17:19:35.630Z - info: transitions: processing enabled
 
2016-12-08T17:19:35.878Z - info: 2016-12-08T17:19:35.877Z - info: startup complete.
```

estellecomment:
@garethbowen where do you see this happening? I don't see it locally with the latest master, which is also the version of the submodule checked in by medic-webapp. (https://github.com/medic/medic-sentinel/commit/c5c4a9908f7c21d7d3b52c6af00222819432f986)

garethbowen:
That log was taken from standard.app which is a recent version. If it's working perfectly locally I suspect it's gardener (which we don't use in a dev environment) adding the extra date.

garethbowen:
Check if API is too.

estellecomment:
Ha! It is gardener. 
Gardener logs timestamps, sentinel logs timestamps, api doesn't log timestamps.
So we see two timestamps in sentinel + gardener logs.

garethbowen:
Let's just remove the timestamps from sentinel and let gardener do it.

garethbowen:
One small change requested.

estellecomment:
Gareth is out tramping, @alxndrsn can you review https://github.com/medic/medic-sentinel/pull/94 please?

garethbowen:
@estellecomment I think this can be merged now, right?

PR #8417, opened 2023-07-29
chore(#8416): Include formatted date in API morgan logging

dianabarsan:
# Description

medic/cht-core#8416

# Code review checklist

# Compose URLs

* __CHT_CORE_COMPOSE_URL__

# License

### DIFF
--- a/api/server.js
+++ b/api/server.js
@@ -2,6 +2,9 @@ const environment = require('./src/environment');
 const serverChecks = require('@medic/server-checks');
 const logger = require('./src/logger');
 const express = require('express');
+const morgan = require('morgan');
+const moment = require('moment');
+morgan.token('date', () => moment().format('YYYY-MM-DD HH:mm:ss'));
 const apiPort = process.env.API_PORT || 5988;
 
 let router;
--- a/api/src/routing.js
+++ b/api/src/routing.js
@@ -138,13 +138,14 @@ app.use(getLocale);
 morgan.token('id', req => req.id);
 
 app.use(
-  morgan('REQ :id :remote-addr :remote-user :method :url HTTP/:http-version', {
+  morgan(':date REQ: :id :remote-addr :remote-user :method :url HTTP/:http-version', {
     immediate: true,
   })
 );
 app.use(
   morgan(
-    'RES :id :remote-addr :remote-user :method :url HTTP/:http-version :status :res[content-length] :response-time ms'
+    ':date RES: :id :remote-addr :remote-user :method :url HTTP/:http-version :status ' +
+    ':res[content-length] :response-time ms'
   )
 );
 
--- a/api/src/services/setup/router.js
+++ b/api/src/services/setup/router.js
@@ -19,7 +19,7 @@ const STATUS = 503;
 router.use(express.static(environment.staticPath));
 router.use(getLocale);
 router.use(
-  morgan('STARTUP RES :remote-addr :remote-user :method :url HTTP/:http-version :status :res[content-length]')
+  morgan(':date STARTUP RES: :remote-addr :remote-user :method :url HTTP/:http-version :status :res[content-length]')
 );
 
 const getEnabledLocales = () => {


api/server.js
```diff
@@ -2,6 +2,9 @@ const environment = require('./src/environment');
 const serverChecks = require('@medic/server-checks');
 const logger = require('./src/logger');
 const express = require('express');
+const morgan = require('morgan');
+const moment = require('moment');
+morgan.token('date', () => moment().format('YYYY-MM-DD HH:mm:ss'));
```
garethbowen, line 7: Is this enough precision? Surely millis would be useful?

tests/e2e/default/db/ongoing-replication.wdio-spec.js
```diff
@@ -36,6 +37,7 @@ describe('ongoing replication', () => {
 
   after(async () => {
     await sentinelUtils.skipToSeq();
+    await sentinelUtils.waitForSentinel();
```
garethbowen, line 40: Is this unrelated??

garethbowen (approved):
Makes sense. One comment about precision but that may have wider ramifications...

However, this is an area where we've regressed before ( #4470 and #3002 but I'm sure there are more that I'm not finding). It would be good to have an e2e test that checks all logs and makes sure that they are formatted as expected. I understand the desire to get this change into 4.3.0 so I'm comfortable if you want to create a new issue to do the test in a later release.

tests/e2e/default/db/ongoing-replication.wdio-spec.js
dianabarsan: Yes, unrelated, I can revert it.

api/server.js
dianabarsan: I was reluctant to change the format, in case _someone_ is already watching dates from winston logs. 
I'm open to adding millis for sure!

garethbowen: It's probably not too important, but I slightly prefer millis, and it only costs us a few bytes...

dianabarsan: I agree, I'll change that and add the e2e test today.

tests/integration/logging/logging.spec.js
```diff
@@ -0,0 +1,35 @@
...
+
+  expect(moment.utc(date, 'YYYY-MM-DDTHH:mm:ss.SSS', true).isValid()).to.equal(true);
+  expect(moment.utc(date).isBetween(before, after)).to.be.true;
+  expect(action).to.be.oneOf(['REQ:', 'RES:', 'DEBUG:', 'INFO:', 'ERROR:', 'WARN:']);
+};
+
+describe('logging', () => {
+  it('logs should include formatted date in API and Sentinel', async () => {
+    const before = moment.utc();
+    await utils.delayPromise(1000); // dates are not exact
```
garethbowen, line 16: Instead of making the test execution 1 second longer, let's just do some moment date math to remove 1 second from the before variable?

```suggestion
    const before = moment.utc().subtract(1, 'second') // dates are not exact
```

tests/integration/logging/logging.spec.js
```diff
@@ -0,0 +1,35 @@
...
+
+    const collectApiLogs = await utils.collectApiLogs(/.*/);
+    const collectSentinelLogs = await utils.collectApiLogs(/.*/);
+    await utils.delayPromise(1000); // log debug message for checking messages
+    await utils.request('/dbinfo');
+    await utils.updateSettings({ test: true }, 'sentinel');
+    const apiLogs = (await collectApiLogs()).filter(log => log.length);
+    const sentinelLogs = (await collectSentinelLogs()).filter(log => log.length);
+
+    await utils.delayPromise(1000); // dates are not exact
```
garethbowen, line 26: Same here - just add a second to `after`.

tests/integration/logging/logging.spec.js
```diff
@@ -0,0 +1,35 @@
+const utils = require('@utils');
+const moment = require('moment');
+
+const validLog = (line, before, after) => {
+  const [date, action] = line.split(/\s/);
+  console.log(date, before.toISOString(), after.toISOString());
```
garethbowen, line 6: Was this left in by accident?

api/server.js
```diff
@@ -2,6 +2,9 @@ const environment = require('./src/environment');
 const serverChecks = require('@medic/server-checks');
 const logger = require('./src/logger');
 const express = require('express');
+const morgan = require('morgan');
+const moment = require('moment');
+morgan.token('date', () => moment().format(logger.DATE_FORMAT));
```
garethbowen, line 7: Can this be moved to `logger.js`? It feels like it fits more there.

garethbowen (changes requested):
Very nice, love the test!

Some suggestions inline.

tests/integration/logging/logging.spec.js
dianabarsan: yes, sorry! I kept measuring the necessary buffer to never get bad intervals.

api/server.js
dianabarsan: Done

garethbowen (approved):
Make it so.

Issue #8039, opened 2023-01-19
Allow users to report client-side crashes via screenshots


kennsippell:
**What feature do you want to improve?**
Currently when a CHT application causes a crash, the user sees a basic error message with no actionable in distinguishable details.
![image](https://user-images.githubusercontent.com/9014751/213537506-df9be9ad-f1f4-4ae0-847e-391802b4478c.png)

Although the details of the crash are recorded via feedback documents - it would sometimes be useful for users to report their users via screenshot. 

Example: an app developer in an offline context is working beside a user who experiences a crash.
Example: a user experiences a crash which cannot be found in feedback docs
Example: an app developer is asked to help fix crashes in an app where they don't have access.

**Describe the improvement you'd like**
It'd be great if the crashing stack was hidden by default but the user could choose to view it. 
![image](https://user-images.githubusercontent.com/9014751/213540367-aa73ea68-90c0-49b1-b267-3b600fcd2e16.png)

Clicking details could show some details would would allow an app developer to fix the problem:
![image](https://user-images.githubusercontent.com/9014751/213541514-511f9aab-f998-419f-ae01-647a07f1108c.png)

I'm not sure what will realistically fit in a screenshot - but it might also be valuable to include:

* top of crash stack
* instance url
* username
* current date time
* last successful sync time

kennsippell:
I also think we should remove `reload` or replace it with `sync`. I'm unclear if reloading ever solves the problem? What users should be doing to resolve this is communicating and syncing fixes.

dianabarsan:
> I'm unclear if reloading ever solves the problem? 

I believe reloading was an acceptable solution in the past, when tasks were calculated in memory. Reloading would be like a complete try again action, in case the error was transitory. 

In the current context, I agree that sync-ing is probably a better action to take. One scenario where reloading might help and syncing would not is when the error is caused by views not being up to date and internal requests timing out.

jonathanbataire:
Hey @n-orlowski could you look at some mock ups weve have for this :) Thanks cc @kennsippell 
![Screenshot from 2023-02-24 10-09-34](https://user-images.githubusercontent.com/32832540/221116627-ce420097-685a-47dc-aa79-727b032a40d4.png)
![Screenshot from 2023-02-24 10-09-19](https://user-images.githubusercontent.com/32832540/221116662-5799de85-f21c-477b-860a-d5c952141539.png)
[...]

PR #8421, opened 2023-07-31
feat(#8039): display client-side crashes error log

jonathanbataire:
# Description

https://github.com/medic/cht-core/issues/8039

medic/cht-core#[number]

# Code review checklist

# Compose URLs

* __CHT_CORE_COMPOSE_URL__

# License


jonathanbataire:
@jkuester i accidentally closed the previous PR sorry
cant reopen
This is the new working PR

jkuester:
Sorry, I was hoping to get this reviewed today, but I did not make it. Will complete this tomorrow!

tests/e2e/default/analytics/analytics.wdio-spec.js
```diff
@@ -105,4 +105,18 @@ describe('Targets', () => {
       'Targets are disabled for admin users. If you need to see targets, login as a normal user.'
     );
   });
+  
+  it('Should show error message for bad config', async () => {
```
jonathanbataire, line 109: Hey @jkuester this test fails when i run the entire file but passes if i run it only
I suspect that settings compilation and update is messing with it anything you can see here? Thanks

tests/e2e/default/analytics/analytics.wdio-spec.js
```diff
@@ -105,4 +105,18 @@ describe('Targets', () => {
...
+    await updateSettings(settings);
+    await analyticsPage.goToTargets();
+
+    const { errorMessage, url, username, errorStack } = await analyticsPage.getErrorLog();
+
+    expect(username).to.equal(chw.username);
+    expect(url).to.equal('localhost');
+    expect(errorMessage).to.equal('Error fetching targets');
+    expect(await (await errorStack.isDisplayed())).to.be.true;
+
```
jkuester, line 120: Minor but can we assert:
```suggestion
    expect(await (await errorStack.isDisplayed())).to.be.true;
    expect(await (await errorStack.getText())).to
      .include('TypeError: Cannot read properties of undefined (reading \'muted\')');
```

tests/e2e/default/analytics/analytics.wdio-spec.js
jkuester: Okay this was really tricky.  After a few hours of tinkering with this, I figured out the source of the problem is in the `tests/utils/cht-conf.js` file!  Basically line 74 is: 

```js
  const appSettings = require(path.join(dir, 'app_settings.json'));
```

For some reason, if this function gets re-run multiple times _in the same test suite_ then `appSettings` does not actually get populated with the _latest_ from the `app_settings.json` file.  Instead, the initial data from the file is getting cached somehow and it will get reused for the subsequent call to this function.  This causes the behavior you describe because the `should display correct message when no target found` test above will call that function and the `appSettings` data gets initialized with the data from that test.  Then, when this test runs, instead of loading the proper config data for this test, the `appSettings` data from the initial test ends up getting used.

To fix it, I think we can just update that line in the `cht-conf.js` to be:

```js
  const appSettings = JSON.parse(fs.readFileSync(path.join(dir, 'app_settings.json'), 'utf-8'));
```

jkuester (approved):
Excellent!  Just added some minor test suggestiong and this code is ready to go!  Thanks!

jkuester (approved):
:rocket:

tests/e2e/default/tasks/tasks.wdio-spec.js
```diff
@@ -0,0 +1,98 @@
...
+];
+
+const chw = {
+  username: 'bob',
+  password: 'medic.123',
+  place: 'fixture:center',
+  contact: { _id: 'fixture:user:bob', name: 'Bob' },
+  roles: ['chw'],
+  known: true,
+};
```
tatilepizs, line 55: Please use factories to create places, people, and users.

An example that you can use is the [send-message.wdio-spec.js ](https://github.com/medic/cht-core/blob/master/tests/e2e/default/sms/send-message.wdio-spec.js) test file, in there we are using `placeFactory`, `personFactory`, and `userFactory`

tests/e2e/default/tasks/tasks.wdio-spec.js
```diff
@@ -0,0 +1,98 @@
...
+  });
+
+  afterEach(async () => {
+    await utils.revertSettings(true);
+  });
+
+  it('Should show error message for bad config', async () => {
+    const settings = await compileTasks('tasks-error-config.js');
+    await updateSettings(settings);
+    await browser.url('/#/tasks');
```
tatilepizs, line 87: Please use the method `goToTasks()` that is in [common.wdio.page.js](https://github.com/medic/cht-core/blob/master/tests/page-objects/default/common/common.wdio.page.js#L217)

tests/e2e/default/tasks/tasks.wdio-spec.js
```diff
@@ -0,0 +1,98 @@
+const path = require('path');
+const chtConfUtils = require('@utils/cht-conf');
+const utils = require('@utils');
+const loginPage = require('@page-objects/default/login/login.wdio.page');
+const tasksPage = require('@page-objects/default/tasks/tasks.wdio.page');
+const { expect } = require('chai');
```
tatilepizs, line 6:
```suggestion
```

This is not needed, you can remove the line 🙂

tests/e2e/default/analytics/config/targets-error-config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = [
+  {
+    id: 'muted-this-month',
+    context: 'user.parent.type === "district_hospital"',
+    type: 'count',
+    goal: 0,
+    appliesTo: 'contacts',
+    appliesIf: (c) => !!c.foo.muted
+  },
+];
```
tatilepizs, line 10: What do you think about creating a `const` with the name `muted` for example, and then exporting that `const`? in that way will be easier for other people to add new config values and to understand the meaning of that specific config.

I know there are a lot of config files where the code is exactly as you did it, and it is not wrong, is just that we are trying to organize a little bit more the tests code in general 🙂 

You can see this file as an example: [tests/factories/cht/reports/generic-report.js](https://github.com/jonathanbataire/cht-core/blob/8039-error-log/tests/factories/cht/reports/generic-report.js)

tests/e2e/default/tasks/config/tasks-error-config.js
```diff
@@ -0,0 +1,18 @@
...
+    appliesIf: (contact) => contact.foo.name,
+    actions: [{ type: 'report', form: 'person_create_7' }],
+    events: [
+      {
+        start: 0,
+        days: 3,
+        end: 4
+      }
+    ]
+  }
```
tatilepizs, line 17: Same here, please insulate  it into a `const` and then exports that `const`

tatilepizs (changes requested):
This looks great @jonathanbataire, thank you so much for adding e2e tests to the feature ✨ 

I just left some comments related to the structure of the code that we are trying to follow 😄

tatilepizs (approved):
Looks great @jonathanbataire, thank you for the updates ✨

Issue #7462, opened 2022-01-10
Make code for Enketo forms reusable outside cht-core


jkuester:
**What feature do you want to improve?**
Support `cht-core@^1.14`, in https://github.com/medic/cht-conf-test-harness requires [updating the enketo-core version](https://github.com/medic/cht-conf-test-harness/issues/150) used by the test harness and the surrounding harness code that is used to bootstrap the Enketo forms. Much of the logic needed (in the test harness `form-host`) [for bootstrapping enketo](https://github.com/medic/cht-conf-test-harness/blob/master/src/form-host/enketo.js) matches what we are already doing in cht-core currently in the webapp [enketo.service](https://github.com/medic/cht-core/blob/master/webapp/src/ts/services/enketo.service.ts).  

Going forwards, as we continue to update the version of enketo-core used by the CHT, it will require extra effort to ensure the logic for bootstrapping Enketo remains consistent before cht-core and cht-conf-test-harness.  

**Describe the improvement you'd like**
Much of this ongoing maintenance work could be removed or at least simplified by moving as much of the core logic as possible from the [enketo.service](https://github.com/medic/cht-core/blob/master/webapp/src/ts/services/enketo.service.ts) into  a new [shared-libs](https://github.com/medic/cht-core/tree/master/shared-libs) folder that could also be consumed by the cht-conf-test-harness.

**Describe alternatives you've considered**
Definitely open to other approaches that might be better for addressing this issue, but this feels like the most straightforward direction that is also inline with patterns we have followed in the past.

**Additional context**
A proof-of-concept refactoring can be seen in [this commit](https://github.com/medic/cht-core/commit/4bc6e12a0924a21182147ad6e7553028981bd6be) where I tried to just decouple the logic from the Angular dependencies.  Ideally it would probably make sense to have this refactoring be more sophisticated and make it easier to consume.

mrjones-plip:
@jkuester  - is this dependent on the uplift in #6345  such that it should be 4.0.0?

jkuester:
Yes, good call! I have updated the milestone.

jkuester:
For the record, after discussing with @garethbowen we have decided not to put the reusable code into `shared-libs` since the code is not actually going to be reused _within_ cht-core components (it is still only used by `webapp`).  

[...]

PR #8455, opened 2023-08-11
chore(#7462): bifurcate EnektoService into FormService

jkuester:
# Description

The goal of this PR is to separate out the logic in the `EnketoService` that is directly related to rendering Enekto forms and recording the report from the extra logic that is specific to the CHT domain.  A new `EnektoFormService` has been created to contain the form-related logic. This is what the coming cht-form web component can depend on (instead of the web component having to worry about stubbing out a large amount of services that are not actually necessary for just displaying the form and recording the result.  

Whereas before, the cht-form web component would need to stub out behavior for `9` different services (and these stubs would have to be maintained whenever the api for those services changed), after this PR, the cht-form web component will only need to stub behavior for the `DbService`. 

<details>
<summary>Original dependency tree</summary>

```mermaid
flowchart LR

subgraph "Direct EnketoService Depenencies"
	attachment
    cht-script-api([cht-script-api])
    contact-summary([contact-summary])
    db([db])
    enekto-prepopulation-data
    enketo-translation
    extract-lineage
    feedback([feedback])
    file-reader
    get-report-content
    language([language])
    lineage-model-generator
    search([search])
    submit-form-by-sms([submit-form-by-sms])
    training-cards([training-cards])
    transitions([transitions])
    translate
    translate-from
    user-contact
    xml-forms([xml-forms])
    z-score([z-score])
    
    get-report-content --> db & file-reader
    lineage-model-generator --> db
end

subgraph "Transitive Dependencies"
auth
changes
contact-types
debug
feedback
form2sms
format-date
get-data-records
location
pipes
route-snapshot
session
settings
telemetry
uhc-settings
uhc-stats
user-settings([user-settings])
validation
version
xml-forms-context-utils
end

db -.-> session & location
enekto-prepopulation-data --> language & enketo-translation & user-settings
language -.-> settings & format-date & telemetry
search -.-> db & session & get-data-records & telemetry
submit-form-by-sms -.-> form2sms & settings
user-contact --> user-settings & lineage-model-generator & db
xml-forms -.-> auth & changes & contact-types & db & file-reader & user-contact & xml-forms-context-utils & feedback
z-score -.-> changes & db
contact-summary -.-> settings & pipes & feedback & uhc-settings & uhc-stats & cht-script-api
transitions -.-> settings & validation
feedback -.-> db & session & version & debug & language & telemetry
cht-script-api -.-> settings & changes & session
training-cards -.-> xml-forms & db & session & route-snapshot & feedback

```

</details>

<details>
<summary>New dependency tree</summary>

```mermaid
flowchart LR

subgraph "Direct EnketoFormService Dependencies"
	attachment
    db([db])
    enekto-prepopulation-data
    enketo-translation
    extract-lineage
    translate
    translate-from
	enekto-prepopulation-data --> enketo-translation
end
```

</details>

## Details:

I have gone to great lengths to resist the urge of doing any unnecessary low-level refactoring. Instead, whenever possible I have just copied functions in their entirety without touching the contacts. 

- Extracted logic specific to displaying forms from `EneketoService` into a new `EnketoFormService`
    - Basically all the data needed for rendering the form (user settings, contact-summary, etc) can be passed in from the EnektoService to the `EnektoFormService.renderForm` function
- Extracted `UserSettingsService` and `LanguageService` from the `EnketoPrepopulationDataService`
    - I needed to do this in order to keep the EnketoPrepopulationDataService as a part of EnketoFormService, but not have to stub out behavior for UserSettingsService and LanguageService in the web compoent
- Update `UserSettings` service to include a function that returns the user's settings with the `language` pre-populated
    - Convenience method for use by the `EnketoService` and the `Forms2SmsService` (basically just recreating the tiny but of code removed from EnketoPrepopulationDataService)
- The EnketoService and EnketoFormService unit tests represent the majority of the code changes, but I have tried to approach these changes in a way that is both safest and easiest to review:
    - Prior to the final commit in this PR, I only made minor changes to the EnektoService tests as necessary to keep the tests running during the refactoring process.  
    - So, after all of the refactoring changes to the implementation code, the EnektoService unit tests still pass with only minor changes (to account for certain functions being moved to the EnektoFormService).
    - In the final commit of the PR, I have bifurcated the unit tests to separate the tests specific to the EnketoFormService from the ones that pertain to the EnektoService
        - When separating these tests, I have tried, as much as possible, to not edit the contents of the tests, but instead to just change the minimum necessary for the test to call the new EnektoFormService APIs.
        - Due to time constraints, I have not updated the remaining EnektoService unit tests to mock out the behavior of the EnektoFormService. Instead, I have left the logic of these tests intact so that they still call through to the EnektoFormService code.  This means that these tests are unnecessarily complex, but it seems an acceptable tradeoff at this point to avoid a huge re-write of the tests at this time.

medic/cht-core#7462

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7462_reduce_enketo_footprint/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7462_reduce_enketo_footprint/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7462_reduce_enketo_footprint/docker-compose/cht-couchdb-clustered.yml)

# License


webapp/src/ts/services/integration-api.service.ts
```diff
@@ -40,7 +40,7 @@ export class IntegrationApiService {
     this.AndroidAppLauncher = androidAppLauncherService;
     this.Language = languageService;
     this.Select2Search = select2SearchService;
-    this.Enketo = enketoService;
+    this.Enketo = enketoFormService;
```
jkuester, line 43: AFAICT, this is only used by the db-object-widget.  That widget calls the `getCurrentForm` function to load the current form in context.

webapp/src/ts/services/enketo.service.ts
```diff
@@ -143,11 +90,6 @@ export class EnketoService {
       ])
       .then(([html, model]) => {
         const $html = $(html);
-        $html.find('[data-i18n]').each((idx, element) => {
-          const $element = $(element);
-          $element.text(this.translateService.instant('enketo.' + $element.attr('data-i18n')));
-        });
```
jkuester, line 149: Extracted to the `EnketoFormService.replaceDataI18nTranslations` method.  The rest of this logic needs to stay here, though, because we have to check the model to see if we need to actual build the contact-summary or not (before calling `EnketoFormService.renderForm`).

jkuester:
@m5r can you have a look at this when you get the chance?  This refactoring is based on [Gareth's comment](https://github.com/medic/cht-core/issues/7462#issuecomment-1673994384) regarding long-term maintainability for an Enketo Angular Element.  DO NOT BE TOO CONCERNED about the line count!  95% of the changes in this PR are just copy-pasted functions.  I have tried to outline the details of the changes as much as possible in the description.  To reiterate, though, there are no functional changes in this PR at all.  It is all just rearranging the code and doing some minor reordering of the call sequence to isolate the logic specific to enekto forms.

dianabarsan:
I'm away for 2 weeks. Requesting a review from @garethbowen .

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -5,7 +5,8 @@ import { isEqual as _isEqual } from 'lodash-es';
 import { ActivatedRoute, Router } from '@angular/router';
 
 import { LineageModelGeneratorService } from '@mm-services/lineage-model-generator.service';
-import { EnketoFormContext, EnketoService } from '@mm-services/enketo.service';
+import { EnketoService } from '@mm-services/enketo.service';
+import { EnketoFormContext } from '@mm-services/enketo-form.service';
```
garethbowen, line 9: Would it be worth renaming this at the same time?

```suggestion
import { EnketoFormService } from '@mm-services/enketo-form.service';
```

garethbowen (approved):
NB: I haven't reviewed thoroughly, just gone over the high level.

Overall this is great - two large services are much better than one huge one. However I do find myself struggling to understand the difference between them, and if adding new code it may be difficult to pick where to put it unless you _know_ that one is used externally. The simple fix would be to include a comment, but I wonder if we can come up with a name for each service which is more specific and makes the distinction clear?

m5r (approved):
A split well done :+1: 

I agree with @garethbowen, a comment would help to kind of encapsulate each class' role. Even something short like that would go a long way:  
```ts
// Services that need to interact with Enketo should do so through the `EnketoService` and underlying form-specific code should reside in `EnketoFormService`
```

webapp/src/ts/modules/contacts/contacts-edit.component.ts
jkuester: TLDR: I don't think we want to rename `EnketoFormContext` (unless I am totally missing @garethbowen point...)

@m5r, I think you are correct here.  Previously, the `enketo.service` file exported both `EnketoFormContext` and `EnketoService`.  The `EnketoService` was the actual Angular service to use for interacting with Enekto. `EnketoFormContext` is just an interface that defined the parameters of some of the public functions on the `EnketoService`. It was exported to allow consumers of the `EnketoService` to build properly typed arguments to pass in.  

After the refactor, I moved `EnketoFormContext` into the `enketo-form.service` file (since it was also being used there internally).

webapp/src/ts/services/enketo.service.ts
```diff
@@ -442,48 +144,25 @@ export class EnketoService {
...
-        isFormInModal,
-        contactSummary,
-      };
-      const form = await this.renderFromXmls(xmlFormContext);
-      const formContainer = xmlFormContext.wrapper.find('.container').first();
-      this.replaceMediaLoaders(formContainer, formDoc);
-      this.registerEnketoListeners(xmlFormContext.wrapper, form, formContext);
-      window.CHTCore.debugFormModel = () => form.model.getStr();
-      return form;
+      return await this.enketoFormService.renderForm(formContext, doc, userSettings, contactSummary);
```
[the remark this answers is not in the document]
jkuester: :+1: many hours of my life were wasted learning this lesson.... :sweat:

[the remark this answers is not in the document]
garethbowen: Fun fact... we're not using the eslint in that blog post and we're failing in a couple of places.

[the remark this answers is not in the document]
jkuester: :sob:

jkuester:
@garethbowen @m5r I have opted for a both-and approach here. I included comments in each of the classes to try and express the intent of the code and guide future code-changes.  Additionally, I updated the class names. Instead of trying to be too specific or descriptive, I ended up going more generic, but I think it actually works (let me know if you disagree, though, and we can try for something more verbose):

- `EnektoService` -> `FormService` - Seems appropriate since this is the primary interface for CHT code to interact with all forms (contact, app, training).  This change is nice since it abstracts out the "Enekto" implementation detail that the rest of the CHT code should not really care about.
- `EnketoFormService` -> `EnketoService` - Instead of trying to get too fancy here, it seemed to make sense to me to actually just call this class `EnektoService`.  It is now the primary interface for wrapping the Enekto functionality. This Enketo functionality is what will get reused externally in the web component.

So, to recap, the summary of the changes in this PR up to this point are that the `EnektoService` has been renamed to the `FormService` and all the Enekto-specific functionality has been extracted to a new `EnketoService` that can also be reused externally.

garethbowen (approved):
Thanks! FormService is perfect.

Looking at the security hotspots, they're all to do with DoS on unit test code. I think we should update our sonar config ignore security risks on tests...

jkuester:
Ah, interesting!  Apparently, when analyzing PRs, SonarCloud will use the `.sonarcloud.properties` file that is present in the branch for the PR (and not the file from `master`).  This is great because it should allow you to see the effects of any changes you make to the `.sonarcloud.properties` file in your PR.  

It was not great for this PR, though, since my branch was not up-to-date with `master` and it did not have _any_ `.sonarcloud.properties` file.  So, SonarCloud just ran with the default configuration which did not differentiate the test code from the implementation code (and so applied all of the rules to all of the code).  When I updated the branch with the latest from `master` (and pulled in the `.sonarcloud.properties` file), SonarCloud now knew that these files were were test code and did not apply the DoS rule.

garethbowen:
That totally makes sense! Thanks for the explanation.

Issue #8457, opened 2023-08-14
API fails to start because of large form attachments exceed POST allowed payload


dianabarsan:
**Describe the bug**
A [step] in our API startup sequence is to re-generate all forms attachments: https://github.com/medic/cht-core/blob/master/api/src/services/generate-xform.js#L258
The code attempts to update all forms at one, in one big `_bulk_docs` request. The payload for this request will include all form attachments and form docs. This includes images and any other media elements that the forms require. 
This request can fail with a `413 Content Too Large`.

**To Reproduce**
1. Create some forms that have very large media attachments. Upload these forms individually to the server. An alternative would be to create many forms. 
2. Restart API
3. See error

![image](https://github.com/medic/cht-core/assets/35681649/eb6e184b-e413-42a8-a836-9d8873021908)

**Expected behavior**
API should not fail to start when there are too many forms to upload. 

**Environment**
- Instance: any
- App: api
- Version: Code was not changed since 3.7.0, it's possible this only affects 4.0.0+ due to architecture changes, but I have not tested. 

**Additional context**
Originally reported on the forum.

job-isabai:
Hello Everyone, we got this error when upgrading from 3.15 to 4.2.2 using the data migration guide on docs. Currently our installation is stuck at: "Installing forms".

magp18:
We tried commenting the parameters "max_document_size" and "max_attachment_size" (they were commented out) in the default ini of the couchdb inside of the couchdb container and restarting the container (thanks  @delcroip ). Now we don't seem to have the 413 anymore but we have this one instead
![image](https://github.com/medic/cht-core/assets/33976946/61825140-4a90-48c4-ac43-726ff715a3e1)

delcroip:
> We tried commenting the parameters "max_document_size" and "max_attachment_size" (they were commented out) in the default ini of the couchdb inside of the couchdb container and restarting the container (thanks  @delcroip ). Now we don't seem to have the 413 anymore but we have this one instead
> ![image](https://github.com/medic/cht-core/assets/33976946/61825140-4a90-48c4-ac43-726ff715a3e1)
> 

This was done on the couchdb container level

I had once a similar issue with cht-conf when pushing a form with lots/big images.

I worked on the image size so I never checked the root cause

Br

dianabarsan:
Thanks for the feedback @magp18 and @delcroip 

Indeed, I believe you are seeing this issue because you have very large media attached to your forms. 
[...]

PR #8458, opened 2023-08-14
fix(#8457): increase couchdb doc size and retry saving forms

dianabarsan:
# Description

- adds couchdb config `max_document_size` equal to 4GB
- raises couchdb `max_http_request_size` to 4GB
- adds a doc-by-doc retry on 413 errors to `db.saveDocs`

medic/cht-core#8457

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8457-api-forms-payload/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8457-api-forms-payload/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8457-api-forms-payload/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/api/src/db.js
+++ b/api/src/db.js
@@ -136,6 +136,19 @@ if (UNIT_TEST_ENV) {
       });
   };
 
+  const saveDocs = async (db, docs) => {
+    try {
+      return await db.bulkDocs(docs);
+    } catch (err) {
+      if (err.status !== 413 || docs.length === 1) {
+        throw err;
+      }
+
+      const results = await Promise.all(docs.map(doc => db.put(doc)));
+      return results.flat();
+    }
+  };
+
   /**
    * @param {Database} db
    * @param {Array<DesignDocument>} docs
@@ -150,18 +163,7 @@ if (UNIT_TEST_ENV) {
       return [];
     }
 
-    let results;
-    try {
-      results = await db.bulkDocs(docs);
-    } catch (err) {
-      if (err.status !== 413 || docs.length === 1) {
-        throw err;
-      }
-
-      results = await Promise.all(docs.map(doc => db.put(doc)));
-      results = results.flat();
-    }
-
+    const results = await saveDocs(db, docs);
     const errors = results
       .filter(result => result.error)
       .map(result => `saving ${result.id} failed with ${result.error}`);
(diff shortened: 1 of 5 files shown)


dianabarsan:
Hey @garethbowen 
I'm a bit of two minds about this. What do you think?

garethbowen:
@dianabarsan First I agree with changing the couchdb settings to align with couchdb 3 defaults.

With the recursive retry I think it's a good concept. Some options are always savings docs one by one, or instead of splitting by 2 on 413 giving up and saving docs individually at that point. What are your reservations?

dianabarsan:
Yea, I agree that saving docs one by one is less complex. I'll just do that instead.

dianabarsan:
Hey @garethbowen . Requesting another review, if you have the bandwidth. The fix was tested by the [submitting team](https://github.com/medic/cht-core/issues/8457#issuecomment-1678773802). 
Thanks a lot!

api/src/db.js
```diff
@@ -150,7 +150,18 @@ if (UNIT_TEST_ENV) {
...
+    try {
+      results = await db.bulkDocs(docs);
+    } catch (err) {
+      if (err.status !== 413 || docs.length === 1) {
+        throw err;
+      }
+
+      results = await Promise.all(docs.map(doc => db.put(doc)));
+      results = results.flat();
+    }
```
garethbowen, line 163: Nit: Let's pull this out as a private function so we can `return results` rather than overwriting, eg:

```
    try {
      return await db.bulkDocs(docs);
    } catch (err) {
      if (err.status !== 413 || docs.length === 1) {
        throw err;
      }
    }
    const results = await Promise.all(docs.map(doc => db.put(doc)));
    return results.flat();
```

garethbowen (approved):
Love it. Thanks for the quick turnaround! ... and outsourcing the acceptance testing! :D 

One nit inline.

api/src/db.js
dianabarsan: Nit indulged.

Issue #8179, opened 2023-04-14
nginx and haproxy should respect `Accept` headers when responding with error templates


dianabarsan:
**What feature do you want to improve?**
Both nginx and haproxy use the default error templates for 4xx and 5xx responses. These templates are html. 
Respecting the `Accept` header would standardize the behavior of the CHT, and this is important since we expect third party software to call CHT APIs. 

To reproduce: 
- start CHT 4.x 
- stop your api container `docker stop <api container name>`
- request a JSON response from API: 
```shell
curl -k https://localhost -H 'Accept: application/json'
```
```
<html>
<head><title>502 Bad Gateway</title></head>
<body>
<center><h1>502 Bad Gateway</h1></center>
<hr><center>nginx/1.19.6</center>
</body>
</html>
```

If you are logged into webapp, you will see a JSON parse error spammed in the console: 
![image](https://user-images.githubusercontent.com/35681649/232050203-f0b152b0-fe75-4156-bece-2765155dc3b0.png)

**Describe the improvement you'd like**
We should change nginx and haproxy configs to respond with JSON when the `Accept: application/json` header is passed. 

**Describe alternatives you've considered**
Leave it as it is. 

**Additional context**
I had a script that passed the result of an API call to `jq`. Because the response was html, my `jq` call failed. 
Similarly, we've seen reports of feedback docs being generated that feedback docs get generated in webapp when with `Unexpected token < in JSON at position 0`:

https://github.com/medic/cht-core/issues/6412
https://github.com/medic/cht-core/issues/7782

garethbowen:
> Similarly, we've seen

The suspense is killing me!

dianabarsan:
> The suspense is killing me!

Apologies. Resuscitation in progress.

mrjones-plip:
Also noting that the lack of a proper JSON and sending HTML instead affects [more tickets than the two above](https://github.com/search?q=repo%3Amedic%2Fcht-core+%22Unexpected+token+%3C+in+JSON%22&type=issues). Spot checking one, I found:  "This indicates that CouhchDb is down and Haproxy responds with html instead of JSON" [#7759](https://github.com/medic/cht-core/issues/7759#issuecomment-1260451501).

A fix to this will be a welcome addition!

nydr:
We have suggested that HAProxy error pages to be JSON only since HAProxy is not externally available

PR #8446, opened 2023-08-10
fix(#8179) Add support for nginx and haproxy returning json error pages

nydr:
# Description

Resolves medic/cht-core#8179 - nginx and haproxy should respect `Accept` headers when responding with error templates

# Code review checklist

- [x] Documented: Configuration and user documentation on [cht-docs](https://github.com/medic/cht-docs/)
- [x] Tested: Unit and/or e2e where appropriate
- [x] Backwards compatible: Works with existing data and configuration or includes a migration. Any breaking changes documented in the release notes.

# Compose URLs

If Build CI hasn't passed, these may 404:

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:dn/8179-nginx-json/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:dn/8179-nginx-json/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:dn/8179-nginx-json/docker-compose/cht-couchdb-clustered.yml)

# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

PR #8481, opened 2023-08-22
chore(#8179) Add json error messages to haproxy

nydr:
# Description

For the HAProxy part of https://github.com/medic/cht-core/issues/8179 - nginx and haproxy should respect `Accept` headers when responding with error templates

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:dn/8179-haproxy-json/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:dn/8179-haproxy-json/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:dn/8179-haproxy-json/docker-compose/cht-couchdb-clustered.yml)

# License


nydr:
Targeting branch in PR #8446 as it builds on changes from there. Will update target if that PR gets merged first. Please let me know if you have another preferred workflow for PRs that builds on other PRs

CC @dianabarsan @jonathanvq @mrjones-plip

garethbowen:
> Targeting branch in PR https://github.com/medic/cht-core/pull/8446 as it builds on changes from there. Will update target if that PR gets merged first.

That process is fine. I did get a bit confused for a second but worked it out.

haproxy/errors/README.md
```diff
@@ -0,0 +1,4 @@
+# nginx_error_messages
+
+This directory contains error pages for haproxy. Do not update the error pages
+manually, instead look in `generate.sh` and the `template.json` file
```
garethbowen, line 4: In general I prefer generated code to not be checked in to the repo, and instead generated in CI and published. This means someone doesn't accidentally manually update the output and saves on duplication. What do you think? Is there an advantage to checking in the build output?

haproxy/compose.yml
```diff
@@ -0,0 +1,34 @@
...
+        volumes:
+            - ./mock-config/initializerJson.json:/config/initializerJson.json
+
+    haproxy-healthcheck:
+        build: ../haproxy-healthcheck
+        environment:
+            COUCHDB_SERVERS: mock-couchdb
+            COUCHDB_USER: someuser
+            COUCHDB_PASSWORD: insecure_pw
+
```
garethbowen, line 34: Can we move this to the `tests` directory?

haproxy/default_frontend.cfg
```diff
@@ -19,6 +40,9 @@ defaults
   timeout server 360000000
   timeout connect 1500000
   timeout http-keep-alive 5m
+
+  errorfiles json
+
```
garethbowen, line 45: If I understand this correctly, this will always respond with JSON regardless of the Accepts header?

haproxy/errors/generate.sh
```diff
@@ -0,0 +1,71 @@
...
+Cache-Control: no-cache
+Content-Type: application/json
+
+EOF
+
+	echo -n "$http_body" >>"$error_file"
+
+	echo "	errorfile $status_code /usr/local/etc/haproxy/errors/${status_code}-json.http"
+
+done
```
garethbowen, line 71: Nice script! Thanks.

haproxy/tests/with_mock.bats
```diff
@@ -0,0 +1,29 @@
...
+}
+
+@test "should receive response from couchdb" {
+	run bash -c "curl -fksm5 http://127.0.0.1:5984/ | jq .couchdb"
+	assert_success
+	assert_output '"Welcome"'
+
+	run curl -Ifksm5 http://127.0.0.1:5984/doesnotexist
+	assert_failure
+	assert_line --partial --index 0 'HTTP/1.1 404 Not Found'
```
garethbowen, line 22: I think this would be better as a separate test.

garethbowen (changes requested):
Great work, thanks! Some thoughts inline.

haproxy/errors/README.md
nydr: I can see arguments on both sides. The main reason I went with the more explicit pre-generated files is that I expect these will change at most 1-2 times per year and I didn't want to add complexity to the general build process. I imagine debugging an issue with the generated files without having been involved in this PR be more difficult if they're generated on build/start.

What if I add a test to ensure the current files are equal to how they will look when running generate.sh? I prefer that over build step as it doesn't complicate the build process and can be run independently from the build. I added a suggestion to what that could look like in this PR in the file `haproxy/tests/errors_generate.bats` feel free to review that one.

haproxy/default_frontend.cfg
nydr: Correct, since HAProxy is not being called directly by browsers or mobile apps adding a conditional can be considered unnecessary logic (feel free to discuss further in [#8179](https://github.com/medic/cht-core/issues/8179#issuecomment-1682863877))

haproxy/tests/with_mock.bats
nydr: Agreed, updated

haproxy/compose.yml
nydr: That's a good idea, I've moved a few more things only relevant to tests to the test directory

garethbowen (approved):
Looks great! Thanks.

Issue #6775, opened 2020-12-08
The wording on Target Tab not accurate


ngaruko:
**What feature do you want to improve?**
On the target tab, we have a `Disable` button and text above stating that _**"If disabled the targets tab will be hidden for all users".**_  
![image](https://user-images.githubusercontent.com/6979995/101438579-90503880-3977-11eb-8c86-10bc9920b5eb.png)

In reality, the target tab will still show for anyone with the right permissions , with a message that `No analytics modules are configured`

**Describe the improvement you'd like**
Edit that sentence .

garethbowen:
Made obsolete by #8498

PR #8482, opened 2023-08-22
fix(#6775): improve default translation

garethbowen:
# Description

[description]

medic/cht-core#6775

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6775-improve-default-translation/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6775-improve-default-translation/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6775-improve-default-translation/docker-compose/cht-couchdb-clustered.yml)

# License


garethbowen:
@michaelkohn Not a big deal but cleaning up some issues and this seemed like an easy one. Issue is assigned to you so picking on you for the review!

garethbowen:
@michaelkohn I don't think it's leftover from then... But for the life of me I cannot remember what it was for. Looking at the code it is only used for showing the message. I think you might also see the message if targets are not configured. I'm happy to remove the feature, but it will mean that any project who had targets configured but disabled, and the tab visible, will have users start seeing the configured targets, but that seems like a bug in the configuration. What do you think?

michaelkohn:
OK thanks for looking into that! Is it even possible to do the "Disable" from cht-conf?

Either way... I'd say if it's pretty quick and easy to remove it, we should do it.

garethbowen:
Yeah you can disable in cht-conf by updating the config, and you really should, rather than just hitting a button in the UI and then next time you push with cht-conf having the setting overwritten...

michaelkohn:
OK... what do you think, is it quick and easy to get rid of the feature altogether?

garethbowen:
I think so... it's on my list to dig into. Thanks the for the push.

garethbowen:
Closing in favour of https://github.com/medic/cht-core/issues/8498

Issue #6514, opened 2020-06-25
Migrate from SW-Precache to Workbox


latin-panda:
**What feature do you want to improve?**
We have some service workers made with SW-Precache but this library is obsolete already. 

**Describe the improvement you'd like**
SW-Precache evolved to the now called Workbox, which has new features that could be useful in the future, additionally it's good to keep our project's technology updated for better support.

**Describe alternatives you've considered**
This link could be a good guide during the migration: https://developers.google.com/web/tools/workbox/guides/migrations/migrate-from-sw

**Additional context**
Additionally, we would like to test the Workbox feature of [preloading](https://developers.google.com/web/tools/workbox/modules/workbox-navigation-preload) while having HTTP/2 enabled, considering that some regions have slow internet connection, it would be beneficial to improve the initial load of resources even though we are [precaching](https://developers.google.com/web/tools/workbox/modules/workbox-precaching) them with the worker.
This issue came from the research about [improving the initial load of login page by using preload and HTTP/2](https://github.com/medic/cht-core/issues/6276), where it can give a background context and the experience during the research. 
There's another issue about [generating the workers during runtime instead of build time](https://github.com/medic/cht-core/issues/6338), it could potentially take place after this migration is done.

kennsippell:
Some relevant information on why sw-precache was used and not workbox: https://github.com/medic/cht-core/pull/5219

## swPrecache vs Workbox 
It is worth checking out the [Support Details for swPrecache](https://github.com/GoogleChromeLabs/sw-precache#support). TLDR: "We would recommend Workbox for new projects" but "we’ll continue to support sw-precache". 
 
Workbox looks awesome. Most notably check out [enabling offline Google Analytics](https://developers.google.com/web/tools/workbox/modules/workbox-google-analytics). So why isn't this PR using Workbox? The short answer is we probably should – but this is a big PR and Workbox is going to complicate things.  
 
Here is why I'm using the swPrecache library: 
 
1. **The community support isn't perfect for either option** - Workbox 3 supports Chrome >51, but modern [Workbox 4 only supports Chrome >56](https://github.com/GoogleChrome/workbox/releases) so really both swPrecache and latests Workbox are [both not supporting us](https://www.youtube.com/watch?v=bVwh8PW6r88). 
[...]

PR #8483, opened 2023-08-23
fix(#6514): replace sw-precache with workbox

garethbowen:
# Description

[description]

medic/cht-core#6514

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6514-workbox/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6514-workbox/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6514-workbox/docker-compose/cht-couchdb-clustered.yml)

# License


garethbowen:
Hi @latin-panda Can I get a review please? I haven't finished testing this, but it works on my machine. Will test more on actual servers if/when you approve.

api/package.json
```diff
@@ -18,18 +18,17 @@
   "dependencies": {
     "async": "^3.2.1",
     "bikram-sambat": "^1.7.0",
-    "bikram-sambat-bootstrap": "^1.5.0",
```
latin-panda, line 21: Are we removing this from `api` because it's only needed by webapp?

api/package.json
```diff
@@ -18,18 +18,17 @@
...
     "compression": "^1.7.4",
     "express": "^4.18.2",
-    "google-libphonenumber": "^3.2.31",
+    "google-libphonenumber": "^3.2.33",
     "gsm": "^0.1.4",
     "helmet": "^3.23.3",
     "http-proxy": "^1.18.1",
     "locale": "^0.1.0",
-    "lodash": "4.17.19",
+    "lodash": "^4.17.21",
```
latin-panda, line 31: I think upgrading this is okay since the [issue](https://docs.communityhealthtoolkit.org/contribute/code/core/update-dependencies/) has been spotted only in the Admin app.

api/src/generate-service-worker.js
```diff
@@ -56,60 +56,64 @@ const getLoginPageContents = async () => {
   } catch (err) {
     logger.error('Error rendering login page %o', err);
     // default to returning the file
-    return [path.join(apiSrcDirectoryPath, 'templates', 'login', 'index.html')];
+    return [path.join(apiSrcDirectoryPath, 'templates', 'login', 'index.html')]; // TODO figure out if this is right
```
latin-panda, line 59: This will be tested on the server, right? Just making sure the `// TODO` wasn't forgotten :)

api/src/generate-service-worker.js
```diff
@@ -56,60 +56,64 @@ const getLoginPageContents = async () => {
...
-      `!${scriptOutputPath}`, // exclude service worker path
-
-      // Fonts
-      path.join(webappDirectoryPath, 'fontawesome-webfont.woff2'),
-      path.join(webappDirectoryPath, 'fonts', 'enketo-icons-v2.woff'),
-      path.join(webappDirectoryPath, 'fonts', 'NotoSans-Bold.ttf'),
-      path.join(webappDirectoryPath, 'fonts', 'NotoSans-Regular.ttf'),
-      path.join(staticDirectoryPath, 'login', '*.{css,js}'),
-      path.join(staticDirectoryPath, 'login', 'images', '*.svg'),
+    cleanupOutdatedCaches: true,
```
latin-panda, line 81: This will clean up incompatible precaches that were created by older versions of Workbox. 
What happens to the old service worker cache ('sw-precache')? 
Do we need a script to delete them for this transition?

latin-panda (approved):
It looks good to me, I left some small questions.
I didn't have time today to test it but I can do it tomorrow if necessary. 
Approving now to not block the testing on server.

api/package.json
garethbowen: Yeah! I couldn't figure out why it was there, I presume by accident.

api/package.json
garethbowen: Hmm.. I'll consider reverting for consistency. Thanks for pointing it out!

api/src/generate-service-worker.js
garethbowen: I checked through the source and it _should_ delete sw-precache too, but I'll need to test an upgrade to be sure.

.eslintrc
```diff
@@ -52,14 +52,6 @@
...
   },
   "overrides": [
-    {
-      "files": [ "build/service-worker.js" ],
-      "rules": {
-        "max-len": "off",
-        "no-console": "off",
-        "no-var": "off"
-      }
-    },
```
garethbowen, line 62: I'm not sure why we were linting a generated file. With workbox the linting fails horribly but I don't think we care because nobody should be reading it.

api/src/services/branding.js
```diff
@@ -14,7 +14,9 @@ const getBrandingDoc = async () => {
   try {
     return await db.medic.get('branding', { attachments: true });
   } catch(e) {
-    logger.warn('Could not find branding doc on CouchDB: %o', e);
+    if (e.status !== 404) {
+      logger.error('Error fetching branding doc from CouchDB: %o', e);
+    }
```
garethbowen, line 19: The error on 404 looks really alarming but no branding doc is handled gracefully.

api/src/controllers/login.js
```diff
@@ -17,8 +17,7 @@ const template = require('../services/template');
 
 const templates = {
   login: {
-    content: null,
-    file: 'index.html',
+    file: path.join(__dirname, '..', 'templates', 'login', 'index.html'),
```
garethbowen, line 20: I decided to inline the whole file path so it calculates it at runtime not every time it's called.

api/src/controllers/login.js
```diff
@@ -92,9 +90,7 @@ const getEnabledLocales = () => {
 };
 
 const getTemplate = (page) => {
-  const filepath = path.join(__dirname, '..', 'templates', 'login', templates[page].file);
-  templates[page].content = template.getTemplate(filepath);
-  return templates[page].content;
+  return template.getTemplate(templates[page].file);
```
garethbowen, line 93: This was caching the result of the getTemplate call but the `template` service already caches it so this helped clean things up a little.

api/src/generate-service-worker.js
```diff
@@ -51,65 +50,63 @@ const getServiceWorkerHash = async () => {
...
 
 const getLoginPageContents = async () => {
-  try {
-    return await loginController.renderLogin();
-  } catch (err) {
-    logger.error('Error rendering login page %o', err);
-    // default to returning the file
-    return [path.join(apiSrcDirectoryPath, 'templates', 'login', 'index.html')];
-  }
+  return await loginController.renderLogin();
```
garethbowen, line 53: I dug down into the renderlogin function and found it was always gracefully handling errors so we don't need this handling code.

api/src/generate-service-worker.js
garethbowen: I checked and this code was unreachable (the other code never throws).

api/src/generate-service-worker.js
garethbowen: I tested on gamma.dev and the old cache got removed as expected.

Issue #8490, opened 2023-08-24
API crashes when CouchDB is overloaded


m5r:
**Describe the bug**
When CouchDB is overloaded it gets unresponsive and ends up responding with 5xx errors that make the API crash.  
It was seen on a training instance backed by a single couch node with 30 cores and 32 GB of RAM dealing with a spike of thousands of users. More info in https://github.com/medic/cht-core/issues/8324

**Expected behavior**
The API should gracefully handle this error and keep running until CouchDB is operational again.

**Logs**
Full logs were shared [on Slack](https://medic.slack.com/archives/C03H77DKY0M/p1692818007329109?thread_ts=1692792957.766869&cid=C03H77DKY0M) 

Part of the logs with the thrown error:
```text
2023-08-22 12:04:18 ERROR: UNCAUGHT EXCEPTION!
2023-08-22 12:04:18 ERROR:   Error: SyntaxError: Unexpected token < in JSON at position 0
    at JSON.parse (<anonymous>)
    at ServerResponse.interceptResponse (/api/src/controllers/bulk-docs.js:29:19)
    at apply (/api/node_modules/lodash/lodash.js:478:17)
    at ServerResponse.wrapper [as interceptResponse] (/api/node_modules/lodash/lodash.js:5337:16)
    at IncomingMessage.<anonymous> (/api/src/routing.js:865:34)
    at IncomingMessage.emit (node:events:525:35)
    at endReadableNT (node:internal/streams/readable:1358:12)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  [stack]: 'SyntaxError: Unexpected token < in JSON at position 0\n' +
    '    at JSON.parse (<anonymous>)\n' +
    '    at ServerResponse.interceptResponse (/api/src/controllers/bulk-docs.js:29:19)\n' +
    '    at apply (/api/node_modules/lodash/lodash.js:478:17)\n' +
    '    at ServerResponse.wrapper [as interceptResponse] (/api/node_modules/lodash/lodash.js:5337:16)\n' +
    '    at IncomingMessage.<anonymous> (/api/src/routing.js:865:34)\n' +
    '    at IncomingMessage.emit (node:events:525:35)\n' +
    '    at endReadableNT (node:internal/streams/readable:1358:12)\n' +
    '    at processTicksAndRejections (node:internal/process/task_queues:83:21)',
  [message]: 'Unexpected token < in JSON at position 0'
}
Info: Starting CHT API
[...]

PR #8491, opened 2023-08-24
fix(#8490): handle errors when reordering intercepted couchdb response

m5r:
# Description

medic/cht-core#8490

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8490-catch-error-in-intercepted-response/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8490-catch-error-in-intercepted-response/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8490-catch-error-in-intercepted-response/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/api/src/controllers/bulk-docs.js
+++ b/api/src/controllers/bulk-docs.js
@@ -26,9 +26,13 @@ const invalidRequest = req => {
 };
 
 const interceptResponse = (requestDocs, req, res, response) => {
-  response = JSON.parse(response);
-  const formattedResults = bulkDocs.formatResults(requestDocs, req.body.docs, response);
-  res.json(formattedResults);
+  try {
+    response = JSON.parse(response);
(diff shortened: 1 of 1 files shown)


garethbowen (commented):
There's nothing wrong with this specific solution, and avoiding unhandled exceptions is definitely a good thing.

But... I think it's treating the symptom and not the cause. CouchDB/haproxy should never return HTML and we have a PR in flight to resolve this issue.

The other problem with this is there are lots of places where the code parses JSON so it only solves this specific case.

In other words, let's test the haproxy fix and confirm this error is not possible once that's merged, and then do that instead.

PR #8523, opened 2023-09-06
fix(#8527): Use duplicate files rather than symlink

nydr:
To support older docker versions

# Description

Since some dev environments have different behaviour when using symlinks and docker

# Code review checklist

# Compose URLs

* __CHT_CORE_COMPOSE_URL__

# License


nydr:
Resolves #8527

Another potential workaround is to copy or download the files on test/build, but since these files won't change often and can be changed independently I oped for copying over adding complexity in the build process. Feel free to discuss the general approach

dianabarsan:
Cool. I'm stumbling into this a lot - I'm applying the "patch" and the accidentally committing the changes and then wasting time on revert. The faster we get this fixed, the better.

dianabarsan (approved):

Issue #7674, opened 2022-07-12
Answers to non-relevant questions in forms are not immediately cleared with new Enekto


jkuester:
**Describe the bug**
This is a known issue that _will be_ introduced by the [Enketo uplift](https://github.com/medic/cht-core/pull/7256). Logging this here ([as requested](https://github.com/medic/cht-core/issues/6345#issuecomment-1167946463) by @garethbowen) as a point of reference and discussion.

Basically, the issue (reported by @tatilepizs [here](https://github.com/medic/cht-core/issues/6345#issuecomment-1164635324)) is that on old versions of enketo-core, if a question that has already been answered becomes non-relevant (such as if the user goes back and changes a previous answer) the answer value for the non-relevant question would be immediately cleared out and its value would be no longer be available to any XPath expressions throughout the form. If the question became relevant again, the user would need to re-enter their answer.

Now, with the new version of enketo-core that we are uplifting to (`5.18.1`), the `clearIrrelevantImmediately` option [has been removed](https://github.com/enketo/enketo-core/commit/26fb7d47921a71a1af75e95e4e75f1556a6d7c6d) and this makes the behavior a little more complex.  When an answered question becomes non-relevant, it retains its answer value and if it becomes relevant again, the user will not have to re-enter the same answer. If the form is submitted while the question is still non-relevant, the value will be cleared out before the form data is saved.  Unfortunately, due to [this Enketo issue](https://github.com/enketo/enketo/issues/82) the answer value for a non-relevant question will still be available to XPath expressions throughout the form _while the form is being filled out_. When the form is submitted, these expressions will be re-evaluated without the answer value from the non-relevant question, but until then, the expressions will still see the non-relevant answer value.  This, of course, can cause inconsistencies between what is displayed in the form and what is actually saved in the report.

The good news is this inconsistency has [been fixed](https://github.com/enketo/enketo-core/pull/870) in version `6.1.0` of enketo-core which introduces a new `excludeNonRelevant` option (which is still off by default).  

The `excludeNonRelevant` option is not going to have identical behavior to the `clearIrrelevantImmediately` option, but I think the new behavior (where the form remembers what the user had answered before) is actually desirable and does not seem likely to cause issues with existing form config (as long as `excludeNonRelevant` is set to `true`).  

---

[...]

Issue #7599, opened 2022-04-25
Update to latest enketo-core


jkuester:
**Describe the issue**
Not to be confused with #6345 which is upgrading to enketo-core to `^5.18.1`, this issue is for upgrading to the next latest version of Enekto (since it already contains several fixes that we know we would like).

**Describe the improvement you'd like**
Once we have upgraded to the latest enketo-core, these are no longer necessary (since they are included in the new Eneketo code) and can be removed:
- [Logic for pausing multimedia when switching pages](https://github.com/medic/cht-core/commit/772a4feac360e8265af15b1fe7292b47ba2f0649)
- [Patch for Enketo repeat name collision issue](https://github.com/medic/cht-core/commit/3cb901c26a2754769577f6f5dce869ce63cb2de1)

jkuester:
Note that we will need to pay attention to https://github.com/enketo/enketo-core/issues/849 since those changes will almost certainly require some changes to how we are handling saving the `inputs` data.

jkuester:
The current CHT/Enketo has a weird behavior (described in [this forum thread](https://forum.communityhealthtoolkit.org/t/unexpected-duplicate-multi-select-drop-down-component-rendered/2668/6)) where some kind of weird conflict occurs when a calculating a `repeat_count` based on a field that has a `choice_filter`.  This issue appears to be fixed in the latest version of Enketo (I cannot recreate the behavior on https://getodk.org/xlsform/), but we should ensure this is fixed in the CHT when upgrading to the latest version of Enketo.  

To recreate the issue, make a form with the following:

survey:

type | name | label::en | appearance | calculation | choice_filter | repeat_count
-- | -- | -- | -- | -- | -- | --
begin group | begin |   | field-list |   |   |  
select_one yes_no | yes_or_no | Yes or No |   |   |   |  
select_multiple yes_no_translations | selected_translations | Translations |   |   | category=${yes_or_no} |  
begin repeat | my_translations | "No" translations |   |   |   | if(${yes_or_no}='yes',0,count-selected(${selected_translations}))
calculate | my_calc |   |   | selected-at(${selected_translations}, position(..)-1) |   |  
note | my_note | Selected: ${my_calc} |   |   |   |  
end repeat | my_translations |   |   |   |   |  
end group | begin |   |   |   |   |  

choices:

list_name | name | label::en | category
-- | -- | -- | --
  |   |   |  
yes_no | yes | Yes |  
yes_no | no | No |  
  |   |   |  
[...]

Issue #8301, opened 2023-06-06
Notes get rendered as disabled inputs


freddieptf:
**Describe the bug**
Some notes inside a repeat group that's in the form's summary page are getting rendered as disabled inputs

**To Reproduce**
Form i used to reproduce [note_bug.xlsx](https://github.com/medic/cht-core/files/11664608/note_bug.xlsx)
1. Fill one child's details and check the summary page, notes are rendered okay
2. Before submitting go back and increment the number of children
3. Fill the children details again. 
4. Check summary page and observe some notes for the new children get rendered as disabled inputs

**Expected behavior**
All notes get rendered as notes/simple text

**Screenshots**
![Screenshot from 2023-05-31 19-44-05](https://github.com/medic/cht-core/assets/10602047/96bb2348-8741-4147-b0bc-5f5279e6cc8e)

**Environment**
- Instance: Local development instance, reproduced on v4.1.1 and v4.2.0
- Browser: Firefox 113.0.2 (64-bit)
- Client platform: Ubuntu 18.04.6
- App: cht-conf 3.18.3
- Version: 4.1.1, 4.2.0

freddieptf:
The behavior was first reported while testing a PNC form in eCHIS staging instance which runs on v4.1.1 and could be a potential V4 upgrade blocker for [production](https://echis.health.go.ke) which is still on 3.17.0

jkuester:
Investigation for this issue is ongoing, but here is an update.  We have been able to re-create the issue. As far as I have been able to figure out so far, it is produced by combining the following circumstances:

- Nested repeats
- Dynamic `repeat_count` values that are evaluated as the form is filled out
- A `calculate` (or perhaps any expression?) in the _inner repeat_ that is dependent on a value produced in another repeat
- Adding a new entry to the _outer repeat_ when an entry already exists that contains an _inner repeat_ with multiple entries

When all of these things come together, it produces [an error in Enekto](https://github.com/enketo/enketo-core/blob/b83271e8cc88fb79ddeb4f79fb1194d2aac9f421/src/js/form-model.js#L1239).  I am still trying to figure out the root cause of this behavior. It seems to be that the logic is getting confused when trying to calculate the value for the new inner repeat group...

I have not been able to recreate this error out on https://getodk.org/xlsform/, but it is not clear yet if this is a bug in the CHT code or an Enketo bug.

derickl:
I tested this on the enketo uplift branch and the bug wasn't reproducible.

[...]

PR #8528, opened 2023-09-07
feat(#7599): uplift enketo-core to v7

m5r:
# Description

Uplifts to enketo-core `7.x`

Closes medic/cht-core#7599
Closes https://github.com/medic/cht-core/issues/7674

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7599-uplift-enketo-7/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7599-uplift-enketo-7/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7599-uplift-enketo-7/docker-compose/cht-couchdb-clustered.yml)

# License


webapp/src/js/enketo/config.js
m5r, line 1: Some forms interactions like increasing the counter in the repeat test form would trigger a code path looking for this `config.experimentalOptimizations.computeAsync` setting and it would result in the form erroring silently

I looked for the minimum recommended set of configuration properties we should have but I couldn't find anything in Enketo's docs.  
The closest thing I found was this [default config example](https://github.com/enketo/enketo-express/blob/master/config/default-config.json) with strangely-named properties :man_shrugging:

tests/e2e/default/enketo/pregnancy-danger-sign.wdio-spec.js
m5r, line 1: I didn't want to change any test file because they're supposed to keep passing but this test was failing for me on master as well

jkuester:
Just recording here that we need to make sure we test that https://github.com/medic/cht-core/issues/8301 is resolved by these changes (and then we can close out that issue with this PR).

webapp/src/js/enketo/config.js
jkuester: In the past, I have looked at this file (https://github.com/enketo/enketo-core/blob/master/config.js) for an example of what config is available to set.  This `computeAsync` value was the one that I think we want to test having it set to `true` and see what happens (since this would be super helpful for performance on huge forms).

jkuester: Depending on the results of the test, we may want to explore allowing projects to configure this value on a per-deployment basis....

jkuester, line 1: The other big config value we need to set here is:

`excludeNonRelevant: true`

https://github.com/medic/cht-core/issues/7674

jkuester (commented):
This is great!  I think the main remaining things I am interested in are around the config values that I mentioned in the comments.  The `excludeNonRelevant`, in particular, might break the `inputs` patch code.  

The effects of both of these config values should be easier to test now that we can hook up the harness.

webapp/src/js/enketo/config.js
jkuester: This is coming in https://github.com/medic/cht-core/pull/8740

jkuester (approved):
Preemptively approving this, pending https://github.com/medic/cht-core/pull/8740  (I will have limited availability next week so do not want to block this!)

In addition to the automated tests included here in cht-core, I have also updated the test-harness in (https://github.com/medic/cht-conf-test-harness/tree/7462_forms_module_uplifted-enketo) to use the latest code from #8740 and then used that harness branch to successfully run the large automated test suite for https://github.com/medic/config-moh-mali-chw!

Assuming that @tatilepizs's other manual testing efforts are successful, I think this is ready to go!

tatilepizs:
For the last two days I have been doing some manual test using [config-moh-mali-chw](https://github.com/medic/config-moh-mali-chw) config and everything look good, no issues, everything has been working as expected.

These are the forms that I have tested:
|  |  |  | 
| --- | --- | --- | 
| c10_central (create/edit) | c12_central (create/edit) | c20_region (create/edit) | 
|c22_region (create/edit) | c30_district (create/edit) | c32_district (create/edit) | 
| c40_health_area (create/edit) | c42_health_area (create/edit) | c50_supervision_area (create/edit) | 
|c52_supervisor (create/edit) | c60_chw_site (create/edit) | c62_chw (create/edit) | 
| c70_village (create/edit) | c80_concession (create/edit) | c90_household (create/edit) | 
| c92_household (create/edit) | chw_paiement_report | individual_feedback | 
| location_gps | supervision_calendar | visit_community_without_chw | 
| chw_site_gps_location | c74_savings | communication | 
| wash | child_longitudinal_followup_2_5 | mother_referral_followup | 
| anc_followup | pregnancy_followup | pregnancy_referral_followup | 
| treatment_referral_followup_under_5 | mam_followup_under_5 | referral_followup_under_5 | 
| child_longitudinal_followup_1_24 | mas_followup_under_5 | vaccine_followup_under_5 | 
| postnatal_followup | newborn_registration | supervision_visit_realization | 
| supervision_without_chw_confirmation | supervision_with_chw_confirmation | individual_feedback_confirmation | 

I think this uplift is ready 🚀 🚀

Issue #7370, opened 2021-10-08
Uplift messageformat to @messageformat/core


jkuester:
The [messageformat](https://www.npmjs.com/package/messageformat) library has been renamed to [@messageformat/core](https://www.npmjs.com/package/@messageformat/core).  Unfortunately, the first (and so far only) tag of `@messageformat/core` (3.0.0) [dropped support for Node <= 10](https://github.com/messageformat/messageformat/releases/tag/%40messageformat%2Fcore%403.0.0). Since this is a dependency of admin/webapp, I think we are blocked on uplifting this until we do not have to support Node 10.

garethbowen:
Just a note to say since CHT 4.0.0 we've dropped support for Node 10 so this should no longer be blocked.

Issue #5705, opened 2019-06-04
Webapp doesn't support languages that are not pluralized by `make-plural`


dianabarsan:
**Describe the bug**
Reports fail to load when the user language isn't pluralized by `make-plural`. 

**To Reproduce**
1. Install 3.5 (or 3.4)
2. Create a new language with a code that is not-existent or is not pluralized by `make-plural`. For example "rn". 
I prefer to copy whole contents of a translation file that already exists and just change the `code` and `_id` fields to match.
3. Login as an online user to the instance and change your language to "rn". Go to reports. **Refresh**
4. Observe that no reports load on the LHS and translations are missing from some of the filters.

**Expected behavior**
Reports LHS should load. 

**Logs**
There are no errors logged in the console.

**Screenshots**
![image](https://user-images.githubusercontent.com/35681649/58900966-dbe1b080-8708-11e9-9318-cf1a82afb049.png)

**Environment**
- Instance: local, [live instance](https://github.com/medic/medic-projects/issues/6310)
- Browser: Chrome
- Client platform: Linux
- App: webapp
- Version: 3.4, 3.5 (possibly earlier)

**Additional context**
There are no errors visible in the console because `angular-translate` [deliberately ignores all translation loading errors](https://github.com/angular-translate/angular-translate/blob/master/src/service/translate.js#L1950):
```
langPromises[key]['finally'](function () {
    clearNextLangAndPromise(key);
})['catch'](angular.noop); // we don't care about errors (clearing)
``` 
If I change the source to actually log the error, replacing `angular.noop` with `console.error`:
```
Error: Localisation function not found for locale "rn"
    at Object.get (plurals.js:31)
    at new MessageFormat (messageformat.js:37)
    at Object.setLocale (angular-translate-interpolation-messageformat:121)
    at eachInterpolator (angular-translate.js:1633)
    at Object.forEach (angular.js:401)
    at useLanguage (angular-translate.js:1638)
    at angular-translate.js:2324
    at processQueue (angular.js:17914)
    at angular.js:17962
    at Scope.$digest (angular.js:19075)
```
We plug in the `message-format` interpolation to handle plurals, which in turn uses `make-plural` library.
We wait for the `$translate` to be `ready` [before we load forms](https://github.com/medic/medic/blob/master/webapp/src/js/controllers/inbox.js#L421), 
and we wait for forms to be loaded [before we load reports in the LHS](https://github.com/medic/medic/blob/master/webapp/src/js/controllers/reports.js#L471).
[...]

PR #8537, opened 2023-09-11
feat: uplift messageformat package

njuguna-n:
# Description

upgrade messageformat package

medic/cht-core#7370

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7370-uplift-messageformat/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7370-uplift-messageformat/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7370-uplift-messageformat/docker-compose/cht-couchdb-clustered.yml)

# License


webapp/patches/messageformat+2.3.0.patch
njuguna-n, line 1: @garethbowen I didn't see a specific test for this patch so I'm not sure if it's still needed.

garethbowen: We'll definitely need to test this. The new code is here: https://github.com/messageformat/messageformat/blob/main/packages/core/src/plurals.ts . It looks like it handles plurals quite differently and doesn't throw the error.

Can you run through the reproduction steps on the original issue to see what happens now: #5705

admin/package.json
```diff
@@ -7,6 +7,7 @@
     "npm": ">=8.3.1"
   },
   "dependencies": {
+    "@messageformat/core": "^3.2.0",
```
garethbowen, line 10: I think we have an opportunity for some more refactoring here!

I was surprised that you updated messageformat but didn't have to change an import somewhere in admin js. Looking through the code I couldn't find it used anywhere! If I'm right I think you can...

1. Uninstall @messageformat/core from admin
2. That led me down the rabbit hole, because the only mention of "messageformat" in admin is in the `formatDate` service, and only for the `getTaskDueDate` which I don't think is used in admin at all and is a hangover from when we shared the code between webapp and admin. So this function can be deleted. It might be worth checking if any other functions in that service are likewise unrequired, and cleaning them up too!
3. That means we no longer need to load the `angular-translate-interpolation-messageformat` module in the main.js of admin so delete that.
4. _That_ means we don't need to do the weird inclusion step in the `browserify-admin` script so delete that line too.

I might be wrong of course, and you might find that it _is_ actually being used in admin but I can't see how if we never use the `messageformat` interpolation id. Needs testing to confirm, but if you do manage to do all 4 steps it'll really help clean up admin!

garethbowen (changes requested):
Nice work! This has the potential to clean up a huge mess... suggestion inline.

webapp/patches/messageformat+2.3.0.patch
njuguna-n: I followed the steps and the reports are showing up as expected.

admin/package.json
garethbowen, line 10: I think this can be removed altogether now, right?

admin/src/js/services/format-date.js
```diff
@@ -42,79 +41,10 @@ const moment = require('moment');
...
-          }
-          return moment(date).fromNow();
-        },
-        age: function(date, options) {
-          options = options || {};
-          return relativeDate(date, options);
-        },
-        time: function(date) {
-          return format(date, 'time');
-        }
```
garethbowen, line 116: This is amazing work, thank you! I love to see admin slowly being slimmed down.

Now that you've eliminated almost all of this code I think we can remove the factory altogether. It looks like it's only used in one place, in the EditUser controller. The only place it's called we're already getting the `settings` so we can just use that and format the date inline or in a private function. This is how the MessageQueue controller works so you can use that as an example.

garethbowen (changes requested):
Loving all this deleted code!!

A couple more suggestions inline to remove even more and then I promise I'll stop  :)

webapp/package.json
```diff
@@ -33,6 +33,7 @@
     "@angular/platform-browser": "^16.2.3",
     "@angular/platform-browser-dynamic": "^16.2.3",
     "@angular/router": "^16.2.3",
+    "@messageformat/core": "^3.2.0",
```
garethbowen, line 36: I think the rebase had an issue because the old `messageformat` is back in this package.json below.

admin/src/js/controllers/edit-user.js
```diff
@@ -53,6 +53,10 @@ angular
       });
     };
 
+    const formatDate = (date) => {
+      return moment(date).format('DD-MMM-YYYY HH:mm:ss');
+    };
```
garethbowen, line 58: We can still use the configured date format, and it's really easy because the calling code already has the `settings`.

```suggestion
    const formatDate = (settings, date) => {
      const format = settings.reported_date_format || 'DD-MMM-YYYY HH:mm:ss';
      return moment(date).format(format);
    };
```

garethbowen (changes requested):
Two small changes requested, thanks!

Issue #8539, opened 2023-09-12
Sentinel stuck in infinite loop when 100 sequential deletions fail to generate tombstones


dianabarsan:
**Describe the bug**
Sentinel only advances its `transitions-seq` sequence if a successful change is processed. 
Sentinel also only loads 100 docs at a time, to prevent an OOM error when there are millions of changes to process. 

If all of the 100 sequential changes fail to be processed by transitions, sentinel will keep requesting the same 100 docs over and over and be stuck in this loop forever. 

**To Reproduce**
This is easiest to reproduce in 4.2.2 and lower, because we can use tombstones to trigger this. Transitions alone will only error when writes fail, but tombstones can fail for a variety of reasons.

1. Launch CHT v 4.2.2. 
2. Stop sentinel. 
3. With a script, create 1000 documents. Then delete those 100 documents, making sure you only save 3 keys in your delete, `_id`, `_rev` and `_deleted`. 
4. Run compaction. 
5. Start Sentinel. 
6. Leave sentinel running for a while and observe that the number of docs to process stays the same - you can use the monitoring endpoint for this. 
If you check the logs, you will see that sections like: 
```
2023-09-12 14:43:05 ERROR: Tombstone: could not process doc id: <doc_id> seq: <SEQ> {
  error: 'not_found',
  reason: 'missing',
  status: 404,
  name: 'not_found',
  message: 'missing',
  docId: <doc_id>
} 
```
If you search the log for the same doc id, you will find multiple appearances for the same doc. 

**Expected behavior**
For tombstones specifically, we could advance the seq even if the tombstone could not be created (?). For other failures, I'm not sure.

**Environment**
- Instance: live training instance
- App: sentinel
- Version: 3.9+, tombstones no longer trigger this in 4.3+. 

**Additional context**
This was affecting a live training instance. 
As a workaround, the sentinel seq can be advanced manually to skip the problematic batches.

dianabarsan:
Given that this mostly affects v4.2 and lower, because of the interaction with how tombstones are generated. 

The only ways I can see this occurring on non-deletes is by:
- bugs in lineage shared library (uncaught errors)
- bugs in infodoc shared library (uncaught errors)
- errors on saving transitions - CouchDb repeatedly failing to write 

For all of these, the infinite sentinel queue is just a symptom of a separare (bigger?) problem. 
We can address it, but I don't see that solution as important as handling tombstones in 4.2. 

Seeking feedback from @garethbowen . (thanks!!)

[...]

PR #8546, opened 2023-09-14
fix(#8539): Fix Sentinel infinite loop on failed transitions

dianabarsan:
# Description

- advances seq when a tombstone fails to be created
- adds failed doc transitions history. Docs that have failed transitions more than 5 times are not enqueued any more. 

medic/cht-core#8539

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8539-advance-sentinel-seq-on-failed-tombstone/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8539-advance-sentinel-seq-on-failed-tombstone/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8539-advance-sentinel-seq-on-failed-tombstone/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -101,7 +101,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        grunt-cmd: ['ci-webdriver-default', 'ci-webdriver-standard', 'ci-e2e-integration', 'ci-e2e', 'ci-e2e-mobile', 'ci-webdriver-default-mobile']
+        grunt-cmd: ['ci-webdriver-default', 'ci-webdriver-standard', 'ci-e2e-integration', 'ci-webdriver-default-mobile']
 
     steps:
     - name: Login to Docker Hub
--- a/sentinel/src/lib/change-retry-history.js
+++ b/sentinel/src/lib/change-retry-history.js
@@ -1,0 +1,36 @@
+const MAX_HISTORY = 1000;
+const MAX_RETRIES = 5;
+const historyKeys = [];
+const history = {};
+
+const getKey = (change) => change && `${change.id}${change.changes && change.changes[0] && change.changes[0].rev}`;
+
+const add = (change) => {
+  const key = getKey(change);
+  if (!key) {
+    return;
+  }
+
+  if (history[key]) {
+    history[key]++;
+  } else {
+    historyKeys.push(key);
+    history[key] = 1;
+  }
+
+  if (historyKeys.length > MAX_HISTORY) {
+    const deletedKey = historyKeys.shift();
+    delete history[deletedKey];
+  }
+};
+
+const shouldProcess = (change) => {
+  const key = getKey(change);
+
+  return !history[key] || history[key] <= MAX_RETRIES;
+};
+
+module.exports = {
+  add,
+  shouldProcess,
+};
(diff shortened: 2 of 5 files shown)


dianabarsan:
Hi @garethbowen 

This is one option for keeping a history of failed transition docs. We can opt to skip this entirely, but I wanted to see how it would look like. Would you mind checking it out and sharing your thoughts? Thanks!

.github/workflows/build.yml
```diff
@@ -101,7 +101,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        grunt-cmd: ['ci-webdriver-default', 'ci-webdriver-standard', 'ci-e2e-integration', 'ci-e2e', 'ci-e2e-mobile', 'ci-webdriver-default-mobile']
+        grunt-cmd: ['ci-webdriver-default', 'ci-webdriver-standard', 'ci-e2e-integration', 'ci-webdriver-default-mobile']
```
garethbowen, line 104: todo: make this a separate commit

sentinel/src/lib/feed.js
```diff
@@ -74,13 +78,16 @@ const changeQueue = async.queue((change, callback) => {
...
   }
   if (change.deleted) {
-    return tombstoneUtils.processChange(Promise, db.medic, change, logger)
-      .then(() => updateMetadata(change, callback))
-      .catch(callback);
+    return tombstoneUtils
+      .processChange(Promise, db.medic, change, logger)
+      .catch(() => {}) // error already logged by tombstoneUtils
+      // advance sentinel seq even when tombstone creation fails
+      .then(() => updateMetadata(change, callback));
```
garethbowen, line 85: todo: make this a separate commit

sentinel/src/lib/change-retry-history.js
```diff
@@ -0,0 +1,36 @@
...
+const shouldProcess = (change) => {
+  const key = getKey(change);
+
+  return !history[key] || history[key] <= MAX_RETRIES;
+};
+
+module.exports = {
+  add,
+  shouldProcess,
+};
```
garethbowen, line 36: You know how much I love caches...

... but I think this is ok.

An alternative would be to only detect when the transition seq isn't advancing at all. That is to say if an entire batch is processed and 100% fails 5 times in a row then skip over them and move on. This way you're only caching 1 number instead of 1000 keys. Simpler and therefore less buggy?

One advantage of your approach is if you had 99 failing docs my approach would never get past them but yours would. 

Are there other advantages of your suggestion?

dianabarsan: I'll check to see how your suggestion looks like. I think it requires adding some conditionals in the `feed.js` code that would increase complexity.

dianabarsan: There's no guarantee that requesting changes from a certains seq will return the same results every time (because shards and because docs can get updated). 
So saying that a "batch" failed 5 times, it would still mean we need to check every doc in that batch and store which docs were in the failed batch. So we'd still be storing 100 changes in memory, along with the seq. 

Do you think this is worth pursuing and would be easier to maintain or understand than straight up caching doc tries without the seq?

garethbowen: Yeah that's a good point. Go with it as is.

Issue #7148, opened 2021-06-16
Feedback docs: "409: Document update conflict"


binokaryg:
**Describe the bug**
In the month of  May 2021, the feedback documents for `brac-ug` (3.9.2) had this error recorded 38 times from 31 users:

`Possibly unhandled rejection: {"status":409,"name":"conflict","message":"Document update conflict","error":true}`

Most of these are from the tasks page, and others from the contacts and targets page:

Location based on URL | Count (May 2021)
-- | --
Tasks page | 26
Contacts page | 6
Targets page | 2
Specific contact | 1
Family survey form from contact | 1
Family survey task| 1
About page| 1

**Logs**

Here is the top-level argument for each of the 38 incidents (brac-ug, May 2021):

Level | Arguments | Count
-- | -- | --
warn | db.type() is deprecated and will be removed in a future version of PouchDB | 18
error | Error updating _local/rulesStateStore: {\\\"status\\\":409,\\\"name\\\":\\\"conflict\\\",\\\"message\\\":\\\"Document update conflict\\\"} | 7
warn | MaxListenersExceededWarning\",\"2\":\"Possible EventEmitter memory leak detected. 11 \\\"destroyed\\\" listeners added. Use emitter.setMaxListeners() to increase limit. | 4
warn | No active pages found. Make sure your form has at least one relevant input. | 4
info | The above 404 is totally normal. PouchDB is just checking if a remote checkpoint exists. | 4
log | Data node: /*/meta/deprecatedID with null-based index: undefined not found. Ignored. | 1

**Environment**
- Instance: brac-ug.app.medicmobile.org
- Version: 3.9.2

**Additional context**
This error is being recorded every month since we started collecting the feedback docs in August 2020, the highest being 80 times in September 2020, followed by 75 times in March 2021 for `brac-ug`.

Similar error is also recorded for `pih-malawi` every month, although less frequent (1-14 times a month).

kennsippell:
roughly 163 docs matching this pattern this month for PIH production data

dianabarsan:
I'm hitting the `Error updating _local/rulesStateStore: {\"status\":409,\"name\":\"conflict\",\"message\":\"Document update conflict\"}` error consistently in an e2e test once I started recording feedback docs every time `console.error` is called.

PR #8547, opened 2023-09-14
chore(#8089): Generate feedback docs on console.error

dianabarsan:
# Description

We catch errors. We log pretty error messages. None of these pretty error messages are seen because they're caught and they don't trigger our feedback doc algorithm. 
All `console.error` events that log an error will generate a feedback doc. 
Resolves some common bugs that generated feedback docs in e2e tests.
Fixes e2e test feedback doc retrieval script. 

medic/cht-core#8089

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8089-generate-feedback-docs/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8089-generate-feedback-docs/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8089-generate-feedback-docs/docker-compose/cht-couchdb-clustered.yml)

# License


shared-libs/rules-engine/src/provider-wireup.js
```diff
@@ -173,8 +173,11 @@ module.exports = {
     }
 
     // this function accepts subject ids, but rulesStateStore accepts a contact id, so a conversion is required
-    return provider.contactsBySubjectId(subjectIds)
-      .then(contactIds => rulesStateStore.markDirty(contactIds));
+    return enqueue(() => {
+      return provider
+        .contactsBySubjectId(subjectIds)
+        .then(contactIds => rulesStateStore.markDirty(contactIds));
+    });
```
dianabarsan, line 180: This fixes a very common feedback doc:
`Possibly unhandled rejection: {"status":409,"name":"conflict","message":"Document update conflict","error":true}`

 https://github.com/medic/cht-core/issues/7148

webapp/src/ts/services/feedback.service.ts
```diff
@@ -76,7 +124,7 @@ export class FeedbackService {
       this.options.console[level] = (...args) => {
         const logEvent = {
           level,
-          arguments: JSON.stringify(args, getCircularReplacer()).substring(0, 1000),
+          arguments: JSON.stringify(args, getCircularReplacer()).substring(0, 5000),
```
dianabarsan, line 127: I had some trouble diagnosing what was happening with a truncated error of just 1000 characters. This limit was arbitrarily picked.

webapp/src/ts/services/hydrate-contact-names.service.ts
```diff
@@ -35,7 +35,7 @@ export class HydrateContactNamesService {
 
   private getMutedState(summaries, contactSummaries) {
     summaries.forEach((summary) => {
-      if (summary.muted || !summary.lineage || !summary.lineage.length) {
+      if (!summary || summary.muted || !summary.lineage || !summary.lineage.length) {
         return;
```
dianabarsan, line 39: Both of these fixes were required, as lots of tests were generating feedback docs because of this missed nullchecks.

webapp/src/ts/services/unread-records.service.ts
```diff
@@ -33,9 +36,13 @@ export class UnreadRecordsService implements OnDestroy {
...
-    return this.dbService
-      .get({ meta: true })
-      .query('medic-user/read', { group: true });
+    return this.inited
+      .then(() => this.dbService.get({ meta: true }).query('medic-user/read', { group: true }))
+      .catch(err => {
+        if (err.status !== 404) {
+          throw err;
+        }
+      });
```
dianabarsan, line 45: Sometimes, this service tries to get unread records before the meta database is actually synced, and there is a ddoc to query,

webapp/src/ts/services/feedback.service.ts
```diff
@@ -30,9 +31,20 @@ export class FeedbackService {
   };
 
   private logIdx = 0;
+  private lastErrorMessage;
   private readonly LEVELS = ['error', 'warn', 'log', 'info'];
   private readonly LOG_LENGTH = 20;
   private readonly logCircularBuffer = new Array(this.LOG_LENGTH);
+  // List of Error messages to not automatically log to feedback
+  // Can be a lower-cased partial string or a regular expression
+  private readonly NO_FEEDBACK_MESSAGES = [
```
kennsippell, line 40: i'm worried about bugs like https://github.com/medic/config-muso/issues/9 where users generated 100k+ feedback docs... mitigating such issues without a core update would be great, it could be really nice if this list was augmentable/configurable via app_settings (or alike)

webapp/tests/karma/ts/services/feedback.service.spec.ts
```diff
@@ -46,14 +45,14 @@ describe('Feedback service', () => {
 
   afterEach(() => {
     sinon.restore();
-    clock.restore();
+    clock?.restore();
   });
 
   it('should submit feedback when there is an unhandled error', async () => {
+    clock = sinon.useFakeTimers();
```
dianabarsan, line 52: useFakeTimers and fakeAsync don't work together, because they both try to mock the passage of time. If used together, tests just hang because fakeAsync requires the microtask queue to be empty before finishing the test.

webapp/src/ts/services/feedback.service.ts
dianabarsan: An alternative is to keep a local history of more than just last message and not allow more than X duplicates to be submitted in one session.

dianabarsan: This way, there's no way to accidentally disable important feedback docs from being created at all.
We can add a telemetry entry for when the X limit is exceeded, so there is an idea that there's a feedback doc landslide.

kennsippell: would you ever consider moving away from the feedback doc format (one error per doc) and move to a daily error summary doc? like a list of "unique errors" + timestamps/logs for first x instances?

dianabarsan: how hard would that be to query?

dianabarsan: this will add a delay between the error and when it makes it to the server, like telemetry documents. it can also be affected by bugs preventing the aggregation - like telemetry documents. it's a risk to add this complexity, it depends how much we value this information or not, I suppose.

kennsippell: i believe it would be pretty easy to write equivalent postgres queries. i can contribute that if we want a path like that.

i agree with your summary of risks. i suspect users value this information highly. i love the idea that people responding to this data within 24hrs; i don't think that is our reality today.  that said, feels bad to make a technology choice which blocks same-day response from becoming possible.

maybe we stay with the hardcoded list for now? perhaps we can prevent an initial flood of feedback docs via testing?

dianabarsan: I've enjoyed this exchange!
 
To be honest, I am expecting a flood of feedback docs, both legitimate (actual flaws in either software or config) and illegitimate (noise). I think it's a big effort to try to anticipate potential sources, and not something that can be realistically done exhaustively. 
I'm inclined to push this change as-is and monitor projects that launch/upgrade to this version closely to see what they produce. 

Open to options and other ideas!

tests/e2e/default/sms/send-message.wdio-spec.js
```diff
@@ -48,7 +48,8 @@ describe('Send message', () => {
   before(async () => {
     await utils.saveDocs([...places.values(), bob]);
     await utils.createUsers([offlineUser]);
-    await loginPage.login(offlineUser);    
+    await utils.updateSettings({ district_admins_access_unallocated_messages: true }, true);
```
dianabarsan, line 51: Some of the sms messages that this file created ended unallocated. 
I considered changing the test and assert that feedback docs get generated. However, only the first test that created an unallocated record produced a feedback doc, and subsequent tests did not, due to us not generating feedback docs with same error message consecutively. 
This meant that test order was significant, as we would only assert feedback doc creation in one test.

dianabarsan:
Hi @jkuester 
Would you mind having a look? Thank you!

dianabarsan:
It looks like some newly added Enketo tests are generating feedback docs after merging master. I'll address those.

dianabarsan:
I _think_ I'm in a pretty good place with this and seeking some feedback. Can you have a glance, @garethbowen ? Thanks!

tests/utils/index.js
```diff
@@ -1218,6 +1219,20 @@ const updatePermissions = async (roles, addPermissions, removePermissions, ignor
...
 };
 
+const saveFeedbackDocs = async (filename, existingFeedbackDocIds) => {
+  const feedBackDocs = await chtDbUtils.getFeedbackDocs();
+  const newFeedbackDocs = feedBackDocs.filter(doc => !existingFeedbackDocIds.includes(doc._id));
+  if (!newFeedbackDocs.length) {
+    return false;
+  }
+
+  const fileName = `feedbackDocs-${filename}.json`;
```
garethbowen, line 1229: This is hard to read, partially because of the difference in case `fileName` vs `filename`. The parameter should change because it's not actually the filename, it's just part of the filename.

My suggestion is to take the `test` as the parameter and then do all the formatting here in one place, eg:

```
const filename = `feedbackDocs-${test.parent}-${test.title}.json`.replace(/\s/g, '-');
```

webapp/src/ts/modals/edit-user/update-password.component.ts
```diff
@@ -127,7 +128,7 @@ export class UpdatePasswordComponent {
   private async validateConfirmPasswordMatches() {
     if (this.editUserModel.password !== this.editUserModel.passwordConfirm) {
       const value = await this.translateService.get('Passwords must match');
-      await this.setError(ErrorType.PASSWORD, value);
+      this.setError(ErrorType.PASSWORD, value);
```
garethbowen, line 131: Thank you!

I feel like eslint should be able to detect this?

webapp/src/ts/services/feedback.service.ts
```diff
@@ -43,6 +61,63 @@ export class FeedbackService {
...
+    }
+
+    if (typeof exception.error === 'string') {
+      return exception.error;
+    }
+
+    return exception.reason || exception.message;
+  }
+
+  private async generateFeedbackOnError(level, ...args) {
```
garethbowen, line 104: For performance, let's do the `    if (this.FEEDBACK_LEVEL !== level) {` check first. 99% of log messages won't be errors and it saves us doing a bunch of other unnecessary checks.

```suggestion
  private async generateFeedbackOnError(level, ...args) {
    if (this.FEEDBACK_LEVEL !== level) {
      return;
    }
```

webapp/src/ts/services/feedback.service.ts
```diff
@@ -43,6 +61,63 @@ export class FeedbackService {
...
+    if (this.FEEDBACK_LEVEL !== level) {
+      return false;
+    }
+
+    // requiring a valid error to be logged to avoid cascades of feedback docs
+    if (!message || !exception) {
+      return false;
+    }
+
+    // don't double-log errors as a basic infinite loop defense
```
garethbowen, line 74: Nice.

webapp/src/ts/services/feedback.service.ts
```diff
@@ -30,9 +31,26 @@ export class FeedbackService {
...
 
   private logIdx = 0;
+  private lastErrorMessage;
   private readonly LEVELS = ['error', 'warn', 'log', 'info'];
   private readonly LOG_LENGTH = 20;
   private readonly logCircularBuffer = new Array(this.LOG_LENGTH);
+  // List of Error messages to not automatically log to feedback
+  // Can be a lower-cased partial string or a regular expression
+  private readonly NO_FEEDBACK_MESSAGES = [
+    'failed to fetch',
```
garethbowen, line 41: Change this to the equivalent regex with `/i` then we don't need to lowercase the error messages and the code gets cleaner because they're all regexes.

webapp/src/ts/services/feedback.service.ts
garethbowen: Let's pull it out as a constant. 5000 is probably ok.

webapp/src/ts/services/unread-records.service.ts
```diff
@@ -97,10 +104,9 @@ export class UnreadRecordsService implements OnDestroy {
 
   init(callback) {
     this.callback = callback;
-    // wait for db.info to avoid uncaught exceptions: #3754
-    this.dbService
-      .get()
-      .info()
+    // wait for meta db sync to avoid querying meta view before the ddoc was synced from the server
+    this.inited = this.dbSyncService
+      .syncMeta()
```
garethbowen, line 109: This breaks offline first, right?

I think the correct way to do this is to `get` the ddoc from the local db and if it doesn't exist, return an unknown state and try again next time count is called. Or just query the view and ignore the error...

garethbowen (changes requested):
This is awesome.

webapp/src/ts/modals/edit-user/update-password.component.ts
dianabarsan: I feel like that as well.

tests/e2e/default/enketo/unauthorized-form.wdio-spec.js
```diff
@@ -54,6 +57,8 @@ describe('Unauthorized form', () => {
     await commonPage.waitForPageLoaded();
 
     expect(await genericFormPage.getErrorMessage()).to.equal(EXPECTED_UNAUTHORIZED_MESSAGE);
+    const feedbackDocs = await chtDbUtils.getFeedbackDocs();
+    expect(feedbackDocs.length).to.equal(0);
```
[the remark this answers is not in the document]
dianabarsan: Correct. I've removed both checks.

api/src/server-utils.js
```diff
@@ -6,7 +6,7 @@ const logger = require('./logger');
 const MEDIC_BASIC_AUTH = 'Basic realm="Medic Web Services"';
 const cookie = require('./services/cookie');
 
-const wantsJSON = req => req.get('Accept') === 'application/json';
+const wantsJSON = req => !!req.get('Accept')?.includes('application/json');
```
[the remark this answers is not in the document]
dianabarsan: There is one difference in behavior, where `req.accepts` will return true for missing headers, where here we're actually looking for the specific header to exist - we have e2e tests that are failing due to this change. 
To keep previous behavior we'd need to do something like: 
```
!!req.get('Accept') && !!req.accepts('application/json'); 
```

What do you think?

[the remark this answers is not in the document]
garethbowen: I think the suggested way to do it is...

```
req.accepts(['text', 'json']) === 'json'
```

My understanding of the API is that we're specifying `text` as the default if there is no header. It also means if the client specifies a list in the header, it'll return json if that's earlier in the list than text. So it gives the client the chance to set a preference.

I think this is slightly better than what you originally had in that it's more complex than I had thought and this makes it express's problem.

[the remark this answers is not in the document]
dianabarsan: Awesome, you're totally right. Updated!

webapp/src/ts/services/geolocation.service.ts
```diff
@@ -29,7 +29,7 @@ export class GeolocationService {
       }
       return !!window.medicmobile_android.getLocationPermissions();
     } catch (err) {
-      console.error(err);
+      console.error('Error when getting localtion permissions', err);
```
garethbowen, line 32:
```suggestion
      console.error('Error when getting location permissions', err);
```

webapp/src/ts/services/feedback.service.ts
```diff
@@ -43,6 +62,56 @@ export class FeedbackService {
...
+
+    const exception = args.find(arg => arg instanceof Error || arg?.stack || arg instanceof HttpErrorResponse);
+    const exceptionMessage = this.getExceptionMessage(exception);
+    const loggedMessage = String(args[0]);
+
+    if (this.shouldGenerateFeedback(loggedMessage, exceptionMessage, exception)) {
+      const message = exceptionMessage || loggedMessage;
+      this.lastErrorMessage = message;
+      try {
+        await this.createAndSave( { message, stack: exception?.stack });
```
garethbowen, line 107: If a loggedMessage and exceptionMessage are both available it would be ideal to capture both in the feedback doc. Is this possible?

You could take this even further by capturing all args (or all args that aren't objects?) so it could capture all the given detail, for example handling a call like this:

```
console.error('Location failed', `Error code ${code}`, err);
```

garethbowen (changes requested):
Awesome.

One typo, and one suggestion inline.

webapp/src/ts/services/geolocation.service.ts
dianabarsan: whops, thanks for noticing!

webapp/src/ts/services/feedback.service.ts
dianabarsan: Yea, I thought about this as well. I'm sort of worried of breaking any existent queries that work with feedback docs. Do you think I should just add an additional field with the message?

dianabarsan: Done.

garethbowen (approved):
Great work! Love it.

dianabarsan:
Hello @garethbowen . 
As discussed, I added a 1k limit to stored feedback docs in the meta db. 
I also added an exception so `unauthorized` errors do not log feedback docs.

garethbowen (approved):
This feels much safer, thanks!

garethbowen:
@dianabarsan Do you think 1000 feedback docs is the right number? Maybe 100 would be better?

dianabarsan:
I think 100 could be too little. Suppose that a CHW only syncs at the end of the month, and there's some common error that happens 10 times each day. In 10 days, their quota is full and a significant error might not be logged. I think 1000 is safe enough.

garethbowen:
> I think 100 could be too little.

I can see both sides of this. It probably comes down to how often people sync, eg: if you're syncing daily like some users in Nairobi are, then 100 is enough, but if you're only syncing monthly on reporting day then 1000 is better.

Let's stick with 1000 for now and we can tighten it up if we start hitting issues.

PR #8550, opened 2023-09-15
fix: update auth for session get

fardarter:
`require_valid_user = true` as a couch config seems to demand this.

Please if you have the correct couchdb config (2.3.1) let me know. I realise on docker is doesn't seem to need to, but I'm wondering if that's somehow a fact of being on the same host. I'm trying to connect via a haproxy vm.

Honestly PR is at least half just looking for an explanation.

### DIFF
--- a/api/src/auth.js
+++ b/api/src/auth.js
@@ -11,7 +11,7 @@ const get = (path, headers) => {
     url: url.toString(),
     headers: headers,
     json: true
-  });
+  }).auth(environment.username, environment.password, false);
 };
 
 const hasPermission = (userCtx, permission) => {


api/src/auth.js
```diff
@@ -11,7 +11,7 @@ const get = (path, headers) => {
     url: url.toString(),
     headers: headers,
     json: true
-  });
+  }).auth(environment.username, environment.password, false);
```
dianabarsan, line 14: The whole point of this request is to sent then authentication headers that the original request carries, not the ones for the admin user stored in env.

dianabarsan: If this fails, it means that your request headers are not valid user credentials. This is the desired behavior here.

fardarter: Couch is rejected anything I try on `_session`. I've tried posting from the haproxy server with a known valid auth token/cookie and it's rejecting.

fardarter:
```

Comment (dianabarsan) [reply api/src/auth.js:14]: I suspect this is an issue with your setup, and not a bug that affects CHT.

Comment (fardarter) [reply api/src/auth.js:14]: Yeah, maybe. I've followed your settings, though. So, to be clear, you expect to be able to hit http://couchdb.org:5984/_session with nothing more than an auth token?

Comment (dianabarsan) [reply api/src/auth.js:14]: yes, correct. 

```
curl 'http://127.0.0.1:6500/_session' -H 'Accept: application/json' --cookie 'AuthSession=YWRtaW46NjREQ0M2RDI6cOfK78w9qFb-XJyV1UWYtRPLV5o' 
{"ok":true,"userCtx":{"name":"admin","roles":["_admin","admin"]},"info":{"authentication_db":"_users","authentication_handlers":["cookie","default"],"authenticated":"cookie"}}
```

fardarter: Including with `require_valid_user = true` set?

I was half hoping it was a bug. This is driving me mad.

dianabarsan: yes, require_valid_user = true is essential for security... otherwise anyone can read heath data

PR #8552, opened 2023-09-16
fix(couchdb): checks cluster_finished before asserting completion

fardarter:
# Description

Since fixing the bug in `verify_membership(`), my cluster deployments have been failing to achieve `cluster_finished` state, getting stuck in `cluster_enabled`. This ensures that `'{"action": "finish_cluster"}'` is run before completion is asserted.

I assume this will be tested via CI testing.

# License

### DIFF
--- a/couchdb/set-up-cluster.sh
+++ b/couchdb/set-up-cluster.sh
@@ -32,10 +32,14 @@ join_node_to_cluster(){
 
 }
 
+verify_cluster_state_finished() {
+  curl -s http://$COUCHDB_USER:$COUCHDB_PASSWORD@$SVC_NAME:5984/_cluster_setup | jq '.state == "cluster_finished"'
+}
+
 complete_cluster_setup()
 {
-    if $(verify_membership); then
-      echo "Cluster Setup Already Finished"
+    if [ "$(verify_membership)" == 'true' ] && [ "$(verify_cluster_state_finished)" == 'true' ]; then
+      echo "Cluster setup already finished"
     else
       curl -X POST -H "Content-Type: application/json" http://$COUCHDB_USER:$COUCHDB_PASSWORD@$SVC_NAME:5984/_cluster_setup \
       -d '{"action": "finish_cluster"}'


garethbowen:
Not sure how this slipped through the cracks, but @dianabarsan would you mind reviewing please?

dianabarsan (changes requested):
Unfortunately, because of the way we use CouchDb, with a single replica, the cluster never reaches the `clustered_finished` state. 
This means that if we add this, we will try to finish the cluster every time we start couchDb. 

Once we add multiple replicas, we should definitely add this.

fardarter:
> Unfortunately, because of the way we use CouchDb, with a single replica, the cluster never reaches the `clustered_finished` state. This means that if we add this, we will try to finish the cluster every time we start couchDb.
> 
> Once we add multiple replicas, we should definitely add this.

It will only run if there are peer provides, which is the multiple node case, no?

dianabarsan:
> > Unfortunately, because of the way we use CouchDb, with a single replica, the cluster never reaches the `clustered_finished` state. This means that if we add this, we will try to finish the cluster every time we start couchDb.
> > Once we add multiple replicas, we should definitely add this.
> 
> It will only run if there are peer provides, which is the multiple node case, no?

It is the multiple node case, but in our case we do multi-node single replica, which makes the cluster be in `single_node_enabled`. 
You can observe this interaction on this branch: https://github.com/medic/cht-core/tree/cluster-demo

fardarter:
> > > Unfortunately, because of the way we use CouchDb, with a single replica, the cluster never reaches the `clustered_finished` state. This means that if we add this, we will try to finish the cluster every time we start couchDb.
> > > Once we add multiple replicas, we should definitely add this.
> > 
> > 
> > It will only run if there are peer provides, which is the multiple node case, no?
> 
> It is the multiple node case, but in our case we do multi-node single replica, which makes the cluster be in `single_node_enabled`. You can observe this interaction on this branch: https://github.com/medic/cht-core/tree/cluster-demo

Off the top of your head and before I dig, is there some fact I could check before running it?

dianabarsan:
I'm not sure what you mean, @fardarter . The PR just demonstrates my point via a new e2e test. You can also check your local cluster.

fardarter:
I mean is there some way to detect using the API which configuration should not prompt the finish call?

dianabarsan:
Yea, you can call the exact endpoint that the setup script you just changed calls and see what output you get when you setup CouchDb using our image and no additional settings.

garethbowen:
@fardarter Did Diana's respond resolve this issue or do you still want to get this merged?

fardarter:
> @fardarter Did Diana's respond resolve this issue or do you still want to get this merged?

I haven't followed up with a resolution. It seems the code is depending on the bug right now.

Issue #7616, opened 2022-05-03
Telemetry records not logged for every day FCHVs are active


1yuv:
**Describe your bug:**

From the [forum](https://forum.communityhealthtoolkit.org/t/is-it-okay-to-assume-that-if-users-dont-have-telemetry-entry-for-particular-day-they-didnt-open-or-use-the-app/1849), it was assumed that version 3.14.2 onwards, it's safe to assume that if there's no telemetry entry for a day, that means there was activity on that day by CHWs. 

However, in one of the instances with version 3.14.2 on live, we are not seeing telemetry docs created for days when users submitted forms. There could be some issues related to #7606 , but we are seeing telemetry docs not created for an interval of more than a month, however we can see that they've reports submitted frequently even during that period. 

**Web app Version:** 3.14.2 
**Device Hardware Detail**:
```json
  "device": "gta7lite",
        "model": "SM-T225",
        "manufacturer": "samsung",
        "hardware": "mt8768",
        "cpuInfo": {
          "cores": 8,
          "arch": "aarch64"
        }
```
**Device Software Detail:**
```json
"software": {
        "androidVersion": "11",
        "osApiLevel": 30,
        "osVersion": "4.19.127-23190912(T225XXU1AUL1)"
      },
```
**App version:** v0.11.0-webview

dianabarsan:
@yrimal 
Are there any updates on this?

yrimal:
Hi @dianabarsan , 
Currently, this CHT instance is on 3.17.1 and android v 1.0.4, yet the issue still exists. I checked for some users and while users were active for 19 days (Counting only days on which they created new contacts), telemetry record was recorded only for 13 days.

latin-panda:
@ngaruko FYI, I'm going to start working on this one. Please let me know if you have any idea how to reproduce this one 🤔

ngaruko:
@latin-panda Not sure if this is consistent across all users and all devices or if it is only happening for particular users/devices. If it is consistent, then we could easily reproduce by shifting system clocks multiple times. If it is not, one has to do a deeper investigation of particular users' environment to learn what's causing that.
@yrimal Are there any patterns you noticed about users who are experiencing this, that we could use as a starting point for investigation?

michaelkohn:
We _should_ be able to look at some data to see how often this is occurring and/or which users are being affected most often. 

I think we'd look at:
1. **_User_**
2. **_Date_**
3. _Has telemetry record?_
[...]

PR #8555, opened 2023-09-17
fix(#7616): telemetry logs not recording every day

latin-panda:
# Description

- Prevents increasing the `syncedSeq` when sync fails.
- Refactors telemetry to use a telemetry db per day to collect records, then it aggregates the records into a doc in the user's meta db and deletes previous days of telemetry db. 
- Adds temporary code to delete old telemetry db and old localStorage properties.

medic/cht-core#7616

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7616_telemetry_logs_not_recording_every_day/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7616_telemetry_logs_not_recording_every_day/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7616_telemetry_logs_not_recording_every_day/docker-compose/cht-couchdb-clustered.yml)

# License


ngaruko (approved):
Looks good > created a placeholder [ticket ](https://github.com/medic/cht-core/issues/8561)to add an e2e test to cover at this scenario

webapp/src/ts/services/telemetry.service.ts
```diff
@@ -164,65 +116,117 @@ export class TelemetryService {
...
+  private getToday(): TodayMoment {
+    const today = moment().startOf('day');
+    return {
+      today,
+      formatted: today.format(this.DATE_FORMAT),
+    };
+  }
+
+  private async getCurrentTelemetryDB(today: TodayMoment) {
+    const telemetryDBs = await this.getTelemetryDBs();
```
jkuester, line 194: Minor, but since we already have this value in context in the `_record` function, it seems like we could take this as a function parameter and not load it again....

webapp/src/ts/services/telemetry.service.ts
```diff
@@ -263,29 +267,35 @@ export class TelemetryService {
...
-    return this.queue;
+    if (!currentDB || currentDB?._destroyed || currentDB?._closed) {
+      return;
+    }
+
+    try {
+      currentDB.close();
+    } catch (error) {
+      console.error('Error closing telemetry DB', error);
+    }
```
jkuester, line 294: Minor, but I always try to avoid `let`.  Would something like this be acceptable?

```suggestion
    try {
      const today = this.getToday();
      const telemetryDBs = await this.getTelemetryDBs();
      await this.submitIfNeeded(today, telemetryDBs);
      const currentDB = await this.getCurrentTelemetryDB(today);
      try {
        await this.storeIt(currentDB, key, value);
      } finally {
        try {
          currentDB.close();
        } catch (error) {
          console.error('Error closing telemetry DB', error);
        }
      }
    } catch (error) {
      console.error('Error in telemetry service', error);
    }
```

jkuester (commented):
Muy bueno!  All this telemetry logic is really tricky for me to understand, so I tried to break it down into a bulleted logic sequence (that I will share here in case it is helpful to anyone in the future).

### Original Logic
	- Get db `medic-username-telemetry-db` as item from `window.localStorage`
		- If that db does not exist, we create one named `medic-username-telemetry-db-uuid`
	- For that DB:
		- Get `medic-username-telemetry-date` date value from `window.localStorage`
			- If it does not exist, set it to `Date.now()`
		- If that aggregated telemetry date is before "the start of today"
			- Aggregate the db
				- Stores the aggregated data into the user's meta DB (which gets replicated to the server)
			- And reset the db
				- Removes the DB name and Date key from localStorage (so everything is reset
	- Get db `medic-username-telemetry-db(-uuid)` again from `localStorage` (could have been reset during aggregation)
	- Store the provided key/value in the DB along with `date_recorded`
	- Close DB
	- *(End Result is non-aggregated telemetry db will exist called `medic-username-telemetry-db(-uuid)` with telemetry data from the last day that telemetry was recorded (might be today) and the `medic-username-telemetry-date` item in localStorage will be set to the last day that telemetry was recorded (might be today))*

### New logic
	- Get the telemetry DB names:
		- Load all the `document.defaultView?.indexedDB?.databases()`
		- For each db name, remove the `_pouch_`
		- Return only the names that start with `telemetry`
	- For each `telemetry` DB:
		- Skip aggregation if:
			- Aggregation is already running
			- The `telemetry` DB is for today
		- Otherwise aggregate the DB:
			- Run original aggregation logic to store data to user's meta DB
		- And `destroy` the Pouch DB
			- Seems to imply total [deletion of db](https://pouchdb.com/api.html#delete_database)
	- Get the current telemetry DB
		- Find the telemetry DB that for today's date
		- If one does not exist, create one called: `telemetry-YYYY-MM-DD-username`
	- Store the provided key/value in the DB along with `date_recorded`
	- Close DB

The only real question I have about this  logic, @latin-panda, is how will things work on a user's device on the day that they take this upgrade?  It seems like if a user has an existing `medic-username-telemetry-db-uuid` database on their device that has not been aggregated yet before the upgrade to this new logic happens then the telemetry in that DB will never actually get aggregated (since the new aggregation logic does not look for that DB).  Also, the `medic-username-telemetry-db-uuid` database and the `medic-username-telemetry-date` date value will never be cleaned up (even though they are no longer used...).  Am I missing something here, or should we consider adding some logic to handle this cut-over period?

webapp/src/ts/services/telemetry.service.ts
```diff
@@ -164,65 +116,117 @@ export class TelemetryService {
...
+      date_recorded: Date.now(),
+    });
   }
 
-  private reset(db) {
-    window.localStorage.removeItem(this.DB_ID_KEY);
-    window.localStorage.removeItem(this.FIRST_AGGREGATED_DATE_KEY);
+  private async submitIfNeeded(today: TodayMoment, telemetryDBs: string[] = []) {
+    for (const dbName of telemetryDBs) {
+      if (this.isAggregationRunning || dbName.includes(today.formatted)) {
```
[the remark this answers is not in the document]
latin-panda: > I wonder if it is here as a locking mechanism to try and avoid multiple calls simultaneous calls to the telemetry record function from doing duplicate aggregation work.

Correct, when calling `telemetry.record(...)` there's a risk of getting this exception: `Failed to execute 'transaction' on 'IDBDatabase': The database connection is closing.`. That's what we're trying to avoid with `isAggregationRunning`, because the other option is to always `await` for `telemetry.record(...)`. So, it's safer to do it here.

> would it make more sense to move the isAggregationRunning logic outside of the for-of loop?

Yes, that makes a lot of sense! Thanks

webapp/src/ts/services/telemetry.service.ts
```diff
@@ -164,65 +116,117 @@ export class TelemetryService {
...
+    };
+  }
+
+  private async aggregate(db, dbName) {
+    const metadata = await this.generateMetadataSection(dbName);
+    const dbInfo = await this.dbService.get().info();
+    const reduceResult = await db.query(
+      { reduce: '_stats', map: (doc, emit) => emit(doc.key, doc.value) },
+      { group: true }
     );
```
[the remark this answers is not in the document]
latin-panda: Yes, I agree with this!

webapp/src/ts/services/telemetry.service.ts
latin-panda: Yeah, I was back and forth on this, I tried to keep it flat and seemed simpler. I think a bit more how to make it prettier

webapp/src/ts/services/telemetry.service.ts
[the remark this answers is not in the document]
latin-panda: I've added a comment to clarify this

webapp/src/ts/services/telemetry.service.ts
latin-panda: I think I got the best of both worlds. 😬

latin-panda:
@jkuester This is ready for review again. Thanks for your time!

> should we consider adding some logic to handle this cut-over period?

Last week Gareth and I had a conversation where we decided just to remove the old Telemetry db and not aggregate those records. The impact of not having that data is small, especially because it took long until someone noticed it and it is already losing data due to this bug. We would like to keep this transition code small and easy to remove in the future (related [slack message](https://medic.slack.com/archives/C024KTGRW/p1697524837377979))

jkuester (approved):
Awesome! Thanks for digging into this!  LGTM

PR #8562, opened 2023-09-19
chore(tests): tear down test containers when e2e tests complete

jkuester:
Currently, after integration or E2E tests, the test CHT instance does not get stopped/removed. So, running the tests locally results in an additional CHT instance that is constantly running on my local machine (that does not get torn down until the next time I run the e2e/integration tests).

This change will cause the containers to be stopped/removed by default.  However, if the `DEBUG` envar is set when running the tests, the containers will be retained.

(As a side note, I have included some documentation about the `DEBUG` flag in https://github.com/medic/cht-docs/pull/1185)

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:tear_down_test_containers/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:tear_down_test_containers/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:tear_down_test_containers/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/tests/utils/index.js
+++ b/tests/utils/index.js
@@ -983,7 +983,7 @@ const prepServices = async (defaultSettings) => {
 
   updateContainerNames();
 
-  await tearDownServices(true);
+  await tearDownServices();
   await startServices();
   await listenForApi();
   if (defaultSettings) {
@@ -1038,10 +1038,8 @@ const saveLogs = async () => {
   }
 };
 
-const tearDownServices = async (removeOrphans) => {
-  if (removeOrphans) {
-    return dockerComposeCmd('down', '-t', '0', '--remove-orphans', '--volumes');
-  }
+const tearDownServices = async () => {
+  await dockerComposeCmd('down', '-t', '0', '--remove-orphans', '--volumes');
   await saveLogs();
 };
 
--- a/tests/utils/index.js
+++ b/tests/utils/index.js
@@ -20,9 +20,9 @@ process.env.COUCHDB_USER = constants.USERNAME;
 process.env.COUCHDB_PASSWORD = constants.PASSWORD;
 process.env.CERTIFICATE_MODE = constants.CERTIFICATE_MODE;
 process.env.NODE_TLS_REJECT_UNAUTHORIZED=0; // allow self signed certificates
+const DEBUG = process.env.DEBUG;
 
 let originalSettings;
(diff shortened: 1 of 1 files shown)


tests/utils/index.js
```diff
@@ -1038,10 +1038,8 @@ const saveLogs = async () => {
   }
 };
 
-const tearDownServices = async (removeOrphans) => {
-  if (removeOrphans) {
-    return dockerComposeCmd('down', '-t', '0', '--remove-orphans', '--volumes');
-  }
+const tearDownServices = async () => {
+  await dockerComposeCmd('down', '-t', '0', '--remove-orphans', '--volumes');
```
dianabarsan, line 1042: I have added this on purpose, because I frequently inspect the existent containers to debug e2e tests. 
I would prefer not to remove it.

jkuester: :laughing: You are a step ahead of me here!  I was going to solicit your help with this [[Chesterton's Fence](https://en.wiktionary.org/wiki/Chesterton%27s_fence)](https://en.wiktionary.org/wiki/Chesterton%27s_fence) once I determined if this was actually required for the CI or not....

My main complaint with the current behavior is that it results in an additional CHT instance that is constantly running on my local machine (that does not get torn down until the *next time* I run the e2e/integration tests).  For me, this is a somewhat minor annoyance (since my laptop is well provisioned), but this could be a bigger issue for someone with less system resources.  Additionally, it just seems like a bad practice to leave the test instance running *by default*.  An external contributor to the CHT might not be super familiar with Docker and may not even realize that, by running the tests locally, they will have a CHT instance eternally running on their computer....

At the _very least_ it seems that we should print a message when the tests complete that includes the command to run for tearing down the instance.

How would you feel about removing the containers by default, but skipping this command and keeping the containers if `process.env.DEBUG` is truthy?  (Seems to align well the the other debugging functionality in the tests that get toggled with that flag....)

dianabarsan: That would work quite well, actually :)

dianabarsan (approved):

Issue #8589, opened 2023-09-26
Users unable to edit the report they created


1yuv:
**Describe the bug**
We've seen CHT user(s) are able to create report but they are not able to edit the same report. 

**To Reproduce**
1. Create a report with offline CHT user for particular contact
2. Go to Reports tab and select same report
3. Click on Edit to edit this report.

**Expected behavior**
Users should be able to edit the report that they created if permission allows that. 

**Screenshots**
See the video:

https://github.com/medic/cht-core/assets/6102813/8c7c7652-efb4-4903-b40b-442438ca53ba

**Environment**
- Instance: walling.dev
- Browser: Chrome
- Client platform: MacOS
- App: Webapp
- Version: 4.3.1 and 4.4.0 both tested

**Additional context**
app settings and forms used for this: medic/config-nssd#406 (internal)

dianabarsan:
@yrimal 
Can you please also share the permissions of the form?

yrimal:
Hi @dianabarsan , 
Here's the forms.properties section 
```json
{
  "title": [
    {
      "locale": "en",
      "content": "Non-communicable diseases survey form"
    },
    {
      "locale": "ne",
      "content": "नसर्ने रोग सम्बन्धी सर्बेक्षण फारम"
    }
  ],
  "context": {
    "place": false,
    "person": true,
    "expression": "contact.contact_type === 'c82_person' && user._id != contact._id"
  },
  "icon": "icon-healthcare-assessment"
}

```

Contact type for which this report was added is `c82_person`.

dianabarsan:
I am able to replicate this with default config and submitting a `death_report` form. It looks like the form expression context does not get populated with current form data when it's evaluated for an edit.

dianabarsan:
I believe this started manifesting when we merged https://github.com/medic/cht-core/issues/6505 and started evaluating form context on edit. The issue is that the context that we sent on edit does not have the same format as on create.

garethbowen:
I'm not 100% convinced by your examples, because in both cases someone is editing a report that is no longer relevant anyway. Why edit a report for a muted contact? Or for a condition that no longer applies to the patient? A particularly fastidious health worker may want to fix the data regardless but I think it's unlikely.

[...]

PR #8598, opened 2023-09-28
fix(#8589): don't check form expression on edit

dianabarsan:
# Description

medic/cht-core#8589

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8589-edit-form-with-expression/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8589-edit-form-with-expression/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8589-edit-form-with-expression/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/webapp/src/ts/services/enketo.service.ts
+++ b/webapp/src/ts/services/enketo.service.ts
@@ -652,7 +652,7 @@ export class EnketoFormContext {
     this.instanceData = instanceData;
   }
 
-  evaluateExpression() {
+  shouldEvaluateExpression() {
     if (this.type === 'report' && this.editing) {
       return false;
     }
--- a/webapp/src/ts/services/enketo.service.ts
+++ b/webapp/src/ts/services/enketo.service.ts
@@ -635,7 +635,7 @@ interface XmlFormContext {
 export class EnketoFormContext {
   selector: string;
   formDoc: Record<string, any>;
-  type: 'contact'|'report'|'task'|'training-card';
+  type: string; // 'contact'|'report'|'task'|'training-card'
   editing: boolean;
   instanceData: null|string|Record<string, any>;
   editedListener: () => void;
@@ -645,7 +645,7 @@ export class EnketoFormContext {
   userContact?: Record<string, any>;
   contactSummary? :Record<string, any>;
 
-  constructor(selector?, type?, formDoc?, instanceData?) {
+  constructor(selector:string, type:string, formDoc:Record<string, any>, instanceData?) {
     this.selector = selector;
     this.type = type;
     this.formDoc = formDoc;
--- a/webapp/src/ts/services/enketo.service.ts
+++ b/webapp/src/ts/services/enketo.service.ts
@@ -165,13 +165,13 @@ export class EnketoService {
   }
 
   private renderFromXmls(xmlFormContext: XmlFormContext, userSettings) {
-    const { doc, instanceData, titleKey, wrapper, isFormInModal, contactSummary } = xmlFormContext;
+    const { doc, data, titleKey, wrapper, isFormInModal, contactSummary } = xmlFormContext;
 
     const formContainer = wrapper.find('.container').first();
     formContainer.html(doc.html.get(0)!);
 
     return this
-      .getEnketoForm(wrapper, doc, instanceData, contactSummary, userSettings)
+      .getEnketoForm(wrapper, doc, data, contactSummary, userSettings)
       .then((form) => {
         this.currentForm = form;
         const loadErrors = this.currentForm.init();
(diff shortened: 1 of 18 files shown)


garethbowen:
@jkuester How are you getting on with this review? This will get backported to 4.4.x and it's blocking the patch release.

jkuester:
Sorry @garethbowen! Looking at this now!

webapp/src/ts/services/form.service.ts
```diff
@@ -142,33 +142,33 @@ export class FormService {
...
-      { doc: instanceData?.contact, contactSummary: contactSummary?.context },
+      formContext.formDoc,
+      formContext.userContact,
+      {
+        doc: typeof formContext.instanceData !== 'string' && formContext.instanceData?.contact,
+        contactSummary: formContext.contactSummary?.context,
+        shouldEvaluateExpression: formContext.shouldEvaluateExpression(),
+      },
     );
   }
```
jkuester, line 155: At this point it seems like we should make the `xmlFormsService.canAccessForm` function take an `EnketoFormContext` as the parameter and then encapsulate the rest of this in that function. That seems more straightforward than doing this interim data translation here....

webapp/src/ts/services/enketo.service.ts
```diff
@@ -629,16 +629,38 @@ interface XmlFormContext {
...
+  editing: boolean;
+  instanceData: null|string|Record<string, any>;
   editedListener: () => void;
   valuechangeListener: () => void;
   titleKey?: string;
   isFormInModal?: boolean;
   userContact?: Record<string, any>;
+  contactSummary? :Record<string, any>;
+
+  constructor(selector?, type?, formDoc?, instanceData?) {
```
jkuester, line 648: Minor, but since this class is getting pretty broad usage, it would be nice to be more specific about the types of the params here and which are _required_.  (e.g. I am pretty sure we should always be getting the first three....)

jkuester: Thinking about this more, I wonder if we should just remove instanceData from the constructor. It can get set in the same way as the other optional params....

webapp/tests/karma/ts/services/form.service.spec.ts
```diff
@@ -350,8 +355,10 @@ describe('Enketo service', () => {
       ContactSummary.resolves({ context: { pregnant: true } });
       Search.resolves([{ _id: 'somereport' }]);
       LineageModelGenerator.contact.resolves({ lineage: [{ _id: 'someparent' }] });
-      return service.render($('<div></div>'), mockEnketoDoc('myform'), instanceData).then(() => {
+      const formContext = new EnketoFormContext('div', 'report', mockEnketoDoc('myform'), instanceData);
+      return service.render(formContext).then(() => {
         expect(EnketoForm.callCount).to.equal(1);
+        console.log(JSON.stringify(EnketoForm.args[0], null, 2));
```
jkuester, line 361:
```suggestion
```

webapp/src/ts/services/enketo.service.ts
jkuester: Just need a builder pattern here to go with the class....  😂

webapp/tests/karma/ts/services/enketo-form.service.spec.ts
jkuester, line 1: :facepalm: Looks like I forgot to rename this test file during the last round of name changes in the Enekto refactor.  Would you mind just renaming this class to be `enketo.service.spec.ts`?

jkuester (approved):
A couple minor comments in-line, then this is good to go!  :tada: I really like that we are further streamlining and simplifying the logic here!

webapp/src/ts/services/enketo.service.ts
dianabarsan: I added `instanceData` after getting frustrated by e2e tests.

webapp/src/ts/services/form.service.ts
dianabarsan: I think the reason for not passing form context in the first place is because we call this function when we filter forms to be displayed for users, in the FAB menu for example or on the reports page. It will not be a form context in that case, and I don't think we should build it as form context because of that. 

https://github.com/medic/cht-core/blob/69601737292037610b1c93f27a6e454281a02cd4/webapp/src/ts/services/xml-forms.service.ts#L239

PR #8600, opened 2023-09-29
fix(#8589): don't check form expression on edit for 4.3

dianabarsan:
# Description

medic/cht-core#8589

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8589-430-backport/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8589-430-backport/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8589-430-backport/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/webapp/src/ts/services/enketo.service.ts
+++ b/webapp/src/ts/services/enketo.service.ts
@@ -434,11 +434,15 @@ export class EnketoService {
     });
   }
 
-  private canAccessForm(formDoc, user, instanceData, contactSummary) {
+  private canAccessForm(formContext: EnketoFormContext) {
     return this.xmlFormsService.canAccessForm(
-      formDoc,
-      user,
-      { doc: instanceData?.contact, contactSummary: contactSummary?.context },
+      formContext.formDoc,
+      formContext.userContact,
+      {
+        doc: typeof formContext.instanceData !== 'string' && formContext.instanceData?.contact,
+        contactSummary: formContext.contactSummary?.context,
+        evaluateExpression: formContext.evaluateExpression(),
+      },
     );
   }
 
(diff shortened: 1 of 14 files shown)


dianabarsan:
I dislike all of this. :(

webapp/src/ts/services/enketo.service.ts
```diff
@@ -901,16 +886,38 @@ interface XmlFormContext {
...
+  contactSummary? :Record<string, any>;
+
+  constructor(selector?, type?, formDoc?, data?) {
+    this.selector = selector;
+    this.type = type;
+    this.formDoc = formDoc;
+    this.instanceData = data;
+  }
+
+  evaluateExpression() {
```
garethbowen, line 912: I think this name should be changed to `shouldEvaluateExpression` or similar.... it reads like it's going to actually do the evaluation.

webapp/src/ts/services/enketo.service.ts
```diff
@@ -434,11 +434,15 @@ export class EnketoService {
...
     return this.xmlFormsService.canAccessForm(
-      formDoc,
-      user,
-      { doc: instanceData?.contact, contactSummary: contactSummary?.context },
+      formContext.formDoc,
+      formContext.userContact,
+      {
+        doc: typeof formContext.instanceData !== 'string' && formContext.instanceData?.contact,
+        contactSummary: formContext.contactSummary?.context,
+        evaluateExpression: formContext.evaluateExpression(),
```
garethbowen, line 444: Same naming concern here... should be a noun not a verb.

webapp/src/ts/services/enketo.service.ts
```diff
@@ -901,16 +886,38 @@ interface XmlFormContext {
...
+
+  constructor(selector?, type?, formDoc?, data?) {
+    this.selector = selector;
+    this.type = type;
+    this.formDoc = formDoc;
+    this.instanceData = data;
+  }
+
+  evaluateExpression() {
+    if (this.type === 'report' && this.editing) {
```
garethbowen, line 913: Wouldn't this also apply when editing contacts?

garethbowen (approved):
Some nits, but otherwise you're good to go.

I think we can improve on this on `master` but for a minimal change this is good. Thanks for doing it!

webapp/src/ts/services/enketo.service.ts
dianabarsan: I'm worried that that might be desired behavior?

dianabarsan: Coming back to this: the contact add and edit forms are different, so the expression can be different between them. I'm tempted to think that we should allow blocking contact edits depending on expression.

Issue #7670, opened 2022-07-09
Same patient id assigned to multiple patients in a same instance


1yuv:
**Describe the bug**
I've seen that with same patient_id has been assigned to more than one patient. This causes a huge impact on the service being delivered as two patients will be treated as a single patient. This also has impact on impact monitoring and analytics. 

**To Reproduce**
1. Go to dho baitadi's production instance.
2. Open document `d935a8c916709b7b13a46e01b97fb9c2` 
3. Open document `64644cba05d134cef802c0202a61c855`
4. They are assigned a same patient_id `97032`. 

**Expected behavior**
Each patient should have an unique patient id generated for them. 

**Environment**
- Instance: baitadi production instance
- App: Web App
- Version: 3.7.1

dianabarsan:
Checking the instance, I'm noticing that the two documents were created in `Sat Feb 17 2018` and `Thu Sep 13 2018` respectively. 
3.0.0 was released in Nov 14, 2018. 
Can you check which version the app was using at the dates  when these documents were created? 
Can you find any more recent duplicates?

dianabarsan:
Very many things have changed in the code that affects how transitions work, that it would be impossible to know what the problem was without knowing exactly which version the server was running at the time the duplicate id documents were created.

yrimal:
Hi @dianabarsan , per [this comment](https://github.com/medic/medic-projects/issues/7043#issuecomment-562000732), upgrade from 2.13.x to 3.7 was completed on December 5, 2019. We saw this document `f87bfb6b-b5f7-4b55-bec7-173d10770be1` assigning previously assigned patient id on December 7, 2019.

dianabarsan:
I'm suspecting something went wrong with a migration. 
I checked the two documents linked in the initial issue, and failed to find the corresponding patients. We have a migration that was supposed to create those patient contacts for every registration. At the same time that migration was added, we switched the way we search for patient_ids for duplicates, from registration searches to contact searches. 
I'm suspecting the migration did not finish, given the missing contacts.

dianabarsan:
is it possible that the initially created contacts were accidentally deleted?

dianabarsan:
I'm asking because I found a couple of issues that talk about deleting training data on this instance.
For example this issue requests a rollback because real data was deleted by mistake, along with training data: https://github.com/medic/medic-projects/issues/3481 . This issue is from February 2018.

dianabarsan:
Looking at these projects can we: 
[...]

PR #8616, opened 2023-10-05
fix(#7670): do not cache generated patient ids

garethbowen:
# Description

This is the minimal fix... I resisted refactoring more to keep the backport easy and safe.

The DB call is quick against a warm view, so I think this is safe to do. If we find it's a limitation we can cache it in a way that's safe in a distributed environment.

An alternative approach would be to keep the ID cache but still check against the DB every time, which would save regenerating IDs every call, but that's the fast bit so I didn't think that optimisation was necessary.

Because we're going to request to the DB every request instead of every 100 requests the chance of finding 0 unused IDs has increased. This means the length of the ID would be increased more frequently which is bad for the user. To compensate I raised the generation size to 200. It's theoretically possible the length will get bumped on the nth request (where n is the size of the set) if it happens to generate random IDs that are already in use. I would prefer this was changed to be deterministic, and the length was bumped only when the total used IDs reached some percentage of available IDs, but I've left this as an exercise for the future.

medic/cht-core#7670

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7670-reduce-duplicate-ids/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7670-reduce-duplicate-ids/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7670-reduce-duplicate-ids/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/shared-libs/transitions/src/lib/ids.js
+++ b/shared-libs/transitions/src/lib/ids.js
@@ -5,7 +5,7 @@ const ID_LENGTH_PARAM = 'current_length';
 const MIN_ID_LENGTH = 5;
 const INITIAL_ID_LENGTH = MIN_ID_LENGTH;
 const MAX_ID_LENGTH = 13;
-const MAX_IDS_TO_CACHE = 100;
+const MAX_IDS_TO_GENERATE = 200;
 
 /*
 Is not used to actually directly check ID validity: instead, it introduces an inherent
@@ -43,7 +43,7 @@ const generateId = length => {
   }
 
   if (length < MIN_ID_LENGTH) {
-    // This minimum is mostly arbitrary. However, MAX_IDS_TO_CACHE is somewhat
+    // This minimum is mostly arbitrary. However, MAX_IDS_TO_GENERATE is somewhat
     // based on what is good for lengths of 5-7 digits (ie realistic default
     // lengths). If we want shorter IDs in the future we'd need to re-calcuate
     // this value, or potentially make it a function of current id length.
@@ -99,15 +99,19 @@ const putIdLengthDoc = (db, idLengthDoc) => {
 /*
  * Given a collection of ids return an array of those not used already
  */
-const findUnusedIds = (db, keys) => {
+const findUnusedId = (db, keys) => {
   return db.medic
     .query('medic/docs_by_shortcode', { keys })
     .then(results => {
-      const uniqueIds = new Set(keys);
+      if (results.rows.length === keys.length) {
+        // there are no unused keys in this batch
+        return;
+      }
+      const used = new Set();
       results.rows.forEach(row => {
-        uniqueIds.delete(row.key);
+        used.add(row.key);
       });
-      return uniqueIds;
+      return keys.find(key => !used.has(key));
     });
 };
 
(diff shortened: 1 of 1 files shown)


dianabarsan (commented):
This is cool. I have a few thoughts: 
- I'm uncertain about generating 200 ids instead of 100. 
- can we globalize the "used" set so we don't end up re-checking the same id over and over when we're close to eliminating all ids from one "range"? 
- how come this didn't require any change in unit tests?

garethbowen:
> I'm uncertain about generating 200 ids instead of 100.

What's concerning you? Just because it'll take twice as long in CPU time or the query will take longer? I was worried that the length would get bumped too often otherwise. Do you think I should take this further and implement the formula that works out the ratio between used and total IDs and bumps when the ratio gets too high instead of relying on bad luck?

> can we globalize the "used" set so we don't end up re-checking the same id over and over when we're close to eliminating all ids from one "range"?

With a long ID length that means storing tens of thousands of IDs in memory instead of relying on couchdb to cache them correctly. Doing that won't save us any DB lookups (we will still do exactly 1) so the benefit is just that we won't increment the length as early because we're less likely to get 200 used IDs in one batch. If we did ratio suggested above the cache would make more sense because then we would recursively check the db until we found an unused ID. Do you think I should do both?

> how come this didn't require any change in unit tests?

Yeah that concerned me too, so I broke some of the unit tests to prove they were actually running. Basically the unit tests were never checking the caching function, ie there was no test like: `getId(); getId(); db.callCount.equals(1);`

dianabarsan:
> Do you think I should take this further and implement the formula that works out the ratio between used and total IDs and bumps when the ratio gets too high instead of relying on bad luck?

I'd be curious to see what happens when the number of "free" shortcodes closes to 0. I'm worried about requesting with many keys, I think 200 is not extreme though. If we can get an easy formula that reduces lookups, I'm down! 

Re unit tests: can we do a unit test that does the above? And tests the rest of the caching?

garethbowen:
> I'd be curious to see what happens when the number of "free" shortcodes closes to 0. I'm worried about requesting with many keys, I think 200 is not extreme though. If we can get an easy formula that reduces lookups, I'm down!

The formula I was thinking of would be for detecting when the density of used patient IDs was getting high and is really simple to implement. This would then be used to decide when to bump the id length. Until this threshold was reached, we would retry lookups until we found a free ID. It would _increase_ lookups, but _reduce_ the average length of ids which is better for users.

> Re unit tests: can we do a unit test that does the above? And tests the rest of the caching?

Because the caching has been removed I don't know what this would test, ie: the test above would not pass on the PR by design.

dianabarsan:
Thanks a lot for explaining. I was worried that when the density of used ids would increase, we would be stuck on a retry cycle for a significant time, before bumping the length, but I see that we are bumping the length immediately after 1 failed try. 
I think we should be fine as is. Sorry for the confusion!

dianabarsan (approved):
🚢

Issue #8627, opened 2023-10-10
Pagination for contacts within the selected contact tab


kennsippell:
**Describe the performance issue**
In this [forum post](https://forum.communityhealthtoolkit.org/t/cht-app-slowing-down/3041/10), performance experiments took ~57seconds to load a place's contact summary but then took another ~55seconds to render the contact list.

**Describe the improvement you'd like**
Consider pagination of the contacts list in the selected contacts tab

**To Reproduce**
Steps to record the performance metrics:
1. Create a place with 5500 contacts under it
1. Load the contact

**Measurements**
Contacts load at 0:07s
Contact summary loads at 0:57s
When I scroll down, I just see white. The list of contacts actually renders at 1:51s

Hamburger is very slow to open
Tab takes ~4 seconds to unload

**Environment**
- Instance: https://itechcares-malawi.dev.medicmobile.org/
- Browser: Chome117
- Client platform: Ubuntu
- App: webapp
- Version: 4.4.1 prerelease

dianabarsan:
I believe we used to have a setting to prevent display of child contacts based on contact type, which _could_ be an option here. 
I think pagination is the best approach, but disabling display of contacts via config can be a short-term solution if this is blocking a project. 
Are you expecting all contacts of this type to have so many child contacts?

kennsippell:
I had a hard time reproducing the ~2 minute load times yesterday; I'm mostly seeing ~25second load times now for the same scenario. I saw 2mins at least 4x in a row (not sure what's changed). One thing that did change is that performance improved in an incognito browser - my local extensions may have been part of the 2min load times. Sorry about that.  

I made a very quick PR #8629 to render only the first 50 contacts within any contact-content pane. My goal was just to quickly see what impact pagination might have for the CHT CARES app.  The timing in `contacts-content.component.ts` from `ngOnInit` to firing of the subscribe event in `subscriptionSelectedContactForms` went down from 10-14 seconds to 2-3 seconds. Note this is only the impact of reduced rendering (pagination would also reduce the cost of db interactions).

> I believe we used to have a setting to prevent display of child contacts based on contact type, which could be an option here.

This must have been removed, right? I don't see any permissions or setting related to this. Would be useful!

> Are you expecting all contacts of this type to have so many child contacts?

There is just one like this. It is a worst-case scenario. Disabling display of contacts would work for this case. This is the top-level facility, so really there is no reason to ever select this contact. But it is selected by default on devices with larger screens.

[...]

Issue #8576, opened 2023-09-22
Low performance in Reports tab for users with thousands of places


latin-panda:
**Describe the bug**
Report tab is slow on instances with thousands of places, some numbers:
```
training-2: 72k takes 20s
training: 1.9M takes 2min
```

**To Reproduce**
1. Log in as admin (desktop, browser)
2. Load the reports tab
3. Load the people tab
4. Load the reports tab
On the second load, the reports tab is very slow, maxes out a CPU, and eventually logs in the browser console `Uncaught RangeError: Maximum call stack size exceeded`. 

**Expected behavior**
Should not be slow.

**Environment**
- Instance: chis training instances
- Browser: Firefox & Chrome
- Client platform: MacOS
- App: webapp
- Version: verified in 4.2.0 and 4.4.0, possible also in 3.17

**Additional context**
Angular seems to have infinite recursion from internal calls from `refreshView(tView, lView, templateFn, context)` to refreshEmbeddedViews(lView) and back again.

latin-panda:
For the record, things that we considered; not necessarily we're going to do it exactly like this, but these are ideas that can help improve performance: 
- Work out what's causing the recursion in angular
- Consider pagination or autocomplete search instead of listing all places in a tree
- Delay rendering all filters until the filter button is clicked - we know they're not used much
- Permission to turn filters on/off to online and offline users (admin users will always see filters)

latin-panda:
@garethbowen I've pushed the last commit, and this is ready for handover and get it to the finish line. Thanks a lot!

garethbowen:
I have reproduced this issue on `3.17.x`.

garethbowen:
Testing plan...

- Create 1000 identical places.
- Ensure one has a hierarchy to test expansion
- Use `top` to watch CPU usage for the browser process
- Load the Reports tab, click People, click Reports

Before the change: CPU spikes to 100% for a few seconds
After the change: CPU never gets to 100% and settles back to 0 quickly

Testing matrix...
- Firefox, Chrome, Android webview
- Feature flag for material UI on vs off

kennsippell:
Hey @garethbowen - I'm getting some questions from both CARES app and NSSD app about how much this fix will impact performance for their offline users. CARES users have ~5600 contacts on their device (dunno if that's helpful).  SInce you have this repro setup already - are there some time measurements before/after the fix which would be helpful to answer this? Maybe CPU throttling?

ngaruko:
[...]

PR #8629, opened 2023-10-10
perf(#8627) - simulation of contact-content pagination

kennsippell:
# Description

Related to #8627 but not a fix for it. This is useful for taking some performance measures and since this change may make sense for the CARES app (which selects contacts using barcode scanning) I'm going to float the idea that they use this feature branch so users can return to work pending proper fix.

# Code review checklist

# Compose URLs

* __CHT_CORE_COMPOSE_URL__

# License

### DIFF
--- a/webapp/src/ts/modules/contacts/contacts-content.component.html
+++ b/webapp/src/ts/modules/contacts/contacts-content.component.html
@@ -80,7 +80,7 @@
         </div>
 
         <ul>
-          <ng-container *ngFor="let child of group?.contacts" >
+          <ng-container *ngFor="let child of group?.contacts?.slice(0, 50)" >
             <mm-content-row *ngIf="!child.deceased"
               [id]="child.doc._id"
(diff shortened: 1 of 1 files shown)


dianabarsan:
Can we maybe directly limit the number of contacts we request from the database instead of the number of contacts we display?

kennsippell:
Yes! But I was just quite surprised by [this impact](https://github.com/medic/cht-core/issues/8627#issuecomment-1758170451) from the reduced rendering alone. Does that seem weird to you? I haven't worked with modern Angular much - so I'm struggling to identify if this is a bug in how we use it.

dianabarsan:
> Yes! But I was just quite surprised by https://github.com/medic/cht-core/issues/8627#issuecomment-1758170451 from the reduced rendering alone.

I think we have already seen this with https://github.com/medic/cht-core/issues/8576 (Low performance in Reports tab for users with thousands of places). <3 Angular maybe? I think we can revisit how we build components, but I'm not holding my breath.

Issue #8633, opened 2023-10-11
Allow schedule to start with multiple fields


1yuv:
**Is your feature request related to a problem? Please describe.**
We need to create one schedule based on multiple forms and SMS. And sadly, not all forms and reports create a field at the same level or name. [`start_from` in schedules.json](https://docs.communityhealthtoolkit.org/apps/reference/app-settings/schedules/) currently accepts only one field. Problem ius explained here in the [forum](https://forum.communityhealthtoolkit.org/t/adding-form-field-outside-of-fields-section/2998). This means we need to repeat the same schedule for different forms for which a field name is different. 

**Describe the solution you'd like**
Allow `start_from` to accept multiple date fields and create schedule best on first non-undefined value.

1yuv:
As a solution to this, I am thinking about CHT to allow array inputs on `start_from` in schedules, then extract the first non-undefined value from the array and create schedule from there.

garethbowen:
Another option we discussed in Slack was having a simple transition that can be configured to copy arbitrary fields from one place to another. This would mean the edd (or whatever) could be copied from `fields.edd` to `edd` and then the same property can be used by `start_from` for both SMS and xform submissions.

jkuester:
I am interested in the idea of a simple transition for making arbitrary changes to docs, but I think I would want to see more clear use-cases where this would be valuable before adding this logic to the CHT.  

Supporting an array for `start_from` seems to be a much more limited feature (and so easier to maintain). Without any further context, this is the direction I would lean.

That being said, I feel like I am still missing something here.  @1yuv, I understand that the app forms can only set `fields.lmp_date` and not just `lmp_date` in the report doc.  Is there a reason why the SMS forms cannot set `fields.lmp_date`?  (I have very little experience with SMS forms, so maybe this is a dumb question....)  I am just hesitant to add more logic to the CHT to address something that could be handled by updating a form config!

1yuv:
>Is there a reason why the SMS forms cannot set fields.lmp_date

[...]

PR #8639, opened 2023-10-13
feat(#8633) accept array on schedule.start_from

1yuv:
# Description

Allow start_from on schedule to be configured using arrays as well for fixing  medic/cht-core#8633

# Code review checklist

# Compose URLs

* __CHT_CORE_COMPOSE_URL__

# License


dianabarsan (commented):
Great! Can you also please add an e2e test that has a schedule with multiple fields?

shared-libs/transitions/src/lib/schedules.js
```diff
@@ -97,8 +97,8 @@ module.exports = {
     if (typeof startFrom === 'undefined') {
       startFrom = 'reported_date';
     }
-
-    const docStart = objectPath.get(doc, startFrom);
+    
+    const docStart = Array.isArray(startFrom)? objectPath.coalesce(doc, startFrom): objectPath.get(doc, startFrom);
```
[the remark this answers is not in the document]
1yuv: coalesce expectes input to be an array and will [fail](https://github.com/mariocasciaro/object-path/blob/master/index.js#L242-L245) if string parameter is passed as of object-path 0.11.8.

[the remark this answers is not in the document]
dianabarsan: uf, thanks for checking.

garethbowen (commented):
Awesome!

I think there are other ways to do it but this is a nice simple solution that solves the problem today. If you can add the e2e test I think we can call this done.

shared-libs/transitions/src/lib/schedules.js
```diff
@@ -97,8 +97,8 @@ module.exports = {
     if (typeof startFrom === 'undefined') {
       startFrom = 'reported_date';
     }
-
-    const docStart = objectPath.get(doc, startFrom);
+
```
garethbowen, line 100: Unnecessary whitespace...

```suggestion

```

shared-libs/transitions/test/integration/schedules.js
```diff
@@ -732,6 +732,132 @@ describe('functional schedules', () => {
...
+      from: contact.phone,
+      contact: contact
+    };
+
+    return transition.onMatch({ doc: doc }).then(complete => {
+      assert.equal(complete, true);
+      assert(doc.tasks);
+      assert.equal(doc.tasks && doc.tasks.length, 1);
+      assert(doc.scheduled_tasks);
+      assert.equal(doc.scheduled_tasks && doc.scheduled_tasks.length, 1);
```
garethbowen, line 788: I think you can use the fancy new optional chaining operator here...

```suggestion
      assert.equal(doc.tasks?.length, 1);
      assert(doc.scheduled_tasks);
      assert.equal(doc.scheduled_tasks?.length, 1);
```

garethbowen (approved):
A couple of nitpicks inline but I thin we're basically there. Thanks!

tests/integration/sentinel/schedules/schedules-start-from.spec.js
```diff
@@ -0,0 +1,253 @@
...
+
+        chai.expect(updWithClinic.scheduled_tasks).to.be.ok;
+        chai.expect(updWithClinic.scheduled_tasks).to.have.lengthOf(3);
+
+        chai.expect(updWithClinic.scheduled_tasks[0]).to.deep.nested.include(expectedMessage2('scheduled'));
+        chai.expect(updWithClinic.scheduled_tasks[1]).to.deep.nested.include(expectedMessage3('scheduled'));
+        chai.expect(updWithClinic.scheduled_tasks[2]).to.deep.nested.include(expectedMessage4('scheduled'));
+      });
+  });
+});
```
dianabarsan, line 253: I think we'll need to assert on the due date as well, to make sure that the correct date is picked, not just that the message gets created.

dianabarsan (changes requested):
Great work, loving the e2e test, thanks a lot for adding it. I just have one small request inline.

tests/integration/sentinel/schedules/schedules-start-from.spec.js
1yuv: I've verified that by calculating difference between duedate and start_date and ensuring it matches our offset [here](https://github.com/medic/cht-core/pull/8639/files/aa30f4ed972750cb45adef30fe9a20f2d1edc6e5#diff-223091154c27112e2c6c6930b63c353c7ab2799567fb69f49546fffcccc5a384R243)

dianabarsan: Yea, I mean assert on every scheduled task, maybe included the expectedMessage function?

shared-libs/transitions/test/integration/schedules.js
1yuv: Thanks, addressed.

dianabarsan (approved):
Awesome! Thank you so much for adding an e2e test, I know schedules were not covered at all so this is great addition to see.

tests/integration/sentinel/schedules/schedules-start-from.spec.js
lorerod, line 1: I was thinking about edge cases or what could go wrong. What happens if we define start_from as an array of more than two elements? What happens if every field in the start_from array is unavailable?
Does this make sense?

lorerod (approved):
Thank you, @1yuv, for this change and for adding an integration test to cover the change. This PR looks excellent!
I left a comment if it makes sense to implement an edge case. But it is definitely not a blocker. 
I couldn't find any issue on cht-docs to update the schedule documentation and include this new feature. I can create one and work with you on the change.
Let me know what you think.

tests/integration/sentinel/schedules/schedules-start-from.spec.js
1yuv: Hi @lorerod , sure we can. Some of these were already in integraiton tests. I've added more tests and also added comment and made variable names more explanatory.

1yuv:
Hi @lorerod , medic/cht-docs#1207 is the releated issue for documentation update and has been approved on medic/cht-docs#1208. Also, this change required cht-conf to be updated. Related issue for cht-conf was created [here](https://github.com/medic/cht-conf/issues/578) and fixed by this [PR](https://github.com/medic/cht-conf/pull/579).

Issue #8159, opened 2023-03-30
Minimum browser requirements for cht-core 4


kennsippell:
We should fail loudly if the minimum browser requirement is not met. In the same way we are clear on the minimum android os version, we should be clear on the minimum chrome version (and other requirements).

_Originally posted by @derickl in https://github.com/medic/cht-core/issues/8143#issuecomment-1489200258_

garethbowen:
I think failing loudly is a good idea. If we know the app will not work (as in the linked issue) then the user should not be able to log in, or if they are logged in, the app should be locked. The idea would be to make it so obvious to whoever is setting up the device that it's caught early before the phone is given to the CHW.

We could also have a warning stage for if the browser is quite old but technically still supported, but I doubt this will prompt any action from the users so it's likely just nagging. This would be an interesting experiment to run to see how often this prompts user action, but probably outside the scope of this issue.

The reason this isn't a quick win is because currently the testing is all done on recent browsers - this is a technical limitation of our test setup not by design. So if we officially support Chrome X then we need to fully test Chrome X before release. This is a good thing, but it's worth highlighting the current state.

Steps I can see for the implementation of this...

1. Decide on what the minimum browser version should be, which is more permissive than the current documented support.  At the same time decide whether the minimum will be a rolling version (last x years of browser), or whether it can only be changed in a breaking version requiring a major release.
2. Work out how to run our e2e test suite on that version
3. Change the current warning to block login completely, and add an in-app screen to block using the app (may be a duplicate of #7770 )

kennsippell:
If the minimum browser changes, will that be labelled appropriately as a CHT core breaking change?

kennsippell:
> whether it can only be changed in a breaking version requiring a major release

Yes please!  This will really maintain clarity in our upgrade process.

garethbowen:
In #8456 it became clear that we need to run browser checking code before attempting to load the app as the Android app bypasses the login page altogether and if the main bundle crashes on an older browser the user is stuck with no useful info.

jkuester:
@lorerod should this issue be closed now because of https://github.com/medic/cht-core/pull/8663?

lorerod:
@jkuester I'm using this issue as a placeholder.  
https://github.com/medic/cht-core/pull/8663 only runs standard config e2e tests on the minimum browser version. But then the standard config was deleted. 
The next step is to resume this work and run default config e2e tests on the minimum browser version. I will create a separate PR for that work.

[...]

PR #8663, opened 2023-10-24
chore(#8159): Run our e2e standard test on minimum browser requirements

lorerod:
# Description

Use WebdriverIO Chromedriver service to run **standard config** e2e tests on minimum browser version, currently set at 90. Organize with a separate config and add command to CI pipeline.

medic/cht-core#8159

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8159-minimum-browser-requirements-chromedriver-standard/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8159-minimum-browser-requirements-chromedriver-standard/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8159-minimum-browser-requirements-chromedriver-standard/docker-compose/cht-couchdb-clustered.yml)

# License


lorerod:
Hi reviewers! I have divided the work on [this](https://github.com/medic/cht-core/issues/8159) issue into different pull requests, one per config/suite. This will make the work to review and merge gradual and manageable. This PR is only for standard config specs. Let me know what you think. Thanks!

api/package-lock.json
```diff
@@ -15,11 +15,13 @@
         "bowser": "^2.11.0",
         "buffer-shims": "^1.0.0",
         "compression": "^1.7.4",
+        "cron-validator": "^1.3.1",
         "express": "^4.18.2",
         "google-libphonenumber": "^3.2.33",
         "gsm": "^0.1.4",
         "helmet": "^3.23.3",
         "http-proxy": "^1.18.1",
+        "later": "^1.2.0",
```
garethbowen, line 24: These have been added to the package lock but not the package? Should this file be reverted?

package.json
```diff
@@ -96,6 +98,7 @@
     "chai-as-promised": "^7.1.1",
     "chai-exclude": "^2.1.0",
     "chai-shallow-deep-equal": "^1.4.6",
+    "chromedriver": "^90.0.1",
```
garethbowen, line 101: I wonder if it would be better to leave this out of the package.json and just `npm install chromedriver@^90.0.1` somewhere in the npm run command. This would mean it only gets downloaded if you're actually running the chrome 90 test.

.github/workflows/build.yml
```diff
@@ -96,7 +96,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        cmd: ['ci-webdriver-standard', 'ci-e2e-integration', 'ci-webdriver-default-mobile']
+        cmd: ['ci-webdriver-standard', 'ci-e2e-integration', 'ci-webdriver-default-mobile', 'ci-webdriver-standard-minimum-chrome']
```
garethbowen, line 99: I think this will cause a lot of duplication once we have all of the other suites also running Chrome 90. Ideally I think we'd have a matrix for the browser versions ([docs](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs)), eg:

```suggestion
        cmd: ['ci-webdriver-standard', 'ci-e2e-integration', 'ci-webdriver-default-mobile'],
        chrome-version: ['90', 'latest']
```

Then pass the chrome version as an environment variable and use that everywhere. This makes it easy to roll out the other test suites without needing a whole new npm run command.

This causes a problem in the short term because it'll automatically run all 3 commands for both chrome versions (6 builds total), and you only have one working for now, but I think you can add something in the suite to skip for unsupported chrome versions until you get to the PR to fix it.

garethbowen (changes requested):
Really exciting stuff.

A few suggestions inline.

.github/workflows/build.yml
lorerod: Yes! I agree. At first, I wanted to implement this using the same wdio.conf.js and add an extra capability for Chrome 90. But I couldn't make it work because we already use Devtools as a service for Chrome latest, and when adding Chromedriver to the list of services, the first failed. So I have to have a separate conf for Chromedriver.
I will use an env variable to choose between wdio.conf.js files. Does this make sense?

garethbowen: I think you should be able to include a switch in the wdio.conf.js to include devtools _or_ chromedriver. For prior art look at how we're switching "headless" based on the environment variable for "DEBUG": https://github.com/medic/cht-core/blob/master/tests/e2e/wdio.conf.js#L86

Would this work here too?

.github/workflows/build.yml
```diff
@@ -109,7 +111,13 @@ jobs:
...
           - cmd: ci-webdriver-default
             suite: workflows
-
+        exclude: # temporary until all suites run on Chrome 90
+          - cmd: ci-e2e-integration
+            chrome-version: 90
+          - cmd: ci-webdriver-default-mobile
+            chrome-version: 90
+          - cmd: ci-webdriver-default
+            chrome-version: 90
```
garethbowen, line 120: Nice!

.github/workflows/build.yml
```diff
@@ -169,9 +177,17 @@ jobs:
...
         cp -r build-history-data/cht-core/allure/${{ matrix.cmd }}-${{ matrix.suite }}/allure-report/history allure-results | true
-    - name: Run tests ${{ matrix.cmd }}-${{ matrix.suite }}
+    - name: Install Chrome 90
+      if: ${{matrix.chrome-version=='90'}}
+      run: |
+        wget -O "google-chrome-stable_current_amd64.deb" https://www.slimjet.com/chrome/download-chrome.php?file=files/90.0.4430.72/google-chrome-stable_current_amd64.deb
+        sudo dpkg -i "google-chrome-stable_current_amd64.deb"
+    - name: Install Chromedriver 90
+      if: ${{matrix.chrome-version=='90'}}
+      run: npm install chromedriver@^90.0.1
```
garethbowen, line 187: Let's merge these two together, ie:

```suggestion
    - name: Install Chrome 90
      if: ${{matrix.chrome-version=='90'}}
      run: |
        wget -O "google-chrome-stable_current_amd64.deb" https://www.slimjet.com/chrome/download-chrome.php?file=files/90.0.4430.72/google-chrome-stable_current_amd64.deb
        sudo dpkg -i "google-chrome-stable_current_amd64.deb"
        npm install chromedriver@^90.0.1
```

tests/e2e/wdio.conf.js
```diff
@@ -83,10 +84,13 @@ const baseConfig = {
...
     //
     browserName: 'chrome',
+    browserVersion: CHROME_VERSION,
     acceptInsecureCerts: true,
     'goog:chromeOptions': {
-      args: DEBUG ? ['disable-gpu', 'deny-permission-prompts', 'ignore-certificate-errors']: 
-        ['headless', 'disable-gpu', 'deny-permission-prompts', 'ignore-certificate-errors']
+      args: DEBUG ? ['disable-gpu', 'deny-permission-prompts', 'ignore-certificate-errors'] :
+        ['headless', 'disable-gpu', 'deny-permission-prompts', 'ignore-certificate-errors', 'no-sandbox',
+          'window-size=1200,900'],
```
garethbowen, line 92: This is a bit complicated now with the additional options, particularly because most of the options are the same. Consider pulling out a function or constants to hold these, eg:

```js
const CHROME_OPTIONS = [ ... ];
const CHROME_OPTIONS_DEBUG = CHROME_OPTIONS.concat([ ... ]);
```

.github/workflows/build.yml
```diff
@@ -92,11 +92,13 @@ jobs:
     runs-on: ubuntu-22.04
     env:
       NODE: 16.x
+      CHROME_VERSION: ${{ matrix.chrome-version }}
```
garethbowen, line 95: Not this line, but can we append `${{ matrix.chrome-version }}` to the test name, otherwise CI runs two builds with the same name. Perhaps we could just append it if it's not 90? For example, I don't know if this is value syntax but something like...

```
name: ${{ matrix.cmd }}-${{ matrix.suite || '' }}${matrix.chrome-version === '90' ? '-minimum-browser' : '' }
```

garethbowen (changes requested):
Yes this is awesome! Minor suggestions inline...

tests/e2e/standard/enketo/pregnancy.wdio-spec.js
```diff
@@ -125,7 +125,9 @@ describe('New pregnancy', () => {
   });
 
   it('Verify the targets page', async () => {
+    await browser.url('/');
     await commonPage.logout();
+    await browser.url('/');
```
garethbowen, line 130: I'm surprised this is needed. Shouldn't logout do that already?

tests/e2e/wdio.conf.js
```diff
@@ -17,6 +17,16 @@ let testTile;
 const DEBUG = process.env.DEBUG;
 const DEFAULT_TIMEOUT = 2 * 60 * 1000;
 const DEBUG_TIMEOUT = 10 * 60 * 1000; //timeout in debug mode, allows more interaction with browser after test
+const CHROME_VERSION = process.env.CHROME_VERSION;
+const CHROME_OPTIONS_ARGS_DEBUG = [
+  'disable-gpu',
+  'deny-permission-prompts',
+  'ignore-certificate-errors',
+  'no-sandbox',
```
garethbowen, line 25: Is this necessary? We didn't need it before and it means we're further away from real-world settings. I couldn't find good docs on why it would be necessary for us...

tests/page-objects/default/contacts/contacts.wdio.page.js
```diff
@@ -181,13 +181,21 @@ const addPerson = async ({
...
   await (await personName()).click(); // blur the datepicker field so the sex field is visible
+  await (await phoneField()).waitForDisplayed();
   await (await phoneField()).addValue(phoneValue);
+  await (await sexField(type, sexValue)).waitForClickable();
   await (await sexField(type, sexValue)).click();
+  await (await roleField(type, roleValue)).waitForClickable();
   await (await roleField(type, roleValue)).click();
+  await (await externalIdField(type)).waitForDisplayed();
   await (await externalIdField(type)).addValue(externalIDValue);
+  await (await notes(type)).waitForDisplayed();
```
garethbowen, line 198: This seems like overkill. Once the whole page is displayed why do we need to wait for each field to be displayed?

tests/page-objects/standard/enketo/delivery.wdio.page.js
```diff
@@ -1,6 +1,7 @@
-const OUTCOME = {liveBirth: 'healthy', stillBirth: 'still_birth', miscarriage: 'miscarriage'};
-const LOCATION = {facility: 'f', homeAttendant: 's', homeNoAttendant: 'ns'};
-
+const OUTCOME = { liveBirth: 'healthy', stillBirth: 'still_birth', miscarriage: 'miscarriage' };
+const LOCATION = { facility: 'f', homeAttendant: 's', homeNoAttendant: 'ns' };
+const MINIMUM_BROWSER_VERSION = '90';
```
garethbowen, line 3: Instead of copying this value which will make life harder when we bump it, let's expose a common function somewhere for `isTestingLegacyChrome` or something?

tests/page-objects/standard/enketo/immunization-visit.wdio.page.js
```diff
@@ -17,13 +17,16 @@ const TYPHOID_VACCINE = 'input[name="/immunization_visit/group_typhoid/g_typhoid
 const VITAMIN_A_VACCINE = 'input[name="/immunization_visit/group_vitamin_a/g_vitamin_a"]';
 const YELLOW_FEVER_VACCINE = 'input[name="/immunization_visit/group_yellow_fever/g_yellow_fever"]';
 
+const SUMMARY_SECTION = 'section[name="/immunization_visit/group_review"]';
+const ACTIVE_SPAN = 'span.active';
```
garethbowen, line 21: This is duplicated in the `delivery.wdio.page.js` - can we pull out a common `enketo.wdio.page.js` or something to have code that is common to all enketo forms?

garethbowen (changes requested):
This is starting to look really clean, thanks!

A few minor suggestions inline.

tests/e2e/wdio.conf.js
lorerod: I faced a problem while working with the default configuration and using Chromedriver instead of Devtools. The tests wouldn't start and kept showing `data:, ` instead of redirecting to `baseUrl` as expected. Thanks for pointing out this. I re-ran the standard configuration tests without the `no-sandbox` option, and they worked perfectly fine.

lorerod: The error I mentioned above appears on the [CI](https://github.com/medic/cht-core/actions/runs/6723769390/job/18276229947?pr=8663), but on my local machine, everything is working correctly.

`ERROR webdriver: session not created: session not created from tab crashed`

garethbowen:
@lorerod Also it looks like there's a conflict that will have to be fixed before this can be merged...

tests/page-objects/standard/enketo/immunization-visit.wdio.page.js
lorerod: In delivery, it is different. The value is `.option-label.active`.
I will check for other common elements and create an enketo.wdio.page.js file that contains them.

tests/e2e/wdio.conf.js
lorerod: I will investigate further how to fix this without adding the extra option. [This](https://stackoverflow.com/questions/59186984/selenium-common-exceptions-sessionnotcreatedexception-message-session-not-crea) is where I read about the `no-sandbox` option.

lorerod: I couldn't fix this without adding a `no-sandbox `option to arguments. To minimize impact, I only added when running tests with Chrome 90.

tests/utils/index.js
```diff
@@ -1250,6 +1251,18 @@ const logFeedbackDocs = async (test) => {
   return true;
 };
 
+const getMinimumChromeVersion = () => {
+  return MINIMUM_BROWSER_VERSION;
+};
+
+const isMinimumChromeVersion = () => {
+  // eslint-disable-next-line no-undef
+  if (driver.capabilities.browserVersion.split('.').shift() === MINIMUM_BROWSER_VERSION) {
```
garethbowen, line 1260: There's a little complexity here with all the different browser versions... we have:

- `driver.capabilities.browserVersion`
- `MINIMUM_BROWSER_VERSION`
- `process.env.CHROME_VERSION`

I think it would be easier if we just don't use `driver.capabilities.browserVersion` if we can, for example, use the env variable here too, eg:

```suggestion
  if (process.env.CHROME_VERSION === MINIMUM_BROWSER_VERSION) {
```

That way it's more consistent throughout the code.

This also means throughout the `tests/e2e/wdio.conf.js` file you can use `isMinimumChromeVersion` rather than `getMinimumChromeVersion` which will simplify the code there a little too.

This might mean you can remove the `getMinimumChromeVersion` function altogether...

What do you think?

tests/page-objects/standard/enketo/enketo.wdio.page.js
```diff
@@ -0,0 +1,15 @@
...
+const FOLLOW_UP_SMS = (formId) => `span[data-value=" /${formId}/chw_sms "]`;
+
+module.exports = {
+  FORM,
+  ACTIVE,
+  ACTIVE_SPAN,
+  ACTIVE_OPTION_LABEL,
+  SMS_NOTE,
+  FOLLOW_UP_SMS,
+};
```
garethbowen, line 15: Nice!

Just one point about naming conventions. Usually we use all caps for primitive constants, not functions, so I think `FORM`, `SMS_NOTE` and `FOLLOW_UP_SMS` should be camel case.

garethbowen (changes requested):
Great work, thanks!

Some minor suggestions inline.

garethbowen (approved):
This is so great. Thanks for putting up with all my nitpicks!

lorerod:
Thank you, Gareth, for caring and all the detailed suggestions.

Issue #6530, opened 2020-07-09
Add rate limiting to authentication endpoints


garethbowen:
Currently it's possible for a malicious actor to make unlimited attempts at logging in until correct credentials are found. This sort of brute force attack should be limited by rate limiting failed authentication requests. This goes for both logging in through the login form, as well as basic authentication via a range of URLs.

Denial of service style attacks are out of scope for this issue.

MaxDiz:
bumping to 3.13 based on discussion with @garethbowen

craig-landry:
I've asked @latin-panda to look into this one.

garethbowen:
I think this got removed from 3.13 when we changed from projects to milestones, so don't read too much into that!

There's some discussion in the private issue about how to go about this. As well as looking into JS libraries we should look at our cloud hosting providers to see if they have something that will just work. We would then need to document how to set it up for our self-hosting partners. Reach out to SRE to discuss options here.

My main concern is about an attacker brute forcing credentials, rather than DOSing the server, so focus on the login and basic auth pipelines and we should be good.

latin-panda:
I've continued the discussion in the other ticket mentioned above.

garethbowen:
@ralfudx This is the [rate limiting PR](https://github.com/medic/cht-core/pull/8675) I was talking about in today's call.

The idea is to stop attackers being able to have unlimited tries at finding a correct username and password.

Basically the idea is that it should block any attacker that is...
1. trying a bunch of passwords for the same user,
2. trying a bunch of users with the same password, or
3. trying a bunch of combinations from the same IP.

I've set it to allow up to 10 tries every 10 seconds so I hoping actual users will never hit the limit.
I'm only blocking authentication requests, so you can still access things like static resources after you've been blocked, but you can't log in until the 10 seconds is up.
The CHT supports logging in via the login page (form submitted to the login api), AND basic auth and it's important that both are protected.

This change does not protect against denial of service type attacks where someone just overloads the system - that's out of scope for now.

Any help with QA to ensure that all attack routes are blocked, and no real users are blocked, would be much appreciated!

ralfudx:
@garethbowen I'm still having having issues with my local env set-up... I've reached out a couple engineers for this (@njuguna-n @Benmuiruri ) once this is resolved i'll proceed with testing this...

garethbowen:
[...]

PR #8675, opened 2023-11-01
feat(#6530): add rate limiting for authentication requests

garethbowen:
# Description

[description]

medic/cht-core#6530

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6530-rate-limiting/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6530-rate-limiting/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6530-rate-limiting/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/api/src/controllers/login.js
+++ b/api/src/controllers/login.js
@@ -298,20 +298,16 @@ module.exports = {
     }
     return createSession(req)
       .then(sessionRes => {
-        if (sessionRes.statusCode !== 200) {
-          res.status(sessionRes.statusCode).json({ error: 'Not logged in' });
-          return;
+        if (sessionRes.statusCode === 200) {
+          return setCookies(req, res, sessionRes)
+            .then(redirectUrl => res.status(302).send(redirectUrl));
         }
-        return setCookies(req, res, sessionRes)
-          .then(redirectUrl => res.status(302).send(redirectUrl))
-          .catch(err => {
-            if (err.status === 401) {
-              return res.status(err.status).json({ error: err.error });
-            }
-            throw err;
-          });
+        res.status(sessionRes.statusCode).json({ error: 'Not logged in' });
       })
       .catch(err => {
+        if (err.status === 401) {
+          return res.status(401).json({ error: err.error });
+        }
         logger.error('Error logging in: %o', err);
         res.status(500).json({ error: 'Unexpected error logging in' });
       });
--- a/api/src/controllers/login.js
+++ b/api/src/controllers/login.js
@@ -296,21 +296,21 @@ module.exports = {
     if (limited) {
       return serverUtils.rateLimited(req, res);
     }
-    return createSession(req)
-      .then(sessionRes => {
-        if (sessionRes.statusCode === 200) {
-          return setCookies(req, res, sessionRes)
-            .then(redirectUrl => res.status(302).send(redirectUrl));
-        }
+    try {
+      const sessionRes = await createSession(req);
+      if (sessionRes.statusCode !== 200) {
         res.status(sessionRes.statusCode).json({ error: 'Not logged in' });
-      })
-      .catch(err => {
-        if (err.status === 401) {
-          return res.status(401).json({ error: err.error });
-        }
-        logger.error('Error logging in: %o', err);
-        res.status(500).json({ error: 'Unexpected error logging in' });
-      });
+      } else {
+        const redirectUrl = await setCookies(req, res, sessionRes);
+        res.status(302).send(redirectUrl);
+      }
+    } catch (e) {
+      if (e.status === 401) {
+        return res.status(401).json({ error: e.error });
+      }
+      logger.error('Error logging in: %o', e);
+      res.status(500).json({ error: 'Unexpected error logging in' });
+    }
   },
   getIdentity: (req, res) => {
     res.type('application/json');
--- a/api/src/controllers/login.js
+++ b/api/src/controllers/login.js
@@ -14,6 +14,8 @@ const cookie = require('../services/cookie');
 const brandingService = require('../services/branding');
 const translations = require('../translations');
 const template = require('../services/template');
+const rateLimitService = require('../services/rate-limit');
+const serverUtils = require('../server-utils');
 
 const templates = {
   login: {
@@ -269,6 +271,10 @@ const renderLogin = (req) => {
   return render('login', req);
 };
 
+const isRateLimited = async req => {
+  return rateLimitService.isLimited([ req.body.user, req.body.password ]);
+};
+
 module.exports = {
   renderLogin,
 
@@ -285,7 +291,11 @@ module.exports = {
       })
       .catch(next);
   },
-  post: (req, res) => {
+  post: async (req, res) => {
+    const limited = await isRateLimited(req);
+    if (limited) {
+      return serverUtils.rateLimited(req, res);
+    }
     return createSession(req)
       .then(sessionRes => {
         if (sessionRes.statusCode !== 200) {
--- a/api/src/controllers/login.js
+++ b/api/src/controllers/login.js
@@ -271,10 +271,6 @@ const renderLogin = (req) => {
   return render('login', req);
 };
 
-const isRateLimited = async req => {
-  return rateLimitService.isLimited([ req.ip, req.body?.user, req.body?.password ]);
-};
-
 const login = async (req, res) => {
   try {
     const sessionRes = await createSession(req);
@@ -310,7 +306,7 @@ module.exports = {
       .catch(next);
   },
   post: async (req, res) => {
-    const limited = await isRateLimited(req);
+    const limited = await rateLimitService.isLimited(req);
     if (limited) {
       return serverUtils.rateLimited(req, res);
     }
@@ -332,7 +328,7 @@ module.exports = {
 
   tokenGet: (req, res, next) => renderTokenLogin(req, res).catch(next),
   tokenPost: async (req, res, next) => {
-    const limited = await isRateLimited(req);
+    const limited = await rateLimitService.isLimited(req);
     if (limited) {
       return serverUtils.rateLimited(req, res);
     }
--- a/api/src/controllers/login.js
+++ b/api/src/controllers/login.js
@@ -275,6 +275,24 @@ const isRateLimited = async req => {
   return rateLimitService.isLimited([ req.body.user, req.body.password ]);
 };
 
+const login = async (req, res) => {
+  try {
+    const sessionRes = await createSession(req);
+    if (sessionRes.statusCode !== 200) {
+      res.status(sessionRes.statusCode).json({ error: 'Not logged in' });
+    } else {
+      const redirectUrl = await setCookies(req, res, sessionRes);
+      res.status(302).send(redirectUrl);
+    }
+  } catch (e) {
+    if (e.status === 401) {
+      return res.status(401).json({ error: e.error });
+    }
+    logger.error('Error logging in: %o', e);
+    res.status(500).json({ error: 'Unexpected error logging in' });
+  }
+};
+
 module.exports = {
   renderLogin,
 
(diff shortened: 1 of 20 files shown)


api/src/controllers/login.js
```diff
@@ -285,23 +291,23 @@ module.exports = {
...
-              return res.status(err.status).json({ error: err.error });
-            }
-            throw err;
-          });
+        res.status(sessionRes.statusCode).json({ error: 'Not logged in' });
       })
       .catch(err => {
+        if (err.status === 401) {
+          return res.status(401).json({ error: err.error });
+        }
```
garethbowen, line 310: Some refactoring needed to make sonar happy..

api/src/middleware/rate-limiter.js
```diff
@@ -0,0 +1,42 @@
+const auth = require('../auth');
+const rateLimitService = require('../services/rate-limit');
+const serverUtils = require('../server-utils');
+
+const getBasicAuthUsername = (req) => auth.basicAuthCredentials(req);
+
+const consume = (req, res) => {
+  res.on('finish', () => {
+    if (res.statusCode === 401 || res.statusCode === 429) {
+      rateLimitService.consume(getKeys(req));
```
garethbowen, line 10: Only consume after an unsuccessful login attempt. This won't protect against DoS style attacks, the idea is to protect the user's password. DoS would require a much more sophisticated process...

api/src/middleware/rate-limiter.js
```diff
@@ -0,0 +1,42 @@
...
+  ];
+
+  const basicAuth = getBasicAuthUsername(req);
+  if (basicAuth) {
+    keys.push(basicAuth.username);
+    keys.push(basicAuth.password);
+  }
+
+  return keys;
+};
```
garethbowen, line 29: The keys I've gone with are
- requesting IP address so a single computer can't spam login attempts,
- the username so if there is a target user or known user the attacker can't spam all combinations of possible passwords, and
- the password so the attacker can't spam common passwords against all existing users.

This is still vulnerable to distributed attacks, for a range of user/pass combinations.

shared-libs/validation/package.json
```diff
@@ -9,6 +9,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@medic/message-utils": "file:../message-utils",
+    "cron-validator": "^1.3.1",
```
garethbowen, line 12: This was missing which caused problems when running locally.

tests/e2e/default/login/login-logout.wdio-spec.js
```diff
@@ -4,12 +4,18 @@ const modalPage = require('@page-objects/default/common/modal.wdio.page');
...
+let brandingDoc;
+
 describe('Login page funcionality tests', () => {
   const auth = {
     username: constants.USERNAME,
     password: constants.PASSWORD
   };
 
+  before(async () => {
+    brandingDoc = await utils.getDoc('branding');
```
garethbowen, line 16: I had to move this here before logout.

tests/integration/api/controllers/all-docs.spec.js
```diff
@@ -299,7 +299,6 @@ describe('all_docs handler', () => {
         ))
       ]))
       .then(([excludeDocs, includeDocs]) => {
-        console.log(JSON.stringify(excludeDocs, null, 2));
```
garethbowen, line 302: Unrelated log spam.

tests/integration/api/controllers/db-doc.spec.js
```diff
@@ -786,7 +786,6 @@ describe('db-doc handler', () => {
               if (reportScenarios[idx].allowed) {
                 chai.expect(result).to.deep.include(reportScenarios[idx].doc);
               } else {
-                console.log(idx, result);
```
garethbowen, line 789: Unrelated log spam.

tests/integration/api/rate-limit.spec.js
```diff
@@ -0,0 +1,34 @@
...
+
+    try {
+      await requestThat401s();
+      expect.fail('should have been rate limited');
+    } catch (e) {
+      expect(e.statusCode).to.equal(429);
+    }
+
+    await new Promise((resolve) => {
+      setTimeout(resolve, 10000);
```
garethbowen, line 22: I'm not a fan of delaying the test by 10 seconds... not sure if there's a better way...

tests/integration/api/routing.spec.js
```diff
@@ -85,140 +82,82 @@ describe('routing', () => {
...
+        { path: '/api/v1/contacts-by-phone' },
+        { path: '/api/v1/contacts-by-phone', method: 'POST' },
+        { path: '/api/v1/hydrate' },
+        { path: '/api/v1/hydrate', method: 'POST' },
+        { path: '/api/v1/forms' },
+      ];
+
+      afterEach(done => {
+        // delay for 1 second to avoid hitting API rate limit
+        setTimeout(done, 1000);
```
garethbowen, line 131: I don't like this solution much but it does work...

api/src/controllers/login.js
```diff
@@ -285,23 +291,23 @@ module.exports = {
       })
       .catch(next);
   },
-  post: (req, res) => {
+  post: async (req, res) => {
+    const limited = await isRateLimited(req);
+    if (limited) {
+      return serverUtils.rateLimited(req, res);
+    }
```
dianabarsan, line 298: Is this not automatically handled by the middleware?

dianabarsan (commented):
This is great stuff! Love the tests too! 
I only left a couple of questions inline.

api/src/routing.js
```diff
@@ -162,6 +163,7 @@ app.use(
     ':res[content-length] :response-time ms'
   )
 );
+app.use(rateLimiterMiddleware);
```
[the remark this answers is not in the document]
garethbowen: Right. Because you can basic auth to any endpoint we have to watch everything. But only 401s and 429s end up consuming points.

The downside is a small tax on every request. At the moment it's an in memory lookup by key so it's supposed to be really fast.

An alternative would be to only do the lookup if someone is logging in either with basic auth or with the login api but quickly skip for every other request. This would have a slight performance benefit, but also allow things like static resources to be returned even when you've hit your rate limit.

What do you think?

api/src/middleware/rate-limiter.js
```diff
@@ -0,0 +1,42 @@
...
+
+const rateLimiterMiddleware = (req, res, next) => {
+  return rateLimitService.isLimited(getKeys(req))
```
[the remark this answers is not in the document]
garethbowen: Commented above...

api/src/controllers/login.js
garethbowen: The problem is the body parser doesn't run before the middleware so we don't have access to the user/pass in the JSON body until now...

dianabarsan: oh, bummer! thanks for explaining

api/src/routing.js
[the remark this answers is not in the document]
dianabarsan: I'm not worried about the small tax, I'm worried about requests that should be honored (like static resources, great example) even when maybe rate limited. So I'm leaning to simply skipping the check for requests that don't carry any login information.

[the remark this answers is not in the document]
garethbowen: Yeah I'm maybe leaning that way too :thinking: 

For a laugh you can look at all the early commits on this branch. I went back and forwards so many times about the correct way to do it without breaking legitimate users, and this may be a hangover from a previous more aggressive approach.

I think about it some more and get back to you...

api/src/controllers/login.js
```diff
@@ -285,23 +291,23 @@ module.exports = {
...
     return createSession(req)
```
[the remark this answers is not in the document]
garethbowen: Done.

api/src/routing.js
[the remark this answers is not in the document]
garethbowen: Done.

dianabarsan (approved):
Sweet!

api/src/middleware/rate-limiter.js
```diff
@@ -0,0 +1,46 @@
+const auth = require('../auth');
+const rateLimitService = require('../services/rate-limit');
+const serverUtils = require('../server-utils');
+
+const getKeys = (req, basicAuth) => {
+  const keys = [
+    req.ip,
```
mrjones-plip, line 7: Which IP is this?  Specifically, is this the Nginx IP or the actual end user's IP?  

My concern is if it's the Nginx IP, then all traffic will come from the same IP. During something like a training, when a lot of users could fail to log in, we could repeatedly hit the rate-limit, which will affect ALL traffic.

If it's the end user's IP - all good!

mrjones-plip:
Does this PR offer any logging? I can imagine the feature being hit and then valid logins not being accepted b/c they're being rate-limited but exposing this in any logs would be confusing.

api/src/middleware/rate-limiter.js
mrjones-plip: Hey again! Testing the code, I can see this is correctly the end user's IP - all good!  I'll see myself out the door and resolve the thread - thanks for listening ;)

api/src/server-utils.js
```diff
@@ -103,6 +103,10 @@ module.exports = {
     }
   },
 
+  rateLimited: (req, res) => {
+    respond(req, res, 429, 'Too Many Requests');
```
mrjones-plip, line 107: While it behaves  well enough when I tested it, this is not returning JSON like other failed logins.  I think this an easy fix though - does this achieve what we want or do we need to dive deeper?

```suggestion
    respond(req, res, 429, '{"error" : "Too Many Requests"}');
```

mrjones-plip:
I see there's no explicit server side logging, bet we do log a `429` in API logs instead of a `401`. As well, client side we do 3 things to let the client know:
* change response from "Incorrect user name or password."  to "Unexpected error while logging in. "
* Respond with a `429`instead of a `401`
* In the browser console, output `Error logging in Too Many Requests`

This seems reasonable enough!  I did note a [minor discrepancy](https://github.com/medic/cht-core/pull/8675#discussion_r1386964570) in our response format.

api/src/middleware/rate-limiter.js
garethbowen: Thank you for taking the time to validate it!

garethbowen:
@mrjones-plip 

> change response from "Incorrect user name or password." to "Unexpected error while logging in. "

I _almost_ added code to the login page to give a clearer message on 429 but we don't expect any real person to hit this and it's a pain to add translations to the login page. What do you think? Is it worth the effort?

api/src/server-utils.js
garethbowen: The first line of the `respond` function is...

```
  if (wantsJSON(req)) {
    return writeJSON(res, code, message, details);
  }
```

This tries to respect the `accepts` header of the request so you get what you asked for.

However you have found a bug in the login code! It turns out we were never specifying the "accepts" header. I've changed that and now the error handling code works as expected. Thanks for being so diligent!

mrjones-plip (approved):
nice work! JSON comin' and goin': 

```
➜  docker-helper-4.x git:(master) ✗ ./login.sh
{"error":"Not logged in"}%                                                                   ➜  docker-helper-4.x git:(master) ✗ ./login.sh
{"error":"Not logged in"}%                                                                   ➜  docker-helper-4.x git:(master) ✗ ./login.sh
{"code":429,"error":"Too Many Requests"}%                                                    ➜  docker-helper-4.x git:(master) ✗ ./login.sh
{"code":429,"error":"Too Many Requests"}%  
```

Issue #8433, opened 2023-08-04
Add Enketo telemetry events for contact forms


kennsippell:
**What feature do you want to improve?**
[Telemetry events](https://docs.communityhealthtoolkit.org/apps/guides/performance/telemetry/) with the format `enketo:*` are recorded when users complete app forms or training forms but are not recorded for contact forms.

**Describe the improvement you'd like**
We are investigating a performance problem related to slow save performance for a contact form. It'd be great if there was telemetry data for contact forms.

**Describe alternatives you've considered**
Shrug

**Additional context**
https://github.com/medic/config-nssd/issues/360#issuecomment-1664178996

I just set breakpoints in telemetry-service on latest cht-core master (~4.3) and confirmed that no telemetry event is created when rendering a contact form by clicking the "+" which opens a health_center contact form. I expected to see a `enketo:*:render` event recorded.

latin-panda:
@michaelkohn We found that there are [telemetry fields ](https://docs.communityhealthtoolkit.org/apps/guides/performance/telemetry/) for when loading report forms inside a contact:`enketo:contacts:<form>:add:<component>`. 

This work is about adding telemetry to the contact forms. What do you think if we prefix it like `enketo:contact-form:<form>:<action>:<component>`? To not mix it with the metrics for the contact's report form

cc: @Benmuiruri to continue with the follow up in the topic :)

michaelkohn:
Couple of questions.... I generally see these prefixes on enketo related telemetry records:

`enketo:contacts:*`
`enketo:reports:*`
`enketo:tasks:*`

I always thought that the second section of that (contacts vs. reports vs. tasks) related to which tab the activity occurred on.  So to stay consistent with that, I would have expected the contact forms to follow that paradigm. Even if not, I don't really think we'd need to differentiate the prefix for contact-specific forms since the name of the contact form will clearly identify what it is `enketo:contacts:clinic_create:add:render` (for example). 
 
Having said that... it also makes me wonder how contact forms will be treated in telemetry since we typically have two different forms* for each contact type....  [one for create / one for edit](https://docs.communityhealthtoolkit.org/apps/reference/forms/contact/). Would the `clinic-create` form only ever have an enketo `add` record and the `clinic-edit` form always just be an enketo `edit` record? 

_* It's not entirely clear to me if it is required to have a different edit form or just common / possible._

latin-panda:
> So to stay consistent with that, I would have expected the contact forms to follow that paradigm.

[...]

PR #8676, opened 2023-11-01
chore(#8433) add telemetry events to contact forms

Benmuiruri:
# Description

- Adds[ Telemetry Events ](https://docs.communityhealthtoolkit.org/apps/guides/performance/telemetry/) to record Telemetry data in the contact forms 

### Personal Targets

- Each "card" is separated into two content areas
- The content areas is shaded differently to delineate them
- The labels and images which used to be on the top, are now at the lower content area
- Instead of using red for unmet targets, the cards now use black

#8433  

# Code review checklist

# Compose URLs

- [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:template-issue-update/docker-compose/cht-core.yml)
- [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:template-issue-update/docker-compose/cht-couchdb.yml)
- [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:template-issue-update/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/webapp/src/ts/modules/contacts/contacts-edit.component.ts
+++ b/webapp/src/ts/modules/contacts/contacts-edit.component.ts
@@ -30,7 +30,7 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
     private contactTypesService:ContactTypesService,
     private dbService:DbService,
     private contactSaveService:ContactSaveService,
-    private telemetryService: TelemetryService,
+    private telemetryService:TelemetryService,
     private translateService:TranslateService,
   ) {
     this.globalActions = new GlobalActions(store);
--- a/webapp/src/ts/modules/contacts/contacts-edit.component.ts
+++ b/webapp/src/ts/modules/contacts/contacts-edit.component.ts
@@ -13,6 +13,7 @@ import { ContactSaveService } from '@mm-services/contact-save.service';
 import { Selectors } from '@mm-selectors/index';
 import { GlobalActions } from '@mm-actions/global';
 import { ContactsActions } from '@mm-actions/contacts';
+import { TelemetryService } from '@mm-services/telemetry.service';
 import { TranslateService } from '@mm-services/translate.service';
 
 
@@ -29,6 +30,7 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
     private contactTypesService:ContactTypesService,
     private dbService:DbService,
     private contactSaveService:ContactSaveService,
+    private telemetryService: TelemetryService,
     private translateService:TranslateService,
   ) {
     this.globalActions = new GlobalActions(store);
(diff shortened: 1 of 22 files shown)


webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -29,6 +30,7 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
     private contactTypesService:ContactTypesService,
     private dbService:DbService,
     private contactSaveService:ContactSaveService,
+    private telemetryService: TelemetryService,
```
garethbowen, line 33: Nit: whitespace consistency.

```suggestion
    private telemetryService:TelemetryService,
```

garethbowen (commented):
Very nice @Benmuiruri . Thanks for remembering to update the unit tests too.

I've put a very minor suggestion in there.

Before I approve it though I had two questions I'd like your take on...

1. Did you think about end-to-end testing this to make sure that the whole stack was working together? Do you think unit testing is sufficient for this change?
2. I see a lot of copied code from the reports component. Did you consider creating a new service that both components could call to make the code a little simpler? What about moving the code down into the enketo service? I'm not even sure that would be possible but curious if you explored it...

tatilepizs (commented):
We have been talking, in our QA syncs, about when it is necessary to include e2e test and when we can add only unit tests or integration tests. And we have talked that maybe it will be a good idea to establish a "rule" that every time that we make changes to a file in the webapp folder we should add an e2e new test or a scenario to an existing one.  (we need to put this in a document, we are working on it 😅 )

What do you think?

We can definitely help you in the process, let me know and we can have a call about it 🙂

Benmuiruri:
Hi @garethbowen, 

1. I didn't consider adding e2e tests given how it had been done in the other components, but here are my thoughts about it.

- Since we are interested in ensuring that the data is not only generated but also sent to and correctly handled by the backend, then e2e would be good to add. 
- A unit test may not be sufficient because it does not how multiple components and external services interact, it only allows us to test that we are recording the data. 
- In the unit tests we are mocking several external interactions, arguably that might not cover all possible conditions and edge cases that we can't see in unit tests. 

2. On the issue of duplication, I did realise it was not the DRYest code. It's definitely a good idea to encapsulate the shared logic in a service and inject it to any component that reuses it.  Perhaps think of `Component Inheritance`, where we have a `Base Component`  that other components can extend.... Maybe...

However, I'll be honest and admit at this time, I do not fully understand `CHT`, `Angular` and `Enketo` enough at this time to evaluate the coupling and cohesion and be able to refactor the implementation and the tests. 

I'd certainly welcome the challenge to improve the code health by making the code DRY at a later time when I have a better understanding of the code.

Benmuiruri:
Hi @tatilepizs,  I would love to have a chat about recommended thinking and approaches to testing `cht`. Please set it up, my calendar is flexible 😃

garethbowen (approved):
@Benmuiruri Thanks for the thoughtful response!

Approved as I'm happy with where this is at now, but feel free to continue to improve testing or DRYness as you see fit!

Issue #8674, opened 2023-11-01
Cannot create user for contacts created via Place Api


freddieptf:
**Describe the bug**
I used the [places api](https://docs.communityhealthtoolkit.org/apps/reference/api/#post-apiv1places) to create a place and contact as shown in the docs. 

```json
{
  "name": "County",
  "type": "contact",
  "contact_type": "a_county",
  "contact": {
    "name": "Paul",
    "phone": "+254883720611",
    "type": "contact",
    "contact_type": "person"
  }
} 
```

That worked but i cannot create a user for the created contact

![Screenshot from 2023-11-01 04-41-02](https://github.com/medic/cht-core/assets/10602047/3ad8959e-4c73-49b9-804d-b1d735532c53)

From the raw contact doc, it seems we don't assign the parent place

```json
{
  "_id": "57abc794c92360562914ff12690118ff",
  "_rev": "1-e40cb19c63fa3b2e161a0aec8ce01fb5",
  "name": "Paul",
  "phone": "+254883720611",
  "type": "contact",
  "contact_type": "person",
  "reported_date": 1698801965100
}
```
The pattern used in the API explains the behavior but it'd be cool if we could assign the place as the person's parent and also include the person id in the response
https://github.com/medic/cht-core/blob/cede018b19239cde2aaeefb30be54db0bfe57f84/shared-libs/contacts/src/places.js#L86-L103

**Expected behavior**
Should be able to create a user for the new contact

**Environment**
- Instance: local instance with moh-ke echis config 
- Browser: Firefox
- Client platform: Linux
- App: webapp, api
- Version: 4.4.1

dianabarsan:
@freddieptf reached out to me, asking if he can work on this. Fred, if this is currently in progress or planning to take on this soon, would you mind assigning this to you, so it doesn't get worked on by another dev. Thanks!

freddieptf:
@dianabarsan yes I'm planning to take this on! I can't assign myself though nor add it to a project

dianabarsan:
Merged to `master`.

mrjones-plip:
@freddieptf - maybe I'm missing why not use the `/api/v2/users` [API](https://docs.communityhealthtoolkit.org/apps/reference/api/#post-apiv2users) which already lets you create a user, contact and place?

PR #8682, opened 2023-11-02
fix(#8674): assign parent place to new contact

freddieptf:
# Description

Contacts created via the [places api](https://docs.communityhealthtoolkit.org/apps/reference/api/#post-apiv1places) don't get a parent place assigned which causes an error: `The contact must be a child of the place`, when trying to create a user for the new contact. This PR fixes that.

medic/cht-core#8674

# Code review checklist

# Compose URLs

* __CHT_CORE_COMPOSE_URL__

# License

### DIFF
--- a/shared-libs/contacts/src/places.js
+++ b/shared-libs/contacts/src/places.js
@@ -17,7 +17,7 @@ const getPlace = id => {
     })
     .catch(err => {
       if (err.status === 404) {
-        err.message  = 'Failed to find place.';
+        err.message = 'Failed to find place.';
       }
       throw err;
     });
@@ -85,6 +85,8 @@ const validatePlace = place => {
 
 const createPlace = place => {
   const self = module.exports;
+  const contact = place.contact;
+  delete place.contact;
   return self._validatePlace(place)
     .then(() => {
       const date = place.reported_date ? utils.parseDate(place.reported_date) : new Date();
@@ -92,14 +94,28 @@ const createPlace = place => {
       if (place.parent) {
         place.parent = lineage.minifyLineage(place.parent);
       }
-      if (place.contact) {
-        // also validates contact if creating
-        return people.getOrCreatePerson(place.contact).then(person => {
-          place.contact = lineage.minifyLineage(person);
-        });
-      }
+      return db.medic.post(place);
     })
-    .then(() => db.medic.post(place));
+    .then((resp) => {
+      if (contact) {
+        contact.place = resp.id;
+        // also validates contact if creating
+        return people.getOrCreatePerson(contact)
+          .then((person) => {
+            return updatePlace(resp.id, { contact: person._id })
+              .then((resp) => {
+                return Promise.resolve({
+                  ...resp,
+                  contact: {
+                    id: person._id,
+                    rev: person._rev
+                  }
+                });
+              });
+          });
+      }
+      return Promise.resolve(resp);
+    });
 };
 
 /*
--- a/shared-libs/contacts/test/unit/places.spec.js
+++ b/shared-libs/contacts/test/unit/places.spec.js
@@ -26,7 +26,7 @@ const contactTypes = [
     group_key: 'contact.type.health_center.plural',
     create_key: 'contact.type.health_center.new',
     edit_key: 'contact.type.place.edit',
-    parents: [ 'district_hospital' ],
+    parents: ['district_hospital'],
     icon: 'medic-health-center',
     create_form: 'form:contact:health_center:create',
     edit_form: 'form:contact:health_center:edit'
@@ -37,7 +37,7 @@ const contactTypes = [
     group_key: 'contact.type.clinic.plural',
     create_key: 'contact.type.clinic.new',
     edit_key: 'contact.type.place.edit',
-    parents: [ 'health_center' ],
+    parents: ['health_center'],
     icon: 'medic-clinic',
     create_form: 'form:contact:clinic:create',
     edit_form: 'form:contact:clinic:edit',
@@ -50,7 +50,7 @@ const contactTypes = [
     create_key: 'contact.type.person.new',
     edit_key: 'contact.type.person.edit',
     primary_contact_key: 'clinic.field.contact',
-    parents: [ 'district_hospital', 'health_center', 'clinic' ],
+    parents: ['district_hospital', 'health_center', 'clinic'],
     icon: 'medic-person',
     create_form: 'form:contact:person:create',
     edit_form: 'form:contact:person:edit',
@@ -338,22 +338,47 @@ describe('places controller', () => {
       };
       sinon.stub(people, 'getOrCreatePerson').resolves({
         _id: 'qwe',
+        _rev: '1',
         name: 'Jim',
         type: 'person'
       });
-      fetchHydratedDoc.resolves({
-        _id: 'ad06d137',
-        name: 'CHP Branch One',
-        type: 'district_hospital'
-      });
       db.medic.post.callsFake(doc => {
-        chai.expect(doc.contact._id).to.equal('qwe');
-        chai.expect(doc.contact.name).to.equal(undefined); // minified
-        chai.expect(doc.contact.type).to.equal(undefined); // minified
-        return Promise.resolve({id: 'ghi'});
+        if (doc.name === 'CHP Family') {
+          return Promise.resolve({ id: 'hc', rev: '1' });
+        }
+        if (doc.name === 'CHP Branch One') {
+          return Promise.resolve({ id: 'ad06d137' });
+        }
+      });
+      fetchHydratedDoc.callsFake(id => {
+        if (id === 'hc') {
+          return Promise.resolve({
+            name: 'CHP Family',
+            type: 'health_center',
+            parent: {
+              _id: 'ad06d137',
+              name: 'CHP Branch One',
+              type: 'district_hospital'
+            },
+          });
+        }
+        if (id === 'ad06d137') {
+          return Promise.resolve({
+            _id: 'ad06d137',
+            name: 'CHP Branch One',
+            type: 'district_hospital'
+          });
+        }
       });
       return controller._createPlaces(place).then(actual => {
-        chai.expect(actual).to.deep.equal({id: 'ghi'});
+        chai.expect(actual).to.deep.equal({
+          id: 'hc',
+          rev: '1',
+          contact: {
+            id: 'qwe',
+            rev: '1'
+          }
+        });
       });
     });
 
@@ -375,10 +400,10 @@ describe('places controller', () => {
         chai.expect(doc.parent._id).to.equal('ad06d137');
         chai.expect(doc.parent.name).to.equal(undefined); // minified
         chai.expect(doc.parent.type).to.equal(undefined); // minified
-        return Promise.resolve({id: 'abc123'});
+        return Promise.resolve({ id: 'abc123' });
       });
       return controller._createPlaces(place).then(actual => {
-        chai.expect(actual).to.deep.equal({id: 'abc123'});
+        chai.expect(actual).to.deep.equal({ id: 'abc123' });
       });
     });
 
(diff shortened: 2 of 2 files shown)


dianabarsan (changes requested):
Thanks for this quick work. I left some readability and test requests. 
Also, we will need to add an e2e  test for this. 
Please have a look at how other API e2e tests work and create one for the places api endpoint: https://github.com/medic/cht-core/blob/b9529a8ba72b56848085b83a5dc3c0aad029ec01/tests/integration/api/controllers/hydration.spec.js

shared-libs/contacts/src/places.js
```diff
@@ -85,21 +85,37 @@ const validatePlace = place => {
...
+                    rev: person._rev
+                  }
+                });
+              });
+          });
       }
-    })
-    .then(() => db.medic.post(place));
+      return Promise.resolve(resp);
+    });
```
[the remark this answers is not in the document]
freddieptf: Yes we can, i was just trying to preserve the code style here with promises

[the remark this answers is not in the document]
freddieptf: Would adding the contact field be a breaking change?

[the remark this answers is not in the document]
dianabarsan: no, it's a fix :D

shared-libs/contacts/test/unit/places.spec.js
```diff
@@ -338,22 +338,47 @@ describe('places controller', () => {
...
-        chai.expect(doc.contact.name).to.equal(undefined); // minified
-        chai.expect(doc.contact.type).to.equal(undefined); // minified
-        return Promise.resolve({id: 'ghi'});
+        if (doc.name === 'CHP Family') {
+          return Promise.resolve({ id: 'hc', rev: '1' });
+        }
+        if (doc.name === 'CHP Branch One') {
+          return Promise.resolve({ id: 'ad06d137' });
+        }
+      });
```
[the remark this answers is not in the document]
freddieptf: Since i know the order of the calls, can i just use `db.medic.post.onCall(0).callsFake` instead?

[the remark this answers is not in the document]
dianabarsan: The problem with using the order of calls is that you end up with testing an implementation detail, and if the calls change order you also need to change the test. Also, from a reader's perspective, it's easier to follow an arguments -> response correlation rather than checking the order of calls, the test will make sense sooner.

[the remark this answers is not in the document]
freddieptf: makes sense, updated to use sinon matcher func

tests/integration/api/controllers/places.spec.js
```diff
@@ -0,0 +1,135 @@
...
+      };
+      return utils.request(onlineRequestOptions)
+        .then(result => {
+          chai.expect(result.id).to.not.be.undefined;
+          return utils.getDoc(result.id);
+        })
+        .then((place) => {
+          chai.expect(place).to.deep.include({
+            name: 'CHP Area One',
+            type: 'health_center',
```
dianabarsan, line 87: Can you please add an assertion about `place.parent`? 
If can be something like: 
```
expect(place.parent._id).to.be.a('string');
```

This way, the test is unambiguous, and if we ever break the code, the test will fail elegantly here instead of in the next call where we might access undefined properties.

tests/integration/api/controllers/places.spec.js
```diff
@@ -0,0 +1,135 @@
...
+    it('should create place with contact', () => {
+      onlineRequestOptions.body = {
+        'name': 'CHP Area One',
+        'type': 'health_center',
+        'parent': 'fixture:online',
+        'contact': {
+          'name': 'Paul',
+          'phone': '+254883720611'
+        }
+      };
```
dianabarsan, line 108: quotes for properties are not needed here

shared-libs/contacts/src/places.js
```diff
@@ -83,23 +83,28 @@ const validatePlace = place => {
...
+  const response = await db.medic.post(place);
+  if (!contact) {
+    return response;
+  }
+  const placeUUID = response.id;
+
+  contact.place = placeUUID;
+  const person = await people.getOrCreatePerson(contact);
+  const result = await updatePlace(placeUUID, { contact: person._id });
+  return { ...result, contact: { id: person._id, rev: person._rev } };
```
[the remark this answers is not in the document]
freddieptf: No the rev isn't used anywhere but it's included in the response for place, so i decided to also include it in contact

tests/integration/api/controllers/places.spec.js
```diff
@@ -0,0 +1,135 @@
...
+            type: 'health_center',
+            contact: { _id: contact._id, parent: contact.parent },
+            parent: {
+              _id: 'fixture:online'
+            }
+          });
+        });
+    });
+
+  });
```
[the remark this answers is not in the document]
freddieptf: > I think we should also test with adding various contact types. Right now, the only test that creates a contact uses the default person type, while I believe any type of person can be created? Would you please add that?

With `validatePlace` and `validatePerson`, i thought that would have that covered?

> Also, what would happen if we pass a type property that is not a person type, so our payload is something like:

Good catch, we can validate the contact model before making the other calls

shared-libs/contacts/src/places.js
```diff
@@ -83,23 +83,29 @@ const validatePlace = place => {
...
-        });
-      }
-    })
-    .then(() => db.medic.post(place));
+const createPlace = async (place) => {
+  const contact = place.contact;
+  delete place.contact;
+
+  await module.exports._validatePlace(place);
+  people._validatePerson(contact);
```
dianabarsan, line 91: I believe that this function will return an error instead of throwing an error. Like here: https://github.com/medic/cht-core/blob/017f395945e9fd1b9c6e5b31041aece7c344f6bf/shared-libs/contacts/src/people.js#L68

So we'll need to treat the error similarly as in the example I linked.

shared-libs/contacts/test/unit/places.spec.js
```diff
@@ -474,7 +516,7 @@ describe('places controller', () => {
       db.medic.post.callsFake(doc => {
         chai.expect(doc.parent._id).to.equal('a');
         chai.expect(doc.parent.name).to.equal(undefined); // minified
-        return Promise.resolve({id: 'x', rev: 'y'});
+        return Promise.resolve({ id: 'x', rev: 'y' });
       });
       return controller.updatePlace('123', data).then(actual => {
         chai.expect(actual).to.deep.equal({ id: 'x', rev: 'y' });
```
dianabarsan, line 522: Can you please add some unit tests where:
- the contact is not valid
- we create a person that is not a `person` type (configurable hierarchies with multiple person types)

shared-libs/contacts/src/places.js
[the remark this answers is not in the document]
dianabarsan: Cool, I think there is a slight risk to changing the API response without versioning though, so if this is not used I think it's safer not to include it - in case integrations require specific responses.

shared-libs/contacts/test/unit/places.spec.js
```diff
@@ -336,24 +336,66 @@ describe('places controller', () => {
...
       });
-      fetchHydratedDoc.resolves({
-        _id: 'ad06d137',
-        name: 'CHP Branch One',
-        type: 'district_hospital'
+      db.medic.post.withArgs(
+        sinon.match(
+          (obj) => {
+            // place is saved first, then updated with contact later
+            return obj.name === 'CHP Family' && !obj.contact;
```
dianabarsan, line 350: Nit: testing that the name is `'CHP Family'` is redundant here, especially since you assert it in the callback. To make this more compact and less of a read, you can change to: 

```
   db.medic.post
     .withArgs(sinon.match((obj) => !obj.contact))
     .callsFake(doc => {
        chai.expect(doc.name).to.equal('CHP Family');
        chai.expect(doc.parent._id).to.equal('ad06d137');
        return Promise.resolve({ id: 'hc', rev: '1' });
      });
```

dianabarsan (changes requested):
This is really really good. I believe I found one minor bug, though and left some nits inline.

shared-libs/contacts/test/unit/places.spec.js
freddieptf: I've added 2 tests that should trigger errors in contact validation, is it worth adding more considering we already test for this elsewhere?

shared-libs/contacts/src/people.js
```diff
@@ -111,3 +111,4 @@ module.exports.isAPerson = isAPerson;
 
 module.exports._getPerson = getPerson;
 module.exports._validatePerson = validatePerson;
+module.exports._getDefaultPersonType = getDefaultPersonType;
```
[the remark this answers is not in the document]
freddieptf: removed

tests/integration/api/controllers/places.spec.js
```diff
@@ -0,0 +1,154 @@
...
+        })
+        .then((parent) => {
+          chai.expect(parent).to.deep.include({
+            name: 'CHP Branch One',
+            type: 'district_hospital'
+          });
+        });
+    });
+
+    it('should create place with contact', () => {
```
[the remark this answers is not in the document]
freddieptf: done

shared-libs/contacts/src/places.js
```diff
@@ -75,6 +75,12 @@ const validatePlace = place => {
     if (!_.isString(place.contact) && !_.isObject(place.contact)) {
       return err(`Property "contact" on place ${placeId} must be an object or string.`);
     }
+    if (_.isObject(place.contact)) {
+      const err = people._validatePerson(place.contact);
+      if (err) {
+        return err(err);
```
lorerod, line 81: When using a contact format:
```
 "contact": {
    "name": "Paula",
    "phone": "+254883720652"
  }
```
At this point we get a `Server error: TypeError: err is not a function`
Could you please improve the way the error is being handled?

shared-libs/contacts/test/unit/places.spec.js
```diff
@@ -336,27 +336,92 @@ describe('places controller', () => {
...
+          id: 'hc',
+          rev: '2',
+          contact: {
+            id: 'qwe',
+          }
+        });
+      });
+    });
+
+    it('returns err if contact does not have name', done => {
```
lorerod, line 391: Thanks, @freddieptf, for adding these unit tests. 
Here, can you please add another scenario similar to this one, to cover the error when the contact does not have the correct format or is missing attributes.

tests/integration/api/controllers/places.spec.js
```diff
@@ -0,0 +1,195 @@
...
+      onlineRequestOptions.body = {
+        name: 'CHP Area One',
+        type: 'health_center',
+        parent: 'fixture:online',
+        contact: {
```
lorerod, line 105: Is this working? without the type attribute?

tests/integration/api/controllers/places.spec.js
lorerod, line 1: This is a great enhancement! Thank you for adding an integration test for this API.

lorerod (changes requested):

shared-libs/contacts/src/places.js
freddieptf: oh i see the issue, let me fix that

tests/integration/api/controllers/places.spec.js
freddieptf: Yes, i believe calls to `createPerson` set a default type if it does not exist

shared-libs/contacts/test/unit/places.spec.js
freddieptf: The value of this test is really only seeing what happens when [contact validation](https://github.com/medic/cht-core/blob/b092815de12f6deb91b7d8ba789925f2ca346ac5/shared-libs/contacts/src/people.js#L32) fails when creating a place, i think adding more would be duplicating already [existing](https://github.com/medic/cht-core/blob/b092815de12f6deb91b7d8ba789925f2ca346ac5/shared-libs/contacts/test/unit/people.spec.js) tests

lorerod (approved):
Thank you, @freddieptf, for addressing my suggestions. 
The problem mentioned [here](https://github.com/medic/cht-core/pull/8682#issuecomment-1840890506) is also fixed.

dianabarsan:
@freddieptf are you happy with the status of this PR? Is it ready to be merged?

freddieptf:
@dianabarsan yep, should be okay to merge!

Issue #8697, opened 2023-11-10
Default config: Consolidate all the enketo commons selectors


tatilepizs:
**What feature do you want to improve?**
Add an `enketo.wdio.page.js` that encapsulates all the commons selectors for the default config. 

**Describe the improvement you'd like**
Create an `enketo.wdio.page.js` file (or move the one that is in the `standard` folder) to consolidate all the commons selectors for the `default` config. Similar to what was done in the standard config [here](https://github.com/medic/cht-core/blob/master/tests/page-objects/standard/enketo/enketo.wdio.page.js)

**Describe alternatives you've considered**

**Additional context**

tatilepizs:
Completed in PR #8720.

A new file, [`common-enketo.wdio.page.js`](https://github.com/medic/cht-core/blob/master/tests/page-objects/default/enketo/common-enketo.wdio.page.js) was created to consolidate all the common selectors related to enketo using mainly the text label.

All the enketo tests were updated to use that file so the other `*.wdio.page.js` files were cleaned and only contain the methods that are specific to a test, this helps us to reduce the duplicated code in the tests and to validate also the text that the user is seeing in every question and option displayed in all the enketo forms.

PR #8720, opened 2023-11-21
chore(#8697): Consolidate enketo common selectors for default config

tatilepizs:
# Description

Create a new file `.wdio.page.js` to consolidate all the common selectors related to enketo, and try to update those selectors using mainly the text label.

Ticket: medic/cht-core#8697

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8697-consolidate-enketo-common-selectors/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8697-consolidate-enketo-common-selectors/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8697-consolidate-enketo-common-selectors/docker-compose/cht-couchdb-clustered.yml)

# License


tests/page-objects/default/enketo/common-enketo.wdio.page.js
```diff
@@ -0,0 +1,35 @@
...
+  await radioButton.waitForClickable();
+  await radioButton.click();
+};
+
+const selectCheckBox = async (text) => {
+  let checkbox = await currentSection();
+  if (await checkbox.$('fieldset.or-branch:not(.disabled)').isExisting()){
+    checkbox = await checkbox.$('fieldset.or-branch:not(.disabled)');
+  }
+  checkbox = await checkbox.$(`label*=${text}`);
```
garethbowen, line 17: I had a little trouble following this... I think it would be clearer if each had their own constant, eg:

```suggestion
  let section = await currentSection();
  if (await checkbox.$('fieldset.or-branch:not(.disabled)').isExisting()){
    section = await checkbox.$('fieldset.or-branch:not(.disabled)');
  }
  const checkbox = await section.$(`label*=${text}`);
```

I'm not quite sure why the `'fieldset.or-branch:not(.disabled)'` bit is needed, so that might need a comment, or a new function with a useful name, eg:

```js
const enabledFieldset = (section) => section.$('fieldset.or-branch:not(.disabled)');

const doSomethingWithUsefulName = async (section) => {
  const fieldset = enabledFieldset(section);
  if (await fieldset.isExisting()) {
    return fieldset;
  }
  return section;
};
```

garethbowen (commented):
I know it's early days, but I like where this is heading! The "common-enketo" API seems super clear so using it will be a breeze, and it's really emulating user behaviour.

tatilepizs:
These are my findings so far about creating the selectors using mainly the text that is shown in the screen.

| | Pros | Cons |
| - | - | - |
| Using text | - We validate not only that the element is present, but also the translation and the visual part of the text. <br /><br/> - Since we are using only text to find the element a single method could work for all the selectos (of the same type) across all the enketo forms. For example, one method to select `radioButtons`, one to add values to `inputs`...  So we are eliminating duplicate selectors and consolidating in a better way the ones that are common through all the forms. <br /><br/> - Readability, it is really easy to understand, when you see the test code, what questions are being evaluated and what is the desired answer <br/><br/> - The `.wdio.page.js` files specific for each form are going to be smaller and simpler | - It was really hard for me to find a unique selector for an element because I didn't find a way to evaluate it before actually executing the code, so took me more time than usual to find the correct one <br /><br/> - In some cases we will need to use something similar to a path to get to the desired element, using methods like [parentElement()](https://webdriver.io/docs/api/element/parentElement/) and [previousElement()](https://webdriver.io/docs/api/element/previousElement/) and I am afraid that if something change in enketo the selectors will be broken because we will depend on the structure of the `html`  |
| Using IDs | - Maybe it is because I am used to creating selectors using IDs, but it is really easy for me to create a unique one and evaluate it using, for example, the chrome extension [Ranorex Selocity](https://chrome.google.com/webstore/detail/ranorex-selocity/ocgghcnnjekfpbmafindjmijdpopafoe) <br /><br/> - It will not depend on the `html` structure, the ID will always point to the same element | - We have a lot of duplicated selectors across all the form page files, where the only difference is one or two words used in the element ID <br /><br/> - It is not always easy to understand what is the element that the selector is pointing to <br /><br/> -  Talking specifically about enketo, it is impossible (or at least I haven't found the way) to add or update a specific ID to an element in the form, so sometimes we have to use an "ugly" path select that no one understands, something like `ul > li > p > span > a` |

Right now I am using a combination of both, I am accessing the correct active page using a selector with IDs and then using a text selector to access the desired element.

I am **not 100% convinced** to use mainly one option, but I would love to create "rules" or "guides" so all of us will know what will be the preferred option to use when we create a new selector to access an element.

@garethbowen, @medic/quality-assurance I will love your thoughts on this 😊

garethbowen:
Great write-up, thanks @tatilepizs !

I've added a comment to the PR.

I think if we can nail the APIs and make it easy and reliable to develop tests then I think the labels are the right approach. It may be however that when we start working with it there are so many edge cases that the APIs become really complicated, in which case IDs probably work better.

We should run the tests a few times to ensure they're stable and not flaky.



> I would love to create "rules" or "guides" so all of us will know what will be the preferred option to use when we create a new selector to access an element.

This would be an amazing resource to add to the Developer Handbook on the docs site.

I don't think we need to do a mass migration to the new system. When the "best practice" is defined then cleaning up the old code can happen over time as we work on the tests. This helps to derisk the effort, so we don't invest a whole lot of time into migration only to find our a few weeks later that our best practice is wrong...

> chrome extension Ranorex Selocity

TIL! Thanks. The Plugin page does imply it provides link text as an option: "Ranorex Selocity generates XPath, link text, RanoreXPath, and CSS selectors. Because only you know which selector fits your needs the best." Does this not work?

tests/page-objects/default/enketo/common-enketo.wdio.page.js
tatilepizs: Thank you Gareth, for your early review, you are right, I am going to exclude it in a separate function.
It is in there because, for the pregnancy form, there is a `pre-init` fieldset that contains also some of the Risk Factors, so if I search for `label*=Heart condition` it will find two, and the first one is the one related to the `pre-init` that is also `disabled` so the test will not be able to click on the checkbox

image

tatilepizs:
> I don't think we need to do a mass migration to the new system. When the "best practice" is defined then cleaning up the old code can happen over time as we work on the tests. This helps to derisk the effort, so we don't invest a whole lot of time into migration only to find our a few weeks later that our best practice is wrong...

That is a good plan, and probably we will get better once we get used to it.

> Does this not work? (Ranorex Selocity)

It doesn't work 😞 

It has different options for selectors:

image

if I use `button.cancel` it will detect 'css' and it find the element

image


But, if I use `button=Cancel` it doesn't detect anything and doesn't find any element

image

tests/page-objects/default/enketo/pregnancy.wdio.page.js
lorerod, line 1: Same here! You ended up with a simple and easy-to-understand code. Thank you.

tests/page-objects/default/enketo/pregnancy.wdio.page.js
```diff
@@ -90,130 +60,66 @@ const setFutureVisitDate = async (value = moment().add(1, 'day').format('YYYY-MM
...
-  breathlessness: breathlessnessValue = 'yes',
-  llin: llinValue = 'yes',
-  ironFolate: ironFolateValue = 'yes',
-  dewormingMedication: dewormingMedicationValue = 'yes',
-  hivTested: hivTestedValue = 'yes'
-} = {}) => {
+const submitDefaultPregnancy = async () => {
   await commonPage.openFastActionReport('pregnancy');
-  await selectGestationAge(gestationalAgeValue);
+  //await selectGestationAge(GESTATION_AGE.edd);
```
lorerod, line 65: Why do we not have to select gestational age anymore?

lorerod (commented):
Thank you for these changes, @Tatilepiz. This looks great. Amazing how it simplifies the code!

tests/e2e/default/enketo/pregnancy.wdio-spec.js
```diff
@@ -31,8 +30,8 @@ describe('Pregnancy registration', () => {
   });
 
   it('Should submit a new pregnancy', async () => {
-    const edd = moment().add(30, 'days');
-    const nextANCVisit = moment().add(1, 'day');
+    const edd = moment().add(8, 'days');
+    const nextANCVisit = moment().add(2, 'day');
```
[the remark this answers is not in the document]
tatilepizs: I changed it because it was simpler for me to calculate the weeks of pregnancy 😅 if the delivery will happen in 8 days the woman will be 38 weeks pregnant. If leave it as it was, incrementing the number of days to 30 I will not know how many weeks she will be, because that will depend on the month and the amount of weeks that each month has.
I just tried to make it simpler 😅

tests/e2e/default/enketo/pregnancy.wdio-spec.js
```diff
@@ -99,7 +128,7 @@ describe('Pregnancy registration', () => {
   });
 
   it('Should verify that all tasks related with the high risk pregnancy were created', async () => {
-    const tasksTitles = ['Health facility ANC reminder', 'Danger sign follow up', 'Pregnancy home visit'];
+    const tasksTitles = ['Health facility ANC reminder', 'Danger sign follow up', 'Delivery'];
```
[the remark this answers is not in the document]
tatilepizs: This happened because I changed the expected delivery date, if it is in a month the program will automatically create some tasks, and one is a `pregnancy home visit`, but since I changed the `edd` to 8 days the program created a `delivery` task instead a `Pregnancy home visit`

tests/e2e/default/enketo/pregnancy.wdio-spec.js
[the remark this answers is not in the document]
lorerod: Great! thanks for the explanation!

tests/page-objects/default/enketo/common-enketo.wdio.page.js
```diff
@@ -0,0 +1,35 @@
+const currentSection = () => $('section[class*="current"]');
+
+const selectRadioButton = async (question, label) => {
+  const radioButton = await currentSection()
+    .$(`legend*=${question}`)
+    .parentElement()
+    .$(`label*=${label}`);
```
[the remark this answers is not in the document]
tatilepizs: I use the `html`, manually search the text inside the section that I am pointing using the id 😄

tests/page-objects/default/enketo/pregnancy.wdio.page.js
tatilepizs: We have to select it, it is just that I am using the method that I created for the radioButtons

``` .js
await commonEnketoPage.selectRadioButton('How would you like to report the pregnancy?',  'Expected date of delivery (EDD)');
```

I need to delete that comment line, thanks!

tatilepizs:
Hi @ralfudx, @lorerod,

This PR is ready for review. The most important changes are in the file [common-enketo.wdio.page.js](https://github.com/medic/cht-core/blob/8697-consolidate-enketo-common-selectors/tests/page-objects/default/enketo/common-enketo.wdio.page.js), all the other changes are the e2e tests using that file to run the tests.

Please check if I left anything using an old selector that needs to be updated, or any new selector that needs to be included in the new `common-enketo.wdio.page`.

Many thanks! 😊 

@jkuester, @m5r, These changes are all related to the enketo e2e tests, so if you have a chance to take a look at it will be great, I would love your feedback.

cc: @jonathanvq

tests/e2e/default/enketo/submit-form.wdio-spec.js
```diff
@@ -118,7 +119,7 @@ describe('Submit Enketo form', () => {
     await commonElements.openFastActionReport(assessmentForm.internalId, false);
 
     // enter name
-    await (await genericForm.nameField()).setValue('Jones');
+    await commonEnketoPage.setInputValue('person.field.name', 'Jones');
```
garethbowen, line 122: Seems odd. Is this label not translated in the config?

tests/e2e/cht-form/default/pregnancy-danger-sign-follow-up.wdio-spec.js
```diff
@@ -15,8 +15,19 @@ describe('cht-form web component - Pregnancy Danger Sign Form', () => {
...
+    await commonEnketoPage.selectRadioButton('Fits', 'Yes');
+    await commonEnketoPage.selectRadioButton('Severe abdominal pain', 'Yes');
+    await commonEnketoPage.selectRadioButton('Severe headache', 'Yes');
+    await commonEnketoPage.selectRadioButton('Very pale', 'Yes');
+    await commonEnketoPage.selectRadioButton('Fever', 'Yes');
+    await commonEnketoPage.selectRadioButton('Reduced or no fetal movements', 'Yes');
+    await commonEnketoPage.selectRadioButton('Breaking of water', 'Yes');
+    await commonEnketoPage.selectRadioButton('Getting tired easily', 'Yes');
+    await commonEnketoPage.selectRadioButton('Swelling of face and hands', 'Yes');
+    await commonEnketoPage.selectRadioButton('Breathlessness', 'Yes');
```
garethbowen, line 30: Overall I think this is an improvement so don't take this as a blocker... but I think there's still value in bundling these up in the page object if it reduces some of the verbiage in the test. As in, this test would call the predanger sign follow up form page, and the predanger sign follow up form page would call the common enketo page.

What do you think?

tests/page-objects/default/enketo/common-enketo.wdio.page.js
```diff
@@ -0,0 +1,74 @@
...
+const setTextareaValue = async (question, value) => {
+  await setValue('textarea', question, value);
+};
+
+const validateSummaryReport = async (textArray) => {
+  const element = await getCurrentPageSection();
+  for (const text of textArray) {
+    expect(await (await element.$(`span*=${text}`)).isDisplayed()).to.be.true;
+  }
+};
```
garethbowen, line 64: I'm trying to figure out if we can do something more specific than just checking the page for the strings, for example, asserting the order or the format somehow. I can't think of anything though. The summary page isn't well structured. Can you think of anything?

One day it'd be great to take a snapshot of the page and compare build to build to make sure nothing changes but that day isn't today...

garethbowen (approved):
Amazing thanks! Some comments but no blockers.

tests/page-objects/default/enketo/pregnancy.wdio.page.js
```diff
@@ -1,201 +1,72 @@
...
-const getSummaryDetails = async () => {
-  return {
-    patientNameSumm: await patientNameSummary().getText(),
-    weeksPregnantSumm: await weeksPregnantSummary().getText(),
-    eddSumm: await eddSummary().getText(),
-    riskFactorsSumm: await riskFactorsSummary().length,
-    dangerSignsSumm: await dangerSignsSummary().length,
-  };
-};
+const submitDefaultPregnancy = async () => {
```
lorerod, line 13: Adding a flag to not submit the form can make this function more reusable. This would allow us to use this code in multiple tests that require validation of the summary report before submitting the default pregnancy form. What do you think?

lorerod (changes requested):
Hi, @tatilepizs; thank you so much for these changes; I think it is an improvement overall. I left some minor suggestions. Let me know what you think.

ralfudx (approved):
Overall looks good to me... not too sure about the changes requested but i guess this is a good starting point for me to delve into these tests - thanks for doing this @tatilepizs

tests/e2e/default/enketo/submit-form.wdio-spec.js
tatilepizs: It has that weird label because it is a custom form, it is not part of the config, and that is the label that the form has -> see [here](https://github.com/medic/cht-core/blob/8697-consolidate-enketo-common-selectors/tests/e2e/default/enketo/submit-form.wdio-spec.js#L31).

But you are right, it might be confusing to see that in the test, I will change it.

tests/e2e/cht-form/default/pregnancy-danger-sign-follow-up.wdio-spec.js
tatilepizs: It will definitely look better, and those danger signs are set in 2 or test, I will create the file and isolate them in a separate function. Thank you!

tests/page-objects/default/enketo/common-enketo.wdio.page.js
tatilepizs: I thought about that when I was making the changes, but most of the summary pages are different depending on the report, and this was the easiest way I found to verify it.

Of course, I am open to suggestions, if you or anyone else on the team have an idea of how can we make it better we can definitely try it 🙂

tests/page-objects/default/enketo/pregnancy.wdio.page.js
tatilepizs: That is a really good idea, I will add the flag.
Thank you! 🙂

lorerod (approved):
Thank you, @tatilepizs, for adding my suggestions!

Issue #8716, opened 2023-11-17
E2E - Barcode happy path scenario 


latin-panda:
**Describe the issue**
Scenario: CHW should be able to scan the barcode using the phone and get search results (contacts) matching the code.

lorerod:
Moving this to TODO until we have a better solution for running our e2e test on mobile.

latin-panda:
Removing from milestone as it wasn't completed on that iteration

Issue #6669, opened 2020-10-09
Barcode scanning using device camera


MaxDiz:
**Is your feature request related to a problem? Please describe.**
As a nurse/CHW, I do home visits to follow-up with patients who are in treatment for disease management, but did not show for clinic appointments. Patients are often taking multiple medications (or using other health products), which are hard to keep track of. Having an accurate list of these products is critical to creating and managing up-to-date health record for patients. The process for collecting this information is currently manual, making it difficult to capture it in a timely and accurate way. 

**Describe the solution you'd like**
As a nurse, I would like the ability to scan barcodes on medications (or other health products) to quickly and accurately acquire patient health information. 

**Describe alternatives you've considered**
- manually inputting information
- taking photographs of products 

**Additional context**
bar code scanner is supported by [XLSForm](http://xlsform.org/en/#question-types) and [ODK Collect](https://docs.getodk.org/form-question-types/#barcode-widget)

Worth noting that ZXing is in maintenance mode, and ODK Collect had to build in the scanner as opposed to an intent since 1.7. There is more investigation needed to know if this is a relatively simple task (handle the XLSForm type and make/handle a new intent) or a big lift.

MaxDiz:
@isaacholeman please add any additional context specific to the project (or correct anything I misunderstood)

isaacholeman:
Looks great @MaxDiz. I added that it's also a requirement to be able to scan barcodes to get patient ID. Thanks!

garethbowen:
Also investigate supporting QR codes.

MaxDiz:
Relevant [CHT Forum conversation](https://forum.communityhealthtoolkit.org/t/qr-code-reading-and-profile-picture/1079)

latin-panda:
@michaelkohn @lorerod @n-orlowski FYI

Nb. This will be delivered in 4 different PRs:
- Search by barcode (this PR)
- Enketo widget to read barcode 
- Autoselect if search by barcode returns 1 result
- Automation test coverage

This work uses the native [Barcode Detector API](https://developer.mozilla.org/en-US/docs/Web/API/Barcode_Detection_API) from the browser by using a hidden input type `file` to natively open the camera in Android devices. This avoids custom code in the CHT Android and allows more accessibility since the feature can be used in CHT Android, PWA, and Android browsers. 

**It works the following way:** 
- The user clicks on the barcode scanner icon
- Prompt is displayed by the OS to take a picture or pic from the gallery 
- User takes a picture
[...]

PR #8732, opened 2023-11-28
chore(#6669): Search by barcode e2e test

lorerod:
# Description

Closes https://github.com/medic/cht-core/issues/8716
# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6669-search-by-barcode-e2e-test/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6669-search-by-barcode-e2e-test/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6669-search-by-barcode-e2e-test/docker-compose/cht-couchdb-clustered.yml)

# License


lorerod:
I'm still trying to figure out a couple of things.
In the `performBarcodeSearch`  method in `tests/page-objects/default-mobile/search/search.wdio.page.js`
I need to manipulate the DOM of the input web element to make it interactable and be able to set the value.

In `tests/e2e/default-mobile/wdio.conf.js`, I changed beforeSuite to beforeHook as testing this locally, I experience that when a test fails, the emulateDevice method loses the mobile view and switches back to desktop.

tests/page-objects/default-mobile/search/search.wdio.page.js
```diff
@@ -1,14 +1,27 @@
...
   await (await openSearchBox()).waitForClickable();
   await (await openSearchBox()).click();
   await searchPageDefault.performSearch(term);
 };
 
+const performBarcodeSearch = async (barcodeImagePath) => {
+  /*In this case the upload file button is hidden,
+  then we need to manipulate the DOM of the respective webelement to make it interactable.*/
+  browser.execute(function () {
+    document.getElementsByClassName('barcode-scanner-input')[0].style.display = 'block';
```
dianabarsan, line 16: Not a fain of this either, because it will definitely obscure if the actual button that makes the input selection no longer works, or is not displayed. So the test only validates the functionality partially. 
Is there no way to use the elements that are displayed on the page to upload the image?

lorerod: If you click on the barcode icon, the OS upload file dialog appears, and Webdriverio has no control over it. I'm investigating how to setValue to a web element that is not interactable without modifying the DOM.

dianabarsan: Yea, I went through that search as well and in the end I couldn't find anything. The idea is that wdio really wants to simulate as close to user interaction as possible. 
Maybe we should just leave it like this.

tests/page-objects/default-mobile/search/search.wdio.page.js
```diff
@@ -1,14 +1,27 @@
...
+  });
+  await (await barcodeSearchInput()).setValue(barcodeImagePath);
+  await browser.pause(1000);
```
dianabarsan, line 19: What is the reason for the 1s delay?

lorerod: We [experienced](https://github.com/medic/cht-core/issues/6669#issuecomment-1812666551) some delay when uploading and processing the barcode image. But I only tried the e2e test with the wait; It may be unnecessary.

lorerod: I try without the wait, and the test fails because it doesn't have time to process the search and show results. [This](https://github.com/medic/cht-core/blob/81f20d3ac93d9ec6b05addddfef4a8641f00d58e/tests/e2e/default-mobile/contacts/barcode-search-contacts.wdio-spec.js#L34) validation fails.

tatilepizs:
Hi @lorerod,

I was trying to run the test locally but it is failing for me, it is not seeing the barcode scanner icon, the same thing that happened on the CI.

image

Let me know if there is anything else can do to help.
Good luck 🤞

tests/e2e/default-mobile/contacts/barcode-search-contacts.wdio-spec.js
```diff
@@ -0,0 +1,50 @@
+const searchPage = require('@page-objects/default-mobile/search/search.wdio.page.js');
```
latin-panda, line 1: Hi @lorerod just popping up here to see if you need my help with the barcode?

lorerod: Hi, @latin-panda. We are experiencing a strange behavior when [running the e2e test](https://github.com/medic/cht-core/pull/8732#issuecomment-1843357767) compared to when[ running the developer setup first and then the e2e test](https://github.com/medic/cht-core/pull/8732#issuecomment-1843642985). When running the e2e test directly in the branch, it does not use the latest changes, and the test fails. This is also happening in CI.

latin-panda: The version in `package.json` and `package-lock.json` is different from the [4.4.1-FR-barcode](https://github.com/medic/cht-core/blob/a17e8987a93aee1d042a57e5fa917d79f42e4769/package.json#L3) branch. We need to pull the latest from `4.4.1-FR-barcode`

lorerod: Thanks @latin-panda! I pulled the latest from `4.4.1-FR-barcode`, but the test still [fails](https://github.com/medic/cht-core/actions/runs/7389173632?pr=8732) on CI because the barcode search icon doesn't show. I will continue looking into this to see if I can find the problem.

lorerod:
The Barcode Scanner API is not supported on Linux, which means this test will not work on CI.

Issue #6395, opened 2020-05-04
Pregnancy registration allows first pregnancy and previous miscarriage risk factors


brad1905:
During a pregnancy registration the user is allowed to specify risk factors of "first pregnancy" and "previous miscarriage" which should not be allowed.  

1. Create female person within a household
Person should be within childbearing age range
2. select new action, Pregnancy registration 
3. Risk factors form allows user to select Y to "first pregnancy" and Y to previous miscarriages or stillbirths 

**Expected behavior**
miscarriages and stillbirths question should not display if first pregnancy is answer Y 

**Environment**
master

Chrome
Android 

**Additional context**
<img width="751" alt="first pregnancy with prev miscarriages" src="https://user-images.githubusercontent.com/62311678/80984555-dbeba400-8de2-11ea-91bd-77b6765c004d.png">

latin-panda:
@Benmuiruri Please check if this is still an issue. 
If yes, then please involve QA from the start (before starting development) and fix the [pregnancy form](https://github.com/medic/cht-core/blob/master/config/default/forms/app/pregnancy.xlsx). There's a possibility that you need to fix e2e tests after fixing this.

latin-panda:
@Benmuiruri some resources that might help:
- Info about form configuration in XLS [here](https://xlsform.org/en/).
- Building cht app forms [here](https://docs.communityhealthtoolkit.org/apps/tutorials/app-forms/)

PR #8742, opened 2023-12-05
chore(#6395): update pregnancy form

Benmuiruri:
Update the [pregnancy form](https://github.com/medic/cht-core/blob/master/config/default/forms/app/pregnancy.xlsx) to not show miscarriages and stillbirths question if first pregnancy is answer `Yes`. Only show the question if the answer to firstPregnancy is `No`

#6395 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6395-fix-pregnancy-form/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6395-fix-pregnancy-form/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6395-fix-pregnancy-form/docker-compose/cht-couchdb-clustered.yml)

# License


Benmuiruri:
Hi @tatilepizs I'd like to include you in the PR for QA. 

- From my understanding fixing the issue includes updating the default config pregnancy xlsx to conditionally show the miscarriages and stillbirths question only if first pregnancy is answer `No`. If it `Yes`, the question should not appear.
- Having made that change, I updated the test to conditionally include the miscarriage answer based on the `firstPregnancy` flag. 

I thought it would be ideal to also update the demo and standard config pregnancy forms to keep it consistent,  what are your thoughts on that? Or should I only update the default config? 

Please share any concerns you may have from a QA perspective. Thank you 🙏

Benmuiruri:
@tatilepizs glad that you agree we should update the standard config as well. However, after testing the standard config, I saw it has a different Pregnancy Risk Factors form

The `Demo` and `Default` configs have the same radio button question for Pregnancy Risk Factors
Screenshot 2023-12-07 at 08 39 04

The Standard config has a check box question for the Pregnancy Risk Factors. I tested the standard config and saw that it's covered. If you choose First Pregnancy as a risk factor you cannot choose the others so I think that's okay. 
Screenshot 2023-12-07 at 09 02 16

Therefore, I think we can just `Demo` and `Default` config. Let me know what you think.

tatilepizs:
Thank you for checking @Benmuiruri.
Sounds good! 👍

Benmuiruri:
Hi @jkuester 

Are you available for this review ?  👀

jkuester:
Sorry @Benmuiruri just catching up here after being OOO last week.  I plan to look at this review tomorrow!  :+1:

config/default/forms/app/pregnancy.xml
```diff
@@ -978,105 +978,105 @@
...
-      <bind nodeset="/pregnancy/risk_factors/risk_factors_present/additional_risk_check" type="select1" required="true()"/>
-      <bind nodeset="/pregnancy/risk_factors/risk_factors_present/additional_risk" type="string" required="true()" relevant="selected(../additional_risk_check, 'yes')" jr:constraintMsg="jr:itext('/pregnancy/risk_factors/risk_factors_present/additional_risk:jr:constraintMsg')" constraint="string-length(.) &lt;= 100"/>
-      <bind nodeset="/pregnancy/risk_factors/r_risk_factor_present" type="string" calculate="if(selected(../risk_factors_history/first_pregnancy, 'yes') or
+      <bind nodeset="/pregnancy/risk_factors/risk_factors_history/first_pregnancy" required="true()" type="select1"/>
+      <bind nodeset="/pregnancy/risk_factors/risk_factors_history/previous_miscarriage" relevant="selected(../first_pregnancy, 'no')" required="true()" type="select1"/>
```
jkuester, line 1055: @Benmuiruri can you confirm if this was the only actual change to the form logic?  I think so, based on context and a brief skim of the rest of the diff, but just want to be sure...

(Also JSYK, this whole mess of changes within the form xml file because of indeterminate arrangement of the attributes will be fixed by updating our pyxform version (whenever we get around to that...).  Until then, I typically try to only commit the xml lines that have actually been updated (though sometimes this is not feasible). No worries on this PR, but something to consider in the future.)

config/default/test/form-inputs.js
```diff
@@ -216,14 +216,20 @@ module.exports = {
         additionalFactors: ['yes', 'underweight']
       };
       Object.assign(content, opts);
+
+      const pregnancyAnswers = [content.firstPregnancy];
+      if (content.firstPregnancy === 'no') {
```
jkuester, line 221:
```suggestion
      if (content.miscarriages) {
```

Instead of baking in some business logic here, can we just check for a `firstPregnancy` value?  I don't think we want the test util code to auto-correct the given answers if someone tries to provide both `firstPregnancy === 'no'` AND a value for `miscarriages`.  Instead the test should fail and they will need to update the provided answers.

config/demo/test/form-inputs.js
```diff
@@ -216,14 +216,20 @@ module.exports = {
         additionalFactors: ['yes', 'underweight']
       };
       Object.assign(content, opts);
+
+      const pregnancyAnswers = [content.firstPregnancy];
+      if (content.firstPregnancy === 'no') {
```
jkuester, line 221:
```suggestion
      if (content.miscarriages) {
```

jkuester (approved):
Very nice!  Just a minor suggestion in the test code.

config/default/forms/app/pregnancy.xml
Benmuiruri: Thanks for the Info Josh.

config/default/test/form-inputs.js
Benmuiruri: Hi @jkuester your suggestion will not work. It will allow the miscarriage question to be included based on the presence of a miscarriage value, regardless of whether firstPregnancy is 'yes' or 'no'.

For context, this is what we want: 

- The miscarriage question should only be asked if this is not the first pregnancy (firstPregnancy === 'no').
- If it is the first pregnancy (firstPregnancy === 'yes'), the miscarriage question should not be included at all.

jkuester: Right, I think we are on the same page here except that I am suggesting we make the test code _less smart_!  It is invalid to try and set `firstPregnancy === 'yes'` and `miscarriages === 'yes'` and now the form will actually not even let you fill out the miscarriage question when `firstPregnancy === 'yes'`.  However, as this test util code is currently written, I could still write a test where I say  `firstPregnancy: 'yes'` AND `miscarriages: 'yes'`.  Then, when I call this util code, it will just ignore my value for `miscarriages` and my test will pass even though I set bad data (then this can get confusing to understand and debug in the future).  Instead, I would rather just have this util code pass through whatever the data is provided by the test and just let it fail later if something is wrong (could alternatively just actively throw an exception here if we see bad data, but that is just more code to maintain....)

Benmuiruri: I get it now.

However, I don't think `if (content.miscarriages) {}` will work as it is because it will attempt to fill in miscarriages even when `firstPregnancy` is `yes` So I think we can go with `if (content.firstPregnancy === 'no' && content.miscarriages) {}`

I believe that will keep the test utility function `simple` by not including logic to correct or ignore data but still respects the form's logic.

jkuester: Hmm, no, if `content.firstPregnancy === 'no'` then they have to provide a value for `miscarriages` to be able to complete the form.  So, checking them both is redundant!  Ultimately, this is just a super minor nitpick and I am good with what you had originally (`if (content.firstPregnancy === 'no')`) if that makes the most sense to you. Let's not let this hold up these changes!

That being said, I will make one more attempt and explaining myself better here with a bit more context!  :smile: 

The automated tests for these forms (e.g. the `pregnancy` form) use the cht-conf-test-harness to fill out the form in a test environment.  The way the `harness.fillForm` function works by taking an array of "answers" to questions on the form.  There is no mapping for which answer goes to which question in the form, so everything is just sequential. The first answer in the array goes to the first _visible_ question.  As you fill out a form, sometimes questions become visible (or get hidden) depending on what inputs you are providing. The test harness will always check for the next _visible_ question to see what it should fill out. If there are answers left at the end of the form page and all the questions have been filled out, then the harness will throw an exception saying that the given answers were invalid.

The tests (e.g. in the `pregnancy.spec.js` file) use the data from this `form-inputs.js` file to fill out forms for various scenarios.  For more complex workflows it can be useful to have the test data from `form-inputs.js` be somewhat _dynamic_ to allow for reuse between similar scenarios.  So in this case we are looking at the `pregnancyRegistrationScenarios.riskDanger` function which will return a set of data for the pregnancy form that is useful for testing various "risks" workflows (and is getting reused in a number of tests). The test data returned from this function is passed directly to the test-harness which will blindly try to fill out the fields on the form one-after-another.

Since this `pregnancyRegistrationScenarios.riskDanger` function is reusable and the consuming tests can pass in various combinations of `content` values, I am trying to avoid a future scenario where a test improperly calls this function with bad content data like `{ firstPregnancy: yes, miscarriages: 'yes' }` (since you cannot answer "yes" to both of these questions.  If a test calls `pregnancyRegistrationScenarios.riskDanger` with content like that, I want the test to FAIL.  The trouble is that with your logic, I think the test will actually PASS!  (You can confirm this by updating the `shows summary when this is the first pregnancy for the woman and there are all the risk factors` test in `pregnancy.spec.js` to set `miscarriages: 'yes'`.  That is an invalid test case that should fail, but I think your logic here will allow it to pass...)

If, instead, we just checked `if (content.miscarriages) {` here, then I think the test would fail since it would include the `miscarriages` answer in the array of answers that gets returned from `pregnancyRegistrationScenarios.riskDanger`. Then that array of answers would be passed to the test harness which would try to fill out the questions on the form.  With `firstPregnancy: yes`, then the miscarriages question should not become visible in the form and the harness will throw an error saying that it cannot fill out the form because it has too many answers.

Benmuiruri: Hi @jkuester thanks for the detailed response. 

To avoid holding this PR up. I will revert to what I had `(if (content.firstPregnancy === 'no'))`

However, I'd like to continue this conversation elsewhere because as much as I understand your concern, I don't see how to write a "passing" test with `if (content.miscarriages) {}`

If that sounds good, let me know and I will merge this PR. Thanks.

jkuester: :+1: Yes, that sounds good!

Issue #7770, opened 2022-09-07
Browser compatability notice for some, block access for others


mrjones-plip:
**Describe the issue**
In #7448 we added a notice at login and on the about page if your browser was incompatible. However, logged users won't see the login screen, we set a year long cookie and logged users rarely go to the about page.  Further, with the release of 4.2, [we now can have users](https://github.com/medic/cht-core/issues/7770#issuecomment-1728466554) that should be fully blocked, regardless of if they're logged in or not.

**Describe the improvement you'd like**
We should:
* block all access for  Chrome <72 (to be verified)
* warn between 72 and 90

NB - out of scope for this ticket is any warnings about CHT Android (eg `0.9.x`) versions or Android versions  (eg `4.2`)

**Describe alternatives you've considered**
NA

mrjones-plip:
Yeah - still should be done!

@n-orlowski - thoughts on the UX for this? Similar to the work you helped out with in [7448](https://github.com/medic/cht-core/issues/7448).  No rush, but if you wanted to throw some comps here, that'd help move things along!

garethbowen:
Basic thoughts here...

- If you have a browser we know will not work we should not allow the user to log in, and if they're already logged in, not try and load the app. Instead show an error page with links to documentation about how to update your webview.
- If you have a browser that is not supported, we should show a warning with a link to the documentation above, but still allow the user to load the app because it might just work.
- For the first iteration I think we can put the documentation on the docs site. It doesn't need to be offline first because you have to be online to access the play store anyway.
- For the first iteration the warning can be as simple as a modal dialog on load. At some point we can do a full UX review and come up with something more user friendly.

mrjones-plip:
> For the first iteration I think we can put the documentation on the docs site. It doesn't need to be offline first because you have to be online to access the play store anyway.

Agreed as long as we have a translatable msg, `Error - your device needs to be updated before it will work with CHT - see here for more information` (or what ever msging we defaulted to).  This way at least offline users know they need to do something likely out of their control in order to proceed

mrjones-plip:
@michaelkohn  - calling this out per our call earlier today, per [Gareth's comment above](https://github.com/medic/cht-core/issues/7770#issuecomment-1685529921) (emphasis mine):

> If you have a browser we know will not work **we should not allow the user to log in**

So this ticket should handle that use case, as we hoped it would, which is great! However, noting that this is different than the title of the ticket "notice for logged in user".  

[...]

Issue #8456, opened 2023-08-11
Loading the webapp fails on with WebView version 74


mrjones-plip:
**Describe the bug**
Starting at CHT 4.2.0, when a self hosted CHT instance uses `nginx` as it's front end, certain Android devices fail with a console error `Uncaught SyntaxError: Unexpected token ? main.js:1`

**To Reproduce**
1. Install CHT Core 4.2.0 in docker helper. Do not use AWS ELB
2. Using CHT Android 1.0.1 with webview 74 or lower, connect to CHT Core
4. See error in console `Uncaught SyntaxError: Unexpected token ? main.js:1` and see full screen spinner.  

**Expected behavior**
You should be prompted to log in to the CHT with no errors in console

**Screenshots**

<details>
<summary>Console tab</summary>

<img width="1319" alt="Captura de pantalla 2023-08-11 a la(s) 13 40 35" src="https://github.com/medic/cht-core/assets/21312057/86f3edbc-bd19-40ba-8f40-afc1b53fad3d">

</details>

<details>
<summary>Network tab</summary>

<img width="1326" alt="Captura de pantalla 2023-08-11 a la(s) 13 40 17" src="https://github.com/medic/cht-core/assets/21312057/9b240b7b-4fe4-4443-8f95-6ff62c91f276">

</details>

<details>
<summary>Security tab</summary>

<img width="1314" alt="Captura de pantalla 2023-08-11 a la(s) 13 40 25" src="https://github.com/medic/cht-core/assets/21312057/a9b6c124-1709-47f9-b66b-0c617a73b6bb">

</details>

**Environment**
- Instance: local dev
- Browser: CHT Android 1.0.1 WebView 74 or 68
- Client platform: See above
- App: webapp
- Version: 4.2.0 and above
  
**Additional context**
This was discovered while testing #8278.  It's assumed that this wasn't caught earlier because we test on Gamma which uses AWS ELB to terminate TLS.  Indeed, testing CHT COre 4.2.0 on Gamma  with the exact same handset as above works.  Testing on 4.1.0 and 4.0.0 on the same handset also work. 

Also, testing the same 4.2.0 instance with WebView 111 worked fine. All desktop browsers worked as well.

garethbowen:
This may be resolved by setting a higher minimum browser version as part of https://github.com/medic/cht-core/issues/7448 but we don't currently know the upper bound on this issue so we need to look into it regardless.

@mrjones-plip It looks like this is happening after login - is that correct?

[...]

PR #8744, opened 2023-12-07
chore(#7770): add browser compatibility notice

Benmuiruri:
# Description

#7770 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7770-browser-compatibility-notice/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7770-browser-compatibility-notice/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7770-browser-compatibility-notice/docker-compose/cht-couchdb-clustered.yml)

# License


api/resources/translations/messages-en.properties
```diff
@@ -391,6 +391,8 @@ branding.title.field = Title
 branding.title.field.help = This text will be displayed in the browser tab and as the PWA title.
 branding.icon.field = Large icon
 branding.icon.field.help = Will show for PWA installations. Should be at least 144 pixels square.
+browser.compatibility.title = Your Browser is outdated.
+browser.compatibility.description = Your device needs to be updated before it will work with CHT.
```
latin-panda, line 395:
```suggestion
browser.compatibility.title = Your Browser is outdated
browser.compatibility.description = Update your browser to use the application.
```

Many CHWs don't know that it's called "CHT" so we usually use "app" or "application" 
Translation in Spanish:
```
browser.compatibility.title = Su navegador está desactualizado
browser.compatibility.description = Actualice su navegador para utilizar la applicación
```
You can find other teammates in Product or Medic in general, to give you the translation of these

webapp/src/ts/app.component.ts
```diff
@@ -646,6 +650,23 @@ export class AppComponent implements OnInit, AfterViewInit {
     });
   }
 
+  private checkBrowserCompatibility(): void {
+    if (!this.isCompatibleChromeBrowser()) {
+      this.modalService.show(BrowserCompatibilityComponent);
+    }
+  }
+
+  private isCompatibleChromeBrowser(): boolean {
```
latin-panda, line 659: Is browserDectorService responsibility for checking the browser? If yes, then you can move this function to that service.

webapp/src/ts/app.component.ts
```diff
@@ -646,6 +650,23 @@ export class AppComponent implements OnInit, AfterViewInit {
...
+    const browserInfo = this.browserDetectorService.detect();
+
+    if (browserInfo.name !== 'Chrome' || !browserInfo.version) {
+      return true;
+    }
+
+    const majorVersion = parseInt(browserInfo.version.split('.')[0]);
+    return !(majorVersion < 72 || (majorVersion >= 72 && majorVersion < 90));
```
latin-panda, line 667: Please use constants for these versions numbers.

Can you [please confirm if it's 72 or 74](https://docs.communityhealthtoolkit.org/contribute/code/core/run-multiple-chrome-versions/)? As per [this ticket, it seems that it was failing in v74](https://github.com/medic/cht-core/issues/8456#issuecomment-1681321031)

latin-panda: Since it's the same message for when `< 72` and` >=72 && < 90`. 
Then, can this be simplified to `if majorVersion < 90 then show modal`?  
Could you double-check with Michael or Nicole if one message for both scenarios is okay? You can also send a demo video to them so they have a better idea of the experience and provide better feedback.

webapp/src/ts/modals/browser-compatibility/browser-compatibility.component.ts
```diff
@@ -0,0 +1,21 @@
+import { Component } from '@angular/core';
+import { MatDialogRef } from '@angular/material/dialog';
+
+@Component({
+  selector: 'browser-compatibility-modal',
+  templateUrl: './browser-compatibility.component.html'
+})
+export class BrowserCompatibilityComponent {
+  static id = 'browser-compatibility-modal';
```
latin-panda, line 9: Can you use this ID in the HTML?

latin-panda (changes requested):
Nice job! I've sent some feedback.

Benmuiruri:
Progress on what I have so far is explained in this [short video ](https://www.loom.com/share/8b9c75db2c214e45b7861187249fb2af?sid=fee00f72-8467-429f-b862-f7b3fb2c937d)

I am currently blocked at how to avoid the forever spinner being [discussed here](https://github.com/medic/cht-core/issues/8456#issuecomment-1677618103)

michaelkohn:
Thanks for the video! For the "Warn" modal, you have this 👇🏼 

Screenshot 2023-12-13 at 1 50 08 PM

Since Cancel and Okay just do the same thing, we should just do like we do with bad date settings, like this (just one button that says "Ok") 👇🏼 

Screenshot 2023-12-13 at 12 47 26 PM

webapp/src/ts/services/browser-detector.service.ts
```diff
@@ -39,6 +39,10 @@ export class BrowserDetectorService {
     });
   }
 
+  public isUsingOutdatedChromeBrowser() {
+    return this.parser.satisfies({ chrome: ">74" }) && !this.parser.satisfies({ chrome: "<90" });
```
latin-panda, line 43: How are we handling when users have a Chrome 68, for example? What did we decide in that scenario?

latin-panda (changes requested):
Nice! I've added a couple of suggestions and a question. Is it possible to add an e2e test?

api/resources/translations/messages-es.properties
```diff
@@ -379,6 +379,9 @@ branding.favicon.field =
 branding.logo.field = 
 branding.title.field = 
 branding.title.field.help =
+browser.compatibility.title = Contacte a su supervisor
+browser.compatibility.description = Por favor contacte a su supervisor para actualizar su navegador.
+browser.compatibility.confirm = Bueno
```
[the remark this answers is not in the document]
michaelkohn: Just want to make sure we are consistent with caps, do we normally use `Ok` or `OK` elsewhere in the app?

webapp/src/ts/services/browser-detector.service.ts
michaelkohn: Also... is it supposed to be 74 or 72? Everywhere in the ticket it mentions 72 (but there may have been other conversations confirming it should be 74).

Benmuiruri: @michaelkohn In this other [issue](https://github.com/medic/cht-core/issues/8456#issuecomment-1677618103) Mr Jones discovered the forever spinner in webview 74. That's why I went with 74 since no version > than 74 has shown the forever spinner.

@latin-panda I think we may need to do extensive testing for Chrome 66-74 to confirm whether we get the forever spinner. What we know for now is we have not seen the forever spinnner for Chrome 75 -90, so for them we just warn them. 

Based on the extensive testing we can come up with the definitive version to completely block and the version to start warning.

api/resources/translations/messages-es.properties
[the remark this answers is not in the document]
Benmuiruri: @michaelkohn the existing `Ok` is the one for [Date confirmation](https://github.com/medic/cht-core/blob/6dec6344c60ca3c36ea267a475336a797a8b4172/api/resources/translations/messages-en.properties#L581) which is small caps. So that's what I used here to keep it consistent.

webapp/src/ts/services/browser-detector.service.ts
michaelkohn: Cool, thanks @Benmuiruri let's be sure to clarify what our final solution was in the [main issue](https://github.com/medic/cht-core/issues/7770) by editing the first comment.

webapp/src/ts/services/browser-detector.service.ts
```diff
@@ -39,6 +39,10 @@ export class BrowserDetectorService {
     });
   }
 
+  public isUsingOutdatedChromeBrowser() {
+    return this.parser.satisfies({ chrome: ">74" }) && this.parser.satisfies({ chrome: "<90" });
```
lorerod, line 43:
```suggestion
    return this.parser.satisfies({ chrome: '>74' }) && this.parser.satisfies({ chrome: '<90' });
```

webapp/src/ts/app.component.ts
```diff
@@ -646,6 +650,12 @@ export class AppComponent implements OnInit, AfterViewInit {
     });
   }
 
+  private warnOutdatedChrome(): void {
+    if (this.browserDetectorService.isUsingOutdatedChromeBrowser()) {
+      this.modalService.show(BrowserCompatibilityComponent);
```
lorerod, line 655: When manually testing, the button text is incorrect for a brief moment. Is this normal?

https://github.com/medic/cht-core/assets/21312057/b06c4b23-6377-4788-9102-c086909f5ff1

lorerod (changes requested):
Hi @Benmuiruri, this looks good overall to me. I left a couple of questions and suggestions. 
Also, it would be helpful if you added an end-to-end test; this is tricky. We are still determining if this works well, but you can look at how you can change the user agent in this [WIP PR](https://github.com/medic/cht-core/blob/8caddd561eb5b5a9c40fb5984fe898a0cfd3eaef/tests/e2e/default-mobile/wdio.conf.js#L25C15-L25C15).

webapp/src/ts/app.component.ts
lorerod: image

webapp/src/ts/services/browser-detector.service.ts
```diff
@@ -39,6 +39,10 @@ export class BrowserDetectorService {
     });
   }
 
+  public isUsingOutdatedChromeBrowser() {
```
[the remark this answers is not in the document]
Benmuiruri: Yes, the  browser compatibility modal is only for Chrome hence the naming. I could change it to `isUsingOutdatedBrowser` but I would not change the implementation of the method.

webapp/src/ts/services/browser-detector.service.ts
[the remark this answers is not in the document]
Benmuiruri: Yes that is correct. The last confirmed version that showed the forever spinner is Chrome 74. So in this PR we are mainly focusing on warning users with 75-90. In another PR we will address users with Chrome 74 and lower

Benmuiruri:
Hi @lorerod I have submitted my draft of the e2e test but it is not totally correct. I temporarily copied the[ wdio.config](tests/e2e/default-mobile/wdio.conf.js) and I changed the user agent according to [WIP PR](https://github.com/medic/cht-core/blob/8caddd561eb5b5a9c40fb5984fe898a0cfd3eaef/tests/e2e/default-mobile/wdio.conf.js#L25C15-L25C15). I even added some console.logs and I can see the Chrome version is being changed 
Screenshot 2024-01-04 at 16 14 10

However, I am seeing the `BrowserCompatibilityModal` when the test runs even when Chrome is version 95. Is it possible how I am changing the user agent is not really changing it correctly?

lorerod:
> However, I am seeing the `BrowserCompatibilityModal` when the test runs even when Chrome is version 95. Is it possible how I am changing the user agent is not really changing it correctly?

Hello, @Benmuiruri. The issue is that the test logs in before changing the user agent, which results in the modal appearing right after logging in. To resolve this problem, you can try rearranging the order of the tests and placing the `newChrome` scenario first. Alternatively, you can change the user agent before logging in to avoid this issue altogether.
For the `newChrome` scenario, add validation to check that the modal is not shown. Maybe using `await modalPage.checkModalHasClosed();` will work? Let me know what you think.

webapp/src/ts/app.component.ts
latin-panda: You can wait for translations like this 
```
private warnOutdatedChrome(): void {
    if (!this.browserDetectorService.isUsingOutdatedChromeBrowser()) {
       return; // Return early to keep code flat
    }
    await this.translationsLoaded;
    this.modalService.show(BrowserCompatibilityComponent);
}
```
The translations can fail to load or be slow.
I guess we still want to show the modal in those scenarios, right?

Benmuiruri: Yes we still want to show the modal. Using your suggestion the translations seems to be loading okay now. Even when I throttle the network. Thanks.

lorerod (approved):

latin-panda (approved):
Ship it 🚀

Issue #7167, opened 2021-06-30
Upgrade CHT-Conf version in CHT-Core


latin-panda:
**Describe the issue**
CHT-Core is using an old version of CHT-Conf, we couldn't upgrade before because of this [issue](https://github.com/medic/cht-conf/issues/396) which is already fixed.

Some app builders run the projects using the CHT-Conf defined in CHT-Core: `>> npx medic-conf [command]`. When doing this with the most recent version of CHT-Core will result of inline code errors (old eslint config) or will show errors while running some commands, etc. 

**Describe the improvement you'd like**
Upgrade CHT-Conf in CHT-Core

**Describe alternatives you've considered**
It's not a show stopper because they can install a newer version of CHT-Conf. But CHT-Core should be aligned with CHT-Conf as features become available. 

**Why Now?**
Partner(s): Internal
Reason(s): Internally needs this ticket to facilitate the release, or reduce technical debt

aidan-plenert-macdonald:
https://github.com/medic/cht-core/pull/7340

All tests are passing now. I marked it ready for review.

latin-panda:
To add a bit more info:
CHT Core is already using the `cht-conf`, but it also has `medic-conf` (see [package.json](https://github.com/medic/cht-core/blob/master/package.json)).

There is [a separate ticket to fix the Standard config](https://github.com/medic/cht-core/issues/7744), as part of that work the `medic-conf` will be uninstalled and we'll keep only `cht-conf`. 

However, there are some other places where we need to fix and it is this ticket work to do. Such as fixing the [grunt file](https://github.com/medic/cht-core/blob/master/Gruntfile.js#L513), fixing the[ GHA and its readme file](https://github.com/medic/cht-core/tree/master/.github/actions/deploy-with-medic-conf), test's [configs](https://github.com/medic/cht-core/blob/9ac9c1d598aa1b1949ed5444063467c4d6fec705/tests/apply-config.js#L2), and some [other appearances](https://github.com/medic/cht-core/blob/master/README.md) in different markup files.

jkuester:
@m5r, after digging into this, I think Jenni's breakdown from the Slack thread is a good representation of how we can split up this work!  Basically we have these 4 separate, but related, pieces that we should be able to work in parallel: 

- [ ] Update `standard` tests to use cht-conf
    -  (As noted above, this has [its own issue](https://github.com/medic/cht-core/issues/7744), but I guess can ship as part of these changes.)
[...]

PR #8756, opened 2023-12-13
chore(#7167): migrate scripts depending on medic-conf

m5r:
- upgraded cht-conf to 3.21.1
- migrated `build-config.sh` to use cht-conf
- removed unused `apply-config.js` script


.github/workflows/build.yml
```diff
@@ -191,6 +191,7 @@ jobs:
     - name: Install cht-conf
       run:  npm install -g cht-conf
     - run: npm ci
+    - run: cd config/default && npm ci
```
jkuester, line 194: Why did this become necessary? I guess a lot of our tests use this default config so it is more surprising to me that we never had to do this before.....

package.json
```diff
@@ -46,7 +46,7 @@
     "unit-haproxy": "cd haproxy/tests && make test",
     "wdio-cht-form": "wdio run ./tests/e2e/cht-form/wdio.conf.js",
     "wdio-default-mobile-local": "export VERSION=$(node ./scripts/build/get-version.js) && ./scripts/build/build-service-images.sh && wdio run ./tests/e2e/default-mobile/wdio.conf.js --suite=all",
-    "wdio-local": "export VERSION=$(node ./scripts/build/get-version.js) && ./scripts/build/build-service-images.sh && wdio run ./tests/e2e/default/wdio.conf.js",
+    "wdio-local": "wdio run ./tests/e2e/default/wdio.conf.js",
```
jkuester, line 49: Did you mean to commit this change?  (Sorry, I know I may have preemptively jumped on this PR before you were done....)

scripts/build/build-config.sh
```diff
@@ -2,9 +2,9 @@
 
 set -e
 
-echo "build-config: running medic-conf"
+echo "build-config: running cht-conf"
 
-node ./node_modules/medic-conf/src/bin/medic-conf.js \
+node ./node_modules/cht-conf/src/bin/index.js \
```
jkuester, line 7:
```suggestion
node ./node_modules/.bin/cht \
```

Minor, but one of the things I picked up from the `check-app-settings-is-committed`  files that might make this a bit cleaner (and less dependent on the position/naming of source files....)

jkuester (approved):
Wanted to get through this before signing off for the year, so sorry if this review is a bit pre-emptive!  Everything looks good! My only concern is the `wdio-local` change that I commented on before....

Issue #8074, opened 2023-02-13
Add a dropdown that gets populated with family members


dianabarsan:
**Is your feature request related to a problem? Please describe.**
In a discussion about interoperability, this question came up: "How do I establish a relationship between two contacts?". For example, selecting the mother/father of a child in a contact create form. 

Right now, we only support db-object searches by doc type, not by parent. So users are left in searching the whole list of contacts of a certain type to find the correct selection. This can prove cumbersome with duplicate names - which is a common occurrence in our deployments. 

**Describe the solution you'd like**
Add support for db-object searches by parent. This can be a compound search by type and parent as well. 

**Describe alternatives you've considered**
Another idea is to inject the children of a place directly in the form (via contact summary for example). 

**Additional context**
Question came up in a session about interoperability with a partner, concerning an existing deployment.

nKataraia:
I believe that solving this issue will enable us to create an accurate report for one of our important indicators. Currently, we are relying on estimates based on various heuristics. Therefore, we are eagerly looking forward to utilise this feature as soon as possible.

iesmail-znz:
@dianabarsan, currently we also have use case where a supervisor chooses a list of CHVs from a drop down menu. The drop down shows a list of chvs + clients. A workaround is to set a replication depth of 2, but that prevents the supervisors from viewing clients.

ChinHairSaintClair:
Related to [this](https://forum.communityhealthtoolkit.org/t/restrict-contact-select-to-a-specific-top-level-place/2803) CHT Forum thread.

In our case a CHW can be assigned work on multiple `households` and their `members`, the latter totaling in about 500 - many of which can have very similar names. 
It would be advantageous to 'restrict' the list of `members` returned by `select-contact` to a specific `household`.
Not only will that reduce the size of the list, it will also mitigate the issue where similarly named `member` from an entirely different `household` may be selected.

michaelkohn:
@latin-panda can you please summarize the solution in the first comment of this issue?

latin-panda:
Commonly, I add it in the PR's description for discussion with the reviewers, with test evidence and/or instructions, and so. 
To keep it practical, I am going to copy that and paste it here. Also, I prefer to not change Diana's text (the original author). 

## Solution

- This feature is applied when using the appearance:`descendant-of-current-contact`
- The scope remains for forms that are opened in the contact tab; it takes the contact ID from the URL. 
- It uses the CouchDB view `contact_by_parent` in combination with the contact type.

[...]

PR #8759, opened 2023-12-14
feat(#8074): add filter by parent to db-object widget

latin-panda:
# Description

- This feature is applied when using the appearance:`descendant-of-current-contact`
- The scope remains for forms that are opened in the contact tab; it takes the contact ID from the URL. 
- It uses the CouchDB view `contact_by_parent` in combination with the contact type.

**Usage**
| type | name | label:en | appearance |
|------|-------|------|------|
| string | a-person | select a person | select-contact type-person descendant-of-current-contact |

#8074

[CHT Docs PR](https://github.com/medic/cht-docs/pull/1254)

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8074-new-dbobject-filter/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8074-new-dbobject-filter/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8074-new-dbobject-filter/docker-compose/cht-couchdb-clustered.yml)

# License


latin-panda:
@jkuester I still need to complete the e2e, but I wanted your feedback on this work. I tried not to make big changes in the code and to keep it simple for app builders.

latin-panda:
@Benmuiruri @tatilepizs, welcome to review as well.

@tatilepizs, do you think waiting for Enketo's e2e refactor is better? I'm still trying to figure out how to check the dropdown options.

tatilepizs:
Hey,

> do you think waiting for Enketo's e2e refactor is better?

Depending on the urgency of this feature. I sent the [PR](https://github.com/medic/cht-core/pull/8720) to review yesterday, and I expect feedback until next year because a lot of us will be on vacation these last weeks of the year. So I don't think that the PR will be merged soon.

> I'm still trying to figure out how to check the dropdown options.

I didn't add a common option to validate dropdowns because I think that only one test that we have right now contains the dropdown widget. 
This is the tests ([enketo-widgets.wdio-spec.js](https://github.com/medic/cht-core/blob/master/tests/e2e/default/enketo/enketo-widgets.wdio-spec.js)) and this is the page object file ([enketo-widgets.wdio.page.js](https://github.com/medic/cht-core/blob/master/tests/page-objects/default/enketo/enketo-widgets.wdio.page.js)) that evaluates the widget.
If you think that it will be a good idea to include a dropdown selector in the common ones please let me know and we can do it 😄 

Thank you @latin-panda

webapp/src/ts/services/select2-search.service.ts
```diff
@@ -195,6 +205,15 @@ export class Select2SearchService {
     }
   }
 
+  private getContactId() {
+    let activeRoute = this.route.firstChild;
+    while (activeRoute?.firstChild) {
+      activeRoute = activeRoute.firstChild;
+    }
+
+    return activeRoute?.snapshot?.params?.id;
```
jkuester, line 214: Currently this `with-same-parent` functionality does not work with `contact` forms because those have a `parent_id` value here instead of `id`.  Not sure if this was intentional, but it seems like it might be useful to have this functionality in a contact form (especially if it is pretty simple to include here in the logic...)

jkuester (changes requested):
Code looks great! :star_struck: Very clean and well tested!

My only concern is around the naming of `with-same-parent`.  To me, that sounds like I will be filtering the searchable contacts to be limited to _siblings_ of the contact in context (the ones that have the same parent as the current contact).  However, with this implementation, it is actually limiting the searchable contacts to _children_ of the contact in context (and all their descendants).  

An appearance name more like `descendant-of-current-contact` would make more sense to me (not a huge fan of that name either, but it seems more accurate and I have not been able to think of a better one)....

Also, just a reminder that we definitely want to update the docs to describe this functionality!

webapp/src/ts/services/select2-search.service.ts
latin-panda: Thank you! Fixed

latin-panda:
@michaelkohn What do you think about Josh's idea of using `descendant-of-current-contact` instead of `with-same-parent`? 

IMO, it's long but descriptive. I prefer things to be descriptive :)

tests/e2e/default/enketo/db-object-widget.wdio-spec.js
```diff
@@ -0,0 +1,82 @@
...
+    _id: 'area2',
+    name: 'area 2',
+    type: 'health_center',
+    parent: { _id: districtHospital._id }
+  });
+
+  const offlineUser = userFactory.build({ place: districtHospital._id, roles: [ 'chw' ] });
+  const personArea1 = personFactory.build({ parent: { _id: area1._id, parent: area1.parent } });
+  const personArea2 = personFactory.build({ name: 'Patricio', parent: { _id: area2._id, parent: area2.parent } });
+
```
tatilepizs, line 42: Please move this section out of the `describe` method. We have seen some really weird behaviors when we have code inside of the `describe` method that does not encapsulate a specific test and as part of our clean process of the test folders, we want to move all of those code sections outside of the `describe` method.

tests/e2e/default/enketo/forms/db-object-widget.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0"?>
```
tatilepizs, line 1: Thank you for separating the `.xml` file ✨

tatilepizs (changes requested):
Thank you @latin-panda for adding the e2e test. 
I just left a comment in place.

tests/e2e/default/enketo/db-object-widget.wdio-spec.js
latin-panda: I made this change but am not buying the idea yet, perhaps because I'm missing the context of `why` it's considered a better practice. 

> We have seen some really weird behaviors when we have code inside of the describe

What kind of weird behaviors? 🤔

> and as part of our clean process of the test folders

They are going to be closer to the "imports", I prefer having a clear separation between imports and test data. And we're going to lose the scoping of variables - like when having more than one `describe`.

tatilepizs: @lorerod could help us here because she read about that while she was doing an investigation related to a flaky test that we were having at some point.

Lore, could you please share more details about why it is not a good idea to add blocks of code inside of the `describe` method that doesn't belong to a specific `it`?

lorerod: Hi @tatilepizs @latin-panda! I came across this when trying to fix a flaky test. You can find the comment and mocha issues in slack [here](https://medic.slack.com/archives/C02H68VMD/p1699645534120339?thread_ts=1699568113.589189&cid=C02H68VMD) 
The unexpected behavior might be caused when using `async` in the `describe` block or having `async` methods called inside the `describe` block. The `describe` block is primarily used for grouping related test cases. If we need an `async` setup for the entire suite, consider using `before` and `beforeEach` and applying async to those functions.
Considering this, it might be a good practice, if you are calling an async function, to organize the code inside `before`, `beforeEach`, `after`, `afterEach`, and it, and use describe only for grouping purposes.

lorerod: I haven't thoroughly checked, but I believe there is no `async` code within this `describe` block. Therefore, it is recommended to leave it as it is. However, we should all exercise caution when adding code inside the `describe` block.

latin-panda: Thanks for explaining! I get it, and it makes sense to me to make async code in `before`, `beforeEach`, `after`, `afterEach`, and `it`.
I have moved the test data declaration to inside the `describe` to make a clear separation with imports and make clear the scope.

jkuester (approved):
LGTM!

Benmuiruri (approved):
Awesome work!

latin-panda:
@tatilepizs When you have time, can you please review this again?

tests/e2e/default/enketo/db-object-widget.wdio-spec.js
tatilepizs: Thank you for the clarification @lorerod, I was confusing things 😅

tatilepizs (approved):
It looks good @latin-panda, thank you!! 🌟

Issue #8757, opened 2023-12-13
Remove standard config


mrjones-plip:
**Describe the issue**
In CHT Core there are some configs that we refer to in the docs and elsewhere, but [the `standard` config](https://github.com/medic/cht-core/tree/master/config/standard) has been abandoned and no end users are  using it that we know of.  It is tech debt which is tedious to maintain.  It may also confuse end users as to which config to build their app off of.

**Describe the improvement you'd like**
Remove the config and leave a `readme.md` in place telling users better places to look.

**Describe alternatives you've considered**
Leaving it in place and maintaining it.

PR #8761, opened 2023-12-14
chore(#7167): update dep versions for default/standard config

jkuester:
#7167 

Updates the various dependency versions for the `default` and `standard` config.  (For the `standard` config, I only bumped as high as possible without breaking anything and did not take any time to debug or try to fix any of the issues (since we're [deleting](https://github.com/medic/cht-core/issues/8757) `standard` soon).

Also updates various other places in the docs were we are referring to "medic-conf".

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7167_uplift_cht-conf_config-tests/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7167_uplift_cht-conf_config-tests/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7167_uplift_cht-conf_config-tests/docker-compose/cht-couchdb-clustered.yml)

# License


config/default/test/targets/pregnancy.spec.js
```diff
@@ -43,6 +43,7 @@ describe('Pregnancy related targets test', () => {
         expect(activePregnancies[0]).to.nested.not.include({ 'value.pass': 1 });
         expect(activePregnancies[0]).to.nested.not.include({ 'value.total': 1 });
       }
+      clock.restore();
```
jkuester, line 46: Theses were needed because the new version of `sinon` did not like calling `useFakeTimers` multiple times without restoring the previous clock first...

config/default/package.json
```diff
@@ -16,17 +16,17 @@
...
-    "cht-conf": "^3.16.2",
-    "cht-conf-test-harness": "^2.4.2",
-    "eslint-plugin-json": "^2.0.1",
-    "mocha": "^5.2.0",
-    "sinon": "^11.1.2"
+    "cht-conf": "^3.21.0",
+    "cht-conf-test-harness": "^2.4.3",
+    "eslint-plugin-json": "^3.1.0",
+    "mocha": "^10.2.0",
+    "sinon": "^12.0.1"
```
jkuester, line 27: For the record, I tried bumping to `^16.1.0` to align with the root package.json.  Unfortunately, there is some weirdness between sinon timers being used in the tests and the sinon timers being used in the cht-conf-test-harness.  The bad interaction does not happen for me with Node 18, but does fail with Node 16 (the current CI verison).  

Honestly, does not really seem worth spending too much time to debug at this point since hopefully it is easily resolved when we move to Node 18. However, if you feel it would be best to align versions, I can spend some more time digging into this!

mrjones-plip:
Updated body of ticket to denote we are indeed [deleting standard config](https://github.com/medic/cht-core/issues/8757)

config/standard/package.json
```diff
@@ -14,13 +14,13 @@
...
-    "eslint-plugin-json": "^1.3.2",
-    "cht-conf": "^3.7.0",
-    "cht-conf-test-harness": "^2.4.1",
-    "medic-conf-test-harness": "^0.1.19",
-    "mocha": "^5.1.1"
+    "chai": "^4.3.8",
+    "chai-exclude": "^2.1.0",
+    "eslint": "^8.49.0",
+    "eslint-plugin-json": "^3.1.0",
+    "cht-conf": "3.18.3",
```
m5r, line 21: I know we're getting rid of the standard config but while we're here updating dependencies, is there anything preventing us from upgrading `cht-conf` to the latest version here?

m5r (approved):
This is good to go, ship it!

config/standard/package.json
jkuester: Yeah, honestly now I cannot remember what it was, but something did break when going past this version...  As you said, though, the standard config is going with the :dodo: so no need to worry about this!

Issue #8771, opened 2023-12-21
Update task expiration recalculation queries


dianabarsan:
**Describe the performance issue**
Premises: 
1. Task emissions expire after 7 days, and require to be recalculated. 
2. When contacts are already loaded into rules-engine, only emissions that are associated to "dirty" contacts are recalculated - for performance reasons
3. When a large amount of contacts become dirty (after 7 days since last recalculation), existent view queries which use `keys` become very inefficient due to implementation details in the PouchDb adaper

**Describe the improvement you'd like**
Update the queries to not use `keys` param when recalculating emissions for a large number of contacts (200+). It's more performant to 

**Describe alternatives you've considered**
Drop task recalculation entirely - more complex. 

**To Reproduce**
Steps to record the performance metrics:
1. Load an user with a large number of contacts - I used one with 6000 contacts
2. Calculate tasks. Check telemetry entry: `rules-engine:tasks:all-contacts` and record value. 
3. Advance clock by 8 days. Reload task tab. Check telemetry entry: `rules-engine:tasks:all-contacts` and record value. 
4. Repeat step 3, adjusting clock back and forth to get multiple measurements. 

**Measurements**
| contacts | reports | Query style                   | duration | duration | duration | duration | AVG    | IDB Crashes |
|------------| ------ |-------------------------------|----------|----------|----------|----------|--------|-------------|
| 6661       | 10000 | keys (current)                | 293s   | 276s  | 270s   | 293s   | 283s | 3           |
| 6661       | 10000 | all rows (proposed)           | 64s    | 63s    | 66s   | 58s    | 63s  | 0           |
| 6661       | 10000  | start_key + end_key (attempt) | 100s   | 95s    | 92s    | 92s    | 95s  | 0           |

**Environment**
- Instance: local
- Browser: Chrome
- Client platform: Android
- App: webapp
- Version: 3.9+

**Additional context**
Instances that have offline users with a large number of contacts have very poor loading times sometimes.

dianabarsan:
Merged to `master`

PR #8772, opened 2023-12-21
perf(#8771): improve performance of task recalculation for many contacts

dianabarsan:
# Description

PouchDb is notoriously slow to query views with a large number of keys. This is what happened when task recalculation was triggered. For users with large numbers of contacts, these requests became so heavy it took minutes to complete, sometimes crashing PouchDb entirely. 

This adds a short-circuit to the key loaded query, when there are more than  500 keys, and instead gets all rows and filters them. 
The limit of 500 can be changed. In my sample test with ~10000 reports, it took roughly the same time to get all rows vs query with 500 keys. 

Performance difference: 
| contacts | reports | Query style                   | duration | duration | duration | duration | AVG    | IDB Crashes |
|------------| ------ |-------------------------------|----------|----------|----------|----------|--------|-------------|
| 6661       | 10000 | keys (current)                | 293s   | 276s  | 270s   | 293s   | 283s | 3           |
| 6661       | 10000 | all rows (proposed)           | 64s    | 63s    | 66s   | 58s    | 63s  | 0           |
| 6661       | 10000  | start_key + end_key (attempt) | 100s   | 95s    | 92s    | 92s    | 95s  | 0           |

#8771

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8771-skip-keys-param/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8771-skip-keys-param/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8771-skip-keys-param/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/shared-libs/rules-engine/src/pouchdb-provider.js
+++ b/shared-libs/rules-engine/src/pouchdb-provider.js
@@ -4,13 +4,14 @@
  * Wireup for accessing rules document data via medic pouch db
  */
 
-// TODO work out how to pass in the logger from node/browser
 /* eslint-disable no-console */
 const moment = require('moment');
 const registrationUtils = require('@medic/registration-utils');
 const uniqBy = require('lodash/uniqBy');
 
 const RULES_STATE_DOCID = '_local/rulesStateStore';
+const MAX_QUERY_KEYS = 500;
+
 const docsOf = (query) => {
   return query.then(result => {
     const rows = uniqBy(result.rows, 'id');
@@ -21,19 +22,31 @@ const docsOf = (query) => {
 const rowsOf = (query) => query.then(result => uniqBy(result.rows, 'id'));
 
 const medicPouchProvider = db => {
+  const dbQuery = async (view, params) => {
+    if (!params?.keys || params.keys.length < MAX_QUERY_KEYS) {
+      return db.query(view, params);
+    }
+
+    const keys = new Set(params.keys);
+    delete params.keys;
+    const results = await db.query(view, params);
+    const rows = results.rows.filter(row => keys.has(row.key));
+    return { ...results, rows };
+  };
+
   const self = {
     // PouchDB.query slows down when provided with a large keys array.
     // For users with ~1000 contacts it is ~50x faster to provider a start/end key instead of specifying all ids
     allTasks: prefix => {
       const options = { startkey: `${prefix}-`, endkey: `${prefix}-\ufff0`, include_docs: true };
-      return docsOf(db.query('medic-client/tasks_by_contact', options));
+      return docsOf(dbQuery('medic-client/tasks_by_contact', options));
     },
 
     allTaskData: userSettingsDoc => {
-      const userSettingsId = userSettingsDoc && userSettingsDoc._id;
+      const userSettingsId = userSettingsDoc?._id;
       return Promise.all([
-        docsOf(db.query('medic-client/contacts_by_type', { include_docs: true })),
-        docsOf(db.query('medic-client/reports_by_subject', { include_docs: true })),
+        docsOf(dbQuery('medic-client/contacts_by_type', { include_docs: true })),
+        docsOf(dbQuery('medic-client/reports_by_subject', { include_docs: true })),
         self.allTasks('requester'),
       ])
         .then(([contactDocs, reportDocs, taskDocs]) => ({ contactDocs, reportDocs, taskDocs, userSettingsId }));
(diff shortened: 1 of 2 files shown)


dianabarsan:
@kennsippell we've discussed about this before. Any suggestions / objections to this?

shared-libs/rules-engine/src/pouchdb-provider.js
```diff
@@ -52,9 +65,9 @@ const medicPouchProvider = db => {
 
     stateChangeCallback: docUpdateClosure(db),
 
-    commitTargetDoc: (targets, userContactDoc, userSettingsDoc, docTag, force = false) => {
-      const userContactId = userContactDoc && userContactDoc._id;
-      const userSettingsId = userSettingsDoc && userSettingsDoc._id;
+    commitTargetDoc: (targets, userContactDoc, userSettingsDoc, docTag, force = false) => { // NOSONAR
```
dianabarsan, line 68: Complains about # of params. Requires some major refactor.

shared-libs/rules-engine/src/pouchdb-provider.js
```diff
@@ -4,13 +4,14 @@
...
  */
 
-// TODO work out how to pass in the logger from node/browser
 /* eslint-disable no-console */
 const moment = require('moment');
 const registrationUtils = require('@medic/registration-utils');
 const uniqBy = require('lodash/uniqBy');
 
 const RULES_STATE_DOCID = '_local/rulesStateStore';
+const MAX_QUERY_KEYS = 500;
```
kennsippell, line 13: great results from this change! seems Pouch should essentially provide thisyes, we have discussed this - so it is nice to see it happen :D
this number was always such a sticking point for me... the results are good, but how did you arrive at `500`?

dianabarsan: I explained it in the PR description.

> The limit of 500 can be changed. In my sample test with ~10000 reports, it took roughly the same time to get all rows vs query with 500 keys. 

I think there is no winning number here, depending on doc counts, other numbers might be more efficient for some users (for a user with only 5000 reports, the best number would be 100 for example, but for a user with 20000 reports, the best number is 700). 

Do you have a suggestion of a different way of selecting this number?

kennsippell: We could plot a curve, but seems like over optimizing.  I'll start by just testing with another heavy dataset and reporting some numbers.

dianabarsan: Perfect,  that would be super helpful. 
If this is indeed blocking loading for heavy users, I'm inclined to just go with a sub-optimal number. Even if we plot a number, we would probably choose a round one, rather than 763. I think the number should be somewhere between 100 and 1000 (maybe 1000 is already way too much). 
I'm keen to plot a curve.

kennsippell:
13393 docs, 1015 contacts
Galaxy Tab A7 Lite (Production Hardware)

Branch | 1st | 2nd | 3rd
-- | -- | --
master | 57s | 68s | 72s
8771-skip-keys-param | 67s |  57s | 63s

Maybe not a huge win for CARES users ...

dianabarsan:
Yea, users with 1000 contacts really isn't the target here (our query limit is 500), the target were deployments where users end up with 5000-6000 contacts.

latin-panda (approved):
LGTM!

lorerod:
Hi @dianabarsan, sorry for the delay. I'm testing this locally. 
Instance up using [cht-couchdb.yml](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8771-skip-keys-param/docker-compose/cht-couchdb.yml) and [cht-core.yml](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8771-skip-keys-param/docker-compose/cht-core.yml). 

I first generated data using [this](https://github.com/medic/cht-core/tree/master/tests/scalability/generate-real-world-data) script, but I got errors when logging in with the created users.

So I decided to check with a simple data set and use the [scalability csv files](https://github.com/medic/cht-core/tree/master/tests/scalability/csv), and I'm getting the same error:

image

I will continue tomorrow to compare with the master branch and continue to try to make this work locally. It may be something on my local machine.

dianabarsan:
Hi @lorerod 

This error is unlikely to be caused by this change. Can you please check your instance and try again?

lorerod:
Ok, finally, I was able to test this with a user with 6.000 contacts. I got a similar average: 69s on `rules-engine:tasks:all-contacts`
LGTM! Thanks!

dianabarsan:
@kennsippell Are there any changes you desire or additional tests to be done? Thanks

kennsippell:
I think the change is excellent. I don't have access to NSSD data for testing so I'm not sure what dataset to try this with. Sorry that the ~1000 contact dataset I tried wasn't very instructive.

Issue #6299, opened 2020-03-15
Sync status sometimes says all reports synced when there are changes yet to sync


garethbowen:
**Describe the bug**
There is a race condition where we show "all reports synced" even though very recent changes may not have yet replicated.

![Screenshot from 2020-03-16 11-41-29](https://user-images.githubusercontent.com/2064938/76712203-4e44ee00-677b-11ea-9be4-1c1326d4f709.png)

**To Reproduce**
1. Set your internet connection to be very slow in dev tools to give yourself time to catch this bug
2. Open the hamburger menu and click "sync now"
3. Make a doc change (eg: review a report) before sync has finished
4. Open the hamburger menu and wait for the sync to finish
5. Notice it says "all reports synced"
6. Shut down the browser to stop any further replication
7. Log in to futon and check if the change made it to the server

**Expected behavior**
The app should only say "all reports synced" if we are sure this is the case.

**Environment**
- Instance: localhost
- Browser: Chrome
- Client platform: Linux
- App: webapp
- Version: 3.8.0

**Additional context**
This happens because after the successful sync we assume all the changes have replicated, but if a change occurs after the sync starts and before it completes then this change hasn't been replicated. 

Options include either 1) keep syncing until there are 0 changes to upload, _or_ 2) record the seq when we start replication and ensure it hasn't changed by the end. The second option also requires that we don't run the `to` and `from` replications in parallel.

garethbowen:
Moved to 3.12.0 to free up engineers to work on 3.11.0.

PR #8773, opened 2023-12-25
fix(#6299): trigger sync if something is left to sync

latin-panda:
# Description

- Refactors to smaller functions
- Fixes the last replicated sequence number to be the last highest `last_seq` from the replication response. 
- Triggers successive sync if something is left to sync, with a max of 2 extra successive syncs. 
- Removes unused imports

#6299

**Tests:**
1.. Sync when no changes to replicate should result in current_seq and last_seq being the same. 
2.. Sync changes that happened during a sync in progress (the goal of this work) 

https://github.com/medic/cht-core/assets/66472237/cf95186b-52f4-45d4-9597-754c092e8bf1

3.. No connection. It can't sync, it shouldn't retry immediately.

https://github.com/medic/cht-core/assets/66472237/65242de3-d215-43a0-836d-f491c07ed134

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6299-fix-sync-status/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6299-fix-sync-status/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6299-fix-sync-status/docker-compose/cht-couchdb-clustered.yml)

# License


webapp/src/ts/services/db-sync.service.ts
```diff
@@ -80,7 +79,7 @@ export class DBSyncService {
     private migrationsService:MigrationsService,
     private replicationService:ReplicationService,
   ) {
-    this.globalActions = new GlobalActions(store);
+    this.globalActions = new GlobalActions(this.store);
```
latin-panda, line 82: The IDE is marking the `store` (L77) as not used. This removes the warning.

webapp/src/ts/services/db-sync.service.ts
```diff
@@ -200,30 +252,9 @@ export class DBSyncService {
...
-      } else {
-        // definitely need to sync something
-        Object.keys(replicationErrors).forEach((directionName: 'to' | 'from') => {
-          syncState[directionName] = SyncStatus.Required;
-        });
-      }
-
+      const syncState = await this.makeBidirectionalReplication(force, quick);
       if (syncState.to === SyncStatus.Success) {
+        console.debug('Finished syncing!');
```
latin-panda, line 257: Nice to have when debugging

webapp/src/ts/services/db-sync.service.ts
```diff
@@ -141,8 +142,9 @@ export class DBSyncService {
   }
 
   private async replicateTo() {
-    if (!await this.authService.has('can_edit')) {
-      // not authorized to replicate - that's ok, skip silently
+    this.canReplicateToServer = await this.authService.has('can_edit');
+    if (!this.canReplicateToServer) {
+      console.debug('Not authorized to replicate - that\'s ok, skip silently');
```
latin-panda, line 147: Good help when debugging e2e tests

webapp/src/ts/services/db-sync.service.ts
```diff
@@ -166,31 +168,76 @@ export class DBSyncService {
...
+      return { to: SyncStatus.Success, from: SyncStatus.Success };
+    }
+
+    if (hasErrors && currentSeq === lastReplicatedSeq) {
+      // No changes to send, but may have some to receive
+      return { state: SyncStatus.Unknown };
+    }
+
+    // Definitely need to sync something
+    return { to: SyncStatus.Required, from: SyncStatus.Required };
```
latin-panda, line 232: For this specific ticket, I thought of including a new status for when there are no errors and the current seq is higher than the last replicated ( `if (!hasErrors && (!this.canReplicateToServer || currentSeq > lastReplicatedSeq))` ) to trigger a successive sync. But it was kind of the same as `SyncStatus.Required`, then for simplicity, I decided to leave it like this.

webapp/src/ts/services/db-sync.service.ts
```diff
@@ -166,31 +168,76 @@ export class DBSyncService {
...
-    return window.localStorage.getItem(LAST_REPLICATED_SEQ_KEY);
+  private setLastReplicatedSeq(sequence: number) {
+    if (!sequence) {
+      return;
+    }
+
+    const lastReplicatedSeq = this.getLastReplicatedSeq();
+    if (lastReplicatedSeq > sequence) {
+      return;
+    }
```
garethbowen, line 192: I don't think this block is necessary. If `lastReplicatedSeq` is greater than `sequence` then writing it to local storage will still mean replication is retried because they're not equal. So the code can be simplified by removing these 4 lines.

It's also more correct to update it because technically we know the replication is complete up to `sequence` so we should store that.

webapp/src/ts/services/db-sync.service.ts
```diff
@@ -166,31 +168,76 @@ export class DBSyncService {
...
+
+    window.localStorage.setItem(LAST_REPLICATED_SEQ_KEY, sequence.toString());
   }
 
   private getLastReplicationDate() {
     return window.localStorage.getItem(LAST_REPLICATED_DATE_KEY);
   }
 
+  private async makeBidirectionalReplication(replicateFromServer: boolean, successiveSyncs = 0) {
```
garethbowen, line 201: Thanks for breaking up the functions!

The naming on this one is a little complicated to me... I think it's because "bidirectional replication" is just a long way of saying "sync" which is what we're using throughout this service. Because we already have a `syncMedic` function it makes the naming of this difficult but feel free to change the naming of that function so how this and that work together is clear.

garethbowen (changes requested):
Very nice!

A couple of suggestions inline...

webapp/src/ts/services/db-sync.service.ts
latin-panda: I struggle with this naming, I renamed them like this:
syncMedic --> startSyncMedic 🫤 (open to suggestions)
makeBidirectionalReplication --> syncMedic

garethbowen (approved):
Very nice, thanks!

latin-panda:
Welcome back @tatilepizs! Can I get your blessing on this, please?

tests/page-objects/default/common/common.wdio.page.js
```diff
@@ -26,6 +26,7 @@ const getTasksButtonLabel = () => $('#tasks-tab .button-label');
 const getAllButtonLabels = async () => await $$('.header .tabs .button-label');
 const loaders = () => $$('.container-fluid .loader');
 const syncSuccess = () => $(`${hamburgerMenuItemSelector}.sync-status .success`);
+const syncInProgress = () => $('*="Currently syncing"');
```
tatilepizs, line 29: Thank you for creating the selector using the text! ⭐

tatilepizs (approved):
It looks really good and it works perfectly, thank you @latin-panda 🌟 

#### Reproducible on `master`

[master_sync-error](https://github.com/medic/cht-core/assets/94494491/755b1eea-24d8-4ef7-a0a3-355cedc91cf5)


#### Fixed on `6299-fix-sync-status`

[6299-fix-sync-status_fixed-sync-error](https://github.com/medic/cht-core/assets/94494491/2b023cd6-6d3f-4f15-894c-45022a813727)

Issue #8096, opened 2023-02-22
Admin pass change in CHT 4.x breaks CHT, shows wrong error message


mrjones-plip:
**Describe the bug**

**To Reproduce**
1. Spin up a CHT 4.1.0 instance using [CHT Docker Helper](https://docs.communityhealthtoolkit.org/apps/guides/hosting/4.x/app-developer/#cht-docker-helper-for-4x)
2. login as user `medic` , password `password`
3. go to hamburger menu -> user settings -> update password
4. change password
5. Password change is accepted, but an error is shown stating it wasn't accepted "Password is not correct."
6. CHT instance stops working because the `COUCHDB_PASSWORD` value in the environment variable file is now wrong

**Expected behavior**
Either password change is not allowed because it will break CHT or password change works and all services use new password

**Logs**
```
<150>Feb 22 00:02:23 haproxy[25]: 172.19.0.7,couchdb,201,12,0,0,PUT,/_users/org.couchdb.user%3Amedic,api,medic,'{"_id":"org.couchdb.user:medic","_rev":"1-5b7bf3319f91c46105aa45fbef077b6c","name":"medic","type":"
user","roles":["admin"],"password":"***"}',414,10,85,'node-fetch/1.0 (+https://github.com/bitinn/node-fetch)'                                                                                                      <150>Feb 22 00:02:23 haproxy[25]: 172.19.0.7,couchdb,200,3,0,0,GET,/_node/_local/_config/admins,-,medic,'-',324,2,97,'-'                                                                                           <150>Feb 22 00:02:23 haproxy[25]: 172.19.0.7,couchdb,200,3,0,0,GET,/_membership,-,medic,'-',253,1,74,'-'                                                                                                 
<150>Feb 22 00:02:23 haproxy[25]: 172.19.0.7,couchdb,200,7,0,0,PUT,/_node/couchdb@127.0.0.1/_config/admins/medic?raw=true,-,medic,'"-pbkdf2-21c4d152c87449a1cf60fd0989ca25f1307b1dc2,5b5034cec58639daf2022ab6e890c1358b604b9d,10"',314,5,87,'-'
[...]

PR #8776, opened 2023-12-29
fix(#8096): block admins from updating own password

garethbowen:
# Description

#8096 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8096-block-admins-from-updating-password/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8096-block-admins-from-updating-password/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8096-block-admins-from-updating-password/docker-compose/cht-couchdb-clustered.yml)

# License


m5r:
I don't know whether it's the desired behavior but I managed to change the admin password from the user table in the admin app
I noticed we have API code specific to this scenario in https://github.com/medic/cht-core/blob/master/shared-libs/settings/src/index.js#L155-L167, if we're deprecating this use case we could block it server-side and remove some of this code

garethbowen:
Excellent point, thanks @m5r !

I've updated the API with the same premise of blocking updating admin passwords which means throwing a 400 error when this is attempted through the API. This means users can't brick their install.

One day if we figure out how to update the env var in the container then we can allow this again, but for now this is safer...

m5r (approved):

tests/e2e/upgrade/admin-user.wdio-spec.js
```diff
@@ -1,34 +0,0 @@
...
-
-    await utils.request({
-      path: `/api/v1/users/${adminUser.username}`,
-      method: 'POST',
-      body: { password: newPassword }
-    });
-
-    await loginPage.login({ username: adminUser.username, password: newPassword, adminApp: true });
-  });
-});
```
garethbowen, line 34: @dianabarsan This test is literally trying to change the admin password which is no longer possible as of this PR. The functionality being testing is a duplicate of an integration test we have so I'm not clear on why we test it in the upgrade suite too. Do you remember? Was it actually testing the password change or was it also doing something else?

dianabarsan: Thanks @garethbowen . I think it served as a check that new admins can still get created and login. I think keeping the test that just creates the new admin user and allows them to login to the admin app should be ok?

garethbowen: Thanks! Added it back and it passed. Too easy.

Issue #8783, opened 2024-01-04
Sometimes manifest.json is empty


garethbowen:
**Describe the bug**

In my dev environment I notice browser console errors about invalid manifest.json and when I navigate there it's a blank file.

**To Reproduce**

1. Start API
2. Run `npm run build-dev`
3. Load the webapp
4. View the error in the console

**Expected behavior**

A valid manifest.json based on your branding configuration.

**Logs**

```
Manifest: Line: 1, column: 1, Syntax error.
manifest.json:1 
```

**Screenshots**

**Environment**
- Instance: local
- Browser: any
- Client platform: Ubuntu
- App: webapp
- Version: master

**Additional context**

Restarting API regenerates the manifest.json and fixes the issue.

garethbowen:
Downgraded to a technical issue because I think it only ever impacts development environments and not production ones.

garethbowen:
Merged.

PR #8784, opened 2024-01-04
fix(#8783): remove empty src manifest

garethbowen:
# Description

#8783

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8783-fix-empty-manifest/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8783-fix-empty-manifest/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8783-fix-empty-manifest/docker-compose/cht-couchdb-clustered.yml)

# License


garethbowen:
@dianabarsan This file has been around for a couple of years... I think the rework of the build scripts meant it got copied more, but I don't think an empty manifest.json is useful anywhee.

dianabarsan (approved):
Cool, You getting an empty manifest instead of the generated one could be a race condition in the dev deployment script, where we might copy files over and overwrite? 
Anyway, I think deleting this empty file is pretty safe.

garethbowen:
> You getting an empty manifest instead of the generated one could be a race condition in the dev deployment script, where we might copy files over and overwrite?

It's not a race condition as far as I can see. What's happening is if you run the dev script you get a blank manifest, then if you start API it generates a full manifest and overwrites the blank one. So it depends which order they run. Because we always want the generated one and never want the blank one it seems safe to rely on API generation only.

Issue #8791, opened 2024-01-08
Update deprecated functions used in generate real-world data script


lorerod:
Update faker library usages.

PR #8792, opened 2024-01-08
chore(#8791) update generate real world data script

lorerod:
# Description

Closes https://github.com/medic/cht-core/issues/8791

# Code review checklist

# Compose URLs

* __CHT_CORE_COMPOSE_URL__

# License

### DIFF
--- a/tests/factories/real-world/reports/create-family-household-survey.js
+++ b/tests/factories/real-world/reports/create-family-household-survey.js
@@ -9,7 +9,7 @@ module.exports = new Factory()
     }
     return Faker.faker.helpers.uniqueArray(
       ['boiling', 'filtration', 'chlorination', 'other'],
-      Faker.faker.datatype.number({ min: 1, max: 4 })
+      Faker.faker.number.int({ min: 1, max: 4 })
     ).join(' ');
   })
   .attr('kitchen', Faker.faker.datatype.boolean)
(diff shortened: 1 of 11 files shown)


tests/factories/real-world/reports/create-family-household-survey.js
```diff
@@ -9,7 +9,7 @@ module.exports = new Factory()
     }
     return Faker.faker.helpers.uniqueArray(
       ['boiling', 'filtration', 'chlorination', 'other'],
-      Faker.faker.datatype.number({ min: 1, max: 4 })
+      Faker.faker.number.int({ min: 1, max: 4 })
     ).join(' ');
   })
   .attr('kitchen', Faker.faker.datatype.boolean)
```
tatilepizs, line 15: According to [the documentation](https://fakerjs.dev/api/datatype.html#boolean) the `faker.datatype.boolean()` is a method that returns the boolean value true or false. 

Can you please change the places where the `datatype.boolean` is not used as a method?

I found these instances:
image

tatilepizs (changes requested):
Thank you @lorerod for updating all these methods that are deprecated ⭐ 

I was reading about `Faker` because I was not sure why we were updating these methods and found the documentation of the release notes to Faker v8, [here](https://fakerjs.dev/guide/upgrading.html#upgrading-to-v8) it is in case anyone else is curious 😅 

I just left one small comment.

tatilepizs (approved):
Thank you for the updates @lorerod ⭐

Issue #8462, opened 2023-08-16
Expose user device details to admins


garethbowen:
**Is your feature request related to a problem? Please describe.**
Technical administrators should be able to find devices which are unsupported or likely to be slow so that they can proactively replace or upgrade them and keep users engaged and productive.

**Describe the solution you'd like**
Provide a feature to show administrators key information about their users and their devices. This could be done in the admin app, or a dashboard, a Mobile Device Management service, or through Watchdog (or some combination of the above).

Some ideas for information to show are...

- Browser version
- Android OS version
- apk version
- Last synced time
- Device specs
- CHT Core version (has the user synced and reloaded since the update was installed)
- Configuration version (has the user synced and reloaded since the config was updated)

The critical thing is for this to be actionable. Some things admins should be working on are,

- Which users have unsupported browsers.
- Which users have browsers which are not supported by the version of the CHT we want to upgrade to.
- Which users have software with known vulnerabilities.
- Which users have old hardware which is due to be replaced.
- etc

Remember that users can have multiple devices and we should list all.

**Describe alternatives you've considered**
This could be managed outside of the CHT but we already collect much of this information, so it makes sense to expose it somewhere.

m5r:
We can retrieve all of this information except "Last synced time" from the telemetry data gathered from users' devices. To build this kind of bird's eye view of a deployment's devices, we need to get each device's last telemetry entry and compile that data to be consumed by the frontend.
It's fair to assume that we will work with thousands of telemetry entries so depending on how that data will be consumed, we are probably going to need to process the data in advance so that queries stay fast.

With this in mind, I see 3 potential solutions right off the bat:
- Solution <span>#</span>1 - naive approach - display data in the admin app
	- admin app calls the API
	- API queries couch for telemetry entries
	- API processes telemetry entries on the fly
- Solution <span>#</span>2 - scheduled data processing - display data in the admin app
	- sentinel queries couch for telemetry entries and processes them on a schedule, preferably when couch is not in high demand (maybe once a day, at night?)
	- admin app calls the API
	- API queries couch for the document produced by sentinel
- Solution <span>#</span>3 - scheduled data processing - display data in some kind of analytics app
	- process telemetry data on a schedule with cht-sync
	- show it in analytics app used by deployment (klipfolio or similar)
even faster query

I briefly experimented with the first two solutions in [this branch](https://github.com/medic/cht-core/compare/master...expose-user-devices-spike).
[...]

PR #8797, opened 2024-01-10
feat(#8462): Export users devices details to CSV

m5r:
# Description

#8462

# Code review checklist

# Compose URLs

* __CHT_CORE_COMPOSE_URL__

# License


m5r:
@jkuester thanks for catching the date comparison bug. I couldn't reproduce the admin UI bug you mentioned but my guess is the export is taking a looong time and there is just no feedback on the UI to reflect that so I went ahead and added some.  
We already had a property in the export template to give a "loading" feedback but the value it relied on was never set.

admin/src/js/controllers/export-dhis.js
```diff
@@ -63,6 +63,10 @@ angular.module('controllers').controller('ExportDhisCtrl',
     });
 
     $scope.export = () => {
+      if ($scope.exporting) {
```
jkuester, line 66: Can we add a unit test to the `export-dhis.spec.js` for this?

admin/src/js/modules/ajax-download.js
```diff
@@ -2,18 +2,37 @@
...
    */
   exports.download = function(url) {
-    const element = document.createElement('a');
-    element.setAttribute('href', url);
-    element.setAttribute('download', DEFAULT_FILE_NAME);
-    element.style.display = 'none';
-    document.body.appendChild(element);
-    element.click();
-    document.body.removeChild(element);
+    return fetch(url)
```
jkuester, line 25: Not a huge deal, but we did change up the logic quite a bit in this file.  Would be nice to have some unit tests if they are feasible.

jkuester: That being said, why did we need to change up the default file name here?  (obviously the current approach is not ideal, but is this just an enhancement on top of our User Devices functionality, or did we need this for some reason?)

jkuester (changes requested):
Very nice!  Got a number of minor questions/comments that I left in-line.  Regarding e2e tests, I wonder if it would be sufficient to just add a test or two to `tests/integration/api/controllers/export-data.spec.js`.  Seems like an integration test where we can manually create the `medic-users-meta` data is probably the best approach to testing this functionality (since we don't want to wait for Sentinal).  

Also, don't forget a cht-docs PR for the new API!

admin/src/js/modules/ajax-download.js
m5r: This is more of an enhancement. With this approach, we know when the download starts and when it ends. It enables us to give feedback to the user and prevent them from requesting a duplicate export while one is being processed.

admin/src/templates/export.html
```diff
@@ -14,5 +14,8 @@
   <li role="presentation" ui-sref-active="active" mm-auth="can_export_dhis">
     <a ui-sref="export.dhis" translate>DHIS2</a>
   </li>
+  <li role="presentation" ui-sref-active="active">
+    <a ui-sref="export.user-devices" translate>User devices</a>
```
latin-panda, line 18: I'd suggest using translation keys. For example `export.tabs.user_devices`
That way we don't need to escape the space, instead of `User\ devices =` we can have `export.tabs.user_devices = User devices`

admin/src/templates/export_user-devices.html
```diff
@@ -0,0 +1,9 @@
+<div class="col-sm-12">
+  <div>
+    <p translate>export.user-devices.description</p>
+    <a class="btn btn-default" ng-click="export()" ng-disabled="exporting">
+      <i class="fa fa-arrow-down"></i>
+      <span translate>Download</span>
```
latin-panda, line 6: Same here about translation keys. Like in L3 to keep consistency: `export.button.download`

api/resources/translations/messages-es.properties
```diff
@@ -259,6 +259,7 @@ Update\ Facility = Actualizar contacto
 Upload\ Forms = 
 Upload\ failed = Falla en el envio
 Upload\ succeeded = Envio exitoso
+User\ devices =
```
latin-panda, line 262: Translation:
Dispositivos de usuario

api/resources/translations/messages-en.properties
```diff
@@ -686,6 +687,7 @@ export.feedback.description = Download a log of detected errors and user feedbac
 export.messages.description = Download all messages that have ever been sent or received in CSV format.
 export.people.description = Download all contacts registered in the system in JSON format.
 export.reports.description = Download a summary of all the reports that have ever been submitted in CSV format.
+export.user-devices.description = Download a summary of the versions of the different pieces of software the CHT uses from users' devices in CSV format.
```
latin-panda, line 690: What do you mean by `pieces of software the CHT uses`?

api/resources/translations/messages-es.properties
```diff
@@ -646,6 +647,7 @@ export.feedback.description =
 export.messages.description = 
 export.people.description = 
 export.reports.description = 
+export.user-devices.description =
```
latin-panda, line 650: Translation:
Descargue un resumen de las versiones de los diferentes software que utiliza el CHT desde los dispositivos de los usuarios en formato CSV

latin-panda (changes requested):
Nice job! Josh cover a lot already, I just added some minor ones I spotted

api/src/services/export/user-devices-mapper.js
```diff
@@ -0,0 +1,22 @@
+const db = require('../../db');
+
+module.exports = async () => {
+  const { rows } = await db.medicUsersMeta.query('users-meta/device_by_user', { group: true });
+  return rows.map(doc => {
+    const user = doc.key;
+    const date = doc.value.date;
+    const deviceId = doc.value.device.deviceId;
+    const webview = /Chrome\/(\d+\.\d+)/.exec(doc.value.device.userAgent)?.[1];
```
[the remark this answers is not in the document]
m5r: I believe it will, good catch! I focused only on CHWs devices that are running the CHT in a webview but it makes sense that an admin would want to know if let's say a supervisor is using an unsupported version of Firefox.

What do you think of renaming this field from `webview` to `browser` and setting the value to `"$browser_name $major_version"`?
It would be safer to rely on a reliable useragent parser like bowser instead of dealing with edge cases ourselves. It's the same dependency we use to detect unsupported browsers on the login page and in the BrowserDetectorService. The `user-devices-mapper.js` code would look like this:
```js
const Bowser = require('bowser');

const { rows } = await db.medicUsersMeta.query('users-meta/device_by_user', { group: true });
return rows.map(doc => {
  // ...
  const { browser } = Bowser.parse(doc.value.device.userAgent);
  return {
    // ...
    browser: `${browser.name} ${browser.version}`,
    // ...
  };
});
```

ddocs/users-meta-db/users-meta/views/device_by_user/reduce.js
```diff
@@ -0,0 +1,11 @@
+function (keys, values) {
+  let latest = { date: '1970-01-01' };
+  values.forEach(function (value) {
+    // using Date.parse to compare dates because they're not well-formatted to compare them as raw strings
+    // e.g. '2020-8-4' instead of '2020-08-04'
+    if (Date.parse(value.date) > Date.parse(latest.date)) {
```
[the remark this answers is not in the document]
m5r: Yes, excellent points!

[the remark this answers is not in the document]
m5r: Actually I think we can skip the date parsing altogether. What we're trying to do here is getting the latest entry, right? We can keep doing string comparisons as long as these date strings are formatted consistently. We can do something like:
```js
const pad = number => number.toString().padStart(2, "0");
const date = `${doc.metadata.year}-${pad(doc.metadata.month)}-${pad(doc.metadata.day)}`;
```
This way we can continue the plain `value.date > latest.date` comparison with just strings

ddocs/.eslintrc
```diff
@@ -7,6 +7,6 @@
     "no-var": "off"
   },
   "parserOptions": {
-    "ecmaVersion": 5,
+    "ecmaVersion": 6,
```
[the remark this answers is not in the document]
m5r: Yes, I got the info from https://blog.couchdb.org/2020/02/26/the-road-to-couchdb-3-0-update-to-javascript-engine/ and it seems to match my experience when testing a few newer syntax. For example the version of SpiderMonkey it's using supports [template literals](https://caniuse.com/template-literals) but fails when trying to parse [optional chaining](https://caniuse.com/mdn-javascript_operators_optional_chaining) and that matches with what caniuse.com is reporting.

api/resources/translations/messages-en.properties
m5r: It represents the version of the user's browser, Android OS, and of the CHT instance. I'm not 100% satisfied with this wording, you got an idea of how to express this more clearly?

cc @mrjones-plip @garethbowen if you have an alternative wording to suggest

ddocs/users-meta-db/users-meta/views/device_by_user/map.js
```diff
@@ -0,0 +1,18 @@
+function(doc) {
+  if (doc.type === 'telemetry') {
+    emit(doc.metadata.user, {
```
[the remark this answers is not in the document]
m5r: Let me know if that amount of null check sounds reasonable for our needs

api/src/services/export/user-devices-mapper.js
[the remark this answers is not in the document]
jkuester: +1000 for `bowser`!  I was a little unsure about combining both `name` and `version` into the same field. :thinking: Trying to decide how a consumer would be most likely to use this data....  My current thought is that what you have done makes the most sense and keeps everything nice and simple.  It is not like consumers can do a simple number/alpha sort on browser versions to "get the earliest" anyway.  So, including the browser name in the string is not going to add much of an extra "burden" to processing the data.

api/resources/translations/messages-en.properties
latin-panda: I think I get what you mean, but I'm not sure how to best say it in proper English; adding what those software are can help explain:
 
`Download a summary of software versions related to CHT (browser, CHT instance and OS) running in the user's devices. The summary is exported in CSV format.`

Once we have the message defined, I'll get the translation in Spanish and Indonesian, Ben can help with the Swahili and Sugat with the Nepali 😎 (our diversity feels powerful)

api/src/services/export/user-devices-mapper.js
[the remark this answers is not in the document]
m5r: I was initially had the CSV formatted export in mind but since we're giving back JSON it'd be fine to nest data. JSON -> CSV tools know how to deal with this scenario anyway.

admin/src/js/modules/ajax-download.js
m5r: Sorry forgot to answer your second point about the file name. The default file name is still the same but we need to explicitly set the file name with this API because it doesn't automatically extract it from the `Content-Disposition` header, so we do it ourselves. The behavior is the same: if there is such a header, it uses the file name provided, otherwise it falls back to the default file name.

m5r:
I opened a [docs PR](https://github.com/medic/cht-docs/pull/1290) to document this new route

api/resources/translations/messages-en.properties
latin-panda: @m5r what do you think about this suggestion?

m5r: Oops sorry for missing it, it's a good one! If @jkuester agrees too, we should go ahead and get it translated  
I'll take care of the French one :smile:

latin-panda: These are some translation you can use:

> User devices

- Spanish:
  - Dispositivos de usuario
- Indonesia:
  - Perangkat pengguna
- Swahili:
  - Vifaa vya mtumiaji

> Download a summary of software versions related to CHT (browser, CHT instance and OS) running in the user's devices. The summary is exported in CSV format.

- Spanish:
  - Descargue un resumen de las versiones de software relacionadas con CHT (navegador, instancia de CHT y sistema operativo) que se ejecutan en los dispositivos del usuario. El resumen se exporta en formato CSV.
- Indonesia:
  - Unduh ringkasan versi perangkat lunak terkait CHT (browser, instansi CHT, dan OS) yang berjalan di perangkat pengguna. Ringkasan tersebut diekspor dalam format CSV.
- Swahili:
  - Pakua muhtasari wa matoleo ya programu yanayohusiana na CHT (kivinjari, mfano wa CHT na tofauti za mifumo) unaoendeshwa kwenye vifaa vya mtumiaji. Muhtasari huo unahamishwa katika umbizo wa CSV.



Swahili by @Benmuiruri, thanks!

latin-panda (commented):
LGTM! Just waiting for the translations

BTW I tried to test it but when downloading the file I got: `missing_named_view` 
What are the steps to test it from a fresh instance?

api/resources/translations/messages-en.properties
m5r: Thanks a lot! :raised_hands:

api/tests/mocha/services/export-data.spec.js
lorerod, line 1: Thank you, @m5r, for adding export user devices scenarios to this spec.

lorerod (approved):
We have the basic cover with the unit test added to this PR. 
As suggested in [the PR comment](https://github.com/medic/cht-core/pull/8797#issue-2074890164), we can evaluate e2e implementation in future iterations.

latin-panda (approved):
LGTM!

admin/src/js/modules/ajax-download.js
latin-panda: > to support the $scope.exporting logic to disable the download buttons wile the current request is processing

Ah okay, so yes, using `fetch` and return a promise makes sense here.

admin/src/js/modules/ajax-download.js
```diff
@@ -4,16 +4,27 @@
...
+        .then(blob => {
+          const blobUrl = URL.createObjectURL(blob);
+          const link = document.createElement('a');
+          link.style.display = 'none';
+          link.href = blobUrl;
+          link.setAttribute('download', getFileName(response.headers));
+          document.body.appendChild(link);
+          link.click();
+          URL.revokeObjectURL(blobUrl);
+        }));
```
latin-panda, line 28: Minor
Maybe I was reading too fast, but my mind automatically closed the first `then` at the end of `response.blob()` and thought the second `then` had bad indentation. This might help readability a tiny bit more while keeping code flat:
```suggestion
      .then(response => response.blob())
      .then(blob => {
        const blobUrl = URL.createObjectURL(blob);
        const link = document.createElement('a');
        link.style.display = 'none';
        link.href = blobUrl;
        link.setAttribute('download', getFileName(response.headers));
        document.body.appendChild(link);
        link.click();
        URL.revokeObjectURL(blobUrl);
      }));
```

admin/tests/unit/controllers/export-user-devices.spec.js
```diff
@@ -0,0 +1,53 @@
+describe('User devices export controller', () => {
+  const { expect } = chai;
+
```
latin-panda, line 3:
```suggestion
describe('User devices export controller', () => {
  'use strict';
  
  const { expect } = chai;

```
Seems we are adding this everywhere.

api/src/services/export/user-devices.js
```diff
@@ -0,0 +1,37 @@
...
+    const deviceId = doc.value.device.deviceId;
+    const browser = doc.value.device.userAgent && getBrowser(doc.value.device.userAgent);
+    const { apk, android, cht, settings } = doc.value.device.versions;
+    return {
+      user,
+      deviceId,
+      date,
+      browser: {
+        name: browser?.name || undefined,
+        version: browser?.version || undefined,
```
latin-panda, line 29: Minor 
Do these sometimes have values like `null` or `false`? We can just omit the `|| undefined` and just let it be falsy.

latin-panda (changes requested):
Nice improvements! Just minor comments below

api/src/services/export/user-devices.js
jkuester: So, if there is a `userAgent` string, but `bowser` cannot parse it, then we get a `browser` object, but `name` and `version` are both empty strings.  I added `|| undefined` to try and unify the behavior for the cases were there is no `userAgent` and the `userAgent` string is not valid. Without  `|| undefined`, this code will return `browser: { name: undefined, version: undefined }` when there is no `userAgent` and `browser: { name: '', version: '' }` when the `userAgent` cannot be parsed.

Let me know if you think it is fine to just have different behavior (or if there is a way we could make this behave better)!

jkuester:
@latin-panda Could I get one more quick review!  The main change is a small behavior switch based on [Gareth's feedback](https://github.com/medic/cht-core/issues/8462#issuecomment-1956005614).

latin-panda (approved):
LGTM! Thanks Josh!

PR #8805, opened 2024-01-11
fix(#7110): update homeplace in LHS contacts list

### LINKED ISSUE #7110: Homeplace in LHS contacts list is not updated on changes
**Describe the bug**
For offline users, the LHS contacts list should update when changes are made to those respective contacts. But this is not happening when the homeplace is edited.

**To Reproduce**
1. Login as an offline user and go to the contacts page. (desktop). 
2. In a different browser, login as an admin and edit the homeplace of the offline user, changing the name in an obvious way.
3. Back as the offline user, sync. 
4. The RHS (if you had your homeplace details on the RHS) is updated to show the new name, but the name is not changed in the RHS. 

**Expected behavior**
Both RHS and LHS should change. 

**Screenshots**
image

**Environment**
- App: webapp
- Version: 3.11 but I suspect this was an earlier bug

Benmuiruri:
# Description

Fix bug to allow `userHomePlace` to update on sync. 

Closes #7110 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7110-update-user-home-place/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7110-update-user-home-place/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7110-update-user-home-place/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/tests/e2e/default/contacts/edit-person-home-place.wdio-spec.js
+++ b/tests/e2e/default/contacts/edit-person-home-place.wdio-spec.js
@@ -1,0 +1,100 @@
+const loginPage = require('@page-objects/default/login/login.wdio.page');
+const usersAdminPage = require('@page-objects/default/users/user.wdio.page');
+const commonPage = require('@page-objects/default/common/common.wdio.page');
+const contactPage = require('@page-objects/default/contacts/contacts.wdio.page');
+const utils = require('@utils');
+
+describe('Edit Person Under Area', () => {
+  const placeFactory = require('@factories/cht/contacts/place');
+  const personFactory = require('@factories/cht/contacts/person');
+  const places = placeFactory.generateHierarchy();
+  const districtHospital = places.get('district_hospital');
+
+  const offlineUsername = 'jack_test';
+  const onlineUsername = 'program_officer';
+  const password = 'Jacktest@123';
+  const editedPlaceName = 'Updated Health Center';
+
+  const healthCenter = placeFactory.place().build({
+    name: 'HealthCenter',
+    type: 'health_center',
+    parent: {
+      _id: districtHospital._id,
+      parent: {
+        _id: '',
+      },
+    },
+  });
+
+  const offlinePerson = personFactory.build({
+    name: 'Jack',
+    parent: {
+      _id: healthCenter._id,
+      parent: healthCenter.parent,
+    },
+  });
+
+  const onlinePerson = personFactory.build({
+    name: 'Program Officer',
+    parent: {
+      _id: healthCenter._id,
+      parent: healthCenter.parent,
+    },
+  });
+
+  const docs = [...places.values(), healthCenter, onlinePerson, offlinePerson];
+
+  beforeEach(async () => {
+    await utils.saveDocs(docs);
+    await loginPage.cookieLogin();
+    await usersAdminPage.goToAdminUser();
+    await usersAdminPage.openAddUserDialog();
+    await usersAdminPage.inputAddUserFields(
+      offlineUsername,
+      'Jack',
+      'chw',
+      healthCenter.name,
+      offlinePerson.name,
+      password
+    );
+    await usersAdminPage.saveUser();
+
+    await usersAdminPage.openAddUserDialog();
+    await usersAdminPage.inputAddUserFields(
+      onlineUsername,
+      'Program Officer',
+      'program_officer',
+      healthCenter.name,
+      onlinePerson.name,
+      password
+    );
+    await usersAdminPage.saveUser();
+
+    await browser.reloadSession();
+  });
+
+  it('can sync and update offlineUser HomePlace', async () => {
+    await browser.url('/');
+    await loginPage.login({ username: offlineUsername, password });
+    await commonPage.waitForPageLoaded();
+    await commonPage.goToPeople();
+    await commonPage.logout();
+
+    await browser.url('/');
+    await loginPage.login({ username: onlineUsername, password });
+    await commonPage.waitForPageLoaded();
+    await commonPage.goToPeople();
+    await contactPage.editPlace(editedPlaceName);
+    await commonPage.waitForPageLoaded();
+    await commonPage.logout();
+
+    await browser.url('/');
+    await loginPage.login({ username: offlineUsername, password });
+    await commonPage.waitForPageLoaded();
+    await commonPage.goToPeople();
+
+    expect(await contactPage.getContactCardText()).to.equal(healthCenter.name);
+    await commonPage.sync();
+    expect(await contactPage.getContactCardText()).to.equal(editedPlaceName);
+  });
+});
--- a/tests/e2e/default/contacts/edit-person-home-place.wdio-spec.js
+++ b/tests/e2e/default/contacts/edit-person-home-place.wdio-spec.js
@@ -1,95 +1,52 @@
 const loginPage = require('@page-objects/default/login/login.wdio.page');
-const usersAdminPage = require('@page-objects/default/users/user.wdio.page');
 const commonPage = require('@page-objects/default/common/common.wdio.page');
 const contactPage = require('@page-objects/default/contacts/contacts.wdio.page');
 const utils = require('@utils');
+const placeFactory = require('@factories/cht/contacts/place');
+const userFactory = require('@factories/cht/users/users');
+const personFactory = require('@factories/cht/contacts/person');
 
 describe('Edit Person Under Area', () => {
-  const placeFactory = require('@factories/cht/contacts/place');
-  const personFactory = require('@factories/cht/contacts/person');
   const places = placeFactory.generateHierarchy();
-  const districtHospital = places.get('district_hospital');
+  const healthCenter = places.get('health_center');
 
-  const offlineUsername = 'jack_test';
-  const onlineUsername = 'program_officer';
-  const password = 'Jacktest@123';
+  const offlineUserContact = personFactory.build({ parent: healthCenter });
+  const onlineUserContact = personFactory.build({ parent: healthCenter });
+  const offlineUser = userFactory.build({
+    username: 'offline_user',
+    place: healthCenter._id,
+    roles: ['chw'],
+    contact: offlineUserContact
+  });
+  const onlineUser = userFactory.build({
+    username: 'online_user',
+    place: healthCenter._id,
+    roles: ['program_officer'],
+    contact: onlineUserContact
+  });
   const editedPlaceName = 'Updated Health Center';
 
-  const healthCenter = placeFactory.place().build({
-    name: 'HealthCenter',
-    type: 'health_center',
-    parent: {
-      _id: districtHospital._id,
-      parent: {
-        _id: '',
-      },
-    },
-  });
+  const docs = [...places.values()];
 
-  const offlinePerson = personFactory.build({
-    name: 'Jack',
-    parent: {
-      _id: healthCenter._id,
-      parent: healthCenter.parent,
-    },
-  });
-
-  const onlinePerson = personFactory.build({
-    name: 'Program Officer',
-    parent: {
-      _id: healthCenter._id,
-      parent: healthCenter.parent,
-    },
-  });
-
-  const docs = [...places.values(), healthCenter, onlinePerson, offlinePerson];
-
-  beforeEach(async () => {
+  before(async () => {
     await utils.saveDocs(docs);
-    await loginPage.cookieLogin();
-    await usersAdminPage.goToAdminUser();
-    await usersAdminPage.openAddUserDialog();
-    await usersAdminPage.inputAddUserFields(
-      offlineUsername,
-      'Jack',
-      'chw',
-      healthCenter.name,
-      offlinePerson.name,
-      password
-    );
-    await usersAdminPage.saveUser();
-
-    await usersAdminPage.openAddUserDialog();
-    await usersAdminPage.inputAddUserFields(
-      onlineUsername,
-      'Program Officer',
-      'program_officer',
-      healthCenter.name,
-      onlinePerson.name,
-      password
-    );
-    await usersAdminPage.saveUser();
-
-    await browser.reloadSession();
+    await utils.createUsers([offlineUser, onlineUser]);
   });
 
   it('can sync and update offlineUser HomePlace', async () => {
-    await browser.url('/');
-    await loginPage.login({ username: offlineUsername, password });
+    await loginPage.login(offlineUser);
     await commonPage.waitForPageLoaded();
     await commonPage.goToPeople();
     await commonPage.logout();
 
-    await browser.url('/');
-    await loginPage.login({ username: onlineUsername, password });
+    await loginPage.login(onlineUser);
     await commonPage.waitForPageLoaded();
     await commonPage.goToPeople();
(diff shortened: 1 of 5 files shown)


webapp/src/ts/modules/contacts/contacts.component.ts
```diff
@@ -356,9 +356,10 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
...
     let docIds;
     if (options.withIds) {
-      docIds = this.contactsList.map((item) => {
-        return item._id;
-      });
+      docIds = this.contactsList.map((item) => item._id);
+      if (this.usersHomePlace && !docIds.includes(this.usersHomePlace._id)) {
+        docIds.push(this.usersHomePlace._id);
+      }
     }
```
latin-panda, line 363: This function is too long and getting complex, I'm wondering why sonar didn't warn

Alternatively, this can be extracted to a function that returns docIds.

webapp/src/ts/modules/contacts/contacts.component.ts
```diff
@@ -369,19 +370,22 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
           const homeIndex = _findIndex(updatedContacts, (contact:any) => {
```
latin-panda, line 370: Please extract from 370 to 389 to a function, this is all about the homeplace

latin-panda (changes requested):
Nice job! I left a couple of suggestion. In the meantime I'm testing it. 
Also, is there any unit test we can add to catch this bug?

latin-panda:
It's working fine for me.

Please add a video showing the test scenarios you covered. You can see how I did it here as example ([PR description](https://github.com/medic/cht-core/pull/8773#issue-2055577618))

tatilepizs:
Hi @Benmuiruri,

I was testing it and you can add an automated test just by playing with the users that are logged in, you don't need to use a different browser to see the error. See the process that I followed using `master` branch:

[video](https://github.com/medic/cht-core/assets/94494491/5ccb7a0d-4cd5-4247-a85c-3a1b71720591)

Now, I also tested it using your branch (`7110-update-user-home-place`), and as Jenny said [here](https://github.com/medic/cht-core/pull/8805#issuecomment-1888694358) I can still see the error. I tried using an offline user with one and two roles and it failed both times, it wasn't until I changed tabs (for example) that I could see the change. See the video below:

[video](https://github.com/medic/cht-core/assets/94494491/47176fdd-afaf-42b4-a173-819fe21bc6a1)

Benmuiruri:
Hi @latin-panda  and @tatilepizs  here is a video of the branch currently working. let me know if it works for you. 

https://github.com/medic/cht-core/assets/2597305/54dddcf1-2b55-4881-820f-5a08ca166da4

tatilepizs:
It is working perfectly now Ben 🎉 

[video](https://github.com/medic/cht-core/assets/94494491/b8758f1f-6c66-420c-84c0-480644acaca7)

We could include an e2e test that the name on both sides is not updated at the beginning and then after the sync verify that it is updated successfully. What do you think?

webapp/src/ts/modules/contacts/contacts.component.ts
```diff
@@ -314,6 +314,51 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
...
+    if (!this.usersHomePlace) {
+      return updatedContacts;
+    }
+
+    const homeIndex = this.findHomeIndex(updatedContacts);
+
+    this.checkAdditionalListItem(homeIndex);
+    this.updateUsersHomePlace(updatedContacts, homeIndex);
+
+    return updatedContacts;
```
latin-panda, line 359: Then you can just return here directly

```suggestion
    return this.updateUsersHomePlace(updatedContacts, homeIndex);
```

latin-panda (changes requested):
Nice job extracting that logic into functions. I tested and it's working!
I left some suggestions inline

Benmuiruri:
> It is working perfectly now Ben 🎉
> 
>  itIsWorking.mov 
> We could include an e2e test that the name on both sides is not updated at the beginning and then after the sync verify that it is updated successfully. What do you think?

That sounds alright (but I'm wondering how to write the test). Would the steps for the e2e be 
- login as admin, update user's name, logout 
- login as user, sync, assert that LHS and RHS names are equal ?

tatilepizs:
> That sounds alright. Would the steps for the e2e be
>  - login as admin, update user's name, logout
>  - login as user, sync, assert that LHS and RHS names are equal ?


Hmm no 🤔 because we want to see that the update of the name is actually happening. I will suggest:

- Create an offline user, online user, and the places using [factories](https://github.com/medic/cht-core/tree/master/tests/factories).
- Loggin as the offline user to see the created place.
- Assert that LHS and RHS names are the same (the first time that you log in as an offline user it will fetch all the docs and that is why we need to loogin as the offline user first, so it will do the fetch and the second time we only see sync).
- Logout.
- Login as the online user.
- Update the name of the place.
- Logout.
- Login as the offline user again.
- Assert that LHS and RHS names are the same (the old names, not the ones updated)
- Sync
- Assert that LHS and RHS names are updated with the edited name and are the same.

What do you think?

Benmuiruri:
@latin-panda and @tatilepizs  The refactored and e2e tested PR is ready for your review. 

Here is a video of the 🟢 e2e test. (Thanks Tatiana for the steps, they came in handy 😃)

https://github.com/medic/cht-core/assets/2597305/0cb95b72-73d5-4fb9-983f-a572bbf144b2

tests/e2e/default/contacts/edit-person-home-place.wdio-spec.js
```diff
@@ -0,0 +1,100 @@
...
+    },
+  });
+
+  const onlinePerson = personFactory.build({
+    name: 'Program Officer',
+    parent: {
+      _id: healthCenter._id,
+      parent: healthCenter.parent,
+    },
+  });
```
tatilepizs, line 43: You don't need to create a person and then manually create the user, you can create directly the user using the [userFactory](https://github.com/medic/cht-core/blob/7110-update-user-home-place/tests/factories/cht/users/users.js), like this:

[Offline user example:](https://github.com/medic/cht-core/blob/7110-update-user-home-place/tests/e2e/default/enketo/pregnancy.wdio-spec.js#L20-L29) 

```.js
const offlineUser = userFactory.build({ place: healthCenter._id, roles: ['chw'] });
await loginPage.login(offlineUser);
```

[Online user example:](https://github.com/medic/cht-core/blob/7110-update-user-home-place/tests/e2e/default/contacts/fab-actionbar.wdio-spec.js#L11-L17)

```.js
const onlineUser = userFactory.build({ place: healthCenter._id, roles: [ 'program_officer' ] });
await utils.createUsers([ onlineUser ]);
await loginPage.login(onlineUser);
```

tests/e2e/default/contacts/edit-person-home-place.wdio-spec.js
```diff
@@ -0,0 +1,100 @@
...
+      onlineUsername,
+      'Program Officer',
+      'program_officer',
+      healthCenter.name,
+      onlinePerson.name,
+      password
+    );
+    await usersAdminPage.saveUser();
+
+    await browser.reloadSession();
```
tatilepizs, line 73: You would not need this if you use the userFactory to create the users 🙂 

```suggestion
```

tests/e2e/default/contacts/edit-person-home-place.wdio-spec.js
```diff
@@ -0,0 +1,100 @@
...
+  });
+
+  it('can sync and update offlineUser HomePlace', async () => {
+    await browser.url('/');
```
tatilepizs, line 77: Is there a reason why you need to do this before login? 🤔

tatilepizs (changes requested):
Thank you for creating the e2e test @Benmuiruri ⭐ 
I left some comments 🙂

webapp/src/ts/modules/contacts/contacts.component.ts
```diff
@@ -310,6 +310,52 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
...
+        return [this.usersHomePlace, ...updatedContacts.slice(1)];
+      }
+
+      if (this.additionalListItem) {
+        return [this.usersHomePlace, ...updatedContacts];
+      }
+
+      return updatedContacts;
+    }
+  }
```
latin-panda, line 346: Nice! Whenever you can return early, the code will be flatter. Also, sonar will count the cognitive complexity of the following `if` statements as 1. 
When `if` statements are nested in a block, sonar counts them as 2, therefore increasing the score of cognitive complexity. 

```suggestion
  private moveUserHomePlaceToTop(updatedContacts: any[], homeIndex: number) {
    if (homeIndex < 0) {
      return;
    }
    
    this.usersHomePlace = updatedContacts[homeIndex];

    if (!this.appending) {
      return [this.usersHomePlace, ...updatedContacts.slice(1)];
    }

    if (this.additionalListItem) {
      return [this.usersHomePlace, ...updatedContacts];
    }

    return updatedContacts;
  }
```

latin-panda: This means the function returns `undefined` and an array of records
I wonder if you can do `Record<string, string|number|bool>[]` instead of `any[]`

latin-panda (changes requested):
Nice job! Additionally to Tatiana's feedback I've added 2 more comments below

webapp/src/ts/modules/contacts/contacts.component.ts
latin-panda: Actually, should that first condition return the updated contact list? 
```
if (homeIndex < 0) {
  return updatedContacts;
}
```
I think it should, right?

webapp/src/ts/modules/contacts/contacts.component.ts
```diff
@@ -310,6 +310,52 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
     return this.sortDirection === 'last_visited_date';
   }
 
+  private getContactsDocIds(contactsList, usersHomePlace): string[] {
+    const docIds = contactsList.map((item) => item._id);
```
latin-panda, line 314: This one is missing the `if (options.withIds) {` that it used to have originally. 
Is that intentional?  If not, it would be good to have an unit test catching that

webapp/src/ts/modules/contacts/contacts.component.ts
```diff
@@ -310,6 +310,52 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
...
+  private checkAdditionalListItem(homeIndex: number) {
+    this.additionalListItem =
+      !this.filters.search &&
+      (this.additionalListItem || !this.appending) &&
+      homeIndex === -1;
+  }
+
+  private moveUserHomePlaceToTop(updatedContacts: any[], homeIndex: number) {
+    if (homeIndex < 0) {
+      return updatedContacts;
```
latin-panda, line 334: If homeIndex is -1 (not found) and not appending, should it add it to `updatedContact`?

latin-panda (changes requested):
I have a couple of questions

webapp/src/ts/modules/contacts/contacts.component.ts
```diff
@@ -310,6 +310,52 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
...
+    }
+    this.usersHomePlace = updatedContacts[homeIndex];
```
latin-panda, line 336:
```suggestion
    this.usersHomePlace = updatedContacts[homeIndex]; // Refreshing user home place with latest info
```

webapp/src/ts/modules/contacts/contacts.component.ts
```diff
@@ -310,6 +310,52 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
...
+
+    if (!this.appending) {
+      return [this.usersHomePlace, ...updatedContacts.slice(1)];
```
latin-panda, line 339:
```suggestion
      return [this.usersHomePlace, ...updatedContacts.slice(homeIndex, 1)];
```

webapp/src/ts/modules/contacts/contacts.component.ts
```diff
@@ -310,6 +310,52 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
...
+    }
+    
+    if (this.additionalListItem) {
+      return [this.usersHomePlace, ...updatedContacts];
```
latin-panda, line 343: I think this isnt reachable if homeIndex is -1

Benmuiruri:
Hi @latin-panda thanks for the questions yesterday, that made me rethink the whole solution. 
- The bug was that in the query method, the logic to move the `usersHomePlace` was not using an updated version of `usersHomePlace` when its value is updated.  (From my review, it seems related to the `options.withIds` flag). 
- New solution, with the least changes to existing logic and avoid unintentional sideeffects is, listen to changes to `usersHomePlace` in `ngOnInit`, whenever it changes, update its value. 

Let me know if this works.

webapp/src/ts/modules/contacts/contacts.component.ts
```diff
@@ -94,6 +94,9 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
           this.contactsActions.removeContactFromList({ _id: change.id });
           this.hasContacts = this.contactsList.length;
         }
+        if (this.usersHomePlace && change.id === this.usersHomePlace._id) {
+          this.getUserHomePlaceSummary(change.id);
```
latin-panda, line 98:
```suggestion
         this.usersHomePlace = await this.getUserHomePlaceSummary(change.id);
```
This function is made to return the home place `getHomePlace....` instead of being a set `setHomePlace...`. Also it makes it closer to a pure function

webapp/src/ts/modules/contacts/contacts.component.ts
```diff
@@ -201,18 +204,20 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
...
+        const facilityId = homePlaceId ?? userSettings.facility_id;
+        if (facilityId) {
+          this.globalActions.setUserFacilityId(facilityId);
+          return this.getDataRecordsService.get(facilityId);
         }
       })
       .then((summary) => {
         if (summary) {
           summary.home = true;
+          this.usersHomePlace = summary;
```
latin-panda, line 220: Based on the previous suggestion, then you won't need this change anymore.

latin-panda (approved):
LGTM! 🤩 Just a small comment but approving now. 
See my screening test below:

https://github.com/medic/cht-core/assets/66472237/06e0d5a2-265b-4789-b9a7-b5c94dd1631d

tatilepizs (approved):
Thank you @Benmuiruri for all the updates.
It looks good to me 🌟

PR #8808, opened 2024-01-15
chore: add an explanation message to 2 assertions in  `routing.spec.js`

m5r:
# Description

Tiny change to get a more helpful message than [`expected null to be truthy`](https://github.com/medic/cht-core/actions/runs/7530916999/job/20498760460?pr=8756#step:17:1036) when these assertions fail

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:explanation-message-on-test-failure/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:explanation-message-on-test-failure/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:explanation-message-on-test-failure/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/tests/integration/api/routing.spec.js
+++ b/tests/integration/api/routing.spec.js
@@ -174,8 +174,14 @@ describe('routing', () => {
         utils.request(Object.assign({ path: '/api/deploy-info' }, offlineRequestOptions)),
         utils.requestOnTestDb('/_design/medic-client'),
       ]).then(([ deployInfoOnline, deployInfoOffline, ddoc ]) => {
-        expect(semver.valid(deployInfoOnline.version)).to.be.ok;
-        expect(semver.valid(deployInfoOffline.version)).to.be.ok;
+        expect(
+          semver.valid(deployInfoOnline.version),
+          `"${deployInfoOnline.version}" is not valid semver`,
+        ).to.be.ok;
+        expect(
+          semver.valid(deployInfoOffline.version),
+          `"${deployInfoOffline.version}" is not valid semver`,
+        ).to.be.ok;
 
         const { BRANCH } = process.env;
         const deployInfo = {


tests/integration/api/routing.spec.js
```diff
@@ -174,8 +174,14 @@ describe('routing', () => {
...
-        expect(semver.valid(deployInfoOnline.version)).to.be.ok;
-        expect(semver.valid(deployInfoOffline.version)).to.be.ok;
+        expect(
+          semver.valid(deployInfoOnline.version),
+          `"${deployInfoOnline.version}" is not valid semver`,
+        ).to.be.ok;
+        expect(
+          semver.valid(deployInfoOffline.version),
+          `"${deployInfoOffline.version}" is not valid semver`,
+        ).to.be.ok;
```
dianabarsan, line 184: I think this clears up the error message nicely, thanks for adding this :) 
I don't particularly like the branch as version convention, but it's what we've done historically and I think there are a lot of ramifications to it, which was why I was quite reluctant to change it in any way.

dianabarsan (approved):

tests/integration/api/routing.spec.js
1yuv: How about using `semver.coerce()` which will parse versions like `4.5.0-7167_uplift_cht-conf_scripts.7530916999` as 4.5.0?

dianabarsan: The point is to _not_ coerce :)

m5r: An alternative would be to use the commit hash of the branch's `HEAD` instead of the branch name in the version but then it would have a different meaning - different branches can have the same commit hash as their `HEAD`, for example when cherry-picking.

I think this is fine as is, we can reconsider if somebody feels strongly about being able to have underscores in branch names

PR #8814, opened 2024-01-17
DNM: Pouchdb Session authentication plugin

dianabarsan:
# Description

This is just a sort of proof of concept about how the plugin should work. 
I've set it as a shared library, but in the end this should be it's separate npm module with a battery of tests. 
This PR is just to showcase what it _could_ look like and get early feedback before I kill it with edge case e2e tests - like expiring sessions, password changes and etc. 

# Code review checklist

# Compose URLs

* __CHT_CORE_COMPOSE_URL__

# License


dianabarsan:
Very interested in your thoughts @garethbowen . Thanks a lot!

shared-libs/pouchdb-http-auth-session/src/index.js
```diff
@@ -0,0 +1,134 @@
...
+  if (!url.username) {
+    return;
+  }
+
+  opts.credentials = {
+    username: url.username,
+    password: url.password
+  };
+
+  url.username = url.password = '';
```
garethbowen, line 89: I don't think this is necessary. In my testing, couchdb uses the session token and falls back to basic auth if the session is expired. If you just leave the basic auth there then the session will be automatically established which greatly simplifies the code because you don't need to:

1. Make a _session request - whatever URL the client uses will just work, saving an additional request to the server.
2. Track session expiry - when it expires a new one will be automatically established using basic auth.
3. Invalidate the session - sending basic auth and an expired cookie will just work.
4. Catch the 401 from the server - if it doesn't work with session cookie and basic auth it's not going to work, just return the error to the client.

garethbowen (commented):
Nice one. This will make rolling it out everywhere so much easier!

I think it can be simplified quite a bit - comment inline.

shared-libs/pouchdb-http-auth-session/src/index.js
dianabarsan: I thought about going this route, but I've been reluctant because this is empirical at best, and there's no documentation around this. I'd rather be comprehensive than assumptive.

garethbowen: It's the logical implementation too. I'd rather not mess with the URL if we can avoid it.

garethbowen:
Another thought... A weird bug would be if the client explicitly changes the BASIC credentials because they want to do something as a different user. This code would automatically attach the session cookie of the old user. I wonder if we need to map the user from basic auth to the session cookie so it can be used with multiple users?

dianabarsan:
Great point about switching users!

garethbowen:
@dianabarsan Now that we have this plugin in a separate repo I think we should close this PR, right?

dianabarsan:
Correct. I believe I kept it as a record of changes I did to switch to sessions, but I believe now everything that this PR covers is already merged. We still have to address requests that are made through RPN (now couch-request).

Issue #8819, opened 2024-01-24
The ci-webdriver-default-workflows job is really slow sometimes


garethbowen:
**Describe the issue**
I haven't managed to work out why it is but sometimes when it fails it takes over an hour to complete. The earliest build I can find is this one which ran on the 5th Jan: https://github.com/medic/cht-core/actions/runs/7417779409/job/20184883373

```
Fri, 05 Jan 2024 04:06:23 GMT [0-3] Error getting userCtx Unexpected token o in JSON at position 1
Fri, 05 Jan 2024 04:06:24 GMT [0-3] PASSED in chrome - file:///tests/e2e/default/contacts/contact-details.wdio-spec.js
Fri, 05 Jan 2024 04:06:26 GMT [0-4] RUNNING in chrome - file:///tests/e2e/default/contacts/edit.wdio-spec.js
Fri, 05 Jan 2024 04:06:34 GMT [0-4] Error getting userCtx Unexpected token o in JSON at position 1
Fri, 05 Jan 2024 04:06:36 GMT [0-4] PASSED in chrome - file:///tests/e2e/default/contacts/edit.wdio-spec.js
Fri, 05 Jan 2024 04:06:38 GMT [0-5] RUNNING in chrome - file:///tests/e2e/default/contacts/export.wdio-spec.js
Fri, 05 Jan 2024 04:06:44 GMT [0-5] Error getting userCtx Unexpected token o in JSON at position 1
Fri, 05 Jan 2024 04:06:45 GMT [0-5] PASSED in chrome - file:///tests/e2e/default/contacts/export.wdio-spec.js
Fri, 05 Jan 2024 04:06:47 GMT [0-6] RUNNING in chrome - file:///tests/e2e/default/contacts/fab-actionbar.wdio-spec.js
Fri, 05 Jan 2024 04:06:58 GMT [0-6] Reload modal not showed up
Fri, 05 Jan 2024 04:15:06 GMT [0-6] 2024-01-05T04:15:06.393Z ERROR webdriver: Request failed with status undefined due to TimeoutError: Timeout awaiting 'response' for 120000ms when running "element" with method "POST" and args "{"using":"css selector","value":"#header-dropdown-link"}"
[...]

PR #8820, opened 2024-01-24
fix(#8819): fix cookie decoding in e2e tests

dianabarsan:
# Description

#8819 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8819-fix-cookie-parsing/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8819-fix-cookie-parsing/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8819-fix-cookie-parsing/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/tests/e2e/default/db/ongoing-replication.wdio-spec.js
+++ b/tests/e2e/default/db/ongoing-replication.wdio-spec.js
@@ -190,6 +190,7 @@ describe('ongoing replication', function() {
       nbrPersons: 1,
       nbrClinics: 1,
     });
+
     await chtDbUtils.createDocs([...docs.clinics, ...docs.persons, ...docs.reports]);
 
     await commonPage.sync();
--- a/tests/e2e/default/db/ongoing-replication.wdio-spec.js
+++ b/tests/e2e/default/db/ongoing-replication.wdio-spec.js
@@ -148,7 +148,7 @@ describe('ongoing replication', function() {
 
   it('should handle deletes', async () => {
     await commonPage.sync();
-    await browser.throttle('offline');
+    // await browser.throttle('offline');
     const waitForServiceWorker = await utils.waitForApiLogs(utils.SW_SUCCESSFUL_REGEX);
     const docIdsToDelete = [
       'form:dummy',
@@ -160,7 +160,7 @@ describe('ongoing replication', function() {
     await sentinelUtils.waitForSentinel();
     await waitForServiceWorker.promise;
 
-    await browser.throttle('online');
+    // await browser.throttle('online');
     await commonPage.sync(true);
     const localDocsPostSync = await chtDbUtils.getDocs();
     const localDocIds = dataFactory.ids(localDocsPostSync);
@@ -169,16 +169,17 @@ describe('ongoing replication', function() {
   });
 
   it('should download settings updates', async () => {
-    await browser.throttle('offline');
+    // await browser.throttle('offline');
+    await commonPage.sync();
     await utils.updateSettings({ test: true }, 'api');
-    await browser.throttle('online');
+    // await browser.throttle('online');
     await commonPage.sync(true);
     const [settings] = await chtDbUtils.getDocs(['settings']);
     expect(settings.settings.test).to.equal(true);
 
-    await browser.throttle('offline');
+    // await browser.throttle('offline');
     await utils.revertSettings(true);
-    await browser.throttle('online');
+    // await browser.throttle('online');
     await commonPage.sync(true);
   });
 
--- a/tests/e2e/default/db/ongoing-replication.wdio-spec.js
+++ b/tests/e2e/default/db/ongoing-replication.wdio-spec.js
@@ -148,7 +148,6 @@ describe('ongoing replication', function() {
 
   it('should handle deletes', async () => {
     await commonPage.sync();
-    // await browser.throttle('offline');
     const waitForServiceWorker = await utils.waitForApiLogs(utils.SW_SUCCESSFUL_REGEX);
     const docIdsToDelete = [
       'form:dummy',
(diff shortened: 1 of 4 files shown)


tests/utils/index.js
```diff
@@ -614,7 +614,7 @@ const revertDb = async (except, ignoreRefresh) => { //NOSONAR
     watcher?.cancel();
     await commonElements.closeReloadModal(true);
   } else if (needsRefresh) {
-    await watcher && watcher.promise; // NOSONAR
+    watcher && await watcher.promise;
```
garethbowen, line 617: Did sonar actually catch this bug?

garethbowen (approved):
Thanks for cleaning this up. For what it's worth I don't think this will fix the slowness (I tried it in a branch already) but definitely worth fixing.

I think we should also stop catching the error because it's too easy to ignore a `warn`. Let's just throw the error and fail loudly.

Also, we use `browser.getCookies` in about a dozen places - do you think it's worth putting this code in a util and reusing it everywhere?

tests/utils/index.js
dianabarsan: No, it complained about not using `?` shortcut.

dianabarsan:
I know it's not fixing the hanging unfortunately. 
It looks like the browser somehow becomes unreachable. 

I'm looking at this log where an element request is failing: 
```
[0-6] 2024-01-24T21:02:30.332Z ERROR webdriver: Request failed with status undefined due to 
TimeoutError: Timeout awaiting 'response' for 
120000ms when running "element" with method "POST" and args "{"using":"css selector","value":"#header-dropdown-link"}"
```

https://github.com/medic/cht-core/actions/runs/7642506050/job/20834034415?pr=8820#step:17:554

And all subsequent browser requests also fail, so I'm suspecting some sort of communication issue between wdio and chrome.

garethbowen:
> And all subsequent browser requests also fail, so I'm suspecting some sort of communication issue between wdio and chrome.

That's my assumption too but I can't figure out why. Due to the timing my main suspect is the commit that adds `browser.reloadSession` linked in the issue - whether that can cause a disconnect between browser and browser driver I'm not sure. It does seem to be random and difficult to reproduce so I can't prove anything yet...

dianabarsan:
I think reloading the session happens _after_ deleting users and after the test is done. The excerpt I linked was triggered while the e2e test was still running, before any after test code.

garethbowen:
> I think reloading the session happens after deleting users and after the test is done. The excerpt I linked was triggered while the e2e test was still running, before any after test code.

You may be on to something as it always seems to happen in the `fab-actionbar.wdio-spec.js` test, which wouldn't be the case if it was something in a shared after test code. But I can't see anything obviously wrong in that test either...

garethbowen:
It is a new test though so it fits the timeline.

dianabarsan:
I'll take a closer look at that test.

tests/e2e/default/db/ongoing-replication.wdio-spec.js
```diff
@@ -169,16 +167,13 @@ describe('ongoing replication', function() {
...
+    await commonPage.sync();
     await utils.updateSettings({ test: true }, 'api');
-    await browser.throttle('online');
     await commonPage.sync(true);
     const [settings] = await chtDbUtils.getDocs(['settings']);
     expect(settings.settings.test).to.equal(true);
 
-    await browser.throttle('offline');
     await utils.revertSettings(true);
-    await browser.throttle('online');
```
dianabarsan, line 181: These tests frequently failed because the update modal showed up while we were clicking around in the hamburger menu. I thought I could control it by forcefully preventing the browser from connecting, or by syncing before the test began, but it didn't help. Setting the browser as offline doesn't also set the service worker instance, which is separate, to use the same setting. 
The time when the modal appeared was unpredictable, and it seemed like I would have to check for its existence before every click.

tests/page-objects/default/common/common.wdio.page.js
```diff
@@ -354,6 +355,7 @@ const closeReloadModal = async (shouldUpdate = false, timeout = 5000) => {
   try {
     shouldUpdate ? await modalPage.submit(timeout) : await modalPage.cancel(timeout);
     await modalPage.checkModalHasClosed();
+    shouldUpdate && await waitForAngularLoaded();
```
dianabarsan, line 358: This was what fixed the browser disconnect issue.

tests/page-objects/default/common/modal.wdio.page.js
```diff
@@ -32,10 +32,21 @@ const cancel = async (timeout) => {
   await checkModalHasClosed();
 };
 
+const hideOverlay = () => {
```
dianabarsan, line 35: I know this is cheap and hacky. It feels like I tried everything else except for checking if the modal exists before every action.

tests/page-objects/default/common/common.wdio.page.js
garethbowen: Nice find! How did you work it out?

I think it would make more sense to have the `await waitForAngularLoaded();` link in the `modalPage.submit` function so nobody else is caught out by this. Unless that breaks other tests that are logging out or something...

Also the above line `await modalPage.checkModalHasClosed();` shouldn't be necessary because the `modalPage.submit` awaits it too.

tests/page-objects/default/common/modal.wdio.page.js
garethbowen: If this is what we have to do then so be it...

It would be nice if the refresh was called in the sync method to guarantee this was reverted.

I think it should also be made a private function inside common.wdio because we don't want it to be used widely, IMO.

garethbowen (changes requested):
Thanks for sticking with this! Two small suggestions.

tests/page-objects/default/common/modal.wdio.page.js
dianabarsan: I moved it to a private function. 
Sync does automatically refresh when we expect an update. Otherwise, I wouldn't overcomplicate the code and tests and force refreshes.

garethbowen (approved):
Thank you!

Issue #8821, opened 2024-01-25
Record Apdex score in webapp


latin-panda:
**Is your feature request related to a problem? Please describe.**
We need a tool to help us understand the performance that CHWs are experiencing. We decided to use ApDex 

**Describe the solution you'd like**
Use telemetry to record the page loads that fall into 'Satisfied', 'Tolerating', and 'frustrating' categories. 

[Docs PR](https://github.com/medic/cht-docs/pull/1278)

jkuester:
What is the current status of this issue?  I see we have a PR open, but are there any blockers on that?  Just wondering because this is one of the 4 remaining open issues for the `4.6.0` Milestone.  At this point, I don't know of any pressing need to rush the release out, but I am just trying to gauge where we are with the remaining projects.  Thanks!

latin-panda:
Thanks @jkuester, these are the pending items before we can merge and close this:
- Resolve and merge the task bug that @Benmuiruri is working ([ticket](https://github.com/medic/cht-core/pull/8838))
- Write the SQL that will consume these telemetry entries ([ticket](https://github.com/medic/care-teams/issues/3))
- See it behaving with way more data on bigger test party ([ticket](https://github.com/medic/care-teams/issues/11))

On Monday, we agree to not hold 4.6 release if this is the last open item. cc: @michaelkohn

PR #8822, opened 2024-01-25
feat(#8821): track Apdex performance

latin-panda:
# Description

Adds service to track performance 
Records Apdex in Telemetry (per user):

**Overall boot time:** including loading the code, purging, and accessing the database
  - boot_time
  - boot_time:apdex:[satisfied|tolerable|frustrated]

**Contacts tab:**
  - contact list
    - contact_list:load
    - contact_list:load:apdex:[satisfied|tolerable|frustrated]
  - contact detail - i.e. when selecting a contact.
    - contact_detail:<contact_type>:load
       - contact_detail:<_form>:load:contact_data
       - contact_detail:<_form>:load:load_descendants
       - contact_detail:<_form>:load:load_reports
       - contact_detail:<_form>:load:load_targets
       - contact_detail:<_form>:load:load_tasks
       - contact_detail:<_form>:load:load_contact_summary
    - contact_detail:<contact_type>:load:apdex:[satisfied|tolerable|frustrated]
  - contact form: this are app forms and contact forms loaded in contacts tab
    - enketo:contacts:<_form>:<_action>:render  
    - enketo:contacts:<_form>:<_action>:render:apdex:[satisfied|tolerable|frustrated]  
    - enketo:contacts:<_form>:<_action>:save  
    - enketo:contacts:<_form>:<_action>:save:apdex:[satisfied|tolerable|frustrated]  
  - search and/or sort (internally the sort is like a filter) 
    - contact_list:query
    - contact_list:query:apdex:[satisfied|tolerable|frustrated]

**Reports tab:**
  - report list
    - report_list:load
    - report_list:load:apdex:[satisfied|tolerable|frustrated]
  - select a report  
    - report_detail:<_form>:load
    - report_detail:<_form>:load:apdex:[satisfied|tolerable|frustrated]
  - report form
    - enketo:reports:<_form>:<_action>:render  
    - enketo:reports:<_form>:<_action>:render:apdex:[satisfied|tolerable|frustrated]  
    - enketo:reports:<_form>:<_action>:save  
    - enketo:reports:<_form>:<_action>:save:apdex:[satisfied|tolerable|frustrated]  
  - search and/or filter
    - report_list:query
    - report_list:query:apdex:[satisfied|tolerable|frustrated]

**Tasks tab:**
  - task list (telemetry should be the same as original)
    - tasks:load 
    - tasks:load:apdex:[satisfied|tolerable|frustrated]
  - task form
    - enketo:tasks:<_form>:<_action>:render  
    - enketo:tasks:<_form>:<_action>:render:apdex:[satisfied|tolerable|frustrated]  
    - enketo:tasks:<_form>:<_action>:save  
    - enketo:tasks:<_form>:<_action>:save:apdex:[satisfied|tolerable|frustrated]  

**Targets tab:**
  - target list
    - analytics:targets:load
    - analytics:targets:load:apdex:[satisfied|tolerable|frustrated]
  - target aggregates
    - analytics:target_aggregates:load
    - analytics:target_aggregates:load:apdex:[satisfied|tolerable|frustrated]

**Message tab:**
  - message list
    - message_list:load
    - message_list:load:apdex:[satisfied|tolerable|frustrated]
 - select conversation
   - messages_detail:load
   - messages_detail:load:apdex:[satisfied|tolerable|frustrated]

[CHT Docs PR](https://github.com/medic/cht-docs/pull/1278)

See telemetry mapping [here](https://docs.google.com/document/d/1iXZybHL56l3Z7Y_aEpRqkZkQYggG97EtpnawzA8N_zA/edit#heading=h.azy63xxyzh10)
#8821

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8821-track-apdex-performance/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8821-track-apdex-performance/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8821-track-apdex-performance/docker-compose/cht-couchdb-clustered.yml)

# License


webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -161,9 +161,7 @@ export class TasksComponent implements OnInit, OnDestroy {
         this.tasksActions.setTasksLoaded(true);
       }
 
-      telemetryData.end = Date.now();
-      const telemetryEntryName = !this.tasksLoaded ? `tasks:load` : `tasks:refresh`;
-      this.telemetryService.record(telemetryEntryName, telemetryData.end - telemetryData.start);
+      track?.stop(recordApdex, false);
```
latin-panda, line 164: @garethbowen, quick question: how important is keeping the old telemetry names? 
We intend to prefix performance records with `perf:<component>:<etc>`, but thinking that we need backward compatibility when adding the ApDex, I added a flag to skip the prefix. 
It's a small thing, but having the name convention across would be nicer and avoid this extra check.

webapp/src/ts/services/performance.service.ts
```diff
@@ -0,0 +1,92 @@
...
+        component: component + this.APDEX_TOLERATING,
+        aggregate: aggregate + this.APDEX_TOLERATING,
+      };
+    }
+
+    return {
+      component: component + this.APDEX_FRUSTRATED,
+      aggregate: aggregate + this.APDEX_FRUSTRATED,
+    };
+  }
```
garethbowen, line 91: This is repetitive and creates a new data structure that's unnecessary. Refactor it so it just returns the satisfaction label, eg:

```suggestion
  private getSatisfactionLabel(time: number) {

    if (time <= this.APDEX_T) {
      return this.APDEX_SATISFIED;
    }
    
    if (time <= (this.APDEX_TOLERANCE * this.APDEX_T)) {
      return this.APDEX_TOLERATING;
    }
    
    return APDEX_FRUSTRATED;
  }
```

webapp/src/ts/services/performance.service.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    const telemetryName = prefix ? this.PERFORMANCE_PREFIX + name : name;
+    const time = this.document.defaultView.performance.now() - startTime;
+    await this.telemetryService.record(telemetryName, time);
+
+    if (recordApdex) {
+      const { component, aggregate } = this.getApdexLabels(name, time);
+      await this.telemetryService.record(component, time);
+      await this.telemetryService.record(aggregate, time);
```
garethbowen, line 65: I don't think we should duplicate this because it doubles the number of records that need to be stored and increases the docs to sync. We can easily aggregate them on the server if we record the `component`.

webapp/src/ts/services/performance.service.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  track(name: string) {
+    if (!this.trackPerformance || !name || !this.document?.defaultView) {
+      return;
+    }
+
+    const startTime = this.document.defaultView.performance.now();
+    return {
+      setName: newName => name = newName,
+      stop: (recordApdex = false, prefix = true) => this.recordPerformance(name, startTime, recordApdex, prefix),
```
garethbowen, line 40: Having multiple booleans in an API is annoying because when you're reading the code you have no idea what `stop(true, false)` means and have to dig down deeper.

Consider changing this to something like...

```suggestion
      stop: ({ recordApdex = false, prefix = true }) => this.recordPerformance(name, startTime, recordApdex, prefix),
```

webapp/src/ts/services/performance.service.ts
```diff
@@ -0,0 +1,92 @@
...
+import { AuthService } from '@mm-services/auth.service';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class PerformanceService {
+  private trackPerformance = false;
+  private readonly CAN_TRACK_PERFORMANCE = 'track_performance';
+  private readonly PERFORMANCE_PREFIX = 'perf:';
+  private readonly APDEX_SUBFIX = ':apdex';
```
garethbowen, line 14: I find it complicated when the separator is part of the value here because if you want to add or change the telemetry name it's easy to put the `:` in the wrong place.

My suggestion is to remove the `:` and instead of concatenating the string use join, eg:

```
[this.PERFORMANCE_PREFIX, 'app', this.APDEX_SUBFIX, this.APDEX_AGGREGATE].join(':');
```

garethbowen, line 14: Also, should this be SUBFIX or SUFFIX? A "subfix" is something else...

webapp/src/ts/modules/tasks/tasks.component.ts
garethbowen: I think we have to keep backwards compatibility here - it's part of the API and someone may be using it.

webapp/src/ts/effects/contacts.effects.ts
```diff
@@ -52,6 +56,8 @@ export class ContactsEffects {
         this.store.select(Selectors.getForms),
       ),
       exhaustMap(([{ payload: { id, silent } }, userFacilityId, forms]) => {
+        this.selectContactTrack = 'select_contact:load_everything';
+        const trackPerformance = this.performanceService.track(this.selectContactTrack);
```
garethbowen, line 60: Move this below the `if(!id)` check - there's no point starting the track and then returning before you stop it.

webapp/src/ts/services/performance.service.ts
```diff
@@ -0,0 +1,92 @@
...
+   *                    Telemetry entries' name.
+   * @private
+   */
+  private async recordPerformance(name: string, startTime: number, recordApdex = false, prefix = true) {
+    if (!this.trackPerformance || !this.document?.defaultView) {
+      return;
+    }
+
+    const telemetryName = prefix ? this.PERFORMANCE_PREFIX + name : name;
+    const time = this.document.defaultView.performance.now() - startTime;
```
garethbowen, line 59: It'd be clearer to me if this were called `duration` not `time`

webapp/src/ts/effects/contacts.effects.ts
```diff
@@ -82,6 +88,10 @@ export class ContactsEffects {
             console.error('Error selecting contact', err);
             this.globalActions.unsetSelected();
             return of(this.contactsActions.clearSelection());
+          })
+          .finally(() => {
+            trackPerformance?.setName(this.selectContactTrack);
+            trackPerformance?.stop(true);
```
garethbowen, line 94: It's really weird to set the name on creation and on stop. Let's change the API so we only do it on stop? eg:

```suggestion
            trackPerformance?.stop(this.selectContactTrack, true);
```

Then I think you can clean up the code even more if instead of storing the name in `this.selectContactTrack` you calculate it only when you need it in the `finally` block.

garethbowen (changes requested):
Can't wait to see the apdex data flowing in!

Some suggestions inline.

webapp/src/ts/services/performance.service.ts
latin-panda: > it doubles the number of records that need to be stored and increases the docs to sync

There was a back-and-forth on the aggregate, but it can also be calculated when building the reports

latin-panda:
@ralfudx @Benmuiruri @garethbowen This is ready, please review, thanks!

latin-panda:
Testing evidence

<details> 
<summary>Messages tab</summary>

select-conversation-tolerable
select-conversation-frustrated
select-conversation-satisfied
message-list-satisfied
</details>

<details> 
<summary>Reports tab</summary>

report-list-all-apedex
select-report-frustrated
select-report-tolerable
select-report-satisfied
report-list-satisfied
report-form-satisfied
</details>

<details> 
<summary>Contacts tab</summary>

contact-form-satisfied
contact-report-forms-edit-add
select-contact-satisfied
select-contact-all
contact-list-satisfied
</details>


<details> 
<summary>Targets tab</summary>

analytics-satisfied
</details>

webapp/src/ts/effects/contacts.effects.ts
```diff
@@ -82,6 +88,16 @@ export class ContactsEffects {
             console.error('Error selecting contact', err);
             this.globalActions.unsetSelected();
             return of(this.contactsActions.clearSelection());
+          })
+          .finally(() => {
+            trackPerformance?.stop({
+              recordApdex: true,
+              name: [
+                'select_contact',
+                selectedContact?.doc?.contact_type || 'contact',
```
garethbowen, line 97: Would it be easier to get the selectedContact from the store rather than in a variable?

garethbowen (commented):
I've had a quick look and left a comment - was there something specific you wanted me to look at?

I'll leave it to the other reviewers to do a complete review.

webapp/src/ts/effects/contacts.effects.ts
latin-panda: I'm afraid of a race condition or retriggering this effect and causing more spaghetti in the execution flow. 
This is not the nicest, but it's safe and simple to debug. Ideally, we refactor all this code to simplify, but not in this PR scope.

Benmuiruri (approved):
Impressive work Jenni. It's great to see those logs.

michaelkohn:
This is looking cool! 

I think the initial comment that says `contact form: this are app forms and contact forms loaded in contacts tab` is a bit confusing, but the data samples above look the way I'd expect them. 

A question about `enketo` related telemetry... here's what I believe we have today:

### Contacts Tab
- `enketo:contacts:<form>:<action>:<component>` for a Contact form
- `enketo:reports:<form>:<action>:<component>` for an App form (report)

### Reports Tab
- `enketo:reports:<form>:action>:<component>` for an App form (report)
_It's not possible to access a Contact form from the Reports tab_

### Tasks Tab
- `enketo:tasks:<form>:action>:<component>` for a Task form (report)
_It's not possible to access a Contact form from the Tasks tab_

It seems like we're kind of inconsistent in that we **don't differentiate** Enketo reports accessed from Reports vs Contacts tabs, but we **do** differentiate Enketo reports accessed from the Tasks tab. I know this is not something new (ie it's how the system has always worked), but wondering if I'm misunderstanding this or if there is a specific reason why it's the same for Contacts/Reports but different for Tasks. Is there a safe way to change it so that we can differentiate between reports accessed from Contacts vs. Reports?

latin-panda:
Hi @ralfudx do you have any feedback?

ralfudx:
Yes @latin-panda this looks great
However, I'm curious to know how we intend to capture the report (especially for multiple users) - In the Testing evidence you attached, i can see min, max, and the sum but the `average response time` which ideally should be a factor of the `count` is not represented
Maybe we can have a chat about this during the teams call?

ralfudx (approved):

latin-panda:
> the Testing evidence you attached, i can see min, max, and the sum but the average response time which ideally should be a factor of the count is not represented

Good question. Telemetry doesn't calculate an average, [see docs](https://docs.communityhealthtoolkit.org/apps/guides/performance/telemetry/). But it has a `sumsqr`, if I remember @michaelkohn's explanation correctly, it's the variance in the occurrences.

The telemetry are saved in the user's meta database in CouchDB, you'll find a telemetry document per day that the CHW was using the app.

webapp/src/ts/modules/contacts/contacts.component.ts
```diff
@@ -397,6 +407,9 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
         this.error = true;
         this.loading = false;
         console.error('Error loading contacts', err);
+      })
+      .finally(() => {
+        trackPerformance?.stop({ name: 'contact_list:query', recordApdex: true });
```
latin-panda, line 412: This differs from the telemetry record `search:contacts:<filter[:filter]>`. 
The `contact_list:query` tracks the query in the Contacts Tab only and includes the whole query process to display the result, like formatting contacts, setting up the scroll etc. This one also tracks the Apdex. 
While `search:contacts:<filter[:filter]>` tracks the process from search service only, which can be called by other components (not just Contact list).

latin-panda:
@michaelkohn @Benmuiruri I have added the Apdex to the search in the Contacts and Reports Tab, as agreed yesterday during the team call.

latin-panda:
Mentioning here a bug that @Benmuiruri resolved [here](https://github.com/medic/cht-core/pull/8838) but still needs confirmation from CT before merging

webapp/tests/karma/ts/modules/messages/messages-content.component.spec.ts
```diff
@@ -150,7 +150,7 @@ describe('MessagesContentComponent', () => {
       expect(performanceService.track.calledOnce).to.be.true;
       expect(stopPerformanceTrackStub.calledOnce).to.be.true;
       expect(stopPerformanceTrackStub.args[0][0]).to.deep.equal({
-        name: 'select_conversation:load',
+        name: 'messages_detail:load',
```
Benmuiruri, line 153: Do we keep this as plural or make it singular?  The other entries are singular i.e `report_detail`, `contact_detail`

latin-panda: Plural, see conversation [here](https://medic.slack.com/archives/C02GME3BQE6/p1707191934904099?thread_ts=1707166546.743469&cid=C02GME3BQE6). Because when clicking on one item from the list, it loads a conversation containing many messages. 
Different from reports that load 1 report, for example.

latin-panda:
@Benmuiruri @ralfudx please review one last time before we merge this work

Benmuiruri (approved):
LGTM 🚀

ralfudx (approved):
LGTM as well

Issue #7993, opened 2022-12-20
Upgrade to Node 20


latin-panda:
**Describe the issue**
After the CHT Core's  v4.x.x has been released, we now support the latest LTS NodeJS.
Upgrade and make sure all is working fine.

garethbowen:
While doing this, also update this code: https://github.com/medic/cht-core/blob/e94df0961fe33f53936cee799e7d8ed196428b0a/tests/cht-conf-utils.js#L20-L39

As well as the TODO in the code, this should `fs.rm` not `fs.rmDir` which is deprecated.

garethbowen:
Blocked on https://github.com/medic/cht-conf/issues/538

garethbowen:
Now, we just need to update cht-conf in cht-core... blocked on https://github.com/medic/cht-core/issues/7167 and https://github.com/medic/cht-core/issues/7744

garethbowen:
Merged!

PR #8824, opened 2024-01-26
feat(#7993): use node v20

garethbowen:
# Description

#7993

# Code review checklist

# Compose URLs

* __CHT_CORE_COMPOSE_URL__

# License


admin/package.json
```diff
@@ -3,8 +3,8 @@
   "private": true,
   "version": "0.1.0",
   "engines": {
-    "node": ">=16.12.0",
-    "npm": ">=8.3.1"
+    "node": ">=20.11.0",
+    "npm": ">=10.2.4"
   },
```
jkuester, line 8: Do we get any value from having the `engines` configured in these sub-package.json files?  Kind of annoying to have to make the same change to all of them when a dev will still get warned when running against the top-level....

sentinel/tests/unit/schedule/reminders.spec.js
```diff
@@ -139,6 +140,9 @@ describe('reminders', () => {
       const reminder = { form: 'formA', some: 'config' };
       reminders.__set__('matchReminder', sinon.stub().resolves(date));
       reminders.__set__('sendReminders', sinon.stub().resolves());
+      sinon.stub(performance, 'now')
+        .onFirstCall().returns(0)
+        .onSecondCall().returns(1);
```
jkuester, line 145: I assume this was required due to a change in the behavior of `perf_hooks`?  Pretty wild that this was the only thing that broke!  :sweat_smile:

jkuester (approved):
Sorry for the delay! Couple minor questions inline, but LGTM!

sentinel/tests/unit/schedule/reminders.spec.js
garethbowen: Yes it seems like the `sinon.useFakeTimers()` call now doesn't work for performance.now because node significantly changed their implementation. This only affects performance.now and only node processes which explains why we only hit it in one place.

admin/package.json
garethbowen: You're absolutely right. I removed them from the web applications (admin and webapp). I think there's still value in having them for the node services, because if anyone (not just a dev) tries to run api on Node 16 this will give them a warning. It shouldn't be possible if it's being run in our official docker image but you never know...

Issue #7288, opened 2021-09-01
Searching for a contact name that has a short value will return no results.


newtewt:
**Describe the bug**
Searching for a contact name that has a short value will return no results.

The name could be  `Steve Kshlerin I` or  `Steve I`

**To Reproduce**
1. Login and navigate to contacts page. 
2. Add a new person that is name `Steve Kshlerin I`
3. Navigate to contacts page.
4. Search in the search bar for `Steve`
5.  The contact is returned
4. Search in the search bar for `Steve Kshlerin`
5.  The contact is returned
6.  Search in the search bar for `Steve Kshlerin  I`
7. Nothing is returned

**Expected behavior**
The search should locate the contact and return a row for that contact. 

**Logs**
If applicable, include the server or browser logs.

**Screenshots**

https://user-images.githubusercontent.com/10531066/131734925-28a1c49a-6791-450a-b85e-4910d1ca0c14.mp4

**Environment**
- Instance: gamma.dev
- Browser: chrome
- Client platform: windows
- App: webapp
- Version: 3.12.x

**Additional context**
Add any other context about the problem here. What have you tried? Is there a workaround?

garethbowen:
This came up on the forum: https://forum.communityhealthtoolkit.org/t/unexpected-search-results/3288/3

garethbowen:
The root cause of this is the server ignores short search terms when indexing the doc ("I" in this case) but the client does not when searching. We need to modify the client code to be identical to the server code. This will also help performance because sending a search for a short word that matches lots of docs is a terrible idea.

garethbowen:
Fixed in 4.6.0

PR #8825, opened 2024-01-26
fix(#7288): ignore short search words

garethbowen:
# Description

#7288 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7288-short-keywords-miss-results/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7288-short-keywords-miss-results/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:7288-short-keywords-miss-results/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/shared-libs/search/src/generate-search-requests.js
+++ b/shared-libs/search/src/generate-search-requests.js
@@ -87,14 +87,16 @@ const freetextRequestParams = (word) => {
   if (word.indexOf(':') !== -1) {
     // use exact match
     params.key = [ word ];
-  } else {
-    // use starts with
-    if (word.length < MINIMUM_SEARCH_TERM_LENGTH) {
-      return;
-    }
-    params.startkey = [ word ];
-    params.endkey = [ word + END_OF_ALPHABET ];
+    return params;
   }
+  
+  // use starts with
+  if (word.length < MINIMUM_SEARCH_TERM_LENGTH) {
+    return;
+  }
+  
+  params.startkey = [ word ];
+  params.endkey = [ word + END_OF_ALPHABET ];
   return params;
 };
 
--- a/shared-libs/search/src/generate-search-requests.js
+++ b/shared-libs/search/src/generate-search-requests.js
@@ -4,7 +4,7 @@ const moment = require('moment');
 const END_OF_ALPHABET = '\ufff0';
 const MINIMUM_SEARCH_TERM_LENGTH = 3;
 
-const getKeysArray = (keys) => keys.map(t => [ t ]);
+const getKeysArray = (keys) => keys.map(key => [ key ]);
 
 // filter = { selected: [...], options: [...]}
 const getRequestForMultidropdown = (view, filter, mapKeysFunc) => {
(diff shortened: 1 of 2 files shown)


garethbowen:
@latin-panda Can I get a review for this please? It's split into 3 commits to make the review easier - 1 just for the fix, 1 for a refactor of the code, and 1 for sonar.

It fixes the bug and helps avoid a significant performance issue, but has a potential unintended consequence that before you could actually search for "ga" and get "gareth" - now that short search will be ignored and you will get all contacts/reports. I'm assuming that most users will be ok typing in at least 3 characters, but there's nothing in the UI to explain what's going on.

Anyway, please have a look!

shared-libs/search/src/generate-search-requests.js
```diff
@@ -1,18 +1,14 @@
...
+
 const END_OF_ALPHABET = '\ufff0';
+const MINIMUM_SEARCH_TERM_LENGTH = 3;
 
-const getKeysArray = function(keys) {
-  return keys.map(function(t) {
-    return [ t ];
-  });
-};
+const getKeysArray = (keys) => keys.map(t => [ t ]);
```
latin-panda, line 7:
```suggestion
const getKeysArray = (keys) => keys.map(term => [ term ]);
```
I wonder why Sonar didn't catch single-letters variable names. Maybe the [rule is off](https://rules.sonarsource.com/javascript/RSPEC-117/?search=name)?

shared-libs/search/src/generate-search-requests.js
```diff
@@ -65,63 +61,60 @@ const reportedDateRequest = function(filters) {
...
+  } else {
+    // use starts with
+    if (word.length < MINIMUM_SEARCH_TERM_LENGTH) {
+      return;
+    }
+    params.startkey = [ word ];
+    params.endkey = [ word + END_OF_ALPHABET ];
+  }
+  return params;
+};
```
latin-panda, line 99:
```suggestion
const freetextRequestParams = (word) => {
  const params = {};
  if (word.indexOf(':') !== -1) {
    // use exact match
    params.key = [ word ];
    return params;
  }
  
  // use starts with
  if (word.length < MINIMUM_SEARCH_TERM_LENGTH) {
    return;
  }
  
  params.startkey = [ word ];
  params.endkey = [ word + END_OF_ALPHABET ];
  return params;
};
```
Flatter code.

shared-libs/search/src/generate-search-requests.js
```diff
@@ -65,63 +61,60 @@ const reportedDateRequest = function(filters) {
...
+
+const freetextRequest = (filters, view) => {
   if (filters.search) {
```
latin-panda, line 102: Return early to make code flatter.

shared-libs/search/src/generate-search-requests.js
```diff
@@ -65,63 +61,60 @@ const reportedDateRequest = function(filters) {
...
     const words = filters.search.trim().toLowerCase().split(/\s+/);
```
latin-panda, line 103:
```suggestion
    const words = filters.search
      .trim()
      .toLowerCase()
      .split(/\s+/);
```

latin-panda (changes requested):
Very nice! Just a couple of minor suggestions

shared-libs/search/src/generate-search-requests.js
garethbowen: Done.

shared-libs/search/src/generate-search-requests.js
garethbowen: Replaced with `key` instead of `term`.

latin-panda (approved):
LGTM! Thanks!

latin-panda:
> I'm assuming that most users will be ok typing in at least 3 characters, but there's nothing in the UI to explain what's going on.

Attaching a video about this topic: https://github.com/medic/cht-core/assets/66472237/f70b3531-f3c1-400c-96f5-adda79fdc133



@garethbowen IMO it's not too bad and we shouldn't hold this improvement. 
We can add feedback like we do in the lookup fields, that we show a `searching...` label to `Type more to search..`. Can you please open a ticket.

garethbowen:
Improvement raised here: https://github.com/medic/cht-core/issues/8832

Issue #8591, opened 2023-09-27
Add SNI details for use in e2e HTTPS environment with TLS termination external to server (eg, kubernetes pod)


fardarter:
When proxying to HTTPS from HTTP (for example where an ingress does TLS termination), not including a 'servername' for a request to the HTTPS server (eg, def.org) the request produces the following error (where abc.com is the calling server):

```
'ERR_TLS_CERT_ALTNAME_INVALID'
"RequestError: Error [ERR_TLS_CERT_ALTNAME_INVALID]: Hostname/IP does not match certificate's altnames:
Host: abc.com. is not in the cert's altnames: DNS:def.org"
The addition of servername as an option passed to the HTTP agent resolves this error.
```

See docs for tls.connect(options[, callback]) (https://nodejs.org/api/tls.html): "Server name for the SNI (Server Name Indication) TLS extension. It is the name of the host being connected to, and must be a host name, and not an IP address.".

This pull request represents the current state of the work: https://github.com/medic/cht-core/pull/8579

As of writing, I need to figure out why the e2e tests are failing.

**Note**: Have not changed the use of request-promise-native outside of calls to Couch.

For us this resolves the breakdown between node and haproxy. The Azure equivalent of ELB which medic uses is not an option for termination for a few reasons:
- We're going to be offboarding our data through a Databricks instance so HTTPS to the haproxy instance is critical.
- Deploying into a much larger network with a lot of access via VPN, so HTTPS as much as possible is useful as defense in depth strategy -- I'm more worried about attacks from inside the network than at my ingress.
- Deploying sentinel, the api, an nginx (+modsec) proxy and a config container for cht-conf (using [container jobs](https://learn.microsoft.com/en-us/azure/container-apps/jobs?tabs=azure-cli) for being able to run the conf in network on changes) into an [Azure Container App](https://learn.microsoft.com/en-us/azure/container-apps/overview) (managed kubernetes) environment. It has a lot of really nice benefits, but it does (as is a good idea, tbh) wrap each container in an HTTPS ingress termination. So while the haproxy instance isn't running in this env, eventually if we want to add sidecars this will be an issue.

garethbowen:
The cht-core change is merged. Needs documentation before closing this issue.

garethbowen:
Feature and documentation merged.

garethbowen:
Thanks again @fardarter !

PR #8833, opened 2024-01-30
feat(#8591): adds option for enabling server name indication

fardarter:
# Description

Have designed a unified wrapper for `request-promise-native` to centralise global options and to add a `servername` by default. It's overridable in the case that you ever want to have a different `servername` or an empty one.

When proxying to HTTPS from HTTP (for example where an ingress does TLS termination), not including a 'servername' for a request to the HTTPS server (eg, def.org) the request produces the following error (where `abc.com` is the calling server):

 ```
'ERR_TLS_CERT_ALTNAME_INVALID'
"RequestError: Error [ERR_TLS_CERT_ALTNAME_INVALID]: Hostname/IP does not match certificate's altnames:
 Host: abc.com. is not in the cert's altnames: DNS:def.org"
```

The addition of `servername` as an option passed to the HTTP agent resolves this error. 

See docs for `tls.connect(options[, callback])` (https://nodejs.org/api/tls.html): "Server name for the SNI (Server Name Indication) TLS extension.  It is the  name of the host being connected to, and must be a host name, and not an IP address.".

Have written unit tests and have run other tests with the changes included. Relying on your CI for E2E and other tests.

**Note:** Have not changed the use of `request-promise-native` outside of calls to Couch.

Some prior discussions under this PR (branch name was throwing an error): https://github.com/medic/cht-core/pull/8579

Closes this issue: https://github.com/medic/cht-core/issues/8591

### DIFF
--- a/api/src/environment.js
+++ b/api/src/environment.js
@@ -1,8 +1,17 @@
 const path = require('path');
 const logger = require('./logger');
 
-const { UNIT_TEST_ENV, COUCH_URL, BUILDS_URL } = process.env;
+const { 
+  UNIT_TEST_ENV,
+  COUCH_URL,
+  BUILDS_URL, 
+  PROXY_CHANGE_ORIGIN = false
+} = process.env;
+
 const DEFAULT_BUILDS_URL = 'https://staging.dev.medicmobile.org/_couch/builds_4';
+
+const isString = value => typeof value === 'string' || value instanceof String;
+const isTrue = value => isString(value) ? value.toLowerCase() === 'true' : value === true;
 
 if (UNIT_TEST_ENV) {
   module.exports = {
@@ -15,6 +24,9 @@ if (UNIT_TEST_ENV) {
     host: '',
     protocol: '',
     buildsUrl: '',
+    proxies: {
+      changeOrigin: isTrue(PROXY_CHANGE_ORIGIN)
+    }
   };
 } else if (COUCH_URL) {
   // strip trailing slash from to prevent bugs in path matching
@@ -36,7 +48,14 @@ if (UNIT_TEST_ENV) {
     db: parsedUrl.pathname.replace('/', ''),
     ddoc: 'medic',
     username: parsedUrl.username,
-    password: parsedUrl.password
+    password: parsedUrl.password,
+    proxies: {
+      // See http-proxy (https://www.npmjs.com/package/http-proxy#options) 
+      // "changeOrigin: true/false, Default: false - changes the origin of the host header to the target URL"
+      // This allows proxying from HTTP to HTTPS without encountering certificate issues 
+      // for environments where TLS termination happens elsewhere.
+      changeOrigin: isTrue(PROXY_CHANGE_ORIGIN)
+    }
   };
 } else {
   logger.error(
--- a/api/src/routing.js
+++ b/api/src/routing.js
@@ -12,15 +12,22 @@ const prometheusMiddleware = require('prometheus-api-metrics');
 const rateLimiterMiddleware = require('./middleware/rate-limiter');
 const logger = require('./logger');
 const isClientHuman = require('./is-client-human');
-const target = 'http://' + environment.host + ':' + environment.port;
-const proxy = require('http-proxy').createProxyServer({ target: target });
+
+const port = typeof environment.port !== 'undefined' && environment.port !== null ? `:${environment.port}` : '';
+const target = `${environment.protocol}//${environment.host}${port}`;
+const proxy = require('http-proxy').createProxyServer({
+  target: target, 
+  changeOrigin: environment.proxies.changeOrigin
+});
 const proxyForAuth = require('http-proxy').createProxyServer({
   target: target,
   selfHandleResponse: true,
+  changeOrigin: environment.proxies.changeOrigin
 });
 const proxyForChanges = require('http-proxy').createProxyServer({
   target: target,
   selfHandleResponse: true,
+  changeOrigin: environment.proxies.changeOrigin
 });
 const login = require('./controllers/login');
 const smsGateway = require('./controllers/sms-gateway');
--- a/api/tests/mocha/services/setup/utils.spec.js
+++ b/api/tests/mocha/services/setup/utils.spec.js
@@ -2,7 +2,7 @@ const sinon = require('sinon');
 const { expect } = require('chai');
 const rewire = require('rewire');
 const fs = require('fs');
-const rpn = require('request-promise-native');
+const request = require('@medic/couch-request');
 
 const db = require('../../../../src/db');
 const upgradeLogService = require('../../../../src/services/setup/upgrade-log');
@@ -14,7 +14,7 @@ let utils;
 let clock;
 let originalEnv;
 
-const buildInfo = (version, namespace='medic', application='medic') => ({ version, namespace, application });
+const buildInfo = (version, namespace = 'medic', application = 'medic') => ({ version, namespace, application });
 
 const mockDb = (db) => {
   sinon.stub(db, 'allDocs');
@@ -396,7 +396,7 @@ describe('Setup utils', () => {
       it('should throw an error when staging doc has no attachments', async () => {
         const version = '4.0.0';
         db.builds.get.resolves({
-          build_info: { },
+          build_info: {},
           version: version,
         });
 
@@ -414,7 +414,7 @@ describe('Setup utils', () => {
       it('should throw an error when ddocs are corrupted', async () => {
         const version = 'theversion';
         db.builds.get.resolves({
-          build_info: { },
+          build_info: {},
           version: version,
           _attachments: {
             'ddocs/medic.json': { data: genAttachmentData([{ _id: 'aaa' }]) },
@@ -625,7 +625,7 @@ describe('Setup utils', () => {
     it('should throw error if json has no ddocs', async () => {
       sinon.stub(env, 'ddocsPath').value('ddocsPath');
       sinon.stub(env, 'ddoc').value('medic');
-      sinon.stub(fs.promises, 'readFile').resolves(JSON.stringify({ }));
+      sinon.stub(fs.promises, 'readFile').resolves(JSON.stringify({}));
 
       try {
         await utils.getPackagedBuildInfo();
@@ -636,7 +636,7 @@ describe('Setup utils', () => {
     });
 
     it('should throw error if json is missing medic ddoc', async () => {
-      const jsonDdocs = [ { _id: '_design/one' }, { _id: '_design/two' }, { _id: '_design/three' } ];
+      const jsonDdocs = [{ _id: '_design/one' }, { _id: '_design/two' }, { _id: '_design/three' }];
       sinon.stub(env, 'ddocsPath').value('ddocsPath');
       sinon.stub(env, 'ddoc').value('medic');
       sinon.stub(fs.promises, 'readFile').resolves(JSON.stringify({ docs: jsonDdocs }));
@@ -686,22 +686,28 @@ describe('Setup utils', () => {
       const deployInfoOld = { user: 'old', upgrade_log_id: '11', timestamp: 100 };
       const deployInfoExpected = Object.assign({ timestamp: 1500 }, deployInfoNew);
 
-      sinon.stub(db.medic, 'allDocs').resolves({ rows: [
-        { doc: { _id: '_design/:staged:medic', _rev: '1', new: true, deploy_info: deployInfoNew } },
-        { doc: { _id: '_design/:staged:medic-client', _rev: '1', new: true, deploy_info: deployInfoNew } },
-        { doc: { _id: '_design/medic', _rev: '2', old: true, deploy_info: deployInfoOld } },
-        { doc: { _id: '_design/medic-client', _rev: '3', old: true, deploy_info: deployInfoOld } },
-      ] }); // all ddocs have match
-      sinon.stub(db.sentinel, 'allDocs').resolves({ rows: [
-        { doc: { _id: '_design/:staged:sentinel1', _rev: '1', isnew: true, deploy_info: deployInfoNew } },
-        { doc: { _id: '_design/sentinel1', _rev: '2', isOld: true, deploy_info: deployInfoOld } },
-        { doc: { _id: '_design/extra', _rev: '3', deploy_info: deployInfoOld } },
-      ] }); // one extra existent ddoc
-      sinon.stub(db.medicLogs, 'allDocs').resolves({ rows: [
-        { doc: { _id: '_design/:staged:logs1', _rev: '1', field: 'a', deploy_info: deployInfoNew } },
-        { doc: { _id: '_design/:staged:logs2', _rev: '1', deploy_info: deployInfoNew } },
-        { doc: { _id: '_design/logs1', _rev: '3', field: 'b', deploy_info: deployInfoOld } },
-      ] }); // one extra staged ddoc
+      sinon.stub(db.medic, 'allDocs').resolves({
+        rows: [
+          { doc: { _id: '_design/:staged:medic', _rev: '1', new: true, deploy_info: deployInfoNew } },
+          { doc: { _id: '_design/:staged:medic-client', _rev: '1', new: true, deploy_info: deployInfoNew } },
+          { doc: { _id: '_design/medic', _rev: '2', old: true, deploy_info: deployInfoOld } },
+          { doc: { _id: '_design/medic-client', _rev: '3', old: true, deploy_info: deployInfoOld } },
+        ]
+      }); // all ddocs have match
+      sinon.stub(db.sentinel, 'allDocs').resolves({
+        rows: [
+          { doc: { _id: '_design/:staged:sentinel1', _rev: '1', isnew: true, deploy_info: deployInfoNew } },
+          { doc: { _id: '_design/sentinel1', _rev: '2', isOld: true, deploy_info: deployInfoOld } },
+          { doc: { _id: '_design/extra', _rev: '3', deploy_info: deployInfoOld } },
+        ]
+      }); // one extra existent ddoc
+      sinon.stub(db.medicLogs, 'allDocs').resolves({
+        rows: [
+          { doc: { _id: '_design/:staged:logs1', _rev: '1', field: 'a', deploy_info: deployInfoNew } },
+          { doc: { _id: '_design/:staged:logs2', _rev: '1', deploy_info: deployInfoNew } },
+          { doc: { _id: '_design/logs1', _rev: '3', field: 'b', deploy_info: deployInfoOld } },
+        ]
+      }); // one extra staged ddoc
       sinon.stub(db.medicUsersMeta, 'allDocs').resolves({ rows: [] }); // no ddocs
       sinon.stub(db.users, 'allDocs').resolves({ rows: [] }); // no ddocs
 
@@ -740,10 +746,12 @@ describe('Setup utils', () => {
 
     it('should throw an error when getting ddocs fails', async () => {
       clock.tick(2500);
-      sinon.stub(db.medic, 'allDocs').resolves({ rows: [
-        { doc: { _id: '_design/:staged:medic', _rev: '1', foo: 1, deploy_info: { deploy: 'info' } } },
-        { doc: { _id: '_design/medic', _rev: '2', bar: 2, deploy_info: { deploy: 'old' } } },
-      ] }); // all ddocs have match
+      sinon.stub(db.medic, 'allDocs').resolves({
+        rows: [
+          { doc: { _id: '_design/:staged:medic', _rev: '1', foo: 1, deploy_info: { deploy: 'info' } } },
+          { doc: { _id: '_design/medic', _rev: '2', bar: 2, deploy_info: { deploy: 'old' } } },
+        ]
+      }); // all ddocs have match
       sinon.stub(db.sentinel, 'allDocs').rejects({ the: 'error' });
 
       sinon.stub(db, 'saveDocs').resolves();
@@ -759,20 +767,24 @@ describe('Setup utils', () => {
         expect(db.medic.allDocs.callCount).to.equal(1);
         expect(db.saveDocs.callCount).to.equal(1);
         expect(db.saveDocs.args[0]).to.deep.equal([db.medic, [
-          { _id: '_design/medic', _rev: '2', foo: 1, deploy_info: { deploy: 'info', timestamp: 2500 }},
+          { _id: '_design/medic', _rev: '2', foo: 1, deploy_info: { deploy: 'info', timestamp: 2500 } },
         ]]);
         expect(db.sentinel.allDocs.callCount).to.equal(1);
       }
     });
 
     it('should throw an error when saving ddocs fails', async () => {
-      sinon.stub(db.medic, 'allDocs').resolves({ rows: [
-        { doc: { _id: '_design/:staged:medic', _rev: '1', foo: 1, deploy_info: { deploy: 'info' } } },
-        { doc: { _id: '_design/medic', _rev: '2', bar: 2, deploy_info: 'omg' } },
-      ] }); // all ddocs have match
-      sinon.stub(db.sentinel, 'allDocs').resolves({ rows: [
-        { doc: { _id: '_design/:staged:sentinel', _rev: '1', foo: 1, deploy_info: { deploy: 'info' } } },
-      ]});
+      sinon.stub(db.medic, 'allDocs').resolves({
+        rows: [
+          { doc: { _id: '_design/:staged:medic', _rev: '1', foo: 1, deploy_info: { deploy: 'info' } } },
+          { doc: { _id: '_design/medic', _rev: '2', bar: 2, deploy_info: 'omg' } },
+        ]
+      }); // all ddocs have match
+      sinon.stub(db.sentinel, 'allDocs').resolves({
+        rows: [
+          { doc: { _id: '_design/:staged:sentinel', _rev: '1', foo: 1, deploy_info: { deploy: 'info' } } },
+        ]
+      });
 
       sinon.stub(db, 'saveDocs').resolves();
       db.saveDocs.onCall(1).rejects({ error: 'an error' });
@@ -928,11 +940,11 @@ describe('Setup utils', () => {
     it('should call default upgrade service url', async () => {
       const payload = { docker_compose: { 'doc.yml': 'payload' }, containers: [] };
       const response = { 'doc.yml': { ok: true } };
-      sinon.stub(rpn, 'post').resolves({ 'doc.yml': { ok: true } });
+      sinon.stub(request, 'post').resolves({ 'doc.yml': { ok: true } });
 
       expect(await utils.makeUpgradeRequest(payload)).to.deep.equal(response);
-      expect(rpn.post.callCount).to.equal(1);
-      expect(rpn.post.args[0]).to.deep.equal([{
+      expect(request.post.callCount).to.equal(1);
+      expect(request.post.args[0]).to.deep.equal([{
         url: 'http://localhost:5008/upgrade',
         json: true,
         body: payload,
@@ -942,13 +954,13 @@ describe('Setup utils', () => {
     it('should call env upgrade service url', async () => {
       const payload = { containers: [], docker_compose: { 'doc.yml': 'payload' } };
       const response = { 'doc.yml': { ok: true } };
-      sinon.stub(rpn, 'post').resolves(response);
+      sinon.stub(request, 'post').resolves(response);
       process.env.UPGRADE_SERVICE_URL = 'http://someurl';
       utils = rewire('../../../../src/services/setup/utils');
 
       expect(await utils.makeUpgradeRequest(payload)).to.deep.equal(response);
-      expect(rpn.post.callCount).to.equal(1);
-      expect(rpn.post.args[0]).to.deep.equal([{
+      expect(request.post.callCount).to.equal(1);
+      expect(request.post.args[0]).to.deep.equal([{
         url: 'http://someurl/upgrade',
         json: true,
         body: payload,
@@ -957,30 +969,31 @@ describe('Setup utils', () => {
 
     it('should throw invalid url error', async () => {
       const payload = { containers: [], docker_compose: {} };
-      sinon.stub(rpn, 'post').resolves({});
+      sinon.stub(request, 'post').resolves({});
       process.env.UPGRADE_SERVICE_URL = 'whatever';
       utils = rewire('../../../../src/services/setup/utils');
 
-      await expect(utils.makeUpgradeRequest(payload)).to.be.rejectedWith('Invalid UPGRADE_SERVICE_URL: whatever');
+      await expect(utils.makeUpgradeRequest(payload))
+        .to.be.rejectedWith('Invalid UPGRADE_SERVICE_URL: whatever');
     });
 
     it('should throw request errors', async () => {
       const payload = { containers: [], 'docker-compose': {} };
-      sinon.stub(rpn, 'post').rejects(new Error('boom'));
+      sinon.stub(request, 'post').rejects(new Error('boom'));
 
       await expect(utils.makeUpgradeRequest(payload)).to.be.rejectedWith('boom');
     });
 
     it('should throw error when response is invalid', async () => {
       const payload = { docker_compose: { 'doc.yml': 'payload' }, containers: {} };
-      sinon.stub(rpn, 'post').resolves();
+      sinon.stub(request, 'post').resolves();
 
       await expect(utils.makeUpgradeRequest(payload)).to.be.rejectedWith('No containers were updated');
     });
 
     it('should throw error when no docker-compose files were updated', async () => {
       const payload = { docker_compose: { 'doc1.yml': 'payload', 'doc2.yml': 'payload' }, containers: [] };
-      sinon.stub(rpn, 'post').resolves({
+      sinon.stub(request, 'post').resolves({
         'doc1.yml': { ok: false },
         'doc2.yml': { ok: false },
       });
@@ -993,7 +1006,7 @@ describe('Setup utils', () => {
         docker_compose: { 'd1.yml': 'p', 'd2.yml': 'p' },
         containers: [{ container_name: 'a' }, { container_name: 'b' }]
       };
-      sinon.stub(rpn, 'post').resolves({
+      sinon.stub(request, 'post').resolves({
         'a': { ok: false },
         'b': { ok: false },
       });
--- a/sentinel/.eslintrc
+++ b/sentinel/.eslintrc
@@ -16,6 +16,7 @@
           "@medic/cht-script-api",
           "@medic/contact-types-utils",
           "@medic/contacts",
+          "@medic/couch-request",
           "@medic/infodoc",
           "@medic/lineage",
           "@medic/message-utils",
--- a/sentinel/src/schedule/reminders.js
+++ b/sentinel/src/schedule/reminders.js
@@ -1,7 +1,7 @@
 const { performance } = require('perf_hooks');
 const _ = require('lodash');
 const moment = require('moment');
-const request = require('request-promise-native');
+const request = require('@medic/couch-request');
 
 const config = require('../config');
 const db = require('../db');
@@ -108,7 +108,7 @@ const getPlaceIds = (keys, startDocId) => {
   // using `request` library because PouchDB doesn't support `start_key_doc_id` in view queries
   // using `start_key_doc_id` because using `skip` is *very* slow
   return request
-    .get(`${db.couchUrl}/_design/medic-client/_view/contacts_by_type`, { qs: query, json: true })
+    .get({ url: `${db.couchUrl}/_design/medic-client/_view/contacts_by_type`, qs: query, json: true })
     .then(result => result.rows.map(row => row.id));
 };
 
--- a/shared-libs/couch-request/package.json
+++ b/shared-libs/couch-request/package.json
@@ -1,0 +1,15 @@
+{
+  "name": "@medic/couch-request",
+  "version": "1.0.0",
+  "description": "Provides a declarative global options interface for http requests",
+  "main": "src/couch-request.js",
+  "scripts": {
+    "test": "mocha ./test"
+  },
+  "author": "",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "lodash": "^4.17.21",
+    "request-promise-native": "^1.0.9"
+  }
+}
--- a/shared-libs/couch-request/src/couch-request.js
+++ b/shared-libs/couch-request/src/couch-request.js
@@ -1,0 +1,111 @@
+const request = require('request-promise-native');
+const isPlainObject = require('lodash/isPlainObject');
+const servername = new URL(process.env.COUCH_URL).hostname;
+
+const isString = value => typeof value === 'string' || value instanceof String;
+const isTrue = value => isString(value) ? value.toLowerCase() === 'true' : value === true;
+
+const addServername = isTrue(process.env.ADD_SERVERNAME_TO_HTTP_AGENT);
+const methods = {
+  GET: 'GET',
+  POST: 'POST',
+  DELETE: 'DELETE',
+  PUT: 'PUT',
+  HEAD: 'HEAD'
+};
+
+
+const mergeOptions = (target, source, exclusions = []) => {
+  for (const [key, value] of Object.entries(source)) {
+    if (Array.isArray(exclusions) && exclusions.includes(key)) {
+      return target;
+    }
+    target[key] = value; // locally, mutation is preferable to spreading as it doesn't
+    // make new objects in memory. Assuming this is a hot path.
+  }
+  return target;
+};
+
+// When proxying to HTTPS from HTTP (for example where an ingress does TLS termination in an SNI environment),
+// not including a 'servername' for a request to the HTTPS server (eg, def.org) produces the 
+// following error:
+// 
+// 'ERR_TLS_CERT_ALTNAME_INVALID'
+// "RequestError: Error [ERR_TLS_CERT_ALTNAME_INVALID]: Hostname/IP does not match certificate's altnames:
+//  Host: abc.com. is not in the cert's altnames: DNS:def.org"
+// 
+// The addition of 'servername' resolves this error. See docs for 'tls.connect(options[, callback])'
+//  (https://nodejs.org/api/tls.html): "Server name for the SNI (Server Name Indication) TLS extension  It is the
+//  name of the host being connected to, and must be a host name, and not an IP address.".
+// 
+
+const validate = (firstIsString, method, first, second = {}) => {
+
+  if (Object.hasOwn(methods, method) === false) {
+    throw new Error(`Unsupported method (${method}) passed to call.`);
+  }
+
+  if (isPlainObject(second) === false) {
+    throw new Error(`"options" must be a plain object'`);
+  }
+
+  if (firstIsString === false && isPlainObject(first) === false) {
+    throw new Error(`"options" must be a plain object'`);
+  }
+};
+
+
+const req = (method, first, second = {}) => {
+
+  const firstIsString = isString(first);
+
+  try {
+    validate(firstIsString, method, first, second);
+  } catch (e) {
+    return Promise.reject(e);
+  }
+
+  const chosenOptions = firstIsString ? second : first;
+
+  const exclusions = firstIsString ? ['url', 'uri', 'method'] : ['method'];
+  const target = addServername ? { servername } : { };
+  
+  const mergedOptions = mergeOptions(target, chosenOptions, exclusions);
+
+  return firstIsString ? getRequestType(method)(first, mergedOptions) : getRequestType(method)(mergedOptions);
+};
+
+const getRequestType = (method) => {
+  // This is intended to simplify testing as sinon does not stub an
+  // exported function (eg, 'export = requestPromise') but only methods.
+  // From reading, proxyquire would need to be used to solve: https://www.npmjs.com/package/proxyquire
+  // See: https://github.com/sinonjs/sinon/issues/562#issuecomment-79227487
+  switch (method) {
+  case methods.GET: {
+    return request.get;
+  }
+  case methods.POST: {
+    return request.post;
+  }
+  case methods.PUT: {
+    return request.put;
+  }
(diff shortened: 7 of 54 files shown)

### CODE REVIEW

**Comment by garethbowen** [comment]
@fardarter Following up here, the build issue you hit previous has been fixed in `master` so if you rebase you should be able to continue. There's now a conflict which will need to be fixed to get back up to date.

After that I'll see if I can figure out why the test is failing.

**Comment by fardarter** [comment]
> @fardarter Following up here, the build issue you hit previous has been fixed in `master` so if you rebase you should be able to continue. There's now a conflict which will need to be fixed to get back up to date.
> 
> After that I'll see if I can figure out why the test is failing.

Run from the rebased version: https://github.com/medic/cht-core/actions/runs/8017517012

Not many tests still failing. 

Branch has a bunch of extra commits for debugging but that stuff would come out for a non-draft PR.

Help is very much appreciated.

**Comment by fardarter** [comment]
> @fardarter Following up here, the build issue you hit previous has been fixed in `master` so if you rebase you should be able to continue. There's now a conflict which will need to be fixed to get back up to date.
> 
> After that I'll see if I can figure out why the test is failing.

@garethbowen Happy to resync this if you will take a look?

**Comment by garethbowen** [comment]
@fardarter Yes let's get this going.

Steps I think are needed before review...

1. Resync and fix the conflicts.
2. Raise or link this PR to an issue to clearly outline the problem you're fixing.
3. Reduce the number of lines changed. I can see why the import change is necessary but there are also some unrelated changes in there. For example, let's not change the eslint config - that warrants some discussion on its own. Also it looks like you've added some debug logging that should be removed.

**Comment by fardarter** [comment]
> @fardarter Yes let's get this going.
> 
> Steps I think are needed before review...
> 
> 1. Resync and fix the conflicts.
> 2. Raise or link this PR to an issue to clearly outline the problem you're fixing.
> 3. Reduce the number of lines changed. I can see why the import change is necessary but there are also some unrelated changes in there. For example, let's not change the eslint config - that warrants some discussion on its own. Also it looks like you've added some debug logging that should be removed.

Hi @garethbowen. This PR is in a working state right now -- ie, I know there things in there that are not intended for merge. The eslint change is to support some of the code written, but I can split out some of the other things into other PRs. The tooling changes shouldn't be affecting the tests though.

Here is the linked issue: https://github.com/medic/cht-core/issues/8591

I have another PR on a branch name that's never going to work, so swapped to this just as a working branch. Happy to clean it all up once I have something that passes tests. 

I'll sync and tidy this up this week.

**Comment by garethbowen** [comment]
> have a positive value but no rows actually returned in the rows array.

That's a valid response... the total_rows is the count of rows in the view before filtering and skipping is done: https://stackoverflow.com/questions/34805659/couchdb-views-total-rows-vs-offset-vs-rows

Does the test fail locally? If so I suggest debugging in there so you can chop the code and work out exactly which line is timing out.

**Comment by fardarter** [comment]
> > have a positive value but no rows actually returned in the rows array.
> 
> That's a valid response... the total_rows is the count of rows in the view before filtering and skipping is done: https://stackoverflow.com/questions/34805659/couchdb-views-total-rows-vs-offset-vs-rows
>
> Does the test fail locally? If so I suggest debugging in there so you can chop the code and work out exactly which line is timing out.

> That's a valid response

Sure, but the test is expecting values, and we're not sure why there are no values. 

>  Does the test fail locally?

It does fail locally (on a VM), yes, but we're lacking context and experience on what it's testing and how it ought to be behaving, so we're finding it difficult to isolate what input might be responsible. 

Locally, what's timing out is this (in `tests/integration/sentinel/schedules/reminders.spec.js`):

```js
if (result.rows.length >= expectedLogs) {
```

Because it is polling for rows to have some length.

garethbowen:
The test that's timing out is "should create reminders". It calls the `getReminderLogs` function multiple times. The first step is to figure out which call is the one that's timing out because that'll indicate which block of code is no longer working.

fardarter:
> The test that's timing out is "should create reminders". It calls the `getReminderLogs` function multiple times. The first step is to figure out which call is the one that's timing out because that'll indicate which block of code is no longer working.

It fails in the first call. 

See the raw logs here: https://[pipelinesghubeus2.actions.githubusercontent.com/Z1TT1tHFtNSb9nWma0rEOKax7paM32FuO78wdpig7IuFd2ZpHF/_apis/pipelines/1/runs/31561/signedlogcontent/24?urlExpires=2024-04-02T07%3A16%3A38.7727271Z&urlSigningMethod=HMACV1&urlSignature=psqzhl1bZ5wslpv9l52zvdYrzxLWdqGDE%2F%2BRaijSTs4%3D](https://pipelinesghubeus2.actions.githubusercontent.com/Z1TT1tHFtNSb9nWma0rEOKax7paM32FuO78wdpig7IuFd2ZpHF/_apis/pipelines/1/runs/31561/signedlogcontent/24?urlExpires=2024-04-02T07%3A16%3A38.7727271Z&urlSigningMethod=HMACV1&urlSignature=psqzhl1bZ5wslpv9l52zvdYrzxLWdqGDE%2F%2BRaijSTs4%3D)

Ctrf-f `----------------------------1` to find. 

I have a number of other log points not being hit. The length of `----------------------------` is always the same in the code. We never get past `1`.


image

garethbowen:
@fardarter I think I've found it... If you go to the most recent build in github you can download an artifact for the build which contains the logs of the run. In the sentinel log there's this exception:

```
2024-02-23T10:14:44.506 DEBUG: Running reminder {
  form: 'FORM1',
  text_expression: 'at 09:11 on Fri',
  message: '{{name}} should do something'
} 
2024-02-23T10:14:44.509 ERROR: Task reminders completed with error: Error: "options" must be a plain object
    at Object.get (/service/shared-libs/couch-request/src/couch-request.js:54:29)
    at getPlaceIds (/service/sentinel/src/schedule/reminders.js:111:6)
    at getValidPlacesBatch (/service/sentinel/src/schedule/reminders.js:197:10)
    at sendReminders (/service/sentinel/src/schedule/reminders.js:249:10)
    at /service/sentinel/src/schedule/reminders.js:293:12
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  [stack]: 'Error: "options" must be a plain object\n' +
    '    at Object.get (/service/shared-libs/couch-request/src/couch-request.js:54:29)\n' +
    '    at getPlaceIds (/service/sentinel/src/schedule/reminders.js:111:6)\n' +
    '    at getValidPlacesBatch (/service/sentinel/src/schedule/reminders.js:197:10)\n' +
    '    at sendReminders (/service/sentinel/src/schedule/reminders.js:249:10)\n' +
    '    at /service/sentinel/src/schedule/reminders.js:293:12\n' +
    '    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)',
  [message]: '"options" must be a plain object'
} 
```

The configuration of this reminder matches the configuration in that test. It looks like the new couch-request lib is not a drop in replacement for `request`, in this case, the reminders function calls `.get` with a URL as the first param, which is valid in the `request` library but not in the new `get` function.

I think you need to update the couch-request library to be api compatible with the request library.

fardarter:
> @fardarter I think I've found it... If you go to the most recent build in github you can download an artifact for the build which contains the logs of the run. In the sentinel log there's this exception:
> 
> ```
> 2024-02-23T10:14:44.506 DEBUG: Running reminder {
>   form: 'FORM1',
>   text_expression: 'at 09:11 on Fri',
>   message: '{{name}} should do something'
> } 
> 2024-02-23T10:14:44.509 ERROR: Task reminders completed with error: Error: "options" must be a plain object
>     at Object.get (/service/shared-libs/couch-request/src/couch-request.js:54:29)
>     at getPlaceIds (/service/sentinel/src/schedule/reminders.js:111:6)
>     at getValidPlacesBatch (/service/sentinel/src/schedule/reminders.js:197:10)
>     at sendReminders (/service/sentinel/src/schedule/reminders.js:249:10)
>     at /service/sentinel/src/schedule/reminders.js:293:12
>     at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
>   [stack]: 'Error: "options" must be a plain object\n' +
>     '    at Object.get (/service/shared-libs/couch-request/src/couch-request.js:54:29)\n' +
>     '    at getPlaceIds (/service/sentinel/src/schedule/reminders.js:111:6)\n' +
>     '    at getValidPlacesBatch (/service/sentinel/src/schedule/reminders.js:197:10)\n' +
>     '    at sendReminders (/service/sentinel/src/schedule/reminders.js:249:10)\n' +
>     '    at /service/sentinel/src/schedule/reminders.js:293:12\n' +
>     '    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)',
>   [message]: '"options" must be a plain object'
> } 
> ```
> 
> The configuration of this reminder matches the configuration in that test. It looks like the new couch-request lib is not a drop in replacement for `request`, in this case, the reminders function calls `.get` with a URL as the first param, which is valid in the `request` library but not in the new `get` function.
> 
> I think you need to update the couch-request library to be api compatible with the request library.

Brilliant, thank you. I'll have a look a bit later.

fardarter:
@garethbowen I have all the tests passing. Would be grateful for some attention when you can spare it.

api/src/routing.js
```diff
@@ -12,15 +12,28 @@ const prometheusMiddleware = require('prometheus-api-metrics');
...
-const proxy = require('http-proxy').createProxyServer({ target: target });
+
+const isString = value => typeof value === 'string' || value instanceof String;
+const isTrue = value => isString(value) ? value.toLowerCase() === 'true' : value === true;
+
+const { 
+  PROXY_CHANGE_ORIGIN = false,
+  PROXY_CHANGE_ORIGIN_AUTH = false,
+  PROXY_CHANGE_ORIGIN_CHANGES = false
+} = process.env;
```
garethbowen, line 23: To make it easier to reuse these, or at least see what options are available, I think these should go in the environment.js

api/src/routing.js
```diff
@@ -12,15 +12,28 @@ const prometheusMiddleware = require('prometheus-api-metrics');
 const rateLimiterMiddleware = require('./middleware/rate-limiter');
 const logger = require('./logger');
 const isClientHuman = require('./is-client-human');
-const target = 'http://' + environment.host + ':' + environment.port;
-const proxy = require('http-proxy').createProxyServer({ target: target });
+
+const isString = value => typeof value === 'string' || value instanceof String;
+const isTrue = value => isString(value) ? value.toLowerCase() === 'true' : value === true;
```
garethbowen, line 17: Move these to environment.js and then call them when reading the process.env so we just store the boolean in memory rather than executing this function every time we want to check.

api/tests/mocha/services/setup/utils.spec.js
```diff
@@ -928,48 +940,49 @@ describe('Setup utils', () => {
...
       await expect(utils.makeUpgradeRequest(payload)).to.be.rejectedWith('boom');
     });
 
     it('should throw error when response is invalid', async () => {
       const payload = { docker_compose: { 'doc.yml': 'payload' }, containers: {} };
-      sinon.stub(rpn, 'post').resolves();
+      sinon.stub(request, 'post').resolves();
 
-      await expect(utils.makeUpgradeRequest(payload)).to.be.rejectedWith('No containers were updated');
+      await expect(utils.makeUpgradeRequest(payload)).to.eventually.be.rejectedWith('No containers were updated');
```
garethbowen, line 962: Why is this change necessary? I expect rpn and @medic/request to behave the same here.

package.json
```diff
@@ -34,6 +34,8 @@
     "integration-sentinel-local": "export VERSION=$(node ./scripts/build/get-version.js) && ./scripts/build/build-service-images.sh && npm run ci-integration-sentinel",
     "integration-api": "cd api && npm ci && cd .. && mocha 'api/tests/integration/**/*.js' -t 10000",
     "lint": "eslint --color --cache . && ./scripts/build/blank-link-check.sh",
+    "lint-fix": "eslint --color --cache . --fix",
+    "standard-wdio-local": "export VERSION=$(node ./scripts/build/get-version.js) && ./scripts/build/build-service-images.sh && wdio run ./tests/e2e/standard/wdio.conf.js --suite=all",
```
garethbowen, line 38: Can we make this change in a separate PR?

scripts/build/index.js
```diff
@@ -160,7 +160,7 @@ const saveServiceTags = () => {
 const updateServiceWorker = () => {
   const updateSWUrl = getApiUrl('/api/v2/upgrade/service-worker');
 
-  return rpn.get(updateSWUrl).catch(err => {
+  return request.get(updateSWUrl).catch(err => {
```
garethbowen, line 163: This is just a name change with no actual change so it complicates the PR and git log. Please revert this and all the other changes in the scripts dir.

sentinel/.eslintrc
```diff
@@ -42,7 +43,8 @@
         "tests/**/*.js"
       ],
       "rules": {
-        "node/no-extraneous-require": "off"
+        "node/no-extraneous-require": "off",
+        "eqeqeq": ["error", "always", { "null": "ignore" }] // foo != null is a convenient way to check for null or undefined.
```
garethbowen, line 47: As above, if you want to lobby for this change it should be done as a separate pr.

sentinel/src/schedule/reminders.js
```diff
@@ -108,7 +108,7 @@ const getPlaceIds = (keys, startDocId) => {
   // using `request` library because PouchDB doesn't support `start_key_doc_id` in view queries
   // using `start_key_doc_id` because using `skip` is *very* slow
   return request
-    .get(`${db.couchUrl}/_design/medic-client/_view/contacts_by_type`, { qs: query, json: true })
+    .get({ url: `${db.couchUrl}/_design/medic-client/_view/contacts_by_type`, qs: query, json: true })
```
garethbowen, line 111: How confident are you that there's nowhere else in the repo that calls rpn with the URL string as the first param? I'm worried there's somewhere else that maybe doesn't have test coverage which will now fail...

shared-libs/couch-request/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@medic/couch-request",
+  "version": "1.0.0",
+  "description": "Provides a declarative global options interface for request-promise-native",
```
garethbowen, line 4:
```suggestion
  "description": "Provides a declarative global options interface for http requests",
```

Let's not tie ourselves to rpn!

shared-libs/couch-request/src/couch-request.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+  const prototype = Object.getPrototypeOf(value);
+  return (prototype === null || prototype === Object.prototype || Object.getPrototypeOf(prototype) === null)
+    && !(Symbol.toStringTag in value) && !(Symbol.iterator in value);
+};
+
+const memoGlobalCouchRequestOptions = () => {
+  const cache = new Map(); // the assumption here is that this will be a fairly hot path, so caching would be valuable
+  return (COUCH_URL) => {
+    if (cache.has(COUCH_URL)) {
```
garethbowen, line 15: The first time through I thought this was checking the wrong key name but it works because you set both. I think this would be more readable.

```suggestion
    if (cache.has('hostname')) {
```

api/src/routing.js
garethbowen: Also can you add more info about what these three are for? Surely if you want to `changeOrigin` for one, you want to do it for all 3?

shared-libs/couch-request/src/couch-request.js
```diff
@@ -0,0 +1,84 @@
...
+  if (typeof value !== 'object' || value == null) {
+    return false;
+  }
+  const prototype = Object.getPrototypeOf(value);
+  return (prototype === null || prototype === Object.prototype || Object.getPrototypeOf(prototype) === null)
+    && !(Symbol.toStringTag in value) && !(Symbol.iterator in value);
+};
+
+const memoGlobalCouchRequestOptions = () => {
+  const cache = new Map(); // the assumption here is that this will be a fairly hot path, so caching would be valuable
```
garethbowen, line 13: Because all we need is the COUCH_URL environment variable and that's immutable at runtime I think we can do even better. Let's calculate this at initialization rather than on call, and then store the `{ servername }` object in a const at the top level?

Also, let's only set `servername` if `env.PROXY_CHANGE_ORIGIN` is set.

shared-libs/couch-request/src/couch-request.js
```diff
@@ -0,0 +1,84 @@
+const request = require('request-promise-native');
+
+const isPlainObject = (value) => {
+  if (typeof value !== 'object' || value == null) {
+    return false;
+  }
+  const prototype = Object.getPrototypeOf(value);
+  return (prototype === null || prototype === Object.prototype || Object.getPrototypeOf(prototype) === null)
+    && !(Symbol.toStringTag in value) && !(Symbol.iterator in value);
+};
```
garethbowen, line 10: Surely there's a lodash function for this?

shared-libs/couch-request/src/couch-request.js
```diff
@@ -0,0 +1,84 @@
...
+const mergeOptions = (target, source) => {
+  for (const [key, value] of Object.entries(source)) {
+    target[key] = value; // locally, mutation is preferable to spreading as it doesn't
+    // make new objects in memory. Assuming this is a hot path.
+  }
+  return target;
+};
+
+module.exports = {
+  get: (options = {}) => {
```
garethbowen, line 52: I don't think you want to default this to an empty object - surely at least the url is required?

```suggestion
  get: (options) => {
```

shared-libs/couch-request/src/couch-request.js
```diff
@@ -0,0 +1,84 @@
...
+      return Promise.reject(Error('"options" must be a plain object'));
+    }
+    return request.delete(mergeOptions(globalCouchRequestOptions(process.env.COUCH_URL), options));
+  },
+  head: (options = {}) => {
+    if (isPlainObject(options) === false) {
+      return Promise.reject(Error('"options" must be a plain object'));
+    }
+    return request.head(mergeOptions(globalCouchRequestOptions(process.env.COUCH_URL), options));
+  },
```
garethbowen, line 83: All these functions are really repetitive... Let's do something like...

```suggestion
  head: (options) => request('HEAD', options);
  request: (method, options) => {
    if (isPlainObject(options) === false) {
      return Promise.reject(Error('"options" must be a plain object'));
    }
    options.method = method;
    options = mergeOptions(globalCouchRequestOptions(process.env.COUCH_URL), options);
    return request(options);
  },
```

shared-libs/couch-request/test/couch-request.js
```diff
@@ -0,0 +1,186 @@
+const couch_request = require('../src/couch-request');
+const chai = require('chai').use(require('chai-as-promised'));
+const request = require('request-promise-native');
+const sinon = require('sinon');
+
+class Unicorn { }
+const notPlainObjects = [[1, 2, 3], new Unicorn(), new Map([[1, 1], [2, 2], [3, 3]]), new Set([1, 2, 3]),
+  () => { }, true, 'foo', null, 1, NaN, Infinity, /foo/, new Date(), new Error(), new Int8Array(), new Float32Array(),
+  new Float64Array(), new Uint8Array(), new Uint8ClampedArray(), new Uint16Array(), new Uint32Array(),
+  new ArrayBuffer(), new WeakMap(), new WeakSet()];
```
garethbowen, line 10: If you use a library like lodash you won't have to test everything that's not an object because you can assume it'll do it for you.

shared-libs/couch-request/test/couch-request.js
```diff
@@ -0,0 +1,186 @@
...
+    requestPut.restore();
+    requestDelete.restore();
+    requestHead.restore();
+    sinon.restore();
+  });
+
+  const options = [{ foo: 'bar' }];
+
+
+  options.forEach(option => {
```
garethbowen, line 126: No need to loop over an array of 1...

shared-libs/couch-request/test/couch-request.js
```diff
@@ -0,0 +1,186 @@
...
+    requestGet = sinon.stub(request, 'get');
+    requestPost = sinon.stub(request, 'post');
+    requestPut = sinon.stub(request, 'put');
+    requestDelete = sinon.stub(request, 'delete');
+    requestHead = sinon.stub(request, 'head');
+    requestGet.returns(Promise.resolve('foo'));
+    requestPost.returns(Promise.resolve('foo'));
+    requestPut.returns(Promise.resolve('foo'));
+    requestDelete.returns(Promise.resolve('foo'));
+    requestHead.returns(Promise.resolve('foo'));
```
garethbowen, line 111: It's better to provide the response you want in the test not the setup - right now if the code accidentally called `get` instead of `post` the test would pass because they're all resolving the same thing!

shared-libs/couch-request/test/couch-request.js
```diff
@@ -0,0 +1,186 @@
...
+  });
+
+  afterEach(() => {
+    requestGet.restore();
+    requestPost.restore();
+    requestPut.restore();
+    requestDelete.restore();
+    requestHead.restore();
+    sinon.restore();
+  });
```
garethbowen, line 121: I don't think this is necessary - sinon should restore automatically.

tests/e2e/upgrade/wdio.conf.js
```diff
@@ -8,7 +8,7 @@ const os = require('os');
 const chai = require('chai');
 const { spawn } = require('child_process');
 chai.use(require('chai-exclude'));
-const rpn = require('request-promise-native');
+const request = require('request-promise-native');
```
garethbowen, line 11: As above - don't rename the variable if nothing actually changed.

sentinel/src/schedule/reminders.js
garethbowen: I think I'd feel more comfortable if we changed the implementation to support passing the URL as a string so it's more compatible with the RPN way.

garethbowen (changes requested):
Thanks for this!

I've suggested a range of tweaks, but nothing too major I hope!

api/tests/mocha/services/setup/utils.spec.js
fardarter: This resolved some flakiness on the testing. It's a fix for the testing not the library.

api/src/routing.js
fardarter: Will do.

.eslintrc
```diff
@@ -53,7 +53,8 @@
     "semi-spacing": ["error", { "before": false, "after": true }],
     "semi-style": ["error", "last"],
     "template-curly-spacing": "error",
-    "unicode-bom": ["error", "never"]
+    "unicode-bom": ["error", "never"],
+    "eqeqeq": ["error", "always", { "null": "ignore" }] // foo != null is a convenient way to check for null or undefined.
```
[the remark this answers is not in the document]
fardarter: Will put in a new commit.

api/src/routing.js
fardarter: I'll look at it.

shared-libs/couch-request/test/couch-request.js
fardarter: I generally do this so that new people can extend instead of having to restructure.

shared-libs/couch-request/test/couch-request.js
fardarter: The lodash implementation pulled in a fairly large set other deps last time I checked it.

shared-libs/couch-request/test/couch-request.js
fardarter: I'll have a look.

api/tests/mocha/services/setup/utils.spec.js
garethbowen: That's what I'm worried about - unit test should not be flake because they control their environment. Let's see if we can fix this reliably...

shared-libs/couch-request/test/couch-request.js
garethbowen: Sure but we're already pulling them in everywhere this is used so should not add to the overall bundle size.

sentinel/src/schedule/reminders.js
fardarter:
```suggestion
  How confident are you that there's nowhere else in the repo that calls rpn with the URL string as the first param? I'm 
  worried there's somewhere else that maybe doesn't have test coverage which will now fail...
```
I've been through it with a fine toothed comb, but can do so again and report.

> I think I'd feel more comfortable if we changed the implementation to support passing the URL as a string so it's more compatible with the RPN way

If you're resolute on this I can change it, but:
- it makes the code more complex
- passing an object is a better API as everything is named and can't be passed in the wrong order. It also makes the code robust to any future API changes that require call site changes (as this one does). I'd argue that it would be a good pattern to enforce over the first param pattern. Libs like rpn will have prioritised ease of use over correctness and robustness, but we don't have to.

Let me know if I've changed your mind or not.

shared-libs/couch-request/src/couch-request.js
fardarter: This is using the value as the key to avoid the value comparison check, so it has to be the same as the passed value from `process.env.COUCH_URL`

I can change the variable name to lower case. I think I was trying to imply that it was an env variable but I can see how that is confusing. Would that be adequate?

shared-libs/couch-request/src/couch-request.js
fardarter: I'll look into it. I think one of the issues I had was that the environment isn't available at all the call sites, but I might be wrong.

sentinel/src/schedule/reminders.js
garethbowen: For API design it's better to have required parameters as explicit and optional ones in an object. In this case only the URL is required, so I would argue that the better API design is to remove the URL from the object completely.

Ultimately (once it's supported in Node) we'll switch to using the native `fetch` spec ( https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch ) which does it this way (eg: `fetch(url, options)`) so I'd prefer to align with that standard rather than enforcing our own.

For backwards compatibility I can live with supporting both the standard way, and having the url in the object, but long term we should standardise on the HTML spec.

shared-libs/couch-request/src/couch-request.js
garethbowen: This could be an issue with testing because the env var will be set before you have a chance to mock it. There are other ways around this for testing so I think try it out and look at how you can fix the testing to keep the production code as clean as possible.

shared-libs/couch-request/src/couch-request.js
fardarter: Last I checked the lodash implementation drags in a bunch of dependencies. Importing lodash for one function seems heavyweight.

shared-libs/couch-request/src/couch-request.js
fardarter: Do you have an example?

shared-libs/couch-request/src/couch-request.js
garethbowen: I don't understand why you need to set `COUCH_URL` AND `hostname` as `hostname` is the only one that's read. If you want to avoid checking again then check if `hostname` is set.

shared-libs/couch-request/src/couch-request.js
fardarter: Let me revisit.

shared-libs/couch-request/src/couch-request.js
garethbowen: You can use `rewire` to set variables directly (search through the code for examples) or you can expose a function like `_init` which is called on init and also can be called by tests (this second option is a bit ugly because you're just trusting devs to ignore underscore prefixed functions).

shared-libs/couch-request/src/couch-request.js
fardarter: Only downside here is that if there are method specific conditions the code gets a bit complex. My instinct is to wait a bit to see if an abstraction is required before adding one. Your call though. If you're firm on it, I'll make the change.

shared-libs/couch-request/test/couch-request.js
fardarter: I seem to remember it also doesn't cover as many cases as this implementation, but let me do a test run and report back for your decision.

.eslintrc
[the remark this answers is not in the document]
fardarter: New commit or new PR?

api/src/routing.js
fardarter: Will look into it.

api/tests/mocha/services/setup/utils.spec.js
fardarter: I'll revisit. It was intermittent enough of a failure that I might just revert the change.

scripts/build/index.js
fardarter: Will do.

sentinel/.eslintrc
fardarter: I'll make that now.

.eslintrc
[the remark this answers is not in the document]
garethbowen: Sorry, new PR. Because we generally "squash and merge" each PR ends up in a single commit to the main branch.

sentinel/src/schedule/reminders.js
fardarter: Disagree on the pattern, as I've seen too many bugs from people changing param ordering both at call site and arguments. I agree with using the low-level API the way you say, but personally wouldn't support it up the stack as we don't have the same kind of API guarantees with code we own as we do with the spec (depending on who is reviewing).

But it's your rodeo -- will make the change. It's a subtle disagreement anyway and hopefully non-consequential.

shared-libs/couch-request/src/couch-request.js
fardarter: Thanks, I'll have a look.

shared-libs/couch-request/src/couch-request.js
fardarter: I see what you're saying.

package.json
fardarter: Sure

.eslintrc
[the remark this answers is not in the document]
fardarter: Have PRed to the `eslint-config` repo. Would it be possible to get that merged if acceptable and then I can version bump here? https://github.com/medic/eslint-config/pull/4

.gitattributes
```diff
@@ -0,0 +1,57 @@
...
+*.pdf binary
+*.png binary
+*.psd binary
+*.webp binary
+
+# Fonts
+*.woff2 binary
+
+# Other
+*.exe binary
```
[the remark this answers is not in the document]
fardarter: @garethbowen Have made a PR on this: https://github.com/medic/cht-core/pull/8991

garethbowen:
@fardarter Let me know when this is ready for another round.

fardarter:
> @fardarter Let me know when this is ready for another round.

@garethbowen I'll tag you directly.

api/src/routing.js
fardarter: By these do you mean the ultimate value for the process.env? Not sure I follow if you're referencing the functions as they're initialised in the file scope and initialised once at `server.js`. Happy to move the variables to  `environment.js` though.

api/src/routing.js
fardarter: > Surely if you want to changeOrigin for one, you want to do it for all 3?

I didn't want to foreclose on optionality here, so I made each client individually configurable. Would you prefer collapsing to one variable or having an override that sets them all while retaining the granularity?

shared-libs/couch-request/src/couch-request.js
fardarter: > Also, let's only set servername if env.PROXY_CHANGE_ORIGIN is set.

These are not strictly equivalent, so coupling them could block someone down the line. 

What about an env variable to enable the behaviour?

shared-libs/couch-request/src/couch-request.js
```diff
@@ -0,0 +1,114 @@
...
+// 'ERR_TLS_CERT_ALTNAME_INVALID'
+// "RequestError: Error [ERR_TLS_CERT_ALTNAME_INVALID]: Hostname/IP does not match certificate's altnames:
+//  Host: abc.com. is not in the cert's altnames: DNS:def.org"
+// 
+// The addition of 'servername' resolves this error. See docs for 'tls.connect(options[, callback])'
+//  (https://nodejs.org/api/tls.html): "Server name for the SNI (Server Name Indication) TLS extension  It is the
+//  name of the host being connected to, and must be a host name, and not an IP address.".
+// 
+
+const testReqInputs = (method, first, second = {}) => {
```
fardarter, line 43: @garethbowen Refactoring this into its own function was purely to please SonarQube. Personally think the code is more readable and less complex in a single function.

.eslintrc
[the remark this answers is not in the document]
garethbowen: It looks like there's some debate about the right approach on that. I suggest making this PR pass lint without that change so it's not blocked.

api/src/routing.js
garethbowen: Yeah let's not add complexity that we don't know there's a requirement for it. Stick with 1 if that works for your use case, and it'll be easy to add the other two if we have a good reason to.

api/src/routing.js
garethbowen: Ahh yes I see what you mean. Now that it's in environment I think this is clearer.

api/src/environment.js
```diff
@@ -1,9 +1,24 @@
...
+const { 
+  UNIT_TEST_ENV,
+  COUCH_URL,
+  BUILDS_URL, 
+  // If set to true, PROXY_CHANGE_ORIGIN_ALL will override the values of PROXY_CHANGE_ORIGIN, 
+  // PROXY_CHANGE_ORIGIN_AUTH and PROXY_CHANGE_ORIGIN_CHANGES and set them to true.
+  PROXY_CHANGE_ORIGIN_ALL = false, 
+  PROXY_CHANGE_ORIGIN = false,
+  PROXY_CHANGE_ORIGIN_AUTH = false,
+  PROXY_CHANGE_ORIGIN_CHANGES = false
```
garethbowen, line 13: Can you think of any time when you would want one proxy to have change origin but not the others? Because they're all going to proxy to the same couch instance the sni will either be required by all or by none. If you set one to true and the others to false you would end up with a really hard to debug setup. This complexity makes life much harder for implementers and unless there's a really good reason to include it we should aim to make their job as easy as possible.

api/src/routing.js
```diff
@@ -12,15 +12,21 @@ const prometheusMiddleware = require('prometheus-api-metrics');
 const rateLimiterMiddleware = require('./middleware/rate-limiter');
 const logger = require('./logger');
 const isClientHuman = require('./is-client-human');
-const target = 'http://' + environment.host + ':' + environment.port;
-const proxy = require('http-proxy').createProxyServer({ target: target });
+
+const port = typeof environment.port !== 'undefined' && environment.port !== null ? `:${environment.port}` : '';
+const target = `${environment.protocol}//${environment.host}${port}`;
+const proxy = require('http-proxy').createProxyServer({
+  target: target, changeOrigin: environment?.proxies?.default?.changeOrigin ?? false
```
garethbowen, line 19: Nit: For formatting let's put each value on its own line just like the two proxies below...

```suggestion
  target,
  changeOrigin: environment?.proxies?.default?.changeOrigin ?? false 
```

api/src/routing.js
```diff
@@ -12,15 +12,21 @@ const prometheusMiddleware = require('prometheus-api-metrics');
...
+
+const port = typeof environment.port !== 'undefined' && environment.port !== null ? `:${environment.port}` : '';
+const target = `${environment.protocol}//${environment.host}${port}`;
+const proxy = require('http-proxy').createProxyServer({
+  target: target, changeOrigin: environment?.proxies?.default?.changeOrigin ?? false 
+});
 const proxyForAuth = require('http-proxy').createProxyServer({
   target: target,
   selfHandleResponse: true,
+  changeOrigin: environment?.proxies?.auth?.changeOrigin ?? false
```
garethbowen, line 24: I think it'd be cleaner to remove the `?? false` and just make sure the value is always defined. I think you just need to provide a default in the unit testing setup. That simplifies this significantly, to

```suggestion
  changeOrigin: environment.proxies.auth.changeOrigin
```

shared-libs/couch-request/src/couch-request.js
```diff
@@ -0,0 +1,114 @@
...
+    return { shouldReject: true, errorMessage: '"options" must be a plain object', firstIsString };
+  }
+  
+  return { shouldReject: false, errorMessage: '', firstIsString };
+};
+
+
+const req = (method, first, second = {}) => {
+
+  const { shouldReject, errorMessage, firstIsString = false } = testReqInputs(method, first, second);
```
garethbowen, line 65: One of the reasons this function is complicated is you're doing two things - one for working out if the first param is a string, and one to validate inputs. This would be cleaner if you cleaned up the parameters (and renamed them from `first`, and `second` before calling the `testReqInputs` function.

Then you can rename `testReqInputs` to `validate` which is actually what you're doing.

Finally you can have validate throw the exception directly which helps get a more accurate stack trace as well as cleaning up the code a lot.

I haven't tested this and it has wider implications but I think you can see what I mean below...

```suggestion
const req = (method, first, options = {}) => {
  if (typeof first === 'string' || first instanceof String) {
    options.url = first;
  }
  try {
    validate(method, options);
  } catch(e) {
    return Promise.reject(e);
  }
```

shared-libs/transitions/package.json
```diff
@@ -35,7 +36,6 @@
     "mustache": "^4.2.0",
     "object-path": "^0.11.8",
     "request": "^2.88.2",
```
garethbowen, line 38: Can we also remove "request" here?

garethbowen (changes requested):
Nice, thanks for simplifying this PR to just the necessary bits!

There are a couple of suggestions to reduce complexity, both for developers and implementers but nothing too complex.

shared-libs/couch-request/src/couch-request.js
fardarter: I'll look at the code, but I don't see a better rename for the inputs. 

`first` can either be the url or the `options` object, and I don't have a unifying name for that. `second` can be the `options` object or nothing, but if first is the `options`, then it doesn't make sense for `second` to be named `options`. 

would be nice to have some decent overloading in js. 

I also suspect that can't add another logic branch to the function without triggering sonarqube, but will test that if that's true if the changes make sense.

api/src/environment.js
fardarter: Well, the existence of the multiple clients means that conceptually they could be pointed at differently configured entities, but if you don't know of any use cases, I'm OK with collapsing them. 

My general bent is to expose configurability/apis where possible and to try not predict usage too much.

Will change.

shared-libs/transitions/package.json
fardarter: Will do.

api/src/routing.js
fardarter: @garethbowen have you ever considered adding prettier to the project?

garethbowen: Yes. It's very challenging to add it to an existing project because every time you modify a file it makes a bunch of unrelated changes that make the review difficult. Furthermore the developers agreed it actually made the code less readable so we removed it.

fardarter: It only makes unrelated changes if it isn't enforced in CI?

I've moved projects over to it. You have a single format change and then there's no noise on PRs. 

In fact, the main reasons I use it is to a) reduce noise on PRs b) avoid formatting issues coming up/being discussed in review. 

If IDE integrated on save, it does have another advantage, which is that you get a snap to format on code that is structurally valid, so it's really great for hunting for closing brackets quickly. 

I agree it's not always the best defaults (the ternary formatting is bad, for example), but there are few things I love less than debating formatting on review. 

Speaking of which, going to make the requested change.

fardarter:
@garethbowen Please won't you rerun the tests? I've made the requested changes. Hope this passes and is ready.

api/src/environment.js
```diff
@@ -1,9 +1,20 @@
 const path = require('path');
 const logger = require('./logger');
 
-const { UNIT_TEST_ENV, COUCH_URL, BUILDS_URL } = process.env;
+const { 
+  UNIT_TEST_ENV,
+  COUCH_URL,
+  BUILDS_URL, 
+  // If set to true, PROXY_CHANGE_ORIGIN_ALL will override the values of PROXY_CHANGE_ORIGIN, 
+  // PROXY_CHANGE_ORIGIN_AUTH and PROXY_CHANGE_ORIGIN_CHANGES and set them to true.
```
garethbowen, line 9: This comment is out of date now I think?

shared-libs/couch-request/src/couch-request.js
```diff
@@ -0,0 +1,112 @@
...
+
+
+const req = (method, first, second = {}) => {
+
+  const firstIsString = typeof first === 'string' || first instanceof String;
+
+  try {
+    validate(firstIsString, method, first, second);
+  } catch (e) {
+    return Promise.reject(Error(e));
```
garethbowen, line 66: There is no need to wrap the error in another error.

```suggestion
    return Promise.reject(e);
```

shared-libs/couch-request/src/couch-request.js
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  if (isPlainObject(second) === false) {
+    throw new Error(`"options" must be a plain object'`);
+  }
+
+  if (firstIsString === false && isPlainObject(first) === false) {
+    throw new Error(`"options" must be a plain object'`);
+  }
+  return;
```
garethbowen, line 55: Unnecessary return (picked up by sonar).

```suggestion
```

shared-libs/couch-request/src/couch-request.js
```diff
@@ -0,0 +1,112 @@
...
+};
+
+
+const req = (method, first, second = {}) => {
+
+  const firstIsString = typeof first === 'string' || first instanceof String;
```
garethbowen, line 61: You created the `isString` function which does this too - call that for reuse. I think this will also reduce the cognitive complexity enough to get sonar to pass.

```suggestion
  const firstIsString = isString(first);
```

garethbowen (changes requested):
Thank you for sticking with this!

There's a small batch of simple changes, and we'll need to pass the sonar check, which I think my suggestions will resolve.

Let me know when these are done and I'll make sure the builds pass.

shared-libs/couch-request/src/couch-request.js
fardarter: Sigh. Can't agree with SQ here. Indirection is a major source of cognitive complexity. I've never liked Uncle Bob's approach, which seems to be driving the ruleset. 

Abstraction IMO is both a cognitive and design overhead and should only be introduced with clear benefit. Use case in the other instance was a consistent way to normalise env inputs.

Some posts on the topic: 
- https://dev.to/wuz/stop-trying-to-be-so-dry-instead-write-everything-twice-wet-5g33
- https://kentcdodds.com/blog/aha-programming

But I'll make the change. Just offering thoughts for future consideration.

shared-libs/couch-request/src/couch-request.js
fardarter: Will fix

api/src/environment.js
fardarter: Yes, thanks

shared-libs/couch-request/src/couch-request.js
fardarter: Will fix

fardarter:
> Thank you for sticking with this!
> 
> There's a small batch of simple changes, and we'll need to pass the sonar check, which I think my suggestions will resolve.
> 
> Let me know when these are done and I'll make sure the builds pass.

We need it, so have to. But I've been around long enough to know one has to listen to the rules of the house. No worries. 

While I have your attention, would you be open to my backporting this to 4.3.x once it is in?

fardarter:
@garethbowen Have amended as requested.

garethbowen (approved):

garethbowen:
> While I have your attention, would you be open to my backporting this to 4.3.x once it is in?

The problem with backporting to `4.3.x` is you'd also have to backport to every release branch between that and `master` (4.4.x, 4.5.x, and 4.6.x) because otherwise someone updating from 4.3 to 4.4 would have a breaking change. Also because it touches so many files it's going to be difficult and risky to do. I don't see the value in that as as far as I know you're the only team who need it so I personally would not approve the backport.

Given the number of changes your team has made it looks like you've got your own fork already so I recommend you backport it to your fork.

Would you be amenable to also providing documentation in the cht-docs repo for this change?

fardarter:
> > While I have your attention, would you be open to my backporting this to 4.3.x once it is in?
> 
> The problem with backporting to `4.3.x` is you'd also have to backport to every release branch between that and `master` (4.4.x, 4.5.x, and 4.6.x) because otherwise someone updating from 4.3 to 4.4 would have a breaking change. Also because it touches so many files it's going to be difficult and risky to do. I don't see the value in that as as far as I know you're the only team who need it so I personally would not approve the backport.
> 
> Given the number of changes your team has made it looks like you've got your own fork already so I recommend you backport it to your fork.
> 
> Would you be amenable to also providing documentation in the cht-docs repo for this change?

We have the repo as a submodule with an explicit patch folder, so trying to reduce the overhead/complexity of the patch folder. I take your point, though. We'll just have to catch up before removing the patch.

> Would you be amenable to also providing documentation in the cht-docs repo for this change?

Sure, point me in the right direction.

garethbowen:
I guess add a section on this page: https://docs.communityhealthtoolkit.org/apps/guides/hosting/4.x/adding-tls-certificates/

If you go there you can click the "edit this page" link at the top right which will take you to the right place in the repo. You can either edit through GH or check out the repo as per usual.

garethbowen:
Merged! Thank you @fardarter !!

fardarter:
> I guess add a section on this page: https://docs.communityhealthtoolkit.org/apps/guides/hosting/4.x/adding-tls-certificates/
> 
> If you go there you can click the "edit this page" link at the top right which will take you to the right place in the repo. You can either edit through GH or check out the repo as per usual.

Do you have a page for environment variables?

garethbowen:
There's some info here: https://docs.communityhealthtoolkit.org/apps/guides/hosting/4.x/self-hosting/multiple-nodes/#prepare-environment-variables-file

Issue #8789, opened 2024-01-08
Target aggregates test is passing but shouldn't be


garethbowen:
**Describe the issue**

Despite the test logging what looks like a fatal error the test is passing.

Build: https://github.com/medic/cht-core/actions/runs/7444119349/job/20250181759

```
[0-22] RUNNING in chrome - file:///tests/e2e/default/targets/target-aggregates.wdio-spec.js
[0-22] 2024-01-08T08:20:34.161Z ERROR webdriver: Request failed with status 400 due to invalid selector: The selector ".cell.what a target! label" used with strategy "css selector" is invalid! For more information on selectors visit the WebdriverIO docs at: https://webdriver.io/docs/selectors
[0-22] 2024-01-08T08:20:35.151Z ERROR webdriver: Request failed with status 400 due to invalid selector: The selector ".cell.what a target! label" used with strategy "css selector" is invalid! For more information on selectors visit the WebdriverIO docs at: https://webdriver.io/docs/selectors
[0-22] Error getting userCtx Unexpected token o in JSON at position 1
[0-22] PASSED in chrome - file:///tests/e2e/default/targets/target-aggregates.wdio-spec.js
```

The CSS selector is invalid probably because it has a `!` in it, but I think it's a bad selector anyway. Even if we remove the `!` you're still left with ".cell.what a target label" which is weird at best. The generated label for this target is "what a target!" so I suspect the selector should be more like `.cell label["what a target!"]` or something.

**Describe the improvement you'd like**

Figure out why the test is failing, and fix the CSS selector to not throw this error.

garethbowen:
Merged

PR #8834, opened 2024-01-31
fix(#8789): targets aggregates test is failing silently

garethbowen:
# Description

#8789 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8789-fix-target-aggregates-test/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8789-fix-target-aggregates-test/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8789-fix-target-aggregates-test/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/tests/e2e/default/targets/target-aggregates.wdio-spec.js
+++ b/tests/e2e/default/targets/target-aggregates.wdio-spec.js
@@ -62,10 +62,15 @@ const updateSettings = async (targetsConfig, user, contactSummary) => {
   await commonPage.goToBase();
 };
 
-const validateCardField = async (label, value) => {
-  const card = await contactsPage.getCardFieldInfo(label);
-  expect(card.label).to.equal(label);
-  expect(card.value).to.equal(value);
+const validateCardFields = async (lastUpdated, firstTarget, secondTarget) => {
+  const fields = await contactsPage.getContactSummaryCardFields();
+  expect(fields.length).to.equal(3);
+  expect(fields[0].label).to.equal('Last updated');
+  expect(fields[0].value).to.equal(lastUpdated);
+  expect(fields[1].label).to.equal('what a target!');
+  expect(fields[1].value).to.equal(firstTarget);
+  expect(fields[2].label).to.equal('the most target');
+  expect(fields[2].value).to.equal(secondTarget);
 };
 
 describe('Target aggregates', () => {
@@ -367,9 +372,8 @@ describe('Target aggregates', () => {
       expect(await contactsPage.getContactInfoName()).to.equal('Clarissa');
       // assert that the activity card exists and has the right fields.
       expect(await contactsPage.getContactCardTitle()).to.equal('Activity this month');
-      await validateCardField('Last updated', 'yesterday Clarissa');
-      await validateCardField('what a target!', '40');
-      await validateCardField('the most target', '50%');
+
+      await validateCardFields('yesterday Clarissa', '40', '50%');
 
       await browser.back();
       await targetAggregatesPage.expectTargetDetails(expectedTargets[0]);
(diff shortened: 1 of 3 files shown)


tests/e2e/default/targets/target-aggregates.wdio-spec.js
```diff
@@ -62,10 +62,16 @@ const updateSettings = async (targetsConfig, user, contactSummary) => {
   await commonPage.goToBase();
 };
 
-const validateCardField = async (label, value) => {
-  const card = await contactsPage.getCardFieldInfo(label);
-  expect(card.label).to.equal(label);
-  expect(card.value).to.equal(value);
+const validateCardFields = async (lastUpdated, firstTarget, secondTarget) => {
+  const fields = await contactsPage.getContactSummaryCardFields();
+  console.log(JSON.stringify(fields, null, 2));
```
garethbowen, line 67: delete me

tests/e2e/default/targets/target-aggregates.wdio-spec.js
```diff
@@ -367,9 +372,8 @@ describe('Target aggregates', () => {
       expect(await contactsPage.getContactInfoName()).to.equal('Clarissa');
       // assert that the activity card exists and has the right fields.
       expect(await contactsPage.getContactCardTitle()).to.equal('Activity this month');
-      validateCardField('Last updated', 'yesterday Clarissa');
-      validateCardField('what a target!', '40');
-      validateCardField('the most target', '50%');
```
garethbowen, line 372: The reason the tests were failing silently is these validations weren't being `await`ed. As soon as I added the `await` the test failed.

tests/page-objects/default/contacts/contacts.wdio.page.js
```diff
@@ -363,11 +363,17 @@ const exportContacts = async () => {
   await (await exportButton()).click();
 };
 
-const getCardFieldInfo = async (label) => {
-  return {
-    label: await (await $(`.cell.${label} label`)).getText(),
```
garethbowen, line 368: This never worked. We didn't have a class for the label, and even if we did, we we using spaces in the label which would have broken this CSS selector completely.

tests/e2e/default/targets/target-aggregates.wdio-spec.js
```diff
@@ -62,10 +62,15 @@ const updateSettings = async (targetsConfig, user, contactSummary) => {
   await commonPage.goToBase();
 };
 
-const validateCardField = async (label, value) => {
-  const card = await contactsPage.getCardFieldInfo(label);
-  expect(card.label).to.equal(label);
-  expect(card.value).to.equal(value);
+const validateCardFields = async (lastUpdated, firstTarget, secondTarget) => {
+  const fields = await contactsPage.getContactSummaryCardFields();
```
lorerod, line 66: Is it possible to use text selectors for contact summary card fields? 
And, validate the text directly instead of first getting the fields and then validating the text. Something similar to what @tatilepizs  has done in [common-enketo page objects](https://github.com/medic/cht-core/blob/648d47728c714816b4e979a519dd744b23ffcb8f/tests/page-objects/default/enketo/common-enketo.wdio.page.js#L60).

lorerod (approved):
Hi @garethbowen, thanks for fixing the e2e test. I left a suggestion, but it's not a blocker.

tests/e2e/default/targets/target-aggregates.wdio-spec.js
garethbowen: Oooh yes I like this. Initially I tried to keep the API as close as possible to the original, but seeing as it's only used in this one place I think I can change it to do whatever I like. Thanks for the prompt!

PR #8838, opened 2024-02-01
Fix bug to correctly record Tasks Telemetry

Benmuiruri:
# Description

Record the correct Telemetry for Tasks Component. 
`tasks:load` on first time loading tasks 
`tasks:refresh` on subsequent visits to Tasks tab.

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:fix-tasks-telemetry/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:fix-tasks-telemetry/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:fix-tasks-telemetry/docker-compose/cht-couchdb-clustered.yml)

# License


webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -155,20 +153,21 @@ export class TasksComponent implements OnInit, OnDestroy {
...
       this.loading = false;
       this.hasTasks = false;
       this.tasksActions.setTasksList([]);
     } finally {
+      const performanceName = this.tasksLoaded ? 'tasks:refresh' : 'tasks:load';
       this.trackPerformance?.stop({
-        name: this.tasksLoaded ? 'tasks:refresh' : 'tasks:load',
-        recordApdex: !this.tasksLoaded,
+        name: performanceName,
+        recordApdex: true,
```
latin-panda, line 166: Here, we only want to record the apdex for `tasks:load`

Benmuiruri: Why would we want to record only `tasks:load` from there?  `recordApdex: true` allows us to record both apdex from one place.

latin-panda: is the `task:refresh` only called when task tab loads? or triggered by something else?

Benmuiruri: First time we record `tasks:load`, subsequent visits we record `tasks:refresh`

https://github.com/medic/cht-core/assets/2597305/9215c155-f38f-4258-b277-3633c1a38522

latin-panda: Got it! Thanks for the video, then this makes sense

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -105,15 +105,13 @@ export class TasksComponent implements OnInit, OnDestroy {
...
     this.currentLevel = this.sessionService.isOnlineOnly() ? Promise.resolve() : this.getCurrentLineageLevel();
-
     this.refreshTasks();
   }
 
   ngOnDestroy() {
     this.subscription.unsubscribe();
 
     this.tasksActions.setTasksList([]);
-    this.tasksActions.setTasksLoaded(false);
```
latin-panda, line 116: Why remove this? Is it because we keep tasks in the Store and not refreshing?

latin-panda (commented):
There is a unit test failing, can you please have a look?

webapp/src/ts/modules/tasks/tasks.component.ts
Benmuiruri: That was part of my understanding of the fix. If we setTasksLoaded as false on `ngOnDestroy()` it means when you navigate to a different tab then back to the Tasks Tab, you will always be logging `tasks:load` everytime

latin-panda: Yes, that's okay; it's the same behavior for the contacts, reports, and messages tab. Whenever the chw navigates to any of those tabs, it will record an Apdex; because the background processes can affect the experience every time the chw visits the tabs.

latin-panda, line 116: @Benmuiruri, based on yesterday's conversation, we rolled back this change. And just keep the one from L163 to L70

webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
```diff
@@ -115,8 +114,6 @@ describe('TasksComponent', () => {
     expect(spySubscriptionsUnsubscribe.callCount).to.equal(1);
     expect(setTasksList.callCount).to.equal(1);
     expect(setTasksList.args[0]).to.deep.equal([[]]);
-    expect(setTasksLoaded.callCount).to.equal(1);
-    expect(setTasksLoaded.args[0]).to.deep.equal([false]);
```
latin-panda, line 119: Please rollback this test too.

latin-panda (changes requested):
Thanks Ben! Just two tiny things and we are good to go

latin-panda (approved):
LGTM! 

@ralfudx you can find the conversation about [this solution here](https://medic.slack.com/archives/C02GME3BQE6/p1706878600642449)

@Benmuiruri please upload a video with the test evidence, and also remember to update the ticket with the solution. Thanks for fixing this!

Benmuiruri:
Here's a video showing we now correctly record `tasks:refresh` and `
https://github.com/medic/cht-core/assets/2597305/53b1ecea-3ce7-4d92-a466-6e93410da9b7
`

Benmuiruri:
Here's a video showing we now correctly record tasks:refresh and tasks:load


https://github.com/medic/cht-core/assets/2597305/2af9c6d2-7351-4a09-a9db-a09308537f39

ralfudx (approved):
LG

Issue #8837, opened 2024-01-31
Telemetry date not logged on telemetry documents


1yuv:
**Describe the bug**
It has been observed that for many users, a telemetry document is captured without a proper date.

Look at this example telemetry doc: 
```json
 "metadata": {
    "year": null,
    "month": null,
    "day": null,
    "user": "_sample_",
    "deviceId": "d496ea42-b24f-4864-8f99-0aa7bcda5179",
    "versions": {
      "app": "4.5.0",
      "forms": {
        "contact:clinic:create": "8-cf7d43ccfbaf887faf1b6550ecde6d8f",
```

**To Reproduce**
There's no particular way to reproduce this, but these documents can be seen in many 4.5.0 production instances. I'll share the link to these documents in slack. 

**Expected behavior**
Proper date should be logged for year, month, day field. 

**Environment**
- Instance: achham, bardiya, and other 4.5 instances
- App: Webapp
- Version: 4.5.0

**Additional context**
This might have been caused after [this fix](https://github.com/medic/cht-core/issues/7616). Issues related to missing days in telemetry document was fixed in that issue. 

This issue has adverse effect in data reporting since we sync telemetry documents by couch2pg and since couch2pg [view](https://github.com/medic/cht-couch2pg/blob/main/libs/medic-users-meta/migrations/202304140052.do.telemetry_devices.sql) doesn't proper date, the couch2pg fails causing entire data transfer failing for all instances running 4.5.0.

latin-panda:
I have been trying to reproduce this issue and see if there is any bad telemetry db running 4.2.x to 4.5.x branches. Like removing cookies (telemetry has the user name), clearing localStorage, offline user, online user, etc.
But so far, no luck. @ralfudx can you please try and see if you can reproduce it? 

I've added a stronger check to only process telemetry db with correct names and throw an exception if the dates are NaN, which can help us track better next time.

garethbowen:
> I've added a stronger check to only process telemetry db with correct names and throw an exception if the dates are NaN, which can help us track better next time.

I don't think this will work, because the `deleteDeprecatedTelemetryDB` function is called before the `getDbDate` function, so the new Error code will never run. I think the unit test proves this because of this line: `expect(consoleErrorSpy.notCalled).to.be.true;`

Right?

latin-panda:
> I don't think this will work, because the deleteDeprecatedTelemetryDB function is called before the getDbDate function, so the new Error code will never run. 

[...]

PR #8839, opened 2024-02-01
fix(#8837): fix NaN in telemetry doc id

latin-panda:
# Description

- Adds regex to check the telemetry name
- Adds warning for telemetry names that don't match the new name format.

Testing: https://github.com/medic/cht-core/assets/66472237/436da61e-ef74-4466-8481-66445beb6932

#8837

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:check-telemetry-db-name/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:check-telemetry-db-name/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:check-telemetry-db-name/docker-compose/cht-couchdb-clustered.yml)

# License


latin-panda:
@garethbowen @ralfudx I ran the v4.2.x and 4.5.x branches, and after several tries, I couldn't get a bad telemetry record. I straightened the validations and added an exception, so next time, we will have more info from the feedback docs. 
Let me know your suggestions. Thanks!

garethbowen:
@latin-panda Thanks for getting right on this!

To do a review I need more information, like how you determined this is the problem, and why this is the right fix. The best place for this is probably a comment on the issue.

latin-panda:
>  how you determined this is the problem

@garethbowen I haven't been able to reproduce it. So maybe @ralfudx can help? 
For now, this is just making the validations stronger by processing the telemetry db names with the correct format. 
I'll add this in the ticket in a couple of minutes

webapp/src/ts/services/telemetry.service.ts
```diff
@@ -345,10 +356,8 @@ export class TelemetryService {
     databaseNames?.forEach(dbName => {
       const nameNoPrefix = dbName?.replace(this.POUCH_PREFIX, '') || '';
 
-      // Skips new Telemetry DB, then matches the old deprecated Telemetry DB.
-      if (!nameNoPrefix.startsWith(this.TELEMETRY_PREFIX)
-        && nameNoPrefix.includes(this.TELEMETRY_PREFIX)
-        && nameNoPrefix.includes(this.sessionService.userCtx().name)) {
```
garethbowen, line 351: Why did you remove the check for `userCtx`? Without this won't we end up deleting someone elses telemetry on phones that are shared between users?

garethbowen:
@latin-panda Sorry I wrote that comment before you updated the PR description and added the comment. I deleted it again once I saw the new details. Sorry for the confusion!

garethbowen (changes requested):
I worry this is going to hide the error, because now any database that doesn't match the name will be deleted. In other words, if this happens again it'll silently fail rather than loudly fail. Because we haven't managed to reproduce this yet, we can't be certain if deletion is the right approach or if there's a bug in how we're creating the dbs in the first place.

Let's put some more time into trying to reproduce it.

webapp/src/ts/services/telemetry.service.ts
latin-panda: ah shared phones, okay putting it back

latin-panda:
@ralfudx and @garethbowen please have a look. The momentjs format was translating the numbers to Nepali, I have updated the ticket with this.  

@ralfudx Please check the PR description, I have added a video showing how to test this error. Let me know any questions or suggestions.

garethbowen (approved):
LGTM. Thanks for adding the bikram sambat test!

ralfudx (approved):
LGTM

Issue #8338, opened 2023-06-22
Research increasing pbkdf2 iterations


mrjones-plip:
**Describe the issue**
Best practices for hashing passwords is to use the highest number of iterations possible, with out impacting server performance (too much ; ).  Currently our `iterations` is [set to](https://github.com/medic/cht-core/blob/0c7399770fdcb94dbbc1764f25e3141e6b79b5b9/shared-libs/settings/src/index.js#L139) `10`.  This is considered very low, [it looks like](https://docs.couchdb.org/en/stable/config/auth.html#chttpd_auth/iterations) CouchDB default is 10k and minimum is 100.

**Describe the improvement you'd like**
Increase the  `iterations` to the highest number reasonable for the [server specs we set](https://docs.communityhealthtoolkit.org/apps/guides/hosting/requirements/).

**Describe alternatives you've considered**
NA

fardarter:
Would bcrypt not be a better option in general?

garethbowen:
@fardarter Somewhat, yes. pbkdf2 with high iterations should be good enough though.

We hand authentication off to couchdb rather than implementing it ourselves and the configuration option to change [hash algorithms](https://docs.couchdb.org/en/stable/config/auth.html#chttpd_auth/hash_algorithms) was only added in v3.3.0. So it's not possible currently. We should evaluate this after #6289 lands.

garethbowen:
Bumping to 4.4.0 as we're almost ready to release 4.3.0 and this has more work needed yet.

fardarter:
See: [Public Comments on SP 800-132, Recommendation for Password-Based Key Derivation: Part 1: Storage 
Applications](https://csrc.nist.gov/csrc/media/Projects/crypto-publication-review-project/documents/initial-comments/sp800-132-initial-public-comments-2023.pdf) via : https://docs.python.org/3/library/hashlib.html#key-derivation

Looks like recommended min for `sha1` is 1300000 iterations.

garethbowen:
One significant downside to increasing this number is it will seriously impact anywhere that uses a lot of basic auth, which at the moment is api, sentinel, and cht-conf in particular. What this means is all of these will need to be changed to use session based auth which is not a trivial amount of work.

fardarter:
@garethbowen This seems like a fairly critical thing to address, though?

garethbowen:
> This seems like a fairly critical thing to address, though?

Definitely.

garethbowen:
[...]

PR #8857, opened 2024-02-12
feat(#8338): use PouchDb session plugin in api sentinel and e2e tests

dianabarsan:
# Description

# Code review checklist

# License

### DIFF
--- a/api/package-lock.json
+++ b/api/package-lock.json
@@ -36,6 +36,7 @@
         "pouchdb-core": "^7.3.1",
         "pouchdb-find": "^7.3.1",
         "pouchdb-mapreduce": "^7.3.1",
+        "pouchdb-session-authentication": "^1.0.0",
         "prometheus-api-metrics": "^3.2.2",
         "properties": "^1.2.1",
         "rate-limiter-flexible": "^3.0.2",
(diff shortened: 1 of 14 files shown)


dianabarsan:
Hi @garethbowen . 
This adds the new pouchdb session plugin hopefully ... everywhere? The next step - which I already have code prepared for but doing in a new PR - is to update e2e tests to not do basic auth requests, and with that I believe we are close to removing basic auth across cht-core.

garethbowen (approved):
So nice! I'm curious if, even with our low iterations, this will result in noticeable performance improvements in api or sentinel...

Also, have we announced the plugin in the pouchdb slack channel? It would be good to get more eyes on it.

dianabarsan:
PouchDb slack feedback was to show interest in baking this directly into the default http adapter. I'm happy to do that, but since the plugin has minimal impact on code, I think we can use the plugin until a new PouchDb version that supports session authentication. 
What do you think?

garethbowen:
Definitely, don't wait for PouchDB to get this merged. It's easy to revert this later.

garethbowen:
@dianabarsan This got merged and released in 4.7.0 but was missing from the release notes because it's attributed to the overarching "increase pbkdk2 iterations" issue which isn't yet closed. It's probably worth raising a new issue which talks about this specific change, adding it to the 4.7.0 milestone, and updating the release notes, to help tracking when this actually landed. What do you think?

dianabarsan:
Yea. I agree. This is a good idea.

dianabarsan:
I've created an issue and updated the milestone: https://github.com/medic/cht-core/issues/9102

Issue #8862, opened 2024-02-12
Bulk password change script causing `503` errors?


mrjones-plip:
**Describe the issue**
@billwambua reports the `bulk-password-update-export.js` [script](https://github.com/medic/cht-core/blob/master/scripts/bulk-password-update-export.js) added in #8843:

> I've tried using this script. There seems to be an issue. Every time I use the script, the instance ends up getting 503 bad gateway errors. It has happened twice now.
>
> The first time I'm not sure whether it was due to the script because I hadn't checked the instances before but the second time the instance was definitely working before I ran the script.
>
> Could it be the scripts doing or do you think coincidence?

**Describe the improvement you'd like**
Using the script doesn't cause errors

**Describe alternatives you've considered**
Not use the script

mrjones-plip:
@billwambua - can  you add details on how to reproduce?
* how many users are you changing the password on?
* what version of CHT are you on?
* is this a local dev instance or a Medic hosted instance or other?

mrjones-plip:
Testing on 256 changes (same 5 users many times in one file) hasn't shown the issue on either [default conf](https://github.com/medic/cht-core/tree/master/config/default/) or [echis ke conf](https://github.com/moh-kenya/config-echis-2.0/).  I'm on CHT `4.4.1` but I'll try on `4.5.2` now.

```
 ****** WARNING ******
Continuing will log 256 users out of the CHT until you provide them with their updated password.

Do you want to continue? [y/N]
```

billwambua:
@mrjones-plip I'm using it on https://bomet.echis.go.ke. 

I can't tell which CHT version it is on right now because due to the 502 Bad Gateway errors, I can't login.

I was only trying to reset the password for 2 users as a test.  I've shared the 2 usernames via DM with @mrjones-plip 

I have not been able to recover.

mrjones-plip:
@billwambua - thanks for the details!

I was able to reproduce the error when I accidentally changed the password for the admin user that couch was using, `medic` in my case.

Do you know if this was the case you?  You can check the contents of your `user-password-change.txt` file.

billwambua:
@mrjones-plip those 2 usernames I've shared via Slack DM are the only usernames in `user-password-change.txt`

mrjones-plip:
I was able to get a `503` when I accidentally changed the password for my main CHT user which API uses with CouchDB (`medic` is this user for me).  This looked like this in the script:

```
[...]

PR #7080, opened 2021-04-28
Script to bulk set passwords

newtewt:
# Description

Script that gets the list of users on an instance and then sets there password to what is passed as an arg. 

medic/cht-core#[number]

# Code review checklist

- [ ] Readable: Concise, well named, follows the [style guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/), documented if necessary.
- [ ] Documented: Configuration and user documentation on [cht-docs](https://github.com/medic/cht-docs/)
- [ ] Tested: Unit and/or e2e where appropriate
- [ ] Internationalised: All user facing text
- [ ] Backwards compatible: Works with existing data and configuration or includes a migration. Any breaking changes documented in the release notes.

# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

PR #8874, opened 2024-02-14
chore(#8862): bulk script: no password changes of admin/medic users, normalize, allow passwords to be specified 

mrjones-plip:
# Description

This PR updates the existing `bulk-password-update-export.js` [script](https://github.com/medic/cht-core/blob/master/scripts/bulk-password-update-export.js)  to 
* not allow changing passwords for users `medic` or `admin`.  While `4.6.0`+ [defend against](https://github.com/medic/cht-core/issues/8096) this already, earlier 4.x versions do not.
* allows you to specify CSV input of users so they can have their passwords specified instead of randomly added
* adds the same warning to both scripts "ARE YOU SURE YOU WANT TO CONTINUE!" 
* removes `bulk-password-functions.js` 

per #8862

###  `bulk-password-update-export.js` 

Now supports the `--use_passes` flag.  If you pass it `false` it will generate random passwords via `generatePassword()`. If you pass it `true` it will read the CSV.  Passwords that are insifiently complex will get refused with a `400` response and `The password is too easy to guess. Include a range of types of characters to increase the score.` message.

Here's the help text:

```
Change passwords of users in user-password-change.txt, show results on screen 

Usage:
      node bulk-password-update-export.js -h | --help
      node bulk-password-update-export.js --url https://localhost --user medic --password adminPass

Options:
    -h --help     Show this screen.
    --url         The url for the instance being changed
    --user        The admin user this operation is run as
    --password    The password for the admin user 
    --use_passes  The passwords are in user-password-change.txt in or not. 
                  "true" - yes they're there in USER, PASSWORD format
                  "false" - no, no password, just usernames, one per line. Generate a random password for me.
```

###  `bulk-password-functions.js` 

this script has been removed as it's been deemed too dangerous.  At a later date, we can add in an `--all` flag on  `bulk-password-update-export.js`  if needed 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8860-defend-admin-pass-change/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8860-defend-admin-pass-change/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8860-defend-admin-pass-change/docker-compose/cht-couchdb-clustered.yml)

# License


scripts/bulk-password-change.js
```diff
@@ -44,24 +45,20 @@ const options = {
...
-    postOptions.body = {
-      'password': 'Secret_1'
-    };
-    postOptions.uri = `${options.uri}/${user.username}`;
-    try {
-      await rpn.post(postOptions);
-    } catch (e) {
-      console.log('An error while updating the password - ', e.message);
-    }
+    await changeUserPass(user.username, 'Secret_1',  options);
```
garethbowen, line 61: I don't understand the use case for this script. When would someone want to update all passwords to the same insecure one? I think we should deprecate this use case and not recommend it for anyone.

It's convenient to have a "reset all" script, but the passwords should be randomly generated (as in the other script).

Let's just delete this script to be safe. Then if this is a useful feature still, work to implement an `--all` flag in the other script so people have to type something extra to do the more destructive option.

mrjones-plip: There [was a use case](https://github.com/medic/cht-core/pull/7080#pullrequestreview-648719995) that was needed for a QA to test back when. but fair! I'll un-refactor the changes in [9b2d197](https://github.com/medic/cht-core/pull/8874/commits/9b2d1976b5faa8c7f6531d1fc2c15c66fe6d0107), delete the dangerous one and resubmit

mrjones-plip: fixed in 32adfe536!

also @billwambua had some changes he wanted to handle quoted strings which he added with a `String.prototype.splitCSV()` call.

scripts/bulk-password-update-export.js
```diff
@@ -106,30 +127,49 @@ const loadUsers = async () => {
...
-  };
-  postOptions.uri = `${options.uri}/${user}`;
-  try {
-    await rpn.post(postOptions);
-    console.log('SUCCESS', user, newPass);
-  } catch (e) {
-    console.log('ERROR', user, e.message);
+
+// Thanks https://stackoverflow.com/a/40672956
+String.prototype.splitCSV = function() {
```
garethbowen: :warning: Don't update the prototype... ever.

https://docs.communityhealthtoolkit.org/contribute/code/style-guide/#do-not-extend-built-in-prototypes

It's just as easy to call a regular named function.

scripts/bulk-password-update-export.js
```diff
@@ -106,30 +127,49 @@ const loadUsers = async () => {
...
+  let matches = this.match(/(\s*"[^"]+"\s*|\s*[^,]+|,)(?=,|$)/g);
+  for (let n = 0; n < matches.length; ++n) {
+    matches[n] = matches[n].trim();
+    if (matches[n] === ',') matches[n] = '';
+  }
+  return matches;
+}
```
garethbowen, line 139: CSVs are notoriously hard to parse so let's not write a bespoke parser when we have the whole of npm at our fingertips. In fact, the `load_messages` script already uses a csv parser so just use that one!

scripts/bulk-password-update-export.js
```diff
@@ -67,6 +69,25 @@ Do you want to continue? [y/N]
...
+  const admins = ['admin', 'medic'];
+  const adminError = new Error('403 - Cannot change password for "medic" or "admin" users.');
+  const postOptions = {...options};
+  postOptions.body = {
+    'password': password
+  };
+  postOptions.uri = `${options.uri}/${user}`;
+  try {
+    if (admins.includes(user)) {
+      throw adminError;
```
garethbowen, line 82: Hmmm interesting. I think because you've created this in line 74 the stacktrace will show line 74 and not 82. I would just inline it here so it's created only if/when necessary (which feels like a tiny performance win too!), and also means you can include the specific bad username in the error message, eg:

```suggestion
      throw new Error(`403 - Cannot change password for administrator user "${user}"`);
```

scripts/bulk-password-update-export.js
```diff
@@ -67,6 +69,25 @@ Do you want to continue? [y/N]
   }));
 };
 
+const changeUserPass = async (user, password, options) => {
+  const admins = ['admin', 'medic'];
```
garethbowen, line 73: I would pull this out as a top level constant. There's no need to init the same array over and over again!

garethbowen (changes requested):

scripts/bulk-password-update-export.js
garethbowen: Hah! I see sonar bet me to it!

scripts/bulk-password-update-export.js
billwambua: [Papaparse](https://www.papaparse.com/docs) is good for this too.

scripts/bulk-password-update-export.js
mrjones-plip: fixed in f0c5b2171 and c237a2f66

scripts/bulk-password-update-export.js
mrjones-plip: @garethbowen  FTW in 63025e3!

garethbowen (approved):

scripts/bulk-password-update-export.js
```diff
@@ -1,29 +1,33 @@
...
+const csvSync = require('csv-parse/sync');
 
 const argv = minimist(process.argv.slice(2), {
   alias: {
     'h': 'help'
   }
 });
 
-if (argv.h|| !argv.url || !argv.user || !argv.password) {
-  console.log(`Change passwords of users in user-password-change.txt to random value, show results on screen
```
billwambua, line 16: `user-password-change.txt` should be `user-password-change.csv`

scripts/bulk-password-update-export.js
```diff
@@ -95,40 +116,42 @@ const generatePassword = async () => {
...
+  }
+  const users = await loadUsers(use_passes);
+  const input = await areYouSure(users.length, extraWarning);
   if (input.toLowerCase() !== 'y') {
     process.exit(0);
   }
 
   const postOptions = {...options};
   for (const user of users) {
+    if (!use_passes) {
```
billwambua, line 152: You should check whether `user` is empty.

Sometimes a csv might have an empty row in the middle of the sheet.

When that happens the script still does the call for password change only to get the following error back from CouchDB.
```
ERROR  404 - {"error":"not_found","reason":"Database does not exist."}
```

It doesn't cause a crash but printing to screen a message like `Error: CSV row is empty` would be more intuitive.

billwambua (changes requested):

scripts/bulk-password-update-export.js
mrjones-plip: good call! fixed in 0bc2eba85

scripts/bulk-password-update-export.js
mrjones-plip: gotcha! fixed in 2c6cba7a0 . 

We now return a more clear `404`:

```
ERROR  404 - Username is blank - check CSV and run again.
```

Issue #6363, opened 2020-04-18
Prevent and/or merge duplicate contacts


ecsalomon:
**Is your feature request related to a problem? Please describe.**
[This excellent report](https://docs.google.com/document/d/1TibEbmNiZPAr9PKHiBw0DTZn-KO2XHtOb6Q702caEBM/) by @marialma shows that duplicate contacts are an issue for a non-trivial amount of CHWs in some deployments. Maria has enumerated a large number of issues that arise for care delivery, privacy, and analytics that arise from this.

**Describe the solution you'd like**
There are several potential solutions:

1. When entering a new contact, search existing contacts for exact or extremely close (say, Levenshtein distance = 2) names and ask the CHW to confirm that they want to create a new contact.
2. Allow CHWs to "move" contacts between parent contacts (e.g., when someone moves to a new household within a catchment area)
3. For older deployments, let's say a data scientist sets up an entity matching service on the backend to identify likely duplicates, allow a CHW/supervisor to review and confirm duplicates and merge them (so that all documents become associated with the merged/new contact) and deconflict contact info (e.g., date of birth).

**Describe alternatives you've considered**
There are likely some training solutions (e.g., advising CHWs of the problems created by duplicate contacts). But training will never address everything.

**Additional context**

ecsalomon:
See also #6309

jkuester:
Additional conversation about the need for this feature [on this forum thread](https://forum.communityhealthtoolkit.org/t/mitigate-duplicate-data-capture/3313/3)!

ChinHairSaintClair:
We've opted to use the `descendant-of-current-contact` feature, although it doesn't use Levenshtein distance, hoping it will help mitigate duplicate captures.

We've also been looking into using the `cht-conf` tool's `contact-move` functionality to consolidate records where possible - as a method of "damage control."

After reviewing the death reporting functionality provided by CHT, we wondered if it is possible to create a custom transition appending some flagging property to a record if its immediate siblings are similarly named. Then, a task would examine this property and flag entries for correction. As noted by @jkuester, this will require some core changes and the CHWs' devices to be online and successfully synced. Additionally, initiating a contact edit form directly via the task will not load the required edit info. This will necessitate an ephemeral form that redirects to the edit form.

Originally mentioned on [this](https://forum.communityhealthtoolkit.org/t/write-custom-transition-and-edit-contact-form-with-task/3334?u=anro) CHT forum thread.

dianabarsan:
[...]

PR #8878, opened 2024-02-16
patch: descendant of current contact

ChinHairSaintClair:
# Description

Similar to ecsalomon's github issue rationale, mentioned below, we've been seeing an increased amount of CHWs creating duplicate contacts.
This is a multi faceted problem, which has spawned quite a bit of conversation, as can be seen by visiting the forum thread linked below.

Backport of a CHT [feature](https://github.com/medic/cht-core/pull/8759/files) originally developed by [latin-panda](https://github.com/latin-panda).
Related to [this](https://forum.communityhealthtoolkit.org/t/mitigate-duplicate-data-capture/3313?u=anro) forum thread attempting to combat duplicate data capturing.
Which in turn is related to the [original duplicate issue](https://github.com/medic/cht-core/issues/6363) raised by [ecsalomon](https://github.com/ecsalomon).

# License

### DIFF
--- a/shared-libs/search/src/generate-search-requests.js
+++ b/shared-libs/search/src/generate-search-requests.js
@@ -73,7 +73,8 @@ const formRequest = function(filters) {
       return forms.map(function(form) {
         return [ form.code ];
       });
-    });
+    }
+  );
 
   if (req) {
     req.params.reduce = false;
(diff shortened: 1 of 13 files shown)


garethbowen:
@ChinHairSaintClair It's usually problematic to backport features, because when you get around to releasing the only way to do it is to do it is to use a patch in semver which is defined as "PATCH version when you make backward compatible bug fixes". This is done on purpose so people upgrading from 4.2.0 to 4.2.10 know that all changes are backwards compatible.

The ideal solution is you upgrade to the latest to get this and all other fixes and improvements. Is there anything blocking you from upgrading to the latest?

garethbowen:
@ChinHairSaintClair If you're running grunt tasks to build the CHT then it sounds like you're already on a complete fork of the CHT. If that's the case then you should be free to apply this PR to your fork, and it's not necessary to merge it into the `4.2.x` branch of the `medic` fork, right?

You've clearly done a lot of work separately which is making it difficult to keep up with the `medic` fork. I would need to know a lot more about your situation to understand how you're using things like tombstones to recommend a way forward. Are these changes that should be contributed upstream for the benefit of the whole community? Are these things that could be supported by supported endpoints and plugins so you don't need a fork?

garethbowen:
@ChinHairSaintClair Sounds great. Feel free to email me at gareth@medic.org and we can set up a call!

Issue #8308, opened 2023-06-08
Add support for Signature Widget in forms


jkuester:
**Is your feature request related to a problem? Please describe.**
As noted on [this forum thread](https://forum.communityhealthtoolkit.org/t/signature-drawing-widget/2176/3) there is a need to collect signatures within Enketo forms.

**Describe the solution you'd like**
An [Enketo Widget](https://docs.getodk.org/form-question-types/#signature-widget) already exists for doing this, but is not one of our [currently enabled widgets](https://github.com/medic/cht-core/blob/master/webapp/src/js/enketo/widgets.js). During the uplift to Enketo 5.x, I did briefly test that widget, but it was ultimately not included. At the time we did not have any usecases where it was needed and so there was not much to justify the increase to the bundle size for the webapp (the drawing libs associated with the widget are not small),

**Describe alternatives you've considered**
We could roll our own drawing widget. This might result in a more lightweight solution, but would be extra code to maintain.  

Another alternative would be to see if the use-cases for the signature widget could be covered by just using the [trigger widget](https://docs.getodk.org/form-question-types/#trigger-acknowledge-widget) (I know a lot of "official" online forms that let me "sign" just by checking a box...).

jkuester:
I have started a branch containing changes for this issue: https://github.com/medic/cht-core/tree/8308_signature_widget

Currently all I did was add the linkage for the [Enketo widget](https://github.com/enketo/enketo-core/blob/master/src/widget/draw/draw-widget.js) so the widget can be used in a form.  Here are the issues that i have noted so far:

- [ ] Something is weird with the widget when it is initially loaded. Drawing works, but the displayed lines do not match with where I am drawing. Reloading the widget fixes this problem.  Not sure exactly what is causing this.
- [ ] Using the widget causes the form to fail to submit.  This is due to an error [when trying to save the attached image](https://github.com/medic/cht-core/blob/0bc141571e6999b969f30d06e929b0766ab9b3a5/webapp/src/ts/services/enketo.service.ts#L662).  I think that code is setup to handle saving uploaded images, but needs to be tweaked to support saving drawn images as well!

jkuester:
@garethbowen do you have any concerns with including this widget (once we get the functional issues resolved)? I know we originally excluded it due to bundle size, but based on the ongoing forum thread linked above, there is community interest in having this widget and seems worth including in the bundle....

latin-panda:
[...]

Issue #8072, opened 2023-02-13
Error when saving form with repeated upload inputs


dianabarsan:
**Describe the bug**
When a form with repeated uploads is submitted, an error is triggered and the report is not saved.

**To Reproduce**
1. Launch 4.x
2. Upload this form: https://gist.github.com/dianabarsan/3696d99ba17fc2ce462d58149618b93b
3. Open the form and add more than one repeat. Save.

**Expected behavior**
The form should be saved.

**Logs**
```
feedback.service.ts:79 Error submitting form data:  TypeError: Cannot read properties of undefined (reading 'files')
    at Element.<anonymous> (enketo.service.ts:616:32)
    at Function.each (jquery.js:381:19)
    at jQuery.fn.init.each (jquery.js:203:17)
    at EnketoService.xmlToDocs (enketo.service.ts:613:8)
    at enketo.service.ts:796:21
    at _ZoneDelegate.invoke (zone.js:372:26)
    at Zone.run (zone.js:134:43)
    at zone.js:1275:36
    at _ZoneDelegate.invokeTask (zone.js:406:31)
    at Zone.runTask (zone.js:178:47)
```

Error triggered here: https://github.com/medic/cht-core/blob/26f3ae818f4d2792eb375e4d65ca6e15ecdec347/webapp/src/ts/services/enketo.service.ts#L616

**Screenshots**
![image](https://user-images.githubusercontent.com/35681649/218465185-15dada0b-9c61-4f51-b974-2f3e4805bb8e.png)

**Environment**
- App: webapp
- Version: 4.0+

garethbowen:
The problem is because the `Xpath.getElementXPath(element);` returns the path with the array position, eg: `/upload_repeat/repeat[1]/my_photo` _but_ the name of the input does not, eg: `/upload_repeat/repeat/my_photo`. This means when we look up the file input it's not found.

There's not obvious way to get to the html input.

garethbowen:
@jkuester I'd appreciate your input on [this potential fix](https://github.com/medic/cht-core/commit/ff16e276c83299814c4c77e81cf2086f079a2a4c) when you have a moment. This approach is nice because it uses enketos own file manager to get the files. However it's not ideal because it changes the file format which means supporting multiple formats (legacy and new) forever. Also I don't think we can use the same approach for the `binary` type because it doesn't have a file name so file upload and binary will not be consistent.

Obviously this needs a lot of cleanup, testing, etc, but I wanted to run it past you first.

[...]

Issue #8865, opened 2024-02-13
Split up webapp JS into multiple smaller bundles


garethbowen:
**Describe the performance issue**

When upgrading the service worker refreshes the asset cache, which tells the browser to download all JS. The browser then checks its cache and only downloads resources that have changed. With all the JS bundled into one module, the browser always downloads the entire JS again. This discourages frequent upgrading and can incur a cost on the CHW or program due to downloading a large JS bundle. Furthermore if the download is interrupted it has to redownload the entire JS bundle. Bundling all JS together was best practice 10 years ago before improvements in the http spec (ie: multiplexing in http2).

**Describe the improvement you'd like**

Use Angular modules to split the one large bundle into a series of small bundles. These bundles should be defined logically with coupled code being together in the same module because this code is likely to be updated in the same release.

There will likely be a tradeoff between having too few modules which cause excessive downloads, and having too many causing many requests and overall complexity.

**Describe alternatives you've considered**

Drastically reducing the size of the JS, but previous audits haven't shown any low hanging fruit here.

**To Reproduce**

1. Clear your service worker cache
2. Open the network tab in the dev console
3. Load the login page (no login necessary)
4. Record the number and size of the JS assets downloaded
5. Record the time it takes to download all the assets
6. Upgrade the CHT (In development, just modify one JS file. In a production like environment do the smallest possible upgrade.)
7. Wait for the service worker to realise there's a new version
8. Record the number and size of the JS assets updated

**Measurements**

- Compare the results in step 4 and step 8 to verify that only the module that changed has to be downloaded
- Compare the results in step 5 with `master` to make sure there's no significant change in load time
- Evaluate the bundle sizes in step 4. Ideally there would be 5-10 bundles of approximately equal size, but it's not completely under our control.

Issue #8755, opened 2023-12-11
Update to latest enketo-core


jkuester:
Now that we are done with #7599, it seems like a good idea to start preparing for updating to the next Enekto version!  :laughing: 

As @m5r pointed out [on Slack](https://medic.slack.com/archives/C02FMC810UE/p1700518578709929), the enketo-core repo has been moved into the big Enketo monorepo: https://github.com/enketo/enketo/blob/main/packages/enketo-core and version `8.0.0` has been released:

> - Breaking change: in the move to a monorepo, some paths in Sass styles have been updated to reflect the new file structure. The impact should be minimal, but custom themes may need to be updated. Generally, paths which cross package boundaries (i.e. imports from Enketo Express to styles in Enketo Core, or imports of third party styles from NPM packages) should use absolute paths from the monorepo root. You may refer to existing theme styles for examples.
> - Breaking change: Support Node 18 and 20, drop support for older versions

It may make sense to try to address this issue at the same time: https://github.com/medic/cht-core/issues/8741

jkuester:
Just noting that the latest version of Enketo (`9.0.0`) contains fixes for the draw widget (so we would not need our custom widget any more).

It also include depends on jquery `3.7+`.  When we upgrade to the new enketo-core version, we should also try lifting our jquery version (making sure https://github.com/select2/select2/issues/5993 does not still present).

jkuester:
When this issue is addressed, we need to remove the dependencies tagged with this issue from the [dependabot ignore list](https://github.com/medic/cht-core/pull/9797).

jkuester:
A [new version](https://github.com/enketo/enketo/releases/tag/7.6.2) of Enketo has dropped.  We should upgrade to this because of https://github.com/medic/cht-conf/issues/797.

jkuester:
Very exciting to see [dynamic read_only expressions](https://github.com/enketo/enketo/pull/1588) coming to an Enekto version near you!

PR #8904, opened 2024-02-21
feat: add support for Enekto draw widget in forms

ChinHairSaintClair:
# Description

Closes https://github.com/medic/cht-core/issues/8308
Closes https://github.com/medic/cht-core/issues/8072

Enables the draw widget, which allows CHT to capture free hand drawings through touch or mouse pointer.
image

# License


garethbowen:
@jkuester I'd love your input here. This feels related to #8308 . Maybe we need to think about how to optionally include specific widgets and/or bundle them separately?

jkuester:
:+1: Yes, I believe that draw/signature is backed by the same libraries (maybe even the same Enekto widget... :thinking: ).  

I will try to carve out some time in the next week to think about optional widgets.  On a related note, @garethbowen I see that the bundle size limits were [removed with Grunt](https://github.com/medic/cht-core/commit/51926eba4171d15fd8da5982402753b2f23896f8#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L147).  Were these replaced with something else or do we have anything right now that is checking the size of the bundle?

garethbowen:
> Were these replaced with something else or do we have anything right now that is checking the size of the bundle?

No they weren't replaced. In reality we weren't blocking on that anyway. It's not a practical solution because every commit gets you closer and closer to the limit until finally one unlucky soul hit it. So every time we breached it we just arbitrarily bumped it until we disabled it anyway.

IMO this is a better solution: https://github.com/medic/cht-core/issues/8865

jkuester:
Have not forgotten about this.  Planning to do some investigating today or tomorrow! Will follow up out here with any findings.

jkuester:
@garethbowen I did a deep dive into the options for modularlizing/lazy-loading widgets and included my findings [in the issue](https://github.com/medic/cht-core/issues/8308#issuecomment-2118206891).  The TLDR with regards to _this PR_, though is: 

- I could not find any low-hanging-fruit that would make it easy to pick and choose widgets. There are a number of options, but they would all take a decent amount of effort to pull off.
- The cost of just merging this PR, as is, will be `+9.28 KB` to the Gzipped bundle size (total webapp Gzipped size going from `2.64 MB` to `2.65 MB`)

Let me know how you want to proceed!

garethbowen:
That size change should not hold up this widget.

jkuester:
:+1: Sounds good!  My plan this week, then, is to do a proper review of these code changes and either suggest or just push some unit tests and some integration/cht-form tests.  

@lorerod just wanted to get you engaged at the start here in case you had any thoughts/concerns going into this. I am thinking we will want at least 1 happy-path e2e test to make sure we cover saving the drawing image _as an attachment in Couch_, but besides that, I will try to cover other test cases in the cht-form integration tests...

webapp/src/ts/services/enketo.service.ts
```diff
@@ -473,12 +473,49 @@ export class EnketoService {
...
         const xpath = Xpath.getElementXPath(element);
         const $input: any = $('input[type=file][name="' + xpath + '"]');
-        const file = $input[0].files[0];
+        const inputElement = $input[0];
+        let file: File | undefined | null;
+
+        if (inputElement?.files?.length > 0 && inputElement.files[0]) {
+          file = inputElement.files[0];
+        } else {
+          file = convertDrawDataToFile(xpath);
```
garethbowen, line 482: I haven't really looked into this at all but something reminded me of [a preexisting bug in this code](https://github.com/medic/cht-core/issues/8072). It might be worth looking into the `FileManager.getCurrentFiles();` function instead of using xpath.

jkuester: For future reference, after quite a bit of tinkering, the only way I found to leverage the functionality in FileManager while maintaining backwards compatibility with the way that we currently do attachment names is by making a small change to the FileManager code (this could be done via a patch or by coping+editing the FileManager code and supplying a custom `enketo/file-manager` value. Either way, we just need to expose the `name` value associated with each of the elements that has a file.  For example, we could set `file.xpath = this.name;` before pushing the file into the array in `getCurrentFiles`.  

Then, that would let us handle all file attachments here in the enketo service with code like this:

```ts
    const files = FileManager.getCurrentFiles();
    for (const file of files) {
      // replace instance root element node name with form internal ID
      const filename = 'user-file' +
        (file.xpath.startsWith('/' + doc.form) ? file.xpath : file.xpath.replace(/^\/[^/]+/, '/' + doc.form));
      this.attachmentService.add(doc, filename, file, file.type, !file.type);
    }
```

jkuester:
Bad news.  The `draw` widget implementation in our current Enketo version (`^7.2.0`) apparently has serious issues and it was reverted in was reverted in `8.1.0`. I believe we will have to block this PR until after we upgrade our Enketo version (but @garethbowen let me know if you agree!).  Read on for more details:

While testing this code, I uncovered an issue where, if the signature is not on the last page (but on one of the earlier pages) and the signature widget is still visible on the page when the "Next" button is hit, the image will not be saved for that signature.  This bug is not a result of the code changes in this PR, but I thought it might be some kind of integration issue with how we are manipulating eneketo-core.  However, after extensive debugging over the last two days, I began to wonder if this was a bug in the actual Enekto widget code. 

This suspicion led me to a review of the latest enketo-core source code where I realized our unlucky timing here...  In enketo-core `7.2.0`, they implemented [a significant shift in the draw widget implementation](https://github.com/enketo/enketo-core/pull/960).  Several issues were identified with this functionality (though I did not find anything logged that reflected the exact behavior I was observing).  Eventually, the decision was made to [completely revert the earlier changes](https://github.com/enketo/enketo/pull/1292) and this was released in enketo-core `8.1.0`.  I did try testing the draw widget from enketo-core `7.1.0` and could not reproduce the bug that I had observed before. My current thinking is that we should hold off on adding this functionality until after we uplift our enketo version in https://github.com/medic/cht-core/issues/8755.

---

For the record, the bug that I was tracking in the implementation of the draw widget from `7.2.0` to `8.0.0` was in the [ResizableSignaturePad.js code](https://github.com/enketo/enketo/blob/enketo-core/8.0.0/packages/enketo-core/src/widget/draw/ResizableSignaturePad.js).  Basically, there are two observers.  The `resizeObserver` listens for changes to the size of the `canvas` entities and then re-draws things accordingly.  The `intersectionObserver` listens for changes in the visibility of the `canvas` entities (e.g. them appearing on a new page or just scrolling in/out of view) and adds/removes the canvases from the `resizeObserver`'s tracking list as they become visible/hidden.  The bugged behavior happens when clicking the "Next" button on a page where the draw widget is still visible (and so the `resizeObserver` is still tracking the canvas).  When the transition to the next page occurs, the current page gets hidden and this causes the `offsetWidth`/`offsetHeight` of the canvas to be set to `0`. I believe what follows is technically a race condition, but I never saw the outcome change. Hiding the current page triggers _both_ the  `intersectionObserver` and the `resizeObserver`, but the `resizeObserver` always wins and executes first. This is a problem because the `intersectionObserver` is supposed to prevent the `resizeObserver` from ever getting executed for a canvas that is not visible.  Calling the  `resizeObserver` with the `offsetWidth`/`offsetHeight` of `0` triggers an unintended flow that ultimately results in the canvas height getting set to `0`. This, in turn, means that when you call `toDataURL()` on the canvas, if [will return `"data:,"`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toDataURL#return_value) instead of the actual base64 canvas data. (And this is why the bugged flow ends with no image attachment getting saved.)

One potential fix I had identified was to update the ResizableSignaturePad code to add a check [in the `resizeObserver`](https://github.com/enketo/enketo/blob/0916592d7e48db3b4f39a541cc35e38d7467edcb/packages/enketo-core/src/widget/draw/ResizableSignaturePad.js#L242) so that it would confirm `pad.isVisible` was true before proceeding to call `pad.redraw()`.

webapp/src/ts/services/enketo.service.ts
garethbowen: I don't believe we need to change the file manager but we would need to handle old and new ways of naming attachments. This branch has my attempts from earlier in the year: https://github.com/medic/cht-core/compare/master...8072-repeat-field-inputs

It basically works but needs complete testing on device and unfortunately got deprioritised.

ChinHairSaintClair:
@jkuester  Am I correct in assuming that CHT version `4.3.x` also uses Enketo `7.2.0`, or is it still on `7.1.0`? 
The reason I ask is that we're currently using this drawing widget in our deployment, and it seems to be functioning as intended.

It would definitely impact our decision/haste in upgrading to newer CHT versions. 
Please see [this](https://docs.google.com/spreadsheets/d/1jh22HDi8xyUOKRpbfOudoIoZ7zcXmfrN/edit?usp=drive_link&ouid=116686630670829159353&rtpof=true&sd=true) form, where our drawing widget is on the 2nd of 8 pages. 
Below, the signature can be seen showing correctly in the report view:
image
The attachment can also viewed by using the below URL:
http://localhost:5984/medic/0b5ec2c2-cded-4375-b09b-e231d1a9002f/user-file/csharp-householdconsentandquestionnaire/hhsignatures/consent_hhrespondentsignature

jkuester:
@ChinHairSaintClair good question!  We [upgraded to Enekto `^7.2.0`](https://github.com/medic/cht-core/commit/314e79061acfa3447fce1a05a784a928a0cf4459) in CHT `4.6.0`. So, CHT `4.3.x` would _not_ include the version of Enekto with the bugged Draw widget.

That being said, your question has made me reconsider our options here!  If we look at the Enketo PR that [reverted the Draw widget changes](https://github.com/enketo/enketo/pull/1292) for Enketo `8.1.0`, we see that the actual code changes were completely isolated to the two files in the `widget/draw` directory and none of the actual core/util code in enketo-core was affected.  This means that it should be trivial for us to copy the working draw widget code from Enekto `8.1.0` into cht-core as a "custom widget" that we use along side our current Enekto `^7.2.0`.  Then, in the future, when we do complete the upgrade to Enekto `8.1.0+` it will be completely passive to remove the custom draw widget code and switch to depending on the Draw widget provided by Enketo.  :thinking: 

@garethbowen let me know if you are open to pursuing that approach (or should we just double-down on efforts to complete the Enketo uplift?).

mrjones-plip:
@jkuester - per our [private chat in slack](https://medic.slack.com/archives/C02H68VMD/p1717082414995929?thread_ts=1717023584.357469&cid=C02H68VMD): let's go ahead with this approach:

>  trivial for us to copy the working draw widget code from Enekto 8.1.0 into cht-core as a "custom widget" 

We can be sure to undo any patchwork we do when we get around to the [Enketo uplift](https://github.com/medic/cht-core/issues/8755)

webapp/src/ts/services/enketo.service.ts
garethbowen: > a change to the attachment name scheme is not a breaking change (as long as the app still functions as it did before)?

Yes I think so. When you say "the app still functions as it did before" we need to still render old reports correctly, so the rendering has to support both naming schemes indefinitely. I think that's ok as we work towards consistency.

I don't think it's a breaking change because I doubt anyone is relying on the file name, but I might be wrong. The problem is to support repeat files we _have_ to change the naming because otherwise they collide. One alternative is to only use the different naming scheme with repeats, but that's complicated and means we can't use enketo's filemanager.

On balance, I think we're ok to change the file name.

jkuester: Could/should we add a DB migration to go back and rename the existing attachments in the DB?  (Not sure exactly how we might balance the migration effort vs having to support both naming schemas indefinitely...)  It seems like it would be possible to write a migration, but it might trigger a lot of syncing....

garethbowen: Yeah we basically don't do mass migrations any more. This is another example of something that could be hidden behind a webapp api so the Report type is standard, and old docs are converted on read so the complexity is buried deep down and not in the UI layer...

webapp/src/js/enketo/lib/window.js
```diff
@@ -0,0 +1,5 @@
+const getCurrentHref = () => window.location.href;
+
+module.exports = {
+  getCurrentHref,
```
jkuester, line 4: Need this to provide an abstraction point for `file-manager` unit test to allow them to mock out the href. (Unless someone has a better way to do this...)

IMHO, this is annoying, but better to just include this vs wasting a ton of time (well even more time since I already wasted a ton here) trying to find a more creative solution....

webapp/tests/karma/ts/modules/reports/reports-add.component.spec.ts
```diff
@@ -275,13 +277,272 @@ describe('Reports Add Component', () => {
...
         expect(component.form).to.equal(undefined);
-        expect(consoleErrorMock.callCount).to.equal(1);
-        expect(consoleErrorMock.args[0][0]).to.equal('Error loading form.');
+        expect(error.callCount).to.equal(1);
+        expect(error.args[0][0]).to.equal('Error loading form.');
       }));
     });
 
     describe('for existent reports', () => {
-      // todo add tests here when migrating the final code about editing reports
```
jkuester, line 284: image

webapp/tests/mocha/unit/enketo/lib/window.spec.js
```diff
@@ -0,0 +1,18 @@
...
+  before(() => originalWindow = global.window);
+  after(() => global.window = originalWindow);
+
+  it('getCurrentHref', () => {
+    global.window = {
+      location: {
+        href: 'http://example.com'
+      }
+    };
+    expect(windowLib.getCurrentHref()).to.equal(global.window.location.href);
```
jkuester, line 16: image

jkuester:
Okay, I am going to try and clarify/recap what is going on in this PR! :sweat_smile: The main goal is to address https://github.com/medic/cht-core/issues/8308 so we can support collecting [signatures](https://docs.getodk.org/form-question-types/#signature-widget) in forms.  (The enketo `draw` widget also supports collecting [general drawings](https://docs.getodk.org/form-question-types/#draw-widget) as well as [`annotation`s](https://docs.getodk.org/form-question-types/#image-widget-with-annotation) where you upload an image and then draw on it.  We get these for "free" along with the signature support.)  Unfortunately, as noted above, the `draw` widget for our current version of enekto-core is bugged, so we have copied the working  version of the widget into our codebase (and will remove it when we upgrade to the latest version of enekto-core).  

The `draw` widget handles all the client-facing/display logic for us, but the cht-core code still needs to handle (in `enketo.service.ts`) taking the signature images saved in the form and storing them as attachments on the Couch document for the report.  Since we were going to need to add a bunch of additional logic here to support attaching the drawing images, we decided instead to take this opportunity to switch to using the Enketo `file-manager` (or really, our custom _extension_ of the file-manager) to load the files from the form. The downside of this is that it means we had to change the naming pattern that we were previously using for Couch doc attachments (which, in turn, meant we had to update the code in a couple places to account for the new naming scheme).  The upside of this is that it fixes https://github.com/medic/cht-core/issues/8072 and we can now support attaching files within `repeat`s!

---

@ChinHairSaintClair I think this is finally ready to go!  When you get the chance, can you pull down the latest version of this branch and try it out to make sure everything still works for you?  

@Benmuiruri I am hoping you might have some time next week to do a code review of this! :pray:   I am not sure how much experience you have had with the Enekto stuff so far, but if you get totally confused by looking at this, let me know and we can jump on a call to discuss! (Our "integration" with the enekto-core library is very _involved_ (read: "tightly coupled").  These changes are not making it much better, but they are not making it too much worse either.  :grimacing: 

@lorerod the testing here got a bit interesting and I would very much appreciate your feedback on it!  In the cht-form integration tests, I added `draw-widget.wdio.spec.ts` which tests a form that has all three types of draw widget appearances (`draw`, `signature`, and `annotate`).  I did manage to get wdio to "draw" a shape on the canvases, but I did not go so far as to actually assert the image contents. Instead, I went with asserting the `size` of the image as a simplification that would at least allow us to confirm that something non-trivial was saved for those questions.  Then, I added `file-upload.wdio-spec.js` with a small form for testing adding image file uploads from within a `repeat` group.  

I thought about adding either of these workflows to the existing `enketo_widgets`  tests either in the cht-form integration tests or the default e2e tests.  However, my current thinking is that we are probably better off, from a long-term maintainability perspective, if we stick with smaller more targeted forms/tests instead of cramming everything in a big catch-all form.  Interested on your thoughts here!  Also, you will note that the `submit-photo-upload-form.wdio-spec.js` file is pretty similar to `file-upload.wdio-spec.js`.  I thought about just adding a repeat to that test, but I did not really want the extra complexity in the e2e tests (instead of in the cht-form integration tests).  Then I thought about just migrating those tests to the  `file-upload.wdio-spec.js`, but ultimately thought it best to leave them in the e2e tests because the `submit-photo-upload-form.wdio-spec.js` is currently covering these workflows (that could not be adequately testing in the integration tests):

- Ensure attached image is _displayed on the reports page_ after the form is submitted
- Ensure that a form with image attachment _can be edited_ and a new image attached (or not)

webapp/src/css/enketo/medic.less
```diff
@@ -60,14 +60,18 @@
...
+  .btn-icon-only:not(.draw-widget__undo) {
     padding: 0;
     width: inherit;
   }
-  .icon {
+  .icon:not(.draw-widget__undo .icon-undo) {
     position: relative;
     top: 1px;
   }
+  // Hide download button for draw widget
```
jkuester, line 71: Hiding the download button (lets you download a copy of your drawing before submitting the form). It did not work for some reason on Android (maybe a permissions problem?) and seemed like extra clutter to include on the page unless we have a use-case where we need it....

api/resources/translations/messages-en.properties
```diff
@@ -638,6 +638,9 @@ email.invalid = Invalid email address.
 empty = It looks like you sent an empty message, please try to resend. If you continue to have this problem please contact your supervisor.
 enketo.constraint.invalid = Value not allowed
 enketo.constraint.required = This field is required
+enketo.drawwidget.signature = signature
+enketo.drawwidget.drawing = drawing
+enketo.drawwidget.annotation = annotation
```
Benmuiruri, line 643: I suggest reaching out to fellow teammates to get Spanish, French and Nepali translations so that we keep the translations in sync. 

I can provide the Swahili translations.

ChinHairSaintClair: The French translations from a French speaking colleague:
Signature (remains the same)
Drawing -> Dessiner
Annotation (remains the same)

Benmuiruri: Swahili translations 
signature: sahihi
drawing: uchoraji
annotation: maelezo

jkuester:
FYI:

- The failing `k3d` test suits are a [known issue](https://github.com/medic/cht-core/issues/9097) caused by this PR originating in an external branch.  I tried pushing the same code into a branch in `medic/cht-core` and the tests pass just fine. For the purposes of this PR we can disregard those test failures.
- I have reached out for additional translations and will include them all in this PR once I get the remaining ones
- @Benmuiruri for manually testing the draw widgets, you can use the `enketo_widgets` form [from this support-scripts PR](https://github.com/medic/support-scripts/pull/14)

api/resources/translations/messages-en.properties
derickl: drawing could also be kuchora @Benmuiruri

lorerod (approved):
@jkuester Thank you for sharing the details of the testing approach. It sounds well thought out!
Regarding the `draw-widget.wdio.spec.ts` test, asserting the image size is a clever simplification. It ensures something substantial is saved and avoids overcomplicating the tests.
Keeping tests focused will make them easier to manage and less prone to breaking with future changes. Regarding the similarity between `submit-photo-upload-form.wdio-spec.js` and `file-upload.wdio-spec.js`, I agree with the justification behind keeping them separate to  keep the tests modular and focused on their specific workflows and avoid adding unnecessary complexity to the e2e tests.
I also manually tested the enketo draw widget. It works! Here is a screenshot of the report, and extract from the saved doc:
image
```
"_attachments": {
    "user-file-drawing-15_5_28.png": {
      "content_type": "image/png",
      "revpos": 1,
      "digest": "md5-X49lAzhPmcGO+8DwomyxSA==",
      "length": 20381,
      "stub": true
    }
  }
```

ChinHairSaintClair:
@jkuester Seems to be working fine on my side too 😃!
Thank you for capturing the investigation and dev journey in such detail.
image
```
"_attachments": {
    "user-file-signature-12_15_38.png": {
      "content_type": "image/png",
      "revpos": 1,
      "digest": "md5-8FXoQ/m15sQB93wznbFYrQ==",
      "length": 7645,
      "stub": true
    }
  }
```

ChinHairSaintClair:
Noticed that the mechanism of viewing the the attachment via a URL no longer seems to work.
URL structure: `<domain>/<database>/<record_id>/user-file/<form_name>/<group_name>/<property>`

On old instance:
`http://localhost:5984/medic/f925d4aa-b4d9-4edf-8c07-cf87e29f8afd/user-file/csharp-householdconsentandquestionnaire/hhsignatures/consent_hhrespondentsignature`
image
image

New instance:
`http://localhost:5984/medic/09625306-6a8a-4af0-906d-508a10cdc6af/user-file/consent/death_details/signature`
image
image

Issue #8868, opened 2024-02-13
Session requests failing after upgrade


jkuester:
**Describe the bug**
Upgrading to `4.6.0-beta.2` somehow seems to break the session/authentication handling in the api server!  You can no longer login to the webapp. Also, even direct REST calls to the api with basic auth fail due to a `400` error.  (Though calls to open endpoints like `api/v2/monitoring` are successful.)  

**To Reproduce**
1. Deploy a CHT instance (e.g. via EKS or docker-helper).
    1. Of course I cannot repro this locally when just building from the `4.6.x` branch....
3. Upgrade it to `4.6.0-beta.2` via the admin app.
    1. If you `Stage` first and then `Install`, the upgrade will complete without any errors being displayed on the admin page.  If you just directly `Install` without staging, then an error will display on the upgrade page, but the upgrade will actually complete successfully.
4. Logout
5. Try to log back in and see error
6. Try directly calling an authenticated endpoint (e.g. `api/v2.upgrade`) and see error
    1. The response I get back is:
        ```
        400 - "<html><body><h1>400 Bad request</h1>\nYour browser sent an invalid request.\n</body></html>\n"
        ```

**Expected behavior**
Upgrading should not break the deployment.  If it does, you are not going to space today....

**Logs**
Here are the logs directly after making 5 back-to-back calls to the `api/v2/upgrade` endpoint (and getting the above 400 error every time)

### api

```
2024-02-13T20:15:29.563 REQ: 1f694f26-1dbf-4b33-b631-776a8a98bb67 216.147.121.215 medic GET /api/v2/upgrade HTTP/1.1
2024-02-13T20:15:29.572 RES: 1f694f26-1dbf-4b33-b631-776a8a98bb67 216.147.121.215 medic GET /api/v2/upgrade HTTP/1.1 400 - 8.494 ms
2024-02-13T20:15:30.355 REQ: d98c78b4-784f-450e-83c3-252e802f0db6 216.147.121.215 medic GET /api/v2/upgrade HTTP/1.1
2024-02-13T20:15:30.362 RES: d98c78b4-784f-450e-83c3-252e802f0db6 216.147.121.215 medic GET /api/v2/upgrade HTTP/1.1 400 - 5.724 ms
2024-02-13T20:15:31.315 REQ: 084cd065-f8f9-48cb-9686-852d51cb9c6d 216.147.121.215 medic GET /api/v2/upgrade HTTP/1.1
[...]

PR #8924, opened 2024-03-12
fix(#8868): don't forward all headers for get requests

dianabarsan:
# Description

Node 19 adds keep-alive header to requests by default. This makes the `content-length` header very significant **and** important because it enables the server to know when one request has been fully received and another one is sent through the same connection. 
When authenticating a user, we take all headers from the original user's request and pass them to a GET /_session request, presumably so we don't have to filter out all the ways through which the user can authenticate (cookie, authorization etc). This meant that if the original request was a POST, and it had a `content-length` header, this header was also forwarded. This meant that when the connection would later be reused, the server (haproxy in this case) would truncate the previous's request `content-length` number of characters from the next request, believing the previous request was not finished. Obviously, this new truncated request was invalid and generated a 400 status code. 

This only seems to happen when accessing API directly or through the AWS load balancer, and never when running requests through nginx. 

To avoid next request truncation, I am no longer forwarding `content-length` headers in the session request. 

#8868

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8868-dont-forward-all-headers-for-get/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8868-dont-forward-all-headers-for-get/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8868-dont-forward-all-headers-for-get/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/api/tests/mocha/auth.spec.js
+++ b/api/tests/mocha/auth.spec.js
@@ -145,13 +145,15 @@ describe('Auth', () => {
     });
 
     it('should clean content-length headers before forwarding', async () => {
-      sinon.stub(rpn, 'get').resolves({ userCtx: { name: 'user', roles: ['userrole'] }});
+      sinon.stub(rpn, 'get').resolves({ userCtx: { name: 'theuser', roles: ['userrole'] }});
 
       req.headers['content-length'] = 100;
       req.headers['Content-Length'] = 22;
+      req.headers['Content-length'] = 44;
+      req.headers['content-Length'] = 82;
 
       const result = await auth.getUserCtx(req);
-      chai.expect(result).to.deep.equal({ name: 'user', roles: ['userrole'] });
+      chai.expect(result).to.deep.equal({ name: 'theuser', roles: ['userrole'] });
       chai.expect(rpn.get.args).to.deep.equal([[{
         url: 'http://abc.com/_session',
         json: true,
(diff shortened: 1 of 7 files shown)


dianabarsan:
Added an e2e test to demonstrate fixed behavior, and an opposite e2e test here: https://github.com/medic/cht-core/compare/test-login?expand=1 that shows the 400 request: https://github.com/medic/cht-core/actions/runs/8274351579/job/22639815092#step:17:286

dianabarsan:
Hi @latin-panda 

Would you mind reviewing this? Thanks a lot!

api/tests/mocha/auth.spec.js
```diff
@@ -116,4 +126,79 @@ describe('Auth', () => {
...
+      }]]);
+    });
+
+    it('should clean content-length headers before forwarding', async () => {
+      sinon.stub(rpn, 'get').resolves({ userCtx: { name: 'theuser', roles: ['userrole'] }});
+
+      req.headers['content-length'] = 100;
+      req.headers['Content-Length'] = 22;
+      req.headers['Content-length'] = 44;
+      req.headers['content-Length'] = 82;
```
latin-panda, line 153:
```suggestion
      req.headers['content-Length'] = 82;
      req.headers['CONTENT-LENGTH'] = 240;
```

latin-panda (approved):
LGTM!

Thanks for explaining in the PR's description box, I learned from it!

Issue #8348, opened 2023-06-28
Adding error object to logger in sentinel/server.js


robin-murphy1:
**What feature do you want to improve?**

In the Cht-Core/Sentinel module the server.js file has a error block that prints out a text message but does not include the error message. 
```js
//filename: sentinel/server.js

24 request({ url, json: true }, (err, response, body) => {
25         if (err) {
26           logger.info('Waiting for API to be ready...');
27           return setTimeout(() => waitLoop(), 10 * 1000);
        }
```

It took some time to realize that Sentinel wasn't connecting and figure out that it was failing to connect to the API. Adding the error object to the log message helped in troubleshooting the issue quickly.

**Describe the improvement you'd like**

Modify the sentinel/server.js file to include the error object in the log message.

```js
//filename: sentinel/server.js

24 request({ url, json: true }, (err, response, body) => {
25         if (err) {
26           logger.warn('Waiting for API to be ready...', err);
27           return setTimeout(() => waitLoop(), 10 * 1000);
        }
```

**Describe alternatives you've considered**

None

**Additional context**

Small change but first time contributor so want to understand the process of contributing to the source base

garethbowen:
Hi @robin-murphy1 and welcome!

This looks like a nice change and you're most welcome to submit a PR for review. If you're interested here's some documentation about how the process works: https://docs.communityhealthtoolkit.org/contribute/code/

If you'd like to contribute the fix, the first step is to assign this issue to yourself, then fork the codebase locally, and start working on a Pull Request.

One thing to think about is the impact this may have on the log size if sentinel can't connect for a long time. A couple of options to consider:

1. caching the error and only outputting it if it's different from the previous error, or
2. only printing the error every 10th attempt

Also a slight change to the logging, because we use `winston` it's better to output the error like so:

```
logger.warn('Waiting for API to be ready...: %o', err);
```

You can see other examples like this in that file too.

Please reach out if you have any other questions.

ElijahBus:
Hi @garethbowen !
What's the status of this issue, did it get resolved ? 

I can see that it didn't go further from here, I am ready to move forward with it.

garethbowen:
@ElijahBus I don't think @robin-murphy1 is working on this so please feel free to pick it up.

billoserra:
Hi! I’d like to work on this issue if it’s available. Could you please assign it to me?

PR #8927, opened 2024-03-12
Add error object to sentinel start up error log (#8348)

ElijahBus:
fix(#8348): add error object to sentinel start-up error log

# Description

When starting the sentinel project in development mode, the message ` Waiting for API to be ready ...` gets printed on the console if the module is not able to reach the dev-api project. 

Adding the `err` object as part of the logs will provide more insights on what could be the reason the app is not able to connect to the dev-api.  To minimize the impact of the log size, and the repetition of the same error over and over,  The error object will be logged at every tenth attempt of connection.

#8348 

# Code review checklist

# License

### DIFF
--- a/sentinel/server.js
+++ b/sentinel/server.js
@@ -19,14 +19,25 @@ const waitForApi = () => new Promise(resolve => {
   // This waits forever, with no escape hatch, because there is no way currently
   // to know what API is doing, and migrations could legitimately take days
   const url = `http://${process.env.API_HOST || 'localhost'}:${process.env.API_PORT || 5988}/setup/poll`;
+
+  let errorLogAttempts = 0;
+
   const waitLoop = () => {
     request({ url, json: true }, (err, response, body) => {
       if (err) {
         logger.info('Waiting for API to be ready...');
+
+        const canLogError = errorLogAttempts % 10 === 0;
+        if (canLogError) {
+          logger.error('%o', err);
+        }
+
+        errorLogAttempts++;
         return setTimeout(() => waitLoop(), 10 * 1000);
       }
 
       logger.info(`API is ready: ${JSON.stringify(body)}`);
+      errorLogAttempts = 0;
       resolve();
     });
   };


dianabarsan (commented):
Hi @ElijahBus 

Thanks for submitting this PR. I think it's really useful to see the connection error between API and Sentinel. However I'm not really sure about only logging every 10 attempts for a couple of reasons: 

1. it's assuming the _same_ error is logged over and over. This might not be true. 
2. If we are to account that the error might not be the same, we would add additional state and complexity. 
3. we retry the connection every 10 seconds, so it's not the most noisy log by default. 

I think my preferred approach wold be to log _enough_ information for every try that we can assess if the error is recurring and what it is, and not so much that we spam logs with MB of repeated request / response stack traces. 

How does the error that you are seeing look like?

garethbowen:
@ElijahBus Have you had a chance to have another look at this?

ElijahBus:
Hi @garethbowen @dianabarsan  !

Sorry for the late reply, I have given this another try; In response to the mentioned points :

The error will most likely be the _same_ at this point with only one reason the "Sentinel service" cannot establish a connection with the API service, and since the error being thrown comes from the API service, Unless the implementation changes on the API Service, probably by returning a custom Error, then we might have a different error. 

Another alternative could be to check  the error code, and we only log the full trace twice if the previous error code isn't the same as the current one, what do you think about that, will it make the solution more complex and probably rise the idea of checking other error codes as well ( now or in the future ) ?

The full trace isn't so big though, 

Here is what the error looks like:

image

andrablaj:
Hi @ElijahBus! Are you still interested in working on this issue? It's been stale for a few months now. Please let us know if you need further support to continue or close the PR if you don't plan to continue working on it. Thank you!

ElijahBus:
Hi @andrablaj, Thanks for reminding of this. Yes, I will find some time to work on it during the week

Issue #8877, opened 2024-02-16
/api/v2/users look up users by `facility_id` and/or `contact_id`


kennsippell:
**Is your feature request related to a problem? Please describe.**
I want to look up the details for a single user by username
I want to look up the details for users with a specific facility_id or contact_id

**Describe the solution you'd like**
~~`GET /api/v2/users/username`~~ (Edit: split out retrieving data for a single user into https://github.com/medic/cht-core/issues/8986)
`GET /api/v2/users?facility_id=foo`
`GET /api/v2/users?contact_id=bar`

**Describe alternatives you've considered**
Admin users (only) can query the database directly and use `/_users/_find`
You can add `mm-online` role to the `_users/_security` document

**Additional context**

garethbowen:
One workaround would be to install a custom ddoc with a map/reduce view to facilitate this. Obviously the API is better but this would work on all versions of the CHT until the API is released.

Something like...

```js
function (doc) {
  if (doc.type === 'user-settings') {
    if (doc.facility_id) emit('facility:' + doc.facility_id);
    if (doc.contact_id) emit('contact:' + doc.contact_id);
  }
}
```

Use a bespoke ddoc so it doesn't get overwritten on upgrade.

Then for the `GET /api/v2/users/username` just fetch the `org.couchdb.user:username` doc from the medic db, which (probably?) has everything you need.

I think this will work but haven't actually tried it out yet...

kennsippell:
Sadly I don't think this will work above 4.4 because non-admin users can't send GET requests to the `_users` database after the CouchDB change.   I'll double check.

garethbowen:
That's the beauty of this solution... the ddoc above is in the `medic` db, not the `_users` db.

kennsippell:
Yes true. Hopefully the data in the `user-settings` doc matches the data in the `user` doc.  I'll use a _find request with payload:

```json
{
  "selector": {
    "type": "user-settings",
    "facility_id": facilityId,
    "$or": [
      { "inactive": false },
      { "inactive": { "$exists": false } }
    ]
  }
}
```

Since we're using this for user management, seems like it is not a best practice since `user-settings` can be edited by users. Here's a nifty scenario. Don't really know if I should worry about stuff like this:

1. Malicious user logs in
2. Malicious user changes facility_id in their user-settings doc (in chrome console or otherwise)
```javascript
const myDoc = await CHTCore.DB.get().get('org.couchdb.user:user_manager');
[...]

Issue #7499, opened 2022-02-04
Relocate API migrations integration tests to the integration suite


dianabarsan:
**Describe the issue**
We have a bunch of [API migration integration tests](https://github.com/medic/cht-core/tree/118c7756eec02b08e88433226f1a39293aef8209/api/tests/integration) that run part of the API code against a real CouchDb database. However these tests are run without having API properly started, and run as part of the [build process](https://github.com/medic/cht-core/blob/master/Gruntfile.js#L1035) before API code is even tested. 

**Describe the improvement you'd like**
Rework the tests so they run against a real working API: 
- before each test, change the migration_log file so the migration would run
- create the documents to be migrated
- restart api
- evaluate results after the migration has run

Run these tests as part of the [e2e-integration suite](https://github.com/medic/cht-core/blob/master/Gruntfile.js#L613). 

**Describe alternatives you've considered**
Leave it as it is.

PR #8928, opened 2024-03-13
feat(#8877): Look up users from their `facility_id` or `contact_id`

m5r:
# Description

Issue: #8877 
Docs PR: medic/cht-docs#1318

This extends the `GET /api/v2/users` route to support passing a `facility_id` and/or a `contact_id` in query parameters and get back a list of users who matched. It's gated behind the same `can_view_users` permission.

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8877-lookup-single-user/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8877-lookup-single-user/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8877-lookup-single-user/docker-compose/cht-couchdb-clustered.yml)

# License


shared-libs/user-management/src/users.js
```diff
@@ -770,8 +786,18 @@ const getUserSettings = async({ name }) => {
...
+  getList: async (filters) => {
+    let users;
+    let settings;
+
+    if (_.isEmpty(filters)) {
+      [users, settings] = await Promise.all([getAllUsers(), getAllUserSettings()]);
+    } else {
+      settings = await getAllUserSettings(filters);
+      const ids = settings.map(({ _id }) => _id);
+      users = await getUsersByIds(ids);
```
m5r, line 798: User docs have a `facility_id` property but no `contact_id`, that's why we have to query the `user-setting` docs first and then get the user docs from their ids

api/src/controllers/users.js
```diff
@@ -233,7 +233,12 @@ module.exports = {
   v2: {
     get: async (req, res) => {
       try {
-        const body = await getUserList(req, res);
+        await auth.check(req, 'can_view_users');
+        const filters = _.chain(req.query)
+          .pick(['facility_id', 'contact_id']) // keep only supported filtering properties
+          .mapKeys((value, key) => _.camelCase(key)) // rename them to keep consistent variable names
+          .value();
```
m5r, line 240: This filters out any property we don't support filtering users on and renames them to camelCase to keep consistent variable names

```ts
{ whatever: true, contact_id: '123' } // => { contactId: '123' }
```

m5r:
@kennsippell could you give it a look and see if this would help with the user management needs raised in #8877 ?

You also mentioned wanting a route to lookup a single user by username. It is not included it in this PR but there is a WIP incremental PR #8959 we could ship next if that's valuable for user management

shared-libs/user-management/src/users.js
jkuester: Nice and efficient to just get the users by id!  

One thing I do want to call out here is that we are filtering by the `facility_id` in the `user-settings` doc (and not the one in the `_users` doc).  This has the [acknowledged drawback](https://github.com/medic/cht-core/issues/8877#issuecomment-1952052847) that currently users can edit their own `user-settings` doc.  Sounds like this will be locked down in the future.  (Also, this may be the reason this code is not running into the issues with non-admin users calling the ` /_users/_find` endpoint....)

I don't think we need to change anything here (any change I can think of will just be worse performance), unless we think it is important that the API code uses the `facility_id` data from the `_users` doc...

shared-libs/user-management/src/users.js
```diff
@@ -105,18 +105,34 @@ const getDocID = doc => {
...
   return db.users.allDocs({ include_docs: true })
-    .then(result => result.rows);
+    .then(result => result.rows.map(({ doc }) => doc));
+};
+
+const getUsersByIds = async (ids) => {
+  const docs = ids.map(id => ({ id }));
+  const { results } = await db.users.bulkGet({ docs });
+  return results
+    .map(result => result.docs && result.docs[0] && result.docs[0].ok)
```
jkuester, line 134:
```suggestion
    .map(result => result?.docs?.[0]?.ok)
```

shared-libs/user-management/src/users.js
```diff
@@ -335,21 +351,21 @@ const hasParent = (facility, id) => {
...
-        username: user.doc.name,
+        id: user._id,
+        rev: user._rev,
+        username: user.name,
         fullname: setting.fullname,
         email: setting.email,
         phone: setting.phone,
-        place: getDoc(user.doc.facility_id, facilities),
-        roles: user.doc.roles,
+        place: getDoc(user.facility_id, facilities),
```
jkuester, line 364: IMHO we really need to switch this to `settings.facility_id` or else change to query the `_users` table by `facility_id`.  Using both like this is going to make for some debugging hell later on....

shared-libs/user-management/test/unit/users.spec.js
```diff
@@ -190,186 +196,296 @@ describe('Users service', () => {
...
+            _id: 'org.couchdb.user:y',
+            doc: {
+              name: 'milan',
+              facility_id: 'b',
+              fullname: 'Milan A',
+              email: 'm@a.com',
+              phone: '987654321',
+              roles: [ 'district-admin' ]
+            }
+          }
```
jkuester, line 481:
```suggestion
          {
            _id: 'x',
            name: 'lucas',
            facility_id: 'c',
            fullname: 'Lucas M',
            email: 'l@m.com',
            phone: '123456789',
            roles: [ 'national-admin', 'data-entry' ]
          },
          {
            _id: 'org.couchdb.user:y',
            name: 'milan',
            facility_id: 'b',
            fullname: 'Milan A',
            email: 'm@a.com',
            phone: '987654321',
            roles: [ 'district-admin' ]
          }
```

shared-libs/user-management/test/unit/users.spec.js
```diff
@@ -190,186 +196,296 @@ describe('Users service', () => {
...
-        },
-        {
-          id: 'org.couchdb.user:y',
-          doc: {
+            phone: '123456789'
+          },
+        ];
+        const getUsersByIds = sinon.stub().resolves(usersByIds);
+        service.__set__('getAllUserSettings', sinon.stub().resolves(allUsersSettings));
+        service.__set__('getUsersByIds', getUsersByIds);
```
jkuester, line 221: We have a good bit of logic in these functions that AFICT is not getting unit tested at all since we are stubbing these out.  

Would it be possible to update your new "with filters" tests to do the stubbing at the db level so these tests can cover the actual behavior of these functions?

tests/integration/api/controllers/users.spec.js
```diff
@@ -1616,6 +1615,43 @@ describe('Users API', () => {
           'contact.name': user.contact.name,
         });
       }
+
+      await Promise.all(savedUsers.map(async (savedUser) => {
```
jkuester, line 1619: I think the probably deserves its own test case where we can be more intentional about writing users with different/same facility_id and contact_id values. (Really this is the only test that is going to validate interaction with the Couch view...) e.g.:

- User 1:
    - Facility A
    - Contact B
- User 2:
    - Facility A
    - Contact B
- User 3:
    - Facility A
    - Contact C
- User 4:
    - Facility D
    - Contact E

Then we can run several GET requests and make sure we get back what we expect:

- facility_id=A, contact_id=B
    - Returns 1,2
- facility_id=A
    - Returns 1,2,3
- contact_id=B
    - Returns 1,2

jkuester (changes requested):
Nice and clean!  Made a few minor test requests.

shared-libs/user-management/src/users.js
jkuester: @m5r did you happen to get @garethbowen's input on using the `user-settings.facility_id` to query on?  The more I think about this, the more I would like to know his thoughts before shipping this.  (But he is OOO all of next week....)

shared-libs/user-management/test/unit/users.spec.js
m5r: Good catch!

tests/integration/api/controllers/users.spec.js
lorerod: I would like to add a scenario using a facility ID and/or contact ID that does not exist. What do you think?

ddocs/medic-db/medic-client/views/doc_by_type/map.js
```diff
@@ -6,5 +6,12 @@ function(doc) {
     });
     return;
   }
+  if (doc.type === 'user-settings') {
```
dianabarsan, line 9: I'm a little worried by this change. This will mean that all clients (users' phones) will need to index this massive view again due to this change that only gets used on the server side.
I find it hard to see a context where we would need this change on the client. I think it makes more sense to put this new view code in `_design/medic` somehow.

m5r: Good catch there. I don't see any existing view in `_design/medic` we could piggyback off of so we'd have to make a new view.
The way I see it, I could either duplicate the `medic-client/doc_by_type` code to a `medic/doc_by_type` view and pop the changes there (and potentially redirect all the `medic-client/doc_by_type` queries in API to this view) or make a specific view named like `medic/user_settings_by_facility_and_contact` that would look like this:
```ts
function(doc) {
  if (doc.type === 'user-settings') {
    emit(doc.facility_id);
    emit(doc.contact_id);
    emit([doc.facility_id, doc.contact_id]);
  }
}
```

I'm leaning towards the first option and duplicate the view and re-map API queries from the client view to have a distinct API-only `doc_by_type` view so that we can make changes to it without forcing clients to index it again.

jkuester: IMHO, we probably only want to do the first option if it would mean that we could avoid having to build the `medic-client/doc_by_type` on the server at all.  Otherwise, having both `medic-client/doc_by_type` and `medic/doc_by_type` would mean a non-trivial addition to the disk space being used on the server.  Do we support having views on the server _that do not get warmed?_

kennsippell:
I'm creating a user, viewing it, deleting it, and then viewing it again. 

1. `curl http://medic:password@localhost:5988/api/v2/users -X POST -d '{ "username": "114_test", "password": "S3cr$t!!!", "type": "chp", "place": "984cd964-4d06-4d96-80a7-21043804153e", "contact": "72095736929b6a7ee751a21337002e43" }' -H "Content-Type: application/json"`
2. `curl "http://medic:password@localhost:5988/api/v2/users?facility_id=984cd964-4d06-4d96-80a7-21043804153e"`                                                                                       
3. `curl -X DELETE "http://medic:password@localhost:5988/api/v1/users/114_test"`
4. `curl "http://medic:password@localhost:5988/api/v2/users?facility_id=984cd964-4d06-4d96-80a7-21043804153e"`

I would expect that the user does not appear after being deleted. But I'm seeing it in the list of returned users.  Is this expected?

m5r:
Right, this is not expected and I got a fix in the works alongside changes to save `contact_id` in the `_users` docs.

To give a bit of context, this happens because when a user is deleted, its related document in `_user` is removed from the database but not its related `user-settings` in `medic`. Along with the `contact_id` change mentioned above, we'll use the `_users` db to query users and this bug won't happen anymore.

lorerod:
> I'm creating a user, viewing it, deleting it, and then viewing it again.
> 
> 1. `curl http://medic:password@localhost:5988/api/v2/users -X POST -d '{ "username": "114_test", "password": "S3cr$t!!!", "type": "chp", "place": "984cd964-4d06-4d96-80a7-21043804153e", "contact": "72095736929b6a7ee751a21337002e43" }' -H "Content-Type: application/json"`
> 2. `curl "http://medic:password@localhost:5988/api/v2/users?facility_id=984cd964-4d06-4d96-80a7-21043804153e"`
> 3. `curl -X DELETE "http://medic:password@localhost:5988/api/v1/users/114_test"`
> 4. `curl "http://medic:password@localhost:5988/api/v2/users?facility_id=984cd964-4d06-4d96-80a7-21043804153e"`
> 
> I would expect that the user does not appear after being deleted. But I'm seeing it in the list of returned users. Is this expected?

@ m5r, is it possible to include this scenario in the API tests? Adding more scenarios like this could be interesting.

m5r:
@lorerod good idea, I've added it to [the integration test](https://github.com/medic/cht-core/blob/8877-lookup-single-user/tests/integration/api/controllers/users.spec.js#L1700-L1704)

m5r:
I just pushed the last bits to complement all your suggestions. There is now a migration in place to save every user's `contact_id` property to their respective `_users` document and it gets updated whenever the same property in the user's settings document is updated.

To answer the issue raised in medic/cht-user-management#76 about non-admin users getting a 403 when running the `_users/_find` query, this route could bypass the firewall and allow them to lookup users by their `facility_id`/`contact_id` as long as they have the `can_view_users` permission.

lorerod (approved):
Thank you @m5r for adding the suggested scenario.

jkuester (approved):
This is  complete from my perspective!  Before merging, it would be good to get some feedback on the following:

@dianabarsan can you have a quick look, when you get the chance, at the new migration and the new views for the `_users` db?  I just want to be sure that is what you expected!  (And we do not add migrations/views all the time so just want to double-check my work here... :crossed_fingers: ).

@m5r can you have a look at the revisions I made?  The biggest change was switching from the `_find` Mango query to normal Couch views.  I ended up doing this for these reasons:

1. Performance for the un-indexed Mango query was 14x worse than with the views with 10000 users.  The performance of the Mango query was not abysmal, but it seemed like we did not have much of a reason to take the performance hit.  (FTR, I did look into _indexing_ the Mango query. That provided equivalent performance as the view, but we do not currently have a pattern of creating/maintaining Mango indexes like we do views, so more work is needed before we can deploy them).
2. As far as I can tell, we do not actually use the Mango `find` call anywhere in our code right now _except_ for the DB migrations (where custom views do not make sense).  This made me reluctant to introduce the pattern here (unless this is a direction we intentionally want to move...)
3. I guess the `pouchdb-find` dependency is required to use that functionality and not all the consumers of `user-management` are currently importing that dependency (e.g. sentinel).  This is not a huge deal since sentinel could always pull in that package if we ever start to hit this code flow, but it added to my unease.  
4. I found that the Mango `find` has a default `25` doc limit on what it returns.  We can set a higher limit, but there does not seem to be a way to go unlimited.  Whenever we do use `find` we should be sure to implement paging logic so we get all the results... (Actually logged [an issue](https://github.com/medic/cht-core/issues/9004) because neither of the other places where we are using `Pouch.find` are actually accounting for this limit either... :facepalm: )

ddocs/users-db/users/views/users_by_contact_id/map.js
```diff
@@ -0,0 +1,5 @@
+function(doc) {
+  if (doc.contact_id) {
+    emit(doc.contact_id);
```
dianabarsan, line 3: A common pattern that we've used in other views that index related properties is to emit an array, like: 
```
function(doc) {
  if (doc.contact_id) {
    emit(['contact_id', doc.contact_id]);
  }
  if (doc.facility_id) {
    emit(['facility_id', doc.facility_id]);
  }
}
```

I think adding multiple views makes sense when the view code is very different. One other advantage is that you could emit both facility and contact and potentially open the possibility to query for both. For this, I think a single view can work ok. What do you think?

shared-libs/user-management/src/users.js
```diff
@@ -770,9 +802,9 @@ const getUserSettings = async({ name }) => {
...
   deleteUser: username => deleteUser(createID(username)),
-  getList: async () => {
-    const [ users, settings ] = await Promise.all([ getAllUsers(), getAllUserSettings() ]);
-    const facilities = await facility.list(users, settings);
+  getList: async (filters) => {
+    const [users, settings] = await getUsersAndSettings(filters);
+    const facilities = await facility.list(users);
     return mapUsers(users, settings, facilities);
   },
   getUserSettings,
```
dianabarsan, line 810: I believe we should also change how this function is implemented, because now it only picks the facility_id from the users doc, and the contact_id from the user-settings doc. I think we should get both from the user doc.

dianabarsan (changes requested):
Great work here! Looking really good. 
I left a few small requests inline.

shared-libs/user-management/src/users.js
```diff
@@ -105,18 +106,45 @@ const getDocID = doc => {
...
+  .then(({ rows }) => rows.map(({ doc }) => doc));
+
+const getAllUserSettings = () => queryDocs(db.medic, 'medic-client/doc_by_type', ['user-settings']);
+
+const getSettingsByIds = async (ids) => {
+  const docs = ids.map(id => ({ id }));
+  const { results } = await db.medic.bulkGet({ docs });
+  return results
+    .map(result => result?.docs?.[0]?.ok)
+    .filter(Boolean);
```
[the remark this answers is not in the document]
jkuester: :+1: The amazing api naming here always makes me forget that `allDocs` can take a `keys` param....

api/tests/integration/migrations/add-contact-id-to-user.js
```diff
@@ -0,0 +1,112 @@
...
+    await utils.initDb(userSettingsDocs);
+    await writeUserDocs(userDocTuples.map(([, userDoc]) => userDoc));
+
+    await utils.runMigration('add-contact-id-to-user-docs');
+
+    await utils.assertDb(userSettingsDocs);
+    for (const [userSettingsDoc, userDoc] of userDocTuples) {
+      const updatedUserDoc = await getUserDoc(userDoc._id);
+      expect(updatedUserDoc).to.deep.include({ ...userDoc, contact_id: userSettingsDoc.contact_id });
+    }
```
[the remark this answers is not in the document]
jkuester: :+1: Added assertion to the last test that verifies the `facility_id` value on the _users doc does not get updated.

api/tests/integration/migrations/utils.js
```diff
@@ -166,6 +168,9 @@ const switchToTestDbs = () => {
   db.sentinel = new PouchDB(
     realSentinelDb.name.replace(/medic-sentinel$/, 'medic-sentinel-test')
   );
+  db.users = new PouchDB(
+    realUsersDb.name.replace(/_users$/, 'users-test')
```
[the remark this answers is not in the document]
jkuester: @dianabarsan I spent a couple hours digging on this today and just want to be sure I am on the right path before investing more time into this.  I agree that it is very not ideal to use the local `COUCH_URL` instance and it would be better to start one in Docker.  After poking at this for awhile, it seems like there are two "viable" approaches here:

1. Move the tests down to `tests/integration/api` and leverage all of the e2e tooling for spooling docker containers.  This seems simple at first, but it is going to be non-trivial to just spin up the Couch containers (and not the rest of the services).  Even if we decide we don't care and will just spin up everything, it still is kinda weird since the tests in  `tests/integration/api`  will be directly referencing the migration script files in `api/src/migrations` to execute them.
2. Keep the tests in `api/tests/integration`, but roll our own custom docker tooling for spinning up Couch that is bespoke to those tests.  This would still need to (I guess?) account for using the freshly generated Couch images, but could maybe avoid docker-compose and clustered couch nodes...

Is either of these close to what you were thinking?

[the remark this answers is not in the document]
dianabarsan: I think we should run these integration tests over the correct stack. 
I've had this issue opened two years ago :))  https://github.com/medic/cht-core/issues/7499
I _don't_ think it should be part of this PR. So we could leave it as-is and address it when we move the tests.

dianabarsan (approved):
Amazing work! Not only is this a nice feature, we're also pushing for consistency and security of data by having the contact_id stored in the user doc. Love it!

Issue #8925, opened 2024-03-12
Enable http-keep-alive in haproxy


dianabarsan:
**What feature do you want to improve?**
`http-keep-alive` is the default option in haproxy. it can be turned off by using `http-server-close`, which we are using at the moment. 
I tried to search git history to find the time when we added this, unsuccessfully. My only thought is that this was added while running early 4.x scalability testing. 

**Describe the improvement you'd like**
Enable and test http-keep-alive in haproxy. 

**Additional context**
Starting with Node 19, Node automatically adds the keep-alive header to all requests.

dianabarsan:
Merged to `master`.

PR #8942, opened 2024-03-17
feat(#8925): enable http-keep-alive in haproxy

dianabarsan:
# Description

#8925

# Code review checklist

# License

### DIFF
--- a/haproxy/default_frontend.cfg
+++ b/haproxy/default_frontend.cfg
@@ -35,7 +35,7 @@ defaults
   option httplog
   option forwardfor
   option redispatch
-  option http-server-close
+  option http-keep-alive
   timeout client 15000000
   timeout server 360000000
   timeout connect 1500000


dianabarsan:
Hi @nydr 

I'm wondering if there's anything that might concern us with this change. As of 4.6.0, NodeJS requests (made through the node:http library) automatically enable `keep-alive`. This is also the default of haproxy: https://cbonte.github.io/haproxy-dconv/2.6/configuration.html#4.2-option%20http-keep-alive

nydr:
> anything that might concern us with this change.

I can't think of anything since it has the configuration line `timeout http-keep-alive 5m` (without that it would default to one of the other timeouts which are unusually long)

dianabarsan:
I ran scalability tests with 300 concurrent users locally against this setup and got no errors and no concerning spikes in CPU usage or memory in any of the services.

PR #8943, opened 2024-03-18
fix(#8944): migration of performance metrics to use performance service

latin-panda:
# Description

Migrate the telemetry recording performance metrics from Date API to the new Performance service. This is better because the service uses `Performance.now` that is based on a [monotonic clock](https://w3c.github.io/hr-time/#dfn-monotonic-clock), which current time never decreases and maybe avoids negative values in telemetry metrics. 

https://github.com/medic/cht-core/issues/8944

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:performance-metrics-migration/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:performance-metrics-migration/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:performance-metrics-migration/docker-compose/cht-couchdb-clustered.yml)

# License


webapp/src/ts/services/db-sync.service.ts
```diff
@@ -391,22 +395,24 @@ class DbSyncTelemetry {
...
+    public performanceService: PerformanceService,
     database,
     direction,
     lastReplicated?,
   ) {
     this.database = database;
     this.direction = direction;
-    this.start = Date.now();
     this.lastReplicated = lastReplicated;
+    this.endLastReplicated = Date.now(); // The lastReplicated is based on Date API, so we use the Date API here too.
```
latin-panda, line 413: The `lastReplicated` is used in the About page too, I decided not to change the Date API to make the calculation using the same thing (apples with apples)

latin-panda:
@Benmuiruri @ralfudx, please have a look. This migration should not affect any pre-existing telemetry metrics, and they should be recorded correctly.

webapp/src/ts/services/rules-engine.service.ts
```diff
@@ -143,50 +148,21 @@ export class RulesEngineService implements OnDestroy {
...
                 }
 
-                initializeTelemetryData.record();
+                trackPerformance?.stop({ name: [ 'rules-engine', 'initialize' ].join(':') });
               });
           });
       });
   }
 
-  private telemetryEntry(entry, startNow = false) {
```
Benmuiruri, line 158: Nice Clean up 👌

Benmuiruri (approved):
Looks good. Reviewing this PR has helped me appreciate the implementation of `PerformanceService`

ralfudx:
@latin-panda qq pls - it seems the telemetry service is being replaced? any major reasons?

latin-panda:
@ralfudx Good question. We are encapsulating the small logic to track performance, ensuring the calculation is the same everywhere and voiding small mistakes. Before, we had inconsistencies since some places were using Date.now() and other areas were using Performance.now(), with the Performance service, we are using the same way (Performance.now(), more reliable than Date API).

The Performance service uses the Telemetry service to record these metrics. In some other places, we record metrics that are not about performance; for those, we use the Telemetry service directly. So, the Telemetry service will stay.

ralfudx (approved):
Ok thanks for the explanation - LG

Issue #8940, opened 2024-03-15
Admin app shows error after successful upgrade


jkuester:
**Describe the bug**

When using the Admin app to upgrade a CHT instance to the latest tagged version (e.g. `4.6.0-beta.3`), the installation will complete successfully, but the UI will still display the installation error card.  

**To Reproduce**

Deploy a CHT instance that can be upgraded via the Admin app (I used EKS)

In the Admin app on the Upgrades tab, select "Install" for the `4.6.0-beta.3` tag.  Wait for the install to complete and see the error message (even though the installation succeeded).  

![image](https://github.com/medic/cht-core/assets/8441903/96547788-fbd9-4881-9546-bfb1989a6a03)

Alternatively, instead of selecting "Install", select "Stage".  Then, once the installation is staged, click "Install" and see the error (note that, for me, the page automatically updates after ~30sec to clear this screen and take me back to the normal Upgrades tab.  I never see the "Deployment Complete" card, but at least the error goes away....):

![image](https://github.com/medic/cht-core/assets/8441903/e4859223-25b7-43c3-82a7-96c28ca6d25c)

**Expected behavior**

Either way, when the installation succeeds, I expect to see "Deployment complete.":

![image](https://github.com/medic/cht-core/assets/8441903/969fa609-454e-4a67-8b19-d1db4883f160)

**Logs**

There are lots of errors in the console log, but I think most of them are normal while polling for the updated api server to come back online.  However, as the error card appears at the end of the upgrade, I do see: `main.js:1 Error triggering update instance.upgrade.error.deploy`

**Additional context**

This behavior only seems to be happening for the `4.6.0` beta tags.   If I start from an older CHT version,  such as `4.3.0`, and then try upgrading to `4.5.2-beta.1`, I do not see any problems.  Then if I upgrade to `4.6.0-beta.3`, I see the error card being displayed.

jkuester:
@garethbowen now that we have resolved https://github.com/medic/cht-core/issues/8868 and there are not any session issues after upgrading, we have still been able to re-create these error display issues.  My main question at this point is if you think this should block the `4.6` release?  [Similar behavior](https://github.com/medic/cht-core/issues/8868#issuecomment-1942449890) has been seen before on older code, but I was not able to reproduce this myself.  As far as I can reproduce, this problem behavior only happens for the 4.6.0 beta tags.

garethbowen:
[...]

Issue #8790, opened 2024-01-08
`/api/deploy-info.version` is not semver valid for final releases


dianabarsan:
**Describe the bug**
Instead of returning the semver valid version of the app, `api/build-info` returns the build version. 

**To Reproduce**

1. Deploy 4.5.1
2. Get `api/deploy-info`
3. Notice that `version` field looks like `4.5.0.7180278601`

**Expected behavior**
Version should be semver valid. 

**Environment**
- App: api
- Version: 4.0.0

dianabarsan:
Merged to `master`.

PR #8965, opened 2024-03-23
fix(#8940): fix upgrade version check

dianabarsan:
# Description

Due to the `deplpy-info` [API change](https://github.com/medic/cht-core/issues/8790), the admin upgrade page ended up comparing a `version` (4.6.0) to a build (4.6.0.432424242) and always reported that the upgrade was not completed.  

#8940 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8940-fix-upgrade/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8940-fix-upgrade/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8940-fix-upgrade/docker-compose/cht-couchdb-clustered.yml)

# License


dianabarsan:
Testing this is very awkward and kind of a nuisance. 
Because the change is in admin app, and when we upgrade we don't actually refresh the admin upgrade page, this means that we will only see the fix in the next version - so upgrading from 4.6.0 -> 4.7.0 will still display the error, because 4.6.0 admin page will be the one running the upgrade. The fix will only be seen when upgrading from 4.7.0 -> 4.8.0. 

To test this, I installed my own staging server locally, and set API to use this staging server. I rebuilt the upgrade service so it doesn't call `docker-compose pull` - which fails if you try to pull locally tagged images that don't exist upstream, and then I built and published two CHT versions: 4.7.0 and 4.8.0. I started 4.6.0 with the modded upgrade service, upgraded to 4.7.0 and saw the error, and then upgraded to 4.8.0 and didn't see the error. 

An alternative to modding the upgrade service is to run your own local docker registry and push images there.

lorerod:
Hi @dianabarsan, do you know if this may fix the failed test I'm having in [this](https://github.com/medic/cht-core/actions/runs/8542367732/job/23415302384) PR?

Issue #3605, opened 2017-07-03
Bulk delete fails when deleting more than a few hundred records


sglangevin:
Original ticket: https://github.com/medic/medic-projects/issues/2409

This has been reported a few times. The workaround has been to delete in smaller batches, but we should probably fix this so that new projects and projects that are upgrading can make better use of this feature. To reproduce, try to delete a few thousand records on any instance and you'll get a 500 error.

@estellecomment has suggested batching as a potential solution.

cc @ashish-medic @bishwas-medic

mandric:
Possibly related: https://github.com/medic/medic-api/issues/96

SCdF:
@sglangevin what is the maximum number of records that should be supported in the bulk delete UI do you think? Naively I'd suggest that if you're deleting hundreds of documents you wouldn't want to be doing it through the UI anyway, and it's something that a script is more suited to. However, I may be missing some use case here.

estellecomment:
Use case : the carrier wants to send new ringtones to the gateway phone. Message loop ensues and 10k messages are created for nothing. Tech lead has to delete them.

sglangevin:
Another use case: hundreds or thousands of messages are sent during training and we want to delete them. Bulk delete was meant to make it easy for PMs to delete training data.

SCdF:
Yes I understand there exists situations where you'd want to delete thousands or documents. I'm sure there are many more than mentioned. The question was whether or not the UI was required to solve them.

Considering that Gareth sounded surprised that bulk delete was being used to delete thousands of documents the requirement for it to scale to the thousands clearly got lost somewhere.

@sglangevin are the search tools adequate for finding these documents?

I am wondering if we'd need to also take another look at the things we're doing that enables us to have a nice UI (for example, every time you select a document you see info about that added to a list), because I worry that is also going to be awful for thousands of documents, especially on poor connections.

SCdF:
OK, I've thought about this some more. Here is what I think we'd need to do to support deleting 10,000 documents in the UI:
 - RHS list of docs selected should follow the same rules as the LHS bar: only query for ~50 of them. If you want to see more you need to scroll.
    - Simpler solution: only ever show the first 50, just say "…and 9950 more" after that
 - We also want to solve #3465 as part of this, as we'll be doing refactoring in this area
 - If it doesn't already (don't have enough local data to test) select all should select all search results, not just results loaded in LHS right now.
[...]

PR #8984, opened 2024-04-02
fix(#8815): fix contact-summary calculation to include all reports

### LINKED ISSUE #8815: Only last 50 reports for contact are provided to contact-summary calculation
**Describe the bug**
The `contact-summary` calculation logic is provided with an array of the reports associated with the contact (and any of the contact's `person` children).  Unfortunately, because of the way the code was implemented (at least as far back as `3.9`) the number of reports that would be retrieved for the contact and passed to the `contact-summary` calculation was [capped at 50](https://github.com/medic/cht-core/blob/3e126356afba2c75e68a5c70dab2eb8bf60e65c5/webapp/src/ts/services/search.service.ts#L133).  So, the `contact-summary` for contacts associated with >50 reports, would not have the full context of all that contacts reports.

**To Reproduce**
1. Register a pregnancy for a contact
2. See that a pregnancy card is shown for the patient based on their contact-summary calculation
3. Submit 50 more reports for the contact
4. See that the pregnancy card is no longer shown

**Expected behavior**
The `search.service` gets used in a number of situations where it probably makes sense to have the 50 report limit (e.g. when displaying reports for the contact on the contact's profile page, we probably do not need to show an unlimited number of reports).  However, for the `contact-summary` calculation, for the sake of accuracy, we should be providing all the reports associated with that contact.

Also, the [contact-summary docs](https://docs.communityhealthtoolkit.org/apps/reference/contact-page/) should be updated to note which version lifted the 50 report limit.

**Additional context**
https://github.com/medic/config-muso/issues/1050
[Forum thread](https://forum.communityhealthtoolkit.org/t/is-contact-reports-limited-to-a-certain-number-of-documents/3247/3)

Benmuiruri:
# Description

This PR fixes the unintended bug where the number of reports that would be retrieved for the contact and passed to the `contact-summary` calculation was [capped at 50](https://github.com/medic/cht-core/blob/3e126356afba2c75e68a5c70dab2eb8bf60e65c5/webapp/src/ts/services/search.service.ts#L133). So, the `contact-summary` for contacts associated with >50 reports, would not have the full context of all that contacts reports.

With this fix, for the sake of accuracy, we provide all the reports associated with that contact (limit of 500 reports).

### Before the fix 
<details>
<summary>Video showing created woman has pregnancy card when reports are less than 50</summary>

https://github.com/medic/cht-core/assets/2597305/d2d2b462-c91a-4509-952c-dd41645a2f71

</details>

<details>
<summary>Video showing pregnancy card for woman does not appear when reports than 50 reports and the pregnancy report is not part of the 50</summary>

https://github.com/medic/cht-core/assets/2597305/00f9d879-ab17-41fa-a9a9-fbf37a39c27c

</details>

### After the fix 

Closes #8815 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8815-fix-contact-summary/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8815-fix-contact-summary/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8815-fix-contact-summary/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/tests/e2e/default/contacts/contact-details.wdio-spec.js
+++ b/tests/e2e/default/contacts/contact-details.wdio-spec.js
@@ -45,15 +45,13 @@ describe('Contact details page', () => {
         }
       ));
 
-    const pregnancyReport = [
-      pregnancyFactory.build({
-        fields: {
-          patient_id: patient._id,
-          patient_uuid: patient._id,
-          patient_name: patient.name,
-        },
-      }),
-    ];
+    const pregnancyReport = pregnancyFactory.build({
+      fields: {
+        patient_id: patient._id,
+        patient_uuid: patient._id,
+        patient_name: patient.name,
+      },
+    });
 
     const updatePermissions = async (role, addPermissions, removePermissions = []) => {
       const settings = await utils.getSettings();
@@ -72,8 +70,7 @@ describe('Contact details page', () => {
       await updatePermissions(role, permissions);
 
       await utils.saveDocs([parent, patient]);
-      await utils.saveDocs(reports);
-      await utils.saveDocs(pregnancyReport);
+      await utils.saveDocs([...reports, pregnancyReport]
 
       await utils.createUsers([user]);
 
--- a/tests/e2e/default/contacts/contact-details.wdio-spec.js
+++ b/tests/e2e/default/contacts/contact-details.wdio-spec.js
@@ -98,8 +98,8 @@ describe('Contact details page', () => {
       'validate that the pregnancy card is always displayed', async () => {
 
         expect(await contactPage.pregnancyCard().isDisplayed()).to.be.true;
-        const pregnancyCardInfo = await contactPage.getTestPregnancyCardInfo();
-        expect(pregnancyCardInfo.weeksPregnant).to.equal('23');
+        const pregnancyCardInfo = await contactPage.getPregnancyCardInfo();
+        expect(pregnancyCardInfo.weeksPregnant).to.equal('12');
         expect(pregnancyCardInfo.risk).to.equal('High risk');
       }
     );
--- a/tests/e2e/default/contacts/contact-details.wdio-spec.js
+++ b/tests/e2e/default/contacts/contact-details.wdio-spec.js
@@ -7,6 +7,7 @@ const userFactory = require('@factories/cht/users/users');
 const placeFactory = require('@factories/cht/contacts/place');
 const personFactory = require('@factories/cht/contacts/person');
 const reportFactory = require('@factories/cht/reports/generic-report');
+const pregnancyFactory = require('@factories/cht/reports/pregnancy');
 
 const updateSettings = async (settings) => {
   await utils.revertSettings(true);
@@ -32,13 +33,25 @@ describe('Contact details page', () => {
     const parent = placeFactory.place().build({ _id: 'dist1', type: 'district_hospital' });
     const user = userFactory.build({ username: 'offlineuser', roles: [role] });
     const patient = personFactory.build({ parent: { _id: user.place._id, parent: { _id: parent._id } } });
-    const report = reportFactory
-      .report()
-      .build(
-        { form: 'pregnancy_danger_sign' },
-        { patient, submitter: user.contact, fields: { t_danger_signs_referral_follow_up: 'yes' },
-        }
-      );
+
+    const reports = Array.from({ length: 60 }).map(() => reportFactory.report().build(
+      { form: 'pregnancy_danger_sign' },
+      {
+        patient,
+        submitter: user.contact,
+        fields: { t_danger_signs_referral_follow_up: 'yes' },
+      }
+    ));
+
+    const pregnancyReport = [
+      pregnancyFactory.build({
+        fields: {
+          patient_id: patient._id,
+          patient_uuid: patient._id,
+          patient_name: patient.name,
+        },
+      }),
+    ];
 
     const updatePermissions = async (role, addPermissions, removePermissions = []) => {
       const settings = await utils.getSettings();
@@ -50,11 +63,16 @@ describe('Contact details page', () => {
       await updateSettings({ roles: settings.roles, permissions: settings.permissions });
     };
 
+    const DOCS_DISPLAY_LIMIT = 50;
+
     before(async () => {
       const permissions = ['can_view_contacts', 'can_view_contacts_tab', 'can_view_reports', 'can_view_tasks'];
       await updatePermissions(role, permissions);
 
-      await utils.saveDocs([parent, patient, report]);
+      await utils.saveDocs([parent, patient]);
+      await utils.saveDocs(reports);
+      await utils.saveDocs(pregnancyReport);
+
       await utils.createUsers([user]);
 
       await loginPage.login(user);
@@ -69,9 +87,17 @@ describe('Contact details page', () => {
       expect(await (await contactPage.rhsReportListElement()).isDisplayed()).to.equal(true);
       expect(await (await contactPage.rhsTaskListElement()).isDisplayed()).to.equal(true);
 
-      expect((await contactPage.getAllRHSReportsNames()).length).to.equal(1);
-      expect((await contactPage.getAllRHSTaskNames()).length).to.deep.equal(1);
+      expect((await contactPage.getAllRHSReportsNames()).length).to.equal(DOCS_DISPLAY_LIMIT);
+      expect((await contactPage.getAllRHSTaskNames()).length).to.deep.equal(DOCS_DISPLAY_LIMIT);
     });
+
+    it(
+      'should show contact summary that has the full context for reports > 50' +
+      'validate that the pregnancy card is always displayed', async () => {
+
+        expect(await contactPage.pregnancyCard().isDisplayed()).to.be.true;
+      }
+    );
 
     it('should not show reports when permission is disabled', async () => {
       await updatePermissions(role, [], ['can_view_reports']);
@@ -83,7 +109,7 @@ describe('Contact details page', () => {
       expect(await (await contactPage.rhsReportListElement()).isDisplayed()).to.equal(false);
       expect(await (await contactPage.rhsTaskListElement()).isDisplayed()).to.equal(true);
 
-      expect((await contactPage.getAllRHSTaskNames()).length).to.deep.equal(1);
+      expect((await contactPage.getAllRHSTaskNames()).length).to.deep.equal(DOCS_DISPLAY_LIMIT);
     });
 
     it('should not show tasks when permission is disabled', async () => {
@@ -97,7 +123,7 @@ describe('Contact details page', () => {
       expect(await (await contactPage.rhsReportListElement()).isDisplayed()).to.equal(true);
       expect(await (await contactPage.rhsTaskListElement()).isDisplayed()).to.equal(false);
 
-      expect((await contactPage.getAllRHSReportsNames()).length).to.equal(1);
+      expect((await contactPage.getAllRHSReportsNames()).length).to.equal(DOCS_DISPLAY_LIMIT);
     });
   });
 
--- a/tests/e2e/default/contacts/contact-details.wdio-spec.js
+++ b/tests/e2e/default/contacts/contact-details.wdio-spec.js
@@ -70,7 +70,7 @@ describe('Contact details page', () => {
       await updatePermissions(role, permissions);
 
       await utils.saveDocs([parent, patient]);
-      await utils.saveDocs([...reports, pregnancyReport]
+      await utils.saveDocs([...reports, pregnancyReport]);
 
       await utils.createUsers([user]);
 
--- a/tests/e2e/default/contacts/contact-details.wdio-spec.js
+++ b/tests/e2e/default/contacts/contact-details.wdio-spec.js
@@ -34,14 +34,16 @@ describe('Contact details page', () => {
     const user = userFactory.build({ username: 'offlineuser', roles: [role] });
     const patient = personFactory.build({ parent: { _id: user.place._id, parent: { _id: parent._id } } });
 
-    const reports = Array.from({ length: 60 }).map(() => reportFactory.report().build(
-      { form: 'pregnancy_danger_sign' },
-      {
-        patient,
-        submitter: user.contact,
-        fields: { t_danger_signs_referral_follow_up: 'yes' },
-      }
-    ));
+    const reports = Array
+      .from({ length: 60 })
+      .map(() => reportFactory.report().build(
+        { form: 'pregnancy_danger_sign' },
+        {
+          patient,
+          submitter: user.contact,
+          fields: { t_danger_signs_referral_follow_up: 'yes' },
+        }
+      ));
 
     const pregnancyReport = [
       pregnancyFactory.build({
@@ -96,6 +98,9 @@ describe('Contact details page', () => {
       'validate that the pregnancy card is always displayed', async () => {
 
         expect(await contactPage.pregnancyCard().isDisplayed()).to.be.true;
+        const pregnancyCardInfo = await contactPage.getTestPregnancyCardInfo();
+        expect(pregnancyCardInfo.weeksPregnant).to.equal('23');
+        expect(pregnancyCardInfo.risk).to.equal('High risk');
       }
     );
 
--- a/tests/e2e/default/contacts/contact-details.wdio-spec.js
+++ b/tests/e2e/default/contacts/contact-details.wdio-spec.js
@@ -92,7 +92,7 @@ describe('Contact details page', () => {
 
     it(
       'should show contact summary that has the full context for reports > 50' +
-      'validate that the pregnancy card is always displayed', async () => {
+      ' validate that the pregnancy card is always displayed', async () => {
 
         expect(await contactPage.pregnancyCard().isDisplayed()).to.be.true;
         const pregnancyCardInfo = await contactPage.getPregnancyCardInfo();
--- a/tests/factories/cht/reports/pregnancy.js
+++ b/tests/factories/cht/reports/pregnancy.js
@@ -42,7 +42,7 @@ const defaultFields = {
   'patient_short_name': 'the woman',
   'patient_short_name_start': 'The woman',
   'lmp_date_8601': lmp,
-  'edd_8601': lmp,
+  'edd_8601': '',
   'days_since_lmp': '77',
   'weeks_since_lmp': '34',
   'weeks_since_lmp_rounded': '34',
@@ -70,7 +70,7 @@ const defaultFields = {
     },
     'g_lmp_date_8601': '',
     'g_lmp_date': lmp,
-    'g_edd_8601': lmp,
+    'g_edd_8601': '',
     'g_edd': ''
   },
   'anc_visits_hf': {
@@ -183,7 +183,7 @@ const defaultFields = {
     '__lmp_date': lmp,
     '__lmp_approx_weeks': '34',
     '__lmp_approx_months': '',
-    '__edd': lmp,
+    '__edd': '',
     '__num_previous_anc_hf_visits': '1',
     '__previous_anc_hf_visit_dates': '',
     '__next_anc_hf_visit_date_known': 'no',
--- a/tests/factories/cht/reports/pregnancy.js
+++ b/tests/factories/cht/reports/pregnancy.js
@@ -1,5 +1,6 @@
 const Factory = require('rosie').Factory;
 const { faker } = require('@faker-js/faker');
+const moment = require('moment');
 const uuid = require('uuid');
 const _ = require('lodash');
 const geolocation = require('./geolocation');
@@ -16,15 +17,8 @@ const defaultSubmitter = {
   }
 };
 
-const edd = faker.date
-  .between({ from: '2024-06-01T00:00:00.000Z', to: '2024-06-30T00:00:00.000Z' })
-  .toISOString()
-  .split('T')[0];
-
-const lmp = faker.date
-  .between({ from: '2023-11-01T00:00:00.000Z', to: '2023-11-30T00:00:00.000Z' })
-  .toISOString()
-  .split('T')[0];
+const nextANCVisit = moment().add(2, 'day');
+const lmp = moment().subtract(3, 'months');
 
 const defaultFields = {
   'inputs': {
@@ -48,7 +42,7 @@ const defaultFields = {
   'patient_short_name': 'the woman',
   'patient_short_name_start': 'The woman',
   'lmp_date_8601': lmp,
-  'edd_8601': edd,
+  'edd_8601': lmp,
   'days_since_lmp': '77',
   'weeks_since_lmp': '34',
   'weeks_since_lmp_rounded': '34',
@@ -56,8 +50,8 @@ const defaultFields = {
   'hiv_status_known': 'no',
   'deworming_med_received': '',
   'tt_received': 'no',
-  't_pregnancy_follow_up_date': '',
-  't_pregnancy_follow_up': 'no',
+  't_pregnancy_follow_up_date': nextANCVisit,
+  't_pregnancy_follow_up': 'yes',
   't_danger_signs_referral_follow_up_date': faker.date.recent({ days: 5 }).toISOString(),
   't_danger_signs_referral_follow_up': 'yes',
   'gestational_age': {
@@ -76,7 +70,7 @@ const defaultFields = {
     },
     'g_lmp_date_8601': '',
     'g_lmp_date': lmp,
-    'g_edd_8601': edd,
+    'g_edd_8601': lmp,
     'g_edd': ''
   },
   'anc_visits_hf': {
@@ -189,7 +183,7 @@ const defaultFields = {
     '__lmp_date': lmp,
     '__lmp_approx_weeks': '34',
     '__lmp_approx_months': '',
-    '__edd': edd,
+    '__edd': lmp,
     '__num_previous_anc_hf_visits': '1',
     '__previous_anc_hf_visit_dates': '',
     '__next_anc_hf_visit_date_known': 'no',
--- a/tests/factories/cht/reports/pregnancy.js
+++ b/tests/factories/cht/reports/pregnancy.js
@@ -1,4 +1,5 @@
 const Factory = require('rosie').Factory;
+const { faker } = require('@faker-js/faker');
 const uuid = require('uuid');
 const _ = require('lodash');
 const geolocation = require('./geolocation');
@@ -14,6 +15,16 @@ const defaultSubmitter = {
     }
   }
 };
+
+const edd = faker.date
+  .between({ from: '2024-06-01T00:00:00.000Z', to: '2024-06-30T00:00:00.000Z' })
+  .toISOString()
+  .split('T')[0];
+
+const lmp = faker.date
+  .between({ from: '2023-11-01T00:00:00.000Z', to: '2023-11-30T00:00:00.000Z' })
+  .toISOString()
+  .split('T')[0];
 
 const defaultFields = {
   'inputs': {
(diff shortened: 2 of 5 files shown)


Benmuiruri:
This fix adopts the [limit of reports](https://github.com/medic/cht-core/blob/79e4268072f19c3f9ae9f8ae1ad5a873260c4fb3/webapp/src/ts/modules/reports/reports.component.ts#L67) for a user used in `reports.component.ts` which sets the limit to `500`. 

However, unlike the `reports.component.ts`  this fix does not implement [lazy loading](https://github.com/medic/cht-core/blob/79e4268072f19c3f9ae9f8ae1ad5a873260c4fb3/webapp/src/ts/modules/reports/reports.component.ts#L7) of the reports.  Therefore, it does not fetch reports in batches of 50, instead if a user has 500 reports, the spinner will wait until all reports are fetched before displaying them in the `contact-detail`. 

Below is a performance test on Samsung Galaxy S23+ using a CHW with just 1 household with 5 people [household used in the video above](https://github.com/medic/cht-core/pull/8984#:~:text=has%202%20reports-,Video,-showing%20pregnancy%20card).


<details>
<summary>Performance (Telemetry) for contact with 2 reports</summary>

```
"contact_detail:contact:load": {
      "sum": 2046,
      "min": 86,
      "max": 557,
      "count": 13,
      "sumsqr": 509392
    },
    "contact_detail:contact:load:apdex:satisfied": {
      "sum": 2046,
      "min": 86,
      "max": 557,
      "count": 13,
      "sumsqr": 509392
    },
    "contact_detail:contact:load:contact_data": {
      "sum": 316,
      "min": 12,
      "max": 52,
      "count": 13,
      "sumsqr": 8754
    },
    "contact_detail:contact:load:load_contact_summary": {
      "sum": 36,
      "min": 2,
      "max": 6,
      "count": 13,
      "sumsqr": 120
    },
    "contact_detail:contact:load:load_descendants": {
      "sum": 352,
      "min": 13,
      "max": 82,
      "count": 13,
      "sumsqr": 13754
    },
    "contact_detail:contact:load:load_reports": {
      "sum": 506,
      "min": 14,
      "max": 138,
      "count": 13,
      "sumsqr": 32494
    },
    "contact_detail:contact:load:load_targets": {
      "sum": 179,
      "min": 6,
      "max": 22,
      "count": 13,
      "sumsqr": 2953
    },
    "contact_detail:contact:load:load_tasks": {
      "sum": 645,
      "min": 20,
      "max": 297,
      "count": 13,
      "sumsqr": 99193
    }
```

</details>

<details>
<summary>Performance (Telemetry) for contact with 52 reports</summary>

```
"contact_detail:contact:load": {
      "sum": 11957,
      "min": 167,
      "max": 7325,
      "count": 11,
      "sumsqr": 57385949
    },
    "contact_detail:contact:load:apdex:satisfied": {
      "sum": 4632,
      "min": 167,
      "max": 1607,
      "count": 10,
      "sumsqr": 3730324
    },
    "contact_detail:contact:load:apdex:tolerable": {
      "sum": 7325,
      "min": 7325,
      "max": 7325,
      "count": 1,
      "sumsqr": 53655625
    },
    "contact_detail:contact:load:contact_data": {
      "sum": 1690,
      "min": 15,
      "max": 775,
      "count": 11,
      "sumsqr": 826898
    },
    "contact_detail:contact:load:load_contact_summary": {
      "sum": 79,
      "min": 3,
      "max": 18,
      "count": 11,
      "sumsqr": 825
    },
    "contact_detail:contact:load:load_descendants": {
      "sum": 983,
      "min": 12,
      "max": 448,
      "count": 11,
      "sumsqr": 330535
    },
    "contact_detail:contact:load:load_reports": {
      "sum": 2437,
      "min": 26,
      "max": 1188,
      "count": 11,
      "sumsqr": 1818973
    },
    "contact_detail:contact:load:load_targets": {
      "sum": 1735,
      "min": 32,
      "max": 578,
      "count": 11,
      "sumsqr": 516949
    },
    "contact_detail:contact:load:load_tasks": {
      "sum": 5017,
      "min": 25,
      "max": 4312,
      "count": 11,
      "sumsqr": 18658219
    }
```

</details>

<details>
<summary>Performance (Telemetry) for contact with 490 reports</summary>

```
"contact_detail:contact:load": {
      "sum": 741392,
      "min": 8435,
      "max": 111124,
      "count": 12,
      "sumsqr": 52127429276
    },
    "contact_detail:contact:load:apdex:frustrated": {
      "sum": 732957,
      "min": 45465,
      "max": 111124,
      "count": 11,
      "sumsqr": 52056280051
    },
    "contact_detail:contact:load:apdex:tolerable": {
      "sum": 8435,
      "min": 8435,
      "max": 8435,
      "count": 1,
      "sumsqr": 71149225
    },
    "contact_detail:contact:load:contact_data": {
      "sum": 2650,
      "min": 2650,
      "max": 2650,
      "count": 1,
      "sumsqr": 7022500
    },
    "contact_detail:contact:load:load_contact_summary": {
      "sum": 24,
      "min": 24,
      "max": 24,
      "count": 1,
      "sumsqr": 576
    },
    "contact_detail:contact:load:load_descendants": {
      "sum": 1658,
      "min": 1658,
      "max": 1658,
      "count": 1,
      "sumsqr": 2748964
    },
    "contact_detail:contact:load:load_reports": {
      "sum": 2669,
      "min": 2669,
      "max": 2669,
      "count": 1,
      "sumsqr": 7123561
    },
    "contact_detail:contact:load:load_targets": {
      "sum": 561,
      "min": 561,
      "max": 561,
      "count": 1,
      "sumsqr": 314721
    },
    "contact_detail:contact:load:load_tasks": {
      "sum": 713553,
      "min": 872,
      "max": 101141,
      "count": 12,
      "sumsqr": 48683546167
    }
```

</details>

The performance when loading all `500` reports would be a huge concern (It would be frustrating every time). Given we want to include `ALL` records for accuracy's sake, I am investigating whether it is possible `to use All 500 records to calculate the contact-summary but use lazy loading to load the contacts in batches`.

cc @latin-panda @michaelkohn

michaelkohn:
> The performance when loading all 500 reports would be a huge concern (It would be frustrating every time). Given we want to include ALL records for accuracy's sake, I am investigating whether it is possible to use All 500 records to calculate the contact-summary but use lazy loading to load the contacts in batches.

Not sure if this is any easier but it would probably also be fine to limit the "3 month" view to 50 (since it is selected by default) but have "6 month" and "View all" show everything applicable.

latin-panda:
> This fix adopts the limit of reports for a user used in reports.component.ts which sets the limit to 500.

Why 500 and not another number? Just curious how it was picked

Benmuiruri:
> > This fix adopts the limit of reports for a user used in reports.component.ts which sets the limit to 500.
> 
> Why 500 and not another number? Just curious how it was picked

@latin-panda  as I was exploring how the search function works I came across the limit set in `reports.component.ts`. Git shows you added the `500 limit` 16 months ago. I'd also be interested to know why 500.

latin-panda:
@Benmuiruri If you use git-blame to track down the changes, it will take you to 7y ago ([commit](https://github.com/medic/cht-core/commit/51bea534d9ba0b35e871d05a878ac8f5d45206d6)) and [this is the conversation](https://github.com/medic/cht-core/issues/3605#issuecomment-313297190) how they came up with this number, previously the report list page was loading 1000 reports but that's too much for the bulk delete (different feature as this one).  

Bulk delete isn't related to what we're doing here. I'm unsure if this number is still relevant for this case - is it too high or too low?

Benmuiruri:
I have adopted @michaelkohn suggestion. 
- When user clicks on the 3 months or 6 month filter for the `Reports` they can only see 50 reports. 
- Same for tasks, when the user clicks 1 week or 2 weeks filter for `Tasks`, they can only see 50 tasks. 

I added the constraint for `Tasks` for consistency. In master, if a contact has 400 tasks and 400 reports. The contact detail page displays all 400 tasks but only 50 reports (whether there is a filter of the reports or not, you cannot see reports > 50). 

With my implementation, if a contact has 400 tasks and 400 reports. 
1. The filtered `Tasks` when 1 week or 2 weeks is selected will display 50 tasks, when `View all` is selected, it will display all 400 tasks. 
2. The filtered `Reports` when 3 months or 6 months is selected will display 50 reports, when `View all` is selected, it will display all 400 reports. 

Two points for discussion. 
1. @latin-panda I am also unsure whether `500` is the right number, perhaps the answer might be in looking at how many reports users typically have in `PostgreSQL` to come up with the number? @michaelkohn.  
2. We will need to open another issue to implement `pagination` / `lazy loading` when `View all` is selected.

michaelkohn:
FWIW... I don't have a problem setting a max of 500, though if it were super easy to implement, it would probably be better to show 50 and if there are more than 50, have a "Show more" link that redirects them to the Reports page, prefiltered to that Contact. It wouldn't be possible to do that for Tasks since the Tasks page doesn't have a filter, but it's also a lot less likely to have that many tasks (unless something is mis-configured).

Benmuiruri:
Hi @michaelkohn based on your comment, here is my suggestion. 
1. Let's limit the scope of this task to fix the bug of including `All Reports` when calculating `contact-summary`.
2. Let's agree on what number represents `All Reports`. Currently I have set it to `500`. 
3. Let's limit the display of Reports and Tasks on the Contact Detail page to 50. Regardless of whether someone has clicked  3 months, 6 months or View All. (The reason to limit is we do not want the app to freeze attempting to fetch `All Reports` )
4. Open another issue to implement efficiently loading `All Reports` if user clicks `View All`

michaelkohn:
> Let's limit the scope of this task to fix the bug of including All Reports when calculating contact-summary.

Good idea

> Let's agree on what number represents All Reports. Currently I have set it to 500.

Feels weird to call it "All Reports" when it's actually "max reports" (since we are limiting it to 500). FWIW, from a functional perspective I think 500 is totally fine. I ran some quick queries and in one big project only `0.2%` of patients had more than 500 reports associated to them and that's server-side (and probably not intentional) so I'd imagine client-side, after purging, there are even fewer patients, if any, that have more than 500.

> Let's limit the display of Reports and Tasks on the Contact Detail page to 50. Regardless of whether someone has clicked 3 months, 6 months or View All. (The reason to limit is we do not want the app to freeze attempting to fetch All Reports )

Cool.

> Open another issue to implement efficiently loading All Reports if user clicks View All

I appreciate the awareness around scope-creep on this ticket, but we aren't going to do the efficient loading work unless we hear the need so I'd actually rather just not even create the ticket since it'll just sit there open/un-addressed until we hear about it.

Benmuiruri:
Hi @ralfudx when you get some time, please review this, let's try and merge it this week. Thanks.

ralfudx (approved):
@Benmuiruri it looks good to me

Benmuiruri:
Hi @latin-panda given Rafa's QA approval. Please give this PR a once over before I merge it. Thanks.

tests/e2e/default/contacts/contact-details.wdio-spec.js
```diff
@@ -69,10 +87,18 @@ describe('Contact details page', () => {
...
-      expect((await contactPage.getAllRHSTaskNames()).length).to.deep.equal(1);
+      expect((await contactPage.getAllRHSReportsNames()).length).to.equal(DOCS_DISPLAY_LIMIT);
+      expect((await contactPage.getAllRHSTaskNames()).length).to.deep.equal(DOCS_DISPLAY_LIMIT);
     });
 
+    it(
+      'should show contact summary that has the full context for reports > 50' +
+      'validate that the pregnancy card is always displayed', async () => {
+
+        expect(await contactPage.pregnancyCard().isDisplayed()).to.be.true;
```
latin-panda, line 98: At least we should check for 3 fields to ensure this card is taking the reports passed.
Fields like: `High risk`, `completed ANC clinic`, `current danger signs`

webapp/src/ts/modules/contacts/contacts-content.component.ts
```diff
@@ -238,14 +239,17 @@ export class ContactsContentComponent implements OnInit, OnDestroy {
 
     const allReports = reports || this.selectedContact?.reports || [];
     this.filteredReports = allReports
-      .filter((report) => !reportStartDate || reportStartDate.isBefore(report.reported_date));
+      .filter((report) => !reportStartDate || reportStartDate.isBefore(report.reported_date))
+      .slice(0, this.DISPLAY_LIMIT);
```
latin-panda, line 243: This is just a note that since before [we were displaying only 50 records](https://github.com/medic/cht-core/blob/c4c3114ed8e17f49a78e1f803a643792b71f2c8e/webapp/src/ts/services/search.service.ts#L133) in all the tabs (3m, 6m, view all)

latin-panda (changes requested):
Well done! Just a feedback below

webapp/src/ts/modules/contacts/contacts-content.component.ts
latin-panda: @Benmuiruri, thinking about this more, it would help in performance if it stops the loop when having 50 reports in the date range, instead of processing 500 reports. 

```
newFilteredReports = [];
allReports.forEach(report => {
  if (!reportStartDate || reportStartDate.isBefore(report.reported_date)) {
     newFilteredReports.push(report)
  }
  if (filteredReports > this.DISPLAY_LIMIT) {
     return;
  }
);

this.filteredReports = newFilteredReports

```

tests/factories/cht/reports/pregnancy.js
```diff
@@ -1,4 +1,5 @@
 const Factory = require('rosie').Factory;
+const { faker } = require('@faker-js/faker');
```
lorerod, line 2:
```suggestion
const { faker } = require('@faker-js/faker');
const moment = require('moment');
```

tests/factories/cht/reports/pregnancy.js
```diff
@@ -15,6 +16,16 @@ const defaultSubmitter = {
   }
 };
 
+const edd = faker.date
```
lorerod, line 19: This is not necessary if you define `lpm`

tests/factories/cht/reports/pregnancy.js
```diff
@@ -15,6 +16,16 @@ const defaultSubmitter = {
...
+  .between({ from: '2024-06-01T00:00:00.000Z', to: '2024-06-30T00:00:00.000Z' })
+  .toISOString()
+  .split('T')[0];
+
+const lmp = faker.date
+  .between({ from: '2023-11-01T00:00:00.000Z', to: '2023-11-30T00:00:00.000Z' })
+  .toISOString()
+  .split('T')[0];
```
lorerod, line 27:
```suggestion
const lmp = moment().subtract(3, 'months');
```
It can be challenging to determine the number of weeks pregnant based on a random date.

tests/factories/cht/reports/pregnancy.js
```diff
@@ -47,8 +58,8 @@ const defaultFields = {
   'tt_received': 'no',
   't_pregnancy_follow_up_date': '',
   't_pregnancy_follow_up': 'no',
```
lorerod, line 60:
```suggestion
  't_pregnancy_follow_up': 'yes',
```

tests/factories/cht/reports/pregnancy.js
```diff
@@ -47,8 +58,8 @@ const defaultFields = {
   'tt_received': 'no',
   't_pregnancy_follow_up_date': '',
```
lorerod, line 59:
```suggestion
  't_pregnancy_follow_up_date': nextANCVisit,
```

tests/factories/cht/reports/pregnancy.js
```diff
@@ -15,6 +16,16 @@ const defaultSubmitter = {
   }
 };
```
lorerod, line 18:
```suggestion
const nextANCVisit = moment().add(2, 'day');

```

tests/e2e/default/contacts/contact-details.wdio-spec.js
```diff
@@ -69,10 +89,21 @@ describe('Contact details page', () => {
...
+        const pregnancyCardInfo = await contactPage.getTestPregnancyCardInfo();
+        expect(pregnancyCardInfo.weeksPregnant).to.equal('23');
```
lorerod, line 102:
```suggestion
        expect(pregnancyCardInfo.weeksPregnant).to.equal('12');
```

tests/e2e/default/contacts/contact-details.wdio-spec.js
```diff
@@ -69,10 +89,21 @@ describe('Contact details page', () => {
...
+      expect((await contactPage.getAllRHSReportsNames()).length).to.equal(DOCS_DISPLAY_LIMIT);
+      expect((await contactPage.getAllRHSTaskNames()).length).to.deep.equal(DOCS_DISPLAY_LIMIT);
     });
 
+    it(
+      'should show contact summary that has the full context for reports > 50' +
+      'validate that the pregnancy card is always displayed', async () => {
+
+        expect(await contactPage.pregnancyCard().isDisplayed()).to.be.true;
+        const pregnancyCardInfo = await contactPage.getTestPregnancyCardInfo();
```
lorerod, line 101:
```suggestion
        const pregnancyCardInfo = await contactPage.getPregnancyCardInfo();
```
With the changes suggested in the pregnancy factory you should be able to use the original method.

lorerod (changes requested):
The main problem was that you needed to define a date for `t_pregnancy_follow_up_date` and set `t_pregnancy_follow_up` to `yes` to be able to see the `Next ANC clinic visit` field.
I left some suggestions inline, could you please have a look and let me know what you think?
As a QA team, we should improve these factories to make them flexible for developers to create your desired scenarios.

Benmuiruri:
Thank you Lorena, that worked. These reports can be tricky, thanks for spotting the missing factor.

tests/e2e/default/contacts/contact-details.wdio-spec.js
```diff
@@ -32,13 +33,27 @@ describe('Contact details page', () => {
...
-        { patient, submitter: user.contact, fields: { t_danger_signs_referral_follow_up: 'yes' },
+        {
+          patient,
+          submitter: user.contact,
+          fields: { t_danger_signs_referral_follow_up: 'yes' },
         }
-      );
+      ));
+
+    const pregnancyReport = [
```
dianabarsan, line 48: Unsure why this is built as an array only containing one element. Are more docs pushed later?

webapp/src/ts/modules/contacts/contacts-content.component.ts
```diff
@@ -54,7 +54,8 @@ export class ContactsContentComponent implements OnInit, OnDestroy {
   relevantReportForms;
   childContactTypes;
   filteredTasks = [];
-  filteredReports = [];
+  filteredReports: any[] = [];
+  DISPLAY_LIMIT = 50;
```
dianabarsan, line 58: I'm unsure about this. Does this mean that we never display more than 50 reports / tasks in the contact detail page and there is no way that this limit is exceeded?

webapp/src/ts/services/contact-view-model-generator.service.ts
```diff
@@ -309,7 +311,7 @@ export class ContactViewModelGeneratorService {
     });
     const filter = { subjectIds: _flattenDeep(subjectIds) };
     return this.searchService
-      .search('reports', filter, { include_docs: true })
+      .search('reports', filter, { include_docs: true, limit: this.LIMIT_SELECT_ALL_REPORTS })
```
dianabarsan, line 314: I think 500 is fine, we already use the limit of 500 in the selectall reorts (https://github.com/medic/cht-core/blob/c4c3114ed8e17f49a78e1f803a643792b71f2c8e/webapp/src/ts/modules/reports/reports.component.ts#L67).

tests/e2e/default/contacts/contact-details.wdio-spec.js
Benmuiruri: No, there are no more docs pushed to it, that one element creates the pregnancy card that is tested later in the test.

webapp/src/ts/modules/contacts/contacts-content.component.ts
Benmuiruri: Yes, that is correct. 
- The scope of this issue was to use all reports to calculate the contact-summary. 
- By increasing the limit of reports fetched to 500, we now use "All 500 Reports" to calculate the contact summary 
- A sideeffect of fetching all 500 reports is when a user clicks on View All in the contact detail page, we do not have pagination or lazy loading, the app might freeze as it fetches all 500. 
- Therefore, we limited the number of reports to 50 whether you are in the 3 month, 6 month or View all tab.  If a user wants to View All reports they have to visit the Reports tab. 
- We could create another issue to add pagination for the View All tab to efficiently load All reports there.

dianabarsan: Cool, thanks for the explanation. I agree that we need a new issue to add pagination to tasks / reports on contact detail pages.

tests/e2e/default/contacts/contact-details.wdio-spec.js
```diff
@@ -32,13 +33,27 @@ describe('Contact details page', () => {
...
+      pregnancyFactory.build({
+        fields: {
+          patient_id: patient._id,
+          patient_uuid: patient._id,
+          patient_name: patient.name,
+        },
+      }),
+    ];
```
latin-panda, line 56:
```suggestion
    const pregnancyReport = pregnancyFactory.build({
        fields: {
          patient_id: patient._id,
          patient_uuid: patient._id,
          patient_name: patient.name,
        },
      });
```
Because it only needs one pregnancy report and the variable name is singular, it seems better to leave it like this because no more reports will be added to this variable.

tests/e2e/default/contacts/contact-details.wdio-spec.js
```diff
@@ -50,11 +65,16 @@ describe('Contact details page', () => {
...
+    const DOCS_DISPLAY_LIMIT = 50;
+
     before(async () => {
       const permissions = ['can_view_contacts', 'can_view_contacts_tab', 'can_view_reports', 'can_view_tasks'];
       await updatePermissions(role, permissions);
 
-      await utils.saveDocs([parent, patient, report]);
+      await utils.saveDocs([parent, patient]);
+      await utils.saveDocs(reports);
+      await utils.saveDocs(pregnancyReport);
```
latin-panda, line 76:
```suggestion
      await utils.saveDocs([ ...reports, pregnancyReport ]);
```
Then, you can save all reports in 1 request.

tests/factories/cht/reports/pregnancy.js
```diff
@@ -15,6 +17,9 @@ const defaultSubmitter = {
...
 
+const nextANCVisit = moment().add(2, 'day');
```
latin-panda, line 20: This is the date when the CI runs and adds two days, right? 

There's a possibility that `nextANCVisit` is next month. If the current date is 30/12/2024, then `nextANCVisit` is 01/01/2025. 
Will that make any of the e2e tests fail? We want to avoid blocking devs because all the PR's CI fails at the end of the month.

webapp/src/ts/modules/contacts/contacts-content.component.ts
```diff
@@ -237,15 +238,35 @@ export class ContactsContentComponent implements OnInit, OnDestroy {
...
+  }
+
+  private getFilteredReports(allReports: any[], reportStartDate, displayLimit): any[] {
+    const filteredReports: any[] = [];
+    for (const report of allReports) {
+      if (filteredReports.length >= displayLimit) {
+        break;
+      }
+
+      if (this.validateReport(report, reportStartDate)) {
```
latin-panda, line 252:
```suggestion
      if (reportStartDate?.isBefore(report.reported_date)) {
```
Can you do this and then delete `validateReport `?

latin-panda (changes requested):
Thanks! A couple of suggestions. Remember to request review to Lorena, otherwise GH won't let you merge this PR

tests/e2e/default/contacts/contact-details.wdio-spec.js
Benmuiruri: Thanks, now I understand Diana's  comment

tests/factories/cht/reports/pregnancy.js
Benmuiruri: No it won't fail. I am not checking the value of the `nextANCVisit` in the `contact-details.wdio-spec.js`.  The value was added so that calling [getPregnancyCardInfo](https://github.com/medic/cht-core/blob/c2393c7ec0beda3802e856a825fb7521e759bee3/tests/page-objects/default/contacts/contacts.wdio.page.js#L314) in the test does not fail

tests/e2e/default/contacts/contact-details.wdio-spec.js
```diff
@@ -69,10 +86,21 @@ describe('Contact details page', () => {
...
 
-      expect((await contactPage.getAllRHSReportsNames()).length).to.equal(1);
-      expect((await contactPage.getAllRHSTaskNames()).length).to.deep.equal(1);
+      expect((await contactPage.getAllRHSReportsNames()).length).to.equal(DOCS_DISPLAY_LIMIT);
+      expect((await contactPage.getAllRHSTaskNames()).length).to.deep.equal(DOCS_DISPLAY_LIMIT);
     });
 
+    it(
+      'should show contact summary that has the full context for reports > 50' +
+      'validate that the pregnancy card is always displayed', async () => {
```
lorerod, line 95:
```suggestion
      ' validate that the pregnancy card is always displayed', async () => {
```

tests/factories/cht/reports/pregnancy.js
```diff
@@ -36,19 +41,19 @@ const defaultFields = {
   'patient_name': 'Adelia Akiyama',
   'patient_short_name': 'the woman',
   'patient_short_name_start': 'The woman',
-  'lmp_date_8601': '2021-02-02',
-  'edd_8601': '2021-11-09',
+  'lmp_date_8601': lmp,
+  'edd_8601': lmp,
```
lorerod, line 45: You can leave `edd` fields without value.

tests/factories/cht/reports/pregnancy.js
```diff
@@ -63,10 +68,10 @@ const defaultFields = {
       'edd_note': '',
       'edd_check_note': ''
     },
-    'g_lmp_date_8601': '2021-02-02',
-    'g_lmp_date': '2 Feb, 2021',
-    'g_edd_8601': '2021-11-09',
-    'g_edd': '9 Nov, 2021'
+    'g_lmp_date_8601': '',
+    'g_lmp_date': lmp,
+    'g_edd_8601': lmp,
```
lorerod, line 73: same here, please leave edd fields without value

tests/factories/cht/reports/pregnancy.js
```diff
@@ -175,15 +180,15 @@ const defaultFields = {
   'data': {
     '__lmp_method': 'method_approx',
     '__no_lmp_registration_reason': '',
-    '__lmp_date': '2021-02-02',
+    '__lmp_date': lmp,
     '__lmp_approx_weeks': '34',
     '__lmp_approx_months': '',
-    '__edd': '2021-11-09',
+    '__edd': lmp,
```
lorerod, line 186: please leave edd fields without value.

lorerod (approved):
Thanks @Benmuiruri for adding my suggestions.

latin-panda (approved):
LGTM! Thanks

michaelkohn:
@Benmuiruri can you please update the docs as well https://docs.communityhealthtoolkit.org/apps/reference/contact-page/

Issue #8987, opened 2024-04-02
Add additional automation to the release process


jkuester:
**Describe the issue**
Our current [release process](https://github.com/medic/cht-core/blob/master/.github/ISSUE_TEMPLATE/z_release_major.md) contains a number of manual steps that are very time consuming and error prone.  We should automate as many of these steps as possible to reduce manual errors, reduce engineering time wasted on the release process, and improve the speed of the release process. All of this will be very helpful in increasing our release cadence. 

**Describe the improvement you'd like**
- [ ] Automatically verify that all commits in the release have an issue/PR included in the Milestone - could be done in the [release notes script](https://github.com/medic/cht-core/blob/master/scripts/release-notes/index.js).
- [ ] Improve handling of the Scalability Test results (many core eng do not have access to the S3 bucket). We could automatically open a PR to add the results into `tests/scalability/previous_results`.
- [ ] Updating the demo instance and pushing the latest config should be something that can be done automatically in a GitHub Action (triggered by the release build).

jkuester:
Another thought I had based on https://github.com/medic/cht-core/pull/8989 is that the commit milestone validation logic could also be included in a CI workflow job for all PRs.  So, basically, when you open a PR, it will check the PR, any linked issues, and your PR title to see if there is a milestone configured. If not, the CI will fail for the PR.

garethbowen:
@jkuester An alternative approach would be more like what cht-conf does which does not use milestones at all. Instead it generates the release notes from the commits and adds a comment to all issues with the release version. As we tend towards smaller and smaller releases I think this is the direction we'll ultimately go in. If that's the case the validation we need is to enforce the PR title and/or commit message complies with the common format so we can parse it programmatically.

jkuester:
Absolutely!  FTR, we implemented a [CI check for PR titles](https://github.com/medic/cht-watchdog/blob/main/.github/workflows/conventional-commits.yml) in watchdog using `commitlint` that works well!  Would be a good place to start when looking to do something similar in cht-core, etc.

dianabarsan:
I believe that only updating the demo instance remains. And we have really detailed instructions in the release template for that. @jkuester do you still think we should do the upgrade in a GH action, especially since it requires running helm commands against our dev cluster?

jkuester:
IMHO, I would _much_ rather just script the helm commands in a GitHub action instead of any developer who does a release of cht-core needing AWS access.  🤷 

[...]

PR #8989, opened 2024-04-03
chore(#8987): validate commit milestones for release

jkuester:
# Description

#8987

Adds validation to the release-notes generation script to validate that all of the commits in the release have an associated milestone.  This is done by checking for a milestone on:
- Any PR associated with the commit
- Any issue closed by a PR associated with the commit
- Any issue/PR whose number is included in the commit message

It is not absolutely air-tight validation, but it is probably more thorough/accurate than any kind of manual validation of the commits (e.g. the release engineer going through each commit to validate it has a milestone). The actual _milestone values_ are not checked. This logic just ensures that each commit is somehow associated with a milestone.

I also added a new `--skip-commit-validation` flag as a backstop so this validation logic will not block the main functionality of the script.  (Maybe there is a commit that we do not care about having in a milestone and it is not worth going back to fix it.)

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8987_validate_commit_milestones/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8987_validate_commit_milestones/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8987_validate_commit_milestones/docker-compose/cht-couchdb-clustered.yml)

# License


scripts/release-notes/index.js
```diff
@@ -1,30 +1,43 @@
...
 const octokit = new Octokit({
   auth: require('../token.json').githubApiToken,
   userAgent: 'cht-release-note-generator',
 });
+const authenticatedGraphql = graphql.defaults({
+  headers: {
+    authorization: `token ${require('../token.json').githubApiToken}`,
+    userAgent: 'cht-release-note-generator',
+  },
+});
```
m5r, line 13: It would be interesting to replace REST API calls so that we have only one way of interacting with GitHub's API

There are only the two calls on [line 203](https://github.com/medic/cht-core/blob/8987_validate_commit_milestones/scripts/release-notes/index.js#L203) and [line 237](https://github.com/medic/cht-core/blob/8987_validate_commit_milestones/scripts/release-notes/index.js#L237).  
I looked into them and we could replace both of them with a single gql query like this:
```js
const getMilestoneIssues = async () => {
  const response = await gitHubRepoQuery(`
    milestones(query: "${MILESTONE_NAME}", first: 1) {
      nodes {
        issues(first: 100, states: [OPEN, CLOSED]) {
          nodes {
            number
            url
            title
            labels(first: 100) {
              nodes {
                name
              }
            }
          }
        }
      }
    }
  `);
  return response.repository.milestones.nodes[0].issues.nodes;
};
```

The only thing I don't like about this solution is the lack of pagination handling. GitHub provides [this utility package](https://www.npmjs.com/package/@octokit/plugin-paginate-graphql) to have the same pagination DX their REST client has with `octokit.paginate(route, params)`

jkuester: Turns out nested pagination (and apparently querying `issues` inside of `milestones` counts as nested pagination) is frowned upon and [not supported](https://github.com/octokit/plugin-paginate-graphql.js/tree/v4.0.1?tab=readme-ov-file#unsupported-nested-pagination) by `plugin-paginate-graphql`, so I had to keep an extra query to get the milestone number, but otherwise this works great!

jkuester:
I forgot to elucidate some tests steps here!  

- `cht-core 4.6.0` should successfully return the release notes. Both the commits and the issues pass validation.
- `cht-core 4.7.0` should fail commit validation for a long list of commits (most of these can be ignored and are coming from a merge commit.  Hopefully will not have this problem in future releases since we have [disabled merge commits](https://medic.slack.com/archives/C024KTGRW/p1712933329462349).  (Things like this are also why I included the `--skip-commit-validation` flag...)
- `cht-core 4.7.0 --skip-commit-validation` should fail because a couple of the milestone issues do not have the proper type label on them
- `cht-core invalid-milestone` should fail because the milestone is not found.

I think these four runs do a pretty good job of checking the major code flows/functionality.

scripts/release-notes/index.js
```diff
@@ -49,11 +56,138 @@ const PREFIXES_TO_IGNORE = [
...
+    `defaultBranchRef { name }`
+  ).then(({ repository }) => repository.defaultBranchRef.name);
+};
+
+// This query calculates the "commits for the release" by comparing the commit history of the latest release with the
+// commit history of the milestone branch and keeping only the commits that are unique to the milestone branch.
+const getCommitsForRelease = async (release, milestoneBranch) => queryRepoPaginated(
+  `ref(qualifiedName: "${release}") {
+      compare(headRef: "${milestoneBranch}") {
+        commits(first: 100, after: $cursor) {
```
m5r, line 95: That's a much better way to get commits for a given release than what I had :sunglasses: :tada:

m5r (approved):
Looks great to me! :heavy_check_mark:

Issue #8990, opened 2024-04-08
Add .gitattributes file for preventing file ending mismatch on different OSes as well as marking certain binaries as binaries for diffing purposes for LFS


fardarter:
**What feature do you want to improve?**

Feature does not yet exist

**Describe the improvement you'd like**

I'd like the addition of a .gitattributes file for preventing file ending mismatch on different OSes as well as marking certain binaries as binaries for diffing purposes for LFS

**Describe alternatives you've considered**

I have not considered alternatives as this is a git standard: https://git-scm.com/docs/gitattributes

**Additional context**

I support a team that operates on WSL + Windows and it is annoying when the files endings get changed by Windows.

PR #8991, opened 2024-04-08
chore(#8990): adds .gitattributes file

fardarter:
Addresses issue: https://github.com/medic/cht-core/issues/8990

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,19 @@
+# Souce: https://rehansaeed.com/gitattributes-best-practices/#line-endings
+
+###############################
+# Git Line Endings            #
+###############################
+
+# Set default behaviour to automatically normalize line endings.
+* text=auto eol=lf
+
+# Force batch scripts to always use CRLF line endings so that if a repo is accessed
+# in Windows via a file share from Linux, the scripts will work.
+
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,57 @@
...
+*.pdf binary
+*.png binary
+*.psd binary
+*.webp binary
+
+# Fonts
+*.woff2 binary
+
+# Other
+*.exe binary
```
garethbowen, line 57: I think we can drop this. Git is smart enough to work this out and means we don't need to manage it. As your linked blog post says...

> ...relying on Git to auto-detect binary files (yes Git is clever enough to do that)

garethbowen (changes requested):
Some simplification recommended but the line ending stuff is great.

.gitattributes
fardarter: We've seen xlsx get diffed on our repo, which is why I lost trust in the auto detect a little. Including it in the file solved that, but will make the change regardless.

fardarter: @garethbowen have made the change

garethbowen (approved):
Lovely!

Issue #8997, opened 2024-04-09
Update browserlist to Chome 110+ or perhaps a last 10 version declaration


fardarter:
Using the issue format to ask a question. I see the current browserlist config has Chrome 90 as a target. Is there a reason that FF is a rolling window but Chrome is a fixed version?

```
# Browsers that we support

Chrome >= 90
last 4 Firefox versions
```

@garethbowen at risk of telling you something you already know, your bundler may be using the browserlist config to determine your target JS version, and with 2021 being the year of release for v90, you may be shipping less efficient/larger code that you desire. If it's already known and there is a reason, maybe a comment in the file could be helpful.

garethbowen:
That matches our documented support matrix here: https://docs.communityhealthtoolkit.org/core/releases/#dependencies

The reason Chrome is set to 90 is because the CHT is mostly used in a webview which is almost always the Android Webview which is Chromium based. Because many android phones are not set to autoupdate we know that lots use old versions of webview. We will continue to progressively increase the minimum browser version when possible but it needs careful communication and is quite expensive for partners to verify which phones need updating and which don't.

Firefox on the other hand is realistically only used on desktops which are much better at keeping up to date so we can safely stick to a rolling version there.

Issue #8936, opened 2024-03-15
Flaky test Service worker cache branding updates trigger login page refresh


dianabarsan:
**Describe the issue**
lowlevel wdio builds frequently fail with: 

```
[chrome-headless-shell 122.0.6261.111 linux #0-1] 1) Service worker cache branding updates trigger login page refresh
[chrome-headless-shell 122.0.6261.111 linux #0-1] element ("#header-dropdown li.sync-status .success") still not displayed after 20000ms
[chrome-headless-shell 122.0.6261.111 linux #0-1] Error: element ("#header-dropdown li.sync-status .success") still not displayed after 20000ms
[chrome-headless-shell 122.0.6261.111 linux #0-1]     at async syncAndWaitForSuccess (/home/runner/work/cht-core/cht-core/tests/page-objects/default/common/common.wdio.page.js:327:3)
[chrome-headless-shell 122.0.6261.111 linux #0-1]     at async Object.sync (/home/runner/work/cht-core/cht-core/tests/page-objects/default/common/common.wdio.page.js:348:3)
[chrome-headless-shell 122.0.6261.111 linux #0-1]     at async Context.<anonymous> (/home/runner/work/cht-core/cht-core/tests/e2e/default/service-worker/service-worker.wdio-spec.js:151:5)
```

https://github.com/medic/cht-core/actions/runs/8211973126/job/22640392518?pr=8921#step:17:464 

**Describe the improvement you'd like**
stabilize the test

dianabarsan:
From artifact, it looks like the app is stuck while syncing and not updating the status: 
![image](https://github.com/medic/cht-core/assets/35681649/d2414e8b-d0a9-4e76-bc50-d56e34259e90)

dianabarsan:
Browser logs reports no errors: 
```
~~~~~~~~~~~~~ confirm initial list of cached resources ~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~~~~~ branding updates trigger login page refresh ~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~~~~~ branding updates trigger login page refresh ~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~~~~~ branding updates trigger login page refresh ~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~~~~~ branding updates trigger login page refresh ~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #8999, opened 2024-04-10
doc(#8997): adds explanatory comment for .browserslistrc settings

fardarter:
I have added a rationale comment for the .browserslistrc file. 

Addresses issue https://github.com/medic/cht-core/issues/8997

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,4 +1,16 @@
 # Browsers that we support
+#
+# Chrome is set to a lagging value because the CHT is mostly used in a webview. Webviews are
+# almost always the Android Webview (Chromium). 
+#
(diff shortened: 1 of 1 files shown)


fardarter:
@garethbowen I see you closed the issue but thought this was still valuable.

garethbowen (approved):

garethbowen:
Good call, thanks. I think I'll also add that rationale to the docs site when I'm back in the office.

fardarter:
> Good call, thanks. I think I'll also add that rationale to the docs site when I'm back in the office.

@garethbowen can't imagine why this would cause a test failure. Would you be able to rerun the failing test?

garethbowen:
Looks like this known issue: https://github.com/medic/cht-core/issues/8936

Issue #9000, opened 2024-04-10
Add eslint fix command to `package.json` as a convenience script for developers preparing pull requests.


fardarter:
Add eslint fix command to `package.json` as a convenience script for developers preparing pull requests.

Eg,    `"lint-fix": "eslint --color --cache . --fix"`

garethbowen:
@fardarter I'm not a big fan of this approach because you end up with a whole lot of lint commands which is really complicated. Instead you can pass parameters directly to the command with the `--` option, eg: `npm run lint -- --fix`.

The problem with this at the moment is the `lint` command actually does two things and the `--fix` gets appended so applied to the second thing and not the `eslint` command. The easy fix is to reverse the order of commands, eg:

```
    "lint": "./scripts/build/blank-link-check.sh && eslint --color --cache .",
```

This then also supports passing other parameters through to the eslint command. What do you think? Is this an easier solution?

fardarter:
My experience is that most developers don't know how to pass args to npm scripts. In general, I like to give devs easy tools to prepare a PR, as it makes my life easier when I have to give feedback. A preset script can be helpful there.

I also like keeping operations discrete and then bundling them into a sort of prepare PR command. This helps with speed, as people can make some fixes quickly, while still being able to run the overall script if they want. 

Personally I like commit hooks but every team I've worked with has voted against them, and they do cause overhead, so I rather try make fixing things easier.

If you're still unpersuaded, then let's swap the calls around. Just not easy to leave a comment in `.json` to explain why the lint must always be at the end. Also, if you introduce other tooling you may end up with tooling conflicts if you actually need them to run after lint.

garethbowen:
> In general, I like to give devs easy tools to prepare a PR

Definitely, but if you fill the package with scripts with various combinations of flags then developers can't find the tools and they're useless. Keep it simple!

PR #9001, opened 2024-04-10
chore(#9000): add convenience eslint tooling script to package.json

fardarter:
Adds convenience eslint tooling script to package.json

Closes issue: https://github.com/medic/cht-core/issues/9000


garethbowen:
This adds complexity by adding an extra build script for an edge case that is easy to run directly. A possible alternative is discussed in the issue.

PR #9014, opened 2024-04-17
DO NOT MERGE - Contact Details' Load Tasks

latin-panda:
# Description

# Code review checklist

# License


webapp/src/ts/services/tasks-for-contact.service.ts
```diff
@@ -82,8 +95,11 @@ export class TasksForContactService {
         }
 
         const contactIds = this.getIdsForTasks(model);
+        const docs = this.getDocsForTasks(model);
+        console.warn('TasksForContactService.get - model', model);
+        console.warn('TasksForContactService.get - contact docs', docs);
         return this.rulesEngineService
-          .fetchTaskDocsFor(contactIds)
+          .fetchTaskDocsFor(contactIds, docs)
```
latin-panda, line 102: Hi @dianabarsan! I opened this PR to show you an idea; this code is not even close to being ready for prod—still a playground.  

When loading Contact Details, [the loadTasks](https://github.com/medic/cht-core/blob/04491c9f93c4689c4207928a339347e863c1d52f/webapp/src/ts/effects/contacts.effects.ts#L181) sends the selected contact to this function. It contains children, lineage, and its doc.  
I noticed those are the same docs that [pouch provider is fetching from the db ](https://github.com/medic/cht-core/blob/04491c9f93c4689c4207928a339347e863c1d52f/shared-libs/rules-engine/src/pouchdb-provider.js#L130) when there are dirty contacts.

What do you think about splitting the process in `this.rulesEngineService.fetchTaskDocsFor(contactIds)`
           and `this.rulesEngineService.fetchTaskDocsForContactDocs(contactDocs)`? Then `contact.effects` sends the docs using the later function

dianabarsan: I'm thinking of assigning myself as a reviewer, otherwise there's a risk I forget about this. Do you mind?

latin-panda: Sure that's okay! Thanks for your time

dianabarsan: I think there might be an issue with a race condition here. 

The rules-engine has a queue for executable jobs, so that we don't end up starting multiple task calculation processes in parallel. This will mean that, if the queue is already populated, we could be executing the task refresh for this list of contacts a little bit later than this call, and at that time, there's a chance that some of the contacts in the list were edited and marked dirty, and we could be calculating tasks for old versions of these docs, and marking the docs as clean, preventing correct tasks from being calculated. 

How resource intensive is the allDocs in rules engine? Is it worth to risk this hard to track behavior for this gain in performance?

PR #9016, opened 2024-04-17
feat(#8986) add api for getting single user

### LINKED ISSUE #8986: /api/v2/users look up data for single user
Split off from: https://github.com/medic/cht-core/issues/8877

The goal is to add api support for retrieving data about a single user.

`GET /api/v2/users/username`

jkuester:
# Description

Closes #8986

# Code review checklist

# License


shared-libs/user-management/test/unit/users.spec.js
```diff
@@ -3020,14 +3149,27 @@ describe('Users service', () => {
       sinon.stub(roles, 'isOffline').returns(false);
 
       const updates = { token_login: true, phone: '+40 755 89-89-89' };
-      db.medic.get.resolves({
+      db.medic.get.onFirstCall().resolves({
```
jkuester, line 3152: This particular test had some very sketchy things happening behind the scenes due to some pass-by-reference effects.  Basically, the asserts below were "working" because the arg objects were getting passed around and continued to be updated as the logic flow continued (because of the `Object.assign` calls in `getUpdatedUserDoc` and `getUpdatedSettingsDoc`).  When I replaced the `Object.assign` calls by spreading the data into a new object, that broke this test.  

However, I think the updated test code is more explicit and accurate anyway.

tests/integration/api/controllers/users.spec.js
```diff
@@ -1605,6 +1608,42 @@ describe('Users API', () => {
     });
   });
 
+  describe('GET api/v2/users/{username}', () => {
```
m5r, line 1611: Could you add a test to ensure only users with the `can_view_users` permission can use this route?
Additionally if we allow users to look themselves up, could you add the scenarios where a user without `can_view_users` queries themselves successfully and where they query another user and get a 401?

m5r (changes requested):

shared-libs/user-management/src/users.js
```diff
@@ -806,6 +779,15 @@ module.exports = {
...
   },
+  getUser: async username => {
+    if (!username) {
+      throw new Error('Username is required.');
+    }
+
+    const [user, userSettings] = await getUserDocsByName(username);
+    const facilities = await facility.list([user]);
+    return mapUser(user, userSettings, facilities);
+  },
```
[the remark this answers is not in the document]
jkuester: :sweat_smile: maybe I am missing something, but AFICT this just returns a single object for the user (which I thought is what we wanted?).  :thinking:

api/src/controllers/users.js
```diff
@@ -235,7 +235,11 @@ module.exports = {
   },
 
   v2: {
-    get: async (req, res) => {
+    get: (req, res) => auth.check(req, 'can_view_users')
```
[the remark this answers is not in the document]
jkuester: Good call! That makes a lot of sense.

tests/integration/api/controllers/users.spec.js
jkuester: Done and done!

lorerod (approved):
Thank you, @jkuester and @m5r, for adding all the necessary tests to this implementation. The work looks fantastic.

tests/integration/api/controllers/users.spec.js
```diff
@@ -1605,6 +1608,116 @@ describe('Users API', () => {
...
+        await utils.request({
+          path: `/api/v2/users/invalidUsername`,
+        });
+      } catch ({ error }) {
+        expect(error.code).to.equal(404);
+        expect(error.error).to.equal('Failed to find user with name [invalidUsername] in the [users] database.');
+        return;
+      }
+
+      expect.fail('Should have thrown an error');
```
dianabarsan, line 1688: If you move this to the try block, after the `await utils.request`, you don't have to `return` in your catch.

dianabarsan (commented):
Great work! I left some minor stylistic comments inline.

tests/integration/api/controllers/users.spec.js
jkuester: So, what I don't like about doing it that way is that the error generated by the `fail` gets caught in the `catch` block.  Then, if `request` call does succeed and the expected error is not thrown, the test will fail, but it will fail with an error like "Expected: Failed to find user... but got Should have thrown an error".  Not a huge thing, but can be a bit confusing when you go to start debugging.

This approach with the `return` is the most convenient I could think of, but open to other ideas!

m5r: It's probably going to throw an error for `cannot read properties of undefined (reading 'code')` since the error being caught is destructured
The early return doesn't throw me off that much :sweat_smile:

m5r (approved):

shared-libs/user-management/src/users.js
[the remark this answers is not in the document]
m5r: Right, my bad! I might have hallucinated additional brackets in the response, I can confirm it's returning a single object

tests/integration/api/controllers/users.spec.js
dianabarsan: I logically prefer `expect.fail`-ing withing the try block. when I read this code, I immediately know: oh, this request is supposed to fail, and I prepare my brain for "why is it supposed to fail" and "how does the error look like". 

If I read the code the way it is now, top down, I see: oh, I'm OK with this request request not failing, but if it fails I run some assertions. Oh, I am not fine with this request failing.
If I read bottom up, I see an `expect.fail` at the end of the test and I think: oh, this is weird, so this test is _not_ supposed to run until the end. Now I have to search for all the returns and throws within the test. 

I totally understand that when reading error messages when the test fails, you get a prettier error message. My comment is about actually reading and understanding the test code. 

This is pretty minor and I'm not having a super strong opinion about this, it's just a preference about what makes sense and is easiest to follow for me when I read code.

PR #9017, opened 2024-04-17
chore(#9015): update date diff xpath tests to work in any TZ

### LINKED ISSUE #9015: difference-in-days xpath date functions fail in CEST timezone
Unit tests should pass regardless of timezone. Running `npm run unit-webapp` on my dev machine in Germany (CEST) I get...

```
  2 failing

  1) medic-xpath-extensions
       #difference-in-days
         should handle cases where end date is before start date and return negative number:

      AssertionError: expected -211 to equal -212
      + expected - actual

      --211
      +-212
      
      at Context.<anonymous> (webapp/tests/mocha/unit/enketo/medic-xpath-extensions.spec.js:179:14)
      at processImmediate (node:internal/timers:466:21)

  2) medic-xpath-extensions
       #difference-in-days
         should return the difference when valid inputs are provided:

      AssertionError: expected 211 to deeply equal 212
      + expected - actual

      -211
      +212
      
      at Context.<anonymous> (webapp/tests/mocha/unit/enketo/medic-xpath-extensions.spec.js:189:14)
      at processImmediate (node:internal/timers:466:21)

```

jkuester:
# Description

Closes #9015

These tests should now work in any TZ! Honestly, I am not super concerned about the exact date comparison assertions since we are just deferring all the comparison logic to `moment`.  These assertions are a "nice to have" if/when we ever got to refactoring this logic, but in reality they are mostly asserting moment's functionality....

I confirmed that these tests all run in at least the following TZs:

- `Canada/Pacific`
- `Africa/Monrovia`
- `Pacific/Auckland`
- `UTC`

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9015_fix_tz_tests/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9015_fix_tz_tests/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9015_fix_tz_tests/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/webapp/tests/mocha/unit/enketo/medic-xpath-extensions.spec.js
+++ b/webapp/tests/mocha/unit/enketo/medic-xpath-extensions.spec.js
@@ -12,6 +12,9 @@ describe('medic-xpath-extensions', function () {
   };
 
   describe('getTimezoneOffsetAsTime()', function () {
+    const getTimeZoneOffset = Date.prototype.getTimezoneOffset;
+    afterEach(() => Date.prototype.getTimezoneOffset = getTimeZoneOffset);
+
     it('returns the time zone offset in hours when given a time zone difference in minutes', function () {
       Date.prototype.getTimezoneOffset = () => -60;
       assert.equal(medicXpathExtensions.getTimezoneOffsetAsTime(new Date()), '+01:00');
@@ -24,6 +27,9 @@ describe('medic-xpath-extensions', function () {
   });
 
   describe('toISOLocalString()', function () {
+    const getTimeZoneOffset = Date.prototype.getTimezoneOffset;
+    afterEach(() => Date.prototype.getTimezoneOffset = getTimeZoneOffset);
+
     it('returns the ISO local string consistent with the time zone offset', function () {
       const date = new Date('August 19, 1975 23:15:30 GMT+07:00');
       Date.prototype.getTimezoneOffset = () => -60;
@@ -93,7 +99,7 @@ describe('medic-xpath-extensions', function () {
       ['2015-10-01T11:11:11.111', '2014-10-01T11:11:11.111', -12,],
       ['2014-10-01T00:00:00.000', '2015-10-01T00:00:00.000', 12,],
       ['August 19, 1975 00:00:00 GMT', 'August 18, 1976 23:15:30 GMT+07:00', 11],
-      ['Sun Sep 25 2005 1:00:00 GMT+0100', 'Sun Oct 25 2005 22:00:00 GMT+2300', 0],
+      ['Sun Sep 25 2005 1:00:00 GMT+0100', 'Sun Oct 25 2005 21:00:00 GMT+2300', 0],
       ['Sun Sep 25 2005 1:00:00 GMT+0100', 'Sun Oct 25 2005 22:00:00 GMT+2200', 1]
     ].forEach(function (example) {
       const d1 = { t: 'str', v: example[0] };
@@ -106,8 +112,8 @@ describe('medic-xpath-extensions', function () {
     });
 
     it('should report difference between date objects', function () {
-      const d1 = { t: 'date', v: new Date('2015-09-22') };
-      const d2 = { t: 'date', v: new Date('2014-09-22') };
+      const d1 = { t: 'date', v: moment('2015-09-22').toDate() };
+      const d2 = { t: 'date', v: moment('2014-09-22').toDate() };
 
       assert.equal(diffInMonths(d1, d2).v, -12);
     });
@@ -199,14 +205,18 @@ describe('medic-xpath-extensions', function () {
 
   describe('#to-bikram-sambat()', () => {
     [
-      [{ t: 'str', v: '2015-9-2' }, '१६ भदौ २०७२'],
-      [{ t: 'str', v: '1999-12-12' }, '२६ मंसिर २०५६'],
-      [{ t: 'date', v: new Date('2015-10-01T00:00:00.000') }, '१४ असोज २०७२'],
-      [{ t: 'num', v: 11111 }, '२१ जेठ २०५७'],
-      [{ t: 'arr', v: [{ textContent: '2014-09-22' }] }, '६ असोज २०७१'],
-    ].forEach(([date, expected]) => {
+      [{ t: 'str', v: '2015-9-2' }, '१६ भदौ २०७२', '१५ भदौ २०७२'],
+      [{ t: 'str', v: '1999-12-12' }, '२६ मंसिर २०५६', '२५ मंसिर २०५६'],
+      [{ t: 'date', v: new Date('2015-10-01T00:00:00.000') }, '१४ असोज २०७२', '१३ असोज २०७२'],
+      [{ t: 'num', v: 11111 }, '२१ जेठ २०५७', '२० जेठ २०५७'],
+      [{ t: 'arr', v: [{ textContent: '2014-09-22' }] }, '६ असोज २०७१', '५ असोज २०७१'],
+    ].forEach(([date, expected, offsetExpected]) => {
       it(`should format the ${date.t} [${JSON.stringify(date.v)}] according to the Bikram Sambat calendar`, () => {
-        assert.equal(func['to-bikram-sambat'](date).v, expected);
+        // The resolved Bikram Sambat date for input that does not have a time+TZ value will be different in
+        // environments where the timezone offset is 12 hours or greater (e.g. `Pacific/Auckland`)
+        const offsetHrs = new Date().getTimezoneOffset() / 60;
+        const tzDateOffset = offsetHrs <= -12;
+        assert.equal(func['to-bikram-sambat'](date).v, tzDateOffset ? offsetExpected : expected);
       });
     });
 
@@ -235,7 +245,7 @@ describe('medic-xpath-extensions', function () {
     [
       [
         [{ t: 'num', v: 1 }],
-        { t: 'date', v: new Date('2015-10-01') },
+        { t: 'date', v: moment('2015-10-01').toDate() },
         '2016-10-01'
       ],
       [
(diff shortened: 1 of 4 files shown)


webapp/tests/mocha/unit/enketo/medic-xpath-extensions.spec.js
```diff
@@ -93,7 +93,7 @@ describe('medic-xpath-extensions', function () {
       ['2015-10-01T11:11:11.111', '2014-10-01T11:11:11.111', -12,],
       ['2014-10-01T00:00:00.000', '2015-10-01T00:00:00.000', 12,],
       ['August 19, 1975 00:00:00 GMT', 'August 18, 1976 23:15:30 GMT+07:00', 11],
-      ['Sun Sep 25 2005 1:00:00 GMT+0100', 'Sun Oct 25 2005 22:00:00 GMT+2300', 0],
+      ['Sun Sep 25 2005 1:00:00 GMT+0100', 'Sun Oct 25 2005 21:59:59 GMT+2300', 0],
```
jkuester, line 96: For reasons that my weak brain cannot comprehend, in the `Pacific/Auckland` TZ, this would fail, saying the dates are `1` month apart (instead of `0`). Backing it up by a second makes this test pass in all TZ's.  Like I mentioned in the description, I think this is acceptable since all of this is just based on Moment's logic...

webapp/tests/mocha/unit/enketo/medic-xpath-extensions.spec.js
```diff
@@ -175,7 +175,7 @@ describe('medic-xpath-extensions', function () {
     const diffInDays = func['cht:difference-in-days'];
 
     it('should handle cases where end date is before start date and return negative number', function () {
-      const result = diffInDays(wrapDate('2023-08-01'), wrapDate('2023-01-01'));
+      const result = diffInDays(wrapDate('2023-08-01 UTC'), wrapDate('2023-01-01 UTC'));
```
garethbowen, line 178: This breaks my brain. Does it really make sense for dates to have timezones? Or should this be called a datezone? I don't think this is sensible.

Secondly the test should reflect what the field value might actually have in an xform. This may be a good edge case test but the common case will be a date without a timezone so we need to decide what we're doing to do with that (ie: the original string).

If the function is being given the same date input and coming up with a different answer depending on which timezone you're in then there's a bug in the function. The way this might manifest in real life is one user may see a valid result and another would see an invalid one. This would make testing configurations difficult.

I think a better solution would be to revert this change and modify the code so for dates it works consistently regardless of timezones.

garethbowen (changes requested):

webapp/tests/mocha/unit/enketo/medic-xpath-extensions.spec.js
jkuester: :sweat:  I think I have finally determined the cause of this behavior.  Basically I think it comes down to Moment's magical handling of [Month and year diffs](https://momentjs.com/docs/#/displaying/difference/)

> moment#diff has some special handling for month and year diffs. It is optimized to ensure that two months with the same date are always a whole number apart.
>
>So Jan 15 to Feb 15 should be exactly 1 month.
>
>Feb 28 to Mar 28 should be exactly 1 month.
>
>Feb 28 2011 to Feb 28 2012 should be exactly 1 year.

On the surface this seem fine and you would not think that you would get inconsistent behavior here _when your date values are declared with explicit time zones_.  However, I think that Moment is going to evaluate this magical date logic based on day/month values for the dates _in the timezone of the runtime environment._  

In `GMT`, these dates resolve to `Sun, 25 Sep 2005 00:00:00 GMT` and `Mon, 24 Oct 2005 23:00:00 GMT` which should clock in as an hour less than `1 month` (if both resolved to the `25th` day of the month, then the Moment magic would kick in and say they are `1 month` apart).  But, when running in an environment where the TZ is set to  `Pacific/Auckland` we need to roll the dates `+12` which gets us: `Sun, 25 Sep 2005 12:00:00 GMT+1200` `Tues, 25 Oct 2005 11:00:00 GMT+1200`.  Now they are on the same day, but the time of the second date is still _before the first_.  So, how are they a month apart?  Yep, you guessed it.  In 2005, DST for `Pacific/Auckland` started on 2 Oct!  That gives the extra hour which bumps this to exactly 1 month apart (according to the Moment magic).... :facepalm:

webapp/tests/mocha/unit/enketo/medic-xpath-extensions.spec.js
```diff
@@ -12,6 +12,9 @@ describe('medic-xpath-extensions', function () {
   };
 
   describe('getTimezoneOffsetAsTime()', function () {
+    const getTimeZoneOffset = Date.prototype.getTimezoneOffset;
+    afterEach(() => Date.prototype.getTimezoneOffset = getTimeZoneOffset);
```
jkuester, line 16: This was the cause of _some_ of the weirdness below (and I think fixing it also maybe caused some of the additional failures I uncovered),  Either way, I figure we def want to be re-setting `getTimezoneOffset` here.

webapp/tests/mocha/unit/enketo/medic-xpath-extensions.spec.js
```diff
@@ -93,7 +99,7 @@ describe('medic-xpath-extensions', function () {
       ['2015-10-01T11:11:11.111', '2014-10-01T11:11:11.111', -12,],
       ['2014-10-01T00:00:00.000', '2015-10-01T00:00:00.000', 12,],
       ['August 19, 1975 00:00:00 GMT', 'August 18, 1976 23:15:30 GMT+07:00', 11],
-      ['Sun Sep 25 2005 1:00:00 GMT+0100', 'Sun Oct 25 2005 22:00:00 GMT+2300', 0],
+      ['Sun Sep 25 2005 1:00:00 GMT+0100', 'Sun Oct 25 2005 21:00:00 GMT+2300', 0],
```
jkuester, line 102: See this old thread for explanation of why this had to change:

https://github.com/medic/cht-core/pull/9017/files/e7ecd56957a509131c2df880b259bc8d5157a8a5#r1569328416

webapp/tests/mocha/unit/enketo/medic-xpath-extensions.spec.js
```diff
@@ -199,14 +205,18 @@ describe('medic-xpath-extensions', function () {
...
+      [{ t: 'str', v: '2015-9-2' }, '१६ भदौ २०७२', '१५ भदौ २०७२'],
+      [{ t: 'str', v: '1999-12-12' }, '२६ मंसिर २०५६', '२५ मंसिर २०५६'],
+      [{ t: 'date', v: new Date('2015-10-01T00:00:00.000') }, '१४ असोज २०७२', '१३ असोज २०७२'],
+      [{ t: 'num', v: 11111 }, '२१ जेठ २०५७', '२० जेठ २०५७'],
+      [{ t: 'arr', v: [{ textContent: '2014-09-22' }] }, '६ असोज २०७१', '५ असोज २०७१'],
+    ].forEach(([date, expected, offsetExpected]) => {
       it(`should format the ${date.t} [${JSON.stringify(date.v)}] according to the Bikram Sambat calendar`, () => {
-        assert.equal(func['to-bikram-sambat'](date).v, expected);
+        // The resolved Bikram Sambat date for input that does not have a time+TZ value will be different in
+        // environments where the timezone offset is 12 hours or greater (e.g. `Pacific/Auckland`)
```
jkuester, line 216: See the [slack thread](https://medic.slack.com/archives/C024KTGRW/p1713461190238749) for more context about these changes....

webapp/tests/mocha/unit/enketo/medic-xpath-extensions.spec.js
```diff
@@ -106,8 +112,8 @@ describe('medic-xpath-extensions', function () {
     });
 
     it('should report difference between date objects', function () {
-      const d1 = { t: 'date', v: new Date('2015-09-22') };
```
jkuester, line 109: Calling the `Date` constructor with date strings of this format creates a date with the time value set to _midnight UTC_.  This can cause weirdness depending on the current TZ when doing diffs/adds via `moment` because `moment` will do its stuff according to the date _as it is in the current TZ_.  

In our real implementation code, this is not a problem because we are either getting a string/number value (that we use `moment` to serialize into a date) or we get a `Date` from OpenRosa (where the local timezone for times [is also handled](https://github.com/eyelidlessness/openrosa-xpath-evaluator/blob/3d088a4c0a240b6abe657a574b9655c1f1ee0e86/src/openrosa-extensions.js#L567)).  So, to keep things working in these tests, I just switched to use `moment` so the local TZ is always accounted for when creating the Date values....

webapp/tests/mocha/unit/enketo/medic-xpath-extensions.spec.js
jkuester: FTR [this change](https://github.com/medic/cht-core/pull/9017/files#r1571480910) fixed the issues here so that these changes were not necessary.

jkuester:
Well I opened the can of works on this one....  That being said, I think I finally got to the bottom of all the weirdness and errors and now all the tests work for me from `GMT-13` to `GMT+11`.

webapp/tests/mocha/unit/enketo/medic-xpath-extensions.spec.js
garethbowen: Ahh DST. That old chestnut...

garethbowen (approved):
Thanks for digging in to this! Great work.

garethbowen (approved):
Awesome, thanks!

Issue #6383, opened 2020-04-29
Patient id not populated in reference app


yrimal:
**Describe the bug**
When I install ANC Reference App on CHT-Core and submit Pregnancy registration form, `patient_id` field is not populated. 

**To Reproduce:**
1. On your development/deployment server, Install ANC reference app (cht-core/config/default) using medic-conf. 
2. Create required persona and one chw who can submit pregnancy registration form.
3. Login with CHW created in step 2 and submit one pregnancy form. 
4. Pick GUID of the form submitted in Step 3.
5. Go to couchdb and browse the GUID from step 4. 
6. Search for `patient_id` field. All 3 occurance of `patient_id` is empty. 

This field is used for analysis once migrated to postgres and is expected to populate. Also, it is easier to use than other options.

**Expected behavior**
`patient_id` field should be populated in document with correct patient id for whom this form was submitted. 

**Environment**
- Instance: Any instances with ANC Reference app
- Browser: Chrome
- Client platform: WSL
- App: Reference App
- Version: 3.8

**Additional context**

dianabarsan:
@yrimal Is the patient you are submitting the form for created on the device (the phone) right before submitting the pregnancy registration form? 
If not, does the patient you are submitting the form for have a `patient_id` field?

yrimal:
@dianabarsan , Yes, it was created right before submitting the pregnancy registration form. Both done from webapp but after Manually syncing the contact.

dianabarsan:
@yrimal In order for your contact on the device to have a `patient_id` you need to sync twice. Once to upload and once to download the updated document. 
To have a patient_id on the report, you should create the contact, sync to upload, wait for the contact to be updated by Sentinel (maybe check in the DB or logged in as an admin) and received a `patient_id`, sync and then create your report.

yrimal:
I tested it the way you mentioned and that sounds right @dianabarsan .

However, it is a likely situation that when registering new pregnancy, CHW first register person and then submit pregnancy registration form after that without update happening. Since we don't force user to update before submitting form, there'll be data integrity issue.

Will it be possible for sentinel or any other module in application to update patient_id field in next run?

dianabarsan:
The you should use `fields.patient_uuid` in analytics instead of or in conjunction with `patient_id`. It's impossible to have a guarantee that you will always have online fields populated in offline workflows.

1yuv:
[...]

PR #9022, opened 2024-04-19
fix(#6383): use correct sms context in webapp and sentinel

dianabarsan:
# Description

Fixes context to include patient information when generating SMS content in Webapp and Sentinel due tasks. 

#6383

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6383-correct-SMS-context/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6383-correct-SMS-context/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6383-correct-SMS-context/docker-compose/cht-couchdb-clustered.yml)

# License


shared-libs/transitions/src/schedule/due_tasks.js
```diff
@@ -1,4 +1,4 @@
-// TODO: this doesn't need to exist. We can just work this out dynamically when
+// ToThinkAbout: this doesn't need to exist. We can just work this out dynamically when
```
dianabarsan, line 1: Sonar complains about this.

garethbowen: As it should! TODOs should be GH issues...

dianabarsan: I personally don't agree with this todo :D

garethbowen (approved):
Looks great!

One question... IIRC you mentioned admin is doing this correctly already. Is there a way to improve reuse so everywhere generates template context with the same code?

shared-libs/transitions/src/schedule/due_tasks.js
garethbowen: Feel free to delete it if it's wrong!

dianabarsan:
> Is there a way to improve reuse 

I would like that. Because we talked about the possibility of backporting this, I chose not to do large refactors - most of the code changes are due to sonar complexity requirements. 
I think this would be a good improvement, but maybe a bigger piece of work.

dianabarsan:
Hi @lorerod 

I'm fixing this as a potentially urgent issue, affecting a production instance, and we might decide to release a patch for supported versions, to get this to users faster. It would be super helpful if you would take a look and advise on testing. 
Thanks!

shared-libs/transitions/src/schedule/due_tasks.js
dianabarsan: I'm not sure if it's wrong. TBH we don't _actually_ keep track of how many messages get processed by one of these runs, and what impact this would have if the `api/sms` endpoint would wait for this processing. 

What would we do in API? Run the due_dates optimistically and only do one batch at a time, and optimistically send from whatever messages we have pending? Run multiple batches until we have at lease one full load (25) pending messages or we have cleared the scheduled queue? 

What if someone "spams"  `api/sms`? Spamming is not even required, it would be enough to have concurrent requests that end up firing updates for the same set of docs and just failing with conflicts? 

I already believe that `api/sms` is bugged to update message states _before_ actually confirming that messages have been sent to gateway, which makes us need to query for two statuses (pending and forwarded-to) to get pending messages.

webapp/tests/karma/ts/services/format-data-record.service.spec.ts
```diff
@@ -626,6 +626,94 @@ describe('FormatDataRecord service', () => {
...
+      });
+    });
+
+    it('should generate messages with patient without patient_id', async () => {
+      const report = {
+        _id: 'report',
+        form: 'frm',
+        fields: { patient_id: '', patient_uuid: 'p_uuid', patient_name: 'alpha' },
+        hidden_fields: ['hidden_field'],
+        patient: { _id: 'p_uuid', patient_id: '12345', name: 'alpha', phone: '222' },
```
lorerod, line 680: What happens if we also don't have patient.patiend_id? Is this possible?

tests/integration/sentinel/schedules/due-tasks.spec.js
```diff
@@ -89,7 +89,7 @@ const contacts = [
     reported_date: reportedDate,
   },
   {
-    _id: 'patient2',
+    _id: 'patient_2',
```
lorerod, line 92: Does adding a scenario where the `patient` doesn't have `patient_id` make sense?

tests/e2e/default/reports/sms-messages.wdio-spec.js
```diff
@@ -0,0 +1,80 @@
...
+        },
+      ),
+  ];
+
+  before(async () => {
+    await utils.saveDocs([ ...places.values(), contact, patient, ...reports ]);
+    await loginPage.cookieLogin();
+  });
+
+  it('should generate SMS report correctly when lacking patient_id', async () => {
```
lorerod, line 67: Same here. Is `patient.patient_id` the only field that can be missing?

tests/e2e/default/reports/sms-messages.wdio-spec.js
```diff
@@ -0,0 +1,80 @@
...
+            }],
+            state: 'scheduled'
+          },]
+        },
+        {
+          patient: { _id: patient._id, patient_id: '' },
+          submitter: contact,
+          fields: { lmp_date: 'Feb 3, 2022', patient_name: patient.name },
+        },
+      ),
```
lorerod, line 59: Can we replicate this scenario by following each step of the workflow with the web app, from patient creation and report generation?

lorerod (changes requested):
The tests implemented are enough; I have a couple of suggestions we can discuss. Thank you, @dianabarsan.

webapp/tests/karma/ts/services/format-data-record.service.spec.ts
dianabarsan: This is possible. But this is not the case we are fixing here. If there is no patient_id at all, then there is nothing to expect in the SMS message.

tests/e2e/default/reports/sms-messages.wdio-spec.js
dianabarsan: It would be a little tricky to simulate the exact scenario, besides needing a lot of setup with config, it would also require editing some dates, to make sure that we simulate the conditions exactly. 

Can you explain the scenario you're requesting in detail and what you expect to see tested?

lorerod: My idea was to do the entire workflow: Create the patient, send the SMS report P, and verify. But if editing dates are required, how you implemented is better. Thanks.

lorerod (approved):
Thank you, @dianabarsan, for the clarifications.

tests/e2e/default/reports/sms-messages.wdio-spec.js
dianabarsan: This is what I'm thinking the test would require: 
- update settings to add the schedule to one of the existent forms, enable `registration` transition for the form to add the schedule. the schedule should have a message that would be sent sometime in the future, but more than 7 days into the future
- enable `generate_shortcode_on_contacts` transition
- create an offline user
- login as the offline user
- while offline, create a patient and then create a report for that patient
- go online and sync 
- wait for Sentinel to process both docs
- sync
- check the report in webapp and validate that webapp generates the message correctly
- if we want to check the sentinel message generating, we would need to edit the report manually, to set the due date of the scheduled task into the near future and restart sentinel, wait for due tasks to process, sync and run the same assertions again after the state has changed to `pending`.

Let me know if you think this is worth writing.

lorerod: Thanks, @dianabarsan, for putting together the test steps. That was what I was thinking, and now I have a clearer idea of how this works. 
If the first two steps are complicated, that can be an issue. But the last step worries me, and it worries me for a couple of things. First, editing the report manually is almost the same as creating and saving the [report](https://github.com/medic/cht-core/blob/b90e34d60fd4091792b5bcc8fa8c28ef48dce3bb/tests/e2e/default/reports/sms-messages.wdio-spec.js#L33) with the necessary values. Second, and most importantly, restarting, waiting, and syncing can easily introduce flakiness. What do you think?

dianabarsan: I think you're right, that the setup is really hard to achieve without it being flaky or requiring manual intervention. 
This was why I was reluctant to add this test.

PR #9027, opened 2024-04-19
fix(#8985): remove extra validation for parent place contact

### LINKED ISSUE #8985: Cannot use /api/v1/people to create a person if any place in the lineage has an invalid primary contact
**Describe the bug**

**To Reproduce**
1. Create a grandparent place with no primary contact
2. Create a parent place with a primary contact
3. Attempt to use the `POST api/v1/people` API to create a person within the parent place
 
**Expected behavior**
In 4.4, you could create this person. In 4.6, you see the error `Wrong type, this is not a person.` 
There is no UUID of the affending contact or place, no hints for a targeted followup. I thought the problem was in my payload; but it is in the hierarchy. 

**Environment**
- Instance: test
- App: webapp
- Version: 4.6.0

**Additional context**
For our Uganda eCHIS app: regions, districts, subdistricts, counties, subcounties all do not have primary contacts. This currently may block cht-user-management from using cht 4.6 via API. 

I believe this was [changed here](https://github.com/medic/cht-core/pull/8682/files#diff-847864fa7d0eecc88050c831b2b0b83933bf7bef90143e48810d7aec0b85a0d9R78)

I'll just note that in the real world it's sometimes quite annoying to need a 100% valid hierarchy in order to complete  task.  When county-level ICT team staff get an error like this and are blocked from creating a user at this place, what are they supposed to do? It's quite burdensome to train every single person creating users in eCHIS to also manage the hierarchy - this app doesn't have forms to create/edit these places or their primary contacts.

freddieptf:
Fixes #8985


dianabarsan:
@freddieptf it looks like sonar is complaining about code complexity here. would you mind the little refactor needed to keep sonar satisfied? Thanks!

freddieptf:
@dianabarsan tests fixed

shared-libs/contacts/src/places.js
```diff
@@ -109,8 +117,9 @@ const createPlace = async (place) => {
     return response;
   }
   const placeUUID = response.id;
-
-  contact.place = placeUUID;
+  if (_.isObject(contact)) {
```
dianabarsan, line 120: I'm seeing a strange behavior here: if I try to create a place with a contact, but I fail to create the payload correctly and I post something like: 
```
{
  _id: 'place_with_contact',
  type: 'district_hospital',
  name: 'placeWithoutContact',
  contact: 'angela',
};
```
I get a failed response saying: 
```
{ code: 404, error: 'Failed to find person.' }
```

However if I correct my mistake and try to push the same payload, but with a "good" contact afterwards , like:
```
{
  _id: 'place_with_contact',
  type: 'district_hospital',
  name: 'placeWithoContact',
  contact: {
    name: 'angela'
  }
}
``` 
I get: 
```
{ code: 409, error: 'Document update conflict.' }
```

Here is a gist with the simulation: https://gist.github.com/dianabarsan/1192777672f4d8545ac6495eac90d4d2

dianabarsan: I think we should cover this use case in some sort of test.

dianabarsan (changes requested):

shared-libs/contacts/src/places.js
freddieptf: Yeah that's gonna happen since the place was getting created with the assumption that a contact is always valid if it's a string. I've updated the code to try and get the contact first if it's a string before creating the place

shared-libs/contacts/src/places.js
```diff
@@ -104,15 +106,25 @@ const createPlace = async (place) => {
...
   }
-  const placeUUID = response.id;
 
-  contact.place = placeUUID;
+  contact = await preparePlaceContact(contact);
+  const placeResponse = await db.medic.post(place);
+  contact.place = placeResponse.id;
   const person = await people.getOrCreatePerson(contact);
-  const result = await updatePlace(placeUUID, { contact: person._id });
+  const result = await updatePlace(placeResponse.id, { contact: person._id });
```
dianabarsan, line 127: Just suggesting a slightly more compact way of writing this code: 
```

const prepareContact = async (contact) => {
  if (!contact) {
    return;
  }

  if (_.isString(contact)) {
    return await people.getOrCreatePerson(contact);
  }

  if (_.isObject(contact)) {
    contact.type = contact.type || people._getDefaultPersonType();
    const errStr = people._validatePerson(contact);
    if (errStr) {
      return err(errStr);
    }
    return contact;
  }
}

const createPlace = async (place) => {
  await module.exports._validatePlace(place);
  const contact = await prepareContact(place.contact);
  delete place.contact;

  const date = place.reported_date ? utils.parseDate(place.reported_date) : new Date();
  place.reported_date = date.valueOf();
  if (place.parent) {
    place.parent = lineage.minifyLineage(place.parent);
  }

  if (!contact) {
    return await db.medic.post(place);
  }

  if (contact._id) {
    place.contact = contact._id;
    return await db.medic.post(place);
  }

  const placeResponse = await db.medic.post(place);
  contact.place = placeResponse.id;
  const person = await people.getOrCreatePerson(contact);
  const result = await updatePlace(placeResponse.id, { contact: person._id });
  return { ...result, contact: { id: person._id } };
};
```

What do you think?

shared-libs/contacts/test/unit/places.spec.js
```diff
@@ -561,4 +561,13 @@ describe('places controller', () => {
 
   });
```
dianabarsan, line 563: Can you please add a unit test that covers the test case I found, where you can pass an invalid contact to the new place? 
Also, can you please add some unit tests to cover the bug you are fixing?

dianabarsan (changes requested):

shared-libs/contacts/src/places.js
freddieptf: Looks good. I was initially trying to keep `prepareContact` as an async wrapper around `validatePerson` that throws on err ~and to please sonar~. One concern with this suggestion is if a user included an `_id` in the contact object then i think we'd assign a non-existent contact to the place

dianabarsan: > One concern with this suggestion is if a user included an _id in the contact object then i think we'd assign a non-existent contact to the place

I'm aware. I'm sorry I didn't mention this in a comment or something, but that check was a placeholder for: assess whether this is a new or existing contact. I'm hoping there is some clean way to do that.

freddieptf: Cool cool, i decided to return an object instead with an `exists` field so we don't check any payload fields

shared-libs/contacts/src/places.js
```diff
@@ -89,30 +83,45 @@ const validatePlace = place => {
...
+  if (errStr) {
+    return err(errStr);
+  }
+  return { exists: false, contact: contact };
+};
 
-  const contact = place.contact;
+const createPlace = async (place) => {
+  await module.exports._validatePlace(place);
+  const { exists, contact } = await preparePlaceContact(place.contact);
```
dianabarsan, line 107: I think `exists` becomes ambiguous in this context. Maybe:
```suggestion
  const { exists: contactExists, contact } = await preparePlaceContact(place.contact);
```

tests/integration/api/controllers/places.spec.js
```diff
@@ -132,16 +166,16 @@ describe('Places API', () => {
...
         type: 'district_hospital',
-        contact: 'fixture:user:online'
+        contact: contactId
       };
       return utils.request(onlineRequestOptions)
         .then(result => {
           chai.expect(result.id).to.not.be.undefined;
-          chai.expect(result.contact.id).to.not.be.undefined;
-          return utils.getDocs([result.id, result.contact.id]);
+          return utils.getDocs([result.id, contactId]);
```
dianabarsan, line 178: Is this changing because now we're _not_ returning this contact id as part of the response? 
I think we should add an assertion for that, so we make the API response very clear in our tests.

shared-libs/contacts/test/unit/places.spec.js
```diff
@@ -404,6 +404,20 @@ describe('places controller', () => {
...
+      const place = {
+        name: 'HC',
+        type: 'district_hospital',
+        contact: 'person'
+      };
+      fetchHydratedDoc.rejects({ status: 404 });
+      const post = db.medic.post;
+      return controller._createPlaces(place).catch(err => {
+        chai.expect(err.message).to.equal('Failed to find person.');
+        chai.expect(post.callCount).to.equal(0);
```
dianabarsan, line 417: Unfortunately, we need more there. Adding assertions in the `catch` _only_ works if you also add a throw if the call is successful. Otherwise, if the call resolves, your test passes and you don't run any assertion: 
```suggestion
try {
  await controller._createPlaces(place);
  chai.expect.fail('Call should throw');
} catch (err) {
  chai.expect(err.message).to.equal('Failed to find person.');
  chai.expect(post.callCount).to.equal(0);
}
```

I'm noticing the same pattern in other unit tests as well. Those should be changed too!

shared-libs/contacts/src/places.js
dianabarsan: Cool! Thanks!

dianabarsan (changes requested):

tests/integration/api/controllers/places.spec.js
freddieptf: Yes, the contact id is now only returned if we create a contact. I did not think this would be a breaking change but will revert so the previous behavior doesn't change

dianabarsan:
Hi @freddieptf 

It looks like there is an e2e test failure that is genuine: https://github.com/medic/cht-core/actions/runs/8919300697/job/24629918331?pr=9027#step:16:1512

freddieptf:
@dianabarsan thanks for pointing that out. I'm going to update the test to align with the new behavior. I think we shouldn't reassign the contact's parent to the place we just created, this way the only side effect is assigning the existing contact as the primary contact for the new place. What you think?

dianabarsan:
Indeed @freddieptf . We should _not_ have a contact assignment side-effect as part of setting that contact as primary for another place.

dianabarsan (approved):
Great work, Fred, thanks for your effort in this! 

I'm approving to not hold off the release, my personal preference would be to add a few more e2e tests to assess behavior, but those can be added later.

Issue #8802, opened 2024-01-11
Flaky e2e test


dianabarsan:
**Describe the issue**

**Describe the improvement you'd like**

**Describe alternatives you've considered**

garethbowen:
Another possibly related failure: https://github.com/medic/cht-core/actions/runs/8780156058/job/24090006364

```
[chrome-headless-shell 123.0.6312.122 linux #0-4] Running: chrome-headless-shell (v123.0.6312.122) on linux
[chrome-headless-shell 123.0.6312.122 linux #0-4] Session ID: ae1d22f9f7c60c060fd495d961830e6f
[chrome-headless-shell 123.0.6312.122 linux #0-4]
[chrome-headless-shell 123.0.6312.122 linux #0-4] » /tests/e2e/default/db/ongoing-replication.wdio-spec.js
[chrome-headless-shell 123.0.6312.122 linux #0-4] ongoing replication
[chrome-headless-shell 123.0.6312.122 linux #0-4]    ✓ should download new documents
[chrome-headless-shell 123.0.6312.122 linux #0-4]    ✓ should handle updates to existing documents
[chrome-headless-shell 123.0.6312.122 linux #0-4]    ✓ should download new forms and update forms
[chrome-headless-shell 123.0.6312.122 linux #0-4]    ✓ should download new languages and language updates
[chrome-headless-shell 123.0.6312.122 linux #0-4]    ✓ should handle deletes
[chrome-headless-shell 123.0.6312.122 linux #0-4]    ✖ should download settings updates (5 retries)
[chrome-headless-shell 123.0.6312.122 linux #0-4]    ✓ should handle conflicts
[chrome-headless-shell 123.0.6312.122 linux #0-4]
[chrome-headless-shell 123.0.6312.122 linux #0-4] 6 passing (4m 9s)
[chrome-headless-shell 123.0.6312.122 linux #0-4] 1 failing
[chrome-headless-shell 123.0.6312.122 linux #0-4]
[chrome-headless-shell 123.0.6312.122 linux #0-4] 1) ongoing replication should download settings updates
[chrome-headless-shell 123.0.6312.122 linux #0-4] A previous test did not call revertSettings
[chrome-headless-shell 123.0.6312.122 linux #0-4] Error: A previous test did not call revertSettings
[chrome-headless-shell 123.0.6312.122 linux #0-4]     at updateCustomSettings (/home/runner/work/cht-core/cht-core/tests/utils/index.js:438:11)
[...]

PR #9062, opened 2024-04-24
chore(#8802): stabilize ongoing replication test

dianabarsan:
# Description

#8802

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8802-stabiize-test/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8802-stabiize-test/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8802-stabiize-test/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/tests/e2e/default/db/initial-replication.wdio-spec.js
+++ b/tests/e2e/default/db/initial-replication.wdio-spec.js
@@ -105,12 +105,11 @@ const refreshAndWaitForAngular = async () => {
 
 describe('initial-replication', () => {
   before(async () => {
+    await utils.stopSentinel();
+
     // we're creating ~2000 docs
     await utils.saveDocs([...userAllowedDocs.places, ...userDeniedDocs.places]);
     await utils.createUsers([userAllowedDocs.user]);
-
-    await sentinelUtils.waitForSentinel();
-    await utils.stopSentinel();
 
     await utils.saveDocs(userAllowedDocs.clinics);
     await utils.saveDocs(userDeniedDocs.clinics);
--- a/tests/e2e/default/db/ongoing-replication.wdio-spec.js
+++ b/tests/e2e/default/db/ongoing-replication.wdio-spec.js
@@ -25,8 +25,6 @@ describe('ongoing replication', function() {
   this.timeout(4 * 60 * 1000); // Sometimes the tests take longer to complete than the original 2 minutes timeout.
 
   before(async () => {
-    await sentinelUtils.skipToSeq();
-
     await utils.saveDocs([...userAllowedDocs.places, ...userDeniedDocs.places]);
     await utils.createUsers([userAllowedDocs.user]);
 
@@ -34,10 +32,14 @@ describe('ongoing replication', function() {
     await saveData(userDeniedDocs);
   });
 
+  after(async () => {
+    await utils.stopSentinel();
+    await sentinelUtils.skipToSeq();
+    await utils.startSentinel();
+  });
+
   afterEach(async () => {
     await browser.throttle('online');
-    await sentinelUtils.skipToSeq();
-    await sentinelUtils.waitForSentinel();
   });
 
   it('should download new documents ', async () => {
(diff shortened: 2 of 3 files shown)


tests/e2e/default/db/initial-replication.wdio-spec.js
```diff
@@ -105,13 +105,12 @@ const refreshAndWaitForAngular = async () => {
 
 describe('initial-replication', () => {
   before(async () => {
+    await utils.stopSentinel();
```
garethbowen, line 108: We start sentinel again in the `after` but we don't wait for it to process the 2k docs. Do you think this will impact other tests or can we just safely leave it running in the background?

tests/e2e/default/db/ongoing-replication.wdio-spec.js
```diff
@@ -35,12 +33,18 @@ describe('ongoing replication', function() {
...
+    await utils.startSentinel();
   });
 
   afterEach(async () => {
     await browser.throttle('online');
+    const isRevertingSettings = utils.revertSettings(true);
+    if (isRevertingSettings) {
+      await isRevertingSettings;
+      await commonPage.sync(true);
+    }
```
garethbowen, line 47: It feels like this should be part of the `revertSettings` function. Surely you always need to sync after changing settings?

tests/utils/index.js
```diff
@@ -1115,32 +1116,38 @@ const waitForDockerLogs = (container, ...regex) => {
...
         resolve();
         clearTimeout(timeout);
         killSpawnedProcess(proc);
       }
     };
 
-    proc.stdout.on('data', checkOutput);
-    proc.stderr.on('data', checkOutput);
+    proc.stdout.on('data', check);
+    proc.stderr.on('data', check);
```
garethbowen, line 1150: This is just a refactor right? Did Sonar get you?

garethbowen (changes requested):
Looks good. A couple of open questions...

tests/e2e/default/db/initial-replication.wdio-spec.js
dianabarsan: The code we run in the after is:
```
 await sentinelUtils.skipToSeq();
 await utils.startSentinel();
```

`skipToSeq` will change the transitions-seq value to the current update_seq of the database. so we're not waiting for sentinel to process docs here because the queue should be empty. 
Are you worried about processing deletes?

tests/e2e/default/db/ongoing-replication.wdio-spec.js
dianabarsan: I thought about that too, I'm not entirely sure if it's always the case. 
I'm afraid this will start generating a lot of other e2e tests to flake or fail because we suddenly introduce this new behavior in a PR that attempts to stabilize an e2e test.

tests/utils/index.js
dianabarsan: Correct. It complained about code nesting depth.

tests/e2e/default/db/initial-replication.wdio-spec.js
garethbowen: Hmm no in retrospect I think it's fine. Thanks for explaining!

tests/e2e/default/db/ongoing-replication.wdio-spec.js
garethbowen: Yeah it feels very fragile. Fine to leave it as is for now.

garethbowen (approved):
Awesome, thanks!

tests/e2e/default/db/ongoing-replication.wdio-spec.js
dianabarsan: I think you're totally right to point to this. I was very close to implementing that as well.

Issue #8437, opened 2023-08-07
Enable `strict` mode for TypeScript Config


jkuester:
**Describe the issue**
Typescript offers a broad range of functionality/checks that it can do to ensure program correctness.  Many of this are only enabled if the [`strict` flag](https://www.typescriptlang.org/tsconfig#strict) is enabled.  Currently we do not have this flag set and are not taking advantage of all the of the guarantees that TS could provide.

**Describe the improvement you'd like**
Enabling `strict` all at once would likely require a significant amount of code changes to our existing code base to be compatible with the strict requirements.  Instead, we can roll out changes for individual sub-rules one-by-one and slowly work our way towards being able to just enable `strict` (and thereby automatically take advantage of future checks added to Typescript).

* [alwaysStrict](https://www.typescriptlang.org/tsconfig#alwaysStrict)
* [noImplicitAny](https://www.typescriptlang.org/tsconfig#noImplicitAny)
* [noImplicitThis](https://www.typescriptlang.org/tsconfig#noImplicitThis)
* [bindCallApply](https://www.typescriptlang.org/tsconfig#strictBindCallApply)
* [strictPropertyInitialization](https://www.typescriptlang.org/tsconfig#strictPropertyInitialization)
* [useUnknownInCatchVariables](https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables)

Each of these will need to be individually evaluated to determine if it is actually desirable/useful for our code-base.

**Describe alternatives you've considered**
We do not have to enable all of these checks if we determine that some are not useful/desirable.  Not enabling any is also an option, though it just means that we are not taking advantage of validation that TS has to offer....

m5r:
I've done some research previously and I found 2 complimentary approaches to incrementally migrate the project to strict mode.

We can use https://github.com/allegro/typescript-strict-plugin to manually migrate the project to strict mode file by file. The files are easily recognizable from the `// @ts-strict-ignore` comment at the top of each file.
Here is a large repo that is using this plugin: https://github.com/adobe/react-spectrum/issues/3183

Additionally, we could use https://github.com/airbnb/ts-migrate to automate bits of the migration like explicitly typing the untyped stuff with `any`. There is [existing tooling](https://gist.github.com/VanTanev/a8fa00edac06442413c15ab8a61d8151) to measure how many `any` are left to type.

[...]

PR #9064, opened 2024-04-24
chore(#8437): add `noImplicitThis` to webapp tsconfig

jkuester:
# Description

https://github.com/medic/cht-core/issues/8437

Very minor change to set [`noImplicitThis` ](https://www.typescriptlang.org/tsconfig/#noImplicitThis) that is one more step in the direction of being able to enable `strict` mode.  

There were only two places where we were violating this rule and both of them were deliberate because we are using a shadowed `this` to be able to reference DOM elements.

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8437_no_implicit_this/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8437_no_implicit_this/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8437_no_implicit_this/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/webapp/src/ts/services/android-api.service.ts
+++ b/webapp/src/ts/services/android-api.service.ts
@@ -50,6 +50,7 @@ export class AndroidApiService {
     $container
       .find('select.select2-hidden-accessible')
       .each(function() {
+        // @ts-expect-error Intentionally referencing shadowed `this` variable.
         const elem = <any>$(this);
         if (elem.select2('isOpen')) {
           elem.select2('close');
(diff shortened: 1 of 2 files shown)


webapp/src/ts/services/android-api.service.ts
```diff
@@ -50,6 +50,7 @@ export class AndroidApiService {
     $container
       .find('select.select2-hidden-accessible')
       .each(function() {
+        // @ts-expect-error Intentionally referencing shadowed `this` variable.
```
m5r, line 53: Thanks for using `@ts-expect-error` and not `@ts-ignore` :smile:

m5r (approved):

Issue #8608, opened 2023-10-02
add a "get all logs" script for k3d deployments


mrjones-plip:
**Describe the issue**
when you're using the [soon merged k3d `troubleshooting/view-logs` script](https://github.com/medic/cht-core/pull/8263), there's no easy way to get all logs and node/cluster/what ever state for k3d deployments 

**Describe the improvement you'd like**
It'd be rad to have a "debug" script [like this one](https://docs.communityhealthtoolkit.org/apps/guides/debugging/sharing-4x-logs/). I could run it now and answer all like 49 bazzilion questions you wanna ask me in one go!

**Describe alternatives you've considered**

henokgetachew:
Completed. PR in review.

Issue #8605, opened 2023-10-02
Show URL when done running k3d script


mrjones-plip:
**Describe the issue**
when you're done using the soon to be [merged k3d `cht-deploy` script](https://github.com/medic/cht-core/pull/8263), you don't know what the URL of your instance is

**Describe the improvement you'd like**
we should output the URL of your instance:

```
  ➜  deploy git:(cht-k8s-configuration) ✗ ./cht-deploy -f values.yml


   --------------    SUCCESS      -------------- 
   https://mrjones.dev.medicmobile.org is live!
   ------------------------------------------------- 
 
 Namespace mrjones already exists.
 Release does not exist. Performing install.
 NAME: mrjones
 LAST DEPLOYED: Fri Sep 15 16:36:31 2023
 NAMESPACE: mrjones
 STATUS: deployed
 REVISION: 1
 TEST SUITE: None
```

**Describe alternatives you've considered**

henokgetachew:
Completed. PR in review.

Issue #8604, opened 2023-10-02
Catch missing values file in k3d script


mrjones-plip:
**Describe the issue**
when you're using the soon to be [merged k3d script](https://github.com/medic/cht-core/pull/8263), and your `values.yml` is missing, you get an ugly error: 

```
 ./cht-deploy -f values.yaml                               
 Traceback (most recent call last):
   File "/home/mrjones/.local/bin/invoke", line 8, in <module>
     sys.exit(program.run())
   File "/home/mrjones/.local/lib/python3.8/site-packages/invoke/program.py", line 384, in run
     self.execute()
   File "/home/mrjones/.local/lib/python3.8/site-packages/invoke/program.py", line 569, in execute 
     executor.execute(*self.tasks)
   File "/home/mrjones/.local/lib/python3.8/site-packages/invoke/executor.py", line 129, in execute
     result = call.task(*args, **call.kwargs)
   File "/home/mrjones/.local/lib/python3.8/site-packages/invoke/tasks.py", line 127, in __call__
     result = self.body(*args, **kwargs)
   File "/home/mrjones/Documents/MedicMobile/cht-core/scripts/deploy/tasks.py", line 160, in install
     prepare(c, f)
   File "/home/mrjones/.local/lib/python3.8/site-packages/invoke/tasks.py", line 127, in __call__
     result = self.body(*args, **kwargs)
   File "/home/mrjones/Documents/MedicMobile/cht-core/scripts/deploy/tasks.py", line 11, in prepare
     with open(f, 'r') as stream:
 FileNotFoundError: [Errno 2] No such file or directory: 'values.yaml
```

**Describe the improvement you'd like**
It would be great if the `cht-deploy` script detected a missing file and just told you instead outputting a bit scary stack trace. 

**Describe alternatives you've considered**

henokgetachew:
Complete. PR in review

Issue #8551, opened 2023-09-15
Convert Python Invoke code in the cht-deploy to its JS/TS equivalent


henokgetachew:
**Describe the improvement you'd like**
The python invoke code found in `tasks.py` should be converted to its equivalent in JS/TS. This would make it align with the coding standards for this repo and would also make it easier to ship as an npm package.

henokgetachew:
I have worked on porting the functionalities in tasks.py over to js to establish feature parity. Working out the bugs and issues right now.

henokgetachew:
Completed. PR in review.

PR #9074, opened 2024-04-29
feat(#8551): removes python from cht-deploy and some bug fixes

henokgetachew:
# Description
Handles the following tickets:

#8551 Remove python - establishing feature parity. No major refactoring.
#8604 Catch Missing Values
#8605 Show URL when done execution
#8608 Add a get-all-logs troubleshooting command

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:cht-deploy-fixes/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:cht-deploy-fixes/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:cht-deploy-fixes/docker-compose/cht-couchdb-clustered.yml)

# License


scripts/deploy/src/install.js
```diff
@@ -0,0 +1,163 @@
+const fs = require('fs');
+const { execSync } = require('child_process');
+const axios = require('axios');
```
dianabarsan, line 3: Unfortunately axios is not part of our stack, and will probably never be. Please use `node-fetch` instead. 
Discussion: https://github.com/medic/cht-core/issues/6250

dianabarsan (changes requested):

scripts/deploy/src/install.js
```diff
@@ -0,0 +1,163 @@
...
+            execSync(
+                `helm upgrade --install ${project_name} ${CHT_CHART_NAME} --version ${chart_version} --namespace ${namespace} --values ${valuesFile} --set cht_image_tag=${image_tag}`, { stdio: 'inherit' });
+            console.log(`Instance at ${values.ingress.host} upgraded successfully.`);
+        } else {
+            console.log("Release does not exist. Performing install.");
+            execSync(`helm install ${project_name} ${CHT_CHART_NAME} --version ${chart_version} --namespace ${namespace} --values ${valuesFile} --set cht_image_tag=${image_tag}`, { stdio: 'inherit' });
+            console.log(`Instance at ${values.ingress.host} installed successfully.`);
+        }
+    } catch (err) {
+        console.error(err.message);
```
nydr, line 113: Applies to all error handling:
I find it a bit inconsistent, I would prefer to exit on any error (`process.exit(1);` as in other places).
Would it make sense to move all error handling to a general error handler instead of having try catch in many places?

scripts/deploy/src/install.js
```diff
@@ -0,0 +1,163 @@
+const fs = require('fs');
+const { execSync } = require('child_process');
+const axios = require('axios');
+const yaml = require('js-yaml');
+const path = require('path');
+
+const MEDIC_REPO_NAME = "medic";
+const MEDIC_REPO_URL = "https://docs.communityhealthtoolkit.org/helm-charts";
+const CHT_CHART_NAME = `${MEDIC_REPO_NAME}/cht-chart-4x`;
+const DEFAULT_CHART_VERSION = "0.2.3";
```
nydr, line 10: Set to wildcard
```suggestion
const DEFAULT_CHART_VERSION = "1.0.*";
```
(or `1.*.*`) you might also need to quote the usage of the version to ensure it doesn't expand the wildcard

scripts/deploy/src/install.js
nydr: Consider native [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) and setting a minimum node version, its API wasn't considered stable until node 21, but it has existed since 18 and I can't think of any breaking changes that would affect this. It's fine to use node-fetch if you don't want to verify compatibility

scripts/deploy/src/install.js
```diff
@@ -0,0 +1,163 @@
...
+        fs.unlinkSync('certificate.crt');
+        fs.unlinkSync('private.key');
+    }
+}
+
+async function get_image_tag(chtversion) {
+    const response = await axios.get(`https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:${chtversion}`);
+    const data = response.data;
+    const tag = data.tags.find(tag => tag.container_name === 'cht-api');
+    if (!tag) throw new Error('cht image tag not found');
```
nydr, line 92: Thrown errors in asynchronous functions in javascript can behave a bit unexpected when coming from other languages, consider returning a rejected promise, error object, or exiting directly

scripts/deploy/package.json
```diff
@@ -0,0 +1,24 @@
+{
```
nydr, line 1: Consider setting a minimum [engines](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines) node/npm version (either same as cht core or node 18+) mostly to set expectations but would also be good to run tests explicitly against all the supported major LTS versions

Consider adding [bin](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#bin) and ensure it is executable using [npx](https://docs.npmjs.com/cli/v7/commands/npx) and similar runtimes, ideally there should also be a publish build step added, but that can be out of scope for this PR, as-is it is already better than the old solution IMO

scripts/deploy/src/install.js
henokgetachew: Done.

scripts/deploy/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "name": "cht-deploy",
+  "version": "1.0.0",
+  "description": "## Deploy the CHT",
+  "main": "cht-deploy",
```
nydr, line 5: minor, I think [main](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#main) should be src/install.js or something similar as cht-deploy is not intended to be imported

nydr (commented):
Great initiative! I'm happy to see significant progress in replacing the old script.

I would love to see a test, even if it's just trying to import the module or testing the easiest to test function as it can catch some syntax errors and make it easier to add more tests later. Since the old script didn't have any tests I don't consider this a blocker, even in the current state I consider it an significant improvement.

I think some [jsdoc](https://jsdoc.app/about-getting-started) style comments on most functions would be great

scripts/deploy/troubleshooting/get-all-logs
```diff
@@ -0,0 +1,46 @@
...
+  echo "Failed to create directory for logs."
+  exit 1
+fi
+
+# Get a list of all pod names in the namespace
+PODS=$(kubectl -n $NAMESPACE get pods -o jsonpath="{.items[*].metadata.name}")
+
+if [ -z "$PODS" ]; then
+  echo "No Pods found in Namespace $NAMESPACE."
+  rm -rf "$NAMESPACE-logs"
```
mrjones-plip, line 32: Instead of creating a directory using the `$NAMESPACE`, let's just use something like `/tmp/cht-eks-logs` or something.  We can create it and abandon it and let `tmp` do it's thing.  

Otherwise, accepting a user input variable and then running `rm -rf` with it is dangerous.  A slightly forced, but real enough, demo below:

```
$ mkdir really-important-stuff-omg-dont-delete-logs

$ file really-important-stuff-omg-dont-delete-logs
really-important-stuff-omg-dont-delete-logs: directory

$ ./troubleshooting/get-all-logs really-important-stuff-omg-dont-delete
No Pods found in Namespace really-important-stuff-omg-dont-delete.

$ file really-important-stuff-omg-dont-delete-logs                     
really-important-stuff-omg-dont-delete-logs: cannot open `really-important-stuff-omg-dont-delete-logs' (No such file or directory)
```

```suggestion
```

scripts/deploy/troubleshooting/get-all-logs
```diff
@@ -0,0 +1,46 @@
...
+  exit 1
+fi
+
+# Loop through all pod names and fetch logs, saving them to individual files
+for POD in $PODS; do
+  kubectl -n $NAMESPACE logs $POD > "$NAMESPACE-logs/$POD.log"
+done
+
+# Create a tar.gz archive of the log files
+tar -czf "$NAMESPACE-logs.tar.gz" -C "$NAMESPACE-logs" .
```
mrjones-plip, line 42: let's add `*-logs.tar.gz` to `.gitignore` [file](https://github.com/medic/cht-core/blob/master/.gitignore) to ensure no one accidentally commits their tarball.

scripts/deploy/troubleshooting/get-all-logs
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
mrjones-plip, line 1: yay!  this worked exactly as promised - thanks for adding!

scripts/deploy/troubleshooting/get-all-logs
```diff
@@ -0,0 +1,46 @@
...
+
+if [ -z "$PODS" ]; then
+  echo "No Pods found in Namespace $NAMESPACE."
+  rm -rf "$NAMESPACE-logs"
+  exit 1
+fi
+
+# Loop through all pod names and fetch logs, saving them to individual files
+for POD in $PODS; do
+  kubectl -n $NAMESPACE logs $POD > "$NAMESPACE-logs/$POD.log"
```
mrjones-plip, line 38: One issue that the docker compose version of this script faced was that some deployments were long running and would return months (years!?) of logs which was way way too much.  By default it would return 24 hrs worth of logs but would allow you to [specify your own range](https://docs.communityhealthtoolkit.org/apps/guides/debugging/sharing-4x-logs/#calling-the-script).

We should add that feature here!

Here's a mock code update.  We'd need to accept a new input instead hard coding it to `24h`. as well, we need to use the new `tmp` directory suggested above:

```bash
kubectl -n $NAMESPACE logs $POD  --since=24h > "/tmp/cht-eks-logs/$POD.log"
```

scripts/deploy/troubleshooting/get-all-logs
```diff
@@ -0,0 +1,46 @@
...
+done
+
+# Create a tar.gz archive of the log files
+tar -czf "$NAMESPACE-logs.tar.gz" -C "$NAMESPACE-logs" .
+
+rm -rf "$NAMESPACE-logs"
+
+echo "Logs archived in $NAMESPACE-logs.tar.gz"
```
mrjones-plip, line 46: Given this has the HAProxy logs, it very likely contains PII/PHI.  We should add a warning:

```
NOTE: Please remove the file when done as it may contain PII/PHI.
```

scripts/deploy/troubleshooting/get-all-logs
```diff
@@ -0,0 +1,46 @@
...
+  exit 1
+}
+
+if [ "$#" -ne 1 ]; then
+  usage
+fi
+
+NAMESPACE=$1
+
+mkdir -p "$NAMESPACE-logs"
```
mrjones-plip, line 21: see below about not creating and deleting a directory with a user input var.  As well, we can clean it up in case there's old logs in there from a different namespace or prior time the called this script on same namespace:

```suggestion
mkdir -p "/tmp/cht-eks-logs"
rm "/tmp/cht-eks-logs/*.log"
```

scripts/deploy/troubleshooting/get-all-logs
```diff
@@ -0,0 +1,46 @@
...
+
+# Loop through all pod names and fetch logs, saving them to individual files
+for POD in $PODS; do
+  kubectl -n $NAMESPACE logs $POD > "$NAMESPACE-logs/$POD.log"
+done
+
+# Create a tar.gz archive of the log files
+tar -czf "$NAMESPACE-logs.tar.gz" -C "$NAMESPACE-logs" .
+
+rm -rf "$NAMESPACE-logs"
```
mrjones-plip, line 44:
```suggestion
```

mrjones-plip (changes requested):
Looking better - thanks for iterating!  

Confirmed that https://github.com/medic/cht-core/issues/9076 is fixed with the latest changes.  As well, sad path and happy paths are well covered!

Requesting some security and changes to new save all logs script.

scripts/deploy/troubleshooting/get-all-logs
nydr: consider using `mktemp` and trap, example:
https://github.com/medic/cht-core/blob/4ced6937a73b105fa2a970d729e3803afd29893a/haproxy-healthcheck/requirements/update.sh#L67-L69

scripts/deploy/cht-deploy
```diff
@@ -1,40 +1,43 @@
-#!/bin/bash
+#!/usr/bin/env node
```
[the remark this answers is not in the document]
henokgetachew: Added.

scripts/deploy/cht-deploy
```diff
@@ -1,7 +1,33 @@
...
+
+import { dirname } from 'path';
+import { fileURLToPath } from 'url';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+function validateNodeVersion() {
+    const packageJsonPath = path.resolve(__dirname, 'package.json');
+    let packageJson;
```
mrjones-plip, line 16: super minor, but we don't change this later, so we can `const` it up!

```suggestion
    const packageJson;
```

henokgetachew: Actually, we do. Check out the code block right below

scripts/deploy/package.json
henokgetachew: handled.

scripts/deploy/src/install.js
```diff
@@ -0,0 +1,163 @@
...
+    console.log("Checking if secret api-tls-secret already exists...");
+    try {
+        execSync(`kubectl get secret api-tls-secret -n ${namespace}`, { stdio: 'inherit' });
+    } catch (err) {
+        console.log("Secret does not exist. Creating Secret from certificate and key...");
+        obtainCertificateAndKey(values);
+        execSync(`kubectl -n ${namespace} create secret tls api-tls-secret --cert=certificate.crt --key=private.key`, { stdio: 'inherit' });
+        fs.unlinkSync('certificate.crt');
+        fs.unlinkSync('private.key');
+    }
```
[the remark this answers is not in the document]
henokgetachew: handled.

scripts/deploy/src/install.js
henokgetachew: handled.

scripts/deploy/package.json
henokgetachew: handled.

scripts/deploy/troubleshooting/get-all-logs
henokgetachew: Good idea. Added.

scripts/deploy/troubleshooting/get-all-logs
henokgetachew: Handled.

scripts/deploy/troubleshooting/get-all-logs
henokgetachew: Probably better to link to some file with other list of requirements? Another one would be they'd need to download to an encrypted HDD. I'm inclined to leave this out of this PR and consider it in depth in the future.

scripts/deploy/troubleshooting/get-all-logs
henokgetachew: handled via mktemp.

scripts/deploy/troubleshooting/get-all-logs
mrjones-plip: Ah yeah, agreed a bigger list of requirements could be more helpful - including FDE.

However, given the incredibly low cost of adding this sentence now (<1min), why not do both?  Add this now, create a requirements doc + link later?

henokgetachew: Added.

henokgetachew:
Comments addressed. A few tests added and build is passing. Also integrated with CI. This puts us in a better place that we can iterate on.

scripts/deploy/cht-deploy
```diff
@@ -1,40 +1,69 @@
...
+        process.exit(1);
+    }
+}
+
+async function runInstallScript(args) {
+    try {
+        const valuesFilePath = args[1];
+        await install(valuesFilePath);
+    } catch (err) {
+        console.error('Error executing the install script:');
```
nydr, line 53:
```suggestion
        console.error('Error executing the install script:', err.message);
```
I assume this is was to print the message

nydr (approved):
Nice job! One small comment for displaying of errors

I'd like to see this package published to a registry or be in it's own repo so it can be called directly using npx or similar, but not a requirement for this PR, still a great step forward!

henokgetachew:
@nydr yes, I've actually already published this to `npm`. https://www.npmjs.com/package/@medic/cht-deploy. It can be run via `npx @medic/cht-deploy -f <path-to-your-yaml>`

scripts/deploy/troubleshooting/get-all-logs
```diff
@@ -0,0 +1,54 @@
...
+  echo
+  echo "Arguments:"
+  echo "  namespace   - The namespace from which to retrieve and archive pod logs."
+  echo "  since       - (Optional) Only return logs newer than a relative duration like 5s, 2m, or 3h. Defaults to all logs."
+  echo
+  echo "Example: $0 mynamespace 24h"
+  echo
+  echo "Note: Logs may contain Personally Identifiable Information (PII). Handle and store them securely."
+  exit 1
+}qq
```
mrjones-plip, line 16: calling this now I get an error instead of the `usage()` output:

```
➜  deploy git:(cht-deploy-fixes) ✗ ./troubleshooting/get-all-logs 
./troubleshooting/get-all-logs: line 55: syntax error: unexpected end of file
```

Aha! There's two errant `qq`s that need to be removed:

```suggestion
}
```

scripts/deploy/troubleshooting/get-all-logs
```diff
@@ -0,0 +1,54 @@
...
+for POD in $PODS; do
+  if [ -z "$SINCE" ]; then
+    kubectl -n $NAMESPACE logs $POD > "$LOG_DIR/$POD.log"
+  else
+    kubectl -n $NAMESPACE logs --since=$SINCE $POD > "$LOG_DIR/$POD.log"
+  fi
+done
+
+# Create a tar.gz archive of the log files
+tar -czf "$NAMESPACE-logs_$TIMESTAMP.tar.gz" -C "$LOG_DIR" .
```
mrjones-plip, line 47: This syntax causes the log files to be exploded in the current directory when uncompressed, which was a surprise.  With a small tweak we can make it be created it in a sub directory:

```suggestion
tar -czf "$NAMESPACE-logs_$TIMESTAMP.tar.gz"  "$LOG_DIR" 
```

scripts/deploy/src/install.js
```diff
@@ -0,0 +1,136 @@
...
+        ` --version ${chart_version} --namespace ${namespace}` + //NoSONAR
+        ` --values ${valuesFile} --set cht_image_tag=${image_tag}`, { stdio: 'inherit' }); //NoSONAR
+      console.log(`Instance at ${values.ingress.host} upgraded successfully.`);
+    } else {
+      console.log('Release does not exist. Performing install.');
+      const createNamespaceFlag = namespaceExists ? '' : '--create-namespace';
+      child_process.execSync(`helm install ${project_name} ${CHT_CHART_NAME}` + //NoSONAR
+        ` --version ${chart_version} --namespace ${namespace} ${createNamespaceFlag}` + //NoSONAR
+        ` --values ${valuesFile} --set cht_image_tag=${image_tag}`, { stdio: 'inherit' }); //NoSONAR
+      console.log(`Instance at ${values.ingress.host} installed successfully.`);
```
mrjones-plip, line 87: It's nice to make this a full link:

```suggestion
      console.log(`Instance installed successfully: https://${values.ingress.host}`);
```

scripts/deploy/src/install.js
```diff
@@ -0,0 +1,136 @@
...
+const DEFAULT_CHART_VERSION = '1.0.*';
+
+import { fileURLToPath } from 'url';
+import { obtainCertificateAndKey, createSecret } from './certificate.js';
+import UserRuntimeError from './error.js';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const readFile = function(f) {
```
mrjones-plip, line 19: I'm not sure if it's worth defending against, but I passed in a project_name and namespace of this:

```
project_name: mrjones-delete
namespace: "mrjones-delete"
...
   host: "mrjones-delete.dev.medicmobile.org"
```

And it created it OK apparently:

```
➜  deploy git:(cht-deploy-fixes) ✗ ./cht-deploy -f  ~/Documents/MedicMobile/medic-infrastructure/terraform/aws/dev/cht-projects/mrjones-dev-cht-deploy-values.yaml
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "medic" chart repository
Update Complete. ⎈Happy Helming!⎈
Release does not exist. Performing install.
NAME: mrjones-delete
LAST DEPLOYED: Tue May 28 15:28:12 2024
NAMESPACE: mrjones-delete
STATUS: deployed
REVISION: 1
TEST SUITE: None
Instance at mrjones-delete.dev.medicmobile.org installed successfully.
```

But I can't get to it at `https://mrjones-delete.dev.medicmobile.org/` - it just 404s. I think this is something endemic to our EKS setup and most other folks won't benefit from improvements here  - but wanted to mention it in case it was helpful!

Switching to sane values it worked as expected \o/

```
project_name: mrjones-dev
namespace: "mrjones-dev"
...
   host: "mrjones.dev.medicmobile.org"
```

mrjones-plip (changes requested):
Awesome work here! Love the follow through.

A few small things to fix up!

scripts/deploy/src/install.js
```diff
@@ -0,0 +1,136 @@
...
+  const response = await fetch(`https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:${chtversion}`);
+  const data = await response.json();
+  const tag = data.tags && data.tags.find(tag => tag.container_name === 'cht-api');
+  if (!tag) {
+    return Promise.reject(new UserRuntimeError('cht image tag not found'));
+  }
+  return tag.image.split(':').pop();
+};
+
+const get_chart_version = function(values) {
```
dianabarsan, line 62: please use camelCase for function names.

scripts/deploy/src/install.js
```diff
@@ -0,0 +1,136 @@
...
+
+const helmInstallOrUpdate = function(valuesFile, namespace, values, image_tag) { //NoSONAR
+  const chart_version = get_chart_version(values);
+  ensureMedicHelmRepo();
+  const project_name = values.project_name || '';
+  const namespaceExists = checkNamespaceExists(namespace);
+
+  try {
+    const releaseExists = child_process.execSync(`helm list -n ${namespace}`).toString(); //NoSONAR
+    if (releaseExists.includes(project_name)) { //NoSONAR
```
dianabarsan, line 74: Again, I don't think we should ever be disabling sonar on new code.

scripts/deploy/src/install.js
```diff
@@ -0,0 +1,136 @@
...
+      console.log('Release exists. Performing upgrade.');
+      child_process.execSync(`helm upgrade` + //NoSONAR
+        ` --install ${project_name} ${CHT_CHART_NAME}` + //NoSONAR
+        ` --version ${chart_version} --namespace ${namespace}` + //NoSONAR
+        ` --values ${valuesFile} --set cht_image_tag=${image_tag}`, { stdio: 'inherit' }); //NoSONAR
+      console.log(`Instance at ${values.ingress.host} upgraded successfully.`);
```
dianabarsan, line 80: We can return early here and drop the else.

scripts/deploy/src/install.js
```diff
@@ -0,0 +1,136 @@
...
+  }
+  return tag.image.split(':').pop();
+};
+
+const get_chart_version = function(values) {
+  return values.cht_chart_version || DEFAULT_CHART_VERSION;
+};
+
+const helmInstallOrUpdate = function(valuesFile, namespace, values, image_tag) { //NoSONAR
+  const chart_version = get_chart_version(values);
```
dianabarsan, line 67: please use camelCase for variables.

scripts/deploy/src/install.js
```diff
@@ -0,0 +1,136 @@
...
+  ensureMedicHelmRepo();
+  const project_name = values.project_name || '';
+  const namespaceExists = checkNamespaceExists(namespace);
+
+  try {
+    const releaseExists = child_process.execSync(`helm list -n ${namespace}`).toString(); //NoSONAR
+    if (releaseExists.includes(project_name)) { //NoSONAR
+      console.log('Release exists. Performing upgrade.');
+      child_process.execSync(`helm upgrade` + //NoSONAR
```
dianabarsan, line 76: I think it would be much cleaner to have a wrapper that runs helm commands, so that it's more visible which parameters we are passing. something like: 

```
const helmCmd = (action, params) => {
  const paramsArray = Object.entries(params).map(([ key, value ]) => {
    if (!value) {
      return `--${key}`;
    }
    return `--${key} ${value}`
  });
  return child_process.execSync(`helm ${action} ${paramsArray.join(' ')}`);
}

const opts = {
  install: `${projectName} ${CHT_CHART_NAME}`,
  version: chartVersion,
  namespace,
  set: `cht_image_tag=${image_tag}`,
};
helmCmd('install', opts);
```

scripts/deploy/src/install.js
```diff
@@ -0,0 +1,136 @@
...
+    } else {
+      console.log('Release does not exist. Performing install.');
+      const createNamespaceFlag = namespaceExists ? '' : '--create-namespace';
+      child_process.execSync(`helm install ${project_name} ${CHT_CHART_NAME}` + //NoSONAR
+        ` --version ${chart_version} --namespace ${namespace} ${createNamespaceFlag}` + //NoSONAR
+        ` --values ${valuesFile} --set cht_image_tag=${image_tag}`, { stdio: 'inherit' }); //NoSONAR
+      console.log(`Instance at ${values.ingress.host} installed successfully.`);
+    }
+  } catch (err) {
+    console.error(JSON.stringify(err));
```
dianabarsan, line 90: Is there a reason to stringify the error? 
There's a risk when stringifying unknown errors, because they might be circular objects, and then your `JSON.stringify` will throw an error. 

```
> let c = { thing: 1  };
undefined
> let a = { c };
undefined
> c.a = a;
<ref *1> { c: { thing: 1, a: [Circular *1] } }
> JSON.stringify(c)
Uncaught TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    |     property 'a' -> object with constructor 'Object'
    --- property 'c' closes the circle
    at JSON.stringify (<anonymous>)
```

scripts/deploy/src/install.js
```diff
@@ -0,0 +1,136 @@
...
+};
+
+const ensureMedicHelmRepo = function() {
+  try {
+    const repoList = child_process.execSync(`helm repo list -o json`).toString();
+    const repos = JSON.parse(repoList);
+    const medicRepo = repos.find(repo => repo.name === MEDIC_REPO_NAME);
+    if (!medicRepo) {
+      console.log(`Helm repo ${MEDIC_REPO_NAME} not found, adding..`);
+      child_process.execSync(`helm repo add ${MEDIC_REPO_NAME} ${MEDIC_REPO_URL}`, { stdio: 'inherit' }); //NoSONAR
```
dianabarsan, line 111: We can early return here.

scripts/deploy/src/install.js
```diff
@@ -0,0 +1,136 @@
...
+    } else if (medicRepo.url.replace(/\/$/, '') !== MEDIC_REPO_URL) {
+      throw new UserRuntimeError(`Medic repo found but url not matching '${MEDIC_REPO_URL}', see: helm repo list`);
+    } else {
```
dianabarsan, line 114: This else is not needed if we either return or throw in the other blocks.

scripts/deploy/src/install.js
```diff
@@ -0,0 +1,136 @@
...
+
+const ensureMedicHelmRepo = function() {
+  try {
+    const repoList = child_process.execSync(`helm repo list -o json`).toString();
+    const repos = JSON.parse(repoList);
+    const medicRepo = repos.find(repo => repo.name === MEDIC_REPO_NAME);
+    if (!medicRepo) {
+      console.log(`Helm repo ${MEDIC_REPO_NAME} not found, adding..`);
+      child_process.execSync(`helm repo add ${MEDIC_REPO_NAME} ${MEDIC_REPO_URL}`, { stdio: 'inherit' }); //NoSONAR
+    } else if (medicRepo.url.replace(/\/$/, '') !== MEDIC_REPO_URL) {
```
dianabarsan, line 112: So ... the expected correct value is `${MEDIC_REPO_URL}/` ? I don't understand why we would be removing a trailing slash in this conditional.

scripts/deploy/src/install.js
```diff
@@ -0,0 +1,136 @@
...
+
+const install = async function(f) {
+  prepare(f);
+  const values = loadValues(f);
+  const namespace = determineNamespace(values);
+  if (values.cluster_type === 'k3s-k3d') {
+    await obtainCertificateAndKey(values);
+    createSecret(namespace, values);
+  }
+  const image_tag = await get_image_tag(values.chtversion);
```
dianabarsan, line 132: Please always use camelCase.

scripts/deploy/tests/helm.test.js
```diff
@@ -0,0 +1,91 @@
...
+  const fakeImageTag = 'latest';
+
+  beforeEach(() => {
+    execSyncStub = sinon.stub(child_process, 'execSync');
+    processExitStub = sinon.stub(process, 'exit');
+  });
+
+  afterEach(() => {
+    execSyncStub.restore();
+    processExitStub.restore();
```
dianabarsan, line 31: you can just use `sinon.restore` and it will restore all stubs, you don't need to be explicit.

scripts/deploy/tests/helm.test.js
```diff
@@ -0,0 +1,91 @@
...
+    project_name: 'test-project',
+    ingress: {
+      host: 'test.host.com'
+    },
+    cht_chart_version: '5.0.0'
+  };
+  const fakeImageTag = 'latest';
+
+  beforeEach(() => {
+    execSyncStub = sinon.stub(child_process, 'execSync');
```
dianabarsan, line 25: You don't need to save these stub names. I'm pretty much against passing this variable to tests and then having trouble knowing what it is. 
You can always access it as: 
```
sinon.stub(child_process, 'execSync');
child_process.execSync.withArgs(`helm list -n ${fakeNamespace}`).returns(Buffer.from('test-project'));
```

scripts/deploy/tests/helm.test.js
```diff
@@ -0,0 +1,91 @@
...
+  it('should upgrade an existing release', () => {
+    // Mock the response to simulate existing release
+    execSyncStub.withArgs(`helm list -n ${fakeNamespace}`).returns(Buffer.from('test-project'));
+    execSyncStub.returns(Buffer.from('Upgrade successful'));
+    helmInstallOrUpdate(fakeValuesFile, fakeNamespace, fakeValues, fakeImageTag);
+    const expectedCommand = `helm upgrade` +
+        ` --install test-project medic/cht-chart-4x` +
+        ` --version 5.0.0 --namespace test-namespace` +
+        ` --values ${fakeValuesFile} --set cht_image_tag=${fakeImageTag}`;
+    expect(execSyncStub.calledWith(expectedCommand)).to.be.true;
```
dianabarsan, line 43: I think we also need to assess that execSync was not called with other arguments. 
```
expect(execSyncStub.calledOnceWith(expectedCommand)).to.be.true;
```

scripts/deploy/tests/helm.test.js
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  it('should install a new release when no release exists', () => {
+    // Mock the response to simulate no existing release
+    execSyncStub.withArgs(`helm list -n ${fakeNamespace}`).returns(Buffer.from(''));
+    execSyncStub.returns(Buffer.from('Install successful'));
```
dianabarsan, line 49: I _think_ this will overwrite your previous rule of `withArgs`. You should call them in the reverse order. 
```
child_process.execSync.returns(Buffer.from('Install successful'));
child_process.execSync.withArgs(`helm list -n ${fakeNamespace}`).returns(Buffer.from(''));
```

scripts/deploy/tests/helm.test.js
```diff
@@ -0,0 +1,91 @@
...
+    helmInstallOrUpdate(fakeValuesFile, fakeNamespace, fakeValues, fakeImageTag);
+    const expectedCommand = `helm install test-project medic/cht-chart-4x` + 
+        ` --version 5.0.0 --namespace test-namespace --create-namespace` +
+        ` --values ${fakeValuesFile} --set cht_image_tag=${fakeImageTag}`;
+    expect(execSyncStub.calledWith(expectedCommand)).to.be.true;
```
dianabarsan, line 54: ```expect(child_process.execSync.calledOnceWith(expectedCommand)).to.be.true;```

scripts/deploy/troubleshooting/get-all-logs
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+function usage() {
+  echo "Usage: $0 <namespace> [since]"
+  echo
+  echo "This script saves logs from all pods within the specified namespace into local files and then creates a tar.gz archive of these logs."
```
dianabarsan, line 6: I think it's useful to mention here that if a pod crashed, its logs will _not_ be retrieved. 
So people can't, unfortunately, use this to debug failing pods. 

I think this script can also pessimistically always try to get the logs from the previously failed pod as well as the current one.

dianabarsan (changes requested):
Cool stuff! This makes for a much easier read!

scripts/deploy/.eslintrc
```diff
@@ -0,0 +1,8 @@
+{
```
dianabarsan, line 1: I think we should find a way to have this file get linted by our default linting rules.

latin-panda:
We need this fix. What's pending so it gets merged?

scripts/deploy/.eslintrc
henokgetachew: That's what we're doing actually. These are just the overrides.

scripts/deploy/cht-deploy
```diff
@@ -1,40 +1,69 @@
...
-if [[ $1 != "-f" || -z $2 ]]; then
-    echo "No values file provided. Please specify a values file using -f <file>"
-    exit 1
-fi
-
-# Pass command line arguments to invoke script
-invoke install $@
+#!/usr/bin/env node
+
+import { install } from './src/install.js';
```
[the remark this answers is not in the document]
henokgetachew: This file was skipped. Fixed. 
The whole thing is linted by our linting rules with overrides configured here: https://github.com/medic/cht-core/blob/cht-deploy-fixes/scripts/deploy/.eslintrc

scripts/deploy/src/certificate.js
```diff
@@ -0,0 +1,48 @@
...
+    return Promise.reject(
+      new UserRuntimeError(`cert_source must be either 'specify-file-path', 'my-ip-co', or 'eks-medic'`)
+    );
+  }
+};
+
+const createSecret = function (namespace, values) {
+  console.log('Checking if secret api-tls-secret already exists...');
+  try {
+    execSync(`kubectl get secret api-tls-secret -n ${namespace}`, { stdio: 'inherit' }); //NoSONAR
```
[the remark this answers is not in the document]
henokgetachew: About using a user supplied value in a shell script. This is not a problem because the user who runs the script is the user who provides this value under his own user/credentials and access levels.

latin-panda:
@henokgetachew @mrjones-plip 
I know it's been busy lately. Is there anything we can do to help this PR? 

We are currently pulling from a commit (`4760b5d30017a86d90ae95b1d570ba9ed6d053dc`) to keep using EKS instances. Especially to verify issues during releases in CHT, so fixing and merging this PR will help the release managers to not rediscover the same issue ` error: InvalidImageName` in the upgrade service. 

Screenshot 2024-08-15 at 4 55 04 PM

sugat009:
I can confirm that this worked for me when testing out the upgrade to release 4.10.0 when the demo instance was not working.

henokgetachew:
This is actually on my plan for today. Until the merge by the way, you should be able to use:
`npx @medic/cht-deploy@latest -f <your-yaml.>`

henokgetachew:
Agreed. That's the plan.

scripts/deploy/troubleshooting/get-all-logs
henokgetachew: Makes sense. Added.

scripts/deploy/src/install.js
henokgetachew: It's to account for helm behavior where the trailing slash exists in some versions and not in others.

scripts/deploy/src/install.js
henokgetachew: Thanks. I like this. Changed.

mrjones-plip (approved):
Nice work on this!  It was a really big effort and we're much better off because of the features added, bugs fixed and totally rewriting it in node.

I'm approving PR after testing the happy paths and not seeing anything amiss.  I note some room for improvements below, but let's merge this! 

-----

### defending against old node 

This works but only if you're running newer node.  Here's what I see running node 12 vs 18:

```
➜  deploy git:(cht-deploy-fixes) node --version
v12.22.12

➜  deploy git:(cht-deploy-fixes) ./cht-deploy  
internal/process/esm_loader.js:74
    internalBinding('errors').triggerUncaughtException(
                              ^

TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension "" for /var/home/mrjones/Documents/MedicMobile/cht-core/scripts/deploy/cht-deploy
    at Loader.defaultGetFormat [as _getFormat] (internal/modules/esm/get_format.js:65:15)
    at Loader.getFormat (internal/modules/esm/loader.js:116:42)
    at Loader.getModuleJob (internal/modules/esm/loader.js:247:31)
    at async Loader.import (internal/modules/esm/loader.js:181:17)
    at async Object.loadESM (internal/process/esm_loader.js:68:5) {
  code: 'ERR_UNKNOWN_FILE_EXTENSION'
}

➜  deploy git:(cht-deploy-fixes) ✗ nvm use 18
Now using node v18.20.4 (npm v10.7.0)

➜  deploy git:(cht-deploy-fixes) ✗ ./cht-deploy 
Invalid Node.js version. Required: >=20.11.0. Current: v18.20.4
```

### not being logged in

This could be smoother.  I don't know if there's an explicit way to check if the user is logged in  and just show "please run `eks-aws-mfa-login` before proceeding", but this is a bit messy:

```

➜  deploy git:(cht-deploy-fixes) ✗ ./cht-deploy -f mrjones-dev.yml 
Helm repo medic not found, adding..
"medic" has been added to your repositories
E0816 15:08:49.499059  101343 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: the server has asked for the client to provide credentials"
E0816 15:08:50.680340  101343 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: the server has asked for the client to provide credentials"
E0816 15:08:51.848484  101343 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: the server has asked for the client to provide credentials"
E0816 15:08:53.082776  101343 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: the server has asked for the client to provide credentials"
E0816 15:08:54.262058  101343 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: the server has asked for the client to provide credentials"
error: You must be logged in to the server (the server has asked for the client to provide credentials)
Error: Kubernetes cluster unreachable: the server has asked for the client to provide credentials
Command failed: helm list -n mrjones-dev
Error: Kubernetes cluster unreachable: the server has asked for the client to provide credentials

Error: Command failed: helm list -n mrjones-dev
Error: Kubernetes cluster unreachable: the server has asked for the client to provide credentials

    at genericNodeError (node:internal/errors:984:15)
    at wrappedFn (node:internal/errors:538:14)
    at checkExecSyncError (node:child_process:890:11)
    at Object.execSync (node:child_process:962:15)
    at helmInstallOrUpdate (file:///var/home/mrjones/Documents/MedicMobile/cht-core/scripts/deploy/src/install.js:88:41)
    at install (file:///var/home/mrjones/Documents/MedicMobile/cht-core/scripts/deploy/src/install.js:165:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async runInstallScript (file:///var/home/mrjones/Documents/MedicMobile/cht-core/scripts/deploy/cht-deploy:55:5)
    at async main (file:///var/home/mrjones/Documents/MedicMobile/cht-core/scripts/deploy/cht-deploy:66:3)
```

### invalid password

This is hard to defend against, but it seems not good to accept a `values.yml` with a value of `password: BankLogin!3`.  While I know we have an extensive section on how to use a valid password in our docs, the system just left with a `503`, and `cht-api` pod not running and this error in the logs:

```
2024-08-16T22:12:41.512 INFO: Configuration loaded successfully 
2024-08-16T22:12:41.513 INFO: Running db migrations… 
2024-08-16T22:12:41.537 INFO: Running migration associate-records-with-people... 
2024-08-16T22:12:41.545 ERROR: Migration associate-records-with-people failed 
2024-08-16T22:12:41.546 ERROR: Fatal error initialising API 
2024-08-16T22:12:41.549 ERROR: StatusCodeError: 401 - {"error":"unauthorized","reason":"Name or password is incorrect."}
    at new StatusCodeError (/service/api/node_modules/request-promise-core/lib/errors.js:32:15)
    at Request.plumbing.callback (/service/api/node_modules/request-promise-core/lib/plumbing.js:104:33)
    at Request.RP$callback [as _callback] (/service/api/node_modules/request-promise-core/lib/plumbing.js:46:31)
    at Request.self.callback (/service/api/node_modules/request/request.js:185:22)
    at Request.emit (node:events:513:28)
    at Request.<anonymous> (/service/api/node_modules/request/request.js:1154:10)
    at Request.emit (node:events:513:28)
    at IncomingMessage.<anonymous> (/service/api/node_modules/request/request.js:1076:12)
    at Object.onceWrapper (node:events:627:28)
    at IncomingMessage.emit (node:events:525:35) {
  name: 'StatusCodeError',
  statusCode: 401,
  message: '401 - {"error":"unauthorized","reason":"Name or password is incorrect."}',
  error: { error: 'unauthorized', reason: 'Name or password is incorrect.' },
  [stack]: 'StatusCodeError: 401 - {"error":"unauthorized","reason":"Name or password is incorrect."}\n' +
    '    at new StatusCodeError (/service/api/node_modules/request-promise-core/lib/errors.js:32:15)\n' +
    '    at Request.plumbing.callback (/service/api/node_modules/request-promise-core/lib/plumbing.js:104:33)\n' +
    '    at Request.RP$callback [as _callback] (/service/api/node_modules/request-promise-core/lib/plumbing.js:46:31)\n' +
    '    at Request.self.callback (/service/api/node_modules/request/request.js:185:22)\n' +
    '    at Request.emit (node:events:513:28)\n' +
    '    at Request.<anonymous> (/service/api/node_modules/request/request.js:1154:10)\n' +
    '    at Request.emit (node:events:513:28)\n' +
    '    at IncomingMessage.<anonymous> (/service/api/node_modules/request/request.js:1076:12)\n' +
    '    at Object.onceWrapper (node:events:627:28)\n' +
    '    at IncomingMessage.emit (node:events:525:35)'
} 
```

PR #9082, opened 2024-05-01
chore(#8437): set strictBindCallApply and strictPropertyInitialization in tsconfig

jkuester:
# Description

Continuing to chip away at https://github.com/medic/cht-core/issues/8437 so we can get `strict` mode enabled on our TS in webapp!

We had no violations of `strictBindCallApply` so nothing had to change to apply this.

For `strictPropertyInitialization`, I had to make some changes to the type definitions for a number of fields. Basically, when you set a type on a field, TS wants that field to be _initialized_ (either immediately or in the constructor).  In our Angular code, some of these fields are initialized later in the Angular lifecycle (e.g. during the `ngOnInit`) or the field is actually only ever initialized in certain workflows and otherwise is `undefined`.  

So, in most cases, I have marked the field as possibly `undefined` by marking it with `?`.  Most of the time this had no down-stream effects (since the other code already protected against this case). In a few places, I had to add some [optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining) for extra protection.  

In a couple special cases, I had to use the [definite assignment assertion](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#definite-assignment-assertions) (`!`) instead. This was because the code referencing the field was very dependent on that field not being `undefined` (to the point that it was clear from the structure of the existing code that the field always ended up being initialized before it was used). This is not the _safest_ thing to do from a typing perspective, but it seems like a reasonable compromise for existing code.

Most importantly, none of these changes should have any affect on the actual code at all!  These should all be type-space changes. They will affect the type that gets inferred, but should not change the JS code that ultimately gets transpiled...

# Code review checklist

# License

### DIFF
--- a/webapp/src/ts/components/fast-action-button/fast-action-button.component.ts
+++ b/webapp/src/ts/components/fast-action-button/fast-action-button.component.ts
@@ -19,8 +19,8 @@ import { Selectors } from '@mm-selectors/index';
 })
 export class FastActionButtonComponent implements OnInit, OnDestroy {
 
-  @Input() config: FastActionConfig;
-  @Input() fastActions: FastAction[];
+  @Input() config?: FastActionConfig;
+  @Input() fastActions?: FastAction[];
   @ViewChild('contentWrapper') contentWrapper;
 
   private subscriptions: Subscription = new Subscription();
@@ -75,7 +75,7 @@ export class FastActionButtonComponent implements OnInit, OnDestroy {
    * Returns a Fast Action that can be executed right away without opening the dialog or bottom sheet.
    */
   getFastExecutableAction(): FastAction | undefined {
-    if (this.fastActions.length === 1 && !this.fastActions[0]?.alwaysOpenInPanel) {
+    if (this.fastActions?.length === 1 && !this.fastActions[0]?.alwaysOpenInPanel) {
       return this.fastActions[0];
     }
     return;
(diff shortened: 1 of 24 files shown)


Benmuiruri (approved):
LGTM! 

Just two comments 
1. I am curious how you are finding these. Are you using one of [Mokhtar's suggestions](https://github.com/medic/cht-core/issues/8437#issuecomment-1875622964), are you setting `strictPropertyInitialization` then fixing everywhere it is flagged ? Or are you reading one file after another ? 
2. I find [definite assignment assertion](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#definite-assignment-assertions) to be strange because if I understand it correctly you can technically declare a property with the definite assignment assertion (!) and then fail to assign it a value before attempting to access it.

jkuester:
Thanks @Benmuiruri!  To answer your questions:

1. Yes, basically I just set the config option, ran the build, and saw where it flagged the failures. Not fancy, but it worked at this scale.
2. Yes, I think it is fair to categorize the definite property assertion as a "code smell".  Essentially you are overwriting the TS system to say "I promise that I know better"!  This was why I preferred to mark the fields as possibly undefined (via `?`) whenever that was feasible. It seemed more correct from the strict TS perspective.  That being said, I think a situation where a property is being initialized during the Angular lifecycle can also be a valid use-case for the definite property assertion.  TS does not know that it your field will always be initialized by `ngOnInit`, but you do. So, using `!` in this case is just telling TS what you already know.  (Now, of course, the risk here is that someone might add code referencing the field that runs _before `ngOnInit`_. This is the tradeoff.  In general, I think we should prefer to avoid the smelly definite property assertions when adding new code. As I mentioned above, though, I took the compromise approach for this PR to avoid spiraling into an unnecessarily large refactoring initiative.)

Issue #9080, opened 2024-05-01
Add @medic/logger shared library


dianabarsan:
**Describe the issue**
The winston configuration for API and Sentinel is straight up duplicated code: 
https://github.com/medic/cht-core/blob/master/api/src/logger.js
https://github.com/medic/cht-core/blob/master/sentinel/src/lib/logger.js

Some shared libraries already use the loggers from API and Sentinel, but we pass them as parameters when we initialize these libraries. Ex:
https://github.com/medic/cht-core/blob/master/api/src/services/config-watcher.js#L45
https://github.com/medic/cht-core/blob/master/shared-libs/transitions/src/transitions/mark_for_outbound.js#L14

The reason for this is that shared libs can (or should be able to) be used across all apps, including browser based webapp. Winston only works in NodeJs. 

**Describe the improvement you'd like**
Add a shared library that exports a logger than can be reused as-is, to remove the need of passing it as a parameter across apps. 
This logger should be environment aware and not load winston if it runs in a browser. 

**Describe alternatives you've considered**
None.

PR #9083, opened 2024-05-01
chore(#9080): adds logger shared lib

dianabarsan:
# Description

#9080

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8338-environment-shared-lib/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8338-environment-shared-lib/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8338-environment-shared-lib/docker-compose/cht-couchdb-clustered.yml)

# License


shared-libs/logger/src/index.js
```diff
@@ -0,0 +1,82 @@
+if (!process || process.browser) {
+  module.exports = console;
+} else {
```
garethbowen, line 3: Having different `module.exports` code depending on this branch makes it really hard to work out what it's going to do, especially if you're calling it from another shared lib and you don't know what context you're in. I think it'd be easier to read if it was something like...

```js
const create() => {
  if (!process || process.browser) {
    return console;
  } else {
    return require('./node-logger.js').create();
  }
};

module.exports = create();
module.exports.DATE_FORMAT = DATE_FORMAT;
```

shared-libs/outbound/src/outbound.js
```diff
@@ -280,41 +278,37 @@ const alreadySent = (payload, configName, recordInfo) => {
...
-            .then(() => logger.info(`Pushed ${record._id} to ${configName}`))
-            .then(() => true);
-        })
-        .catch(err => {
-          logSendError(configName, record._id, err);
-          throw err;
-        });
-    },
-  };
+module.exports = {
```
garethbowen, line 281: :heart:

shared-libs/transitions/test/unit/transitions/mark_for_outbound.js
```diff
@@ -1,19 +1,19 @@
-const rewire = require('rewire');
```
garethbowen, line 1: Woohoo! Death to rewire!

garethbowen (changes requested):
One refactor suggestion, but otherwise this is great!

shared-libs/transitions/test/unit/transitions/mark_for_outbound.js
dianabarsan: yes! i concur!

garethbowen:
@dianabarsan One more random thought... in node services we often log like `log.error("got error: %o", err)` - but that won't work in the browser, right? Do we have to just be careful to not use that format in shared-libs?

dianabarsan:
> in node services we often log like log.error("got error: %o", err)

Correct, this kind of formatting won't work in the browser. Nothing is going to break, but we won't achieve the desired result. I suppose we can continue to use this format when we know the code will only be used in nodejs - in libraries like server-checks for example.

garethbowen:
@dianabarsan What about this: https://www.npmjs.com/package/winston-transport-browserconsole ? It might mean the logging API is consistent between node and browser?

dianabarsan:
> @dianabarsan What about this: https://www.npmjs.com/package/winston-transport-browserconsole ? It might mean the logging API is consistent between node and browser?

I've read about it. People said it bulks the browser bundle. I'll try it out though.

dianabarsan:
It looks like `winston-transport-browserconsole` it doesn't work out of the box:

image

It looks like `setImmediate` has been deprecated for a while: https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate and the library hasn't received any update in years. I'm reluctant to use it. 

Do you think it's worth more time working out how to generalize logging in webapp in this PR?

garethbowen (approved):
Awesome, thanks!

Issue #9065, opened 2024-04-24
API Endpoint for getting person by UUID


jkuester:
NOTE: the get-place workflow has been split into https://github.com/medic/cht-core/issues/9194

---

**Is your feature request related to a problem? Please describe.**

Currently the only REST api options for getting contact data are:

- [`GET /api/v2/export/contacts`](https://docs.communityhealthtoolkit.org/apps/reference/api/#get-apiv2exportcontacts) which returns a little bit of data about all the contacts
- [`GET /api/v1/contacts-by-phone`](https://docs.communityhealthtoolkit.org/apps/reference/api/#get-apiv1contacts-by-phone) - which returns a list of fully hydrated contacts for a given phone number

A more basic API is needed for being able to retrieve contact data for just a specific contact. (e.g. you just got the data for a user via `GET /api/v2/users/{{username}}` and now you have their `contact_id` value, but you want to get all the data for that contact.)

**Describe the solution you'd like**

We already have [`POST /api/v1/places`](https://docs.communityhealthtoolkit.org/apps/reference/api/#post-apiv1places) and [`POST /api/v1/places`](https://docs.communityhealthtoolkit.org/apps/reference/api/#post-apiv1places), so we could have `GET /api/v1/places/{{id}}` and `GET /api/v1/people/{{id}}`.

**Additional context**

As noted in https://github.com/medic/cht-core/issues/8889, we should implement the core data access aspects of this via a new data source library that can be shared between API and webapp with a well defined interface.

cc @m5r @mrjones-plip

jkuester:
Okay, I figured I would throw out a bit of a technical design.  This is all entirely speculative (riffing on the content/discussion in [this doc](https://docs.google.com/document/d/1opDXST8-NTqRUl716UPn0N_hz2syXPq1B4YDagl7yUY/edit)), but I figured the most practical way to proceed here was to start with a specific proposal that we could discuss.

@garethbowen the main goal here is to hammer out an over-arching design. Once we have an idea of what that should look like, I can iterate on this proposal with more detail.  I am happy to jump on a call to discuss this further if that would be more effective!

## Tech Design

The actual `api` code for this issue is trivial, so this proposal is mostly focused on the structure/content of the new "Webapp API library" (that I am calling `cht-datasource`).

### `cht-datasource`

It seems to make sense to put this new lib into the `shared-libs` folder where it can be referenced by api/sentinel/webapp.

- `index.ts`

	```ts
	const init = (sourceDb) => {...}
[...]

PR #9090, opened 2024-05-02
feat(#9065): add cht-datasource to support get person by uuid

jkuester:
# Description

https://github.com/medic/cht-core/issues/9065

## cht-datasource

This PR converts `shared-libs/cht-script-api` into `shared-libs/cht-datasource` as a general purpose data-access layer intended for use within cht-core both in server and the client-side code as well as within custom config code for (tasks, targets, etc).

There are two modes of functionality exported from `cht-datasource`.  Both require you to first acquire a `DataContext` (either `remote` or `local` depending on if you can interact directly with the CHT api server or if offline functionality is required).  

```ts
import { getRemoteDataContext, getLocalDataContext } from '@medic/cht-datasource';

const dataContext = isOnlineOnly
  ? getRemoteDataContext(...)
  : getLocalDataContext(...);
```

### Imperative

The imperative mode provides a hierarchical factory object that you can drill down into and interact with the datasource.

```ts
import { getDatasource } from '@medic/cht-datasource';

const datasource = getDatasource(dataContext);
const myUuid = 'my-uuid';
const myPerson = await datasource.v1.person.getByUuid(myUuid);
```

This mode is useful for exposing functionality to custom configuration code (tasks, targets, etc) by passing the datasource object directly into the custom function.

### Declarative

The declarative mode provides more flexible access to API functions.  The `dataContext` must be curried when using the functions.

```ts
import { Person, Qualifier } from '@medic/cht-datasource';

const getPerson = Person.V1.get(dataContext);
const myUuid = 'my-uuid';
const myPerson = await getPerson(Qualifier.byUuid(uuid));
```

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9065_get_contact_by_id/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9065_get_contact_by_id/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9065_get_contact_by_id/docker-compose/cht-couchdb-clustered.yml)

# License


shared-libs/cht-datasource/src/index.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+const getDataSource = async (sourceDatabases?: SourceDatabases) => {
+  const localEnv = await getLocalEnvironment(sourceDatabases);
+
+  return {
+    v1: {
+      hasPermissions,
+      hasAnyPermission,
+      person: adapt(localEnv, PersonV1.remote, PersonV1.local),
```
garethbowen, line 39: I'm slightly on the fence on this one. I think there are two options...
1. what you've done - Person knows how to get a person from local or remote dbs, or
2. what pouchdb does - local and remote are datasource adapters that implement the same interface (getPerson)

This distinction doesn't change the external API at all - this is all internal to the datasource lib.

I prefer option 2 personally. I think it's a better separation of concerns, ie: Person knows about people, the remote datasource cares about http, the local datasource cares about pouch.

It feels a little dirty because it's a second layer of abstraction but on balance I think it's cleaner.

jkuester: So, funny story, I was originally trying to avoid #2 for a couple reasons:

- I thought maybe I would push the local/remote abstraction down so it was internal to `person.ts`
- I did not want to have to maintain a common interface

As I iterated on this code, I decided it seemed better to make the local/remote distinction at the highest level possible and not encapsulate it in `person.ts`. Then I also ended up with a common interface anyway since that was the best way to keep everything in sync.  

Thinking about this all again, I totally agree with you that the code will be a lot cleaner if we separate local/remote into different files (and maybe different directories)....  In reality, they will not be sharing code anyway.

shared-libs/cht-datasource/src/local/libs/local-environment.ts
```diff
@@ -0,0 +1,40 @@
...
+      settings = await getSettingsDoc(medicDb);
+    }
+  });
+
+export interface LocalEnvironment {
+  medicDb: PouchDB.Database<Doc>;
+  getSettings: () => Doc;
+}
+
+export const getLocalEnvironment = async function(sourceDatabases?: SourceDatabases): Promise<Nullable<LocalEnvironment>> {
```
m5r, line 28:
```suggestion
export const getLocalEnvironment = async (sourceDatabases?: SourceDatabases): Promise<Nullable<LocalEnvironment>> => {
```

minor nitpick but let's stay consistent on the syntax

shared-libs/cht-datasource/src/libs/person.ts
```diff
@@ -0,0 +1,9 @@
+import { v1 as doc } from './doc';
+const { isUuidIdentifier } = doc;
+
+export const validateIdentifier = (identifier: unknown): identifier is V1.UuidIdentifier => {
+  if (isUuidIdentifier(identifier)) {
+    return true;
+  }
+  throw new Error('Invalid identifier');
+};
```
m5r, line 9:
```suggestion
// eslint-disable-next-line func-style
export function validateIdentifier(identifier: unknown): asserts identifier is V1.UuidIdentifier {
  if (!isUuidIdentifier(identifier)) {
    throw new Error('Invalid identifier');
  }
}
```

we can leverage typescript's [assertion functions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions) to nudge the control flow analysis and make it understand more easily the type of the variable
it needs to be defined as a regular `function` though, assertion functions [don't support arrow functions](https://github.com/microsoft/TypeScript/issues/34523)

shared-libs/cht-datasource/src/remote/libs/remote-environment.ts
```diff
@@ -0,0 +1,9 @@
+export const fetchIdentifiedResource = async <T>(resourcePath: string, identifier: string): Promise<Nullable<T>> => {
+  const response = await fetch(`${resourcePath}/${identifier}`);
```
m5r, line 2: What's our nodejs support for CHT 4.x situation again? I couldn't find any info in our docs
I see that fetch was shipped with node v18

m5r (commented):
+1 on what Gareth said about the API design it makes more sense to have the two functions implement the same interface. I wish the TS team had implemented https://github.com/microsoft/TypeScript/issues/420, it would have been pretty nice in our scenario but we can surely make `cht-datasource` work without!

shared-libs/cht-datasource/src/libs/doc.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const getDocsByIds = (db: PouchDB.Database<Doc>) => async (ids: NonEmptyArray<string>): Promise<Doc[]> => {
+  const results = await db.allDocs({ keys: ids, include_docs: true });
+  return results.rows
+    .map(row => row.doc)
+    .filter(isDoc);
+};
+
+export const getDocById = (db: PouchDB.Database<Doc>) => async (uuid: string): Promise<Nullable<Doc>> => {
+  return (await getDocsByIds(db)([uuid]))[0];
```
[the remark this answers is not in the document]
jkuester: Honestly I originally used `allDocs` just because I was copying code from `shared-libs/lineage` and that function was handling the case of grabbing multiple docs.  Switched this back up to use `get`.

shared-libs/cht-datasource/src/libs/person.ts
jkuester: Ahh fancy!  I did not know about assertion functions!

Also looks like things work with arrow functions if you give enough context in the signature:

```
  const assertPersonQualifier: (qualifier: unknown) => asserts qualifier is UuidQualifier = (qualifier: unknown) => {
    if (!isUuidQualifier(qualifier)) {
      throw new Error(`Invalid identifier [${JSON.stringify(qualifier)}].`);
    }
  }
```

shared-libs/cht-datasource/src/remote/libs/remote-environment.ts
jkuester: We are rolling [Node 20](https://github.com/medic/cht-core/blob/900218e9b4594c11e01809d1f1333cd955e7d310/api/Dockerfile#L6)!!!!!

garethbowen: Last time I checked fetch isn't API locked until Node 22 so I've been holding off.

Does `resourcePath` include the host and credentials? If so this will have issues with basic auth and you should use the pouchdb instance which has the cookie auth plugin to fetch remotely instead.

jkuester: > Last time I checked fetch isn't API locked until Node 22 so I've been holding off.

Ahh, good to know.  (FWIW, looks like it actually became stable [in Node 21](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V21.md#21.0.0), but still not helpful for us at this moment.)

> Does resourcePath include the host and credentials? If so this will have issues with basic auth and you should use the pouchdb instance which has the cookie auth plugin to fetch remotely instead.

@garethbowen the idea of the `remote` adapter here was to be able to leverage our REST apis and not require Pouch _on the client side_.  (See my [data flow diagram](https://github.com/medic/cht-core/issues/9065#issuecomment-2090866133) for how this fits together in my head...)  Honestly I have been kicking the authentication questions down the road till now because I was not sure how the specifics should work here.  FWIW, when I originally smoke tested making `fetch` calls from webapp code, I found that I could just run something like `fetch('api/v1/person/my-uuid')` and it would somehow magically make the request with the correct root url and credentials (I assume it was using the session cookie somehow).  Of course, things will be different for sentinel/3rd-party apps....

I assume, for the HTTP requests to the API, that we can use `request-promise-native` instead of `fetch`, but that will not solve the authentication problem.  :thinking:   I plan to do more digging here, but def welcome any thoughts/suggestions!

admin/src/js/services/auth.js
```diff
@@ -1,4 +1,5 @@
-const chtScriptApi = require('@medic/cht-script-api');
+const cht = require('@medic/cht-datasource');
+const chtDatasource = cht.getDatasource(cht.getRemoteDataContext());
```
jkuester, line 2: Need to decide if Sentinel should use the `remote` context and talk to api, of if we need to be passing in the `local` Pouch dbs, etc.  On one hand, we have Pouch already in context here in Sentinel, but on the other hand, we should be able to leverage the REST endpoints for all the cht-datasource interactions (since we don't have to consider offline functionality...).

shared-libs/cht-datasource/src/auth.js
jkuester, line 1: Had to refactor this file a bit to make Sonar happy (original functions were too complex...)

shared-libs/cht-datasource/src/libs/core.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const hasField = (value: Record<string, unknown>, field: { name: string, type: string }): boolean => {
+  const valueField = value[field.name];
+  return typeof valueField === field.type;
+};
+
+export const hasFields = (
+  value: Record<string, unknown>,
+  fields: NonEmptyArray<{ name: string, type: string }>
+): boolean => fields.every(field => hasField(value, field));
```
jkuester, line 31: Not super excited to have logic like this.  I think it is important to be able to check the shape of data provided at the edges of the library (e.g. getting passed in as an argument or coming back from Couch).  I think a "schema validation" library (e.g. https://zod.dev/) basically does this (and a lot more).  We should consider pulling in something like that, but perhaps not for this MVP... :thinking:  :shrug:

webapp/tsconfig.base.json
```diff
@@ -11,6 +11,7 @@
     "importHelpers": true,
     "target": "es2020",
     "module": "es2020",
+    "skipLibCheck": true,
```
jkuester, line 14: I had to put this in because I started getting TS compile errors coming from PouchDB types here in webapp. I am not exactly sure what changed so that this started being a problem.  :thinking:   Perhaps is related to now Pouch types are also transitive deps from `@medic/cht-datasource`....  :shrug:   Either way, I guess we don't really need to be validating the types for deps anyway...

jkuester:
@m5r @sugat009 this is ready for a proper review now! I updated the PR description with a bit of a high-level intro, so have a look at that.  The implementation code and unit tests in cht-datasource for _retrieving a person without lineage from a local data context_ are complete.  I have stubbed in some code for getting a person from the _remote context_, but the implementation details [need more investigation](https://github.com/medic/cht-core/pull/9090#discussion_r1602248215).  

The goal is to get this functionality reviewed and stable. Then it will form the base for the branch and future PRs can add:

- Proper support for the _remote context_
- Getting a person with lineage
- Getting a place with and without lineage
- Adding new endpoints to `api` 
- Adding new integration tests
- Updating various parts of the codebase to use cht-datasource for getting contacts

shared-libs/cht-datasource/src/remote/libs/remote-environment.ts
garethbowen: Sorry you're absolutely right - ignore my auth comment. Requests sent from the browser will get the cookie auth automatically.

I think I was getting ahead of myself thinking about how this could also be a wrapper tool for Node based scripts because it'll be easier to deal with than pure REST (ie: Oktokit to GitHubs REST API) but this is way out of scope for today.

shared-libs/cht-datasource/src/libs/core.ts
m5r: I had great experiences with zod in the past. We use [joi](https://joi.dev/) in cht-conf which is pretty similar but for typescript codebases zod has the upper hand. You're right that we don't need it right now, it's an implementation detail that can easily be switched out at a later point :+1:

admin/src/js/services/auth.js
m5r: My immediate gut feeling would be to use the `remote` context, we don't need the offline-first functionality of pouch here

shared-libs/cht-datasource/src/libs/core.ts
jkuester: Have to say, it seems pretty rare to hear a developer say they "had great experiences" with a particular library!  We all have a million libs that we hate to use, but ones that were great?  Those are precious!  :laughing:

m5r (changes requested):
Amazing progress! I left a few thoughts and change requests :eyes:

shared-libs/cht-datasource/src/libs/context.ts
```diff
@@ -0,0 +1,68 @@
...
+    throw new Error(`Invalid settings service [${JSON.stringify(settings)}].`);
+  }
+};
+
+const assertSourceDatabases: (sourceDatabases: unknown) => asserts sourceDatabases is SourceDatabases =
+  (sourceDatabases: unknown) => {
+    if (!isRecord(sourceDatabases) || !hasField(sourceDatabases, {name: 'medic', type: 'object'})) {
+      throw new Error(`Invalid source databases [${JSON.stringify(sourceDatabases)}].`);
+    }
+  };
```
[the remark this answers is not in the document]
jkuester: Just so I am clear on what you are worried about here, is it around the janky `isRecord` and `hasField` stuff?  I feel like the general idea of these `asserts sourceDatabases is SourceDatabases` is solid since it fills a dual purpose.  It both allows for nudging the TS type compiler when needed, but more importantly provides a semantically proper way to do some schema validation for arguments coming in at the edge.  Let me know if you are seeing things differently here though!

Then, regarding the `isRecord` and `hasField` stuff, that just goes back to eventually [offloading schema validation](https://github.com/medic/cht-core/pull/9090#discussion_r1602262068) to a library...

shared-libs/cht-datasource/src/local/libs/doc.ts
```diff
@@ -0,0 +1,12 @@
+import { Nullable } from '../../libs/core';
+import { Doc, isDoc } from '../../libs/doc';
+
+export const getDocById = (db: PouchDB.Database<Doc>) => async (uuid: string): Promise<Nullable<Doc>> => db
```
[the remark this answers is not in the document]
jkuester: I have gone back and forth with this a couple times.  My main concern is that we don't accidentally change the return types of _functions exported and used outside of cht-datasource_.  That could be a bit tricky to detect and lead to some interesting debugging.  

Currently, I am [using the `explicit-module-boundary-types` eslint rule](https://github.com/medic/cht-core/blob/5196de4b1cb132c90cb996ab2d705d45907e58b0/shared-libs/cht-datasource/.eslintrc.js#L18) to require return types on _exported_ functions (applies to all exported functions, not just ones that end up getting exported from `index.ts`, but it was the best I could find).  The [rule docs](https://typescript-eslint.io/rules/explicit-module-boundary-types/) do note that return types on module boundaries "can also improve TypeScript type checking performance on larger codebases". Not something I am too concerned about at this point, but is a nice feature...

 At the end of the day, I feel like our unit tests should be a sufficient backstop to prevent us from accidentally breaking things with regards to the data returned from public functions (and we def should not be heavily relying on the _types_ to save us here... :sweat_smile: ). So, I am happy to go either way. Let me know if the `explicit-module-boundary-types` seems valuable given the above rational, otherwise I can just remove it!  (Another interesting datapoint is that `explicit-module-boundary-types` was not included in the `strict-type-checked` or the `stylistic-type-checked` eslint rulesets which tells me there is not broad consensus that it is a "best practice".)

shared-libs/cht-datasource/src/libs/contact.ts
```diff
@@ -0,0 +1,15 @@
...
+  parent?: DehydratedParent;
+}
+
+export interface Contact extends Doc {
+  contact_type?: string;
+  name?: string;
+  parent?: DehydratedParent;
+  reported_date?: Date;
+  type: string;
+}
```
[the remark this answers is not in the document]
jkuester: So, my thought with these interfaces is that conceptually they should not depend on the db (or on data needed for querying the db). Instead they just represent the contract with any consumers of cht-datasource of what we _guarantee_.  Unfortunately, for `V1` of the API, we cannot really guarantee anything since the existing code we are trying to replace does not do much schema validation when looking up data.  That is why I ended up making most of these fields optional here since the existing code does not do any kind of checking that the contact actually has a `name` value.  In `V2`, we should definitely come up with a list of _required_ fields that we would check for when pulling data from the DB. Then our V2 interfaces can guarantee those fields to be valid.

Does that make sense, or have I missed your point entirely?  :sweat_smile:

jkuester:
I expect to trigger some folks with the latest commit which adds the `jsdoc/require-jsdoc` eslint rule, requiring docs on _all exported entities_, but hear me out!  :sweat_smile: 

We want docs on our public API. AFICT, there is no way to just require docs on stuff that gets exported publicly from the cht-datasource module (eslint does not seem to be that magically smart).  So, the next best thing is just requiring docs on anything that gets exported from a file.  This might feel like overkill, but as I was going through the errors, I started marking the stuff that I knew was not exported public from cht-datasource as `/** @internal */`.  I ended up really liking the result because, before hand, it was pretty confusing to understand if a function exported from a file ever ended up being exported from the whole library or not.  That info is important because it determines if you can trust the arguments being passed in or not. Internal functions do not need to check their args (and can assume they match the declared types). Publicly accessible functions should always check the args.

All in all, I am happy to continue discussing this and there are lots of various options and tweaks that can be done for the linting, but it seemed like a good place to start....

shared-libs/cht-datasource/src/local/libs/doc.ts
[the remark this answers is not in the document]
m5r: So I went back to [this video](https://www.totaltypescript.com/tips/dont-use-return-types-unless) (the author made a great typescript course!) and he summarized it at the end:
> To sum up, in application code, don't by default use return types. Use them if your function has multiple branches. Use them if you're writing a library, definitely. Use them if you have any weird niche performance concerns to do with really big objects being passed and inferred through your code.

In the video he made an example of why exported library functions should be typed and that made a lot of sense! Plus, by explicitly typing these functions, we get an additional safety layer for javascript projects that import the compiled code without any types. Let's keep it that way :+1:

shared-libs/cht-datasource/src/libs/contact.ts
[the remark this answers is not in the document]
m5r: Yes it totally makes sense, thanks for sharing your thought process here :raised_hands:

garethbowen:
> I expect to trigger some folks... 

You got me! I agree that all public APIs should have good and validated docs, but in that commit there's a whole lot of JSDoc that's been added just to satisfy the linter and adds no value, eg:

```
    /**
     * Version 1 of the datasource.
     */
    v1: {
```

```
      /**
       * Person datasource.
       */
      person: {
```

```
/**
 * Service providing access to the app settings.
 */
export type SettingsService = Readonly<{ getAll: () => Doc }>;
```

This makes the code much harder to read because much less of it fits on a screen at a time. Also, it trains people to ignore jsdocs when most of the time they're useless.

This reminds me of 99% of javadocs...

```java
/**
 * Gets the foo
 * @returns foo
 */
public Foo getFoo() { ... }
```

Unless we can tweak the linter to only fail when appropriate let's leave this to the humans (coder + reviewer) to catch.

shared-libs/cht-datasource/src/libs/context.ts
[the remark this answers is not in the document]
m5r: Yes my concern is that the validation is kind of fragile but there is no way to validate a function or an object "identity" in JS :man_shrugging: 
Let's keep it that way and if we come up with a better approach along the way we can always come back and change this code

shared-libs/cht-datasource/src/local/libs/doc.ts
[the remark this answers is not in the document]
jkuester: :+1: I am going to start a list of "do's" and "don'ts" for TS code in medic repos where I will capture some of these design thoughts.  Maybe can put it out in the Dev Handbook on the docs site...

jkuester:
:+1: This has been all part of my evil plan to start with crazy amounts of jsdoc and then agree to "moderate" things back to where I actually wanted them.... :laughing:   

But seriously, :100: agree regarding the unnecessary redundant docs. Most of those got added late on Friday afternoon in my feverish zeal for consistency :sweat_smile:, but are actually not required by current linting rules.  I have deleted those.  I also removed the linting requirement for documenting `namespace`s.  I think we will eventually want to mark previous `namespace`s as `@deprecated`, but we should not really needs docs for current ones.  I will push back a bit on the `SettingsService` type documentation. What I had originally written was not very helpful documentation, but I think that _proper docs_ here could be actually useful vs just not having any documentation.  I tried to illustrate this by adding more helpful context to the docs.  Let me know what you think!

> Unless we can tweak the linter to only fail when appropriate let's leave this to the humans (coder + reviewer) to catch.

I think the big question here is if folks agree with my assessment on the value of the `/** @internal */` docs.  I don't see any way to tweak the linting rules so we could avoid these, but as I said before, I find them valuable _in their own right._ (But my tolerance threshold for "spurious" content in the codebase is def higher than many other folks....)  I am happy to just drop the `require-jsdoc` rule entirely if folks just cannot stand the extra `/** @internal */` docs.

On the other hand, if we are okay with at least some `/** @internal */` docs, there are a number of other levers we can pull to further tweak when docs are required.  Specifically we could decide which _kinds_ of exported things we want to require docs on:

- Functions - seems like this is a clear "yes" if we want docs anywhere...
- Types/Interfaces - this feels like a place where we might decide we should not _require_ docs (since in many places these docs could be redundant (e.g. the `SettingsService`).  
- Interface property signatures - (e.g. require docs for each field/method in an interface).  This I left off from the start since it feels excessive, but can turn it on if folks think it would be useful...

shared-libs/cht-datasource/src/libs/contact.ts
```diff
@@ -0,0 +1,16 @@
+import { Doc } from './doc';
+import { DataObject } from './core';
+
+interface DehydratedParent extends DataObject {
```
sugat009, line 4: question: What is the significance of `Dehydrated*`?

jkuester: Yeah, not totally happy with this name, but the "Hydrate" terminology comes from the [existing codebase](https://github.com/medic/cht-core/blob/899e30df5fc6b7c52c99893609887ae714c87cd0/shared-libs/lineage/src/hydration.js#L5010.  Basically, the parent lineage of contacts is normalized onto the contact docs in the DB so that each contact records the hierarchy of its parents/grandparents/etc.  But, instead of having each contact doc in the DB store the _whole data_ about the parent contact, we just normalize the `_id` and the `parent` fields.  However, when you want to get data about a contact and their parents, often you want to have all the data about the parent (and not just the normalized data).  The process of getting the rest of that parent data has been called "hydrating". 

Now that I have typed this all out, I am wondering if maybe "NormalizedParent" or something like that might be a better name for this interface (and a bit less specific to the historical context of the CHT...)

sugat009:
@jkuester Not to be the variable name reviewer guy 😅 but usage-wise would it be possible to have the version access name consistent? 

Meaning that for the declarative one, we do:

```typescript
import { getDatasource } from '@medic/cht-datasource';

const myPerson = await getDatasource().v1.person.getByUuid(myUuid);
```

Notice the lowercase `v1` whereas, for imperative, we do:

```typescript
import { Person, Qualifier } from '@medic/cht-datasource';

const getPerson = Person.V1.get(dataContext);
```
An uppercase `V1`.
Would it be possible to have both of them either as `v1` or `V1`?

And as for the JSdoc conversation, IMO having additional/extra comments/docs is always useful than having fewer docs, and IDEs and editors these days can always fold them out if the concern is that they take up screen space. 😀

shared-libs/cht-datasource/src/libs/contact.ts
```diff
@@ -0,0 +1,16 @@
...
+  parent?: DehydratedParent;
+}
+
+/** @internal */
+export interface Contact extends Doc {
+  contact_type?: string;
+  name?: string;
+  parent?: DehydratedParent;
+  reported_date?: Date;
+  type: string;
```
sugat009, line 15: note: A heads-up on this. `type` is a keyword in Typescript but this is allowed and in the future, if we decide to raise strictness this will raise an error at some point.

jkuester: Good point!  We need this as `type` on v1 for backwards compatibility, but on future iterations, we should avoid keywords like this.

shared-libs/cht-datasource/src/remote/libs/remote-environment.ts
jkuester: So, I ended up having to back out the `request-promise-native` code. :disappointed: It was causing some janky errors when getting transitively included in webapp. E.g.:

```
../node_modules/assert-plus/assert.js:5:13-37 - Error: Module not found: Error: Can't resolve 'stream' in '/var/home/jlkuester7/git/cht-core/node_modules/assert-plus'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "stream": require.resolve("stream-browserify") }'
        - install 'stream-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "stream": false }
```

The best solutions I could come up with were:

- Inject a HTTP "service" into cht-datasource so that api/sentinel could use rpn, but webapp could inject the Angular HTTP service.
- Pollyfill the node stuff into webapp so we don't get these errors.
- Just switch back to the `fetch` api and run the risk that something changes between 20 and 21

I have gone with the last option for now since it seemed like the least time wasted overall (open to feedback here, though). The cht-datasource code that is using `fetch` is structured to be reusable so references to the actual `fetch` function should be isolated and easy to update if necessary in the future.

admin/src/js/services/auth.js
jkuester: After thinking about this more, I am now actually thinking that we want to use the _local_ data context in Sentinel for two reasons:

- Performance - With the remote context, every action will end up being two http "hops"; one from Sentinel to api and one from api to Couch. When using Pouch in the local context, the http calls will just go directly from Sentinel to Couch. It should not be a huge performance problem, but I cannot come up with any benefit we would get out of calling api instead of just Couch directly....
- Access - the main value of the _remote context_ is that it can function without direct access to the Couch apis.  So, if we use the remote context in the admin app and a sysadmin is running the admin app in their browser and doing stuff that uses cht-datasource, all the interactions will be going through the api server and not directly against Couch.  This supports our eventual goal of locking down the Couch apis from being directly accessed.  However, this concern does not apply to Sentinel where we can continue to support direct access to Couch since we have full control over what happens.

jkuester:
> Not to be the variable name reviewer guy

:laughing:  @sugat009 the names are one of the the most important parts of a new API, so please, by all means, be the variable name reviewer guy!  :100:   I agree that more consistency here would be nice, so I have updated the namespaces to be a lowercase `v1`. :+1:

shared-libs/cht-datasource/src/remote/libs/remote-environment.ts
garethbowen: Another option is to have an optional injection for fetch so API can inject [node-fetch](https://www.npmjs.com/package/node-fetch) which should be a drop in replacement.

Let's go with `fetch` for now and rely on e2e testing to find any issues in the Node 20 implementation or when we upgrade to Node 22.

tests/integration/api/controllers/person.spec.js
```diff
@@ -0,0 +1,62 @@
...
+    await utils.createUsers([userNoPerms]);
+  });
+
+  after(async () => {
+    await utils.revertDb([], true);
+    await utils.deleteUsers([userNoPerms]);
+  });
+
+  describe('GET /api/v1/person/:uuid', async () => {
+    const getPerson = Person.v1.get(dataContext);
```
jkuester, line 42: @lorerod my initial approach here is to end up with the minimal amount of new integration tests by _combining_ tests for the cht-datasource with tests for the person API controller.  Basically, these tests use cht-datasource with the `remote` data context to call the controller via the `/api/v1/person/:uuid` REST endpoint. Then the api controller code uses cht-datasource with the `local` data context (aka Pouch) to do the data interactions with Couch.  So we end up being about to test the following "integration":

cht-datasource (remote) -> api Person controller -> cht-datasource (local) -> Couch

To me, this seems like a valid way to test everything at the same time without compromising the effectiveness/fidelity of the tests at all.  Using the remote cht-datasource to make the REST calls is functionally not much different than using the `utils.request` function to make the REST calls...  But, IDK, let me know what you think about all this!

tests/utils/index.js
```diff
@@ -63,6 +63,9 @@ const MINIMUM_BROWSER_VERSION = '90';
 const KUBECTL_CONTEXT = `-n ${PROJECT_NAME} --context k3d-${PROJECT_NAME}`;
 const cookieJar = rpn.jar();
 
+// Cookies from the jar will be included on Node `fetch` calls
+global.fetch = require('fetch-cookie').default(global.fetch, cookieJar);
```
jkuester, line 67: After some tinkering and thinking, I realized we are in a pretty stable place for authentication in cht-datasource. 

Webapp (online) and the admin app are using the _remote adapter_ which makes `fetch` calls from the browser. These `fetch` calls can be made relative to the "current page" (so the URL does not need to be provided) and they inherit the cookies for the browser session (so we don't need to worry about authentication).

Sentinel and api are using the _local adapter_ which is just `Pouch` under the hood and the injected Pouch instances are already authenticated.

---

However, our integration tests (and any future 3rd-party usage of cht-datasource) need to provide a root URL to use for calling the api server (done via  the `getRemoteDataContext` function).  This all works great.  Where things get tricky for the tests is the NodeJS implementation of the `fetch` api does not supply cookies from the `Set-Cookie` of previous responses when it makes new requests (this happens automatically in the browser environment).  So, the choices were either update the cht-datasource code to make it possible to inject the session cookie just so the tests would work, or get a bit creative with `fetch` during the test runs to be sure the cookies are there when `fetch` is called.  I have opted for the latter option for now. It keeps the impl code the most clean.  This will probably need to be re-visited in the future, either when we go to replace all our `request-promise-native` with `fetch` or when we release cht-datasource for 3rd-party consumption and we need a more proper way to do authentication.  But, for now, this seems like a reasonable approach especially since we can just piggy-back on our existing rpn cookie jar....

jkuester:
@m5r @sugat009 This is ready for another review! Actually, IMHO this code is  "complete" in that I would be comfortable merging it to `master`.  It is the complete flow for adding an endpoint to "get person without lineage" along with the necessary unit/integration tests.  That being said, before we actually merge this, we want to also include:

- [X] Get person without lineage
- [ ] Get person with lineage
- [ ] Get place without lineage
- [ ] Get place with lineage

tests/integration/api/controllers/person.spec.js
lorerod: @jkuester , thank you for clarifying this. I agree with your approach to testing using the remote data context. This is enough for testing this change.

lorerod (approved):
From a testing perspective, this looks good. I just have a small suggestion. Thank you!

shared-libs/cht-datasource/src/libs/core.ts
```diff
@@ -0,0 +1,44 @@
...
+/** @internal */
+export const hasFields = (
+  value: Record<string, unknown>,
+  fields: NonEmptyArray<{ name: string, type: string }>
+): boolean => fields.every(field => hasField(value, field));
+
+/** @internal */
+export abstract class AbstractDataContext implements DataContext {
+  readonly get = <T>(fn: (ctx: DataContext) => T): T => fn(this);
+}
```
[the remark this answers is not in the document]
m5r: I moved it to `shared-libs/cht-datasource/src/libs/data-context.ts` and changed the imports accordingly in the remote and local `data-context.ts` files and it doesn't seem to throw any error. Maybe something you did in recent commits fixed it?

shared-libs/cht-datasource/src/libs/data-context.ts
```diff
@@ -0,0 +1,41 @@
...
+ * offline. Or it may represent a remote data context where all data operations are performed against a remote CHT
+ * instance.
+ */
+export interface DataContext {
+  /**
+   * Executes the provided function with this data context as the argument.
+   * @param fn the function to execute
+   * @returns the result of the function
+   */
+  get: <T>(fn: (ctx: DataContext) => T) => T
```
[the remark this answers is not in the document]
m5r: What do you think of `ctx.bind(fn)` or `ctx.wrap(fn)`? This `ctx.get` function reminds me a lot of [`Function.prototype.bind`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind)

shared-libs/cht-datasource/src/libs/core.ts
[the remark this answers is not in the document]
jkuester: Hmm, I will have to try it again and see if I can recreate the error.  However, I did forget to mention that the error only happened _at runtime_ when running against the transpiled JS files.  So, `tsc` would run fine and my unit tests would work, but the api server would fail to start with an error about null constructors or something...

shared-libs/cht-datasource/src/libs/data-context.ts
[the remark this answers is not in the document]
jkuester: :heart_eyes: I really like `bind`!  That is perfect!

shared-libs/cht-datasource/src/libs/core.ts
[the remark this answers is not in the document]
m5r: Okay I eventually managed to reproduce this issue! Yes there *is* a circular dependency in `shared-libs/cht-datasource/src/libs/data-context.ts`, two actually:
- `import { isLocalDataContext } from '../local/libs/data-context';`
- `import { assertRemoteDataContext, isRemoteDataContext } from '../remote/libs/data-context';`

Those two files then import `DataContext` (and eventually `AbstractDataContext`) from `'../../libs/data-context'` which creates the circular dependency leaving all three exports from `src/libs/data-context.ts` undefined: `exports.adapt`, `exports.assertDataContext`, and `exports.AbstractDataContext`.

I think what you did is correct. The only other alternative is to move `isLocalDataContext`, `assertRemoteDataContext`, and `isRemoteDataContext` to `src/libs/data-context.ts`. Both way get us the same result so let's keep it the way it currently is :+1:

api/src/services/data-context.js
```diff
@@ -0,0 +1,7 @@
+const { getLocalDataContext } = require('@medic/cht-datasource');
+const db = require('../db');
+const config = require('../config');
+
+const dataContext = getLocalDataContext(config, db);
+
+Object.assign(module.exports, dataContext);
```
m5r, line 7: I like the brevity of this pattern but I would prefer if it was [greppable](https://mrcoles.com/is-it-greppable/). It's a bit more tedious to re-export each property but explicit > implicit here

shared-libs/cht-datasource/README.md
```diff
@@ -0,0 +1,24 @@
...
+1) Implement the interaction in the [`local`](./src/local) and [`remote`](./src/remote) adapters.
+2) Expose a unified interface for the interaction from the relevant top-level [concept module](./src).
+3) Expose the new concept interaction by adding it to the datasource returned from the [index.ts](./src/index.ts).
+4) Implement the necessary endpoint(s) in [api](../../api) to support the new interaction (these are the endpoints called by the remote adapter code).
+
+### Updating functionality
+
+Only passive changes should be made to the versioned public API's exposed by cht-datasource. Besides their usage in cht-core, these API's are available to custom configuration code for things like purging, tasks, targets, etc. If a non-passive change is needed, it should be made on a new version of the API.
+
+The previous version of the functionality should be marked as `@deprecated` and, where possible, all usages in the cht-core code should be updated to use the new API.
```
m5r, line 24: Thanks for documenting this process, it's gonna make things much easier when we'll come back to it months later! Could you touch a word or two about the relevant project's npm scripts like `build` and `build-watch` and their use cases?
Let's say I'm adding a new API route that uses the places API from `cht-datasource`, how can I run the API and have my changes in datasource "hot-reloaded" without having to manually cd into datasource, compile it, restart the API?

m5r (approved):
I did a thorough pass of the PR and it looks great! I know you're expecting to add a few more APIs to the Person and Place APIs but the groundwork you laid here is fantastic and ready to merge IMO :clap: 

I wouldn't mind having the additional APIs in another PR, it would make the changes easier to review

webapp/src/ts/services/cht-script-api.service.ts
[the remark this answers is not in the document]
jkuester: :+1: I had originally hesitated since it would mean updating the references in a bunch of other files, but considering this again, I guess there will never be a better time to make this change (since this service is only going to get used in more places in the future.  So, I have gone ahead with the renaming.

api/src/services/data-context.js
jkuester: This is good advice. We can be explicit here with what we are assigning and so we should be.

jkuester: Actually, looking at this again, I guess we can simplify this code here by just directly exporting the dataContext object, itself, and not trying to re-bind anything to any new object....

jkuester:
@m5r this should be ready for a final review!  :pray: 

> I wouldn't mind having the additional APIs in another PR, it would make the changes easier to review

Good call!  Once this PR is approved, I will merge and then we can open the subsequent PRs against `master`.  :+1:

shared-libs/cht-datasource/README.md
m5r, line 1: Great readme :raised_hands: very helpful

api/src/services/data-context.js
```diff
@@ -0,0 +1,5 @@
+const { getLocalDataContext } = require('@medic/cht-datasource');
+const db = require('../db');
+const config = require('../config');
+
+module.exports = getLocalDataContext(config, db);
```
m5r, line 5: That works too! :smile:

m5r (approved):
:ship: it! Well done Josh

Issue #6543, opened 2020-07-16
Allow for multiple places to be assigned to users


garethbowen:
### Overview

Currently replication permissions are based on a user's facility_id. This is a single UUID that references a place in the hierarchy. The user is permitted to access all people and places underneath that place. The user also has access to all reports about the people and places they can access.

There are a few caveats to this, including
- depth: restricting how many levels below your facility you can access
- report depth: restricting how many levels you will access reports for
- privacy: whether the report contains restricted information
- sensitivity: whether the report is about the user

The hierarchy is also used to represent the geographical areas being served and is useful for reporting on a place's catchment.
The facility_id also dictates the default selected place when loading the Contacts tab, and registering new places starts here.

### Problems

1. CHPs don't also manage all families in one geographical area. To work around this child areas are created (eg: Mary's area) to contain all the families a CHP is working with, but these are arbitrary and not useful for reporting purposes. To move a family from one area to another then requires updating their hierarchy which is difficult.
2. CHPs don't necessarily stick to their assigned geographical area. To work around this CHPs are given multiple logins, and multiple apps, so they can switch between areas. This performs really poorly.

### Proposal

Instead of a user having a single facility_id allow configurations with multiple IDs. Keep the facility_id to reference the home place and add a new called something like "assigned_contacts". For backwards compatibility if you have no assigned contacts it defaults to the facility_id. The assigned_contacts behave the same as the facility_id but instead of querying the contacts_by_depth view with just one ID it'll use an array.

Each assigned contact will be able to be configured with a depth and report_depth component so they can be assigned contacts at different levels of the hierarchy. This means we can configure users who are assigned all places under their home place (as they are now) which means they will automatically be granted access to new children. Alternatively you can manually assign places to users where there is shared responsibility. One example is supervisors for an area will likely have access to view all CHPs in their area, but families may be assigned individually to CHPs and be reassigned as needed.

As well as replicating the assigned places, we will also allow read only access to all the parents, grandparents, etc of the assigned places so you can view but not edit the hierarchy. This means they can create valid children of places they can access.
When a user creates a new place we need to make sure it is assigned to them even if they're not assigned to the place's parent. This will require code to update the user's assigned places when a new place is created. This could be done in a sentinel transition but sentinel processes changes in order and can get behind which would be bad for assigned places. Instead, make a change to the replication code that updates the assignees when new places are replicated up. This will require a sentinel transition to update the user doc on place creation.

[...]

PR #9094, opened 2024-05-03
feat(#6543): only show assigned facilities without children for users with multiple facilities

Benmuiruri:
# Description

 (WIP) 
- I have traced the execution flow  and experimented quite a bit (_alot - I have tried different implementations_)but I still have not found an answer.
- I need to dig some more, changing `facility_id` from a `string` to an `array` requires changing several parts of the codebase. I am still tracing which parts need to change

#6543 

# Code review checklist

# License

### DIFF
--- a/webapp/src/ts/modules/contacts/contacts.component.ts
+++ b/webapp/src/ts/modules/contacts/contacts.component.ts
@@ -130,6 +130,7 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
       ])
       .then(([homePlaceSummary, viewLastVisitedDate, settings, contactTypes]) => {
         this.usersHomePlace = homePlaceSummary;
+        console.log('homeplace', this.usersHomePlace);
         this.lastVisitedDateExtras = viewLastVisitedDate;
         this.visitCountSettings = this.UHCSettings.getVisitCountSettings(settings);
         if (this.lastVisitedDateExtras && this.UHCSettings.getContactsDefaultSort(settings)) {
@@ -192,6 +193,7 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
       selectedContact,
     ]) => {
       this.contactsList = contactsList;
+      console.log('contactList', this.contactsList);
       this.filters = filters;
       this.listContains = listContains;
       this.selectedContact = selectedContact;
@@ -217,19 +219,24 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
     return this.userSettingsService
       .get()
       .then((userSettings:any) => {
-        const facilityId: string = homePlaceId ?? userSettings.facility_id;
+        // eslint-disable-next-line max-len
+        const facilityId = Array.isArray(userSettings.facility_id) ? userSettings.facility_id : [userSettings.facility_id];
+        console.log('facility', facilityId);
         if (!facilityId) {
           return;
         }
 
         this.globalActions.setUserFacilityId(facilityId);
         return this.getDataRecordsService
-          .get([ facilityId ])
-          .then(places => places?.length ? places[0] : undefined);
+          .get(facilityId)
+          .then(places => {
+            console.log('places', places); // Log the fetched places.
+            return places;
+          });
       })
       .then((summary) => {
         if (summary) {
-          summary.home = true;
+          summary.forEach(place => place.home = true);
         }
         return summary;
       });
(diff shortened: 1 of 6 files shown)


webapp/src/ts/modules/contacts/contacts.component.ts
```diff
@@ -213,31 +213,36 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
...
   }
 
-  private getUserHomePlaceSummary(homePlaceId?: string) {
+  private getUserHomePlaceSummary() {
     return this.userSettingsService
       .get()
-      .then((userSettings:any) => {
-        const facilityId: string = homePlaceId ?? userSettings.facility_id;
+      .then((userSettings: any) => {
+        const facilityId = Array.isArray(userSettings.facility_id)
```
Benmuiruri, line 220: HI @latin-panda give me your quick thought on this. This is what I meant by treating all `facility_id` as arrays. If it is a string I typecast it here to treat all as arrays in the component. 

Any concerns about that ?

latin-panda: It's okay.

Does the ternary fit in 1 line? It's better seeing like that

latin-panda: Maybe you can make the code flat with async?

webapp/src/ts/modules/contacts/contacts.component.ts
```diff
@@ -213,31 +213,36 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
...
+          ? userSettings.facility_id
+          : [userSettings.facility_id];
+
         if (!facilityId) {
           return;
         }
 
         this.globalActions.setUserFacilityId(facilityId);
```
latin-panda, line 228: Because now this is saving an array. You'll need to update the contact.effects and the contact-component
search by `getUserFacilityId`, you'll find the selectors

webapp/src/ts/modules/contacts/contacts.component.ts
Benmuiruri: It doesn't fit in 1 line. I will try the async

Benmuiruri:
Hi @Diana one change I did in the webapp is to add a method to typecast string facility_id to array

```
private getUserFacilityId(userSettings) {
    return Array.isArray(userSettings.facility_id) ? userSettings.facility_id : [userSettings.facility_id];
  }

```
So that everywhere in the code we work with one data type. I'm curious whether that is something we should change at the API level for consistency, or its fine just making the change in webapp  ?

webapp/src/ts/modules/contacts/contacts-content.component.ts
```diff
@@ -104,14 +104,14 @@ export class ContactsContentComponent implements OnInit, OnDestroy {
...
-        const shouldDisplayHomePlace = userFacilityId &&
+      .subscribe((userFacilityIds) => {
+        const shouldDisplayHomePlace = userFacilityIds &&
           !this.filters?.search &&
           !this.route.snapshot.params.id &&
           !this.responsiveService.isMobile();
 
         if (shouldDisplayHomePlace) {
-          this.contactsActions.selectContact(userFacilityId);
+          this.contactsActions.selectContact(userFacilityIds[0]);
```
dianabarsan, line 114: Just philosophically: do we think this is the correct behavior? For all intents and purposes, this is a randomly picked facility out of the users' facilitires list.

dianabarsan (commented):
> So that everywhere in the code we work with one data type. I'm curious whether that is something we should change at the API level for consistency, or its fine just making the change in webapp ?

When I push my latest changes, the API will only create users with array facility_id fields, and I'm probably going to add a migration too, but I think it's safe to have the typecast in webapp just in case. 

I left one philosophical question inline. Nice work!

webapp/src/ts/modules/contacts/contacts-content.component.ts
Benmuiruri: For the most part it is the right behavior. 
1. If the user has one `facility_id` then getting `userFacilityIds[0]` of `['id-1']` as their homeplace is fine
2. If user has several `facility_ids`, we will randomly pick one of the facilities from `['id-1', 'id-2']`. But that should be fine because both of them are `homeplaces` so we can pick either. We also have to pick one to display the details of the `homeplace` on the RHS (Contact Detail view )

ralfudx:
Code looks good to me - trying to test this locally

ralfudx (approved):
Tested this locally alongside @Benmuiruri and it looks good functionally 👍

PR #9099, opened 2024-05-07
feat(#6543): disable aggregate targets for users with multiple facility_ids

latin-panda:
# Description

- Users with many associated facilities cannot see target aggregates 

- To assign more than one facility to a user:
  - Go to couchdb, find the `_users` database, find the offline user `org.couchdb:<username>`
  - Make `facility_id` an array and add the IDs: `"facility_id: [ "id-1", "id-2" ]`
      - the facility should be at the same level in the hierarchy. 
  - Go find the `medic` database, find the offline user `org.couchdb:<username>`
  -  Make `facility_id` an array and add the IDs: `"facility_id: [ "id-1", "id-2" ]`. It should match with the one in `_users` db. 

Videos:

- The `can_aggregate_targets` is disabled. Users with 1 and with many associated facilities cannot see the aggregate targets page. 

Uploading permission-disabled.mov…

- The `can_aggregate_targets` is enabled. Users with 1 associated facility can see the page, but those with many associated facilities cannot see the aggregate targets page. 

Uploading permission-enabled.mov…

#6543

# Code review checklist

# License

### DIFF
--- a/webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.ts
+++ b/webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.ts
@@ -1,5 +1,5 @@
 import { AfterContentChecked, AfterContentInit, Component, Input, OnDestroy } from '@angular/core';
-import { ActivatedRoute, Router } from '@angular/router';
+import { ActivatedRoute } from '@angular/router';
 import { Subscription } from 'rxjs';
 
 @Component({
@@ -12,8 +12,7 @@ export class AnalyticsFilterComponent implements AfterContentInit, AfterContentC
   subscriptions: Subscription = new Subscription();
 
   constructor(
-    private router: Router,
-    private route: ActivatedRoute
+    private route: ActivatedRoute,
   ) { }
 
   ngAfterContentInit() {
--- a/webapp/src/ts/services/target-aggregates.service.ts
+++ b/webapp/src/ts/services/target-aggregates.service.ts
@@ -228,13 +228,20 @@ export class TargetAggregatesService {
       });
   }
 
-  private async getHomePlace() {
+  private async getUserFacilityIds(): Promise<string[] | undefined> {
     const { facility_id }:any = await this.userSettingsService.get();
-    if (!facility_id) {
+    if (!facility_id?.length) {
       return;
     }
+    return Array.isArray(facility_id) ? facility_id : [ facility_id ];
+  }
 
-    const places = await this.getDataRecordsService.get([ facility_id ]);
+  private async getHomePlace() {
+    const facilityIds = await this.getUserFacilityIds();
+    if (!facilityIds?.length) {
+      return;
+    }
+    const places = await this.getDataRecordsService.get(facilityIds);
     return places?.length ? places[0] : undefined;
   }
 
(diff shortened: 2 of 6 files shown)


webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.ts
```diff
@@ -12,8 +12,7 @@ export class AnalyticsFilterComponent implements AfterContentInit, AfterContentC
   subscriptions: Subscription = new Subscription();
 
   constructor(
-    private router: Router,
```
latin-panda, line 15: Unused dependency

webapp/src/ts/services/target-aggregates.service.ts
```diff
@@ -269,8 +276,14 @@ export class TargetAggregatesService {
...
-  isEnabled() {
-    return this.authService.has('can_aggregate_targets');
+  async isEnabled() {
+    const canSeeAggregates = await this.authService.has('can_aggregate_targets');
+    if (!canSeeAggregates) {
+      return false;
+    }
+
+    const facilityIds = await this.getUserFacilityIds();
+    return !!facilityIds && facilityIds.length <= 1;
```
latin-panda, line 286: I tried `facilityIds?.length <= 1`, but the static checks didn't like it.

latin-panda:
@Benmuiruri @lorerod please have a look at this small PR. I attached a couple of videos and added coverage with unit tests.

Thanks!

Benmuiruri (approved):
Hi @latin-panda 

I expected the user who should have multiple`facility_ids` is the CHA not the CHP. In your video I see you enabling and disabling `can_aggregate_targets` for the CHP. I wonder, are the users in your test CHA or a CHP  ? 

However, in my test I have tested with CHA and it works as expected. The one with 1 `facility_id` is the only one who can see `targets_aggregate`

latin-panda:
> are the users in your test CHA or a CHP ?

@Benmuiruri the users in my video are CHP

webapp/tests/karma/ts/services/target-aggregates.service.spec.ts
```diff
@@ -84,22 +84,48 @@ describe('TargetAggregatesService', () => {
   });
 
   describe('isEnabled', () => {
-    it('should return false when user does not have permission', async () => {
-      authService.has.resolves(false);
+    it('should return true when user has permission and user has one facility assigned as array', async () => {
+      authService.has.resolves(true);
+      userSettingsService.get.resolves({ facility_id: [ 'facility-1' ] });
```
lorerod, line 89: I believe that an array with multiple elements would be better. Alternatively, if there is a specific reason for testing an array with only one element, please create a new test case using an array with multiple elements.

lorerod (changes requested):
What do you think about adding an end-to-end test in `/tests/e2e/default/targets/target-aggregates.wdio-spec.js` for users with multiple facility IDs?

webapp/tests/karma/ts/services/target-aggregates.service.spec.ts
latin-panda: > if there is a specific reason for testing an array with only one element

Must users (CHWs) will have only 1 element (facility) in this array, we are just including this case to play safe. 

> please create a new test case using an array with multiple elements.

There's one in L121 `it('should return false when user has more than one facility assigned',`
It should not return true when multiple elements are present; that's why we have this test case :)

latin-panda:
@lorerod thanks for reviewing! I have added the e2e test

lorerod (approved):
Thank you @latin-panda !

latin-panda:
@Benmuiruri, this has been rebased, and the tests are working. It's ready. Just waiting for [6543-multi-facility](https://github.com/medic/cht-core/tree/6543-multi-facility).

PR #9128, opened 2024-05-16
feat(#9116): update user place field in admin to allow setting multiple places

### LINKED ISSUE #9116: Make place field in the user form a multiselect dropdown
This is a continuation of https://github.com/medic/cht-core/issues/6543

Now, users can be associated with more than one facility. The CHT's Admin app should provide a way to associate those facilities with the user. 

- The admin user enables `can_have_multiple_places` to some specific roles (e.g Supervisor CHW, etc.)

Whenever a new user is added:
- The admin user clicks on  `add user`
- The dialog displays, and the place field is a multiselect dropdown. 
- The admin user fills in the information and selects roles and multiple places. 
- The admin submits the form and CHT validates:
   - At least one of the selected roles has `can_have_multiple_places`
   - Selected places are at the same hierarchy level
   - There's currently validation that says the associated contact must be a child of the selected place. In multiselect case, it should be a child of at least one of the selected places. Validate with the team. 
- CHT displays error in form when validations doesn't pass successfully.  
- Make sure to edit existing user prefills the multiselect field, and saves changes correctly 

[MK EDIT 2024-05-14] We should also update the "Place" text. 

- When the validation happens, the user is not created yet, then probably use the settings service to validate `can_have_multiple_places` permission. 
- The place dropdown shows the hierarchy, check if we have access to that data easily to validate all places are from the same hierarchy level
Screenshot 2024-05-14 at 1 43 59 PM

Benmuiruri:
# Description

This is a continuation of https://github.com/medic/cht-core/issues/6543

Now, users can be associated with more than one facility. This PR enables the CHT's Admin app to provide a way to associate those facilities with the user.

To get the feature to work, the admin user must enable `can_have_multiple_places` to some specific roles (e.g Supervisor, community_health_assistant, etc)

Whenever a new user is added:
- The dialog displays, and the place field is a multiselect dropdown. 
- The admin submits the form and CHT validates:
   - At least one of the selected roles has `can_have_multiple_places`
   - Selected places are at the same hierarchy level
   - The user should be a child of at least one of the selected places.
- CHT displays error in form when above validations don't pass successfully.  
- When editing an existing user, the form should load with the multiselect field prefilled with the user's current place or places. On editing a user the form saves changes correctly 

Closes #9116 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9116-update-admin-place-field/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9116-update-admin-place-field/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9116-update-admin-place-field/docker-compose/cht-couchdb-clustered.yml)

# License


latin-panda:
@Benmuiruri, this is just a reminder to please enable the test `tests/e2e/default/more-options-menu/offline-user/all-permissions.wdio-spec.js` so that we can have all the tests working again.

ralfudx (approved):
All AC's in the description look good to me

admin/src/js/controllers/edit-user.js
```diff
@@ -232,6 +236,27 @@ angular
       return true;
     };
 
+    const validatePlacesPermission = () => {
+      const userRoles = $scope.editUserModel.roles;
+
+      if (!$scope.editUserModel.place || $scope.editUserModel.place.length <= 1) {
```
dianabarsan, line 242: is `$scope.editUserModel.place` expected to always be an array? Are there situations when it can also be a string?

admin/src/js/controllers/edit-user.js
```diff
@@ -232,6 +236,27 @@ angular
       return true;
     };
 
+    const validatePlacesPermission = () => {
+      const userRoles = $scope.editUserModel.roles;
```
dianabarsan, line 240: If we return early in the next if, this can be defined after the `if`.

admin/src/js/controllers/edit-user.js
```diff
@@ -232,6 +236,27 @@ angular
...
+
+      if (!$scope.editUserModel.place || $scope.editUserModel.place.length <= 1) {
+        return true;
+      }
+
+      const allowedRoles = $scope.permissions.can_have_multiple_places;
+
+      const userHasPermission = userRoles.some(role => allowedRoles.includes(role));
```
dianabarsan, line 248: I think we're duplicating logic that we already have in the cht script api. This should already be available in the admin app. https://github.com/medic/cht-core/blob/master/admin/src/js/services/auth.js#L65

admin/src/js/controllers/edit-user.js
```diff
@@ -232,6 +236,27 @@ angular
...
+
+      if (!userHasPermission) {
+        $translate('permission.description.can_have_multiple_places.not_allowed').then(value => {
+          $scope.errors.multiFacility = value;
+        });
+        return false;
```
dianabarsan, line 254: I think this return can be removed. We would already return the falsy value of `userHasPermission` on the next line.

admin/src/js/controllers/edit-user.js
```diff
@@ -254,24 +279,63 @@ angular
       return hasPlace && hasContact;
     };
 
+    const validateFacilityHierarchy = () => {
+      const placeIds = $scope.editUserModel.place;
+
+      if (!placeIds || placeIds.length === 1) {
```
dianabarsan, line 285: Is this _always_ an array?

admin/src/js/controllers/edit-user.js
```diff
@@ -254,24 +279,63 @@ angular
...
+        return $q.resolve(true);
+      }
+
+      const getPlace = (placeId) => {
+        return DB().get(placeId);
+      };
+
+      return Promise.all(placeIds.map(getPlace))
```
dianabarsan, line 293: I think we should use an _all_docs call here, instead of individually getting every place with a single request.

admin/src/js/controllers/edit-user.js
```diff
@@ -254,24 +279,63 @@ angular
...
+        .then((places) => {
+          const contactTypes = places.map(place => place.contact_type);
+          const isSameHeirarchy = contactTypes.every(type => type === contactTypes[0]);
```
dianabarsan, line 296: Um. This is pretty hard to understand. 
I think it would be easier to do something like this: 
```
const contactTypes = new Set(places.map(place => place.contact_type));
const isSameHeirarchy = contactTypes.size === 1;
```

admin/src/js/controllers/edit-user.js
```diff
@@ -254,24 +279,63 @@ angular
...
-              valid = true;
-              break;
-            }
-            parent = parent.parent;
-          }
+
+      const getParent = (contactId) => {
+        return DB()
+          .get(contactId)
+          .then(function (contact) {
```
dianabarsan, line 321: Please use an arrow function here. 
```
return DB().get(contactId).then(contact => contact.parent);
```

admin/src/js/controllers/edit-user.js
```diff
@@ -254,24 +279,63 @@ angular
...
+        }
+        if (placeIds.includes(parent._id)) {
+          return true;
+        }
+        return checkParent(parent.parent, placeIds);
+      };
+
+      return getParent(contactId)
+        .then(function (parent) {
+          const valid = placeIds.some((placeId) => checkParent(parent, [placeId]));
```
dianabarsan, line 338: I think you _should_ be able to just call this function by passing all placeIds, right?
```suggestion
const valid = checkParent(parent, placeIds);
```

admin/src/js/controllers/edit-user.js
```diff
@@ -371,9 +435,8 @@ angular
     };
 
     const computeFields = () => {
-      $scope.editUserModel.place = $(
-        '#edit-user-profile [name=facilitySelect]'
-      ).val();
+      const placeValue = $('#edit-user-profile [name=facilitySelect]').val();
+      $scope.editUserModel.place = Array.isArray(placeValue) && placeValue.length === 0 ? null : placeValue;
```
dianabarsan, line 439: I'm not sure id this works as expected. 
If `placeValue` is a string, than this will return a string. And in lots of places we assume this will be an array.

admin/src/js/services/create-user.js
```diff
@@ -20,6 +20,7 @@
        * @param      {Object}  updates        Updates you wish to make
        */
       const createSingleUser = (updates) => {
+        URL = '/api/v3/users';
```
dianabarsan, line 23: Because this is a singleton, we will end up changing this value for all the following calls. 
I think we should use a scoped variable here, instead of overwriting the global one. Or have two global variables.

admin/src/js/services/select2-search.js
```diff
@@ -26,11 +27,31 @@ angular.module('inboxServices').factory('Select2Search',
...
-      return row.text;
+      if (selection.doc) {
+        return (
+          selection.doc.name +
+          (selection.doc.muted
+            ? ' (' + $translate.instant('contact.muted') + ')'
+            : '')
+        );
+      }
+      return selection.text;
```
dianabarsan, line 54: This is extremely hard to follow. Was this a change requested by sonar? 
I think we can make this more readable: 
```suggestion
const defaultTemplateSelection = (selection) => {
  const formatRow = (row) => {
    if (!row.doc) {
      return row.text;
    }
    
    let formatted = row.doc.name;
    if (row.doc.muted) {
      formatted += `(${$translate.instant('contact.muted')})`;
    }
    return formatted;
  };
  if (Array.isArray(selection)) {
    return selection.map(row => formatRow(row)).join(', ');
  }
  return formatRow(selection);
}
```

admin/src/js/services/select2-search.js
```diff
@@ -114,9 +135,17 @@ angular.module('inboxServices').factory('Select2Search',
...
-          if (!selectEl.children('option[value="' + initialValue + '"]').length) {
+          if (Array.isArray(initialValue)) {
+            initialValue.forEach(function (val) {
+              if (!selectEl.children('option[value="' + val + '"]').length) {
+                selectEl.append($('<option value="' + val + '"/>'));
+              }
+            });
+          } else if (
+            !selectEl.children('option[value="' + initialValue + '"]').length
+          ) {
```
dianabarsan, line 148:
```suggestion
const addValue = (val) => {
  if (!selectEl.children('option[value="' + val + '"]').length) {
    selectEl.append($('<option value="' + val + '"/>'));
  }
}

if (Array.isArray(initialValue)) {
  initialValue.forEach(val => addValue(val);
} else {
  addValue(initialValue);
}
```

admin/src/js/services/select2-search.js
```diff
@@ -125,15 +154,42 @@ angular.module('inboxServices').factory('Select2Search',
...
           if (Array.isArray(value)) {
-            // NB: For now we only support resolving one initial value
-            // multiple is not an existing use case for us
-            value = value[0];
+            // NB: We now support an Array of IDs for places
+            const docPromises = value.map(function (val) { //NoSONAR
+              return getDoc(val).then(function (doc) { //NoSONAR
+                return { id: val, doc: doc };
+              });
+            });
```
dianabarsan, line 167: We should be using _all_docs here, instead of getting docs one by one.

admin/src/js/services/select2-search.js
```diff
@@ -125,15 +154,42 @@ angular.module('inboxServices').factory('Select2Search',
...
+                  if (selected) {
+                    selected.doc = doc.doc;
+                  } else {
+                    select2Data.push({
+                      id: doc.id,
+                      doc: doc.doc,
+                      text: doc.doc.name,
+                    });
+                  }
+                });
```
dianabarsan, line 184: We should extract this to a separate function.

admin/src/js/services/select2-search.js
```diff
@@ -125,15 +154,42 @@ angular.module('inboxServices').factory('Select2Search',
...
+                selectEl.select2('data', select2Data);
+                $timeout(() => { //NoSONAR
+                  selectEl.trigger('change');
+                }, 1000);
```
dianabarsan, line 188: What is the reason for adding the 1s delay to this and the next timeout?

admin/src/js/services/select2-search.js
```diff
@@ -186,16 +246,30 @@ angular.module('inboxServices').factory('Select2Search',
...
+                    doc: doc,
+                    text: doc.name,
+                  });
+                }
+                selectEl.select2('data', select2Data);
+                $timeout(() => { //NoSONAR
+                  selectEl.trigger('change');
+                }, 1000);
+
+                return selectEl;
```
dianabarsan, line 265: It looks like this block is very similar to a block above (line 180). I think we should find a way to reuse this code.

api/resources/translations/messages-en.properties
```diff
@@ -1024,6 +1024,8 @@ permission.description.can_view_unallocated_data_records = Allowed to see report
 permission.description.can_view_users = Allowed to get a list of all configured users.
 permission.description.can_write_wealth_quintiles = Allowed to update contacts with their wealth quintiles.
 permission.description.can_upgrade = Allowed to upgrade the CHT Core Framework version via the API or admin interface.
+permission.description.can_have_multiple_places.not_allowed = The person with selected role cannot have multiple places.
+permission.description.can_have_multiple_places.incompatible_place = The selected places must be at the same hierarchy level.
```
dianabarsan, line 1028: I think this message can be misleading, because what we are actually checking is whether the selected places are of the same type, not that they are at the same hierarchy level. 
Because a certain contact_type can have multiple parent types, these two assertions are not equivalent. 
I think this message should convey what we are actually checking, that the places have the same type. 

If we want to switch our verification to check for the level, then code changes are required.

shared-libs/user-management/src/users.js
```diff
@@ -342,15 +342,15 @@ const hasParent = (facility, id) => {
  */
 const mapUser = (user, setting, facilities) => {
   const facilityIds = Array.isArray(user.facility_id) ? user.facility_id : [user.facility_id];
-  const places = facilityIds.map(facility => getDoc(facility, facilities));
+  const places = facilityIds.filter(facilityId => facilityId).map(facility => getDoc(facility, facilities));
```
dianabarsan, line 345: I believe this is already covered by the changes in code I have done. Can you please rebase off of my branch. 
You should _not_ need to make changes to any api / shared libs code.

tests/integration/api/controllers/users.spec.js
```diff
@@ -1718,7 +1716,7 @@ describe('Users API', () => {
   });
 
 
-  describe('POST/GET api/v2/users', () => {
+  describe.skip('POST/GET api/v2/users', () => {
```
dianabarsan, line 1719: This should not be skipped.

webapp/src/ts/modules/contacts/contacts.component.ts
```diff
@@ -136,8 +140,8 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
       this.contactsActions.removeContactFromList({ _id: change.id });
       this.hasContacts = !!this.contactsList.length;
     }
-    if (this.usersHomePlace && change.id === this.usersHomePlace._id) {
-      this.usersHomePlace = await this.getUserHomePlaceSummary(change.id);
+    if (this.usersHomePlace?.some(homePlace => homePlace._id === change.id)) {
```
dianabarsan, line 143: I think `find` is a better choice here than `some`, because `find` will stop at the first found match, which `some` will continue iterating even after finding one match.

webapp/src/ts/modules/contacts/contacts.component.ts
```diff
@@ -218,31 +222,42 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
...
+  private getUserHomePlaceSummary() {
     return this.userSettingsService
       .get()
-      .then((userSettings:any) => {
-        const facilityId: string = homePlaceId ?? userSettings.facility_id;
-        if (!facilityId) {
+      .then((userSettings: any) => {
+        const facilityId = this.getUserFacilityId(userSettings);
+
+        if (!facilityId.length || facilityId.some(id => id === undefined)) {
```
dianabarsan, line 235: I don't understand. We're not loading the home place summary if any of the places are undefined? 
I have trouble understanding the logic here.

webapp/src/ts/modules/contacts/contacts.component.ts
```diff
@@ -484,7 +514,7 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
     this.globalActions.setLeftActionBar({
       exportFn: () => this.exportContacts(),
       hasResults: this.hasContacts,
-      userFacilityId: this.usersHomePlace?._id,
+      userFacilityId: this.usersHomePlace?.[0]?._id,
```
dianabarsan, line 517: I'm seeing that this `this.usersHomePlace?.[0]?._id` is used in multiple places. 
While I accept that this is for the MVP, I think we should have it in a function and show that this is intentional, and reuse that function throughout.

dianabarsan (changes requested):
Cool stuff! Great work!
I left some comments mostly about readability and reusability.

admin/src/js/controllers/edit-user.js
Benmuiruri: No it is always an array. When it was a string, there were situations it was `null` for certain users. Right now I believe it will always be an array even if it is an empty array but for safety I retained `!$scope.editUserModel.place`

admin/src/js/controllers/edit-user.js
Benmuiruri: Hmm, the value of `userCtx.roles` is admin. It ends up being a check whether current user (Admin) has the permission

admin/src/js/controllers/edit-user.js
Benmuiruri: I believe so because the html uses a multi-select

admin/src/js/controllers/edit-user.js
Benmuiruri: I added this after realizing an admin user or certain users can have `null` for their `placeValue`. If `placeValue` is a string, then it is typecasted to an `Array`. I don't see the situation it would be a `string`.

admin/src/js/services/select2-search.js
Benmuiruri: Without it, there was an error from select in the browser console. it seemed like an overlapping of the digest cycles. So I added the small delay.

dianabarsan: I think the delay is really bad practice. We should find out why the delay is required.

webapp/src/ts/modules/contacts/contacts.component.ts
Benmuiruri: This was change was I related to the base branch, not this PR. 

When working with strings the check was `if (!facilityId)`, when we changed to Arrays, I changed it to `if(!facilityId.length)` but I realized there are situations where we have an array with `[undefined]` and that check fails. That's why I added that if the element in the Array is `undefined` then disregard it.

Benmuiruri:
Hi @dianabarsan I have made almost all of the requested changes. The one I could not figure out is the suggestion to use `_all_docs`. Do you mind sharing a suggestion for that one. Thank you 🙏

api/resources/translations/messages-en.properties
Benmuiruri: Could we have it as `The selected places must be of the same type (be at the same hierarchy level).` ? That conveys technically what we are checking, but also explaining to the user in a way they understand.

admin/src/js/controllers/edit-user.js
dianabarsan: Yea, I think you have to call it like: 
```
return chtScriptApi.v1.hasPermissions(['can_have_multiple_places'], $scope.editUserModel.role, settings.permissions);
```

admin/src/js/controllers/edit-user.js
```diff
@@ -254,24 +277,57 @@ angular
...
+      const docsFetched = 10;
+
+      if (!placeIds || placeIds.length === 1) {
+        return $q.resolve(true);
+      }
+
+      return DB().allDocs({ keys: placeIds, include_docs: true, limit: docsFetched })
+        .then(result => {
+          const places = result.rows.map(row => row.doc);
+          const contactTypes = new Set(places.map(place => place.contact_type));
```
dianabarsan, line 291: I believe that checking the `contact_type` is not enough here. 
A contact can have a `type` and a `contact_type` as well. 

Have a look at the contact type utils library: https://github.com/medic/cht-core/blob/master/shared-libs/contact-types-utils/src/index.js#L13

I think this function, that checks whether a list of contacts are of the same type can belong there, instead of here.

admin/src/js/controllers/edit-user.js
```diff
@@ -254,24 +277,57 @@ angular
       return hasPlace && hasContact;
     };
 
+    const validateFacilityHierarchy = () => {
+      const placeIds = $scope.editUserModel.place;
+      const docsFetched = 10;
```
dianabarsan, line 282: It's unclear why we are fetching less docs than we are actually selecting?

admin/src/js/services/select2-search.js
```diff
@@ -114,26 +127,72 @@ angular.module('inboxServices').factory('Select2Search',
...
 
-      const resolveInitialValue = function(selectEl, initialValue) {
+      const addValue = (val) => {
+        if (!selectEl.children('option[value="' + val + '"]').length) {
+          selectEl.append($('<option value="' + val + '"/>'));
+        }
+      };
+
+      const getDocs = function (ids) {
+        const docsFetched = 10;
```
dianabarsan, line 137: It's unclear why we are limiting to 10 documents, when we can clearly have a longer selection.

admin/src/js/services/select2-search.js
```diff
@@ -114,26 +127,72 @@ angular.module('inboxServices').factory('Select2Search',
...
+
+        return DB()
+          .allDocs({ keys: ids, include_docs: true, limit: docsFetched })
+          .then((docs) => processDocs(docs));
+      };
+
+      const processDocs = function(result) {
+        const docs = result.rows.map(row => row.doc);
+        const docIds = docs.map(doc => doc._id);
```
dianabarsan, line 146: I think you can do this in a single line: 
```suggestion
const docIds = result.rows.map(row => row.doc._id);
```

admin/src/js/services/select2-search.js
```diff
@@ -147,12 +206,43 @@ angular.module('inboxServices').factory('Select2Search',
           }
         }
 
-        return resolution.then(function() {
-          selectEl.trigger('change');
+        return resolution.then(function() { //NoSONAR
+          $timeout(() => { //NoSONAR
+            selectEl.trigger('change');
+          }, 1000);
```
dianabarsan, line 212: I think we should figure out why this timeout is needed. I'm not comfortable with keeping this at all.

admin/src/js/services/select2-search.js
```diff
@@ -147,12 +206,43 @@ angular.module('inboxServices').factory('Select2Search',
...
+          select2Data.push({
+            id: docId,
+            doc: doc,
+            text: doc.name,
+          });
+        }
+        selectEl.select2('data', select2Data);
+        $timeout(() => { //NoSONAR
+          selectEl.trigger('change');
+        }, 1000);
```
dianabarsan, line 232: Same. We should not have delays like this, we need to figure out what is happening here.

webapp/src/ts/modules/contacts/contacts.component.ts
```diff
@@ -257,6 +256,7 @@ export class ContactsComponent implements OnInit, AfterViewInit, OnDestroy {
 
   private canViewLastVisitedDate() {
     if (this.sessionService.isAdmin()) {
+      // disable UHC for DB admins
```
dianabarsan, line 259: Is this added accidentally?

dianabarsan (changes requested):
Nice work! Some minor things inline to unblock you. I'll give this a try tomorrow!

admin/src/js/controllers/edit-user.js
Benmuiruri: Right, I forgot to take it out after some experimentation.

admin/src/js/services/select2-search.js
Benmuiruri: This seems what's mainly left. I will dedicate tomorrow morning to it.

admin/src/js/services/select2-search.js
Benmuiruri: Is the problem using `timeout` with a 1 sec delay or is it using `timeout` all together ? 

From my understanding of [ some research](https://stackoverflow.com/questions/12729122/angularjs-prevent-error-digest-already-in-progress-when-calling-scope-apply) and the [docs](https://docs.angularjs.org/error/$rootScope/inprog) `timeout`is one way to deal with the `Error: $rootScope:inprog
Action Already In Progress`. 

```
$timeout(() => {
          selectEl.trigger('change');
        });
```

With the event being triggered outside AngularJS's digest cycle by JQuery, I don't see a simpler workaround. But I am open to suggestions.

admin/src/js/controllers/edit-user.js
Benmuiruri: I have moved the work of checking the `contact_types` to the contact type utils library. But I don't understand what you mean by 

> checking the contact_type is not enough here.

the check confirms that all the places have the same contact_type (community_health_unit), why is that not enough for ensuring they are at the same hierarchy level ?

admin/src/js/services/select2-search.js
dianabarsan: Using $timeout is ok. Using $timeout with a delay is not ok.

admin/src/js/controllers/edit-user.js
dianabarsan: If you look at the code from the library, you'll see that a contact can have a `type` with a hardcoded contact type and a `contact_type` with a configurable contact type. 

Additionally, one type of contact can have multiple types of parents. so you can have a certain place at different levels of the hierarchy, but I don't think we have to worry about this ... yet.

Benmuiruri: Sorry, I don't understand the issue (difference) with `type` and `contact_type`. I don't see where `HARDCODED_TYPES` comes in the equation. 

Since all places have the same type `contact` and the hierarchy level is defined by `contact_type`, why is the check to ensure they all have the same `contact_type`  insufficient ?

dianabarsan: Not all places have type `contact`.... 
Please take the time to read the code of https://github.com/medic/cht-core/blob/master/shared-libs/contact-types-utils/src/index.js#L13 to understand or read the docs: https://docs.communityhealthtoolkit.org/apps/reference/app-settings/hierarchy/#changing-the-configuration
Or look at our sample data. 

It perfectly valid to have a place that has: { type: 'clinic' } and no `contact_type`.

Benmuiruri: Thanks for the patience with my daftness 😅 . I read the docs and added some logs, now I get what you mean. Now it's ready for review.

admin/src/js/services/select2-search.js
```diff
@@ -26,18 +28,29 @@ angular.module('inboxServices').factory('Select2Search',
       return $(format.sender(row.doc, $translate));
     };
 
-    const defaultTemplateSelection = function(row) {
-      if (row.doc) {
-        return row.doc.name + (row.doc.muted ? ' (' + $translate.instant('contact.muted') + ')': '');
+    const defaultTemplateSelection = (selection) => {
```
dianabarsan, line 31: I'm noticing that when one of the facilities is deleted (because I deleted it as my user in the frontend - another but that needs to be addressed) - this is how the edit user UI looked: 
image

The list of facilities had no names and there was an error in the console.

dianabarsan (changes requested):

shared-libs/contact-types-utils/src/index.js
```diff
@@ -67,6 +67,11 @@ const isPlace = (config, contact) => {
   return isPlaceType(type);
 };
 
+const isSameContactType = (contacts) => {
```
dianabarsan, line 70: Can we have a unit test suite for this new function?

admin/src/js/services/select2-search.js
dianabarsan: I'm still seeing this behavior :(

dianabarsan: I think you need to check the results from your all docs request to see if any of the docs has been deleted.

dianabarsan (changes requested):

Benmuiruri:
Hi @dianabarsan, when you say you are still getting same behavior, do you mean on the [care teams instance](https://care-teams-kenya.dev.medicmobile.org/) ? 

- I have not updated the instance, I was testing it locally.

Benmuiruri:
@dianabarsan 
-  Unit test done ✅ 
- The other bug, I cannot reproduce it. Here is a video showing the fix working as expected 

https://github.com/medic/cht-core/assets/2597305/54cfb5b6-30c3-4246-9c5e-51e291d92922

- The other thing is the code you highlighted is for the select dropdown in the admin, how is that related to the delete issue. (Or maybe we are not on the same page about which behavior you are referring to.)

dianabarsan:
This is not related to the delete issue. But it is triggered when one of the user's facilities has been deleted.

dianabarsan:
The issue is strictly in the admin app.
image

dianabarsan:
Hi @Benmuiruri 

I think you need to merge current master into your branch to get the haproxy fix.

shared-libs/contact-types-utils/test/index.js
```diff
@@ -155,6 +155,21 @@ describe('ContactType Utils', () => {
...
+      chai.expect(utils.isSameContactType([
+        { type: 'contact', contact_type: 'health_center' },
+        { type: 'contact', contact_type: 'health_center' },
+      ])).to.equal(true);
+    });
+    it('should return false for contacts of different type', () => {
+      chai.expect(utils.isSameContactType([
+        { type: 'contact', contact_type: 'health_center' },
+        { type: 'contact', contact_type: 'district_hospital' },
+      ])).to.equal(false);
```
dianabarsan, line 169: Can we have a test where we use a mix of hardcoded and configurable hierarchies? so so using both only `type` and both `type` and `contact_type` together.

dianabarsan (approved):
Great stuff! I left one minor comment about unit test coverage.

Issue #9122, opened 2024-05-14
Protect against redirection attack


garethbowen:
Details in private issue: https://github.com/medic/medic-projects/issues/8184

PR #9131, opened 2024-05-19
fix(#9122): detect unsafe redirection on login

garethbowen:
# Description

#9122

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9122-detect-double-slash-redirection/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9122-detect-double-slash-redirection/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9122-detect-double-slash-redirection/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/api/src/controllers/login.js
+++ b/api/src/controllers/login.js
@@ -79,7 +79,7 @@ const resolveUrl = (requested) => {
     // invalid url
     return;
   }
-}
+};
 
 const sanitizeRequestedRedirect = (requested) => {
   const resolved = resolveUrl(requested);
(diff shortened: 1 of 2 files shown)


garethbowen:
@sugat009 Would you mind reviewing this? Check out the issue for details on what it's trying to protect against.

It's more complicated than I'd like. The "proper" solution is not to take a URL as parameter but rather take enough context to go to redirect to the right place. Alternatively instead of having the redirect in the http response we could just load the app and pass the path to angular to sort out. But it's fiddly to do this and may break backwards compatibility so this is the safer option for now.

sugat009 (approved):

Issue #8557, opened 2023-09-19
Investigate using shellcheck for script linting


garethbowen:
**Describe the issue**
Shell scripts are hard to validate, so if we can add linting or similar build time checks we can detect bugs before merging.

**Describe the improvement you'd like**
Try out https://www.shellcheck.net/

**Describe alternatives you've considered**
There may be other linting tools.

PR #9142, opened 2024-05-23
chore(#8557): add linting for shell scripts

nydr:
# Description
Adds linting to shell script and update scripts to pass, there's some cases where this could change the functionality so pay extra attention to the sections of the script relying on unexpected behavior 

per #8557

Affected bash scripts - please check off when you have confirmed they work with the additions in this PR:

 - [x] .husky/pre-commit - just has `#shellcheck` comment
 - [x] couchdb/docker-entrypoint.sh - just has `#shellcheck` comment
 - [x] nginx/ssl-install.sh
 - [x] nginx/tests/ssl-install.bats
 - [x] scripts/add-local-ip-certs-to-docker-4.x.sh
 - [x] scripts/build/blank-link-check.sh
 - [x] scripts/ci/list-shellscripts.sh
 - [x] scripts/compress_and_archive_docker_logs.sh
 - [x] scripts/deploy/cht-deploy - just has `#shellcheck` comment
 - [x] scripts/deploy/troubleshooting/describe-deployment
 - [x] scripts/deploy/troubleshooting/list-all-resources
 - [x] scripts/deploy/troubleshooting/list-deployments
 - [x] scripts/deploy/troubleshooting/restart-deployment
 - [x] scripts/deploy/troubleshooting/view-logs
 - [x] scripts/docker-helper-4.x/cht-docker-compose.sh - fixed in a025cd9 74a9368 cd709da 
 - [x] scripts/docker-helper/cht-docker-compose.sh
 - [x] scripts/docker-helper/docker-status.sh
 - [x] tests/integration/haproxy/keep-alive-script/cmd.sh
 - [x] tests/scalability/prepare-ec2.sh
 - [x] tests/scalability/run_suite.sh
 - [x] tests/scalability/start-ec2-cht.sh

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8557-shellcheck/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8557-shellcheck/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8557-shellcheck/docker-compose/cht-couchdb-clustered.yml)

# License


.github/workflows/test_shell.yml
```diff
@@ -0,0 +1,27 @@
+name: Test Shellscripts (Conditional)
+on:
+  pull_request:
+    paths:
+      - '**/*.bats'
+      - '**/*.sh'
```
nydr, line 6: Not convinced conditional run is the right thing to do here as it only takes 30 sec including install step, but I'm hesitant to add another dependency on the main test since it's already fairly cluttered.

garethbowen: I think it should be part of the `npm run lint` command so devs just run one command and everything is linted, but it would be a shame to add more to the dev environment setup requirements. There is an [npm package](https://www.npmjs.com/package/shellcheck) that might be useful to do the installation for us. Might be worth looking in to...

garethbowen:
Changed the PR to "chore" not "feat" as it's not a feature from an end user perspective.

.github/workflows/test_shell.yml
nydr: That's a good suggestion, I'm also making use of shfmt to list all shell scripts which is needed for shellcheck to find and test those files. I can't use `*.sh` and `*.bats` since not all of them have those extensions, but something like 

```shell
for f in $(git ls-files); do
    head -n1 $f | egrep -q '#!/bin/(ba)?sh' && echo $f;
done
```
might work as an alternative
(also need to consider scripts with /usr/bin/env bash and `*.bats` extensions so not exactly this script)

nydr:
PR updated with adding shellcheck npm package dependency as suggested and replacing the usage of `shfmt` with a shell script for listing files to test, I considered using js, but didn't want start a project of manually parsing `.gitignore` or `.git/objects` with git worktrees at this point

Feel free to take over this PR if anyone has the bandwith. Keep in mind that it is technically changing the behaviour of ~22 scripts with low to no related tests even if most of these behaviour changes are unlikely to be hit and if they were the previous behaviour is most likely the undesired one

mrjones-plip:
While I do concur - and think this is worthwhile effort - we're touching a LOT of scripts.  I think we should smoke test each script that got updated - skipping any that will hit CI right away that might be hard to trigger locally.

I'll grab this next week!

mrjones-plip (approved):
after a careful review, and some fix-it commits since `nydr` isn't working on this anymore, this is good to go!

PR #9143, opened 2024-05-24
fix(na): update z_release_major.md forum post template

michaelkohn:
Updated the forum post template when releasing.

### DIFF
--- a/.github/ISSUE_TEMPLATE/z_release_major.md
+++ b/.github/ISSUE_TEMPLATE/z_release_major.md
@@ -52,13 +52,26 @@ If all is good, then in 24h, I will start the release. Thanks!
 ```
 *We're excited to announce the release of {{version}} of {{product}}*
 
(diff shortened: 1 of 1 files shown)


michaelkohn:
> Is this a copy and paste job from the docs release notes?

I think most of the content is coming from the release notes, though it is obviously in a bit different format. If it's possible to automagically populate the forum post template with content from the release notes, that'd be cool!

mrjones-plip:
gotcha!  let's not overthink this and just go for the change - we're iterating so quickly on releases that we'll know if we need to change it or not soon!

mrjones-plip (approved):

mrjones-plip:
whoa!! are you a GH admin @michaelkohn ?!  CI wasn't passing and i didn't think you could merge to `master` without passing CI?!

image

michaelkohn:
¯\_(ツ)_/¯

No clue!

Issue #9123, opened 2024-05-14
Add e2e test for upgrading from build


garethbowen:
**Describe the issue**

We have an e2e test that ensures it's possible to upgrade to the build version, but we don't have anything that checks it's possible to upgrade _from_ the build version. This meant a recent [critical regression](https://github.com/medic/cht-core/issues/9117) got missed in CI and released to the market and only got detected when other branch builds tried to upgrade from the released version.

**Describe the improvement you'd like**

Add a CI test/suite that after upgrading to the build version it is possible to upgrade to the next version. This is a little tricky because the next version doesn't exist yet. We could use the `master` version or possibly install the build version over the top of itself.

**Describe alternatives you've considered**

tatilepizs:
Completed in PR #9215

PR #9215, opened 2024-06-26
chore(#9123): add scenario that update from build version to 'master'

tatilepizs:
# Description

Add a scenario to the individual test `should upgrade to current branch` to verify that it is possible to update from a specific build version, `master` branch it is being used for this test.

Related to issue #9123 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:update-from-build-version/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:update-from-build-version/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:update-from-build-version/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/tests/e2e/upgrade/upgrade.wdio-spec.js
+++ b/tests/e2e/upgrade/upgrade.wdio-spec.js
@@ -49,6 +49,24 @@ const deleteUpgradeLogs = async () => {
   await utils.logsDb.bulkDocs(logs);
 };
 
+const upgradeVersion = async (branchVersion) => {
+  await upgradePage.goToUpgradePage();
+  await upgradePage.expandPreReleasesAccordion();
+
+  await (await upgradePage.getInstallButton(branchVersion, TAG)).click();
+  await (await upgradePage.upgradeModalConfirm()).click();
+
+  await (await upgradePage.cancelUpgradeButton()).waitForDisplayed();
+  await (await upgradePage.deploymentInProgress()).waitForDisplayed();
+  await (await upgradePage.deploymentInProgress()).waitForDisplayed({ reverse: true, timeout: 100000 });
+
+  if (testFrontend) {
+    // https://github.com/medic/cht-core/issues/9186
+    // this is an unfortunate incompatibility between current API and admin app in the old version
+    await (await upgradePage.deploymentComplete()).waitForDisplayed();
+  }
+};
+
 describe('Performing an upgrade', () => {
   before(async () => {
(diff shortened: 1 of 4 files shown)


tests/e2e/upgrade/upgrade.wdio-spec.js
```diff
@@ -145,6 +149,28 @@ describe('Performing an upgrade', () => {
...
+    await upgradeVersion('master');
+
+    expect(await upgradePage.getBuild()).to.include('alpha');
+    await commonPage.goToAboutPage();
+    await commonPage.waitForPageLoaded();
+    await (await aboutPage.aboutCard()).waitForDisplayed();
+    expect(await aboutPage.getVersion()).to.include('master');
+
+    // installing again the BRANCH version for further validations
+    await upgradeVersion(BRANCH);
```
tatilepizs, line 171: @garethbowen I had to upgrade the version to the build version after I installed `master` because if I didn't do it the next tests would fail. See [this](https://github.com/medic/cht-core/actions/runs/9684357961/job/26723377655?pr=9215) failed job

Do you think that it will be better just to re-install the build version over the top of itself instead of installing `master`? 
The problem with doing that is that we can not verify that the second time that we installed it was successfully installed it

garethbowen: Yeah great point. Whatever we do is really risky because if the downgrade fails for any reason then the rest of the tests will fail in a confusing way, and we don't really need to test the downgrade option because we don't support that in production!

You could either run this test at the end of the suite so it doesn't matter what state it's left in, OR you could run this test in its own suite. I think I prefer the second option because upgrade testing is so challenging to get right, you really want to know you're starting with a clean slate.

What do you think?

tatilepizs: Yes, I am agree
I will isolate it and also add it at the end, just in case 😅 
Thank you!

lorerod (approved):
What a great comeback @tatilepizs! Thank you so much for this.

tatilepizs:
@lorerod can you please do a quick check again? I made some changes since you approved it. thanks! 😊

lorerod (approved):
LGTM @tatilepizs

PR #9220, opened 2024-06-28
feat(#9227): add validation function for luhn algorithm

ChinHairSaintClair:
# Description

#9227

The Luhn algorithm is classified as:
> designed to protect against accidental errors, not malicious attacks. Most credit cards and many government identification numbers use the algorithm as a simple method of distinguishing valid numbers from mistyped or otherwise incorrect numbers. 

It has several applications, as shown below. Specifically, we use it to validate South African ID numbers.

Other uses:
* Credit card numbers
* IMEI numbers
* National Provider Identifier numbers in the United States
* Canadian social insurance numbers
* Israeli ID numbers
* South African Tax reference numbers
* Swedish national identification numbers
* Swedish Corporate Identity Numbers (OrgNr)
* Greek Social Security Numbers (ΑΜΚΑ)
* ICCID of SIM cards
* European patent application numbers
* Survey codes appearing on McDonald's, Taco Bell, and Tractor Supply Co. receipts
* United States Postal Service package tracking numbers use a modified Luhn algorithm

https://en.wikipedia.org/wiki/Luhn_algorithm

# Code review checklist

# License


webapp/tests/mocha/unit/enketo/medic-xpath-extensions.spec.js
```diff
@@ -380,4 +380,39 @@ describe('medic-xpath-extensions', function () {
       });
     });
   });
+
+  describe('validateIdNumber()', () => {
```
garethbowen, line 384: Let's rename this to the name of the function it's testing

```suggestion
  describe('cht:luhn()', () => {
```

webapp/tests/mocha/unit/enketo/medic-xpath-extensions.spec.js
```diff
@@ -380,4 +380,39 @@ describe('medic-xpath-extensions', function () {
...
+    const test = func['cht:luhn'];
+
+    it('should return true for a valid ID number', () => {
+      const idNumber = { t: 'arr', v: [{ textContent: '6505285125086' }] };
+      const result = test(idNumber);
+      assert.strictEqual(result.v, true);
+    });
+    it('should return false for an ID number that does not contain 13 digits', () => {
```
garethbowen, line 392: As far as I can tell the luhn algorithm doesn't have a length restriction and in other contexts having different lengths would be useful, eg: credit cards have 16. Can we remove this and have length validation done through a separate xpath call?

webapp/src/js/enketo/medic-xpath-extensions.js
```diff
@@ -190,6 +213,7 @@ module.exports = {
     'difference-in-months': (d1, d2) => dateDiff(d1, d2, MOMENT_KEYS.MONTHS), // To be deprecated
     'cht:difference-in-weeks': (d1, d2) => dateDiff(d1, d2, MOMENT_KEYS.WEEKS),
     'cht:difference-in-days': (d1, d2) => dateDiff(d1, d2, MOMENT_KEYS.DAYS),
+    'cht:luhn': (number) => luhn(number),
```
garethbowen, line 216: Can we use a verb instead of a noun for the fn name? eg: `validate-luhn`? I think it'd be more self explanatory.

webapp/src/js/enketo/medic-xpath-extensions.js
```diff
@@ -162,6 +163,28 @@ const dateDiff = function (startDateObj, endDateObj, key) {
...
+const luhn = function (number) {
+  number = getValue(number);
+  number = number.toString().trim().replace(/\s/g, '');
+
+  if (number.length !== 13 || !/^\d+$/.test(number)) {
+    return XPR.bool(false);
+  }
+
+  const digits = number.split('').map(Number);
+  const sum = digits.reduce((acc, digit, index) => {
```
garethbowen, line 175: Our style guide advises to [avoid reduce](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/#avoid-reduce) for readability reasons. The wikipedia reference you linked has a typescript version which is more readable (IMO): https://en.wikipedia.org/wiki/Luhn_algorithm#TypeScript

garethbowen (changes requested):
Very cool, thanks!

I've added some comments inline.

Could you also raise an issue to explain this feature? That way it can get documented and included in the release notes so others can hear about it.

webapp/tests/mocha/unit/enketo/medic-xpath-extensions.spec.js
ChinHairSaintClair: The function also does the job of trimming the string, and ensuring that no other characters other than digits are present. Would it not be an opportune place, after those initial cleanups/checks to also check the intended length?
Rather than having to do all the cleanup again in a separate function. An expected length number could optionally be passed in. Since American Express credit cards have a length of 15 (and interestingly CVV of 4 🤔).

ChinHairSaintClair: I've pushed some updates based on the implementation explained above. If the length validation doesn't meet expectations, I'll be happy to adjust it accordingly.

webapp/src/js/enketo/medic-xpath-extensions.js
```diff
@@ -163,24 +163,28 @@ const dateDiff = function (startDateObj, endDateObj, key) {
   return XPR.number(endDate.diff(startDate, key));
 };
 
-const luhn = function (number) {
+const luhn = function (number, expLength) {
   number = getValue(number);
   number = number.toString().trim().replace(/\s/g, '');
 
-  if (number.length !== 13 || !/^\d+$/.test(number)) {
+  if (!/^\d+$/.test(number) || (expLength && number.length !== expLength)) {
```
garethbowen, line 170: I think this would be more reusable if the length validation was done as a separate constraint, eg: `cht:validate-luhn(.) and string-length(.) === 13`.

This gives more flexibility in the form construction, for example, you might want `<=` instead of `===`.

What do you think?

garethbowen (changes requested):
One suggestion left and then I think this will be ready.

webapp/src/js/enketo/medic-xpath-extensions.js
ChinHairSaintClair: I've, albeit with limited exposure, seldomly seen the luhn algorithm used without some sort of length validation when it comes to the real world applications as mentioned in the PR description.
That being said, you have a valid point that the type of length check required might differ from impl to impl (`===` vs `<=`).
Still, having the length check outside would mean that the user would also need to strip spaces themselves.
For example `6505  2851  2508 6` and `  6505285125086  ` would satisfy the luhn check as it is written now, but would fail the proposed `cht:validate-luhn(.) and string-length(.) === 13` check in the form. Unless the `string-length` automatically removes all spaces (`.replace(/\s/g, '')`) ?
Thinking about it now, perhaps stripping spaces would also be a useful utility function.
What do you think?

garethbowen: Yeah that's a good point.

In the implementation you've made it optional so maybe that's the way to go - it's an optional parameter and you can use it to validate the trimmed length, OR you can omit it if you want to do your own length validation.

If you agree with this optional approach, could you add a test or two that proves it works when you don't provide a length?

ChinHairSaintClair: You're right on the money there! Awesome, I'll get those tests in for you ASAP! 🚀 
Thank you for taking the time to talk through the various approaches; it really is appreciated.

garethbowen (approved):
Nice one, thanks!

PR #9232, opened 2024-07-01
chore(#9231): Enable place filter aggregate targets for users with multiple facility_ids

### LINKED ISSUE #9231: Allow users to view aggregate targets and filter by facility_id and reporting period
After changes to allow users to have multiple places in [4.9.0](https://docs.communityhealthtoolkit.org/core/releases/4.9.0/) we temporarily disabled users assigned to multiple facilities to view aggregate targets. 

Currently in `4.9.0`, users assigned to multiple `facility_ids` cannot view aggregate targets.  

 - Enable users assigned to multiple facilities to view target aggregates page 
 - Update Target Aggregates Component to display filter button
 - Add sidebar filter that is displayed when the filter button is clicked
 - Allow users to filter aggregate targets by reporting period (current month or previous month)
 - Add radio buttons to the sidebar filter that allows users assigned to multiple facilities to filter aggregate targets by facility or reporting period
 -  Add radio buttons to the sidebar filter that allows users assigned to one facility to filter aggregate targets by reporting period  
**Additional context**

screens

Benmuiruri:
# Description

Closes #9231 

# Code review checklist

# License

### DIFF
--- a/webapp/src/css/inbox.less
+++ b/webapp/src/css/inbox.less
@@ -406,6 +406,20 @@ body {
     left: 0;
     right: 0;
     display: flex;
+    .right-side-filter {
+      margin-left: auto;
+      .open-filter-label {
+        color: @analytics-filter-text-color;
+      }
+      .btn.open-filter {
+        background-color: transparent;
+        margin: 0;
+        .fa {
+          font-size: @font-extra-large;
+          color: @filter-icon-color;
+        }
+      }
+    }
   }
 
   .freetext-filter {
@@ -1712,6 +1726,10 @@ mm-search-bar {
     display: none;
   }
 
+  .open-filter-label {
+    display: none;
+  }
+
   mm-search-bar {
     padding: 5px;
     .mm-search-bar-container {
@@ -1729,10 +1747,6 @@ mm-search-bar {
 
       .btn.open-filter {
         margin: 0;
-
-        .open-filter-label {
-          display: none;
-        }
 
         .filter-counter {
           position: absolute;
--- a/webapp/src/css/variables.less
+++ b/webapp/src/css/variables.less
@@ -42,6 +42,7 @@
 @sidebar-background-color: @white;
 @filter-icon-color: @gray-ultra-dark;
 @filter-text-color: @gray-medium;
+@analytics-filter-text-color: @black;
 @progress-bar-background-color: @white;
 @targets-progress-bar-unmet-goal: @gray-medium-dark;
 @backdrop-color: @black;
--- a/webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.ts
+++ b/webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.ts
@@ -1,4 +1,11 @@
-import { AfterContentChecked, AfterContentInit, Component, Input, OnDestroy } from '@angular/core';
+import {
+  AfterContentChecked,
+  AfterContentInit,
+  Component,
+  EventEmitter,
+  Input,
+  OnDestroy,
+  Output } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
 import { Subscription } from 'rxjs';
 
@@ -8,6 +15,8 @@ import { Subscription } from 'rxjs';
 })
 export class AnalyticsFilterComponent implements AfterContentInit, AfterContentChecked, OnDestroy {
   @Input() analyticsModules: any[] = [];
+  @Input() showFilterButton: boolean = false;
+  @Output() toggleFilter: EventEmitter<any> = new EventEmitter();
   activeModule;
   subscriptions: Subscription = new Subscription();
 
--- a/webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.html
+++ b/webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.html
@@ -1,0 +1,61 @@
+<section class="sidebar-filter">
+  <div class="sidebar-backdrop" [ngClass]="{ hidden: !isOpen }"
+    (click)="toggleSidebarFilter()"></div>
+
+  <div class="sidebar-main" [ngClass]="{ hidden: !isOpen }">
+    <div class="sidebar-header">
+      <p class="sidebar-title">Filter</p>
+      <a class="sidebar-reset" [class.disabled]="disabled"
+        (click)="resetFilters()">Reset</a>
+      <a class="sidebar-close fa fa-times" (click)="toggleSidebarFilter()"></a>
+    </div>
+
+    <div class="sidebar-body">
+      <mat-accordion multi>
+
+        <!-- 3. Date -->
+        <mat-expansion-panel id="date-filter-accordion"
+          [class.filter-error]="dateFilterError">
+          <mat-expansion-panel-header>
+            <ng-container [ngTemplateOutlet]="headerTemplate"
+              [ngTemplateOutletContext]="{
+                label: 'reports.sidebar.filter.date',
+                filters: [ 'fromDateFilter', 'toDateFilter' ]
+              }">
+            </ng-container>
+          </mat-expansion-panel-header>
+          <mat-panel-description>
+            <label>{{ 'reports.sidebar.filter.from_date' | translate }}</label>
+            <mm-date-filter #fromDate class="filter" fieldId="fromDateFilter"
+              (search)="applyFilters()" (onError)="showDateFilterError($event)"
+              [disabled]="disabled" [isStartDate]="true">
+            </mm-date-filter>
+
+            <label>{{ 'reports.sidebar.filter.to_date' | translate }}</label>
+            <mm-date-filter #toDate class="filter" fieldId="toDateFilter"
+              (search)="applyFilters()" (onError)="showDateFilterError($event)"
+              [disabled]="disabled">
+            </mm-date-filter>
+            <p class="filter-error-message" *ngIf="dateFilterError">{{
+              dateFilterError | translate }}</p>
+          </mat-panel-description>
+        </mat-expansion-panel>
+      </mat-accordion>
+    </div>
+
+    <div class="sidebar-footer">
+      <button type="button" class="btn btn-primary"
+        (click)="toggleSidebarFilter()">Submit</button>
+    </div>
+  </div>
+</section>
+
+<ng-template #headerTemplate let-numSelected="numSelected" let-label="label"
+  let-filters="filters">
+  <p class="title">{{ label | translate }}</p>
+  <div class="chip" *ngIf="numSelected" [class.disabled]="disabled">
+    <span>{{ numSelected | translate }}</span>
+    <span class="fa fa-times"
+      (click)="clearFilters(filters); $event.stopPropagation();"></span>
+  </div>
+</ng-template>
--- a/webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
+++ b/webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
@@ -1,0 +1,88 @@
+import { AfterViewInit, Component, EventEmitter, Input, OnDestroy, Output, ViewChild } from '@angular/core';
+import { Store } from '@ngrx/store';
+
+import { GlobalActions } from '@mm-actions/global';
+import { DateFilterComponent } from '@mm-components/filters/date-filter/date-filter.component';
+import { TelemetryService } from '@mm-services/telemetry.service';
+
+type FilterComponent = DateFilterComponent ;
+
+@Component({
+  selector: 'mm-analytics-target-aggregates-sidebar-filter',
+  templateUrl: './analytics-target-aggregates-sidebar-filter.component.html'
+})
+export class AnalyticsTargetAggregatesSidebarFilterComponent implements AfterViewInit, OnDestroy {
+  @Output() search: EventEmitter<any> = new EventEmitter();
+  @Input() disabled;
+
+
+  @ViewChild('fromDate') fromDateFilter!: DateFilterComponent;
+  @ViewChild('toDate') toDateFilter!: DateFilterComponent;
+
+  private globalActions;
+  private filters: FilterComponent[] = [];
+  isResettingFilters = false;
+  isOpen = false;
+  dateFilterError = '';
+
+  constructor(
+    private store: Store,
+    private telemetryService: TelemetryService,
+  ) {
+    this.globalActions = new GlobalActions(store);
+  }
+
+  ngAfterViewInit() {
+    this.filters = [
+      this.fromDateFilter,
+      this.toDateFilter,
+    ];
+  }
+
+  applyFilters() {
+    if (this.isResettingFilters || this.disabled) {
+      return;
+    }
+    this.search.emit();
+  }
+
+  clearFilters(fieldIds?) {
+    if (this.disabled) {
+      return;
+    }
+    const filters = fieldIds ? this.filters.filter(filter => fieldIds.includes(filter.fieldId)) : this.filters;
+    filters.forEach(filter => filter.clear());
+  }
+
+  resetFilters() {
+    if (this.disabled) {
+      return;
+    }
+    this.isResettingFilters = true;
+    this.globalActions.clearFilters('search');
+    this.clearFilters();
+    this.isResettingFilters = false;
+    this.applyFilters();
+  }
+
+  toggleSidebarFilter() {
+    console.log('clicked from filter');
+    this.isOpen = !this.isOpen;
+    console.log('Sidebar isOpen:', this.isOpen);
+    this.globalActions.setSidebarFilter({ isOpen: this.isOpen });
+
+    if (this.isOpen) {
+      // Counting every time the user opens the sidebar filter in analytics_targets_aggregrate tab.
+      this.telemetryService.record('sidebar_filter:analytics_target_aggregates:open');
+    }
+  }
+
+  showDateFilterError(error) {
+    this.dateFilterError = error || '';
+  }
+
+  ngOnDestroy() {
+    this.globalActions.clearSidebarFilter();
+    this.globalActions.clearFilters();
+  }
+}
--- a/webapp/src/ts/modules/analytics/analytics-target-aggregates.component.html
+++ b/webapp/src/ts/modules/analytics/analytics-target-aggregates.component.html
@@ -20,6 +20,11 @@
     <p class="loading-status" *ngIf="!error && !aggregates?.length">{{'analytics.target.aggregates.no.targets' | translate}}</p>
   </div>
 
+  <mm-analytics-target-aggregates-sidebar-filter
+    class="col-sm-3 sidebar-filter-wrapper"
+    *ngIf="useSidebarFilter">
+  </mm-analytics-target-aggregates-sidebar-filter>
+
   <router-outlet></router-outlet>
 </div>
 
--- a/webapp/src/ts/modules/analytics/analytics-target-aggregates.component.ts
+++ b/webapp/src/ts/modules/analytics/analytics-target-aggregates.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnDestroy, OnInit } from '@angular/core';
+import { Component, OnDestroy, OnInit, ViewChild } from '@angular/core';
 import { Store } from '@ngrx/store';
 import { combineLatest, Subscription } from 'rxjs';
 
@@ -6,12 +6,18 @@ import { TargetAggregatesActions } from '@mm-actions/target-aggregates';
 import { TargetAggregatesService } from '@mm-services/target-aggregates.service';
 import { Selectors } from '@mm-selectors/index';
 import { PerformanceService } from '@mm-services/performance.service';
+import { SidebarFilterService } from '@mm-services/sidebar-filter.service';
+import { AnalyticsTargetAggregatesSidebarFilterComponent }
+  from './analytics-target-aggregates-sidebar-filter.component';
 
 @Component({
   selector: 'analytics-target-aggregates',
   templateUrl: './analytics-target-aggregates.component.html',
 })
 export class AnalyticsTargetAggregatesComponent implements OnInit, OnDestroy {
+  @ViewChild(AnalyticsTargetAggregatesSidebarFilterComponent) analyticsTargetAggregatesFilterComponent?:
+   AnalyticsTargetAggregatesSidebarFilterComponent;
+
   private targetAggregatesActions: TargetAggregatesActions;
   private trackPerformance;
   subscriptions: Subscription = new Subscription();
@@ -20,11 +26,14 @@ export class AnalyticsTargetAggregatesComponent implements OnInit, OnDestroy {
   aggregates: any = null;
   selected = null;
   error = null;
+  useSidebarFilter = true;
+  //To-do : What was this for in Reports?
 
   constructor(
     private store: Store,
     private targetAggregatesService: TargetAggregatesService,
     private performanceService: PerformanceService,
+    private sidebarFilterService: SidebarFilterService
   ) {
     this.targetAggregatesActions = new TargetAggregatesActions(store);
   }
@@ -32,6 +41,7 @@ export class AnalyticsTargetAggregatesComponent implements OnInit, OnDestroy {
   ngOnInit(): void {
     this.trackPerformance = this.performanceService.track();
     this.subscribeToStore();
+    this.subscribeSidebarFilter();
     this.getTargetAggregates();
   }
 
@@ -87,5 +97,19 @@ export class AnalyticsTargetAggregatesComponent implements OnInit, OnDestroy {
           recordApdex: true,
         });
       });
+  }
+
+  private subscribeSidebarFilter() {
+    if (!this.useSidebarFilter) {
+      return;
+    }
+
+    const subscription = this.sidebarFilterService.getSidebarToggleEvents()
+      .subscribe(() => {
+        if (this.analyticsTargetAggregatesFilterComponent) {
+          this.analyticsTargetAggregatesFilterComponent.toggleSidebarFilter();
+        }
+      });
+    this.subscriptions.add(subscription);
   }
 }
--- a/webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.html
+++ b/webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.html
@@ -20,4 +20,10 @@
     </ul>
   </span>
 
+  <div class="right-side-filter" *ngIf="showFilterButton">
+    <button class="btn open-filter right-side-filter" (click)="toggleFilter.emit()">
+      <span class="fa fa-sliders"></span>
+      <span class="open-filter-label">{{ 'search_bar.filter.label' | translate }}</span>
+    </button>
+  </div>
 </div>
--- a/webapp/src/ts/modules/analytics/analytics.component.html
+++ b/webapp/src/ts/modules/analytics/analytics.component.html
@@ -1,6 +1,10 @@
 <div class="filters">
   <div class="inner">
-    <mm-analytics-filters [analyticsModules]="analyticsModules"></mm-analytics-filters>
+    <mm-analytics-filters
+    [analyticsModules]="analyticsModules"
+    [showFilterButton]="showFilterButton"
+    (toggleFilter)="toggleFilter()">
+    </mm-analytics-filters>
     <mm-navigation></mm-navigation>
   </div>
 </div>
--- a/webapp/src/ts/modules/analytics/analytics.component.ts
+++ b/webapp/src/ts/modules/analytics/analytics.component.ts
@@ -1,24 +1,37 @@
-import { Component, OnDestroy, OnInit } from '@angular/core';
+import { Component, ViewChild, OnDestroy, OnInit } from '@angular/core';
 import { Store } from '@ngrx/store';
-import { Subscription } from 'rxjs';
+import { Subscription, filter } from 'rxjs';
 
 import { GlobalActions } from '@mm-actions/global';
 import { Selectors } from '@mm-selectors/index';
+import { AnalyticsTargetAggregatesSidebarFilterComponent }
+  from './analytics-target-aggregates-sidebar-filter.component';
+import { SidebarFilterService } from '@mm-services/sidebar-filter.service';
+import { NavigationEnd, Router } from '@angular/router';
 
 @Component({
   templateUrl: './analytics.component.html'
 })
 export class AnalyticsComponent implements OnInit, OnDestroy {
+  @ViewChild(AnalyticsTargetAggregatesSidebarFilterComponent) analyticsTargetAggregatesFilterComponent?:
+   AnalyticsTargetAggregatesSidebarFilterComponent;
+
   private globalActions;
   subscriptions: Subscription = new Subscription();
   analyticsModules = [];
+  showFilterButton = false;
 
-  constructor(private store: Store) {
+  constructor(
+    private store: Store,
+    private router: Router,
+    private sidebarFilterService: SidebarFilterService
+  ) {
     this.globalActions = new GlobalActions(store);
   }
 
   ngOnInit() {
     this.subscribeToStore();
+    this.subscribeToRouteChanges();
     this.globalActions.unsetSelected();
   }
 
@@ -31,6 +44,23 @@ export class AnalyticsComponent implements OnInit, OnDestroy {
       .select(Selectors.getAnalyticsModules)
       .subscribe(analyticsModules => this.analyticsModules = analyticsModules);
     this.subscriptions.add(subscription);
+  }
+
+  private subscribeToRouteChanges() {
+    const routeSubscription = this.router.events.pipe(
+      filter(event => event instanceof NavigationEnd)
+    ).subscribe((event: NavigationEnd) => {
+      this.canDisplayFilterButton(event.urlAfterRedirects);
+    });
+    this.subscriptions.add(routeSubscription);
+  }
+
+  private canDisplayFilterButton(currentRoute) {
+    this.showFilterButton = currentRoute.includes('target-aggregates');
+  }
+
+  toggleFilter() {
+    this.sidebarFilterService.toggleSidebarFilter();
   }
 }
 
--- a/webapp/src/ts/modules/modules.module.ts
+++ b/webapp/src/ts/modules/modules.module.ts
@@ -39,6 +39,8 @@ import { PrivacyPolicyComponent } from '@mm-modules/privacy-policy/privacy-polic
 import { AnalyticsTargetsComponent } from '@mm-modules/analytics/analytics-targets.component';
 import { AnalyticsModulesComponent } from '@mm-modules/analytics/analytics-modules.component';
 import { AnalyticsTargetAggregatesComponent } from '@mm-modules/analytics/analytics-target-aggregates.component';
+import { AnalyticsTargetAggregatesSidebarFilterComponent }
+  from '@mm-modules/analytics/analytics-target-aggregates-sidebar-filter.component';
 import {
   AnalyticsTargetAggregatesDetailComponent
 } from '@mm-modules/analytics/analytics-target-aggregates-detail.component';
@@ -76,6 +78,7 @@ import { DirectivesModule } from '@mm-directives/directives.module';
     AnalyticsModulesComponent,
     AnalyticsTargetAggregatesComponent,
     AnalyticsTargetAggregatesDetailComponent,
+    AnalyticsTargetAggregatesSidebarFilterComponent,
     TasksComponent,
     TasksContentComponent,
     TasksGroupComponent,
--- a/webapp/src/ts/services/sidebar-filter.service.ts
+++ b/webapp/src/ts/services/sidebar-filter.service.ts
@@ -1,0 +1,17 @@
+import { Injectable } from '@angular/core';
+import { Subject } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class SidebarFilterService {
+  private sideBarToggleSubject = new Subject<void>();
+
+  toggleSidebarFilter(): void {
+    this.sideBarToggleSubject.next();
+  }
+
+  getSidebarToggleEvents() {
+    return this.sideBarToggleSubject.asObservable();
+  }
+}


webapp/src/css/variables.less
```diff
@@ -42,6 +42,7 @@
 @sidebar-background-color: @white;
 @filter-icon-color: @gray-ultra-dark;
 @filter-text-color: @gray-medium;
+@analytics-filter-text-color: @black;
```
latin-panda, line 45:
```suggestion
@filter-text-color: @black;
```

webapp/src/ts/modules/analytics/analytics-target-aggregates.component.ts
```diff
@@ -1,17 +1,23 @@
...
+import { SidebarFilterService } from '@mm-services/sidebar-filter.service';
+import { AnalyticsTargetAggregatesSidebarFilterComponent }
+  from './analytics-target-aggregates-sidebar-filter.component';
 
 @Component({
   selector: 'analytics-target-aggregates',
   templateUrl: './analytics-target-aggregates.component.html',
 })
 export class AnalyticsTargetAggregatesComponent implements OnInit, OnDestroy {
+  @ViewChild(AnalyticsTargetAggregatesSidebarFilterComponent) analyticsTargetAggregatesFilterComponent?:
```
latin-panda, line 18:
```suggestion
  @ViewChild(AnalyticsTargetAggregatesSidebarFilterComponent) sidebarFilter?:
```

latin-panda (commented):
Some feedback we reviewed during the call

latin-panda:
@Benmuiruri please change the PR title from chore to `feat`

webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.ts
```diff
@@ -1,19 +1,54 @@
...
-import { Subscription } from 'rxjs';
+import {
+  AfterContentChecked,
+  AfterContentInit,
+  Component,
+  EventEmitter,
+  Input,
+  OnDestroy,
+  OnInit,
+  Output } from '@angular/core';
```
latin-panda, line 9:
```suggestion
  Output 
} from '@angular/core';
```
We usually close things in the next like, so it looks more like block

webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.ts
```diff
@@ -30,9 +65,67 @@ export class AnalyticsFilterComponent implements AfterContentInit, AfterContentC
     this.subscriptions.unsubscribe();
   }
 
+  private subscribeToStore() {
+    const subscription = this.store.select(Selectors.getSidebarFilter).subscribe(({ isOpen }) => {
+      this.isOpen = isOpen;
+    });
+    this.subscriptions.add(subscription);
+  }
```
latin-panda, line 73:
```suggestion
  private subscribeToStore() {
    const subscription = this.store
      .select(Selectors.getSidebarFilter)
      .subscribe(({ isOpen }) => this.isOpen = isOpen);
    this.subscriptions.add(subscription);
  }
```
On function chaining, each function goes in separate lines

webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.ts
```diff
@@ -30,9 +65,67 @@ export class AnalyticsFilterComponent implements AfterContentInit, AfterContentC
...
       return module.id === this.route.snapshot?.firstChild?.data?.moduleId;
     });
   }
+
+  private subscribeToRouteChanges() {
+    const routeSubscription = this.router.events.pipe(
+      filter(event => event instanceof NavigationEnd)
+    ).subscribe(() => {
+      this.canDisplayFilterButton();
+    });
```
latin-panda, line 86: Same:
```suggestion
    const routeSubscription = this.router.events
      .pipe(filter(event => event instanceof NavigationEnd))
      .subscribe(() => this.canDisplayFilterButton());
```

webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.ts
```diff
@@ -30,9 +65,67 @@ export class AnalyticsFilterComponent implements AfterContentInit, AfterContentC
...
+  }
+
+  private hasMultipleFacilities() {
+    return this.userSettingsService
+      .get()
+      .then((userSettings: any) => {
+        const facility = userSettings.facility_id;
+        return Array.isArray(facility) && facility.length > 1;
+      });
+  }
```
latin-panda, line 97: We have this kind of check everywhere in the code. 
Please move the `hasMultipleFacilities` to `userSettingsService` to centralize it for maintenance and reuse it.

webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.ts
```diff
@@ -30,9 +65,67 @@ export class AnalyticsFilterComponent implements AfterContentInit, AfterContentC
...
+
+  private checkPermissions() {
+    return Promise
+      .all([
+        this.authService.has('!can_view_old_filter_and_search'),
+        this.authService.has('!can_view_old_action_bar')
+      ])
+      .then(([cannotViewOldFilter, cannotViewOldActionBar]) => {
+        return cannotViewOldFilter && cannotViewOldActionBar;
+      })
```
latin-panda, line 107: The auth service supports an array for checking multiple permissions, so it doesn't need 2 separate calls.

webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.ts
```diff
@@ -30,9 +65,67 @@ export class AnalyticsFilterComponent implements AfterContentInit, AfterContentC
...
+      .catch(error => {
+        console.error('Error checking permissions:', error);
+        return false;
+      });
+  }
+
+  private isTargetAggregates() {
+    return this.router.url.startsWith(AGGREGATE_TARGET_URL);
```
latin-panda, line 115: You can keep the constant at this level and check by the `moduleId` that is part of the router's data object. 

Screenshot 2024-07-09 at 11 26 01 AM

That is already available in this component. See `setActiveModule()` in l77

latin-panda, line 115: 1. You can keep the constant at this level instead of outside the class, because it's only used here.  
2. and check by the `moduleId` that is part of the router's data object. 

Screenshot 2024-07-09 at 11 26 01 AM

3. That is already available in this component. See `setActiveModule()` in L77, so you can reuse that check.

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.html
```diff
@@ -0,0 +1,40 @@
...
+    </div>
+
+    <div class="sidebar-body">
+      <mat-accordion multi>
+        <div class="facility-filter-wrap">
+          <div class="facility-filter-title">
+            <span class="facility-title"> {{ facilityLabel | translate }} </span>
+          </div>
+          <div class="facility-filter-body">
+            <div *ngFor="let facility of userFacilities" class="facility-filter-place">
```
latin-panda, line 17: The double wrap might be unnecessary here. You can use it like this: 
`<ng-container *ngFor=".....`

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.html
```diff
@@ -0,0 +1,40 @@
...
+              </label>
+            </div>
+          </div>
+        </div>
+      </mat-accordion>
+    </div>
+
+    <div class="sidebar-footer">
+      <button type="button" class="btn btn-primary"
+        (click)="toggleSidebarFilter()">Apply</button>
```
latin-panda, line 33: same line
```suggestion
      <button type="button" class="btn btn-primary" (click)="toggleSidebarFilter()">Apply</button>
```

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.html
```diff
@@ -0,0 +1,40 @@
...
+          <div class="facility-filter-title">
+            <span class="facility-title"> {{ facilityLabel | translate }} </span>
+          </div>
+          <div class="facility-filter-body">
+            <div *ngFor="let facility of userFacilities" class="facility-filter-place">
+              <label class="facility-label">
+                <input type="radio" name="facility" [value]="facility._id" class="facility-radio"
+                  [(ngModel)]="selectedFacilityId"
+                  [checked]="facility._id === selectedFacilityId"
+                  (click)="fetchAggregateTargets(facility._id)">
```
latin-panda, line 22: It's usually every argument in separate lines, but this one has too many. So at least we do the angular ones in separate lines.
```suggestion
                <input type="radio" name="facility" class="facility-radio"
                  [value]="facility._id"
                  [(ngModel)]="selectedFacilityId"
                  [checked]="facility._id === selectedFacilityId"
                  (click)="fetchAggregateTargets(facility._id)">
```
Question: Are you sure you need `[value]="facility._id"`? The ` [(ngModel)]="selectedFacilityId"` is double-way binding, so for default values, it's set in the ngModel, and the value isn't needed.

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
```diff
@@ -0,0 +1,110 @@
...
+      console.error('Error initializing Target Sidebar component', err);
+    }
+  }
+
+  ngOnDestroy() {
+    this.subscriptions.unsubscribe();
+  }
+
+  private subscribeToStore() {
+    const subscription = this.store.select(Selectors.getSidebarFilter).subscribe(({ isOpen }) => {
```
latin-panda, line 55: Same, one line per chain function

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+  }
+
+  private getUserFacility() {
+    return this.userSettingsService
+      .get()
+      .then((userSettings: any) => {
+        return this.getPlaces(userSettings.facility_id);
+      });
+  }
```
latin-panda, line 77: This is another function that can be moved to `userSettingsService`. We have in other places the same need to get the user facility data; we can centralize it for maintenance 
That includes `private getPlaces(facilityId) { return this.getDataRecordsService` part, so all in 1 function in `userSettingsService`

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
```diff
@@ -0,0 +1,110 @@
...
+      .then(places => {
+        this.userFacilities = places;
+        if (this.userFacilities) {
+          this.selectedFacilityId = this.userFacilities[0]?._id;
+        }
+      });
+  }
+
+  private getFacilityType() {
+    return this.contactTypesService.getTypeId(this.userFacilities[0]);
```
latin-panda, line 91: It doesn't seem necessary to extract 1 function to a function. It's like too much modularization. Same with the other functions in this PR

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  private setFacilityLabel() {
+    return this.settingsService.get()
+      .then((settings) => {
+        const place = settings.contact_types.find(type => type.id === this.getFacilityType());
+        this.facilityLabel = place.name_key;
```
latin-panda, line 98: 1. It's more efficient to calculate the user facility first into a variable.
2. Clearer names
```suggestion
      .then((settings) => {
      const userFacilityType = this.contactTypesService.getTypeId(this.userFacilities[0])
        const placeType = settings.contact_types.find(type => type.id === userFacilityType);
        this.facilityFilterLabel = placeType.name_key;
```

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
```diff
@@ -0,0 +1,110 @@
...
+        }
+      });
+  }
+
+  private getFacilityType() {
+    return this.contactTypesService.getTypeId(this.userFacilities[0]);
+  }
+
+  private setFacilityLabel() {
+    return this.settingsService.get()
```
latin-panda, line 95: you can await this to make code flatter. Same with the other functions

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
```diff
@@ -0,0 +1,110 @@
...
+      .then((settings) => {
+        const place = settings.contact_types.find(type => type.id === this.getFacilityType());
+        this.facilityLabel = place.name_key;
+      })
+      .catch(error => {
+        this.error = true;
+        console.error('Error fetching facility label', error);
+        this.facilityLabel = FACILITY;
```
latin-panda, line 103: This constant is only used in this function and not exported. So better move it to inside this function

webapp/src/ts/modules/analytics/analytics-target-aggregates.component.html
```diff
@@ -4,8 +4,10 @@
   </div>
 </div>
 
-<div class="page target-aggregates" *ngIf="enabled && !loading">
-  <div id="target-aggregates-list" class="col-sm-4 inbox-items left-pane">
+<div class="inbox page" *ngIf="enabled && !loading">
+  <div id="target-aggregates-list"
+    class="inbox-items left-pane"
+    [ngClass]="{ 'col-sm-3': isSidebarFilterOpen, 'col-sm-4': !isSidebarFilterOpen }">
```
latin-panda, line 10:
```suggestion
  <div 
    id="target-aggregates-list"
    class="inbox-items left-pane"
    [ngClass]="{ 'col-sm-3': isSidebarFilterOpen, 'col-sm-4': !isSidebarFilterOpen }">
```

webapp/src/ts/modules/analytics/analytics-target-aggregates.component.html
```diff
@@ -20,7 +22,14 @@
     <p class="loading-status" *ngIf="!error && !aggregates?.length">{{'analytics.target.aggregates.no.targets' | translate}}</p>
   </div>
 
-  <router-outlet></router-outlet>
+  <div class="content-pane right-pane">
+    <div class="item-content"
+      [ngClass]="{ 'col-sm-6': isSidebarFilterOpen, 'col-sm-8': !isSidebarFilterOpen }">
+    <router-outlet></router-outlet>
+    </div>
```
latin-panda, line 29: One line
identation
```suggestion
    <div class="item-content" [ngClass]="{ 'col-sm-6': isSidebarFilterOpen, 'col-sm-8': !isSidebarFilterOpen }">
      <router-outlet></router-outlet>
    </div>
```

latin-panda (commented):
I did a quick pass and left some feedback but I still need to do a proper review specially the CSS part.

Please add unit tests to the new code.
Also make sure to test the view without this feature, including when the old design is activated to verify it still looks good (can_view_old_filter, can_view_old_actionbar)

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.html
Benmuiruri: I think we still need value to know when the facility is changed. What it seems we don't need is the `[checked]` . From what I have read, `ngModel` handles the checked state.

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
Benmuiruri: Cool, I have removed that one, but I will leave the separation of `getUserFacility` and `getPlaces`

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  private async setFacilityLabel() {
+    const FACILITY = 'Facility';
+    try {
+      const settings = await this.settingsService.get();
+      const userFacilityType = this.contactTypesService.getTypeId(this.userFacilities[0]);
+      const placeType = settings.contact_types.find(type => type.id === userFacilityType);
+
+      this.facilityFilterLabel = placeType.name_key;
```
latin-panda, line 91:
```suggestion
      this.facilityFilterLabel = placeType.name_key || FACILITY;
```

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
```diff
@@ -0,0 +1,102 @@
...
+    } catch (err) {
+      this.error = true;
+      console.error('Error fetching facility label', err);
+      this.facilityFilterLabel = FACILITY;
+    }
+  }
+
+  fetchAggregateTargets(facilityId) {
```
latin-panda, line 99: Remove, not used

webapp/src/ts/modules/analytics/analytics-target-aggregates.component.html
```diff
@@ -4,8 +4,13 @@
   </div>
 </div>
 
-<div class="page target-aggregates" *ngIf="enabled && !loading">
-  <div id="target-aggregates-list" class="col-sm-4 inbox-items left-pane">
+<div
+  class="inbox page target-aggregates" *ngIf="enabled && !loading"
+  [ngClass]="{'sidebar-open': isSidebarFilterOpen, 'sidebar-closed': !isSidebarFilterOpen}">
```
latin-panda, line 9:
```suggestion
<div
  class="inbox page target-aggregates" 
  *ngIf="enabled && !loading"
  [ngClass]="{'sidebar-open': isSidebarFilterOpen, 'sidebar-closed': !isSidebarFilterOpen}">
```

webapp/src/ts/modules/analytics/analytics-target-aggregates.component.html
```diff
@@ -20,7 +25,13 @@
     <p class="loading-status" *ngIf="!error && !aggregates?.length">{{'analytics.target.aggregates.no.targets' | translate}}</p>
   </div>
 
-  <router-outlet></router-outlet>
+  <div class="content-pane right-pane">
+    <div class="item-content" [ngClass]="{ 'col-sm-6': isSidebarFilterOpen, 'col-sm-8': !isSidebarFilterOpen }">
+    <router-outlet></router-outlet>
```
latin-panda, line 30: I think the indentation of <router-outlet></router-outlet> needs fix

webapp/src/ts/services/user-settings.service.ts
```diff
@@ -60,6 +62,22 @@ export class UserSettingsService {
     });
   }
 
+  async hasMultipleFacilities(): Promise<boolean> {
+    return this.get()
+      .then((userSettings:any) => {
+        const userFacility = userSettings.facility_id;
+        return Array.isArray(userFacility) && userFacility.length > 1;
+      });
```
latin-panda, line 70:
```suggestion
    return this
      .get()
      .then(userSettings => {
        const userFacility = userSettings.facility_id;
        return Array.isArray(userFacility) && userFacility.length > 1;
      });
```

webapp/src/ts/services/user-settings.service.ts
```diff
@@ -60,6 +62,22 @@ export class UserSettingsService {
...
+  }
+
+  async getUserFacility(): Promise<any[]> {
+    return this.get()
+      .then((userSettings:any) => {
+        const userFacility = userSettings.facility_id;
+        return this.getDataRecordsService.get(userFacility);
+      });
```
latin-panda, line 78:
```suggestion
    return this
      .get()
      .then(userSettings => this.getDataRecordsService.get(userSettings.facility_id));
```

webapp/src/ts/services/user-settings.service.ts
latin-panda: We avoid `any` as we're working toward defining the data for clarity. Instead, define a type

latin-panda (changes requested):
@Benmuiruri I did a quick pass and I have reviewed 50% only.
I'll review the other 50% later

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.html
```diff
@@ -0,0 +1,38 @@
+<section class="sidebar-filter">
+  <div class="sidebar-backdrop" [ngClass]="{ hidden: !isOpen }" (click)="toggleSidebarFilter()"></div> <!-- //NOSONAR -->
```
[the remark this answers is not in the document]
Benmuiruri: When sonar sees a click it complains about needing a key-up event for keyboard interaction. 

`Add a 'onKeyPress|onKeyDown|onKeyUp' attribute to this <div> tag.`

But in this case for closing the sidebar by clicking the backdrop, I don't see why we would need to offer keyboard interaction.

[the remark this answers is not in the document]
latin-panda: Well, for desktop users, they could have done `(KeyPress.Escape)=close()` to close it, and still valid exp. But if you decide to skip it, then add a comment explaining the reason

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
```diff
@@ -0,0 +1,102 @@
...
+      });
+
+    this.subscriptions.add(subscription);
+  }
+
+  toggleSidebarFilter() {
+    this.isOpen = !this.isOpen;
+    this.globalActions.setSidebarFilter({ isOpen: this.isOpen });
+
+    if (this.isOpen) {
```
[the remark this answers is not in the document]
latin-panda: Actually, just move it to analytics-filter.component.ts where it open the sidebar

ralfudx (approved):
LGTM

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.html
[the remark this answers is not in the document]
Benmuiruri: My rationale is: users on desktop wont be on the view that requires clicking on the backdrop to close the sidebar. That action is typically for someone on a mobile device. For such a user, a key press does not seem practical . That's why in this instance I disagree with Sonar and left the behavior as it was by adding  NoSonar.

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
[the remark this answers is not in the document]
Benmuiruri: Please elaborate what you mean by it's unreachable. 

My understanding is we use the `openSidebar` method in `analytics-filter.component.ts` to open and close the sidebar filter by clicking the Icon. 
Then in `analytics-target-aggregates-sidebar-filter.component.ts` we have the `toggleSidebarFilter` to close the sidebar filter when the X is clicked.

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.html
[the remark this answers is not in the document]
latin-panda: > My rationale is: users on desktop will not typically be on the view that requires clicking on the backdrop to close the sidebar.

Some administrative roles might be using target aggregates in computers, like nurses, or pharmacists, and managers. Remember that CHT is an open-source framework, and there are many projects worldwide configuring and using it in different ways (even for stock monitoring)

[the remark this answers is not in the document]
Benmuiruri: Cool, I'll remove the comment and address the Sonar warning.

webapp/tests/karma/ts/components/filters/analytics-filter.component.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+  }));
+
+  it('should not display filter button if module is not target aggregates', fakeAsync(() => {
+    route.snapshot.firstChild.data.moduleId = 'not-target-aggregates';
+
+    component.ngOnInit();
+    flush();
+
+    expect(component.showFilterButton).to.be.false;
+  }));
```
latin-panda, line 135: Please add a unit test asserting that `openSidebar() {` calls the global actions and telemetry with the right values and the right amount of time (call count of each function), when opening and closing the filter.

webapp/tests/karma/ts/modules/analytics/analytics-target-aggregates-sidebar-filter-component.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+  it('should toggle sidebar filter', () => {
+    component.toggleSidebarFilter();
+    component.toggleSidebarFilter();
+    component.toggleSidebarFilter();
+
+    expect(globalActions.setSidebarFilter.calledThrice).to.be.true;
+    expect(globalActions.setSidebarFilter.args[0][0]).to.deep.equal({ isOpen: true });
+    expect(globalActions.setSidebarFilter.args[1][0]).to.deep.equal({ isOpen: false });
+    expect(globalActions.setSidebarFilter.args[2][0]).to.deep.equal({ isOpen: true });
+    expect(telemetryService.record.calledTwice).to.be.true;
```
latin-panda, line 93: this one should fail because telemetry isn't called anymore in that component

webapp/tests/karma/ts/modules/analytics/analytics-target-aggregates-sidebar-filter-component.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+import { TelemetryService } from '@mm-services/telemetry.service';
+import { UserSettingsService } from '@mm-services/user-settings.service';
+import { GlobalActions } from '@mm-actions/global';
+
+describe('Analytics Target Aggregate Sidebar Filter Component', () => {
+  let component: AnalyticsTargetAggregatesSidebarFilterComponent;
+  let fixture: ComponentFixture<AnalyticsTargetAggregatesSidebarFilterComponent>;
+  let contactTypesService;
+  let settingsService;
+  let telemetryService;
```
latin-panda, line 22: Remove telemetry from here too

webapp/tests/karma/ts/modules/analytics/analytics-target-aggregates-sidebar-filter-component.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    expect(globalActions.setSidebarFilter.args[0][0]).to.deep.equal({ isOpen: true });
+    expect(globalActions.setSidebarFilter.args[1][0]).to.deep.equal({ isOpen: false });
+    expect(globalActions.setSidebarFilter.args[2][0]).to.deep.equal({ isOpen: true });
+    expect(telemetryService.record.calledTwice).to.be.true;
+  });
+
+  it('should set userFacilities and selectedFacilityId when user has multiple facilities', fakeAsync(() => {
+    const userFacilities = [{ _id: 'facility_1' }, { _id: 'facility_2' }];
+    userSettingsService.getUserFacility.resolves(userFacilities);
+    userSettingsService.hasMultipleFacilities.resolves(true);
```
latin-panda, line 99: The test data is set here, but the unit of code isn't invoked, so this test data has no effect. 
The reason it's working is because it's set in L36. The correct place to set the test data is here, inside the `it(` function.

latin-panda: Please add one test case, where there is only 1 facility in `userFacilities`, with different data from this test.

webapp/tests/karma/ts/modules/analytics/analytics-target-aggregates-sidebar-filter-component.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    };
+    settingsService.get.resolves(settings);
+    contactTypesService.getTypeId.returns('district_hospital');
+    component.userFacilities = [{ _id: 'facility_1', type: 'district_hospital' }];
+
+    component.ngOnInit();
+    flush();
+
+    expect(component.facilityFilterLabel).to.equal(DEFAULT_FACILITY_LABEL);
+    expect(component.error).to.be.true;
```
latin-panda, line 142: Check the expected functions are called with the correct amount of time

webapp/tests/karma/ts/modules/analytics/analytics-target-aggregates-sidebar-filter-component.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+  }));
+
+  it('should set error and default facilityFilterLabel when settings service fails', fakeAsync(() => {
+    const DEFAULT_FACILITY_LABEL = 'Facility';
+    settingsService.get.rejects({ some: 'err' });
+
+    component.ngOnInit();
+    flush();
+
+    expect(component.facilityFilterLabel).to.equal(DEFAULT_FACILITY_LABEL);
```
latin-panda, line 123: Same here, assert functions are called with the correct amount of time (call count)

webapp/tests/karma/ts/modules/analytics/analytics-target-aggregates-sidebar-filter-component.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    expect(component.userFacilities).to.deep.equal(userFacilities);
+    expect(component.selectedFacilityId).to.equal('facility_1');
+    expect(userSettingsService.getUserFacility.calledOnce).to.be.true;
+    expect(userSettingsService.hasMultipleFacilities.calledOnce).to.be.true;
+  }));
+
+  it('should set user facility name_key as facilityFilterLabel', fakeAsync(() => {
+    contactTypesService.getTypeId.returns('district_hospital');
+    component.userFacilities = [{ _id: 'facility_1', type: 'district_hospital' }];
+
```
latin-panda, line 110: Same here, call the unit, try with different set of data, like 3 facilities, different ids. The more variation of data the better.

webapp/tests/karma/ts/services/user-settings.service.spec.ts
```diff
@@ -74,6 +79,40 @@ describe('UserSettings service', () => {
     });
   });
 
+  it('hasMultipleFacilities returns true when user has multiple facilities', async () => {
+    userCtx.returns({ name: 'jack' });
+    get.returns(Promise.resolve({ facility_id: ['facility_1', 'facility_2']}));
+    hasMultipleFacilities = sinon.spy(service, 'hasMultipleFacilities');
+    const result = await service.hasMultipleFacilities();
+    expect(result).to.be.true;
```
latin-panda, line 87: Extra understanding by separating the test data definition, the unit execution and the asserts. Same with the ones below
```suggestion
    hasMultipleFacilities = sinon.spy(service, 'hasMultipleFacilities');
    
    const result = await service.hasMultipleFacilities();
    
    expect(result).to.be.true;
```

webapp/tests/karma/ts/services/user-settings.service.spec.ts
```diff
@@ -74,6 +79,40 @@ describe('UserSettings service', () => {
...
+    expect(get.callCount).to.equal(1);
+    expect(hasMultipleFacilities.callCount).to.equal(1);
+    expect(get.args[0][0]).to.equal('org.couchdb.user:jack');
+  });
+
+  it('hasMultipleFacilities returns false when user has one facility', async () => {
+    userCtx.returns({ name: 'jack' });
+    get.returns(Promise.resolve({ facility_id: ['facility_1'] }));
```
latin-panda, line 95: Test case when it's undefined

webapp/tests/karma/ts/services/user-settings.service.spec.ts
```diff
@@ -74,6 +79,40 @@ describe('UserSettings service', () => {
...
+    hasMultipleFacilities = sinon.spy(service, 'hasMultipleFacilities');
+    const result = await service.hasMultipleFacilities();
+    expect(result).to.be.false;
+    expect(get.callCount).to.equal(1);
+    expect(hasMultipleFacilities.callCount).to.equal(1);
+    expect(get.args[0][0]).to.equal('org.couchdb.user:jack');
+  });
+
+  it('getUserFacility returns user facility from GetDataRecordService', async () => {
```
latin-panda, line 104: case: should not break if GetDataRecordService throws error

webapp/tests/karma/ts/modules/analytics/analytics-target-aggregates-sidebar-filter-component.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    settingsService.get.resolves(settings);
+    contactTypesService.getTypeId.returns('district_hospital');
+    component.userFacilities = [{ _id: 'facility_1', type: 'district_hospital' }];
+
+    component.ngOnInit();
+    flush();
+
+    expect(component.facilityFilterLabel).to.equal(DEFAULT_FACILITY_LABEL);
+    expect(component.error).to.be.true;
+  }));
```
latin-panda, line 143: Another test case, should not break when any of the other services throws error

webapp/tests/karma/ts/modules/analytics/analytics-target-aggregates-sidebar-filter-component.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    expect(component.userFacilities).to.deep.equal(userFacilities);
+    expect(component.selectedFacilityId).to.equal('facility_1');
+    expect(userSettingsService.getUserFacility.calledOnce).to.be.true;
+    expect(userSettingsService.hasMultipleFacilities.calledOnce).to.be.true;
+  }));
+
+  it('should set user facility name_key as facilityFilterLabel', fakeAsync(() => {
+    contactTypesService.getTypeId.returns('district_hospital');
+    component.userFacilities = [{ _id: 'facility_1', type: 'district_hospital' }];
```
latin-panda, line 109: case: should not break if user facilities is undefined - like contacts with no homeplaces

webapp/src/css/inbox.less
```diff
@@ -1151,13 +1172,12 @@ a.fa:hover {
 .analytics {
   .content {
     .inner {
-      /* allow the scrollable content to grow to full width */
```
latin-panda, line 1154: Did you test the scroll in these places?
- the aggregate list 
- the aggregates on the right side (the content)the cards below the main section 
Both work fine when the filter is closed and opened.
Also, please check the scroll inside the sidebar filter. 
Devices: Desktop, mobile, tablet.

latin-panda (changes requested):
Very nice!! Almost there, just small feedback

webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.ts
```diff
@@ -28,11 +66,62 @@ export class AnalyticsFilterComponent implements AfterContentInit, AfterContentC
...
+    ]);
+
+    this.showFilterButton = !isAdmin && hasMultipleFacilities && checkPermissions && this.isTargetAggregates();
+  }
+
+  openSidebar() {
+    this.globalActions.setSidebarFilter({ isOpen: !this.isOpen });
+    if (this.isOpen) {
+      // Counting every time the user opens the sidebar filter in analytics_targets_aggregrate tab.
+      this.telemetryService.record('sidebar_filter:analytics_target_aggregates:open');
```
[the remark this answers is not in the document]
Benmuiruri: I've opened a doc [issue](https://github.com/medic/cht-docs/issues/1453) I will work on immediately after this PR is merged to the main PR (Not merging to Master)

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
```diff
@@ -0,0 +1,88 @@
...
+    this.globalActions.setSidebarFilter({ isOpen: this.isOpen });
+  }
+
+  private async loadUserFacility() {
+    const hasMultipleFacilities = await this.userSettingsService.hasMultipleFacilities();
+    this.userFacilities = await this.userSettingsService.getUserFacility();
+
+    if (hasMultipleFacilities) {
+      this.selectedFacility = this.userFacilities[0];
+      this.selectedFacilityId = this.userFacilities[0]._id;
```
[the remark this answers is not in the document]
Benmuiruri: I was trying to avoid passing the whole facility to the HTML. But I could bind the entire facility object to the radio button instead of just the ID.

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  private async setFacilityLabel() {
+    const FACILITY = 'Facility';
+    try {
+      const settings = await this.settingsService.get();
+      const userFacilityType = this.contactTypesService.getTypeId(this.selectedFacility);
+      const placeType = settings.contact_types.find(type => type.id === userFacilityType);
+
+      this.facilityFilterLabel = placeType.name_key || FACILITY;
```
latin-panda, line 82: I get this error when running the app as an admin user - please add a unit test to cover when the user has no home place assigned

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
```diff
@@ -0,0 +1,88 @@
...
+import { Selectors } from '@mm-selectors/index';
+import { ContactTypesService } from '@mm-services/contact-types.service';
+import { SettingsService } from '@mm-services/settings.service';
+import { UserSettingsService } from '@mm-services/user-settings.service';
+
+@Component({
+  selector: 'mm-analytics-target-aggregates-sidebar-filter',
+  templateUrl: './analytics-target-aggregates-sidebar-filter.component.html'
+})
+export class AnalyticsTargetAggregatesSidebarFilterComponent implements OnInit, OnDestroy {
```
latin-panda, line 15: I found this error. 
Offline user gets an upgrade app popup and the app reloads while in targets aggregate page. So for example:
1. This user didn't have the targets aggregate enabled, so it displays `target aggregates are disabled`. 
2. Then admin enables, 
3. and the while the offline user is in the page with `target aggregates are disabled`, the user gets an upgrade popup, 
4. app reloads, and see error in browser console, and filer doesn't show. 
Screenshot 2024-07-16 at 12 16 34 PM
Screenshot 2024-07-16 at 12 20 37 PM

The filter should display with no errors, because the user already has the permission to access it

latin-panda: Wait, this user has only 1 facility assigned. So it should not show the filter, but it should not throw an error. 
So it still needs a fix.

webapp/src/css/sidebar-filter.less
```diff
@@ -99,6 +99,68 @@
       overflow-y: auto;
       overflow-x: hidden;
 
+      .filter-options-container {
```
latin-panda, line 102: Screenshot 2024-07-16 at 12 44 07 PM
Long facility names is making the radio button an oval, it should keep as circle

latin-panda: Also in desktop
Screenshot 2024-07-16 at 12 46 40 PM

webapp/src/css/sidebar-filter.less
```diff
@@ -99,6 +99,68 @@
       overflow-y: auto;
       overflow-x: hidden;
```
latin-panda, line 100: You can decrease the margin a bit to give more space to the content in small screens, that's more comfortable for the users. See reports:
Screenshot 2024-07-16 at 12 49 11 PM

Screenshot 2024-07-16 at 12 50 17 PM

latin-panda (changes requested):

latin-panda:
Feedback above ⬆️ 
This is what I tested ⬇️ 

<details><summary>Admin user (with 1 facility assigned and no facility assigned)</summary>

- Device
   - Mobile
   - Desktop
   - Tablet
- Back button
- Reload
- Navigate target
- Navigate target aggregates
- Check the browser's console for errors or suspicious warnings

</details>

<details><summary>Offline user (with 1 facility and 2 facilities))</summary>
- Device
   - Mobile
   - Desktop
   - Tablet
- Back button
- Reload
- With no aggregate permissions
   - Navigate target
   - Navigate target aggregates by changing url - should not allow and show `target aggregates are disabled`

- Enable permissions and wait for the upgrade message should work; no errors 
   - Navigate target
   - Navigate target aggregates
   - Aggregate targets details view - like click on one from the list. 
- With old filter and old action bar

</details>

<details><summary>Online user (with 1 facility and 2 facilities))</summary>
- Device
   - Mobile
   - Desktop
   - Tablet
- Back button
- Reload
- With no aggregate permissions
   - Navigate target
   - Navigate target aggregates by changing url - should not allow and show `target aggregates are disabled`

- Enable permissions and wait for the upgrade message should work; no errors 
   - Navigate target
   - Navigate target aggregates
   - Aggregate targets details view - like click on one from the list. 
- With old filter and old action bar

</details>

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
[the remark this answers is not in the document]
latin-panda: can you try this: `([ngModel])="this.selectedFacility._id"` ? Does it work?

webapp/src/css/sidebar-filter.less
Benmuiruri: This is why I was using the `!important` what's another way I can ensure it is always a circle without using important ?

latin-panda: Make the selector more specific

latin-panda: this fixes it: `min-width: 20px;` in the input

webapp/tests/karma/ts/modules/analytics/analytics-target-aggregates-sidebar-filter-component.spec.ts
Benmuiruri: I don't fully understand this test case, could you please be more specific.

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
Benmuiruri: Good catch. 


https://github.com/user-attachments/assets/f244060d-d775-4c0d-958e-c79a9ffece47

I have fixed it. Let me know if the video captures the fix

webapp/tests/karma/ts/modules/analytics/analytics-target-aggregates-sidebar-filter-component.spec.ts
latin-panda: a unit test to assert that it doesn't break when any of these throws errors 
settingsService.get
    contactTypesService.getTypeId

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
latin-panda: Yeah, but also, if the user has only 1 user assigned, it shouldn't throw that error after reloading the app on upgrade.

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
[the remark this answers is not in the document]
Benmuiruri: No that does not work.

latin-panda (approved):
I reviewed the changes, but couldn't test since I'm running out of time for this week. 
LGTM! Thanks for fixing all the feedback

Issue #9237, opened 2024-07-03
Add functionality of getting people with pagination in cht-datasource


sugat009:
**Is your feature request related to a problem? Please describe.**

This feature is part of the bigger feature of the creation of API endpoints for getting contacts by type. (#9193)
The target of this feature is to have the functionality of getting people using `cht-datasource` with the option of pagination.

**Describe the solution you'd like**

```javascript
Person.v1.getAll(limit, skip);
```
As mentioned in the parent ticket:
- `limit` - default value is `100`
- `skip` - default value is `0`

**Describe alternatives you've considered**

**Additional context**

jkuester:
@sugat009 Thinking some more about API naming here (along with the [iterator function](https://github.com/medic/cht-core/issues/9238), I propose these function signatures:

```ts
const getPage = (limit = 100, skip = 0) => Promise<Person.v1.Person[]>

const getAll = () => AsyncGenerator<Person.v1.Person, void>
```

Under the hood, `getAll` can call `getPage`.  The main thing here is that I would like to encourage use of the `getAll` instead of `getPage`, so we do not have consumers trying to manually page through results unless it is necessary.  For this issue, I think all this means is that you would need to change your function name from `getAll` to `getPage`.  Let me know if this seems reasonable. Happy to discuss more!

sugat009:
@jkuester Of course, that sounds good.

jkuester:
I am removing this from the 4.11 milestone since it is covered already by https://github.com/medic/cht-core/issues/9241.

PR #9266, opened 2024-07-16
feat(#9237): add functionality of getting people with pagination in cht-datasource

sugat009:
# Description

Add functionality of getting people with pagination in `cht-datasource`.

Issue: #9237 

Usage:

```typescript
Person.v1.getPage(limit, skip); // limit and skip are optional with defaults as 100 and 0 respectively
```

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9237-add-functionality-of-getting-people-with-pagination-in-cht-datasource/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9237-add-functionality-of-getting-people-with-pagination-in-cht-datasource/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9237-add-functionality-of-getting-people-with-pagination-in-cht-datasource/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/shared-libs/cht-datasource/src/index.ts
+++ b/shared-libs/cht-datasource/src/index.ts
@@ -85,6 +85,14 @@ export const getDatasource = (ctx: DataContext) => {
          * @throws Error if no UUID is provided
          */
         getByUuidWithLineage: (uuid: string) => ctx.bind(Person.v1.getWithLineage)(Qualifier.byUuid(uuid)),
+
+        /**
+         * Returns a list of people.
+         * @param limit the total number of records to retrieve
+         * @param skip the total number of records to skip
+         * @returns array of `Person`
+         */
+        getPage: (limit = 100, skip = 0) => ctx.bind(Person.v1.getPage)(limit, skip),
       }
     }
   };
--- a/shared-libs/cht-datasource/src/index.ts
+++ b/shared-libs/cht-datasource/src/index.ts
@@ -96,7 +96,7 @@ export const getDatasource = (ctx: DataContext) => {
          * @throws Error if the provided limit is `<= 0`
          * @throws Error if the provided skip is `< 0`
          */
-        getPage: (personType: string, limit = 100, skip = 0) => ctx.bind(Person.v1.getPage)(
+        getPageByType: (personType: string, limit = 100, skip = 0) => ctx.bind(Person.v1.getPage)(
           Qualifier.byContactType(personType), limit, skip
         ),
       }
--- a/shared-libs/cht-datasource/src/index.ts
+++ b/shared-libs/cht-datasource/src/index.ts
@@ -88,11 +88,14 @@ export const getDatasource = (ctx: DataContext) => {
 
         /**
          * Returns a list of people.
+         * @param personType the string that represents the person type
          * @param limit the total number of records to retrieve
          * @param skip the total number of records to skip
          * @returns array of `Person`
          */
-        getPage: (limit = 100, skip = 0) => ctx.bind(Person.v1.getPage)(limit, skip),
+        getPage: (personType: string, limit = 100, skip = 0) => ctx.bind(Person.v1.getPage)(
+          Qualifier.byContactType(personType), limit, skip
+        ),
       }
     }
   };
--- a/shared-libs/cht-datasource/src/index.ts
+++ b/shared-libs/cht-datasource/src/index.ts
@@ -86,16 +86,16 @@ export const getDatasource = (ctx: DataContext) => {
          */
         getByUuidWithLineage: (uuid: string) => ctx.bind(Person.v1.getWithLineage)(Qualifier.byUuid(uuid)),
 
-    /**
-     * Returns an array of persons for the provided page specifications.
-     * @param personType the type of persons to return
-     * @param limit the maximum number of persons to return. Default is 100.
-     * @param skip the number of persons to skip. Default is 0.
-     * @returns an array of persons for the provided page specifications
-     * @throws Error if no type is provided or if the type is not for a person
-     * @throws Error if the provided limit is `<= 0`
-     * @throws Error if the provided skip is `< 0`
-     */
+        /**
+         * Returns an array of people for the provided page specifications.
+         * @param personType the type of people to return
+         * @param limit the maximum number of people to return. Default is 100.
+         * @param skip the number of people to skip. Default is 0.
+         * @returns an array of people for the provided page specifications
+         * @throws Error if no type is provided or if the type is not for a person
+         * @throws Error if the provided limit is `<= 0`
+         * @throws Error if the provided skip is `< 0`
+         */
         getPage: (personType: string, limit = 100, skip = 0) => ctx.bind(Person.v1.getPage)(
           Qualifier.byContactType(personType), limit, skip
         ),
--- a/shared-libs/cht-datasource/src/index.ts
+++ b/shared-libs/cht-datasource/src/index.ts
@@ -86,13 +86,16 @@ export const getDatasource = (ctx: DataContext) => {
          */
         getByUuidWithLineage: (uuid: string) => ctx.bind(Person.v1.getWithLineage)(Qualifier.byUuid(uuid)),
 
-        /**
-         * Returns a list of people.
-         * @param personType the string that represents the person type
-         * @param limit the total number of records to retrieve
-         * @param skip the total number of records to skip
-         * @returns array of `Person`
-         */
+    /**
+     * Returns an array of persons for the provided page specifications.
+     * @param personType the type of persons to return
+     * @param limit the maximum number of persons to return. Default is 100.
+     * @param skip the number of persons to skip. Default is 0.
+     * @returns an array of persons for the provided page specifications
+     * @throws Error if no type is provided or if the type is not for a person
+     * @throws Error if the provided limit is `<= 0`
+     * @throws Error if the provided skip is `< 0`
+     */
         getPage: (personType: string, limit = 100, skip = 0) => ctx.bind(Person.v1.getPage)(
           Qualifier.byContactType(personType), limit, skip
         ),
--- a/shared-libs/cht-datasource/src/local/libs/lineage.ts
+++ b/shared-libs/cht-datasource/src/local/libs/lineage.ts
@@ -18,11 +18,7 @@ import logger from '@medic/logger';
  * sorted such that the identified document is the first element and the parent documents are in order of lineage.
  * @internal
  */
-export const getLineageDocsById = (
-  medicDb: PouchDB.Database<Doc>
-): (
-  id: string
-) => Promise<Nullable<Doc>[]> => {
+export const getLineageDocsById = (medicDb: PouchDB.Database<Doc>): (id: string) => Promise<Nullable<Doc>[]> => {
   const fn = queryDocsByRange(medicDb, 'medic-client/docs_by_id_lineage');
   return (id: string) => fn([id], [id, {}]);
 };
--- a/shared-libs/cht-datasource/src/local/libs/lineage.ts
+++ b/shared-libs/cht-datasource/src/local/libs/lineage.ts
@@ -10,7 +10,7 @@ import {
   Nullable
 } from '../../libs/core';
 import { Doc } from '../../libs/doc';
-import { queryDocsByKey } from './doc';
+import { queryDocsByRange } from './doc';
 import logger from '@medic/logger';
 
 /**
@@ -20,7 +20,10 @@ import logger from '@medic/logger';
  */
 export const getLineageDocsById = (
   medicDb: PouchDB.Database<Doc>
-): (id: string) => Promise<Nullable<Doc>[]> => queryDocsByKey(medicDb, 'medic-client/docs_by_id_lineage');
+): (
+  startkey: unknown,
+  endkey: unknown
+) => Promise<Nullable<Doc>[]> => queryDocsByRange(medicDb, 'medic-client/docs_by_id_lineage');
 
 /** @internal */
 export const getPrimaryContactIds = (places: NonEmptyArray<Nullable<Doc>>): string[] => places
--- a/shared-libs/cht-datasource/src/local/libs/lineage.ts
+++ b/shared-libs/cht-datasource/src/local/libs/lineage.ts
@@ -21,9 +21,11 @@ import logger from '@medic/logger';
 export const getLineageDocsById = (
   medicDb: PouchDB.Database<Doc>
 ): (
-  startkey: unknown,
-  endkey: unknown
-) => Promise<Nullable<Doc>[]> => queryDocsByRange(medicDb, 'medic-client/docs_by_id_lineage');
+  id: string
+) => Promise<Nullable<Doc>[]> => {
+  const fn = queryDocsByRange(medicDb, 'medic-client/docs_by_id_lineage');
+  return (id: string) => fn([id], [id, {}]);
+};
 
 /** @internal */
 export const getPrimaryContactIds = (places: NonEmptyArray<Nullable<Doc>>): string[] => places
--- a/shared-libs/cht-datasource/src/local/person.ts
+++ b/shared-libs/cht-datasource/src/local/person.ts
@@ -65,11 +65,11 @@ export namespace v1 {
       const personTypes = contactTypeUtils.getPersonTypes(settings.getAll());
       const personTypesIds = personTypes.map(item => item.id);
 
-      const getDocsByPage = queryDocsByKey(medicDb, 'medic-client/contacts_by_type');
-
       if (!personTypesIds.includes(personType.contactType)) {
         throw new Error(`Invalid person type: ${personType.contactType}`);
       }
+
+      const getDocsByPage = queryDocsByKey(medicDb, 'medic-client/contacts_by_type');
 
       const docs = await getDocsByPage([personType.contactType], limit, skip);
 
--- a/shared-libs/cht-datasource/src/local/person.ts
+++ b/shared-libs/cht-datasource/src/local/person.ts
@@ -1,6 +1,6 @@
 import { Doc } from '../libs/doc';
 import contactTypeUtils from '@medic/contact-types-utils';
-import { deepCopy, isNonEmptyArray, Nullable } from '../libs/core';
+import { deepCopy, isNonEmptyArray, Nullable, Page } from '../libs/core';
 import { ContactTypeQualifier, UuidQualifier } from '../qualifier';
 import * as Person from '../person';
 import { getDocById, getDocsByIds, queryDocsByKey } from './libs/doc';
@@ -61,9 +61,13 @@ export namespace v1 {
 
   /** @internal */
   export const getPage = ({ medicDb, settings }: LocalDataContext) => {
-    return async (personType: ContactTypeQualifier, limit: number, skip: number): Promise<Person.v1.Person[]> => {
+    return async (
+      personType: ContactTypeQualifier,
+      limit: number,
+      skip: number
+    ): Promise<Page<Person.v1.Person>> => {
       const personTypes = contactTypeUtils.getPersonTypes(settings.getAll());
-      const personTypesIds = personTypes.map(item => item.id);
+      const personTypesIds = personTypes.map((item) => item.id);
 
       if (!personTypesIds.includes(personType.contactType)) {
         throw new Error(`Invalid person type: ${personType.contactType}`);
@@ -71,9 +75,41 @@ export namespace v1 {
 
       const getDocsByPage = queryDocsByKey(medicDb, 'medic-client/contacts_by_type');
 
-      const docs = await getDocsByPage([personType.contactType], limit, skip);
+      const fetchAndFilter = async (
+        currentLimit: number,
+        currentSkip: number,
+        personDocs: Person.v1.Person[],
+        totalDocsFetched = 0,
+      ): Promise<Page<Person.v1.Person>> => {
+        const docs = await getDocsByPage([personType.contactType], currentLimit, currentSkip);
+        if (docs.length === 0) {
+          return { data: personDocs, cursor: '-1' };
+        }
 
-      return docs.filter((doc): doc is Person.v1.Person => isPerson(settings, doc, doc?._id));
+        const tempFilteredDocs = docs.filter((doc): doc is Person.v1.Person => isPerson(settings, doc, doc?._id));
+
+        personDocs.push(...tempFilteredDocs);
+        totalDocsFetched += docs.length;
+
+        if (personDocs.length >= limit) {
+          let cursor: number;
+          if (docs.length < currentLimit) {
+            cursor = -1;
+          } else {
+            cursor = skip + totalDocsFetched - (personDocs.length - limit);
+          }
+          return { data: personDocs.slice(0, limit), cursor: cursor.toString() };
+        }
+
+        return fetchAndFilter(
+          (currentLimit - tempFilteredDocs.length) * 2,
+          currentSkip + currentLimit,
+          personDocs,
+          totalDocsFetched
+        );
+      };
+
+      return fetchAndFilter(limit, skip, []);
     };
   };
 }
--- a/shared-libs/cht-datasource/src/local/person.ts
+++ b/shared-libs/cht-datasource/src/local/person.ts
@@ -10,7 +10,7 @@ import { getLineageDocsById, getPrimaryContactIds, hydrateLineage, hydratePrimar
 
 /** @internal */
 export namespace v1 {
-  const isPerson = (settings: SettingsService, uuid: string, doc: Nullable<Doc>): doc is Person.v1.Person => {
+  const isPerson = (settings: SettingsService, doc: Nullable<Doc>, uuid = ''): doc is Person.v1.Person => {
     if (!doc) {
       logger.warn(`No person found for identifier [${uuid}].`);
       return false;
@@ -28,7 +28,7 @@ export namespace v1 {
     const getMedicDocById = getDocById(medicDb);
     return async (identifier: UuidQualifier): Promise<Nullable<Person.v1.Person>> => {
       const doc = await getMedicDocById(identifier.uuid);
-      if (!isPerson(settings, identifier.uuid, doc)) {
+      if (!isPerson(settings, doc, identifier.uuid)) {
         return null;
       }
       return doc;
@@ -41,7 +41,7 @@ export namespace v1 {
     const getMedicDocsById = getDocsByIds(medicDb);
     return async (identifier: UuidQualifier): Promise<Nullable<Person.v1.PersonWithLineage>> => {
       const [person, ...lineagePlaces] = await getLineageDocs(identifier.uuid);
-      if (!isPerson(settings, identifier.uuid, person)) {
+      if (!isPerson(settings, person, identifier.uuid)) {
         return null;
       }
       // Intentionally not further validating lineage. For passivity, lineage problems should not block retrieval.
@@ -61,19 +61,19 @@ export namespace v1 {
 
   /** @internal */
   export const getPage = ({ medicDb, settings }: LocalDataContext) => {
-    const personTypes = contactTypeUtils.getPersonTypes(settings.getAll());
-    const personTypesIds = personTypes.map(item => item.id);
-
-    const getDocsByPage = queryDocsByKey(medicDb, 'medic-client/contacts_by_type');
-
     return async (personType: ContactTypeQualifier, limit: number, skip: number): Promise<Person.v1.Person[]> => {
+      const personTypes = contactTypeUtils.getPersonTypes(settings.getAll());
+      const personTypesIds = personTypes.map(item => item.id);
+
+      const getDocsByPage = queryDocsByKey(medicDb, 'medic-client/contacts_by_type');
+
       if (!personTypesIds.includes(personType.contactType)) {
         throw new Error(`Invalid person type: ${personType.contactType}`);
       }
 
       const docs = await getDocsByPage([personType.contactType], limit, skip);
 
-      return docs.filter((doc): doc is Person.v1.Person => isPerson(settings, doc?._id ?? '', doc));
+      return docs.filter((doc): doc is Person.v1.Person => isPerson(settings, doc, doc?._id));
     };
   };
 }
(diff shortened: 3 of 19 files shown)


jkuester (changes requested):
Sorry, I was not able to make it through everything today (just ran out of time), but want to share my thoughts so far.  Will continue going through this tomorrow!

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -58,4 +58,22 @@ export namespace v1 {
...
+  /** @internal */
+  export const getPage = ({ medicDb, settings }: LocalDataContext) => {
+    const personIdentifierRecord = contactTypeUtils.getPersonTypes(settings.getAll());
+
+    let personIdentifier: string;
+    if (Object.entries(personIdentifierRecord).length > 0) {
+      personIdentifier = personIdentifierRecord[0]?.id as string;
+    } else {
+      throw new Error('Person type not found');
+    }
```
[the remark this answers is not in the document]
sugat009: additionally, let's have the parameter be named `personType` instead because `type` is a reserved word in typescript

shared-libs/cht-datasource/src/person.ts
```diff
@@ -44,6 +44,18 @@ export namespace v1 {
...
     };
 
+  const getPeople = <T>(
+    localFn: (c: LocalDataContext) => (limit: number, skip: number) => Promise<T>,
+    remoteFn: (c: RemoteDataContext) => (limit: number, skip: number) => Promise<T>
+  ) => (context: DataContext) => {
+      assertDataContext(context);
+      const fn = adapt(context, localFn, remoteFn);
+
+      return async (limit = 100, skip = 0): Promise<T> => {
```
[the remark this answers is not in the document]
sugat009: sure, let's do that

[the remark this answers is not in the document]
jkuester: As you mentioned above, `type` is a keyword, so maybe `contactType` would be a better property name.

shared-libs/cht-datasource/src/remote/person.ts
```diff
@@ -19,4 +19,13 @@ export namespace v1 {
     identifier.uuid,
     { with_lineage: 'true' }
   );
+
+  /** @internal */
+  export const getPage = (remoteContext: RemoteDataContext) => (
+    limit: number,
+    skip: number
+  ): Promise<null> => getPerson(remoteContext)(
```
jkuester, line 27: For this function, I don't think we want to reuse the `getPerson`/`getResource` code.  I guess there are some natural differences between getting a specific entity vs getting a list of entities and the most straightforward way to handle this seems like it would be by adding a new function:

```ts
/** @internal */
export const getResources = (context: RemoteDataContext, path: string) => async <T>(
queryParams?: Record<string, string>
): Promise<T> => {
const params = new URLSearchParams(queryParams).toString();
try {
const response = await fetch(`${context.url}/${path}?${params}`);
if (!response.ok) {
throw new Error(response.statusText);
}

    return (await response.json()) as T;
} catch (error) {
logger.error(`Failed to fetch resources from ${context.url}/${path} with params: ${params}`, error);
throw error;
}
};
```

The most important difference is that there is no special handling for a `404`.

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -58,4 +58,22 @@ export namespace v1 {
...
+
+    const getDocsByPage = queryDocsByKey(medicDb, 'medic-client/contacts_by_type');
+
+    return async (limit: number, skip: number): Promise<Nullable<Doc>[]> => {
```
jkuester, line 75: The array returned by `queryDocsByKey` might contain null values and this is fine since sometimes those may have semantic meaning in the context of the query.  However, for this case, we can eliminate any null values (and anything that is not actually a person) by filtering the returned values via `isPerson`.  The return value for this function should actually just be `Promise<Person.v1.Person[]>`.

jkuester (changes requested):

shared-libs/cht-datasource/src/person.ts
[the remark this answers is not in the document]
sugat009: Now that I think about this, maybe the type could also be named `ContactTypeQualifier` since we're explicitly dealing with `Contact Types` with this qualifier.

shared-libs/cht-datasource/src/person.ts
```diff
@@ -44,6 +44,18 @@ export namespace v1 {
...
+        return fn(limit, skip);
+      };
```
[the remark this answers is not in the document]
sugat009: Cool, lemme try something as well.

shared-libs/cht-datasource/src/index.ts
```diff
@@ -85,6 +85,14 @@ export const getDatasource = (ctx: DataContext) => {
...
          */
         getByUuidWithLineage: (uuid: string) => ctx.bind(Person.v1.getWithLineage)(Qualifier.byUuid(uuid)),
+
+        /**
+         * Returns a list of people.
+         * @param limit the total number of records to retrieve
+         * @param skip the total number of records to skip
+         * @returns array of `Person`
+         */
+        getPage: (limit = 100, skip = 0) => ctx.bind(Person.v1.getPage)(limit, skip),
```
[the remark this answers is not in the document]
sugat009: I think we should have both options and let the consumers choose based on what their preference or use-case is.

sugat009:
@jkuester I did not add unit tests for this function(https://github.com/medic/cht-core/blob/de63197bb8545bcd07492d8f30cf691f7b6ddf02/shared-libs/cht-datasource/src/remote/libs/data-context.ts#L63) as NodeJS 20+ has an issue in which mocking `fetch` is not possible. Also, existing tests that have mocked fetched are failing as well. For eg: https://github.com/medic/cht-core/blob/3c2b140a33fd16f87e11663da07037fbc827a3c9/shared-libs/cht-datasource/src/remote/libs/data-context.ts#L41
References: [here](https://github.com/sinonjs/sinon/issues/2590) and [here](https://github.com/nodejs/node/pull/51598#issuecomment-1984341350).

shared-libs/cht-datasource/src/person.ts
```diff
@@ -59,4 +80,12 @@ export namespace v1 {
    * @throws Error if the provided context or qualifier is invalid
    */
   export const getWithLineage = getPerson(Local.Person.v1.getWithLineage, Remote.Person.v1.getWithLineage);
+
+  /**
+   * Returns an array of people.
+   * @param context the current data context
+   * @returns an array of people
+   * @throws Error if the provided context or personType qualifier is invalid
+   */
```
jkuester, line 89: As I mentioned in the comment above, after seeing how the docs site gets generated for this file, I am thinking we should change our documentation pattern a bit here.  This function documentation can specifically indicate it returns _a function_ and then we can get the documentation for the returned function _to be pulled through_ and also displayed on the site.

```suggestion
  /**
   * Returns a function for retrieving a paged array of persons from the given data context.
   * @param context the current data context
   * @returns a function for retrieving a paged array of persons
   * @throws Error if a data context is not provided
   */
```

shared-libs/cht-datasource/src/index.ts
```diff
@@ -85,6 +85,17 @@ export const getDatasource = (ctx: DataContext) => {
...
          */
         getByUuidWithLineage: (uuid: string) => ctx.bind(Person.v1.getWithLineage)(Qualifier.byUuid(uuid)),
+
+        /**
+         * Returns a list of people.
+         * @param personType the string that represents the person type
+         * @param limit the total number of records to retrieve
+         * @param skip the total number of records to skip
+         * @returns array of `Person`
+         */
```
jkuester, line 95:
```suggestion
    /**
     * Returns an array of persons for the provided page specifications.
     * @param personType the type of persons to return
     * @param limit the maximum number of persons to return. Default is 100.
     * @param skip the number of persons to skip. Default is 0.
     * @returns an array of persons for the provided page specifications
     * @throws Error if no type is provided or if the type is not for a person
     * @throws Error if the provided limit is `<= 0`
     * @throws Error if the provided skip is `< 0`
     */
```

shared-libs/cht-datasource/test/index.spec.ts
```diff
@@ -124,6 +125,24 @@ describe('CHT Script API - getDatasource', () => {
         expect(personGet.calledOnceWithExactly(qualifier)).to.be.true;
         expect(byUuid.calledOnceWithExactly(qualifier.uuid)).to.be.true;
       });
+
+      it('getPage', async () => {
+        const expectedPeople: Index.Nullable<Doc>[] = [];
```
jkuester, line 130: Minor, the type here can be `Person.v1.Person[]` instead of `Index.Nullable<Doc>[]`.

jkuester:
> NodeJS 20+ has an issue in which mocking fetch is not possible. Also, existing tests that have mocked fetched are failing as well. For eg: 

:thinking: These existing tests are working fine for me with node `20.11.0`  (and are running in CI with that same version).  What version of Node are you using?

jkuester (changes requested):
Looking great!  Got a couple other nit-picks and suggestions, but all in all we are close to done here.

sugat009:
> > NodeJS 20+ has an issue in which mocking fetch is not possible. Also, existing tests that have mocked fetched are failing as well. For eg:
> 
> 🤔 These existing tests are working fine for me with node `20.11.0` (and are running in CI with that same version). What version of Node are you using?

I'm on Node version v20.12.2

shared-libs/cht-datasource/src/local/libs/lineage.ts
```diff
@@ -20,7 +20,12 @@ import logger from '@medic/logger';
  */
 export const getLineageDocsById = (
   medicDb: PouchDB.Database<Doc>
-): (id: string) => Promise<Nullable<Doc>[]> => queryDocsByKey(medicDb, 'medic-client/docs_by_id_lineage');
+): (
+  id: string
+) => Promise<Nullable<Doc>[]> => {
```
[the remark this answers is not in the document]
sugat009: sure. but in the long run, we need to configure a formatted like `prettier` for consistent formatting.

shared-libs/cht-datasource/test/local/person.spec.ts
```diff
@@ -224,5 +224,70 @@ describe('local person', () => {
...
+
+      it('returns a page of people', async () => {
+        const doc = { type: 'person'};
+        const docs = [doc, doc, doc];
+        queryDocsByKeyInner.resolves(docs);
+
+        const res = await Person.v1.getPage(localContext)(personTypeQualifier, limit, skip);
+
+        expect(res).to.deep.equal(docs);
+        expect(settingsGetAll.callCount === 4).to.be.true;
```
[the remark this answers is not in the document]
sugat009: this is very useful, thanks @jkuester

jkuester:
> I'm on Node version v20.12.2

:sweat: I can reproduce the error locally when I switched to `20.12.2`.  Going to do some more digging to see what the best response is here....

jkuester:
Actually, digging in more to the [issues you linked](https://github.com/sinonjs/sinon/issues/2590) it looks like this was caused by Node changes introduced in `20.12.0` and then subsequently fixed in `20.13.1`.

Since this issue is localized to a few minor Node versions and is fixed in the latest releases of `20.x` and `22.x`, I think it is fine to not worry about it and @sugat009 you just need to update your Node version to `20.13.1+`. :sweat_smile:   (I confirmed it works fine on `v20.15.1`.)

shared-libs/cht-datasource/src/person.ts
```diff
@@ -32,14 +32,23 @@ export namespace v1 {
...
     qualifier: unknown
-  ) => asserts qualifier is ContactTypeQualifier = (qualifier: unknown) => {
+  ) => {
     if (!isContactTypeQualifier(qualifier)) {
       throw new Error(`Invalid type [${JSON.stringify(qualifier)}].`);
     }
   };
 
+  const assertLimitAndSkip = (limit: unknown, skip: unknown) => {
+    if (typeof limit !== 'number' || limit <= 0) {
```
jkuester, line 44: Can you also include a `Number.isInteger` check? Just don't want to let anyone be passing decimal numbers here.

shared-libs/cht-datasource/src/person.ts
```diff
@@ -32,14 +32,23 @@ export namespace v1 {
...
+      throw new Error('limit must be a positive number');
```
jkuester, line 45: Minor, but lets also print out the invalid value.  Something like: 

```suggestion
      throw new Error(`The limit must be a positive integer: [${limit}].`);
```

shared-libs/cht-datasource/src/person.ts
```diff
@@ -32,14 +32,23 @@ export namespace v1 {
...
+  const assertTypeQualifier: (qualifier: unknown) => asserts qualifier is ContactTypeQualifier = (
     qualifier: unknown
-  ) => asserts qualifier is ContactTypeQualifier = (qualifier: unknown) => {
+  ) => {
     if (!isContactTypeQualifier(qualifier)) {
       throw new Error(`Invalid type [${JSON.stringify(qualifier)}].`);
     }
   };
 
+  const assertLimitAndSkip = (limit: unknown, skip: unknown) => {
```
jkuester, line 43: Minor, but for the sake of consistency, I think we should just break this into two assert functions.

shared-libs/cht-datasource/src/person.ts
```diff
@@ -59,10 +68,23 @@ export namespace v1 {
...
       const fn = adapt(context, localFn, remoteFn);
 
-      return async (personType: ContactTypeQualifier, limit = 100, skip = 0): Promise<T> => {
+      /**
+       * Returns an array of people for the provided page specifications.
+       * @param personType the type of people to return
+       * @param limit the maximum number of people to return. Default is 100.
+       * @param skip the number of people to skip. Default is 0.
+       * @returns an array of people for the provided page specifications.
+       * @throws Error if `personType` qualifier is invalid
```
jkuester, line 77:
```suggestion
       * @throws Error if no type is provided or if the type is not for a person
```

shared-libs/cht-datasource/test/local/libs/lineage.spec.ts
```diff
@@ -16,12 +16,12 @@ describe('local lineage lib', () => {
   beforeEach(() => {
     debug = sinon.stub(logger, 'debug');
   });
-
+  //
```
jkuester, line 19:
```suggestion

```

shared-libs/cht-datasource/test/local/libs/lineage.spec.ts
```diff
@@ -16,12 +16,12 @@ describe('local lineage lib', () => {
...
   afterEach(() => sinon.restore());
 
   it('getLineageDocsById', async () => {
     const uuid = '123';
-    const queryFn = sinon.stub().returns(Promise.resolve([]));
+    const queryFn = sinon.stub().returns([]);
```
jkuester, line 24:
```suggestion
    const queryFn = sinon.stub().resolves([]);
```

Minor, but this is technically more correct since the real version of this function will return a promise.  I guess the test will pass either way since the `await` on line 31 does not really care if you give it a function or not.

shared-libs/cht-datasource/test/local/person.spec.ts
```diff
@@ -287,6 +288,28 @@ describe('local person', () => {
           queryDocsByKeyOuter.calledOnceWithExactly(localContext.medicDb, 'medic-client/contacts_by_type')
         ).to.be.true;
         expect(queryDocsByKeyInner.calledOnceWithExactly([personIdentifier], limit, skip)).to.be.true;
+        expect(isPerson.notCalled).to.be.true;
+      });
+
+      it('returns null valued array if rows returned from database are not docs', async () => {
```
jkuester, line 294:
```suggestion
      it('returns empty array if rows returned from database are not persons', async () => {
```

shared-libs/cht-datasource/src/local/libs/lineage.ts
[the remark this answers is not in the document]
jkuester: 100% agree. Unfortunately, `prettier` (or anything else with similar functionality) is somewhat controversial here at Medic.  :sweat: Coming from a Java background, I am used to having no control over code formatting and just letting the Eclipse auto-formatter do its thing!  :magic_wand: 

_However_, others have a more profound appreciation for being able to manually improve the readability of some code by carefully formatting it for maximum clarity.

Currently, we have settled on something of a middle ground where we use _eslint_ to enforce quite a few formatting constraints, but we have not totally given control of things over to `prettier`....

jkuester (approved):
Looks like CI is failing for a simple linting error. Should be an easy fix. I added a few more nit-pick comments below, but otherwise LGTM!  :rocket:

shared-libs/cht-datasource/src/person.ts
```diff
@@ -44,6 +61,32 @@ export namespace v1 {
       };
     };
 
+  const getPeople = <T>(
```
jkuester, line 64: Just a random final thought here.  I cannot come up with a scenario in the near future where we will be re-using this function.  It might make sense to just combine this back in with `getPage`.  Will definitely simplify things a bit.

(Also, with how we had decided to go with this documentation, it really is going to be single-purpose...)

shared-libs/cht-datasource/src/index.ts
```diff
@@ -85,6 +85,20 @@ export const getDatasource = (ctx: DataContext) => {
...
+         * Returns an array of people for the provided page specifications.
+         * @param personType the type of people to return
+         * @param limit the maximum number of people to return. Default is 100.
+         * @param skip the number of people to skip. Default is 0.
+         * @returns an array of people for the provided page specifications
+         * @throws Error if no type is provided or if the type is not for a person
+         * @throws Error if the provided limit is `<= 0`
+         * @throws Error if the provided skip is `< 0`
+         */
+        getPage: (personType: string, limit = 100, skip = 0) => ctx.bind(Person.v1.getPage)(
```
jkuester, line 99: Another after-thought I had here was that we should probably name this function _getPageByType_. 

My vision is that in the future the `Person.v1.getPage` function could be expanded to accept different Qualifiers without changing the function name (e.g. you could get a page of persons by birthDate or something).  We should be able to keep things from getting too confusing there because of the structure provided by the qualifier objects/types.  

However, for these functions we are directly accepting the parameters (and don't want to do anything sketchy like `string | Date`).  So, we will end up with one function per type of parameter combination.  With that in mind, we should name things accordingly....

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -58,4 +58,22 @@ export namespace v1 {
...
+
+      if (!personTypesIds.includes(personType.contactType)) {
+        throw new Error(`Invalid person type: ${personType.contactType}`);
+      }
+
+      const getDocsByPage = queryDocsByKey(medicDb, 'medic-client/contacts_by_type');
+
+      const docs = await getDocsByPage([personType.contactType], limit, skip);
+
+      return docs.filter((doc): doc is Person.v1.Person => isPerson(settings, doc, doc?._id));
```
jkuester, line 76: @m5r @sugat009 I have realized a complication here while looking at the code over in https://github.com/medic/cht-core/pull/9281.  Basically, we pass the `limit` through to the Couch view query and so we know `docs.length === limit` _or we have reached the last page of results_.  However, before this function returns the persons, we are _filtering_ the array. In theory it is possible that some of the results returned from `queryDocsByKey` could get filtered out (with our current validation logic I cannot come up with a feasible scenario where this could happen, but if/when we add more validations, this will become an issue).  

The reason I think this filtering is a problem is because it could result in us giving back an array for the page that is `< limit` when there are still additional results to retrieve.  Of course, this is bad for our `getAll` logic, but it will also probably be confusing to any other consumers of this code. (How would anyone know when to stop calling `getPage`?).  

The only feasible option I can come up with here is to change the return type to  `Promise<Nullable<Person.v1.Person>[]>` and just pass through `null` when a returned doc fails the `isPerson` check.  If we try to do anything more fancy, like calling the `queryDocsByKey` again until we get the full limit of persons, it is going to mess up the `skip` value for any subsequent calls from the consumer...  The only comfort I have here is that we do not really want folks directly calling this function anyway (and the `getAll` can still abstract away the `Nullable`).  The REST api is going to be a bit janky, but I just don't see any other way around it.

Do either of you have other thoughts here?  Ideas on better options?

sugat009: TBH, I'm surprised that we even need a `isPerson` validator since we explicitly pass a `personType` to the database. Anyways, @jkuester is right, if in some way an entry is filtered by the `isPerson` function then the sanctity of the pagination would be disturbed, so let's have a `null` value in the place of the non-person doc. 

As an alternative approach, instead of setting the document entry to null, we could introduce an invalid property or similar indicator. This would signify that the `person` document contains incorrect or problematic data, without completely removing the entry.

jkuester: >  I'm surprised that we even need a isPerson validator since we explicitly pass a valid personType to the database

This is true. I have gone back and forth on if we should keep the `isPerson` check since all it is doing currently is validating the contact type (which, in theory should always be correct since that is what we are querying one...).  _However_, regardless of if we need to do any checks like this currently, I do think it is very important that we design these API patterns in such a way that we _can_ filter data we get from Couch before returning it to the user.  One of the main points of having cht-datasource is that not everything needs to be modeled directly in the db....  With that in mind, and seeing as the performance cost of `isPerson` is minimal, it makes sense to me to keep it for the sake of code consistency (and better support for future validations).  

>  the sanctity of the pagination would be disturbed

Quote of the day!  :fire: 

> we could introduce an invalid property or similar indicator

This is an interesting idea! :thinking:   Similar to the Pouch API which returns the row objects with `id` value, etc, but may or may not actually have a `doc`.  Ultimately, though, it seems like it might not actually be useful to consumers of this API (and anyone trying to do data validation will probably want to connect directly to the database).

---

Another approach we could take here would be to return some kind of page object like this:

```ts
Promise<{ entries: Person.v1.Person[], cursor: string }>
```

This is similar to the paging in GitHub's GraphQL api, where each request returns a different `cursor` that  you can pass back in on subsequent requests to get the next page.  In our case we could just return the proper `skip` value for the next Couch query, but the meaning/value of the `cursor` should be completely abstracted away from consumers of the cht-datasource api. Consumers can just treat it as a token string for getting the next page.  That would give us flexibility in the future if we needed to change up some underlying details of how things work.  In the context of the above discussion regarding `isPerson`, this approach would let us run followup queries to populate a full list of persons to return so that `entries.length === limit` while returning the proper `skip` value for the next page request as the `cursor`.  

I don't know guys, the more I think about this approach, the more I like it!  It decouples us more from the underlying Couch implementation, which I think is a win, giving us more flexibility for future innovation....

jkuester: One variation on this new page object proposal would be to include a `hasNextPage: boolean` in the returned page object so that consumers know when more data is available.  Then, we could clarify that the provided `limit` would be the _max_ value possible for `entries`, (but some pages might be returned with less).  This would free us from having to implement any kind of followup queries in the `getPage` logic (at least at this time).  We could run the Couch query, do any necessary filtering, and return what is left.

sugat009: @jkuester Cursor-based pagination sounds great and would help if any record is filtered by the `isPerson` or any future filters. As for the implementation, should we do it via:
1.  `overfetching` method where we fetch more docs than the limit specifies such that if records are filtered out then, the over-fetched docs can fill in
1. `re-fetching` method where we re-query the database with the amount of docs that were filtered out.

Usually, it is said to be better to minimize the no. of queries to the database which is even more true in our case. 😃 So, my vote is in the first method. What do you think?

m5r: Cursor-based pagination always felt a bit technically clunky but they make a lot of sense from a consumer point of view. I found a couple of resources to grok and compare these two types of pagination and these are the most relevant ones for us:
- https://use-the-index-luke.com/no-offset
- https://guide.couchdb.org/draft/recipes.html#pagination

A scary but interesting bit from CouchDB's guide to pagination (related to [this comment](https://github.com/medic/cht-core/blob/master/sentinel/src/lib/purging.js#L404) in our purging code):
> The `skip` argument works like this: in addition to going to the first row and starting to read, `skip` will skip as many rows as specified, but CouchDB will still read from the first row; it just won’t return any values for the skipped rows. If you specify `skip=100,` CouchDB will read 100 rows and not create output for them. This doesn’t sound too bad, but it is *very* bad, when you use `1000` or even `10000` as `skip` values. CouchDB will have to look at a lot of rows unnecessarily. 

That doesn't really answer "what should we do when we have entries that shouldn't be in the response?".
I personally don't support potentially returning `null` entries in the middle of the response when the user requested 100 contacts... I would much rather do some plumbing to fill up the response with the desired amount of contacts but I can only see this approach working with cursor-based pagination

jkuester: Nice! This discussion is great!  (Also, super helpful resources Mokhar, thanks!  :+1: )  I think that definitely seals the deal that we should ditch `skip` and use a `cursor`.  So, `getPage` could return something like:

```ts
export interface Page<T> {
  readonly data: T[];
  readonly cursor: string;
} 
```

With that in mind, I do think it makes sense to put the extra effort in to avoid returning null.  @sugat009 I like your breakdown of `overfetching` vs `re-fetching`. I guess we cannot _only_ rely on overfetching, though, because there might be a case where we end up filtering out a bunch of docs and so we still don't have enough to return (even though we originally fetched extra...).  I am considering a hypothetical future scenario where maybe we start validating that all returned persons have a `name` value.  Docs with `person` type but no name will be returned from the Couch view.  (And maybe you have a CHT instance where someone ran a script to dump a bunch of invalid `person` contacts with no name.)  In that case, we need to be able to deal with disregarding a bunch of those person contacts.

What do you think of an approach where we first just run a view query for the given limit (no overfetch).  Then, we take the returned docs and filter them.  If we ended up filtering any docs (lets say we filtered out `n` docs), then we should re-fetch for the next `n*2` docs.  For our followup query, we are overfetching by a factor of 2, which should significantly reduce the chances of needing to do additional re-fetches.  Here are the benefits I see of this approach:

- Optimizes for the happy-path, good data workflow.  The only thing we should really be filtering for is bad data.  (Any more specific queries for expected good data should be implemented as a new Couch view and not by querying a broader view and filtering the results).  So, hopefully we are discussing edge case behavior with this filtering
- The  "algorithm" for overfetching on the follow-up calls is simple, but still dynamically tailored to what we know about the state of the data being queried.  (If we filtered `5/100`, we only re-query for the next `10`, but if we filtered `75/100`, we know there is a lot of bad data and we instead query for `150` on the re-fetch)

sugat009: I read through the linked sites by @m5r  and tried a bit on our couchdb view(`medic-client/contacts_by_type`). The `startkey` as a cursor approach does not seem to work for the view as all the keys are person types. We probably need to create a new view designed specifically for pagination.

jkuester: I have not tried it yet, but at the end of that Couch guide, it notes that for views where the key is not unique, you will need to use a combination of `startkey`, `startkey_docid`, and `limit`.  The `startkey_docid` is what will actual do the heavy lifting for our contacts_by_type query since we are not actually looking for a _range_ of keys...

m5r: Good point Josh, the fact that the `key` is the same is what enables couch to do the pagination here. I gave it a try locally and it should do exactly what we need:

Querying the view for persons:
```sh
$ curl "http://admin:password@localhost:5984/medic/_design/medic-client/_view/contacts_by_type?key=\[\"person\"\]"
{"total_rows":6,"offset":3,"rows":[
{"id":"3e93a66c-06c9-4766-b68b-eaec896ee935","key":["person"],"value":"false false 3 grootchw"},
{"id":"6892b0eb-7398-47a9-8e92-d40d3012f9f6","key":["person"],"value":"false false 3 grootvisor"},
{"id":"de324069-da24-4ffa-89f3-f7f67b859d70","key":["person"],"value":"false false 3 grootpatient"}
]}
```

Then using the person with id `6892b0eb-7398-47a9-8e92-d40d3012f9f6` as cursor:
```sh
$ curl "http://admin:password@localhost:5984/medic/_design/medic-client/_view/contacts_by_type?key=\[\"person\"\]&startkey_docid=6892b0eb-7398-47a9-8e92-d40d3012f9f6"
{"total_rows":6,"offset":4,"rows":[
{"id":"6892b0eb-7398-47a9-8e92-d40d3012f9f6","key":["person"],"value":"false false 3 grootvisor"},
{"id":"de324069-da24-4ffa-89f3-f7f67b859d70","key":["person"],"value":"false false 3 grootpatient"}
]}
```

Then limiting the number of rows that I get back:
```sh
$ curl "http://admin:password@localhost:5984/medic/_design/medic-client/_view/contacts_by_type?key=\[\"person\"\]&startkey_docid=6892b0eb-7398-47a9-8e92-d40d3012f9f6&limit=1"
{"total_rows":6,"offset":4,"rows":[
{"id":"6892b0eb-7398-47a9-8e92-d40d3012f9f6","key":["person"],"value":"false false 3 grootvisor"}
]}
```

sugat009: Ahhh! Gotcha. I was under the impression that the `key` also needed to be a uuid. Thanks!

jkuester: Okay, so just recapping further discussion from the [Slack Thread](https://medic.slack.com/archives/C024KTGRW/p1722374179017329), it turns out that Pouch does not have any support for the `startkey_docid` parameter.  :sweat:   Also from reading the [Couch docs](https://docs.couchdb.org/en/stable/ddocs/views/pagination.html#paging-alternate-method) it seems like the performance for `skip` significantly improved after Couch `1.2` (I guess the guide linked above was old?).  The [Pouch Docs](https://pouchdb.com/api.html#query_database) still note that `skip` has "poor performance on IndexedDB/LevelDB", but there does not seem to be any other way to actually page queries with Pouch, so... :shrug:

So the decision is to stick with just using `skip` to page the queries for now.  We should still do the refactor to return the skip value as a generic `cursor` to consumers instead of actually exposing `skip`.  That will let us passively refactor our implementation once https://github.com/pouchdb/pouchdb/issues/7326 is completed.

Issue #9238, opened 2024-07-03
Add functionality of getting people as an iterator in cht-datasource


sugat009:
**Is your feature request related to a problem? Please describe.**

This feature is part of a bigger feature of the creation of API endpoints for getting contacts by type. (#9193)
The target of this feature is to have the functionality of getting people using `cht-datasource` as an `AsyncIterator` or something similar that can fetch data in batches.

**Describe the solution you'd like**

```javascript
Person.v1.getAllAsync();
```

**Describe alternatives you've considered**

**Additional context**

jkuester:
As noted here: https://github.com/medic/cht-core/issues/9237#issuecomment-2221081736

I think we should go for a function signature like:

```ts
const getAll = () => AsyncGenerator<Person.v1.Person, void>
```

Also, FTR, I have found [this blog post](https://javascript.info/async-iterators-generators) to be very helpful in wrapping my mind around async interators/generators.

garethbowen:
@sugat009 Should this be in the release milestone?

sugat009:
@garethbowen yes.

jkuester:
I am removing this from the 4.11 milestone since it is covered already by https://github.com/medic/cht-core/issues/9241.

PR #9281, opened 2024-07-24
feat(#9238): add functionality of getting people as an AsyncGenerator in cht-datasource

sugat009:
# Description

Add functionality of getting people as an `AsyncGenerator` in `cht-datasource`.

Ticket: #9238 

Usage:

```typescript
const getAllIterator = Person.v1.getAll(ctx)(Qualifier.byContactType('person'));
for await (const page of getAllIterator) {
  for (const doc of page) {
    ...
  }
}
```

_NOTE: The target branch is also set to the other feature(#9237) branch temporarily and not the main feature branch as #9237's PR #9266 has not been merged yet and this would give a better diff in this PR for the current changes._

# Code review checklist

# License


sugat009:
@jkuester @m5r I've created a draft PR for discussion.
Currently, the state of the code is that the natural setup of the functionality has been done according to the pattern in `cht-datasource`.
The current plan is to write the iterable logic in [this function](https://github.com/medic/cht-core/blob/03e19f1e4e08afb260fa63847d13226d1559999f/shared-libs/cht-datasource/src/local/person.ts#L81). 
In contrast, the remote logic has been written in [this function](https://github.com/medic/cht-core/blob/03e19f1e4e08afb260fa63847d13226d1559999f/shared-libs/cht-datasource/src/remote/person.ts#L35). The thing that I'm thinking about right now is that if a new endpoint different than the one in which the `getPage(remote)` would be requesting for i.e. `/api/v1/person` is necessary like a `/api/v1/personAsync` or something.
What do you think about this?

sugat009:
Additionally, the plan to implement the iterable for the local function `Person.v1.getAll` is to first query the view `medic-client/contacts_by_type?<personType>` for total row count and then, subsequently query page by page using the already existing `Person.v1.getPage` function until the total row count is reached.

jkuester:
So, my thinking here was that we might be able to get away without having any separate remote/local logic at all. Instead we implement `getAll` in `src/person.ts` in such a way that it just calls the `getPage` function in `src/person.ts`.  I am not sure if there is some extra complexity that would prevent this, but at first glace it seems possible.  With this approach, we should be able to tell when we have reached the end of the data and can stop yielding when we get back an empty page, _or a page with a length `< 100`_.  

Should not need any separate REST endpoint for this functionality.  In reality, all we are implementing here is a TS convenience  wrapper over top of `getPage`.

sugat009:
> So, my thinking here was that we might be able to get away without having any separate remote/local logic at all.
That is an interesting one and I agree that the whole implementation would be simpler and is also feasible. I might be a bit out of context here but why not the remote logic?

jkuester:
> but why not the remote logic?

If the `src/person.ts/getAll` function is just calling the `src/person.ts/getPage`, then all of the remote vs local behavior will be handled by `src/person.ts/getPage` and can be transparent to `src/person.ts/getAll`.  `getAll` is just getting arrays of Persons back. It should not even need to care if they are coming from the remote or the local adapter.  At least that is my thinking anyway...  :thinking:

sugat009:
@jkuester @m5r PR is ready for review.

m5r (changes requested):
So far the API you came up with is good! :clap:

shared-libs/cht-datasource/src/person.ts
```diff
@@ -87,6 +87,35 @@ export namespace v1 {
...
+     * @param personType - The type of person to fetch
+     * @throws throws an error if the provided personType is invalid.
+     * @returns An iterator-like object with a next method for fetching data.
+     */
+    const curriedFn = async function* (personType: ContactTypeQualifier): AsyncGenerator<Person[], void> {
+      assertTypeQualifier(personType);
+      const limit = 100;
+      let skip = 0;
+
+      while (true) {
```
m5r, line 104: ESLint complains about an unnecessary conditional. As much as I like ESLint, I think this is one of the few exceptions where we silence it. What do you think @jkuester @sugat009 ?
```suggestion
      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
      while (true) {
```

shared-libs/cht-datasource/src/person.ts
```diff
@@ -87,6 +87,35 @@ export namespace v1 {
...
+
+    /**
+     * Creates an iterator-like object for fetching batches of people data.
+     * @param personType - The type of person to fetch
+     * @throws throws an error if the provided personType is invalid.
+     * @returns An iterator-like object with a next method for fetching data.
+     */
+    const curriedFn = async function* (personType: ContactTypeQualifier): AsyncGenerator<Person[], void> {
+      assertTypeQualifier(personType);
+      const limit = 100;
```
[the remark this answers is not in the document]
jkuester: I have gone back and forth on this. I am having trouble coming up with a clear example of when someone would want to change from the default batch size....  Maybe if you are writing a script that you know is only going to run with a good internet connection. So you want to bump the batch size to get more data at one time?  IDK, still seems like an edge case.  

At this point, my vote would be to not make the limit configurable for `getAll` at this time.  We can wait until we have an actual use case and then add an optional param for it.  Meanwhile, if someone needs very specific batching functionality, they can always just call `getPage`.

shared-libs/cht-datasource/src/person.ts
jkuester: My vote is for the venerable do-while:

```ts
    const curriedFn = async function* (personType: ContactTypeQualifier): AsyncGenerator<Person, void> {
  const getPage = context.bind(v1.getPage);
  const limit = 100;
  let skip = 0;
  const hasMoreResults = () => skip && skip % limit === 0;
  do {
    const docs = await getPage(personType, limit, skip);
    for (const person of docs) {
      yield person;
    }

    skip += docs.length;
  } while (hasMoreResults());
};
return curriedFn;
};
```

Note that I also think we need to change from returning an array of persons (yielding each batch) to just yielding one person at a time.  I think that is going to be the cleanest for code trying to consume this (otherwise the consuming code ends up in these wierd nested loops).

jkuester: Also, this generator logic is pretty generic.  I guess we could probably re-use it for the `Place.v1.getAll` if you allow the actual `getPage` function to be injected.  If we go that round, I guess we could put the shared function in `libs/data-context.ts`.

shared-libs/cht-datasource/src/person.ts
```diff
@@ -87,6 +87,35 @@ export namespace v1 {
       return curriedFn;
     };
 
+  const getPeopleGenerator = () => (context: DataContext) => {
```
jkuester, line 90: I think we can just combine this function with `getAll`, since I don't think it is going to get reused anywhere...

shared-libs/cht-datasource/src/person.ts
```diff
@@ -87,6 +87,35 @@ export namespace v1 {
...
+    assertDataContext(context);
+
+    /**
+     * Creates an iterator-like object for fetching batches of people data.
+     * @param personType - The type of person to fetch
+     * @throws throws an error if the provided personType is invalid.
+     * @returns An iterator-like object with a next method for fetching data.
```
jkuester, line 97:
```suggestion
     * Returns a generator that yeilds the persons with the given type.
     * @param personType The type of person to fetch
     * @returns a generator that yeilds the persons with the given type
     * @throws Error if no person type is provided or if the type is not for a person
```

shared-libs/cht-datasource/src/person.ts
```diff
@@ -110,4 +139,18 @@ export namespace v1 {
...
+   * Returns a people generator function for fetching batches of people data.
+   * @param context the current data context
+   * @returns a generator function that creates an iterator-like object for fetching people data.
+   * @example
+   * const getAllIterator = Person.v1.getAll(ctx)(Qualifier.byContactType('person'));
+   * for await (const page of getAllIterator) {
+   *   for (const doc of page) {
+   *     ...
+   *   }
+   * }
```
jkuester, line 153: Based on my earlier suggestion, I think the generator should yield individual Person objects (and not batches).  Without the batches, consuming this function should be much more straightforward and I do not think we need an example.

```suggestion
   * Returns a function for retrieving people from the given data context.
   * @param context the current data context
   * @returns a function for retrieving people
   * @throws Error if a data context is not provided
```

jkuester (changes requested):

shared-libs/cht-datasource/src/person.ts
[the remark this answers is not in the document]
sugat009: While a configurable batch size could be beneficial in certain scenarios and aligns with good design principles, I don't see immediate value in it for our current use case. A fixed batch size of 100 seems reasonable. We can revisit this if we encounter performance issues or specific requirements that necessitate adjustment.

shared-libs/cht-datasource/src/person.ts
sugat009: I did a performance test(nothing serious just timed the entry and exit where the function `getAll` is called and looped until the end) for sanity for the current implementation of returning arrays of people VS retuning singular people for data with 27k docs and almost 10k people and the performance is pretty much the same. So, I think if consumers could avoid the nested for loop, we should implement @jkuester's suggestion from above.

shared-libs/cht-datasource/src/libs/data-context.ts
```diff
@@ -39,3 +39,26 @@ export const adapt = <T>(
...
+    limit: number;
+    skip: number;
+  } & Record<string, unknown>
+): AsyncGenerator<unknown, void> {
+  const { limit } = fetchFunctionArgs;
+  let { skip } = fetchFunctionArgs;
+  const hasMoreResults = () => skip && skip % limit === 0;
+
+  do {
+    const docs = await fetchFunction(...(Object.values(fetchFunctionArgs) as Args));
```
m5r, line 56: I'm not a fan of using `...Object.values(args)` as arguments because it now creates an implicit dependency on the order of the properties. In practice, it means that passing `{ limit: 100, skip: 500 }` will yield different results from `{ skip: 500, limit: 100 }` based on how `fetchFunction` expects parameters to be ordered

m5r: So I've been mulling over this and I believe we could simplify it by making `fetchFunction` expect a single object argument, that way the ordering doesn't matter and we get a clear type inference

```ts
interface PaginationArgs {
  limit: number;
  skip: number;
}

type DefaultArgs = PaginationArgs & Record<string, unknown>;

/** @internal */
export const getDocumentStream = async function* <T, Args extends DefaultArgs>(
  fetchFunction: (args: Args) => Promise<T[]>,
  fetchFunctionArgs: Args
): AsyncGenerator<unknown, void> {
  const { limit } = fetchFunctionArgs;
  let { skip } = fetchFunctionArgs;
  const hasMoreResults = () => skip && skip % limit === 0;

  do {
    const docs = await fetchFunction(fetchFunctionArgs);

    for (const doc of docs) {
      yield doc;
    }

    skip += docs.length;
  } while (hasMoreResults());
};
```

and `getPeople()` in `src/person.ts` could look like this to accommodate for the different function signature:

```ts
  const getPeople = <T>(
    localFn: (c: LocalDataContext) => (personType: ContactTypeQualifier, limit: number, skip: number) => Promise<T>,
    remoteFn: (c: RemoteDataContext) => (personType: ContactTypeQualifier, limit: number, skip: number) => Promise<T>
  ) => (context: DataContext) => {
      assertDataContext(context);
      const fn = adapt(context, localFn, remoteFn);

      /**
       * Returns an array of people for the provided page specifications.
       * @param params parameters
       * @param params.personType the type of people to return
       * @param params.limit the maximum number of people to return. Default is 100.
       * @param params.skip the number of people to skip. Default is 0.
       * @returns an array of people for the provided page specifications.
       * @throws Error if `personType` qualifier is invalid
       * @throws Error if the provided `limit` value is `<=0`
       * @throws Error if the provided `skip` value is `<0`
       */
      const curriedFn = async ({personType, limit = 100, skip = 0}: {
      personType: ContactTypeQualifier,
      limit: number,
      skip: number
    }): Promise<T> => {
        assertTypeQualifier(personType);
        assertLimitAndSkip(limit, skip);

        return fn(personType, limit, skip);
      };
      return curriedFn;
    };
```

The formatting is very rough but let me know if the idea makes sense

m5r (changes requested):

shared-libs/cht-datasource/src/person.ts
```diff
@@ -26,6 +26,12 @@ export namespace v1 {
     readonly parent?: Place.v1.PlaceWithLineage | NormalizedParent,
   }
 
+  interface getPageParams {
```
m5r, line 29:
```suggestion
  interface GetPageParams {
```
make sure your interface name is capitalized

shared-libs/cht-datasource/src/libs/data-context.ts
```diff
@@ -39,3 +39,30 @@ export const adapt = <T>(
...
+type FetchFunctionArgs<TFields extends Record<string, unknown>> = PaginationArgs & TFields;
+
+/** @internal */
+export const getDocumentStream = async function* <T, TFields extends Record<string, unknown>>(
+  fetchFunction: (args: FetchFunctionArgs<TFields>) => Promise<T[]>,
+  fetchFunctionArgs: FetchFunctionArgs<TFields>
+): AsyncGenerator<T, void> {
+  const { limit } = fetchFunctionArgs;
+  let { skip } = fetchFunctionArgs;
+  const hasMoreResults = () => skip && skip % limit === 0;
```
m5r, line 57: Thinking out loud, and definitely not necessary at this stage, but we could re-use the numbers in CouchDB's response to know if we've hit the limit here. That would save us one extra query with an empty page in the rare scenario where `total_rows % 100 === 0`
I shared what CouchDB's response looks like in [this comment](https://github.com/medic/cht-core/pull/9266#discussion_r1697035442). For this particular query, the `hasMoreResults` function could be
```ts
const hasMoreResults = () => response.total_rows > (response.offset + response.rows.length) 
```

Again, you don't need to change anything right now, the current logic is good enough for our needs today

shared-libs/cht-datasource/src/person.ts
```diff
@@ -109,5 +92,50 @@ export namespace v1 {
...
+
+      return fn(personType, limit, skip);
+    };
+    return curriedFn;
+  };
+
+  /**
+   * Returns a generator function for retrieving people from the given data context.
+   * @param context the current data context
+   * @returns a generator function for retrieving people
```
jkuester, line 125:
```suggestion
   * Returns a function for getting a generator that fetches people from the given data context.
   * @param context the current data context
   * @returns a function for getting a generator that fetches people
```

shared-libs/cht-datasource/src/person.ts
```diff
@@ -109,5 +92,50 @@ export namespace v1 {
...
+   * @throws Error if a data context is not provided
+   */
+  export const getAll = (
+    context: DataContext
+  ): (personType: ContactTypeQualifier) => AsyncGenerator<v1.Person, void> => {
+    assertDataContext(context);
+
+    return async function* (personType: ContactTypeQualifier): AsyncGenerator<Person, void> {
```
jkuester, line 133: As with the `curriedFn` in  `getPage`, I think we should try to get some docs for this function to show up!  

```ts
    /**
     * Returns a generator for fetching all people with the given type
     * @param personType the type of people to return
     * @returns a generator for fetching all people with the given type
     * @throws Error if no type is provided or if the type is not for a person
     */
```

shared-libs/cht-datasource/src/person.ts
```diff
@@ -109,5 +92,50 @@ export namespace v1 {
    * @returns a function for retrieving a paged array of people
    * @throws Error if a data context is not provided
    */
-  export const getPage = getPeople(Local.Person.v1.getPage, Remote.Person.v1.getPage);
+  export const getPage = (
```
jkuester, line 95: As in the index, it would be nice to add some docs here encouraging folks to use the other function:

> @see {@link getByType} which provides the same data, but without having to manually account for paging

jkuester (changes requested):

shared-libs/cht-datasource/src/libs/data-context.ts
```diff
@@ -39,3 +39,30 @@ export const adapt = <T>(
...
+
+type FetchFunctionArgs<TFields extends Record<string, unknown>> = PaginationArgs & TFields;
+
+/** @internal */
+export const getDocumentStream = async function* <T, TFields extends Record<string, unknown>>(
+  fetchFunction: (args: FetchFunctionArgs<TFields>) => Promise<T[]>,
+  fetchFunctionArgs: FetchFunctionArgs<TFields>
+): AsyncGenerator<T, void> {
+  const { limit } = fetchFunctionArgs;
+  let { skip } = fetchFunctionArgs;
```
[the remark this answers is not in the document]
sugat009: Good catch. The `personType` for the case of a person or maybe `placeType` for places is still required.

shared-libs/cht-datasource/src/person.ts
```diff
@@ -109,5 +92,50 @@ export namespace v1 {
...
+     * @param params the function params
+     * @param params.personType the type of people to return
+     * @param params.limit the maximum number of people to return. Default is 100.
+     * @param params.skip the number of people to skip. Default is 0.
+     * @returns an array of people for the provided page specifications.
+     * @throws Error if no type is provided or if the type is not for a person
+     * @throws Error if the provided `limit` value is `<=0`
+     * @throws Error if the provided `skip` value is `<0`
+     */
+    const curriedFn = async ({ personType, limit = 100, skip = 0}: GetPageParams): Promise<Person[]> => {
```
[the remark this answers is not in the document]
sugat009: Yeah, that was the initial implementation. However, `getDocumentStream` needs to call the function passed in function with the passed-in parameter too. The initial implementation included spreading the passed parameters without modifying the function in question but as @m5r flagged that would lead to having the consumer of the function `getDocumentStream` have to pass the arguments in an ordered way. [Ref](https://github.com/medic/cht-core/pull/9281#discussion_r1695425097). Since `getPage` is public whereas `getDocumentStream` is not, makes the consumers of `getDocumentStream` ourselves which might make us aware that we need to send in the params in order.

shared-libs/cht-datasource/src/person.ts
sugat009: I tried adding the docs to this curried function with and without the pattern we are doing for the other ones but it didn't work. I think it's because it's a generator function and I didn't want to spend too much time on this.

shared-libs/cht-datasource/src/person.ts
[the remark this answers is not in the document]
jkuester: So, ultimately I think we should still be able to get this to work via generics.  One think that should make our current code more simple is that [we only need to pass the person type](https://github.com/medic/cht-core/pull/9281#discussion_r1705349811) to `getDocumentStream` (and not `limit` or `skip`).  

I recognize that this may cause trouble in the future when we want to reuse this with a function that needs multiple parameters, but honestly I would rather take that risk than have to change our public `getPage` api just to support this now.  (Especially if we don't actually need the change.)

(Also I apologize for bringing this up late in the game here. :sweat: I did not fully realize the implications of Mokhtar's earlier suggestion until I saw it  fully coded up here...)

shared-libs/cht-datasource/src/person.ts
```diff
@@ -108,6 +91,50 @@ export namespace v1 {
    * @param context the current data context
    * @returns a function for retrieving a paged array of people
    * @throws Error if a data context is not provided
+   * @see {@link getAll} which provides the same data, but without having to manually account for paging
    */
-  export const getPage = getPeople(Local.Person.v1.getPage, Remote.Person.v1.getPage);
+  export const getPage = (
+    context: DataContext
+  ): ({ personType, limit, skip }: GetPageParams) => Promise<Person[]> => {
```
jkuester, line 98:
```suggestion
  ): typeof curriedFn => {
```

Got to do it this way to get the docs from `curriedFn` to show up in the site... :sweat:

shared-libs/cht-datasource/src/person.ts
jkuester: I managed to get this to work!  The key is, as I mentioned above, you have to set the return type of the outer function to be `typeof` the curried function variable name.  (for some reason this works thanks to the fact that types are not real... :shrug: )

```ts
  /**
   * Returns a generator function for retrieving people from the given data context.
   * @returns a generator function for retrieving people
   */
  export const getAll = (
    context: DataContext
  ): typeof curriedGen => {
    assertDataContext(context);

    /**
     * Returns a generator for fetching all people with the given type
     * @param personType the type of people to return
     * @returns a generator for fetching all people with the given type
     * @throws Error if no type is provided or if the type is not for a person
     */
    const curriedGen = async function* (personType: ContactTypeQualifier): AsyncGenerator<Person, void> {
      assertTypeQualifier(personType);
      const getPage = context.bind(v1.getPage);
      const limit = 100;
      const skip = 0;
      yield* getDocumentStream(getPage, { personType, limit, skip });
    };
    return curriedGen;
  };
```

sugat009: ahhh! excellent find. Thank you.

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -61,19 +61,55 @@ export namespace v1 {
...
+      const personTypesIds = personTypes.map((item) => item.id);
 
       if (!personTypesIds.includes(personType.contactType)) {
         throw new Error(`Invalid person type: ${personType.contactType}`);
       }
 
-      const docs = await getDocsByPage([personType.contactType], limit, skip);
+      const getDocsByPage = queryDocsByKey(medicDb, 'medic-client/contacts_by_type');
+
+      const fetchAndFilter = async (
```
jkuester, line 78: Solid logic here!  :+1:  I think we can tighten up the cursor calculation a bit (at least reduce the nesting if-statements):

```ts
const fetchAndFilter = async (
  currentLimit: number,
  currentSkip: number,
  currentPersons: Person.v1.Person[] = []
): Promise<Page<Person.v1.Person>> => {
  const docs = await getDocsByPage([personType.contactType], currentLimit, currentSkip);
  const noMoreResults = docs.length < currentLimit;
  const newPersons = docs.filter((doc): doc is Person.v1.Person => isPerson(settings, doc, doc?._id));

  // Need to slice here because we retrieve extra docs on re-fetches and may end up with too many.
  const overFetchCount = currentPersons.length + newPersons.length - limit || 0;
  const totalPersons = [...currentPersons, ...newPersons].slice(0, limit);
  if (noMoreResults) {
    return { data: totalPersons };
  }
  if (totalPersons.length === limit) {
    const nextSkip =  currentSkip + currentLimit - overFetchCount;
    return { data: totalPersons, cursor: (nextSkip).toString() };
  }

  // Re-fetch twice as many docs as we need to limit number of recursions
  const nextLimit = (currentLimit - newPersons.length) * 2;
  const nextSkip = currentSkip + currentLimit;
  return fetchAndFilter(nextLimit, nextSkip, totalPersons);
};

return fetchAndFilter(limit, skip);
```

shared-libs/cht-datasource/src/remote/person.ts
```diff
@@ -27,7 +27,7 @@ export namespace v1 {
     personType: ContactTypeQualifier,
     limit: number,
     skip: number
-  ): Promise<Person.v1.Person[]> => getPeople(remoteContext)(
+  ): Promise<Page<Person.v1.Person>> => getPeople(remoteContext)(
     {'limit': limit.toString(), 'skip': skip.toString(), 'contactType': personType.contactType}
```
jkuester, line 31: Need to make sure to update `skip` to `cursor`.

jkuester (changes requested):

shared-libs/cht-datasource/src/libs/core.ts
```diff
@@ -112,3 +112,9 @@ export const findById = <T extends Identifiable>(values: T[], id: string): Nulla
 export abstract class AbstractDataContext implements DataContext {
   readonly bind = <T>(fn: (ctx: DataContext) => T): T => fn(this);
 }
+
+/** @internal */
+export interface Page<T> {
+  readonly data: T[];
+  readonly cursor: string;
```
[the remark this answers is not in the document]
m5r: Good catch! I believe `cursor`'s type should be
```ts
interface Page {
  readonly cursor: string | null;
}
```
That's more explicit to express "you've reached the end"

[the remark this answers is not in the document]
sugat009: @m5r and I commented simultaneously, `null` should do as well.

m5r (approved):
Nothing to add to Josh's comments, I'll approve to not block merging after Josh's approval

PR #9282, opened 2024-07-24
feat(#9231): add facility indicator in aggregates targets

latin-panda:
# Description

Fixes colors in report's sidebar filter - it should look like in 4.9 branch 
Adds indicator of selected facility in aggregate targets list and details.

<details><summary>Screenshots Aggregate Target</summary>

Screenshot 2024-07-24 at 8 21 00 PM
Screenshot 2024-07-24 at 8 20 51 PM
Screenshot 2024-07-24 at 8 17 15 PM

</details>

<details><summary>Screenshot - Report filter with colors</summary>

Screenshot 2024-07-26 at 3 12 29 PM

</details>

#9231
https://github.com/medic/care-teams/issues/144

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:add-facility-name-in-aggregates-targes/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:add-facility-name-in-aggregates-targes/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:add-facility-name-in-aggregates-targes/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
+++ b/webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
@@ -1,4 +1,4 @@
-import { Component, EventEmitter, OnDestroy, OnInit, Output } from '@angular/core';
+import { Component, EventEmitter, OnDestroy, OnInit, Output, Input } from '@angular/core';
 import { Subscription } from 'rxjs';
 import { Store } from '@ngrx/store';
 
@@ -6,7 +6,6 @@ import { GlobalActions } from '@mm-actions/global';
 import { Selectors } from '@mm-selectors/index';
 import { ContactTypesService } from '@mm-services/contact-types.service';
 import { SettingsService } from '@mm-services/settings.service';
-import { UserSettingsService } from '@mm-services/user-settings.service';
 
 @Component({
   selector: 'mm-analytics-target-aggregates-sidebar-filter',
@@ -14,21 +13,19 @@ import { UserSettingsService } from '@mm-services/user-settings.service';
 })
 export class AnalyticsTargetAggregatesSidebarFilterComponent implements OnInit, OnDestroy {
 
+  @Input() userFacilities;
   @Output() facilitySelected = new EventEmitter<string>();
   private globalActions;
   subscriptions: Subscription = new Subscription();
   error;
   isOpen = false;
-  userFacilities;
   selectedFacility;
-  selectedFacilityId;
   facilityFilterLabel;
 
   constructor(
     private store: Store,
     private contactTypesService: ContactTypesService,
     private settingsService: SettingsService,
-    private userSettingsService: UserSettingsService,
   ) {
     this.globalActions = new GlobalActions(store);
   }
@@ -36,7 +33,6 @@ export class AnalyticsTargetAggregatesSidebarFilterComponent implements OnInit, 
   async ngOnInit() {
     try {
       this.subscribeToStore();
-      await this.loadUserFacility();
       await this.setFacilityLabel();
     } catch (err) {
       this.error = true;
@@ -45,7 +41,6 @@ export class AnalyticsTargetAggregatesSidebarFilterComponent implements OnInit, 
   }
 
   ngOnDestroy() {
-    this.globalActions.clearSidebarFilter();
     this.subscriptions.unsubscribe();
   }
 
@@ -54,6 +49,9 @@ export class AnalyticsTargetAggregatesSidebarFilterComponent implements OnInit, 
       .select(Selectors.getSidebarFilter)
       .subscribe((filterState) => {
         this.isOpen = filterState?.isOpen ?? false;
+        if (!this.selectedFacility && filterState?.defaultFilters?.facility) {
+          this.selectedFacility = filterState.defaultFilters.facility;
+        }
       });
 
     this.subscriptions.add(subscription);
@@ -62,16 +60,6 @@ export class AnalyticsTargetAggregatesSidebarFilterComponent implements OnInit, 
   toggleSidebarFilter() {
     this.isOpen = !this.isOpen;
     this.globalActions.setSidebarFilter({ isOpen: this.isOpen });
-  }
-
-  private async loadUserFacility() {
-    const hasMultipleFacilities = await this.userSettingsService.hasMultipleFacilities();
-    this.userFacilities = await this.userSettingsService.getUserFacility();
-
-    if (hasMultipleFacilities) {
-      this.selectedFacility = this.userFacilities[0];
-      this.selectedFacilityId = this.userFacilities[0]._id;
-    }
   }
 
   private async setFacilityLabel() {
@@ -94,7 +82,8 @@ export class AnalyticsTargetAggregatesSidebarFilterComponent implements OnInit, 
     }
   }
 
-  fetchAggregateTargets(facilityId) {
-    this.facilitySelected.emit(facilityId);
+  fetchAggregateTargets(facility) {
+    this.selectedFacility = facility;
+    this.facilitySelected.emit(this.selectedFacility);
   }
 }
--- a/webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
+++ b/webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
@@ -16,9 +16,10 @@ export class AnalyticsTargetAggregatesSidebarFilterComponent implements OnInit, 
   @Input() userFacilities;
   @Output() facilitySelected = new EventEmitter<string>();
   private globalActions;
+  DEFAULT_FACILITY_LABEL = 'Facility';
   subscriptions: Subscription = new Subscription();
-  error;
   isOpen = false;
+  error;
   selectedFacility;
   facilityFilterLabel;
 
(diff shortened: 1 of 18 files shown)


latin-panda:
@jkuester @ralfudx Can you please review? I'm adding you since you reviewed Ben's work before. This is a continuation of that, it adds the selected facility name to the list and detail views. 
This filter appears only when `can_have_multiple_places` permission is enabled, and the user has multiple facilities assigned. Multiple facilities are assigned by using the Admin app > Users

webapp/src/ts/modules/analytics/analytics-target-aggregates.component.ts
```diff
@@ -1,110 +1,119 @@
...
-      return;
-    }
-
     const subscription = this.store
       .select(Selectors.getSidebarFilter)
-      .subscribe((filterState) => this.isSidebarFilterOpen = filterState?.isOpen ?? false);
+      .subscribe(sidebarFilter => {
+        this.sidebarFilter = sidebarFilter;
+        if (!this.initialLoad && this.sidebarFilter?.defaultFilters) {
+          this.initialLoad = this.getTargetAggregates(this.sidebarFilter.defaultFilters.facility);
```
jkuester, line 107: I am probably missing something here, but this feels like an odd case to use a subscription.  We only want to set the target aggregates to the default facility once (when we first load into the target aggregates component).  But, we are subscribed to `getSidebarFilter`, so this will get called each time something changes with the filter.  So, you put in the `initialLoad` variable so we only run this once...  For something that should only happen once, it seems like we do not need a subscription.  

Could we maybe just move the `getTargetAggregates` call down into `setDefaultFilters` and get rid of this function (and `initialLoad`) altogether?

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
```diff
@@ -1,51 +1,47 @@
...
 
+  @Input() userFacilities;
   @Output() facilitySelected = new EventEmitter<string>();
   private globalActions;
+  DEFAULT_FACILITY_LABEL = 'Facility';
   subscriptions: Subscription = new Subscription();
-  error;
   isOpen = false;
-  userFacilities;
+  error;
```
jkuester, line 22: What is this `error` property doing anyway?  I cannot seem to find it used anywhere...

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
```diff
@@ -1,51 +1,47 @@
...
   ) {
     this.globalActions = new GlobalActions(store);
   }
 
   async ngOnInit() {
     try {
       this.subscribeToStore();
-      await this.loadUserFacility();
-      await this.setFacilityLabel();
+      this.facilityFilterLabel = await this.setFacilityLabel() || this.DEFAULT_FACILITY_LABEL;
```
jkuester, line 37: You don't have to change anything here, but I wanted to say that for the record I preferred having the `DEFAULT_FACILITY_LABEL` logic encapsulated down in `setFacilityLabel`.  I find it cleaner and easier to understand.  

(Also would not complain if you renamed the function to `getFacilityLabel`...)

jkuester (changes requested):
Very nice!  This feature is looking great! Just a few minor comments/question...

webapp/src/ts/modules/analytics/analytics-target-aggregates.component.ts
```diff
@@ -1,110 +1,119 @@
...
-        this.enabled = enabled;
-
-        if (!this.enabled) {
-          return;
-        }
+  async getTargetAggregates(userFacility) {
+    try {
+      const aggregates = this.enabled ? await this.targetAggregatesService.getAggregates(userFacility?._id) : [];
+      if (this.sidebarFilter.hasFacilityFilter) {
+        aggregates?.forEach((aggregate) => aggregate.facility = userFacility?.name);
```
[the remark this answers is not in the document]
latin-panda: You're right. I removed the optional chaining, taking those extra points by defining a type for Targets and Aggregate Targets. I'm leaving a question in that code, I'd like your opinion on those types please.

webapp/src/ts/modules/analytics/analytics-target-aggregates.component.ts
```diff
@@ -1,110 +1,119 @@
...
 export class AnalyticsTargetAggregatesComponent implements OnInit, OnDestroy {
-  @ViewChild(AnalyticsTargetAggregatesSidebarFilterComponent) sidebarFilter?:
-   AnalyticsTargetAggregatesSidebarFilterComponent;
 
   private targetAggregatesActions: TargetAggregatesActions;
+  private globalActions: GlobalActions;
   private trackPerformance;
   subscriptions: Subscription = new Subscription();
+  userFacilities;
+  initialLoad;
```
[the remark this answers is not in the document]
latin-panda: I've removed the `initialLoad`

webapp/src/ts/modules/analytics/analytics-target-aggregates.component.ts
```diff
@@ -1,110 +1,119 @@
...
+        this.sidebarFilter = sidebarFilter;
+        if (!this.initialLoad && this.sidebarFilter?.defaultFilters) {
+          this.initialLoad = this.getTargetAggregates(this.sidebarFilter.defaultFilters.facility);
+        }
+      });
     this.subscriptions.add(subscription);
   }
+
+  private async setDefaultFilters() {
+    this.userFacilities = await this.userSettingsService.getUserFacility();
```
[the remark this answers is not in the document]
latin-panda: Nice catch! I've fixed it

webapp/src/ts/modules/analytics/analytics-target-aggregates.component.ts
latin-panda: Yes, totally! That's how it started, but later made that setDefaultFilters function and didn't see this simplification. 

We're adding another filter (by period) with a default value, so this is perfect. Thanks!

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
latin-panda: Me neither, I've removed it

webapp/src/ts/modules/analytics/analytics-target-aggregates-sidebar-filter.component.ts
latin-panda: I wanted to set the default `DEFAULT_FACILITY_LABEL` once, but also I don't like it in the `ngOnInit`. 
I have reverted the change and left it in the set function.

webapp/src/ts/services/target-aggregates.service.ts
```diff
@@ -1,6 +1,7 @@
 import { Injectable, NgZone } from '@angular/core';
 import * as moment from 'moment';
 import { isString as _isString } from 'lodash-es';
+import { Contact } from '@medic/cht-datasource/dist/libs/contact';
```
latin-panda, line 4: @jkuester is it okay to import the type directly from here?

latin-panda:
@jkuester thanks for the review, please have a look again, I've added some questions too

webapp/src/ts/services/target-aggregates.service.ts
jkuester: Probably not a bid deal either way, but for the sake of consistency and completeness it would be better to export the interface from the cht-datasource index.ts (similar to what we did [here](https://github.com/medic/cht-core/blob/282faee191a448bbcff8f67baa8b3fc844ef14dc/shared-libs/cht-datasource/src/index.ts#L35) and then remove `@internal` from the JSDoc for `Contact` (can add a bit of real docs that indicate a contact could either be a place or a person...)

I think in the long-term types will be a bit more manageable if we can see which ones are explicitly exported.

webapp/src/ts/services/target-aggregates.service.ts
```diff
@@ -350,3 +351,29 @@ export class TargetAggregatesService {
...
+  values: ContactTargetValue[];
+  aggregateValue: AggregateTargetValue;
+  dhis?: {
+    dataElement?: string;
+    categoryOptionCombo?: string;
+    attributeOptionCombo?: string;
+  };
+}
+
+export interface AggregateTargetValue extends TargetValue {
```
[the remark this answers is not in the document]
jkuester: As before, to me the main question is if `AggregateTargetValue`'s fields are actually a superset of `TargetValue`.  If so, then extending makes total sense.  If, in reality, aggregate target value objects never actually contain some of `TargetValue`'s properties, then yeah, probably a bad idea to extend.

jkuester (approved):
Most excellent!  (:sweat_smile: for the record, when I suggested adding an actual return type for the aggregate targets, in my head I was just thinking something like `Record<string, unknown>[]` (just so we could know if we were getting an object or an array, or something that could be null)...  But I have a dream that one day we can have nice types in webapp and these changes are a great step in that direction! Thank you for taking the time to create them.)

webapp/src/ts/services/target-aggregates.service.ts
latin-panda: Got it! Fixed

webapp/src/ts/services/target-aggregates.service.ts
```diff
@@ -350,3 +351,29 @@ export class TargetAggregatesService {
   }
 
 }
+
+export interface AggregateTarget extends Target {
```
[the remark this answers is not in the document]
latin-panda: Thanks for adding all these insights. It is very helpful!

PR #9295, opened 2024-08-01
feat(#9241): create REST API endpoint for getting people

sugat009:
# Description
Add a new REST API endpoint that gets people using `cht-datasource`.

Ticket: #9241 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9241-create-api-endpoint-for-getting-people/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9241-create-api-endpoint-for-getting-people/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9241-create-api-endpoint-for-getting-people/docker-compose/cht-couchdb-clustered.yml)

# License


shared-libs/cht-datasource/src/libs/error.ts
```diff
@@ -0,0 +1,17 @@
...
+ * that doesn't meet the expected criteria or constraints.
+ */
+export class InvalidArgumentError extends Error {
+  /**
+   * Constructor
+   * @param message a descriptive error message why the error was raised
+   */
+  constructor(message: string) {
+    super(message);
+    this.name = message;
```
m5r, line 13:
```suggestion
    this.name = 'InvalidArgumentError';
```

I would expect the name property to hold the name of the class. The message can always be accessed in 
the message property since the `Error` constructor assigns it there.

shared-libs/cht-datasource/src/libs/error.ts
```diff
@@ -0,0 +1,17 @@
...
+
+    Object.setPrototypeOf(this, InvalidArgumentError.prototype);
```
m5r, line 15: This used to be how to do [prototypal inheritance](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Inheritance_and_the_prototype_chain#building_longer_inheritance_chains) before ES6 classes came along. Since we're using classes, this doesn't need to be set
```suggestion
```

api/tests/mocha/controllers/person.spec.js
```diff
@@ -154,5 +154,113 @@ describe('Person Controller', () => {
...
+    describe('getPageByType', () => {
+      let personGetPageByType;
+      let qualifierByContactType;
+      const personType = 'person';
+      const invalidPersonType = 'invalidPerson';
+      const personTypeQualifier = { contactType: personType };
+      const person = { name: 'John Doe' };
+      const limit = 100;
+      const skip = 0;
+      const people = Array.from({ length: 3 }, () => ({ ...person }));
```
m5r, line 167: That's neat! :+1:

api/tests/mocha/controllers/person.spec.js
m5r, line 1: Thanks for adding unit tests to cover this route! Could you add integration tests as well in `tests/integration/api/controllers/` ?

m5r (changes requested):

m5r (approved):
Thanks for adding the integration test as well! I left a note but it's a minor change request, feel free to merge once you resolve it

api/src/controllers/person.js
```diff
@@ -8,20 +8,45 @@ const getPerson = ({ with_lineage }) => ctx.bind(
...
-    })
-  }
+    }),
+    getPageByType: serverUtils.doOrError(async (req, res) => {
+      await checkUserPermissions(req);
+
+      try {
+        const personType  = Qualifier.byContactType(req.query.personType);
+        const limit = req.query.limit !== undefined ? Number(req.query.limit) : 100;
+        const skip = req.query.skip !== undefined ? Number(req.query.skip) : 0;
```
jkuester, line 37: skip > cursor

jkuester, line 37: The value of putting all the overkill parameter validation (and default values) into cht-datasource is that we should be able to avoid extra logic here.  I think JS will even handle auto-boxing the string values from the query parameters into number values.  Unless I am missing something, I think we should just pass the query parameters straight to the cht-datasource function.

api/src/controllers/person.js
```diff
@@ -8,20 +8,45 @@ const getPerson = ({ with_lineage }) => ctx.bind(
...
+
+        // TODO: change this when #9281 gets merged
+        const docs = await getPageByType()( personType, limit, skip );
+
+        return res.json(docs);
+      } catch (err) {
+        if (err instanceof InvalidArgumentError) {
```
jkuester, line 44: Can you just put this logic down in the `server-utils.error` function?  (Basically just make sure the `code` gets set to `400` when we have an InvalidArgumentError).

Then, we would not need this try-catch block up here in `person.js`...

jkuester (changes requested):

api/src/controllers/person.js
sugat009: Actually, it doesn't.
I removed the number validation and conversion stuff and did this.
```javascript
const docs = await getPageByType()( personType, req.query.limit, req.query.skip );
```

As you can see in the screenshots below, the values are in strings.
image

which then throws the error
image

However, the above code can be a bit shortened to:

```javascript
const limit = Number(req.query.limit) || 100;
const skip = Number(req.query.skip) || 0;
```

Issue #9106, opened 2024-05-10
Investigate moving common dependencies to root to make updates easier


garethbowen:
**Describe the issue**

Updating dependencies is necessary for patching vulnerabilities quickly, but...
- Updating dependencies manually is a big effort because it means navigating to a dozen directories (root, api, sentinel, webapp, admin, and all shared-libs) and doing an npm install. It's inevitable that we get inconsistencies with some dependencies being updated and some not.
- Automatic updating through dependabot isn't working well because it does one directory at a time so you get a dozen PRs every time lodash updates.
- Node services are packaged very crudely by just including all shared libs rather than just those that are needed for that service. This is not a big deal because if the code doesn't run then it shouldn't be exploitable, and it doesn't contribute much to image size.
- Finally this continues to get worse as we add more and more shared libs

**Describe the improvement you'd like**

The way monorepos should work is all common dependencies (and maybe _all_ dependencies) are defined at the root level. This works fine for bundled apps (admin and webapp) because the bundler pulls in only needed code. But we don't have a good solution for node services. This _should_ be solvable by `npm pack` with `bundledDependencies`.

**Describe alternatives you've considered**

- Bundling service code (eg: TS) but that adds another step to the dev build that I'd rather avoid.

garethbowen:
Also check out the `npm-pack-all-zip` package..

garethbowen:
Note to self... commands I've been using to test this out:

Using the npm-pack-all-zip package
```
rm -r ../../api & npm install && npx npm-pack-all-zip --dist=../../ && unzip ../../medic-api.zip -d ../../api && node ../../api/server.js
```

Using native npm pack
```
rm -r ../../api/package & npm install && npm pack --pack-destination=../../ && tar -Pxvzf ../../medic-api-0.1.0.tgz -C ../../api && node ../../api/package/server.js
```

garethbowen:
More todo list items that can be investigated during or after this work:

- [ ] adding more dependabot locations, eg: GH actions, docker, etc
- [ ] adding dependabot ignores for the list of [things we should not update](https://docs.communityhealthtoolkit.org/contribute/code/core/update-dependencies/#problems)

garethbowen:
I've played around with a variety of options, and nothing is perfect. Basically monorepos work fine for webapps where the distributed code is bundled/compiled but as far as I can see nobody is doing monorepos for node services.

These are my findings...

## 1. Status quo

[...]

PR #9300, opened 2024-08-05
chore(#9106): move npm dependencies to root

garethbowen:
# Description

#9106

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9106-just-share-node-modules/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9106-just-share-node-modules/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9106-just-share-node-modules/docker-compose/cht-couchdb-clustered.yml)

# License


garethbowen:
@m5r Would you mind reviewing this change? I think there are more improvements that could be made but it's in a stable spot so if you agree it's moving in the right direction I'd like to merge before moving on to the next step.

package.json
m5r, line 1: Would it be possible to either move all `@angular/*` dependencies from webapp to the root or move all `@angular/*` devDependencies from the root to webapp?

Right now running `npx ng update @angular/core@18 @angular/cli@18` from root fails with `Error: This command is not available when running the Angular CLI outside a workspace.`. And running the same command from webapp goes a bit further but eventually fails:
```
The installed Angular CLI version is outdated.
Installing a temporary Angular CLI versioned 18.1.3 to perform the update.
✔ Packages successfully installed.
Using package manager: npm
Collecting installed dependencies...
Found 50 dependencies.
Package '@angular/cli' is not a dependency.
```
This split of packages between root and webapp makes angular updates a bit awkward and I was hoping this PR would fix that.
To be clear this isn't a regression, we have the same behavior on `master`. I couldn't figure out why we have this split in the first place.

m5r (changes requested):
Thanks for making dependencies easier to maintain. I left one comment about angular dependencies, let me know what you think about it

package.json
garethbowen: Yes getting all angular dependencies in the same package.json is definitely one of the next steps I mentioned. The split used to be that all dev depdendencies went into the root, and all prod dependencies went in the subfolder. Now the pattern is to put all dependencies in the root, unless they conflict with other dependencies. In the case of angular because it's not shared, and because it potentially conflicts with admin (angularjs) I'm tempted to move the angular dev dependencies down into webapp, that way they can be versioned together as expected.

The other next step is I think I can move all "shared-lib" dependencies to the root.

Do you think this is a blocker, or can we merge this as is and do the next steps in subsequent PRs?

m5r: Sounds like a good plan! It's definitely not blocking, we've had this behavior for as long as I can remember. I'll approve the PR so you can move ahead and merge it.

For the shared-lib dependencies, I haven't tried it yet but I believe we will run into some circular dependency issue if we move everything to the root package.json. It's just a hunch, hopefully I'm wrong

m5r (approved):

PR #9306, opened 2024-08-07
chore(#9106): fix dependencies for e2e tests that runs in local environment

latin-panda:
# Description

1. The problem is that the `cp -r` doesn't preserve symbolic links. So I changed to this `await exec('cp', ['-a', '${workingDir}/node_modules', service]);`

2. Then, the docker files still need the node-modules and the shared-libs folders to find the dependencies during run time. I'm copying those folders into the working directory of the container. 

#9106

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:fixes-e2e-for-local-run/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:fixes-e2e-for-local-run/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:fixes-e2e-for-local-run/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/scripts/build/index.js
+++ b/scripts/build/index.js
@@ -244,8 +244,9 @@ const buildServiceImages = async () => {
 
   console.log('working directory', workingDir);
   for (const service of versions.SERVICES) {
-    await exec('cp', ['-r', `${workingDir}/node_modules`, service]);
-    await exec('cp', ['-r', './shared-libs', service]);
+    await exec('rsync', ['-r', '-p', './shared-libs', service]);
+    await exec('rsync', ['-r', '-p', `${workingDir}/node_modules`, service]);
+
 
     const tag = versions.getImageTag(service);
     if (INTERNAL_CONTRIBUTOR) {
--- a/scripts/build/index.js
+++ b/scripts/build/index.js
@@ -244,7 +244,6 @@ const buildServiceImages = async () => {
 
   console.log('working directory', workingDir);
   for (const service of versions.SERVICES) {
-    await exec('rsync', ['-r', '-p', './shared-libs', service]);
     await exec('rsync', ['-r', '-p', `${workingDir}/node_modules`, service]);
 
     const tag = versions.getImageTag(service);
--- a/scripts/build/index.js
+++ b/scripts/build/index.js
@@ -244,7 +244,7 @@ const buildServiceImages = async () => {
 
   console.log('working directory', workingDir);
   for (const service of versions.SERVICES) {
-    await exec('rsync', ['-r', '-p', `${workingDir}/node_modules`, service]);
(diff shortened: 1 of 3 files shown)


scripts/build/index.js
```diff
@@ -244,8 +244,8 @@ const buildServiceImages = async () => {
 
   console.log('working directory', workingDir);
   for (const service of versions.SERVICES) {
-    await exec('cp', ['-r', `${workingDir}/node_modules`, service]);
-    await exec('cp', ['-r', './shared-libs', service]);
+    await exec('rsync', ['-r', '-p', './shared-libs', service]);
+    await exec('rsync', ['-r', '-p', `${workingDir}/node_modules`, service]);
```
latin-panda, line 248: @m5r This is how I found to keep the link reference, the `rsync` is available in the containers. 
What do you think? Is it okay?

latin-panda:
@m5r I'm still waiting for the CI to pass, but these are the minimum changes we need to run the e2e in the local environment. I only tested it on my Mac. Does it work for you?

scripts/build/index.js
m5r: `rsync` does the job well but it's not always available. If this script is only ever executed in docker images, I think this is fine. But if this gets executed on the host machine, it would be safer to make it work with `cp`

I can't give it a shot locally at the moment but have you tried some of the `cp` options mentioned [here](https://superuser.com/a/138588)? I think one of them could solve our symlink issue here

latin-panda: yes, I tried `cp -Pr`, `cp -r -a`, but it doesn't work in mac at least

scripts/build/index.js
```diff
@@ -244,8 +244,7 @@ const buildServiceImages = async () => {
 
   console.log('working directory', workingDir);
   for (const service of versions.SERVICES) {
-    await exec('cp', ['-r', `${workingDir}/node_modules`, service]);
-    await exec('cp', ['-r', './shared-libs', service]);
```
latin-panda, line 248: These 2 lines leave a `node-module` and `shared-libs` folders inside `api` and `sendinel` directories. Then git shows like +600 new files to add in the diff.  
It doesn't need to copy the `shared-libs`. While the `node_modules` is already in the git ignore. 

@garethbowen is this okay? I wonder if I'm not seeing something about copying `shared-libs`.

garethbowen: This was an attempt to simplify the Dockerfile and get development closer to prod, but clearly this was a step too far. Thanks for finding it!

garethbowen (approved):

Issue #9293, opened 2024-07-31
Research freetext view use to inform hosting TCO reduction


mrjones-plip:
**Describe the issue**
it's been decided that storage is the highest cost for hosting total cost of ownership (TCO).  It's suspected that CHT's use of CouchDB freetext views cause a large amount of disk space.  But how much? How might the size of the view be reduced? Where does the UI surface the use of these views?

**Describe the improvement you'd like**
research and document use of freetext views so we know how to safely reduce the disk use caused by using them

**Describe alternatives you've considered**
na

-----------------------

**Next Steps:**

 * [x] Check Couch/HAProxy logs for instances of the views being used
 * [ ] [community research](https://github.com/medic/cht-core/issues/9365)
 * [ ] Remove views and run e2e tests and see what happens
 * [ ] Add logging for the view and manually poke around the app

Open Questions:
* Where are the views being used programmatically?
* What fields are being searched for?
* What are users actually searching for?

----------------------------------

**CouchDB views for freetext searching**
* [contacts_by_freetext](https://github.com/medic/cht-core/blob/67bb273e2d20b775b27a6c0bdf20b169284aff1d/ddocs/medic-db/medic-client/views/contacts_by_freetext/map.js#L11)
* [contacts_by_type_freetext](https://github.com/medic/cht-core/blob/master/ddocs/medic-db/medic-client/views/contacts_by_type_freetext/map.js)
* [reports_by_freetext](https://github.com/medic/cht-core/blob/master/ddocs/medic-db/medic-client/views/reports_by_freetext/map.js)

These get queried by the @medic/search shared lib:
* code that generates queries against the 2 contact-related views: [shared-libs/search/src/generate-search-requests.js](https://github.com/medic/cht-core/blob/master/shared-libs/search/src/generate-search-requests.js#L264-L276)
* code that generates queries against the report-related view: [shared-libs/search/src/generate-search-requests.js](https://github.com/medic/cht-core/blob/master/shared-libs/search/src/generate-search-requests.js#L251)

These generated queries are then executed one level higher in [shared-libs/search/src/search.js](https://github.com/medic/cht-core/blob/master/shared-libs/search/src/search.js#L124-L129)This is then imported and ran by:

* the contact exporter in the API: [api/src/services/export/contact-mapper.js](https://github.com/medic/cht-core/blob/master/api/src/services/export/contact-mapper.js)
[...]

PR #9308, opened 2024-08-08
feat(#9293): refactor freetext search views

m5r:
# Description

#9293 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9293-research-freetext-views/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9293-research-freetext-views/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9293-research-freetext-views/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/ddocs/medic-db/medic-client/views/contacts_by_freetext/map.js
+++ b/ddocs/medic-db/medic-client/views/contacts_by_freetext/map.js
@@ -20,7 +20,9 @@ function(doc) {
     }
   };
 
-  var include = ['name', 'case_id', 'place_id', 'patient_id', 'external_id', 'house_number', 'notes'];
+  var include = [
+    'name', 'case_id', 'place_id', 'patient_id', 'external_id', 'house_number', 'notes', 'search_keywords',
+  ];
   var types = ['district_hospital', 'health_center', 'clinic', 'person'];
   var idx;
   if (doc.type === 'contact') {
--- a/ddocs/medic-db/medic-client/views/contacts_by_freetext/map.js
+++ b/ddocs/medic-db/medic-client/views/contacts_by_freetext/map.js
@@ -1,5 +1,5 @@
 function(doc) {
-  var skip = [ '_id' ];
+  var include = [ 'name', 'external_id', 'notes' ];
 
   var usedKeys = [];
   var emitMaybe = function(key, value) {
@@ -16,7 +16,7 @@ function(doc) {
       return;
     }
     key = key.toLowerCase();
-    if (skip.indexOf(key) === -1 || /_date$/.test(key)) {
+    if (include.indexOf(key) === -1 || /_date$/.test(key)) {
       return;
     }
     if (typeof value === 'string') {
--- a/ddocs/medic-db/medic-client/views/contacts_by_freetext/map.js
+++ b/ddocs/medic-db/medic-client/views/contacts_by_freetext/map.js
@@ -20,7 +20,7 @@ function(doc) {
     }
   };
 
-  var include = ['name', 'external_id', 'notes']; // TODO: add the other fields
+  var include = ['name', 'case_id', 'place_id', 'patient_id', 'external_id', 'house_number', 'notes'];
   var types = ['district_hospital', 'health_center', 'clinic', 'person'];
   var idx;
   if (doc.type === 'contact') {
--- a/ddocs/medic-db/medic-client/views/contacts_by_freetext/map.js
+++ b/ddocs/medic-db/medic-client/views/contacts_by_freetext/map.js
@@ -1,36 +1,29 @@
 function(doc) {
-  var include = [ 'name', 'external_id', 'notes' ];
-
-  var usedKeys = [];
-  var emitMaybe = function(key, value) {
-    if (usedKeys.indexOf(key) === -1 && // Not already used
-        key.length > 2 // Not too short
-    ) {
-      usedKeys.push(key);
-      emit([key], value);
-    }
-  };
-
   var emitField = function(key, value, order) {
     if (!key || !value) {
       return;
     }
+
     key = key.toLowerCase();
-    if (include.indexOf(key) === -1 || /_date$/.test(key)) {
-      return;
-    }
+
     if (typeof value === 'string') {
       value = value.toLowerCase();
       value.split(/\s+/).forEach(function(word) {
-        emitMaybe(word, order);
+        if (word.length <= 2) {
+          return;
+        }
+
+        emit([word], order);
       });
     }
+
     if (typeof value === 'number' || typeof value === 'string') {
-      emitMaybe(key + ':' + value, order);
+      emit([key + ':' + value], order); // TODO: only `case_id`
     }
   };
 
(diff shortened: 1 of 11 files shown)


m5r:
@jkuester could you give this a brief look when you have time? I made some changes in the view to simplify the code (like iterating over the `include` keys instead of each document object keys) and I'd like a sanity check to make sure the numbers I shared in the issue are coherent

ddocs/medic-db/medic-client/views/contacts_by_freetext/map.js
```diff
@@ -41,12 +37,14 @@ function(doc) {
...
+  var isContactOrPlace = idx !== -1;
+  if (isContactOrPlace) {
     var dead = !!doc.date_of_death;
     var muted = !!doc.muted;
     var order = dead + ' ' + muted + ' ' + idx + ' ' + (doc.name && doc.name.toLowerCase());
-    Object.keys(doc).forEach(function(key) {
+    include.forEach(function(key) {
       emitField(key, doc[key], order);
     });
+    emitMaybe('case_id:' + doc.case_id, order);
```
jkuester, line 48: Minor, but I don't think this is strictly required since the only programmatic `case_id:` search is for the Reports tab. So technically it should not hit this view.  I am not opposed to keeping it for the sake of consistency, but I guess we don't actually need it....

ddocs/medic-db/medic-client/views/contacts_by_type_freetext/map.js
```diff
@@ -43,12 +39,15 @@ function(doc) {
...
+  var isContactOrPlace = idx !== -1;
+  if (isContactOrPlace) {
     var dead = !!doc.date_of_death;
     var muted = !!doc.muted;
     var order = dead + ' ' + muted + ' ' + idx + ' ' + (doc.name && doc.name.toLowerCase());
-    Object.keys(doc).forEach(function(key) {
+    include.forEach(function(key) {
       emitField(type, key, doc[key], order);
     });
+    emitMaybe('case_id:' + doc.case_id, order);
```
jkuester, line 51: Same here.  Probably not required that we index the `case_id:` for contacts...

ddocs/medic-db/medic-client/views/reports_by_freetext/map.js
```diff
@@ -1,10 +1,13 @@
 function(doc) {
-  var skip = [ '_id', '_rev', 'type', 'refid', 'content' ];
+  if (doc.type !== 'data_record' || !doc.form) {
+    return;
```
jkuester, line 3: Just ruined the great nesting pattern we had going here.... :sweat_smile:

ddocs/medic-db/medic-client/views/reports_by_freetext/map.js
```diff
@@ -15,32 +18,33 @@ function(doc) {
...
-    if (typeof value === 'number' || typeof value === 'string') {
-      emitMaybe(key + ':' + value, reportedDate);
-    }
   };
 
-  if (doc.type === 'data_record' && doc.form) {
-    Object.keys(doc).forEach(function(key) {
-      emitField(key, doc[key], doc.reported_date);
-    });
+  var include = ['patient_id', 'patient_uuid', 'place_id', 'case_id', 'patient_name'];
```
jkuester, line 30: > `patient_uuid`

Is there a specific reason to include this field? In the discussion on the issue thread, it was noted that there is probably minimal value in indexing UUID values (since who is going to actually search for those)

ddocs/medic-db/medic-client/views/reports_by_freetext/map.js
```diff
@@ -15,32 +18,33 @@ function(doc) {
...
-      });
-    }
-    if (doc.contact && doc.contact._id) {
-      emitMaybe('contact:' + doc.contact._id.toLowerCase(), doc.reported_date);
+      emitField(key, doc.fields[key], doc.reported_date);
     }
+  });
+
+  if (doc.contact && doc.contact._id) {
+    emitMaybe('contact:' + doc.contact._id.toLowerCase(), doc.reported_date);
```
jkuester, line 40: This is an interesting case I had not noticed before....  :thinking: [Looks like](https://github.com/medic/cht-core/commit/b4cd9a541ee076e140700fe9302ae043f220692b) this is to support the ancient ANC workflow Gareth mentioned [here](https://github.com/medic/cht-core/issues/9293#issuecomment-2311658854).  I cannot see any place this is currently used in the code (but after my adventures with the `case_id`, who knows...).  

I guess maybe it is safer to leave it... :shrug:

ddocs/medic-db/medic-client/views/reports_by_freetext/map.js
```diff
@@ -15,32 +18,33 @@ function(doc) {
...
+  }
+
+  if (doc.case_id) {
+    emitMaybe('case_id:' + doc.case_id, doc.reported_date);
```
jkuester, line 44: Looks like the `emitMaybe` function does not check the `value` parameter at all.  Do we care?  What happens if we `emit` an `undefined` value?  Maybe it would be better to still call the `emitField` function for all these special cases...

jkuester (commented):
Sorry I missed this last week!

ddocs/medic-db/medic-client/views/contacts_by_freetext/map.js
m5r: Right, let's keep those views lean then :+1:

ddocs/medic-db/medic-client/views/reports_by_freetext/map.js
m5r: You're right, I wrongly assumed `patient_id` and `patient_uuid` represented the same thing :sweat_smile:

ddocs/medic-db/medic-client/views/reports_by_freetext/map.js
m5r: We already had this bit before that emitted `doc.reported_date`
```ts
if (doc.contact && doc.contact._id) {
  emitMaybe('contact:' + doc.contact._id.toLowerCase(), doc.reported_date);
}
```
It worked this far so I think it's safe to assume `doc.reported_date` is always defined and we can use `emitMaybe()` for `case_id` as well

jkuester: :facepalm: Looking back and this now, I am not sure exactly what I was thinking when I left that comment...  You are correct that the logic seems good!

m5r:
I rebased the branch off `master` to fix [this weird CI issue](https://github.com/medic/cht-core/actions/runs/10995508967/job/30526481940#step:11:1358) but I kept a snapshot of the branch rebased off `4.10.x` in the branch `9293-freetext-views-4.10`

Issue #9193, opened 2024-06-21
API Endpoints for getting contacts by type


jkuester:
**Is your feature request related to a problem? Please describe.**
As noted in https://github.com/medic/cht-core/issues/8889, getting the contacts of a specific type is needed for cht-user-managment.  

**Describe the solution you'd like**

```
HTTP GET api/v1/person

HTTP GET api/v1/place?type=clinic
```

Endpoints should also accept the following query parameters to allow for paging through the data: 
- `limit` - default is `100`
- `skip` - default `0`

When it comes to the cht-datasource api, things will be a bit interesting.  On one hand, _most_ consumers of the lib will not want to manually deal with paging, but instead we should return something like an `async Iterable` so the paging it completely abstracted away.  However, the api code will need to be able to make a targeted request for a specific limit/skip since that is what it will be receiving as a request and for best performance we will need to be able to pass those though directly to the Couch view query... Perhaps we just provide both functions in cht-datasource:

- One that takes `limit/skip` and returns an array of values.
- One that just returns an Iterable - Under the hood, this would probably just use the first function, but wrap it up with a generator function or something.

**Describe alternatives you've considered**
Other ways to structure the API include:
- `api/v1/clinic`
- `api/v1/place/clinic`

However, these do not seem any more clear than just using a query parameter and are definitely less flexible.

sugat009:
After a bit of assessment, I've decided to break this task down into smaller tasks in the following way.
1. Add functionality of getting people with pagination in the `cht-datasource`.
2. Add functionality of getting people as an `async Iterable` in the `cht-datasource`.
3. Add functionality of getting places with pagination in the `cht-datasource`.
4. Add functionality of getting places as an `async Iterable` in the `cht-datasource`.
5. Add API `api/v1/person`
6. Add API `api/v1/place?type=clinic`

_TODO: Create tickets for these._

sugat009:
I will be pushing each sub-task changes to this branch `9193-api-endpoints-for-getting-contacts-by-type`.

PR #9311, opened 2024-08-12
feat(#9193): add functionality of getting people as pages or async iterables in cht-datasource

sugat009:
# Description
Added functionality of getting people as using pages or as `AsyncGenerator` using `cht-datasource` along with addition of REST endpoint `/api/v1/person` which provides people as pages.

Issues: #9237 , #9238 , #9241 

Usage:
1. Pagination API
```typescript
Person.v1.getPageByType(Qualifier.byContactType(personType), cursor, limit); // limit and cursor are optional with defaults as 100 and "0" respectively
```
2. Async Generator API
```typescript
const getAllIterator = Person.v1.getByType(ctx)(Qualifier.byContactType('person'));
```
3. REST API endpoint
```
HTTP GET api/v1/person
```
Query Params:
- personType - the `person_type` to fetch
- limit - default is 100
- cursor - default is 0

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9193-api-endpoints-for-getting-contacts-by-type/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9193-api-endpoints-for-getting-contacts-by-type/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9193-api-endpoints-for-getting-contacts-by-type/docker-compose/cht-couchdb-clustered.yml)

# License


sugat009:
The CI is failing for a param's JSDoc which exists but ESLint is marking it as does not exist. 
https://github.com/medic/cht-core/actions/runs/10352638921/job/28653693985?pr=9311#step:11:88
Looking into it right now.
This might have been due to the upgrade of ESLint as this was passing before.

jkuester (commented):
This is looking great!  Today I was able to make it through the cht-datasource implementation code (where I left some minor comments, mostly related to the null/undefined cursor discussion from Slack).  Tomorrow I will continue with the rest!

shared-libs/cht-datasource/src/index.ts
```diff
@@ -85,6 +86,29 @@ export const getDatasource = (ctx: DataContext) => {
...
+         * @param personType the type of people to return
+         * @param cursor a value representing the index of the first person to return
+         * @param limit the maximum number of people to return. Default is 100.
+         * @returns an array of people for the provided page specifications
+         * @throws Error if no type is provided or if the type is not for a person
+         * @throws Error if the provided limit is `<= 0`
+         * @throws Error if the provided cursor is `< 0`
+         * @see {@link getByType} which provides the same data, but without having to manually account for paging
+         */
+        getPageByType: (personType: string, cursor = '0', limit = 100) => ctx.bind(Person.v1.getPage)(
```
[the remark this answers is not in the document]
sugat009: Let's go with this implementation:
```typescript
getPageByType: (personType: string, cursor: Nullable<string>, limit = 100)
```

tests/integration/api/controllers/person.spec.js
```diff
@@ -104,4 +112,111 @@ describe('Person API', () => {
...
+          ...userNoPerms.contact,
+          ...onlineUserPlaceHierarchy
+        },
+        {
+          type: personType,
+          ...offlineUser.contact,
+          ...offlineUserPlaceHierarchy
+        }
+      ];
+      const responsePage = await getPage(Qualifier.byContactType(personType), cursor, limit);
```
jkuester, line 155:
```suggestion
      const responsePage = await getPage(Qualifier.byContactType(personType));
```

I don't want to get too deep into edge cases with these integration tests, but can we have this test cover the case where we use the default cursor/limit values.  Then add another test where we call `getPage` twice.  On the first call we give it a `null` cursor and a limit of `4`.  Then on the second call, we pass in the cursor value returned from the first call (and again a limit of `4`).  Then we can assert that between the two calls, we got the expected 7 persons back.  One of the key things I want to be careful to assert in the new test is the _value_ returned for the page cursor.  Should be `'4'` on the first call and then `null` on the second.

jkuester (changes requested):
This is great!  I have made it though the rest of the implementation code and all the tests.  Just a number of nitpicks thoughout and a few logic tweaks, but otherwise this is nearly complete.

tests/integration/api/controllers/person.spec.js
jkuester: One more case that I thought of that we should probably have is to call the `Person.v1.getAll` function and load up all the returned persons and make sure we get all 7 of them.  Just a safety check that our generator code integrates fine.

shared-libs/cht-datasource/test/index.spec.ts
```diff
@@ -124,6 +125,44 @@ describe('CHT Script API - getDatasource', () => {
...
+        const returnedPeople = await person.getPageByType(personType, cursor, limit);
+
+        expect(returnedPeople).to.equal(expectedPeople);
+        expect(dataContextBind.calledOnceWithExactly(Person.v1.getPage)).to.be.true;
+        expect(personGetPage.calledOnceWithExactly(personTypeQualifier, cursor, limit)).to.be.true;
+        expect(byContactType.calledOnceWithExactly(personType)).to.be.true;
+      });
+
+      it('getByType', async () => {
+        // eslint-disable-next-line @typescript-eslint/require-await
```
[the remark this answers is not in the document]
sugat009: That didn't work.

image

I've added a fake `Promise.resolve()` to solve this. So, the function looks like:
```typescript
const mockAsyncGenerator = async function* () {
  await Promise.resolve();
  yield [];
};
```

shared-libs/cht-datasource/test/person.spec.ts
```diff
@@ -123,5 +126,162 @@ describe('person', () => {
...
+        await expect(Person.v1.getPage(dataContext)(invalidQualifier, cursor, limit))
+          .to.be.rejectedWith(`Invalid contact type [${JSON.stringify(invalidQualifier)}].`);
+
+        expect(assertDataContext.calledOnceWithExactly(dataContext)).to.be.true;
+        expect(adapt.calledOnceWithExactly(dataContext, Local.Person.v1.getPage, Remote.Person.v1.getPage)).to.be.true;
+        expect(isContactTypeQualifier.calledOnceWithExactly(invalidQualifier)).to.be.true;
+        expect(getPage.notCalled).to.be.true;
+      });
+
+      it('throws an error if limit is invalid', async () => {
```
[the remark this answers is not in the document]
sugat009: I've used everything except `undefined` here as it will trigger the default argument of the function.

api/src/controllers/person.js
```diff
@@ -8,20 +8,35 @@ const getPerson = ({ with_lineage }) => ctx.bind(
...
       }
       return res.json(person);
-    })
-  }
+    }),
+    getAll: serverUtils.doOrError(async (req, res) => {
+      await checkUserPermissions(req);
+
+      const personType  = Qualifier.byContactType(req.query.personType);
+      const limit = Number(req.query.limit) || 100;
```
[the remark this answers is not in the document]
sugat009: We should probably be forwarding anything that is passed here and not `undefined` because if `null` or `false` is passed here then, `undefined` is passed into `cht-datasource` instead which is probably not what we want.

```javascript
const limit = req.query.limit ? Number(req.query.limit) : req.query.limit;
```

tests/integration/api/controllers/person.spec.js
```diff
@@ -2,13 +2,16 @@ const utils = require('@utils');
 const placeFactory = require('@factories/cht/contacts/place');
 const personFactory = require('@factories/cht/contacts/person');
 const { getRemoteDataContext, Person, Qualifier } = require('@medic/cht-datasource');
+const chai = require('chai');
 const { expect } = require('chai');
+const deepEqualInAnyOrder = require('deep-equal-in-any-order');
 const userFactory = require('@factories/cht/users/users');
+chai.use(deepEqualInAnyOrder);
```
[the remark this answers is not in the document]
sugat009: Turns out integrations tests use this [mocha config file](https://github.com/medic/cht-core/blob/master/tests/integration/.mocharc-base.js) instead. I've added it there.

tests/integration/api/controllers/person.spec.js
sugat009: > I don't want to get too deep into edge cases with these integration tests, but can we have this test cover the case where we use the default cursor/limit values. Then add another test where we call `getPage` twice. On the first call we give it a `null` cursor and a limit of `4`. Then on the second call, we pass in the cursor value returned from the first call (and again a limit of `4`). Then we can assert that between the two calls, we got the expected 7 persons back. One of the key things I want to be careful to assert in the new test is the _value_ returned for the page cursor. Should be `'4'` on the first call and then `null` on the second.

The reason I kept the integration test with the "non-paging" values for `limit` and `cursor` is that the objects being returned from the database are not in a consistent order for the order that we are passing into the `utils.saveDocs()` function. In our case, we pass `const allDocItems = [contact0, contact1, contact2, place0, place1, place2, patient];` into `utils.saveDocs` but it is not guaranteed that contacts will be returned in the same order. The order is undeterministic and random; it could be the order in which they were indexed into the view, so the page cannot be compared. For example, if I request a page with 4 items, it could be any 4 items.

shared-libs/cht-datasource/test/index.spec.ts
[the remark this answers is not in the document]
jkuester: :sweat_smile: Sorry, bad explaining on my part!  What I was trying to suggest was:

```ts
        const mockAsyncGenerator = async function* () {
          yield await Promise.resolve([]);
        };
```

It is all basically the same functionally, but that code seems the least janky.

api/src/controllers/person.js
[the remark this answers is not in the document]
jkuester: Good call!  Yes, you code is precisely what we want.

tests/integration/api/controllers/person.spec.js
jkuester: > The order is undeterministic and random; it could be the order in which they were indexed into the view

Minor, but I just want to clarify that order the data is _returned_ should not be random.  My understanding is that data returned from couch views is deterministically ordered based on the _key value_ and in cases like this where we have many results for the same key, the order is by `_id` value.  Each call to `getPage` with the same parameters, should return the same set of data (otherwise our skip-based paging would not work :sweat:).  However, this does not help us much for these integration tests since, as you have observed, the UUID's assigned as the `_id` value for our new contacts _are random_ and so, the order of the returned data could change between test runs.  

All that being said, I don't think we need to worry too much about trying to assert the _order_ of the data returned.  We don't even need to assert the _exact docs_ returned for each page.  For the test where we get multiple pages, it seems enough to just assert the size of the `data` array for each page.  Then at the end combine the `data` arrays into one array that contains all the contacts and do your `equalInAnyOrder` assertion to make sure that ultimately we got back all the contacts we expected.

shared-libs/cht-datasource/test/local/person.spec.ts
```diff
@@ -228,6 +228,7 @@ describe('local person', () => {
     describe('getPage', () => {
```
jkuester, line 228: Do we need to add some test cases for throwing an exception when there is an invalid cursor value?  (The new logic you added in `local/person.ts`).

tests/integration/api/controllers/person.spec.js
```diff
@@ -133,33 +130,44 @@ describe('Person API', () => {
...
 
+    it('returns a page of people when limit and cursor is passed and cursor can be reused', async () => {
+      const firstPage = await getPage(Qualifier.byContactType(personType), cursor, limit);
+      const secondPage = await getPage(Qualifier.byContactType(personType), firstPage.cursor, limit);
+
+      const allPeople = [...firstPage.data, ...secondPage.data];
+
+      expect(allPeople).excludingEvery(['_rev', 'reported_date']).to.deep.equalInAnyOrder(expectedPeople);
+      expect(firstPage.cursor).to.be.equal('4');
+      expect(secondPage.cursor).to.be.equal(null);
```
jkuester, line 168: Minor, but can you also assert the `data.length` for each page?  Just want to confirm that we got `4` back on the first page and `3` on the second.

tests/integration/cht-datasource/person.spec.js
```diff
@@ -0,0 +1,43 @@
...
+  before(async () => {
+    await utils.saveDocs(allDocs);
+  });
+
+  after(async () => {
+    await utils.revertDb([], true);
+  });
+
+  describe('v1', async () => {
+    describe('getAll', async () => {
```
jkuester, line 28: Good test! Since this is technically still just calling `GET /api/v1/person` underneath the hood, I would vote we just add this test case into the `'GET /api/v1/person'` describe block of the `controllers/person.spec.ts` file (next to the rest of the new test cases).  It is true that it is one more step removed from the controller, but it really is not actually adding any additional functionality besides abstracting away the paging logic of the REST calls...

If we had a bunch of other tests to add here that did not really fit anywhere else, that would be one thing, but as things stand now, I think we can just keep it simple.  Happy to discuss more if you all see it differently.  @sugat009 @tatilepizs

jkuester (approved):
Left a couple other minor comments around the tests, but otherwise this is ready to go!  

One thing we do not want to forget, @sugat009, is that we need to update the [api reference docs](https://docs.communityhealthtoolkit.org/apps/reference/api/) to include documentation for the new REST endpoint.  [Here](https://github.com/medic/cht-docs/pull/1423) is one of the previous PRs for example.

tests/integration/cht-datasource/person.spec.js
sugat009: I'm ok with any as technically it's still calling the API under the hood, however, to keep things separated the current one would be the way to go but then again even for `Person.v1.get` or `Person.v1.getPage` we technically aren't directly calling the REST API endpoints so moving this also makes sense.
Let's move this test there for now and if things tend to branch out more we'll move it back out.

shared-libs/cht-datasource/test/local/person.spec.ts
sugat009: ahh. yes. good catch.

jkuester (approved):

PR #9305, opened 2024-08-06
chore(143): e2e tests for aggregate targets (place, period)

tatilepizs:
# Description

This PR will cover the following scenarios:

- Verify that the aggregate filter is displayed
- Verify that when a user with just one place is logged in, the aggregate filter only shows the "Reporting Period" section
- Verify that when a user with two places is logged in, the aggregate filter shows the "Place" and "Reporting Period" sections
- Verify that the filter would show, by default, the first place and the monthly period with the correct data in the aggregate target list and detail view and that the data changes when the filter changes.

Task from EPIC - [#143](https://github.com/medic/care-teams/issues/143)
Started work from [chore(#9231): filter aggregate targets by period](https://github.com/medic/cht-core/pull/9283)

This PR is the replacement of the PR [chore(143): e2e tests for aggregate targets](https://github.com/medic/cht-core/pull/9273), that PR was closed because it was missing the changes of the "Reporting Period" section.

# Code review checklist

- [ ] Readable: Concise, well named, follows the [style guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/), documented if necessary.
- [ ] Documented: Configuration and user documentation on [cht-docs](https://github.com/medic/cht-docs/)
- [ ] Tested: Unit and/or e2e where appropriate
- [ ] Internationalised: All user facing text
- [ ] Backwards compatible: Works with existing data and configuration or includes a migration. Any breaking changes documented in the release notes.

# Compose URLs
If Build CI hasn't passed, these may 404:

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:e2e-tests-filter-aggregate-targets/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:e2e-tests-filter-aggregate-targets/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:e2e-tests-filter-aggregate-targets/docker-compose/cht-couchdb-clustered.yml)

# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

PR #9283, opened 2024-07-24
chore(#9231): filter aggregate targets by period

Benmuiruri:
# Description

After the completion of https://github.com/medic/cht-core/pull/9267 which implements the actual filtering of the Aggregate Targets through the selection of the facilities in the radio buttons. 

This PR adds the feature of filtering by reporting period. This filtering will be available for all supervisors (With one facility or with multi-facility). 

<details>
<summary>Video demo and details of expected content on filtering</summary>

### The first half of the video shows user with multi-facilities

- Toggling the Health Facility on the sidebar changes the label of the LHS list from `Ben's Health Facility` to `Second Facility`
Screenshot 2024-08-06 at 12 04 39
- Toggling the Reporting Period on the sidebar changes the label of the LHS list from `Ben's Health Facility` to `Ben's Health Facility - {{ Previous month }}` when you select `Last month`
Screenshot 2024-08-06 at 12 06 41

### The second hald of the video shows user with one facility 

- When user has one facility, it does not show the facility name on the LHS and we don't show the Health Facility filter
- When user has one facility, it shows the  `{{ Previous month }}` on the LHS when you select `Last month`

https://github.com/user-attachments/assets/2616ddc6-3870-4f91-b566-b02b41120b2a

Mobile View 

https://github.com/user-attachments/assets/547c170b-6d72-4398-9dee-ad0a6c433127

</details>

# Code review checklist

- [ ] Readable: Concise, well named, follows the [style guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/), documented if necessary.
- [ ] Documented: Configuration and user documentation on [cht-docs](https://github.com/medic/cht-docs/)
- [ ] Tested: Unit and/or e2e where appropriate
- [ ] Internationalised: All user facing text
- [ ] Backwards compatible: Works with existing data and configuration or includes a migration. Any breaking changes documented in the release notes.

# Compose URLs
If Build CI hasn't passed, these may 404:

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:filter-aggregate-period/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:filter-aggregate-period/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:filter-aggregate-period/docker-compose/cht-couchdb-clustered.yml)

# License

[...]

PR #9317, opened 2024-08-13
feat(#9231): add filter in aggregate targets

### LINKED ISSUE #9231: Allow users to view aggregate targets and filter by facility_id and reporting period

Benmuiruri:
# Description

This PR combines 5 PRs to allow: 
- Filtering Aggregate Targets based on facility and reporting period for users assigned to multiple areas.
- Filtering Aggregate Targets based on reporting period for users assigned to one area.

The PRs being combined are: 
1. [Enable place filter aggregate targets for users with multiple](https://github.com/medic/cht-core/pull/9232)
2. [Filter target aggregate data using sidebar filter](https://github.com/medic/cht-core/pull/9267)
3. [Add facility indicator in aggregates targets](https://github.com/medic/cht-core/pull/9282)
4. [Filter aggregate targets by period](https://github.com/medic/cht-core/pull/9283)
5. [e2e tests for aggregate targets (place, period)](https://github.com/medic/cht-core/pull/9305)

For details about what each PR does, please check the description of the individual PRs. 

Closes #9231 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9231-add-aggregate-filter/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9231-add-aggregate-filter/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9231-add-aggregate-filter/docker-compose/cht-couchdb-clustered.yml)

# License


webapp/src/ts/services/analytics-modules.service.ts
```diff
@@ -3,6 +3,8 @@ import { Injectable } from '@angular/core';
 import { SettingsService } from '@mm-services/settings.service';
 import { TargetAggregatesService } from '@mm-services/target-aggregates.service';
 
+export const AGGREGATE_TARGETS_ID: string = 'target-aggregates';
```
Benmuiruri, line 6: Inferred type

```suggestion
export const AGGREGATE_TARGETS_ID = 'target-aggregates';
```

Benmuiruri (commented):
I tested these scenarios 

- Does undefined break anything ? ✅
- If 2025 jan check previous ✅
- Check leap year. ✅
- Confirm admin. ✅
- Online vs offline users ✅
- place with very long name ✅

I didn't find anything apart from those two small code suggestions.

latin-panda:
I tested:
- Reports filter should work and look the same as we had before ✅
- User with one facility assigned - the facility_id is a string (backward compatibility) ✅
- User with two facilities assigned ✅
- Aggregate targets should reset the filter and work fine if there is an app update and the user reloads the app. ✅
- Aggregate targets should reset the filter and work fine if the user clicks a back button or navigates back and forth from one page to another.  ✅
- Admin user gets expected error when trying to access aggregate targets ✅
- Mobile view and Desktop view looking according to Figma  ✅
- We have a test server with a config from a partner and it's working. ✅

webapp/src/ts/services/target-aggregates.service.ts
```diff
@@ -233,23 +255,28 @@ export class TargetAggregatesService {
     if (!facility_id?.length) {
```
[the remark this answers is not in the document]
Benmuiruri: That's a good idea but I would suggest we just leave it as getUserFacilityIds. 
- If we change to using the new this.userSettingsService.getUserFacilities, we have to map the result to extract the id from the facility objects.... we just need id, not whole facility
- We will need to update all tests in `webapp/tests/karma/ts/services/target-aggregates.service.spec.ts` that use `userSettingsService.get`
- We still have `userSettingsService.get` as an existing function even if we dont use it here. 

Therefore for the scope of this task I would say lets leave it.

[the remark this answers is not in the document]
latin-panda: Sorry, I didn't understand this one:

> We still have userSettingsService.get as an existing function even if we dont use it here.

This is the only place we use it in this file. Can you please explain a bit more what do you mean by `We still have userSettingsService.get as an existing function`?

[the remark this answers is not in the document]
latin-panda: I mean changing this function to this, which is 1 line
```
const facility_ids: string[] = (await this.userSettingsService. getUserFacilities()).map(facility => facility._id);
```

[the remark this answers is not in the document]
latin-panda: This comment is about updating just this file `target-aggregates.service.ts`, the `getUserFacilityIds()` function only. We don't need to update other files.

tatilepizs (commented):
I was doing some exploratory testing around Reports, Contacts, and Targets, and two things caught my attention.

1. If I have a place with a really long name and I apply the period filter, the month is not going to be shown in list view. Is that expected? Wouldn't be a better idea to cut the place name in way that the last month is always shown no matter how big the name of place is?

[video](https://github.com/user-attachments/assets/1ee28416-60f2-48ef-b432-8037f3039f55)

2. I created a user with 3 places, at the area level, and everything is working fine, but in the reports tab, the place filter is showing the health facility level as `unavailable`

image

Everything else looks perfectly 😊

tests/e2e/default/targets/target-aggregates.wdio-spec.js
```diff
@@ -109,317 +49,309 @@ describe('Target aggregates', () => {
...
       await commonPage.logout();
       await utils.revertDb(DOCS_TO_KEEP, true);
     });
 
-    it('should disable content when user has many facilities associated', async () => {
-      await loginPage.login({ password: userWithManyPlacesPass, username: userWithManyPlaces.name });
-      await commonPage.waitForPageLoaded();
-      await targetAggregatesPage.checkContentDisabled();
-    });
+    describe('Online user with one place associated', () => {
```
lorerod, line 125: Is it the same behavior for an `offline` user with only one place associated?
If it is not, we should add the scenario.

tests/e2e/default/targets/target-aggregates.wdio-spec.js
```diff
@@ -109,317 +49,309 @@ describe('Target aggregates', () => {
...
+        await browser.back();
+        const secondTargetItem =
+          await (await targetAggregatesPage.getTargetItem(expectedTargets[1], CURRENT_PERIOD, districtHospital1.name));
+        await helperFunctions.assertTitle(secondTargetItem.title, expectedTargets[1].title);
+      });
+    });
 
-      await commonPage.goToAnalytics();
-      await targetAggregatesPage.goToTargetAggregates(true);
+    describe('Offline user with multiple places associated', () => {
```
lorerod, line 290: Same question here. Is it the same behavior for `online` users with multiple places associated?

lorerod (commented):
I only review the e2e test part of the code. Great work, @tatilepizs. Thank you for completing this work. I left some comments related to the scenarios covered.

tests/e2e/default/targets/target-aggregates.wdio-spec.js
tatilepizs: Yes it is. The different behavior depends on whether the user has one or more than one place associated.

latin-panda:
@tatilepizs Thanks for exploring the feature more! 

> I created a user with 3 places, at the area level, and everything is working fine, but in the reports tab, the place filter is showing the health facility level as unavailable

CHT expects the facility to have a `name` property in the doc ([see the condition in code](https://github.com/medic/cht-core/blob/f40feb1d9ac2894ec138f2ec887d839f19e2270d/webapp/src/ts/components/filters/facility-filter/facility-filter.component.ts#L240)). When it doesn't have it and user is online, CHT displays `[deleted]`. When the user is offline, it displays `[unavailable]`. I guess this is to handle edge cases, to allow users to keep using the filter (like selecting descendants or seeing what's associated with that unknown place?) 

This is a component only used in Reports page.

tests/e2e/default/targets/target-aggregates.wdio-spec.js
latin-panda: Correct. It's based on associated places to a user

Benmuiruri: Yes Tatiana you are correct. The difference in behavior is whether user has one place or more than one place.

tests/e2e/default/targets/target-aggregates.wdio-spec.js
Benmuiruri: Yes, you are correct. It is the same behavior for online and offline users

latin-panda:
> If I have a place with a really long name and I apply the period filter, the month is not going to be shown in the list view. Is that expected? Wouldn't it be a better idea to cut the place name in a way that the last month is always shown no matter how big the name of the place is?

@tatilepizs This list uses the "breadcrumbs", the same component as Reports, Contacts, and Tasks. It was designed to show 2 lines of text, then cut it with an ellipsis, so small screens can fit more items in the list (instead of pushing down with long text). 
I tried to add the ellipsis in each section of the breadcrumb, but, users will see less text. It can't go in 2 lines and put the ellipsis at the same time because the css `display` conflicts. 

Screenshot 2024-08-14 at 12 31 43 PM
<br>

We prefer to leave it as it is today, because, it fits more text (reasonably long names) and the reporting period still fits. So,  this as an edge case. 

Screenshot 2024-08-14 at 12 57 22 PM

<br>
Users can still click on a target, and see the reporting period in the details view. 
Screenshot 2024-08-14 at 12 59 28 PM

latin-panda (approved):
@Benmuiruri this looks good to me!

Benmuiruri:
Thanks everyone for chipping in to get this over the line.

Issue #9339, opened 2024-08-16
Phone Widget display formatted phone number value


jkuester:
**Is your feature request related to a problem? Please describe.**
Currently the logic for the custom phone widget for collecting phone numbers in forms, will [automatically format](https://github.com/medic/cht-core/blob/1b9b6527c77d8668ca4963ddc7a2d5e6aa6b7273/webapp/src/js/enketo/widgets/phone-widget.js#L54) the number value before saving the data into the form's model.  However, the value _displayed to the user_ in the form does not change (the input text box still just contains the exact value entered by the user).  

**Describe the solution you'd like**
More UX consideration here is needed to decide what would be most helpful (and least surprising) to users, but the original phone widget code included a TODO that suggested updating the displayed phone number value in the form to also be the formatted value stored in the model.  

**Describe alternatives you've considered**
One reason we may not want to do this is because it would result in the user seeing the value change _after they enter it_.  This could result in user confusion as to what happened with the value the entered.  (Note that the only automatic changes that could happen to the value would be with regards to its formatting. The actual number, itself, would not be automatically updated.)

**Additional context**
I am logging this issue now for better visibility and discussion of the feature. I plan to remove the TODO comment from the code.

Abhishek9639:
Hi @jkuester,
I’d like to work on this issue. The behavior around formatting and display looks interesting, and I’d be happy to explore a solution.
Could you please assign this to me?
Thanks

Abhishek9639:
Thanks for assigning this issue to me. I'll start working on it.

Issue #7514, opened 2022-02-11
Error message on duplicate phone numbers doesn't say it's a duplicate


mrjones-plip:
**Describe the bug**
If you try to create a contact with a phone number that is already in use by another contact, the error message doesn't tell you it's a duplicate.

**To Reproduce**
1. Create a contact in the CHT app (not admin app) with the phone number `+254712345678`
2. Try to create a second contact with the same phone number `+254712345678`
3. See the input validation error that says
   > Please enter a valid local number, or use the standard international format, which includes a plus sign (+) and country code. For example: +254712345678

**Expected behavior**
The input validation error should tell you that the phone number is in use by another contact.

**Logs**
NA

**Screenshots**

https://user-images.githubusercontent.com/8253488/153676369-7ec2a6c9-3bac-4ba2-b383-004ce99bab2c.mp4

**Environment**
- Instance:  local dev, off of CHT Docker Helper off of `medic-os`
- Browser: latest firefox
- Client platform:  Ubuntu 20.04
- App: webapp?
- Version: 3.14.0

**Additional context**
* I'm running the default app that ships with CHT
* @abbyad mentioned this was part of Enketo and not something in the default app, hence the bug against `cht-core` repo

abbyad:
Thanks for raising @mrjones-plip! I just noticed that improving the validation message has been mentioned [in this related issue](https://github.com/medic/cht-core/issues/6390#issuecomment-622514266), and can be done by updating the constraint message in the XForm to include something about the number needing to be "unique" as well.

mrjones-plip:
Cools! let's use this ticket then as an call to arms to update the default app.

mrjones-plip:
Good thinking!  For the future person, find the IDs of the forms to edit on google drive in the `forms-on-google-drive.json` file (see below).

Also - what about the wording? It gets awkward in a hurry as seen here with me mocking int up.:
![error wording](https://user-images.githubusercontent.com/8253488/153682952-4e4399e3-ac35-4795-9f7c-a01570841eac.png)

Finally - whoever takes this on should know that we need at least three translations of this: English, French and Nepali

```
cat forms-on-google-drive.json 
{
  "app/death_report.xlsx": "1azFHCMTMehxuSg_LWOgsJZxrJo0yhseLp9FK4gTT38M",
  "app/delivery.xlsx": "1KdJEp-iErhLtgILdrafIvK0vvkOsEGRKVlsWBB3ysmQ",
[...]

PR #9340, opened 2024-08-16
feat(#6390): update phone widget to allow not checking for dupes

### LINKED ISSUE #6390: Allow for phone number validation but also allow for duplicate phone numbers
Currently if an XForm uses our [Phone Widget](https://github.com/medic/cht-core/blob/master/webapp/src/js/enketo/widgets/phone-widget.js), this widget both validates the phone number's format and ensures that no two contacts have the same phone number.

MSF-Goma relies on an SMS-workflow of "shared phones". They wish to validate the input but allow for duplicate phone numbers.

Can we allow for validation but not uniqueness?

MSF-Goma team is currently working around this by using a text input without validation

jkuester:
# Description

Closes #6390

https://github.com/medic/cht-docs/pull/1498

This PR add support for a new way to configure a phone number field in a form.  Previously, the phone widget was configured in an xlsxform field by setting `type: tel`.  By default, the phone widget both validates the phone number value and checks for existing contacts that have the same number.  

With these new changes, the existing (now deprecated) functionality stays the same, but going forwards, phone number fields should be configured in an xlsxform by setting `type: string` with `appearance: number tel`.  Fields with this new configuration will always validate the phone number value, but by default they will _not_ check for existing contacts with the same number.  To enforce a unique phone number with the new style of field, you should add a new column with `instance::cht:unique_tel: true`

## Testing considerations

We had some existing coverage of the phone widget in the enekto_widgets tests both in `e2e` and `integration/cht-form`.  However, as we determined when we did the countdown timer feature, it seems more sustainable to break out tests/forms for specific widgets instead of trying to cram everything into the `enketo_widgets` tests.  So, I have created a new `phone_widget` form and added tests for that, both in `e2e` and `integration/cht-form`. I covered as much as possible in the cht-form tests, but part of the phone-widget functionality involves querying the database to check for other contacts with the same phone number.  This functionality cannot be properly validated in the cht-form tests, so I added a few cases to the `e2e` tests.

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6390_optional_dup_check_for_tel/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6390_optional_dup_check_for_tel/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6390_optional_dup_check_for_tel/docker-compose/cht-couchdb-clustered.yml)

# License


webapp/.eslintrc
```diff
@@ -14,7 +14,7 @@
   "overrides": [
     {
       "files": [
-        "src/js/bootstrapper/*.js"
+        "src/js/**/*.js"
```
jkuester, line 17: Seems like it is fine to let any of the js files log to the console. Right?

webapp/src/js/enketo/widgets/phone-widget.js
```diff
@@ -5,38 +5,49 @@ const $ = require( 'jquery' );
...
+  validate: ( fieldValue ) => window.CHTCore.Settings.get()
+    .then(settings => isContactPhoneValid(settings, fieldValue)),
+};
+FormModel.prototype.types.unique_tel = {
+  validate: ( fieldValue ) => window.CHTCore.Settings.get()
+    .then(settings => isContactPhoneValid(settings, fieldValue) && isContactPhoneUnique(settings, fieldValue)),
+};
+
+// Remove this when we no longer support the tel xlsxform type
+const deprecated = {
```
jkuester, line 46: Following a similar pattern we used in the countdown timer widget. I put all the deprecated functionality in this object, so when we remove support for the `tel` type, it is more clear what parts of the code need to be updated (basically just the places where we use this object).

webapp/src/js/enketo/widgets/phone-widget.js
```diff
@@ -45,13 +56,13 @@ FormModel.prototype.types.tel = {
    * @extends Widget
    */
 class PhoneWidget extends Widget {
-  constructor( element, options, Settings = window.CHTCore.Settings ) {
-    super(element, options);
-
+  _init() {
```
jkuester, line 59: The Enketo Widget class has [an _init function](https://github.com/enketo/enketo/blob/351944c491cd59c6a0574d40393a4bb9fca7f81e/packages/enketo-core/src/js/widget.js#L48) that is intended to be overridden to include initialization logic.  We use that method in our other custom widgets, so just cleaning up this old code to follow the new convention.

webapp/src/js/enketo/widgets/phone-widget.js
```diff
@@ -45,13 +56,13 @@ FormModel.prototype.types.tel = {
...
     const $input = $( this.element );
+    const $wrapper = $input.closest('.question');
+    const uniqueTel = $wrapper.attr('data-cht-unique_tel') === 'true' || deprecated.isDeprecated($wrapper);
+    $input.attr('data-type-xml', uniqueTel ? 'unique_tel' : 'tel');
 
     // Add a proxy input field, which will send its input, formatted, to the real input field.
-    // TODO(estellecomment): format the visible field onBlur to user-friendly format.
```
jkuester, line 54: https://github.com/medic/cht-core/issues/9339

webapp/src/js/enketo/widgets/phone-widget.js
```diff
@@ -5,38 +5,49 @@ const $ = require( 'jquery' );
...
+
+
+// Set up enketo validation for data types
 FormModel.prototype.types.tel = {
-  validate: function( fieldValue ) {
-    return window.CHTCore.Settings
-      .get()
-      .then( function( settings ) {
-        if ( !phoneNumber.validate( settings, fieldValue ) ) {
-          throw new Error( 'invalid phone number: "' + fieldValue + '"' );
```
jkuester, line 15: Looking at the implementation of the other [Enketo Type validations](https://github.com/enketo/enketo/blob/351944c491cd59c6a0574d40393a4bb9fca7f81e/packages/enketo-core/src/js/types.js), it seems that none of them are deliberately throwing an error.  Instead they just return `false` when the validation fails.  The main behavior difference that I could find between returning `false` and just throwing an error is that an `invalidated` event is not properly emitted for the input when an error is thrown.  

To remain more consistent with existing Enketo behavior, I have updated the validation logic to just return `false` instead of throwing errors.

tests/page-objects/default/contacts/contacts.wdio.page.js
```diff
@@ -205,12 +205,16 @@ const addPerson = async ({
...
+const openEditContactForm = async () => {
   await waitForContactLoaded();
   await commonPage.openMoreOptionsMenu();
   await (await menuSelectors.editContactButton()).waitForClickable();
   await (await menuSelectors.editContactButton()).click();
+};
+
+const editPerson = async (currentName, { name, phone, dob }) => {
+  await selectLHSRowByText(currentName);
+  await openEditContactForm();
```
tatilepizs, line 217: 🤩

tatilepizs (approved):
It looks good Josh, thank you for adding all of the new tests scenarios ⭐

webapp/src/js/enketo/widgets/phone-widget.js
```diff
@@ -5,38 +5,48 @@ const $ = require( 'jquery' );
...
+  }
 
-        const DB = window.CHTCore.DB;
-        return DB.get().query('medic-client/contacts_by_phone', { key: phoneNumber });
-      } )
-      .then( function( res ) {
-        if ( res.rows.length === 0 ) {
-          return true;
-        }
+  console.error( 'invalid phone number: "' + fieldValue + '"' );
```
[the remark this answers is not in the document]
jkuester: Good question.  Yes, this is logging to the console in prod.  The tricky part here is that currently there is no way without debugging into the code to know if the provided phone number validation failed because it is not a proper phone number for the current locale or if it is a duplicate phone number.  On the front-end, the app just shows a constraint failure, but the user does not know which it is.  The plan is to address this issue in https://github.com/medic/cht-core/issues/7514.  

With the previous implementation of this validation logic where an exception was thrown instead of returning `false`, the Enekto code would [log the exception message](https://github.com/enketo/enketo/blob/534bf69ef84a56df0f41bfffd69401ee25439284/packages/enketo-core/src/js/form.js#L1442) to `console.error`. So, I am essentially trying to re-create that behavior here.  Printing the error to the console is not really helpful to the user, but it could be useful to an app-dev or someone else trying to troubleshoot unexpected behavior.  Ideally, we can remove this logging when https://github.com/medic/cht-core/issues/7514 is addressed.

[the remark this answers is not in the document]
sugat009: Cool, that makes sense.

sugat009 (approved):

Issue #8698, opened 2023-11-10
Clean up and organize e2e tests: folders /about, /admin, /analytics


tatilepizs:
**Describe the issue**
We need to clean up and organize the e2e test, this ticket is to do it for folders [/about](https://github.com/medic/cht-core/tree/master/tests/e2e/default/about), [/admin](https://github.com/medic/cht-core/tree/master/tests/e2e/default/admin), [/analytics](https://github.com/medic/cht-core/tree/master/tests/e2e/default/analytics)

**Describe the improvement you'd like**
The plan for this first phase is just to clean up and organize the e2e test that we already have, we don't plan to remove or implement any new scenarios, we just want to try to reduce the flakiness of the tests that we have. 

**Describe alternatives you've considered**

PR #9355, opened 2024-08-19
chore(##8698): clean up and organize e2e tests: folders /about, /admin, /analytics

lorerod:
# Description

closes [#8698](https://github.com/medic/cht-core/issues/8698)

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8698-clean-up-and-organize-about-admin-analytics-folder/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8698-clean-up-and-organize-about-admin-analytics-folder/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8698-clean-up-and-organize-about-admin-analytics-folder/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/tests/e2e/default/analytics/analytics.wdio-spec.js
+++ b/tests/e2e/default/analytics/analytics.wdio-spec.js
@@ -32,7 +32,7 @@ describe('Targets', () => {
 
   const compileTargets = async (targetsFileName = 'targets-config.js') => {
     await chtConfUtils.initializeConfigDir();
-    const targetFilePath = path.join(__dirname, `config/${ targetsFileName }`);
+    const targetFilePath = path.join(__dirname, `config/${targetsFileName}`);
 
     return chtConfUtils.compileNoolsConfig({ targets: targetFilePath });
   };
--- a/tests/e2e/default/analytics/analytics.wdio-spec.js
+++ b/tests/e2e/default/analytics/analytics.wdio-spec.js
@@ -38,8 +38,8 @@ describe('Targets', () => {
   };
 
   before(async () => {
-    await utils.saveDocs([ ...places.values(), owl ]);
-    await utils.createUsers([ chw ]);
+    await utils.saveDocs([...places.values(), owl]);
+    await utils.createUsers([chw]);
     await sentinelUtils.waitForSentinel();
 
     await loginPage.login(chw);
@@ -112,7 +112,7 @@ describe('Targets', () => {
 
     const feedbackDocs = await chtDbUtils.getFeedbackDocs();
     expect(feedbackDocs.length).to.equal(1);
-    expect(feedbackDocs[ 0 ].info.message).to.include('Cannot read properties of undefined (reading \'muted\')');
+    expect(feedbackDocs[0].info.message).to.include('Cannot read properties of undefined (reading \'muted\')');
     await chtDbUtils.clearFeedbackDocs();
   });
 });
--- a/tests/e2e/default/analytics/analytics.wdio-spec.js
+++ b/tests/e2e/default/analytics/analytics.wdio-spec.js
@@ -10,15 +10,7 @@ const placeFactory = require('@factories/cht/contacts/place');
 const personFactory = require('@factories/cht/contacts/person');
 const chtConfUtils = require('@utils/cht-conf');
 const chtDbUtils = require('@utils/cht-db');
-const { TARGET_MET_COLOR, TARGET_UNMET_COLOR } = analyticsPage;
-
-
-const compileTargets = async (targetsFileName = 'targets-config.js') => {
-  await chtConfUtils.initializeConfigDir();
-  const targetFilePath = path.join(__dirname, `config/${targetsFileName}`);
-
-  return chtConfUtils.compileNoolsConfig({ targets: targetFilePath });
-};
+const {TARGET_MET_COLOR, TARGET_UNMET_COLOR} = analyticsPage;
 
 describe('Targets', () => {
   const places = placeFactory.generateHierarchy();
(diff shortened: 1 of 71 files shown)


tests/e2e/default/analytics/analytics.wdio-spec.js
```diff
@@ -12,11 +12,6 @@ const chtConfUtils = require('@utils/cht-conf');
 const chtDbUtils = require('@utils/cht-db');
 const { TARGET_MET_COLOR, TARGET_UNMET_COLOR } = analyticsPage;
 
-const updateSettings = async (settings) => {
-  await utils.updateSettings(settings, 'api');
-  await commonPage.sync(true);
-  await browser.refresh();
-};
 
 const compileTargets = async (targetsFileName = 'targets-config.js') => {
```
tatilepizs, line 16: Please move this method inside the first `describe`.

And on L39, please remove the `roles: ['chw']` because the `role` is already set as `cht`, in the [userFactory](https://github.com/medic/cht-core/blob/master/tests/factories/cht/users/users.js#L24)

tatilepizs (changes requested):
Thank you Lore for refactoring the `updateSettings` method, it looks great ⭐ 

I left one comment in line and was checking and in the `admin-access.wdio-spec.js` test the `offlineUser` and `parent` definitions are outside of the `describe` method, please move them inside to the describe, so everything will be wrapped in the main `describe`.

There is one more thing, I saw that you removed a lot of blank spaces between the `{` and the data (maybe the editor that you are using modified that 🤔 ) but I am not sure if we should change that, I was checking the code, and those spaces are pretty consistent over the `cht-core`, see these examples:

- [analytics-target-aggregates-sidebar-filter-component.spec.ts](https://github.com/medic/cht-core/blob/master/webapp/tests/karma/ts/modules/analytics/analytics-target-aggregates-sidebar-filter-component.spec.ts#L54-L55)
- [contacts-content.component.spec.ts](https://github.com/medic/cht-core/blob/master/webapp/tests/karma/ts/modules/contacts/contacts-content.component.spec.ts#L104-L111)
- [reports-add.component.spec.ts](https://github.com/medic/cht-core/blob/master/webapp/tests/karma/ts/modules/reports/reports-add.component.spec.ts#L87-L96)

And if you check in the [Coding Style Guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/), in the `right` sections, there are always spaces:

image
image

What do you think?

tests/integration/api/controllers/all-docs.spec.js
```diff
@@ -192,10 +192,10 @@ describe('all_docs handler', () => {
 
   it('filters offline users when requested with keys param', () => {
     const docs = [
-      { _id: 'allowed_contact', parent: { _id: 'fixture:offline'}, type: 'clinic' },
```
dianabarsan, line 195: I'm not sure what is adding these changes, there seem to be a lot of them and it makes reviewing this so much more cumbersome. I also like having spaces around curlies. Are these style changes (and IO'm referring to all changes in this PR that remove or add spaces around curlies) necessary in this already large PR?

tests/utils/index.js
```diff
@@ -526,15 +526,25 @@ const waitForSettingsUpdateLogs = (type) => {
...
-    return await commonElements.closeReloadModal(true);
+    await commonElements.closeReloadModal(true);
+  } else if (watcher) {
+    await watcher.promise;
+  }
+  if (sync) {
+    await commonElements.sync(true);
+  }
+  if (refresh) {
+    await browser.refresh();
```
dianabarsan, line 546: Cool! This cleans up code quite nicely. One comment is that when we choose not to ignoreReload, it means that we have an online user that will never have the sync button, and the "reload" modal will also refresh the page. So we can just return after `await commonElements.closeReloadModal(true);` to make the function simpler.
I'm not sure if we will ever have `sync: false` and `refresh: true`, we could consider connecting these somehow.

dianabarsan (commented):
Nice work! 
I reviewed without whitespace because it was super overwhelming. I'm not sure if that change is necessary here. 
Otherwise I just have one tiny comment.

tests/integration/api/controllers/all-docs.spec.js
lorerod: I'm so sorry about this. I had to reinstall my IDE because of an update going wrong, and now it seems to be changing the spaces! I will fix this for the next round of review.

lorerod:
> There is one more thing, I saw that you removed a lot of blank space
@tatilepizs I will fix this for the next round of review. Sorry about that.

tests/utils/index.js
lorerod: Until now, we have used these options in three ways:
`sync:false` and` refresh:false` (default values)
`sync:true` and `refresh:true`
`sync:true `and `refresh: false` (only once in an e2e default tasks test)

I can implicitly set the sync option to true if a refresh is true, but I think it is better to leave the flexibility. What do you think?

tatilepizs (approved):
Looks good to me! Thanks for the updates Lore ⭐

tests/utils/index.js
dianabarsan: If flexibility is non-sensical it can also create confusion, because then we would never have function parameters, we would just pass unlimited args to every function. In the end, this isn't so important in this case.

tests/utils/index.js
```diff
@@ -521,20 +521,56 @@ const waitForSettingsUpdateLogs = (type) => {
...
+  if (revert) {
+    await revertSettings(true);
+  }
+  const watcher = ignoreReload && Object.keys(updates).length && await waitForSettingsUpdateLogs(ignoreReload);
   await updateCustomSettings(updates);
   if (!ignoreReload) {
-    return await commonElements.closeReloadModal(true);
+    await commonElements.closeReloadModal(true);
+    return;
+  } else if (watcher) {
```
dianabarsan, line 565: You don't need the else here, because you return in the previous block. 
Also please use `return await` for one line less.

dianabarsan (commented):
I think I left one tiny comment,

dianabarsan (approved):
<3

Issue #9362, opened 2024-08-21
Publish cht-deploy to npm on version change


henokgetachew:
**Is your feature request related to a problem? Please describe.**

The `cht-deploy` publish process should be automated.
**Describe the solution you'd like**

The publish should happen via GitHub Actions.
**Describe alternatives you've considered**

Publishing directly from engineer's machines. Certainly not ideal.

mrjones-plip:
we're likely planning on [removing this script](https://github.com/medic/cht-core/pull/10249#issuecomment-3239748265).

PR #9363, opened 2024-08-21
feat(#9362): publish cht-deploy to npm on version change

henokgetachew:
# Description

Publishes to npm on version change in cht-deploy `package.json` (Under `scripts/deploy/`). The version change has to happen manually and no automated commits to github happens. This was done on purpose, to avoid triggering the build process. It runs the cht-deploy tests to make sure they pass before publishing.

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:cht-deploy-publish/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:cht-deploy-publish/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:cht-deploy-publish/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/.github/workflows/npm-publish-cht-deploy.yml
+++ b/.github/workflows/npm-publish-cht-deploy.yml
@@ -2,8 +2,6 @@ name: Publish cht-deploy to npm
 
 on:
   push:
-    branches:
-      - '**'
     paths:
       - 'scripts/deploy/**'
 
@@ -35,4 +33,4 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-        run: npx semantic-release --branches=${{ github.ref_name }} --config ./scripts/deploy/.releaserc.json
+        run: npx semantic-release --config ./scripts/deploy/.releaserc.json
--- a/.github/workflows/npm-publish-cht-deploy.yml
+++ b/.github/workflows/npm-publish-cht-deploy.yml
@@ -3,7 +3,9 @@ name: Publish cht-deploy to npm
 on:
   push:
     branches:
-      - master
+      - '**'
+    paths:
+      - 'scripts/deploy/**'
 
 jobs:
   publish:
@@ -14,7 +16,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      
+        with:
+          fetch-depth: 0
+
       - name: Set up Node
         uses: actions/setup-node@v3
         with:
@@ -26,9 +30,8 @@ jobs:
       - name: Run tests
         run: npm test
 
-      - name: Publish to npm
-        uses: JS-DevTools/npm-publish@v3
-        with:
-          token: ${{ secrets.NPM_TOKEN }}
-          access: public
-          package: ./scripts/deploy/package.json
+      - name: Semantic Release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: npx semantic-release --config ./scripts/deploy/.releaserc.json
--- a/.github/workflows/npm-publish-cht-deploy.yml
+++ b/.github/workflows/npm-publish-cht-deploy.yml
@@ -1,0 +1,34 @@
+name: Publish cht-deploy to npm
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./scripts/deploy
+
+    steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/npm-publish-cht-deploy.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Set up Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '20.x'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Run tests
+        run: npm test
```
garethbowen, line 27: This build only runs on master, right? Do we run these tests from some other CI process? If they only run in master then we won't find out we've broken something until after merge. Consider adding this to the main build.yml so we can test before merging too.

.github/workflows/npm-publish-cht-deploy.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Publish to npm
+        uses: JS-DevTools/npm-publish@v3
+        with:
+          token: ${{ secrets.NPM_TOKEN }}
+          access: public
+          package: ./scripts/deploy/package.json
```
garethbowen, line 34: In cht-conf we use semantic release to do this and it would be great to use the same in both places, but that library is designed to also update the version and generate release notes so maybe it's overkill. What do you think?

garethbowen (changes requested):
A couple of questions inline.

I'm assuming you haven't tested this at all, because it only runs on `master`? What do you think about tweaking it slightly so we can run it in this branch and check if it actually works and then switch back to `master` before merging?

For example, it would be good to check that, a) it actually builds and publishes correctly, and 2) it doesn't publish if it's run a second time and the version hasn't changed.

Otherwise we might end up cluttering up the commit log trying to get this to work in master.

.github/workflows/npm-publish-cht-deploy.yml
henokgetachew: Yes, semantic release is cool and more comprehensive. I liked the automated changelog.md. I was assuming that it would try to include all the CHT related commits in the release notes. That [may not be the case.](https://stackoverflow.com/questions/69342978/is-it-possible-to-exclude-some-commits-from-semantic-release-release-notes). If that's not the case, there is a [commits analyzer](https://github.com/semantic-release/commit-analyzer) to filter out commits that don't have anything to do with cht-deploy. I'll give it a look.

.github/workflows/npm-publish-cht-deploy.yml
henokgetachew: Yup, they are [run](https://github.com/medic/cht-core/blob/master/package.json#L53) before PRs are merged. I guess that effectively makes running the tests here redundant as tests must pass before PRs are merged anyway. I will remove it.

garethbowen: Perfect!

henokgetachew:
Thanks @garethbowen!

Current state of things:
* We're using semantic-release
* The workflow triggers only on changes to the `scripts/deploy/**` path.
* If the branch is master, it doesn't run the tests.
* If the branch is any other branch, it runs the tests and it makes a pre-release to npm

We need to configure secrets - I don't have those rights on this repo: 

```
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
That's what's remaining to see this in action.

scripts/deploy/.releaserc.json
```diff
@@ -0,0 +1,24 @@
+{
+  "branches": [
+    "master",
+    {
+      "name": "*",
+      "prerelease": "${name}"
+    }
+  ],
+  "plugins": [
+    "@semantic-release/commit-analyzer",
```
garethbowen, line 10: I suspect this won't work as expected because it'll get the changelog for all commits on the master branch, not just commits that impact this directory. What do you think?

.github/workflows/npm-publish-cht-deploy.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Set up Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '20.x'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Run tests
+        if: github.ref != 'refs/heads/master'
```
garethbowen, line 29: Once we merge this check should be at the top level instead of here because the entire publish job should not run if not on master, right?

garethbowen (changes requested):
Nice one! A couple changes suggested, and I'll look at getting the tokens added.

henokgetachew:
Closing this in favor of moving cht-deploy to its own repo.

PR #9364, opened 2024-08-22
fix(#9341): strip invisible characters

### LINKED ISSUE #9341: Strip invisible characters from the fields bsYear, bsMonth, bsDay
With issue #7654 (PR #7656), we solved the problem of zero-width (invisible) characters ruining the SMS forms.

With #7676 (PR #7718), we made it less strict and allowed the zero-width characters in strings.

With #8447 (PR #8450), we accepted Nepali (Devanagari) numerals in more fields

But it looks like, we missed to strip the invisible characters from some Bikram Sambat date related fields:

https://github.com/medic/cht-core/blob/1b9b6527c77d8668ca4963ddc7a2d5e6aa6b7273/api/src/services/report/smsparser.js#L227-L235

SMS: https://www.fontspace.com/unicode/analyzer#e=TCA5NzI4NyAyMDgxIDEgMeKAjQ

image
image

aloundoye:
# Description
fix #9341

# Code review checklist

# License


aloundoye:
@garethbowen, can you look at this one?

garethbowen (changes requested):
Thanks, this looks good to me!

Can you please also add few unit tests to prove that this works as expected. One technique I like to use is red/green testing which means you write the test which is failing, and then write the code to solve it, and see that the test goes green. Because you've already written the code I recommend commenting it out, and writing the test, and then uncommenting the code again. This gives me a lot of confidence that I've actually solved the issue.

The best place to put these unit tests is smsparser.spec.js

garethbowen (changes requested):
I didn't realise this function wasn't tested at all! Thank you for filling that hole in the coverage.

Can you add one more test which tests that invisible characters are stripped? For example, copy one of the tests you added and add an invisible character to the input string?

garethbowen (approved):
Excellent! Thanks for the quick turnaround.

I'll merge this as soon as the build passes.

aloundoye:
> I'll merge this as soon as the build passes.

Thanks @garethbowen

Issue #8225, opened 2023-05-08
Cannot use db-object-widget to load contact data in non-relevant `inputs` group


jkuester:
**Describe the bug**
If the `inputs` group in a form is never relevant, then the `select-contact` feature from the db-object-widget cannot be used to load contact data into a sub-group of the `inputs` group.

Take the following form:

type | name | label::en | relevant | appearance | calculation
-- | -- | -- | -- | -- | --
begin group | inputs | NO_LABEL | ./source = 'user' | field-list |  
begin group | contact | NO_LABEL |   | field-list |  
begin group | parent | NO_LABEL |   |   |  
hidden | _id | Parent ID |   |   |  
end group | parent |   |   |   |  
end group | contact |   |   |   |  
begin group | my_parent_data |   |   |   |  
string | _id |   |   | select-contact | ../../contact/parent/_id
string | name |   |   |   |  
end group | my_parent_data |   |   |   |  
end group | inputs |   |   |   |  

This does not work in situations where `./source = 'user'` is false (does not work when opening the form from the "People" tab with a contact in context.  

I think this behavior is the result of some incomplete [patching](https://github.com/medic/cht-core/tree/master/webapp/patches) of the `enketo-inputs-always-relevant` situation where Enketo is still not actually evaluating the calculations in the non-relevant `inputs` group.  [Other `inputs` data](https://docs.communityhealthtoolkit.org/apps/guides/forms/form-inputs/) (such as `inputs/contact`) will show up just fine since it is pre-loaded into the form.  All of this kind of makes sense from the Enketo logic perspective, but is _really confusing_ for any consumer trying to get their form to work!
�
The whole `inputs` functionality probably needs to be re-thought in `5.0`, but for now, it might be enough to just add an additional patch(es) in enketo-core for places [like this](https://github.com/enketo/enketo-core/blob/b83271e8cc88fb79ddeb4f79fb1194d2aac9f421/src/js/calculate.js#L199) that would allow for calculations to run inside the non-relevant `inputs` group. 

**Environment**
- Version: `master` (`4.1.x`)

**Additional context**

The work-around here is to not try to use the `select-contact` in the `inputs` group unless you know that it is going to be relevant when you need it.

[forum thread](https://forum.communityhealthtoolkit.org/t/query-parent-contact-place-data-in-child-form/2711/4)

garethbowen:
@jkuester Please check the Affects labels I've added here - I assume this is related to the enketo-core upgrade in 4.x?

[...]

PR #9382, opened 2024-08-27
fix(#8225): fix handling of `inputs` group with contact selector

jkuester:
# Description

Closes https://github.com/medic/cht-core/issues/8225

This PR updates our patch for the Eneketo relevant logic to make it so that the `inputs` group never can get set to non-relevant (and so calculations and widgets and things involving `inputs` all work as expected even when the `inputs` fields are not _shown_ in the form).  

## Testing considerations

I have expanded our db-object widget e2e tests to cover using the widget to load contact data inside the `inputs` group.  Also, the other existing tests all work as expected.  The risk here is that the new Enekto patch somehow breaks Enekto code in other unexpected ways.  I have done some manual testing to try to shake out any issues, but everything has worked as expected.

It would be good, though, to perhaps do more manual testing of any workflows that involve passing `inputs` data to forms (e.g. forms triggered from tasks).  

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8225_update_relevant_patch/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8225_update_relevant_patch/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8225_update_relevant_patch/docker-compose/cht-couchdb-clustered.yml)

# License


tests/e2e/default/enketo/pregnancy-complete-a-delivery.wdio-spec.js
```diff
@@ -247,7 +247,7 @@ describe('Contact Delivery Form', () => {
       source_id: '',
       user: {
         contact_id: offlineUser.contact._id,
-        name: offlineUser.username,
+        name: offlineUser.contact.name,
```
jkuester, line 250: This change actually demonstrates this fix in action!  The regression scenario [from the forum](https://forum.communityhealthtoolkit.org/t/current-user-data-hydration-with-db-object-issue-using-cht4/3880/4) was doing the exact same thing as the `delivery` form where they are trying to load the user's contact document directly into the `inputs/user` group.  The complication is that the `inputs/user` group _already contains data from the users `user-settings` doc._  So, when you use the contact selector to load the user's contact into the same group, it will override the data that is already there.  

In this case, both the user's contact and the `user-settings` document contain a `name` field.  For the `user-settings` it is the username. For the contact, it is just the normal contact name value.  Because the contact-selector functionality in non-relevant `inputs` was broken before this PR, the user's contact data was not actually being loaded and so the `name` was just populated with the username from the `user-settings` document.  Now that this PR has fixed the contact selector functionality, we see `inputs/user/name` getting overridden with the value from the user's contact (which is the proper behavior that we expect).

webapp/patches/enketo-core+7.2.5.patch
```diff
@@ -14,20 +14,24 @@ index 62003af..db02cba 100644
...
+             }
+ 
+-            if (
++            // CHT-CORE PATCH
++            // /inputs is ALWAYS relevant #4875
++            if(!result && /^\/[^/]+\/inputs$/.test(path)) {
++                branchChange = this.enable(branchNode, path, options) || branchChange;
++                branchNode.classList.add('disabled');
++            }
++            else if (
```
jkuester, line 31: The previous patch here was attempting a much more surgical modification where the `inputs` group would actually still get set to be non-relevant, but we would not clear the data in the `inputs` group (as a part of [this logic](https://github.com/enketo/enketo-core/blob/9551bf19b42db3ee4637a2e7067b6e9ba62ced0e/src/js/relevant.js#L461). It would remain available to other expressions in the form.  Unfortunately, the side effect of this was that if any of the data in the non-relevant `inputs` changed, that change did not get stored/propagated to other expressions since technically the fields that changed were non-relevant.  So, when the db-object widget was triggered inside the non-relevant `inputs` group, it would load the data as expected, but it would not get saved into the form data model.  

This new patch takes a more heavy handed approach where we actually just never let the `inputs` group get set to be non-relevant (by hijacking [this logic flow](https://github.com/enketo/enketo-core/blob/9551bf19b42db3ee4637a2e7067b6e9ba62ced0e/src/js/relevant.js#L268)).  The hope is that this will let us avoid any other kinds of weird side effects at the data-model level (especially since Enekto has some additional logic that gets trippy when some fields _have never been relevant_). Basically this lets is side-step all of that and as far as the rest of the form is concerned the `inputs` group is always relevant.  The downside is that when the `inputs` relevant expression evaluates to `false`, _we do not want the user to actually see the `inputs` fields in the form.  So, that is why we still slap the `disabled` class onto the `inputs` group to make sure it is not shown in the form.

tests/e2e/default/enketo/forms/db_object_form.xlsx
latin-panda, line 1: In the /enketo/forms folder we have a mix where we use underscore and dashes to separate words in the filename. But it seems like the most common is using dashes `tests/e2e/default/enketo/forms/db-object-form.xlsx`

latin-panda (approved):
Thanks for the great level of details in the ticket and in here, it's super clear and informative ✨

mrjones-plip:
Given this has been approved - should we move this to full PR instead of a Draft?

jkuester:
@mrjones-plip Done!

tests/e2e/default/enketo/forms/db_object_form.xlsx
jkuester: I am good going either way, but would push back a bit on dashes. The `config/default` forms use underscores and so you could say that is the "common pattern".  :sweat_smile: 

Maybe @tatilepizs can be the tie breaker here!  :rofl: :grimacing:

tatilepizs (approved):
Thank you for all the details @jkuester ⭐ 

I manually tested and created a pregnancy with danger signs that triggered (all as tasks):

- Pregnancy danger sign follow-up
- Health facility ANC reminder 
- Delivery
- Postnatal danger sign follow-up - mother
- Postnatal danger sign follow-up - baby

And everything worked as expected 🚀

tests/e2e/default/enketo/forms/db_object_form.xlsx
tatilepizs: 😐 You are right about the `config/default/forms/app` @jkuester, but if you check in `config/default/forms/contact` they use dashes.

I lean a little bit more towards dashes just because the percentage of use is bigger compared to underscores generally over the project (I think) 😅

jkuester:
@tatilepizs @latin-panda could you have one more look through this?  

I ended up making a change to the patch (to always set `branchChange = true;` when `inputs` gets set to be non-relevant).  This helped to fix some behavior issues I found with the following workflow:

- Open form where `inputs` in relevant (and so the user can see questions in the `inputs` group.
- Change a value in the form such that `inputs` is no longer relevant.
    - In this case, the user should no longer be able to see the questions in the `inputs` group.  

With my original patch in this PR, those non-relevant `inputs` questions would still be visible in to the user.  Now, with this change, the are hidden from view as expected. (But the values in the `inputs` group are still available to other expressions and will be saved in the report as expected.)

jkuester:
Test form file name changes coming soon!  :sweat_smile: Forgot to post those changes before requesting a review!

tatilepizs (approved):
🚀 
Thanks!

Issue #9104, opened 2024-05-08
Provide children contacts of current contact to contact-summary calculation


jkuester:
**Is your feature request related to a problem? Please describe.**
When calculating the contact summary for a place, it would be useful to be able to include data about the child contacts of the place in the calculation.  (E.g. [to show a warning](https://forum.communityhealthtoolkit.org/t/expose-child-count-to-contact-summary/3493) if the number of contacts does not match what is expected).  Currently, there is no way to access any information about those child contacts.

**Describe the solution you'd like**
In webapp, when we have a contact in context (e.g. we are viewing that contact's profile page), we have already done the heavy lifting to [look up the contact's children](https://github.com/medic/cht-core/blob/6fb510359c95eaaf1c751451046490268d35d2b8/webapp/src/ts/effects/contacts.effects.ts#L82).  (This involves finding which contact types are valid children of the current contact and then querying `medic-client/contacts_by_parent` for each of the types.)  We only calculate the contact-summary when we have a contact in context, so I do not think adding support for this functionality would negatively impact performance (except perhaps when loading a form for a _task_.  I believe we are currently calculating a the contact-summary in that case, but we might not have the children loaded already...)

**Describe alternatives you've considered**
We could rework our contact hierarchy model to be doubly linked so contacts contain the _ids both for their parent and their children.  Seems like that would cause more trouble than it would resolve....

jkuester:
@garethbowen after taking a deep dive into the code here and what it would take to make sure we have the contact children everywhere that we are currently calculating the contact summary, I have confirmed that we would need to add an additional call to `medic-client/contacts_by_parent` _each time we open an app form_ either from a contact profile page or from a task.  This is the only way we could guarantee we have up-to-date data about the contact's children.  

Let me know if you think this qualifies as "much of a lift"!  On one hand, the view query should be pretty straightforward and efficient. On the other hand, I know that we are already pretty sensitive about the form opening times and this will def not improve that. But, on the other hand, we would be able to do the query simultaneously with these other three lookups that we are already doing for the contact:

- getting all the contact's reports
- getting the contact's lineage
- getting the contact's target docs

These 4 things would all be happening at the same time and it seems unlikely the new query would be any slower than `medic-client/reports_by_subject`. 🤔 🤷

dianabarsan:
> happening at the same time

Unfortunately, the indexeddb adapter is single-threaded and queues all requests. 

>  pretty sensitive about the form opening times

[...]

Issue #9392, opened 2024-08-29
Expose child count to contact summary


ChinHairSaintClair:
**What feature do you want to improve?**
In the contact-summary there's no way to check the number of items that have been captured under a parent.
For example, the amount of household members captured for a household.

**Describe the improvement you'd like**
I would like to expose the children of a contact and use it to verify the enumerated items against the expected amount. 
Should there be a mismatch, I want to notify the CHW that further action is required via a contact flag.

**Describe alternatives you've considered**
NA

**Additional context**
After some pointers from @jkuester we managed to achieve the desired functionality:
![image](https://github.com/user-attachments/assets/5200e050-51db-4922-a03a-0de47f0e6150)
Just as a note the `config/<your_project>/.eslintrc` will need to have `children` added to its `globals` block.

The count can be used as follows:
```
const getChildCount = (items, key) => {
  for (let i = 0; i < items.length; i++) {
    const element = items[i];
    if (element.type.id === key) {
      return element.contacts.length;
    }
  }
  return 0; // No children exist
};
const contactChildren = children;
getChildCount(contactChildren ? contactChildren : [],'hhm')
```

https://forum.communityhealthtoolkit.org/t/expose-child-count-to-contact-summary/3493

jkuester:
@ChinHairSaintClair I am closing this in favor of https://github.com/medic/cht-core/issues/9104.  `9104` is about the general functionality of providing the contact's children to the contact-summary calculation.  This seems to be essentially what is implemented in [your PR](https://github.com/medic/cht-core/pull/9393).  Using those children in a workflow to calculate the actual "count" (as described in this issue) is functionality that will be enabled by `9104` (as implemented in your PR).  

Please respond here, though, if you feel like I have misunderstood something!

PR #9393, opened 2024-08-29
fix(#9104): expose children to contact summary

ChinHairSaintClair:
# Description

Closes https://github.com/medic/cht-core/issues/9104

One of our stakeholder requirements is to make sure our CHWs enumerate all places and people within their work area.
Exposing the child count in the contact summary and comparing it to the captured `expected_children` value in the parent doc, it'll allow us to dynamically show a contact flag to highlight additional action is required.

[](https://github.com/medic/cht-core/issues/9392)

# Code review checklist

# License


ChinHairSaintClair:
Also relates to:
https://github.com/medic/cht-core/issues/9104

jkuester:
Okay, I have taken a deep dive into the code here with the following findings:

- The current functionality works for calculating the contact-summary for contact profile pages, but the `children` array that is exposed to the contact-summary calculation is not optimized.  
  - That array is not a simple array of contact docs. Instead it is a more complex data structure from the [contact-view-model-generator.service](https://github.com/medic/cht-core/blob/a2fe0b43d4161d80c93d7fd8980d56afcaea01b6/webapp/src/ts/services/contact-view-model-generator.service.ts#L244) (intended for populating UI components).  
  - Ideally I think we should keep things simple and just pass an array of contact docs.
- The current approach is not working for the contact-summary calculated for app forms.  Basically I think we need to load the children for the contact similar to how we are already [loading the reports](https://github.com/medic/cht-core/blob/d18a40253c2f9fe12384529f86aa3cd8a1c536e3/webapp/src/ts/services/form.service.ts#L132) for the contact.  

Where things get complicated here is that between the [contacts.effects](https://github.com/medic/cht-core/blob/ebee299bef721a00d965e7eebedd7120f55746e8/webapp/src/ts/effects/contacts.effects.ts#L154) logic and the [contact-view-model-generator.service](https://github.com/medic/cht-core/blob/a2fe0b43d4161d80c93d7fd8980d56afcaea01b6/webapp/src/ts/services/contact-view-model-generator.service.ts#L198) logic, there is some non-trivial filtering being done regarding which children are returned for a contact (presumably to support UI workflows).  This makes things tricky for our contact-summary needs, though, because we really need the list of children to be consistent everywhere that the contact summary could be calculated.  Somehow we need to de-tangle the loading of the children in `contacts.effects`/`contact-view-model-generator` from the UI model logic in a way that still avoids repeated calls to the `medic-client/contacts_by_parent` view.

My current line of thinking is that perhaps we can encapsulate `medic-client/contacts_by_parent` in cht-datasource apis.  Then, contact returned from there could get passed to the both the `contact-summary.service` and the `contact-view-model-generator.service.loadChildren` service (which would then filter the contacts based on the current logic so the UI needs are still met).

These changes would considerably expand the surface area of this PR.  @ChinHairSaintClair I am not sure about your availability to work on them at this time.  If you are interested, I can try to put together a more clear technical design that would give you something to work on.

Issue #9269, opened 2024-07-18
Expose the user's target documents into the contact summary


kennsippell:
**Is your feature request related to a problem? Please describe.**
Automated payment workflows and stock monitoring workflows are often limited in their functionality because of the data that is available to the contact summary and data which is available to forms. A contact summary can only see data directly associated with a single place and even after 3.7.x the data is limited to 500 reports which can be insufficient when reviewing all data created by a CHP in a month. 

We are seeking a performance way to expose more data to contact summaries and to forms.

**Describe the solution you'd like**
Targets have full access to all data on a user's device. Targets are efficient because 1) the data is not exposed at once and 2) the results target computations are efficiently cached. `targetDocs` store the computed results of targets.  If we were to expose the user's current and recent target docs to the contact summary, this would be fast, powerful, and performant way to allow contact summaries and forms to effectively access the results of computations which requirei all data on the device. 

**Describe alternatives you've considered**

**Additional context**
eCHIS Kenya Digital Payments go-live is awaiting this feature

mrjones-plip:
@garethbowen  - I'd love to merge this to `master` so we have a solid feature in `4.11`  and then do the release that @witash  is release manager for.

Can you complete the merge now/create a new PR to merge/assign an engineer to do the work?  

Thanks!

garethbowen:
@dianabarsan is assigned already. What do you think? Is this ready for the mainline?

dianabarsan:
This has been merged to `master` and backported to `4.11` and UAT-ed on 4.11.

Issue #9391, opened 2024-08-29
Workflows repeatedly fail in e2e tests for muting and unmuting


witash:
**Describe the issue**
`for an offline user should mute and unmute a clinic while "offline", without processing in between` and `for an offline user should mute and unmute a person while "offline", with processing in between` in `/tests/e2e/default/transitions/client-side-muting.wdio-spec.js` fail for about 2/3 workflow runs

```
[chrome 128.0.6613.84 linux #0-2] Muting 
[chrome 128.0.6613.84 linux #0-2]     for an online user
[chrome 128.0.6613.84 linux #0-2]        ✓ should not process client-side when muting as an online user
[chrome 128.0.6613.84 linux #0-2] 
[chrome 128.0.6613.84 linux #0-2]     for an offline user
[chrome 128.0.6613.84 linux #0-2]        ✓ should not process muting client-side if not enabled
[chrome 128.0.6613.84 linux #0-2]        ✖ should mute and unmute a person while "offline", with processing in between
[chrome 128.0.6613.84 linux #0-2]        ✖ "after each" hook for for an offline user
[chrome 128.0.6613.84 linux #0-2]     
[chrome 128.0.6613.84 linux #0-2] 2 passing (1m 50.4s)
[chrome 128.0.6613.84 linux #0-2] 1 failing
[chrome 128.0.6613.84 linux #0-2]
[chrome 128.0.6613.84 linux #0-2] 1) for an offline user should mute and unmute a person while "offline", with processing in between
[chrome 128.0.6613.84 linux #0-2] element ("#header-dropdown li.sync-status .success") still not displayed after 20000ms
[chrome 128.0.6613.84 linux #0-2] Error: element ("#header-dropdown li.sync-status .success") still not displayed after 20000ms
[chrome 128.0.6613.84 linux #0-2]     at async syncAndWaitForSuccess (/home/runner/work/cht-core/cht-core/tests/page-objects/default/common/common.wdio.page.js:329:3)
[chrome 128.0.6613.84 linux #0-2]     at async Object.sync (/home/runner/work/cht-core/cht-core/tests/page-objects/default/common/common.wdio.page.js:350:3)
[chrome 128.0.6613.84 linux #0-2]     at async updateClientSideMutingSettings (/home/runner/work/cht-core/cht-core/tests/e2e/default/transitions/client-side-muting.wdio-spec.js:225:9)
[chrome 128.0.6613.84 linux #0-2]
[...]

PR #9279, opened 2024-07-23
DO NOT MERGE feat(#9269): 4.8.1 FR improved contact summary targets

dianabarsan:
# Description

Adds cht-script-api v2 which has a context property. Right now, context only has a `targetDocs` property containing the last 6 target documents for the contact. 
Passes target documents to contact summary in forms (all forms, including tasks) as well as contact details. 
When loading contact summary of the user's facility, loads the target documents for the user. 

#9269

# Code review checklist

- [ ] Readable: Concise, well named, follows the [style guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/), documented if necessary.
- [ ] Documented: Configuration and user documentation on [cht-docs](https://github.com/medic/cht-docs/)
- [ ] Tested: Unit and/or e2e where appropriate
- [ ] Internationalised: All user facing text
- [ ] Backwards compatible: Works with existing data and configuration or includes a migration. Any breaking changes documented in the release notes.

# Compose URLs
If Build CI hasn't passed, these may 404:

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:4.8.1-FR-improved-contact-summary-targets/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:4.8.1-FR-improved-contact-summary-targets/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:4.8.1-FR-improved-contact-summary-targets/docker-compose/cht-couchdb-clustered.yml)

# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

PR #9394, opened 2024-08-29
feat(#9269): pass last three months target docs to contact summary

dianabarsan:
# Description

Adds new `analytics` property to CHTDatasourceAPI in webapp. 
This analytics property gets a `targetDocs()` list passed to contact-summary. This list has the latest 3 target docs for the contact. 
When generating the contact summary of one of the logged in user's facilities, pass the logged in user's target docs to contact-summary. 

#9269 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9269-last-three-months-targets/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9269-last-three-months-targets/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9269-last-three-months-targets/docker-compose/cht-couchdb-clustered.yml)

# License


dianabarsan:
Hi @latin-panda 

This became sort of urgent https://medic.slack.com/archives/C044NKVMD4Z/p1725011773530029?thread_ts=1724962197.807179&cid=C044NKVMD4Z
Can you please review? These changes are already approved by Gareth for the FR branch here: https://github.com/medic/cht-core/pull/9279 and is already deployed on several prod instances. This PR does involve some minimal refactors to the target-aggregates service, to reconcile the changes between the FR and the recent last months target aggregates.

dianabarsan:
cc @lorerod 

Appreciate it!

tests/e2e/default/transitions/client-side-muting.wdio-spec.js
```diff
@@ -11,7 +11,7 @@ const personFactory = require('@factories/cht/contacts/person');
 
 /* global window */
 
-describe('Muting', () => {
+describe.skip('Muting', () => {
```
dianabarsan, line 14: Skipping this test due to https://github.com/medic/cht-core/issues/9391

lorerod:
@dianabarsan, is this feature only applicable to online users?

dianabarsan:
@lorerod no, it's applicable for all users.

tests/e2e/default/targets/target-aggregates.wdio-spec.js
```diff
@@ -278,13 +279,71 @@ describe('Target aggregates', () => {
...
+          [onlineUser.contact.name]: [
+            { id: 'a_target', value: { total: 1, pass: 1 } },
+            { id: 'b_target', value: { total: 1, pass: 1 } }
+          ],
+        };
+
+        const targetsForContact = (contact) => {
+          return helperFunctions.docTags.map(tag => ({
+            _id: `target~${tag}~${contact._id}~irrelevant`,
+            reporting_period: tag,
```
lorerod, line 318: Can you also add targets over three months and validate those are not visible in the activity card fields?

tests/e2e/default/targets/target-aggregates.wdio-spec.js
```diff
@@ -278,13 +279,71 @@ describe('Target aggregates', () => {
...
-        await helperFunctions.validateCardFields(['yesterday Prometheus', '18', '15%']);
+        await helperFunctions.validateCardFields(['yesterday Prometheus', moment().format('YYYY-MM'), '18', '15%']);
 
         await browser.back();
         const secondTargetItem =
           await (await targetAggregatesPage.getTargetItem(expectedTargets[1], CURRENT_PERIOD, districtHospital1.name));
         await helperFunctions.assertTitle(secondTargetItem.title, expectedTargets[1].title);
       });
+
+      it('should display targets of current user on home place', async () => {
```
lorerod, line 290: Have you considered testing this with an offline user instead?

lorerod (approved):
Thanks, @diana. This looks great. I just left a couple of suggestions, but none are a blocker.

tests/e2e/default/targets/target-aggregates.wdio-spec.js
dianabarsan: it should be the same for the offline user. I just used one user that was available.

webapp/src/ts/services/cht-datasource.service.ts
```diff
@@ -121,6 +121,9 @@ export class CHTDatasourceService {
         },
         getExtensionLib: (id) => {
           return this.extensionLibs[id];
+        },
+        context: {
+          targetDocs: []
         }
```
latin-panda, line 127: This isn't very clear to me. When looking at dataSource, it's about exposing functions on scope, for sample `[person](https://github.com/medic/cht-core/blob/94ac934a05dfa1b1ebe35d4ffa6fe7673179ad9d/shared-libs/cht-datasource/src/index.ts#L74)`. 
It's unclear to me what entity this `context`  is about. Also, why is there a `targetDocs: []` directly instead of having an accessor function?

webapp/src/ts/services/contact-types.service.ts
```diff
@@ -98,6 +98,16 @@ export class ContactTypesService {
     return contactTypesUtils.isPersonType(type);
   }
 
+  async isPerson(contact) {
```
latin-panda, line 101: It's confusing to have `isPerson` and `isPersonType` exposed simultaneously in this service. It makes someone think we can use `isPersonType` wrongly because a type can be an object.  

Instead, it seems safer to have this function only since it already handles possible datatypes. And then remove `isPersonType(type?) {`

webapp/src/ts/services/enketo.service.ts
```diff
@@ -632,6 +632,8 @@ export class EnketoFormContext {
   isFormInModal?: boolean;
   userContact?: Nullable<Person.v1.Person>;
   contactSummary?: Record<string, any>;
+  userContactId?: string;
+  userFacilityId?: string;
```
latin-panda, line 636: Double checking: is this just one facility or can it be many?

webapp/src/ts/services/target-aggregates.service.ts
```diff
@@ -37,6 +36,23 @@ export class TargetAggregatesService {
     private ngZone:NgZone,
   ) { }
 
+  private readonly NBR_MONTHS = 3;
```
latin-panda, line 39: What is NBR?

webapp/src/ts/services/target-aggregates.service.ts
```diff
@@ -37,6 +36,23 @@ export class TargetAggregatesService {
...
+  private readonly INTERVAL_TAG_FORMAT = 'Y-MM';
+
+  private getIntervalTag(targetInterval) {
+    return moment(targetInterval.end).format(this.INTERVAL_TAG_FORMAT);
+  }
+
+  private getCurrentInterval(settings) {
```
latin-panda, line 46:
```suggestion
  private getCurrentInterval(appSettings) {
```
I guess this is app settings and not user settings, the parameter name can be like this for clarification

webapp/src/ts/services/target-aggregates.service.ts
```diff
@@ -315,19 +322,20 @@ export class TargetAggregatesService {
...
 
-  getReportingMonth(reportingPeriod) {
+  getReportingMonth(reportingPeriod:ReportingPeriod) {
     return this.settingsService
       .get()
       .then(settings => {
-        const tag = this.getIntervalTag(settings, reportingPeriod);
-        return moment(tag, 'YYYY-MM').format('MMMM');
+        const tag = this.getTargetIntervalTag(settings, reportingPeriod);
+        return moment(tag, this.INTERVAL_TAG_FORMAT).format('MMMM');
```
latin-panda, line 330: Why it's changing from 'YYYY-MM' to 'Y-MM'?

webapp/src/ts/services/target-aggregates.service.ts
```diff
@@ -359,30 +367,35 @@ export class TargetAggregatesService {
     return aggregates.find(aggregate => aggregate.id === targetId);
   }
 
-  getCurrentTargetDoc(contact?) {
-    return this.ngZone.runOutsideAngular(() => this._getCurrentTargetDoc(contact));
+  getTargetDocs(contact, userFacilityId:string[]|string|undefined, userContactId:string|undefined):Promise<any[]> {
```
latin-panda, line 370:
```suggestion
  getTargetDocs(contact, userFacilityIds:string[]|undefined, userContactId:string|undefined):Promise<any[]> {
```
Can it only receive an array of strings? Then `isUserFacility` only checks the array with includes. The consumer can pass the array with 1 id if necessary, `getTargetDocs([id])`.  Or the user settings service can return an array of facilities always.

latin-panda (changes requested):
Thanks @dianabarsan! ✨ 
Mostly minor suggestions, and a couple of questions where I didn't fully understand

latin-panda:
Just curious: do you think this will make the contact details page slower? Something we need to monitor?

webapp/src/ts/services/cht-datasource.service.ts
dianabarsan: This is the design choice we made for the feature release. There have been lots of requests over time to add more data to contact summary. 
It's not an accessor function because the new code replaces how we get the target that always gets passed and was part of the original interface, so the computation happens anyway. We can make it an accessor function, but it would just be candy.

webapp/src/ts/services/target-aggregates.service.ts
dianabarsan: short for NUMBER :)

webapp/src/ts/services/target-aggregates.service.ts
dianabarsan: I wanted less magic strings. I'm unsure why there was a difference in the first place.

webapp/src/ts/services/target-aggregates.service.ts
dianabarsan: I'll check if we ever get the single facility id as a string. I thought this was possible. Is it not?

dianabarsan:
> do you think this will make the contact details page slower? Something we need to monitor?

It does add a two extra queries. I had performance in mind when I designed this and rewrote it three times until I was satisfied.

webapp/src/ts/services/contact-types.service.ts
dianabarsan: Afaic, if you have a contact document and you want to know if it's a person, you now have to: 
1. call `getTypeId`
2. call `getTypeById`
3. call `isPersonType`

Do you think a shortcut for these three calls is not worth for this interface?

webapp/src/ts/services/cht-datasource.service.ts
latin-panda: It doesn't seem fit like that 🤔. [Looking at the documentation:](https://github.com/medic/cht-core/blob/94ac934a05dfa1b1ebe35d4ffa6fe7673179ad9d/shared-libs/cht-datasource/src/index.ts#L4)

> This module provides a simple API for interacting with CHT data.

Looks like it needs a [target file here](https://github.com/medic/cht-core/tree/94ac934a05dfa1b1ebe35d4ffa6fe7673179ad9d/shared-libs/cht-datasource/src) to return the list of target [docs type](https://github.com/medic/cht-core/blob/94ac934a05dfa1b1ebe35d4ffa6fe7673179ad9d/shared-libs/cht-datasource/src/libs/doc.ts#L6) by contact. 

Or have there a contact's file, to return the contact summary and that would define the contact summary context

dianabarsan: i think i'll just add an analytics section.

webapp/src/ts/services/contact-types.service.ts
latin-panda: Worth it! I mean to remove `isPersonType` since it doesn't handle when the type is an object. It's only used in Target Aggregate Service 's getSupervisedContacts anyway. 

Then keep this `isPerson(` as the only public function:
```
async isPerson(contact) {
  let type;
    if (typeof contact.type === 'object') {
      type = contact.type;
    } else {
      type = await this.get(this.getTypeId(contact));
    }
    return contactTypesUtils.isPersonType(type);
  }
```

latin-panda: Or `isPersonType` can be private

webapp/src/ts/services/enketo.service.ts
dianabarsan: it can be many. as you do :D

webapp/src/ts/services/target-aggregates.service.ts
latin-panda: I didn't think about `number`. What about `TARGET_MAX_MONTHS_OLD` or `MAX_MONTHS_OLD`

webapp/src/ts/services/target-aggregates.service.ts
latin-panda: Is at the end the same thing `YYYY-MM` to `Y-MM` for the usage in targets? 
If yes, I think `YYYY-MM` is something more people will understand when looking at it.

dianabarsan: it's the same thing.

webapp/src/ts/services/contact-types.service.ts
dianabarsan: > getSupervisedContacts anyway

I checked out this code and we still need `isPersonType` there. 
I'm leaving it as is.

dianabarsan: Ok, I refactored it and added a `getPersonChildTypes` so we don't have type-related code in the target aggregate service

webapp/src/ts/services/target-aggregates.service.ts
dianabarsan: I'm enforcing the the array in the reducer now. Previously it could show up as array or string.

webapp/src/ts/services/contact-types.service.ts
latin-panda: Because in `getSupervisedContacts` the `type` is string, and `async isPerson(contact) {` doesn't handle the case when `type` is provided as string, right?

dianabarsan: yea. I changed it in a different way. it's better now.

latin-panda (approved):
Great job! 🤩 Ship it 🛫

oh, if you haven't already, remember to document it in cht-docs

Issue #9398, opened 2024-08-30
Streamline /config dependencies


garethbowen:
**Describe the issue**

Having dependencies defined in the /config directories makes managing updates difficult. Currently these packages are triggering 65 of 118 dependabot warnings, which at worst mean there is a vulnerability, but at best is hiding the signal in the noise of overreporting.

**Describe the improvement you'd like**

Remove unnecessary dependencies, remove dependencies that are already defined at the root, and move dependencies needed to the root devDependencies

**Describe alternatives you've considered**

We could add monitoring to each subdir and do the updates but because most packages are duplicated at the top level this would just be duplication of effort.

PR #9403, opened 2024-09-02
chore(#9398): move config dependencies to root

garethbowen:
# Description

#9398

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9398-clean-up-config-dependencies/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9398-clean-up-config-dependencies/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9398-clean-up-config-dependencies/docker-compose/cht-couchdb-clustered.yml)

# License


jkuester (approved):
Changes all look good!  

My only comment is to just recognize that our maintainability gains that we are getting here by centralizing the dependencies are a tradeoff with the reference value that was being provided by having these config projects maintain their own dependencies.  For example, previously folks could easily reference the `config/default/package.json` file to see which combination of test dependencies was working together.  (Have seen partners struggle with this in the past.) 

Ultimately, I think the tradeoff is still worth it, especially since folks can still just look to the root package.json for dependency versions if they need to.

Issue #9413, opened 2024-09-04
Deprecated mutation event


ChinHairSaintClair:
We've experienced an new test failure in our pipeline (where we run many of your tests) without a change in code which we think may be related to removal of support for the mutation event `DOMSubtreeModified` https://developer.chrome.com/blog/mutation-events-deprecation from Chrome 127, which appears in the following file: https://github.com/medic/cht-core/blob/master/webapp/src/js/enketo/widgets/dynamic-url.js. 
This file is identical in the version (`"enketo-core": "^5.18.1"`) we're running.

`should set the href to the dynamically generated URL even when the dynamic value is empty`. outputs:
```
+ expected - actual
-[[http://google.com?q=helloWorld](http://google.com/?q=helloWorld)](http://google.com/?q=helloWorld](http://google.com/?q=helloWorld))
+[[http://google.com?q=worldHello](http://google.com/?q=worldHello)](http://google.com/?q=worldHello](http://google.com/?q=worldHello))
```
This only fails in the pipeline and does not fail locally.

The failures starting loosely align with an update to the runner image: https://github.com/actions/runner-images/commit/0d632219e5a750d8d1b9974d1e8713a2fa111ce9

We have looked at your pipelines and don't see failures with the same Chrome version, but we're a bit out of ideas otherwise.

In support of this theory, using the following to ensure that the assert occurs as the last action on the macro-task queue seems to resolve the issue.
```
setTimeout(() => {
   expect($(DynamicUrlWidget.selector).attr('href')).to.equal(`[http://google.com?q=${dynamic}](http://google.com/?q=${dynamic})`);
}, 0);
```

If there is a polyfill using MutationObserver (something like https://github.com/mfreed7/mutation-events-polyfill) buried somewhere in the dependency tree (shadydom may be adding one), then this may have changed the behaviour from sync to async and scheduled on the micro-task queue.

That said, we're not sure why this issue wouldn't be affecting your pipelines. However, because `DOMSubtreeModified` is being removed, we thought it useful to let you know about it.

We understand this isn't your bug, so our PR with the setTimeout is just a suggestion, but it might be something of a minor future proof for you and this issue is in part just a note that a breakage may be coming down the pipe.

Otherwise, we're hoping you may know something about it.

**Additional context:**
Environment/Image:
```
Agent name: 'Azure Pipelines 2'
Current agent version: '3.243.1'
Operating System
Ubuntu
20.04.6
LTS
Runner Image
Image: ubuntu-20.04
[...]

PR #9414, opened 2024-09-04
feat(#9413): update dynamic url widget to use MutationObserver

ChinHairSaintClair:
# Description
This approach ensures that the assertion occurs as the last action on the macro-task queue, which seems to address our pipeline issue without introducing too much additional code.

Closes https://github.com/medic/cht-core/issues/9413

# Code review checklist

# License


webapp/tests/karma/js/enketo/widgets/dynamic-url.spec.ts
```diff
@@ -47,7 +47,9 @@ describe('Enketo: Dynamic URL Widget', () => {
     const dynamic = 'worldHello';
     $('.url span').text(dynamic);
 
-    expect($(DynamicUrlWidget.selector).attr('href')).to.equal(`http://google.com?q=${dynamic}`);
+    setTimeout(() => {
+      expect($(DynamicUrlWidget.selector).attr('href')).to.equal(`http://google.com?q=${dynamic}`);
+    }, 0);
```
garethbowen, line 52: My take on this is you've made the expectation async, so the test thread will complete before the `expect` executes. This can lead to nasty issues where the test is marked as passed and then an error message is output in the log but doesn't cause CI to fail. While I don't see this happening in your CI build if this expectation was ever broken in the future then it probably wouldn't be picked up. You can test this out locally by changing the expected URL and making sure the test fails.

The fix for this is probably to add the `done` parameter to the function definition and calling `done()` inside the `setTimeout` callback.

```suggestion
    setTimeout(() => {
      expect($(DynamicUrlWidget.selector).attr('href')).to.equal(`http://google.com?q=${dynamic}`);
      done();
    }, 0);
```

garethbowen (changes requested):

garethbowen:
@jkuester As I'm about to head on leave can I hand this over to you for review?

I've approved so I don't block the eventual merge...

garethbowen (approved):
Approved to unblock.

jkuester:
@ChinHairSaintClair did you mean to push that last commit with the contact summary changes to this branch?

ChinHairSaintClair:
@jkuester no. Accidentally pushed the [other branch's](https://github.com/medic/cht-core/pull/9393) work to this one 🤦 .
Will have it fixed in the next few. Thank you for the heads up!

webapp/src/js/enketo/widgets/dynamic-url.js
```diff
@@ -17,7 +17,20 @@ class DynamicUrlWidget extends Widget {
     const urlElement = currentElement.find('.url');
     const setHref = () => currentElement.attr('href', urlElement.text());
     setHref();
-    urlElement.on('DOMSubtreeModified', setHref);
+
+    const observer = new MutationObserver(mutationList => {
+      mutationList.forEach(mutation => {
+        if (mutation.type === 'childList' || mutation.type === 'characterData') {
```
jkuester, line 23: Minor, but do we actually need this check? (I am not a MutationObserver expert, so I am trying to understand what all might be happening here!)

It seems like since we are locking down the observations below to only have `childList: true` and `characterData: true`, we should not need to filter anything here.  Any time the observer picks up a change, we should call `setHref`.  (And testing this locally, I only saw `childList` mutations coming through here).

webapp/tests/karma/js/enketo/widgets/dynamic-url.spec.ts
```diff
@@ -40,14 +40,17 @@ describe('Enketo: Dynamic URL Widget', () => {
...
 
     new DynamicUrlWidget($(DynamicUrlWidget.selector)[0]);
     const dynamic = 'worldHello';
     $('.url span').text(dynamic);
 
-    expect($(DynamicUrlWidget.selector).attr('href')).to.equal(`http://google.com?q=${dynamic}`);
+    setTimeout(() => {
+      expect($(DynamicUrlWidget.selector).attr('href')).to.equal(`http://google.com?q=${dynamic}`);
+      done();
+    }, 0);
```
jkuester, line 53: So just for the record (for anyone that is looking into this in the future) the normal Angular way of testing async UI updates via `fakeAsync` does not seem to work here.  And from reading online this seems to be expected (I guess the MutationObserver falls outside of what Angular has control of).  In general the consensus seems to be against trying too hard to unit test MutationObserver behavior.  However, since this works (at least for now) I say we ship it as is!  :+1:

jkuester (commented):
Very nice, thank you for these updates!  (We are only 10+ years behind on removing the deprecated `DOMSubtreeModified` code....)  :sweat_smile: 

I left one minor question about the type check, but besides that this should be good to go!

webapp/src/js/enketo/widgets/dynamic-url.js
ChinHairSaintClair: From my testing, and as you've noted, the check seems to be a bit redundant and no `characterData` mutation is observed.

For future further reading:
https://stackoverflow.com/a/10719879
https://lists.w3.org/Archives/Public/public-webapps/2011JulSep/1622.html
https://davidwalsh.name/mutationobserver-api (MutationObserver options)
https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/observe#removed_descendants_when_using_subtree
https://developer.mozilla.org/en-US/docs/Web/API/MutationRecord/target#logging_the_target_of_a_mutation

jkuester (approved):
LGTM!  Thanks @ChinHairSaintClair!

Issue #9327, opened 2024-08-14
Update Navigation to follow Material Design standards


Benmuiruri:
**What feature do you want to improve?**
As part of the full evolution of the CHT UI , the final step deals with updating how Navigation works. The specific changes are illustrated below.

1. Implement [bottom navigation](https://m3.material.io/components/navigation-bar/overview#856e8950-0398-4252-aec9-43327a94fb60)
2. Implement [navigation drawer](https://m3.material.io/components/navigation-drawer/overview#973f9227-7d99-4c5d-9908-7e7ed64cb9b3)
3. Implement [navigation rail](https://m3.material.io/components/navigation-rail/overview#582c4596-a472-4d1b-9b78-0abff9d0675a)
4. Align [app bars: top](https://m3.material.io/components/top-app-bar/overview#491fd00e-910c-47fe-b0bb-3a6b53d3cbf8) to material design standard

**Additional context**
[Figma mockups of all screens](https://www.figma.com/design/3wFzMcNEGxV6se1QWftckk/UX-Stages?node-id=3471-28116&t=40p1CPzFAuXoSO4f-1)

PR #9418, opened 2024-09-06
feat(#9327): refactor the colored bar to be the top bar

Benmuiruri:
chore: refactor messages tasks reports

chore: refactor contacts and analytics

Refactor colored bar to be the top bar in mobile and desktop.  (WIP)

Issue: https://github.com/medic/care-teams/issues/130

# Code review checklist

# License


Benmuiruri:
Last minute I noticed I used the wrong value for `@filters-bar-desktop-height:` it was the value for the old-nav. I wont push the fix since you have started the review. I'm just making the adjustments and I will push the correct value with feedback changes.

webapp/tests/karma/ts/modules/error/error.component.spec.ts
```diff
@@ -45,6 +49,11 @@ describe('ErrorComponent', () => {
           { provide: DOCUMENT, useValue: documentMock },
         ]
       })
+      .overrideComponent(ErrorComponent, {
+        set: {
+          template: `<div></div>`
```
latin-panda, line 54: Well played! 😁 
Just a comment on the type of quote
```suggestion
          template: '<div></div>'
```

webapp/src/css/inbox.less
```diff
@@ -1141,11 +1177,44 @@ div.old-nav .app-menu-button {
...
 
   mm-search-bar,
   contacts-filters,
   reports-filters {
     flex-grow: 2;
+    padding-left: 30px;
+  }
+
+  .app-menu-button {
+    order: 4;
```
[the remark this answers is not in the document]
Benmuiruri: yes it does.

webapp/src/css/tooltip.less
```diff
@@ -7,7 +7,6 @@
    */
   [title] {
     position: relative;
-    display: inline-flex;
```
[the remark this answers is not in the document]
Benmuiruri: That is to fix this alignment 
Screenshot 2024-09-13 at 14 27 46

When I added `display flex `to `inner` div it causes that misalignment. I had to remove the `inline-flex` to fix the alignment. 

Where else is that title used in the dates or times display?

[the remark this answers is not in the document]
latin-panda: Maybe it's better to make a specific rule for that case unless you see the tooltips displaying wrong, and we need to fix them.

We use it a lot and usually in mobile displays. The user long press a date or a label to see more details; some places on top of my head are these: 
Screenshot 2024-09-16 at 1 08 52 PM
Screenshot 2024-09-16 at 1 08 45 PM
Screenshot 2024-09-16 at 1 08 11 PM
Screenshot 2024-09-16 at 1 07 48 PM
Screenshot 2024-09-16 at 1 07 57 PM

webapp/src/ts/components/tool-bar/tool-bar.component.html
```diff
@@ -0,0 +1,10 @@
+<div class="filters tab-bar">
```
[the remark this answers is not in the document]
Benmuiruri: Not a bad idea, how about the class `tool-bar`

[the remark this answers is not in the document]
latin-panda: Sounds good!

webapp/src/ts/components/search-bar/search-bar.component.ts
```diff
@@ -102,4 +103,9 @@ export class SearchBarComponent implements AfterContentInit, AfterViewInit, OnDe
   ngOnDestroy() {
     this.subscription.unsubscribe();
   }
+
+  @HostBinding('class.search-open')
+  get isSearchOpen(): boolean {
+    return this.openSearch;
+  }
 }
```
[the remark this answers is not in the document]
Benmuiruri: This is the structure of the html without that code. The problem is the `inner` div does not know when the section inside `mm-search-bar` has the class `open-search` to be able to resize the `mm-search-bar`. 

<details>
<summary>Original HTML</summary>

Screenshot 2024-09-16 at 16 21 10


</details>

With that you end up with this 


<details>
<summary>Incorrect search width</summary>

Screenshot 2024-09-16 at 16 20 32

</details>

What that code does, is pass on the `open-search` class from the section to the parent `mm-search-bar`. 

<details>
<summary>Updated HTML</summary>

Screenshot 2024-09-16 at 16 21 34

</details>


Then from there, the `inner` div can see when the class is applied to `mm-search-bar` and adjust it accordingly 


<details>
<summary>Correct Search width</summary>

Screenshot 2024-09-16 at 16 28 15

</details>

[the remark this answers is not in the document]
latin-panda: Can we use the Store for that? It's similar to when the filter is open, so it's evident that it subscribes to an X-component state.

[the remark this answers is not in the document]
Benmuiruri: Yes, the store is another solution. I'll switch to that, perhaps it is more self-explanatory.

webapp/src/css/tooltip.less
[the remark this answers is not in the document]
Benmuiruri: I've tested it on Chrome, Firefox and Tecno Mobile, removing or setting the `display: inline-flex; ` does not affect the position of long pressing the date. It still displays in the same position. 

But having it affects the position of the Title at the top. I maintain we delete it as part of the changes of the new nav.

latin-panda (changes requested):
Nice job! I found some things please see below

Benmuiruri:
@latin-panda This is ready for review. 

The default mobile test is failing because the call to `waitForAngularLoaded` is looking for the `await (await hamburgerMenu()).waitForDisplayed({ timeout });`

After the refactor we no longer have this menu icon.

Screenshot 2024-09-20 at 07 51 49

 I will follow it up with @kapelo , but you can proceed with the review

latin-panda:
@Benmuiruri, those issues were fixed in Tati's latest PR, which was merged on Friday. I updated your branch to get the GHA restarted with those fixes.

webapp/src/css/inbox.less
```diff
@@ -1589,28 +1778,64 @@ mm-sidebar-menu .mat-sidenav-container {
...
       }
+    }
+  }
 
-      &.open-search {
-        background-color: rgba(255, 255, 255, 0.8);
+  .app-root.search-bar-active mm-search-bar {
+    width: 100%;
+    padding-left: 5px;
+    max-width: initial;
```
latin-panda, line 1802: L1174 is already setting `max-width: initial;`, are you sure we still need it here?

latin-panda (changes requested):
Thanks Ben! I've added some comments and questions below

webapp/src/css/old-nav.less
latin-panda, line 1: @Benmuiruri The user settings page has a very big header 
Screenshot 2024-09-23 at 5 36 04 PM

It should show like this:

Screenshot 2024-09-23 at 5 37 26 PM

webapp/src/css/inbox.less
```diff
@@ -1367,18 +1515,30 @@ mm-sidebar-menu .mat-sidenav-container {
...
+    .app-menu-button {
+      min-width: 0;
+    }
+
     .navigation {
-      display: inline-block;
+      .mat-icon-close {
+        width: @icon-XXS;
+        height: @icon-XXS;
+        overflow: initial;
```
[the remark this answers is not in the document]
latin-panda: It's weird the `overflow` fixes the alignment issue. It means there are some side effects and that fix is patching the problem.

Looking quickly, it seems like the anchor has some font size and line height that make it worse, also some `float`, the float is very old stuff and should be avoided if possible. This seems to work, but I didn't test all cases:

Screenshot 2024-09-23 at 9 11 19 PM
Screenshot 2024-09-23 at 9 11 22 PM
Screenshot 2024-09-23 at 9 11 35 PM
Screenshot 2024-09-23 at 9 11 46 PM

webapp/src/css/inbox.less
```diff
@@ -1299,6 +1391,28 @@ mm-sidebar-menu .mat-sidenav-container {
   }
 
   .app-root:not(.old-nav) {
+    &.messages,
+    &.tasks,
+    &.reports,
+    &.contacts {
+      .loading-status {
+        margin: 10px 10px calc(@tab-navbar-size + 10px);
```
latin-panda, line 1399: The privacy policy on mobile isn't showing the full page, it needs more margin bottom
Screenshot 2024-09-24 at 11 15 25 AM

webapp/src/css/inbox.less
```diff
@@ -206,10 +225,12 @@ body {
   }
 }
 
-.filters {
-  height: @filters-bar-desktop-height;
+.tool-bar {
```
latin-panda, line 228: Something broke the search in reports for mobile display
Screenshot 2024-09-24 at 11 29 47 AM

webapp/src/ts/components/search-bar/search-bar.component.ts
```diff
@@ -55,9 +59,11 @@ export class SearchBarComponent implements AfterContentInit, AfterViewInit, OnDe
     const subscription = combineLatest(
       this.store.select(Selectors.getSidebarFilter),
       this.store.select(Selectors.getFilters),
-    ).subscribe(([sidebarFilter, filters]) => {
+      this.store.select(Selectors.getSearchBar),
+    ).subscribe(([sidebarFilter, filters, searchBar]) => {
       this.activeFilters = sidebarFilter?.filterCount?.total || 0;
       this.filters = filters;
+      this.openSearch = searchBar.isOpen;
```
latin-panda, line 66: It's not closing properly in navigation:
https://github.com/user-attachments/assets/40681843-ee3d-4e51-9b40-b2b743e5c001

In the master branch:
https://github.com/user-attachments/assets/fd2c948d-faae-4dd8-a7b0-7dedda579cf7

webapp/src/css/inbox.less
```diff
@@ -206,10 +225,12 @@ body {
...
+  display: flex;
+  height: @toolbar-desktop-height;
```
latin-panda, line 230: There is a "jumping" effect in the topbar
- the top bar height changes
- the elements are displayed higher rather than vertically centered. Note the hamburger icon, and title are vertically centered, while the back chevron icon and title are higher.

latin-panda: https://github.com/user-attachments/assets/129c7646-f51d-48a0-9fdc-2a3cb7681651

webapp/src/css/old-nav.less
latin-panda, line 1: about page and testing page is being cut at the bottom 

Uploading Screen Recording 2024-09-24 at 12.23.59 PM.mov…

webapp/src/ts/components/search-bar/search-bar.component.ts
Benmuiruri: This was caused by the approach to manage search bar using state. I'm thinking of this solution. 

```
ngOnDestroy() {
  if (this.openSearch) {
    this.globalActions.setSearchBar({ isOpen: false });
  }
}
```

The other option is 

```
ngOnDestroy() {
  this.clear();
}
```

But clear() does more than just close the searchBar so I am leaning towards the first

latin-panda: It the state was already `false` it won't emit, so you can do this a bit simpler: 
```
ngOnDestroy() {
  this.globalActions.setSearchBar({ isOpen: false });
}
```

webapp/src/css/inbox.less
Benmuiruri: <details>
<summary>Alignment 1 </summary>

Screenshot 2024-09-24 at 11 31 51

</details>

<details>
<summary>Alignment 2 </summary>

Screenshot 2024-09-24 at 11 32 04

</details>


<details>
<summary>Alignment 3 </summary>

Screenshot 2024-09-24 at 11 33 53

</details>

<details>
<summary>Alignment 4 </summary>

Screenshot 2024-09-24 at 11 35 13

</details>

webapp/src/css/inbox.less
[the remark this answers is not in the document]
Benmuiruri: One reason the alignment was off is we use `mat-icon X svg`  to close the form. And we use `font icon back` for the back button. You fix one you break the other. 

I've changed the back button to be mat icon as well. Then made the adjustments.

latin-panda (approved):
LGTM! 🤩 Thanks for all the effort on this work, it looking really good! 

There's a tiny adjustment we can do in the about, privacy policy and testing pages for old nav, that we can submit in another PR just after this one is merged.

PR #9461, opened 2024-09-24
chore(#9327): update mobile e2e tests to work with the top bar

kapelo:
# Description

https://github.com/medic/cht-core/issues/9327
[care-teams/#161](https://github.com/medic/care-teams/issues/160)

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:160-e2e-tests-top-bar/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:160-e2e-tests-top-bar/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:160-e2e-tests-top-bar/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/tests/page-objects/default/common/common.wdio.page.js
+++ b/tests/page-objects/default/common/common.wdio.page.js
@@ -5,7 +5,7 @@ const aboutPage = require('@page-objects/default/about/about.wdio.page');
 const hamburgerMenu = () => $('aria/Application menu');
 const closeSideBarMenu = () => $('.panel-header-close');
 const FAST_ACTION_TRIGGER = '.fast-action-trigger';
-const fastActionFAB = () => $(`${FAST_ACTION_TRIGGER} .fast-action-fab-button`);
+const fastActionFAB = () => $$(`${FAST_ACTION_TRIGGER} .fast-action-fab-button`);
 const fastActionFlat = () => $(`${FAST_ACTION_TRIGGER} .fast-action-flat-button`);
 const multipleActions = () => $(`${FAST_ACTION_TRIGGER}[test-id="multiple-actions-menu"]`);
 const FAST_ACTION_LIST_CONTAINER = '.fast-action-content-wrapper';
@@ -81,12 +81,24 @@ const clickFastActionById = async (id) => {
   await (await fastActionById(id)).click();
 };
 
+/**
+ * There are two FABs, one for desktop and another for mobile. This finds the visible FAB.
+ * @returns {Promise<HTMLElement>}
+ */
+const findVisibleFAB = async () => {
+  for (const button of await fastActionFAB()) {
+    if (await button.isDisplayed()) {
+      return button;
+    }
+  }
+};
+
 const clickFastActionFAB = async ({ actionId, waitForList }) => {
   await closeHamburgerMenu();
-  await (await fastActionFAB()).waitForDisplayed();
-  await (await fastActionFAB()).waitForClickable();
+  const fab = await findVisibleFAB();
+  await fab.waitForClickable();
   waitForList = waitForList === undefined ? await (await multipleActions()).isExisting() : waitForList;
-  await (await fastActionFAB()).click();
+  await fab.click();
   if (waitForList) {
     await clickFastActionById(actionId);
   }
@@ -94,9 +106,9 @@ const clickFastActionFAB = async ({ actionId, waitForList }) => {
 
 const getFastActionItemsLabels = async () => {
   await closeHamburgerMenu();
-  await (await fastActionFAB()).waitForDisplayed();
-  await (await fastActionFAB()).waitForClickable();
-  await (await fastActionFAB()).click();
+  const fab = await findVisibleFAB();
+  await fab.waitForClickable();
+  await fab.click();
 
   await browser.pause(500);
   await (await fastActionListContainer()).waitForDisplayed();


latin-panda (approved):
LGTM! Just waiting for the CI to pass

tests/page-objects/default/common/common.wdio.page.js
```diff
@@ -94,6 +95,17 @@ const clickFastActionFAB = async ({ actionId, waitForList }) => {
   }
 };
 
+const clickFastActionFABMobile = async ({ actionId, waitForList }) => {
```
tatilepizs, line 98: I am a little confused about this method. 
Did the messages test start failing?
Because if we need to add differentiation between how we click on the flat or fab button we should use a method similar to the one created by Jenny for the reports, the [openFastActionReport](https://github.com/medic/cht-core/blob/master/tests/page-objects/default/common/common.wdio.page.js#L115-L124).

But I wonder why it stated failing 🤔

tests/e2e/default-mobile/navigation/more-options-menu.wdio-spec.js
```diff
@@ -60,9 +63,19 @@ describe('More Options Menu - Offline User', () => {
...
 
   describe('Contact tab', () => {
+
+    beforeEach(async () => {
+      await commonPage.goToBase();
+    });
+
+    after(async () => {
+      await searchPage.clearSearch();
+    });
```
tatilepizs, line 73:
```suggestion
```

You don't need this after method to clean the search because after all the tests in this file are executed a new chrome window will be created and it will started clean

tests/e2e/default-mobile/old-navigation/old-navigation.wdio-spec.js
```diff
@@ -42,7 +42,8 @@ describe('Old Navigation', () => {
     permissions.can_view_old_navigation = offlineUser.roles;
     await utils.updateSettings({ tasks, permissions }, true);
 
-    await loginPage.login(offlineUser);
+    await loginPage.login({username: offlineUser.username, password: offlineUser.password, loadPage: false});
```
tatilepizs, line 45: Why did you put the loadPage as `false`? 🤔

tatilepizs (changes requested):
Thanks @kapelo 
I left some questions 🤔

tatilepizs:
@kapelo I was checking why the `more-options-menu` test was still failing for you, and it is because the kebab menu is not working on mobile, something happened 😕 
cc: @Benmuiruri 

[video](https://github.com/user-attachments/assets/9556e506-9dff-4a20-bdb0-ef9d591e7922)

tests/page-objects/default/common/common.wdio.page.js
latin-panda: The FAB didn't change nor was it affected by the new navigation work. The original code to handle FAB (before the Angular exception fix) works fine. These changes were reverted.

tests/e2e/default-mobile/navigation/more-options-menu.wdio-spec.js
latin-panda: Confirmed and removed.

tests/e2e/default-mobile/old-navigation/old-navigation.wdio-spec.js
latin-panda: After the login action, this login function [waits for an element](https://github.com/medic/cht-core/blame/68cab433ccb537f19aca644ca87ab207b9d4e563/tests/page-objects/default/common/common.wdio.page.js#L304) to verify that the webapp is loaded. That element is different in the old nav from the new nav. Passing `false` skips [the `wait` that works in new nav](https://github.com/medic/cht-core/blob/128c1161aa22239ad90f5e224d0bef80e9d68a19/tests/page-objects/default/common/common.wdio.page.js#L332), and then we can use [a wait for the old nav](https://github.com/medic/cht-core/blob/128c1161aa22239ad90f5e224d0bef80e9d68a19/tests/page-objects/default/old-navigation/old-navigation.wdio.page.js#L109).

latin-panda:
Thanks @tatilepizs

> I was checking why the more-options-menu test was still failing for this branch (160-e2e-tests-top-bar), and it is because the kebab menu is not working on mobile, something happened 😕

@Benmuiruri already fixed that issue

tests/page-objects/default/common/common.wdio.page.js
Benmuiruri: CI is failing with `element (".fast-action-trigger .fast-action-fab-button") still not displayed after 15000ms`. 

I was facing that issue after the Angular exception fix. That's why I changed the selector to 

```
const NOT_MOBILE_ONLY = 'mm-fast-action-button:not(.mobile-only)';
const fastActionFAB = () => $(`${NOT_MOBILE_ONLY} ${FAST_ACTION_TRIGGER} .fast-action-fab-button`);
```

We had the conversation on [Slack](https://medic.slack.com/archives/D05QJT74F6E/p1725949112415229?thread_ts=1725948866.424229&cid=D05QJT74F6E)

latin-panda: Yes, but it's breaking the mobile tests, so I'm doing a different approach

Benmuiruri (approved):
LGTM & tests passing.

PR #9512, opened 2024-10-05
feat(#8119): add route guard to show training card confirm message

### LINKED ISSUE #8119: Show confirmation message in training cards when navigating to another page
When a user quits the training cards by clicking the "Cancel" button or the top-right "X" button, the app displays a confirmation message that says: `This training is not finished. You will lose your progress if you leave now. Are you sure you want to leave?`.

We should show this confirmation message when the user navigates to another page, for example:

1. The user logs in, and training cards show up.
2. The user clicks on the back button of the browser.
3. The app stops navigation and shows a confirmation message.
    3.1. The user confirms to quit training, and the app resumes navigation to the other page.
    3.2. The user prefers to stay, and the app resumes the training where it was interrupted. 

**Additional context**
The training cards were added to CHT Core as part of this [work](https://github.com/medic/cht-core/issues/6598).

Benmuiruri:
# Description

- When the training card is open and you enter a different route on the url bar, it displays confirm message before navigating. 
- When the training card is open and you click back, it displays confirm message before navigating. 
- This fix has been tested and works for Desktop Browser (Firefox and Chrome) && CHT Android app && Android Chrome PWA && Chrome on Android.

<details>
<summary>Video</summary>

https://github.com/user-attachments/assets/c00e1d18-6f90-4528-857d-8453f616168d

</details>

Closes #8119 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8119-training-card/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8119-training-card/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8119-training-card/docker-compose/cht-couchdb-clustered.yml)

# License


Benmuiruri:
Hi @kapelo now that I got this working. As I consider which unit tests I can add to `webapp/src/ts/modals/training-cards/training-cards.component.ts` could you help with e2e ? 

Perhaps adding a test case in `tests/e2e/default/enketo/training-cards.wdio-spec.js`.

webapp/src/ts/training-card.guard.provider.ts
```diff
@@ -0,0 +1,40 @@
...
+import { Selectors } from '@mm-selectors/index';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class TrainingCardDeactivationGuardProvider implements CanDeactivate<any> {
+  private globalActions;
+
+
+  constructor(private store: Store) {
```
latin-panda, line 14:
```suggestion
  private globalActions;
  
  constructor(private store: Store) {
```

latin-panda (changes requested):
Thanks Ben! I left a couple of comments below

webapp/src/ts/modals/training-cards/training-cards.component.ts
```diff
@@ -154,7 +162,7 @@ export class TrainingCardsComponent implements OnInit, OnDestroy {
   }
 
   close() {
-    this.globalActions.setTrainingCardFormId(null);
+    this.globalActions.setTrainingCard({ formId: null, isOpen: false, showConfirmExit: false, nextUrl: null });
```
latin-panda, line 165: I think it's safer to clean this in ngOnDestroy too.

webapp/src/ts/training-card.guard.provider.ts
```diff
@@ -0,0 +1,47 @@
...
+@Injectable({
+  providedIn: 'root'
+})
+export class TrainingCardDeactivationGuardProvider implements CanDeactivate<any> {
+  private globalActions;
+  private trainingCardOpen = false;
+  private formId: string | null = null;
+
+  constructor(private store: Store) {
+    this.globalActions = new GlobalActions(store);
```
latin-panda, line 17: The IDE marks the constructor's parameter `private store: Store)` as unused. The warning goes away by doing this: 
```suggestion
    this.globalActions = new GlobalActions(this.store);
```

webapp/src/ts/training-card.guard.provider.ts
```diff
@@ -0,0 +1,47 @@
...
+    currentState: RouterStateSnapshot,
+    nextState: RouterStateSnapshot,
+  ) {
+
+    this.store
+      .select(Selectors.getTrainingCard)
+      .subscribe(trainingCard => {
+        this.trainingCardOpen = trainingCard.isOpen;
+        this.formId = trainingCard.formId;
+      }).unsubscribe();
```
latin-panda, line 32:
```suggestion
      })
      .unsubscribe();
```

webapp/src/ts/training-card.guard.provider.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  canDeactivate(
+    component: any,
+    currentRoute: ActivatedRouteSnapshot,
+    currentState: RouterStateSnapshot,
+    nextState: RouterStateSnapshot,
+  ) {
+
+    this.store
```
latin-panda, line 27: This subscription is asynchronous code. It can get slightly delayed and execute the `  if (this.trainingCardOpen) {` code block below first with old values in `this.trainingCardOpen` and `this.formId = trainingCard.formId;`

How can we make it safer?

webapp/tests/karma/ts/modals/training-cards/training-cards.component.spec.ts
```diff
@@ -132,6 +140,8 @@ describe('TrainingCardsComponent', () => {
     xmlFormsService.get.resolves(xmlForm);
     formService.render.resolves(renderedForm);
     store.overrideSelector(Selectors.getTrainingCardFormId, 'training:a_form_id');
+    store
+      .overrideSelector(Selectors.getTrainingCard, { formId: '', isOpen: false, showConfirmExit: false, nextUrl: '' });
```
latin-panda, line 144:
```suggestion
    store.overrideSelector(
      Selectors.getTrainingCard, 
      { formId: '', isOpen: false, showConfirmExit: false, nextUrl: '' }
    );
```

webapp/tests/karma/ts/modals/training-cards/training-cards.component.spec.ts
```diff
@@ -149,8 +159,49 @@ describe('TrainingCardsComponent', () => {
...
+      showConfirmExit: false,
+      nextUrl: null,
+    }]);
+  }));
+
+  it('should get training card state from store', fakeAsync(() => {
+    sinon.resetHistory();
+    store
+      .overrideSelector(Selectors
+        .getTrainingCard, { formId: '', isOpen: true, showConfirmExit: true, nextUrl: '/next/page' });
```
latin-panda, line 174:
```suggestion
    store.overrideSelector(
      Selectors.getTrainingCard, 
      { formId: '', isOpen: true, showConfirmExit: true, nextUrl: '/next/page' }
    );
```

latin-panda (changes requested):
@Benmuiruri I did 50%, I'll review the other 50% later today

webapp/src/ts/training-card.guard.provider.ts
Benmuiruri: The magic of Webstorm. Vscode does not complain :(. 

How about the sonar complain that [router](https://github.com/medic/cht-core/pull/9512/files#diff-94d07e089bc01aba387ccbb407e1b7ed08b021c77672da72e2511b1a92b4cef9:~:text=private%20router%3A%20Router%2C) should be marked as readonly. Is that correct ? I was checking other parts of the code and I didn't see that.

webapp/src/ts/training-card.guard.provider.ts
Benmuiruri: How about updating canDeactivate to return an Observable<boolean> and use `take` to wait for the store's latest value before executing `if (this.trainingCardOpen)` ?

webapp/src/ts/training-card.guard.provider.ts
```diff
@@ -0,0 +1,47 @@
...
+      .select(Selectors.getTrainingCard)
+      .subscribe(trainingCard => {
+        this.trainingCardOpen = trainingCard.isOpen;
+        this.formId = trainingCard.formId;
+      }).unsubscribe();
+
+    if (this.trainingCardOpen) {
+      this.globalActions.setTrainingCard({
+        formId: this.formId,
+        isOpen: true,
```
latin-panda, line 37: So far, these two properties seem to have the same values as in the state, and there is no need to set them in the Guard. Did you spot a weird behavior that maybe I missed? 
```suggestion
      this.globalActions.setTrainingCard({
```

webapp/src/ts/training-card.guard.provider.ts
latin-panda, line 1: Sometimes, the guard doesn't display the confirmation modal, but in the background is navigating. Video:
https://github.com/user-attachments/assets/8814994c-fd8c-4803-b3b8-078e82e06749

latin-panda (changes requested):
@Benmuiruri a couple more comments:

webapp/src/ts/training-card.guard.provider.ts
Benmuiruri: No I did not, that was an oversight. I agree the guard does not need to update those values.

webapp/src/ts/training-card.guard.provider.ts
Benmuiruri: I added the route guard to the child routes. But that means guarding individual routes in different contexts, spread out in the different routing files. 

The other alternative we talked about is adding it to the app.routing module which is not ideal. But for consideration, this is how it would be. 

It's sort of  better since it's one place but it feels like it is polluting the file. I'd like to hear your thoughts. 


<details>
<summary>Possble app routing</summary>

```
const addTrainingRouteGuard = (routes: Routes) => {
  return routes.map(route => ({
    ...route,
    canDeactivate: [
      ...(route.canDeactivate || []),
      TrainingCardDeactivationGuardProvider
    ]
  }));
};

const routes: Routes = [
  ...homeRoutes,
  ...aboutRoutes,
  ...confUserRoutes,
  ...analyticsRoutes,
  ...addTrainingRouteGuard(reportRoutes),
  ...addTrainingRouteGuard(messagesRoutes),
  ...contactsRoutes,
  ...privacyPolicyRoutes,
  ...tasksRoutes,
  ...testingRoutes,
  ...errorRoutes,
];
```

</details>

latin-panda: It's a bit tricky. Wearing the shoes of someone unfamiliar with this feature and looking at the report route file, I wouldn't imagine that some other file is adding new guards like this. Instead, I would check the app guard file, thinking that's where all app-wide guards are located.

latin-panda: Did you try using to guard the parent route and child route? [For example here](https://github.com/medic/cht-core/blob/4470796bc1712297e520c7f22c14a999249fbbd0/webapp/src/ts/modules/analytics/analytics.routes.ts#L20)

```
canActivate: [ AppRouteGuardProvider ],
canActivateChild: [ AnalyticsRouteGuardProvider ],
```

webapp/src/ts/training-card.guard.provider.ts
```diff
@@ -0,0 +1,41 @@
...
+  providedIn: 'root'
+})
+export class TrainingCardDeactivationGuardProvider implements CanDeactivate<any> {
+  private globalActions;
+
+  constructor(private store: Store) {
+    this.globalActions = new GlobalActions(this.store);
+  }
+
+  canDeactivate(
```
latin-panda, line 19: [The privacy policy route also needs it ](https://github.com/medic/cht-core/blob/master/webapp/src/ts/modules/privacy-policy/privacy-policy.routes.ts)

webapp/src/ts/training-card.guard.provider.ts
Benmuiruri: Thanks for sharing your thoughts. I didn't know about `canActivateChild`. Let me test it out.

Benmuiruri: `canActivateChild` does not work, it prevents access into the route. It's not ideal for our case for navigating out of the child route. 

So I've added the guard directly to the child routes.

latin-panda (approved):
LGTM! ✨✨✨
Thanks Ben!

Issue #9489, opened 2024-10-02
Add telemetry for offline freetext searching


jkuester:
**Is your feature request related to a problem? Please describe.**

During our research for https://github.com/medic/cht-core/issues/9293 it has become clear that we currently cannot make any changes to the offline freetext search experience without potentially affecting user workflows because we simply have no way to be sure how users are actually interacting with the freetext search (for both contacts and reports).

For example, in interviews with users, they mention searching by contact name.  However, in practice we know that contact names are often de-normalized into various custom fields on both contacts and reports.  So, simply indexing the `name` fields on contacts may not be enough to support user workflows.  Most importantly, we just do not have any way to know the answer here one way or the other.  The users, themselves, are probably not aware of the exact field structure on the documents they are searching for.  

**Describe the solution you'd like**

One reason we are not currently collecting any search telemetry data is because it is likely that the search terms provided by the user will contain PII and we want to avoid logging PII in telemetry docs.

The proposal here is to add telemetry that is more holistically focused on _the search workflow_.  Instead of merely logging what terms the user searches for, we want to focus on the whole process whereby a user searches for something _and then actually selects ones of the filtered reports/contacts._  Once a user chooses a report/contact to navigate to, we do a comparison between the contents of the report/contact with the user's search term and find the _doc fields_ that were matched by the search query.  Then we can log telemetry indicating a search happened (but do not  include the exact search terms).  Instead we can include information about the report/contact that the user ended up selecting and espeically _the names of the doc fields that were matched by the search query_.  In this way, we avoid including PII in the telemetry, but can still construct an (imperfect) image of the kinds of things users are searching for....

**Describe alternatives you've considered**

Not collecting any telemetry means that any changes we make to offline freetext search functionality will be done without any way of realistically quantifying their impact.

OTOH, if we log PII in telemetry, it would make it more difficult to collect telemetry data from partners.  Extra care must be taken when storing/transmitting this kind of info.

PR #9525, opened 2024-10-10
feat(#9489): collect telemetry for offline freetext searching

m5r:
# Description

#9489 

# Code review checklist

# License

### DIFF
--- a/webapp/src/ts/modules/contacts/contacts-content.component.ts
+++ b/webapp/src/ts/modules/contacts/contacts-content.component.ts
@@ -1,5 +1,5 @@
 import { Component, OnDestroy, OnInit } from '@angular/core';
-import { Store } from '@ngrx/store';
+import {createSelector, Store} from '@ngrx/store';
 import { ActivatedRoute, Router } from '@angular/router';
 import { combineLatest, Subscription } from 'rxjs';
 import { first, take } from 'rxjs/operators';
@@ -112,15 +112,35 @@ export class ContactsContentComponent implements OnInit, OnDestroy {
     this.subscriptions.add(subscription);
   }
 
+  private recordSearchTelemetry(selectedContact, nextSelectedContact, nextFilters) {
+    if (
+      !nextFilters?.search ||
+      !nextSelectedContact
+    ) {
+      console.log("SKIP no filters or no selected contact");
+      return null;
+    }
+
+    // user searched for something and now selects a contact
+    if (
+      selectedContact === null || // had no contact selected
+      selectedContact._id !== nextSelectedContact._id // or had a different contact selected
+    ) {
+      console.log({ nextFilters, nextSelectedContact });
+      // TODO: find which properties matched user's query
+      // TODO: record telemetry
+    }
+  }
+
   private subscribeToStore() {
-    const reduxSubscription = combineLatest(
+    const reduxSubscription = combineLatest([
       this.store.select(Selectors.getSelectedContact),
       this.store.select(Selectors.getForms),
       this.store.select(Selectors.getLoadingContent),
       this.store.select(Selectors.getLoadingSelectedContactReports),
       this.store.select(Selectors.getContactsLoadingSummary),
       this.store.select(Selectors.getFilters),
-    ).subscribe(([
+    ]).subscribe(([
       selectedContact,
       forms,
       loadingContent,
@@ -128,6 +148,7 @@ export class ContactsContentComponent implements OnInit, OnDestroy {
       contactsLoadingSummary,
       filters,
     ]) => {
+      this.recordSearchTelemetry(this.selectedContact, selectedContact, filters);
       if (this.selectedContact?._id !== selectedContact?._id) {
         // reset view when selected contact changes
         this.resetTaskAndReportsFilter();


shared-libs/search/src/generate-search-requests.js
```diff
@@ -191,7 +191,7 @@ const sortByLastVisitedDate = () => {
 const makeCombinedParams = (freetextRequest, typeKey) => {
   const type = typeKey[0];
   const params = {};
-  if (freetextRequest.key) {
+  if (freetextRequest.params.key) {
```
m5r, line 194: This is a fix for `key:value` queries against the view `contacts_by_type_freetext`. Without this, these queries would fail silently and the select2 input would show an infinite loader until a different query would return results. If this has never been noticed before I believe the `key:value` search feature isn't used very much...

webapp/src/ts/modules/contacts/contacts-content.component.ts
```diff
@@ -55,21 +55,21 @@ export class ContactsContentComponent implements OnInit, OnDestroy {
...
-    private store: Store,
-    private route: ActivatedRoute,
-    private router: Router,
-    private changesService: ChangesService,
-    private contactChangeFilterService: ContactChangeFilterService,
-    private xmlFormsService: XmlFormsService,
-    private modalService: ModalService,
-    private contactTypesService: ContactTypesService,
-    private settingsService: SettingsService,
-    private userSettingsService: UserSettingsService,
```
m5r, line 67: `userSettingsService` wasn't used so I did some cleaning

webapp/src/ts/modules/contacts/contacts-content.component.ts
```diff
@@ -112,22 +112,36 @@ export class ContactsContentComponent implements OnInit, OnDestroy {
...
-    ).subscribe(([
+    ]).subscribe(([
       selectedContact,
       forms,
       loadingContent,
       loadingSelectedContactReports,
       contactsLoadingSummary,
       filters,
     ]) => {
+      this.recordSearchTelemetry(this.selectedContact, selectedContact, filters);
```
m5r, line 144: We don't need to block the main thread by awaiting for this promise here, it can run in the background just fine

webapp/src/ts/services/select2-search.service.ts
```diff
@@ -77,17 +83,39 @@ export class Select2SearchService {
     this.searchService
       .search('contacts', filters, searchOptions)
       .then((documents) => {
-        if (currentQuery === params.data.q) {
```
m5r, line 80: This would always be `true` because `currentQuery` is defined and assigned a value in the same scope here. By pulling it into a class field, it can now be used to discard outdated results

tests/e2e/default/telemetry/telemetry.wdio-spec.js
```diff
@@ -6,41 +6,55 @@ const placeFactory = require('@factories/cht/contacts/place');
...
+    },
+    contact: {
+      _id: contact._id,
+      parent: contact.parent,
+    },
+    from: contact.phone,
+  });
+  let reportDocs;
 
   before(async () => {
```
tatilepizs, line 55: Please create an after hook to delete the users and DB, something like this:

```.js
after(async () => {
    await utils.deleteUsers([ user ]);
    await utils.revertDb([/^form:/], true);
  });
```

tatilepizs (approved):
Looks great @m5r thank you for reorganizing the `telemetry.wdio-spec.js` test ⭐ 
I only left one comment, that is not a blocker

_PS: I only checked the tests section._

shared-libs/search/src/generate-search-requests.js
jkuester: 🤯

webapp/src/ts/services/select2-search.service.ts
```diff
@@ -10,21 +10,26 @@ import { SessionService } from '@mm-services/session.service';
...
 import { TranslateService } from '@mm-services/translate.service';
+import { SearchTelemetryService } from '@mm-services/search-telemetry.service';
 
 @Injectable({
   providedIn: 'root'
 })
 export class Select2SearchService {
+  private selectEl;
+  private currentQuery: string | null = null;
+  private onContactSelect;
```
jkuester, line 21: I don't think it is going to work to have this state here at the service level.  I double-checked and confirmed that if you have a form with more than one `select-contact` field, then the `init` function below is going to get called for each of the contact selectors in the form.  (And this state will only be set for the last one).

Maybe I am missing something here, but is there a reason we cannot setup the `select2:select` event listener somewhere like the `initSelect2` method (for each given selector) instead of doing it in the generic `query` method?

webapp/src/ts/services/select2-search.service.ts
```diff
@@ -77,17 +83,39 @@ export class Select2SearchService {
...
+        this.onContactSelect = async (event) => {
+          this.onContactSelect = null;
+
+          const docId = event.params?.data?.id;
+          if (!search || !docId) {
+            return;
+          }
+
+          const doc = await this.getDoc(docId);
+          await this.searchTelemetryService.recordContactByTypeSearch(doc, search);
```
jkuester, line 104: Minor, but technically I think this will only hit the `contacts_by_type_freetext` view if the `types` param above is valued.  Otherwise, I guess `shared-libs/search` will just route this to `contacts_by_freetext`...

webapp/src/ts/services/search-telemetry.service.ts
```diff
@@ -0,0 +1,71 @@
...
+  private readonly REPORT_SKIPPED_PROPERTIES = ['_id', '_rev', 'type', 'refid', 'content'];
+
+  constructor(
+    private readonly telemetryService: TelemetryService,
+  ) {}
+
+  public async recordContactSearch(contactDoc: Record<string, any>, search: string) {
+    const matchingProperties = await this.findMatchingProperties(contactDoc, search, this.CONTACT_SKIPPED_PROPERTIES);
+
+    for (const key of matchingProperties) {
```
jkuester, line 19: Minor, but is there any reason we need to do this sequentially instead of all at once via `Promise.all`?

webapp/src/ts/services/search-telemetry.service.ts
```diff
@@ -0,0 +1,71 @@
...
+  private async findMatchingProperties(
+    doc: Record<string, any>,
+    search: string,
+    skip: string[],
+    basePropertyPath = '',
+  ) {
+    const matchingProperties = new Set<string>();
+    const colonSearch = search.split(':');
+    if (colonSearch.length > 1) {
+      matchingProperties.add(`${colonSearch[0]}:$value`);
```
jkuester, line 53: This is just capturing the fact that someone did a colon search right?  

In this case, should we be adjusting the `_search` value to be `colonSearch[1]`?  OTOH, maybe we should just return with [`${colonSearch[0]}:$value`] and not bother checking the rest of the doc fields...

webapp/src/ts/services/search-telemetry.service.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    const _search = search.toLowerCase();
+    Object.entries(doc).forEach(([key, value]) => {
+      const _key = key.toLowerCase();
+      if (skip.includes(_key) || _key.endsWith('_date')) {
+        return;
+      }
+
+      const propertyPath = basePropertyPath ? `${basePropertyPath}.${key}` : key;
```
jkuester, line 63: Minor, can move this up inside the if-block

jkuester (changes requested):
Very nice!  Great e2e tests!

webapp/src/ts/services/select2-search.service.ts
m5r: That's a good catch... I landed with registering this event listener in the `query` method because this is where I had access to the user's search to hook up with the contact the user would end up selecting.
I also couldn't find anything in [select2 docs](https://select2.org/) to retrieve the search text programmatically without writing code that looks like a pile of hacks with a mix of jQuery and DOM element querying :sweat_smile: 

But! I think I found [a solution](https://stackoverflow.com/a/44678950). I'll come back and resolve this convo if this works.

webapp/src/ts/services/select2-search.service.ts
m5r: Right, but `select2-search` is only ever used to retrieve contacts by type, no?

m5r: Ah no you're right, there is one case where this service is used without type `types` parameter
https://github.com/medic/cht-core/blob/b7b2629bfe92329f9974e94adc6786bc687f3202/webapp/src/ts/modals/send-message/send-message.component.ts#L187

I'll handle this scenario too, thanks!

m5r: `Select2SearchService` is initialized twice in the codebase. First occurrence in https://github.com/medic/cht-core/blob/b7b2629bfe92329f9974e94adc6786bc687f3202/webapp/src/ts/modals/send-message/send-message.component.ts#L174-L189
And the second occurrence in https://github.com/medic/cht-core/blob/59b42e2fa3271a69c713b4720b9ee9fb7ea3dc61/webapp/src/ts/modals/edit-report/edit-report.component.ts#L42-L53

Both times it's initialized with the `types` parameter and then that gets translated into a query against `medic-client/contacts_by_type` in https://github.com/medic/cht-core/blob/843888c13456919f50daf12449611ed28c732e5c/shared-libs/search/src/generate-search-requests.js#L137-L148
and its results get "merged" with the freetext query results in https://github.com/medic/cht-core/blob/843888c13456919f50daf12449611ed28c732e5c/shared-libs/search/src/generate-search-requests.js#L274-L276

webapp/src/ts/services/search-telemetry.service.ts
m5r: You're right we can just return early and not bother looping over the doc fields

webapp/src/ts/modules/contacts/contacts-content.component.ts
m5r: I've only ever seen the `void` operator used in minified JS code to shorten `undefined` to `void 0` and gain 3 bytes for each occurrence :joy: 

But looking into your suggestion, it seems to be what typescript devs @ microsoft [recommend](https://github.com/microsoft/TypeScript/issues/13376#issuecomment-273289748) when you want to fire and forget a promise

+1 to start doing that

webapp/src/ts/services/select2-search.service.ts
```diff
@@ -177,16 +183,25 @@ export class Select2SearchService {
...
-
-        if (docId) {
-          this.getDoc(docId)
-            .then(doc => {
-              this.setDoc(selectEl, doc);
-              selectEl.trigger('change');
-            })
-            .catch(err => console.error('Select2 failed to get document', err));
+      let search: string | undefined;
+      selectEl.on('select2:selecting', () => search = selectEl.data('select2').dropdown.$search.val());
```
m5r, line 187: @jkuester since the search input gets cleared before the `select2:select` event is called, this is the next best place to get its content before it gets cleared

webapp/src/ts/services/select2-search.service.ts
jkuester: Select2Search is also inited by the db-object-widget: https://github.com/medic/cht-core/blob/45478853727d40b60b78c6359fb11fc7606d4a9c/webapp/src/js/enketo/widgets/db-object-widget.js#L59

I did some double-checking and confirmed that when you use the `select-contact` appearance _without a `type-*` appearance, it will call through to just the `contacts_by_freetext` view (since the types in the search request are empty).

webapp/src/ts/services/search-telemetry.service.ts
jkuester: :+1:  In that case, seems reasonable to just do this however you feel is more readable.

webapp/src/ts/services/select2-search.service.ts
jkuester: > search input gets cleared before the select2:select

That is super annoying, but yeah, in that case, this seems like a fine approach!  I assume `select2:selecting` is just getting repeatedly triggered as the user types in the search box?

m5r: No that's the good part, it gets triggered right before a result is selected aka right before the `select2:select` event we have a listener for one line below.

webapp/src/ts/services/select2-search.service.ts
m5r: 🫡 your enketo-fu is always appreciated, I'll handle this case

m5r: Okay I pushed what I think will cover this scenario but I'm about to drop off. I've noted to myself to manually test it and also add unit tests for this particular case

m5r: Okay there was no straightforward way to unit test this so I went for an e2e test :slightly_smiling_face:

shared-libs/search/src/generate-search-requests.js
jkuester: Thinking about this again, I feel like we really should include a unit test for this case (if possible).

webapp/src/ts/services/select2-search.service.ts
```diff
@@ -58,8 +61,7 @@ export class Select2SearchService {
   }
 
   private query(params, successCb, failureCb, options, types) {
-    const currentQuery = params.data.q;
-
+    this.currentQuery = params.data.q;
```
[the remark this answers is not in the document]
jkuester: Hmm, wait, I see that the `this.currentQuery !== params.data.q` check is happening _async_ in the code-flow below (after the `searchService.search` call).  So, referencing back to your earlier comment, having this as a class field allows a newer search to preempt an earlier one.  But, since we now know it is possible to have multiple references to this service from multiple elements at the same time, I am concerned that having this as a class field is just inviting weird conflicts/race-conditions (perhaps around enketo auto-triggering updates to contact selectors as things get recalculated?).  

Maybe it would be better to just leave this as a local variable and then remove the unnecessary check below...

webapp/src/ts/services/select2-search.service.ts
jkuester: :+1: You would think in 2024 we would have a half-way decent way to unit-test code that involves jquery managed browser events.  So painful....

webapp/tests/karma/ts/modules/contacts/contacts-content.component.spec.ts
```diff
@@ -138,8 +144,10 @@ describe('Contacts content component', () => {
           { provide: ModalService, useValue: modalService },
           { provide: ResponsiveService, useValue: responsiveService },
           { provide: ContactMutedService, useValue: contactMutedService },
+          { provide: TelemetryService, useValue: telemetryService },
```
jkuester, line 147: Since we are stubbing out the `SearchTelemetryService`, I would not expect anything to be hitting the `TelemetryService`?

webapp/tests/karma/ts/modules/contacts/contacts-content.component.spec.ts
```diff
@@ -241,6 +249,23 @@ describe('Contacts content component', () => {
...
+    store.overrideSelector(Selectors.getFilters, { search });
+    store.refreshState();
+    expect(searchTelemetryService.recordContactSearch.callCount).to.equal(0);
+
+    const contact = { _id: 'contact_id', doc: { case_id: 'abc-1234' } };
+    store.overrideSelector(Selectors.getSelectedContact, contact);
+    store.refreshState();
+    expect(searchTelemetryService.recordContactSearch.callCount).to.equal(1);
+    expect(searchTelemetryService.recordContactSearch.getCall(0).args).to.deep.equal([contact.doc, search]);
+  }));
```
jkuester, line 267: Is it also possible to hit the case where the same contact is re-selected?

webapp/tests/karma/ts/modules/reports/reports-content.component.spec.ts
```diff
@@ -88,7 +93,9 @@ describe('Reports Content Component', () => {
           FormIconPipe,
           TitlePipe,
           RelativeDatePipe,
-          FastActionButtonComponent
+          FastActionButtonComponent,
+          SenderComponent,
+          ReportVerifyInvalidIconComponent,
```
jkuester, line 98: I am curious about why we needed to add these here?  🤔  Seems like the only new code flows we should be hitting involve the SearchTelemetryService, but I guess maybe the existing tests were not covering everything?

webapp/tests/karma/ts/modules/reports/reports-content.component.spec.ts
```diff
@@ -490,5 +498,25 @@ describe('Reports Content Component', () => {
...
+      fixture.detectChanges();
+      expect(searchTelemetryService.recordReportSearch.callCount).to.equal(0);
+
+      const report = { _id: 'report_id', doc: { case_id: 'abc-1234' } };
+      store.overrideSelector(Selectors.getSelectedReport, report);
+      store.refreshState();
+      fixture.detectChanges();
+      expect(searchTelemetryService.recordReportSearch.callCount).to.equal(1);
+      expect(searchTelemetryService.recordReportSearch.getCall(0).args).to.deep.equal([report.doc, search]);
+    }));
```
jkuester, line 520: Is it possible to also cover the edge cases here? 

- Multiple reports selected
- Same report re-selected

jkuester (changes requested):
Sorry, accidentally submitted the previous review before I was done making comments!  (I blame WebStorm.  😓)

Things are looking really good here! The main concern is just the possible PII leak with the colon separated logic. 🙏

webapp/src/ts/services/search-telemetry.service.ts
```diff
@@ -0,0 +1,73 @@
...
+      matchingProperties.map(key => this.telemetryService.record(`search_match:contacts_by_type_freetext:${key}`)),
+    );
+  }
+
+  public async recordReportSearch(reportDoc: Record<string, any>, search: string) {
+    const [matches, fieldsMatches] = await Promise.all([
+      this.findMatchingProperties(reportDoc, search, this.REPORT_SKIPPED_PROPERTIES),
+      this.findMatchingProperties(reportDoc.fields, search, this.REPORT_SKIPPED_PROPERTIES, 'fields'),
+    ]);
+    const matchingProperties = Array.from(new Set([...matches, ...fieldsMatches]));
```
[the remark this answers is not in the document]
m5r: There could be a collision with the `column:value` search here

webapp/src/ts/services/select2-search.service.ts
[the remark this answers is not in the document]
m5r: Right I think you're correct here... but at the same time this local variable becomes irrelevant as it's just an alias for `params.data.q` and the checks below are always true
Maybe I'm missing something about the scope of this variable but the way I read it, it can never be unequal to `params.data.q`. I'll remove `currentQuery` and its associated checks and we'll see if the CI breaks :man_shrugging: but I expect all tests to keep passing

webapp/tests/karma/ts/modules/reports/reports-content.component.spec.ts
m5r: Because opening a report renders these two components on the page and the test runner was complaining about "`<mm-sender>` is not a known element"

webapp/src/ts/services/search-telemetry.service.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  private async findMatchingProperties(
+    doc: Record<string, any>,
+    search: string,
+    skip: string[],
+    basePropertyPath = '',
+  ) {
+    const colonSearch = search.split(':');
+    if (colonSearch.length > 1) {
+      return [`${colonSearch[0]}:$value`];
```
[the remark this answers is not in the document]
m5r: After reading your proposal I was a bit worried that we'd lose out on the `key:value` searches but we kind of reached the conclusion that it's not used by anyone so it makes sense to not handle these queries. It would have been nice to have the data to back up that claim but risking to leak PII to get this data is not worth it.
So instead of continuing the search with `colonSearch[1]`, I think we should remove the `colonSearch`-related code from line 50 to 53, what do you think @jkuester?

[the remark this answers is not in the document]
m5r: Coming back to this, I think your second approach is better. So removing line 50-53 and at the end of the function, re-run it with `colonSearch[1]` only if `matchingProperties` is empty.

jkuester (approved):
Most excellent!  The updated logic in `select2-search.service.ts` is super clean!

Issue #9523, opened 2024-10-10
Upgrade WebdriverIO to v9 and get the new features!


latin-panda:
**Describe the issue**
WebdriverIO v9 is out with some features we can use, [like fake timers ](https://webdriver.io/blog/2024/08/15/webdriverio-v9-release/#fake-timers-support)

[But there are some deprecations that we might need to test and resolve.](https://webdriver.io/blog/2024/08/15/webdriverio-v9-release/#notable-breaking-changes)

PR #9556, opened 2024-10-17
chore(#9523): upgrade WebdriverIO to v9

kapelo:
# Description

This PR upgrades wdio to the v9 and refactors tests that will fail as a result of this upgrade.

# Code review checklist

# License


tatilepizs:
Hi @kapelo, 
Since there is already a bug open in the webdriverio project for the issue with the `reloadSession` method, I think that we should continue with the upgrade but add the `'wdio:enforceWebDriverClassic': true` in the [capabilities](https://github.com/medic/cht-core/blob/9523-upgrade-wdio-v9/tests/wdio.conf.js#L92-L115) to disable the Bidi protocol. I already ran the tests with that flag, and they passed fine. And once the bug is fixed, we can remove the flag.

What do you think?

PR #9563, opened 2024-10-19
feat(#9426): validate contact_type against parent_id in url when creating contact

### LINKED ISSUE #9426: Prevent creating contacts under facilities that are not the direct parent
It's possible to create a contact under another that is not its direct parent. 

In app-settings, we define the hierarchy. Take the one in the `default` config as an example, [where `clinic` direct parent is the `health_center`](https://github.com/medic/cht-core/blob/488bbd6148075ea8cbc575729061e019ecda7004/config/default/app_settings.json#L331). It's possible to create a `clinic` under `district_hospital` by changing the URL like this: 
`#/contacts/[district-hospital-id]/add/clinic`

The Contact Edit component can validate that the parent_id in the URL is a contact type directly associated of contact_type form. 

**Additional context**
It can help preventing errors [like this one reported in the forum](https://forum.communityhealthtoolkit.org/t/bizarre-recursive-place-case/3825/9)

Benmuiruri:
# Description

This PR  updates the Contact Edit component to validate that the `parent_id` in the URL is a contact type directly associated of `contact_type` form. 

It validates the contact_type being created should be a direct child of the parent based on the `parent_id` in the URL. 

<details>
<summary>Creating Contact Type that's a direct child of parent works correctly without errors</summary>

https://github.com/user-attachments/assets/b32d62d8-f6eb-47d0-8514-46ac66abe40a

</details>

<details>
<summary>Creating Contact Type that is not a direct child of parent fails</summary>

https://github.com/user-attachments/assets/841f5569-bbce-4941-8914-2a41c87bc0a2

</details>

Closes #9426 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9425-create-contact/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9425-create-contact/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9425-create-contact/docker-compose/cht-couchdb-clustered.yml)

# License


Benmuiruri:
Hey @jkuester 👋 

This is ready for review when you have some time. 

Let me know if it adequately addresses the concern raised in the forum. It took me a second to understand the problem. 

PS - My thinking is that backward compatibility won't be an issue since if you have an already existing `contact_type` that is not a direct child of the parent at the time of creation, it will still be in the same "incorrect place" and you can still view, edit or delete it. This fix is just for new contacts to come in the future.

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -221,6 +228,21 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
     return formId;
   }
 
+  private async validateParent(parentId, childTypeId) {
```
jkuester, line 231: These are just suggestions, so take what makes sense to you and leave the rest.

- If a function is called 'validate...' I assume it is going to _fail_ for invalid data (instead of just return `false`).
- IMHO the `LineageModelGeneratorService` is overkill here since we don't care about hydrating the lineage of the parent.  Can just pull in the `ChtDatasourceService` instead.

```ts
  private async validateParentForCreateForm() {
    const getPlace = await this.chtDatasourceService.bind(Place.v1.get);
    const parent = await getPlace({ uuid: this.contact.parent });
    const parentType = this.contactTypesService.getTypeId(parent);
    const validParentType = !this.contact.parent || parentType;
    const isValidChildType = () => this.contactTypesService
      .getChildren(parentType)
      .then((validChildTypes) => validChildTypes.some(({ id }) => id === this.contact.contact_type));
    if (!validParentType || !await isValidChildType()) {
      throw new Error(`Cannot create a ${this.contact.contact_type} as a child of a ${parentType}.`);
    }
  }
```

jkuester (approved):
Couple suggestions, but otherwise I tested this all locally and worked great!

garethbowen:
@Benmuiruri I think this should not be a "chore" type, right? It's seems like a "feat"?

This matters because at some point commit messages will be used to generate release notes and version numbers - "chore" won't even trigger a release.

Benmuiruri:
@garethbowen I usually think of `feat` as something "new". But, I've just looked at the documentation and you're right. `feat` refers to changes users will notice. I'll change it

webapp/src/ts/modules/contacts/contacts-edit.component.ts
Benmuiruri: I've gone with `ChtDatasourceService`, thanks for the suggestion.  But I went with the imperative mode, which I find easier to read and write.

garethbowen:
@Benmuiruri Yeah, I do too. With issues we had a difference between features (something new) and improvements, but the standard commitlint just has "feat". I guess for release notes and version numbers it doesn't make difference - it's a minor release either way...

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -162,6 +162,13 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
         throw new Error(`Unknown contact type "${contactTypeId}"`);
       }
 
+      const parentId = this.routeSnapshot.params?.parent_id;
+      const childTypeId = this.routeSnapshot.params?.type;
+      const validateParent = await this.validateParent(parentId, childTypeId);
```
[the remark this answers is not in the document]
Benmuiruri: I'd initially done the validation in `getForm` but error propagation influenced me to move the logic. On second thought after your suggestion I think it's better to validate after creating the contact in `getForm`

I've refactored it, please give it a look-see before I merge.

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -216,11 +218,29 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
...
       formId = contactType.create_form;
+
+      await this.validateParentForCreateForm();
     }
 
     return formId;
   }
 
+  private async validateParentForCreateForm() {
+    if (!this.contact.parent) {
```
jkuester, line 229: So, the edge-case logic here is super jank.  

- `datasource.v1.place.getByUuid` will return `null` if no _place_ is found for the identifier (e.g. if `this.contact.parent` is a `person`).
- `contactTypesService.getTypeId` will return `undefined` when you give it a nullish `parent`
- `contactTypesService.getChildren` will return the "orphan" (aka "top level") types (e.g. `district_hospital`) when you give it a nullish `parentType`.  

The combination of these means that with your current logic it will still let me add a `district_hospital` as the _child_ of a `person`.

---

Additionally, another edge case that just occurred to me that is not covered is that we should throw an error if someone tries to add a contact _to the top level_ when the type of that contact is not an "orphan" (aka when there are parents configured for the contact type).  So, the user should be able to fill out a form at `contacts/add/district_hospital` since there are no parents for that type, but they should not be able to fill out `contacts/add/person` since `person` is configured as a child type.

jkuester (changes requested):

webapp/src/ts/modules/contacts/contacts-edit.component.ts
Benmuiruri: Thanks @jkuester . I wanted to get your review early so that we don't skip a day. I've made the checks very explicit.

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -216,11 +218,44 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
...
+    }
+
+    const parent = await datasource.v1.place.getByUuid(this.contact.parent);
+    if (!parent) {
+      throw new Error(`Parent with UUID ${this.contact.parent} does not exist.`);
+    }
+
+    if (await this.contactTypesService.isPerson(parent)) {
+      throw new Error(`Cannot create a ${this.contact.contact_type} as a child of a person.`);
+    }
```
jkuester, line 246: Sorry for going back and forth on this. :facepalm: Seeing this `isPerson` check got me to wondering and then digging/testing.  It appears that we actually do not have anything currently preventing a configuration where `person` contacts are parents of other types of contacts.  I have logged https://github.com/medic/cht-core/issues/9584 seeking more clarity on this issue for the future, but for now, I don't think we want to take a strong position in this code.

Instead, I think we should switch to just use the `dbService` instead of the `datasource` to get the parent (since the `datasource` does not yet support just getting generic "contacts").  One "advantage" to this approach is that Pouch should throw an error if it cannot find the parent. That means we can eliminate these two if-blocks and make Sonar happy!  :+1: 

```suggestion
    const parent = await this.dbService
      .get()
      .get(this.contact.parent);
```

jkuester (approved):
The explicit usage of the `contactTypeService` functions is a good idea and will make the behavior of this code less surprising to readers in the future.  

I left one more comment in the code regarding the `person` contact check.  Once that is done, this should be good to go!

Issue #9552, opened 2024-10-16
Error when accessing tasks


dianabarsan:
**Describe the bug**
When loading tasks or targets for the first time after the upgrade to 4.13, this error can occur. 
![Image](https://github.com/user-attachments/assets/6e1faf66-5271-45f0-b693-29237b922a5e)

**To Reproduce**
1. Install 4.12
2. Create an offline user. Log in with this offline user and create one pregnancy. Logout. 
4. Login as admin and upgrade to 4.13.x . Logout. 
5. Log in as the offline user again and navigate to tasks or targets page. 
6. See error. 

**Expected behavior**
No error. 

**Logs**
```
main.js:1 Error getting targets TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at Object.aggregateStoredTargetEmissions (main.js:1:1309050)
    at Object.<anonymous> (main.js:1:1306901)
    at Generator.next (<anonymous>)
    at w (main.js:11:600733)
    at x (main.js:11:600945)
    at main.js:11:601004
    at new r (polyfills.js:1:22887)
    at Object.<anonymous> (main.js:11:600886)
    at A.aggregateStoredTargetEmissions (main.js:1:1307030)
```

**Screenshots**

**Environment**
- App: webapp
- Version: 4.13

**Additional context**
This is due to a change in structure of the local rulesStateStore document.

dianabarsan:
Fixed in master and backported to 4.13.x

kennsippell:
Reactivating due to similar bug:

1. Install 4.12
1. Create an offline user. Log in with this offline user and create one pregnancy. Logout.
1. Login as admin and upgrade to 4.13.x . Logout.
1. **Move calendar forward to start of next month**
1. Log in as the offline user again and navigate to tasks or targets page.
1. See error.

kennsippell:
Yes

garethbowen:
Second patch has been merged and backported.

PR #9569, opened 2024-10-19
fix(#9552): migrate stale target state for interval turnover

dianabarsan:
# Description

#9552

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9552-interval-turnover/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9552-interval-turnover/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9552-interval-turnover/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/shared-libs/rules-engine/src/rules-state-store.js
+++ b/shared-libs/rules-engine/src/rules-state-store.js
@@ -42,6 +42,7 @@ const self = {
 
     const rulesConfigHash = hashRulesConfig(settings);
     if (state.rulesConfigHash !== rulesConfigHash || targetState.isStale(state.targetState)) {
+      state.targetState = targetState.migrateStaleState(state.targetState);
       state.stale = true;
     }
 
--- a/shared-libs/rules-engine/src/target-state.js
+++ b/shared-libs/rules-engine/src/target-state.js
@@ -102,6 +102,13 @@ const mergeEmissions = (state, contactIds, targetEmissions) => {
   return isUpdated;
 };
 
+const createState = (existentStaleState) => {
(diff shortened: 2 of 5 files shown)


dianabarsan:
I was really tempted to do this last time I "fixed" this, but thought it was too hard-handed. 
And now I'm only doing it on load, because I think migrating it pre-calculation is also too hard-handed. 
Leaving this for the reviewer to decide.

shared-libs/rules-engine/src/rules-state-store.js
```diff
@@ -42,6 +42,7 @@ const self = {
 
     const rulesConfigHash = hashRulesConfig(settings);
     if (state.rulesConfigHash !== rulesConfigHash || targetState.isStale(state.targetState)) {
+      state.targetState = targetState.migrateStaleState(state.targetState);
```
garethbowen, line 45: It's weird to call this and immediately check `isStale` again - it would be better to split this into two `if` statements so we only call `migrateStale` when it's stale.

garethbowen (approved):
One stylistic preference, but approved to expedite the patch.

shared-libs/rules-engine/src/rules-state-store.js
dianabarsan: thanks for the review!!!

Issue #9558, opened 2024-10-18
Household filtering within task view


ShaunKrog:
Refers to #5886 (Make it easier to complete all tasks at a household in the same visit), which suggested a possible solution where tasks can be grouped by household so that tasks for the same household always appear next to each other. 

**Is your feature request related to a problem? Please describe.**
The solution to issue 5886 focuses on implementing task grouping, which is certainly useful, but does not necessarily improve filtering within the task view itself. We anticipate an overload of tasks within this view, making it difficult for the CHW to effectively plan their field trip.

**Describe the solution you'd like**
We have down streamed the integration of a slide filter similar to the one in the reporting view in our local implementation. We have kept the user experience consistent with how the filter functions in the reporting view and have exposed the ability for the CHW to filter tasks by household. However, just to note that we have not implemented the freetext search bar, as this is currently out of scope, but this may be a useful inclusion in a future release.

1. Include a filter icon on the task view (filter icon has been extracted from the reporting view search bar)

![Image](https://github.com/user-attachments/assets/138e8a86-d346-485a-8423-292e393275fe)

2. Include slide filter, allowing filter by place

![Image](https://github.com/user-attachments/assets/56f5eaeb-6677-4acc-b397-b82c37281734)

3. Resulting in an overall task view that looks as follows

![Image](https://github.com/user-attachments/assets/8cdc7b59-6d03-4d10-b8ea-08c5299588e3)

As an introduction to myself, I am new to the CHT development environment and would very much like to contribute to the community. I have about 20 years of experience as a full-stack C# developer, focusing on system implementation in the health and social domains (across Africa). We are using CHT in South Africa (Western Cape Department of Health) and I work with Anro Swart, who is already active within this community.

We would like to see if there is any possibility of including these changes into cht-core? Looking forward to being part of the community.

n-orlowski:
Hi @ShaunKrog ! Thank you for joining the community and for your contribution, we're glad to have you 🎉 

> We have kept the user experience consistent with how the filter functions in the reporting view and have exposed the ability for the CHW to filter tasks by household. 

Just to confirm after looking at the screenshots, did you mean filter by _place_? 

I'm also curious to learn more about the content in the 3rd screenshot where the task details are displayed. Is this a workaround for a user journey that's not currently supported? (also assuming the task itself isn't part of this feature but please correct me if I'm wrong!)

Some quick design notes:
- The bounding box of the current filter is actually the search input and not a button, so without the search bar it should look like:
[...]

PR #9576, opened 2024-10-23
feat(#9558): Task filter by household

ShaunKrog:
# Description

We have down streamed the integration of a slide filter similar to the one in the reporting view in our local implementation. We have kept the user experience consistent with how the filter functions in the reporting view and have exposed the ability for the CHW to filter tasks by household. However, just to note that we have not implemented the freetext search bar, as this is currently out of scope, but this may be a useful inclusion in a future release.

1. Include a filter icon on the task view (filter icon has been extracted from the reporting view search bar)
2. Include slide filter, allowing filter by place

Issue Number #9558 

# Code review checklist

# License

### DIFF
--- a/webapp/src/css/inbox.less
+++ b/webapp/src/css/inbox.less
@@ -473,7 +473,7 @@ mm-analytics-filters {
   background-color: @background-color;
 }
 
-.reports .inbox-items, .contacts .inbox-items, .messages .inbox-items {
+.reports .inbox-items, .contacts .inbox-items, .messages .inbox-items, .tasks .inbox-items {
   overflow-y: hidden;
   overflow-x: hidden;
   height: 100%;
@@ -1110,6 +1110,49 @@ mm-search-bar {
     }
 
     &.disabled .search-bar-left-icon .fa {
+      color: @inactive-color;
+    }
+  }
+}
+
+mm-filter-slider-icon {
+  padding: 8px 0;
+  position: relative;
+  display: block;
+
+  .mm-filter-slider-container {
+    display: flex;
+    justify-content: flex-start;
+    border-radius: 4px;
+    background-color: rgba(255, 255, 255, 0.8);
+
+    *, *:active, *:focus {
+      outline: none !important;
+      box-shadow: none;
+    }
+
+    .open-filter {
+      background-color: transparent;
+      margin: 0 3px;
+      font-size: @font-large;
+      position: relative;
+      border-radius: 0 px;
+
+      &:active {
+        box-shadow: none;
+      }
+
+      .filter-counter {
+        color: @chip-text-color;
+        font-size: @font-extra-extra-small;
+        background: @chip-background-color;
+        border-radius: 50%;
+        padding: 1px 5px;
+        vertical-align: middle;
+      }
+    }
+
+    &.disabled {
       color: @inactive-color;
     }
   }
--- a/webapp/src/ts/components/components.module.ts
+++ b/webapp/src/ts/components/components.module.ts
@@ -32,6 +32,7 @@ import { ReportImageComponent } from '@mm-components/report-image/report-image.c
 import { NavigationComponent } from '@mm-components/navigation/navigation.component';
 import { EnketoComponent } from '@mm-components/enketo/enketo.component';
 import { SearchBarComponent } from '@mm-components/search-bar/search-bar.component';
+import { FilterSliderIconComponent } from '@mm-components/filter-slider-icon/filter-slider-icon.component';
 import { MultiselectBarComponent } from '@mm-components/multiselect-bar/multiselect-bar.component';
 import {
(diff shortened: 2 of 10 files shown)


andrablaj:
Hi @ShaunKrog! Are you still interested in working on this issue? It's been stale for a few months now. Please let us know if you need further support to continue or close the PR if you don't plan to continue working on it. Thank you!

ChinHairSaintClair:
Hi @andrablaj ,
I plan on pushing some work into this PR sometime this week. It might not be the most refined code, but it should give the CHT team an idea of our approach 🙂

andrablaj:
Excellent, @ChinHairSaintClair! Can you please comment on https://github.com/medic/cht-core/issues/9558 so I can assign it to you?

ChinHairSaintClair:
@andrablaj Here's what we're working on:
https://github.com/ChinHairSaintClair/cht-core/tree/task-filter

It’s still a work in progress, but:
- The report filter has been repurposed into a reusable component for both the Tasks and Reports tabs.
- We added the ability to hide specific filters, like the "status" filter on the Tasks tab.
- Task names can now be fuzzy searched, piggybacking on the existing deduplication functionality.

The work will be pushed to the PR soon.

The original report filter tests should still apply - we’d probably just need to tweak them slightly.
We also noticed that `rulesEngineCore.fetchTasksFor` accepts a list of IDs, so it might be worth exploring whether we can get a reduced result set from that API.

What we’d like to clarify:
1. When filtering by date, should the `start/end` date be considered, or just the `due date`?
2. Since task querying works fundamentally differently from its report counterpart, we're using fuzzy matching. Would a default matching threshold of 3 differences be acceptable?

Please let me know if I've missed anything @ShaunKrog

andrablaj:
Thank you for the comprehensive explanation, @ChinHairSaintClair!

@dianabarsan Do you have any thoughts or suggestions related to @ChinHairSaintClair's [design above](https://github.com/medic/cht-core/pull/9576#issuecomment-2880899307)?

dianabarsan:
> We also noticed that rulesEngineCore.fetchTasksFor accepts a list of IDs, so it might be worth exploring whether we can get a reduced result set from that API.

I don't know if the additional roundabout operation would be necessary for filtering, since tasks only get regenerated AUTOMATICALLY when some change is made. So I would have the filter not reload the tasks list, just filter the one that is already loaded through the existent mechanisms. Of course, the filters should be applied when the task list gets refreshed due to changes. 

> When filtering by date, should the start/end date be considered, or just the due date?

I believe start/end are completely obfuscated for the user and would be completely confusing to filter by them. I would say the filters should be about fields that the user can see. 

> Since task querying works fundamentally differently from its report counterpart, we're using fuzzy matching. Would a default matching threshold of 3 differences be acceptable?

As above, I think the filter should work over the list that gets returned by the RulesEngine instead of making additional queries. I think indexing tasks by freetext is something completely novel and it could be designed in 20 different ways. I would probably skip the search entirely until we know _what_ the community wants to search for. 

Great work for reusing the reports filters!

ChinHairSaintClair:
Thank you for the review @dianabarsan, we're already chipping away at the changes. One point in particular stood out: that we don’t really know what the community wants to search for. That’s absolutely true, our implementation so far has been loosely guided by historical feedback from site managers. I've posted a follow-up on the [task filter](https://forum.communityhealthtoolkit.org/t/filter-option-for-task/4808/7) thread to help get broader alignment from the community.

ChinHairSaintClair:
Hi @andrablaj do have any further thoughts on the direction of the task filters?
I’ve opened a CHT forum thread, but unfortunately, it has also become relatively stale:
https://forum.communityhealthtoolkit.org/t/filter-option-for-task/4808

andrablaj:
Thank you for the nudge, @ChinHairSaintClair! I shared your question in the current Task Prioritization squad Slack channel to get more input from the programs involved in that project. It would be wonderful if you could join one of that squad's weekly calls to share your thoughts and questions, if you are available. 
cc @antonykhaemba

ChinHairSaintClair:
Hi @andrablaj, apologies for the delayed response. We've been tied up with some deployment blockers and syncing issues that have taken most of our time. As we wrap up those fixes, we’ll get back to this PR - it remains a priority for us.

@ShaunKrog, please take note of the "weekly squad call" mentioned above by Andra 🙂

ShaunKrog:
Hi @andrablaj, just to follow up on @ChinHairSaintClair's message, we are still tied up with our upgrade to CHT v4.15.x and alongside this have been experiencing several issues with our deployment pipeline. 

I would just like to note that we plan on attending to this PR shortly, particularly as we wrap our existing upgrade and deployment issues.

I also note the weekly squad call and will arrange to attend.

andrablaj:
Un-stale 😄

ShaunKrog:
Un-stale, planning to devote time to this in December

PR #9581, opened 2024-10-24
[DO NOT MERGE]

### LINKED ISSUE #9547: Change password on first login
System admin users create accounts for CHWs and then share the password with them. To enhance the security of these accounts, there should be a way to prompt a password change on the first login. 

On the login page, create a feature to change the password with the following considerations: 
- All UI texts should be translatable in CHT's supported languages
- This feature is enabled by default but it can be disabled for user roles based on the `can_skip_password_change` permission
- Only new users will be prompted to change their password on the first login when the permission is enabled.
- Subsequent logins won't require a password change. 
- Users will be prompted to change their password on their next login when the admin manually updates the user password 
- Changing the password in this step is an online action (they need internet)
- If the CHW can't change the password for some reason, they can't access the app
- If CHW logins for first login and closes the app before changing password, the next time they login (even though it is not technically first login) they are prompted to change password.

[Figma design](https://www.figma.com/design/3S7e2khzFNkdT7PLlf153H/Cybersecurity?node-id=202-92&node-type=canvas&t=J1kyJfBTdPxJmCUP-0)

The scope of this work is to enable changing the password on the first login. It doesn't include changing the password rules or current authentication mechanisms.

Benmuiruri:
# Description

Closed this PR after discussions and clarification of the feature in favor of a cleaner PR #9731 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9547-change-password/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9547-change-password/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9547-change-password/docker-compose/cht-couchdb-clustered.yml)

# License


Benmuiruri:
Hi @garethbowen I worked on the first part to understand the API side routing and display the password page. 
Now I'm updating the form to meet the Figma design and checking out the update password logic to use it in the password change form. As I do that, I'd like your input of my approach / understanding of this feature. 

- The approach I would like to take for this is to edit the `add user` form in the admin to add a checkbox option, which is unchecked by default (hence `false / null`). When checked it sets` require_password_change to true` in the user object.
-  In the login controller I have `password_change_required = true;`. 
- The login controller will check the permission (does the user's role have the permission `can_change_password_first_login` && is `require_password_change = true` for the user && is  `password_change_required = true;` ) - If all those are true, then it is the user's first time login and require password change. 
- Once user changes password the password update function will be in the login controller so that it can update `password_change_required = false;` and possibly update user settings to update `require_password_change flag to false` That way subsequent logins will not require password change. 

My Thoughts 
- I believe this would work with CHT User Management tool to add the checkbox option when creating single or multiple users. 
- I believe this would work if a user loses their phone because the admin can select the `require_password_change` checkbox in the admin. 
- I believe this would not be a breaking change for existing users because `require_password_change` would be null for them. 
- When making it mandatory for projects in future, we would make the `require_password_change` field to be mandatory true. 
- A forgotten password feature would also reuse the same `require_password_change` in the user object.

garethbowen:
> The approach I would like to take for this is to edit the add user form in the admin to add a checkbox option, which is unchecked by default (hence false / null). When checked it sets require_password_change to true in the user object.

Unfortunately this does not solve the requirement. The requirement is that every time a password is set by anyone other than the user themselves (ie: admin), the user is required to set a new password on login. This should be enforced in the API somewhere to ensure this happens regardless of whether the password is set via the admin app, the user management app, or any other method. Therefore there must be no checkbox in the admin app to turn this off. You'll have to check how the [user management app](https://github.com/medic/cht-user-management/) is written to make sure it applies there as well.

Because this may be disruptive we will include a feature flag to turn it off, just like we do for the new UX features like the action bar -> floating action button. In some time if/when this has been proven to work we will remove the feature flag and make this the only option for all user password setting.

All of this must default to true, otherwise we cannot claim the CHT is secure by default.

> I believe this would work with CHT User Management tool to add the checkbox option when creating single or multiple users.

It's also possible to set up the user with scripts and so on, so I prefer the API approach so we can catch all cases where the password is set.

> I believe this would work if a user loses their phone because the admin can select the require_password_change checkbox in the admin.

If the phone is lost then the administrator MUST change the password, not set the checkbox. You have to change the password in order to invalidate the session so anyone finding the phone is kicked out of the app. If they're changing the password then the process above (mandatory password reset) will mean you don't need the checkbox at all.

> I believe this would not be a breaking change for existing users because require_password_change would be null for them.

Thanks for thinking about this! That's an absolute requirement.

api/src/controllers/login.js
```diff
@@ -300,15 +314,27 @@ const renderLogin = (req) => {
...
     if (sessionRes.statusCode !== 200) {
       res.status(sessionRes.statusCode).json({ error: 'Not logged in' });
-    } else {
-      const redirectUrl = await setCookies(req, res, sessionRes);
-      res.status(302).send(redirectUrl);
     }
+    const sessionCookie = getSessionCookie(sessionRes);
+    const options = { headers: { Cookie: sessionCookie } };
+    const userCtx = await getUserCtxRetry(options);
+    await setCookies(req, res, sessionRes);
```
garethbowen, line 330: I think this is a bug. If you create the session and set the cookie _before_ they change their password then it's trivial to bypass the reset. I haven't tried this but if you go through this flow, and when you get to the "password reset" form, you just refresh the page, I think you'll be redirected right into the app!

Benmuiruri: Yes that's true, this was a quick work around to just see the redirect happen and see the form. So that I can now design the form 😄

api/src/controllers/login.js
```diff
@@ -51,6 +51,20 @@ const templates = {
...
     ],
   },
+  passwordReset: {
+    file: path.join(__dirname, '..', 'templates', 'login', 'password-reset.html'),
+    translationStrings: [
+      'login',
+      'Password',
+      'privacy.policy'
+    ],
+  }
```
garethbowen, line 61: I'm debating whether it's better/easier to do this as a separate page, or include it in the initial login form with show/hide based on JS.

I added a comment below that you should not create the session until the password has been changed which adds a bit of complexity because the password reset submission will need to know a) the user, b) the old password, and c) the new password in order to do the password reset and log them in in one go.

Because you need all three pieces of info, and the user has just entered the first two, then that will be lost when you redirect to the new page. However if you just show another form then you can easily access the user and old pass from the hidden form for the second submission.

If you do it as a separate page then you'll have to pass the old password back in clear text which is insecure.

Does that make sense?

Benmuiruri:
Hi @latin-panda this is ready for an early review even though it is not 100% done. A couple of things: 
- I tried a couple of things to display the param for the translation, but I could not get the `minimum` to display, any idea how to pass it from the [js](api/src/public/login/password-reset.js) file to the [html](api/src/templates/login/password-reset.html) file ?
- I split the translation for login and password reset to `api/src/public/login/auth-utils.js` to avoid duplication. Do I change the eslint to allow compile app to run without the `Parsing error: 'import' and 'export' may appear only with 'sourceType: module'` error ?
- I'll work on the styling of the password reset validation user feedback to match Figma on Monday. 

With those three points, I think the solution design is ready for an early review.

api/src/templates/login/password-reset.html
latin-panda, line 1: HTML indentation is 2 spaces

api/src/templates/login/password-reset.html
```diff
@@ -0,0 +1,50 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="theme-color" content="#323232">
+    <title>{{ branding.name }}</title>
+    <link href="/login/style.css" media="all" rel="stylesheet">
+    <link rel="shortcut icon" href="/favicon.ico">
+</head>
+<body data-default-locale=" {{ defaultLocale }}" data-translations="{{ translations }}">
+<div class="center">
+    <form id="form" action="/medic/password-reset" method="POST">
+        <img class="logo" src="{{ branding.logo }}" alt="logo" title="{{ branding.name }}">
```
latin-panda, line 13: One test is to run it in a different config with branding that is not CHT, make sure the logo displayed nicely

api/src/public/login/script.js
latin-panda, line 1: Nice extracting these for reusability

api/src/public/login/auth-utils.js
latin-panda, line 1: 1. JS files uses 2 spaces indentation 
2. Can it use arrow function?

api/src/public/login/auth-utils.js
```diff
@@ -0,0 +1,90 @@
...
+};
+
+const replaceTranslationPlaceholders = (text, translateValues) => {
+    if (!text || !translateValues) {
+        return text;
+    }
+
+    try {
+        const values = JSON.parse(translateValues);
+        console.log(values);
```
latin-panda, line 62: remove console log

api/src/public/login/auth-utils.js
```diff
@@ -0,0 +1,90 @@
...
+        return Object.entries(values).reduce((result, [key, value]) =>
```
latin-panda, line 63:
```suggestion
        return Object
          .entries(values)
          .reduce((result, [key, value]) =>
```

api/src/public/login/password-reset.js
```diff
@@ -0,0 +1,84 @@
...
+let translations;
+
+const PASSWORD_INPUT_ID = 'password';
+const CONFIRM_PASSWORD_INPUT_ID = 'confirm-password';
+
+const translate = () => {
+    baseTranslate(selectedLocale, translations);
+};
+
+const togglePassword = () => {
```
latin-panda, line 19: I think we have the same in login, can this be in the utils?

api/src/public/login/password-reset.js
```diff
@@ -0,0 +1,84 @@
...
+    document.getElementById('confirm-password-container').classList.toggle('hidden-password');
+};
+
+const displayPasswordValidationError = (serverResponse) => {
+    const { error, params } = JSON.parse(serverResponse);
+    setState(error);
+
+    if (params?.minimum) {
+        const passwordError = document.querySelector('.error.password-short');
+        if (passwordError) {
```
latin-panda, line 36:
```suggestion
    const passwordError = document.querySelector('.error.password-short');
    if (params?.minimum && passwordError) {
```

latin-panda (changes requested):
Nice job Ben! 🤩

Please test that password values don't appear in logs for weird error cases

api/src/controllers/login.js
```diff
@@ -51,6 +55,30 @@ const templates = {
       'privacy.policy'
     ],
   },
+  passwordReset: {
+    file: path.join(__dirname, '..', 'templates', 'login', 'password-reset.html'),
+    translationStrings: [
+      'login.show_password',
+      'login.hide_password',
+      'change.password.title',
+      'change.password',
```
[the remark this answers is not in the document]
Benmuiruri: - This one doesn't have anything after 'change.password' like the others

I'm sorry, I don't understand this comment.

api/src/controllers/login.js
```diff
@@ -51,6 +55,30 @@ const templates = {
...
+      'change.password.hint',
+      'change.password.submit',
+      'change.password.new.password',
+      'change.password.confirm.password',
+      'change.password.required',
+      'password.weak',
+      'password.length.minimum',
+      'Passwords must match'
```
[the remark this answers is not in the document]
Benmuiruri: We already have an [existing translation key](https://github.com/medic/cht-core/blob/0d70a9880a712a087fc284fdbdac9f062041bfdb/api/resources/translations/messages-en.properties#L184), which is the one I am using there. But I can change it and update places it is being used

api/src/controllers/login.js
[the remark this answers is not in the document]
latin-panda: In other places the key has 3 sections, for example: `change.password.title` (1. change 2. password 3.title).
In L64 only has 2 sections `change.password` (1. change 2. password)  Is it missing something? Like `message` so it's `change.password.message`

api/src/controllers/login.js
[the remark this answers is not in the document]
latin-panda: At some point CHT should move to keys for translations, we can add a new key with the translation and leave the old one for backward compatibility, make a small note in the release notes asking people to use the new key 🤔
What do you think? Another idea to slowly deprecate the old way?

api/src/controllers/login.js
```diff
@@ -355,6 +433,61 @@ module.exports = {
...
+    const limited = await rateLimitService.isLimited(req);
+    if (limited) {
+      return serverUtils.rateLimited(req, res);
+    }
+
+    try {
+      const validation = validatePassword(req.body.password, req.body.confirmPassword);
+      if (!validation.isValid) {
+        return res.status(400).json({
+          error: `password${validation.error}`,
```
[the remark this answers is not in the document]
Benmuiruri: Yes, the validatePassword tells us which validation has failed whether it is `password-weak`, `password-mismatch` etc.

api/src/public/login/style.css
```diff
@@ -101,7 +101,11 @@ form {
 .tokenmissing .error.missing,
 .tokentimeout .error.timeout,
 .tokenexpired .error.expired,
-.tokenerror .error.unknown
+.tokenerror .error.unknown,
+.passwordweak .error.password-weak,
+.passwordshort .error.password-short,
+.passwordmismatch .error.password-mismatch,
+.passwordrequired .error.password-required
```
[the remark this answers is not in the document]
Benmuiruri: I decided to write it like that to keep the style consistent. In the file we have existing `tokenexpired`, `tokenerror`, `loginoffline` etc. If I change just for the password the styles will be inconsistent.

api/src/controllers/login.js
```diff
@@ -355,6 +433,61 @@ module.exports = {
...
+
+      await db.users.put(user);
+
+      cookie.clearCookie(res, 'login');
+      cookie.clearCookie(res, 'AuthSession');
+
+      req.body = {
+        ...req.body,
+        user: user.name,
+        password: req.body.password,
```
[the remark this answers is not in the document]
Benmuiruri: This was to create the new session with the new password, but maybe I can call `validateSession`. I'll look into that.

api/src/public/login/style.css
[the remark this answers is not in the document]
latin-panda: Yeah, but it's not following the CSS standard, and it's harder to read

api/src/controllers/login.js
```diff
@@ -355,6 +433,61 @@ module.exports = {
...
+      }
+      const userCtx = await auth.getUserCtx(req);
+      const user = await db.users.get(`org.couchdb.user:${userCtx.name}`);
+      user.password = req.body.password;
+      user.password_change_required = false;
+
+      await db.users.put(user);
+
+      cookie.clearCookie(res, 'login');
+      cookie.clearCookie(res, 'AuthSession');
```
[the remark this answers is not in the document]
latin-panda: login service or auth service. Which one do you think fit better?

api/src/controllers/login.js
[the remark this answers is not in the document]
latin-panda: Maybe like this: [ 'password', validation.error].join('-') 
Or: `password-${validation.error}` <-- adding the dash here instead, so it's clearer it's a key and not a text or prop name

api/src/controllers/login.js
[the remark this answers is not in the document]
Benmuiruri: I forgot to clean it up. It's not being used.

api/src/controllers/login.js
```diff
@@ -355,6 +431,60 @@ module.exports = {
...
```
garethbowen, line 469: This line implies the AuthSession is already created at this point based on the old user/pass. When we last talked about it I think we agreed you cannot create a session using the password that requires reset, because it would be trivial to use that session and bypass the password reset.

api/src/controllers/login.js
```diff
@@ -300,26 +333,70 @@ const renderLogin = (req) => {
...
+  return sessionRes;
+};
+
+const sendLoginErrorResponse = (e, res) => {
+  if (e.status === 401) {
+    return res.status(401).json({ error: e.error });
+  }
+  logger.error('Error logging in: %o', e);
+  return res.status(500).json({ error: 'Unexpected error logging in' });
+};
```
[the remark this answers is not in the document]
Benmuiruri: I don't think so. From what I could find like here in this feature [request](https://github.com/expressjs/express/issues/3942) is [https-status-codes](https://www.npmjs.com/package/http-status-codes) being used. 

We could `http.STATUS_CODES` provided by Node by creating our own constants and doing a lookup. 

 `http.STATUS_CODES` gives us ⬇️ 

```
{
  '100': 'Continue',
  '101': 'Switching Protocols',
  '102': 'Processing',
  '103': 'Early Hints',
  '200': 'OK',
 }

File: api/src/controllers/login.js (hunk 14 above)
Comment (Benmuiruri) [reply api/src/controllers/login.js]: I get it now, the current setCookie creates the AuthSession so I need to should refactor it to only set AuthCookie for users who do not need password reset.

**Comment by Benmuiruri** [comment]
Hi @garethbowen and @latin-panda I believe this is ready for a review. 

The password reset logic , unit tests and integration tests are ready for review. 

Unfortunately the e2e are a bit flaky with the password reset. Sometimes a test hangs on the password reset page causing the test to fail with  I want to reach out to QA to get possibly a more robust way of doing the password reset. 

While I work on the e2e I think the review can go ahead.

File: api/src/controllers/login.js [hunk 16]
```diff
@@ -17,6 +17,15 @@ const translations = require('../translations');
 const template = require('../services/template');
 const rateLimitService = require('../services/rate-limit');
 const serverUtils = require('../server-utils');
+const { validatePassword } = require('@medic/user-management/src/users');
```
garethbowen, line 20: Rather than linking to the internals of the user-management shared-lib treat it like all you can see is the main index.js file. If you just require `@medic/user-management` and update the index.js to export the `validatePassword` function then it's a cleaner interface.

api/src/controllers/login.js
```diff
@@ -190,36 +221,46 @@ const setUserCtxCookie = (res, userCtx) => {
...
-      throw { status: 401, error: 'Error getting authCtx' };
-    });
+  try {
+    const userCtx = await getUserCtxRetry(options);
+    if (roles.isDbAdmin(userCtx)) {
+      await users.createAdmin(userCtx);
+    }
+
+    const user = await users.getUserDoc(userCtx.name);
+    if (user?.password_change_required && !user?.token_login?.active && !await skipPasswordChange(userCtx)) {
```
garethbowen, line 237: This `if` condition is quite complicated (even more so because it has an await!). Let's change it so all the conditions are part of `skipPasswordChange`, eg:

```suggestion
    if (!await skipPasswordChange(user, userCtx)) {
```

api/src/controllers/login.js
```diff
@@ -190,36 +221,46 @@ const setUserCtxCookie = (res, userCtx) => {
...
+
+const redirectToPasswordReset = (req, res, userCtx) => {
+  setUserCtxCookie(res, userCtx);
+  setUserLocale(req, res);
+  return PASSWORD_RESET_URL;
+};
+
+const setUserLocale = (req, res) => {
+  const selectedLocale = req.body.locale || config.get('locale');
+  cookie.setLocale(res, selectedLocale);
```
garethbowen, line 263: This is so much easier to read. Good call!

api/src/controllers/login.js
```diff
@@ -300,26 +341,85 @@ const renderLogin = (req) => {
...
 
+const updatePassword = async (user, newPassword) => {
+  const updatedUser = {
+    ...user,
+    password: newPassword,
+    password_change_required: false
+  };
+  await db.users.put(updatedUser);
+  // creating new session immediately after changing a password might 401
+  await new Promise(resolve => setTimeout(resolve, 50));
```
garethbowen, line 399: This isn't a great solution because while it might work on your machine, what happens if couchdb is over a laggy network, or under load? A better solution would be to retry creating the session with 50ms delay until it works or some limit is reached. You'll have to be careful of rate limiting here...

api/src/public/login/auth-utils.js
```diff
@@ -0,0 +1,121 @@
+/* eslint-disable */
```
garethbowen, line 1: Remove this.

api/src/public/login/script.js
```diff
@@ -1,40 +1,35 @@
...
+  setState,
+  request,
+  getCookie,
+  getLocale,
+  parseTranslations,
+  baseTranslate,
+  togglePassword,
+  getUserCtx
+} from './auth-utils.js';
+
```
garethbowen, line 12: I'm nervous about all the modern code you've added here because we don't have any transpilation to ensure backwards compatibility. Have you checked that all of these features are supported?

api/src/public/login/script.js
```diff
@@ -1,40 +1,35 @@
+/* eslint-disable */
```
garethbowen, line 1: Remove.

api/src/public/login/password-reset.js
```diff
@@ -0,0 +1,77 @@
+/* eslint-disable */
```
garethbowen, line 1: Remove.

api/src/controllers/login.js
```diff
@@ -355,6 +456,49 @@ module.exports = {
...
+    try {
+      const validation = validatePasswordReset(req.body.password, req.body.confirmPassword);
+      if (!validation.isValid) {
+        return res.status(400).json({
+          error: validation.error,
+          params: validation.params,
+        });
+      }
+      const user = await db.users.get(`org.couchdb.user:${req.body.user}`);
+      await updatePassword(user, req.body.password);
```
garethbowen, line 487: I think there's a big security hole here! What if I navigate to the reset password page, set my userCtx cookie to `{ name: ben }`, submit a new password. I haven't tried but I _think_ that will update YOUR password to whatever I want.

I think the submission of the password reset form MUST contain the existing password as well as the new password otherwise how can you verify the user is who they claim to be?

api/src/public/login/password-reset.js
```diff
@@ -0,0 +1,77 @@
...
+  setState('loading');
+  const url = document.getElementById('form')?.action;
+  const password = document.getElementById(PASSWORD_INPUT_ID)?.value;
+  const confirmPassword = document.getElementById(CONFIRM_PASSWORD_INPUT_ID)?.value;
+  const userCtx = getUserCtx();
+
+  const payload = JSON.stringify({
+    user: userCtx.name,
+    password: password,
+    confirmPassword: confirmPassword,
```
garethbowen, line 48: There's no point submitting both the `password` and the `confirmPassword`. Just do a quick `===` check on the client side before submitting and just send through one copy. This is better UX because it can fail without the round trip to the server.

Ideally we'd also check password strength on the client side but if it's too hard to share the password strength library on the client side then it's fine to leave this for another day.

garethbowen (changes requested):
Really extensive work! I haven't looked at the flaky test problem but left some comments on the other bits.

tests/e2e/default/login/login-logout.wdio-spec.js
```diff
@@ -162,7 +169,78 @@ describe('Login page funcionality tests', () => {
...
+      await browser.url('/');
+      await loginPage.setPasswordValue(user.password);
+      await loginPage.setUsernameValue(user.username);
+      await (await loginPage.loginButton()).click();
+      await browser.url('/medic/password-reset');
+      await loginPage.getPasswordResetTranslations();
+      await loginPage.setPasswordValue('');
+      await loginPage.setConfirmPasswordValue(user.password);
+      await (await loginPage.updatePasswordButton()).click();
+      expect(await loginPage.getPasswordResetErrorMessage('password-short')).to.equal(PASSWORD_MISSING);
```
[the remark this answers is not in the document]
Benmuiruri: Not really, the first check in user management is for password length before it compares with confirmPassword

garethbowen:
@Benmuiruri Another thing to think about is what happens if the user connects with basic auth, either through curl or directly in the browser. Can they bypass the password reset this way?

api/src/public/login/script.js
Benmuiruri: I've checked for Chrome v.90 and Firefox v.74

Arrow functions [yes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#browser_compatibility)
Optional chaining [yes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#browser_compatibility)
Export [yes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export#browser_compatibility)
Import [yes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#browser_compatibility)

But the one for imports shows a higher support version for `Available in service workers` which is v.91 for Chrome.

Benmuiruri:
Hi @garethbowen Thanks for your review. 
- I refactored the auth-utils.js file to drop the use of modules 
- I changed the updatePassword in the login controller to safe guard against timeouts over a network
- I added the currentPassword field in the password reset form 
- I added a check for password reset in the authorization middleware 


<details>
<summary>This video shows the progress so far </summary>

https://github.com/user-attachments/assets/16a4fbcf-d003-421e-b13a-388db8f9d072


</details>

I'm currently updating failing e2e tests (almost done). As I continues, could I get your review of my implementation of your requested changes. Thanks

api/src/controllers/login.js
```diff
@@ -18,6 +18,15 @@ const template = require('../services/template');
 const rateLimitService = require('../services/rate-limit');
 const serverUtils = require('../server-utils');
 
+const PASSWORD_RESET_URL = '/medic/password-reset';
+
+const ERROR_KEY_MAPPING = {
+  'password.weak': 'password-weak', //NoSONAR
+  'password.length.minimum': 'password-short', //NoSONAR
+  'password.current.incorrect': 'current-password-incorrect', //NoSONAR
+  'password.same': 'password-same' //NoSonar
```
latin-panda, line 27: What is the sonar warning?

api/src/auth.js
```diff
@@ -76,6 +76,32 @@ module.exports = {
...
+      throw error;
+    }
+
+    if (req.url.includes('/password-reset')) {
+      return true;
+    }
+
+    if (environment.isTesting) {
+      return true;
+    }
```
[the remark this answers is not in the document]
Benmuiruri: Hi @garethbowen , I cleaned this up, but spent the better part of Friday figuring out how to do the password reset in integration tests. I was tripping over session management with password reset but got there in the end. 

I'm doing final checks to get the PR ready for review.

Benmuiruri:
Hey @jkuester 👋 

With @garethbowen out of office this week, I was wondering if you could review this PR, we work together to get it merged.

I addressed the feedback, but I'd appreciate a fresh pair of eyes given it was my first time in the API side of things. 

The PR description has a video of the state of the feature. Happy to work with you on this.

jkuester:
:+1:   Sounds good @Benmuiruri.  I can have a look at things and let you know if it would be helpful to sync on the details.  

Unfortunately, I am pretty stacked up this week with PRs.  I should be able to get to this by EOW, but if you need it sooner, feel free to replace me with a different reviewer.

jkuester:
FYI, got dug in on this and started reviewing it today, but was not able to make it all the way through. I will be OOO on Fri/Mon, but will pick this back up next Tues.  :+1:

Benmuiruri:
Thanks for the heads up @jkuester

shared-libs/user-management/src/index.js
```diff
@@ -17,7 +17,8 @@ module.exports = (sourceConfig, sourceDb, sourceDataContext) => {
     bulkUploadLog,
     roles,
     tokenLogin,
-    users
+    users,
+    validatePassword: users.validatePassword
```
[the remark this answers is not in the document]
jkuester: Ahh, I see [Gareth's comment](https://github.com/medic/cht-core/pull/9581#discussion_r1840355510)...  :thinking: 

Still not sure I understand how `validatePassword` is any different from the rest of the functions on `users`, but no need to change anything here....

jkuester (changes requested):
Alright, going to go ahead and post what I have reviewed so far.  You have done a TON of great work here! (Hard to believe how much code it took to implement this feature... :sweat_smile: )  

I have tried to focus just on important issues/questions and/or easy code suggestions. But, knowing the time constraints we have here, I will definitely accept "This works fine as it is" as a response to any of my refactoring comments!

api/src/controllers/login.js
```diff
@@ -51,6 +62,31 @@ const templates = {
...
+      'password.same'
+    ],
+  }
+};
+
+const skipPasswordChange = async (user, userCtx) => {
+  return !user?.password_change_required ||
+    user?.token_login?.active ||
+    roles.isDbAdmin(userCtx) ||
+    await auth.hasAllPermissions(userCtx, 'can_skip_password_change');
```
jkuester, line 89: thought: referencing back to my [previous comment](https://github.com/medic/cht-core/pull/9581#discussion_r1880794413) about how this feature feels like a bit of a breaking change to have enabled by default, I initially expected this behavior to be something _toggled ON by a permission_.  E.g. the system would trigger the reset password flow only for users with the `can_reset_password_assigned_by_admin` permission, or something like that...

jkuester (changes requested):
Okay, made it through the rest of the implementation code and unit tests.

For the e2e/integration tests, I don't like how all the random tests that need a non-admin user have to deal with the password change complexity. Ideally we would have dedicated tests for the password reset workflow and then none of the rest of the tests woud have to worry about it.  

Now that is, of course, easier said than done.  I guess the most naive way to avoid the password reset would be to add the roles to `can_skip_password_change` in `config/default/app_settings.json`, but that is not super elegant.

The main idea I have come up with so far involves updaing the `createUsers` function in `utils/index.js` to do either of these two things:

- Manually update the user doc after creating the user to remove the `password_change_required` field. This is nice and simple. The downside is that it is a bit artificial, but I am not sure it rises to the level of affecting the integrety of the tests.
- The other alternative would be to have the `createUsers` function write the initial user with with a _different password_ than the one provided in the parameter. Then, after the user has been created the `createUsers` function could do a login and run the reset-password flow to set the password to the _original password value_ provided for the user.  The end result after calling `createUsers` would be that the user exists and has the specified password and a reset will not be triggered the next time the user logs in.

Ideally the deafult behavior of `createUsers` would result in the user existing in the db with the provided password and no password reset would happen the next time the user logs in.  For the actual password-reset tests, we could either parameterize the `createUsers` function to be able to flag off that behavior (so the test can do the reset) or we could just not use this function for creating users for those tests...

shared-libs/user-management/src/users.js
```diff
@@ -452,14 +459,31 @@ const getSettingsUpdates = (username, data) => {
...
+  const admins = await couchSettings.getCouchConfig('admins');
+  return admins && !!admins[username];
+};
+
+const requirePasswordChange = async (username, fullAccess) => {
+  if (!fullAccess) {
+    return false;
+  }
+
+  return !(await isTargetAdminUser(username));
```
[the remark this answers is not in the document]
Benmuiruri: You're correct, previous discussions with @garethbowen was to require password reset every time an admin changes a user's password for security purposes like you mention.

api/src/controllers/login.js
```diff
@@ -300,26 +346,139 @@ const renderLogin = (req) => {
...
+
+const passwordResetValidation = async (username, currentPassword, password) => {
+  const validation = validatePasswordReset(password);
+  if (!validation.isValid) {
+    return {
+      isValid: false,
+      status: 400,
+      error: validation.error,
+      params: validation.params
+    };
```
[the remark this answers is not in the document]
Benmuiruri: Nice. Thanks

api/src/controllers/login.js
```diff
@@ -355,6 +515,47 @@ module.exports = {
...
+      const validationResult = await passwordResetValidation(username, currentPassword, password);
+      if (!validationResult.isValid) {
+        return res.status(validationResult.status).json({
+          error: validationResult.error,
+          params: validationResult.params
+        });
+      }
+
+      const userDoc = await db.users.get(`org.couchdb.user:${username}`);
+      await updatePassword(userDoc, password);
```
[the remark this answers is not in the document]
Benmuiruri: It's because of a race condition, when testing it out, in some instances  requesting session immediately after changing a password was giving back a 401. Hence I safe guarded it with the retry to avoid that.

api/src/controllers/login.js
```diff
@@ -51,6 +62,31 @@ const templates = {
...
+      'password.current.incorrect',
+      'password.same'
+    ],
+  }
+};
+
+const skipPasswordChange = async (user, userCtx) => {
+  return !user?.password_change_required ||
+    user?.token_login?.active ||
+    roles.isDbAdmin(userCtx) ||
```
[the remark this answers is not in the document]
Benmuiruri: Yes, you're correct. It _might_  be extra validation since those checks are already in `user.js`. Maybe lets keep it because after the after the POST in login which needs to decide where to go next, these checks make it clear in this controller.

api/src/controllers/login.js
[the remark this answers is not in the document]
Benmuiruri: Thanks for catching this. I now see the disconnect between when we set the requirement and when we check if it can be skipped. Let me move the logic to users.js then in login controller just check value of `password_change_required`.

Benmuiruri: Great observation, I initially thought that as well. Even started working on it from that approach. However, Gareth raised a fair point. Our purpose for this feature is to improve security to safe guard against a password leak. 

If we leave it as an opt in feature, because people are generally "lazy" when it comes to security steps, a project will not turn it on. Therefore, in the event of a password leak, the feature won't really help them. 

But having it as an opt out then that way the system is designed to safe guard, but you can opt out knowingly at your own risk.

api/src/middleware/authorization.js
```diff
@@ -40,7 +40,9 @@ module.exports = {
       return serverUtils.error('Authentication error', req, res);
     }
 
-    next();
+    return auth.checkPasswordChange(req)
```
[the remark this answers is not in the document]
Benmuiruri: The reason for adding this was to block users who require password change from bypassing the check via basic auth or curl. 
- The first result is with the guard
- The second result is without it

Screenshot 2024-12-13 at 17 11 20

shared-libs/user-management/src/users.js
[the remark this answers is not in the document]
jkuester: Hmm, that does not seem to be the direction of the initial conversation out on the issue thread.  I added [a comment](https://github.com/medic/cht-core/issues/9547#issuecomment-2541646693) to the main issue to give this question a bit more visibility.  I am just a bit extra wary here since this feels a bit uncharacteristic compared to how we usually roll out CHT changes...

api/src/controllers/login.js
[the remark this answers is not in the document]
jkuester: So, yeah, the `createSessionRetry` logic makes sense to me (and I guess already existed for token login).  But here is where we are actually _changing_ the password.  Seems like we should not be getting a `401` from the `db.users.put(updatedUser);` call.  If I understand things correctly, the user session that is actually used to make that update would be the api server's admin user (`COUCH_USER`) anyway since we have not yet created a session for the current user trying to login....

api/src/middleware/authorization.js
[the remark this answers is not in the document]
jkuester: Ahh, I see.  :thinking:   Honestly, I wonder if we actually care about the basic auth case...  :grimacing: They are giving is a _valid_ password, so it is not like this is a security hole.  Maybe it is acceptable to say that the password-reset flow is only triggered by the login-flow (and not when making random requests with basic auth...).  IMHO that would be preferable to adding an extra round-trip to `_users` for each server request....

Going to push this question up to the main issue too!

Benmuiruri:
Closed this PR after discussions and clarification of the feature in favor of a cleaner PR https://github.com/medic/cht-core/pull/9731

Issue #9582, opened 2024-10-24
Update cht-datasource `bind` in webapp so it returns a wrapper function instead of a Promise


jkuester:
**Describe the issue**

Currently the experience of consuming the `CHTDatasourceService` in webapp is annoying.  Both the `get` and `bind` methods on the service are async and so code must await them before going on to make _another async call_ to get the data they actually want.

```ts
const datasource = await this.chtDatasourceService.get();
const person = await datasource.v1.person.getByUuid(id);

// OR

const getPerson = await this.chtDatasourceService.bind(Person.v1.get);
const person = await getPerson(Qualifier.byUuid(id));
```

It is annoying to have to `await` twice and it really limits the re-usability of the "bound" functions since they cannot be initialized in a the constructor of a consuming class.  

**Describe the improvement you'd like**
Thanks to the wonders of TypeScript and the built-in [Parameters](https://www.typescriptlang.org/docs/handbook/utility-types.html#parameterstype) and [ReturnType](https://www.typescriptlang.org/docs/handbook/utility-types.html#returntypetype) types, it is pretty simple to update the `CHTDatasourceService.bind` method to return a function (not a Promise) that wraps the provided function in a type-safe manner. 

```ts
Person.v1.get
// (context: DataContext) => (qualifier: Readonly<{ uuid: string; }>) => Promise<Nullable<Person. v1.Person>>

const getPerson = chtDatasourceService.bind(Person.v1.get);
// (qualifier: Readonly<{   uuid: string }>) => Promise<Nullable<Person.v1.Person>>

const person = await getPerson(Qualifier.byUuid(id));
// Nullable<Person.v1.Person>
``` 

The wrapper function preserves all the typing from the provided function, but basically delays the async dataContext binding until the function is actually called.

garethbowen:
@jkuester Remember to add this to a release milestone.

PR #9583, opened 2024-10-24
chore(#9582): update datasource `bind` to not return promise

jkuester:
# Description

Closes https://github.com/medic/cht-core/issues/9582

See issue for details about the change.

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:datasource_bind/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:datasource_bind/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:datasource_bind/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/webapp/src/ts/services/cht-datasource.service.ts
+++ b/webapp/src/ts/services/cht-datasource.service.ts
@@ -106,12 +106,16 @@ export class CHTDatasourceService {
    */
   bind<R, F extends (arg?: unknown) => Promise<R>>(fn: (ctx: DataContext) => F):
     (...p: Parameters<F>) => ReturnType<F> {
-    return (...p) => new Promise((resolve, reject) => this
-      .isInitialized()
-      .then(() => this.dataContext
-        .bind(fn)(...p)
-        .then(resolve)
-        .catch(reject))) as ReturnType<F>;
+    return (...p) => {
+      return new Promise((resolve, reject) => {
+        this.isInitialized().then(() => {
+          const contextualFn = this.dataContext.bind(fn);
+          contextualFn(...p)
+            .then(resolve)
+            .catch(reject);
+        });
+      }) as ReturnType<F>;
+    };
   }
 
   async get() {


webapp/src/ts/services/cht-datasource.service.ts
```diff
@@ -97,9 +97,21 @@ export class CHTDatasourceService {
...
+   * @returns a "context-aware" version of the function that is bound to the data context and ready to be used
+   */
+  bind<R, F extends (arg?: unknown) => Promise<R>>(fn: (ctx: DataContext) => F):
+    (...p: Parameters<F>) => ReturnType<F> {
+    return (...p) => new Promise((resolve, reject) => this
+      .isInitialized()
+      .then(() => this.dataContext
+        .bind(fn)(...p)
+        .then(resolve)
+        .catch(reject))) as ReturnType<F>;
```
garethbowen, line 114: Personally I find the autoreturn style with multi line body really difficult to follow. Is this just me?

Secondly I think you can remove an indent from lines 113 and 114. My suggestion would be...

```suggestion
    return (...p) => new Promise((resolve, reject) => {
      return this.isInitialized()
        .then(() => this.dataContext.bind(fn)(...p))
        .then(resolve)
        .catch(reject)) as ReturnType<F>;
    });
```

I _think_ I got all the braces right...

garethbowen (commented):
Nice improvement on the API!

One readability suggestion in the implementation.

Benmuiruri (approved):
Is that TS code less Janky 😅  ? 

I need to do more TS exercism. 

But thanks for this, hides the initialization and just gives the function one wants to call.

webapp/src/ts/services/cht-datasource.service.ts
jkuester: :sweat_smile: Generally I don't mind autoreturns for promise chains (and other places where it is pretty clear where things are going).  But, yeah, for this craziness we should make it as readable as possible (even if it means I actually have to type `return` a few more times.... :sob: ).

Going to take this a couple steps further and explicitly return the constructed Promise as well as assigning the contextual function to a variable to avoid the double-parenthesis from the currying. :+1:

jkuester:
@Benmuiruri :laughing:  I should have been more specific!  This makes the _code consuming cht-datasoure_ less janky (by encapsulating all the jankyness into the `bind` function....) :facepalm:

Issue #8806, opened 2024-01-11
Multiple validation functions in rule doesn't work


1yuv:
**Describe the bug**
If you apply Logical operation to multiple [cht validation functions](https://docs.communityhealthtoolkit.org/apps/reference/app-settings/#cht-validation-functions) it doesn't work. 

**To Reproduce**
1. Prepare an app settings with forms for validations. 
2. On config, ensure that two validations are combined together, for example we are trying to check whether L or G form has been submitted previously before f1 form is submitted. Add following validation property in `f1` form. 

```js
validations = [
      {
        property: 'parent_id',
        rule: '"exists('L','patient_id') || exists('G','patient_id')',
        message: [
          {
            content: 'Pregnancy not registered already',
            locale: 'en',
          },
        ],
      }
```
3. Compile app setting, upload to instance. 
4. Ensure patient doesn't have `G` or `L` form submitted already.
5. Submit `f1` form. Since `L` or `G` form doesn't exist, it will work properly and report `Pregnancy not registered already`. This is expected. 
6. Now, submit `G` form for this patient. 
7. After submitting G form, submit `f1` form again. 
8. This will again throw error `pregnancy not registered already`.

**Expected behavior**
1. In step 8 above, it should not throw an error since `G` form is already submitted. 

**Environment**
- Instance: Any
- Browser: Any
- Client platform: Any
- App: Sentinel
- Version: 4.5.0, but this likely exists in all previous supported versions. 

**Additional context**
If this issue is fixed, we won't need to work on #8402 .

dianabarsan:
I'm afraid this is known and not a bug: https://github.com/medic/cht-core/blob/master/shared-libs/validation/src/validation.js#L230-L231

> Validation settings may consist of Pupil.js rules and custom rules.
> These cannot be combined as part of the same rule.

I think the best course of action here is to: 
1. clearly document this limitation
2. consider adding the possibility of combining rules as a feature.

1yuv:
Thanks @dianabarsan , the [documentation](https://docs.communityhealthtoolkit.org/apps/reference/app-settings/#rules) did not mention that custom rules can not be combined, so I expected it to be desired feature. 

>Validation settings may consist of Pupil.js rules and rules specific to the CHT. These two types of rules cannot be combined as part of the same rule.

It'd be good to have multiple custom rules to be combined together which would allow more flexibility to validation.

garethbowen:
[...]

Issue #8402, opened 2023-07-21
Add support to check if either of multiple forms is submitted for a patient. 


1yuv:
There's a validation method in pupil like [exists](https://docs.communityhealthtoolkit.org/apps/reference/app-settings/#cht-validation-functions), which allows us to check multiple conditions, for example if a particular form has been submitted before for the same patient or not. 

However, this function only supports one form and there's no provision to check multiple forms. Our scenario is: 

1. We want to ensure that pregnancy registration form has been submitted for a patient before pregnancy visit or delivery form is submitted. 
2. We've 4 ways of registering pregnancy form i.e. Any form out of those 4 form submitted means a pregnancy form is submitted. 
3. exists() method currently accepts only one form. 
4. Also looks like rules like this is not supported:
```json
"rule": "exists('A', 'patient_id') || exists('L', 'patient_id')",
```

We need a mechanism to check if any of those forms has been submitted. More on this [here](https://github.com/medic/config-moh-nepal/issues/962)

1yuv:
Hi @dianabarsan , @garethbowen , 

The cht-core documentation for [rules](https://docs.communityhealthtoolkit.org/apps/reference/app-settings/#rules) mentions that Pupil.js rules and rules specific to the CHT can not be combined. However, it should allow similar type of rules to be combined together.  For example something like this is a cht specific rule and should be allowed.  
`exists('A', 'patient_id') || exists('L', 'patient_id')`. CHT doesn't complain about the rule, but it's not actually working.

If we had that functionality, we could have tested multiple functions with this.

PR #9602, opened 2024-11-01
fix(#8806): merge extra validations with pupil so rules can have both

garethbowen:
# Description

#8806 
#8402

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8806-combining-validation-rules/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8806-combining-validation-rules/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8806-combining-validation-rules/docker-compose/cht-couchdb-clustered.yml)

# License


shared-libs/transitions/test/unit/pregnancy_registration.js
```diff
@@ -379,7 +379,7 @@ describe('pregnancy registration with weeks since LMP', () => {
     return transition.onMatch({ doc: doc }).then(function(changed) {
       assert.equal(changed, true);
       assert.equal(doc.patient_id, undefined);
-      assert.equal(getMessage(doc), 'Invalid patient name.  Invalid LMP; must be between 0-40 weeks.');
+      assert.equal(getMessage(doc), 'Invalid LMP; must be between 0-40 weeks.  Invalid patient name.');
```
garethbowen, line 382: This is technically broken backwards compatibility because the errors are returned in a different order. This is because instead of the "medic" validations being run separately, they are run at the same time as the pupil validations. I doubt anyone is relying on the order of the validation errors but it's worth pointing it out!

shared-libs/validation/test/validations.js
```diff
@@ -197,7 +193,7 @@ describe('validations', () => {
     return validation.validate(doc, validations).then(errors => {
       assert.deepEqual(errors, [
         {
-          code: 'invalid_xyz_unique',
+          code: 'invalid_xyz',
```
garethbowen, line 196: This is the other potential regression - medic validation errors used to append the validation rule name on to the code. This was to make them work alongside pupil validations. Now that they're merged this is no longer required, and standardising on the code makes sense. However, if anyone relied on the validation code this would be a regression, but I doubt anyone is as it _should_ be an implementation detail and nothing more.

garethbowen:
@1yuv Would you mind reviewing this? There are two regressions commented in the PR but I don't think anyone's going to notice. If you could run some of your configs through this update to confirm that would be ideal!

1yuv:
I tested this with few configurations and I can combine multiple validations easily and is working perfectly.

1yuv (approved):
Thanks for this addition Gareth. This makes SMS workflows more versatile and powerful as well. I've tested this against different configs and everything seems fine

1yuv:
There's an #9604  I've filed that was discovered while testing this, but is not caused by this change.

Issue #9601, opened 2024-11-01
Prevent duplicate sibling contact capture


ChinHairSaintClair:
**Is your feature request related to a problem? Please describe.**
Currently, there appears to be no built-in deterrent against creating records with names similar to existing siblings.

**Describe the solution you'd like**
Prevent duplicate place/person creation and display possible duplicates for consideration. On record submission (through create or edit flow), we want to show the possible duplicate items to our user. They can then navigate to the possible duplicate item via a link, and proceed with record changes there, or circumvent the duplicate check & proceed with record submission.

**Describe alternatives you've considered**
Despite improving our search functionality, and training the CHWs to use it, usage of the search feature before creating new records remains low, leading to frequent duplicates.

**Additional context**
We noticed that our CHWs either forget they've previously captured items or miss previously captured items due to it being slightly mistyped. This has resulted in quite a few duplicate records being created on all user created levels of our hierarchy. We want to fix this at the source before tasks are rolled out to make sure no unnecessary/incorrect tasks fill up our CHWs worklist. This will naturally also improve the accuracy of our data for reporting purposes.

We have a working prototype that we will soon upstream, which can produce the following:
![Image](https://github.com/user-attachments/assets/fe5df62f-f56b-4dd9-89a1-268006a92824)

Please see the following related discussions for more info:
https://forum.communityhealthtoolkit.org/t/mitigate-duplicate-data-capture/3313
https://github.com/medic/cht-core/issues/6363

As a "damage control" step, as discussed with the medic team, we plan to use Databricks (our tool primarily responsible for pulling couchDB data into our community database0 to also push "flags" to potential duplicate items. That will, in turn, cause tasks to trigger on the app. CHWs are expected to then confirm/deny possible duplicate and determine what should happen to the record (delete, merge, other).

jkuester:
## Overview

First I just want to clarify the space and existing problem-set to clarify exactly what should be addressed _in this issue_.  I think there are three separate, but related, problems to solve under the heading of "Duplicate prevention": 

1. Prevent user from creating a duplicate contact - (this current issue)
	- Runs in the webapp. Works offline.
	- Only checks against contact's *siblings* due to performance and data-access considerations.
2. Flag/merge duplicate contacts on the server - https://github.com/medic/cht-core/issues/6363
	- Sentinel transition? Historically, not possible to do much interesting here because of the limitations of Couch views.  However, [Lucene](https://github.com/medic/cht-core/issues/9542) may change the game.
[...]

PR #9609, opened 2024-11-04
fix(#9601): prototype duplicate prevention

ChinHairSaintClair:
# Description

This feature prevents duplicate hierarchy contact siblings from being created, as discussed with @jkuester and @mrjones-plip in a "technical working session" and through interactions on the linked issue thread.

To achieve this, we hook into duplicate detection strategies through "configuration", updated the `saveContact` method in the `form.service.ts` file to include an additional check, and display potential duplicates in a `duplicate_info` section added to the `enketo.component.html` file. Each duplicate is displayed as a card with expandible/collapsible segments, accompanied by an "acknowledgement" prompt to allow form submission.

## Configuration:
```json
{
   "expression":"levenshteinEq(3, current.name, existing.name)"
}
```

Currently two strategies are supported: `levenshteinEq` and `normalizedLevenshteinEq`, with the ability to customize properties based on implementation needs.

## Example implementation:
```json
{
    "title": [
      {
        "locale": "en",
        "content": "New Household"
      },
    ],
    "icon": "household-1",
    "context": {
      "expression": "contact.type === 'Indawo'",
      "permission": "can_register_household",
      "duplicate_check": {
          "expression":"levenshteinEq(4, current.name, existing.name)"
      }
    }
  }
```
Here, the `duplicate_check.expression` defines the logic for comparing the current record with its sibling. If no `duplicate_check` is provided, the system defaults to evaluating the `name` field. E.g:
- **For a "location"**, you might match on `street_number`, `street_name`, and `postal_code`:
```
"expression": "current.street_number === existing.street_number && levenshteinEq(3, current.street_name, existing.street_name) && current.postal_code == existing.postal_code"
```
- **For a "person"**, you might match on `name`, `sex`, and `date_of_birth`:
```
"expression": "levenshteinEq(3, current.name, existing.name) && current.sex === existing.sex && current.postal_code === existing.postal_code"`
```
In these expressions, `current` refers to the created/edited form, while `existing` refers to a "sibling" loaded from the database. 

## Conditional Duplicate Check:

This is a key requirement. For now, the `is_canonical` form question will be used to control conditional duplicate checking. When the backend flags a record as a duplicate, the CHW can mark the record accordingly. Downstream it will allow us to `merge` or `delete` the record based on the specified action.

## Opt-out:
Use the following configuration to disable the duplicate-checking functionality for a specific form:
```json
{
   "duplicate_check":{
      "disabled":true
   }
}
```

## Misc:
We use the CHT provided `medic-client/contacts_by_parent` view to query for siblings.

@kennsippell, since we've touched on the [duplicate topic before](https://github.com/medic/cht-core/issues/6363), it would be great to get your thoughts as this as well.

#Issue
Closes https://github.com/medic/cht-core/issues/9601

# Code review checklist

# License


webapp/src/ts/main.ts
```diff
@@ -7,6 +7,22 @@ window.startupTimes.firstCodeExecution = performance.now();
...
+    health_center: {
+      ...Levenshtein,
+      props: [
+        {form_prop_path: `/data/health_center/name`, db_doc_ref: 'name'},
+        {form_prop_path: '/data/health_center/external_id', db_doc_ref: 'external_id'}
+      ],
+      queryParams: {
+        valuePaths: ['/data/health_center/is_user_flagged_duplicate', '/data/health_center/duplicate/action'],
+        // eslint-disable-next-line eqeqeq
+        query: (duplicate, action) => duplicate === 'yes' && action != null
```
fardarter, line 21: @jkuester @ChinHairSaintClair My feeling here is that this would need to end up with some kind of operator syntax support that lets the user do logic, maybe something like (though doesn't have to be) JsonLogic: https://jsonlogic.com/

The way I see it there are two point of config -- the excel files and the configuration json -- and probably neither should take serialised JS (?).

If we return the `key => { key, value }` as a tuple or object on key lookup, we could let the user express logic maybe like this (with the value **found** at the key tested against the value **provided** (`with`) using the operator):

```js
let andOnly = {
  logic: [
    [
      // array bracket denotes a logical grouping
      {
        key1: { with: value, op: $in },
        key2: { with: value, op: $contains },
        key3: { with: value, op: $eq },
      },
      {
        key1: { with: value, op: $in },
        key2: { with: value, op: $ne },
        key3: { with: value, op: $startsWith },
      },
    ],
  ],
};

let orWithNestedAnds = {
  logic: [
    [
      {
        key1: { with: value, op: $in },
      },
      {
        key2: { with: value, op: $eq },
      },
    ],
    // each grouping in its own bracket
    [
      {
        key1: { with: value, op: $in },
      },
      {
        key2: { with: value, op: $in },
      },
    ],
  ],
};

let orWithNestedOrWithNestedAnds = {
  logic: [
    [
      {
        key1: { with: value, op: $in },
      },
      {
        key2: { with: value, op: $in },
      },
    ],
    // OR
    [
      [
        // nested groupings acceptable
        {
          key1: { with: value, op: $ne },
        },
        // AND
        {
          key2: { with: value, op: $ne },
        },
      ],
      // OR
      [
        {
          key1: { with: value, op: $contains },
        },
        // AND
        {
          key1: { with: value, op: $startsWith },
        },
      ],
    ],
  ],
};
```

webapp/src/ts/polyfills.ts
```diff
@@ -88,7 +88,14 @@ declare global {
     CHTCore: any;
     angular: any;
     EnketoForm:any;
+    _phdcChanges: { // Additional namespace
+      // Specify your own contact_types here
+      hierarchyDuplicatePrevention: Partial<{[key in 'person' | 'health_center']: Strategy;}>;
```
fardarter, line 93: @jkuester @ChinHairSaintClair Following on from my above comment:

> The way I see it there are two point of config -- the excel files and the configuration json.

We probably couldn't expect the user to do config here. I agree the type safety is lovely and I'm not sure whether it's possible to generate a type from the config, but if not I don't see how we can keep the type safety and still maintain the existing configuration contract.

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -299,6 +376,152 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
     };
   }
 
+  private parseXmlForm(form): Document | undefined {
```
fardarter, line 379: I'd move as much of this sort of stuff as is possible to a lib/file so it can be unit tested easily and then just called internally. Not sure if `private` is needed here. 

(I know this was just a conceptual prototype not productionised, but for forward looking feedback.)

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -299,6 +376,152 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
...
+
+    for (const field of props) {
+      if (Object.prototype.hasOwnProperty.call(sibling, field.compareToDocProp)) {
+        const test = this.strategy(field.value.toLowerCase(), sibling[field.compareToDocProp].toLowerCase());
+        totalScore += test;
+        count++;
+      }
+    }
+
+    return count > 0 ? totalScore / count : null;
```
fardarter, line 428: Does the alternative return type need to be `null` or is there an appropriate default value of type `int`?

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -299,6 +376,152 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
...
+  // Note: Look into 'https://www.npmjs.com/package/fuzzball' to perform various matching calculations
+
+  private cleanUpDuplicateSection() {
+    const $duplicateInfoElement = $('#contact-form').find('#duplicate_info');
+    $duplicateInfoElement.empty(); // Remove all child nodes
+    $duplicateInfoElement.hide();
+  }
+
+  // Promise.allSettled is not available due to the app's javascript version
+  private allSettledFallback(promises: Promise<Exclude<any, null | undefined>>[]): Promise<ReturnType[]> {
```
fardarter, line 466: @jkuester @ChinHairSaintClair  Would promise `allSettled` not be transpiled to a target version?

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -299,6 +376,152 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
...
+      Created on: ${date} <br>
+      <a class="duplicate-navigate-link" href="${link}" rel="noopener">Take me there</a> 
+      </p>`;
+  }
+
+  private outputDuplicates(duplicates: Duplicate[]){
+    const $duplicateInfoElement = $('#contact-form').find('#duplicate_info');
+    $duplicateInfoElement.empty(); // Remove all child nodes
+    $duplicateInfoElement.show();
+    // TODO: create a template component where these values are fed into.
```
fardarter, line 492: Does this need DOM mutation? Can the angular template not handle it?

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -299,6 +376,152 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
...
+    // TODO: Use angular instead of jQuery?
+    let content = this.buildDuplicateHeader(duplicates.length);
+    let count = 0;
+    for (const d of duplicates){
+      content += this.buildDuplicateItem(d, count+1);
+      count++;
+    }
+    $duplicateInfoElement.append(content);
+    $duplicateInfoElement.on('click', '.duplicate-navigate-link', () => {
```
fardarter, line 501: No anon functions on event listeners please. They don't get properly cleaned up by the browser. They need to be named.

webapp/src/ts/polyfills.ts
```diff
@@ -98,6 +105,39 @@ declare global {
...
+}
+
+export const Levenshtein: Strategy = {
+  type: 'Levenshtein',
+  threshold: 1.5,
+};
+
+export const NormalizedLevenshtein: Strategy = {
+  type: 'NormalizedLevenshtein',
+  threshold: 0.334,
```
fardarter, line 138: You'll probably want to set this in config

webapp/tests/karma/ts/modules/contacts/contacts-edit.component.spec.ts
```diff
@@ -680,13 +692,21 @@ describe('ContactsEdit component', () => {
...
       };
       formService.render.resolves(form);
 
       await createComponent();
       await fixture.whenStable();
 
       formService.saveContact.resolves({ docId: 'new_clinic_id' });
+      
+      // TODO: figure out why this test's dbLookupRef is null despite being set in the beforeEach
+      component.dbLookupRef = Promise.resolve({
```
fardarter, line 705: You may want `sinon.resolves`

webapp/src/ts/main.ts
```diff
@@ -7,6 +7,22 @@ window.startupTimes.firstCodeExecution = performance.now();
...
 window.$ = window.jQuery = require('jquery');
+window._phdcChanges = {
+  hierarchyDuplicatePrevention: {
+    health_center: {
+      ...Levenshtein,
+      props: [
+        {form_prop_path: `/data/health_center/name`, db_doc_ref: 'name'},
+        {form_prop_path: '/data/health_center/external_id', db_doc_ref: 'external_id'}
+      ],
+      queryParams: {
```
fardarter, line 18: maybe `formQuestion`? Name is still bothering me.

webapp/src/ts/main.ts
```diff
@@ -7,6 +7,22 @@ window.startupTimes.firstCodeExecution = performance.now();
 
 window.PouchDB = require('pouchdb-browser').default;
 window.$ = window.jQuery = require('jquery');
+window._phdcChanges = {
+  hierarchyDuplicatePrevention: {
+    health_center: {
+      ...Levenshtein,
+      props: [
+        {form_prop_path: `/data/health_center/name`, db_doc_ref: 'name'},
```
fardarter, line 15: camelCase is the js standard.

webapp/src/ts/main.ts
```diff
@@ -7,6 +7,22 @@ window.startupTimes.firstCodeExecution = performance.now();
...
+        {form_prop_path: '/data/health_center/external_id', db_doc_ref: 'external_id'}
+      ],
+      queryParams: {
+        valuePaths: ['/data/health_center/is_user_flagged_duplicate', '/data/health_center/duplicate/action'],
+        // eslint-disable-next-line eqeqeq
```
fardarter, line 20: @jkuester Continuing to lobby for `!=` as the most excellent exception to the strict equality lint rule. Checks for both `null` and `undefined`.

webapp/src/ts/services/form.service.ts
```diff
@@ -326,7 +331,39 @@ export class FormService {
...
   }
 
-  async saveContact(form, docId, type, xmlVersion) {
+  async checkForDuplicates(doc, duplicateCheck, acknowledged) {
+    const parentId = doc ? doc.parent?._id : undefined;
+    const contactType = doc ? doc.contact_type ?? doc.type : undefined;
+    const siblings = await requestSiblings(this.dbService, parentId, contactType);
+    const expression = extractExpression(duplicateCheck);
+    const isCanonical = doc.is_canonical ? doc.is_canonical === 'true' : false;
+    acknowledged = acknowledged ?? false;
```
jkuester, line 340: question: I had expected when I checked "Acknowledge" in the form that it would set `is_canonical` to `true` on my contact doc. That way, if I go back and edit the same contact later, it does not hit me with the dupe error again since I already confirmed the contact was real.  How were you planning to have `is_canonical` get set?

webapp/src/ts/services/form.service.ts
```diff
@@ -326,7 +331,39 @@ export class FormService {
     }, null);
   }
 
-  async saveContact(form, docId, type, xmlVersion) {
+  async checkForDuplicates(doc, duplicateCheck, acknowledged) {
+    const parentId = doc ? doc.parent?._id : undefined;
+    const contactType = doc ? doc.contact_type ?? doc.type : undefined;
+    const siblings = await requestSiblings(this.dbService, parentId, contactType);
```
jkuester, line 337: issue: we should definitely move this down after the if-check below since we do not want to query the contact siblings unless we are really going to be doing the dupe check.

webapp/src/ts/services/form.service.ts
```diff
@@ -326,7 +331,39 @@ export class FormService {
...
+        doc, 
+        siblings, 
+        {
+          expression, 
+          parseProvider: this.parseProvider, 
+          xmlFormsContextUtilsService: this.xmlFormsDuplicateUtilsService
+        }
+      );
+      return duplicates;
+    }
```
jkuester, line 353: nitpick: we can simplify/flatten this a bit by reversing the if-check:

```suggestion
    if (isCanonical || !expression || acknowledged) {
      return [];
    }
    return getDuplicates(
      doc, 
      siblings, 
      {
        expression, 
        parseProvider: this.parseProvider, 
        xmlFormsContextUtilsService: this.xmlFormsDuplicateUtilsService
      }
    );
```

webapp/src/ts/services/form.service.ts
```diff
@@ -335,6 +372,16 @@ export class FormService {
     const preparedDocs = await this.applyTransitions(docs);
 
     const primaryDoc = preparedDocs.preparedDocs.find(doc => doc.type === type);
+
+    const duplicates = await this.checkForDuplicates(
+      primaryDoc || preparedDocs.preparedDocs[0], 
+      duplicateCheck, 
+      acknowledged
+    );
+    if (duplicates && duplicates.length > 0){
```
jkuester, line 381: nitpick: if we keep the return type for `checkForDuplicates` consistent, then we can simplify this to just:

```suggestion
    if (duplicates.length){
```

jkuester (changes requested):
Wow, this is amazing! I think we are 100% on the right track here and I am excited to get this functionality released and in the hands of users.  

I have not had a chance to look at all the changes yet, but I wanted to go ahead and share my comments so far (mostly focused on the internal service stuff and have not looked much yet at the UI code). Mostly just a bunch of minor questions/suggestions.

I provided some suggestions regarding the Sonar complaints for `hasOwnProperty`, but I have not had a chance to dig into the circular navigation issue yet.

Thank you for the great PR and the patience to work with us and get the across the finish line!

webapp/tests/karma/ts/modules/contacts/contacts-edit.component.spec.ts
jkuester, line 1: suggestion: can we get a couple unit tests in here around calling the `onNavigateToDuplicate` and `onAcknowledgeChange` functions?  And a test for the `save` flow where calling `formService.saveContact` throws a `DuplicatesFoundError`?

jkuester (changes requested):
Okay, I made it through the rest of the files.  Added a few more comments, but all in all this is good stuff! I really appreciate the quality unit tests!  

I am going to make a post out on the forum that demos the current "Duplicates Found" UI we have here. The idea would be to maybe crowd-source some UX design in case other folks have input on the look/feel/function.

Finally, before we merge this code (but probably after we finalize the UX) we will need to add some e2e tests. Since there are a lot of moving parts here (config + db docs + forms), we should probably be thorough in validating the main flows. My current thought is to add some new tests cases to `tests/e2e/default/contacts/edit.wdio-spec.js` and maybe create a new `tests/e2e/default/contacts/create.wdio-spec.js`.

jkuester:
Okay, I created [a forum post](https://forum.communityhealthtoolkit.org/t/coming-soon-duplicate-contact-detection/4604/2) demoing some of the functionality we are building here.  Would love any UX feedback from your team on the look/feel/functionality!  :pray:

webapp/src/ts/services/xml-forms-context-utils.service.ts
```diff
@@ -31,4 +32,12 @@ export class XmlFormsContextUtilsService {
     return this.getDateDiff(contact, 'years');
   }
 
+  levenshteinEq(threshold: number, current: string, existing: string){
+    return current && existing ? levenshteinEq(current, existing) < threshold : false;
```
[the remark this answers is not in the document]
ChinHairSaintClair: Originally, I thought of the distance calculation and threshold as defining a "range" (0–3), where the upper bound is typically exclusive in programming.

I can see how that might be confusing, so I've updated the code to make the threshold inclusive. Good catch!

webapp/src/ts/services/xml-forms-context-utils.service.ts
```diff
@@ -31,4 +32,12 @@ export class XmlFormsContextUtilsService {
     return this.getDateDiff(contact, 'years');
   }
 
+  levenshteinEq(threshold: number, current: string, existing: string){
```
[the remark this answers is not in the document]
ChinHairSaintClair: Moved the threshold parameter as suggested. You raise a good point. While I’m hesitant to define a universal threshold since it depends heavily on the domain, as noted, a default of `3` seems reasonable for names based on:

1. **Name length considerations** – Shorter names (e.g., Amy) tolerate fewer edits than longer names (e.g., Christopher).
2. **Common typos and variations** – A threshold of `3` allows for small spelling differences (Jon vs. John) without excessive false positives.
3. **Phonetic similarity** – Names with different but phonetically similar spellings (Katherine vs. Catherine) would still be matched.

A stricter threshold (`2`) could be useful for high-precision matching as data normalizes over time, whereas a threshold of `4+` might introduce too many false positives.

One thing to keep in mind is that these considerations are based primarily on English names. If we want a more general approach across different languages, we may need to revisit this in the future.

That said, I don’t have extensive experience working with Levenshtein distance, so I’m open to adjustments if we find a better general default.

webapp/package.json
```diff
@@ -61,7 +61,9 @@
     "select2": "4.0.3",
     "signature_pad": "2.3.x",
     "tslib": "^2.5.3",
-    "zone.js": "^0.14.4"
+    "zone.js": "^0.14.4",
+    "levenshtein": "1.0.5",
```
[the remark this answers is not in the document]
ChinHairSaintClair: Maybe the lib author got it right the first time, why mess with perfection 😆 (just kidding). 
You raise a valid point, and I've updated the implementation to use `fastest-levenshtein`. Since this calculation will see significant traffic, especially on lower-spec devices, prioritizing performance makes sense.
Interestingly, `fast-levenshtein` uses `fastest-levenshtein` under the hood - yet ironically has more downloads. I'm not sure how much weight download count carries here.
The key difference is that the `levenshtein` library does not depend on any other library, whereas `fastest-levenshtein` depends on `levenshtein-edit-distance`, one of the libraries it compares itself to in its benchmarking (specifically version "2.0.5" `cli.js`).
Another difference would be that the levenshtein lib allows you to pretty print the [distance table](https://en.wikipedia.org/wiki/File:Levenshtein_distance_animation.gif), but that's just nice to have 🙂.

webapp/src/ts/services/utils/deduplicate.ts
```diff
@@ -0,0 +1,94 @@
...
+// The Levenshtein distance is a measure of the number of edits (insertions, deletions, and substitutions) 
+// required to change one string into another.
+const levenshteinEq = function (str1: string, str2: string): number {
+  return new Levenshtein(str1, str2).distance;
+};
+
+
+const requestSiblings = async function (dbService: DbService, parentId: string, contactType: string) {
+  const siblings: Doc[] = [];
+  const results = parentId && contactType && await dbService.get().query('medic-client/contacts_by_parent', {
```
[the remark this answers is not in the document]
ChinHairSaintClair: Agreed, code updated accordingly. Nice spot!

webapp/src/ts/services/utils/deduplicate.ts
```diff
@@ -0,0 +1,94 @@
...
+  if (results) {
+    // Desc order - reverse order by switching props
+    siblings.push(...results.rows.map((row: { doc: Doc }) => row.doc)
+      .sort((a: Doc, b: Doc) => (b.reported_date || 0) - (a.reported_date || 0)));
+  }
+  return siblings;
+};
+
+export type DuplicateCheck = { expression?: string; disabled?: boolean } | undefined;
+const extractExpression = function (duplicateCheck: DuplicateCheck) {
```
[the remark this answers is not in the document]
ChinHairSaintClair: That makes sense. The `| undefined` was originally included to maintain a specific property order (thinking about it now, not that it matters) and make it easier to use in multiple places, though I agree it’s less expressive.

Updating it to an optional parameter required reordering some properties since required parameters can’t follow optional ones, but the change makes the type cleaner and more intuitive. Thanks for the suggestion!

webapp/src/ts/services/utils/deduplicate.ts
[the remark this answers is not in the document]
ChinHairSaintClair: Converted to a service, it is much cleaner! 🙂

webapp/src/ts/services/utils/deduplicate.ts
```diff
@@ -0,0 +1,94 @@
...
+  // eslint-disable-next-line eqeqeq
```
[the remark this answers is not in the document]
ChinHairSaintClair: Good question! A simple falsy check should be fine here. While `{}` is truthy, the logic inside the function will handle it appropriately, so there should be no unexpected behaviour. The overhead difference is negligible.

webapp/src/ts/services/utils/deduplicate.ts
```diff
@@ -0,0 +1,94 @@
...
+
+export {
+  normalizedLevenshteinEq,
+  levenshteinEq,
+
+  requestSiblings,
+  extractExpression,
+  getDuplicates,
+
+  DEFAULT_CONTACT_DUPLICATE_EXPRESSION
```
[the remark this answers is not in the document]
ChinHairSaintClair: It does get used in the unit test. Should remove the export?

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -153,6 +165,10 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
     this.contentError = false;
     this.errorTranslationKey = false;
 
+    // Reset when when navigated to duplicate
```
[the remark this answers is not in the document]
ChinHairSaintClair: Agreed. Clearing the values will also be unnecessary if we navigate to the profile page instead of switching out the edit screen content.

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -55,6 +55,18 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
...
   private trackMetadata = { action: '', form: '' };
 
+  private duplicateCheck;
+  acknowledged = false;
+  onAcknowledgeChange(value: boolean) {
+    this.acknowledged = value;
+  }
+
+  onNavigateToDuplicate(_id: string){
+    this.router.navigate(['/contacts', _id, 'edit']);
```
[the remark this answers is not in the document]
ChinHairSaintClair: Initially we thought that the user might not have enough details around why the sibling was flagged as a duplicate, which can cause frustration.
The edit form "will" display all the information, where the profile page "could" display all the info depending on config. 
In this version of the duplicate item all values pertaining to that document are displayed before clicking off, which is then a compelling argument for simply navigating to the profile page. It's also a noticeable change, since it can be a bit jarring to the user if the page content just switches out.
That said, since the content of the duplicate item is now being reconsidered, maybe we should revisit this?

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -345,6 +364,11 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
             this.router.navigate(['/contacts', result.docId]);
           })
           .catch((err) => {
+            if (err instanceof DuplicatesFoundError){
+              this.duplicates = err.duplicates;
+              err = Error(err.message);
```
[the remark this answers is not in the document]
ChinHairSaintClair: I believe the idea was to omit dumping the duplicate data to console.

webapp/src/ts/services/form.service.ts
```diff
@@ -326,7 +331,39 @@ export class FormService {
     }, null);
   }
 
-  async saveContact(form, docId, type, xmlVersion) {
+  async checkForDuplicates(doc, duplicateCheck, acknowledged) {
+    const parentId = doc ? doc.parent?._id : undefined;
+    const contactType = doc ? doc.contact_type ?? doc.type : undefined;
+    const siblings = await requestSiblings(this.dbService, parentId, contactType);
+    const expression = extractExpression(duplicateCheck);
+    const isCanonical = doc.is_canonical ? doc.is_canonical === 'true' : false;
```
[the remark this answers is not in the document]
ChinHairSaintClair: I was under the impression that everything in `xlsx` resolves to string.With that reasoning in mind, the `is_canonical` value would need to be compared to `'true'`. If that's not always the case, maybe we should check for both `true` and `'true'`. What are your thoughts?

webapp/src/ts/services/utils/deduplicate.ts
```diff
@@ -0,0 +1,94 @@
...
+import { XmlFormsContextUtilsService } from '@mm-services/xml-forms-context-utils.service';
+
+
+export type Doc = { _id: string; name: string; reported_date: number;[key: string]: any };
+
+const DEFAULT_CONTACT_DUPLICATE_EXPRESSION = 'levenshteinEq(3, current.name, existing.name)';
+
+// Normalize the distance by dividing by the length of the longer string. 
+// This can make the metric more adaptable across different string lengths
+const normalizedLevenshteinEq = function (str1: string, str2: string) {
```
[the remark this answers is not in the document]
ChinHairSaintClair: That's a great question. As noted above, it's difficult to find a "one-size-fits-all" approach when comparing strings, especially when dealing with different domains. The `normalizedLevenshteinEq` method becomes useful in cases where the string length should play a role in the comparison. For example, with shorter strings (like nicknames or initials), small differences might not be as significant as they would be with longer strings (like full names or addresses).

In environments where community health workers (CHWs) often capture variations such as nicknames, short forms, or even common spelling errors, using a normalized approach can provide better results. It adjusts for the relative length of the strings being compared, which helps in cases where you might expect a more lenient match for shorter strings and stricter criteria for longer ones.

webapp/src/ts/services/utils/deduplicate.ts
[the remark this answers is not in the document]
fardarter: @ChinHairSaintClair If it's only used in a unit test, hardcode it in the test file. Code should be written to be testable, but that's the extent to which code should be adjusted for tests.

webapp/src/ts/modules/contacts/contacts-edit.component.ts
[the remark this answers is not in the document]
fardarter: Destructive reassignment is usually a code smell (the go community heard what I said). You don't know who else wants to use a variable or why and now they have to work around your choices.

webapp/src/ts/services/form.service.ts
ChinHairSaintClair: No, these are two separate concepts.

- Acknowledge: This allows the user to review and compare possible duplicates, then proceed with saving the record—knowing a duplicate may exist—without blocking progress. It does not change how future duplicate checks behave.
- `is_canonical`: This explicitly marks the contact as the true record, signalling to the backend that this is the authoritative version and that other duplicate records should be dealt with (either merged, deleted, or archived).

Acknowledgment is available to all users and primarily serves to let them continue working despite potential duplicates. In contrast, `is_canonical` is an opt-in mechanism that bypasses duplicate checks entirely for that record.

While we're on the subject, have we decided on a prefix standard for xlsx questions? I have not seen any additional comments on the [forum thread](https://forum.communityhealthtoolkit.org/t/how-to-differentiate-between-custom-fields-in-documents-and-fields-used-by-the-cht/4547).

webapp/src/ts/services/xml-forms-context-utils.service.ts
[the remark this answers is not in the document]
jkuester: :+1: My vote is we start with 3 as the default and we can adjust later if needed.

webapp/src/ts/services/utils/deduplicate.ts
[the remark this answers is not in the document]
jkuester: Awesome!  Thank you for the explanation!  I have made a note of it for future docs updates.

webapp/package.json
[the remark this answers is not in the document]
jkuester: I was looking at `fast-levenshtein` as well. It seemed like it maybe just added some extra functionality on top of `fastest-levenshtein`. Ultimately, I agree we want to optimize for performance (especially on low-spec devices). Additionally, we should be sensitive about adding more kb to our app bundle size. :grimacing: IMHO it seems like `fastest-levenshtein` can provide a good balance for us and a good place to start.  However, as you noted in a previous comment, we may decide in the future that the [locale sensitive comparisons](https://github.com/hiddentao/fast-levenshtein?tab=readme-ov-file#examples) from `fast-levenshtein` are needed for non-English use cases...

webapp/src/ts/modules/contacts/contacts-edit.component.ts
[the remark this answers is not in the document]
jkuester: Yeah, that is a good point that this could be affected by the conclusions of the [UX discussion](https://forum.communityhealthtoolkit.org/t/coming-soon-duplicate-contact-detection/4604).  

> The edit form "will" display all the information, where the profile page "could" display all the info depending on config.

Speaking purely from my opinion (with no real-world data to back me up), my _expectation_ would be that any info important enough to be dupe-checking on would also be important enough to display via the contact-summary on a contact's profile page.  

I can see where you are coming from with the consistency of moving from entering data in a contact create form to now being in the contact edit form were we can review the data and make any necessary changes.  On the other hand, though, I am sure we will get cases like this:

- The user needs to submit an app form for a child (e.g. a u5 assessment)
- The user forgot they already created a contact for the child and proceeds to open a create-person form to add a new person to the household
- Their new child contact is flagged as a duplicate when trying to submit the create-person form
- The user selects "Take me there" for the exiting child, intending to go to that contact and launch the app form.
- Instead the user is taking to the edit form for the existing contact and has to cancel out of that before proceeding to the app form

---

Ultimately, you and your team are closer to the end-users then I am and I will happily defer to your expertise with these kinds of things!  I hope it is useful for me to raise questions and point out other possible behaviors we can discuss, but ultimately I am not any kind of UX expert nor do I have direct access to regularly observe/interact-with CHT users.

webapp/src/ts/modules/contacts/contacts-edit.component.ts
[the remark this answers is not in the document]
jkuester: > I believe the idea was to "box" the value and omit dumping the duplicate data in console.

:+1: Thanks for the context. I understand where you are coming from here now.

FWIW, I don't know that we really have any solid patterns around "destructive reassignment" in the CHT code-base. :sweat_smile: So, I am happy to go with which ever way seems most clean to you.

webapp/src/ts/services/form.service.ts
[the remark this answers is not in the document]
jkuester: Ah, yes, you are 100% correct here. I forgot that Enketo does weird stuff with booleans....  :grimacing: 

All good to just leave this checking for `'true'`.

webapp/src/ts/services/form.service.ts
jkuester: Okay, so conceptually your distinction here makes sense. Can you help me understand a bit more about the `is_canonical` workflow? How will the value get set? Who has permission to mark a contact as canonical?  If two contacts match as duplicates, but are both real people (e.g. a father and son with the exact same name) should both be marked as `is_canonical`?

The main reason I expected the "Acknowledge" to set `is_canonical` was because I was hoping to avoid the user being warned about duplicates multiple times for the same contact.  (Each time the contact was edited it would trigger the same dupe warnings if `is_canonical` was not set....)

---

Regarding the prefix standard, I have bumped my forum question to some of the other devs that have done a lot of CHT work.

webapp/src/ts/services/utils/deduplicate.ts
```diff
@@ -0,0 +1,94 @@
...
+  return siblings;
+};
+
+export type DuplicateCheck = { expression?: string; disabled?: boolean } | undefined;
+const extractExpression = function (duplicateCheck: DuplicateCheck) {
+  // eslint-disable-next-line eqeqeq
+  if (duplicateCheck != null) {
+    if (Object.prototype.hasOwnProperty.call(duplicateCheck, 'expression')) {
+      return duplicateCheck.expression as string;
+    } else if (Object.prototype.hasOwnProperty.call(duplicateCheck, 'disabled') && duplicateCheck.disabled) {
```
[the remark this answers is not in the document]
ChinHairSaintClair: Good catch!

webapp/src/ts/modules/contacts/contacts-edit.component.ts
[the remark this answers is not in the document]
ChinHairSaintClair: You make a good point. Making decisions for the user without fully understanding their behavior and needs could backfire. Navigating to the profile screen, where they have the option to do x, y, or z, is better than forcing them into an edit.

That said, IMO we should provide the user with enough information before navigating to the item, as they will lose the form data. This is especially important when considering multiple duplicates, as discussed in the [forum thread](https://forum.communityhealthtoolkit.org/t/coming-soon-duplicate-contact-detection/4604).

webapp/src/ts/services/form.service.ts
ChinHairSaintClair: ## Who Sets `is_canonical`?
CHWs, who have the most contextual knowledge, are responsible for marking records as canonical.

**Captured in the following ways: **
1. **No duplicate found during submission** → Set `is_canonical=true` automatically on save.
2. **Duplicate found** → Prompt the user to acknowledge that the current record is canonical before proceeding.
3. **Editing an existing record** → Display the current `is_canonical` status (if any) and allow the user to reset it. Additionally, provide a "Mark for Deletion" option for records that need cleanup (particularly useful for users without delete permissions).

For scenario 3, consider a user reviewing records after this feature is introduced. They may initially mark a record as canonical but later decide to investigate further. The ability to reset the status allows for flexibility. Meanwhile, marking a record for deletion signals that it should be removed from the system.

## Duplicate Checks and Warnings
The goal is to eventually remove non-canonical records, so they no longer trigger duplicate warnings or remain available for editing. 

Until then, we should not suppress duplicate warnings. Doing so would prevent users from revisiting and resolving potential issues later.
1. For non-canonical records → Duplicates will continue to be displayed as usual.
2. For canonical records → Introduce a "Check for Possible Duplicates" button to allow users to manually trigger a duplicate check. Useful if you want to fact check as mentioned in the previous section.

## Handling Similar but Distinct Contacts (e.g., Father and Son with the Same Name)
Even if two contacts appear similar, they should both be marked as `is_canonical` if they are distinct. For instance, in the father-son scenario, a key differentiator like date of birth should be used. Implementers may need to fine-tune duplicate detection criteria to reduce false positives.

## Detail on duplicate items
We're of the opinion that duplicate items should contain enough data, not all, to defer navigating until absolutely necessary - as losing the current form data could cause frustration. 
If we don't five the user enough info ahead of time, we're worried users might feel forced to submit incorrect data just to bypass the duplicate screen and not lose data. Leading to bad records in the system. 
The `is_canonical` property should be displayed in the duplicate summary.

jkuester: Thank you for the additional context here. This is super helpful!

> For instance, in the father-son scenario, a key differentiator like date  
of birth should be used. Implementers may need to fine-tune duplicate   
detection criteria to reduce false positives.  

This statement has got me to thinking that perhaps we do not actually need embedded CHT logic for handling false positives. Instead, I think most of your needed `is_canonical` workflows could be supported by custom config. This would give you (and other projects) total control over the specifics of the `is_canonical` logic without adding additional data-model overhead in the CHT to other projects that do not need it.

You can already include custom logic in your `contact` forms for adding/editing an `is_canonical` field on your contact docs. Then, in your custom `duplicate_check.expression` logic, you can always include logic for checking the `is_canonical` field. This would let you avoid duplicate warnings on a contact marked as `is_canonical`.  For these workflows, I don't see a clear need/benefit to have extra functionality built into the CHT.

> Additionally, provide a "Mark for Deletion" option for records that need  
cleanup (particularly useful for users without delete permissions).  

The same goes for this logic.  I guess your current `contact` forms can support marking contacts for review/deletion, right? (Thinking that this would basically just be a custom field set on contact docs that your custom server-side scripting could detect.)

There is definitely *a lot* more possible functionality that could be built into the CHT around this to make managing contacts more reasonable and less of a manual/custom process, but ultimately, that probably needs its own project and dedicated design discussions... (One idea that would be interesting to pursue in a future project would be a Sentinel transition that could perhaps be responsible for flagging duplicate contacts. However, I think it is important to recognize that detecting/handling duplicate contacts on the server is somewhat distinct from the main goal of this PR, which is simply to reduce the number of duplicates contacts that get created in the first place (by warning the user)).  

---

## Detail on duplicate items

I really appreciate all the great discussion on this! Forcing the user to navigate away from the form and loose everything they entered just to discover that the contact is not actually a duplicate is a big issue.  Lets definitely continue the conversation in the forum thread and I am confident we can find the optimal balance between data and simplicity for this MVP!

fardarter: > For these workflows, I don't see a clear need/benefit to have extra functionality built into the CHT.

@jkuester Maybe I'm not following the discussion closely enough, but for me one of the core things the `is_canonical` achieves is helping us determine whether we need to run the duplicate check. If the duplicate check is to be built in (which I think we agree would be great?), then how else would you decide not to run the check? The purpose of `is_canonical` afaik isn't to help reduce false positives, but those moving away from the default config (which would just check names) would have more control over what flags when a check is done.

So, the benefit to the platform would be that you wouldn't get unpleasant/annoying/non-performant user behaviour with the system and also wouldn't get a prompt that might lead to an error on an item already considered resolved. 


Just to clarify the intended setting logic for `is_canonical`:

Create:

- No duplicate found during submission → Set is_canonical=true automatically on save.
- Duplicate found → Prompt the user to acknowledge (wording can be changed) *that they are asserting the new data they are adding has no duplicates in the system and therefore the new entry can validly be marked as canonical*.

Edit:
- No duplicate found during submission → Set is_canonical=true automatically on save.
- Duplicate found → Prompt the user to acknowledge (wording can be changed) *that they are asserting the new data they are adding has no duplicates in the system and therefore the new entry can validly be marked as canonical*.

Additional data management features:

Caveat: I don't know the UI very well so take these as loose ideas from a UI perspective.

- Unset `is_canonical`. This can't be on the edit page without adding complexity as the above flow would simply reset it. Semantically, if we added it to the create/edit flow this should be something like "Create/mark as draft", which is largely unsupported. So the reset would need to take place as a data management operation somewhere on the UI, maybe close to delete.
- Mark for deletion: This should only appear if the user does not have delete permissions, otherwise it should just be the delete button. 

Editing an existing record → Display the current is_canonical status (if any) and allow the user to reset it. Additionally, provide a "Mark for Deletion" option for records that need cleanup (particularly useful for users without delete permissions). From a data management perspective, in general, I think we need to see something not marked as `is_canonical` as being in an uncertain/undecided state rather being deleteable, so resetting the `is_canonical` field just move it from a "we're sure" to a "we're unsure", where mark for deletion is a "we're sure this was wrong".

For me, these are useful core platform primitives that can be evolved. Data accuracy is critical to the success of products like CHT (at least for us) and my view is that effective delegated tooling for data management would be a major contributor to success. 

> the main goal of this PR, which is simply to reduce the number of duplicates contacts that get created in the first place (by warning the user)).

I think this isn't exactly correct for us. We *also* need the tools to provide affordance to our CHWs to give them the correct level of agency here -- leaving them in (well-mediated) control is *critical*. And I would suggest that there are very few people in the world (ie, not just our use case) who wouldn't appreciate that.

We're obviously in a position to make use of a marked for deletion field immediately, and I agree it would take a larger project to integrate them with something like Sentinel and that's a constraint. 

However, it feels to me like a really desirable *direction*, so what about adding some of the data management elements we need under a feature flag? 

One of the core reasons we're engaging with the CHT code base is that from a maintainability perspective we'd like to own as little code and config as possible, which will also help us stay aligned with future CHT choices.

ChinHairSaintClair: Implementing `is_canonical` as a built-in functionality in CHT ensures that users make informed decisions before bypassing the duplicate check. If users could set `is_canonical` directly from the form, they might do so without evaluating potential duplicates. This could lead to more duplicates, which, as noted, is one of the things this feature aims to prevent. Instead, by requiring users to first review duplicates (if any) before confirming a record as canonical we ensure that the decision is intentional - "I have looked at the duplicate results and I'm happy with this record being saved as canonical".

jkuester: > - No duplicate found during submission → Set is_canonical=true automatically on save.
> - Duplicate found → Prompt the user to acknowledge (wording can be changed) that they are asserting the new data they are adding has no duplicates in the system and therefore the new entry can validly be marked as canonical.

> by requiring users to first review duplicates (if any) before confirming a record as canonical we ensure that the decision is intentional - "I have looked at the duplicate results and I'm happy with this record being saved as canonical".

😅 So, circling back to my initial comment in this thread, when a user checks "Acknowledge" and continues with the creation/edit even when duplicates are found, then we _will_ be marking the contact as `is_canonical`?  Sorry if I misunderstood parts of the above responses and have made this more confusing than it needed to be, but I thought the idea with the original implementation here was to not trigger the `is_canonical` from the "Acknowledge" checkbox. (But maybe that idea has evolved as a part of this discussion...)

If that is the case, though, it would effectively mean we would be setting `is_canonical` on _every_ new contact, right?  (Contacts with no dupes found would get `is_canonical` and for contacts with dupes found, if the user selects "Acknowledge" and proceeds to add the contact we will also be setting `is_canonical`.)  If that is the case, I am unsure of the benefits of `is_canonical` from an overall data management perspective (since all the contacts in the db will have that field set to true).

I do not think `is_canonical` contains enough information to be useful in terms of data validation. Marking contacts as `is_canonical` could produce some complex scenarios as config evolves.  There is no proposed mapping/history between `is_canonical` and any identified duplicate contacts and/or which version of the `duplicate_check.expression` was used when setting `is_canonical`. So, just seeing that a contact is marked as `is_canonical` does not tell us if the contact has been checked by the latest version of the `duplicate_check.expression` or if maybe other duplicate contacts have been added/removed since `is_canonical` was set. 

To me, it seems like if a user added a contact, then the user considered that contact to be canonical (and we don't need another field to indicate this).  As you noted, `is_canonical` could prevent future duplicate check warnings, but I am not even confident in that workflow anymore. If all contacts get `is_canonical: true` when they are created, then they all will have `is_canonical: true` when they are edited and so we would never perform a dupe-check when editing a contact.  Admittedly, if we do not have `is_canonical`, we have no way of preventing follow-up dupe checks for a particular contact.  I think this might be okay because:

- If there is no performance problem running the dupe-check when creating a contact, there should not be a performance _need_ to prevent subsequent dupe checks.
- The duplicate check logic should be setup such that "false positive" duplicates are extreme edge cases. (I think we should change our default check to validate both `name` and `date_of_birth` (if it exists).) Basically the dupe check should be based on logic that can uniquely identify a contact. A properly tuned dupe-check expression should minimize the number of times a user every sees a false-positive.  

---

Ultimately, I am not trying to make a definitive statement here about the future of the CHT data model. I am very eager to continue collaborating on ways to improve data validation and data management in the CHT. For the sake of this PR, though, I am not seeing a critical problem that `is_canonical` is solving (while I have imagined at least a few headaches that it might trigger in the future....).

I understand that the needs of your deployment go way beyond just adding some user decision support around duplicate contacts. But this is a useful feature and one that we are very close to completing in this PR! I think we should focus on solving this smaller issue in this PR and continue working towards solutions to other problems in other issues/PRs. (It seems like not including `is_canonical` in this PR gives us the freedom to add it later if we conclude it would be useful, or to take a different approach since we would not be limited by a previous decision to add `is_canonical`...)  With that in mind, I have logged https://github.com/medic/cht-core/issues/9779 where we can refine a data management approach!

---

P.S. As I finish writing this, I can sense some tunnel vision in my own perspective here from spending too long thinking about this :sweat_smile:.  I am going to reach out to some collegues for another fresh perspective on this question that will hopefully be less muddled than my brain right now....

mrjones-plip: Wow!  There's thoughtful discussion here - folks are taking a lot of time to weigh the benefits and drawbacks of how `is_canonical` might work. Thank you all!

@ChinHairSaintClair - are you open to shipping this PR without `is_canonical`? I think bifurcating the ideas here will give us more time to think about how to best proceed.  This PR would likely ship relatively (entirely?) isolated in [4.18](https://github.com/medic/cht-core/milestone/141) which sets up nicely for easily testing in production to get good data without needing to take other features too.

Speaking of good data - is there telemetry gathered on how often we find a possible dupe, how many possible dupes we found etc.?  Sorry, I haven't followed the PR that closely to know either way.

Finally, two other options might be:
* ship this as a [Feature Release](https://docs.communityhealthtoolkit.org/contribute/code/releasing/feature_releases/) to test in production (would very much rely on good telemetry!) 
* have a real time call to discuss best path forward for this PR

fardarter: @jkuester @mrjones-plip 

>  So, circling back to my initial comment in this thread, when a user checks "Acknowledge" and continues with the creation/edit even when duplicates are found, then we will be marking the contact as is_canonical?

You're right that this has evolved in the discussion. I took the view that it doesn't make a lot of sense to think that a duplicate-checked user isn't canonical by default, especially given your concern about spamming checks (which we've walked back on, I guess). Given the right tools to help prevent dupes, we should assume our user is acting in good faith -- it's just that hitherto they've not had any defences.

@ChinHairSaintClair have chatted and I think we can come on board with dropping `is_canonical` from the data structure and moving it to the form field and running the check each time on edit. We'll have to tone down/rephrase some of the warnings on edit, I think. 

The issue we're left with in this flow is when (as in our case) we have a user performing an edit that can't do a delete. If they're editing a contact, unlike for create, that contact already exists. So, let's say they're now shown a list of duplicates, identify one of the contacts in the list as the duplicate, then... what? They have no valid action to take by default.

In our flavour of CHT, a delete restriction was put in place on CHWs to ensure contact deletion was double-checked by our digital health officers/CHW supervisors, given the risk of significant data loss on a delete action (and given the low average of digital literacy of our CHWs). We also found CHWs were able to (and did!) delete themselves or colleagues, which caused problems (some form of user management RBAC would be valuable here). I suspect restricting delete might be something others want to do. 

In the meantime, we can use a form field for a "Mark for deletion", but it seems to me that this would be a valuable platform level UI feature where a user doesn't have delete or maybe has an explicit `allowedActions: ['MarkForDeletion']` permission (and a confirmation flow to a supervisor). I'd see it as a platform enhancement rather than something just for us, even if we're the only one's using it at present.

@mrjones-plip We're not added any telemetry. Do you have a suggestion on how?

fardarter: I'm not sure if it means we need to be cautious, but think it's good information to know in case someone can think of a concern.

jkuester: > The issue we're left with in this flow is when (as in our case) we have a user performing an edit that can't do a delete. If they're editing a contact, unlike for create, that contact already exists. So, let's say they're now shown a list of duplicates, identify one of the contacts in the list as the duplicate, then... what? They have no valid action to take by default.

Ahhh, now I understand how the "Mark for deletion" issue connects back to the duplicate check!  I agree it is bad to leave the user with no options in this case.  (Though it may be acceptable to ship the dupe check feature while we continue to work on a resolution for this elsewhere.)

> In the meantime, we can use a form field for a "Mark for deletion", but it seems to me that this would be a valuable platform level UI feature where a user doesn't have delete or maybe has an explicit allowedActions: ['MarkForDeletion'] permission (and a confirmation flow to a supervisor). I'd see it as a platform enhancement rather than something just for us, even if we're the only one's using it at present.

One thing I have been considering here is how the [muting workflow](https://docs.communityhealthtoolkit.org/building/workflows/muting/) might be able to help us here.  That already triggers platform-level functionality to do things like grey-out the contact in the contact list and avoid tasks or notifications for the muted contact. I wonder if you could have a "Mark for deletion" form that the CHW could submit for a contact that would mute the contact and then notify a supervisor that the contact has been marked for deletion... :thinking: 

However, the muting approach would only be helpful in the context of dupe checks if we did not include muted contacts in the duplicates check (or if you included custom logic in your `expression` to avoid matching on muted contacts).  I have gone back and forth in my head on whether to include dead/muted contacts in the dupe checking by default.  At this point I think it makes more sense to include them, but depending on your muting workflows, it may be useful for your `expression` to not match on contacts with a `muted` property....

That being said, lets continue the discussion regarding the Mark For Deletion workflow over in https://github.com/medic/cht-core/issues/9779.  If the existing functionality is not sufficient, then I do agree that it would be best to enhance the platform to better support this kind of thing.  :+1: 

---

> We should be cautious when using the date_of_birth field for default comparisons, as it may be an approximation. We've found that patients often don't know or are unwilling to share their exact date of birth due to varying levels of trust. This results in different ways of capturing dates, such as:

This is a good point that in many cases the DOB is not super reliable. But it seems like it would be better than nothing!  Maybe, for the default check, we could even use a "1 year" threshold (or even more).  Something like `levenshteinEq(current.name, existing.name, 3) && ageInYears(current) === ageInYears(existing)`

jkuester: > We're not added any telemetry. Do you have a suggestion on how?

Sorry, I realized this comment was never addressed. Practically speaking, you record telemetry by calling the [telemetry.service `record` function](https://github.com/medic/cht-core/blob/23da4f74773288b6b2e7820263e49aec14642edf/webapp/src/ts/services/telemetry.service.ts#L309).  We use telemetry to collect usage data, particularly from offline users, so that it gets eventually replicated back to the server where it can be analyzed when necessary.

That being said, exactly what data to collect and where/when to collect it is somewhat subjective for a new feature like this (trying to anticipate what might be useful to know...).  If it works for you, once this PR is ready to review again, I would be happy to branch off of it and add some telemetry calls that I think would be important. then I can open a PR for you to review and get that branch merged back in here....

jkuester:
Sorry for the delayed response here! I will not be able to have a look at the revisions this week, but I plan to do so early next week.  :+1:

webapp/src/ts/components/duplicate-info/duplicate-info.component.html
```diff
@@ -0,0 +1,48 @@
...
+      <hr>
+      <p>
+        <strong>{{ 'duplicate_check.contact.name'  | translate }}</strong> {{ duplicate.name }}
+        <br>
+        <strong>{{ 'duplicate_check.contact.created_on' | translate }}</strong> {{ duplicate.reported_date | date: 'EEE MMM dd yyyy HH:mm:ss' }}
+      </p>
+      <p *ngIf="summaryRequestInfo?.contact_id === duplicate._id && summaryRequestInfo?.error" class="error_msg">
+        <strong>{{ summaryRequestInfo?.error?.toString() }}</strong>
+      </p>
+      <button *ngIf="(!duplicate._summary && !summaryRequestInfo?.isLoading) || (!duplicate._summary && summaryRequestInfo?.contact_id === duplicate._id && summaryRequestInfo?.error)" class="load-more-button" (click)="_loadContactSummary(duplicate._id)">
```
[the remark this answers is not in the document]
jkuester: Actually, what do you think about getting rid of this button altogether and making the whole `card` here have a click-listener that triggers the contact-summary details to be shown?  I am concerned about the usability being sub-par for trying to tap this "button" on a small  touch-screen.  Instead of adding a bigger button, we just make the whole row the button.

Then, we could also nest the `_navigateToDuplicate` button down inside of the `nested-section`. Then the flow for the user would be:

- Review the list of duplicate contact names/creation-dates
- Tap one of the rows - it displays the contact summary data and the nav button
- The user decides that contact is the desired one and taps the "Take me there" button to navigate to the existing contact.

api/resources/translations/messages-en.properties
```diff
@@ -576,6 +576,13 @@ contact.type.place.edit = Edit place
 contact.type.place.new = New place
 contact.type.wrong = Wrong type, contact is not a person.
 contact.updated = Contact updated
+duplicate_check.contact.duplication_message = {{count}} duplicate siblings found
+duplicate_check.contact.acknowledge = Acknowledge duplicate siblings and proceed with submission
+duplicate_check.contact.item_number = Item number:
+duplicate_check.contact.name = Name:
+duplicate_check.contact.created_on = Created on:
+duplicate_check.contact.load_summary_info = Load contact summary info
+duplicate_check.contact.go_to = Take me there
```
jkuester, line 585: nitpick: A more common wording would probably just be "Open".  I think it would still be clear enough that by selecting that button you would be leaving the current form and going to that contact.

jkuester (changes requested):
Things are coming along great here!  :star_struck:

ChinHairSaintClair:
I've pushed up the UI changes, translation strings, duplicate-contacts expand/collapse functionality, and the shift of state management to the child component for your review. I'll finish up the deduplication and form service next week.

webapp/src/ts/components/duplicate-info/duplicate-info.component.html
```diff
@@ -0,0 +1,48 @@
+<div
+  id="duplicate_info"
+  [ngStyle]="{ display: duplicates.length > 0 ? 'block' : 'none' }">
+  <p class="results_header">
+    {{ ('duplicate_check.contact.' + entityType + '.duplication_message' | translate: { count: duplicates.length }) !== ('duplicate_check.contact.' + entityType + '.duplication_message') 
+     ? ('duplicate_check.contact.' + entityType + '.duplication_message' | translate: { count: duplicates.length }) 
+     : ('duplicate_check.contact.duplication_message' | translate: { count: duplicates.length }) }}
```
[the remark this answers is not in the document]
ChinHairSaintClair: If the `duplication_message` is meant to provide context about what the list contains - such as explaining the properties used to identify duplicates - we'll need a separate message for each contact type, as the expression for each type may differ. Having a specific message per type could be useful, especially early on when the CHW might not fully understand why certain items are flagged. This would help them determine whether they are false positives and if they can proceed with submission. A generic message like "Possible duplicate contacts found:" might not provide enough context.

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -59,6 +60,47 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
...
+    this.summaryRequestInfo = { contact_id: id, isLoading: true, error: undefined };
+    try {
+      const contact = this.duplicates.find(x => x._id === id);
+
+      if (!contact) {
+        throw new Error(`Contact with ID ${id} not found`);
+      }
+
+      // Remove "reserved" fields
+      const sanitizedContact = Object.keys(contact).reduce((acc, key) => {
```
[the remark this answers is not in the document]
ChinHairSaintClair: The goal is to avoid showing the same properties in both the contact summary and the general view. Right now, `name` and `reported_date` are displayed before the summary info loads, so we’re making sure they don’t get repeated. If needed, we can add more properties to the general view.

webapp/src/ts/services/xml-forms-context-utils.service.ts
```diff
@@ -31,4 +32,20 @@ export class XmlFormsContextUtilsService {
     return this.getDateDiff(contact, 'years');
   }
 
+  // The Levenshtein distance is a measure of the number of edits (insertions, deletions, and substitutions) 
+  // required to change one string into another.
+  levenshteinEq(current: string, existing: string, threshold: number = 3){
+    return current && existing ? distance(current, existing) <= threshold : false;
```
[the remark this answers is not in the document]
ChinHairSaintClair: Good point! I’ve updated the function to check if both values are strings before calling `distance()`, ensuring we only apply Levenshtein logic when it makes sense. Otherwise, we now fall back to a direct `===` comparison, which also covers non-string truthy and falsy values. This keeps behaviour consistent and avoids forcing implementers to do extra checks beforehand.

webapp/src/ts/components/duplicate-info/duplicate-info.component.html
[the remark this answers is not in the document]
jkuester: Ohhh yes!  Thanks for reminding me about that.  You are right. With different custom duplicate checking logic, it could be very necessary to give different messages. :+1:

webapp/src/css/enketo/medic.less
```diff
@@ -443,6 +443,72 @@
...
+
+      .btn.submit.btn-primary {
+        justify-self: flex-end;
+      }
+    }
+
+    .error_msg {
+      color: #e33030;
+    }
+  }
```
jkuester, line 511: suggestion: these are the css changes needed to go with my suggested html:

```suggestion
  #duplicate_contacts {
    .results_header {
      font-size: @font-large;
      padding: 0;
    }

    .acknowledge_label {
      -webkit-user-select: none; -ms-user-select: none; user-select: none;
      margin: 15px 0 0 0;
    }

    mat-expansion-panel {
      mat-expansion-panel-header {
        height: auto;

        .mat-content {
          font-size: @font-large;
          display: flex;
          flex-direction: column;
          padding: 0.5em 0;
        }

        .mat-expansion-indicator::after {
          border-width: 0 3px 3px 0;
          padding: 4px;
          color: @text-normal-color;
        }
      }

      .nested-section {
        font-size: @font-medium;
        justify-content: center;
        display: flex;

        .error_msg {
          color: #a94442;
        }
      }

      mat-action-row {
        font-size: @font-medium;
      }
    }
  }
}
```

webapp/src/ts/components/duplicate-contacts/duplicate-contacts.component.ts
```diff
@@ -0,0 +1,71 @@
...
+      return;
+    }
+
+    this.isLoading = true;
+    this.error = undefined;
+
+    try {
+      this._summary = await this.loadContactSummary?.(duplicateId);
+    } catch (err) {
+      this.error = `Unable to load summary data for contact ${duplicateId}`;
```
jkuester, line 58: issue: Can we load this message text from the translations?  (Also, I don't think the `duplicateId` will be super helpful to the user. They will already have the name/reported_date for the contact.)

jkuester (changes requested):
This is coming along great!  Thank you for the updates!

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -59,6 +61,40 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
...
+  private duplicateCheck?: DuplicateCheck;
+  duplicatesAcknowledged = false;
+
+  duplicates: Contact.v1.Contact[] = [];
+  entityType: string = '';
+
+  private readonly omitProperties = ['reported_date', 'name'];
+
+  toggleDuplicatesAcknowledged() {
+    this.duplicatesAcknowledged = !this.duplicatesAcknowledged;
```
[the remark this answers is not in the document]
ChinHairSaintClair: Clearing the duplicate list will remove the entire duplicate section from view - including the 'duplicate acknowledge toggle'. This means users won't be able to toggle the checkbox if they click it accidentally. I concerned this might do more harm than good. What are your thoughts?

webapp/src/ts/services/deduplicate.service.ts
```diff
@@ -0,0 +1,68 @@
+import { Injectable } from '@angular/core';
+import { DbService } from '@mm-services/db.service';
+import { ParseProvider } from '@mm-providers/parse.provider';
+import { XmlFormsContextUtilsService } from '@mm-services/xml-forms-context-utils.service';
+
+export type Doc = { _id: string; name: string; reported_date: number;[key: string]: any };
+const DEFAULT_CONTACT_DUPLICATE_EXPRESSION =
+  'levenshteinEq(current.name, existing.name, 3) && ' +
```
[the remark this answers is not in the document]
ChinHairSaintClair: I think we should lead with the standard `levenshteinEq`. Explaining that the matching works by allowing up to three differences between strings is much easier for users to understand as default behaviour. In contrast, the normalized method, which also considers string length, is harder to explain. More advanced users will likely explore that aspect over time as their understanding of the data & fuzzy mathcing improves.

webapp/src/ts/modules/contacts/contacts-edit.component.ts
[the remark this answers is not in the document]
jkuester: As with all things, I blame GitHub Copilot....  😅

ChinHairSaintClair:
@jkuester I ran into a bug in the top-level place duplicate check. Create two top-level places with the name 'test'. On the second creation, you will NOT be prompted with a duplicate check. Logging out/in fixes this. I believe this might be related to the cache used in the `get` function. Do you know of an effective workaround? Lower-level checks still work as expected.

jkuester:
Oh wow @ChinHairSaintClair this is a good find!  After debugging this, it seems like the cache invalidation for the `ContactsService` is just broken in general.... :facepalm:   You can re-create similar behavior (where the cache is not busted) by adding a new top-level place and then going to the Reports tab and trying to filter on "Place".  Your new top-level place will not show up in the list of places until after you log out and log back in....  :sweat: 

The good news is that the fix should be simple.  Basically, the `invalidate` function for the cache is actually called with a _change_ object, not the actual doc itself.  In most cases, the _change_ object will have a `doc` property that does contain the doc that changed.  We just need to call `contactTypesService.getTypeId` with that inner doc instead of with the change object.  

So, line `39` in the `contacts.service.ts` just becomes:

```ts
            invalidate: ({ doc }) => type.id === this.contactTypesService.getTypeId(doc),
```

ChinHairSaintClair:
@jkeuster, the PR should now be ready for the telemetry additions and a bundle size bump.

jkuester:
Alright we are very close to getting this landed!  :tada: Here are the remaining tasks before I think we can consider this done:

- [ ] Update cht-conf to support the new `duplicate_check` property: https://github.com/medic/cht-conf/issues/675
    - @ChinHairSaintClair would you possibly be able to look into putting up a PR for this? Should be a very small change. :+1: 
- [ ] Update cht-docs to document the new feature and configuration options.
    - I am currently working on these changes and hope to raise a PR for review tomorrow.
- [ ] Create a forum thread to collect translation feedback/suggestions for our newly added strings.  I am assuming, @ChinHairSaintClair, that you did not source those 9 different translations all from native speakers in your organization :sweat_smile: (but maybe you actually did!?!).  If not, we can crowdsource this and get feedback from other folks in Medic and the broader community.  :+1: 
    - I will create this forum post and tag @ChinHairSaintClair so hopefully you can help support by making any suggested edits here in this PR :pray:

jkuester:
Here is the docs PR: https://github.com/medic/cht-docs/pull/1819
And here is the forum post regarding translations: https://forum.communityhealthtoolkit.org/t/translation-support-request-duplicate-contact-prevention-strings/4841

jkuester (approved):
Code is done! Tests are done! Docs are done! cht-conf updates are done!  We were only missing translations for two languages and those langs are not required for merging/releasing this PR (can add the translations later when they are available).  

So, I think I am going to hit the merge button on this and call it complete! :rocket: 

@ChinHairSaintClair thank you for your tireless efforts here to get this right! I am super proud of what we have built! (Also a big thank you to @fardarter for your support and assistance!)

mrjones-plip:
wow!  This PR was an amazing collaborative journey and a thing of real beauty to see in open source and to help the CHT.  Thanks all!!

Issue #8216, opened 2023-05-05
Propagate request uuid from API to audit logging layer


dianabarsan:
**What feature do you want to improve?**
When API receives a request from a client, we assign a unique ID to it, which gets logged along with some basic request / response information. 

**Describe the improvement you'd like**
Propagate this unique UUID to the audit layer (right now haproxy, but this is likely to change) as a header, and include it in the log. This should also apply for requests that compound and not directly proxied - for example an offline user request to changes ends up generating multiple requests to CouchDb, and these requests should have the same uuid as the original client request. 

**Describe alternatives you've considered**
None. 

**Additional context**
This is extremely helpful when analyzing logs. Without this, the correlation between the API incoming request and the cascade of couchdb requests is loose and sometimes impossible to determine.

PR #9613, opened 2024-11-05
feat(#8216): propagate api request id to haproxy

dianabarsan:
# Description

Adds custom header to pass the id of the original client request from API to all cascading requests to CouchDb. 
Changes request ID to being a 12 char long string (uuid slice) instead of the whole uuid. 

#8216

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8216-propagate-req-uuid-to-haproxy/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8216-propagate-req-uuid-to-haproxy/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:8216-propagate-req-uuid-to-haproxy/docker-compose/cht-couchdb-clustered.yml)

# License


api/src/routing.js
```diff
@@ -158,7 +159,8 @@ if (process.argv.slice(2).includes('--allow-cors')) {
 
 app.use((req, res, next) => {
   req.id = uuid.v4();
-  next();
+  req.headers[serverUtils.REQUEST_ID_HEADER] = req.id;
+  asyncLocalStorage.run({ clientRequest: req }, () => next());
```
garethbowen, line 163: Feels bloated to store the whole `req` here. Why not just the req.id?

garethbowen: I'd have to trace it through - it'd be nice if req could be GCed earlier but this may not be possible...

dianabarsan: I don't think it can get GC-ed before sending a response to the client, which is when the async local storage also "expires"

dianabarsan:
Since you've already been here, @garethbowen , mind a proper review? Thanks!

api/src/routing.js
```diff
@@ -158,7 +159,8 @@ if (process.argv.slice(2).includes('--allow-cors')) {
 
 app.use((req, res, next) => {
   req.id = uuid.v4();
-  next();
+  req.headers[serverUtils.REQUEST_ID_HEADER] = req.id;
```
garethbowen, line 162: Is this needed at all? Surely req.id is available everywhere req.headers is?

api/src/services/async-storage.js
```diff
@@ -0,0 +1,8 @@
+const { AsyncLocalStorage } = require('node:async_hooks');
+const asyncLocalStorage = new AsyncLocalStorage();
+const { REQUEST_ID_HEADER } = require('../server-utils');
+
+const request = require('@medic/couch-request');
+request.initialize(asyncLocalStorage, REQUEST_ID_HEADER);
+
+module.exports = asyncLocalStorage;
```
garethbowen, line 8: I think this is too leaky, which is particularly dangerous because most JS devs won't be expecting a thread local obj. Can we hide the implementation behind an API so it's more like...

```js
modules.exports = {
  getRequestId = () => {
    const localStorage = asyncLocalStorage.getStore();
    return localStorage?.clientRequest?.id;
  },
  setRequestId = (req, callback) => {
    asyncLocalStorage.run({ clientRequest: req }, () => callback());
  }
}
```

Maybe this doesn't work (I haven't tested), or maybe it ends up just as complicated?

garethbowen (changes requested):
TIL! That's cool. Java thread local flashbacks...

Anyway, some suggestions inline.

tests/integration/api/server.spec.js
```diff
@@ -230,4 +233,176 @@ describe('server', () => {
...
+        });
+        await utils.delayPromise(500); // wait for everything to get logged
+
+        const apiLogs = (await collectApiLogs()).filter(log => log.length);
+        const haproxyLogs = (await collectHaproxyLogs()).filter(log => log.length);
+
+        const reqUuid = getReqId(apiLogs[0]);
+
+        const haproxyRequests = haproxyLogs.filter(entry => getReqId(entry) === reqUuid);
+        console.warn(haproxyRequests);
```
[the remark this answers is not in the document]
dianabarsan: whoops!

tests/integration/api/server.spec.js
```diff
@@ -230,4 +233,176 @@ describe('server', () => {
...
+        expect(haproxyRequests.length).to.equal(2);
+        expect(haproxyRequests[0]).to.include('_session');
+        expect(haproxyRequests[1]).to.include(`/medic-test-user-${offlineUser.username}-meta/the_doc`);
+      });
+    });
+
+  });
```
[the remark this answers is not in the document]
dianabarsan: great suggestion!

api/src/routing.js
dianabarsan: This is needed because we might immediately proxy the request to couchdb with the router proxy thing. and those requests need the header as well.

tests/integration/api/server.spec.js
```diff
@@ -230,4 +233,176 @@ describe('server', () => {
...
+      }
+    };
+
+    describe('for online users', () => {
+      it('should propagate uuid via proxy', async () => {
+        const collectApiLogs = await utils.collectApiLogs(/\/_all_docs\?limit=1/);
+        const collectHaproxyLogs = await utils.collectHaproxyLogs(/\/_all_docs\?limit=1/);
+
+        const result = await utils.request('/medic/_all_docs?limit=1');
+        await utils.delayPromise(500); // wait for everything to get logged
```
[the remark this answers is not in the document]
dianabarsan: This will need to introduce some kind of passing what we expect and a timeout. 
It adds some complexity for half a second. Do you think it's justified?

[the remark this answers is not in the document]
garethbowen: Definitely. It's a pattern we use a lot so if we nail it then we've added some complexity in a util somewhere, but save 0.5s in many places. It doesn't have to be part of this issue - feel free to raise a new one and replace this pattern everywhere at once.

api/src/routing.js
```diff
@@ -158,7 +159,8 @@ if (process.argv.slice(2).includes('--allow-cors')) {
 
 app.use((req, res, next) => {
   req.id = uuid.v4();
```
garethbowen, line 161: A shower thought I had this morning... a UUID is IMO overkill for this. It doesn't matter if very occasionally there's a conflict, or even if we reuse IDs from week to week, right? The downside of such overkill is adding 16k to every log line, which we were already doing in API but now also doing in the audit log. Again maybe this is a separate issue but let's find a shorter alternative...

api/src/db.js
```diff
@@ -6,6 +6,8 @@ PouchDB.plugin(require('pouchdb-adapter-http'));
 PouchDB.plugin(require('pouchdb-session-authentication'));
 PouchDB.plugin(require('pouchdb-find'));
 PouchDB.plugin(require('pouchdb-mapreduce'));
+const asyncLocalStorage = require('./services/async-storage');
+const { REQUEST_ID_HEADER } = require('./server-utils');
```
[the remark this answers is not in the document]
dianabarsan: yea, I think so too.

api/src/routing.js
dianabarsan: I know, I thought about this as well. uuids are really convenient, we could just truncate the uuid to 1/4?

garethbowen: I was thinking hash it rather than truncate because it feels randomer but yeah, basically! There's packages to do this ( https://www.npmjs.com/package/short-uuid ) but if it's as simple as I expect then we can just do our own implementation.

garethbowen:
It would also be worth checking that whatever we land on here works with 3rd party tools to pull logs from multiple services, eg: https://grafana.com/docs/grafana/latest/explore/simplified-exploration/logs/ (released like a month ago!)

api/src/routing.js
dianabarsan: as far as I can tell, `short-uuid` just shortens the uuid to a specific length. I wouldn't bother hashing and adding a potentially complex operation.

tests/integration/api/server.spec.js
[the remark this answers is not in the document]
dianabarsan: I've created the issue. https://github.com/medic/cht-core/issues/9620
I think this is quite an intricate and possibly e2e test destabilizing. I'd rather have this PR clean and merged.

api/src/routing.js
dianabarsan: I added a 12 character limit to request ids.

dianabarsan:
> It would also be worth checking that whatever we land on here works with 3rd party tools to pull logs from multiple services, eg: https://grafana.com/docs/grafana/latest/explore/simplified-exploration/logs/ (released like a month ago!)

I think we already have a custom logging structure, we don't follow some specific standard.

api/src/db.js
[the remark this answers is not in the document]
dianabarsan: https://github.com/medic/cht-core/issues/9621

garethbowen:
> I think we already have a custom logging structure, we don't follow some specific standard.

I agree. IMO the killer feature is for our observability server to be able to show all things that happened across all services for a specific request. I'm assuming this change will make that work, but if you felt like testing loki to make sure that'd be valuable. It _should_ work with custom logging structures.

api/src/routing.js
garethbowen: Did you choose 12 because the 13th character is always a 4? Because that's what I would've done!

Anyway, nice work, quick win.

dianabarsan: hahaha

Actually, I chose 12 because docker containers have unique 12 char identifiers.

tests/integration/api/server.spec.js
```diff
@@ -230,4 +233,201 @@ describe('server', () => {
...
+          .map((log) => getReqId(log))[0];
+
+        const haproxyConfigReqs = haproxyLogs.filter(entry => getReqId(entry) === configReqId);
+        expect(haproxyConfigReqs.length).to.equal(3);
+
+        const haproxyHydrateReqs = haproxyLogs.filter(entry => getReqId(entry) === hydrateReqId);
+        expect(haproxyHydrateReqs.length).to.equal(2);
+
+        expect(hydrateReqId).not.to.equal(configReqId);
+      });
```
garethbowen, line 329: :heart:

garethbowen (approved):
Awesome! Nice one.

Issue #9636, opened 2024-11-11
Update to Node 22


garethbowen:
**Describe the issue**

Node 22 went LTS late October so we should look into updating.

PR #9638, opened 2024-11-12
chore(#9636): update to node 22

garethbowen:
# Description

#9636 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9636-node-22/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9636-node-22/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9636-node-22/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -3,7 +3,7 @@ FROM alpine:3.19 AS base_build
 RUN apk add --update --no-cache \
   build-base \
   curl \
-  nodejs~=20 \
+  nodejs~=22 \
   npm~=10 \
   tzdata \
   libxslt \
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.19 AS base_build
+FROM alpine:3.20 AS base_build
 
 RUN apk add --update --no-cache \
   build-base \
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -1,10 +1,8 @@
-FROM alpine:3.20 AS base_build
+FROM node:22-alpine AS base_build
 
 RUN apk add --update --no-cache \
   build-base \
   curl \
-  nodejs~=22 \
-  npm~=10 \
   tzdata \
   libxslt \
   bash \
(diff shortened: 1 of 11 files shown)


api/Dockerfile
```diff
@@ -1,10 +1,8 @@
-FROM alpine:3.19 AS base_build
+FROM node:22-alpine AS base_build
```
garethbowen, line 1: This feels like a significant change. alpine haven't yet added node 22 to their packages, but this official node alpine package works fine. What do you think?

dianabarsan (approved):
NICE! I think we're good to go. We can always change back to `alpine:xx` later if we think that's the best image choice for us.

api/Dockerfile
dianabarsan: Oh, also, I saw that there are two 22 node alpine images, one based on alpine:3.19 and one based on alpine:3.20 https://github.com/nodejs/docker-node/tree/main/22
Which one are we getting?

dianabarsan: The image is a bit more bloat (like auto adding yarn) in these images: https://github.com/nodejs/docker-node/blob/main/22/alpine3.20/Dockerfile

garethbowen: Yeah I couldn't figure out why Node was adding yarn. That seems crazy...

garethbowen: > Which one are we getting?

According to [the tags](https://hub.docker.com/_/node), `22-alpine` is the same as `22-alpine3.20`

garethbowen: > It looks like alpine releases are really rare, but if a critical security fix gets pushed, we might need it asap.

Yeah you're right. It's a crazy circular dependency - if there's a critical flaw in Node then we need a node image to get it asap, but if there's a critical flaw in something else in alpine then we need the alpine image to get it asap.

We can always switch back to alpine if there's a flaw especially if they've published node 22 package.

dianabarsan: agree.

PR #9646, opened 2024-11-14
chore(#8437): enable useUnknownInCatchVariables rule in tsconfig

jkuester:
# Description
https://github.com/medic/cht-core/issues/8437

# Code review checklist

# License

### DIFF
--- a/webapp/src/ts/modals/edit-user/update-password.component.ts
+++ b/webapp/src/ts/modals/edit-user/update-password.component.ts
@@ -8,6 +8,7 @@ import { UserSettingsService } from '@mm-services/user-settings.service';
 import { UpdatePasswordService } from '@mm-services/update-password.service';
 import { UserLoginService } from '@mm-services/user-login.service';
 import { TranslateService } from '@mm-services/translate.service';
+import { getProperty } from '../../libs/schema';
 
 const PASSWORD_MINIMUM_LENGTH = 8;
 const PASSWORD_MINIMUM_SCORE = 50;
@@ -64,7 +65,7 @@ export class UpdatePasswordComponent {
       try {
         await this.userLoginService.login(username, newPassword);
       } catch (err) {
-        if (err.status === 302) {
+        if (typeof err === 'object' && err !== null && 'status' in err && err.status === 302) {
           this.close();
           const snackText = await this.translateService.get('password.updated');
           this.globalActions.setSnackbarContent(snackText);
@@ -73,12 +74,13 @@ export class UpdatePasswordComponent {
         }
       }
     } catch (error) {
-      if (error.status === 0) { // Offline status
+      const status = getProperty(error, 'status');
+      if (status === 0) { // Offline status
         const message = await this.translateService.get('online.action.message');
         this.setError(ErrorType.SUBMIT, message);
         return;
       }
-      if (error.status === 401) {
+      if (status === 401) {
         const message = await this.translateService.get('password.incorrect');
         this.setError(ErrorType.CURRENT_PASSWORD, message);
         return;
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -14,7 +14,7 @@
   },
   "scripts": {
     "postinstall": "patch-package && ng cache clean",
-    "unit:mocha": "UNIT_TEST_ENV=1 mocha 'tests/mocha/**/*.spec.js'",
+    "unit:mocha": "UNIT_TEST_ENV=1 mocha 'tests/mocha/**/*.spec.js' && mocha --config tests/mocha/ts/.mocharc.js",
     "unit:mocha:tz": "TZ=Canada/Pacific npm run unit:mocha && TZ=Africa/Monrovia npm run unit:mocha && TZ=Pacific/Auckland npm run unit:mocha",
     "unit:cht-form": "ng test cht-form",
     "unit": "UNIT_TEST_ENV=1 ng test webapp",
--- a/webapp/src/ts/libs/schema.ts
+++ b/webapp/src/ts/libs/schema.ts
@@ -1,0 +1,10 @@
+export const hasProperty = <T extends string> (obj: unknown, prop: T): obj is Record<T, unknown> => {
+  return typeof obj === 'object' && obj !== null && prop in obj;
+};
+
+export const getProperty = <T extends string> (obj: unknown, prop: T): unknown => {
+  if (hasProperty(obj, prop)) {
+    return obj[prop];
+  }
+  return undefined;
+};
--- a/webapp/src/ts/modules/contacts/contacts-edit.component.ts
+++ b/webapp/src/ts/modules/contacts/contacts-edit.component.ts
@@ -13,7 +13,7 @@ import { Selectors } from '@mm-selectors/index';
 import { GlobalActions } from '@mm-actions/global';
 import { PerformanceService } from '@mm-services/performance.service';
 import { TranslateService } from '@mm-services/translate.service';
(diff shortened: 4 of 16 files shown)


webapp/src/ts/modals/edit-user/update-password.component.ts
```diff
@@ -73,12 +74,13 @@ export class UpdatePasswordComponent {
         }
       }
     } catch (error) {
-      if (error.status === 0) { // Offline status
+      const status = getProperty(error, 'status');
+      if (status === 0) { // Offline status
```
m5r, line 78: I'm not sold on this change yet... I agree we shouldn't assume anything thrown/caught is an `Error` object and the `unknown` type is best for this purpose because it forces us to do proper validation in the `catch` clause. But this change barely gets us halfway there: it checks that the property exists but it doesn't help with type-safety.
I don't think a generic lib file can solve this problem on its own. I think this sort of validation should be limited to its respective `catch` clause. For example here this could look like:

```ts
import { HttpErrorResponse } from '@angular/common/http';

try {
	// ...
} catch {
	if (error instanceof HttpErrorResponse) {
		const status = error.status;
//			^? number
	}
}

Comment (jkuester) [reply webapp/src/ts/modals/edit-user/update-password.component.ts:78]: :raised_hands: This is just the kind of push back I was hoping for! I am very interested in digging in deeper here to find the best solution.  

The challenge I see with doing something like `error instanceof HttpErrorResponse` in existing code is that we would have to be _absolutely certain of what is in scope for the try-block and what kind of errors might be thrown._  Taking this code as a good example, there is quite a bit of logic wrapped in the try-block and I am not confident that _anything that was being thrown with a `status` property,_ is definitely going to be `instanceof HttpErrorResponse`.  For example, even with a local Pouch db, if you call `get` with a docId that does not exist, it will throw an error with `status: 404`, but I do not think it will be a `HttpErrorResponse`...

This was the main reason I settled on this half-baked pattern matching approach.  :grimacing: 

> his change barely gets us halfway there: it checks that the property exists but it doesn't help with type-safety.

100% agree.  But, I really think that if we are going to do nested type-checking, we should revisit the idea of adding a schema validation library.... :grimacing: I have found `effect/schema` to be super powerful in chtoolbox and I think we could reduce a lot of manual logic that we have by pulling in something like that.  (FTR, not saying we go with `effect/schema` here in cht-core.... :sweat_smile:  Zod, seems like the default option, but I find https://github.com/fabian-hiller/valibot to be very interesting as well, particularly with its claims of superior tree-shaking....)

Comment (jkuester) [reply webapp/src/ts/modals/edit-user/update-password.component.ts:78]: Did a simple side-by-side comparison of `zod` vs `valibot` for evaluating the schema of this error.  The code is nearly identical:

`zod`:

```ts
      const MyErrorSchema = z.object({
        status: z.number().optional(),
      });
      const myError = MyErrorSchema.parse(error);
```

`valibot`:

```ts
      const MyErrorSchema = v.object({
        status: v.optional(v.number()),
      });
      const myError = v.parse(MyErrorSchema, error);
```

Then, I used `npm run build` to build the prod assets (with tree-shaking) and compared the sizes of `api/build/static/webapp`:

- Current: `8384K`
- Zod: `8444K` - `+60K`
- Valibot: `8388K` - `+4K`

This is inline with [valibot's claims](https://valibot.dev/guides/comparison/) to be significantly more efficient than Zod because the structure of the project is optimized for tree shaking.

IMHO `+4K` is worth it for sane schema validation.  In reality, we should be able to replace a lot of janky manual code checks with valibot.

jkuester:
@sugat009 would love your input on this PR.  I think I would rather not merge things as they are currently, but am very interested in continuing the discussion Mokhtar started regarding a schema validation library. This PR provides a really good example use-case for schema validation, but I would also be interested in making heavy use of validation in the cht-datasource code....

webapp/src/ts/modals/edit-user/update-password.component.ts
sugat009: I think Mokhtar does have a point in finding the object type of `err`. It is the general pattern in other programming languages as well. For eg: 
1. Java
```java
    try {  
        // Code that may throw an exception  
    } catch (IOException e) {  
        // Handle IOException  
    } catch (NumberFormatException e) {  
        // Handle NumberFormatException  
    } catch (Exception e) {  
        // Handle any other exceptions  
    }  
```
2. python
```python
try:
    n = 0
    res = 100 / n
    
except ZeroDivisionError:
    print("You can't divide by zero!")
    
except ValueError:
    print("Enter a valid number!")
    
else:
    print("Result is", res)
    
finally:
    print("Execution complete.")
```
I think we should apply these here as well and I agree that we probably won't know what kind of error the error object is. We would check for the most likely errors in the conditionals. If we are not sure what the error object would be a type of then, the `getProperty` function could be used in the default clause. For eg: when making a network call to an endpoint it's most likely not gonna throw a FileNotFound kind of error.

I'm not sure if zod and valibot are required here.

jkuester: > finding the object type of err. It is the general pattern in other programming languages as well.

One reason this is a general practice in other programming languages, though, is because those languages support things like checked exceptions and type-specific catch blocks (neither of which exist in JS).  (Effect has some [really cool features](https://effect.website/docs/error-management/expected-errors/) for handling this kind of thing, but that is not super helpful here.)

What we really need is some proper [duck type](https://en.wikipedia.org/wiki/Duck_typing) checking that would give us a clean way to say "do this if we have the kind of thing that has a `status` property"....

I am happy with checking the error class, particular when catching errors that we produce ourselves (like we did with the `InvalidArgumentError` in the api code), or in narrowly scoped places where we can be confident of exactly what to expect.  However, I think we will still end up with situations as I pointed out above where it seems risky to assume the the actual interface for the error.  Like you said, in those cases we need to fall back to some kind of duck-typing.

> I'm not sure if zod and valibot are required here.

Absolutly true that they are not _required_ to solve this problem.  My janky `getProperty` code is sufficient to make things "work".  The bigger question for me is if we should stop writing janky code like `getProperty` (and all the crazy core field checking code [in cht-datasource](https://github.com/medic/cht-core/blob/9673d78a406c8bbc56922e747a53388e5cf33320/shared-libs/cht-datasource/src/libs/core.ts#L93)) and instead just adopt a proper schema validation lib.  Maybe it is worth converting this PR to use `valibot` just to give us a feel for what that might look like?

sugat009: I've thought of this differently. We may not need to know the exception type unless we are doing something different for different types. Looking through the changes there is no special handling for exceptions and getting the properties from `err` object from `catch` might be sufficient for which the current implementation of `getProperty` is good enough. When there arises a case for special handling for different exceptions, we can consider `valibot`.

jkuester: > the current implementation of getProperty is good enough. When there arises a case for special handling for different exceptions, we can consider valibot.

This is fair.  I think this PR is not sufficient to provide a good enough rational for validot on its own.  That being said, I think there is good enough justification for valibot elsewhere in our codebase (specifically  in cht-datasource).  Instead of moving ahead with the hacky `getProperty` solution here, I am going to go ahead and move this PR back to Draft.  

Then, I plan to open a new PR where I leverage valibot functionality in cht-datasource in a meaningful way. That will provide a much better opportunity to weight the costs/benefits of a schema validation lib. Depending on the outcome of that PR, I can circle back here and update this branch accordingly....

Issue #9561, opened 2024-10-18
cht-form printing errors to console


jkuester:
**Describe the issue**
It appears that the `changes.service.ts` has leaked into the cht-form web component.  Now, when it tries to initialize, it fails and just spams the logs endlessly every 5 sec with 

```
Error initialising watching for db changes TypeError: Cannot read properties of undefined (reading 'update_seq')
    at eval (main.js:2)
    at H.invoke (polyfills.js:1)
    at Object.onInvoke (main.js:2)
    at H.invoke (polyfills.js:1)
    at ne.run (polyfills.js:1)
    at eval (polyfills.js:1)
    at H.invokeTask (polyfills.js:1)
    at Object.onInvokeTask (main.js:2)
    at H.invokeTask (polyfills.js:1)
    at ne.runTask (polyfills.js:1)
```

jkuester:
Looks like the culprit is the [extract-lineage.service](https://github.com/medic/cht-core/commit/1e8264a1d0d244749c688a6d4ba93ffc8257d5be#diff-8529f945fd15c5451df9dca706ac7121cd33e7cc5641951b63995126334cb5a7) which now depends on:

- UserSettingsService
- UserContactService
- AuthService

`EnektoService` > `ExtractLineageService` > `UserSettingsService` > `CacheService` > `ChangesService`.

PR #9647, opened 2024-11-14
chore(#9561): refactor services to fix cht-from error

jkuester:
# Description

Closes https://github.com/medic/cht-core/issues/9561

More details on source of the problem are in the issue, but TLDR is that a dependency on the `changes.service` leaked into cht-form. The `changes.service` tries to listen to the db for changes and prints an error if it cannot, before re-trying to listen again.  In cht-form, there is no db to listen to and the result is that when cht-form runs, the `changes.service` code just contantly spams error messages to the logs.  

My fix here is to update the service dependency tree to move some code out of the `extract-lineage.service` (which the `enketo.service` needs) and put it in the `user-contact.service`.  This essentially removes the dependency on the `changes.service` (and a number of other things not needed by `enketo.service`).   I think the `user-contact.service` is a reasonable place for this functionality anyway, but I am happy to hear other ideas for how to address this!

Additionally, I have added an after-hook to the cht-form integration tests to validate the browser console and check for any unexpected logs. Hopefully this will help catch issues like this in the future.

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9561_fix_cht-form_logs/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9561_fix_cht-form_logs/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9561_fix_cht-form_logs/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/tests/integration/cht-form/default/person-edit.wdio-spec.js
+++ b/tests/integration/cht-form/default/person-edit.wdio-spec.js
@@ -40,9 +40,6 @@ describe('cht-form web component - Edit Person Form', () => {
             created_by_place_uuid: ''
           }
         },
-        meta: {
-          instanceID: 'uuid:c558e232-0951-4ed8-8392-5ed8ac3c81b3'
-        }
       };
     });
 
--- a/tests/integration/cht-form/wdio.conf.js
+++ b/tests/integration/cht-form/wdio.conf.js
@@ -11,6 +11,17 @@ const logPath = path.join('tests', 'logs');
 const browserLogPath = path.join(logPath, 'browser.console.log');
 const mockConfig = require('./mock-config');
 
+const ALLOWED_LOG_MSGS = [
+  'favicon.ico - Failed to load resource: the server responded with a status of 404 (Not Found)',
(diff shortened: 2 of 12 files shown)


tests/integration/cht-form/default/person-edit.wdio-spec.js
```diff
@@ -40,9 +40,6 @@ describe('cht-form web component - Edit Person Form', () => {
             created_by_place_uuid: ''
           }
         },
-        meta: {
```
jkuester, line 43: I do not think this block should have been included here in the first place. It was causing a warning to be printed to the browser logs.

latin-panda:
Thanks @jkuester! I'll review this tomorrow. How can I reproduce the issue? Will any form submission in the `default` config do?

latin-panda (approved):
Sorry for the delayed on this review. LGTM! Thanks Josh!

Issue #9585, opened 2024-10-25
k3d tests frequently timeout


garethbowen:
**Describe the issue**

There are two suites that run with k3d instead of docker compose. These suites are copies of other suites so they run exactly the same tests with the only difference being how the CHT is being launched. This is better because it's more like a production environment and therefore more likely to find issues that actual projects would hit. Eventually the goal was to remove all docker compose versions to reduce the number of suites we execute.

However, the k3d suites frequently fail due to hitting the 60 minute GHA limit.

The original issue covering this migration is https://github.com/medic/cht-core/issues/8909

**Describe the improvement you'd like**

We need to find a way to stablise these suites so that we can trust the results of our CI runs. The ideal solution is to figure out why k3d is so much slower than docker compose and fix that. If it's not possible to speed the test up we could execute a minimal k3d suite, and retain docker compose suites for the bulk of the testing.

**Describe alternatives you've considered**

We could just increase the GHA timeout and that would probably work but then we would have to wait over an hour for each build to complete which is not acceptable.

dianabarsan:
From my experience, k3d is a lot slower when starting and stopping pods, additionally logs are logs when a pod is stopped, so to keep a log trail, we need to save logs for all pods everytime we stop a pod (because a crash in haproxy will also crash API). 

I think k3d generally requires more resources than docker on the tiny CI machine, so I do expect tests to be a little slower. 
I think the solution is to work out a way to change the tests so frequent or regular restarts are not required. 
We frequently stop sentinel just so we stop it from processing docs, for example, or forcing it to run some startup queue, instead of waiting for the regular 5 minute timer to elapse. 

One intermediary solution would be to skip all tests that require pod restarts in k3d and see how that improves stability, but ideally we would find a way to change these tests so that they don't need service restarts, or if we could somehow disconnect a service in some other way.

dianabarsan:
Discussing with @garethbowen, we've come up with a plan to:

- separate tests that genuinely require service restarts into their separate suite. this suite should run both over docker and k3d. 
- update service (api? sentinel?) to respond to process signals. instead of stopping and starting the service to achieve some behavior, setup hooks that run that code that are triggered by signals. add code in e2e tests that pass these signals to the containers/pods.

dianabarsan:
Added signals to sentinel to: 
- toggle transition processing on or off
- run scheduled tasks

Added a killswitch in mocha when API is offline in a beforeEach. 
[...]

PR #9658, opened 2024-11-19
chore(#9585): teach Sentinel sign language

dianabarsan:
# Description

# Code review checklist

# License

### DIFF
--- a/sentinel/src/lib/feed.js
+++ b/sentinel/src/lib/feed.js
@@ -15,6 +15,7 @@ const MAX_QUEUE_SIZE = 100;
 
 let request;
 let processed = 0;
+let processing;
 
 const enqueue = change => changeQueue.push(change);
 
@@ -100,33 +101,32 @@ changeQueue.drain(() => {
   resumeProcessing();
 });
 
+/**
+ * Start listening from the last processed seq. Will restart
+ * automatically on error.
+ */
 const listen = () => {
+  processing = true;
   changeQueue.resume();
   return resumeProcessing();
 };
 
+/**
+ * Stops listening for changes. Must be restarted manually
+ * by calling listen.
+ */
+const cancel =  () => {
+  processing = false;
+  changeQueue.pause();
+  if (request) {
+    request.cancel && request.cancel();
+    request = null;
+  }
+};
+
 module.exports = {
-
-  /**
-   * Start listening from the last processed seq. Will restart
-   * automatically on error.
-   */
   listen,
+  cancel,
 
-  /**
-   * Stops listening for changes. Must be restarted manually
-   * by calling listen.
-   */
-  cancel: () => {
-    changeQueue.pause();
-    if (request) {
-      request.cancel && request.cancel();
-      request = null;
-    }
-  },
-
-  // exposed for testing
-  _changeQueue: changeQueue,
-  _transitionsLib: transitionsLib,
-  _enqueue: enqueue,
+  toggle: () => processing ? cancel() : listen()
 };
--- a/sentinel/src/process-hooks.js
+++ b/sentinel/src/process-hooks.js
@@ -1,0 +1,13 @@
+const feed = require('./lib/feed');
+const schedule = require('./schedule');
+
+module.exports = {
+  init: () => {
+    process.on('SIGUSR1', () => {
+      feed.toggle();
+    });
+    process.on('SIGCONT', () => {
+      schedule.runTasks();
+    });
+  }
+};
--- a/tests/utils/index.js
+++ b/tests/utils/index.js
@@ -868,7 +868,7 @@ const listenForApi = async () => {
 
 const dockerComposeCmdExec = (params) => {
   const composeFiles = COMPOSE_FILES.map(file => ['-f', getTestComposeFilePath(file)]).flat();
-  params = ['docker compose', ...composeFiles, '-p', PROJECT_NAME].join(' ') + ' ' + params;
+  params = `docker compose ${composeFiles.join(' ')} -p ${PROJECT_NAME} ${params}`;
 
   return new Promise((resolve, reject) => {
     exec(params, { env }, (error, stdout, stderr) => {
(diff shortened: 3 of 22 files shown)


dianabarsan:
hey @garethbowen . Wanna do a PR review for a change? 

I've added signals to sentinel so we don't need to stop and start it. 
It doesn't entirely stabilize the k3d suites, because it doesn't address moments when API is not starting, but for that I've added one attempt to at least stop the process from hanging and instead quit early and hopefully we'll get more info from logs. 

Appreciate your time and your ideas :)

dianabarsan:
Also fat fingered @latin-panda . Sorry for the noise!

tests/utils/allure.js
```diff
@@ -0,0 +1,29 @@
+const allure = require('allure-commandline');
+
+const generateReport = () => {
+  const reportError = new Error('Could not generate Allure report');
+  const timeoutError = new Error('Timeout generating report');
```
garethbowen, line 5: Are you generating these upfront to capture the stacktrace or something? Otherwise it'd be better to do them on demand, but I can't remember if modern node carries the stack into the promise or not...

tests/utils/index.js
```diff
@@ -1179,29 +1169,18 @@ const startServices = async () => {
   }
 };
 
-const runCommand = (command, silent) => {
-  const [cmd, ...params] = command.split(' ');
+const runCommand = (command, verbose = true, overrideEnv = false) => {
```
garethbowen, line 1172: It's slightly easier to read to take an object here, so instead of passing `boolean, boolean` you can pass `{ verbose: boolean, overrideEnv: boolean }`.

tests/utils/index.js
```diff
@@ -1179,29 +1169,18 @@ const startServices = async () => {
...
+  verbose && console.warn(command);
```
garethbowen, line 1173: Why warn?

```suggestion
  verbose && console.log(command);
```

tests/utils/index.js
```diff
@@ -1562,6 +1553,9 @@ const isMinimumChromeVersion = process.env.CHROME_VERSION === MINIMUM_BROWSER_VE
 
 const escapeBranchName = (branch) => branch?.replace(/[/|_]/g, '-');
 
+const toggleSentinelTransitions = () => sendSignal('sentinel', 'USR1');
```
garethbowen, line 1556: I'm not sure about toggle. Wouldn't it be better to have an explicit start and an explicit stop? It would be very easy to accidentally call toggle twice (eg: test retry) which leaves sentinel in an unexpected state. This would be very hard to debug because the double call could be in one test, but the failure only show up many tests later. I think having two separate signals would make it explicit. Then we can decide if calling `startProcessing` when already processing would be ignored, logged, or throw an error because someone forgot to clean up.

sentinel/src/process-hooks.js
```diff
@@ -0,0 +1,13 @@
+const feed = require('./lib/feed');
+const schedule = require('./schedule');
+
+module.exports = {
+  init: () => {
+    process.on('SIGUSR1', () => {
```
garethbowen, line 6: I struggled to find any documentation on this but I think we can call the signals whatever we want, right? If so let's use clearer names. What about `SIGLISTEN`, `SIGSTOPLISTENING`, and `SIGRUNTASKS`? Or even `SIGSENTINELLISTEN` etc? It's getting long but at least it'll be clear that it's a custom signal and not a Node thing?

Also I think `SIGUSR1` is reserved and according to this documentation will trigger Node into debugging mode: https://nodejs.org/en/learn/getting-started/debugging - so if we can't rename as above, let's use a different number at least.

sentinel/src/lib/feed.js
```diff
@@ -100,33 +101,32 @@ changeQueue.drain(() => {
...
   return resumeProcessing();
 };
 
-module.exports = {
+/**
+ * Stops listening for changes. Must be restarted manually
+ * by calling listen.
+ */
+const cancel =  () => {
+  processing = false;
```
garethbowen, line 119: I think we should add logging here and in `listen` to make debugging easier.

garethbowen (changes requested):
Awesome work as always.

Looks like the successful builds are pretty quick, though you did have to re-run to get a green build. Any idea why they're still timing out?

tests/utils/allure.js
dianabarsan: I just moved some code that already existed, to have these run for the upgrade e2e test, which could not call the generic teardown because we have different containers.

sentinel/src/process-hooks.js
dianabarsan: Unfortunately, you can't send custom signals: 
```
➜  ~ kill -S SIGSIGSTOPLISTENING 26360
kill: unknown signal: SIGS
kill: type kill -l for a list of signals
```

So you need to select one of the signals in the list. There are some random ones that might not kill the process, but most do, and it felt weird to send a `SIGURG` to stop feed listening. 
This is why there's a toggle to stop/start processing, because we can't send custom signals and there are only two which are "user customizable" - SIGUSR1 and SIGUSR2, which I am using.

tests/utils/index.js
dianabarsan: As above, we can't send custom signals.

dianabarsan:
> though you did have to re-run to get a green build. Any idea why they're still timing out?

Yea. Sometimes API fails to start. No repro locally and because the test times out, we don't get an artifact for logs. 
I added a timeout to try to get these logs and figure out what's going on. But for now this stabilizes the Sentinel job.

sentinel/src/process-hooks.js
garethbowen: Ugh that's disgusting. Ok. The only other way I can see to do it would be to have a single signal which then looks somewhere else (file or db) to reload config, but that almost defeats the purpose!

Let's go with it for now.

garethbowen (approved):

Issue #9544, opened 2024-10-15
Make freetext search view indexes "offline only"


jkuester:
**Is your feature request related to a problem? Please describe.**
If we move to Nouveau for freetext searching on the Couch server (aka "online" use), then we still need a solution for freetext searching "offline" (what offline users would experience via their local Pouch instance).  

**Describe the solution you'd like**
The most straightforward approach should be to just have a design doc that is "client-only" similar to how we have docs that are currently "server-only".  This design doc will be indexed by PouchDB clients, but will not be indexed by the Couch server.  

Since we are likely going to require a complete re-indexing of these views on the client devices one way or the other, we should take this opportunity to remove the unused `key:value` emissions from the freetext views [as originally proposed](https://github.com/medic/cht-core/issues/9293) for the "freetext-lite" views. 

**Describe alternatives you've considered**
One interesting alternative would come [after our upgrade to Pouch `9`](https://github.com/medic/cht-core/issues/9208) and the `indexeddb` adapter.  IIRC, Mango queries should be more performance on Pouch with `indexeddb`. If that is the case, then we could evaluate possibly using Mango queries for freetext searching.  [This one blog post](https://reintech.io/blog/efficient-full-text-search-couchdb-mango-queries) suggests it works well... 😅

jkuester:
Here are my conclusions after taking a deep dive into possible implementations. The main approach options seem to be:

1. Add ddoc to Couch as normal but with `"autoupdate": false` so it does not get indexed on the server. Then in webapp we would need to warm the index when the app starts.
    - Pros:
      - Design doc is managed/controlled along with the rest. Gets all the metadata, etc.
    - Cons:
      - Run the risk of someone calling the view (even from Fauxton) and then it will be built on the server-side.
      - Have to manually warm the view on the client side (since it will not auto-update).
      - Need a bit of additional logic to allow replication of the new ddoc to the clients.
2. Add ddoc directly on the client side. Hardcode the ddoc directly in the webapp code and have it build the index on the client side.
    - Pros:
      - With the ddoc only on the client, there is no danger of it getting accidentally indexed on the server.
      - Might also simplify logic in `shared-libs/search` since that code could just check if the offline ddoc exists. If it does, use it, otherwise call nouveau.
    - Cons:
      - New code flow - not tracking ddoc with rest of the server-side ones.  View code will not all be in the same place in the repo.
      - Need custom logic in webapp to make sure the index is up-to-date when the app starts.
[...]

PR #9661, opened 2024-11-22
feat(#9544): add offline freetext search indexes

jkuester:
# Description

Closes https://github.com/medic/cht-core/issues/9544

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9544_offline_freetext/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9544_offline_freetext/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9544_offline_freetext/docker-compose/cht-couchdb-clustered.yml)

# License


webapp/src/js/bootstrapper/index.js
```diff
@@ -108,58 +109,50 @@
...
-          }
-          setUiError(err);
-          throw (err);
+    try {
+      const [isInitialReplicationNeeded] = await Promise
+        .all([
+          initialReplicationLib.isReplicationNeeded(localDb, userCtx),
+          swRegistration,
+          setReplicationId(POUCHDB_OPTIONS, localDb),
+          offlineDdocs.init(localDb)
```
jkuester, line 118: This is the only real change here. The rest is just a quick re-factor to use async/await.

shared-libs/search/src/generate-search-requests.js
jkuester, line 1: These changes in `shared-libs/search` are just temporary. The proper changes will need to be made against the cht-datasource code once we re-base on top of that.  

This is why I have not added any additional unit tests to cover this logic (or worried too much about code structure).

webapp/tests/mocha/unit/views/contacts_by_freetext.spec.js
```diff
@@ -1,95 +1,214 @@
...
+const medicOfflineFreetext = require('../../../../src/js/bootstrapper/offline-ddocs/medic-offline-freetext');
+const { expect } = require('chai');
+
+const expectedValue = (
+  {typeIndex, name, dead = false, muted = false } = {}
+) => `${dead} ${muted} ${typeIndex} ${name}`;
+
+describe('contacts_by_freetext', () => {
+  [
+    ['online view', loadView('medic-db', 'medic-client', 'contacts_by_freetext')],
```
jkuester, line 11: We will need to remove this once the `medic-client/contacts_by_freetext` view is deleted, but for now I thought it was useful to keep this in here to prove that the new offline indexes behave exactly the same as the original ones.

jkuester:
c.c. @m5r  and @tatilepizs Particularly regarding how I have updated the e2e search tests to run both for an online and an offline user.  I think this means these same tests can cover the offline indexes and the new Lucene functionality. :+1:

webapp/src/js/bootstrapper/index.js
jkuester, line 1: This bootstrapper code gets run [during the main initialization](https://github.com/medic/cht-core/blob/d2593c4f80a98d4a816afd87278d247d2ee9d0f4/webapp/src/ts/main.ts#L48) of the app (before any of the Angular code spins up.  This bootstrapping gets run during initial login and also when reloading the app (e.g. when the webapp-code/app-settings/ddocs change).  

So, any time in the future when we make updates/additions/removals of views here, we can be confident that everything will be initialized properly in the local db when the app reloads.

m5r, line 1: Thinking out loud here, do you think we could (should?) warm up the freetext views during bootstrap?

m5r (approved):
Nothing to add to Sugat's comments. I left an open-ended question about warming up views in the bootstrapper but it's not a blocker

webapp/src/js/bootstrapper/index.js
```diff
@@ -83,8 +93,18 @@
     return hasRole(userCtx, '_admin') || hasRole(userCtx, ONLINE_ROLE);
   };
 
+  const doInitialReplication = async (remoteDb, localDb, userCtx) => {
+    const replicationStarted = performance.now();
+    // Polling the document count from the db.
+    await initialReplicationLib.replicate(remoteDb, localDb);
+    if (await initialReplicationLib.isReplicationNeeded(localDb, userCtx)) {
```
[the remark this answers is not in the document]
jkuester: Technically yes.  I think it would be more clear to say that we always want to try to replicate when bootstrapping (just to be sure the app has the latest data from the server).  Then we have this extra call to `isReplicationNeeded` here to verify the replication we just did was actually successful.  (If everything replicated properly, then `isReplicationNeeded` should be `false`.)

tests/e2e/default/contacts/search-contacts.wdio-spec.js
```diff
@@ -5,63 +5,101 @@ const contactPage = require('@page-objects/default/contacts/contacts.wdio.page')
...
+    name: 'Sittu Health Center',
+    type: 'health_center',
+    parent: { _id: districtHospitalId, parent: { _id: '' } }
   });
 
-  const potuHospital = placeFactory.place().build({
-    name: 'Potu Hospital',
-    type: 'district_hospital',
-    parent: { _id: '', parent: { _id: '' } }
+  const potuHealthCenter = placeFactory.place().build({
```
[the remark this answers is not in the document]
jkuester: :sweat_smile:   I have no idea!  Just copying what was here before....

webapp/src/js/bootstrapper/index.js
jkuester: We definitely cannot block the bootstrapping process to wait for the warming to finish since the whole app will be waiting for this to finish.  That being said, it _should_ be possible to trigger some async warming (maybe even from within the app code itself instead of here in the bootstrap).  

Since there is no direct dependency on this functionality, I am going to log this as a different ticket that we can work on separately (if we decide we should go that direction).  https://github.com/medic/cht-core/issues/9728

webapp/src/js/bootstrapper/index.js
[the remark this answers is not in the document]
sugat009: ahh, I see. It's a check if the first one succeeded.

sugat009 (approved):

jkuester:
For clarity, this PR is now is now blocked waiting on https://github.com/medic/cht-core/pull/9541.  They should both be merged together before going to `master`.  We cannot ship this one earlier because it would involve adding new offline indexes without actually removing the existing indexes that exist both online and offline.  The effect would be to double the amount of space used on offline devices by the freetext indexes....

PR #9668, opened 2024-11-26
feat: add couch configs to enable proxy authentication

rmayore:
# Description

- Simple couchdb configuration update to enable [proxy authentication ](https://docs.couchdb.org/en/stable/api/server/authn.html#proxy-authentication)

## Building couchdb image with proxy auth enabled
1. Clone this fork, and pull from this branch
2. `npm ci`
3. `npm run build-dev`
4. Build images locally with the command  `npm run local-images`
5. `docker image ls` to get the new, timestamped, locally build couchdb image
6. Start your couchdb container using the new image. If you're running your couch using the [docker compose](https://docs.communityhealthtoolkit.org/contribute/code/core/dev-environment/#couchdb-setup-in-cht-4x) template, replace the image name with the local one in the `docker-compose.yml` file

## Restarting existing couchdb container with proxy auth enabled (thanks @mrjones-plip)
1. get to the couchdb shell with `docker exec -it cht-docker-couchdb-1 bash`
2. edit the config in `vi /opt/couchdb/etc/default.d/10-docker-default.ini`. You might have to install your favorite editor if it's not present in the container
3. Have your config changes take effect by restarting. do this by exiting the container shell and running `COUCHDB_USER=medic COUCHDB_PASSWORD=password docker compose -f docker-compose.yml -f couchdb-override.yml restart`

## Generating proxy auth token
Use the secret to generate a HMAC of the username... this will become our proxy token.
`echo -n "username" | openssl dgst -sha256 -hmac "the_secret"
`

# Code review checklist

# License

### DIFF
--- a/couchdb/10-docker-default.ini
+++ b/couchdb/10-docker-default.ini
@@ -18,6 +18,15 @@ bind_address = 0.0.0.0
 server_options = [{recbuf, 262144}]
 socket_options = [{sndbuf, 262144}, {nodelay, true}]
 require_valid_user = true
+authentication_handlers = {chttpd_auth, cookie_authentication_handler}, {chttpd_auth, proxy_authentication_handler}, {chttpd_auth, default_authentication_handler}
+
+[chttpd_auth]
+require_valid_user = true
+secret = 5ec56423-0856-4b13-8ab7-9b476fea2c39 ; CHANGE ME  - proxy secret
+x_auth_roles = X-Auth-CouchDB-Roles ; http header name for passing the comma separated couch user roles
+x_auth_username = X-Auth-CouchDB-UserName ; http header name for passing the couch username
+proxy_use_secret = true
+x_auth_token = X-Auth-CouchDB-Token ; http header name for passing the proxy token
 
 [httpd]
 secure_rewrites = false


couchdb/10-docker-default.ini
```diff
@@ -18,6 +18,15 @@ bind_address = 0.0.0.0
 server_options = [{recbuf, 262144}]
 socket_options = [{sndbuf, 262144}, {nodelay, true}]
 require_valid_user = true
+authentication_handlers = {chttpd_auth, cookie_authentication_handler}, {chttpd_auth, proxy_authentication_handler}, {chttpd_auth, default_authentication_handler}
+
+[chttpd_auth]
+require_valid_user = true
```
jkuester, line 24: I don't think we need to set this here. It is already set on line `20` (and I think it belongs in the `chttpd` section antyway).
```suggestion
```

couchdb/10-docker-default.ini
```diff
@@ -18,6 +18,15 @@ bind_address = 0.0.0.0
...
+secret = 5ec56423-0856-4b13-8ab7-9b476fea2c39 ; CHANGE ME  - proxy secret
```
jkuester, line 25: I also do not think we need to hard-code the `secret` here.  Unless I am mistaken, the [COUCHDB_SECRET envar](https://docs.communityhealthtoolkit.org/hosting/4.x/production/docker/#prepare-environment-variables-file) you can supply to the couch container [gets set as the `httpd_auth.secret` value](https://github.com/medic/cht-core/blob/987947839f1a01538c7abecbfb9ca5b410a60e41/couchdb/docker-entrypoint.sh#L39). 

```suggestion
```

couchdb/10-docker-default.ini
```diff
@@ -18,6 +18,15 @@ bind_address = 0.0.0.0
...
+x_auth_roles = X-Auth-CouchDB-Roles ; http header name for passing the comma separated couch user roles
```
jkuester, line 26: It looks like you just kept the default values for these header names. As far as I know, the default values should work fine for our case, but do we even need to include this config here if we just use the default values?

jkuester (changes requested):
@rmayore this is great!  I pulled down the changes and was able to get everything running locally.  Thank you for the detailed instructions!

I just left some minor comments inline. :+1: 

One question  I am interested in getting everyones' thoughts on is if we think it would be a problem to have the `proxy_authentication_handler` always enabled (even when a deployment is not using SSO)?  It seems like you can enable it without any disruption to the `default/cookie_authentication_handler`s, so probably the only concern would be around security.  The proxy auth could theoretically increase the attack surface since it allows admin access to anyone with the secret and direct access to the Couch instance.  However, I am not sure this is a major concern.  The Couch secret is already carefully managed with the other sensitive values such as the admin user/pass.  Additionally, on a proper production deployment, the Couch ports should not be accessible. Instead all traffic should be proxyed through the api server.  Thoughts?

jkuester:
Closing this PR since we have determined not to go this direction for SSO.  The code changes will still be available out here to reference in the future if we want to return to this idea of using Couch Proxy Auth.

Issue #9704, opened 2024-12-11
Crash when loading tasks/targets: Cannot read properties of undefined (reading '_rev')


kennsippell:
**Describe the bug**
Crash

**To Reproduce**
Crash is in `J=>!(!q && J._rev`

```
                commitTargetDoc: (B,U,{userContactDoc: H, userSettingsDoc: G},q)=>{
                    const z = null == H ? void 0 : H._id
                      , Y = null == G ? void 0 : G._id
                      , K = `target~${U}~${z}~${Y}`;
                    return k.get(K).catch(J=>{
                        if (404 === J.status)
                            return {
                                _id: K,
                                type: "target",
                                user: Y,
                                owner: z,
                                reporting_period: U
                            }
                    }
                    ).then(J=>!(!q && J._rev) && (J.targets = B,
                    J.updated_date = y().startOf("day").valueOf(),
                    k.put(J)))
                }
```

**Screenshots**
![Image](https://github.com/user-attachments/assets/a2541ca5-0569-4389-b5bc-72973f816165)

**Environment**
- Instance: Nairobi
- App: webapp
- Version: 4.13

**Additional context**

dianabarsan:
It looks like this is a minor issue here: https://github.com/medic/cht-core/blob/master/shared-libs/rules-engine/src/pouchdb-provider.js#L76 where we catch and handle 404 errors, but don't explicitly throw on other errors. 
What we are seeing here is a symptom of another pouchdb crash. 
Since the username is obfuscated, can someone who has access to the unedited image check this user's feedback docs and find the underlying PouchDb error? 

I will make the change to explicitly throw non 404 errors, so the next time this happens we will have visibility into the underlying error.

kennsippell:
The user doesn't appear to have any telemetry docs in 2024. 

```
curl -X POST https://nairobi.echis.go.ke/medic-users-meta/_find -H 'Content-Type: application/json' -d '{ "selector": { "metadata.user": "mjanethiah", "type": "telemetry" } }'
```

All tabs are throwing errors ("Error fetching people" on contacts tab), so I'm guessing `IDBDatabase` connection closing. I'll look deeper with a SQL query.

kennsippell:
Sorry `_find` just has a default limit of 25.  Confirmed this is a global database failure.

```
        {
          "level": "error",
[...]

PR #9705, opened 2024-12-12
fix(#9704): throw pouch errors when saving target docs

dianabarsan:
# Description

#9704

# Code review checklist

# License


shared-libs/rules-engine/src/pouchdb-provider.js
```diff
@@ -82,6 +81,7 @@ const medicPouchProvider = db => {
               reporting_period: docTag,
             };
           }
+          throw err;
```
dianabarsan, line 84: Only real change. Everything else is appeasing sonar complaints for code depth, and fixing them by switching functions to use async/await.

jkuester (approved):
:+1: LGTM!

Issue #9691, opened 2024-12-04
Plug Nouveau APIs with API lifecycle


m5r:
**What feature do you want to improve?**

**Describe the improvement you'd like**
- warm index and track indexing progress (maybe?) in https://github.com/medic/cht-core/blob/couchdb-nouveau/api/src/services/setup/upgrade-steps.js#L100
- clean up in https://github.com/medic/cht-core/blob/couchdb-nouveau/api/src/services/setup/utils.js#L55

**Additional context**
Track indexing progress: [`GET /_active_tasks`](https://docs.couchdb.org/en/latest/api/server/common.html#get--_active_tasks)

Example response with nouveau indexing in progress:
```
[
  {
    "node": "couchdb@127.0.0.1",
    "pid": "<0.22282.4>",
    "process_status": "running",
    "changes_done": 915433,
    "database": "shards/20000000-3fffffff/medic.1559595003",
    "design_document": "_design/medic-nouveau",
    "index": "contacts_by_freetext",
    "progress": 92,
    "started_on": 1732718731,
    "total_changes": 984938,
    "type": "search_indexer",
    "updated_on": 1732719464
  },
  {
    "node": "couchdb@127.0.0.1",
    "pid": "<0.22283.4>",
    "process_status": "waiting",
    "changes_done": 864788,
    "database": "shards/20000000-3fffffff/medic.1559595003",
    "design_document": "_design/medic-nouveau",
    "index": "reports_by_freetext",
    "progress": 87,
    "started_on": 1732718731,
    "total_changes": 984938,
    "type": "search_indexer",
    "updated_on": 1732719464
  },
  {
    "node": "couchdb@127.0.0.1",
    "pid": "<0.22358.4>",
    "process_status": "waiting",
    "changes_done": 914830,
    "database": "shards/00000000-1fffffff/medic.1  559595003",
    "design_document": "_design/medic-nouveau",
    "index": "contacts_by_freetext",
    "progress": 92,
    "started_on": 1732718731,
    "total_changes": 987896,
    "type": "search_indexer",
    "updated_on": 1732719464
  },
  {
    "node": "couchdb@127.0.0.1",
    "pid": "<0.22359.4>",
    "process_status": "waiting",
    "changes_done": 865463,
    "database": "shards/00000000-1fffffff/medic.1559595003",
    "design_document": "_design/medic-nouveau",
    "index": "reports_by_freetext",
    "progress": 87,
    "started_on": 1732718731,
    "total_changes": 987896,
[...]

PR #9717, opened 2024-12-17
feat(#9691): plug nouveau APIs with API lifecycle

m5r:
# Description

#9691 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9691-nouveau-lifecycle/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9691-nouveau-lifecycle/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9691-nouveau-lifecycle/docker-compose/cht-couchdb-clustered.yml)

# License


api/src/services/setup/view-indexer.js
```diff
@@ -60,23 +82,27 @@ const getViewsToIndex = async () => {
...
+ * Returns a promise that resolves when a nouveau index is indexed.
+ * Retries querying the index until no error is thrown
+ * @param {String} dbName
+ * @param {String} ddocId
+ * @param {String} indexName
+ * @return {Promise}
+ */
+const indexNouveauIndex = async (dbName, ddocId, indexName) => {
+  return await waitForRequest({
+    uri: `${environment.serverUrl}/${dbName}/_design/${ddocId}/_nouveau/${indexName}`,
```
jkuester, line 102: issue: the `_design` part of this URL is already included in the ddocid!

```suggestion
    uri: `${environment.serverUrl}/${dbName}/${ddocId}/_nouveau/${indexName}`,
```

jkuester (approved):
Very nice!  1 small issue with the URL, but otherwise this is good to go!

api/tests/mocha/services/setup/utils.spec.js
```diff
@@ -155,6 +156,10 @@ describe('Setup utils', () => {
   });
 
   describe('cleanup', () => {
+    beforeEach(() => {
+      sinon.stub(db, 'nouveauCleanup').returns(new Promise(() => {}));
```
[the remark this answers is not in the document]
witash: added a test for nouveauCleanup and assertions to the `should catch errors` tests that other compact and cleanup functions are called if one fails, and that the error is logged

api/src/services/setup/view-indexer.js
witash: So this means I wasn't testing the indexing properly...the thing is that nouveau indexes are built any time a design document is created or updated, so the index will be built and "warmed" even if nothing queries the endpoint. Which is what I was testing for.

Actually the point of querying the endpoint in upgrade-steps is to make sure that upgrade **waits** until that is actually done before swapping the staged and live ddoc.

Fixed the url and re-tested properly.

jkuester: Yeah, I think the [ddoc versioning issue](https://github.com/medic/cht-core/issues/9691#issuecomment-2790367531) was making this super tricky to test.  I actually only caught this because I noticed the api unexpectedly crashed the first time I switched to your branch (with a `404` error).  But, that error only happened for the _first_ time I ran the api server. If I restated it after the crash it would run fine (since the ddoc version in Couch was matching).

api/tests/mocha/services/setup/utils.spec.js
```diff
@@ -175,19 +181,52 @@ describe('Setup utils', () => {
...
 
-    it('should catch errors', async () => {
-      mockDb(db.medic);
-      mockDb(db.sentinel);
-      mockDb(db.medicLogs);
-      mockDb(db.medicUsersMeta);
-      mockDb(db.users);
+    it('should catch compact errors and log them', async () => {
+      const error = { some: 'error' };
+      db.sentinel.compact.returns(Promise.reject(error));
```
sugat009, line 190:
```suggestion
      db.sentinel.compact.rejects(error);
```

api/tests/mocha/services/setup/utils.spec.js
```diff
@@ -175,19 +181,52 @@ describe('Setup utils', () => {
...
+      expect(db.users.viewCleanup.callCount).to.equal(1);
+
+      expect(logger.error.callCount).to.be.at.least(1);
+      expect(logger.error.args[0][0]).to.include('Error while running cleanup');
+      expect(logger.error.args[0][1]).to.deep.equal(error);
+    });
+
+    it('should catch nouveau cleanup errors and log them', async () => {
+      const error = { some: 'error' };
+      db.nouveauCleanup.returns(Promise.reject(error));
```
sugat009, line 219:
```suggestion
      db.nouveauCleanup.rejects(error);
```

sugat009 (approved):
LGTM! A minor nitpick.

api/tests/mocha/services/setup/utils.spec.js
witash: I think it has to return a promise that will reject, instead of rejecting directly.

db.sentinel.compact is async so while it could throw an error directly, what this test is really testing for is when there's some problem compacting the db, which will return a Promise that then throws the actual error.

witash: I guess the difference is that the cleanup function is using `Promise.all` instead of `await`? If it was using `await` mocha would mock the function throwing an error directly, but because it's called with Promise.all that doesn't test what its supposed to, and it has to return a Promise that will reject instead.

sugat009: They are the same thing,  `.rejects` is the shorthand. It rejects the `error` object as a promise.
https://sinonjs.org/releases/v20/stubs/. _Search for `rejects`. There is no hyperlink._

witash: Ok thanks, I misunderstood `.rejects`

witash:
unless anyone objects going to merge this even with some tests failing because it has been open for so long; the failing tests are either k3d or don't seem to have anything to do with this ticket specifically.

Issue #9714, opened 2024-12-17
Targets report 0/0 for some users after interval turnover


kennsippell:
**Describe the bug**
I'm struggling a bit to pin this down to a specific repro, but I think the evidence below indicates quite strongly a problem in `handleIntervalTurnover`.  

**To Reproduce**
[November target document](https://nairobi-echis.health.go.ke/medic/target~2024-11~992120f8-d1b9-54f7-a422-f6ce7ccd5820~org.couchdb.user:crebeccahonii)

* All targets report a value of 0
* This target doc has 121 revs, so presumably it wasn't 0 all month. Trying to verify [via audit log](https://forum.communityhealthtoolkit.org/t/understanding-the-history-a-document-through-haproxy-audit-log/4471) but stuck.
* If I login as this user in production and set the date to November, I see 95 household visits and 14 referrals in the targets tab. I can also see these values in the target doc during the month of November.
* Last write is Dec 2 midnight EAT (interval turnover on Dec 2)

[Nov 30th telemetry](https://nairobi.echis.go.ke/medic-users-meta/telemetry-2024-11-30-crebeccahonii-5f0441c4-13db-475e-af24-9ecfd26289e8) indicates:

* User loaded the targets tab successfully! 
* No feedback documents reported Nov 30th +/- many days
* They are running CHT Core 4.13
* No reports completed or contacts created. They seem to just login and sync (which is what they are told to do by Nairobi coordinators to ensure end-of-month reporting)

[Dec 2nd telemtry](https://nairobi.echis.go.ke/medic-users-meta/telemetry-2024-12-2-crebeccahonii-5f0441c4-13db-475e-af24-9ecfd26289e8)

* No activity on Dec 1. 
* They load the targets tab like 10 times
* They completed about 32 reports on this single day
* Same DeviceID as Nov 30th telemetry

**Expected behavior**
Target doc should not report values with all zeros when target tab is showing valid sums

**Environment**
- Instance: nairobi.echis.go.ke
- App: webapp
- Version: 4.13.0

**Additional context**
About 7% of eCHIS Nairobi users are reporting that the values on their Nov target documents are incorrect during the first week of December. They are running the fix for https://github.com/medic/cht-core/issues/9431. This is causing users to continue to not trust the performance data presented by the eCHIS digital payment system.

Issue #6209, opened 2020-01-20
Improve target doc data fidelity - Close gap between last write and end of the reporting interval


kennsippell:
**What feature do you want to improve?**
Webapp v3.8 included a Rules Engine v2 which resulted in documents with type `target` to be written to PouchDB. This document updates a maximum of once per day. An update for the current reporting interval's target document is triggered by the user viewing the tasks/analytics tab, or by a timer which triggers a minute after the app loads. This was a "best-effort" approach to reporting which conservatively prevents writes to the document.

A few scenarios may lead to the target document of a reporting interval to not include a fully accurate summary of the user's target data. For example, if the user navigates to the task/target tab on the day prior to the end of a reporting interval, then creates/edits reports, then views the task/target tab again that day. The document only updates once per day, so it will not recieve the latest updates. The next day is the next reporting period, so the out of date target doc won't be updated. Other scenarios are possible if the user doesn't view the tasks/targets page for several days prior to the end of the reporting period.

**Describe the improvement you'd like**
I'd like to update the target document for the `n`th reporting interval once in the `n+1`th reporting interval.

**Additional context**
This is an existing problem in 3.8 that will impact the fidelity of our impact metrics. I'm scheduling this in 3.9, because Target documents will also be used to drive our DHIS2 workflows. #6002

kennsippell:
I attempted a fix for this measurement gap using the method suggested above.

> I'd like to update the target document for the nth reporting interval once in the n+1th reporting interval.

However, this approach is perhaps not viable since it isn't currently safe to aggregate targets for a time interval which is not the current interval. This works fine for target emissions which expicitly set their date attribute, but doesn't work with target emissions which leave the date undefined or which set it to Date.now() (which is the default for declarative configuration).

While investigating, I've discovered that this same issue actually presents in a critical scenario. See #6267. Check out the PR which illustrates the problem a bit more clearly (see the failing test). The resolution for this may be possible depending on how we handle #6267.

dianabarsan:
Ready for AT on `6209-recalculate-targets`. 
Recalculation of the old target doc should happen when the reporting interval changes, either if it's while you're logged out or while you're using the app. In order to trigger the target recalculation, you can visit either target or tasks pages.

newtewt:
[...]

PR #9718, opened 2024-12-18
feat(#9714): remove rules-engine's interval turnover

kennsippell:
# Description

I'm proposing this change based on a) the evidence from #9714 which demonstrates that interval turnover is likely causing incorrect values in target docs, and b) [this comment](https://github.com/medic/cht-core/pull/9486#discussion_r1802337684) from #9486 which suggests interval turnover is no longer needed.  The linked comment also suggests this is maybe not the correct approach, so I'm open to all suggestions.

Removing interval turnover reopens the hole documented via #6209 but it should be significantly smaller after #9486.

#9714 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9714-no-interval-turnover/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9714-no-interval-turnover/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9714-no-interval-turnover/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/shared-libs/rules-engine/src/provider-wireup.js
+++ b/shared-libs/rules-engine/src/provider-wireup.js
@@ -12,7 +12,6 @@ const refreshRulesEmissions = require('./refresh-rules-emissions');
 const rulesEmitter = require('./rules-emitter');
 const rulesStateStore = require('./rules-state-store');
 const updateTemporalStates = require('./update-temporal-states');
-const calendarInterval = require('@medic/calendar-interval');
 
 let wireupOptions;
 
@@ -31,7 +30,7 @@ module.exports = {
    * @param {number} settings.monthStartDate reporting interval start date
    * @param {Object} userDoc User's hydrated contact document
    */
-  initialize: (provider, settings) => {
+  initialize: async (provider, settings) => {
     const isEnabled = rulesEmitter.initialize(settings);
     if (!isEnabled) {
       return Promise.resolve();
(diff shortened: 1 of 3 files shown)


dianabarsan:
I think interval turnover is still needed for project that upgrade from a version older than 4.13 to 4.13, in case some users end up upgrading _after_ the previous month ended. 
Otherwise I think interval turnover is safe to be removed.

kennsippell:
I'm planning to deploy this in Nairobi and see if the target doc data quality improves in January. I've created FR branch `4-13-FR-no-interval-turnover` for this. I'd appreciate a review prior to deploying to confirm I haven't missed anything.

These are the options I see to move this forward in main:

1. Remove interval turnover completely. One-time reduction in accuracy of target docs for users who sync the core upgrade near turnover.
2. Move interval turnover to be a public interface of RulesEngine. Run it once after upgrade from <4.13. 
3. Keep IntervalTurnover and continue deeper down the path to hammer out the problems.

My inclination is toward Option 1 because:

1. The data in target docs is simply untrustworthy today. We are kinda breaking something that is already broken.
1. Seems somewhat rare that a user would sync the change in a manner where they would be impacted.
2. Option 2 requires maintenance forever.
3. I'd discuss Option 3 further but probably not until we experiment with this change in January and confirm there is still a significant unaddress measurement gap.

How would you proceed?

dianabarsan (approved):
I agree with option 1, if we're not worried about projects potentially missing one recalculation after upgrade. 

I hope this fixes the data quality, because I not entirely convinced it will. Interval turnover would have saved a snapshot of the last calculation, which now _should_ be accurate. Wrong targets means that the snapshot was wrong already :|

kennsippell:
Reviewing data from Nairobi, I pulled the # of CHPs who disagreed with their target data. Then, their supervisor reviewed their monthly performance and agreed that the data was significantly wrong. 

Before this change (November Targets) it was 462 CHPs (6.7%). After (December targets), it was 215 CHPs (3.5%).
Of these, 275 CHPs had all 0s before the change. vs 68 CHPs having all 0s after the change.

So this does seem to result in a significant improvement.

PR #9731, opened 2025-01-08
feat(#9547): add password change feature on first time login

### LINKED ISSUE #9547: Change password on first login

Benmuiruri:
# Description

WIP

Closes #9547 

# Code review checklist

# License


Benmuiruri:
Hi @jkuester I resolved the feedback you gave me and this implementation is leaner than the previous one. It is now ready for review.

tests/e2e/default/login/login-logout.wdio-spec.js
```diff
@@ -165,4 +167,99 @@ describe('Login page funcionality tests', () => {
       await (await commonPage.tabsSelector.messagesTab()).waitForDisplayed();
     });
   });
+
+  describe('Password Reset', () => {
```
jkuester, line 171: praise: very nice! It is great to have thorough e2e tests for this interaction!

jkuester (changes requested):
This is coming together nicely!  Couple additional comments/suggestions here.  

I think my only remaining workflow-level question is if you think we need some kind of message in the admin app when changing a user's password that will alert the admin that the user will be prompted to change their password again?  I am just concernd that we will still catch folks off guard with this functionality change...  🤔

api/resources/translations/messages-en.properties
```diff
@@ -654,8 +659,8 @@ enketo.geopicker.altitude = altitude (m)
 enketo.geopicker.closepolygon = close polygon
 enketo.geopicker.kmlcoords = KML coordinates
 enketo.geopicker.kmlpaste = paste KML coordinates here
-enketo.geopicker.latitude = latitude (x.y �)
-enketo.geopicker.longitude = longitude (x.y �)
+enketo.geopicker.latitude = latitude (x.y °)
+enketo.geopicker.longitude = longitude (x.y °)
```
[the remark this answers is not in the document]
Benmuiruri: No, I think what happened is there was a encoding mismatch somewhere, either Github or my IDE. That degree symbol looked different after I opened the pR (Without having edited those lines). 

This is what I was seeing 

Screenshot 2025-01-14 at 10 34 52


So I copied the lines from master.

shared-libs/user-management/test/unit/users.spec.js
```diff
@@ -1301,6 +1330,41 @@ describe('Users service', () => {
   });
 
   describe('createUser', () => {
+    it('should set password_change_required to true for new user creation', () => {
```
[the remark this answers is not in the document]
Benmuiruri: I would say let's keep it since the other two check place/contact relationship validation.

api/src/controllers/login.js
```diff
@@ -190,36 +225,46 @@ const setUserCtxCookie = (res, userCtx) => {
...
+  try {
+    const userCtx = await getUserCtxRetry(options);
+    if (roles.isDbAdmin(userCtx)) {
+      await users.createAdmin(userCtx);
+    }
+
+    const user = await users.getUserDoc(userCtx.name);
+    if (!skipPasswordChange(user)) {
+      return redirectToPasswordReset(req, res, userCtx);
+    }
```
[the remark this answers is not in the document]
Benmuiruri: Granted it is not possible to do password reset on admin users.

But, I would say leaving it outside the else block keeps the logic simpler and flatter (More readable)

Benmuiruri:
Hi @jkuester Good point about adding a sort of warning. I came up with this 

```
update.password.help = User will be prompted to reset password after login unless can_skip_password_change permission is enabled for the user role.
```

Screenshot 2025-01-14 at 19 21 46

jkuester:
@Benmuiruri Yes, that was exactly what I was thinking.  Would it be possible to just hide it for users that have the `can_skip_password_change` permission?  I would rather not mention the permission at all in the message and it seems it would be most helpful to just see it in the case where we know a password reset will be necessary.  But, I am not sure if we have everything we need in context here to know which permissions the user is going to have...

api/src/controllers/login.js
```diff
@@ -300,26 +345,112 @@ const renderLogin = (req) => {
...
+    return sendLoginErrorResponse(e, res);
+  }
+};
+
+const updatePassword = (user, newPassword) => {
+  const updateData = {
+    password: newPassword,
+    password_change_required: false,
+  };
+  return users.updateUser(user.name, updateData, true, null);
```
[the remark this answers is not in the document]
Benmuiruri: Without full access the password reset would not work. It has to do with what is allowed in `RESTRICTED_USER_EDITABLE_FIELDS` inside the user management library.

api/resources/translations/messages-en.properties
```diff
@@ -1272,6 +1280,7 @@ translation.add = Add new translation key
 translation.key = Translation key
 unique.id = Unique ID
 unknown.contact = Unknown contact
+update.password.help = User will be required to reset their password on their next login
```
Benmuiruri, line 1283: Hey folks could I please get a review of this translation. Feel free to add the translation in this comment. 

French - @andrablaj 
Nepali - @sugat009 
Spanish - @lorerod 

Thank you.

andrablaj: French: `L'utilisateur devra réinitialiser son mot de passe lors de sa prochaine connexion`

lorerod: Spanish: El usuario deberá restablecer su contraseña en su próximo inicio de sesión.

api/src/controllers/login.js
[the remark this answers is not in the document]
jkuester: :thinking: Okay, so I had spent a bunch of time staring at those arrays before leaving this message and was pretty sure `password` was allowed to be edited by the user.  (And after double-checking I confirmed this is true.  `ALLOWED_RESTRICTED_EDITABLE_FIELDS` contains `RESTRICTED_USER_EDITABLE_FIELDS` which contains `password`.)

_However,_ this problem is actually that we are also trying to edit the `password_change_required` field (which is not contained in `ALLOWED_RESTRICTED_EDITABLE_FIELDS`.  That is why you get an error if you do not pass `fullAccess`.  

We could add `password_change_required` to the user editable fields list. This might make sense since the user can edit their own password.  At the same time, since it was the system that flagged the user as needing a password reset, maybe only the "system" should be able to write to this field.  This is very much a fully "managed" workflow being orchestrated by the system, so I am fine just leaving it with `fullAccess = true`.

shared-libs/user-management/src/users.js
```diff
@@ -582,9 +561,14 @@ const missingFields = data => {
 const getUpdatedUserDoc = async (username, data, fullAccess) => {
   return getUserDoc(username, 'users')
     .then(async doc => {
-      return {
+      const baseUser = {
         ...doc,
-        ...(await getUserUpdates(username, data, fullAccess)),
+        name: username,
+        type: 'user',
+      };
```
jkuester, line 568: suggestion: I don't think we actually need this here.  `doc` will always have `name` and `type`.  I have tried manually creating docs in the `_users` db in Couch and it rejects anything without a `name` or when `type != 'user'`.

jkuester (approved):
Fantastic!  I had one minor suggestion about the update user code, but otherwise this is good to go!

shared-libs/user-management/src/users.js
Benmuiruri: That's true. In production it would not be necessary, but the tests were breaking because the mocks in the test for the doc were empty objects `db.users.get.resolves({})`. 

I've removed the baseUser which was setting the name and updated the mocks to be more realistic

```
db.users.get.resolves({
  name: 'paul',
  type: 'user'
});
```

Benmuiruri:
Hi @jkuester thanks for the follow up. 

I've pinged @PhilipNgari on the echis-ke [repo](https://github.com/moh-kenya/config-echis-2.0/issues/2655) to get clarification on merging it.

jkuester:
@Benmuiruri I think we should go ahead and merge this.  We have confirmed through broad and extensive discussion that this functionality is our intended path forward. Stakeholders have been made aware of the details of the changes. The `can_skip_password_change` permission provides instance administrators the ability to toggle off the behavior. This allows them the time/flexibility to update their administrative workflows as necessary.  As I [mentioned in the issue thread](https://github.com/medic/cht-core/issues/9547#issuecomment-2590572756) I think we all recognize this functionality as just being one piece to the whole user security puzzle, but we do not need to wait for all the pieces to be in place before releasing this one. (In fact, I think releasing this feature will make it easier to identify other features necessary to ensure user security.)

PhilipNgari:
We raised the same issue again last week and it was put on hold. Given the @jkuester  mention on the ability to "Toggle off" the functionality at the admin level, we can go ahead and merge but not deploy for eCHIS Kenya until the consumer Department at MoH gves a nod. I imagine its a similar concept to the new UI changes where they appear differently on both phone or laptop but can be toggled on or off. 

We just need to be sure that it's harmonious across the instances so that this does not skip and deploy in some. cc @alexosugo

andrablaj:
@Benmuiruri Per @PhilipNgari's comment above, could you please solve the conflicts and merge the PR, so this can be released in 4.17? Thank you!

Issue #9732, opened 2025-01-08
Navigation Issue on Privacy Policy Page When Bottom Back Button is Clicked First


paulpascal:
**Describe the bug**
When accessing the Privacy Policy page from the login page, clicking the bottom back button causes a navigation issue. When the user reaches the bottom of the page and clicks the bottom back button first, the page scrolls to the top. Then, clicking the top back button returns the user to the bottom, and vice versa.  On a browser, the user needs to click the browser back button to return to the login page. On the mobile app, this behavior persists even after closing and reopening the app; clearing the cache temporarily resolves the issue.

**To Reproduce**
1. Open the app (having privacy policy set up) without logging in.
2. Navigate to the Privacy Policy page from the login screen
3. Scroll down and click the bottom back button
4. Observe the behavior: The page scrolls to the top, but clicking the top back button returns the user to the bottom

**Expected behavior**
Clicking either back button should consistently return the user to the previous page (login page), regardless of which button is clicked first.

**Screenshots**
https://github.com/user-attachments/assets/0d486fa3-33c0-42fc-a7d4-f372339982d8
https://github.com/user-attachments/assets/3114cec8-9765-495c-a2d6-141be55f6726

**Environment**
- App: webapp
- Version: all

**Additional context**
This issue occurs only if the bottom back button is clicked first. If the top back button is clicked first, the issue does not occur.

PR #9733, opened 2025-01-08
fix(#9732): infinite loop on privacy policy page

paulpascal:
# Description

Fixes Navigation Issue with Back Buttons on Privacy Policy Page

This fix addresses the navigation issue observed on the Privacy Policy page where clicking the bottom back button first leads to an infinite loop between the top and bottom of the page. Specifically:

- When accessing the Privacy Policy from the login page, clicking the bottom back button first would scroll the page to the top.
- The top back button, when clicked after this, would return the user to the bottom, and vice versa, creating an endless loop.

Changes Made:

- Adjusted the event listeners to ensure that the back button consistently takes the user back to the login page, without causing an infinite loop between the top and bottom of the Privacy Policy page.
- The top back button now returns to the login page as expected when clicked, regardless of whether the bottom or top button is clicked first.

Related Issue: [#9732](https://github.com/medic/cht-core/issues/9732)

# Code review checklist

# License


dianabarsan (changes requested):
Very cool! Thanks for this :) 
I think it would be super useful to add an e2e test for the login privacy policy. We have a bunch of e2e tests that cover privacy policies, but none target this page: https://github.com/medic/cht-core/blob/5157ee973030631ae5beb81828a7f07388580814/tests/e2e/default/privacy-policy/privacy-policy.wdio-spec.js#L9

Additionally, it will be useful to test this code automatically in our minimum browser (since you're switching to arrow functions) - which will happen automatically if you add an e2e test in that file. 

Thanks again! Nice work!

dianabarsan (approved):
Excellent work!!! Thank you for taking the time to add a comprehensive e2e test and delivering code at the highest quality. 

Am I free to merge this as is?

Issue #9735, opened 2025-01-10
Single sign on (SSO) using identity provider


mrjones-plip:
**Is your feature request related to a problem? Please describe.**
Some deployments don't want to have to use a decentralized set of users for program, of which CHT is just one piece of. Instead, they would like to use a centralized system to allow users to login using a single sign on approach. 

In the context of  OpenID Connect (OIDC), this would mean that CHT authenticates via an OAuth2 identity provider over OIDC.

This ticket will be the parent ticket for the MVP effort - [from the design doc](https://docs.google.com/document/d/1LUn1ZRetAmYE04CtdcTmp-bEBvl37AZ0CvFXZChXqfU/edit?tab=t.0):

> This is a minimum viable product (MVP) effort.  This means no extra features beyond what is needed to achieve the goal.  A bare minimum of effort should be put into extensibility beyond the MVP as it's assumed any code that needs to be extensible will be refactored later. Keep this really focused to literally only what is needed for the MVP!

**Describe the solution you'd like**
Allow users to login using single sign on

**Describe alternatives you've considered**
There was a larger effort earlier on more generic [Oauth 2.0 support](https://github.com/medic/cht-core/issues/8849).  This ticket is a MUCH smaller scope.

**Additional context**
This is a unified effort of the larger community who's interested in actively help develop this feature.  A [weekly meeting](https://docs.google.com/document/d/1cnrlchDo_TNvil3-2VAr5ekRWTutLqmczSPPHvfXgMI/edit?tab=t.0) is held to help move the process forward. Please comment on this ticket if you're interested in joining!

rmayore:
### Features to consider for testing
Free and Open source.
Docker based, meaning it can be installed locally and on the cloud
Support for common protocols such as OIDC, Oauth 2.0, SAML
Ability to run e2e tests
IAM
Single Sign-On
Documentation and support

### Extras (Beyond Testing)

Social Login
User federation
Multi-factor authentication

### Options Considered
Shibboleth - https://www.shibboleth.net/
Keycloak - https://www.keycloak.org/
OpenIAM - https://www.openiam.com/
Evolveum - https://evolveum.com/midpoint/#self-service
FusionAuth - https://fusionauth.io/
Apache Synocope - https://syncope.apache.org/
Auth0 - https://auth0.com/


### 1. Shibboleth
Free only for development and testing https://www.shibboleth.net/membership/consortium-membership-fees/
Open source
[...]

PR #9740, opened 2025-01-10
feat(#9706): use Proxy Auth for communicating with Couch

### LINKED ISSUE #9706: Proxy user requests to Couch with common (non-admin) api user

Currently the API server uses several different approaches for authenticated communication with Couch. All users have their sessions authenticated against the `_session` endpoint to ensure they are a valid user, but then:

- The configured admin `COUCH_USER` is used for:
  - Doing admin stuff like creating/updating design docs, adding/removing users, etc
  - MITM "attacking" the replication flow of _offline users_ (to control which data they are authorized to access). When an offline user syncs, the api server uses the admin user to actually read/write from Couch.
  - Other various API functions triggered by calls to REST endpoints, etc.
- The actual user logged in on the client is used for:
  - _Online users'_ direct interactions with Couch endpoints are proxied through the API server to Couch with the online user's actual session.

For better simplicity and consistency in these various workflows, I want to propose that the API server be updated to use a new non-admin user for performing all operations except those that explicitly require admin access:

- Admin `COUCH_USER`:
  - Doing admin stuff like creating/updating design docs, adding/removing users, etc
- Non-admin `api` user:
  - MITM "attacking" the replication flow of _offline users_ (to control which data they are authorized to access). When an offline user syncs, the api server uses the admin user to actually read/write from Couch.
  - Other various API functions triggered by calls to REST endpoints, etc.
  - _Online users'_ interactions with Couch endpoints are proxied through the API server and to Couch using the `api` user

With this approach, the `_session` endpoint would still be used to _authenticate_ the current user, but it would be trivial to enhance this authentication flow to support SSO by authenticating against a 3rd-party IDM service instead.

In the future when we [no longer proxy any requests directly to Couch](https://github.com/medic/cht-core/issues/8889), maybe it would not matter to just have all API interactions with Couch be via the admin user... 🤔   Then the `api` user would be unnecessary and redundant.  On the other hand, maybe it is better for data security/consistency to prefer using a non-admin user either way....

Alternatively, instead of needing to create/maintain a new `api` user for the api server to use when talking to Couch, we could enable [Proxy Authentication](https://docs.couchdb.org/en/stable/api/server/authn.html#api-auth-proxy) on Couch which would allow the api server to interact with Couch via one or more "virtual users" by means of a secret shared between api and Couch.  E.g. for each request that the API gets from the client, it would:

- Authenticate the user session by calling `_session` for the current user
- Hash the user's username with the secret
- Call the necessary Couch resource endpoints with the proxy auth headers set (including preset role(s) for the api server that are not specific to the current user).  The idea is that all _authorization_ happens in the API server and the user's roles are not relevant for the Couch interactions....

These approach seems really nice since we don't have to maintain an additional user and we could also have additional traceability for each request since "username" for each proxy auth request sent to Couch could actually match the actual end user (instead of having all Couch traffic be just from the generic `api` user).  

**Additional context**

Not sure how relevant this actually is, but seems worth noting that I think we are moving away from using session cookies on the server side?!?  https://github.com/medic/cht-core/issues/8338

jkuester:
# Description

Closes #9706

This is my prototype attempt at "simplifying" communication between api/sentinel and CouchDB.  As noted in the ticket, there are currently several different user modes used when api/sentinel connect to Couch.  Significantly, there is currently not really any way for api/sentinel to connect to Couch with just ["member" access](
https://docs.couchdb.org/en/stable/api/database/security.html#get--db-_security).  Currently all access (not using creds provided by a user) is full "admin".  Most of the time, admin access is unnecessary.

This PR tries to address both challenges by introducing [Couch Proxy Auth](https://docs.couchdb.org/en/stable/api/server/authn.html#api-auth-proxy) as the sole method used to authenticate Couch traffic from api/sentinel (except for the actual user auth checks to authenticate users).  The username `cht-system` gets added as a "member" user to the various Couch dbs that api/sentinel need to access.  Then, all traffic from api/sentinel includes the proxy auth headers needed to authenticate this `cht-system` user (including a token calculated via the `COUCHDB_SECRET`).  This means that by default all traffic has "member" access.  For requests that need admin access (or no access, e.g. for the auth requests) this must be explicitly configured in the request. (A request for the `cht-system` user is made "admin" simply by specifying `_admin` in the role header.)  IMHO this gives a nice balance of security and convenience.  

Including a `COUCHDB_SECRET` envar when deploying via Docker Compose is part of our [prod hosting instructions](https://docs.communityhealthtoolkit.org/hosting/4.x/production/docker/#prepare-environment-variables-file) (not positive about the k8s setup). However, it does not seem to actually be _required_ to run the compose config unless you are doing a multi-node Couch cluster.  Otherwise, if no secret is configured, then the Couch container will generate one when it runs.  This means that even if we update the compose config to inject the `COUCHDB_SECRET` envar, there is not a 100% guarantee it will be there.  So, I added some fallback code to ping CouchDB and lookup the secret. If the `COUCHDB_SECRET` envar is set, then the `COUCH_URL` envar for api/sentinel does not actually need to include the user/pass.  

The overall goal of simplifying Couch communication is to hopefully make the implementation of SSO Auth in https://github.com/medic/cht-core/issues/9735 straightforward so that SSO users can be authenticated against the 3rd party identity provider and then the api server does all the talking to Couch with no need of an actual Couch user session.

## TODO

- [ ] Update `couch-request` functionality to not use `_admin` role by default on the non-pouch requests.  We should be able to make "member" level requests by default. The `_admin` requests should be made by explicitly setting that header. 
- [ ] Fix the server-checks code to kill the server if the Couch secret cannot be found.
- [ ] Update bundled docker compose files to pass `COUCHDB_SECRET` in to the api and sentinel containers
- [ ] Update api routing to intercept online user Couch traffic and update it to use proxy auth headers

# Code review checklist

# License


api/src/db.js
```diff
@@ -193,34 +201,33 @@ if (UNIT_TEST_ENV) {
...
     }
 
-    logger.info(`Adding "${role}" role to ${dbname} ${property}`);
-    securityObject[property].roles.push(role);
+    logger.info(`Adding "${value}" role to ${dbname} members`);
+    securityObject.members[securityField].push(value);
     await request.put({ url: securityUrl.toString(), json: true, body: securityObject });
   };
 
-  module.exports.addRoleAsAdmin = (dbname, role) => addRoleToSecurity(dbname, role, true);
```
jkuester, line 224: This function was not being used anywhere.

api/src/services/config-watcher.js
```diff
@@ -143,14 +144,19 @@ const updateServiceWorker = () => {
   });
 };
 
-const load = () => {
+const addSystemUserToDbs = () => Promise.all([...DATABASES, { name: `${environment.db}-purged-cache` }]
```
jkuester, line 147: Technically this is not currently "config related", so it probably does not belong in here.  I just stuffed it in here for now since it was the right place in the startup and nowhere else seemed better.

Also probably need a more centralized approach to listing all the DBs that the system user should have member access too....

couchdb/10-docker-default.ini
```diff
@@ -19,6 +19,10 @@ bind_address = 0.0.0.0
 server_options = [{recbuf, 262144}]
 socket_options = [{sndbuf, 262144}, {nodelay, true}]
 require_valid_user = true
+authentication_handlers = {chttpd_auth, cookie_authentication_handler}, {chttpd_auth, proxy_authentication_handler}, {chttpd_auth, default_authentication_handler}
```
jkuester, line 22: The default value is just:

```ini
authentication_handlers = {chttpd_auth, cookie_authentication_handler}, {chttpd_auth, default_authentication_handler}
```

shared-libs/couch-request/src/couch-request.js
```diff
@@ -18,18 +20,19 @@ const methods = {
   HEAD: 'HEAD'
 };
 
+const promisedAdminAuthHeaders = getAuthHeaders(environment.username, '_admin');
```
jkuester, line 23: This needs to be changed to not set the `_admin` role by default.  I have it like this so everything runs, but the intended approach is to just use the default username with its "member" access.  Then, for requests that need admin access, they will have to explicitly set the `_admin` role header.

shared-libs/couch-request/src/proxy-auth.js
jkuester, line 1: I put the code for dealing with the `COUCHDB_SECRET` in here.  It could probably be its own shared-lib if we wanted but it seemed at least adjacent to the couch-request functionality so I started with it here.

jkuester:
@dianabarsan Could you please let me know what you think about the approach in this prototype? :pray:   The code is not done yet, but I think it proves enough of the functionality for us to be able to make a decision on if we want to pursue this approach.  

The api/sentinel dev instances do run locally for me and I have done _some_ smoke testing, but there are definitely still broken edge cases and incomplete flows....

shared-libs/couch-request/src/proxy-auth.js
jkuester: _Actually_, after more digging, it seems like `shared-libs/settings` might be a the most logical place to put this. _However_ seems like it could cause a circular dependency since that lib already depends on `couch-request`.  I need to look more into this.

dianabarsan:
Hi @jkuester 

I requested a review, so I get reminded about this. I had read the email over the weekend and it skipped my memory today. Apologies for the delay.

shared-libs/couch-request/src/proxy-auth.js
```diff
@@ -0,0 +1,44 @@
...
+  return {
+    'X-Auth-CouchDB-UserName': username,
+    'X-Auth-CouchDB-Roles': roles,
+    'X-Auth-CouchDB-Token': token,
+  };
+};
+
+const getAuthHeaders = async (username, roles) => {
+  if (!username) {
+    return createHeaders(username, roles);
```
dianabarsan, line 31: So, if the username is `null` we'll create headers with the username `null`? I don't follow.

dianabarsan (commented):
I left a few questions inline. 
Nice work on this. Overall I think this could be useful to overcome the difficulties in handling sso, however I'm still not convinced about how secure this will be, because it seems that right now everyone gets a free pass.

shared-libs/couch-request/src/proxy-auth.js
```diff
@@ -0,0 +1,44 @@
...
+
+const { COUCH_URL, COUCHDB_SECRET } = process.env;
+
+const promisedCouchSecret = (async () => {
+  const envarSecret = COUCHDB_SECRET?.trim();
+  if (envarSecret) {
+    return envarSecret;
+  }
+
+  // Need to look up the secret (assuming the COUCH_URL contains user/pass)
```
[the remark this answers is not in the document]
jkuester: In order to make things passive, we need to support the case where no `COUCHDB_SECRET` is set, but the `COUCH_URL` contains the username/password (this is likely the case for many existing single-node servers).  With these changes the "recommended" case would be where `COUCHDB_SECRET` is set and `COUCH_URL` does not contain username/password.

The case we do not support is where `COUCHDB_SECRET` is NOT set and `COUCH_URL` does not contain username/password. (Since in this case we have no way of actually connecting to Couch...)

shared-libs/couch-request/src/proxy-auth.js
```diff
@@ -0,0 +1,44 @@
...
+// TODO Should we also load configured values for these header names (in case someone customized them)?
+const createHeaders = (username, roles, token) => {
+  return {
+    'X-Auth-CouchDB-UserName': username,
+    'X-Auth-CouchDB-Roles': roles,
+    'X-Auth-CouchDB-Token': token,
+  };
+};
+
+const getAuthHeaders = async (username, roles) => {
```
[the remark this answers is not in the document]
jkuester: "everyone" here would be "all the calls coming from api/sentinel" and "free pass" would be "member-level access to Couch dbs (upgradeable to admin by changing the roles header value)".  

The goal is to use proxy auth for all Couch traffic _after the user making the request is authenticated_.  (More on this in my response to your next comment.)

shared-libs/couch-request/src/proxy-auth.js
jkuester: Yeah, so this really needs its own code path instead of the confusing overloading I am doing here.  But practically, yes. When you pass in `null`, you get `null` values for the headers which means that the proxy auth headers are not actually set on the request.  

So, when doing the `_session` checks to authenticate the user making the api REST call, we pass `null` in here to make sure the `_session` call is done without any proxy auth headers (and is just authenticated via the users username/password or their session cookie).  

For all the normal traffic between api/sentinal and Couch we want to use Proxy Auth. But, for the specific user authorization checks (made against `_session`) we explicitly disable the proxy auth (by passing `null`), so we can actually check the user.  For SSO flows, instead of calling `_session` to validate the user, we would call the 3rd party identity provider.  Either way, once the user is authenticated, the subsequent communication between api and couch should be the same and should happen via proxy auth.

shared-libs/couch-request/src/proxy-auth.js
jkuester: After more digging/thinking about this, my current conclusion is that it is best to leave this header/secret logic here in `couch-request`.  Putting it in `shared-libs/settings` would mean that lib could no longer depend on `couch-request`.  Looking up the secret probably needs to be a one-off type of thing anyway since we have to be able to do it without passing through the normal `couch-request` flow.

PR #9748, opened 2025-01-16
fix(#9747): throw error if outbound config is undefined

### LINKED ISSUE #9747: Misconfigurations in outbound push config result in unexpected behavior.
**Describe the bug**
Misconfigurations in outbound push config result in unexpected behavior.
If a mapping value does not have type "string" and is not an Object with a property "expr", this is a misconfiguration; it is not defined what the mapping should do. 
What it actually does is include the entire document in the mapped field in the payload.

**To Reproduce**
1. create an outbound push config with a mapping field that is neither a string nor an Object with property "expr", e.g.
```json
...
"outbound": {
    "relevant_to": "doc.type == 'data_record'",
    "mapping": {
        "example_path": {
            "not_expr": "[{ \"example\": doc.phone }]",
            "optional": true
        }
    }
}
```
2. create any document that will trigger the outbound push (any report with the example above)
3. Look at the mapped field in the outbound push body ("example_path" in the example above); if will contain the entire document

**Expected behavior**
If the configuration has a mapping value that is undefined, mapping should raise error `OutboundError(`Mapping error for '${key}/${dest}' JS error on source document: '${doc._id}': ${err}`);` similar to what happens with other config misconfigurations.
This alerts users to fix the configuration instead of allowing the outbound push to silently continue with a payload that is probably not what they intended.

witash:
closes #9747

### DIFF
--- a/shared-libs/outbound/test/outbound.spec.js
+++ b/shared-libs/outbound/test/outbound.spec.js
@@ -210,6 +210,20 @@ describe('outbound shared library', () => {
 
       assert.throws(() => mapDocumentToPayload(doc, conf, 'test-doc'), /Mapping error/);
     });
+
+    it('throws an exception if the expression does not have either path or expr', () => {
+      const doc = {
+        _id: 'test-doc',
+      };
+
+      const conf = {
+        mapping: {
+          is_gonna_fail: {not_path_or_expr: 'doc.fields.null.pointer'},
+        }
+      };
+
+      assert.throws(() => mapDocumentToPayload(doc, conf, 'test-doc'), /Mapping error/);
+    });
   });
 
   describe('updateInfo', () => {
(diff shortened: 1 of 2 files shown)


shared-libs/outbound/test/outbound.spec.js
```diff
@@ -210,6 +210,20 @@ describe('outbound shared library', () => {
...
+        _id: 'test-doc',
+      };
+
+      const conf = {
+        mapping: {
+          is_gonna_fail: {not_path_or_expr: 'doc.fields.null.pointer'},
+        }
+      };
+
+      assert.throws(() => mapDocumentToPayload(doc, conf, 'test-doc'), /Mapping error/);
```
dianabarsan, line 225: Can you please make these error assertions more specific, so we know which error is triggered. Something like: 

```suggestion
      assert.throws(() => mapDocumentToPayload(doc, conf, 'test-doc'), /Mapping error.*'expr' or 'path' is required/);
```

andrablaj:
@witash I see you added new commits to the PR. Is this work ready to be reviewed?

dianabarsan (approved):
Sweet!!

Issue #9653, opened 2024-11-15
Refactor `shared-libs/validations` to call cht-datasource instead of directly querying freetext index


jkuester:
**Describe the issue**

Currently, the `share-libs/validations` code is [directly querying](https://github.com/medic/cht-core/blob/0d70a9880a712a087fc284fdbdac9f062041bfdb/shared-libs/validation/src/validation_utils.js#L12) the `medic-client/reports_by_freetext` view index.  This index is going away eventually.

**Describe the improvement you'd like**

We should update this logic to just use the new cht-datasource api being added in https://github.com/medic/cht-core/issues/9586.  Then, the future "under-the-hood" changes to the freetext indexes will be abstracted away from the `shared-libs/validations` code.

sugat009:
thanks @jkuester for the heads-up.

PR #9755, opened 2025-01-22
feat(#9653): refactor shared-libs/validations to call cht-datasource

sugat009:
# Description

Refactored the `shared-libs/validations` library to query `cht-datasource` when possible while still querying freetext indexes for when there is whitespace in the freetext search string.

Ticket: #9653 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9653-refactor-shared-libsvalidations-to-call-cht-datasource-instead-of-directly-querying-freetext-index/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9653-refactor-shared-libsvalidations-to-call-cht-datasource-instead-of-directly-querying-freetext-index/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9653-refactor-shared-libsvalidations-to-call-cht-datasource-instead-of-directly-querying-freetext-index/docker-compose/cht-couchdb-clustered.yml)

# License


sugat009:
I've set the target branch to [9586-implement-freetext-search-in-cht-datasource](https://github.com/medic/cht-core/tree/9586-implement-freetext-search-in-cht-datasource) because it has not been merged to master yet and will show lots of diffs.

sugat009:
@jkuester I've not added an actual e2e test for this one because the unit tests feel like integration tests for the whole package. Let me know what you think though.

jkuester:
I don't think we need integration tests _specific_ to `shared-libs/validations`.  However, if it is possible it would be good to have at least one `sentinel` integration test that triggers this validation code (basically validating the overall workflow that users the validations).

sugat009:
@jkuester I've added an integration test for uniqueness in the integration sentinel which eventually calls the changes in `shared-libs/validation`. However, I've gone with the `async/await` pattern for the tests as it was a pain to debug and follow the promise chain pattern currently there in other tests. 😅

shared-libs/validation/src/validation_utils.js
```diff
@@ -33,21 +47,49 @@ const parseStartDate = (duration) => {
...
 
-const exists = async (doc, fields, options = {}) => {
-  if (!fields.length) {
-    return Promise.reject('No arguments provided to "exists" validation function');
-  }
-  const requestOptions = fields.map(field => {
-    return { key: [`${field}:${lowerCaseString(doc[field])}`] };
+const getExistsRequestOptions = (fields, doc) => {
+  return fields.map(field => {
+    if (doc[field].includes(' ')) {
```
jkuester, line 52: As I mentioned [in the cht-datasource PR](https://github.com/medic/cht-core/pull/9625#discussion_r1939941973) we need to update our cht-datasource handling of keyed freetext queries that contain white-space.  

We cannot have this logic fall back to the `medic-client/reports_by_freetext` view query since that view will literally not exist when running on the server.

jkuester (changes requested):

shared-libs/validation/test/validations.js
```diff
@@ -18,15 +20,22 @@ const stubMe = (functionName) => {
...
+    dataContext = { bind: sinon.stub() };
     config = {};
     translate = sinon.stub().returnsArg(0);
     sinon.stub(logger, 'debug');
     sinon.stub(logger, 'error');
 
-    validation.init({ db, config, translate });
+    reportGetUuids = sinon.stub();
+    dataContext.bind.returns(reportGetUuids);
+    qualifier = sinon.stub(Qualifier, 'byFreetext');
```
jkuester, line 36: note: no need to make any changes here, but in the `api` unit tests we have tended to not mock out the `Qualifier` functions.  In my experience, there comes a point of diminishing returns for mocking lower level functions where it takes more effort to mock the behavior than it does to just manage the behavior in the test code.  Personally, I think the qualifier code fits that case where we might as well just use the real implementation code in the tests....

jkuester (changes requested):
Excellent refactor!  :star: 

The only piece missing is we need to pass in the data context when we init the validations lib [in webapp](https://github.com/medic/cht-core/blob/899e30df5fc6b7c52c99893609887ae714c87cd0/webapp/src/ts/services/validation.service.ts#L32).  I think we will need to add a new public function on the [cht-datasource.service](https://github.com/medic/cht-core/blob/master/webapp/src/ts/services/cht-datasource.service.ts) that will return the data context. Then, the validation.service can just use that to get the context.

webapp/src/ts/services/cht-datasource.service.ts
```diff
@@ -144,4 +144,8 @@ export class CHTDatasourceService {
       }
     };
   }
+
+  getDataContext() {
+    return this.dataContext;
```
jkuester, line 149: issue: this `dataContext` will  be null until we actually init the service.  I think this function needs to be `async` and then before returning the data context we just do: `await this.isInitialized();`

jkuester (approved):
:+1: One more minor addition and we should be good to go!

Issue #9759, opened 2025-01-29
Uplift to Angular 19


jkuester:
**Describe the issue**
Angular 19 [has been released](https://blog.angular.dev/meet-angular-v19-7b29dfd05b84)!  We should upgrade to stay current with our Angular version.

**Describe the improvement you'd like**

The [Angular Migration Guide](https://angular.dev/update-guide?v=18.0-19.0&l=3) notes the following steps for upgrading:

- [x] In the application's project directory, run `ng update @angular/core@19 @angular/cli@19` to update your application to Angular v19.
- [x] Angular directives, components and pipes are now standalone by default. Specify "standalone: false" for declarations that are currently declared in an NgModule. The Angular CLI will automatically update your code to reflect that.
- [x] Remove `this.` prefix when accessing template reference variables. For example, refactor `<div #foo></div>{{ this.foo }}` to `<div #foo></div>{{ foo }}`
- [x] Replace usages of `BrowserModule.withServerTransition()` with injection of the `APP_ID` token to set the application `id` instead.
- [x] The `factories` property in `KeyValueDiffers` has been removed.
- [x] In angular.json, replace the "name" option with "project" for the `@angular/localize` builder.
- [x] Rename `ExperimentalPendingTasks` to `PendingTasks`.
- [x] Update tests that relied on the Promise timing of effects to use await `whenStable()` or call `.detectChanges()` to trigger effects. For effects triggered during change detection, ensure they don't depend on the application being fully rendered or consider using `afterRenderEffect()`. Tests using faked clocks may need to fast-forward/flush the clock.
- [x] Upgrade to TypeScript version 5.5 or later.
- [x] Update tests using `fakeAsync` that rely on specific timing of zone coalescing and scheduling when a change happens outside the Angular zone (hybrid mode scheduling) as these timers are now affected by `tick` and `flush`.
- [x] When using `createComponent` API and not passing content for the first `ng-content`, provide `document.createTextNode('')` as a `projectableNode` to prevent rendering the default fallback content.
- [x] Update tests that rely on specific timing or ordering of change detection around custom elements, as the timing may have changed due to the switch to the hybrid scheduler.
- [x] Migrate from using `Router.errorHandler` to `withNavigationErrorHandler` from `provideRouter` or `errorHandler` from `RouterModule.forRoot`.
- [x] Update tests to handle errors thrown during `ApplicationRef.tick` by either triggering change detection synchronously or rejecting outstanding `ComponentFixture.whenStable` promises.
[...]

PR #9767, opened 2025-01-30
chore(#9759): uplift dependency versions

jkuester:
# Description

Docs PR: https://github.com/medic/cht-docs/pull/1754

### Changed `@tsconfig/node20` to `@tsconfig/node22`

Targeting our current node version. Updated the cht-datasource `tsconfig.json`.

### Added `q` as a dev-dependency in `admin`

This was needed for the unit tests which were piggy-backing on `q` in the root node_modules. This broke, though, because I guess `q` was no longer getting pulled in transitively.  

### Removed `diff`

It was added in [this commit](https://github.com/medic/cht-core/commit/19b4eadf948cdd64032b02fcd6be9c2621b14dd2#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519), but I cannot figure out where it was ever being used... :thinking: 

### `@faker-js/faker` - 8 > 9

Checked [migration guide](https://v9.fakerjs.dev/guide/upgrading.html#upgrading-to-v9). Nothing concerning as long as we are not using any deprecated functions.

### `@types/node` 20 > 22

Should be safe as Node 22 is now our base version.

### `cht-conf` 3 > 4

Just [dropping support](https://github.com/medic/cht-conf/blob/main/CHANGELOG.md#400-2024-07-01) for Node < 18.

### `inquirer` 9 > 12

We do not seem to be affected by any breaking changes:
- https://github.com/SBoudrias/Inquirer.js/releases/tag/inquirer%4010.0.0
- https://github.com/SBoudrias/Inquirer.js/releases/tag/inquirer%4011.0.0
- https://github.com/SBoudrias/Inquirer.js/releases/tag/inquirer%4012.0.0

### `mocha` 10 > 11

Just [dropping support](https://github.com/mochajs/mocha/releases/tag/v11.0.0) for older Node versions.

### `node-html-parser` 6 > 7

Now [targeting es6](https://github.com/taoqf/node-html-parser/blob/main/CHANGELOG.md#700-2024-12-26). Should not impact us since this is only used in api.

### `nyc` 15 > 17

Just [dropping support](https://github.com/istanbuljs/nyc/blob/main/CHANGELOG.md#1700-2024-04-19) for older Node versions.

### `rate-limiter-flexible` 3 > 5

Just dropping support for older Node versions:

- https://github.com/animir/node-rate-limiter-flexible/releases/tag/v4.0.0
- https://github.com/animir/node-rate-limiter-flexible/releases/tag/v5.0.0

### `shellcheck` 2 > 3

Just [dropping support](https://github.com/gunar/shellcheck/blob/main/CHANGELOG.md#v300---2024-06-06) for older Node versions.

### `uuid` 8 > 10

Mostly just dropping support for old Node versions.

- https://github.com/uuidjs/uuid/blob/main/CHANGELOG.md#900-2022-09-05
- https://github.com/uuidjs/uuid/blob/main/CHANGELOG.md#1000-2024-06-07

There are issues with going to version `11+`. Have logged https://github.com/medic/cht-core/issues/9769

### `sinon` 16 > 18

Just [dropping support](https://sinonjs.org/guides/migration-guide.html) for old Node versions and removing some deprecated code that does not affect us.  

We will need to do some migration work to get to `19+`, so I have logged https://github.com/medic/cht-core/issues/9768

### `helmet` 3 > 8

Not super clear why we were so behind on this one.  Helmet 4 [changed some things](https://helmetjs.github.io/faq/helmet-4-upgrade/), but none of those seem to affect us (at least not anymore).  

- https://github.com/helmetjs/helmet/blob/main/CHANGELOG.md#400---2020-08-02
- https://github.com/helmetjs/helmet/blob/main/CHANGELOG.md#500---2022-01-02
- https://github.com/helmetjs/helmet/blob/main/CHANGELOG.md#600---2022-08-26
- https://github.com/helmetjs/helmet/blob/main/CHANGELOG.md#700---2023-05-06
- https://github.com/helmetjs/helmet/blob/main/CHANGELOG.md#800

# Code review checklist

# License


tests/e2e/default/telemetry/telemetry.wdio-spec.js
```diff
@@ -190,11 +189,6 @@ describe('Telemetry', () => {
       expect(await getTelemetryEntryByKey('search_match:contacts_by_type_freetext:name')).to.have.lengthOf(2);
       expect(await getTelemetryEntryByKey('search_match:contacts_by_type_freetext:phone')).to.have.lengthOf(2);
 
-      const searchField = await $('.select2-search__field');
```
jkuester, line 193: Not sure exactly what it was that changed with this test.  I think it has always been a bit flaky around the timing of opening and shutting the search drop-down.  My guess is that something changed with the timing to exacerbate the flakiness (or I just got really unlucky).  Either way, I think my fix here is more stable anyway where I have updated the `generic-form` code to always "close" the search box when we clear the data from it (by clicking on the form title). This is the same way we close the date-picker.

webapp/custom-webpack.config.js
```diff
@@ -34,10 +34,6 @@ module.exports = {
       // openrosa-xpath-evaluator, but keeping it in adds 72.51KB to the bundle size.
       // https://github.com/medic/cht-core/issues/7324
       'node-forge': false,
-      // Only include the jquery version from the package.json (and not any different versions pulled in transitively).
-      // Once https://github.com/select2/select2/issues/5993 is resolved, we should try to coalesce back on one version
-      // of jquery and remove this alias.
-      'jquery': __dirname + '/node_modules/jquery',
```
jkuester, line 40: I have bumped us to use the same version of jquery as Enekto. Additionally, I suppose this _should_ be unnecessary with the `resolutions` section that we have in the package.json, right?

jkuester:
@dianabarsan can you look at this when you have some time?  Mostly it is straightforward, but you might pay special attention to:

- `helmet` - not sure why we were so far behind (maybe we used to use features that broke in `4.x`). But as things are now, I don't see anything preventing us from upgrading.  Anyway, an upgrade seems less risky than just staying on the super old version....
- `select2` - In webapp we had this pinned to `4.0.3` with warnings in the docs to not upgrade this.  However, everything seemed to work fine for me when I bumped this to `4.0.13`.  :shrug: (I did spend a bunch of time tinkering with different versions of `jquery`, but ended up having to leave that one at `3.5.1` since that was the only version that worked for both `select2` and `enketo-core`....  When we update to the latest enketo version, we should be able to bump to jquery `3.7` since that should fix the select2 issue.)

dianabarsan (approved):
Really appreciate you doing this work!

PR #9784, opened 2025-02-07
chore(#9759): upgrade to Angular 19 - use standalone components

jkuester:
# Description

Closes https://github.com/medic/cht-core/issues/9759

The main goal of this PR is to uplift Angular to `19.x` (see issue for more details).  Along the way I also uplifted TS to `5.7` (needed as part of the Angular uplift).  

The vast majority of the code line changes, though, were cased by the shift to [standalone components](https://blog.angular.dev/the-future-is-standalone-475d7edbc706). Angular 19 changed so that components were standalone by default (see more details in [this blog post](https://blog.angular.dev/meet-angular-v19-7b29dfd05b84) under the "Standalone defaults to true" heading).  Our existing Angular code was setup such that we had a bunch of interdependent `@Component` classes that were all rolled up into centralized `@NgModule` files.  To maintain that structure I was going to have to go through and add `standalone: false` to all of these `@Component` files.  Instead of doing that, I went the other direction and used the [migration scripts](https://angular.dev/reference/migrations/standalone) to convert all our components to actually just be standalone.  In the code this means that each component now imports all the different components that it depends on and the over-arching NgModules files are now gone.  I am no Angular expert, but just reading a bit it sounds like the idea of standalone components it to make your application structure less rigid/hierarchical/centrally-defined and more flexible/decentralized/reusable. Now, I did not make any real code changes, so we are not taking advantage of this _yet_, but the Angular ecosystem seems to be leaning in this direction (to the point where I think it is fair to consider standalone modules as a "best practice".)   For all the lines of code that were changed, you will notice that there was not really any changes to the actual implementation code, HTML, or tests.  To the best of my knowledge (and our automated tests measurements/validations) none of this code restructuring had any impact on the actual functionality of the app.

# Code review checklist

- [X] UI/UX backwards compatible: Test it works for the new design (enabled by default). And test it works in the old design, enable `can_view_old_navigation` permission to see the old design.

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9759_angular_19/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9759_angular_19/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9759_angular_19/docker-compose/cht-couchdb-clustered.yml)

# License


webapp/.eslintrc
```diff
@@ -32,7 +32,6 @@
         "@angular-eslint/contextual-lifecycle": "error",
         "@angular-eslint/directive-class-suffix": "error",
         "@angular-eslint/no-conflicting-lifecycle": "error",
-        "@angular-eslint/no-host-metadata-property": "error",
```
jkuester, line 35: This linting rule was removed in the newer version of the package.

webapp/src/ts/main.ts
```diff
@@ -1,3 +1,14 @@
+import { IntegrationApiService } from '@mm-services/integration-api.service';
+
+
+const logger = reducer => {
```
jkuester, line 4: This was originally over in the `app.module.ts`. Pulled it over here where it is being used now.

webapp/src/ts/main.ts
```diff
@@ -45,6 +85,10 @@ Object.defineProperties($, {
   htmlPrefilter: { value: (html) => html.replace(rxhtmlTag, '<$1></$2>') }
 });
 
+class MissingTranslationHandlerLog implements MissingTranslationHandler {
```
jkuester, line 88: Same here. This was initially defined in the `app.module.ts` file.

webapp/src/ts/main.ts
```diff
@@ -53,19 +97,45 @@ bootstrapper(POUCHDB_OPTIONS)
...
+        DatePipe,
+        provideHttpClient(withInterceptorsFromDi()),
+        provideAnimations()
+      ]
+    })
       .then((moduleRef) => {
-        window.CHTCore = moduleRef.instance.integration;
-        // backwards compatibility with the old way of reaching these services, the syntax looked like:
-        // angular.element(document.body).injector().get(<serviceName>);
-        window.angular = {
```
jkuester, line 62: AFICT this is not being used anymore.

webapp/tsconfig.json
```diff
@@ -1,17 +1,47 @@
...
+      "@mm-services/*": ["src/ts/services/*"]
     },
-    {
-      "path": "./tsconfig.spec.json"
-    }
-  ]
+    "strictNullChecks": true,
+    "alwaysStrict": true,
+    "noImplicitThis": true,
+    "outDir": "dist",
```
jkuester, line 37: Also somewhere along the way with upgrading TS and Angular and tweaking the tsconfig, it started outputting the `.js` and `.js.map` files directly next to all the `.ts` source files (in the same directories).  This is not desired since we do not actually want to commit these files.  

Following the pattern from cht-datasource, I added this config to just put these files into `webapp/dist`.  

The weird/slightly-concerning part, though, is that I could not figure out where these files were going beforehand.  I _think_ you could avoid creating the files during the TS build by setting `noEmit: true` and `sourceMap: false`, but there was no such existing configuration.  

I tinkered with this for awhile, but could not figure out what was the root cause of the behavior. Everything I read indicated:

- It is useful to have the files (when doing debugging/development) and setting the above config is not encouraged
- It is expected for those files to be generated as part of the Angular build
- You can use `outDir` to specify where you want them to be placed.

Ultimately, I am unsatisfied, but not sure how much more effort to pour into solving this mystery vs just moving ahead with some working config here....

webapp/tests/karma/ts/reducers/global.spec.ts
```diff
@@ -263,7 +263,7 @@ describe('Global Reducer', () => {
       recordTelemetry: '.entry',
       cancelTranslationKey: 'a key',
       extra: 'property',
-      is: !'saved',
+      is: true,
```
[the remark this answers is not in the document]
sugat009: This is a weird one. Eslint in my IDE is also giving me a `this will always be truthy` error, but when I run this in typescript, it evaluates to `false`. I also ran the test with both `true` and `false` values, and in both cases, the test passed, so the test is not dependent on this value. We can go with `true` for now.

image

package.json
```diff
@@ -165,7 +165,7 @@
     "tail": "^2.2.6",
     "ts-node": "^10.9.2",
     "typedoc": "^0.27.6",
-    "typescript": "5.5.4",
+    "typescript": "5.7.3",
```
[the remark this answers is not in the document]
jkuester: :skull:   
image
:sweat_smile:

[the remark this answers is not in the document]
jkuester: I bumped all our `@typescript-eslint` dependency versions and fixed the warning. That did result in some linting errors, but was pretty easy to fix those.

shared-libs/cht-datasource/.eslintrc.js
```diff
@@ -64,6 +64,12 @@ module.exports = {
         }],
         ['jsdoc/check-tag-names']: ['error', { definedTags: ['typeParam']}],
       }
+    },
+    {
+      files: ['*.spec.ts'],
+      rules: {
+        ['@typescript-eslint/no-unused-expressions']: 'off',
```
jkuester, line 71: The vast majority of the new linting errors were due to the calls in the test like:

```ts
expect(isContactTypeQualifier.notCalled).to.be.true;
```

I could not find any eslint/chai-native approach that would resolve this issue, so seems file to just disable this rule for the tests...

shared-libs/cht-datasource/src/libs/core.ts
```diff
@@ -33,13 +33,14 @@ const isDataPrimitive = (value: unknown): value is DataPrimitive => {
...
 
-interface DataArray extends Readonly<DataValue[]> { }
+type DataArray = readonly DataValue[]
 
 const isDataArray = (value: unknown): value is DataArray => {
   return Array.isArray(value) && value.every(v => isDataPrimitive(v) || isDataArray(v) || isDataObject(v));
 };
 
 /** @internal */
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
```
jkuester, line 43: I cannot convert this interface to a type alias because TS does not let you have recursive type aliases (but you can do that with an interface).  

This `DataObject` interface is recursive because DataObject > DataValue > DataObject.

webapp/.eslintrc
```diff
@@ -45,10 +44,12 @@
         "no-restricted-syntax": "off",
         // Prevent TypeScript-specific constructs from being erroneously flagged as unused
         "no-unused-vars": "off",
-        "@typescript-eslint/no-unused-vars": "error",
-        // Require a specific member delimiter style for interfaces and type literals
-        // Default Semicolon style
-        "@typescript-eslint/member-delimiter-style": "error",
```
jkuester, line 51: This rule was removed from typescript-eslint.

PR #9818, opened 2025-02-24
fix(#9799): add user-agent header to outbound

### LINKED ISSUE #9799: Allow configuring headers in outbound push
Outbound push may sometimes require different HTTP headers such as User-Agent; currently its not possible to add or change HTTP headers other than authorization.

Allow headers to be added or changed in configuration like
```
        "headers": {
          "User-Agent": "CHT-Integration/1.0 libwww/2.17b3"
        },
```

witash:
adds user-agent header to outbound
User-Agent headers should include versions, which means a little extra stuff; it tries to get the CHT version in the same way as api/src/services/deploy-info.js, but that code is not directly available from shared-libs, so there is some small duplication. 
Added getVersion() to shared-libs/settings. It doesn't quite fit there, but I don't think its worth it to make another separate shared-lib

We could also just omit the version if it adds too much mess, but I think these small additions are ok.

closes #9799

### DIFF
--- a/shared-libs/environment/src/index.js
+++ b/shared-libs/environment/src/index.js
@@ -77,23 +77,20 @@ const getDeployInfo = async () => {
     return deployInfoCache;
   }
 
-  let ddoc;
   try {
-    ddoc = await request.get({
+    const ddoc = await request.get({
       url: `${couchUrl}/_design/${module.exports.ddoc}`,
-      json: true
     });
+    deployInfoCache = {
+      ...ddoc.build_info,
+      ...ddoc.deploy_info,
+      version: getVersionFromDdoc(ddoc)
+    };
+    return deployInfoCache;
   } catch (err) {
     logger.error('Error getting deploy info: %o', err);
     throw err;
   }
-
-  deployInfoCache = { 
-    ...ddoc.build_info, 
-    ...ddoc.deploy_info, 
-    version: getVersionFromDdoc(ddoc) 
-  };
-  return deployInfoCache;
 };
 
 const getVersion = async () => {
--- a/shared-libs/environment/src/index.js
+++ b/shared-libs/environment/src/index.js
@@ -1,4 +1,6 @@
 const logger = require('@medic/logger');
+const request = require('@medic/couch-request');
+const semver = require('semver');
 
 const { COUCH_URL, BUILDS_URL, PROXY_CHANGE_ORIGIN = false } = process.env;
 const DEFAULT_BUILDS_URL = 'https://staging.dev.medicmobile.org/_couch/builds_4';
@@ -19,12 +21,16 @@ if (!COUCH_URL) {
   logError();
 }
 
+let couchUrl;
+let serverUrl;
+let serverUrlNoAuth;
+
 try {
   // strip trailing slash from to prevent bugs in path matching
-  const couchUrl = COUCH_URL.replace(/\/$/, '');
+  couchUrl = COUCH_URL.replace(/\/$/, '');
   const parsedUrl = new URL(couchUrl);
-  const serverUrl = new URL('/', parsedUrl);
-  const serverUrlNoAuth = new URL('/', parsedUrl);
+  serverUrl = new URL('/', parsedUrl);
+  serverUrlNoAuth = new URL('/', parsedUrl);
   serverUrlNoAuth.username = '';
   serverUrlNoAuth.password = '';
 
@@ -57,5 +63,48 @@ try {
   process.exit(1);
 }
 
+let deployInfoCache;
+
+const getVersionFromDdoc = (ddoc) => {
+  return semver.valid(ddoc.build_info?.version) ||
+    semver.valid(ddoc.deploy_info?.build) ||
+    ddoc.version ||
+    'unknown';
+};
+
+const getDeployInfo = async () => {
+  if (deployInfoCache) {
+    return deployInfoCache;
+  }
+
+  let ddoc;
+  try {
+    ddoc = await request.get({
+      url: `${couchUrl}/_design/${module.exports.ddoc}`,
+      json: true
+    });
+  } catch (err) {
+    logger.error('Error getting deploy info: %o', err);
+    throw err;
+  }
+
+  deployInfoCache = { 
+    ...ddoc.build_info, 
+    ...ddoc.deploy_info, 
+    version: getVersionFromDdoc(ddoc) 
+  };
+  return deployInfoCache;
+};
+
+const getVersion = async () => {
+  try {
+    const deployInfo = await getDeployInfo();
+    return deployInfo.version;
+  } catch (err) {
+    return 'unknown';
+  }
+};
 
 module.exports.isTesting = module.exports.db === 'medic-test';
+module.exports.getDeployInfo = getDeployInfo;
+module.exports.getVersion = getVersion;
--- a/shared-libs/environment/test/index.spec.js
+++ b/shared-libs/environment/test/index.spec.js
@@ -116,7 +116,7 @@ describe('environment', () => {
           timestamp: '2023-01-02'
         });
         expect(request.get.callCount).to.equal(1);
-        expect(request.get.args[0][0].url).to.equal('http://admin:pass@localhost:5984/medicdb/_design/medic');
+        expect(request.get.args).to.deep.equal([[{url: 'http://admin:pass@localhost:5984/medicdb/_design/medic'}]]);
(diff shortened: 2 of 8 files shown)


dianabarsan (commented):
Such a quick turnaround! nice! 
I left some minor suggestions inline.

gitcliff:
hello @dianabarsan @witash should we upgrade to this branch in the UI in the meantime as the work here still goes on ?

witash:
@dianabarsan this is ready for re-reivew, moved deploy-info to shared-libs/environment so it could be used in both the API and other shared-libs, and refactored outbound.js to use aysnc/await instead of explicit Promises.

@gitcliff I think it should be ok to upgrade your instance to this branch temporarily

shared-libs/environment/src/index.js
```diff
@@ -57,5 +63,48 @@ try {
...
+const getDeployInfo = async () => {
+  if (deployInfoCache) {
+    return deployInfoCache;
+  }
+
+  let ddoc;
+  try {
+    ddoc = await request.get({
+      url: `${couchUrl}/_design/${module.exports.ddoc}`,
+      json: true
```
dianabarsan, line 84: This should not be required (it should be enabled by default).

shared-libs/environment/src/index.js
```diff
@@ -57,5 +63,48 @@ try {
...
+};
+
+const getDeployInfo = async () => {
+  if (deployInfoCache) {
+    return deployInfoCache;
+  }
+
+  let ddoc;
+  try {
+    ddoc = await request.get({
```
dianabarsan, line 82: I believe `ddoc` can be a scoped `const` here, and we can return within the try-catch block.

shared-libs/environment/test/index.spec.js
```diff
@@ -38,7 +39,7 @@ describe('environment', () => {
 
     const env = rewire('../src/index');
 
-    expect(env).to.deep.equal({
+    expect(env).to.deep.include({
```
dianabarsan, line 42: I think we should keep the strict `equal` test here, and be definitive of which properties should be exported, and we can use `chai-exclude` to exclude the functions.

shared-libs/outbound/src/outbound.js
```diff
@@ -92,101 +95,118 @@ const mapDocumentToPayload = (doc, config, key) => {
...
+    await handleMusoSihAuth(authConf, config, sendOptions);
+  }
+};
+
+// Attempts to send a given payload using a given push config
+const sendPayload = async (payload, config) => {
+  const sendOptions = {
+    url: urlJoin(config.destination.base_url, config.destination.path),
+    body: payload,
+    json: true,
```
dianabarsan, line 186: `json` is `true` by default.

shared-libs/outbound/src/outbound.js
```diff
@@ -92,101 +95,118 @@ const mapDocumentToPayload = (doc, config, key) => {
...
-  return auth().then(() => {
-    if (logger.isDebugEnabled()) {
-      logger.debug('About to send outbound request');
-      const clone = JSON.parse(JSON.stringify(sendOptions));
-      if (clone.auth && clone.auth.password) {
-        // mask password before logging
-        clone.auth.password = '*****';
-      }
-      logger.debug(JSON.stringify(clone, null, 2));
+  sendOptions.headers['user-agent'] = await getUserAgent();
```
dianabarsan, line 191: Why not add this directly when declaring `sendOptions`?

dianabarsan (commented):
Very nice. Just some minor comments here.

shared-libs/outbound/src/outbound.js
```diff
@@ -16,22 +16,25 @@ const urlJoin = require('url-join');
...
-    return password;
-  });
+const CHT_AGENT = 'CHT';
+
+const fetchPassword = async key => {
+  const password = await secureSettings.getCredentials(key);
+  if (!password) {
+    throw new OutboundError(
+      `Credentials for '${key}' have not been configured. See the Outbound documentation.`
+    );
```
gitcliff, line 35: hello @witash been testing this out and what is the sample payload for this new refactoring i am hitting this error message with an outbound payload like 
```
 "patient triggers enroll workflow in Textit": {
      "relevant_to": "doc.type === 'person' && doc.docket",
      "destination": {
        "base_url": "https://textit.com",
        "auth": {
          "type": "header",
          "name": "Authorization",
          "value_key": "rapidpro.api"
        },
        "headers": {
          "User-Agent": "CHT/4.17.0"
        },
        "path": "/api/v2/flow_starts.json"
      },
      "mapping": {
        "flow": {
          "expr": "'650567d6-6662-4c0e-9ea3-98eba8ef'"
        },
        "urns": {
          "expr": "['tel:' + doc.phone]"
        }
      }
    },

```
Error log `2025-03-03T20:28:13.381 ERROR: Failed to push 356ffef1-29d8-44a7-8861-dc5d3637 to patient triggers enroll workflow in Textit: Credentials for 'rapidro.api' have not been configured. See the Outbound documentation. 
`

andrablaj:
Per our discussion today, @witash, re-requesting a review from @dianabarsan

shared-libs/environment/test/index.spec.js
```diff
@@ -74,4 +80,98 @@ describe('environment', () => {
...
+
+        const result = await env.getDeployInfo();
+        expect(result).to.deep.equal({
+          version: '4.1.0',
+          date: '2023-01-01',
+          base_version: '4.0.0',
+          timestamp: '2023-01-02'
+        });
+        expect(request.get.callCount).to.equal(1);
+        expect(request.get.args[0][0].url).to.equal('http://admin:pass@localhost:5984/medicdb/_design/medic');
```
dianabarsan, line 119: I think we should be more strict about the assertion here, and assert all params that are passed to request.get. 
We could easily make a `DELETE` request and the test would still pass

shared-libs/environment/test/index.spec.js
```diff
@@ -74,4 +80,98 @@ describe('environment', () => {
...
+        expect(request.get.callCount).to.equal(1);
+      });
+
+      it('should throw error if request fails', async () => {
+        sinon.stub(request, 'get').rejects(new Error('Failed to fetch'));
+        sinon.stub(logger, 'error');
+
+        const env = rewire('../src/index');
+
+        try {
```
dianabarsan, line 144: A nicer way to assert on calls we know will fail is do something like: 
```suggestion
await expect(env.getDeployInfo()).to.eventually.be.rejectedWith('Failed to fetch');
```

This requires `chai.use(require('chai-as-promised''))`; you should be able to find examples of usage throughout our codebase.

shared-libs/outbound/test/outbound.spec.js
```diff
@@ -340,10 +341,7 @@ describe('outbound shared library', () => {
           assert.equal(request.post.callCount, 1);
           assert.equal(request.post.args[0][0].url, 'http://test/foo');
           assert.deepEqual(request.post.args[0][0].body, {some: 'data'});
-          assert.equal(request.post.args[0][0].json, true);
-          assert.deepEqual(request.post.args[0][0].headers, {
```
dianabarsan, line 344: I generally prefer to keep test assertions with as much parameter coverage as possible, So here, instead of reducing coverage, can you please update the test to assert other potential headers too instead of reducing the assertion to just one header?

shared-libs/outbound/test/outbound.spec.js
```diff
@@ -267,14 +268,15 @@ describe('outbound shared library', () => {
...
+      sinon.stub(environment, 'getVersion').resolves('4.18.0');
       sinon.stub(request, 'post').resolves();
 
       return outbound.__get__('sendPayload')(payload, conf)
         .then(() => {
           assert.equal(request.post.callCount, 1);
           assert.equal(request.post.args[0][0].url, 'http://test/foo');
           assert.deepEqual(request.post.args[0][0].body, {some: 'data'});
-          assert.equal(request.post.args[0][0].json, true);
+          assert.match(request.post.args[0][0].headers['user-agent'], /^CHT\/4\.18\.0/);
```
dianabarsan, line 279: Can you please add a test where os.platform and arch are also stubbed, so we see the complete user agent header as well?

dianabarsan (commented):
Very nice! Just a few minor requests inline :) Thanks for your patience.

dianabarsan (approved):
Approving to unblock this. Good luck!!

Issue #9770, opened 2025-01-31
Switch from CommonJS to ES Modules


jkuester:
**Describe the issue**

For better or worse it seems that [ES Modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) are not going to go away.  Various parts of the JS ecosystem have started dropping support for CommonJS modules (which is what we are using currently).

These are the currently known dependency packages whose latest versions only support ESM:

- @octokit/core
- @octokit/plugin-paginate-graphql
- @types/chai
- @types/chai-as-promised
- chai
- chai-as-promised
- chai-exclude
- node-fetch
- url-join

**Describe the improvement you'd like**

Unfortunately, I expect converting to ESM to be very painful.  I did this recently with CHToolbox and [documented the experience](https://gist.github.com/jkuester/26339ac292d036134148d9153e6b6f47). The main issue for cht-core is probably going to be the same `sinon` troubles.  I think `esmock` is a reasonable solution here, but it will still involve _a lot_ of changes to the test setup code....

**Describe alternatives you've considered**

We could just give up on any dependencies that are ESM only.

dianabarsan:
Thank you for summarizing this and flagging all dependencies that are ESM only. 

I agree that this will require a titanic effort to convert everything. 
What we have done, so far, was to use older versions of these updated dependencies that used to support CommonJS. 

But, for some of these dependencies that you have listed: 

-> @octokit/core
-> @octokit/plugin-paginate-graphql

These are only used in our release notes script, which we could independently switch to ESM. 

-> @types/chai
-> @types/chai-as-promised
-> chai
-> chai-as-promised
-> chai-exclude

Testing frameworks are not high priority in terms of upgrading, unless there are some serious security concerns. Putting it _this much effort_ for syntactic sugar or small improvements in the framework seems like a bad deal to me. 

-> node-fetch

Afaik only PouchDb uses this, so it's not our direct dependency, and Pouch locks to the CommonJS version. 
We use bundled `fetch` for our HTTP communication. 

-> url-join

This is only used in `outbound`. I have used bundled `path` package to join URLs before, and it seems to be an ... "accepted" (tolerated) way of doing this. So this seems like the only dependency that we can consider dropping.

jkuester:
[...]

Issue #9738, opened 2025-01-10
Figure how to run e2e tests


mrjones-plip:
**Describe the issue**
As we start to write code, we need to start writing end to end (e2e) tests for #9735.  

**Describe the improvement you'd like**
This is an unknown process so some prototyping needs to be done.  May be dependent on #9737 

**Describe alternatives you've considered**
no tests FTW \o/ (that's a bad joke, we need tests)

tasawar-hussain:
Can i look into this?

jkuester:
Just to cross-link things here, we are leaning towards [using KeyCloack](https://github.com/medic/cht-core/issues/9737) for initial testing/development/prototyping of the OIDC functionality.

That does not mean KeyCloak will necessarily be preferable for e2e testing, though.  Among other things, it is pretty heavy and full featured, so it would be a lot to spin up for a few automated tests.  Not sure what other options might be available for a lightweight OIDC provider, but I did want to throw out [Kanidm](https://kanidm.com/) as an option.  The main benefits I see in Kanidm for automated tests are that it is more lightweight (Rust vs Java 😬), has first-class support for configuring everything (users/services/clients/etc) via a CLI interface, and the project has first-class support for deploying with Docker.

jkuester:
Another thought here is that our existing e2e tests for our RapidPro integration simply uses `express` to [mock a RapidPro server](https://github.com/medic/cht-core/blob/58a5e7d901c5946c4cdebd23611bf13db1c35b3f/tests/e2e/default/sms/rapidpro.wdio-spec.js#L17).  We _should_ be able to do the same and use `express` as a mock OIDC provider.

benkags:
:white_check_mark: Test SSO enabling in ICRC test infrastructure

Summary steps

- deploy cht-api vesion 4.19.0-9735-sso docker image 
- update settings (base_settings) to add `app_url` & `oidc_provider` settings, compile and upload to instance
- configure Entra ID  application in the application Authentication tab
  - configure redirect url URI: `https://<public_app_url>/medic/login/oidc`
  - check "ID tokens (used for implicit and hybrid flows)" option

    ![Image](https://github.com/user-attachments/assets/11a1e102-c7d0-402f-ba9b-acb578a58279)

  - create application secret
  - Add user to the application AZ group associated with Entra ID app
- push Entra ID app OIDC credentials to CouchDB
[...]

PR #9833, opened 2025-03-10
feat(#9765): add sso login

### LINKED ISSUE #9765: Add support to the CHT api for new endpoint for OIDC login
- The `/medic/login/oidc` endpoint should support `GET` requests with a `code` query param. The `code` should be the `authorization_code` coming from the OIDC Provider.
- This is the endpoint the OIDC Provider should re-direct back to.
- Add a new `getOidc` function in `api/src/controllers/login.js` that will handle the request and call through to code in the `openid-client` to validate the `authorization_code`.
- For a good `authorization_code`, the OIDC Provider should return an `id_token` containing the username for the authenticated user.
- Lookup the user's doc in `_user`and confirm `oidc_provider` value is set
- Generate a Couch session cookie for user and respond with valid user session.
  - This will require access to the Couch `secret` value which can be requested from the Couch instance.  ~~There is some code in [the proxy auth PR](https://github.com/medic/cht-core/pull/9740) that does this.  I _think_ we might be able to put that functionality for getting the secret into `shared-libs/settings` (or maybe that code is already sufficient for retrieving the secret value?).~~
  - See the [`getKey` function](https://github.com/medic/cht-core/blob/2c5a640c59a17d2da2bccc71cf84791e4c90bab2/shared-libs/settings/src/index.js#L38) in shared-libs/settings. That is loading/using the Couch key already. I think we just need to extract that part of the function into a new `getCouchSecret` function and export that as a new function for the shared-libs/settings library.  Then we can call that function from the api code to get the Couch secret.
- Add integration tests for the [login controller](https://github.com/medic/cht-core/blob/58a5e7d901c5946c4cdebd23611bf13db1c35b3f/tests/integration/api/controllers/login.spec.js)

### LINKED ISSUE #9907: Handle situation where CHT user does not exist for SSO user
What should we do when the SSO user authenticates successfully with the OIDC provider, but once they are redirected back to the CHT, the api server cannot find any existing CHT user for the SSO user?

It seems like a good default action is just to redirect them back to the login page.  It would be even better if they ended up on the login page AND were shown some helpful message indicating the issue.  

My main concern is that an SSO login failure like this could be really hard for the personnel in the field to understand/troubleshoot without some kind of error message that helps you understand why it is that you "logged in" to the OIDC provider successfully, but still cannot access the CHT....

benkags:
# Description

Closes #9765 
Closes #9907

# Code review checklist

# License


benkags:
@jkuester this is ready for review

package.json
```diff
@@ -43,7 +43,7 @@
     "test": "npm run lint && npm run unit && npm run integration-api",
     "unit": "node scripts/build/cli npmCiModules && npm run unit-webapp && npm run unit-admin && npm run unit-shared-lib && npm run unit-api && npm run unit-sentinel",
     "unit-admin": "node ./scripts/ci/run-karma.js",
-    "unit-api": "UNIT_TEST_ENV=1 nyc --nycrcPath='api/nyc.config.js' mocha 'api/tests/mocha/**/*.js'",
+    "unit-api": "UNIT_TEST_ENV=1 nyc --nycrcPath='api/nyc.config.js' mocha --exclude '**/controllers/sso.spec.js' 'api/tests/mocha/**/*.js' 'api/tests/mocha/controllers/sso.spec.js' && UNIT_TEST_ENV=1 nyc --nycrcPath='api/nyc.config.js' mocha 'api/tests/mocha/controllers/sso.spec.js'",
```
benkags, line 46: Context: [Here's](https://github.com/medic/cht-core/blob/3c6343ea5f9e71bf19bf373b47e073723699452e/api/src/openid-client-wrapper.js) the culprit. AFAICT, the openid-client is written as an ECMAScript Module which is not a problem because node is able to import it fine for the most part but for some reason, when loading tests it's complaining with `SyntaxError: Cannot use import statement outside a module` because of an import statement that is in the node package openid-client (in the node_modules folder): `import * as oauth from 'oauth4webapi';`; however, when I run the test file directly, all is fine. So, this is basically running the sso tests separately until I can figure this out.

jkuester: :disappointed: I was hoping we were not going to hit this so soon (https://github.com/medic/cht-core/issues/9770)...

Before things get too interesting here @benkags could you perhaps try rolling our version of openid-client back to `5.x` and see how much of the code you have written here would need to be updated to still work?  It looks like https://github.com/panva/openid-client/releases/tag/v6.0.0 was when they changed to an ESM module.  My thought is that if the apis that we are using here are not significantly changed between 5.x and 6.x (such that we are not incurring significant technical debt by using 5.x) then we can unblock ourselves for now by using `5.x` and revisit uplifting this dependency at a later date...

If there is a huge difference between the openid-client api from 5.x to 6.x such that we are going to have to re-work a bunch of logic, then I think we will have to perhaps explore more creative solutions, but if we can get away with easily just using the `5.x` version for now, then I would be more comfortable with that than with trying to hack an ESM module into our current CJS dependency tree...

(Also, FWIW according to the [npm download stats](https://www.npmjs.com/package/openid-client?activeTab=versions) the `5.7.1` version has 3-4x more downloads in the last week than all the `6.x` versions combined.  So, it is not like we would be on some crazy old code that no one else is using...)

jkuester (commented):
@benkags can you be sure to run the linting on this branch and get that passing?  `npm run lint`.

Getting [an eslint plugin](https://docs.communityhealthtoolkit.org/community/contributing/code/static-analysis/#eslint) setup in your local IDE makes it a lot more convenient to see the linting issues while you are typing too (instead of having to come back later and fix them).

benkags:
@jkuester fixing the linting. tried using openid-client@5.7.1 and the test are now failing. would need to look into that further.

package.json
jkuester: Okay @benkags several updates here!  (TLDR is that we should be good to keep using openid-client `6.x`.)

First, I found the [NodeJS docs I needed](https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require) regarding the feature to load ES Modules via require.  These docs are super helpful because they clarify how "experimental" this all is.  TLDR is that the feature exists as of NodeJS `v20.17.0`/`v22.0.0` (which means it is supported for the version of Node that we deploy). It is technically still an "experimental" feature, but with the status of `1.2 - Release candidate` which is the highest [level of stability](https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require) before you get to `Stable`.  `1.2 - Release candidate` means the feature is "hopefully ready to become stable. No further breaking changes are anticipated."  

This info, combined with the fact that the [openid-client release notes](https://github.com/panva/openid-client/releases/tag/v6.0.0) are explicitly supporting this feature, are enough to convince me that I think it is better to leverage this functionalty and stay current with our version of openid-client instead of pinning to an older `5.x` version of the library.  :+1: 

---

That does leave us with the issue of making the tests happy.  :sweat_smile: After poking around in this for awhile, I have learned a few important things.  The problem seems to be a weird confluence of:

- Loading an ESM module with `require`
- Instrumenting the code with `nyc`
- Specifying the mocha tests to run via the generic globbing (`**/*.js`)

If you remove any one of these elements the problem goes away. My theory is that `nyc` is probably doing janky things when it instruments the files (since nyc really [does not support ESM anyway](https://github.com/istanbuljs/nyc/issues/1287)) and that puts the files in some kind of state that confuses mocha when it tries to do the globbing.  If all else fails we could consider replacing nyc with `c8` since that seems like what everyone else is doing (but _technically we are not trying to get coverage for ESM code_ so that seems heavy handed).

_However,_ it seems like we can actually short circuit the mocha globbing logic so that it does not get confused by updating this npm script code to just be:

```json
    "unit-api": "UNIT_TEST_ENV=1 nyc --nycrcPath='api/nyc.config.js' mocha 'api/tests/mocha/controllers/sso.spec.js' 'api/tests/mocha/**/*.js'",
```

I am not 100% sure _why_ this works, but I have double-checked the count of the test run and everything seems to be fine! :+1:   Not a perfect fix, but seems like the best workaround I can come up with....

api/src/routing.js
```diff
@@ -450,6 +451,10 @@ app.postJson('/api/v3/users/:username', users.v3.update);
 app.delete('/api/v1/users/:username', users.delete);
 app.get('/api/v1/users-info', authorization.handleAuthErrors, authorization.getUserSettings, users.info);
 
+sso.init();
+app.get(routePrefix + 'login/oidc', sso.authorize);
+app.get(routePrefix + 'oidc/get_token', sso.login);
```
jkuester, line 456: question: the structure of these end-points seems a bit confusing (I would have no idea which is which until I actually dig into the code and try to figure it out....)

One option would be to map `sso.authorize` to something like `login/oidc/authorize` and then `sso.login` could point to `login/oidc`. So, the flow would be to GET `login/oidc/authorize` which would redirect you to the OIDC provider. Once you complete the authentication there you would get redirected back to `login/oidc` with your authorization code.  

Alternatively, we could just have one `login/oidc` endpoint.  If you call it _without_ a `code` query parameter, then it will automatically redirect you to the OIDC provider.  Then when the OIDC Provider auth is complete it would redirect you _back_ to `login/oidc` with the `code` query parameter set (at which point the controller would send you to `sso.login`).  

Does either of these approaches seem like a good idea?

api/src/controllers/sso.js
```diff
@@ -0,0 +1,200 @@
...
+    throw new Error({ status: 503, error: err });
+  }
+
+  logger.info('Authorization server config auth config loaded successfully.');
+
+  return ASConfig;
+};
+
+const getAuthorizationUrl = async (serverConfig, redirectUrl) => {
+  code_verifier = client.randomPKCECodeVerifier();
```
jkuester, line 69: We decided not to worry about the PKCE process in the MVP code. The main challenge is around handling concurrent OIDC login requests. Your current code here will not work for that since the `code_verifier` will be re-written if one login request starts before the other one finishes.  

If you check the version of the [design doc](https://docs.google.com/document/d/1LUn1ZRetAmYE04CtdcTmp-bEBvl37AZ0CvFXZChXqfU/edit?tab=t.0#heading=h.w42npptux3pi) called `PKCE Removed`, that is when I pull all the PKCE from the design.  We decided that because we were using the confidential client flow and PKCE is not _required_ by the OIDC spec (only "recommended" for confidential clients) we would not try to worry about caching code verifiers for our MVP implementation.

api/src/controllers/sso.js
```diff
@@ -0,0 +1,200 @@
...
+
+  const clientSecret = await secureSettings.getCredentials(OIDC_CLIENT_SECRET_KEY);
+  if (!clientSecret) {
+    const err = `No OIDC client secret '${OIDC_CLIENT_SECRET_KEY}' configured.`;
+    logger.error(err);
+    throw new Error(err);
+  }
+
+  try {
+    ASConfig = await networkCallRetry(
```
jkuester, line 51: question: I like that you have set this up so that we do not have to hammer the discovery URL for each request, but the more I think about this, the more concerns I have around caching this config.  The two main issues are:

- What happens if our `oidc_provider` settings change in the CHT `settings` doc? We do not restart the api server when this happens. So, we would need to bust this cache somehow if we switch to new `oidc_provider` settings.  
    - This should not be too difficult since we can use the [`config-watcher.watch`](https://github.com/medic/cht-core/blob/5152cc5e542c607c5d05e19ae8a30f15b8c4a4df/api/src/services/config-watcher.js#L189) function to know when the settings doc changes. But it does add extra logic/moving-parts here...
- Perhaps a more difficult concern is how would we know when config might change in the OIDC provider?  (meaning the config we are caching from the discovery is now stale/inaccurate) I have not had a chance to dig too much on what the standard practice here is (and if it is normal to just call the discovery for each request), but I think we want to be careful how we proceed here....

api/src/controllers/sso.js
jkuester, line 1: suggestion: I would like to keep it so that all the traffic going through the `/login` endpoints gets routed through functions in the `login` controller.  What do you think about making this file a service (in `src/services`)?  We keep all  the oidc-specific logic in this file, but it gets consumed by a bit of code in the `login` controller than is responsible for handling the current req/rep.  I think that would give us a good alignment with the existing code while still maintaining a separation of concerns.

jkuester (changes requested):
One other thing we will need to figure out here is the integration tests.  I am still hoping we can add some to the [login controller integration tests](https://github.com/medic/cht-core/blob/58a5e7d901c5946c4cdebd23611bf13db1c35b3f/tests/integration/api/controllers/login.spec.js) to validate the behavior of the REST endpoints.  

However, I think that is going to require us figuring out how to mock the behavior of an OIDC provider (as noted [here](https://github.com/medic/cht-core/issues/9738#issuecomment-2655095251)).

package.json
benkags: Thanks for looking into this.

api/src/routing.js
benkags: The first options sounds better: separation of concerns.

api/src/controllers/sso.js
benkags: I think here the issue was the identity provider enforcing PKCE but I have not revisited this since initial experiments. Looking into this.

api/src/controllers/sso.js
benkags: Good point. This is definitely not thread safe and does not handle oidc_provider config update. I struggled with the right approach here a bit. All things considered a I think the best thing here is to hammer the oidc provider for every request to guarantee the updated oidc config at the time of making the request ( both from conf & oidc provider end). It's a bit more inefficient but probably not too much of a price to pay.

api/src/controllers/sso.js
benkags: Good suggestion. Let me look into it.

package.json
benkags: @jkuester should we just bump node to [22.14.0](https://nodejs.org/en/about/previous-releases#looking-for-the-latest-release-of-a-version-branch)?

api/src/openid-client-wrapper.js
```diff
@@ -0,0 +1,16 @@
+// lib/oidc.js
```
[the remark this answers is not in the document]
benkags: It allow us mock the openid-client; otherwise, we get an error in tests
image

api/src/routing.js
benkags: Fixed.

benkags:
@jkuester I'd like to handle the integration tests in a different PR. We do need to build and use a keycloak docker image and I think that is a good unit of work that can be reviewed independently. What do you think? If you are ok with the suggestion, this is ready for another round of review.

api/src/controllers/sso.js
jkuester: :+1: Yeah, that makes sense to me. Lets go with the inefficient but safe route first and we can iterate on this in the future as needed.

package.json
jkuester: Either way is seems fine, just so long as we do not go higher than what will be used in the docker image (which should always be the latest version of `22.x`).  You are probably right that there is not much sense in bumping from one outdated version to another outdated version when we could just go to `22.14`...

jkuester (changes requested):
:+1: Your service implementation is great! The logic flow is super straightforward.

api/src/services/sso-login.js
```diff
@@ -0,0 +1,191 @@
...
+
+/**
+ * Establishes connection to oidc server config to get server metadata.
+ * 
+ * @returns {object} OIDC server config.
+ */
+const oidcServerSConfig = async () => {
+  const settings = await settingsService.get();
+  if (!settings.oidc_provider) {
+    return;
```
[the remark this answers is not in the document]
benkags: Is should be a server error: 501 is perhaps the most appropriate.

api/src/services/sso-login.js
```diff
@@ -0,0 +1,191 @@
...
+
+const OIDC_CLIENT_SECRET_KEY = 'oidc:client-secret';
+
+const SERVER_ERROR = new Error({ status: 500, error: 'An error occurred when logging in.' });
+
+const networkCallRetry = async (call, retryCount = 3) => {
+  try {
+    return await call();
+  } catch (err) {
+    if (retryCount === 1) {
```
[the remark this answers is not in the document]
benkags: This mostly serves timeouts as I noticed Keycloak has but it's likely a local set up edge case.

api/src/services/sso-login.js
```diff
@@ -0,0 +1,191 @@
...
+const getAuthorizationUrl = async (redirectUrl) => {
+  const params = {
+    redirect_uri: redirectUrl,
+    scope: 'openid'
+  };
+
+  try {
+    const serverConfig = await oidcServerSConfig();
+    return client.buildAuthorizationUrl(serverConfig, params);
+  } catch (err) {
```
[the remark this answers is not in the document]
benkags: I am not sure we want to return the openid-client errors to the user. The errors are logged as they are before wrapping returning a processed user-friendly error to the user. For example in this case, the next line is `logger.error(err);`. In the case of the example you pointed out: no client secret, implementation errors are logged in full and a generic user message returned to show the user.

api/src/services/sso-login.js
```diff
@@ -0,0 +1,191 @@
...
+  let secret;
+  let authTimeout;
+  let salt;
+
+  try {
+    const userDoc = await users.getUserDoc(username);
+    salt = userDoc.salt;
+  } catch (err) {
+    logger.error(err);
+    throw new Error({ status: 401, error: 'You are not enabled for log in using SSO.'});
```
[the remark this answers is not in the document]
benkags: I was thinking substituting `sendLoginErrorResponse` [here](https://github.com/medic/cht-core/blob/36290e05d7af9f9bf4052a256c881151f54fe78a/api/src/controllers/login.js#L558) would be the right place.

api/package-lock.json
```diff
@@ -0,0 +1,50 @@
+{
```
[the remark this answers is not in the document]
benkags: Removed.

api/src/services/sso-login.js
```diff
@@ -0,0 +1,191 @@
...
+ * @param {string} currentUrl Current url that contains authorization code grant.
+ * @returns {object} Token id and user details.
+ */
+const getIdToken = async (currentUrl) => {
+
+  try {
+    const serverConfig = await oidcServerSConfig();
+    const tokens = await networkCallRetry(
+      () => client.authorizationCodeGrant(serverConfig, currentUrl, { idTokenExpected: true }),
+      3
```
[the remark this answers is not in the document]
benkags: fixed

api/src/services/sso-login.js
```diff
@@ -0,0 +1,191 @@
...
+ * @returns {string} Session cookie.
+ */
+const getCookie = async (username) => {
+  let secret;
+  let authTimeout;
+  let salt;
+
+  try {
+    const userDoc = await users.getUserDoc(username);
+    salt = userDoc.salt;
```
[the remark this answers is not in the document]
benkags: Added the check for salt value and if missing error 401 is raised.

api/src/services/sso-login.js
```diff
@@ -0,0 +1,210 @@
...
+ * @param {string} username Username.
+ * @returns {string} Session cookie.
+ */
+const getCookie = async (username) => {
+  let secret;
+  let authTimeout;
+  let salt;
+
+  try {
+    const userDoc = await users.getUserDoc(username);
```
jkuester, line 182: If we are resolving users by `email`, we will not be able to use this function to get the user doc.  Instead, I am thinking we should update the [`users_by_field` view query](https://github.com/medic/cht-core/blob/61a04a62d19323f87b2020c62ee07ed5dbd56695/ddocs/users-db/users/views/users_by_field/map.js) to emit the users by their `email` value.  Then we can query here with code like:

```
    const [userDoc, ...additionalUsers] = await db.users.query(
      'users/users_by_field',
      {
        key: ['email', email],
        include_docs: true
      }
    ).then(({ rows }) => {
      return rows
        .map(({ doc }) => doc)
        .filter(doc => doc?.oidc);
    });
```

Then, we should throw an exception if `userDoc` is undefined OR if `additionalUsers` _is NOT empty_ (that means more than one user has this email value).

jkuester: @benkags I found an issue with this proposed implementation. :disappointed: It turns out that we do not actually set the user's `email` value onto their `_users` document.  Instead that gets stored in their `user-settings` document in the `medic` database.  

My new proposal is that we drop the email stuff from this PR and just wrap this up by keeping the `preferred_username` mapping for now as you have it here.   Then, I can follow up with another PR for https://github.com/medic/cht-core/issues/9890 that can add a new Couch View index for looking up the `user-settings` by email.

benkags: Thanks for looking into the email support in `_users` objects. Agree on wrapping this up with current `preferred_username` approach.

api/src/services/sso-login.js
```diff
@@ -0,0 +1,210 @@
...
+    const { name, preferred_username, email } = tokens.claims();
+
+    return {
+      id_token,
+      user: {
+        name,
+        username: preferred_username,
+        email
+      }
+    };
```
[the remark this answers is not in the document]
benkags: Deferring this to medic/cht-core#9890 as per the conversation below.

jkuester (approved):
:rocket: :rocket:  LGTM :rocket: :rocket:

PR #9840, opened 2025-03-14
fix: update cht-form to not fail when loading a form using extension-…

jkuester:
# Description

Closes https://github.com/medic/cht-core/issues/9844

- Adds support for injecting extension-lib functions into cht-form
- Fixes runtime errors in cht-form by mocking out the language service
- Updates cht-form "cancel" logic to work properly with new behavior (presumably caused by the recent Angular uplift) where the web component framework de-bounces multiple calls to set a web component property with the same value.
    - As a part of this, I re-worked the integration test logic a bit to allow multiple forms to be submitted in the same test without actually reloading the page for each form. This should help ensure the cancel logic continues working as desired in the future.

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:cht-form_extension-libs/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:cht-form_extension-libs/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:cht-form_extension-libs/docker-compose/cht-couchdb-clustered.yml)

# License


webapp/web-components/cht-form/src/stubs/language.service.ts
```diff
@@ -0,0 +1,11 @@
+import { Injectable } from '@angular/core';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class LanguageService {
```
dianabarsan, line 6: This is only here so this code works in the config test harness?

webapp/web-components/cht-form/src/app.component.ts
```diff
@@ -236,15 +247,36 @@ export class AppComponent {
...
+      ?.closest('cht-form');
+    const component = myForm || this;
+    // @ts-expect-error it does exist
+    component.formXml = undefined;
+    // @ts-expect-error it does exist
+    component.formHtml = undefined;
+    // @ts-expect-error it does exist
+    component.formModel = undefined;
+    // @ts-expect-error it does exist
+    component.contactSummary = undefined;
```
dianabarsan, line 270: I have a PR that adds an additional field to formContext. This link between form-context AND the cht-form sems very loose. I wouldn't know I had to reset this new field if I had not read this code. 

Is it possible to somehow iterate over the possible properties of the formContext interface, to get their names, and reset their corresponding value in this component instead of hardcoding them here as well?

webapp/web-components/cht-form/tests/karma/stubs/cht-datasource.service.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    consoleError = sinon.stub(console, 'error');
+    service = TestBed.inject(CHTDatasourceService);
+  });
+
+  afterEach(() => sinon.restore());
+
+  it('getSync', () => {
+    const chtApi = service.getSync();
+
+    expect(chtApi).to.have.property('v1');
```
dianabarsan, line 21: I prefer to be exclusive here instead of inclusive, and go for something like: 
```suggestion
expect(chtApi).to.have.keys(['v1']);
expect(chtApi.v1).to.have.keys(['getExtensionLib']);
```

dianabarsan (commented):
Great work! i think I have one question inline.

webapp/web-components/cht-form/src/stubs/language.service.ts
jkuester: The real implementation of the `LanguageService` pulls in a bunch of other services that it depends on (specifically, I think the changes.service does not work in cht-form and keeps spewing errors into the console logs).  This change just isolates cht-form from the entire LanguageService dependency tree.  So, this is not just for the test-harness, but is required generally for proper operation of cht-form.

dianabarsan (approved):
Approving to unblock, since all questions have been addressed! 
Thanks @jkuester !

webapp/web-components/cht-form/src/stubs/language.service.ts
jkuester: > I think this in-concrete coupling between cht-form and other webapp services confuses me

Yeah, not just you.  :sweat:   Conceptually we are just generating two separate artifacts (an Angular `webapp` and a generic `cht-form` web component) from basically the same code base. But, that does mean that changes to the codebase can affect both artifacts.  Frustratingly, I have not been able to find much in the way of Angular tooling to help manage/detect changes in the service dependency hierarchy. It would be nice to at least get a clear notification when new service dependencies end up pulled in to cht-form.

So far, the best we have is trying to keep the unit/integration tests tight so we know exactly when stuff breaks...

webapp/web-components/cht-form/src/app.component.ts
jkuester: In theory I think we could make `EnketoFormContext` an interface and then include an implementation here in cht-form.  Then, when the interface changes, the cht-form build should break if its data model is not updated with the new interface changes.  I do kind of like this idea!  :thinking:

webapp/src/ts/services/form.service.ts
```diff
@@ -351,3 +352,40 @@ export class FormService {
   }
 }
 
+export class WebappEnketoFormContext implements EnketoFormContext {
```
dianabarsan, line 355: Nice moving this to forms service instead of enketo service! I was very tempted to do the same thing in a PR of mine, but it was too much an unrelated change. Thanks for making this switch!

webapp/web-components/cht-form/src/app.component.ts
dianabarsan: If something forces the two types to be synced, I'm happy!

dianabarsan (approved):
Sweeeeeet!

Issue #9751, opened 2025-01-21
Refactor `shared-libs/search` to call cht-datasource freetext search apis


jkuester:
**What feature do you want to improve?**
The `shared-libs/search` library currently directly queries the freetext search views when someone performs a freetext search with the library.  

In https://github.com/medic/cht-core/issues/9586 @sugat009 is adding support for querying the freetext search views to `shared-libs/cht-datasource`.  This is going to be further enhanced by changes from https://github.com/medic/cht-core/pull/9541 which will update the cht-datasource code to toggle between querying the offline freetext views vs the server-side lucene indexes via Nouveau.  

**Describe the improvement you'd like**

Instead of duplicating the toggling logic and view/index queries in both `cht-datasource` and `search`, we should just refactor `search` to consume the `cht-datasource` apis.

A freetext search is triggered in `shared-libs/search` by passing a `search` value in the `filters` object.  Depending on whether the `type` value is set to `'report'` or `'contact'`, this will cause the code to query either the `reports_by_freetext` or `contacts_by_freetext` views.  Alternatively, when doing a contact search, if the `filters.types.selected` is also set, the `contacts_by_type_freetext` view will be the one queried.  Each of these views have corresponding cht-datasource api calls that should be used the replace the Couch view queries. 

The tricky part is going to be trying to make the switch with the least amount of disruption to the `shared-libs/search` code. The existing logic leans pretty heavily on this generic array of `requests` returned from the `generate` function.  These `requests` represent the various view queries that need to be made as a part of the search and they are dealt with together because of the need to be able to do paging across them.  My initial thought is to try and maybe have a wrapper shim around the cht-datasource calls that allows them to mimic normal view requests.

Another complexity to keep in mind when approaching this code is the fact that AFIK Couch view queries have no default `limit` value.  So, if you query a Couch view with no custom `limit`, then you get back _all results_ for the query from the entire db.  This currently comes into play in the multi-request searches because to do the "paging" `shared-libs/search` actually just gets literally all the results for all the requests, does a big union of them, and then "pages" the union.  On the other hand, `cht-datasource` does have a default limit. For the multi-request searches, we will need to make repeated calls to cht-datasource to collect all the results (or loop with the the `Generator` functionality to let cht-datasource handle all the paging).  

**Describe alternatives you've considered**

[...]

PR #9845, opened 2025-03-18
feat(#9751): refactor shared libssearch to call cht datasource freetext search APIs

sugat009:
# Description

Refactor `shared-libs/search` to call `cht-datasource` for freetext searching. 

Closes: #9751 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9751-refactor-shared-libssearch-to-call-cht-datasource-freetext-search-apis/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9751-refactor-shared-libssearch-to-call-cht-datasource-freetext-search-apis/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9751-refactor-shared-libssearch-to-call-cht-datasource-freetext-search-apis/docker-compose/cht-couchdb-clustered.yml)

# License


sugat009:
Changed target branch to `couchdb-nouveau` temporarily to get the actual changes only.

sugat009:
@jkuester I'm a bit confused. I was going through the implementation in `shared-libs/search`. In these lines:
https://github.com/medic/cht-core/blob/b716bb18dd14f3f3026717a99eee54eba84e256e/shared-libs/search/src/search.js#L110:L117

The `request.freetext` value is checked for which if exists then `queryFreetext` function is called. However, inside [this file](https://github.com/medic/cht-core/blob/couchdb-nouveau/shared-libs/search/src/freetext-query.js) the value is never used and instead `startKey` and `key` are used.

I'm asking this because in `cht-datasource`, for freetext searching the value `freetext` is required. Which value should be passed from `shared-libs/search` to `cht-datasource` as the `freetext` value?

sugat009:
> @jkuester I'm a bit confused. I was going through the implementation in `shared-libs/search`. In these lines: https://github.com/medic/cht-core/blob/b716bb18dd14f3f3026717a99eee54eba84e256e/shared-libs/search/src/search.js#L110:L117
> 
> The `request.freetext` value is checked for which if exists then `queryFreetext` function is called. However, inside [this file](https://github.com/medic/cht-core/blob/couchdb-nouveau/shared-libs/search/src/freetext-query.js) the value is never used and instead `startKey` and `key` are used.
> 
> I'm asking this because in `cht-datasource`, for freetext searching the value `freetext` string value is required. Which value should be passed from `shared-libs/search` to `cht-datasource` as the `freetext` value?

Nevermind, `freetext` turns out to be a boolean value. 😅

shared-libs/cht-datasource/src/local/libs/data-context.ts
```diff
@@ -19,7 +20,8 @@ export class LocalDataContext extends AbstractDataContext {
   /** @internal */
   constructor(
     readonly medicDb: PouchDB.Database<Doc>,
-    readonly settings: SettingsService
+    readonly settings: SettingsService,
+    readonly url: string
```
sugat009, line 24: @jkuester I've added a `url` property in the `LocalContext` similar to `RemoteContext` as it will be used to query the Nouveau views. However, I'm getting a `url` error when querying Nouveau since it is set to blank. What is the correct way to initialize this value?

shared-libs/cht-datasource/src/local/libs/doc.ts
```diff
@@ -175,3 +178,78 @@ export const fetchAndFilterUuids = (
...
+    const response = await fetch(
+      url, {
+        method: 'POST',
+        body: getRequestBody(viewName, params, bookmark)
+      }
+    );
+
+    // this is to match what clouseau view is doing when
+    // limit and cursor values are given which do not exist for eg: cursor 100000
+    if (response.status === 500) {
```
sugat009, line 216: I ran into a case where if a `cursor`/`bookmark` value is invalid then, Nouveau will directly throw a 500 error, unlike Clouseau view which returns an empty array.

shared-libs/cht-datasource/src/local/libs/request-utils.ts
```diff
@@ -0,0 +1,65 @@
...
+};
+
+/** @internal */
+export const getAuthenticatedFetch = (view: string) => {
+  const nouveauPath = getNouveauPath(view);
+  const headers = new Headers();
+  headers.set('Content-Type', 'application/json');
+
+  return (url: string, options: RequestInit | undefined): Promise<Response> => {
+    return global.fetch(`${url}/medic/${nouveauPath}`, { headers, ...options });
```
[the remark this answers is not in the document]
jkuester: Yeah so auth here is going to be a bit of a pain since we now need to make `fetch` calls _from the local data context_.  Our existing cht-datasource `fetch` calls were for the `remote` data contet and are currently only being made from the browser, which handles setting the session cookie on all requests.  We don't have the same functionality guaranteed for the local context. This is the case where api/sentinel is consuming cht-datasource and using it to make a `fetch` call to the nouveau endpoint.

 For this local data context code, the "cleanest" thing I could come up with was to [piggy-back on the PouchDb's fetch call](https://github.com/medic/cht-core/blob/474ab9ad28af51a553ed6faee7d9f663ef18b355/shared-libs/search/src/freetext-query.js#L103) (since it already handles authentication) instead of directly using `global.fetch`.  Any other option I could come up with seemed more janky than that...

shared-libs/cht-datasource/src/local/libs/doc.ts
sugat009: Yes, a 4XX would have been ideal here.

We do have a case currently in the getPages* APIs where if there are 100 items in the database and the limit is also set to 100 then the cursor will be returned as "100" but when the same endpoint is requested with the cursor as "100" then, it will return an empty result. This is due to our previous [implementation](https://github.com/medic/cht-core/blob/9bcc58a5b8b1c8311a4c70c039dd393e43754dd4/shared-libs/cht-datasource/src/local/libs/doc.ts#L127) of pagination and presence of more results. We should probably change this first if we want to throw an error when passing in an invalid `bookmark` value. _The change will be an easy fix though, just adding a +1 to the limit value._

jkuester: So, for that same case were there are only 100 results and we return all 100 of them from nouveau, what will the `cursor`/bookmark value be that we return?  As long as we return a `null` cursor here for the nouveau flow, the consumer should no not to run another fetch.  I do not see an issue (at least yet) with this behavior being a bit different from the Couch view query flow where after we get the 100 results, we return 100 as the cursor since we don't actually know if there is more or not....

sugat009: For the above case, a base64-encoded string is returned from Couch as a bookmark for the Nouveau endpoint. In the case of Nouveau, since the bookmark itself is returned from the database, we are not calculating the cursor to return. However, when this bookmark is used the same as Closeau, an empty array is returned instead of an error. I guess we can throw errors for the 500 error case from Nouveau.

shared-libs/cht-datasource/src/local/libs/request-utils.ts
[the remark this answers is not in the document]
sugat009: UPDATE:
I tried this one like [here](api/src/db.js) and [here](sentinel/src/db.js). I found out that the `PouchDB` package requires `pouchdb-adapter-http` and `pouchdb-session-authentication` plugins to be able to use the authentication. So, naturally, I added them to the import list in `cht-datasource` as well. `pouchdb-adapter-http` did have a type declaration so that was easy to import but `pouchdb-session-authentication` did not have a type declaration so I declared it in a new  `.d.ts` file with the content:
```typescript
declare module 'pouchdb-session-authentication'
``` 
I also added this file to the `files` key in the `tsconfig.json` file which enabled me to import `pouchdb-session-authentication` and plugin into PouchDB. However, I'm still getting redirected to the login page.

Usage Code:
```typescript
import PouchDB from 'pouchdb-core';
import PouchDBAdapterHttp from 'pouchdb-adapter-http';
import PouchDBSessionAuth from 'pouchdb-session-authentication';

PouchDB.plugin(PouchDBAdapterHttp);
PouchDB.plugin(PouchDBSessionAuth);

/** @internal */
export const getAuthenticatedFetch = (db: PouchDB.Database<Doc>, view: string) => {
  const nouveauPath = getNouveauPath(view);
  const headers = new Headers();
  headers.set('Content-Type', 'application/json');

  return (url: string, options: RequestInit | undefined): Promise<Response> => {
    return PouchDB.fetch(`/medic/${nouveauPath}`, { headers, ...options });
  };
};
```
Also, without using `url` in `PouchDB.fetch` throws an error that says that `Only absolute URLs are supported`.

couchdb-nouveau/nouveau.yaml
```diff
@@ -0,0 +1,27 @@
+maxIndexesOpen: 3000
```
witash, line 1: this file was added in 55ba2bade but removed in 342faa4d4. it looks like this branch was branched off 9542_freetext_tco in between the two; to prevent the file being recreated, I think we can merge 9542_freetext_tco into this branch to get 342faa4d4.

shared-libs/search/package.json
```diff
@@ -7,5 +7,9 @@
     "test": "nyc --nycrcPath='../nyc.config.js' mocha ./test"
   },
   "author": "",
-  "license": "Apache-2.0"
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@medic/couch-request": "file:../couch-request",
+    "@medic/environment": "file:../environment"
```
witash, line 13: These dependencies are unused and should be removed; I think this is also branch confusion and will be fixed by merging the target

sugat009: Merging the target branch did not work for some reason, so I've made all the requested changes manually.

Issue #9854, opened 2025-03-24
Bug: Duplication in outbound push


jonathanbataire:
**Describe the bug**
When configuring more than one out bound pushes without a cron configuration eg obp1 and obp2 the second configuration (obp2) will be pushed twice. Once immediately and another after about 5 mins

**To Reproduce**
- create out bound push with atleast 2 configurations to 2 target end points eg `push-target1` and `push-target2`
- Configure both outbound pushes to be triggered by the same form type i.e `relevant_to: doc.form === malaria_screening`
- Leave the  "cron" empty so pushes occur immediately as soon as the report is submitted
- Login as a user and submit the sample form (malaria_screening)
- Outbound push will push records to both `push-target1` and `push-target2` and create another  task:outbound for `push-target2`
- After about 5 mins the push-target2 task will be executed and the same payload will be pushed to target2 again 
**Expected behavior**

- Target2 task shouldnt be created as it causes payload to be pushed twice

**Logs**

**Screenshots**

**Environment**
- Instance: dev instance on v4.9.0, local environment with v4.18.0
- Browser: Firefox and Chrome
- Client platform: Linux
- Version: 4.9.0 and the latest code 4.18.0

**Additional context**

andrablaj:
4.9 is [not supported anymore](https://docs.communityhealthtoolkit.org/core/releases/), so backporting to 4.9 is not a recommended approach. The ideal would be to upgrade MoH Uganda to a more recent version. Is the upgrade option viable for MoH Uganda?

dianabarsan:
Hi @jonathanbataire 
The advised path is to upgrade, but I understand it's very hard to get MOH signoff. 

I think we should not do selective backports (for example release a 4.9.x with the fix, but not release a 4.10.x with the fix), and there are technical limitations to these backports - like fixes for "external" issues that you would need to identify and also cherry pick in order to get your build to pass (one example quickly comes to mind: https://github.com/medic/cht-core/commit/5d50e060e57b985a85004a03ee3d78114bd6dabe ) and I am sure this is not the only one. 

Creating hotpatch branches with fixes is risky, because it further decreases the chances for future upgrades. We have done this very sparingly (maybe twice?) and never pushed more than one change, and it was always to fix absolutely critical bugs (and I don't believe this bug is critical). 

How badly is the instance affected by this bug? 
What is the logistical aspect of a potential upgrade that is preventing MOH approval?

PR #9855, opened 2025-03-24
fix: out bound push fails while writing to info doc

jonathanbataire:
# Description
When two out bound pushes running against the same document (same doc id) there is an uncaught error as shown below while writing to the sentinel info doc that leads to duplication in the push.

while looping thru the outbound keys eg push-1 and push-2, 
push-1 will successfully write the info doc for the report
when it get to push-2 it attempts to `put` the info doc but runs into a conflict error
This will lead to creation of an unnecessary task:outbound which leads to push-2 being pushed again.

Fixes https://github.com/medic/cht-core/issues/9854 
```
{
  error: 'conflict',
  reason: 'Document update conflict.',
  status: 409,
  name: 'conflict',
  message: 'Document update conflict.',
  stack: 'Error\n' +
    '    at Object.generateErrorFromResponse (/home/jonathan/Downloads/cht-core-4.9.0/sentinel/node_modules/pouchdb-errors/lib/index.js:104:18)\n' +
    '    at /home/jonathan/Downloads/cht-core-4.9.0/sentinel/node_modules/pouchdb-adapter-http/lib/index.js:254:33\n' +
    '    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)',
  docId: 'c4628ff1-e3ac-4a2c-a499-f2f5fe8d81b6-info'
}
```

# Code review checklist

# License


jonathanbataire:
ready for review @dianabarsan

shared-libs/transitions/src/transitions/mark_for_outbound.js
```diff
@@ -48,8 +48,17 @@ const markForOutbound = (change) => {
       .then(sent => {
         if (sent) {
           // Successfully sent, outbound.send wrote to the infodoc
-
-          return db.sentinel.put(change.info);
+          return db.sentinel.get(change.info._id)
```
dianabarsan, line 51: I think this conflict resolution is a bit verbose to be here, and it doesn't treat an  eventual conflict directly - in your error, the status that the PUT request got was 409, and here you're only resolving for 404s (not found). 

I think the best solution here is to use the `infodoc` shared lib, which already has infodoc conflct resolution built in _and_ already has a `saveCompletedTasks` function  (https://github.com/medic/cht-core/blob/master/shared-libs/infodoc/src/infodoc.js#L285). This function doesn't receive parameters now, but that can be changed. 

Please let me know what you think!

tests/integration/sentinel/transitions/mark-for-outbound.spec.js
```diff
@@ -133,6 +133,78 @@ describe('mark_for_outbound', () => {
         });
     });
 
+    it('correctly creates and sends multiple outbound requests for the same report immediately', () => {
```
dianabarsan, line 136: Love this test!

dianabarsan (changes requested):
Really appreciate you resolving on this! I left one suggestion inline.

shared-libs/infodoc/src/infodoc.js
```diff
@@ -155,8 +155,8 @@ const saveTransitions = change => {
   return saveProperty(change.id, change.info, 'transitions', {});
 };
 
-const saveCompletedTasks = (id, infodoc) => {
-  return saveProperty(id, infodoc, 'completed_tasks', {});
+const saveCompletedTasks = (id, infodoc, completedTasks = []) => {
```
jonathanbataire, line 158: like this?

dianabarsan: Oh... now that I check the implementation, I see that the property value was taken from the infodoc that was already passed, my bad!

dianabarsan (approved):
Awesome!

PR #9877, opened 2025-04-03
feat(#9762): setup SSO login button and redirect

### LINKED ISSUE #9762: Update API login controller to redirect to OIDC provider when one is configured
- Get the redirect URL from `openid-client` (might be the `buildAuthorizationUrl` function?)
	- Consider making an oidc service in the API to manage the `openid-client` functionality
- When a user visits `/medic/login`, the `get` function in `api/src/controllers/login.js` should trigger the redirect
- Can add integration test(s) for the [login controller](https://github.com/medic/cht-core/blob/58a5e7d901c5946c4cdebd23611bf13db1c35b3f/tests/integration/api/controllers/login.spec.js)
    - For the integration tests, I think we will have to consider if we need to spin up an OIDC provider, or if we can just create a fake one...

Kymoraa:
# Description

Updates CHT login page to have button for redirecting to OIDC provider

Closes #9762

# Code review checklist

# License


api/src/public/login/script.js
```diff
@@ -125,6 +125,10 @@ const getRedirectUrl = function() {
   }
 };
 
+const getSSOLoginUrl = function(){
```
benkags, line 128: When does this get invoked? From local tests, the controller code seems to take precedence.

api/src/controllers/login.js
```diff
@@ -496,6 +500,10 @@ module.exports = {
       })
       .catch(next);
   },
+  getSSOLogin: (req, res) => {
+    // Placeholder url. Change to redirect to KeyCloak UI for SSO login
+    res.redirect('https://www.google.com');
```
benkags, line 505: update to `{base_url}/medic/login/oidc`

api/src/routing.js
```diff
@@ -296,6 +296,7 @@ app.get('/extension-libs/:name', extensionLibs.get);
 app.get(routePrefix + 'login', login.get);
 app.get(routePrefix + 'login/identity', login.getIdentity);
 app.postJson(routePrefix + 'login', login.post);
+app.get(routePrefix + 'login/oidc', login.getSSOLogin);
```
benkags, line 299: This causes a recursion because `login.getSSOLogin` should redirect to this endpoint.

benkags: Remove this, it's added when addressing #9765

api/src/public/login/script.js
Kymoraa: This method is being invoked when the ssoLoginButton is clicked. Still on the same file, on line 221
```
  const ssoLoginButton = document.getElementById('login-sso');
  ssoLoginButton.addEventListener('click', function(){
    getSSOLoginUrl();
  });
```

benkags: I see. Which in turn invokes the function in the controller via the /medic/login/oidc route.

api/src/controllers/login.js
benkags: you can ignore this comment. this function will not be necessary because #9765 adds the odic login path to the route and handles the redirection to oidc provider.

api/src/routing.js
Kymoraa: PR updated @benkags

api/package.json
```diff
@@ -13,6 +13,6 @@
   },
   "scripts": {
     "toc": "doctoc --github --maxlevel 2 README.md",
-    "run-watch": "TZ=UTC nodemon --inspect=0.0.0.0:9229 --ignore 'build/static' --ignore 'build/public' --watch ./ --watch '../shared-libs/**/src/**' server.js -- --allow-cors"
+    "run-watch": "TZ=UTC nodemon -e js,json,html,properties --inspect=0.0.0.0:9229 --ignore 'build/static' --ignore 'build/public' --watch ./ --watch '../shared-libs/**/src/**' server.js -- --allow-cors"
```
benkags, line 16: Did you intend for this update to remain?

jkuester: We do want to keep this! Apparently nodemon [does not watch html files by default](https://medic.slack.com/archives/C07U4E31K40/p1738686544487489?thread_ts=1738671082.341249&cid=C07U4E31K40).

api/src/services/settings.js
```diff
@@ -49,6 +49,17 @@ const getDeprecatedTransitions = () => {
...
+const hasOidcProvider = async () => {
+  const settingsDoc = await getDoc();
+  if(!settingsDoc){
+    return false;
+  }
+  if(!!settingsDoc.settings && !!settingsDoc.settings.oidc_provider){
+    return true;
+  }
+  return false;
+};
```
jkuester, line 61: suggestion: we should be able to simplify this a bit.  The server will actually fail to start without a settings doc, so we do not need to explicitly handle that case here.  

```suggestion
const hasOidcProvider = async () => {
  const settings = await module.exports.get();
  return !!settings?.oidc_provider;
};
```

jkuester (approved):
Great work @Kymoraa (and thanks for the close review @benkags).  I had one simplification suggestion that I will go ahead and apply and then we should be able to merge this up!

PR #9887, opened 2025-04-09
feat(#9763): update default login flow to not allow sso users

### LINKED ISSUE #9763: Update non-SSO login flow to not allow login if `oidc_provider` set for user
- When a user visits the login page and enters credentials in non-SSO username/password boxes, the  ~`get`~ `post` function in `api/src/controllers/login.js` should check the user's `oidc_provider` property before allowing them to login.
- Do not allow a user to login with the normal login page if they have a value set for ~`oidc_provider`~ `oidc`.
- Include integration test(s) for the [login controller](https://github.com/medic/cht-core/blob/58a5e7d901c5946c4cdebd23611bf13db1c35b3f/tests/integration/api/controllers/login.spec.js)

rmayore:
# Description

Updates `login controllers`'s default password login to block sso users.
Updates `login controllers`'s reset password to block sso users.

Closes #9763

# Code review checklist

# License


rmayore:
@jkuester referring to [this](https://github.com/medic/cht-core/pull/9800#issuecomment-2812529196) comment, it still not clear of when in tests we should use `api/v1/users`, `api/v2/users` or `api/v3/users` to create users. I assume it doesn't matter a lot since they both create users.

benkags (approved):

tests/integration/api/controllers/login.spec.js
```diff
@@ -78,6 +78,11 @@ const setupTokenLoginSettings = (configureAppUrl = false) => {
     .then(() => utils.addTranslations('en', { login_sms: 'Instructions sms' }));
 };
 
+const setupOidcLoginSettings = async (clientId) => {
+  const settings = { oidc_provider: { client_id: clientId } };
+  return await utils.updateSettings(settings, { ignoreReload: true });
```
jkuester, line 83: suggestion: can you add a call to `utils.revertCustomSettings` in the `afterEach`?  Just want to make sure any settings changes do not accidentally leak across tests.

tests/integration/api/controllers/login.spec.js
```diff
@@ -157,6 +162,31 @@ describe('login', () => {
...
+        place: {
+          type: 'health_center',
+          name: 'Online place',
+          parent: 'PARENT_PLACE'
+        },
+        contact: {
+          name: 'OnlineUser'
+        },
+        roles: ['district_admin'], 
+      };
```
jkuester, line 182: suggestion: none of these other properties matter, right?  (Can just be the defaults from `user`).

```suggestion
      const newUser = {
        ...user,
        oidc: true,
      };
```

tests/integration/api/controllers/login.spec.js
```diff
@@ -157,6 +162,31 @@ describe('login', () => {
...
+      await utils.request({ path: '/api/v2/users', method: 'POST', body: newUser });
```
jkuester, line 183: thought: regarding your question about which create user endpoint to use, I think you are correct that it does not matter for this test (as long as it works).  Typically, when I am writing tests, I try to use the latest version of the endpoint (since if we ever remove support for endpoints it will be the earlier versions), but functionally it makes no difference here.

jkuester (changes requested):
:rocket: Looking great!  Added some requrests inline.

jkuester (approved):
Okay, that turned into a bit more of a mess than I expected!  :sweat_smile: Thanks for the work here @rmayore!  I had to get a bit creative with the logic in the login controller because of how things got changed from Bernard's branch, but everything seems to be working together now!  :+1:

PR #9895, opened 2025-04-12
fix(#7990): prevent credentials exposure in error response

### LINKED ISSUE #7990: Credentials exposed in logs
**Describe the bug**
Credentials exposed in error response when there is an error in the uri.

**To Reproduce**
1. Add the following code to a file in our code
```
const username = 'medic-user';
const password = 'medic-pass';
const message = 'Testing request';
const sender = '093764582649003';

const sendMessage = async () => {
  await console.log('testing seeder');
  await utils.request({
    method: 'POST',
    uri:'https://gamma-cht.dev.medicmobile.org',
    auth:{username:username, password:password},
    headers: {
      'Content-type': 'application/x-www-form-urlencoded'
    },
    body:`message=${message}&from=${sender}`,
  }).then(response => console.log(response));    
};
sendMessage();
```
2. Notice the uri is missing path (with should be `/api/v2/records`, in this case)
3. Run the file `node <path>/file.js`
4. See error in the logs

**Expected behavior**
Credentials/password should be either not logged in or masked
**Logs**
If applicable, include the server or browser logs.

**Screenshots**
image
**Environment**
- Instance: gamma
- Browser: -
- Client platform:  MacOS
- App:  api
- Version: 4.x

**Additional context**
Also notice that those logs are duplicated 
```
auth: {
      username: 'medic',
      password: '*****',
      user: 'medic',
      pass: '****'
    },
```

Satyamyaduvanshi:
# Description

   This PR fixes a security vulnerability where credentials were being exposed in error logs when there was an error in the URI or other request parameters.

   The fix adds a sanitization function that:
   1. Detects and redacts sensitive fields (like passwords, usernames, auth tokens) in error objects
   2. Uses regex to find and replace credentials in string error content
   3. Preserves the error message format while ensuring credentials are masked

   Fixes #7990

   # Code review checklist


shared-libs/couch-request/src/couch-request.js
```diff
@@ -180,8 +204,16 @@ const request = async (options = {}) => {
     return responseObj.body;
   }
 
-  const err = new Error(response.error || `${response.status} - ${JSON.stringify(responseObj.body)}`);
-  Object.assign(err, responseObj);
+  const sanitizedBody = sanitizeErrorResponse(responseObj.body);
```
sugat009, line 207: question: Would it make sense to not log the response body at all?

sugat009 (changes requested):

shared-libs/couch-request/src/couch-request.js
sugat009: I think we should remove the sensitive fields. The fields will not be able to be used further after masking. For instance, if I get an error object like below:
```javascript
{
  username: '*****',
  password: '*******'
}
```
I will not be able to do anything with it.

Satyamyaduvanshi: Okay, thank you for the feedback. I can remove the body, although it might be useful elsewhere. Also, instead of masking sensitive fields, I’ll remove them entirely. Does this sound correct, or are there any other changes needed?

Satyamyaduvanshi: and is there any discord channel or something else where i can join?

sugat009: > I can remove the body, although it might be useful elsewhere.

This is for confirmation. I meant this as the case when there is an error. The non-error workflow should not be changed.

sugat009: > and is there any discord channel or something else where i can join?

We have a [forum](https://forum.communityhealthtoolkit.org/) where there is an option for chat.

Satyamyaduvanshi: > > I can remove the body, although it might be useful elsewhere.
> 
> This is for confirmation. I meant this as the case when there is an error. The non-error workflow should not be changed.


To confirm I understand correctly:
For error responses only, you'd like me to remove sensitive authentication fields entirely rather than masking them with *****
For successful requests, I should keep the existing behavior unchanged - no modifications to how authentication data is handled in successful responses
The "body" of error responses should still be included for debugging, just with sensitive data removed
Is this what you meant? If so, I'll update the PR accordingly.

sugat009: Yes, that is correct. In case of errors, the auth-related sensitive fields should be removed.

Satyamyaduvanshi: test2
test1

Satyamyaduvanshi: @sugat009 I have raised the updated PR—please check it out and let me know if it needs any changes.

sugat009 (changes requested):
LGTM!

Can you add tests for these changes?
The test file is located [here](https://github.com/medic/cht-core/blob/master/shared-libs/couch-request/test/couch-request.js).

Satyamyaduvanshi:
yes sure!!

shared-libs/couch-request/test/couch-request.js
```diff
@@ -705,4 +705,122 @@ describe('couch-request', () => {
...
+    });
+
+    it('should return other data types unchanged', () => {
+      expect(sanitizeErrorResponse(123)).to.equal(123);
+      expect(sanitizeErrorResponse(true)).to.equal(true);
+      // Arrays may be converted to objects in the sanitization process
+    });
+  });
+
+  describe('error handling', () => {
```
sugat009, line 772: suggestion: I think this test suite is unnecessary, and the tests can be written outside of this suite in the outer suite.

shared-libs/couch-request/test/couch-request.js
```diff
@@ -705,4 +705,122 @@ describe('couch-request', () => {
...
+        
+        // Verify error body has sensitive fields removed
+        expect(error.body).to.not.have.property('password');
+        expect(error.body).to.not.have.property('auth');
+        expect(error.body).to.not.have.property('username');
+        
+        // But other fields should remain
+        expect(error.body).to.have.property('error', 'auth_error');
+        expect(error.body).to.have.property('reason', 'Invalid credentials');
+      }
```
sugat009, line 802: issue(non-blocking): We generally use this pattern when testing for exceptions being thrown.

```javascript
await expect(request(options)).to.be.rejectedWith(Error)
      .then(err => {
        expect(err.status).to.equal(500);
        expect(err.body).to.deep.equal({ error: 'Internal server error' });
      });
```

sugat009 (changes requested):

shared-libs/couch-request/test/couch-request.js
Satyamyaduvanshi: okay thnx

sugat009 (approved):

PR #9901, opened 2025-04-16
feat(#9764): update token_login flow to not allow oidc user

### LINKED ISSUE #9764: Update `token_login` flow to not allow login if `oidc_provider` value is set for user.
The `tokenPost` function in `api/src/controllers/login.js` should do the check before giving the user a valid session.

kanishkIIITD:
# Description

Updated token_login flow to not allow login if oidc_provider value is set for user. Made consistent async/await changes in loginByToken function.

Closes #9764

# Code review checklist

# License

### DIFF
--- a/api/src/controllers/login.js
+++ b/api/src/controllers/login.js
@@ -319,15 +319,21 @@ const loginByToken = (req, res) => {
         throw { status: 401, error: 'invalid' };
       }
 
-      return tokenLogin.resetPassword(userId).then(({ user, password }) => {
-        req.body = { user, password, locale: req.body.locale };
+      return users.getUserDoc(userId)
+        .then(userDoc => {
+          if (userDoc.oidc_provider) {
+            throw { status: 400, error: 'Token login not allowed for OIDC users' };
+          }
+          return tokenLogin.resetPassword(userId).then(({ user, password }) => {
+              req.body = { user, password, locale: req.body.locale };
 
-        return createSessionRetry(req)
-          .then(sessionRes => setCookies(req, res, sessionRes))
-          .then(redirectUrl => {
-            return tokenLogin.deactivateTokenLogin(userId).then(() => res.status(302).send(redirectUrl));
-          });
-      });
+              return createSessionRetry(req)
+                .then(sessionRes => setCookies(req, res, sessionRes))
+                .then(redirectUrl => {
+                  return tokenLogin.deactivateTokenLogin(userId).then(() => res.status(302).send(redirectUrl));
+                });
+            });
+        });
     })
     .catch((err = {}) => {
       logger.error('Error while logging in with token', err);
--- a/api/src/controllers/login.js
+++ b/api/src/controllers/login.js
@@ -331,7 +331,7 @@ const loginByToken = async (req, res) => {
     const status = err.status || err.code || 400;
     const message = err.error || err.message || 'Unexpected error logging in';
     const response = { error: message };
-    if(err.reason) {
+    if (err.reason) {
       response.reason = err.reason;
     }
     res.status(status).json(response);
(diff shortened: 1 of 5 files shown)


api/src/controllers/login.js
```diff
@@ -312,29 +312,32 @@ const loginByToken = (req, res) => {
...
       if (!userId) {
-        throw { status: 401, error: 'invalid' };
+        return res.status(401).json({ error: 'invalid', reason: 'Invalid token' });
       }
 
-      return tokenLogin.resetPassword(userId).then(({ user, password }) => {
-        req.body = { user, password, locale: req.body.locale };
+      const userDoc = await users.getUserDoc(userId);
+      if (userDoc.oidc) {
+        return res.status(400).json({ error: 'Token login not allowed for OIDC users' });
```
benkags, line 323: `Token log in no allowed for SSO users`

api/src/controllers/login.js
```diff
@@ -312,29 +312,32 @@ const loginByToken = (req, res) => {
     return res.status(400).json({ error: 'missing', reason: 'Missing required param' });
   }
 
-  return tokenLogin
-    .getUserByToken(req.params.token)
-    .then(userId => {
+    try {
```
benkags, line 315: check on the indentation here.

api/src/controllers/login.js
benkags: I'd also have this as the first check in this method.

benkags:
please also add a test for this scenario

api/src/controllers/login.js
```diff
@@ -303,38 +303,40 @@ const createSessionRetry = (req, retry=10) => {
...
+    if (!userId) {
+      return res.status(401).json({ error: 'invalid', reason: 'Invalid token' });
+    }
 
-      return tokenLogin.resetPassword(userId).then(({ user, password }) => {
-        req.body = { user, password, locale: req.body.locale };
+    const userDoc = await users.getUserDoc(userId);
+    if (userDoc.oidc) {
+      return res.status(400).json({ error: 'invalid', reason: 'Token login not allowed for SSO users' });
+    }
```
jkuester, line 320: suggestion: calling the `users.getUserDoc` here is adding an additional db round-trip.  Looking at this more closely, I think we could actually just move this validation down into the `tokenLogin.getUserByToken` function (over in the [`shared-libs/user-managment` code](https://github.com/medic/cht-core/blob/d47a578a1ad4fa6b8e59293a79a42ff8fa0cc15e/shared-libs/user-management/src/token-login.js#L286)).  Right before we return the `user._id` we can just check:

```js
      if (user.oidc && config.get().oidc_provider?.client_id) {
  const err = new Error('Token login not allowed for SSO users');
  err.status = 401;
  throw err;
}
```

jkuester (changes requested):
Thank you @kanishkIIITD for the contribution (and @benkags for the review)!  I have just adding a few more requests inline. :+1:

tests/integration/api/controllers/login.spec.js
```diff
@@ -257,5 +257,38 @@ describe('login', () => {
         .then(() => loginWithTokenLink(tokenLogin.token))
         .then(response => expectLoginToFail(response)); // fails after being activated the 1st time
     });
+
+    it('should reject token login for SSO users', () => {
```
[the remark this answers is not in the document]
jkuester: One thing to watch our for with this test is that once you add the check for the `oidc_provider` config setting in the implementation code, we are going to need to make sure that setting is enabled during the test. Should be able to do something similar to: https://github.com/medic/cht-core/pull/9887/files#diff-a59bc866714f32aa414cde186f492daf434c6ed36acc9531ff890528df6ddaa7

kanishkIIITD:
I have updated the code as suggested, and all unit and integration tests are passing locally. Ready for re-review!

PR #9909, opened 2025-04-18
feat(#9885): adds new meta audit database

### LINKED ISSUE #9885: Implement metadata audit log
Historically, the CHT has always had (or tried to have) an audit log. 
An older implementation involved a secondary database that stored, for each document, a copy of the current body along with metadata of when the change happened. 
This solution became very problematic with documents with large bodies or that were saved frequently. 

We had an issue where an SMS service would infinitely update a certain document, every time it was processed. This should not be an issue normally, but the audit log saved copy after copy of the same document in the same audit document, creating absurdly large docs that eventually overwhelmed CouchDb. 

A second iteration of the Audit log was haproxy. However this solution had another issue, where haproxy would not log whole payloads of requests (makes sense, since payloads can be gigantic! and logging them in text can be overwhelming for the I/O system), but limit the size to a maximum of 65k characters and only log first body chunk. 

More details in the parent issue. 

Iterating back on the solution where audit logs are stored in the database, except: 
- only metadata should be saved
- Make sure that we never get too large documents, this might mean that there should not be a 1-1 ratio between audit doc and medic doc (think of this as a log rotation, where once an audit doc is large enough, a new one is created)
- access should be fast and writing the audit should not be blocking

Continue to log to STDIO. Because we can only add this to API, it would mean that API would log every document change. In my opinion, this would pollute API logs significantly, especially in periods of high load. 
Another option is to stream this log through another service.

dianabarsan:
# Description

Adds new mechanism to audit document changes, through a new shared library, `audit`. The library is injected into every call that handles requests (PouchDb `fetch` and `couch-request`). 
When a monitored API is called (POST to create a doc, PUT to update a doc or _bulk_docs), the response is analyzed and entries are added into a new database `medic-audit`. 
In `medic-audit`, each document has an audit doc which contains the last 10 entries of received changes. Each entry includes the `rev` of the change in question, the user that made the change, the `request_id` (in case this was a direct client request), the service that made the change (api or sentinel) and the date when the change was registered. 
When an audit doc has 10 history entries, a rotation mechanism will save the old audit entry in a new doc, and the existing document trail will only contain the new change. 

Moves `getDeployInfo` into its own shared library to avoid circular dependencies. 

closes #9885 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9885-new-audit/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9885-new-audit/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9885-new-audit/docker-compose/cht-couchdb-clustered.yml)

# License


dianabarsan:
Hi @witash . I'd like another set of eyes on this. Appreciate your time and feedback!

witash:
Looks great, I can see how this could useful in a lot of different ways.

My only question is about the growth rate of the history db, since history entries are never deleted or archived.

Tasks get edited a lot because of automatic changes, and we know from cht sync that these changes are much more frequent than creations or updates for any other type of document. And there may be other examples that have a similar pattern of a lot of small, automatic updates.

So the history db will be mostly small changes to tasks or other automatically updated things. It should still be small compared to the main db but could still be significant...

Is it worth it to add a way to archive or delete old history?

dianabarsan:
Good questions! 

> Is it worth it to add a way to archive or delete old history?

I usually prefer to not overload an initial feature with lots of features. IMO this is quite linear in terms of storage usage: most documents are never edited and will have only one history entry, would you ask the same storage question if this PR added 5 fields to a document? probably not. 

There are some docs (forms for example) that do receive lots of changes, but these are also limited in terms of quantity. 
So I would not overthink or overdo this feature with an archive option until after we prove that this is indeed significantly disk consuming.

dianabarsan:
I'll handle the conflicts and re-request a review!

Thanks for your feedback!

dianabarsan:
Hi @sugat009 ! Because @witash is not available, would you mind having a look at this? 
Thanks a lot!

tests/integration/api/audit.spec.js
```diff
@@ -0,0 +1,232 @@
...
+      const audit = await getAudit(docIds);
+      docIds.forEach((id) => checkAudit(audit[id]));
+    });
+
+    it('should not audit service worker cache', async () => {
+      const audit = await getAudit('service-worker-meta');
+      expect(audit.audit).to.be.undefined;
+    });
+
+    it('should only store 20 history entries', async () => {
```
sugat009, line 215:
```suggestion
    it('should only store 10 history entries', async () => {
```

sugat009 (approved):
LGTM!
Very small suggestion though.

tests/integration/api/audit.spec.js
dianabarsan: Oh yea. it was 20 initially and then I reduced it to 10.

dianabarsan:
@lorerod I would love some feedback on this from you. Would you mind having a look?

lorerod:
Hi @dianabarsan,
Thank you for adding me to check out this feature! I'm currently testing it locally and have a question:

I'm testing a scenario where an offline user updates the same contact multiple times while offline. When the user comes back online, I see that only one update is recorded in the history. Is this the expected behavior, or should all the individual updates be registered?

<details> <summary>Audit record before multiples offline changes</summary>

```json
{
  "_id": "bfdd7097-ed29-43af-8a0d-3cc40904ebc0",
  "_rev": "5-01baf3df6aa0ac2f9a0863e8696146ee",
  "history": [
    {
      "rev": "2-25d468164345e5a19221115e5964d261",
      "request_id": "46d1a8e5bebd",
      "user": "adele.walter",
      "date": "2025-05-16T17:12:49.273Z",
      "service": "api"
    },
    {
      "rev": "3-58c4fec719640a8a753a6b2844f58857",
      "request_id": "ff2ee6cdf748",
      "user": "adele.walter",
      "date": "2025-05-16T17:18:01.050Z",
      "service": "api"
    },
    {
      "rev": "4-5d6b6372c6853160d944e93ac97c8588",
      "request_id": "ec9a921b1ef0",
      "user": "adele.walter",
      "date": "2025-05-16T17:21:15.680Z",
      "service": "api"
    },
    {
      "rev": "5-cd5fe25ba871db4645fba2584bb12df3",
      "request_id": "b5273bc43dfa",
      "user": "adele.walter",
      "date": "2025-05-16T17:24:13.724Z",
      "service": "api"
    },
    {
      "rev": "6-a2979343ee6fc3463db90e5868745efa",
      "request_id": "f96f63571a19",
      "user": "adele.walter",
      "date": "2025-05-16T17:32:45.991Z",
      "service": "api"
    }
  ]
}
```
</details>


<details> <summary>Audit record after sync</summary>

```json
{
  "_id": "bfdd7097-ed29-43af-8a0d-3cc40904ebc0",
  "_rev": "6-b1dbf7dd18e5cea7dc8492b2aee88a47",
  "history": [
    {
      "rev": "2-25d468164345e5a19221115e5964d261",
      "request_id": "46d1a8e5bebd",
      "user": "adele.walter",
      "date": "2025-05-16T17:12:49.273Z",
      "service": "api"
    },
    {
      "rev": "3-58c4fec719640a8a753a6b2844f58857",
      "request_id": "ff2ee6cdf748",
      "user": "adele.walter",
      "date": "2025-05-16T17:18:01.050Z",
      "service": "api"
    },
    {
      "rev": "4-5d6b6372c6853160d944e93ac97c8588",
      "request_id": "ec9a921b1ef0",
      "user": "adele.walter",
      "date": "2025-05-16T17:21:15.680Z",
      "service": "api"
    },
    {
      "rev": "5-cd5fe25ba871db4645fba2584bb12df3",
      "request_id": "b5273bc43dfa",
      "user": "adele.walter",
      "date": "2025-05-16T17:24:13.724Z",
      "service": "api"
    },
    {
      "rev": "6-a2979343ee6fc3463db90e5868745efa",
      "request_id": "f96f63571a19",
      "user": "adele.walter",
      "date": "2025-05-16T17:32:45.991Z",
      "service": "api"
    },
    {
      "rev": "12-4b0b634161d3a00038e8c2d17967b31f",
      "request_id": "9d8da44eb1ad",
      "user": "adele.walter",
      "date": "2025-05-16T17:44:01.805Z",
      "service": "api"
    }
  ]
}
```
</details>

It was only added on 2025-05-16T17:44:01.805Z, when I synced, but I performed 5 or 6 editions to the same contact. I was trying to see the rotation mechanism.

lorerod:
I checked the DB sizes with and without audit enabled. The audit records are pretty small — for example, medic-audit had ~280 KB for 734 records, which feels reasonable compared to the 15.2 MB in medic. So I don’t see any significant overhead added by this change.

I agree it makes sense [to keep the feature simple for now](https://github.com/medic/cht-core/pull/9909#issuecomment-2854765589). Still, it might be worth keeping an eye on storage growth over time, just to be cautious and ensure it doesn’t become a concern down the line or for deployments with unexpected editing patterns.
<details> <summary>database size 4.19</summary>

Screenshot 2025-05-16 at 1 40 06 PM

</details>

<details> <summary>database size on this branch</summary>

Screenshot 2025-05-16 at 2 29 09 PM

</details>

lorerod:
@dianabarsan sorry for the separate messages, but I'm commenting as I test.

I noticed that the `_id` used for audit records in **tasks** is quite long and includes a lot of parts, like:
`task~org.couchdb.user:adele.walter~e4b36423-b4cd-4206-b475-968d7154f527~delivery-reminder~anc.delivery~1747413220361`
The audit record seems to reuse the document’s `_id` for **contacts** and **reports**, making searching easier, just wondering if there’s a cleaner way to generate the audit `_id` for these as well.

dianabarsan:
Hi @lorerod 

Thanks for the deep testing!

> I see that only one update is recorded in the history. Is this the expected behavior, or should all the individual updates be registered?

Yes, correct. This is the expected behavior, because only the last version from the client is uploaded to the server. 

> I noticed that the _id used for audit records in tasks is quite long and includes a lot of parts

The _id of the audit doc is **identical** to the original doc _id. Nothing is changed for tasks.

lorerod (approved):
Nothing else to add here. This LGTM. Thank you @dianabarsan

dianabarsan:
Thank you so much @lorerod ! I get a little anxious and unsure about merging these big changes, and it helps me a lot to get confirmation from someone whose judgement and technical opinions I trust and value. I appreciate your time and thorough testing, I'm confident about this now :D

PR #9924, opened 2025-04-24
chore(#9915): replace deprecated appearance fields in forms

### LINKED ISSUE #9915: Replace usage of `db-object` appearance with `select-contact`
**Describe the bug**
We should replace the `db-object` appearance  and use `select-contact` instead which supports multiple types of contacts as a follow-up to: https://github.com/medic/cht-conf/issues/502.

`db-object` instances in the XML forms in `/config` : https://pastebin.com/0pB1PmTN

Additional tip by @jkuester, originally mentioned [here](https://forum.communityhealthtoolkit.org/t/warn-when-using-deprecated-db-object-appearance/4890/4?u=apoorvapendse):

Just one tip when it comes to uplifting forms to use `select-contact`, you can find the docs [here](https://docs.communityhealthtoolkit.org/building/forms/configuring/form-inputs/#contact-selector). But basically for the form to work you will need to update both the `appearance` column (from `db-object` to `select-contact`) AND the `type` column. The type changes from `db:{{contact_type}}` to just be `string`. *Then*, you need to add an additional value in the `appearance` column (it is a space-separated list) that is `type-{{contact_type}}`. 

For example, if you start with:

|type|appearance|
| --- | --- |
|db:person|db-object|

This should be updated to:

|type|appearance|
| --- | --- |
|string|select-contact type-person|

**To Reproduce**
Using the branch in https://github.com/medic/cht-conf/pull/682 might be a good start, as that branch introduces the `db-object` deprecation warnings.

The steps will be similar to the ones mentioned in https://github.com/medic/cht-conf/issues/502
We should ensure that the `db-object` warnings are not displayed on running for the applicable forms.
```
cht --url=$(chtx instance ls <instance_name>) upload-app-forms -- <form_name>
```

**Expected behavior**
`db-object` usage is replaced with `select-contact` which can be tested running the form validation using a `cht-conf` version that introduces the `db-object` warning.

**Additional context**
Related Discussion: https://forum.communityhealthtoolkit.org/t/warn-when-using-deprecated-db-object-appearance/4890

apoorvapendse:
# Description
This refactor primarily replaces the `db-object`
appearance field with `select-contact`.
It also replaces two other deprecated appearance
fields namely `horizontal` and `horizontal-compact`.

Fixes #9915

# License


apoorvapendse:
@jkuester can you take a look when you find time?

jkuester (changes requested):
@apoorvapendse thanks for raising this PR!  I have gone through all the changes and they look correct.  The only problem is that we also need to include updates to the `.xslx` files.  Did you just manually update these `xml` files?  

The correct procedure is to update the configuration in the `.xlsx` form files and then use the cht-conf `convert-contact/app-forms` actions to update the `xml` files based on the changes to the `xlsx` files.  I think the only extra complication to look out for when updating the form `xlsx` files is that for the _place contact_ forms, instead of editing the individual create/edit form `xlsx` files directly, you can just edit the `PLACE_TYPE-create/edit.xlsx` files.

apoorvapendse:
Thanks for the review @jkuester!
>Did you just manually update these xml files?

Yes, pretty much. I automated a few parts, for instance, for the files that only contained `db:person`. 


> The correct procedure is to update the configuration in the .xlsx form files

So, something like this?
original
updated

jkuester:
@apoorvapendse exactly!  Yes.

apoorvapendse:
Hi @jkuester
I've updated the PR as per your suggestions, please take a look at your convenience.

>The correct procedure is to update the configuration in the .xlsx form files and then use the cht-conf convert-contact/app-forms actions to update the xml files based on the changes to the xlsx files. 

I kept the initial commit, which replaces the deprecated fields in `<place>-create/edit.xml` files.

jkuester (approved):
Most excellent!  Thank you for your attention to detail here!

PR #9928, opened 2025-04-25
chore(#9523): upgrade WebdriverIO to v9

lorerod:
# Description

This PR upgrades WebdriverIO from v8 to v9
closes https://github.com/medic/cht-core/issues/9523

**Main Changes**
- WebdriverIO version updates
- Removed `waitForClickable` before `.click()` calls, as WebdriverIO v9 now automatically waits for elements to become interactable. [Reference](https://webdriver.io/blog/2024/08/15/webdriverio-v9-release#automatic-waiting-for-elements-to-become-interactable)
- Simplified element access by removing double `await`, following the updated approach for accessing element properties. [Reference](https://webdriver.io/blog/2024/08/15/webdriverio-v9-release#accessing-element-properties)
- Replaced `isDisplayedInViewport` with `isDisplayed({ withinViewport: true })` as the former has been removed. [Reference](https://webdriver.io/blog/2024/08/15/webdriverio-v9-release#removal-of-isdisplayedviewport)
- Using WebDriver Classic mode by setting `'wdio:enforceWebDriverClassic': true` in capabilities, as BiDi support is currently unstable. I'm getting [this](https://github.com/webdriverio/webdriverio/issues/14430) error.

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9523-upgrade-webdriver-v9/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9523-upgrade-webdriver-v9/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9523-upgrade-webdriver-v9/docker-compose/cht-couchdb-clustered.yml)

# License


lorerod:
Hi @jkuester and @dianabarsan 
I’ve added you both as reviewers — feel free to take a look whenever you have time. No pressure at all; whoever gets to it first can go ahead with the review. I expect just one review. Maybe just leave a 👍 here when you start, so we avoid double effort. Thanks so much!

tests/e2e/default/enketo/extension-lib-form.wdio-spec.js
```diff
@@ -16,6 +16,11 @@ describe('Extension lib xpath function', () => {
     await commonPage.reloadSession();
   });
 
+  after(async () => {
+    await utils.deleteDocs(['form:extension-lib-average-calculator']);
+    await utils.revertDb([/^form:/], true);
```
dianabarsan, line 21: I believe that `utils.revertDb` is already called in the generic `after` defined here: https://github.com/medic/cht-core/blob/master/tests/wdio.conf.js#L336 SO this should not be be necessary.

tests/e2e/default/enketo/enketo-widgets.wdio-spec.js
```diff
@@ -83,6 +83,12 @@ describe('Enketo Widgets', () => {
     await loginPage.login(offlineUser);
   });
 
+  after(async () => {
+    await utils.deleteUsers([offlineUser]);
```
dianabarsan, line 87: Deleting users and calling `revertDb` is already part of the generic after defined here: https://github.com/medic/cht-core/blob/master/tests/wdio.conf.js#L336

I think you would only need to remove the form. Even though the generic revertdb should already cover that too.

tests/e2e/default/enketo/pregnancy-visit.wdio-spec.js
```diff
@@ -26,6 +27,26 @@ describe('Pregnancy Visit', () => {
...
   });
 
+  afterEach(async () => {
+    await commonPage.goToReports();
+    const reports = await reportsPage.reportsListDetails();
+    for (const report of reports) {
+      if (report.heading === pregnantWoman.name) {
+        await utils.deleteDocs([report.dataId]);
+      }
+    }
```
dianabarsan, line 37: Can we somehow make this more robust? Like save the doc ids when they are created (somehow) instead of going over this visible list after the test?

tests/page-objects/default/common/common.wdio.page.js
```diff
@@ -65,38 +65,32 @@ const userSettingsSelectors = {
 const getJsonErrorText = async () => await $('pre').getText();
 
 const isHamburgerMenuOpen = async () => {
-  return await (await $('mat-sidenav-container.mat-drawer-container-has-open .mat-drawer-opened')).isDisplayed();
+  return await hamburgerMenuSelectors.closeSideBarMenu().isDisplayed();
```
dianabarsan, line 68: Nice!

tests/page-objects/default/common/common.wdio.page.js
```diff
@@ -146,9 +140,22 @@ const hideSnackbar = () => {
...
+  const loaders = await $$('.loader').getElements();
+  if (loaders.length) {
+    // Add a small pause to let the DOM stabilize
+    await browser.pause(100);
+    // Instead of iterating through the loaders array, query for each loader individually
+    // This avoids issues with stale references
+    for (let i = 0; i < loaders.length; i++) {
+      // Use a more specific selector to get a fresh reference to each loader
+      // This avoids the index out of bounds issue
+      const loaderSelector = `.loader:nth-child(${i + 1})`;
```
dianabarsan, line 154: I don't think this will work. 
`nth-child` matches elements based on the indexes of the elements in the child list of their parents (reference: https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child)
So this won't be matching your 4th existent loader, it will be matching _a loader_ that is the 4th element in its parent.

tests/page-objects/default/tasks/tasks.wdio.page.js
```diff
@@ -9,20 +9,47 @@ const FORM_TITLE_SELECTOR = `${TASK_FORM_SELECTOR} h3#form-title`;
...
+      if (!(await task.isDisplayed())) {
+        return false;
+      }
+    }
+    return true;
+  }, {
+    timeout: 10000,
+    interval: 500,
+    timeoutMsg: 'Expected task elements to be present and displayed'
+  });
```
dianabarsan, line 36: Can this be pulled out into its own function "waitForTaskLoading" ?

tests/page-objects/default/tasks/tasks.wdio.page.js
```diff
@@ -9,20 +9,47 @@ const FORM_TITLE_SELECTOR = `${TASK_FORM_SELECTOR} h3#form-title`;
...
+    timeoutMsg: 'Expected tasks list to be displayed'
+  });
+
+  await browser.waitUntil(async () => {
+    const tasks = await $$(`${TASK_LIST_SELECTOR} li.content-row`);
+    if (!tasks.length) {
+      return false;
+    }
+    for (const task of tasks) {
+      if (!(await task.isDisplayed())) {
```
dianabarsan, line 27: I'm curious.. how can we have a li.content-row element that we just got a reference for _not_ be displayed?

dianabarsan (changes requested):
Awesome work!! 

I think I left one question and found one issue. My biggest comment is about all the new `after` test hooks, where I wish we could leverage the generic `after` hook more instead.

tests/page-objects/default/common/common.wdio.page.js
lorerod: I changed it to `nth-of-type`, what do you think about it? What I understand about this selector is for example: `.loader:nth-of-type(1) ` would select the first loader because it's the 1st element with class "loader"

lorerod: Thank you so much for catching this.

tests/page-objects/default/tasks/tasks.wdio.page.js
lorerod: I’ve already simplified it significantly based on your earlier suggestion to remove the redundant isDisplayed() checks on individual tasks. Additionally, I combined the two waitUntil blocks into one. Let me know what you think.

tests/page-objects/default/tasks/tasks.wdio.page.js
lorerod: Thanks! 
I removed the redundant` isDisplayed()` checks on individual tasks. Additionally, I combined the two `waitUntil` blocks into one.

lorerod:
Thank you, @dianabarsan, for your review.
It’s ready for another round. Kindly take a look at the changes when you have a moment.

tests/e2e/default/enketo/pregnancy-visit.wdio-spec.js
```diff
@@ -19,13 +19,25 @@ describe('Pregnancy Visit', () => {
     date_of_birth: moment().subtract(25, 'years').format('YYYY-MM-DD'),
     parent: { _id: healthCenter._id, parent: healthCenter.parent }
   });
+  const DOCS_TO_KEEP = [
+    ...Array.from(places.values()).map(doc => doc._id),
+    pregnantWoman._id,
+    'fixture:user:user1',
+    'org.couchdb.user:user1',
+    [/^form:/],
+  ];
```
dianabarsan, line 28: I sometimes wish we would have a pre-test snapshot of the docs and the after test would just revert to that snapshot. 
Not requesting a change here, just daydreaming.

tests/e2e/default/enketo/unauthorized-form.wdio-spec.js
```diff
@@ -34,6 +34,10 @@ describe('Unauthorized form', () => {
     await loginPage.login(offlineUser);
   });
 
+  after(async () => {
+    await utils.revertSettings(true);
```
dianabarsan, line 38: This should not be required, as the generic `after` also reverts settings.

dianabarsan (approved):
Awesome work! I think I just left some minor comment inline.

tests/page-objects/default/common/common.wdio.page.js
dianabarsan: Sweet! I think `nth-of-type` is the way to go here!

PR #9937, opened 2025-04-30
feat(#9936): add user-agent for all outgoing RapidPro requests

### LINKED ISSUE #9936: Add user-agent header to RapidPro requests
A recent change in TextIt (or all RapidPro) blocks all requests that are lacking a `user-agent` header. We have recently added a `user-agent` header to outbound requests, and this needs to be applied to rapidpro requests as well. 

Add `user-agent` header to RapidPro requests. 

Add user-agent to all outgoing requests, not just rapid-pro. 

**Additional context**
Reported on [the forum](https://forum.communityhealthtoolkit.org/t/user-agent-header-in-rapidpro-integration-code/4928/11)

shashi-sah2003:
This pull request introduces a new feature for automatically adding a `user-agent` header to HTTP requests in the `couch-request` library, while also removing redundant code from the `outbound` library. The changes include implementing a `getUserAgent` function, modifying request handling to include the `user-agent` header, and adding corresponding unit tests to ensure proper functionality.

fixes #9936 
### Changes to `couch-request` library:

* **Added `user-agent` header generation and inclusion**:
  - Introduced a `getUserAgent` function to dynamically create a `user-agent` string based on the CHT version, platform, and architecture. (`shared-libs/couch-request/src/couch-request.js`, [shared-libs/couch-request/src/couch-request.jsR29-R35](diffhunk://#diff-7c51430a26d41894740c0f40a3f2477eeea8b9d56e3adb60075f647cdecf9cb4R29-R35))
  - Updated the `setRequestOptions` function to include the `user-agent` header if it is not already set. (`shared-libs/couch-request/src/couch-request.js`, [shared-libs/couch-request/src/couch-request.jsL129-R150](diffhunk://#diff-7c51430a26d41894740c0f40a3f2477eeea8b9d56e3adb60075f647cdecf9cb4L129-R150))
  - Ensured `setRequestOptions` is awaited in the `request` function to handle the asynchronous `getUserAgent` call. (`shared-libs/couch-request/src/couch-request.js`, [shared-libs/couch-request/src/couch-request.jsL186-R200](diffhunk://#diff-7c51430a26d41894740c0f40a3f2477eeea8b9d56e3adb60075f647cdecf9cb4L186-R200))

* **Added unit tests**:
  - Mocked dependencies (`os` and `@medic/environment`) to simulate platform and version data. (`shared-libs/couch-request/test/couch-request.js`, [shared-libs/couch-request/test/couch-request.jsR36-R41](diffhunk://#diff-8406c6aa90be792d652a8e1a0df0824f3923709dba9ecdb690406cde25c2894bR36-R41))
  - Added tests to verify that the `user-agent` header is automatically added and does not override an existing `user-agent` header. (`shared-libs/couch-request/test/couch-request.js`, [shared-libs/couch-request/test/couch-request.jsR715-R738](diffhunk://#diff-8406c6aa90be792d652a8e1a0df0824f3923709dba9ecdb690406cde25c2894bR715-R738))

### Changes to `outbound` library:

* **Removed redundant `user-agent` code**:
  - Deleted the `getUserAgent` function and the associated `CHT_AGENT` constant, as this functionality is now centralized in the `couch-request` library. (`shared-libs/outbound/src/outbound.js`, [[1]](diffhunk://#diff-4cd39b6dcdad1347e201fe0deaad5a6f78e739ce2c38078ffdf07a23d67bd49dL28) [[2]](diffhunk://#diff-4cd39b6dcdad1347e201fe0deaad5a6f78e739ce2c38078ffdf07a23d67bd49dL98-L103)
  - Removed the inclusion of the `user-agent` header in the `sendPayload` function, delegating this responsibility to the `couch-request` library. (`shared-libs/outbound/src/outbound.js`, [shared-libs/outbound/src/outbound.jsL186-L188](diffhunk://#diff-4cd39b6dcdad1347e201fe0deaad5a6f78e739ce2c38078ffdf07a23d67bd49dL186-L188)

# License


shared-libs/couch-request/src/couch-request.js
```diff
@@ -1,10 +1,12 @@
 const environment = require('@medic/environment');
 const path = require('path');
+const os = require('os');
 let asyncLocalStorage;
 let requestIdHeader;
 
 const JSON_HEADER_VALUE = 'application/json';
 const CONTENT_TYPE = 'content-type';
+const CHT_AGENT = 'CHT';
```
mrjones-plip, line 9: @dianabarsan - what do you think about this suggestion?

```suggestion
const CHT_AGENT = 'Community Health Toolkit';
```

My thinking is that the if the CHT is misconfiguration and starts accidentally spamming a remote system and the operator there wants to know what `CHT` is it might be hard to look up.  `Community Health Toolkit` will more quickly lead them us.

Don't feel strongly about this!

shashi-sah2003: Hey @mrjones-plip  Yes, I think we can have this for more clarity at operator end.

andrablaj: @shashi-sah2003, if you agree with @mrjones-plip's suggestion, please commit it and resolve this conversation! Thank you!

andrablaj:
@shashi-sah2003 the `Compile the app` action fails with:

```
/home/runner/work/cht-core/cht-core/shared-libs/outbound/src/outbound.js
  19:7  error  'os' is assigned a value but never used           no-unused-vars
  21:7  error  'environment' is assigned a value but never used  no-unused-vars
```

Please make the suggested changes accordingly and we can rerun it. Thanks!

shashi-sah2003:
> @shashi-sah2003 the `Compile the app` action fails with:
> 
> ```
> /home/runner/work/cht-core/cht-core/shared-libs/outbound/src/outbound.js
>   19:7  error  'os' is assigned a value but never used           no-unused-vars
>   21:7  error  'environment' is assigned a value but never used  no-unused-vars
> ```
> 
> Please make the suggested changes accordingly and we can rerun it. Thanks!


@andrablaj made the necessary changes, it should work now

andrablaj:
@shashi-sah2003 the `Compile the app` action if unfortunately failing again. Please follow the documentation to [build the app](https://docs.communityhealthtoolkit.org/community/contributing/code/core/build-commands/) and [test it locally](https://docs.communityhealthtoolkit.org/community/contributing/code/core/automated-tests/) and reach out if you encounter any blocker.

shashi-sah2003:
@andrablaj there was an issue with circular dependency so I have modified the codebase to use lazy loading i.e. importing environment inside function block. I think it should be able work now.

shashi-sah2003:
@witash @mrjones-plip @andrablaj I can see the issue is not related to `environment` but is related to something else `moment.js`
I think I would need help on this debugging

witash:
@shashi-sah2003 I don't think lazy loading is necessary here; running this branch locally, I didn't have any circular dependency issues after reverting `fe633d2d02c722693e6405719539c13d9c883658`?

But what I did get was test failures. These are kind of hard to see in the full ci-compile task because they are buried behind all the other tests, but if you run `npm run test` from `shared-libs/couch-request` and `shared-libs/outbound` it will run only the tests for those modules which makes errors easier to see.

witash:
it looks like the test failures are mostly because the tests in `shared-libs/couch-request/test/couch-request.js` use deeply equal assertions on the request headers. So we have to add the user agent to most of those tests (looks like you already did this while I was typing 🙂).

Then, for the tests in `shared-libs/outbound/test/outbound.spec.js`, remove the user-agent header (its testing what is passed to `couch-request`, not what is actually sent, so it's ok to remove it). And then there is one small change on line 107
```
    sendOptions.headers = {
      'authorization': value
    }
```
because empty `headers` was removed from `sendPayload`

shashi-sah2003:
hey @witash thank you so much for helping me out. It was a silly mistake from my end. I had to add `user-agent` headers in `couch-request.js` and remove it from `outbound.js`. 
I have tested it locally and confirm that these two files are not throwing any error.
thanks

shashi-sah2003:
@witash why it is throwing error `non-existent function` for the function which exists already?

shashi-sah2003:
@witash we can implement try and catch block inside `getUserAgent` function and in the catch block we can define a default version number? but what would be the version number? any idea? I'll implement it.

witash:
@shashi-sah2003 I see now the problem is its not just the circular require; the functions actually call each other. `getVersion` will call `getUserAgent`, which will call `getVersion` again, which will call `getUserAgent` again.

Maybe instead
1. call require from `getDeployInfo` to avoid the circular require, and then leave the require at the top in couch-request. couch-request is used in a lot of different places, so lazy loading `environment` there could get complicated. But `getDeployInfo` is used only in a few places, is already a bit weird, and caches its result so the lazy load will only happen once. So maybe we can live with the lazy loading there?
2. set a dummy user agent in `getDeployInfo`. This will prevent `getUserAgent` from being called in `getVersion` (because it doesn't try to overwrite existing user agent) and so prevent the infinite loop.

OR

omit the version from the user-agent.
the circular dependency would still be there because `couch-request` still requires environment for something else.
But, then we could remove the dependency of `environment` on `couch-request` by moving `getDeployInfo` back to `api/src/services/deploy-info.js`. It was moved to environment in 9bcc58a5b8 so that it would be accessible in shared-libs for the User-Agent; if thats not required anymore, we can just move it back.

shashi-sah2003:
@witash thanks for your suggestion. I have implemented your first suggestion by lazy loading inside `getDeployInfo` and tested it locally by running `npm run ci-compile` and it's working perfectly fine.

shashi-sah2003:
@andrablaj @witash now the compile test is successfully passing.

shared-libs/couch-request/test/couch-request.js
```diff
@@ -699,13 +726,45 @@ describe('couch-request', () => {
...
         servername: 'test.com',
         uri: 'http://test.com:5984/b',
       }
     ]]);
   });
 
+  it('should automatically add user-agent header to requests', async () => {
+    const getUserAgent = couchRequest.__get__('getUserAgent');
+    
+    // Temporarily restore the NODE_ENV to allow user-agent header for this test only
```
witash, line 740: tests passed for me locally without this, is it necessarry?

shared-libs/environment/src/index.js
```diff
@@ -78,8 +77,13 @@ const getDeployInfo = async () => {
   }
 
   try {
+    // Lazy load couch-request only when needed
+    const request = require('@medic/couch-request');
     const ddoc = await request.get({
       url: `${couchUrl}/_design/${module.exports.ddoc}`,
+      headers: {
+        'user-agent': 'Community Health Toolkit/4.18.0 (test-platform,test-arch)',
```
witash, line 85: Sorry to nitpick, but can we change this so it doesn't say 4.18.0? It doesn't matter what it is, but would still prefer it to not be inaccurate

```suggestion
        'user-agent': 'Community Health Toolkit',
```

shared-libs/environment/test/index.spec.js
```diff
@@ -116,7 +116,16 @@ describe('environment', () => {
           timestamp: '2023-01-02'
         });
         expect(request.get.callCount).to.equal(1);
-        expect(request.get.args).to.deep.equal([[{url: 'http://admin:pass@localhost:5984/medicdb/_design/medic'}]]);
+        expect(request.get.args).to.deep.equal([
+          [
+            {
+              url: 'http://admin:pass@localhost:5984/medicdb/_design/medic',
+              headers: {
+                'user-agent': 'Community Health Toolkit/4.18.0 (test-platform,test-arch)'
```
witash, line 124: if accepting the above suggestion, this would also change
```suggestion
                'user-agent': 'Community Health Toolkit'
```

witash (changes requested):
@shashi-sah2003 looks good. The latest ci failures seem to be just flaky tests, they passed after re-running them.

Left a few nit picks but otherwise looks good, thanks for pushing through the unexpected complications for what should have been a quick fix

shared-libs/couch-request/test/couch-request.js
shashi-sah2003: @witash you're right it's supposed not to be there. It might have slipped my mind to remove that section of code
thanks for pointing out. I have committed the changes

shashi-sah2003:
@witash there is error in publish for testing `CERT_HAS_EXPIRED`. that's why tests are now failing.

PR #9939, opened 2025-05-01
fix(na): refactor copy static files to be futureproof

mrjones-plip:
# Description

This PR refactors the the [copy static file script](https://github.com/medic/cht-core/blob/master/scripts/build/copy-static-files.sh#L10) so that it is compatible with both GNU Coreutils and uutils coreutils implementations of the `cp` command.  Yes, here we are - believe it.  See [forum post](https://forum.communityhealthtoolkit.org/t/how-to-lint-on-the-command-line/4939/6) for more info.

the fix was to break out each recursive `cp` on to it's own line. 

If you have both versions of `cp` handy, you can prefer one vs the other by setting your `$PATH` correctly.  I had the uutils coreutils as default and the GNU one in `/usr/bin/cp`.  So I called this to make the GNU be called:

```
export PATH=/usr/bin/cp:$PATH
```

Then to test for regressions, call `./scripts/build/copy-static-files.sh`.  Success should show this when using GNU or uutils `cp`:

```
copy-static-files: copying
copy-static-files: done
```

Failure should show this only when using uutils `cp`:

```
copy-static-files: copying
cp: will not overwrite just-created 'api/build/static/admin/fonts' with 'webapp/src/fonts'
```

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:refactor-copy-static-files/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:refactor-copy-static-files/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:refactor-copy-static-files/docker-compose/cht-couchdb-clustered.yml)

# License


jkuester (approved):
:grapes:

mrjones-plip:
that's just the way I roll apparently?!?!?  :bug:

PR #9944, opened 2025-05-03
fix(#9739): deselecting "Select all" deselects all reports

### LINKED ISSUE #9739: Can't deselect all reports after selecting all reports
**Describe the bug**
Webapp has a checkbox to select all reports. Deselecting that checkbox should automatically deselect all reports. But that does not happen. 

**To Reproduce**

1. Login as an offline user with at least 50 reports. 
1. Go to reports page. 
2. Select all report. Wait until reports are selected.
3. Deselect the "Select all" checkbox.
4. All reports are still selected.

**Expected behavior**
Deselecting "Select all" should deselect all reports. 

**Logs**
No errors. 

**Environment**
- App: webapp
- Version: 4.15 (haven't checked older versions)

anurag2787:
# Description

The issue was happening because of scroll loading when we select all reports using checkbox it only selects the reports that are loaded but many reports are still not loaded due to lazy loading so when we uncheck the select all checkbox it doesn't actually deselect anything.

To fix this i added a change so that when select all is clicked it fetches all reports first. That way deselecting works properly on the full list.

### ISSUE NUMBER
Closes #9739 

### Before Changes 
https://github.com/user-attachments/assets/8fe7df15-71e2-4845-be0c-78538ab68b10

### After Changes
https://github.com/user-attachments/assets/4b95a26b-c7aa-4e02-a361-251a7b64644b

# License


anurag2787:
Hy @dianabarsan I have fixed the issue, please review Thank you!

anurag2787:
Hi @dianabarsan I have fixed the sonar issue to reduce cognitive complexity and updated to use nullish coalescing.

webapp/src/ts/modules/reports/reports.component.ts
```diff
@@ -465,30 +465,65 @@ export class ReportsComponent implements OnInit, AfterViewInit, OnDestroy {
...
-      const preparedReports = await this.prepareReports(reports, true);
-      this.reportsActions.setSelectedReports(preparedReports);
-      this.globalActions.unsetComponents();
-
+      await this.finalizeReportSelection();
     } catch (error) {
       console.error('Error selecting all', error);
     }
   }
+
```
dianabarsan, line 481: I'm slightly concerned about the performance of this approach, because it effectively loads the same reports _twice_, once for the LHS and once for the contents. 
Is there any way we can show a loading icon and disable the deselection of the checkbox while the reports are loading, instead of loading the whole lhs list?

dianabarsan (changes requested):

webapp/src/ts/modules/reports/reports.component.ts
anurag2787: I’ve implemented the change so that when the Select All checkbox is clicked it gets disabled once all the reports are fully loaded the checkbox is re-enabled and the Deselect All option becomes visible. I’ve attached a quick video demo how this works. 

Let me know if this approach works for you.

Thanks!


https://github.com/user-attachments/assets/7ce3b8bf-f0b7-4aed-bd07-b80f16969fbf

dianabarsan: Looking at the code, we are _still_ loading the reports twice, once in `loadRemainingReports` and once in `finalizeReportSelection`. This is a serious performance concern. 

The solution should avoid doing reports search twice.

anurag2787:
Hy @dianabarsan i have pushed my code please review. Implemented the change so that when the Select All checkbox is clicked it gets disabled once all the reports are fully loaded the checkbox is re-enabled

webapp/src/ts/modules/reports/reports.component.ts
```diff
@@ -468,7 +469,7 @@ export class ReportsComponent implements OnInit, AfterViewInit, OnDestroy {
     if (this.areAllReportsSelected()) {
       return;
     }
-
+    this.isSelectingAllReports =true
```
dianabarsan: Please fix linting here: 
```suggestion
    this.isSelectingAllReports = true;
```

webapp/src/ts/modules/reports/reports.component.ts
```diff
@@ -533,6 +534,10 @@ export class ReportsComponent implements OnInit, AfterViewInit, OnDestroy {
     }
 
     const isMaxReportsSelected = this.selectedReports?.length >= this.LIMIT_SELECT_ALL_REPORTS;
-    return isMaxReportsSelected || this.reportsList?.length === this.selectedReports?.length;
+    const check = isMaxReportsSelected || this.reportsList?.length === this.selectedReports?.length;
+    if(check){
+      this.isSelectingAllReports = false
+    }
+    return check;
```
dianabarsan: Please fix linting here: 
```suggestion
    const finishedSelecting = isMaxReportsSelected || this.reportsList?.length === this.selectedReports?.length;
    this.isSelectingAllReports = !finishedSelecting;
    return finishedSelecting;
```

This also has a renaming suggestion and reduces code depth.

dianabarsan (changes requested):
Nice thinking about the indeterminate state! I think that it might be the nicest solution technically, but I'm not sure if it communicates to nontechnical users that they should not press the select again. 
I haven't tested how it works yet, ideally when indeterminate clicking it will do nothing (neither stop or start a selection), but I'm seeing some linting issues that will fail your build once you start it.

anurag2787:
Hi, i have fixed linting issues and renamed it as per your suggestion. Please review it whenever you get time Thanks!

andrablaj:
@anurag2787 are you still planning to work on this issue? If yes, can you please review @dianabarsan's comments? If not, please close the PR. Thank you!

Issue #9760, opened 2025-01-29
Update `shared-libs/user-management` to accept an `oidc_provider` property for users


jkuester:
- The `oidc_provider` property should be set *instead of* `password` or `token_login`.  Cannot set both.
- Validate that the value provided for `oidc_provider` matches the `oidc_provider.client_id` from the app_settings.
- Set a random value for the `password` of the user.
- This will affect user create/update via the [/api/v2/users API](https://docs.communityhealthtoolkit.org/building/reference/api/#post-apiv2users)
- Update [integration tests](https://github.com/medic/cht-core/blob/58a5e7d901c5946c4cdebd23611bf13db1c35b3f/tests/integration/api/controllers/users.spec.js)

mrjones-plip:
@rmayore - I don't think I can assign you this ticket until you  comment on the ticket?  Please comment "I'll take this!" or something and I think I can assign it to you then.

rmayore:
Assign to me please @mrjones-plip

mrjones-plip:
Perfect - you commenting enabled me to assign - thanks Robert!

benkags:
> - Validate that the value provided for oidc_provider matches the oidc_provider.client_id from the app_settings.


We should consider removing this tight coupling with the client id for the simple reason that updating the client id necessitates updating the user and client id can change for reasons outside the control of the CHT.

jkuester:
Hmm, this is a good point.  😓   I was hoping to be able to avoid needing any kind of user migration if we add support for multiple oidc providers in the future and so wanted to be able to use the `client_id` to map to the oidc provider associated with that user.  However, I think you are right @benkags that we run the risk of requiring a migration of the user docs if we have to change the `client_id` value for some reason....

Do you think we should add another arbitrary property to the `oidc_provider` settings object (like `id` or `key`) that we could use to map the users (with the idea that this property should never change, but the client_id could still be updated without needing to update the user docs)?  Or, am I just over-architecting and over-complicating things here and we could just have a simple `oidc` boolean on the user docs to indicate they are SSO users?

mrjones-plip:
>  Or, am I just over-architecting and over-complicating things

This sounds like a feature for the next release and not the MVP!  The idea being that there will be a potential painful process to upgrade to N OIDC providers from the MVPs support of just 1 OIDC provider.

My vote is no extra complexity, but defer to both Josh and Bernard who are closer to the code and will appreciate the cost of the upgrade should we take the simple approach!

[...]

PR #9955, opened 2025-05-08
feat(#9760): add support for authenticating users via OIDC

### LINKED ISSUE #9735: Single sign on (SSO) using identity provider

jkuester:
# Description

Closes #9735

https://github.com/medic/cht-docs/pull/1875

This PR includes several interlocking changes all required for supporting SSO authentication via OIDC in the CHT.

1. Add api support for creating/updating users with `oidc_username` property.
    - `shared-libs/user-management` updated to support this property on requests made to the `/api/v?/users` endpoints. 
    - The property value is returned when retrieving users via the `/api/v?/users` endpoint.
    - When creating/updating a user, the `oidc_username` property cannot be combined with `password` or `token_login`.
        - Instead, when `oidc_username` is set, a random `password` value will be automatically generated for the user. 
    - The `oidc_username` value must be unique. There cannot be any existing users that already have the same value.
    - The server must be configured with `oidc_provider` in the app_settings.json when setting an `oidc_username` on a user.
1. Add support for creating/updating SSO users via the admin app UI
    - The create/update user modal in the admin app has been updated to include a new "SSO Email Address" text box.
    - The box is only displayed when `oidc_provider` is configured in the app_settings.json and the user does not have `token_login` enabled.
    - If a value is provided for "SSO Email Address", the user is not able to enter values for Token Login or the Password.
1. Add support for API endpoints to login as an OIDC user
    - `/medic/login/oidc/authorize` is used for starting the OIDC flow. Submit a GET request to this endpoint and a redirect url will be returned pointing to the OIDC provider with all the necessary query parameters.
    - `/medic/login/oidc` is used for finishing the OIDC flow and completing the login process. The OIDC provider will redirect the user back to this endpoint with all the necessary query parameters needed for finishing the OIDC flow.
        - Most importantly, the `authorization_code` query parameter should be provided (by the OIDC Provider) when calling this endpoint. The CHT api server will take this auth code and use it to make a back-channel connection to the OIDC Provider and get the user's id_token.
        - The id_token will include the user's `email` claim. The value for this `email` claim is then used to query for users with a matching `oidc_username`.
        - When one (and only one) user is found with a matching `oidc_username` value, the API server will generate a new Couch session cookie for that user (this mimics what happens during a normal logic process where the Couch `_session` endpoint is called with the user's credentials). 
        - The new Couch session cookie is provided on the response and from that point the user is validly logged in as a normal CHT user.
    - These endpoints can only be called when the server has been properly configured for SSO:
        - The `app_url` must be set in the app_settings.json
        - The `oidc_provider` object must be configured in the app_settings.json and include:
            - `discovery_url` - the URL of the OIDC provider's discovery endpoint
            - `client_id` - the unique identifier for the CHT server in the OIDC provider
        - The `client_secret` value for the CHT server from the OIDC provider must also be set in the CHT via the `/api/v1/credentials` endpoint.
1. Add support for showing a "Login with SSO" button on login page
    - When `oidc_provider` is configured in the app_settings.json, the normal CHT login page with be rendered with a "Login with SSO" button. 
    - Clicking this button will cause the user's browser to be automatically redirected to the OIDC Provider where they can authenticate before being automatically redirected back to the CHT app.
    - Selecting a locale on the login page will not affect the locale used in the app once the user has completed the OIDC login flow. Instead the locale used in the app will match the `locale` claim configured for the user in the OIDC provider (or will fall back to the default locale for the CHT instance). 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9735_sso/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9735_sso/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9735_sso/docker-compose/cht-couchdb-clustered.yml)

# License


.github/workflows/build.yml
```diff
@@ -374,7 +374,7 @@ jobs:
       if: ${{ failure() }}
 
   publish:
-    needs: [tests, config-tests, test-cht-form, translations]
+    needs: [tests, config-tests, test-cht-form]
```
jkuester, line 377: Proposed this change in a [separate PR](https://github.com/medic/cht-core/pull/9984). Leaving this here for now to keep image builds while collecting missing translation strings.

admin/src/js/controllers/edit-user.js
```diff
@@ -86,25 +87,41 @@ angular
...
           }
 
           // Start with the password masked.
-          $scope.model.passwordFieldType  = 'password';
+          $scope.model.passwordFieldType = 'password';
 
           $scope.model.showPasswordIcon = SHOW_PASSWORD_ICON;
           $scope.model.hidePasswordIcon = HIDE_PASSWORD_ICON;
-
+          $scope.model.oidc_username = oidcUsername;
```
jkuester, line 124: The other `$scope.model` properties here are coming from the `user-settings` document for the user being edited.  The `oidc_username` is stored on the user's doc in the `_users` db and not on the `user-settings` doc in the `medic` db.  So, to get the current `oidc_username` value we have to retrieve the user's data above via the `/api/v2/users` api.

api/resources/translations/messages-en.properties
```diff
@@ -1307,6 +1309,9 @@ user.password.current = Current password
 user.phone.help = Must be a valid phone number without dashes or punctuation.
 user.place.help = All users must be associated with a place in the system to be correctly located in the hierarchy and see appropriate data when they log in.
 user.username.help = This is what you will use to log in to the app.
+user.sso.username = SSO Email Address
```
jkuester, line 1312: Note that we are intentionally calling this field in the admin UI "SSO Email Address" even though it populates the `oidc_username` field on the backend.  This is because for the current MVP implementation, this value _must_ map to the `email` claim value provided by the OIDC provider.  It is possible we could support different mapping in the future, but for now it seems most clear to just call this an "Email Address".

package.json
```diff
@@ -43,7 +43,7 @@
     "test": "npm run lint && npm run unit && npm run integration-api",
     "unit": "node scripts/build/cli npmCiModules && npm run unit-webapp && npm run unit-admin && npm run unit-shared-lib && npm run unit-api && npm run unit-sentinel",
     "unit-admin": "node ./scripts/ci/run-karma.js",
-    "unit-api": "UNIT_TEST_ENV=1 nyc --nycrcPath='api/nyc.config.js' mocha 'api/tests/mocha/**/*.js'",
+    "unit-api": "UNIT_TEST_ENV=1 nyc --nycrcPath='api/nyc.config.js' mocha 'api/tests/mocha/services/sso-login.spec.js' 'api/tests/mocha/**/*.js'",
```
jkuester, line 46: This is more weirdness caused by `openid-client` being an ES Module. :disappointed: As [described here](https://github.com/medic/cht-core/pull/9833#discussion_r2049603474), there is some weird behavior when combining `require`ing an ESM + `nyc` + mocha tests _with file globbing_.  Explicitly targeting `sso-login.spec.js` works around this issue (and all the unit tests are still only run once).

shared-libs/user-management/src/token-login.js
```diff
@@ -193,8 +194,7 @@ const generateTokenLoginDoc = (user, userSettings, token, appUrl) => {
       tasks: [],
     };
 
-    appUrl = (config.get('app_url') || appUrl).replace(/\/+$/, '');
-    const url = `${appUrl}/medic/login/token/${token}`;
+    const url = `${appUrl.replace(/\/+$/, '')}/medic/login/token/${token}`;
```
jkuester, line 197: FTR, I doulbe/triple checked to confirm that the `appUrl` being passed in here is populated from `config.get('app_url')`. So, the duplicate call here is unnecessary.

webapp/tests/karma/karma-unit.base.conf.js
```diff
@@ -46,7 +46,7 @@ module.exports = {
         statements: 87,
         lines: 87,
         branches: 79,
-        functions: 86,
+        functions: 85,
```
jkuester, line 49: Open to push-back here, but I wasted a bunch of hours trying to figure out how the actual coverage value from the tests went down from `86` to `85` when in reality all I did was add _more_ test coverage.  

I think there is perhaps some unexpected impact from the `skipFilesWithNoCoverage` config above since my added unit tests were for files that previously not covered at all.  Maybe somehow that equates to more test coverage equaling less function coverage percentage... :sweat_smile: :facepalm:. 

Either way, I am confident the changes in this PR _improve_ the test coverage in webapp and so am not too worried about updating this value (particularly since these values were only [recently configured](https://github.com/medic/cht-core/commit/6b78651b9d7a2af7d531992c1b98ba9c19ee3c61), so perhaps we are still discovering the nuances here...)

admin/tests/unit/controllers/edit-user.spec.js
```diff
@@ -159,6 +161,11 @@ describe('EditUserCtrl controller', () => {
   const mockContactGet = facilityId => {
     dbGet.resolves({ parent: { _id: facilityId } });
   };
+  const editUser = async () => {
+    await scope.editUser();
+    // Allow time for async validations to complete
+    await timeout.flush(2000);
```
jkuester, line 167: It seems like the additional Promise I added to the edit-user flow pushed the tests over the edge of some race-condition where previously the async validations were magically completing before the test tried to assert anything.

Now some assertions around the async validates were failing because the validates did not "have time" to execute.  With this flush, the async queue is cleared and everything can be properly asserted.

witash (approved):
tested locally with Keycloak and allow_insecure_requests, docs were easy to follow.
LGTM, no comments.

dianabarsan:
This work is amazing! Excellent job! Appreciate all the collaboration to make this happen. Thanks @jkuester for an outstanding job at leading this work into this clean, concise and maintainable state!

api/src/services/sso-login.js
```diff
@@ -0,0 +1,201 @@
...
+ *  
+ * @param {string} username Username.
+ * @param {string} salt CouchDB user salt value.
+ * @param {string} secret  CouchDB secret value.
+ * @param {number} authTimeout Cooke timeout.
+ * @returns {string} Cookie string.
+ */
+const makeCookie = (username, salt, secret, authTimeout) => {
+  // an adaptation of https://medium.com/@eiri/couchdb-cookie-authentication-6dd0af6817da
+  const expiry = Math.floor(Date.now() / 1000) + authTimeout;
```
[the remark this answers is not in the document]
mrjones-plip: woah - nice catch!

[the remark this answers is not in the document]
jkuester: Okay, did some more testing here and was able to confirm my theory.  If I generated the cookie with a date that was _more than a year in the past_, the login process would fail because the generated cookie was actually expired.  If I generated the cookie with a date _slightly less than one year in the past_ the login process would succeed, but subsequent Couch requests with the cookie would result in a new cookie being generated (by Couch) and set in the browser session (presumably because the existing cookie was about to expire).  Finally, if I generate the cookie with the current date, the login is successful and the cookie is not immediately replaced by Couch.

So, based on this behavior and on the date values I observe being included on the normal Couch AuthSession cookies, I am confident that we should just be setting the _current date_ on the new cookie and not the current date + `couch_httpd_auth/timeout` value.  When Couch validates the cookie, it is comparing the date in the cookie against the `couch_httpd_auth/timeout` value.

jkuester:
@witash could I get a quick re-review of the latest changes (particularly of the shift in the cookie logic [described here](https://github.com/medic/cht-core/pull/9955#discussion_r2105094871))?  :pray:

api/src/services/sso-login.js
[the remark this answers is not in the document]
benkags: If we set the current date as expiry date in the cookie, doesn't that mean that the next call to CouchDB will use an expired cookie? What are we missing here?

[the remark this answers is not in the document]
jkuester: Excellent question! (And this is why I want to make sure we all take a second and have a closer look at this since it does have an important impact on authentication.  

My interpretation of things is that the date on the cookie is the _issued date_ and NOT the _expiration date_.  Couch will treat a cookie as "expired" if its _issued date_ is BEFORE now - `couch_httpd_auth/timeout`.  (Which is why I could use a cookie with a date in the past as long as the date was less than one year in the past.)  Let me know if that makes sense or if something still seems off here. Now is def the time to be confident in our approach here!

Issue #9974, opened 2025-05-20
Support opening contact edit form from task


ChinHairSaintClair:
**What feature do you want to improve?**
Enable editing a contact directly from a task. 
Currently, while the CHT allows creating a new contact based on type via a task, it does not support editing an existing contact directly through a task entry. This enhancement would build on existing functionality and streamline the "follow-up" process where one has to maintain patient information.

**Describe the improvement you'd like**
Setting an `action`'s `type` to "contact" and unpacking the entire contents of the linked contact into the `content` variable, facilitated by `modifyContent`, should allow for editing that contact directly.

**Describe alternatives you've considered**
Create an "intermediary" app form that gets opened by the task, passing in the relevant contact `_id`. This form contains a "button", essentially an anchor tag pointing to `contacts/<your_contact_id>/edit`.

Here's an example layout:
| type              | name            | label          | calculation                                                               |
| -------------- | ------------- | ------------ | ------------------------------------------------------ |
| begin group | inputs           |                   |                                                                                 |
| begin group | contact         |                   |                                                                                 |
| hidden          | _id                |                   |                                                                                 |
| hidden          | name            |                   |                                                                                 |
| end group    | contact         |                   |                                                                                 |
| end group    | inputs           |                   |                                                                                 |
| calculate       | _id                |                   | ../inputs/contact/_id                                                |
| calculate       | link               |                   | concat(concat("#/contacts/", string(../_id)),"/edit") |

The "redirect" `note`, placed below the `link` entry, displays as a button with the following HTML content:
```
<a target="_self" href="#" rel="noopener" class="dynamic-url btn btn-primary" style=”display:inline-block;">
    Got to "${name}"
    <div class=”url hidden">${link}</div>
</a>
```

Problems with this approach:
1. It is unintuitive and difficult to train for CHWs.
2. It allows for unintended submission of intermediary forms, cluttering the reports view.
3. It loses the ability to use `modifyContent` with the intended `contact` form.

**Additional context**
https://forum.communityhealthtoolkit.org/t/write-custom-transition-and-edit-contact-form-with-task/3334
https://forum.communityhealthtoolkit.org/t/temporarily-increase-future-task-180-limit/4026

PR #9975, opened 2025-05-20
feat(#9974): support opening contact edit form from task

ChinHairSaintClair:
# Description

Allow editing of a contact via the linked task entry. The CHT already allows creating a new contact based on type through a task; this builds on that functionality.

Closes https://github.com/medic/cht-core/issues/9974

# Code review checklist

# License

### DIFF
--- a/webapp/src/ts/modules/tasks/tasks-content.component.ts
+++ b/webapp/src/ts/modules/tasks/tasks-content.component.ts
@@ -304,8 +304,8 @@ export class TasksContentComponent implements OnInit, OnDestroy {
     }
 
     if (action.type === 'contact') {
-      if (action.content?._id){
-        this.router.navigate(['/contacts', action.content._id, 'edit']);
+      if (action.content?.id){
+        this.router.navigate(['/contacts', action.content.id, 'edit']);
       } else if (action.content?.parent_id) {
         this.router.navigate(['/contacts', action.content.parent_id, 'add', action.content.type || '']);
       } else {
--- a/webapp/src/ts/modules/tasks/tasks-content.component.ts
+++ b/webapp/src/ts/modules/tasks/tasks-content.component.ts
@@ -304,7 +304,7 @@ export class TasksContentComponent implements OnInit, OnDestroy {
     }
 
     if (action.type === 'contact') {
-      if (action.content?.parent_id) {
+      if (action.content?.parent_id && action.content?.type) {
         this.router.navigate(['/contacts', action.content.parent_id, 'add', action.content.type || '']);
       } else if (action.content?.type) {
         this.router.navigate(['/contacts', 'add', action.content.type]);
--- a/webapp/src/ts/modules/tasks/tasks-content.component.ts
+++ b/webapp/src/ts/modules/tasks/tasks-content.component.ts
@@ -304,12 +304,14 @@ export class TasksContentComponent implements OnInit, OnDestroy {
     }
 
     if (action.type === 'contact') {
-      if (action.content?.id){
-        this.router.navigate(['/contacts', action.content.id, 'edit']);
-      } else if (action.content?.parent_id) {
+      if (action.content?.parent_id) {
         this.router.navigate(['/contacts', action.content.parent_id, 'add', action.content.type || '']);
+      } else if (action.content?.type) {
+        this.router.navigate(['/contacts', 'add', action.content.type]);
+      } else if (action.content?.edit_id) {
+        this.router.navigate(['/contacts', action.content.edit_id, 'edit']);
       } else {
-        this.router.navigate(['/contacts', 'add', action.content?.type || '']);
+        this.router.navigate(['/contacts', action.content?.contact?._id ?? '']);
       }
     }
   }
--- a/webapp/src/ts/modules/tasks/tasks-content.component.ts
+++ b/webapp/src/ts/modules/tasks/tasks-content.component.ts
@@ -304,7 +304,9 @@ export class TasksContentComponent implements OnInit, OnDestroy {
     }
 
     if (action.type === 'contact') {
-      if (action.content?.parent_id) {
+      if (action.content?._id){
+        this.router.navigate(['/contacts', action.content._id, 'edit']);
+      } else if (action.content?.parent_id) {
         this.router.navigate(['/contacts', action.content.parent_id, 'add', action.content.type || '']);
       } else {
         this.router.navigate(['/contacts', 'add', action.content?.type || '']);
--- a/tests/e2e/default/tasks/config/tasks-contact-config.js
+++ b/tests/e2e/default/tasks/config/tasks-contact-config.js
@@ -1,0 +1,44 @@
+
+module.exports = [
+  {
+    name: 'add_household_members_task',
+    title: 'Add Household Members',
+    appliesTo: 'contacts',
+    appliesToType: ['clinic'],
+    resolvedIf: () => false,
+    events: [{
+      id: 'add_household_members_task_event_id',
+      days: 0,
+      start: 0,
+      end: 1,
+    }],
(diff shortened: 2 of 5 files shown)


jkuester:
@ChinHairSaintClair Thanks for the PR! I love the sound of this!  Just wanted to note that I plan to have a look at this by EOW.  :+1:

ChinHairSaintClair:
Awesome! Thank you for taking the time to review @jkuester, I'll get on that asap. 

Yeah, the missing docs surprised me a little too. I know there HAD to be something given the [option](https://docs.communityhealthtoolkit.org/building/tasks/tasks-js/#tasksjs) to set `action.type` to "contact". So, I did some digging. Necessity really is the mother of invention! 😆. 

I don't have an example for "add-contact-from-task", but I did check that the existing checks for that functionality still pass after my tweaks. For “edit-contact-from-task”, my current config is still a work in progress, but here’s a (very) rough version:
```
{
    name: 'my_test_task',
    icon: 'icon-follow-up',
    title: 'my_test_task',
    appliesTo: 'contacts',
    appliesToType: ['person'],
    appliesIf: function (contact, report) {
      return contact.contact.name !== null && contact.contact.name !== undefined;
    },
    resolvedIf: function (contact, report, event, dueDate) {
      const start = Utils.addDate(dueDate, -event.start).getTime();
      const end = Utils.addDate(dueDate, event.end + 1).getTime();
      return false;
    },
    actions: [
       {
         type: 'contact',
         modifyContent: function (content, contact, report, event) {
           const followupCount = this.definition.events.findIndex(e => event.id === e.id) + 1;
           content.t_followup_count = followupCount.toString();

           Object.assign(content, contact.contact);
           Object.assign(content, { type: contact.contact.contact_type });
           Object.assign(content, { is_edit: true });
         }
       }
    ],
    events: [
      {
        id: 'initial',
        days: 0,
        start: 50,
        end: 50,
      },
    ],
    priority: { 
      level: 'medium', 
      label: 'This is a medium priority label'
    },
  }
```

You’ll notice it doesn’t yet have a solid `resolvedIf` implementation, but the general idea is that the task’s due date would check if a `follow_up_date` for a contact has been recorded within the last 4 months. We could also combine this with time-interval requirements, like shown in the [complex tasks docs](https://docs.communityhealthtoolkit.org/building/tasks/complex-tasks/#developing-the-task).

In our case, determining if the task is resolved would involve checking multiple forms submitted (e.g., contact follow-up, CSHARP, and COPC). We’d likely need to verify multiple reported dates, and `Utils.isFormSubmittedInWindow` would probably help with those app forms.

I do wish individual actions could have their own `resolvedIf` functions, as that would help prevent duplicates by removing completed action buttons from the UI, and allow the task to automatically close once all actions are done.

I’ve been pulled onto other priorities, but I’ll try to get a solid, real-world example ready for you soon.

jkuester:
Thanks for  the config code! This is very helpful context!  (And yeah, the `resolvedIf` gets pretty messy here... :thinking: )  

However, I am not sure about this `status` property approach. :thinking: This task config stuff is already impossible to understand and so I hate to add any more moving parts here.  At the very least I think we might want to address that (and any other `resolvedIf` issues) in a separate issue/PR).  After seeing your config, I was able to cook up a couple bare-bones examples [here](https://github.com/jkuester/cht-config-library/pull/25/files#diff-e21b637c28b66207b94de89a5df802964bad276817043c88020c4716c74238fc) that I think demonstrate the add/modify contact functionality in a useful way (without the `status` manipulation).  Let me know if these make sense to you!  :pray: 

Feel free to push back here and help me understand the necessity behind the `status` piece, or go ahead and log a separate issue for that and we can continue the conversation there. :+1:

webapp/src/ts/modules/tasks/tasks-content.component.ts
```diff
@@ -304,7 +304,9 @@ export class TasksContentComponent implements OnInit, OnDestroy {
     }
 
     if (action.type === 'contact') {
-      if (action.content?.parent_id) {
+      if (action.content?.id){
+        this.router.navigate(['/contacts', action.content.id, 'edit']);
+      } else if (action.content?.parent_id) {
```
jkuester: Thinking about this even more, several more considerations occurred to me:

1. First of all, I think we want to be careful about backwards compatibility here so that we don't accidentally break existing working configuration. With that in mind, we should probably prefer the existing navigations if `parent_id` or `type` exist in the `content`.
2. Also, the existing "default" behavior of what happens when you have a `contact` action, but don't set anything in the `content` seems really bad, right?  Navigating the user to `/contacts/add` (with no type on the end) just ends up failing to load a form and then falling back to showing the contacts list (in the "People" tab). Seems like we could have a better default behavior of just navigating to the contact associated with the task (`action.content?.contact?._id`).
3. Finally, I know I suggested `id`, but the more I think about it (and read the task config code that is using it) the more confusing it seems.  What do you think of `edit_id` or something similarly less generic?

So, putting it all together, this code could look something like:

```ts
      if (action.content?.parent_id) {
        this.router.navigate(['/contacts', action.content.parent_id, 'add', action.content.type || '']);
      } else if (action.content?.type) {
        this.router.navigate(['/contacts', 'add', action.content.type]);
      } else if (action.content?.edit_id) {
        this.router.navigate(['/contacts', action.content.edit_id, 'edit']);
      } else {
        this.router.navigate(['/contacts', action.content?.contact?._id ?? '']);
      }
```

ChinHairSaintClair: Sound thinking. I also found the default behaviour quite confusing - I'm glad we can improve it as part of this. Having an `edit_id` is more descriptive, I like that. In my original task content, I played around with adding `Object.assign(content, { is_edit: true });`, but this is much cleaner.

ChinHairSaintClair:
Thank you for the review and for putting those examples together, @jkuester! 

I'm happy to open a new PR for the `status` functionality, as you're right - it doesn't strictly belong in the "edit contact from task" scope. That said, I do think it's a useful piece of functionality. It could support data quality in a similar vein to the dedupe work.

Take a task that requires a CHW to perform multiple actions:
- Manage contact details for a place/person
- Capture available facilities/medical info
- Answer research-related questions

That’s quite a bit of data to collect, and it’s not always realistic to expect a CHW to complete everything in one sitting. If they need to return to the task later - **maybe a day or more later** - we shouldn't expect them to remember exactly where they left off. Requiring them to leave the task flow and dig into the person view just to verify what’s been done feels like a poor UX.

A lightweight indicator like `status` could help show progress at a glance, help prevent duplicate form capture, and in so doing support data quality - all without burdening the CHW.

jkuester (approved):
Looking great!  @ChinHairSaintClair can you have a look through the e2e tests I added and let me know if they make sense. If all looks good, I will hit the merge button. :+1:

ChinHairSaintClair:
Thank you @jkuester! Sorry for not getting around to the e2e tests. They look great! I'll use them as a guide going forward. 
I'm wondering if we shouldn't add ` && action.content?.type` to the `parent_id` if case.
When I attempt to navigate to `http://localhost:5988/#/contacts/a89128b4-a465-48e2-990e-1f36282de0db/add`, which would happen if the `type` is not defined, I'm met with a "Not Found" screen:
image
I'm wondering if the user would much rather just have been directed to the ~parent~ contact itself, as that would be the fallback.

jkuester (changes requested):
>  I'm wondering if we shouldn't add  && action.content?.type to the parent_id if case.

Ohh yeah, good call!  That is a pretty bad experience. I agree, best to just send them to the contact profile page. Please go ahead and add the additional check (and maybe a unit test it if makes sense. Probably don't need to e2e test that scenario...)!

jkuester (approved):
LGTM!

Issue #5997, opened 2019-10-03
Sort/Filter tasks by priority


not-the-fish:
**Is your feature request related to a problem? Please describe.**
CHVs often have many tasks to do and currently, the only way they can prioritize them is by due date sorting or by reading all of the tasks doing their own prioritization.

**Describe the solution you'd like**
We already flag some tasks as high priority, and with predictive algorithms may be able to identify additional high priority tasks (e.g., important tasks unlikely to completed or tasks where we expect the patient to show danger signs). Similar to medic/cht-core#5886, we can help CHWs plan and prioritize their activities by one (or more) of the following:

- providing the ability to order tasks by priority
- providing the ability to order tasks by due date-weighted priority
- providing the ability to filter tasks by high priority

**Describe alternatives you've considered**
Alternatives include the current mechanisms to flag a task as high priority and change its presentation to the user, including its icon and text. However, for a CHW with many tasks, this may not surface high priority tasks as easily as sorting or filtering would.

Another alternative is for urgent tasks (e.g., those where we expect danger signs) to use SMS or a dialog box upon opening the app to remind the CHW of the importance of the task. However, this may be overwhelming for a CHW if too many tasks are high priority.

**Additional context**
This is relevant as we consider expansion and scale-up of our predictive algorithms initiatives through recently launched data science partnerships. These partnerships would be looking to close out in mid-2020, with implementation possibly in late 2020.

MaxDiz:
Related to #5886

michaelkohn:
This need also came up on some contact tracing conversations whereby you might want to prioritize tracing tasks based on certain risk criteria (age, health history, etc....)

kennsippell:
This issue came up again for eCHIS digital payment work. A task must be resolved by each CHPs before they can be paid. 

Many (50?) CHPs have messaged their supervisor this month saying that the task is not on the list. When we investigate, we can find the task but need to scroll down.

The head of the Nairobi community health program asked us "make all the other tasks disappear until this one is resolved".  This request underscores the importance of high-priority tasks and that the existing CHT functionality is insufficient for this programmatic needs.

n-orlowski:
@kennsippell , what about using the [high priority label](https://docs.communityhealthtoolkit.org/building/tasks/tasks-js/#tasksjs) with a [filtering functionality](https://github.com/medic/cht-core/issues/9558) to solve for this?

kennsippell:
[...]

PR #9978, opened 2025-05-21
feat(#5997): dynamic sorting of tasks by priority

sookwalinga:


sookwalinga:
Hi @dianabarsan please find the updates to the task sorting function here

Issue #9957, opened 2025-05-09
Implement task sorting in ascending or descending order


andrablaj:
#### Is your feature request related to a problem? Please describe.
Currently tasks are sorted only by due date, so users have to manually identify tasks that are most urgent. The update takes into account both a priority score and due date to prioritize tasks.

#### Describe the solution you'd like
Extend the sorting logic to sort tasks as follows:
1. Priority values are positive numbers
2. Higher priorities sort first
3. Equal priorities sort by due date
4. Invalid/missing values sort last while adhering to the above

#### Describe alternatives you've considered
1. Store an array of priorities per task for e.g. [not due, due, overdue], then let the WebApp pick the correct one. This turned out to be rigid and complex to maintain.
2. Sort entirely by due date and then priority. This does not fully capture the intended urgency based on individual conditions.
3. Sort overdue tasks by due date, and all other tasks by priority. This sounded great, but less consistent for borderline-due tasks.

kalamakitsao:
I can help update the description. Please assign it to me.

kalamakitsao:
Some description was added to the [pull request](https://github.com/medic/cht-core/pull/9980#issue-3082777948) instead. Will work with @sookwalinga to get this out.

kalamakitsao:
Some minor updates/rephrasing:

#### Is your feature request related to a problem? Please describe.
Currently tasks are sorted only by due date, so users have to manually identify tasks that are most urgent. The update takes into account both a priority score and due date to prioritize tasks.

#### Describe the solution you'd like
Extend the sorting logic to sort tasks as follows:
1. Priority values are positive numbers
2. Higher priorities sort first
3. Equal priorities sort by due date
4. Invalid/missing values sort last while adhering to the above

#### Describe alternatives you've considered
1. Store an array of priorities per task for e.g. [not due, due, overdue], then let the WebApp pick the correct one. This turned out to be rigid and complex to maintain.
2. Sort entirely by due date and then priority. This does not fully capture the intended urgency based on individual conditions.
3. Sort overdue tasks by due date, and all other tasks by priority. This sounded great, but less consistent for borderline-due tasks.

PR #9980, opened 2025-05-22
feat(#9957): dynamic sorting of tasks by priority and due date

sookwalinga:
Closes https://github.com/medic/cht-core/issues/9957

  Currently tasks are sorted only by due date. The update takes into account both a priority score and due date.

  Implemented a refined sorting algorithm that (Sort by priority and then due date within the same priority):
  - Primarily sorts tasks by priority (higher numbers first)
  - For tasks with equal priority, sorts by due date (earlier dates first)
  - Handles invalid values consistently:
      - Tasks with invalid priorities move to end
      - Within same priority, invalid dates move to end
      - Maintains original order when both priority and date are invalid

****Alternative Solution 1****
Similar to the proposed solution, but with the priority property of a task not be a number, but an array, with one number representing the priority of the task when it is not due, one number for when it’s due and one number for when it’s overdue. Then web-app will just select the right number when sorting tasks, and not recompute priority based on due date.

****Alternative Solution 2****
Options for frontend sorting: 
- Sort by due date and then priority within the same due date.
- Sorting by Due date first for tasks above the due date, then by priority for tasks that are not due. 
- Sort by priority for tasks that are above due date, then by due date for tasks that are not due.

**Additional context**
- The solution maintains backwards compatibility with existing task structures
- Includes comprehensive test cases covering edge cases like:
  - Invalid date formats (null, false, undefined).
  - Invalid priority values (strings, negative numbers).
  - Duplicate priorities and dates.
  - Missing fields.

### DIFF
--- a/config/default/tasks.js
+++ b/config/default/tasks.js
@@ -133,8 +133,7 @@ module.exports = [
       },
     ],
     //every two weeks from reported date until 42nd week, show before due date: 6 days, show after due date: 7 days
-    events: [...Array(21).keys()].map((i) =>
-      generateEventForHomeVisit((i + 1) * 2, 6, 7)
+    events: [...Array(21).keys()].map(i => generateEventForHomeVisit((i + 1) * 2, 6, 7)
     ),
   },
 
--- a/config/default/tasks.js
+++ b/config/default/tasks.js
@@ -1,3 +1,4 @@
+// @ts-nocheck
 const extras = require('./nools-extras');
 
 const {
@@ -14,7 +15,7 @@ const {
   addDays,
   getRecentANCVisitWithEvent,
   isPregnancyTaskMuted,
-  getField
+  getField,
 } = extras;
 
 const generateEventForHomeVisit = (week, start, end) => ({
@@ -23,36 +24,52 @@ const generateEventForHomeVisit = (week, start, end) => ({
   end,
   dueDate: function (event, contact, report) {
     const recentLMPDate = getMostRecentLMPDateForPregnancy(contact, report);
-    if (recentLMPDate) { return addDays(recentLMPDate, week * 7); }
+    if (recentLMPDate) {
+      return addDays(recentLMPDate, week * 7);
+    }
     return addDays(report.reported_date, week * 7);
-  }
+  },
 });
 
 function checkTaskResolvedForHomeVisit(contact, report, event, dueDate) {
   //delivery form submitted
-  if (report.reported_date < getNewestDeliveryTimestamp(contact)) { return true; }
+  if (report.reported_date < getNewestDeliveryTimestamp(contact)) {
+    return true;
+  }
 
   //old pregnancy report
-  if (report.reported_date < getNewestPregnancyTimestamp(contact)) { return true; }
+  if (report.reported_date < getNewestPregnancyTimestamp(contact)) {
+    return true;
+  }
 
   //miscarriage or abortion
-  if (getRecentANCVisitWithEvent(contact, report, 'abortion') || getRecentANCVisitWithEvent(contact, report, 'miscarriage')) { return true; }
+  if (
+    getRecentANCVisitWithEvent(contact, report, 'abortion') ||
+    getRecentANCVisitWithEvent(contact, report, 'miscarriage')
+  ) {
+    return true;
+  }
 
   //Due date older than reported day
-  if (dueDate <= getTimeForMidnight(report.reported_date)) { return true; }
+  if (dueDate <= getTimeForMidnight(report.reported_date)) {
+    return true;
+  }
 
   //Tasks cleared
-  if (isPregnancyTaskMuted(contact)) { return true; }
-  const startTime = Math.max(addDays(dueDate, -event.start).getTime(), report.reported_date);
+  if (isPregnancyTaskMuted(contact)) {
+    return true;
+  }
+  const startTime = Math.max(
+    addDays(dueDate, -event.start).getTime(),
+    report.reported_date
+  );
   const endTime = addDays(dueDate, event.end + 1).getTime();
-  return isFormArraySubmittedInWindow(contact.reports, ['pregnancy_home_visit'], startTime, endTime);
-}
-
-function getPriorityCategory(score) {
-  if (score < 6){ return 'Low priority';}
-  else if (score >= 6 && score <= 8){ return 'Medium priority';}
-  else if (score > 8){ return 'High priority';}
-  return '';
+  return isFormArraySubmittedInWindow(
+    contact.reports,
+    ['pregnancy_home_visit'],
+    startTime,
+    endTime
+  );
 }
 
 module.exports = [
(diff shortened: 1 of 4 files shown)


sookwalinga:
Hi @dianabarsan please find the rebased pr here

webapp/src/ts/reducers/tasks.ts
```diff
@@ -14,20 +14,60 @@ const initialState = {
...
+    if (t?.priority === 'high') {
+      return 10;
+    } else if (t?.priority === 'medium') {
+      return 6;
+    } else if (typeof t?.priority === 'string') {
+      return 0;
+    } else {
+      return t?.priority ?? 0;
+    }
+  };
```
dianabarsan, line 43: I think we should not hardcode **any values**. 
If the priority is a string, we always consider it the lowest priority (for example, have this function return NaN, and check for NaN when comparing priorities.

webapp/src/ts/reducers/tasks.ts
```diff
@@ -14,20 +14,60 @@ const initialState = {
...
+
+  const now = new Date();
+  const startOfToday = new Date(now.getFullYear(), now.getMonth(), now.getDate()).getTime();
```
dianabarsan, line 46: Please use `moment` functions instead of JS date.

webapp/src/ts/reducers/tasks.ts
```diff
@@ -14,20 +14,60 @@ const initialState = {
...
+  
+  const getDateScore = (t) => {
+    if (!t?.dueDate) return Infinity;
+    const date = new Date(t.dueDate).getTime();
+    return Math.floor((date - startOfToday) / (1000 * 60 * 60 * 24));
```
dianabarsan, line 51: `moment` would have good date diff functionality that you can use here.

webapp/src/ts/reducers/tasks.ts
```diff
@@ -14,20 +14,60 @@ const initialState = {
...
+  };
+
+  const p1 = getPriorityScore(t1);
+  const p2 = getPriorityScore(t2);
```
dianabarsan, line 55: Please don't use one letter variables. This can be `priority1`.

webapp/src/ts/reducers/tasks.ts
```diff
@@ -14,20 +14,60 @@ const initialState = {
...
+  const days1 = getDateScore(t1);
+  const days2 = getDateScore(t2);
```
dianabarsan, line 57: I think `days` doesn't really convey what this value is. Maybe `dueInDays` ?

webapp/src/ts/services/format-date.service.ts
```diff
@@ -107,14 +107,18 @@ export class FormatDateService {
...
+    if (diff < 0) {
       if (this.config.taskDaysOverdue) {
         return this.translateService.instant('task.overdue.days', { DAYS: Math.abs(diff) });
       }
 
       return this.translateService.instant('task.overdue');
     }
 
+    if (diff === 0) {
+      return this.translateService.instant('Due today');
```
dianabarsan, line 119: Please follow the conventions for adding new translation fields here. 
Please use a dot separated name (like `task.due.today`) and add a translation for it in all the translation files: https://github.com/medic/cht-core/tree/master/api/resources/translations

Because we don't want to change behavior for existent deployments when adding this new translation field, when adding the translation please make it the same values as the one for `task.overdue`

config/default/tasks.js
```diff
@@ -48,8 +48,14 @@ function checkTaskResolvedForHomeVisit(contact, report, event, dueDate) {
   return isFormArraySubmittedInWindow(contact.reports, ['pregnancy_home_visit'], startTime, endTime);
 }
 
-module.exports = [
+function getPriorityCategory(score) {
```
dianabarsan, line 51: I don't think we should be adding any of these changes to the default config. Please revert them.

dianabarsan (changes requested):
Nice work! I left a couple of requests below. we're getting close :)

webapp/tests/karma/ts/reducers/tasks.spec.ts
```diff
@@ -221,7 +221,7 @@ describe('Tasks reducer', () => {
     it('should sort provided tasks by due date', () => {
       const tasks = [
         { _id: 'task1', dueDate: false, state: 'Ready', field: 1 },
-        { _id: 'task2', dueDate: undefined, state: 'Ready', field: 2 },
+        { _id: 'task2', dueDate: null, state: 'Ready', field: 2 },
```
dianabarsan, line 224: Is this change required? What happens if the dueDate is undefined? Does the behavior change?

webapp/tests/karma/ts/reducers/tasks.spec.ts
```diff
@@ -242,15 +242,215 @@ describe('Tasks reducer', () => {
...
+    });
+
+    it('should sort provided tasks by priority and due date', () => {
+      const tasks = [
+        {
+          _id: 'task1',
+          dueDate: '2025-05-30',
+          priority: 'invalid',
+          state: 'Ready',
+          field: 1,
```
dianabarsan, line 264: Can you please add a task that has both invalid due date _and_ invalid priority? Also please don't add it as last in the initial list, it would be good to see it actually pushed down to the end.

dianabarsan (changes requested):
Very nice and timely! 
I left some minor requests inline.

webapp/tests/karma/ts/reducers/tasks.spec.ts
sookwalinga: For some reason, test output shows `"dueDate": [undefined]`, yet the input data has `dueDate: undefined`. Now in the conditional statement `moment([undefined]).isValid()` returns `true ` so the test fails. I tried figuring out why `undefined` is returned as `[undefined]` but I haven't yet figured that out.

dianabarsan: Ok. 
I believe that `moment(undefined).isValid();` is actually correct to be true, because that falls back to `moment().isValid()` which is true because moment instantiates using the current date when no input is provided. 
This means that to protect for `undefined`, you need to add a specific check in your date function.

config/default/tasks.js
```diff
@@ -308,35 +426,58 @@ module.exports = [
...
       const endTime = addDays(dueDate, event.end + 1).getTime();
-      return isFormArraySubmittedInWindow(contact.reports, ['pnc_danger_sign_follow_up_baby'], startTime, endTime);
+      return isFormArraySubmittedInWindow(
+        contact.reports,
+        ['pnc_danger_sign_follow_up_baby'],
+        startTime,
+        endTime
+      );
+    },
+    priority: function (contact, report, event, dueDate) {
```
dianabarsan, line 452: I believe these changes should be reverted, I don't think we should be updating the default config with random priority scores. 
I will do this as part of a secondary PR once this is merged if we choose to merge this quickly.

PR #9987, opened 2025-05-26
feat(#9255):  adds androidApi service to calculate task notifications

### LINKED ISSUE #9255: Enable android notifications for task reminders/alerts
Users sometimes miss critical app updates such as tasks since they need to open the app first to view updates. 

CHT/cht-android needs to have some notifications for users as a way to remind/ inform users on app activities such as pending tasks and incoming messages. 
Currently the user needs to open the app's tasks/messages tab to get and update on these items.
A notification feature can help in the usability by managing interactions with the app.

jonathanbataire:
# Description
- Adds a task notification service  that returns task notifications to allow notification within the web app or android app
- Adds new androidApi method to allow cht-android access these notifications. Android app access this API [here](https://github.com/medic/cht-android/pull/394) 
-  Adds new translation key `android.notification.tasks.contentText` for translating notification content text.  

fixes #9255 

# Code review checklist

# License


jonathanbataire:
Hey @andrablaj this ready for review :)

andrablaj:
@dianabarsan, would you be available to review this PR? If not, please reach out to/assign other teammates who might be available. Thank you!

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,103 @@
...
+    private readonly rulesEngineService: RulesEngineService,
+    private readonly translateService: TranslateService,
+    private readonly dbSyncService: DBSyncService
+  ) { }
+
+  private async fetchNotifications(): Promise<Notification[]> {
+    try {
+      const today = moment().format('YYYY-MM-DD');
+      let latestNotificationTimestamp = this.getLatestNotificationTimestamp();
+      const isEnabled = await this.rulesEngineService.isEnabled();
```
dianabarsan: We can just return early here, because we won't have notifications at all if there are no tasks.

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,103 @@
...
+      const taskDocs = isEnabled ? await this.rulesEngineService.fetchTaskDocsForAllContacts() : [];
+      
+      let notifications = taskDocs
+        .filter(task => {
+          return task.state === 'Ready' && task.emission.dueDate === today &&
```
dianabarsan: I believe that all tasks returned by `fetchTaskDocsForAllContacts` will be in the `Ready` state, so this filter is not needed.

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,103 @@
...
+    try {
+      const today = moment().format('YYYY-MM-DD');
+      let latestNotificationTimestamp = this.getLatestNotificationTimestamp();
+      const isEnabled = await this.rulesEngineService.isEnabled();
+      const taskDocs = isEnabled ? await this.rulesEngineService.fetchTaskDocsForAllContacts() : [];
+      
+      let notifications = taskDocs
+        .filter(task => {
+          return task.state === 'Ready' && task.emission.dueDate === today && 
+            task.authoredOn > latestNotificationTimestamp;
```
dianabarsan: I'm not sure what purpose this serves. `authoredOn` is when the task gets first created. This can be up to 60 days before the task is ready. How is the date when the task doc was **created** relevant for showing notifications?

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,103 @@
...
+        .map(task => ({
+          _id: task._id,
+          authoredOn: task.authoredOn,
+          state: task.state,
+          title: task.emission.title,
+          contentText: this.translateContentText(task.emission.title, task.emission.contact.name),
+          dueDate: task.emission.dueDate,
+        }));
+
+      notifications = orderBy(notifications, ['authoredOn'], ['desc']);
```
dianabarsan: Again, `authoredOn` is when the task was created. The is completely random and depends on things like: when the user synced, when the configuration changed and other events that are not relevant to task sorting in any logical way.

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,103 @@
...
+      notifications = notifications.slice(0, MAX_NOTIFICATIONS);
+      latestNotificationTimestamp = notifications[0]?.authoredOn ?? latestNotificationTimestamp;
```
dianabarsan: Ok, I think I see where this is going. You're trying to somehow only notify for new tasks. 
Unfortunately, we don't store task recalculation dates, we only store task state change dates - so a task might suffer changes (for example it's icon could change or a label etc), but the timestamp of the change is only recorded if the task state changed. 

You could use the last entry of the task's `stateHistory` to get the timestamp of when the task became `Ready`.

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  private isNewDay(): boolean {
+    const now = moment();
+    const timestampToday = Number(window.localStorage.getItem(TODAY_TIMESTAMP));
+    if (!now.isSame(timestampToday, 'day')) {
+      window.localStorage.setItem(TODAY_TIMESTAMP, String(moment().startOf('day').valueOf()));
+      return true;
+    }
+    return false;
+  }
```
dianabarsan, line 102: Window local storage is site-wide. Storing a `TODAY_TIMESTAMP` that has a value from a different day is super confusing. 
I suggest we rename the stored variables and store: `LATEST_NOTIFICATION_TIMESTAMP` with the value of the last time the user received a notification, `TASK_STATE_NOTIFICATION_TIMESTAMP` (or something like this) that stores the task state calculation timestamp. 

```suggestion
  private isNewDay(): boolean {
    const timestampToday = Number(window.localStorage.getItem(LATEST_NOTIFICATION_TIMESTAMP));
    return !moment().isSame(timestampToday, 'day')  
  }
```

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  private translateContentText(task: string, contact: string): string {
+    const key = 'android.notification.tasks.contentText';
+    return this.translateService.instant(key, { task, contact });
+  }
+
+  async get(): Promise<Notification[]> {
+    return Promise.race([
+      this.dbSyncService.sync(),
+      new Promise(resolve => setTimeout(() => resolve([]), 5 * 1000))
```
dianabarsan: I don't believe sync should be called here. Sync should be controlled by other events. 
Also, please don't use `then` in async functions: 
```suggestion
get(): Promise<Notification[]> {
  return this.fetchNotifications();
}
```

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,103 @@
...
+
+import { RulesEngineService } from '@mm-services/rules-engine.service';
+import { TranslateService } from '@mm-services/translate.service';
+import { DBSyncService } from '@mm-services/db-sync.service';
+
+/* 
+** avoid overloading app with too many notifications especially at once
+** 24 for android >= 10
+*/
+const MAX_NOTIFICATIONS = 24;
```
dianabarsan: I'm curious how this number was selected. This seems like an incredibly high number. I believe we should consult with some UI experienced designers for the number of notifications that is acceptable, and doesn;t become overwhelming. 
I believe we should also make this configurable.

webapp/src/ts/services/task-notifications.service.ts
dianabarsan: What about tasks that are overdue? I'm curious what this system is really trying to achieve. 
If a user doesn't log in for one day, when a specific task is due, and logs in the next day when the task is overdue, they never get notified for it.

dianabarsan (changes requested):
Great start and awesome work! I have some questions and suggestions inline. Please let me know if you have any questions.

webapp/src/ts/services/task-notifications.service.ts
jonathanbataire: - Starting with android 10 android dropped [max number of notifications](https://support.google.com/pixelphone/thread/23619509/notification-limits-dropped-to-24-in-android-10-can-i-change-that?hl=en) to 24. Anything after 24 is discarded. This is what i went off of
- for the android UI if notifications are [more than 4 they'll be grouped](https://developer.android.com/develop/ui/views/notifications#bundle) like below

LMK what you think especially on the making this number configurable

Grouped by default
Screenshot_20250610_095750_One UI Home

when user expands them
Screenshot_20250610_095801_One UI Home

webapp/src/ts/services/task-notifications.service.ts
jonathanbataire: i see that's one scenario that wasn't considered 
yeah we can add overdue tasks

webapp/src/ts/services/task-notifications.service.ts
jonathanbataire: @dianabarsan how can we handle tasks triggered server side without dbsync eg tasks triggered by an integration
this has come up as a very important use case for notifications

dianabarsan: sync will be handled by other mechanisms, it happens automatically every 5 minutes and when users make changes to docs.

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,94 @@
...
+      if (!isEnabled) {
+        return [];
+      }
+      const today = moment().format('YYYY-MM-DD');
+      let latestNotificationTimestamp = this.getLatestNotificationTimestamp();
+      const taskDocs = await this.rulesEngineService.fetchTaskDocsForAllContacts();
+      let notifications = taskDocs
+        .map(task => ({
+          _id: task._id,
+          readyAt: get(find(task.stateHistory, { state: 'Ready' }), 'timestamp') || 0,
```
dianabarsan: Can you please use built-in javascript functions here? it makes the code clearer and more readable. it will also be easier to switch to some other library later.

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,94 @@
...
+          title: task.emission.title,
+          contentText: this.translateContentText(
+            task.emission.title,
+            task.emission.contact.name,
+            task.emission.dueDate
+          ),
+          dueDate: task.emission.dueDate,
+        }))
+        .filter(notification => notification.dueDate <= today &&
```
dianabarsan: We can get away with a single iteration here: 
```suggestion
const notifications = [];
taskDocs.forEach(task => {
  const readyAt = task.stateHistory.find(state => state.state === 'Ready')?.timestamp || 0;
  if (readyAt < latestNotificationTimestamp || task.emission.dueDate > today) {
    return;
  }
  notifications.push({
     _id: task._id,
     readyAt,
     title: task.emission.title,
     contentText: this.translateContentText(
       task.emission.title,
       task.emission.contact.name,
       task.emission.dueDate
     ),    
  });
})
```

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,94 @@
...
+import { TranslateService } from '@mm-services/translate.service';
+import { FormatDateService } from '@mm-services/format-date.service';
+
+/* 
+** avoid overloading app with too many notifications especially at once
+** 24 for android >= 10
+*/
+const MAX_NOTIFICATIONS = 24;
+const TASK_NOTIFICATION_STATE_TIMESTAMP = 'cht-task-notification-state-timestamp';
+const LATEST_NOTIFICATION_TIMESTAMP = 'cht-latest-notification-timestamp';
```
dianabarsan:
```suggestion
const LATEST_NOTIFICATION_TIMESTAMP = 'cht-task-notification-timestamp';
```

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,94 @@
...
+import { RulesEngineService } from '@mm-services/rules-engine.service';
+import { TranslateService } from '@mm-services/translate.service';
+import { FormatDateService } from '@mm-services/format-date.service';
+
+/* 
+** avoid overloading app with too many notifications especially at once
+** 24 for android >= 10
+*/
+const MAX_NOTIFICATIONS = 24;
+const TASK_NOTIFICATION_STATE_TIMESTAMP = 'cht-task-notification-state-timestamp';
```
dianabarsan:
```suggestion
const TASK_NOTIFICATION_DAY = 'cht-task-notification-day';
```

webapp/src/ts/services/task-notifications.service.ts
dianabarsan: If an app sent me 24 notifications, I'd uninstall it :))) 

Maybe we could consult with a UX designer about this.

webapp/tests/karma/ts/services/task-notifications.service.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+          dueDate: moment().format('YYYY-MM-DD')
+        },
+        stateHistory: [
+          {
+            state: 'Ready',
+            timestamp: moment().add(100, 'milliseconds').valueOf()
+          }
+        ],
+      }
+    ];
```
dianabarsan, line 71: Please add a task that has more entries in the state history.

webapp/tests/karma/ts/services/task-notifications.service.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+    rulesEngine.fetchTaskDocsForAllContacts.resolves([...tasks, taskDueTomorrow]);
+    const clock = sinon.useFakeTimers(moment().valueOf());
+    const notifications = await service.fetchNotifications();
+    expect(notifications).to.be.an('array').that.has.lengthOf(2);
+
+    clock.tick(DAY_IN_MS);
+
+    const updatedNotifications = await service.fetchNotifications();
+    expect(updatedNotifications).to.be.an('array').that.has.lengthOf(3);
+    clock.restore();
```
dianabarsan: The issue with restoring the clock at the end of the test is that, if the test fails in an assertion, the clock is not restored. 
My preferred approach is to declare `clock` as a higher scope variable, and do a `clock?.restore()` in the `afterEach` hook

webapp/tests/karma/ts/services/task-notifications.service.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+  it('should fetch notifications with no tasks ready', async () => {
+    rulesEngine.fetchTaskDocsForAllContacts.resolves([]);
+    const notifications = await service.fetchNotifications();
+    expect(notifications).to.be.an('array').that.is.empty;
+    expect(rulesEngine.fetchTaskDocsForAllContacts.callCount).to.equal(1);
+    expect(translateService.instant.callCount).to.equal(0);
+    expect(consoleErrorMock.callCount).to.equal(0);
+  });
+
+  it('should fetch notifications with no tasks due today', async () => {
```
dianabarsan: The test title is confusing :)) I kept reading and assuming that the task is due in the future and we should not return any notifications. Can you please change it to something like "should fetch notifications for overdue tasks"?

webapp/tests/karma/ts/services/task-notifications.service.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+    const pastTask = {
```
dianabarsan, line 240: Can you please add a test where you show that tasks that are due in the future do not generate notifications?

dianabarsan (changes requested):
Very nice! I added some minor comments inline. 
Does this have a corresponding `cht-android` change?

jonathanbataire:
> Very nice! I added some minor comments inline. Does this have a corresponding `cht-android` change?

Yeah here is the cht-android PR https://github.com/medic/cht-android/pull/394
but still evolving depending on the final form of this PR

webapp/src/ts/services/task-notifications.service.ts
jonathanbataire: @dianabarsan sure 
for now i've made it configurable thru settings file and defaults to 10

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,118 @@
...
+    private readonly formatDateService: FormatDateService,
+    private readonly settingsService: SettingsService,
+  ) { }
+
+  private async fetchNotifications(): Promise<Notification[]> {
+    try {
+      const isEnabled = await this.rulesEngineService.isEnabled();
+      if (!isEnabled) {
+        return [];
+      }
```
dianabarsan, line 40: I just realized that we don't have anything to disable notifications for some projects / users. 
Can you please add something like: 
```suggestion
const canGetTaskNotifications = await this.authService.has('can_get_task_notifications');
```
similar to how it's done in the rules-engine service: https://github.com/medic/cht-core/blob/51da792b7bc0c6e9b212bc61a7ff0afe05fa440a/webapp/src/ts/services/rules-engine.service.ts#L97

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,118 @@
...
+  private getReadyStateTimestamp(stateHistory): number {
+    const readyState = stateHistory.find(state => state.state === 'Ready');
+    return readyState ? readyState.timestamp : 0;
+  }
+
+  private getMaxNotificationSettings(): Promise<number> {
+    return this.settingsService
+      .get()
+      .then((res) => {
+        if (typeof res?.max_task_notifications === 'number' && res.max_task_notifications >= 0) {
```
dianabarsan: this is great! I'm wondering if we can somehow put this in the settings.tasks, so they're logically grouped instead of scattered. 
It would probably need a change here: https://github.com/medic/cht-conf/blob/main/src/fn/compile-app-settings.js#L126

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,118 @@
...
+      return [];
+    }
+  }
+
+  private getReadyStateTimestamp(stateHistory): number {
+    const readyState = stateHistory.find(state => state.state === 'Ready');
+    return readyState ? readyState.timestamp : 0;
+  }
+
+  private getMaxNotificationSettings(): Promise<number> {
```
dianabarsan: Can you please use async/await here instead. Also ... if the settingsService fails we're in huge trouble. I don't think we should be catching this error, but failing loudly.

webapp/tests/karma/ts/services/task-notifications.service.spec.ts
```diff
@@ -0,0 +1,305 @@
...
+    expect(notifications[0]._id).to.equal('task2');
+    expect(notifications[1]._id).to.equal('task1');
+    expect(rulesEngine.fetchTaskDocsForAllContacts.callCount).to.equal(1);
+    expect(translateService.instant.callCount).to.equal(2);
+    expect(consoleErrorMock.callCount).to.equal(0);
+
+    notifications.forEach((notification) => {
+      const task = getTask(notification._id);
+      expect(notification).to.be.an('object');
+      expect(notification).to.eql({
```
dianabarsan, line 142: I love this! So clean!

webapp/src/ts/services/task-notifications.service.ts
jonathanbataire: added it here https://github.com/medic/cht-conf/pull/717

webapp/tests/karma/ts/services/task-notifications.service.spec.ts
```diff
@@ -282,7 +311,27 @@ describe('TasksNotificationService', () => {
...
+
   it('should get() notifications', async () => {
+    sessionService.userCtx.returns({ roles: ['chw'] });
+    settingsService.get.resolves({
+      permissions: {
+        can_get_task_notifications: ['chw']
+      },
+    });
     const notifications = await service.get();
     expect(notifications).to.be.an('array').that.has.lengthOf(2);
```
dianabarsan, line 307: Please add an assertion for what the authService is called with: 
```suggestion
    expect(notifications).to.be.an('array').that.has.lengthOf(2);
    expect(authService.has.calledOnceWith('can_get_task_notifications')).to.be.true;
```

dianabarsan (approved):
Excellent work!

dianabarsan (approved):
Very clean work!

dianabarsan:
@lorerod Do you have some time to acceptance test this new feature? There is a corresponding cht-android PR: https://github.com/medic/cht-android/pull/394 but that seems to not have been reviewed yet, and is required to fully acceptance test this.

lorerod:
Thanks for looping me in, @dianabarsan !
Happy to help with acceptance testing, but as you mentioned, the Android PR still looks like a work in progress. Since that part is required to fully test the feature, I’ll hold off until it's ready and integrated. I’ll be keeping an eye on the status of the Android PR so I can jump in once it’s ready for testing.

jkuester:
Are we collecting any telemetry data about how often we are checking for notifications?  Basically, I would like to be able to tell from telemetry if notifications-checks are happening regularly throughout the day.  (Looking for measurable data to be able to tell if the background notification checks in cht-android are happening as expected...)  I did not see anything in the `TasksNotificationService`, but maybe I missed something... :thinking:

jkuester:
On a similar note to :point_up:, @jonathanbataire @dianabarsan what do you think about exposing the [`TelemetryService.record` function](https://github.com/medic/cht-core/blob/da5fe7f731303c6f69315b3992455125f539f91a/webapp/src/ts/services/telemetry.service.ts#L309) on the `AndroidApiService`?  This would let us push telemetry info from the Android app (as long as we can load the webapp in the WebView...).  Maybe this should be a separate PR, though... :thinking:   Let me know what you think and maybe I can get some new tickets logged around this.  (Might make some "Good First Issues"...)

jkuester:
@dianabarsan had an interesting though when we were discussing this functionality today and I wanted to continue the conversation here in case it is worth a pivot.  

The general idea is that the `fetchNotifications()` function could _pre-load future tasks_ and return them to the cht-android app.  So, the function could return tasks that were currently due _and_ tasks that will be due in the future (for some set time window).  Then, on the cht-android side, we could take the future tasks and cache them for future notifications.  

Then, the background worker in cht-android would not need a WebView/webapp to calculate the current notifications. When the cht-android app is closed, the notifications would only be powered based on the pre-loaded future notifications. This seems potentially viable since, in most cases, the user's data that would affect tasks is not be changing unless the user actually opens the app.  So, future tasks should remain accurate enough.

I do have a couple questions here:

1. When we calculate future tasks (by giving a future date to the rules engine), that will still result in new task docs getting written to the DB for those future tasks, right?
2. If the data in the DB changes between the time we calculate the task and the time the task is actually relevant such that the task is no longer relevant, will it be automatically resolved/deleted?  (I think so, right?)
3. Will suddenly pre-calculating a bunch of tasks trigger another task doc explosion?  :grimacing: (Seems like it will add a bit of initial load, but if we keep a reasonably short window it should not be a huge problem...)

dianabarsan:
Thanks @jkuester ! 
I tagged @jonathanbataire in a similar comment here: https://github.com/medic/cht-android/pull/394#issuecomment-3379519054

Just so we don't have the same conversation in multiple threads

Issue #9992, opened 2025-05-29
Remove haproxy-healthcheck service from single-node CouchDb deployments


dianabarsan:
**Describe the issue**
[Haproxy-healthcheck](https://github.com/medic/cht-core/tree/master/haproxy-healthcheck) is a service that will periodically query CouchDb servers and insure that all servers in a cluster are up.
This service is only required when CouchDb is clustered and not for CouchDb single node, but it is deployed even for single node. It requires its own container and is relatively low resource consuming. 

**Describe the improvement you'd like**
Remove haproxy-healthcheck from single-node deployments, and have haproxy use the `_up` endpoint directly. 

**Describe alternatives you've considered**
Leave it as it is.

PR #10006, opened 2025-06-02
feat(#9992): remove haproxy-healthcheck from single node deployments

dianabarsan:
# Description

#9992 

# Code review checklist

# License


dianabarsan:
@Hareet any change you have the bandwidth to give this a look?

dianabarsan:
It seems I had no luck with @hareet. @jkuester would you mind having a look?

jkuester:
:+1: got a couple others in the queue, but will add this one and try to have a look by EOW.

haproxy/entrypoint.sh
```diff
@@ -1,30 +1,46 @@
...
+      echo "  option httpchk"
+      echo "  http-check send meth GET uri /_up hdr Authorization 'Basic ${basic_auth}'" >> "$BACKEND"
+      echo "  http-check expect status 200" >> "$BACKEND"
+      config="  server ${SERVERS[0]} ${SERVERS[0]}:5984 check inter 5s"
+      setResolver "$config"
+    } >> "$BACKEND"
+else
+    for server in "${SERVERS[@]}"; do
+        [[ -z "$server" ]] && continue  # Skip empty entries
+        config="  server ${server} ${server}:5984 check agent-check agent-inter 5s agent-addr ${HEALTHCHECK_ADDR:-localhost} agent-port 5555"
```
jkuester, line 34: `${HEALTHCHECK_ADDR:-localhost}` is the default to `localhost` here intentional/necessary?  We always need to be calling out to the healthcheck container here, right?

haproxy/entrypoint.sh
```diff
@@ -1,30 +1,46 @@
...
-  echo "${server}" >> $BACKEND
-done
+cp "${BACKEND}.template" "$BACKEND"
+
+# Read servers into array, trim whitespace
+readarray -td, SERVERS < <(echo -n "${COUCHDB_SERVERS}" | tr -d '[:space:]')
 
-# Place environment variables into config
-envsubst < $DEFAULT
-envsubst < $BACKEND
```
jkuester, line 22: I am having trouble understanding what this was even for....  😬  Just wanted to double-check, though, that we do not need it now?

haproxy/tests/Makefile
```diff
@@ -1,16 +1,36 @@
...
 clean:
-	docker compose down --rmi local || :
+	docker compose -f $(COMPOSE_SINGLE) down --rmi local || :
+	docker compose -f $(COMPOSE_CLUSTER) down --rmi local || :
+
+# Cleanup handler
+.PHONY: emergency-cleanup
+emergency-cleanup:
+	-docker compose -f $(COMPOSE_SINGLE) down
+	-docker compose -f $(COMPOSE_CLUSTER) down
```
jkuester: Okay, I think I am missing something here (not super fluent in make), but how are these targets actually used? I don't see them referenced anywhere....

jkuester (approved):
Okay, so I am def no expert on bash/Haproxy/make, so just a few (probably silly questions), but otherwise these changes seems to do what was intended.  👍

haproxy/entrypoint.sh
dianabarsan: Yes, we always need to call out healthcheck here, and we should _always_ need to have  `HEALTHCHECK_ADDR` set, I think the `localhost` was probably used for testing at some point, or when network is set to be host mode in docker. 

I think that if we remove the default value and not set HEALTHCHECK_ADDR, we would get some weird config error in haproxy, and if we keep the value then we would get a better error that checking the server failed (because nothing runs on port 5555 on haproxy host). 
Maybe the best is to always require HEALTHCHECK_ADDR to be set. 

Honestly I just took this code from the previous version and didn't think about it at all.

haproxy/entrypoint.sh
dianabarsan: I don't know why this was needed either.

haproxy/tests/Makefile
dianabarsan: you're right. I think the purpose was to run them while debugging failing tests.

PR #10007, opened 2025-06-02
Task UI design.  And it also has Sam's New sorting Log

Edwin-Rugoogamu:
**Describe the solution**
 In cooperated the UI design changes .

1. The High Priority Tasks contain a red border 
2. The High Priority Label comes last On the Task layout.


webapp/src/css/inbox.less
```diff
@@ -1958,3 +1958,6 @@ mm-sidebar-menu .mat-sidenav-container {
     display: none;
   }
 }
+.high-priority {
+  border: 1px solid #d9534f !important; // Red border
```
dianabarsan, line 1962: There is an attempt to structure the CSS by concern, so first of all, this change should be in another file, not the generic `inbox.less`. This would be a better spot: https://github.com/medic/cht-core/blob/e33cd8738be9bbe1e3c34566fe07d7bf9fa9da6f/webapp/src/css/content-list.less#L127

Secondly, we should use the CHT color scheme. In the link above, I direct specifically to a line that uses the already defined high priority color.

webapp/src/ts/modules/tasks/tasks.component.html
```diff
@@ -9,6 +9,7 @@
           [attr.data-record-id]="task._id"
           [class.selected]="task.selected"
           [class.overdue]="task.overdue"
+          [class.high-priority]="task.priority >= 8"
```
dianabarsan, line 12: We should not hardcode this. 

There are several considerations here: 
1. Insure that projects can use priority scores _without_ the visual UI change. 
2. Insure that existent projects do not accidentally get this UI change. 
3. Make sure that the high level priority threshold is configurable. This configuration can be generic (for all tasks) or per task (this can be an additional field in the priority property),

webapp/src/css/inbox.less
Edwin-Rugoogamu: Okay Diana.  Thankyou for those insights .  Let me have a look at it and modify as required

PR #9960, opened 2025-05-09
feat(#9882): upgrade couchdb to version 3.5.0

sugat009:
# Description

Upgrades CouchDB to version 3.5.0

Closes #9882 

# Code review checklist

- [ ] UI/UX backwards compatible: Test it works for the new design (enabled by default). And test it works in the old design, enable `can_view_old_navigation` permission to see the old design. Test it has appropriate design for RTL languages. 
- [ ] Readable: Concise, well named, follows the [style guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/), documented if necessary.
- [ ] Documented: Configuration and user documentation on [cht-docs](https://github.com/medic/cht-docs/)
- [ ] Tested: Unit and/or e2e where appropriate
- [ ] Internationalised: All user facing text
- [ ] Backwards compatible: Works with existing data and configuration or includes a migration. Any breaking changes documented in the release notes.

# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

Issue #9882, opened 2025-04-07
Upgrade to latest version of Couch/Nouveau


jkuester:
**Is your feature request related to a problem? Please describe.**
Couch/Nouveau `3.4.3` has now been released. Currently our development has been targeting `3.4.2`.  

**Describe the solution you'd like**
Once we have the [9542_freetext_tco](https://github.com/medic/cht-core/tree/9542_freetext_tco) branch stable and the lifecycle code from https://github.com/medic/cht-core/issues/9691 working like we want, we should have a new PR to upgrade to Couch/Nouveau `3.4.3`.  In this process we fully validate the process of upgrading an instance with existing Nouveau indexes.

sugat009:
UPDATE: that target version has been changed to 3.5.0.

mrjones-plip:
Hey all - I updated the title to show 3.5.0 so the release notes are accurate. lemme know if that's not correct!

andrablaj:
Done via https://github.com/medic/cht-core/pull/9960

PR #10014, opened 2025-06-04
feat(#9882): upgrade couchDB version to 3.5.0

sugat009:
# Description

Upgrading CouchDB to version 3.5.0. This is a partial change for #9882 . This PR only upgrades CouchDB whereas the target is to include the same version of CouchDB Nouveau. There is another PR #9960 which has done upgrades to both CouchDB and CouchDB Nouveau, but we decided that CouchDB upgrade did not have to wait for the whole TCO effort to be merged which will be done in v5.0.0.

# Code review checklist

# License


sugat009:
That would work but there's another [PR](https://github.com/medic/cht-core/pull/9960) open to upgrade Nouveau as well. Should it be handled differently?

jkuester:
I am not so worried about https://github.com/medic/cht-core/pull/9960 as it is just merging to a base branch. Since this is going to `master`, we need to treat it as a full stand-alone feature (and it seems like it should have its own dedicated issue).  So, we can either repurpose https://github.com/medic/cht-core/issues/9882 or just log a new issue. :+1:

sugat009:
cool, I'll create another issue for this PR as I think that would be the least amount of effort.

Issue #9835, opened 2025-03-12
Add cht-datasource APIs for creation and update of contacts and reports


sugat009:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

jkuester:
I like the direction we are going with this design!  Couple of thoughts here:

1. We had to support and API for _reading_ generic contacts because there are certain application contexts where you have a contact `_id` value, but do not know if it is a person or place (so you need some way to get the data to check). However, I do not think we should support creating/updating via a generic contacts API.  Instead these should be supported via the person and place APIs.
2. You are currently modeling updates as a `PATCH`, which is definitely valid from an HTTP method perspective.  However, CouchDB only supports updates [via `PUT`](https://docs.couchdb.org/en/stable/api/document/common.html#put--db-docid).  (The difference being, with the `PUT` you have to pass the whole version of the new doc whereas with the `PATCH`, you only pass the changes.)  I am concerned that going with `PATCH` would add extra complexity to our API code (which would need to handle translating the PATCH changes to something Couch/Pouch can deal with). Additionally, the `PATCH` approach would be a pretty significant break in the current data model pattern in cht-core. It could make it more complex to _consume_ the cht-datasource API in different parts of the code-base.  I have happy to continue this discussion if you see particular value in going with `PATCH`, but otherwise I think I would lean towards supporting updates via `PUT`.
3. You don't explicitly mention it, but one interesting detail to consider is the parameter type that the `create`/`update` functions will accept in the cht-datasource code.  For example, for a person, the raw JSON data we want to end up sending to Couch should conform to the `Person` interface (where all the lineage is "de-hydrated"). So, it makes sense to accept that.  However, when actually doing things with person data and using the cht-datasource apis, it can be pretty easy to end up with `PersonWithLineage` objects that contain "hydrated" lineage.  A good example of this is when you go to write a new person, you probably already have a `parent` object containing the doc for the parent place. Then you want to create a new object for the new person with the parent doc set as the `parent` field on the new person doc. What you have done is just created a person with "hydrated" lineage (since the `parent` field contains more than just the `_id` and `parent` fields).  We could lock down our create/update APIs so that they only accept de-hydrated data, but that will just force the consuming code to handle the de-hydration process.  I think it would probably be work accepting either hydrated or de-hydrated data in our cht-datasource create/update APIs and then the cht-datasource code can handle properly de-hydrating before writing to Couch.

[...]

PR #10015, opened 2025-06-04
feat(#9835): add `ContactQualifier` functions

apoorvapendse:
This is a WIP version for adding the `ContactQualifer`.

# Description

# Code review checklist

# License

### DIFF
--- a/shared-libs/cht-datasource/src/qualifier.ts
+++ b/shared-libs/cht-datasource/src/qualifier.ts
@@ -1,4 +1,4 @@
-import { isString, hasField, isRecord, Nullable } from './libs/core';
+import { isString, hasField, isRecord, Nullable, hasFields } from './libs/core';
 import { InvalidArgumentError } from './libs/error';
 
 /**
@@ -139,4 +139,72 @@ export const and = <
     ...(qualifierC ?? {}),
     ...(qualifierD ?? {}),
   };
+};
+
+/** 
+ * A qualifier for a contact
+ */
+type ContactQualifier = Readonly<{
+  type: string,
+  name: string,
+  reported_date?: string | number,
+  _id?: string,
+  _rev?: string
+}>;
+
+/**
+ * Builds a qualifier for creation and update of a contact with
+ * the given fields.
+ * @param data object containing the fields for a contact
+ * @returns the contact qualifier
+ * @throws Error if data is not an object
+ * @throws Error if type is not provided or is empty
+ * @throws Error if name is not provided or is empty
+ * @throws Error if reported_date is not in a valid format. 
+ * Valid formats are 'YYYY-MM-DDTHH:mm:ssZ', 'YYYY-MM-DDTHH:mm:ss.SSSZ', or <unix epoch>.
+ */
+export const byContactQualifier = (data: unknown): ContactQualifier => {
+  if (!isRecord(data)){
+    throw new InvalidArgumentError('Invalid "data": expected an object.');
+  }
+  const qualifier = {...data};
+  if ('reported_date' in qualifier && !isValidReportedDate(qualifier.reported_date)){
+    throw new InvalidArgumentError(
+      // eslint-disable-next-line max-len
+      `Invalid reported_date. Expected format to be 'YYYY-MM-DDTHH:mm:ssZ', 'YYYY-MM-DDTHH:mm:ss.SSSZ', or a Unix epoch.`
+    );
+  }
+  if (!isContactQualifier(qualifier)){
+    throw new InvalidArgumentError(`Missing required fields [${JSON.stringify(data)}].`);
+  }
+  return qualifier;
+};
+
+/**
+ * Returns `true` if the given qualifier is a {@link ContactQualifier} otherwise `false`.
+ * @param qualifier the qualifier to check
+ * @returns `true` if the given type is a {@link ContactQualifier}, otherwise `false`.
+ */
+export const isContactQualifier = (qualifier: unknown): qualifier is ContactQualifier => {
+  if (isRecord(qualifier) && hasFields(qualifier, [{name: 'type', type: 'string'}, {name: 'name', type: 'string'}])){
+    if ('reported_date' in qualifier && !isValidReportedDate(qualifier.reported_date)){
+      return false;
+    }
+    return true;
+  }
+  return false;
+};
+
+/** @internal */
+const isValidReportedDate = (value: unknown): boolean => {
+  if (typeof value === 'number') {
+    return Number.isFinite(value);
+  }
+
+  if (typeof value === 'string') {
+    const isoRegex = /^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d{3})?Z)$/;
+    return isoRegex.test(value);
+  }
+
+  return false;
 };
--- a/shared-libs/cht-datasource/test/qualifier.spec.ts
+++ b/shared-libs/cht-datasource/test/qualifier.spec.ts
@@ -4,7 +4,9 @@ import {
   byUuid,
   isContactTypeQualifier,
   isFreetextQualifier,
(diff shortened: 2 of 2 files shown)


shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -140,3 +140,66 @@ export const and = <
...
+const isContactQualifier = (qualifier: unknown) : qualifier is ContactQualifier => {
+  if (isRecord(qualifier) && hasFields(qualifier, [{name: 'type', type: 'string'}, {name: 'name', type: 'string'}])){
+    if ('reported_date' in qualifier){
+      const date = qualifier.reported_date;
+      if (!isValidReportedDate(date)) {
+        throw new InvalidArgumentError(
+          `Invalid reported_date format. Expected 'YYYY-MM-DDTHH:mm:ssZ', 'YYYY-MM-DDTHH:mm:ss.SSSZ', or a Unix epoch.`
+        );
+      }
+    }
```
apoorvapendse, line 192: Just returning `false` from `isContactQualifier` might not be helpful to determine what kind of error to throw in `byContactQualifer`. Which is why I throw an error here, which doesn't seem to be the best way to do this.
One solution might be doing the type checks and date formatting checks individually inside `byContactQualifer` without reusing `isContactQualifier`, to prevent ambiguity regarding which error to throw for the generic `false` return we would get on reusing `isContactQualifier`.

apoorvapendse: I believe we want to throw a different error based on the input.
I've identified two types:
1. Missing required `name` or `type` fields.
2. Invalid date format.

apoorvapendse: This was discussed on today's call and here is the summary:
1. We perform the date validation before calling `isContactQualifier` in `byContactQualifier` to throw the correct exception.
2. We will only return a boolean value from `isContactQualifier` similar to other `is.*` functions.

Thanks @sugat009 for unblocking me on this.

apoorvapendse:
Hi @sugat009, I've updated the commit as per the [discussion](https://github.com/medic/cht-core/pull/10015/#discussion_r2126094597).
Please take a look when you find time.

Also should this be merged to some other branch?

sugat009:
@apoorvapendse Can you target this as the base branch for all the consecutive PRs? `9835-add-cht-datasource-apis-for-creation-and-update-of-contacts-and-reports`

Also, the type of this PR should be `feat` and not `chore`.

apoorvapendse:
Updated the base branch, commit and PR message.

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -140,3 +140,66 @@ export const and = <
...
+type ContactQualifier = Readonly<{
+  type: string,
+  name: string,
+  reported_date?: string | number,
+  _id?: string,
+  _rev?: string
+}>;
+
+/**
+ * Builds a qualifier for creation and updation of a contact with
```
sugat009:
```suggestion
 * Builds a qualifier for creation and update of a contact with
```

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -140,3 +140,66 @@ export const and = <
...
+ * @throws Error if type is not provided or is empty
+ * @throws Error if name is not provided or is empty
+ * @throws Error if reported_date is not in a valid format. 
+ * Valid formats are 'YYYY-MM-DDTHH:mm:ssZ', 'YYYY-MM-DDTHH:mm:ss.SSSZ', or <unix epoch>.
+ */
+export const byContactQualifier = (data: object) :ContactQualifier => {
+  const qualifier = {...data};
+  if ('reported_date' in qualifier && !isValidReportedDate(qualifier.reported_date)){
+    throw new InvalidArgumentError(
+      `Invalid reported_date format. Expected 'YYYY-MM-DDTHH:mm:ssZ', 'YYYY-MM-DDTHH:mm:ss.SSSZ', or a Unix epoch.`
```
sugat009:
```suggestion
      `Invalid reported_date. Expected format to be 'YYYY-MM-DDTHH:mm:ssZ', 'YYYY-MM-DDTHH:mm:ss.SSSZ', or a Unix epoch.`
```

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -140,3 +140,66 @@ export const and = <
...
+ * Builds a qualifier for creation and updation of a contact with
+ * the given fields.
+ * @param data object containing the fields for a contact
+ * @returns the contact qualifier
+ * @throws Error if type is not provided or is empty
+ * @throws Error if name is not provided or is empty
+ * @throws Error if reported_date is not in a valid format. 
+ * Valid formats are 'YYYY-MM-DDTHH:mm:ssZ', 'YYYY-MM-DDTHH:mm:ss.SSSZ', or <unix epoch>.
+ */
+export const byContactQualifier = (data: object) :ContactQualifier => {
```
sugat009:
```suggestion
export const byContactQualifier = (data: unknown): ContactQualifier => {
```

sugat009: This is because we are verifying stuff and we need to assume that `data` could be anything.
_I'm aware that in the implementation details in the ticket it was `object`._

shared-libs/cht-datasource/test/qualifier.spec.ts
```diff
@@ -109,4 +111,80 @@ describe('qualifier', () => {
       });
     });
   });
+  describe('byContactQualifier', () => {
```
sugat009, line 115:
```suggestion

  describe('byContactQualifier', () => {
```

sugat009: space before a code block

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -140,3 +140,71 @@ export const and = <
...
+  _rev?: string
+}>;
+
+/**
+ * Builds a qualifier for creation and update of a contact with
+ * the given fields.
+ * @param data object containing the fields for a contact
+ * @returns the contact qualifier
+ * @throws Error if data is not an object
+ * @throws Error if type is not provided or is empty
```
apoorvapendse, line 161: Based this [comment](https://github.com/medic/cht-core/pull/10015#discussion_r2126791174), I think we should include this error too.
Does that seem fine?

sugat009: Yes, this and the error for the name not existing is included in the exception thrown in line 177-178, right?

apoorvapendse: Line [168](https://github.com/medic/cht-core/pull/10015/files#diff-e8efbeca6ec5922fb090952571f5b6ee11fb68a4f177116de42e99462395426cR168) for this and [178](https://github.com/medic/cht-core/pull/10015/files#diff-e8efbeca6ec5922fb090952571f5b6ee11fb68a4f177116de42e99462395426cR178) for missing `name` or `type`.

sugat009: I think I misunderstood something here.

> @throws Error if type is not provided or is empty

I'm simply saying that this highlighted error is already cover in line 177 because this is the line in which the comment is tagged. 

As for the [change](https://github.com/medic/cht-core/pull/10015#discussion_r2126791174) of type of argument `data` from `object` to `unknown`, the check that was added at the beginning of the function is nice as well.

sugat009 (approved):

shared-libs/cht-datasource/src/qualifier.ts
apoorvapendse: I meant I added this error, which wasn't part of the original TSDoc.
>   @throws Error if data is not an object

PR #10022, opened 2025-06-05
feat(#9835): add `ReportQualifier` functions and update `hasField`

apoorvapendse:
# Description
Commit 1 adds the `ReportQualifier` based on the description in #9835.
An additional error is included in cases where the `data` is not a valid `object`.

Commit 2 tweaks `hasField` and `hasFields` to perform an optional
truthy check on values, which is utilized to detect empty strings 
in `isContactQualifier` and `isReportQualifier` to return `false`
for such field values.

# Code review checklist

# License

### DIFF
--- a/shared-libs/cht-datasource/src/libs/core.ts
+++ b/shared-libs/cht-datasource/src/libs/core.ts
@@ -87,16 +87,19 @@ export const isRecord = (value: unknown): value is Record<string, unknown> => {
 /** @internal */
 export const hasField = <T extends Record<string, unknown>>(
   value: T,
-  field: { name: keyof T, type: string }
+  field: { name: keyof T, type: string, ensure_truthy_value?: boolean},
 ): value is T & Record<typeof field.name, string> => {
   const valueField = value[field.name];
+  if (field.ensure_truthy_value) {
+    return typeof valueField === field.type && !!valueField;
+  }
   return typeof valueField === field.type;
 };
 
 /** @internal */
 export const hasFields = (
   value: Record<string, unknown>,
-  fields: NonEmptyArray<{ name: string, type: string }>
+  fields: NonEmptyArray<{ name: string, type: string, ensure_truthy_value?: boolean}>,
 ): boolean => fields.every(field => hasField(value, field));
 
 /** @internal */
--- a/shared-libs/cht-datasource/src/libs/core.ts
+++ b/shared-libs/cht-datasource/src/libs/core.ts
@@ -84,22 +84,28 @@ export const isRecord = (value: unknown): value is Record<string, unknown> => {
   return value !== null && typeof value === 'object';
 };
 
+interface FieldDescriptor<T> {
+  name: keyof T;
+  type: string;
+  ensureTruthyValue?: boolean
+}
+
 /** @internal */
 export const hasField = <T extends Record<string, unknown>>(
   value: T,
-  field: { name: keyof T, type: string, ensure_truthy_value?: boolean},
-): value is T & Record<typeof field.name, string> => {
-  const valueField = value[field.name];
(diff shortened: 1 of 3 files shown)


shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -195,6 +195,70 @@ export const isContactQualifier = (qualifier: unknown): qualifier is ContactQual
...
+  if (isRecord(qualifier) && 
+      hasFields(qualifier, [{name: 'type', type: 'string'}, {name: 'form', type: 'string'}])
+  ){
+    if (
+      typeof qualifier.type !== 'string' || qualifier.type.trim() === '' ||
+      typeof qualifier.form !== 'string' || qualifier.form.trim() === ''
+    ) {
+      return false;
+    }
+
```
apoorvapendse: I missed doing this check for required fields that might be empty in `isContactQualifier`.
Should I add that fix as a part of this PR in a new commit?

shared-libs/cht-datasource/src/libs/core.ts
```diff
@@ -87,16 +87,19 @@ export const isRecord = (value: unknown): value is Record<string, unknown> => {
 /** @internal */
 export const hasField = <T extends Record<string, unknown>>(
   value: T,
-  field: { name: keyof T, type: string }
+  field: { name: keyof T, type: string, ensure_truthy_value?: boolean},
```
sugat009:
```suggestion
  field: { name: keyof T, type: string, ensureTruthyValue?: boolean},
```
We use `camelCase` for [variables](https://docs.communityhealthtoolkit.org/community/contributing/code/style-guide/#use-lowercamelcase-for-variables-properties-and-function-names). Can you make changes wherever necessary accordingly?

sugat009: issue: instead of having this variable as an optional field, let's make it a default parameter with the default value of `false`.

sugat009 (changes requested):

shared-libs/cht-datasource/test/qualifier.spec.ts
```diff
@@ -136,26 +138,32 @@ describe('qualifier', () => {
...
         {
           name: 'A'
         },
+        {
+          name: '', type: 'person'
+        },
         {
           type: 'person', reported_date: '2025-06-03T12:45:30Z'
         }
       ].forEach((quantifier) => expect(() => byContactQualifier(quantifier))
```
sugat009:
```suggestion
      ].forEach((qualifier) => expect(() => byContactQualifier(qualifier))
```

shared-libs/cht-datasource/test/qualifier.spec.ts
```diff
@@ -136,26 +138,32 @@ describe('qualifier', () => {
...
-        .to.throw(`Missing required fields [${JSON.stringify(quantifier)}].`));
+        .to.throw(`Missing or empty required fields [${JSON.stringify(quantifier)}].`));
```
sugat009:
```suggestion
        .to.throw(`Missing or empty required fields [${JSON.stringify(qualifier)}].`));
```

shared-libs/cht-datasource/test/qualifier.spec.ts
```diff
@@ -188,4 +196,75 @@ describe('qualifier', () => {
       ].forEach((quantifier) => expect(isContactQualifier(quantifier)).to.be.true);
```
sugat009:
```suggestion
      ].forEach((qualifier) => expect(isContactQualifier(qualifier)).to.be.true);
```

apoorvapendse: oops

PR #10026, opened 2025-06-06
feat(#10019): refactor shared-libs/lineage library to use cht-datasource

### LINKED ISSUE #10019: Refactor code that is loading contacts via shared-libs/lineage to use cht-datasource
... when possible (additional common logic around structuring the contact data should remain in shared-libs/lineage)

@shashi-sah2003 

**Additional context**

shashi-sah2003:
# Description

This pull request refactors the lineage module to replace direct database queries with calls to a new dataContext object, improving modularity and testability. It also updates the associated unit tests to mock the dataContext methods.

fixes #10019 

# License

### DIFF
--- a/shared-libs/lineage/src/hydration.js
+++ b/shared-libs/lineage/src/hydration.js
@@ -46,7 +46,7 @@ const getContactIds = (contacts) => {
   return _.uniq(ids);
 };
 
-module.exports = function(Promise, DB) {
+module.exports = function(Promise, DB, dataContext) {
   const fillParentsInDocs = function(doc, lineage) {
     if (!doc || !lineage.length) {
       return doc;
@@ -114,10 +114,11 @@ module.exports = function(Promise, DB) {
       }
     });
 
-    return fetchDocs(contactsToFetch)
-      .then(function(fetchedContacts) {
-        return lineageContacts.concat(fetchedContacts);
-      });
+    const getContact = dataContext.Contact.v1.get;
+
+    return Promise.all(contactsToFetch.map(id => getContact(id)))
+      .then(fetchedContacts => lineageContacts.concat(fetchedContacts.filter(Boolean)));
+    
   };
 
   const mergeLineagesIntoDoc = function(lineage, contacts, patientLineage, placeLineage) {
@@ -229,17 +230,10 @@ module.exports = function(Promise, DB) {
   };
 
   const fetchLineageById = function(id) {
-    const options = {
-      startkey: [id],
-      endkey: [id, {}],
-      include_docs: true
-    };
-    return DB.query('medic-client/docs_by_id_lineage', options)
-      .then(function(result) {
-        return result.rows.map(function(row) {
-          return row.doc;
-        });
-      });
+    if (dataContext.Contact?.v1?.getWithLineage){
+      return dataContext.Contact.v1.getWithLineage(id)
+        .then(lineageArr => Array.isArray(lineageArr) ? lineageArr : []);
+    }
   };
 
   const fetchLineageByIds = function(ids) {
@@ -257,9 +251,9 @@ module.exports = function(Promise, DB) {
   };
 
   const fetchDoc = function(id) {
-    return DB.get(id)
-      .catch(function(err) {
-        if (err.status === 404) {
+    return dataContext.Contact.v1.get(id)
+      .catch(function(err){
+        if (err.status === 404){
           err.statusCode = 404;
         }
         throw err;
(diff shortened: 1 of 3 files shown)


shashi-sah2003:
@sugat009  @jkuester  I have tried to make changes for `shared-libs/lineage`, the main changes was in the `hydration.js` file.
I have made changes to few functions but I'm a little confused with `contactUuidByShortcode` and `fetchHydratedDocs` in these functions [here](https://github.com/medic/cht-core/blob/master/shared-libs/lineage/src/hydration.js), what do you think about these functions, their definition and utility in `cht-datasource` doesn't align, so should I left these as it is directly querying database?

One more doubt is also there what should be the value of `dataContext`? For now I have passed it in lineage parameter but that's not working entirely.

sugat009:
@shashi-sah2003 Can you checkout this [sample](https://github.com/medic/cht-core/blob/9542_freetext_tco/shared-libs/search/src/search.js) from another `shared-lib` that uses `cht-datasource`.

> I have made changes to few functions but I'm a little confused with contactUuidByShortcode and fetchHydratedDocs in these functions [here](https://github.com/medic/cht-core/blob/master/shared-libs/lineage/src/hydration.js), what do you think about these functions, their definition and utility in cht-datasource doesn't align, so should I left these as it is directly querying database?

Do you mean that the functions from `cht-datasource` do not provide enough data for replacement or the opposite? I've not checked them out in too much detail myself but can you compare the input and output between the functions as a start?

shashi-sah2003:
@sugat009 for the `fetchHydratedDocs` function I am thinking of using `Promise.all` with `getWithLineage` utility but for the `contactUuidByShortcode` function I amn't able to find any direct implementation of utility in [contact.ts](https://github.com/medic/cht-core/blob/master/shared-libs/cht-datasource/src/contact.ts) (pls correct me if I am wrong). I have also seen your code [here](https://github.com/medic/cht-core/blob/9542_freetext_tco/shared-libs/search/src/search.js) which is using `dataContext` as a parameter, I am also using `dataContext` as a parameter. I have a question regarding this, the dataContext is being set to `LocalContext` or `RemoteContext` based on the online availability.

Don't we need to also pass `dataContext` when calling that function if yes then how `dataContext` needs to be initialized? Also it seems the ci pipeline is failing due to undefined dataContext so most probably the dataContext isn't being passed correctly?

sugat009:
@shashi-sah2003 Yes, you are correct. `contactUuidByShortcode` function queries the `medic-client/contacts_by_reference` view which has a different purpose and is not implemented in `cht-datasource`. You can leave this as it is for now.

> Don't we need to also pass dataContext when calling that function if yes then how dataContext needs to be initialized? Also it seems the ci pipeline is failing due to undefined dataContext so most probably the dataContext isn't being passed correctly?

You'll need to check where `shared-libs/lineage` is being called from and send in the `dataContext` similar to how `shared-libs/search` is called. Here are some samples:
1. https://github.com/medic/cht-core/blob/9542_freetext_tco/api/src/services/export/report-mapper.js
2. https://github.com/medic/cht-core/blob/9542_freetext_tco/api/src/services/export/contact-mapper.js
3. https://github.com/medic/cht-core/blob/b732aa3f6e049474b889bebfe63696cde1c57923/webapp/src/ts/services/search.service.ts#L135

Usually, a `datacontext` already exists there but if there isn't you can create it like in the above samples and pass it into lineage lib.

shashi-sah2003:
@sugat009 thanx for telling me how to use `dataContext`. In the [index.ts](https://github.com/medic/cht-core/blob/master/shared-libs/cht-datasource/src/index.ts) it has been mentioned to use `dataContext` in a different way (see 10-12), what do you think about that? is that also correct way for this or that just defines what `dataContext` actually means? if not should we change that to importing dataContext from `api/src/services`?

Also is something wrong with ci/cd pipeline? It seems unusual error to me.

sugat009:
@shashi-sah2003 There is no right or wrong way to initialize `cht-datasource`. It all depends on the situation like if you need a remote adapter or local adapter. I'd just go with how `shared-libs/search` is being initialized, unless this requires something extra. As we know that it will work and the `shared-libs/lineage` will be used in the same manner.

> Also is something wrong with ci/cd pipeline? It seems unusual error to me.

I've never seen this one. I'm looking into this.

shashi-sah2003:
@sugat009 Did you find the solution of above ci failure?

sugat009:
@shashi-sah2003 Can you rename your branch to a simpler name? Your branch can not be checked out into.

image

shashi-sah2003:
@sugat009 can you try with something like this `git checkout "10019-refactor(shared-libs/lineage)-to-use-(cht-datasource)"` wrapping the branch name in inverted comma. Let me know if it works or not, if not I will close this PR and raise a new PR with simpler branch name? or you can directly checkout using PR `gh pr checkout 10026` can be found on top right side of this page when clicked on code.

sugat009:
@shashi-sah2003 This is not happening locally. I've rerun the CI, if it fails, create a new branch.

PR #10028, opened 2025-06-06
feat(#9835): `createDoc` function for PouchDB.

apoorvapendse:
# Description

WIP `createDoc` function to insert documents in PouchDB.
Related to #9835.

# Code review checklist

# License


shared-libs/cht-datasource/src/local/libs/doc.ts
```diff
@@ -175,3 +175,19 @@ export const fetchAndFilterUuids = (
...
+  if (!ok) {
+    throw new Error('Error creating document.');
+  }
+  const createdDoc = (db as PouchDB.Database<Doc>)
+    .get(id)
+    .then(doc => {
+      return doc;
+    }).catch(() => {
+      throw new Error('Failed to fetch created document.');
+    });
```
sugat009: question: we can you the `async/await` pattern here as well, right? and enclose it with a `try/catch` block and return the same error.

shared-libs/cht-datasource/src/local/libs/doc.ts
```diff
@@ -175,3 +175,19 @@ export const fetchAndFilterUuids = (
     limit
   );
 };
+
+/** @internal */
+const createDoc = async(db: PouchDB.Database, data:Record<string, unknown>) : Promise<Doc> => {
```
sugat009:
```suggestion
const createDoc = (db: PouchDB.Database) => async (data:Record<string, unknown>) : Promise<Nullable<Doc>> => {
```

sugat009: Here, we should apply the concept of `currying` to store the database object in the createDoc function and use that function to return another function which can be used multiple times later on.

sugat009 (changes requested):

shared-libs/cht-datasource/src/local/libs/doc.ts
```diff
@@ -175,3 +175,12 @@ export const fetchAndFilterUuids = (
     limit
   );
 };
+
+/** @internal */
+export const createDoc = (db: PouchDB.Database) => async (data: Record<string, unknown>): Promise<Nullable<Doc>> => {
```
sugat009, line 180:
```suggestion
export const createDoc = (db: PouchDB.Database<Doc>) => async (data: Record<string, unknown>): Promise<Nullable<Doc>> => {
```

if you change here then, the type cast in line 185 is not necessary

shared-libs/cht-datasource/test/local/libs/doc.spec.ts
```diff
@@ -587,4 +591,39 @@ describe('local doc lib', () => {
       expect(() => fetchAndFilterUuids(getFunction, 3)).to.throw('API Error');
     });
   });
+
+
```
sugat009:
```suggestion
```

shared-libs/cht-datasource/test/local/libs/doc.spec.ts
```diff
@@ -587,4 +591,39 @@ describe('local doc lib', () => {
...
+  describe('createDoc', () => {
+    it('should create and retrieve the doc', async () => {
```
sugat009, line 617: issue: another test that tests when the database returns `not ok` should be added

sugat009 (changes requested):

shared-libs/cht-datasource/src/local/libs/doc.ts
apoorvapendse: But it is not certain that what we are posting is a `Doc` right?

shared-libs/cht-datasource/test/local/libs/doc.spec.ts
apoorvapendse: Done.

shared-libs/cht-datasource/src/local/libs/doc.ts
sugat009: ahh! yes but `db: PouchDB.Database<Doc>` does not mean that the data we are going is posting is going to be a `Doc`. It indicates that the database that is used stores the documents in the form of `Doc` and when it returns something it will be of type `Doc`.

apoorvapendse:
```
Argument of type 'Record<string, unknown>' is not assignable to parameter of type 'PostDocument<Doc & Record<string, unknown>>'.
  Type 'Record<string, unknown>' is missing the following properties from type 'Doc': _rev, _idts(2345)
 ```
 
 This is what I see when I try do that, not sure if I am missing something.

apoorvapendse: Basically the object we pass to Pouch won't contain a `_rev` and `_id` right?

sugat009: OK, got it. The `db` expects the passed in value to be a `Doc` as well. You can leave this as it is for now.

Issue #10036, opened 2025-06-10
To have API that can create people


sugat009:
@apoorvapendse 
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

apoorvapendse:
I plan to pick this up next, once #10028 is merged.
Just noting down what we discussed on 10th June's call:
1. Throw an error is the provided object's parent hierarchy isn't minified, when the parent's lineage contains anything other than an `_id` or `_rev`.
2. Optimistically assume that the lineage provided is accurate in terms of the parent ids as well the depth of the hierarchy.

Paraphrasing the reasoning behind this as @jkuester mentioned:
The reason behind staying rigid initially and keeping it simple is that we can expand `cht-datasource`'s compatibility with other "complex-to-process" data later on, but doing that in the other direction(lenient to rigid) would introduce breaking changes.

cc @jkuester, @sugat009

apoorvapendse:
I'd like to know how we want to model `createPerson` in shared-libs/cht-datasource/src/local/person.ts
Should I create a `PersonQualifier` that extends `ContactQualifier` which would be created using something like `byPersonQualifier` in shared-libs/cht-datasource/src/person.ts before calling `createPerson` in `local` /`remote` or is there some other way that is preferred?

I think the `PersonQualifier` would look something like this:
```typescript
type PersonQualifier = ContactQualifier & Readonly<{
  parent: NormalizedParent,
  date_of_birth?: Date;
  phone?: string;
  patient_id?: string;
  sex?: string;
}>
```

PR #10043, opened 2025-06-11
feat(#10036): add `PersonQualifier` and related functions

apoorvapendse:
# Description
WIP to integrate `PersonQualifier` to eventually add support for creating `person` documents.

Related to #10036.

# Code review checklist

# License

### DIFF
--- a/shared-libs/cht-datasource/src/qualifier.ts
+++ b/shared-libs/cht-datasource/src/qualifier.ts
@@ -1,4 +1,4 @@
-import { isString, hasField, isRecord, Nullable, hasFields } from './libs/core';
+import { isString, hasField, isRecord, Nullable, hasFields, NormalizedParent, isNormalizedParent } from './libs/core';
 import { InvalidArgumentError } from './libs/error';
 
 /**
@@ -266,4 +266,100 @@ const isValidReportedDate = (value: unknown): boolean => {
   }
 
   return false;
+};
+
+/** 
+ * A qualifier for a person
+ */
+type PersonQualifier = ContactQualifier & Readonly<{
+  parent: NormalizedParent;
+  date_of_birth?: Date;
+  phone?: string;
+  patient_id?: string;
+  sex?: string;
+  contact_type?: string
+}>
+
+/**
+ * Builds a qualifier for creation and update of a person with
+ * the given fields.
+ * @param data object containing the fields for a person
+ * @returns the person qualifier
+ * @throws Error if data is not an object
+ * @throws Error if type is not provided or is empty
+ * @throws Error if name is not provided or is empty
+ * @throws Error if parent is not provided or is empty
+ * @throws Error if parent is not in a valid de-hydrated format. 
+ * @throws Error if reported_date is not in a valid format. 
+ * Valid formats are 'YYYY-MM-DDTHH:mm:ssZ', 'YYYY-MM-DDTHH:mm:ss.SSSZ', or <unix epoch>.
+ */
+export const byPersonQualifier = (data: unknown): PersonQualifier => {
+  if (!isRecord(data)) {
+    throw new InvalidArgumentError('Invalid "data": expected an object.');
+  }
+  const qualifier = { ...data };
+  if ('reported_date' in qualifier && !isValidReportedDate(qualifier.reported_date)) {
+    throw new InvalidArgumentError(
+      // eslint-disable-next-line max-len
+      `Invalid reported_date. Expected format to be 'YYYY-MM-DDTHH:mm:ssZ', 'YYYY-MM-DDTHH:mm:ss.SSSZ', or a Unix epoch.`
+    );
+  }
+  if (!isContactQualifier(qualifier) || !hasField(data, { name: 'parent', type: 'object' })) {
+    throw new InvalidArgumentError(`Missing or empty required fields [${JSON.stringify(data)}].`);
+  }
+  if (!isNormalizedParent(data.parent)) {
+    throw new InvalidArgumentError(`Missing required fields in the parent hierarchy [${JSON.stringify(data)}].`);
+  }
+
+  if (data.type === 'contact' && !hasField(data, { name: 'contact_type', type: 'string' })) {
+    throw new InvalidArgumentError(`Missing or empty required fields [${JSON.stringify(data)}].`);
+  } else if (!(data.type === 'person')) {
+    throw new InvalidArgumentError('Expected `type` to be `person`.');
+  }
+
+  // Ensure parent lineage doesn't have any additional properties other than `_id` and `parent`.
+  let parent = data.parent;
+  while (parent) {
+    if (Object.keys(parent).length > 2) {
+      // This means that the parent certainly has extra fields and is not minfied/de-hydrated as per
+      // our liking as `isNormalized` check ensures that it does have two keys `_id` and `parent`.
+      throw new InvalidArgumentError(`Additional fields found in the parent lineage [${JSON.stringify(data)}].`);
+    }
+    parent = data.parent;
+  }
+
+  return qualifier as PersonQualifier;
+};
+
+/** @internal */
+export const isPersonQualifier = (data: unknown): data is PersonQualifier => {
+  if (!isRecord(data)) {
+    return false;
+  }
+  if (!hasField(data, { name: 'parent', type: 'object' }) || !isNormalizedParent(data.parent)) {
+    return false;
+  }
+
+  // Ensure parent lineage doesn't have any additional properties other than `_id` and `parent`.
+  let parent = data.parent;
+  while (parent) {
+    if (Object.keys(parent).length > 2) {
+      // This means that the parent certainly has extra fields and is not minfied/de-hydrated as per
+      // our liking as `isNormalized` check ensures that it does have two keys `_id` and `parent`.
+      return false;
+    }
+    parent = data.parent;
+  }
+
+  if (!isContactQualifier(data)) {
+    return false;
+  }
+
+  if (data.type === 'contact' && !hasField(data, { name: 'contact_type', type: 'string' })) {
+    return false;
+  } else if (!(data.type === 'person')) {
+    return false;
+  }
+
+  return true;
 };


shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -267,3 +267,99 @@ const isValidReportedDate = (value: unknown): boolean => {
...
+    parent = data.parent;
+  }
+
+  if (!isContactQualifier(data)) {
+    return false;
+  }
+
+  if (data.type === 'contact' && !hasField(data, { name: 'contact_type', type: 'string' })) {
+    return false;
+  } else if (!(data.type === 'person')) {
```
sugat009:
```suggestion
  } else if (data.type !== 'person') {
```

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -267,3 +267,99 @@ const isValidReportedDate = (value: unknown): boolean => {
...
+  const qualifier = { ...data };
+  if ('reported_date' in qualifier && !isValidReportedDate(qualifier.reported_date)) {
+    throw new InvalidArgumentError(
+      // eslint-disable-next-line max-len
+      `Invalid reported_date. Expected format to be 'YYYY-MM-DDTHH:mm:ssZ', 'YYYY-MM-DDTHH:mm:ss.SSSZ', or a Unix epoch.`
+    );
+  }
+  if (!isContactQualifier(qualifier) || !hasField(data, { name: 'parent', type: 'object' })) {
+    throw new InvalidArgumentError(`Missing or empty required fields [${JSON.stringify(data)}].`);
+  }
```
sugat009: question: can the `byContactQualifier` function be reused here?

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -267,3 +267,99 @@ const isValidReportedDate = (value: unknown): boolean => {
...
+      // This means that the parent certainly has extra fields and is not minfied/de-hydrated as per
+      // our liking as `isNormalized` check ensures that it does have two keys `_id` and `parent`.
+      return false;
+    }
+    parent = data.parent;
+  }
+
+  if (!isContactQualifier(data)) {
+    return false;
+  }
```
sugat009: issue: this should be at the top of the file because the first two checks are included in that function

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -267,3 +267,99 @@ const isValidReportedDate = (value: unknown): boolean => {
...
+  if (!isNormalizedParent(data.parent)) {
+    throw new InvalidArgumentError(`Missing required fields in the parent hierarchy [${JSON.stringify(data)}].`);
+  }
+
+  if (data.type === 'contact' && !hasField(data, { name: 'contact_type', type: 'string' })) {
+    throw new InvalidArgumentError(`Missing or empty required fields [${JSON.stringify(data)}].`);
+  } else if (!(data.type === 'person')) {
+    throw new InvalidArgumentError('Expected `type` to be `person`.');
+  }
```
sugat009: issue: duplicate code with lines 358-362, refactor into a function

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -267,3 +267,99 @@ const isValidReportedDate = (value: unknown): boolean => {
...
+  // Ensure parent lineage doesn't have any additional properties other than `_id` and `parent`.
+  let parent = data.parent;
+  while (parent) {
+    if (Object.keys(parent).length > 2) {
+      // This means that the parent certainly has extra fields and is not minfied/de-hydrated as per
+      // our liking as `isNormalized` check ensures that it does have two keys `_id` and `parent`.
+      throw new InvalidArgumentError(`Additional fields found in the parent lineage [${JSON.stringify(data)}].`);
+    }
+    parent = data.parent;
+  }
```
sugat009: issue: duplicate code with lines 344-352, refactor into a function

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -267,3 +267,99 @@ const isValidReportedDate = (value: unknown): boolean => {
...
+  // Ensure parent lineage doesn't have any additional properties other than `_id` and `parent`.
+  let parent = data.parent;
+  while (parent) {
+    if (Object.keys(parent).length > 2) {
+      // This means that the parent certainly has extra fields and is not minfied/de-hydrated as per
+      // our liking as `isNormalized` check ensures that it does have two keys `_id` and `parent`.
+      return false;
+    }
+    parent = data.parent;
+  }
```
sugat009: question: this is more of a design decision. Personally, I think it's cleaner to have just a immediate parent as an field and not allow more nests. We could add the grandparents and so on from the server side. There's no benefit of the current way other than not having to query the database for the grandparents.

@jkuester What are you thoughts on this?

sugat009 (changes requested):

shared-libs/cht-datasource/src/qualifier.ts
apoorvapendse: That causes TypeScript to think that `data` is a ContactQualifer, which results in a lot of assertions for further checks.
Is there some way to deal with that?

shared-libs/cht-datasource/src/qualifier.ts
apoorvapendse: Same [problem](https://github.com/medic/cht-core/pull/10043#discussion_r2140463550) mentioned here.

shared-libs/cht-datasource/src/qualifier.ts
apoorvapendse: The reason I didn't refactor into a function is because this check depends on first checking `isNormalizedParent` which is done in the previous steps of both the functions, so it cannot be used independently.
also including `isNormalizedParent` call in the new extracted function wouldn't be convenient because of the fact that it throws a different error and the same throwing error or returning false dilemma we have been facing for the qualifiers.

shared-libs/cht-datasource/src/qualifier.ts
apoorvapendse: Same argument as the one I made [here](https://github.com/medic/cht-core/pull/10043#discussion_r2140475176), we throw different errors for different stuff and booleans in case of the `is` functions.

I think we did something similar for the ContactQualifier, where we redundantly copied the checks to throw errors/false returns based on whether it was a `is` or `by` function.

shared-libs/cht-datasource/src/qualifier.ts
sugat009: yes, there a few ways to deal with this.
1. reset type
```typescript
export const isPersonQualifier = (data: unknown): data is PersonQualifier => {
  if (!isContactQualifier(data)) {
    return false;
  }
  
  // Reset data back to unknown for further type checking
  const unknownData = data as unknown;
  
  // Continue with your additional checks on unknownData
  // ...
};
```
2. Double refactor the code. Put the common code from `isContactQualifier` and `isPersonQualifier` into another function.
```typescript
const checkContactQualifierFields = (data: unknown): data is Record<string, unknown> => {
  return isRecord(data) && 
    hasFields(data, [
      {name: 'type', type: 'string', ensureTruthyValue: true}, 
      {name: 'name', type: 'string', ensureTruthyValue: true}
    ]) &&
    (!('reported_date' in data) || isValidReportedDate(data.reported_date));
};

export const isContactQualifier = (qualifier: unknown): qualifier is ContactQualifier => {
  return checkContactQualifierFields(qualifier);
};

export const isPersonQualifier = (data: unknown): data is PersonQualifier => {
  if (!checkContactQualifierFields(data)) {
    return false;
  }
  
  // data is still treated as Record<string, unknown> here, not ContactQualifier
  // Add your PersonQualifier-specific checks
  // ...
  
  return true;
};
```

sugat009: Try out the 2nd option as I think it will get the ball rolling for the other comments.

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -164,6 +164,13 @@ type ContactQualifier = Readonly<{
  * Valid formats are 'YYYY-MM-DDTHH:mm:ssZ', 'YYYY-MM-DDTHH:mm:ss.SSSZ', or <unix epoch>.
  */
 export const byContactQualifier = (data: unknown): ContactQualifier => {
+  // This will throw errors if any.
+  byContactQualifierNonAssertive(data);
+  return data as ContactQualifier;
```
sugat009: question: shouldn't this be returning the function instead like this?

```suggestion
  return byContactQualifierNonAssertive(data) as ContactQualifier;
```

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -267,3 +277,99 @@ const isValidReportedDate = (value: unknown): boolean => {
...
+
+  if (hasBloatedLineage(data)) {
+    return false;
+  }
+
+  if (!hasValidContactType(data) && !hasValidLegacyContactType(data, 'person')) {
+    return false;
+  }
+
+  return true;
```
sugat009:
```suggestion
  return hasValidContactType(data) || hasValidLegacyContactType(data, 'person');
```

equivalent?

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -267,3 +277,99 @@ const isValidReportedDate = (value: unknown): boolean => {
...
+  }
+  return true;
+};
+
+/** @internal */
+const hasValidContactType = ( data: Record<string, unknown> ): boolean => {
+  if (data.type === 'contact' && hasField(data, { name: 'contact_type', type: 'string' })) {
+    return true;
+  }
+  return false;
```
sugat009:
```suggestion
  return data.type === 'contact' && hasField(data, { name: 'contact_type', type: 'string' })
```

shared-libs/cht-datasource/src/qualifier.ts
apoorvapendse: yup

shared-libs/cht-datasource/src/qualifier.ts
apoorvapendse: Yeah makes sense.

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -267,3 +275,92 @@ const isValidReportedDate = (value: unknown): boolean => {
...
+  
+  if (!hasField(qualifier, { name: 'parent', type: 'object' })) {
+    throw new InvalidArgumentError(`Missing or empty required fields [${JSON.stringify(qualifier)}].`);
+  }
+
+  if (!isNormalizedParent(qualifier.parent)) {
+    throw new InvalidArgumentError(`Missing required fields in the parent hierarchy [${JSON.stringify(qualifier)}].`);
+  }
+
+  if (!hasValidContactType(qualifier) && !hasValidLegacyContactType(qualifier, 'person')) {
```
sugat009, line 324:
```suggestion
  if (hasValidContactType(qualifier) || hasValidLegacyContactType(qualifier, 'person') {
```

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -186,14 +191,17 @@ export const byContactQualifier = (data: unknown): ContactQualifier => {
...
+};
+
+/** @internal */
+const checkContactQualifierFields = (data: unknown): data is Record<string, unknown> => {
+  return isRecord(data) && 
+    hasFields(data, [
+      {name: 'type', type: 'string', ensureTruthyValue: true}, 
+      {name: 'name', type: 'string', ensureTruthyValue: true}
+    ]) &&
+    (!('reported_date' in data) || isValidReportedDate(data.reported_date));
```
sugat009, line 211: question: if the `reported_date` does not exist, shouldn't it be added to it? Or are you planning on adding it elsewhere?

sugat009 (changes requested):

shared-libs/cht-datasource/src/qualifier.ts
apoorvapendse: Makes sense, added in the latest commit.

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -222,8 +237,10 @@ export const byReportQualifier = (data: unknown): ReportQualifier => {
   if (!isRecord(data)) {
     throw new InvalidArgumentError('Invalid "data": expected an object.');
   }
+  insertReportedDateIfMissing(data);
+
   const qualifier = {...data};
-  if ('reported_date' in qualifier && !isValidReportedDate(qualifier.reported_date)) {
+  if (!isValidReportedDate(qualifier.reported_date)) {
```
sugat009: issue: the addition of date should be done after line 242, because the whole point of deep copying the `data` object into `qualifier` is to not mutate the passed in `data` object

shared-libs/cht-datasource/test/qualifier.spec.ts
```diff
@@ -267,4 +284,205 @@ describe('qualifier', () => {
...
+              }
+            }
+          }
+        }
+      };
+      expect(() => byPersonQualifier(data)).to
+        .throw(`Additional fields found in the parent lineage [${JSON.stringify(data)}].`);
+    });
+
+    it('builds qualifier for valid objects', () => {
```
sugat009, line 367: issue: can you add a test for when a `reported_date` already exists within the passed `body`

sugat009 (changes requested):
Almost there. Some final touches.

shared-libs/cht-datasource/src/qualifier.ts
apoorvapendse: Ah, right.

shared-libs/cht-datasource/test/qualifier.spec.ts
apoorvapendse: Yes, added in the last commit.

sugat009 (approved):

shared-libs/cht-datasource/src/qualifier.ts
apoorvapendse: (Cross posting from the call with hopefully better clarity)
I think going from the current version to the version @sugat009 mentioned would be easier for the user.
Transitioning from a stricter to more lenient data requirement would introduce complexity and we can make it broader after doing the "MVP" version.
The ideal version would probably be the one where someone can just pass the immediate parent_id or the entire lineage(normalized or hydrated) and we just ignore everything apart from the immediate parent to do our own lineage checks up the hierarchy.

apoorvapendse: The current assumption with `createPerson` and  `PersonQualifier` methods is that the user passed in the correct lineage. We do throw errors if it is bloated or misses the required `_id`, `parent` fields.
I'd like to keep it this way and maybe come back to building the more forgiving and robust version later down the line.
Would love to hear your thoughts @jkuester, @sugat009.

jkuester: @sugat009, what do you think about perhaps modeling our `v1` api for creating a person (and the associated `POST /api/v1/person` REST api) on what we already have for the existing [`POST /api/v1/people`](https://docs.communityhealthtoolkit.org/building/reference/api/#post-apiv1people) api?  (With the goal of being able to deprecate/remove that API in favor of `/api/v1/person`).  If it was a drop-in replacement, it would be easy for consumers to migrate.  

(Same would go for [POST /api/v1/places](https://docs.communityhealthtoolkit.org/building/reference/api/#post-apiv1places))

sugat009: I took a look at the `/api/v1/people` API. I think it has a bad design for the option `Create new person and place hierarchy`. An endpoint named `people` ideally should not be creating `places` and violates Single Responsibility Principle. The other available option of `Create new person and assign existing place.` seems reasonable to me.

jkuester: Ahh, :sweat_smile: thanks for digging deeper into this.  I 100% agree with your objection here. This kind of stuff is why I want to deprecate/remove the `/api/v1/people` api in the first place... But yeah, seems better not to replicate the bad practices into the new api... :facepalm: 

So, circling back to the beginning, what if we just accept a string value for `parent` that must be the _id of the parent contact?  That seems like the most simple, locked-down approach that will involve the least logic. :+1:

sugat009: Yes, thanks Josh. That would be the most simplest API that we could create.
@apoorvapendse Some minor changes will be required throughout the codebase for this. I think it's probably best if it can be done before the integration tests [PR](https://github.com/medic/cht-core/pull/10070) get merged.

apoorvapendse: Sounds good.
I'll append commits to that PR.

PR #10047, opened 2025-06-12
feat(#10019): refactor shared-libs/lineage library to use cht-datasource

### LINKED ISSUE #10019: Refactor code that is loading contacts via shared-libs/lineage to use cht-datasource

shashi-sah2003:
# Description

This pull request refactors the lineage module to replace direct database queries with calls to a new dataContext object, improving modularity and testability. It also updates the associated unit tests to mock the dataContext methods.

fixes #10019 

# License

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
 
   build:
     name: Compile the app
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     timeout-minutes: 60
 
     steps:
@@ -86,6 +86,7 @@ jobs:
   config-tests:
     needs: build
     name: ${{ matrix.cmd }}
+    # these tests fail on newer os. https://github.com/medic/cht-conf-test-harness/issues/279
     runs-on: ubuntu-22.04
     timeout-minutes: 60
     
@@ -109,7 +110,7 @@ jobs:
 
   test-cht-form:
     needs: build
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v4
@@ -129,7 +130,7 @@ jobs:
   tests-k3d:
     needs: build
     name: ${{ matrix.cmd }}
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     timeout-minutes: 60
 
     strategy:
@@ -201,7 +202,7 @@ jobs:
   translations:
     needs: build
     name: Lint translations
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
     steps:
@@ -216,7 +217,7 @@ jobs:
   tests:
     needs: build
     name: ${{ matrix.cmd }}-${{ matrix.suite || '' }}${{ matrix.chrome-version == '90' && '-minimum-browser' || '' }}
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     timeout-minutes: 60
     
     env:
@@ -390,7 +391,7 @@ jobs:
   publish:
     needs: [tests, config-tests, test-cht-form]
     name: Publish branch build
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     timeout-minutes: 60
     
     if: ${{ github.event_name != 'pull_request' }}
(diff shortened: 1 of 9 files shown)


shashi-sah2003:
@sugat009 I have created this new branch with simpler name but still its failing ;). Could it be possible that this is failing for this base branch only because for PR of master branch I don't see this issue.

sugat009:
I'll update the feature branch with the latest master and see how it goes.

sugat009:
@shashi-sah2003 That did not work. Can you create a temp PR from your branch directly to master?

sugat009:
https://github.com/medic/cht-core/pull/10049
I created a dummy PR to master and it does not have the same error. Can you also check if your changes are affecting anything? Try running the command `npm run ci-compile` and it's sub commands one-by-one.

shashi-sah2003:
@sugat009 if that's the case I'll open a new pr pointing to master branch for this issue but before that I'll undo my changes first to see if everything is working correctly or this error is because of my changes

shashi-sah2003:
hey @sugat009 I have found the issue, it is caused by this line `const dataContext = require('../../../api/src/services/data-context.js');`, currently I have found other `data-context.js` files inside `shared-libs` which has been imported like [this](https://github.com/medic/cht-core/blob/master/shared-libs/transitions/src/data-context.js), so I will try to do the same for importing dataContext.

EDIT: I have a thorough analysis of this `shared-libs/lineage` part
My findings:
1. If we want to use `dataContext` in `shared-libs` directory then it couldn't be imported directly like `const dataContext = require('../../../api/src/services/data-context.js');`  rather we have to use these approach.

`data-context.js`

```
module.exports = {
  init: function(dataContext) {
    Object.assign(module.exports, dataContext, { init: this.init });
  },
};
```

`db.js`

```
module.exports = {
  init: function(db) {
    Object.assign(module.exports, db, { init: this.init });
  },
};
```

`config.js`

```
module.exports = {
  init: function(config) {
    Object.assign(module.exports, config, { init: this.init });
  },
};

```

The above three files can be created inside `shared-libs/lineage/src/libs`.

2. Now we have to change the parameter of `lineage` module from something like this `module.exports = (Promise, DB)` to `module.exports = (Promise, DB, sourceConfig, sourceDataContext)`

3. After the above changes as we have changed the parameter of lineage module we need to change this everywhere wherever this lineage module has been imported. 

Note: If lineage module is required in `api/src` directory then we have to pass the original definition of db, config, and dataContext just like done [here](https://github.com/medic/cht-core/blob/master/api/src/services/config-watcher.js), see line 47 and if lineage module is required in `shared-libs` directory then we have to follow the 1st step of passing the different parameters in the respective folders.

for this task we can take example of `@medic/transitions` module, see [here](https://github.com/medic/cht-core/blob/master/shared-libs/transitions/src/index.js) how it has utilized dataContext, db and config inside `shared-libs` directory and also see [here](https://github.com/medic/cht-core/blob/master/api/src/services/config-watcher.js) when `@medic/transitions` has been imported on line 47 inside `api/src` directory the original values have been passed.

sugat009:
Quick question. Why is the data context gonna be initialized in `shared-libs/lineage` and not where it is being called like for `shared-libs/search`?

shashi-sah2003:
We need to incorporate changes in `hydration.js` to use `cht-datasource` which requires `dataContext`, the hydration is required by lineage module so we need to pass the dataContext parameter to lineage which can also be passed to hydration module thus dataContext can be used in that way.

The fact that why we aren't able to import dataContext directly in hydration.js file is doubtful to me as well.  Maybe @jkuester can help us or we can discuss it in tomorrow meet. While examining the current shared-libs directory on the master branch, the dataContext isn't being imported directly.

jkuester:
Ideally, updating `shared-libs/lineage` to use cht-datasource should be one of the last things we do. My hope with the other refactoring is that we can avoid using `shared-libs/lineage` as much as possible (in favor of directly consuming cht-datasource).  Then, at the end we can circle back and see what is left in `shared-libs/lineage` that is being used. At that point we can work out the details of refactoring it to use cht-datasource. :+1:

shashi-sah2003:
@jkuester @sugat009 do you think this issue needs to be closed as now our new goal is to update the `cht-datasource` (`getWithLineage`) module to use `fetchHydratedDoc` as it is hyperoptimized to fetching multiple docs.

jkuester:
Yeah, closing this out as we have reversed directions here (as noted by @shashi-sah2003). :+1:

PR #10056, opened 2025-06-14
feat(#10036): implement `createPerson` for local

apoorvapendse:
# Description
WIP Pull request to add support for creating Person Doc in Pouch.

Related: #10036.

# Code review checklist

# License

### DIFF
--- a/shared-libs/cht-datasource/src/local/person.ts
+++ b/shared-libs/cht-datasource/src/local/person.ts
@@ -104,20 +104,21 @@ export namespace v1 {
   export const createPerson = ({
     medicDb,
     settings
-  } : LocalDataContext) => async (
-    qualifer: PersonQualifier
-  ) :Promise<Person.v1.Person> => {
-    if (hasField(qualifer, { name: '_rev', type: 'string', ensureTruthyValue: true })) {
-      throw new InvalidArgumentError('Cannot pass `_rev` when creating a person.');
-    }
+  } : LocalDataContext) => {
+    const createPersonDoc = createDoc(medicDb);
+    return async (qualifer: PersonQualifier) :Promise<Person.v1.Person> => {
+      if (hasField(qualifer, { name: '_rev', type: 'string', ensureTruthyValue: true })) {
+        throw new InvalidArgumentError('Cannot pass `_rev` when creating a person.');
+      }
     
-    // This check can only be done when we have the contact_types from LocalDataContext.
-    if (!contactTypeUtils.isPerson(settings.getAll(), qualifer)) {
-      throw new InvalidArgumentError('Invalid contact type.');
-    }
+      // This check can only be done when we have the contact_types from LocalDataContext.
+      if (!contactTypeUtils.isPerson(settings.getAll(), qualifer)) {
+        throw new InvalidArgumentError('Invalid contact type.');
+      }
       
-    const createPerson = createDoc(medicDb);
-    const person = await createPerson(qualifer);
-    return person as Person.v1.Person;
+      const person = await createPersonDoc(qualifer);
+      return person as Person.v1.Person;
+    };
   };
 }
+
--- a/shared-libs/cht-datasource/src/local/person.ts
+++ b/shared-libs/cht-datasource/src/local/person.ts
@@ -113,7 +113,7 @@ export namespace v1 {
     
       // This check can only be done when we have the contact_types from LocalDataContext.
       if (!contactTypeUtils.isPerson(settings.getAll(), qualifier)) {
-        throw new InvalidArgumentError('Invalid contact type.');
+        throw new InvalidArgumentError('Invalid person type.');
       }
       
       return await createPersonDoc(qualifier) as Person.v1.Person;
--- a/shared-libs/cht-datasource/src/local/person.ts
+++ b/shared-libs/cht-datasource/src/local/person.ts
@@ -16,8 +16,12 @@ import { validateCursor } from './libs/core';
 /** @internal */
 export namespace v1 {
   /** @internal */
-  // eslint-disable-next-line max-len
-  export const isPerson = (settings: SettingsService) => (doc: Nullable<Doc>, uuid?: string): doc is Person.v1.Person => {
+  export const isPerson = (
+    settings: SettingsService
+  ) => (
+    doc: Nullable<Doc>,
+    uuid?: string
+  ): doc is Person.v1.Person => {
     if (!doc) {
       if (uuid) {
         logger.warn(`No person found for identifier [${uuid}].`);
--- a/shared-libs/cht-datasource/src/local/person.ts
+++ b/shared-libs/cht-datasource/src/local/person.ts
@@ -101,8 +101,12 @@ export namespace v1 {
 
   
   /** @internal */
-  // eslint-disable-next-line max-len
-  export const createPerson = ({ medicDb, settings } : LocalDataContext) => async(qualifer: PersonQualifier) :Promise<Person.v1.Person> => {
+  export const createPerson = ({
+    medicDb,
+    settings
+  } : LocalDataContext) => async (
+    qualifer: PersonQualifier
+  ) :Promise<Person.v1.Person> => {
     if (hasField(qualifer, { name: '_rev', type: 'string', ensureTruthyValue: true })) {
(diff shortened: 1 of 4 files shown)


shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -15,7 +15,9 @@ import { validateCursor } from './libs/core';
 
 /** @internal */
 export namespace v1 {
-  const isPerson = (settings: SettingsService) => (doc: Nullable<Doc>, uuid?: string): doc is Person.v1.Person => {
+  /** @internal */
+  // eslint-disable-next-line max-len
+  export const isPerson = (settings: SettingsService) => (doc: Nullable<Doc>, uuid?: string): doc is Person.v1.Person => {
```
```suggestion
  export const isPerson = (
    settings: SettingsService
  ) => (
    doc: Nullable<Doc>,
    uuid?: string
  ): doc is Person.v1.Person => {
```

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -92,4 +94,22 @@ export namespace v1 {
       )(limit, skip) as Page<Person.v1.Person>;
     };
   };
+
+  
+  /** @internal */
+  // eslint-disable-next-line max-len
+  export const createPerson = ({ medicDb, settings } : LocalDataContext) => async(qualifer: PersonQualifier) :Promise<Person.v1.Person> => {
```
```suggestion
  export const createPerson = ({
    medicDb,
    settings
  } : LocalDataContext) => async (
    qualifer: PersonQualifier
  ) :Promise<Person.v1.Person> => {
```

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -92,4 +94,22 @@ export namespace v1 {
...
+    if (hasField(qualifer, { name: '_rev', type: 'string', ensureTruthyValue: true })) {
+      throw new InvalidArgumentError('Cannot pass `_rev` when creating a person.');
+    }
+    
+    // This check can only be done when we have the contact_types from LocalDataContext.
+    if (!contactTypeUtils.isPerson(settings.getAll(), qualifer)) {
```
sugat009: issue: there is a typo in the word `qualifier` -> `qualifer`

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -92,4 +94,22 @@ export namespace v1 {
...
+      throw new InvalidArgumentError('Invalid contact type.');
+    }
+      
+    const createPerson = createDoc(medicDb);
```
sugat009: issue: this is one of those places where we are following a pattern of currying. This variable assignment needs to be outside the inner curried function but inside the outer function.
Refer to this: https://github.com/medic/cht-core/blob/master/shared-libs/cht-datasource/src/local/person.ts#L34:L43

Here, similar to `getMedicDocById`, `createPerson` needs to be created outside the curried function.

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -92,4 +94,22 @@ export namespace v1 {
...
+    const person = await createPerson(qualifer);
+    return person as Person.v1.Person;
```
sugat009: suggestion: this could be a one liner

```suggestion
    return await createPerson(qualifer) as Person.v1.Person;
```

shared-libs/cht-datasource/test/local/person.spec.ts
```diff
@@ -315,5 +322,109 @@ describe('local person', () => {
...
+              _id: '1-id',
+              parent: {
+                _id: '2-id'
+              }
+            },
+            reported_date: 20202020
+          }
+        ];
+        
+        for (const qualifier of qualifiers) {
```
sugat009: issue: this is generally considered antipattern when writing unit tests. Suppose if the 2nd one fails then, the third one will not run. Code repetition is not a problem in testing. Each input can have their own test case.

sugat009 (changes requested):

shared-libs/cht-datasource/src/local/person.ts
apoorvapendse: OK, so that will save us the overhead of recreating it everytime?

apoorvapendse: When we pass in the qualifier to inner fn that forms the closure with `createPerson`.

sugat009: yes, the connection does not need to be passed in everytime it is used.

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -92,4 +98,26 @@ export namespace v1 {
...
+  } : LocalDataContext) => {
+    const createPersonDoc = createDoc(medicDb);
+    return async (qualifier: PersonQualifier) :Promise<Person.v1.Person> => {
+      if (hasField(qualifier, { name: '_rev', type: 'string', ensureTruthyValue: true })) {
+        throw new InvalidArgumentError('Cannot pass `_rev` when creating a person.');
+      }
+    
+      // This check can only be done when we have the contact_types from LocalDataContext.
+      if (!contactTypeUtils.isPerson(settings.getAll(), qualifier)) {
+        throw new InvalidArgumentError('Invalid contact type.');
```
```suggestion
        throw new InvalidArgumentError('Invalid person type.');
```

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -92,4 +98,26 @@ export namespace v1 {
...
+    settings
+  } : LocalDataContext) => {
+    const createPersonDoc = createDoc(medicDb);
+    return async (qualifier: PersonQualifier) :Promise<Person.v1.Person> => {
+      if (hasField(qualifier, { name: '_rev', type: 'string', ensureTruthyValue: true })) {
+        throw new InvalidArgumentError('Cannot pass `_rev` when creating a person.');
+      }
+    
+      // This check can only be done when we have the contact_types from LocalDataContext.
+      if (!contactTypeUtils.isPerson(settings.getAll(), qualifier)) {
```
sugat009, line 115: issue(non-blocking): i'm getting an error on this line for the `qualifier` object 
```
ESLint: Unsafe argument of type error typed assigned to a parameter of type `Record&lt;string, unknown&gt;`. (@typescript-eslint/no-unsafe-argument)
```
can you check if `hasField` function is changing the type of `qualifier`?

shared-libs/cht-datasource/test/local/person.spec.ts
```diff
@@ -315,5 +322,100 @@ describe('local person', () => {
...
+        dbPost.resolves({ id: qualifier_id, ok: true });
+        dbGet.resolves({ reported_date: qualifier_reported_date, ...qualifier });
+      
+        try {
+          const person = await Person.v1.createPerson(localContext)(qualifier);
+          expect(Person.v1.isPerson(localContext.settings)(person)).to.be.true;
+        } catch (e) {
+          logger.info('Failed creating person for valid qualifier', e);
+          throw e;
+        }
```
sugat009: question: why is this `try/catch` block required?

shared-libs/cht-datasource/test/local/person.spec.ts
```diff
@@ -315,5 +322,100 @@ describe('local person', () => {
...
+          const qualifier_reported_date = new Date().toISOString();
+          dbPost.resolves({ id: '1-id', ok: true });
+          dbGet.resolves({ reported_date: qualifier_reported_date, ...qualifier });
+          try {
+            const person = await Person.v1.createPerson(localContext)(qualifier);
+            expect(Person.v1.isPerson(localContext.settings)(person)).to.be.true;
+          } catch (e) {
+            logger.info('Failed creating person for valid qualifier', e);
+            throw e;
+          }
```
sugat009: same question as above

shared-libs/cht-datasource/test/local/person.spec.ts
apoorvapendse: I was using it for debugging something, thanks for pointing out.

shared-libs/cht-datasource/src/local/person.ts
apoorvapendse: It does assert the value having the property we pass to check, but it should anyway throw an error if `hasField` is true, so I don't think it is an issue.

Issue #10038, opened 2025-06-10
To have API that can create places


sugat009:
@apoorvapendse 
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

apoorvapendse:
commenting for assignment.

PR #10057, opened 2025-06-15
feat(#10038): add `PlaceQualifier` and related methods

apoorvapendse:
# Description
WIP Pull request to add `PlaceQualifier` and related validations.

Related: #10038
# Code review checklist

# License

### DIFF
--- a/shared-libs/cht-datasource/src/qualifier.ts
+++ b/shared-libs/cht-datasource/src/qualifier.ts
@@ -359,6 +359,9 @@ const hasBloatedLineage = ( data: Record<string, unknown> ): boolean => {
       // our liking as `isNormalized` check ensures that it does have two keys `_id` and `parent`.
       return true;
     }
+    if (!parent.parent) {
+      return Object.keys(parent).length > 1;
+    }
     parent = parent.parent;
   }
   return false;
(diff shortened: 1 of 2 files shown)


shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -359,6 +359,9 @@ const hasBloatedLineage = ( data: Record<string, unknown> ): boolean => {
       // our liking as `isNormalized` check ensures that it does have two keys `_id` and `parent`.
       return true;
     }
+    if (!parent.parent) {
+      return Object.keys(parent).length > 1;
+    }
```
apoorvapendse, line 368: This was a case I missed addressing in #10043

apoorvapendse: Basically for the final parent.

sugat009: nice. can you add a comment above this as well for future selves.

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -373,3 +376,73 @@ const hasValidContactType = ( data: Record<string, unknown> ): boolean => {
 const hasValidLegacyContactType = ( data: Record<string, unknown>, type:string): boolean => {
   return data.type === type;
 };
+
+/** 
+ * A qualifier for a place
+ */
+export type PlaceQualifier = ContactQualifier & Readonly<{
+  parent?: NormalizedParent;
+  contact?: NormalizedParent;
```
sugat009, line 389: question: Isn't contact supposed to be either a string containing a UUID string or a object that contains the key `_id` which has the value of a UUID string. Using `Identifiable` interface would be more appropriate here.

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -373,3 +376,73 @@ const hasValidContactType = ( data: Record<string, unknown> ): boolean => {
...
+
+  if (hasField(qualifier, {name: 'contact', type: 'object'})) {
+    if (!isNormalizedParent(qualifier.contact)) {
+      // eslint-disable-next-line max-len
+      throw new InvalidArgumentError(`Missing required fields in the contact hierarchy [${JSON.stringify(qualifier)}].`);
+    }
+    if (hasBloatedLineage(qualifier.contact)){
+      throw new InvalidArgumentError(`Additional fields found in the contact lineage [${JSON.stringify(qualifier)}].`);
+    }
+  }
```
sugat009: suggestion: there's probably a way to refactor this

sugat009 (changes requested):

shared-libs/cht-datasource/src/qualifier.ts
apoorvapendse: Yes, I was a bit unclear about this.
But I ended up seeing and following:
https://github.com/medic/cht-core/blob/5c1f9ba4a149041dd2a16e58420546731cd27b25/shared-libs/cht-datasource/src/place.ts#L19

shared-libs/cht-datasource/src/qualifier.ts
apoorvapendse: done

shared-libs/cht-datasource/src/qualifier.ts
sugat009: I re-read and re-ran the code for this. I have an explanation for this. It's true that the `contact` will have a value of `string` or an object with `_id` with value `string`. However, when requesting a place along with it's lineage, the `contact` object is expected to be a person typed object with `parent` as a field and so on. In which `NormalizedParent` is a nice fit-in. [Example](https://docs.communityhealthtoolkit.org/building/reference/api/#examples-11).
This can be left as it is.

PR #10064, opened 2025-06-18
10052 report lineage

### LINKED ISSUE #10052: Add support for getting report with lineage
Currently we support getting a report doc [via cht-datasource](https://docs.communityhealthtoolkit.org/cht-datasource/interfaces/Report.v1.Report.html) and our [REST API](https://docs.communityhealthtoolkit.org/building/reference/api/#getapiv1reportuuid).  However, this report data only includes the minified (dehydrated) data about the report's `contact` lineage.

Similar to the `PersonWithLineage`/`PlaceWithLineage`/`ContactWithLineage` APIs, we need a `ReportWithLineage` interface.  The normal report interface should be updated to be (with the new `contact` field):

```ts
  export interface Report extends Doc {
    readonly form: string;
    readonly reported_date: Date;
    readonly fields: DataObject;
    readonly contact?: NormalizedParent;
  }
```

This should not require any changes to the actual implementation code, though, since the minified contact data (that fits the `NormalizedParent` interface) should already be included in the returned report data (it just was not called out on the interface).

Then, we need to add a new interface:

```ts
  export interface ReportWithLineage extends Report {
    readonly contact?: Contact.v1.ContactWithLineage | NormalizedParent;
  }
```

In addition to the interface, we need to add a new `getWithLineage` function to the [`Report.v1` namespace](https://github.com/medic/cht-core/blob/5c1f9ba4a149041dd2a16e58420546731cd27b25/shared-libs/cht-datasource/src/report.ts#L16).  

I am thinking the `local/report.ts` implementation for this function would look something like:

```ts
  /** @internal */
  export const getWithLineage = ({ medicDb, settings }: LocalDataContext) => {
    const getLineageDocs = getLineageDocsById(medicDb);
    const getLineage = getContactLineage(medicDb);

    return async (identifier: UuidQualifier): Promise<Nullable<Report.v1.ReportWithLineage>> => {
      const [report, ...contacts] = await getLineageDocs(identifier.uuid);
      if (!isReport(report, identifier.uuid)) {
        return null;
      }

      if (!isNonEmptyArray(contacts)) {
        logger.debug(`No lineage contacts found for report [${identifier.uuid}].`);
        return report;
      }

      const [baseContact, ...lineageContacts] = contacts;
      if (baseContact && contactTypeUtils.isPerson(settings.getAll(), baseContact)) {
        return {
          ...report,
          contact: isNonEmptyArray(lineageContacts)
            ? await getLineage(lineageContacts, baseContact as Person.v1.Person)
            : baseContact
        };
      }

      return {
        ...report,
        contact: await getLineage(contacts)
      };
    };
  };
```

This _may_ just work with the current implementation in `local/libs/lineage.ts`, but I have not actually tested it yet... 😅 

As with the other cht-datasource functions we will want to update the [api report controller](https://github.com/medic/cht-core/blob/5c1f9ba4a149041dd2a16e58420546731cd27b25/api/src/controllers/report.js#L19) to handle returning the report with lineage when the `with_lineage` query param is true. (This will be used by the remote adapter implementation in cht-datasource.)

shashi-sah2003:
This pull request introduces functionality to retrieve reports with lineage information, enhancing the data retrieval capabilities of the system. The changes primarily involve adding new methods and interfaces to support this feature, as well as modifying existing code to integrate the lineage functionality.

fixes #10052 

# License

### DIFF
--- a/api/src/controllers/report.js
+++ b/api/src/controllers/report.js
@@ -6,6 +6,7 @@ const { PermissionError } = require('../errors');
 
 const getReport = () => ctx.bind(Report.v1.get);
 const getReportIds = () => ctx.bind(Report.v1.getUuidsPage);
+const getReportWithLineage = () => ctx.bind(Report.v1.getWithLineage);
 
 const checkUserPermissions = async (req) => {
   const userCtx = await auth.getUserCtx(req);
@@ -19,7 +20,13 @@ module.exports = {
     get: serverUtils.doOrError(async (req, res) => {
       await checkUserPermissions(req);
       const { uuid } = req.params;
-      const report = await getReport()(Qualifier.byUuid(uuid));
+      
+      let report;
+      if(req.query.with_lineage === 'true'){
+        report = await getReportWithLineage()(Qualifier.byUuid(uuid));
+      }else{
+        report = await getReport()(Qualifier.byUuid(uuid));
+      }
 
       if (!report) {
         return serverUtils.error({ status: 404, message: 'Report not found' }, req, res);
--- a/shared-libs/cht-datasource/src/local/report.ts
+++ b/shared-libs/cht-datasource/src/local/report.ts
@@ -6,12 +6,16 @@ import {
   queryDocUuidsByRange
 } from './libs/doc';
 import { FreetextQualifier, UuidQualifier, isKeyedFreetextQualifier } from '../qualifier';
-import { Nullable, Page } from '../libs/core';
+import { Nullable, Page, isNonEmptyArray, NormalizedParent } from '../libs/core';
 import * as Report from '../report';
 import { Doc } from '../libs/doc';
 import logger from '@medic/logger';
 import { normalizeFreetext, validateCursor } from './libs/core';
 import { END_OF_ALPHABET_MARKER } from '../libs/constants';
+import { getContactLineage, getLineageDocsById } from '../local/libs/lineage';
+import contactTypeUtils from '@medic/contact-types-utils';
+import * as Person from '../person';
+
 
 /** @internal */
 export namespace v1 {
@@ -41,6 +45,39 @@ export namespace v1 {
         return null;
       }
       return doc;
+    };
+  };
+
+    /** @internal */
+  export const getWithLineage = ({ medicDb, settings }: LocalDataContext) => {
+    const getLineageDocs = getLineageDocsById(medicDb);
+    const getLineage = getContactLineage(medicDb);
+
+    return async (identifier: UuidQualifier): Promise<Nullable<Report.v1.ReportWithLineage>> => {
+      const [report, ...contacts] = await getLineageDocs(identifier.uuid);
+      if (!isReport(report, identifier.uuid)) {
+        return null;
+      }
+
+      if (!isNonEmptyArray(contacts)) {
+        logger.debug(`No lineage contacts found for report [${identifier.uuid}].`);
+        return report;
+      }
+
+      const [baseContact, ...lineageContacts] = contacts;
+      if (baseContact && contactTypeUtils.isPerson(settings.getAll(), baseContact)) {
+        return {
+          ...report,
+          contact: isNonEmptyArray(lineageContacts)
+                      ? (await getLineage(lineageContacts, baseContact as Person.v1.Person))!
+                      : baseContact as  NormalizedParent
+        };
+      }
+
+      return {
+        ...report,
+        contact: (await getLineage(contacts))!
+      };
     };
   };
 
--- a/shared-libs/cht-datasource/src/report.ts
+++ b/shared-libs/cht-datasource/src/report.ts
@@ -119,4 +119,30 @@ export namespace v1 {
     };
     return curriedGen;
   };
+
+  /**
+   * Returns a function for retrieving a report with lineage from the given data context.
+   * @param context the current data context
+   * @returns a function for retrieving a report with lineage
+   * @throws Error if a data context is not provided
+   */
+
+  export const getWithLineage = (context: DataContext): typeof curriedFnWithLineage => {
+    assertDataContext(context);
+    const fn = adapt(context, Local.Report.v1.getWithLineage, Remote.Report.v1.getWithLineage);
+
+    /**
+     * Returns a report with lineage for the given qualifier.
+     * @param qualifier identifier for the report to retrieve
+     * @returns the report with lineage or `null` if no report is found for the qualifier
+     * @throws Error if the qualifier is invalid
+     */
+    const curriedFnWithLineage = async (
+      qualifier: UuidQualifier
+    ): Promise<Nullable<ReportWithLineage>> => {
+      assertUuidQualifier(qualifier);
+      return await fn(qualifier);
+    };
+    return curriedFnWithLineage;
+  };
 }
--- a/shared-libs/cht-datasource/src/report.ts
+++ b/shared-libs/cht-datasource/src/report.ts
@@ -1,6 +1,7 @@
 import {
   DataObject,
   getPagedGenerator,
+  NormalizedParent,
   Nullable,
   Page
 } from './libs/core';
@@ -11,6 +12,7 @@ import { FreetextQualifier, UuidQualifier } from './qualifier';
 import * as Remote from './remote';
 import { DEFAULT_IDS_PAGE_LIMIT } from './libs/constants';
 import { assertCursor, assertFreetextQualifier, assertLimit, assertUuidQualifier } from './libs/parameter-validators';
+import * as Contact from './contact';
 
 /** */
 export namespace v1 {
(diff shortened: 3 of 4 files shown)


shashi-sah2003:
@jkuester could you share me with the postman collection so that I can create and test reports locally? Also it would be really helpful if you could help me with running and testing it locally like by running only `npm run dev-api` will it be able to test on `localhost` or do I need to have setup with something else as well.

jkuester:
Nice! This is looking good!  

> could you share me with the postman collection so that I can create and test reports locally?

Unfortunately, we do not have a common postman collection. The api server does support basic auth, though, so when you have `dev-api` running you can easily send REST calls to `http://medic:password@localhost:5988/api/v1/report/{{uuid}}` (with Postman or anything else).

> by running only npm run dev-api will it be able to test on localhost or do I need to have setup with something else as well

The short answer is, yes, once you have your local development environment setup and working (e.g. so if you run `npm run dev-api`) you are able to login and use the app), then any changes you make in `shared-libs/cht-datasource` or in `api` should automatically get applied to the running server (the one you started via `npm run dev-api`). :+1:  Please reach out on Discord, though, if you run into trouble here. Happy to jump on a call and work through this together!

---

The final piece we will need in this PR (besides some unit tests for the new code) is to add a few integration tests in:

- https://github.com/medic/cht-core/blob/master/tests/integration/shared-libs/cht-datasource/report.spec.js
- https://github.com/medic/cht-core/blob/master/tests/integration/api/controllers/report.spec.js

To run the integration tests locally you can use `npm run integration-all-local`.  However, since running all the tests takes a long time, I usually update the `all` array in this file: https://github.com/medic/cht-core/blob/c8985c111d2d3ab1e5724fa187dd70467d16d7c7/tests/integration/specs.js#L3

Update the file so that `all` just contains the path to the test file you want to run. Then when you run `npm run integration-all-local` only that test file will run.  (Just don't commit the changes to that `specs.js` file)

shashi-sah2003:
@jkuester pls take a look at the integration test. locally I am not able to test it due to my docker is broken. lmk where I am making the mistake?

jkuester:
:+1: Your tests were very close. I pushed a couple updates just now that should hopefully get everything passing.  _However,_ I got to looking closer at the behavior modeled in the tests and I think we still need to be loading the parent hierarchy for the `contact`. :thinking: I will do some more digging tomorrow on the design side on exactly what we _should_ be returning and let you know.

shashi-sah2003:
@sugat009 marking it as ready to review. pls review it. Made changes regarding to using `fetchHydratedDoc` to be used in `getWithLineage` function.

shared-libs/cht-datasource/src/local/report.ts
```diff
@@ -44,6 +46,26 @@ export namespace v1 {
...
 
+    /** @internal */
+    export const getWithLineage = ({ medicDb }: LocalDataContext) => {
+      const { fetchHydratedDoc } = lineage(Promise, medicDb) as { 
+                                    fetchHydratedDoc: (
+                                      uuid: string, 
+                                      options?: { throwWhenMissingLineage?: boolean }, 
+                                      callback?: (err: Error | null, result?: Doc) => void
+                                    ) => Promise<Doc> 
+                                  };
```
sugat009:
```suggestion
      const { fetchHydratedDoc } = lineage(Promise, medicDb) as {
        fetchHydratedDoc: (
          uuid: string,
          options?: { throwWhenMissingLineage?: boolean },
          callback?: (err: Error | null, result?: Doc) => void
        ) => Promise<Doc>
      };
```

api/src/controllers/report.js
```diff
@@ -19,7 +20,13 @@ module.exports = {
...
       await checkUserPermissions(req);
       const { uuid } = req.params;
-      const report = await getReport()(Qualifier.byUuid(uuid));
+      
+      let report;
+      if (req.query.with_lineage === 'true'){
+        report = await getReportWithLineage()(Qualifier.byUuid(uuid));
+      } else {
+        report = await getReport()(Qualifier.byUuid(uuid));
+      }
```
sugat009: suggestion: traditionally we've had this pattern of binding the `Report.v1.*` function in the constant definition part. [Reference](https://github.com/medic/cht-core/blob/8c5ed645a4252c9bfebe5fcbf0a9693d0f7eb780/api/src/controllers/contact.js#L7)

shared-libs/cht-datasource/src/report.ts
```diff
@@ -21,8 +23,19 @@ export namespace v1 {
...
   }
 
+  /**
+   * A report document with lineage information.
+   */
+  export interface ReportWithLineage extends Report {
+  readonly contact?: Contact.v1.ContactWithLineage | NormalizedParent;
+  readonly patient?: Contact.v1.ContactWithLineage | NormalizedParent;
+  readonly place?: Contact.v1.ContactWithLineage | NormalizedParent;
+}
```
sugat009:
```suggestion
  export interface ReportWithLineage extends Report {
    readonly contact?: Contact.v1.ContactWithLineage | NormalizedParent;
    readonly patient?: Contact.v1.ContactWithLineage | NormalizedParent;
    readonly place?: Contact.v1.ContactWithLineage | NormalizedParent;
  }
```

tests/integration/api/controllers/report.spec.js
```diff
@@ -139,6 +139,55 @@ describe('Report API', () => {
     });
   });
 
+  describe('GET /api/v1/report/:uuid with lineage', async () => {
+    const endpoint = '/api/v1/report';
```
sugat009: issue: endpoint-wise speaking this is the same endpoint as `/api/v1/report` but with the added param of `with_lineage`. All these tests cases can be added in the existing `/api/v1/report` test suite. [Reference](https://github.com/medic/cht-core/blob/master/tests/integration/api/controllers/person.spec.js#L95:L150)

sugat009 (changes requested):

shashi-sah2003:
@sugat009 I have made the necessary changes requested from you. lmk if you want any more changes from my end.

sugat009:
@shashi-sah2003 The app-compilation step is failing for some reason. Can you look into it?

shashi-sah2003:
@sugat009 yeah I found the failing unit test 
```
1) local report
       v1
         getWithLineage
           "before each" hook for "returns a report with contact lineage when found":
     TypeError: sinon_1.default.stub(...).returns is not a function
      at Context.<anonymous> (test/local/report.spec.ts:112:50)
      at processImmediate (node:internal/timers:485:21)
```

I'll try updating it today, commit the changes.

shashi-sah2003:
@jkuester @sugat009 I tried making changes to the failing test case of `getWithLineage`, but there's some issue which I am facing with it regarding the usages of `fetchHydratedDoc` which is inside `lineage` module. I might need to look at any example on how `fetchHydratedDoc` has been used for testing in other test files. If you find anything then do lmk

jkuester:
@shashi-sah2003, I just pushed a commit that should fix the tests (and tweaks a few more things).  Trying to deal with `shared-libs/lineage` in unit tests is painful because of how it just has a function as the default export (so really hard to mock with sinon).  To make the unit tests more convenient (and to centralize the logic for future code), I moved the `shared-libs/lineage` call into a dedicated function in [shared-libs/cht-datasource/src/local/libs/lineage.ts](https://github.com/medic/cht-core/pull/10064/files#diff-db41364ac4b93b74e1554e4ddb5964e213460caad3c5ec56528024a24b81319a).  Unit tests for other parts of cht-datasource can easily mock out that function. :+1: 

@sugat009 could you give another quick review of the changes I made here?  I thing these changes can stand alone and can just be merged directly to `master`, so I updated the base of this PR. But, I since I made a few non-trivial changes in the last commit, it would be good to get another review before merging. :pray:

shashi-sah2003:
@jkuester @sugat009 what exactly needs to be imported to solve this `'PouchDB' is not defined` I tried importing `PouchDB` like this `import PouchDB from 'pouchdb-core';` but it's not reading this PouchDB value?

Also I need your advise on how can I work on this PR #10107 such that I can utilize this latest commit by @jkuester to make changes on test cases on that PR. Will rebasing that PR branch on this PR branch works or cherry-pick will work too, which one would be good?

sugat009 (approved):

sugat009:
> what exactly needs to be imported to solve this 'PouchDB' is not defined I tried importing PouchDB like this import PouchDB from 'pouchdb-core'; but it's not reading this PouchDB value?

@shashi-sah2003 It's a type definition error where the linter is not finding where the `PouchDB` is being referenced from. A separate type declaration to link `PouchDB` might be required.

jkuester:
Sorry, the TS compiler was happy, but forgot to check the linting.  I think the PouchDB problem is fixed now. :crossed_fingers:

Issue #10032, opened 2025-06-10
Convert nginx unit tests to javascript


dianabarsan:
**Describe the issue**
Currently nginx tests use bash and bats. These are limiting because developers are not familiar with this testing framework and they require the supported software (bats) to be committed to the repository.

Describe the improvement you'd like
Convert all nginx tests to javascript. Since these tests are integration tests already - and require spinning up docker containers with the tested software, this is possible.

We should also remove the installation of bats from the CI build pipeline. 

Related: https://github.com/medic/cht-core/issues/9948

sugat009:
@dianabarsan I'll work on this one.

PR #9963, opened 2025-05-13
feat(#9948): couchdb haproxy unit tests

shashi-sah2003:
This draft pull request introduces the CouchDB unit testing setup.

Closes #9948 

# Code review checklist

- [ ] UI/UX backwards compatible: Test it works for the new design (enabled by default). And test it works in the old design, enable `can_view_old_navigation` permission to see the old design. Test it has appropriate design for RTL languages. 
- [ ] Readable: Concise, well named, follows the [style guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/), documented if necessary.
- [ ] Documented: Configuration and user documentation on [cht-docs](https://github.com/medic/cht-docs/)
- [ ] Tested: Unit and/or e2e where appropriate
- [ ] Internationalised: All user facing text
- [ ] Backwards compatible: Works with existing data and configuration or includes a migration. Any breaking changes documented in the release notes.

# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

PR #10067, opened 2025-06-19
chore(#10032): convert nginx unit tests to javascript

sugat009:
# Description

Closes: #10032 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10032-convert-nginx-unit-tests-to-javascript/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10032-convert-nginx-unit-tests-to-javascript/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10032-convert-nginx-unit-tests-to-javascript/docker-compose/cht-couchdb-clustered.yml)

# License


nginx/tests/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "nginx-tests",
+  "private": true,
+  "version": "1.0.0",
+  "scripts": {
+    "test": "npx mocha *.spec.js --no-config"
+  },
+  "dependencies": {
```
dianabarsan: I'm going to echo a comment I made on a similar PR that changed couchdb and haproxy tests: https://github.com/medic/cht-core/pull/9963#discussion_r2091021790 

```
Can you please change this so the tests are run on the main machine, and not in a container? This will mean that the couchdbs will have to expose their ports (i normally use custom ports like: 15984, 25984, 35984).
The main advantage will be that we will have one less container to create and install and that we will not need to worry about upgrading or maintaining these dependencies, as the test will use the ones already installed on the main machine.
```

I believe that the PR I linked to does the ideal changes for the requirements. Can you please use that PR as a model? 
Thanks!

sugat009: Quick confirmation, for the `ssl-install.spec.js` tests, I decided not to mock anything and let the test file run `ssl-install.sh` in the docker container. Now that the tests will be running on the host, this script and it's internals will have to mocked right? Or can it's run be shifted to the existing test container `cht-nginx`?

dianabarsan: I'm not 100% sure I'm understanding the question. 
I think we can stop/start the container using different ENV to trigger various behaviors in ssl-install.sh, and we can check the container logs for success/failure messages.

sugat009: Let me try again. 😅
There are two containers in the `nginx/tests` directory's compose [file](https://github.com/medic/cht-core/blob/bab4032e2bf314116d143e5ef3a56db2aacb2156/nginx/tests/compose.yml#L3) which run during the tests for the `with_mock.bats` tests triggered with the `npm run unit-nginx` command. For the [nginx_error_pages_generate.bats](https://github.com/medic/cht-core/blob/master/nginx/tests/nginx_error_pages_generate.bats) file container is not really necessary.

However, there is another test file `ssl-install.bats` which tests the `ssl-install.sh` file. This test runs in a separate docker [container](https://github.com/medic/cht-core/blob/master/nginx/Dockerfile#L19:L27). This test is not running at the moment. Since `ssl-install.sh` is used, I decided to port `ssl-install.bats` to JS as well. Now, circling back to my question. The tests for `ssl-install.sh` can be written to have the commands inside mocked but I did not find it beneficial so, I ran it in another container because `ssl-install.sh` writes to `/etc/nginx/` folder. My question was: should the new tests in `ssl_install.spec.js` mock the shell commands in `ssl-install.sh` like `shellmock` somewhat does or should it  run `ssl-install.sh` in an already existing container(`cht-nginx`) from the `with_mock.bats` test?

dianabarsan: I think we should have no .bats tests at all. That's the purpose of this PR. 
AFAICT all tests in `ssl-install.bats` can done by starting the `cht-nginx` container with various environment settings, so that it runs `ssl-install.sh` with the correct parameters. 
It's more of an integration  test than unit test, but I don't think it matters. 
So tests in `ssl-install.bats` -> `ssl-install.spec.js` will each start the `cht-nginx` container with different `.env` to handle certificates, and we can assess correctness by tailing `cht-nginx` logs and actually calling the container.

sugat009: Yeah, the tests have already been written. I'm just gonna change it a bit to start the containers each time from the host machine instead of running the tests in the container itself. Are there utility functions in cht-core for working with docker from the host machine?

dianabarsan: There are some utilities that we use in e2e tests (we basically just run docker commands: https://github.com/medic/cht-core/blob/master/tests/utils/index.js#L1215 or https://github.com/medic/cht-core/blob/master/tests/utils/index.js#L1392)

mrjones-plip:
Since the review is well underway and I'll be OOO next week, removing myself from review process.

nginx/tests/integration/ssl_install.spec.js
```diff
@@ -0,0 +1,316 @@
...
+
+  it('should succeed when CERTIFICATE_MODE is OWN_CERT and certificates are provided', async function() {
+    let { stdout, stderr, code } = await runCommand(
+      `docker compose -f ${COMPOSE_FILE_PATH} run -d -e CERTIFICATE_MODE=OWN_CERT --entrypoint "sleep infinity" ${CHT_NGINX_SERVICE_NAME}`
+    );
+    expect(code).to.equal(0);
+
+    // Pre-create dummy cert files inside the container
+    const dummyCert = '---BEGIN CERTIFICATE---\ndummy cert content\n---END CERTIFICATE---';
+    const dummyKey = '---BEGIN PRIVATE KEY---\ndummy key content\n---END PRIVATE KEY---';
```
dianabarsan, line 172: Most secure cert I've ever seen!! <3

nginx/tests/integration/ssl_install.spec.js
```diff
@@ -0,0 +1,316 @@
...
+    await writeFileInContainer(CONTAINER_SSL_CERT_FILE_PATH, dummyCert);
+    await writeFileInContainer(CONTAINER_SSL_KEY_FILE_PATH, dummyKey);
+
+    ({ stdout, stderr, code } = await execInContainer(`/docker-entrypoint.d/ssl-install.sh main`, {
```
dianabarsan, line 176: I'm really enjoying this interface!! Great job setting this up!

nginx/tests/unit/nginx_error_pages_generate.spec.js
```diff
@@ -0,0 +1,135 @@
...
+    it('should be possible to run', async () => {
+      const generateScript = path.join(tmpErrorsDir, 'generate.sh');
+
+      try {
+        await execAsync(generateScript, {
+          cwd: tmpErrorsDir
+        });
+
+        // If we get here, the script ran successfully
+        expect(true).to.be.true;
```
dianabarsan: haha this is sonar complaining that there should be at least one assertion in the test?

nginx/tests/unit/nginx_error_pages_generate.spec.js
```diff
@@ -0,0 +1,135 @@
...
+      } catch (error) {
+        throw new Error(`generate.sh failed to run: ${error.message}`);
```
dianabarsan: Maybe do `expect.fail` instead of a `throw`.

nginx/tests/unit/nginx_error_pages_generate.spec.js
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * Helper function to get all .html and .json files from a directory
+ */
+const getErrorFiles = async (directory) => {
+  try {
+    const files = await fs.readdir(directory);
+    return files.filter(
+      file => file.endsWith('.html') || file.endsWith('.json')
+    ).map(file => path.join(directory, file));
```
dianabarsan: Suggesting an indentation change. 

```suggestion
 return files
  .filter(file => file.endsWith('.html') || file.endsWith('.json'))
  .map(file => path.join(directory, file));
```

nginx/tests/unit/with_mock.spec.js
```diff
@@ -0,0 +1,216 @@
...
+  let original_NODE_TLS_REJECT_UNAUTHORIZED;
+
+  before(function () {
+    original_NODE_TLS_REJECT_UNAUTHORIZED = process.env.NODE_TLS_REJECT_UNAUTHORIZED;
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
+  });
+
+  beforeEach(function() {
+    // Reset any sinon stubs/spies before each test
+    sinon.restore();
```
dianabarsan, line 29: Customary we would do this in the `afterEach`

nginx/tests/unit/with_mock.spec.js
```diff
@@ -0,0 +1,216 @@
+const { expect } = require('chai');
+const sinon = require('sinon');
+
+// Create custom fetch wrapper with default options
+const apiRequest = async (url, options = {}) => {
+  const defaultOptions = {
+    timeout: 5000,
```
dianabarsan: There is no fetch timeout afaict. To use a timeout, you need to use an `AbortSignal`. Please see https://github.com/medic/cht-core/blob/master/shared-libs/couch-request/src/couch-request.js#L135

nginx/tests/unit/with_mock.spec.js
```diff
@@ -0,0 +1,216 @@
...
+          // If we get any response, the API is up
+          expect(response).to.exist;
+          return; // Success!
+
+        } catch (error) {
+          lastError = error;
+          attempt++;
+
+          // Wait 100ms before next attempt
+          await new Promise(resolve => setTimeout(resolve, 100));
```
dianabarsan: There is a node:timers/promises now!!!

https://nodejs.org/docs/latest-v22.x/api/timers.html#timerspromisessettimeoutdelay-value-options

```suggestion
const { setTimeout: setTimeoutPromise } = require('node:timers/promises');
await setTimeoutPromise(100); 
```

nginx/tests/unit/with_mock.spec.js
```diff
@@ -0,0 +1,216 @@
...
+        }
+      }
+
+      throw new Error(`API did not respond within 15 seconds. Last error: ${lastError.message}`);
```
dianabarsan: Does `expect.fail` work better here?

nginx/tests/unit/with_mock.spec.js
```diff
@@ -0,0 +1,216 @@
...
+  });
+
+  describe('Response Header Validation', () => {
+    it('should have proper response headers for HTML content', async () => {
+      const response = await apiRequest(HTTPS_BASE_URL + '/error/drop');
+
+      expect(response.headers.get('content-type')).to.exist;
+
+      const contentType = response.headers.get('content-type');
+      if (contentType.includes('text/html')) {
```
dianabarsan: So ... we just let it slide if the content-type is not html? What could it be?

nginx/tests/Makefile
```diff
@@ -1,10 +1,9 @@
 .PHONY: test
 test:
-	bats nginx_error_pages_generate.bats
-
-	docker compose up --wait --build
-	bats with_mock.bats
+	docker compose up --build --wait --remove-orphans
```
dianabarsan: Can this `makefile` be removed, and have the npm command get the containers up and down?

nginx/tests/unit/nginx_error_pages_generate.spec.js
sugat009: yup

nginx/tests/unit/with_mock.spec.js
sugat009: ohh, this should have been an `expect`.

nginx/tests/package.json
```diff
@@ -3,7 +3,12 @@
...
   "scripts": {
-    "test:unit": "npx mocha unit/*.spec.js --no-config",
-    "test:integration": "npx mocha integration/*.spec.js --no-config"
+    "docker-up": "docker compose up --build --wait --remove-orphans",
+    "docker-down": "docker compose down",
+    "test:generate": "mocha unit/nginx_error_pages_generate.spec.js",
+    "test:api": "npm run docker-up && mocha unit/with_mock.spec.js --no-config && npm run docker-down",
+    "test:unit": "npm run docker-up && mocha unit/*.spec.js && npm run docker-down",
+    "test:integration": "mocha integration/*.spec.js --no-config",
+    "test": "npm run docker-up && npm run test:unit && npm run docker-down && npm run test:integration"
```
dianabarsan: I believe `docker-up` and `docker-down` are already part of `test:unit`, so not needed again.

dianabarsan (approved):
Excellent!

nginx/tests/package.json
sugat009: yes, missed this one

Issue #10041, opened 2025-06-10
To have API that can update reports


sugat009:
@apoorvapendse 
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

manaskumar3003:
can i work on this feature issue ?

sugat009:
@manaskumar3003 Thanks for you interest in this task. However, this is a sub-task of #9835 and has already been assigned. If you are interested in contributing to other tasks, then please go through this first-time-contributors [guide](https://docs.communityhealthtoolkit.org/community/contributing/first-time-contributors/) where you will find some "Good first issues" to work on.

apoorvapendse:
Commenting for assignment.

PR #10083, opened 2025-06-24
feat(#9835): add cht datasource apis for creation and update of contacts and reports

sugat009:
# Description

Adds APIs for creating and updating contacts and reports.

Ticket: #9835 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9835-add-cht-datasource-apis-for-creation-and-update-of-contacts-and-reports/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9835-add-cht-datasource-apis-for-creation-and-update-of-contacts-and-reports/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9835-add-cht-datasource-apis-for-creation-and-update-of-contacts-and-reports/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/tests/page-objects/default/common/common.wdio.page.js
+++ b/tests/page-objects/default/common/common.wdio.page.js
@@ -206,8 +206,11 @@ const clickFastActionById = async (id) => {
  * @returns {Promise<HTMLElement>}
  */
 const findVisibleFAB = async () => {
-  const buttons = [...(await fabSelectors.fastActionFAB())];
-  return Promise.race(buttons.map(button => button.waitForDisplayed().then(() => button)));
+  for (const button of await fabSelectors.fastActionFAB()) {
+    if (await button.isDisplayed()) {
+      return button;
+    }
+  }
 };
 
 const clickFastActionFAB = async ({ actionId, waitForList }) => {


shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -140,3 +140,261 @@ export const and = <
...
+    return isoRegex.test(value);
+  }
+
+  return false;
+};
+
+/** 
+ * A qualifier for a person
+ */
+export type PersonQualifier = ContactQualifier & Readonly<{
```
jkuester: Not a _huge_ deal, but I feel like "Qualifier" is not quite the correct term here.  For the READ apis, we have qualifiers because they filter what data is actually returned. In this case, the type is defining the data that we accept.  

What do you think about calling this someting like `NewPerson` or `PersonInput`?  Also, can this (or really,  `ContactQualifier`) also extend `DataObject`? I think that would help to clarify from a typing perspective that other fields can exist on the object besides just what is specified.

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -140,3 +140,261 @@ export const and = <
...
+ * @param data object containing the fields for a person
+ * @returns the person qualifier
+ * @throws Error if data is not an object
+ * @throws Error if type is not provided or is empty
+ * @throws Error if name is not provided or is empty
+ * @throws Error if parent is not provided or is empty
+ * @throws Error if reported_date is not in a valid format. 
+ * Valid formats are 'YYYY-MM-DDTHH:mm:ssZ', 'YYYY-MM-DDTHH:mm:ss.SSSZ', or <unix epoch>.
+ */
+export const byPersonQualifier = (data: unknown): PersonQualifier => {
```
jkuester: Is there a reason to have this as a separate function outside of `Person.createPerson` (besides that you were just following the pattern of the other qualifiers)?  The main point of the separate qualifier objects for the READ functions was to allow for several different (perhaps non-overlapping) types of qualifiers to be passed to the same READ function. So, having a `Qualifier.byUuid` function made it easy to tell at a glace what was being used to retrieve the patient.  In the case of our CREATE functions, it seems like we really are only going to have one basic creation path that everything has to follow (e.g. I guess that `Person.createPerson` would never accept any other type of parameter besides a `PersonQualifier`, right?) and the validation for that pathway is non-trivial (requires reading from the DB/settings/etc). If that is the case, then I don't see much value in exporting this function. With `PersonQualifier` (or whatever we call that type) as the type of the parameter for `Person.createPerson`, consumers will know the basic shape of the data they have to provide.  (and we can include the validation from this function directly in the `Person.createPerson` implementation....

shared-libs/cht-datasource/src/qualifier.ts
jkuester: It also feels like this could just be moved into the `src/person.ts` file if we are not naming it "Qualifier" any more.

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -140,3 +140,261 @@ export const and = <
...
+ * Builds a qualifier for creation and update of a person with
+ * the given fields.
+ * @param data object containing the fields for a person
+ * @returns the person qualifier
+ * @throws Error if data is not an object
+ * @throws Error if type is not provided or is empty
+ * @throws Error if name is not provided or is empty
+ * @throws Error if parent is not provided or is empty
+ * @throws Error if reported_date is not in a valid format. 
+ * Valid formats are 'YYYY-MM-DDTHH:mm:ssZ', 'YYYY-MM-DDTHH:mm:ss.SSSZ', or <unix epoch>.
```
jkuester: I am really tempted to say we should just start setting the current date to `reported_date` if no other value is provided.  Do you see any reason we should not do that?  It would be nice if, under normal circumstances, consuming code did not even have to mess with that value.

shared-libs/cht-datasource/src/qualifier.ts
sugat009: All the validation logic has been added to this separate function. Also, this function checks the existence of `reported_date` and validates it and adds the current time if it does not exist. The naming and place of the file where this lives could be changed if it is not a Qualifier.

> include the validation from this function directly in the Person.createPerson implementation

sonar probably won't allow that 😅

shared-libs/cht-datasource/src/qualifier.ts
sugat009: we already do that from inside `byContactQualifierNonAssertive` function

shared-libs/cht-datasource/src/qualifier.ts
sugat009: `PersonInput` seems like a good name for this.

apoorvapendse: > Also, can this (or really, ContactQualifier) also extend DataObject

Yes.

shared-libs/cht-datasource/src/qualifier.ts
apoorvapendse: Yeah, we have some additional validations particularly on `contact_type` where we need to have access to the settings, which are done in `createPerson`. Sonar lint is also a reason. 

The general idea is these `by.*` functions do initial validation to check required types and formats and add some missing ones like `reported_date` and validations requiring context are done in the `createPerson`.

jkuester: > sonar probably won't allow that 😅

Sorry, to clarify, by "directly in the Person.createPerson implementation" I just meant that the validation gets run when the user calls `createPerson` (and not `byPersonQualifier`). Of course it will likely be useful to have the validation in a separate reusable function that gets called by `createPerson`.  I am not so worried about the internal structure of the code as I am about how many functions consumers have to call in order to create a person.

jkuester: I like the idea of doing the validation as soon as possible in the process, but I guess I am not sure about the value of the `byPersonQualifier` when there would never be any other kind of qualifier that we pass to `createPerson` (and so the validation will never be different...)

shared-libs/cht-datasource/src/qualifier.ts
jkuester: Ohhh, so you don't _have_ to provide a `reported_date`. But, if you do provide a value and the value is an invalid format we will throw this error?

jkuester: Yeah, okay, I see now that `reported_date` is an optional field on the `ContactQualifier`.  Nice!

shared-libs/cht-datasource/src/qualifier.ts
sugat009: ohh, ok. `byPersonQualifier`(name not changed yet) would be called from inside the `createPerson` and not before to pass into `createPerson`.

sugat009:
@apoorvapendse ESLint was updated recently and it's throwing linting errors because some of the existing code in this PR was aligned with the older ESLint rules. Can you update the code in this PR to align with the new ESLint rules?

apoorvapendse:
[Fixed in #10041](https://github.com/medic/cht-core/pull/10141)

sugat009:
@apoorvapendse I was testing this out manually for verifying the documentation and the update `place` and `report` endpoints are throwing errors when `reported_date` are in string format.

1. Place

image

2. Report

image

Can you check it out?

shared-libs/cht-datasource/src/person.ts
```diff
@@ -116,4 +118,53 @@ export namespace v1 {
...
+  /**
+   * Returns a function for creating a person from the given data context.
+   * @param context the current data context
+   * @returns a function for creating a person.
+   * @throws Error if a data context is not provided
+   */
+  export const create = (context: DataContext): typeof curriedFn => {
+    assertDataContext(context);
+    const fn = adapt(context, Local.Person.v1.create, Remote.Person.v1.create);
+    /**
```
jkuester, line 140:
```suggestion
    /**
     * Creates a new person record.
```

shared-libs/cht-datasource/src/person.ts
```diff
@@ -116,4 +118,53 @@ export namespace v1 {
...
+     * @param input input fields for creating a person
+     * @returns a person doc
```
jkuester:
```suggestion
     * @returns the created person record
```

shared-libs/cht-datasource/src/person.ts
```diff
@@ -116,4 +118,53 @@ export namespace v1 {
...
+     * @throws InvalidArgumentError if the input does not contain required fields
+     * @throws InvalidArgumentError if the input's parent field is not one of the allowed parents in the config
+     * @throws InvalidArgumentError if the required fields do not have the expected type
+     */
+    const curriedFn = async (input: unknown): Promise<Person> => {
```
jkuester: Can we use `PersonInput` as the type of the input parameter here?  We should still run it through the `validatePersonInput` function since the typing is not a guarantee (especially when this is called from JS code), but it would be nice for TS code and for the TSDocs.

```js
    const curriedFn = async (input: PersonInput): Promise<Person> => {
```

shared-libs/cht-datasource/src/person.ts
```diff
@@ -116,4 +118,53 @@ export namespace v1 {
...
+      const personInput = validatePersonInput(input);
```
jkuester: I see in the person api controller you are calling 

```js
Input.validatePersonInput(req.body)
```

before calling `createPerson`.  Then, here in create person we are making a second call to `validatePersonInput`.  I don't think I see the benefit of having `validatePersonInput` as public function outside of cht-datasource.  From the api perspective, it seems like having the type of `input` as `PersonInput` be enough to indicate what should be passed in.  Then we can call the `validatePersonInput` here to ensure the provided data is valid. But, having consumers of `createPerson` also call `validatePersonInput` seems like an extra step...

shared-libs/cht-datasource/src/input.ts
```diff
@@ -0,0 +1,270 @@
...
+    return isoRegex.test(value);
+  }
+
+  return false;
+};
+
+/**
+ * An input for a person
+ */
+export type PersonInput = ContactInput & Readonly<{
```
jkuester: So, this input data is the kind of thing that I think we need to version, right?  I am thinking that if we have a `v2` of the person apis, we might want to make non-passive changes to this interface (e.g requiring additional fields).  If that is the case, I think we should either move this to just be in the `person.ts` file (in the `v1` namespace), or we should add a `v1` namespace here.  

I am leaning towards just putting this in the `person.ts` file since it really is related to "Person", but what do you think?

shared-libs/cht-datasource/src/input.ts
```diff
@@ -0,0 +1,270 @@
...
+ * Builds an input object for creation and update of a person with
+ * the given fields.
+ * @param data object containing the fields for a person
+ * @returns the person input
+ * @throws Error if data is not an object
+ * @throws Error if type is not provided or is empty
+ * @throws Error if name is not provided or is empty
+ * @throws Error if parent is not provided or is empty
+ * @throws Error if reported_date is not in a valid format.
+ * Valid formats are 'YYYY-MM-DDTHH:mm:ssZ', 'YYYY-MM-DDTHH:mm:ss.SSSZ', or <unix epoch>.
```
jkuester: If we end up not exposing this function publicly outside cht-datasource, then we should note these errors on the inner `createPerson` function so they are clearly documented for consumers.

shared-libs/cht-datasource/src/input.ts
jkuester: Minor, for the most part `interface` and `type` can be used interchangeably in TS.  However, based on the reading I have done in the past, `interface` is generally recommended for defining object types (when you don't need the extra power of `type`).  Since our other object types in cht-datasource are already using `interface`s, I would lean in this direction for any new ones...

shared-libs/cht-datasource/src/input.ts
```diff
@@ -0,0 +1,270 @@
...
+ */
+export const validatePersonInput = (data: unknown): PersonInput => {
+  const input = validateContactInputNonAssertive(data);
+
+  if (!hasField(input, { name: 'parent', type: 'string', ensureTruthyValue: true })) {
+    throw new InvalidArgumentError(`Missing or empty required field (parent) [${JSON.stringify(input)}].`);
```
jkuester: Let's not log the entire `input` here. Seems good to avoid PII in the logs when it is not necessary.

shared-libs/cht-datasource/src/input.ts
```diff
@@ -0,0 +1,270 @@
...
+ * @param data object containing the fields for a person
+ * @returns the person input
+ * @throws Error if data is not an object
+ * @throws Error if type is not provided or is empty
+ * @throws Error if name is not provided or is empty
+ * @throws Error if parent is not provided or is empty
+ * @throws Error if reported_date is not in a valid format.
+ * Valid formats are 'YYYY-MM-DDTHH:mm:ssZ', 'YYYY-MM-DDTHH:mm:ss.SSSZ', or <unix epoch>.
+ */
+export const validatePersonInput = (data: unknown): PersonInput => {
```
jkuester: Also, if we end up not using this function to provide input for the `createPerson` function, then I think maybe we can just convert this to be a type assertion (like the `assertDataContext` function)

shared-libs/cht-datasource/src/input.ts
```diff
@@ -0,0 +1,270 @@
...
+ * @throws Error if name is not provided or is empty
+ * @throws Error if reported_date is not in a valid format.
+ * Valid formats are 'YYYY-MM-DDTHH:mm:ssZ', 'YYYY-MM-DDTHH:mm:ss.SSSZ', or <unix epoch>.
+ */
+export const validateContactInput = (data: unknown): ContactInput => {
+  return validateContactInputNonAssertive(data) as ContactInput;
+};
+
+/** @internal*/
+export const validateContactInputNonAssertive = (data: unknown): Record<string, unknown> => {
```
jkuester: I might be missing something, but I don't think I understand why this needs to be a separate function from `validateContactInput`?  Not sure it matters if we return the type as `ContactInput` instead of the more generic `Record<string, unknown>`.

shared-libs/cht-datasource/src/input.ts
```diff
@@ -0,0 +1,270 @@
...
+    return true;
+  }
+  return false;
+};
+
+
+/** @internal */
+const isValidReportedDate = (value: unknown): boolean => {
+  if (typeof value === 'number') {
+    return Number.isFinite(value);
```
jkuester: Typically you would never have an epoch value with decimal places right?  I wonder if we should just be checking `Number.isInteger` here?

shared-libs/cht-datasource/src/input.ts
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  if (typeof value === 'string') {
+    const isoRegex = /^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d{3})?Z)$/;
```
jkuester: Minor, but would be nice to just have this pattern as a file-level constant.  Maybe `ISO_8601_DATE_PATTERN`?

shared-libs/cht-datasource/src/input.ts
```diff
@@ -0,0 +1,270 @@
...
+  }
+  const input = { ...data };
+  insertReportedDateIfMissing(input);
+  if (!isValidReportedDate(input.reported_date)) {
+    throw new InvalidArgumentError(
+      'Invalid reported_date. Expected format to be ' +
+      '\'YYYY-MM-DDTHH:mm:ssZ\', \'YYYY-MM-DDTHH:mm:ss.SSSZ\', or a Unix epoch.'
+    );
+  }
+  input.reported_date = convertToUnixTimestamp(input.reported_date as string | number);
```
jkuester: We don't necessarily have to change anything here, but just wanted to call this out in case you had similar feelings.  The tricky part about this validation is that with the remote adapter, it is going to be run twice (once on the client side when the remote adapter code is called, then again running in the api server when the local adapter is called.)  This is true of all the front-line validation we do here in cht-datasource and it is always a tradeoff between failing early vs avoiding duplicate logic.

I think I am fine leaving this how it is, but I would also be okay either moving all this code down to the local adapter OR splitting it so if a valid is provided for `reported_date`, we validate the value here, but if no value is provided, we wait to set one (and/or convert to unix timestamp) until the local adapter...  I feel like this logic is right on the edge of what I would want to possibly duplicate...

shared-libs/cht-datasource/src/input.ts
```diff
@@ -0,0 +1,270 @@
...
+};
+
+
+/** @internal */
+const isValidReportedDate = (value: unknown): boolean => {
+  if (typeof value === 'number') {
+    return Number.isFinite(value);
+  }
+
+  if (typeof value === 'string') {
```
jkuester: We probably had this conversation before and I just forgot. Do we have a specific use-case that requires us to accept the `reported_date` as a string value at all?  Can we just require it to be a unix epoch value?

shared-libs/cht-datasource/src/person.ts
apoorvapendse: Yes, makes sense.

shared-libs/cht-datasource/src/person.ts
apoorvapendse: #10308 is a step towards this

shared-libs/cht-datasource/src/input.ts
apoorvapendse: Hmm, this is mostly for convenience and to be more lenient with the data we allow the user to send as per my reasoning.
We haven't really had a discussion around this AFAIK, but the [original spec ](https://github.com/medic/cht-core/issues/9835#:~:text=whereas%20reported_date%20should%20be%20of%20format%20%27YYYY%2DMM%2DDDTHH%3Amm%3AssZ%27%2C%20%27YYYY%2DMM%2DDDTHH%3Amm%3Ass.SSSZ%27%2C%20or%20%3Cunix%20epoch%3E.) made sense to me based on that line of reasoning.

apoorvapendse: I'll push an update to make this a unix epoch, if you think that's better.

jkuester: @sugat009 do you have any thoughts/context here?  If we think it is more convenient or consistent to support dates like that then I am fine leaving this.

Otherwise, if we don't have a particular reason to support it, then YAGNI would say to just get rid of it.

sugat009: When I was creating the ticket, I also explored some of the instances for the kind of data that exist in the fields for validation. In some cases, for the `reported_date` field and some datetime related fields, there were dates being stored as strings. So, I thought to have some backward compatibility or API flexibility, it would be nice to have this feature.

jkuester: Got it. In that case, lets just leave this as-is with support for the string date.

shared-libs/cht-datasource/src/input.ts
sugat009: I did some re-reading of the code myself and I think we should do the `v1` namespace for this as well but keep it here, unless we are to get rid of the file itself. This file contains most of the validation code for the inputs and having the interface that is being validated makes sense.

Although I think an additional comment to indicate why this `PersonInput` interface and `Person` both exist could be added. It will be helpful for future readers of the code. @apoorvapendse

shared-libs/cht-datasource/src/input.ts
sugat009: I looked at the usage of the `validateContactInput` and it seems it is not used anywhere and can be removed.

shared-libs/cht-datasource/src/input.ts
sugat009: I think we can remove the validation from the API controller because the validation in both the `src/person.ts` and the controller are identical. If they were validation different stuff, it would make sense to keep both but in this case the validation in `cht-datasource/src/person.ts` would be called anyways with the same payload so, removing the validation in the controller makes sense. Thanks @jkuester for flagging this. And agreed with the inserting `reported_date` not existing case in the local adapter function.

shared-libs/cht-datasource/src/input.ts
jkuester: >  but keep it here, unless we are to get rid of the file itself. 

So, I am more thinking of this in terms of what the external _imports_ will need to look like.  As a dev, would I expect to find this interface under `Input.v1` or just under `Person.v1` with the other person APIs?  Honestly, either place should work and the difference is mostly semantic.  We do already have `Qualifier` were we export things used across the different entities, so `Input` would be similar to that.  

IMHO, since we already have `Person` for stuff that is specific to person entities and `PersonInput` is only for creating these person entities, then putting it in `Person` makes the most sense to me.  (And just get rid of `input.ts`.)  But, if it makes more sense to you all to have a dedicated `Input`, then lets go with that. :+1:

sugat009: I get your point about developer expectations. I think we should keep `input.ts` separate.
The difference from `Qualifier` is that `input.ts` has actual validation logic with inheritance. `PersonInput`, `PlaceInput`, and `ReportInput` all extend `ContactInput` and reuse functions like `validateContactInputNonAssertive`, `insertReportedDateIfMissing`, and `isValidReportedDate`.

If we split this up, we either duplicate code everywhere or end up creating something like `input-validator.ts` anyway.

`Person` is our domain model (immutable, already validated), while `PersonInput` is for external data coming in (mutable, needs validation). They're used at different stages:

`External Request → PersonInput → Validation → Person → Business Logic`

**Clearer imports**
```typescript
// Current approach - somewhat
import { Person, Input } from '@medic/cht-datasource';
const personInput = Input.v1.validatePersonInput(data);
const person = await Person.v1.create(context)(personInput);

// If we change
const personInput = Person.v1.validatePersonInput(data);  // mixing concerns
// or
// const personInput = Person.v1.validateInput(data);
```

When I see `Input.v1`, I know it's about validation. When I see `Person.v1`, I know it's domain operations.

For "namespacing" the `src/input.ts` file, we could consider the following design:

```typescript
// input.ts
export namespace v1 {
  export type PersonInput = ContactInput & Readonly<{...}>;
  export const validatePersonInput = (data: unknown): PersonInput => {...};
  
  export type PlaceInput = ContactInput & Readonly<{...}>;
  export const validatePlaceInput = (data: unknown): PlaceInput => {...};
  
  // etc.
}
```

jkuester: :+1: Sounds good!  I do agree that we should include the `v1` namespace in the `Input` module.

apoorvapendse: > Although I think an additional comment to indicate why this PersonInput interface and Person both exist could be added. It will be helpful for future readers of the code. @apoorvapendse

Hmm, does adding a generic comment at the top of `input.ts` make sense here?
The crux is that we have these `<Entity>Input` interfaces/types because that's the data users have to send for create/update, which I think is applicable to all the interfaces present here, except something like `ContactInput` which isn't something we can create/update as of now.

apoorvapendse: > Person is our domain model (immutable, already validated), while PersonInput is for external data coming in (mutable, needs validation). They're used at different stages:

> External Request → PersonInput → Validation → Person → Business Logic

Thanks for putting it so nicely, I see it in a similar way!

shared-libs/cht-datasource/src/input.ts
apoorvapendse: I don't think we do validation in the controller for person?
@jkuester, I think you were pointing out how we call:

https://github.com/medic/cht-core/blob/df47bbb85be1f3c7facdbfd70e84aff3becb357f/shared-libs/cht-datasource/src/person.ts#L140

Which does redundant validation if a remote call was made from the client.
So should we keep this on the server, i.e. only call `validatePersonInput` if we are running the local version of the code?

jkuester: Yes @apoorvapendse, that is exactly what I am talking about. :+1: Any "expensive" validation should just be done in the cht-datasource local adapter code. As I noted in my original comment, this logic here could probably go either way (it is not super expensive, but there is enough logic going on that I am not sure it is worth running it twice...).  

It seems like we are all on the same page here, so yeah, lets move this validation to the local person adapter code.

apoorvapendse: Cool, thanks for confirming.

apoorvapendse: The third commit in #10308 is an attempt towards this.

shared-libs/cht-datasource/src/input.ts
apoorvapendse: commit 6 of #10308 tries to address this.

shared-libs/cht-datasource/src/input.ts
apoorvapendse: Commits 1, 8, 10 from #10308 make this change.

shared-libs/cht-datasource/src/input.ts
sugat009: Note to self: TODO

sugat009: Note to self: TODO

shared-libs/cht-datasource/src/input.ts
sugat009: Note to self: TODO

shared-libs/cht-datasource/src/input.ts
sugat009: Note to self: TODO

shared-libs/cht-datasource/src/input.ts
sugat009: Note to self: TODO

apoorvapendse: Addressed in commit 13 of #10308.

shared-libs/cht-datasource/src/input.ts
apoorvapendse: Addressed in commit 12 of #10308.

shared-libs/cht-datasource/src/input.ts
apoorvapendse: Hmm, this will result in a lot of 
```
Index signature for type 'string' is missing in <InterfaceName>
```
type errors where the function expects Record<string, unknown> and we pass in something like a PersonInput object.
And we have quite a few functions that expect a Record<string,unknown>

Are we OK with adding the extra 
`[key: string]: unknown;`
index signature in each of these interfaces?

apoorvapendse: IMO, it would look ugly
To make a case for using types here, we already use them in `qualifier.ts`, so I don't think it would be sacrilegious to do so in `input.ts` as well.

Happy to hear your thoughts.

sugat009: In our case, I think it is valid to have a `[key: string]: unknown` when converting to an `interface` because our data is not fixed except for some fields.

apoorvapendse: Sure, I'll refactor this.

apoorvapendse: Done in commit 14 of #10308

apoorvapendse: Commit 15 of #10308 for this.

shared-libs/cht-datasource/src/input.ts
apoorvapendse: Commit 16 (#10308)

sugat009:
I just realized that the update endpoints are named like `PUT /api/v1/person` and not `PUT /api/v1/person/:uuid`. Won't be a massive change. Working on this now.

shared-libs/cht-datasource/src/person.ts
```diff
@@ -116,4 +118,59 @@ export namespace v1 {
...
+     * @returns the created person record
+     * @throws InvalidArgumentError if the input does not contain required fields
+     * @throws InvalidArgumentError if the input's parent field is not one of the allowed parents in the config
+     * @throws InvalidArgumentError if the required fields do not have the expected type
+     * @throws Error if input is not an object
+     * @throws Error if type is not provided or is empty
+     * @throws Error if name is not provided or is empty
+     * @throws Error if parent is not provided or is empty
+     * @throws Error if reported_date is not in a valid format.
+     * Valid formats are 'YYYY-MM-DDTHH:mm:ssZ', 'YYYY-MM-DDTHH:mm:ss.SSSZ', or <unix epoch>.
```
jkuester: TODO lets be completely explicit here.

shared-libs/cht-datasource/src/person.ts
```diff
@@ -116,4 +118,59 @@ export namespace v1 {
...
+    const fn = adapt(context, Local.Person.v1.update, Remote.Person.v1.update);
+    /**
+     * Returns the updated Person Doc for the provided updateInput
+     * @param updateInput the Doc containing updated fields
+     * @returns updated person Doc
+     * @throws InvalidArgumentError if updateInput has changes in immutable fields
+     * @throws InvalidArgumentError if updateInput does not contain required fields
+     * @throws InvalidArgumentError if updateInput fields are not of expected type
+     */
+    const curriedFn = async (updateInput: unknown): Promise<Person> => {
```
jkuester: The input type here can still just be `Person` right?  Since this function just _replaces_ the existing person doc with the new data, then the input needs to already be a valid person.  

```suggestion
    const curriedFn = async (updated: Person): Promise<Person> => {
```

shared-libs/cht-datasource/src/person.ts
```diff
@@ -116,4 +118,59 @@ export namespace v1 {
...
+      if (!isRecord(updateInput)) {
+        throw new InvalidArgumentError('Invalid person update input');
```
jkuester:
```suggestion
        throw new InvalidArgumentError('Updated person data not provided.');
```

shared-libs/cht-datasource/src/input.ts
```diff
@@ -0,0 +1,267 @@
...
+  /**
+   * An input for a contact
+   */
+  interface ContactInput {
+    readonly type: string,
+    readonly name: string,
+    readonly reported_date?: string | number,
+    readonly _id?: string,
+    readonly _rev?: string
+    [key: string]: unknown
```
jkuester: Instead of this, can we just have `ContactInput` extend `DataObject`?

shared-libs/cht-datasource/src/input.ts
```diff
@@ -0,0 +1,267 @@
...
+  };
+
+  /**
+   * Builds an input object for creation and update of a contact with
+   * the given fields.
```
jkuester:
```suggestion
   * Validates the input data for creating a person.
```

shared-libs/cht-datasource/src/input.ts
```diff
@@ -0,0 +1,267 @@
...
+      throw new InvalidArgumentError(
+        `Missing or empty required fields (name, type)`
+      );
+    }
+    return input;
+  };
+
+  /**
+   *
+   */
```
jkuester:
```suggestion
  /** @internal */
```

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -82,4 +89,138 @@ export namespace v1 {
...
+      }
+      input = addParentToInput(input, 'parent', parentDoc);
+      return input;
+    };
+
+    return async (input: Input.v1.PersonInput): Promise<Person.v1.Person> => {
+      input = Input.v1.validatePersonInput(input);
+      if (hasField(input, { name: '_rev', type: 'string', ensureTruthyValue: true })) {
+        throw new InvalidArgumentError('Cannot pass `_rev` when creating a person.');
+      }
```
jkuester: Can we do this in the `Input.v1.validatePersonInput` check?  (Maybe even down in the ContactInput logic...)

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -82,4 +89,138 @@ export namespace v1 {
...
+
+      // Append `contact_type` for newer versions.
+      if (typeFoundInSettingsContactTypes) {
+        input = {
+          ...input,
+          contact_type: input.type,
+          type: 'contact'
+        } as unknown as Input.v1.PersonInput;
+      }
```
jkuester: Minor but this can be simplified a bit. There is a dedicated function for getting person types:

```suggestion
      const customPersonType = contactTypeUtils
        .getPersonTypes(settings.getAll())
        .find(type => type.id === input.type);
      if (!customPersonType && input.type !== 'person') {
        throw new InvalidArgumentError('Invalid person type.');
      }
      if (customPersonType) {
        input = {
          ...input,
          contact_type: input.type,
          type: 'contact'
        };
      }
```

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -82,4 +89,138 @@ export namespace v1 {
...
+
+      input = await appendParent(typeFoundInSettingsContactTypes, input);
+      return await createPersonDoc(input) as Person.v1.Person;
```
jkuester: I think your forking logic in `appendParent` is a bit unnecessarily complicated. 

Something like this should work, right?

```ts
const { minify } = lineageFactory(Promise, medicDb);

const getNormalizedParent = async (
  input: Input.v1.PersonInput,
  customPersonType?: Record<string, unknown>,
): Promise<NormalizedParent> => {
  const supportedParentTypes = customPersonType?.parents as string[] | undefined ?? [
    'district_hospital',
    'health_center',
    'clinic',
  ];
  const parentDoc = await getPersonDoc(input.parent);
  if (!parentDoc) {
    throw new InvalidArgumentError(`Parent with _id ${input.parent} does not exist.`);
  }
  const parentType = contactTypeUtils.getTypeId(parentDoc);
  if (!parentType || !supportedParentTypes.includes(parentType)) {
    throw new InvalidArgumentError(
      `Parent of type ${parentType} is not allowed for ${contactTypeUtils.getTypeId(input)} type`
    );
  }
  minify(parentDoc);
  return parentDoc;
};
```

```ts
minify: (doc: Record<string, unknown>) => void,
```


```ts
      const parent = await getNormalizedParent(input, customPersonType);
      return createPersonDoc({ ...input, parent }) as Promise<Person.v1.Person>;
```

apoorvapendse:
Great!!
Thank you for taking this to completion @jkuester, @sugat009.

I just flipped through [this gist](https://gist.github.com/apoorvapendse/348987ae70bad27e8a805ef23e765fa5) and it brought back a lot of happy memories.
I miss the dev hours and the fun discussions we had during our weekly syncs and on the forum.
I'll try to show up sometime if the dev hour is still a thing.

Thank you once again for the C4GT experience and for integrating this.
@dianabarsan, @andrablaj , @antonykhaemba, @sugat009, @jkuester, @lorerod and all the other wonderful folks I met during that time; I hope you have a great year ahead and all the best for the upcoming C4GT season!

sugat009:
I reviewed and tested this. We can merge it to `master`. The docs [PR](https://github.com/medic/cht-docs/pull/1962) needs to be updated as well with the changes in permissions. Working on it.

PR #10085, opened 2025-06-24
chore(#10074): update api to use contact from cht-datasource

### LINKED ISSUE #10074: Update `api` to read contacts with `cht-datasource`
Use `api/src/services/data-context.js`  to interact with cht-datasource.  
 
- `api/src/migrations/associate-records-with-people.js`
    - Update `getContact` and `getClinic` functions to use cht-datasource to get contact by id.  
- `api/src/migrations/extract-person-contacts.js`  
    -  Update `createPerson` , `resetParent`, and `updateParents` function to use cht-datasource to get contacts by id.  
    - Update `migrateOneType` function to use cht-datasource to get contacts by type instead of querying `medic-client/contacts_by_type`.

@shashi-sah2003 can you comment on this issue and I will get it assigned to you?

shashi-sah2003:
fixes #10074 

This pull request refactors database interaction in two migration scripts to use the `dataContext` service and `cht-datasource` library instead of direct calls to `db.medic`. 

# License


api/src/migrations/associate-records-with-people.js
```diff
@@ -1,9 +1,12 @@
...
 const batch = require('../db-batch');
+const { Contact, Qualifier } = require('@medic/cht-datasource');
+const dataContext = require('../services/data-context');
 
 const getClinic = id => {
-  return db.medic.get(id)
+  return dataContext
+    .bind(Contact.v1.get)(Qualifier.byUuid(id))
     .then(clinic => {
       const contact = clinic.contact;
```
jkuester, line 14: One big difference in behavior between the Pouch `get` function and the cht-datasource `get` functions is that the Pouch `get` will throw an error (with `status: 404`) if a doc with the given id is not found. But, the cht-datasource `get` functions will just return `null`.  So, we need to add a null-check here:

```js
if (!clinic) {
    return;
}
```

Then, we can remove the `catch` block below.

api/src/migrations/associate-records-with-people.js
```diff
@@ -27,7 +30,9 @@ const getContact = (contactId, clinicId) => {
   if (!contactId) {
     return getClinic(clinicId);
   }
-  return db.medic.get(contactId).catch(err => {
+  return dataContext
+    .bind(Contact.v1.get)(Qualifier.byUuid(contactId))
+    .catch(err => {
```
jkuester: Same goes here.  Need to update this `catch` to be a `then` with a null-check before calling `getClinic`.

api/src/migrations/extract-person-contacts.js
```diff
@@ -274,28 +275,26 @@ const updateParents = function(id, callback) {
...
-    db.medic.get(parentId, function(err) {
-      if (err) {
+    dataContext
+      .bind(Contact.v1.get)(Qualifier.byUuid(parentId))
+      .then(() => {
+        return places.updatePlace(facilityId, { parent: parentId });
+      })
+      .then(() => callback())
+      .catch(err => {
         if (err.status === 404) {
```
jkuester: Same here.  This `catch` needs to actually be a null-check.

jkuester (changes requested):
Can you fix the linting error here that are breaking the build?  Also, probably want to run `npm unit-api` to see if any of the unit tests need updated.

jkuester:
@shashi-sah2003 Sorry, I might have gotten ahead of you here with this review :sweat_smile: I see now that this PR is still  in Draft.  Once you are done with the code and ready for me to have another look, just set it to Ready and re-request a review from me. :+1:

shashi-sah2003:
hey @sugat009 I'm marking it as ready for review, I have also made unit test cases change for `associate-records-with-people.spec.js`, I'm facing issue with `extract-person-contacts.spec.js` unit test case. for this I think I need your help, else implementation I think it's good to go.

shashi-sah2003:
@sugat009 you can review this now. All the test cases have been modified

sugat009:
@shashi-sah2003 sonarlint is throwing an error. Can you check it?

shashi-sah2003:
@jkuester @sugat009 the warning from sonar cloud is to not nest functions more than 4 levels deep. Does this warning serious or can be avoided?

shashi-sah2003:
@jkuester I don't understand you have closed this PR? Do I need to rebase this as well making `reportWithLineage` as the base branch?

jkuester:
:thinking: I did not mean to close this PR.  The original base branch was deleted when I merged https://github.com/medic/cht-core/pull/10143, but I thought that should result in the base for PR's like this to automatically become `master`... :shrug: Not sure what GitHub is up to today, but I resurrected the branch and will get this reset here....

jkuester:
> the warning from sonar cloud is to not nest functions more than 4 levels deep. Does this warning serious or can be avoided?

We can always work around Sonar errors if they are blocking. But, in this case I think it is valid to say that the deep nesting of functions here does make this code less readable.  I would recommend trying to extract things to be something like this:

```js
const migrate = (id) => new Promise((resolve, reject) => {
  updateParents(id, function(err) {
    if (err) {
      return reject(err);
    }
    createPerson(id, function(err) {
      if (err) {
        return reject(err);
      }
      resolve();
    });
  });
});

const migrateOneType = function(type, callback) {
  const processIds = async () => {
    
    const generator = dataContext.bind(Contact.v1.getUuids)(Qualifier.byContactType(type));
    for await (const id of generator) {
      await migrate(id);
    }
    
  };

  processIds()
    .then(() => callback())
    .catch(err => callback(err));
};
```

shashi-sah2003:
@sugat009 @jkuester I have fixed the nested function now we're gtg with this PR.

api/tests/mocha/migrations/extract-person-contacts.spec.js
```diff
@@ -5,81 +5,117 @@ const config = require('../../../src/config');
...
+    insertDoc.onCall(1).callsArgWith(1, null);
+    
+    // createPerson: create person doc - success
     createPerson.onCall(0).resolves({ id: 'c'});
-    // Update doc : reset the contact field
-    updatePlace.onCall(1).returns(Promise.reject('error'));
-    // Restore the parent
+    
+    // resetContact: update doc with new contact - FAIL
+    updatePlace.onCall(1).rejects(new Error('error'));
```
jkuester: suggestion: can we revert this back to:

```js
    updatePlace.onCall(1).returns(Promise.reject('error'));
```

Then the error message we assert below should stay the same.

jkuester (changes requested):
This is great! I really appreciate your work with the generators in the unit tests!

I have left one minor request inline, but after that we should be ready to merge!

shashi-sah2003:
@jkuester made the suggested change

Issue #10023, opened 2025-06-05
Update `release-notes` script to include all commit co-authors in contributors list


jkuester:
**Describe the issue**
Our script for automatically generating release notes generates a [list of "Contributors"](https://github.com/medic/cht-core/blob/80d075e0c3b46866a76ac63262a6bdd3abb34ddd/scripts/release-notes/index.js#L316) with the GitHub user details of each user listed as the author of one of the commits to `master`.  

However, many commits to `master` are the result of a lot of collaborative involving many commits to a base branch that is then squash-merged to `master` (see [this commit](https://github.com/medic/cht-core/commit/2cbe9c10991de77e5ed7408432474800ea5185d4) as an example).  For commits like this we want to include all of the "Co-authors" in the list of "Contributors" (and not just the actual "Author" of the commit, which was probably just whoever happened to merge the branch.)

### Background

As a part of the cht-core [release process](https://github.com/medic/cht-core/blob/master/.github/ISSUE_TEMPLATE/z_release_patch.md#releasing---release-engineer) we use the release notes generation script to automatically generate the base release notes content to be [included on our Docs site](https://docs.communityhealthtoolkit.org/releases/).  

**Describe the improvement you'd like**

We need to check the GitHub [graphql API](https://docs.github.com/en/graphql) and see if we can get access to the co-author data for these commits to `master`.  If so, then it should just be a matter of updating the processing in the script's `formatCommits` function to also include the co-authors in the list.

**Describe alternatives you've considered**
We could require the release manager to add this co-author information manually, but ain't nobody got time for that.

shashi-sah2003:
hey @jkuester GitHub graphql API doesn't provide any direct implementation for these but there's an alternative for this which comes from `commit` message. The GitHub `commit` message includes a `Co-authored-by` [message trailer](https://docs.github.com/en/graphql/reference/objects#:~:text=author%20and%20the%20Co%2Dauthored%2Dby%20message%20trailer.%20The%20git) which can be simply extracted using a regex. Something like this, you can also try running this (This is just a dummy code to test).

```
const { Octokit } = require("@octokit/core");
const fs = require("fs");

let token;
try {
  token = require("./token.json").githubApiToken;
} catch (e) {
  console.error("Please create a token.json file with your GitHub token");
  process.exit(1);
}

const octokit = new Octokit({ auth: token });

async function getCommitDetails() {
  try {
[...]

PR #10096, opened 2025-06-27
Fix: Include co-authors in release notes

Saurav-gitus:
# Description

This PR updates the `release-notes` script to correctly include **co-authors** in the list of contributors.

Previously, only the primary author of a commit (typically the person who merged the PR) was listed. This caused contributors to be underrepresented, especially when using squash merges where many co-authors are included via `Co-authored-by:` trailers.

With this update, the script:
- Uses the GitHub GraphQL API to fetch all commit `authors` for each relevant commit
- Ensures each unique contributor is listed once (based on login or name+email)
- Falls back to parsing `Co-authored-by:` if API data is incomplete (optional, if implemented)

This change makes the **contributors list more complete and fair**, especially for collaborative PRs.

Fixes: #10023

# Code review checklist
- [ ] UI/UX backwards compatible: Not applicable
- [x] Readable: Follows style guide and is well organized
- [x] Documented: The purpose and logic are clear in comments
- [x] Tested: Verified against releases `4.1.0` and `4.22.0`
- [ ] Internationalised: Not applicable
- [x] Backwards compatible: The script update is safe and non-breaking

release md

# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

PR #10106, opened 2025-06-28
chore(#10023): include commit co-authors in release notes

Saurav-gitus:
# Description

This PR updates the `release-notes` script to correctly include **co-authors** in the list of contributors.

Previously, only the primary author of a commit (typically the person who merged the PR) was listed. This caused contributors to be underrepresented, especially when using squash merges where many co-authors are included via `Co-authored-by:` trailers.

With this update, the script:
- Uses the GitHub GraphQL API to fetch all commit `authors` for each relevant commit
- Ensures each unique contributor is listed once (based on login or name+email)
- Falls back to parsing `Co-authored-by:` if API data is incomplete (optional, if implemented)

This change makes the **contributors list more complete and fair**, especially for collaborative PRs.

Fixes: #10023

# Code review checklist
- [ ] UI/UX backwards compatible: Not applicable
- [x] Readable: Follows style guide and is well organized
- [x] Documented: The purpose and logic are clear in comments
- [x] Tested: Verified against releases `4.1.0` and `4.22.0`
- [ ] Internationalised: Not applicable
- [x] Backwards compatible: The script update is safe and non-breaking

release md

# License


Saurav-gitus:
Hi @jkuester — I've implemented the fix for https://github.com/medic/cht-core/issues/10023 to include co-authors in the contributors section. Ready for review whenever you get a chance 🙌

mrjones-plip:
Hey @Saurav-gitus - is this a clone of #10096?  If yes, please see @jkuester's [feedback](https://github.com/medic/cht-core/pull/10096#issuecomment-3013780035) which I believe should still apply

Saurav-gitus:
no it's not the clone of the #10096 in the #10096 i was make some changes in the package.json & package-lock.json file but in this i didn't do that again just the implementation which is required

scripts/release-notes/release.md
jkuester, line 1: Can remove this file since it is not needed/used.

scripts/release-notes/index.js
```diff
@@ -1,14 +1,14 @@
-const minimist = require('minimist');
-const { Octokit } = require('@octokit/core');
-const { paginateGraphql } = require('@octokit/plugin-paginate-graphql');
+const minimist = require("minimist");
```
jkuester, line 1: The single-quotes are required by our eslint rules.  We need to roll back these changes to double-quotes.  Then, you can run `npm run lint` to check for any other linting issues in this file before pushing a new version. :+1:

scripts/release-notes/index.js
```diff
@@ -317,10 +372,26 @@ ${formatCommits(commits)}
 `);
 };
 
+const extractCoAuthorsFromMessage = (message) => {
```
jkuester, line 375: As I [mentioned on the issue thread](https://github.com/medic/cht-core/issues/10023#issuecomment-2968207202) last month, I think it would be better to just use the GitHub GraphQL api to get the commit authors instead of parsing the commit message with regex. 

Practically speaking, I think we need to update the query above at the current line `119`.  We should replace `author { user { login, name, url } }` with:

```
        authors(first: 50) {
          edges {
            node { user { login, name, url } }
          }
        }
```

Then we just need to update the down-stream code to use the `authors` array from the response instead of the `author` field.

jkuester (changes requested):

PR #10124, opened 2025-07-03
feat(#10038): store parent/contact with lineage instead of string

apoorvapendse:
WIP PR to store `parent` and `contact` fields with a dehydrated lineage.

We would need to fetch the parent and check its type and ensure it is one of the preconfigured/ allowedparents for the currently created place/person's contact_type
This parent if valid(meaning has a `contact_type` from one of the allowed `parents`  in the contact type object for the given input) will later be used to store `parent` field with de-hydrated lineage in a person/place doc.
Previously, we just stored the `string` containing the parent `id` instead of the de-hydrated lineage.

TODO: Store `contact` with lineage in a place, which shouldn't be as tricky given that there is no validation we'd have to do based on settings.
# Description

# Code review checklist

# License

### DIFF
--- a/shared-libs/cht-datasource/src/local/person.ts
+++ b/shared-libs/cht-datasource/src/local/person.ts
@@ -115,24 +115,25 @@ export namespace v1 {
         throw new InvalidArgumentError(
           `Missing or empty required field (parent) for [${JSON.stringify(input)}].`
         );
-      } else {
-        const parentWithLineage = await getDocById(medicDb)(input.parent);
-        if (parentWithLineage === null){
-          throw new InvalidArgumentError(
-            `Parent does not exist for [${JSON.stringify(input)}].`
-          );
-        }
-        // Check whether parent doc's contact_type matches with any of the allowed parents type.
-        const parentTypeMatchWithAllowedParents = (contactTypeObject.parents as string[])
-          .find(parent => parent===(parentWithLineage as PersonInput).contact_type);
-        
-        if (!(parentTypeMatchWithAllowedParents)) {
-          throw new InvalidArgumentError(
-            `Invalid parent type for [${JSON.stringify(input)}].`
-          );
-        }
-        return parentWithLineage;
+      } 
+
+      const parentWithLineage = await getDocById(medicDb)(input.parent);
+      if (parentWithLineage === null){
+        throw new InvalidArgumentError(
+          `Parent does not exist for [${JSON.stringify(input)}].`
+        );
       }
+      // Check whether parent doc's contact_type matches with any of the allowed parents type.
+      const parentTypeMatchWithAllowedParents = (contactTypeObject.parents as string[])
+        .find(parent => parent===(parentWithLineage as PersonInput).contact_type);
+        
+      if (!(parentTypeMatchWithAllowedParents)) {
+        throw new InvalidArgumentError(
+          `Invalid parent type for [${JSON.stringify(input)}].`
+        );
+      }
+      return parentWithLineage;
+      
     };
 
       const validatePersonParent = 
--- a/shared-libs/cht-datasource/src/local/person.ts
+++ b/shared-libs/cht-datasource/src/local/person.ts
@@ -122,7 +122,7 @@ export namespace v1 {
           );
         }
         if (!((contactTypeObject.parents as string[])
-          .find(parent => parent===(parentWithLineage as PersonInput).type))) {
+          .find(parent => parent===(parentWithLineage as PersonInput).contact_type))) {
           throw new InvalidArgumentError(
             `Invalid parent for [${JSON.stringify(input)}].`
           );
--- a/shared-libs/cht-datasource/src/local/person.ts
+++ b/shared-libs/cht-datasource/src/local/person.ts
@@ -111,7 +111,7 @@ export namespace v1 {
     const ensureHasValidParentFieldAndReturnParentDoc = async(
       input:Record<string, unknown>, 
       contactTypeObject: Record<string, unknown>
-    ): Promise<Doc|null> => {
+    ): Promise<Nullable<Doc>> => {
       const parentDoc = await getDocById(medicDb)(input.parent as string);
       if (parentDoc === null){
         throw new InvalidArgumentError(
--- a/shared-libs/cht-datasource/src/local/person.ts
+++ b/shared-libs/cht-datasource/src/local/person.ts
@@ -115,7 +115,7 @@ export namespace v1 {
       const parentDoc = await getDocById(medicDb)(input.parent as string);
       if (parentDoc === null){
         throw new InvalidArgumentError(
-          `Parent does not exist for [${JSON.stringify(input)}].`
+          `Parent with _id ${input.parent as string} does not exist.` //NoSONAR
         );
       }
       // Check whether parent doc's `contact_type` or `type`(if `contact_type` is absent) 
@@ -160,7 +160,7 @@ export namespace v1 {
     
       if (parentDoc === null){
         throw new InvalidArgumentError(
-          `Parent does not exist for [${JSON.stringify(input)}].`
+          `Parent with _id ${input.parent} does not exist.`
         );
       }
       input = {...input, parent: {
--- a/shared-libs/cht-datasource/src/local/person.ts
+++ b/shared-libs/cht-datasource/src/local/person.ts
@@ -110,19 +110,8 @@ export namespace v1 {
     return async (input: PersonInput) :Promise<Person.v1.Person> => {
 
       const ensureHasValidParentFieldAndReturnParentWithLineage =
-    async(input:Record<string, unknown>, contactTypeObject: Record<string, unknown>):Promise<Doc> => {
-      if (!hasField(input, {name: 'parent', type: 'string', ensureTruthyValue: true})){
-        throw new InvalidArgumentError(
-          `Missing or empty required field (parent) for [${JSON.stringify(input)}].`
-        );
-      } 
-
-      const parentWithLineage = await getDocById(medicDb)(input.parent);
-      if (parentWithLineage === null){
-        throw new InvalidArgumentError(
-          `Parent does not exist for [${JSON.stringify(input)}].`
-        );
-      }
+    async(input:Record<string, unknown>, contactTypeObject: Record<string, unknown>):Promise<Doc|null> => {
+      const parentWithLineage = await getDocById(medicDb)(input.parent as string);
       // Check whether parent doc's contact_type matches with any of the allowed parents type.
       const parentTypeMatchWithAllowedParents = (contactTypeObject.parents as string[])
         .find(parent => parent===(parentWithLineage as PersonInput).contact_type);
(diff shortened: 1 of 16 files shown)


shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -108,6 +108,56 @@ export namespace v1 {
   } : LocalDataContext) => {
     const createPersonDoc = createDoc(medicDb);
     return async (input: PersonInput) :Promise<Person.v1.Person> => {
+
+      const ensureHasValidParentFieldAndReturnParentWithLineage =
```
sugat009: issue: we try not to have nested functions, unless there are absolutely necessary. Here, these functions can be taken out from the `inner async` function. [Reference](https://github.com/medic/cht-core/blob/54dc6c4019452f6b65a9b3f1465c736f1543ea27/shared-libs/cht-datasource/src/local/contact.ts#L78). Does this function and `local/people.ts` function have nothing in common?

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -108,6 +108,56 @@ export namespace v1 {
...
+    async(input:Record<string, unknown>, contactTypeObject: Record<string, unknown>):Promise<Doc|null> => {
+      const parentWithLineage = await getDocById(medicDb)(input.parent as string);
```
sugat009: issue: as with `createPersonDoc` from line 109, this assignment can be taken out from the inner function and be declared in line 110

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -108,6 +108,56 @@ export namespace v1 {
   } : LocalDataContext) => {
     const createPersonDoc = createDoc(medicDb);
     return async (input: PersonInput) :Promise<Person.v1.Person> => {
+
+      const ensureHasValidParentFieldAndReturnParentWithLineage =
+    async(input:Record<string, unknown>, contactTypeObject: Record<string, unknown>):Promise<Doc|null> => {
```
sugat009: issue: can you do some code formatting with reference to some of the existing functions? we do not have a specific formatter and the formatting rules that come with it. you can format to something that looks like this.

```typescript
const ensureHasValidParentFieldAndReturnParentWithLineage = async (
  input: Record<string, unknown>,
  contactTypeObject: Record<string, unknown>
): Promise<Doc | null> => {
  const parentWithLineage = await getDocById(medicDb)(input.parent as string);
  // Check whether parent doc's `contact_type` or `type`(if `contact_type` is absent) 
  // matches with any of the allowed parents type.
  const typeToMatch = (parentWithLineage as PersonInput).contact_type ?? (parentWithLineage as PersonInput).type;
  const parentTypeMatchWithAllowedParents = (contactTypeObject.parents as string[])
    .find(parent => parent === typeToMatch);

  if (!(parentTypeMatchWithAllowedParents)) {
    throw new InvalidArgumentError(`Invalid parent type for [${ JSON.stringify(input) }].`);
  }
  return parentWithLineage;
};
```

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -108,6 +108,56 @@ export namespace v1 {
...
+        return await ensureHasValidParentFieldAndReturnParentWithLineage(input, contactTypeObject);
+      }
+    };
+
+      const appendParentWithLineage = async() => {
+        let parentWithLineage: Doc | null = null;
+        if (typeFoundInSettingsContactTypes){
+          parentWithLineage = await validatePersonParent(typeFoundInSettingsContactTypes, input);
+        } else if (input.parent){
+          parentWithLineage = await getDocById(medicDb)(input.parent);
```
sugat009: issue: as with createPersonDoc from line 109, this assignment can be taken out from the inner function and be declared in line 110

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -108,6 +108,56 @@ export namespace v1 {
...
+        );
+      } else {
+        return await ensureHasValidParentFieldAndReturnParentWithLineage(input, contactTypeObject);
+      }
+    };
+
+      const appendParentWithLineage = async() => {
+        let parentWithLineage: Doc | null = null;
+        if (typeFoundInSettingsContactTypes){
+          parentWithLineage = await validatePersonParent(typeFoundInSettingsContactTypes, input);
```
sugat009: question: this code is just weird to read. `typeFoundInSettingsContactTypes` and `input` is used from the parent scope in `appendParentWithLineage` whereas for the function that exists in the same scope, `validatePersonParent`, params are being passed. What is the reasoning behind this?

shared-libs/cht-datasource/src/local/person.ts
sugat009: this comment applies to all the inner functions declared in this PR

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -108,6 +108,56 @@ export namespace v1 {
...
+        );
+      }
+      return parentWithLineage;
+      
+    };
+
+      const validatePersonParent = 
+    async(contactTypeObject: Record<string, unknown>
+      , input:Record<string, unknown> ):Promise<Doc|null> => {
+      if (!hasField(contactTypeObject, {name: 'parents', type: 'object'})) {
```
sugat009, line 141: question: `contactTypeObject` is a match found from the `input`'s type with the value from the `settings` object. Is this clause checking if the settings object's item is valid?

shared-libs/cht-datasource/src/local/person.ts
sugat009: issue: when `parentWithLineage` is empty/`null`, an error should be thrown, else an unhandled exception will occur

image

sugat009: issue: also this variable should be named `parentDoc` or something similar, because the returning value from `getDocById` will not have the lineage

shared-libs/cht-datasource/src/local/place.ts
```diff
@@ -100,6 +100,111 @@ export namespace v1 {
...
+        } else if (hasField(input, {name: 'parent', type: 'string', ensureTruthyValue: true})){
+          // The current input type is meant to be at the top of the hierarchy.
+          throw new InvalidArgumentError(
+            `Unexpected parent for [${JSON.stringify(input)}].`
+          );
+        }
+        return null;
+      };
+
+      const ensureHasValidParentField =
```
sugat009: question: Can this function and the one in `local/person.ts` be extracted to a common function?

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -108,6 +108,56 @@ export namespace v1 {
...
+        throw new InvalidArgumentError(
+          `Invalid type of person, cannot have parent for [${JSON.stringify(input)}].`
+        );
+      } else {
+        return await ensureHasValidParentFieldAndReturnParentWithLineage(input, contactTypeObject);
+      }
+    };
+
+      const appendParentWithLineage = async() => {
```
sugat009: issue: `lineage` has a different meaning. It's having the object whole object of the parent and it's parent and so on. In this context, it's just parent's IDs and not `lineage`.

For better visualization, checkout `GET /api/v1/person/<uuid>` endpoint. Do 2 requests, one without `with_lineage=true` and one with `with_lineage=true` as query params.

sugat009 (changes requested):

shared-libs/cht-datasource/src/local/person.ts
apoorvapendse: Yeah, it must have `parents`
Meaning `settings.contact_types.[object's contact type].parents` contains the array of allowed parent types.

apoorvapendse: If it doesn't, it means that this type of contact is not allowed to have a parent.

shared-libs/cht-datasource/src/local/person.ts
apoorvapendse: That was a mistake I made, there is no need to pass them at all.

shared-libs/cht-datasource/src/local/person.ts
sugat009: Yes, but you are checking if the matched settings object w.r.t. the `input` type has a `parents` field. Based on which an exception is being thrown, without checking if the `input` also has a `parents` field.

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -107,6 +107,67 @@ export namespace v1 {
     settings
   } : LocalDataContext) => {
     const createPersonDoc = createDoc(medicDb);
+
+    const ensureHasValidParentFieldAndReturnParentDoc = async(
+      input:Record<string, unknown>, 
+      contactTypeObject: Record<string, unknown>
+    ): Promise<Doc|null> => {
+      const parentDoc = await getDocById(medicDb)(input.parent as string);
```
sugat009: issue: `getDocById` should be declared outside this function like `createDoc` and the same function can be used in line 157

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -107,6 +107,67 @@ export namespace v1 {
     settings
   } : LocalDataContext) => {
     const createPersonDoc = createDoc(medicDb);
+
+    const ensureHasValidParentFieldAndReturnParentDoc = async(
+      input:Record<string, unknown>, 
+      contactTypeObject: Record<string, unknown>
+    ): Promise<Doc|null> => {
```
```suggestion
    ): Promise<Nullable<Doc>> => {
```

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -107,6 +107,67 @@ export namespace v1 {
...
+      const parentDoc = await getDocById(medicDb)(input.parent as string);
+      if (parentDoc === null){
+        throw new InvalidArgumentError(
+          `Parent does not exist for [${JSON.stringify(input)}].`
```
sugat009: suggestion: instead of having the whole `input` object in the exception stack trace, it would be useful to just have the `parent` ID that was passed because that was the base upon which the parent is being searched

sugat009: the error message in the same case for when creating a place can be used here. 

image

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -107,6 +107,67 @@ export namespace v1 {
     settings
   } : LocalDataContext) => {
     const createPersonDoc = createDoc(medicDb);
+
+    const ensureHasValidParentFieldAndReturnParentDoc = async(
```
sugat009, line 111: question: can these new functions that checks for `parents` and `contacts` in person, places and reports be refactored into a more common function? They look very similar.

sugat009 (changes requested):

shared-libs/cht-datasource/src/local/person.ts
apoorvapendse: Yeah, makes sense.
with_lineage returns the fully populated docs.

shared-libs/cht-datasource/src/local/person.ts
apoorvapendse: They can be, but I'd like to keep them isolated, b ecause of certain nuances like missing parent field being checked by validatePersonInput before hand for persons, where we do a check during ensureHasValidParent for it in place, since it is optional in place.

PR #10126, opened 2025-07-03
chore(#10078): update user-management use cht-datasource

### LINKED ISSUE #10078: Update `shared-libs/user-management` to read contacts with cht-datasource
Use the existing local data-context utility to interact with cht-datasource [like this](https://github.com/medic/cht-core/blob/2cbe9c10991de77e5ed7408432474800ea5185d4/shared-libs/user-management/src/users.js#L226)  

- `shared-libs/user-management/src/users.js`   
    - Update `validateContact` function to use cht-datasource to get the contact by id.

@shashi-sah2003 can you comment on this issue and I will get it assigned to you?

shashi-sah2003:
# Description

This pull request introduces a refactor to replace direct database calls for fetching contacts with a `getContact` method bound to the `Contact.v1.get` function from the `@medic/cht-datasource` library. Additionally, it updates the corresponding unit tests to mock and validate the behavior of this new method. 
fixes #10078 

# License


shashi-sah2003:
@jkuester @sugat009 could you help me with this failing test cases, all the failing test cases are failing with the same error, I have just replaced `db.medic.get` to use `dataContext` instead and also fixed the failing unit test cases accordingly but the result it is giving after replacement is showing `Contact is not a person`. 

Also when I had added the null check for doc after using dataContext [here](https://github.com/medic/cht-core/pull/10126/commits/87dadfabd3d88d8d9c26fd73cf6d5700c0287b1a) it was throwing this error `Contact not found` which indicates that it wasn't passing null check and doc was returned empty. Have I not replaced it correctly with dataContext?

shared-libs/user-management/test/unit/users.spec.js
```diff
@@ -2333,7 +2340,11 @@ describe('Users service', () => {
         roles: ['national-manager'],
         password: 'password.123'
       };
-      db.medic.get.withArgs('h').resolves({ parent: { _id: 'u', parent: { _id: 'z' } } });
+      getContact.withArgs(Qualifier.byUuid('h')).resolves({
+        _id: 'h',
+        type: 'person',
+        parent: { _id: 'u', parent: { _id: 'x' } }  // ✅ matches place: ['x']
```
jkuester:
```suggestion
        parent: { _id: 'u', parent: { _id: 'x' } }
```

Issue #10037, opened 2025-06-10
To have API that can update people


sugat009:
@apoorvapendse 
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

apoorvapendse:
commenting for assignment.

PR #10141, opened 2025-07-09
feat(#10037): implement update person local

apoorvapendse:
# Description

Implementation of update person in local module

Related: #10037

# Code review checklist

# License

### DIFF
--- a/shared-libs/cht-datasource/src/local/libs/core.ts
+++ b/shared-libs/cht-datasource/src/local/libs/core.ts
@@ -1,4 +1,4 @@
-import { Nullable } from '../../libs/core';
+import { isRecord, Nullable } from '../../libs/core';
 import { Doc } from '../../libs/doc';
 import { InvalidArgumentError } from '../../libs/error';
 
@@ -58,4 +58,56 @@ export const addParentToInput = <T extends HasParentOrContact>(
     ...input,
     [key]: value,
   };
+};
+
+/** @internal*/
+export const dehydrateDoc = (lineage:Record<string, unknown>):Record<string, unknown> => {
+  if (isRecord(lineage.parent)){
+    return {
+      _id: lineage._id,
+      parent: dehydrateDoc(lineage.parent)
+    };
+  }
+  return {
+    _id: lineage._id
+  };
+};
+  
+/** @internal*/
+export const isSameLineage = (
+  a: Record<string, unknown> | null | undefined,
+  b: Record<string, unknown> | null | undefined
+): boolean => {
+  if (!a && !b) {
+    return true;
+  }
+  if (!a || !b) {
+    return false;
+  }
+
+  if (a._id !== b._id) {
+    return false;
+  }
+
+  const aParent = a.parent as Record<string, unknown> | undefined;
+  const bParent = b.parent as Record<string, unknown> | undefined;
+  return isSameLineage(aParent, bParent);
+};
+
+/** @internal*/
+export const ensureHasRequiredImmutableFields=(
+  fields:Set<string>, 
+  originalDoc: Doc, 
+  updateInput:Record<string, unknown>
+):void => {
+  // ensure required immutable fields have the same value as the original doc.
+  for (const field of Array.from(fields)) {
+    if (updateInput[field] !== originalDoc[field]){
+      throw new InvalidArgumentError(
+        `Value ${JSON.stringify(
+          updateInput[field]
+        )} of immutable field '${field}' does not match with the original doc`
+      );
+    }
+  }
 };
--- a/shared-libs/cht-datasource/src/local/libs/core.ts
+++ b/shared-libs/cht-datasource/src/local/libs/core.ts
@@ -16,3 +16,24 @@ export const normalizeFreetext = (
 ): string => {
   return freetext.trim().toLowerCase();
 };
+
+/** @internal*/
+export const getUpdatedFields = (
+  originalDoc:Record<string, unknown>,
+  updatedDoc : Record<string, unknown>,
+  ignoreUpdateFields : Set<string>
+):Record<string, unknown> => {
+  const updatedFields:Record<string, unknown> = {};
+  for (const key of Object.keys(originalDoc)) {
+    if (ignoreUpdateFields.has(key)) {
+      continue;
+    }
+    if (updatedDoc[key] ===undefined) {
+      // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+      delete originalDoc[key];
+      continue;
+    }
+    updatedFields[key] = updatedDoc[key];
+  }
+  return updatedFields;
+};
--- a/shared-libs/cht-datasource/src/local/libs/core.ts
+++ b/shared-libs/cht-datasource/src/local/libs/core.ts
@@ -1,4 +1,4 @@
-import { isRecord, Nullable } from '../../libs/core';
+import { hasField, isRecord, Nullable } from '../../libs/core';
 import { Doc } from '../../libs/doc';
 import { InvalidArgumentError } from '../../libs/error';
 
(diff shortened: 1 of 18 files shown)


apoorvapendse:
Tagging @sugat009 for review.

sugat009:
@apoorvapendse The CI is failing at the compile step. Can you check it out?

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -195,5 +195,48 @@ export namespace v1 {
...
+  }:LocalDataContext) => {
+    const updatePerson = updateDoc(medicDb);
+    const getPerson = get({medicDb, settings} as LocalDataContext);
+    return async(personInput: PersonInput):Promise<Nullable<Doc>> => {
+      if (!isDoc(personInput)){
+        throw new InvalidArgumentError(`Document for update is not a valid Doc ${JSON.stringify(personInput)}`);
+      }
+      const originalDoc = await getPerson({uuid: personInput._id});
+      if (originalDoc===null){
+        throw new InvalidArgumentError(`Invalid person _id, no such person exists.`);
```
```suggestion
        throw new InvalidArgumentError(`Person not found`);
```

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -195,5 +195,48 @@ export namespace v1 {
...
+      }
+      const originalDocDeepCopy = JSON.parse(JSON.stringify(originalDoc)) as unknown as Person.v1.Person;
```
```suggestion
      const originalDocDeepCopy = structuredClone(originalDoc) as unknown as Person.v1.Person;
```

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -195,5 +195,48 @@ export namespace v1 {
...
+      const updatedFields = getUpdatedFields(originalDocDeepCopy, personInput);
+      const updatedDoc = {
+        ...originalDocDeepCopy, ...updatedFields
+      };
+      return await updatePerson(updatedDoc);
+    };
+  };
+
+  /** @internal*/
+  const getUpdatedFields = (originalDoc:Record<string, unknown>, updatedDoc : Record<string, unknown>) => {
```
sugat009: suggestion: this function can be used for `place` and `report`, so it needs to be put in a common place. A change to this function is to have the `ignoreFields` as a parameter.

shared-libs/cht-datasource/test/local/person.spec.ts
```diff
@@ -495,5 +499,88 @@ describe('local person', () => {
         expect(createDocInner.called).to.be.false;
       });
     });
+    describe('updatePerson', () => {
```
sugat009, line 503:
```suggestion

    describe('updatePerson', () => {
```

shared-libs/cht-datasource/src/local/libs/core.ts
```diff
@@ -16,3 +17,53 @@ export const normalizeFreetext = (
...
+    Object.assign(value, { parent: docParent });
+  }
+  return {
+    ...input,
+    [key]: value,
+  };
+};
+
+/** @internal*/
+export const addParentToInput = <T extends HasParentOrContact>(
```
sugat009: suggestion: I think this function can be removed and `buildWithParent` can do everything. Or `buildWithParent` can be renamed to `addParentToInput`.

apoorvapendse: Done.

sugat009 (approved):

sugat009:
@apoorvapendse The changes have been approved but I'll hold on the PR merge until the [discussion](https://forum.communityhealthtoolkit.org/t/exposing-update-apis-as-patch-vs-put/5114) on the `updates` have a final conclusion.

sugat009:
@apoorvapendse While this PR is pending stuff, I ran the unit tests in `cht-datasource` and found out that some test cases are missing and some code blocks are missing the unit tests. Can you check it out?

To get the unit test coverage report in html format. Add the `html` key in `cht-core/shared-libs/nyc.config.js` like this:
```javascript
'use strict';

module.exports = {
  reporter: ['text-summary', 'html']
};
```
and run `npm run test` in the `cht-datasource` folder. The coverage file will be in `cht-datasource/coverage/index.html`.

Preview:
image

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -194,14 +195,38 @@ export namespace v1 {
...
+
+  const validateUpdatePersonPayload = (originalDoc: Person.v1.Person, updatePersonInput: Record<string, unknown>) => {
+    ensureHasRequiredImmutableFields(new Set(['_rev', '_id', 'reported_date']), originalDoc, updatePersonInput);
+    if (!hasField(updatePersonInput, {type: 'object', name: 'parent', ensureTruthyValue: true})){
+      throw new InvalidArgumentError(
+        `Value ${JSON.stringify(
+          updatePersonInput.parent
+        )} of immutable field 'parent' does not match with the original doc`
+      );
+    }
```
sugat009: question: is this just checking if the `parent` field is an object and it is a truthy value? if it is true, then the error message might be a mismatch.

shared-libs/cht-datasource/src/local/libs/core.ts
```diff
@@ -59,3 +59,55 @@ export const addParentToInput = <T extends HasParentOrContact>(
...
+  }
+
+  if (a._id !== b._id) {
+    return false;
+  }
+
+  const aParent = a.parent as Record<string, unknown> | undefined;
+  const bParent = b.parent as Record<string, unknown> | undefined;
+  return isSameLineage(aParent, bParent);
+};
```
sugat009, line 74: suggestion: `lodash`'s utility functions might already have implemented this. Use it if possible.

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -194,14 +195,38 @@ export namespace v1 {
...
+    if (!isSameLineage(
+      updatePersonInput.parent as unknown as Record<string, unknown>,
+      originalDoc.parent
+    )){
+      throw new InvalidArgumentError('Lineage does not match with the lineage of the doc in the db');
+    } 
+    const dehydratedUpdatePersonInput = dehydrateDoc(updatePersonInput);
+    return {...updatePersonInput, parent: dehydratedUpdatePersonInput.parent} as UpdatePersonInput;
```
```suggestion
    return {
      ...updatePersonInput,
      ...dehydratedUpdatePersonInput
    } as UpdatePersonInput;
```

does this work?

sugat009: also, wouldn't the current implementation fail if `dehydratedUpdatePersonInput` does not have a parent?

shared-libs/nyc.config.js
```diff
@@ -1,5 +1,5 @@
 'use strict';
 
 module.exports = {
-  reporter: 'text-summary'
+  reporter: ['text-summary', 'html']
```
sugat009: issue: do not commit this

shared-libs/cht-datasource/src/local/libs/core.ts
apoorvapendse: OK, I'll check.

shared-libs/cht-datasource/src/local/person.ts
apoorvapendse: Every person must have a parent, so it should not fail according to me.

shared-libs/cht-datasource/src/local/person.ts
apoorvapendse: Makes sense, I'll update.

shared-libs/cht-datasource/src/local/libs/core.ts
```diff
@@ -111,3 +114,16 @@ export const ensureHasRequiredImmutableFields=(
...
+  updateInputLineage:Record<string, unknown>, 
+  originalDocLineage: Record<string, unknown>
+) :void => {
+  if (!isSameLineage(
+    updateInputLineage,
+    originalDocLineage
+  )){
+    throw new InvalidArgumentError('Lineage does not match with the lineage of the doc in the db');
+  } 
+};
```
sugat009, line 129: question: does this not work here?

image

sugat009 (changes requested):

shared-libs/cht-datasource/src/local/libs/core.ts
apoorvapendse: One reason why this using `isSameLineage` might be little more efficient is because we don't need to dehydrate first to do this comparison.
Basically we first check the lineage match and only then dehydrate the lineage of the updateDoc. Checking deep equality will require dehydrating the updateDoc first.

PR #10153, opened 2025-07-12
fix(#10062): address race-condition in user edit model for SSO

### LINKED ISSUE #10062: Rendering issue in edit user modal for SSO user
**Describe the bug**
As [described in this forum thread](https://forum.communityhealthtoolkit.org/t/unable-to-edit-place-for-sso-enabled-user-on-cht-v4-20-0/5077) there is an issue with rendering the edit user modal in the admin app that can occur (not all the time, but I am able to reproduce).

Image

**To Reproduce**

1. Enable [SSO login](https://docs.communityhealthtoolkit.org/hosting/sso/)
2. In the admin app, add an "SSO Email Address" value to a user
3. Save the user and reopen the edit user modal for that user
4. Repeat the action of closing the modal and reopening it as necessary until you observe the behavior in the screenshot

**Expected behavior**
The edit user modal should always render properly (even for SSO users) with the "Place:" selector initialized and usable.

**Additional context**
When testing this locally, I was able to work around the issue by just re-opening the modal. 

I suspect there is a race condition caused by [this new logic](https://github.com/medic/cht-core/blob/3912847eff9c8b5e7f51cad74e0e43a307067915/admin/src/js/controllers/edit-user.js#L109) for loading the user's `oidc_login` value. I am not sure exactly how this produces the observed behavior, but is pretty much the only thing that has changed here (and I cannot seem to re-produce the behavior with a non-SSO user).  My current theory is that there was a latent race condition in how the edit user modal is loading and initializing the select2 widgets for place/contact that never manifested itself because the Promise I linked to always resolved fast enough. But, now that we have to make an extra REST call for an SSO user, the promise does not complete fast enough.

benkags:
# Description
Fix broken population of Facility and Associated contact fields in the edit user modal in the admin app.

Closes #10062 

# Code review checklist

# License


jkuester:
@benkags I was testing this out locally and I can confirm that this seems to have addressed the rendering issues for the `Place` field.  However, I am still seeing a related problem with the `Associated contact` field.  Again, it seems to be a race condition here where it can work fine and I see the modal render correctly like this:

image

However, sometimes (particularly if I throttle the network speed for the tab) the `Associated contact` field does not actually have the contact name:

image

I do not see a similar issue on non-SSO contacts when rendering with slow network.

tests/e2e/default/users/common.js
jkuester, line 1: I like the idea of centralizing this logic so it is not duplicated!  However, I don't really see a similar pattern for files like this in the rest of our test code...

@lorerod I was wondering if you had any thoughts or examples of how we have handled things in similar situations?  This is one of those cases where code-reuse is useful, but things are not really so generic that we can just make them general-purpose test utilities...

One thought I did have was to actually just combine the `add-user.wdio-spec.js`, `edit-user.wdio-spec.js` and `common.js` files here all together into a `user-modal.wdio-spec.js` file (the idea being that if they want to share this code, it should all just be in the same file...).

benkags: Happy to rename `add-user.wdio-spec.js` to `user-modal.wdio-spec.js` and combine the logic into one file.

benkags:
Pushed a fix for "Associated Contact". For some reason, it seems rendering any `Select2Search` will trigger some logic for each of the fields have this configure and as a result, if the Associated Contact's `Select2Search` will attempt to look up the place.

tests/e2e/default/users/common.js
benkags: @jkuester do you want me to proceed with refactoring this to one file?

jkuester: @benkags if that approach makes sense to you, then lets proceed to do that.  :+1:   I will wait to merge this PR until we get some feedback from @lorerod next week, but I anticipate that will be her preferred approach too.

jkuester (approved):
Tested this out locally and it works as advertised!  :rocket: Also really appreciate the additional e2e test cases! :sparkles: 

Like I mentioned inline, I will wait for finalization on the e2e file structure before merging this, but otherwise LGTM!

tests/e2e/default/users/common.js
benkags: Alright. I will go ahead and get rid of the `edit-user.wdio-spec.js` file in favor of one file and to avoid a big diff in PR, I think it makes sense to add it this file as a separate test case.

lorerod: Thanks @jkuester, for tagging me on the PR and for waiting until I was back from time off! Yeah, I agree with you that combining the tests into one file sounds like the right move. Renaming it to something like `user-modal.wdio-spec.js` makes sense, and using `describe` blocks to keep the different flows organized should work well. I think the spec will still be easy to manage with everything in one place.

tests/page-objects/default/users/user.wdio.page.js
```diff
@@ -44,6 +45,26 @@ const openAddUserDialog = async () => {
...
 
+const getUsernameRow = async (username) => {
+  await usernameText().waitForDisplayed();
+  const elems = await userList();
+  const found = await elems.filter(async elem => await elem.$$(usernameTextSelector)[0].getText() === username);
+  return found.length === 0? undefined : found[0];
+};
+
+const openEditUserDialog = async (username) => {
+  const el = await getUsernameRow(username);
```
lorerod: Minor suggestion: consider using a more descriptive variable name like `element` instead of `el`. This also aligns with our [style guide](https://docs.communityhealthtoolkit.org/community/contributing/code/style-guide/#use-lowercamelcase-for-variables-properties-and-function-names), which encourages clear and descriptive naming for variables.

tests/e2e/default/users/edit-user.wdio-spec.js
```diff
@@ -0,0 +1,50 @@
...
+      const chwIsSelected = await (await $('input[value="chw"]')).isSelected();
+      expect(chwIsSelected).to.be.true;
+
+      const place = await (await $('[id="facilitySelect"]')).getValue();
+      expect(place).to.equal(districtHospital._id);
+
+      const contact = await (await $('[id="contactSelect"]')).getValue();
+      expect(contact).to.equal(person._id);
+
+      const ssoEmail = await (await $('[id="sso-login"]')).getValue();
```
lorerod: One suggestion for improving maintainability: it would be great to move these hardcoded selectors into the users page object (user.wdio.page.js). We could also add a helper method similar to [getModalDetails](https://github.com/medic/cht-core/blob/cd950f33067dec65312987833b136e1d03a64487/tests/page-objects/default/common/modal.wdio.page.js#L10) (from modal.wdio.page.js), maybe something like `getEditUserDialogDetails()` that returns an object with all the relevant fields (e.g., username, role, place, contact, ssoEmail). Then the test could simply call that method and assert on the returned values, making the code easier to reuse and maintain.

benkags: great suggestion!

tests/page-objects/default/users/user.wdio.page.js
benkags: Updated

tests/e2e/default/users/edit-user.wdio-spec.js
benkags: Updated.

benkags:
Thank you @lorerod, @jkuester. I have pushed updates to address the review comments.

lorerod:
Thanks @benkags for considering my feedback. I reviewed the changes and it looks good to go!
cc: @jkuester

jkuester (approved):
LGTM!

PR #10171, opened 2025-07-22
fix(#10132): Update how API handles form updates on change

### LINKED ISSUE #10132: Update how API handles form updates on change
**Describe the issue**
When CHT-API receives a form update, it reads the full contents of the form document, including _all_ attachments. 
If attachments are large, or only media attachments have received changes, this operation is expensive and pointless. 

Relevant code: https://github.com/medic/cht-core/blob/989c9d67cfa95ff14fa1164400cfecc8b9dcf065/api/src/services/generate-xform.js#L260

Update this code so that only the relevant xml attachments are read, instead of all media attachments. Same, if the xml attachment was changed, only save that attachment instead of the whole form. 

Keep it as it is.

nKataraia:
closes #10132

### DIFF
--- a/api/src/services/generate-xform.js
+++ b/api/src/services/generate-xform.js
@@ -195,6 +195,7 @@ const generateModel = formXml => {
   });
 };
 
+
 const getEnketoForm = doc => {
   const collect = doc.context && doc.context.collect;
   return !collect && formsService.getXFormAttachment(doc);
@@ -226,6 +227,18 @@ const updateAttachmentsIfRequired = (doc, updated) => {
   return formUpdated || modelUpdated;
 };
 
+const addGeneratedAttachments = (doc, updated) => {
+  doc._attachments['form.html'] = {
+    data: Buffer.from(updated.form),
+    content_type: 'text/html'
+  };
+  doc._attachments['model.xml'] = {
+    data: Buffer.from(updated.model),
+    content_type: 'text/xml'
+  };
+  return doc;
+};
+
 const updateAttachments = (accumulator, doc) => {
   return accumulator.then(results => {
     const form = getEnketoForm(doc);
@@ -233,11 +246,16 @@ const updateAttachments = (accumulator, doc) => {
       results.push(null); // not an enketo form - no update required
       return results;
     }
-    logger.debug(`Generating html and xml model for enketo form "${doc._id}"`);
-    return module.exports.generate(form.data.toString()).then(result => {
-      results.push(result);
-      return results;
-    });
+    const name = formsService.getXFormAttachmentName(doc)
+    return db.medic.attachment.get(doc._id, name, { rev, binary: true })
+      .then(rawXML => {
+        logger.debug(`Generating html and xml model for enketo form "${doc._id}"`);
+        return module.exports.generate(rawXML.toString()).then(result => {
+          results.push(result);
+          result && addGeneratedAttachments(doc, result);
+          return results;
+        });
+      });
   });
 };
 
@@ -251,6 +269,7 @@ const updateAllAttachments = docs => {
   });
 };
 
+
 module.exports = {
 
   /**
@@ -258,7 +277,7 @@ module.exports = {
    * @param {string} docId - The db id of the doc defining the form.
    */
   update: docId => {
-    return db.medic.get(docId, { attachments: true, binary: true })
+    return db.medic.get(docId)
       .then(doc => updateAllAttachments([ doc ]))
       .then(docs => {
         const doc = docs.length && docs[0];
@@ -269,7 +288,6 @@ module.exports = {
         logger.info(`Form with ID "${docId}" does not need to be updated.`);
       });
   },
-
   /**
    * Updates the model and form attachments for all forms if necessary.
    */
--- a/api/src/services/forms.js
+++ b/api/src/services/forms.js
@@ -8,13 +8,21 @@ const DOC_TYPE = 'form';
 module.exports = {
 
   /**
+   * @param {Object} doc The doc to find the name of xform attachment on.
+   * @return {String} The name of xform xml attachment.
+   */
+  getXFormAttachmentName: doc => {
+    return Object.keys(doc && doc._attachments || {})
+      .find(name => name === 'xml' ||
+            (name.endsWith('.xml') && name !== 'model.xml'));
+  },
+
+  /**
    * @param {Object} doc The doc to find the xform attachment on.
    * @return {String} The xform xml content.
    */
   getXFormAttachment: doc => {
-    const name = Object.keys(doc && doc._attachments || {})
-      .find(name => name === 'xml' ||
-            (name.endsWith('.xml') && name !== 'model.xml'));
+    const name = module.exports.getXFormAttachmentName(doc);
     return name && doc._attachments[name];
   },
 


jkuester:
@dianabarsan Do you have bandwidth to review this PR (you logged the original issue)?  If not, I can have a look. :+1:

api/src/services/generate-xform.js
```diff
@@ -226,18 +227,35 @@ const updateAttachmentsIfRequired = (doc, updated) => {
...
     if (!form) {
       results.push(null); // not an enketo form - no update required
       return results;
     }
-    logger.debug(`Generating html and xml model for enketo form "${doc._id}"`);
-    return module.exports.generate(form.data.toString()).then(result => {
-      results.push(result);
-      return results;
-    });
+    const name = formsService.getXFormAttachmentName(doc);
```
dianabarsan: Very nice! Unfortunately this function has some very deep promise chains that are pretty hard to follow visually. Can you please convert it to async/await? I think that would make it much more readable.

dianabarsan (changes requested):
Very nice! One request below and please add unit test coverage for this!

api/src/services/generate-xform.js
```diff
@@ -226,30 +228,36 @@ const updateAttachmentsIfRequired = (doc, updated) => {
...
-// Returns array of docs that need saving.
-const updateAllAttachments = docs => {
-  // spawn the child processes in series so we don't smash the server
-  return docs.reduce(updateAttachments, Promise.resolve([])).then(results => {
-    return docs.filter((doc, i) => {
-      return results[i] && updateAttachmentsIfRequired(doc, results[i]);
-    });
-  });
+const updateAttachments = async (doc) => {
+  let generated = null;
```
dianabarsan: You don't need to define this variable here, as it's not reused outside of the inner block.

api/tests/mocha/services/generate-xform.spec.js
```diff
@@ -326,6 +329,8 @@ describe('generate-xform service', () => {
         'form.html': { data: Buffer.from(currentForm) },
         'model.xml': { data: Buffer.from(currentModel) }
       } });
+
```
dianabarsan, line 348: In this file, I think you need to change all `db.medic.get` stubs to not return attachment data, because you changed the call parameters. 

So instead of: 
```
inon.stub(db.medic, 'get').resolves({ _attachments: {
        'xform.xml': { data: Buffer.from(formXml) },
        'form.html': { data: Buffer.from(currentForm) },
        'model.xml': { data: Buffer.from(currentModel) }
      } });
```

you'll have: 
```
sinon.stub(db.medic, 'get').resolves({ _attachments: {
        'xform.xml': { stub: true },
        'form.html': { stub: true },
        'model.xml': { stub: true }
      } });
```

api/tests/mocha/services/generate-xform.spec.js
```diff
@@ -334,8 +339,147 @@ describe('generate-xform service', () => {
...
+      const newForm = '<html><title>Hello</title></html>';
+      const currentModel = '<xml/>';
+      const newModel = '<instance><multimedia/></instance>';
+      
+      sinon.stub(db.medic, 'getAttachment').resolves({ _attachments: {}});
+
+      sinon.stub(db.medic, 'get').resolves({ _attachments: {
+        'xform.xml': { data: Buffer.from(formXml) },
+        'form.html': { data: Buffer.from(currentForm) },
+        'model.xml': { data: Buffer.from(currentModel) }
```
dianabarsan: As above, the `get` will only return attachment stubs, not full contents.

api/src/services/generate-xform.js
```diff
@@ -205,19 +206,20 @@ const generate = formXml => {
     .then(([ form, model ]) => ({ form, model }));
 };
 
-const updateAttachment = (doc, updated, name, type) => {
-  const attachmentData = doc._attachments &&
+const updateAttachment = (doc, update, name, type) => {
+  const attachmentData = doc &&
```
dianabarsan: This function will end up comparing new attachment contents with the existent ones. The problem now is that you're not reading all attachments, because you've changed the `get` command, and you're only reading the main xml attachment. This means that you never have the data of the `model.xml` or the `form.html` attachments, so the app will always think they need to updated, and you end up in an infinite loop. you will need to read the contents of those two attachments too to compare.

api/tests/mocha/services/generate-xform.spec.js
```diff
@@ -334,8 +357,135 @@ describe('generate-xform service', () => {
       });
     });
 
+    it('should not update doc if attachment is not xml/html', async () => {
+      const newForm = '<html><title>Hello</title></html>';
+      const newModel = '<instance><multimedia/></instance>';
+      
+      sinon.stub(db.medic, 'getAttachment').resolves({ _attachments: {}});
```
dianabarsan: How would `getAttachment` return a an empty list of attachments?

api/tests/mocha/services/generate-xform.spec.js
```diff
@@ -334,8 +357,135 @@ describe('generate-xform service', () => {
...
+      sinon.stub(db.medic, 'get').resolves({ _attachments: {
+        'xform.xml': {stub: true},
+        'form.html': { stub: true },
+        'model.xml': { stub: true }
+      } });
+      sinon.stub(service, 'generate').resolves({ form: newForm, model: newModel });
+      sinon.stub(db.medic, 'put');
+
+      await service.update('form:exists');
+      expect(db.medic.put.callCount).to.equal(1);
```
dianabarsan: The test title says that the form should not be updated if the attachment is not xml, but then this assertion contradicts the test name, with the form actually getting updated. 
I'm confused.

api/tests/mocha/services/generate-xform.spec.js
```diff
@@ -27,6 +27,22 @@ const expectAttachments = (doc, form, model) => {
   expect(modelAttachment.content_type).to.equal('text/xml');
 };
 
+const fakeGetAttachment = (id, name, opt) => {
```
dianabarsan, line 30: I love this stub! Nice work here!

api/src/services/generate-xform.js
```diff
@@ -205,51 +206,39 @@ const generate = formXml => {
...
-  const formUpdated = updateAttachment(doc, updated.form, 'form.html', 'text/html');
-  const modelUpdated = updateAttachment(doc, updated.model, 'model.xml', 'text/xml');
-  return formUpdated || modelUpdated;
-};
+const updateAttachments = async (doc) => {
+  const enketoForm = getEnketoForm(doc);
+  if (enketoForm) {
+    const name = formsService.getXFormAttachmentName(doc);
+    const rawXML = await db.medic.getAttachment(doc._id, name, { rev: doc._rev });
+    const form = await db.medic.getAttachment(doc._id, 'form.html', { rev: doc._rev });
```
dianabarsan: It's possible that any of these two calls fail with a 404 if the form has just been uploaded. This is causing a lot of e2e test failures now. 
You will need to wrap this in a try-catch and always create these attachments if they don't yet exist.

dianabarsan (changes requested):
Nice! Great improvements! 
I did find one issue which I've detailed in the comments.

api/src/services/generate-xform.js
```diff
@@ -224,23 +223,31 @@ const addGeneratedAttachments = (doc, updated, outdated) => {
   return doc;
 };
 
+const getAttachment = async (doc, name) => {
+  try { 
+    return await db.medic.getAttachment(doc._id, name, { rev: doc._rev }); 
+  } catch (error) {
```
dianabarsan: I think we shouldn't catch all errors, just 404s, and throw all others.

api/src/services/generate-xform.js
```diff
@@ -258,18 +261,16 @@ module.exports = {
...
-    return db.medic.get(docId, { attachments: true, binary: true })
+    return db.medic.get(docId)
       .then(doc => updateAllAttachments([ doc ]))
       .then(docs => {
         const doc = docs.length && docs[0];
         if (doc) {
           logger.info(`Updating form with ID "${docId}"`);
           return db.medic.put(doc);
         }
-        logger.info(`Form with ID "${docId}" does not need to be updated.`);
```
dianabarsan, line 269: In E2E tests we are expecting this line to be logged, to check that the forms were actually generated. Please add it back.

api/tests/mocha/services/generate-xform.spec.js
```diff
@@ -286,11 +302,16 @@ describe('generate-xform service', () => {
...
-      } });
+      sinon.stub(db.medic, 'get').resolves({
+        _attachments: {
+          'xform.xml': { stub: true },
+          'form.html': { stub: true },
+          'model.xml': { stub: true },
+        },
+        _rev: '1-rev'
+      });
+      // sinon.stub(db.medic, 'getAttachment').resolves(Buffer.from(formXml));
```
dianabarsan: Please remove this commented code.

api/tests/mocha/services/generate-xform.spec.js
```diff
@@ -334,8 +357,137 @@ describe('generate-xform service', () => {
...
+        'model.xml': { stub: true }
+      } });
+      sinon.stub(service, 'generate').resolves({ form: newForm, model: newModel });
+      sinon.stub(db.medic, 'put');
+
+      await service.update('form:exists');
+      expect(db.medic.getAttachment.callCount).to.equal(3);
+      expect(service.generate.callCount).to.equal(0);
+      expect(db.medic.put.callCount).to.equal(0);
+      // expectAttachments(db.medic.put.args[0][0], newForm, newModel);
```
dianabarsan: Please remove this commented code.

dianabarsan (approved):
Thanks for all the changes! Awesome work! 
I believe this is done and we can proceed to acceptance test this along with the changes to [cht-conf](https://github.com/medic/cht-conf/pull/724) to support uploading large attachments. 
The focus of this testing is: 
- forms with large attachments get uploaded
- the model.xml and form.html *do not* get updated when subsequent attachments are uploaded

api/src/services/generate-xform.js
```diff
@@ -205,51 +206,56 @@ const generate = formXml => {
...
-const updateAttachmentsIfRequired = (doc, updated) => {
-  const formUpdated = updateAttachment(doc, updated.form, 'form.html', 'text/html');
-  const modelUpdated = updateAttachment(doc, updated.model, 'model.xml', 'text/xml');
-  return formUpdated || modelUpdated;
+const getAttachment = async (doc, name) => {
+  try {
+    return await db.medic.getAttachment(doc._id, name, { rev: doc._rev });
+  } catch (error) {
+    if (error.status === 404) {
+      logger.error(error);
```
dianabarsan: I'm seeing this error being logged in a really ambiguous way unfortunately: 
```
2025-10-13T13:15:46.955 INFO: Detected form change for form:pregnancy
2025-10-13T13:15:46.955 INFO: Detected form change - generating attachments
2025-10-13T13:15:46.963 ERROR: [object Response]
2025-10-13T13:15:46.964 ERROR: [object Response]
```

I think we should either not log this error or change it to something like: 
```suggestion
logger.error(`Attachment ${name} not found for doc ${docId}`);
```

dianabarsan:
closing in favor of https://github.com/medic/cht-core/pull/10248

Issue #9707, opened 2024-12-12
Support Nouveau in CHT Deploy script and upgrades docs


mrjones-plip:
**Describe the issue**
Current, the [CHT Deploy](https://github.com/medic/cht-core/tree/master/scripts/deploy) script for k8s is not aware of the upcoming [CHT Core Nouveau feature](https://github.com/medic/cht-core/issues/9542).  

**Describe the improvement you'd like**
We should update  the deploy script such that a happy path new deployment works when we ship CHT 4.x with Nouveau.  Additionally, upgrade paths from prior versions of k8s based deployments should account for this change to ensure services don't break.  This may be just documentation improvements or it maybe code additions or helm chart additions.

mrjones-plip:
@jkuester or @sugat009 - do we know how a deployment looks when it has both a clustered couch deployment AND Nouveau?  That is - if there's 3 couch containers running, is there 1 Nouveau container or 3 Nouveau containers?

cc @dianabarsan

jkuester:
😓 Mokhtar might know the answer to that question, but unfortunately I do not have any hard answers yet.  Definitely something we need to confirm.

mrjones-plip:
Ah - I don't think CI has ever passed, so the only way to POC is in [dev mode](https://docs.communityhealthtoolkit.org/contribute/code/core/dev-environment/).  I'll see what I can do!  I'll also see if CouchDB docs have any info.

mrjones-plip:
@jkuester or @sugat009 - are there steps on how to run Nouveau against a developer setup?

Otherwise, I can't find an answer through a POC (b/c I can't bootstrap the system in the first place) and don't see [Nouveau setup instructions](https://docs.couchdb.org/en/stable/install/nouveau.html) involving a cluster anywhere - I'm hopping in there Slack server to see what I can find!

> Hey all! I'm continuing [Mokhtar's work](https://couchdb.slack.com/archives/C49LEE7NW/p1730901561531279) at Medic with Nouveau .  We currently offer a performance boost by deploying our app against a clustered couch install with 3 couch nodes behind HAProxy (these are a naive "raid 0" type cluster with `q=12` and `n=`1 - so just for speed, not for redundancy)
>
> My question is - with our clustered setup - how many instances of Nouveau would we need: 1:1 (so 3 in our case) or can multiple couch nodes speak to a single Nouveau since the data is all the same?

mrjones-plip:
[...]

PR #10181, opened 2025-07-29
feat(#9707): add nouveau to helm charts

witash:
This adds a new pod and service for nouveau.

some notes:

- it does not add a new volume for nouveau data, instead it uses the same volume that couchdb uses. Nouveau is supposed to be mostly transparent, essentially just a part of couchdb, so we don't want to ask deployers to provision space for a separate volume, the size of which will be unknown and difficult to estimate. Also the main point of nouveau is to reduce storage space, so we don't want to have to ask for more instead. For multi-node couchdb this is a little complicated, because of course each couch node has its own volume, but we don't want nouveau to duplicate nouveau data for each node. So one solution is to just use the volume from the first couchdb node. This means both the storage size and IO will be imbalanced; the first node will be larger and busier in terms of IO operations than other nodes. For now, I'm assuming that this won't significant, but its something we should verify somehow; is the performance of the first couchdb node going to be significantly worse because of the extra IO operations from nouveau?
- For the nouveau instance it does add a new pod; if we really wanted to have all the resources for nouveau be combined with couchdb, we could add a nouveau container to each existing couchdb pod, instead of a new pod. This is also a change from the current performance; all freetext searches will use a single nouveau instance even if couchdb is clustered. We should also verify that that won't become a bottleneck, but it seems unlikely.
- Its mounted on a hardcoded subdirectory and will ignore the settings for preexisting data. Which means nouveau indexes will have to rebuild after a new deployment even if using preexisting data, which is a change from previous behavior where views would come along with the other data. But I don't think its worth it to add new settings for nouveau data paths, which no one is going to know or want to care about.
- I'm not sure what the tolerations are for, so I copied them from the couchdb template.


lorerod (approved):
LGTM ✅ The nouveau Helm templates are properly integrated and pass all validation tests across all deployment scenarios. Thank you @witash

PR #10182, opened 2025-07-29
bug: fix needs_signoff bug

jonathanbataire:
# Description

# Code review checklist

# License


api/tests/mocha/services/authorization.spec.js
```diff
@@ -2904,7 +2904,7 @@ describe('Authorization service', () => {
           doc: {
             _id: 'r1', type: 'data_record',
             contact: { _id: 'c1', parent: { _id: 'p1', parent: { _id: 'facility_id' } } },
-            fields: { patient_id: 'patient1', needs_signoff: true }
+            fields: { patient_id: 'patient1', needs_signoff: 'true' }
```
dianabarsan, line 2907: Unfortunately, I believe this is a breaking change. I know that our documentation says that the value needs to be `true` (boolean `true`), but I'm afraid that in the real world there will be many deployments that use other values, and we would be breaking their workflows if we deploy this change.

jonathanbataire: I see true
@dianabarsan do we push this to v5?

PR #10187, opened 2025-07-31
chore(#9639): remove old migrations [5.0]

### LINKED ISSUE #9639: Remove pre 3.x db migrations and protect `migration-log` doc from edits
Database migrations are messy. Some of the early database migrations were so heavy, they took down deployments for days when they ran. They're also quite complex, can be destructive and are gatekept by a document `migration-log`. If a migration doesn't exist in this log, it will run the next time api starts. Editing the `migration-log` doc is not validated in any way, it's not protected by validate_doc_update, so deletes or edits can be extremely dangerous to a deployment. 

1. Make sure only admins can edit the migration log. And never allow it to be deleted. We can also make some specific rule about never removing a migration from being run. 
2. Remove migrations that were added before 3.x. These are the heaviest and most destructive. Nobody runs CHT 2.x any more, we don't support it. The migrations can be kept for historical reasons, but their code should never run over a database. 

Leave it as it is.

dianabarsan:
# Description

Removes migrations that were added before 4.x, with some very few exceptions that add the resources document and the ones that change db security. 
Adds `migration-log` to protected admin documents. 
Refactors some migrations to not use db.query, instead use db.allDocs (because we run migrations before starting API, and we start API without requiring views to be warmed, and db.query does require the views to be warmed). 

closes #9639

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9639-remove-migrations/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9639-remove-migrations/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9639-remove-migrations/docker-compose/cht-couchdb-clustered.yml)

# License


dianabarsan:
Care to throw an eye on this cleanup job, @jkuester ? Appreciate it!

tests/integration/api/audit.spec.js
```diff
@@ -225,6 +231,7 @@ describe('auditing', () => {
         include_docs: true
       });
       expect(auditDocs.rows.length).to.equal(Math.ceil(revCount / MAX_HISTORY_LIMIT));
+      console.warn(JSON.stringify(auditDocs.rows, null, 2));
```
jkuester: Did you leave this in here on purpose?

jkuester:
image

Sooo satisfying! :heart_eyes:

tests/integration/api/audit.spec.js
dianabarsan: No, this was by mistake. I'll remove it.

Issue #8026, opened 2023-01-16
Admin app fails to update `branding` if existing document is malformed


dianabarsan:
**Describe the bug**
A malformed `branding` doc will block the user from updating branding from the admin app. No error is shown in the UI, only in console. 

**To Reproduce**
1. Create an `branding.json` document in your config. Set the contents of this document to be `{}`. run `cht upload-branding`. 
2. Login to admin app. Go to Images -> Branding.
3. Change the logo file. 
4. Click Submit. Nothing happens.

**Expected behavior**
The UI should inform the user that the existing document is malformed. 
Even when the document is malformed, the user should have the possibility to "fix" it by editing from the Admin UI. 

**Logs**
```
angular.js:15697 Error fetching resources file TypeError: Cannot read properties of undefined (reading 'favicon')
    at images-branding.js:24:29
    at processQueue (angular.js:18075:1)
    at angular.js:18123:1
    at Scope.$digest (angular.js:19242:1)
    at angular.js:19562:1
    at TaskTracker.completeTask (angular.js:21403:1)
    at angular.js:6879:1
```

**Environment**
- Instance: any
- App: admin
- Version: 3.4.0+

**Additional context**
[Form post](https://forum.communityhealthtoolkit.org/t/branding-not-working-using-admin-interface/2399/21)

jarias:
Hi this is Julio from Edify can you assign this to me

mrjones-plip:
Note - I added all the "Affects" labels, but we don't have them going all the back to `3.4`.  I'm not adding more labels though.

PR #10198, opened 2025-08-04
fix(#8026): validate empty branding doc

jarias:
# Description

When an empty branding JSON is uploaded via the `cht` tool the branding controller assumed certain keys existed in order to populate various $scope objects thus throwing an exception when they where were missing. Now the code checks if they exists exist and if they don't it creates a minimal set of keys in the objects that the template needs to work.

I created a test to validate the issue following similar conventions as other tests since this controller was lacking tests.

Closes medic/cht-core#8026

# License

### DIFF
--- a/admin/src/js/controllers/images-branding.js
+++ b/admin/src/js/controllers/images-branding.js
@@ -25,8 +25,13 @@ angular.module('controllers').controller('ImagesBrandingCtrl',
       return DB().get(DOC_ID, { attachments: true })
         .then(doc => {
           $scope.doc = doc;
-          $scope.favicon = doc._attachments[doc.resources.favicon];
-          $scope.icon = doc._attachments[doc.resources.icon];
+          if (!$scope.doc.resources) {
+            $scope.doc.resources = {};
+          }
+          if (doc._attachments && doc.resources) {
+            $scope.favicon = doc._attachments[doc.resources.favicon];
+            $scope.icon = doc._attachments[doc.resources.icon];
+          }
         })
         .catch(err => {
           $log.error('Error fetching resources file', err);
@@ -36,7 +41,7 @@ angular.module('controllers').controller('ImagesBrandingCtrl',
         });
     };
 
-    getResourcesDoc();
+    this.$onInit = getResourcesDoc;
 
     const validateTitle = () => {
       if (!$scope.doc.title) {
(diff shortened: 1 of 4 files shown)


jarias:
@dianabarsan this requires a review again cause I updated the code with the master branch since they where out of sync

admin/src/js/controllers/images-branding.js
```diff
@@ -25,8 +25,16 @@ angular.module('controllers').controller('ImagesBrandingCtrl',
       return DB().get(DOC_ID, { attachments: true })
         .then(doc => {
           $scope.doc = doc;
-          $scope.favicon = doc._attachments[doc.resources.favicon];
-          $scope.icon = doc._attachments[doc.resources.icon];
+          if (!Object.hasOwn($scope.doc, 'title')) {
```
dianabarsan: This is going to be a simple object, obtained form a JSON.parse call. So using `Object.hasOwn` is slight overkill.

admin/src/js/controllers/images-branding.js
```diff
@@ -25,8 +25,16 @@ angular.module('controllers').controller('ImagesBrandingCtrl',
...
+            $scope.doc.title = '';
+          }
+          if (!Object.hasOwn($scope.doc, 'resources')) {
+            $scope.doc.resources = {};
+          }
+          if (Object.hasOwn(doc, '_attachments') && Object.hasOwn(doc, 'resources')) {
+            $scope.favicon = doc._attachments[doc.resources.favicon];
```
dianabarsan, line 32: This might still throw an error if `doc.resources` exists but it has a value that cannot have the `favicon` property, for example `null`. 

image

dianabarsan (changes requested):
Very nice! Really appreciate adding the test file! 

I did find one issue that is not covered by this fix, comments below.

admin/src/js/controllers/images-branding.js
jarias: Not sure I understand, if the JSON is `{}` it won't have a title property and that breaks the form, that's why I added an empty title prop so it won't be undefined

admin/src/js/controllers/images-branding.js
jarias: I'll check this one, I didn't see this error.

jarias: Added some `ng-if` attribute in the HTML to prevent this errors

admin/src/js/controllers/images-branding.js
jarias: You where right the title is not needed, not sure why during my tests I thought it was.

jarias:
There still is a warning shown 

screenshot-20250805-12:56:43

This happens here https://github.com/medic/cht-core/blob/master/admin/src/js/services/resource-icons.js#L54 seams to me that the data hasn't been load yet thus the warning, but it doesn't affect the form functionality

And it happens even if the branding doc is not empty

admin/src/js/controllers/images-branding.js
dianabarsan: This one I got when my branding doc was: 
```
{
  ....
  resources: null,
  ...
}
```

dianabarsan:
I'm aware of the resource-icons library warning. I don't think you should concern yourself with them as part of this PR.

admin/src/js/controllers/images-branding.js
jarias: This case is now cover, sorry for the delay I was sick for almost a week :sneezing_face:

Issue #10225, opened 2025-08-15
Support Centralized SMS Gateway of Nepal run by Department of IT


binokaryg:
**Is your feature request related to a problem? Please describe.**

The Department of IT, Nepal has been running the centralized SMS gateway for a few years. Initially, it was focused on sending Bulk SMS, but now it can send A2P (Application to Person) SMS via API. Our goal is to enable CHT core to use this gateway for sending SMS messages in our Ministry of Health projects.

**Describe the solution you'd like**

We would like the CHT core to natively support the government-run SMS gateway of Nepal. By adding this support, our Ministry of Health projects can utilize the provided REST API to send SMS messages directly through this official gateway.

**Describe alternatives you've considered**
1. RapidPro Messaging Gateway: We explored using RapidPro's External API, but encountered a limitation. The DoIT API does not support the international phone number format, which is a requirement for RapidPro.
2. RapidPro Webhook: Another option was to call the SMS API webhook via a RapidPro flow. However, this approach is inconvenient and introduces intermediate dependencies like CHT outbound and RapidPro flows, complicating the setup and maintenance.
3. CHT outbound: We could call the DoIT endpoint directly using CHT outbound, but it would not work for scheduled messages. Also, it would be difficult to handle multiple SMS per report (although DoIT API supports this).

**Additional context**
- DoIT Centralized SMS gateway: [Integration Guide](https://sms.doit.gov.np/developer-guide/#/)
- Limitations:
  - The SMS service is currently outgoing only (A2P: Application-to-Person). It does not support replies or incoming messages (SMS from a person).
  - It does not provide the status of sent messages via API; status checks are available through a web dashboard.
  - This service is restricted to government entities. They must sign an agreement with the Department of IT and potentially incur upfront and operational costs.

PR #10230, opened 2025-08-16
feat(#10225): support for centralized sms gateway of nepal

binokaryg:
# Description

This PR adds support for Nepal Centralized SMS gateway.

#10225

# Code review checklist

# License

### DIFF
--- a/api/src/services/nepal-doit-sms.js
+++ b/api/src/services/nepal-doit-sms.js
@@ -1,0 +1,138 @@
+const request = require('@medic/couch-request');
+const secureSettings = require('@medic/settings');
+const logger = require('@medic/logger');
+const config = require('../config');
+
+const STATUS_MAP = {
+  // success
+  queued: { success: true, state: 'forwarded-by-gateway', detail: 'Processed' },
+  sent: { success: true, state: 'sent', detail: 'Sent' },
+
+  // failure
+  unauthenticated: { success: false, state: 'failed', detail: 'InvalidCredentials' },
+  invalid_number: { success: false, state: 'failed', detail: 'InvalidPhoneNumber' }
+};
+
+const getUrl = () => {
+  const settings = config.get('sms');
+
+  const url = settings &&
+    settings.nepal_doit_sms &&
+    settings.nepal_doit_sms.url;
+  if (!url) {
+    // invalid configuration
+    return Promise.reject('No URL configured. Refer to the Nepal DoIT SMS configuration documentation.');
+  }
+  return url;
+};
+
+const getCredentials = () => {
+  return secureSettings.getCredentials('nepal_doit_sms:outgoing')
+    .then(apiKey => {
+      if (!apiKey) {
+        return Promise.reject('No api key configured. Refer to the Nepal DoIT SMS configuration documentation.');
+      }
+      return { apiKey };
+    });
+};
+
+const getResponseMessage = res => {
+  return res &&
+    res.message;
+};
+
+const getStatus = responseMessage => {
+  // Response messages according to: https://sms.doit.gov.np/developer-guide
+  if (typeof responseMessage === 'string') {
+    if (responseMessage.toLowerCase().includes('queued')) {
+      return STATUS_MAP.queued;
+    }
+    if (responseMessage.toLowerCase().includes('sent')) {
+      return STATUS_MAP.sent;
+    }
+    if (responseMessage.toLowerCase().includes('unauthenticated')) {
+      return STATUS_MAP.unauthenticated;
+    }
+    if (responseMessage.toLowerCase().includes('invalid number')) {
+      return STATUS_MAP.invalid_number;
+    }
+  }
+}
+
+const generateStateChange = (message, res) => {
+  const responseMessage = getResponseMessage(res);
+  if (!responseMessage) {
+    return;
+  }
+  const status = getStatus(responseMessage);
+  return {
+    messageId: message.id,
+    // gatewayRef: recipient.messageId,
+    state: status.state,
+    details: status.message
+  };
+};
+
+
+const sendMessage = (credentials, message) => {
+  const url = getUrl();
+  logger.debug(`Sending message to "${url}"`);
+
+  // Strip the country code from recipient number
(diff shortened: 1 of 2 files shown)


binokaryg:
Message sent from the Messages tab in CHT: `This is a test message.`

API Logs:
```
2025-08-18T21:46:57.268 DEBUG: Checking for pending outgoing messages 
2025-08-18T21:46:57.320 INFO: Sending 1 messages 
2025-08-18T21:46:57.339 DEBUG: Sending message to "https://sms.doit.gov.np/api/sms" 
2025-08-18T21:46:58.192 DEBUG: SMS API Response: {
  message: 'Success! SMS queued !',
  status: 1,
  ntc: 1,
  ncell: 0,
  invalid_number: [ [length]: 0 ]
} 
```
CHT (Messages tab):
image

DoIT SMS dashboard:
image

Phone:
image


Multiply reply messages from a pregnancy registration containing long multi-SMS parts.

image

api/src/services/nepal-doit-sms.js
```diff
@@ -0,0 +1,147 @@
...
+  422: { success: false, state: 'failed', detail: 'UnprocessableContent' },
+  500: { success: false, state: 'failed', detail: 'InternalServerError', retry: true },
+};
+
+const getUrl = () => {
+  const settings = config.get('sms');
+
+  const url = settings &&
+    settings.nepal_doit_sms &&
+    settings.nepal_doit_sms.url;
```
sugat009:
```suggestion
  const url = settings?.nepal_doit_sms?.url;
```

api/src/services/nepal-doit-sms.js
```diff
@@ -0,0 +1,147 @@
...
+  if (!url) {
+    // invalid configuration
+    return Promise.reject('No URL configured. Refer to the Nepal DoIT SMS configuration documentation.');
```
sugat009:
```suggestion
    throw new Error('No URL configured. Refer to the Nepal DoIT SMS configuration documentation.');
```

The `getUrl()` function has inconsistent return types - it returns `Promise<rejected>` on error but a `string` on success. This breaks Promise chaining. 

Same issue exists in `getCredentials()`.

api/src/services/nepal-doit-sms.js
```diff
@@ -0,0 +1,147 @@
...
+    return;
+  }
+  return {
+    messageId: message.id,
+    state: status.state,
+    details: status.detail
+  };
+};
+
+const sendMessage = (credentials, message) => {
```
sugat009: suggestion: let's change this function into an async function. it's easier to debug and read.

api/src/services/nepal-doit-sms.js
```diff
@@ -0,0 +1,147 @@
...
+   */
+  send: messages => {
+    // get the credentials every call so changes can be made without restarting api
+    return getCredentials().then(credentials => {
+      const changes = [];
+      let promise = Promise.resolve();
+      
+      messages.forEach(message => {
+        promise = promise.then(() => processMessage(credentials, message, changes));
+      });
```
sugat009: question: how many messages are we expecting here? I think if it's a lot then, the messages can be sent in parallel and then can be awaited.

sugat009 (changes requested):

api/src/services/nepal-doit-sms.js
binokaryg: It depends on the configuration. `send` is called once per document from [here](https://github.com/binokaryg/cht-core/blob/9256c3623e37d3f741149f0f40bd7965b7055c72/api/src/services/messaging.js#L135) after getting `pendingMessages` from [here](https://github.com/binokaryg/cht-core/blob/10225-nepal-doit-sms/api/src/services/messaging.js#L214-L222):

```js
  send: docId => {
    const service = getOutgoingMessageService();
    if (!service) {
      return Promise.resolve();
    }
    return db.medic.get(docId).then(doc => {
      return sendMessages(service, getPendingMessages(doc));
    });
  },
```

```js
const sendMessages = (service, messages) => {
  if (!messages.length) {
    return;
  }
  return service
    .send(messages)
    .then(statusUpdates => module.exports.updateMessageTaskStates(statusUpdates));
};
```

Sometimes, a document can have several messages to be sent at once. For example:
- thank the sender
- alert the HF
- notify the patient

There can be more scheduled messages.

sugat009: ~~Cool cool, we can think about this later, if it causes some performance bottleneck.~~

I read this comment before I went through the changed code.

api/src/services/nepal-doit-sms.js
```diff
@@ -0,0 +1,163 @@
...
+    return;
+  }
+  return {
+    messageId: message.id,
+    state: status.state,
+    details: status.detail
+  };
+};
+
+const sendMessage = async (credentials, message) => {
```
sugat009, line 85: issue: ok, my bad on this one. I was thinking in the line that if the function was converted to `async` function then, the `promise chaining` below(Lines 75 -> 114) would be converted to `await` calls as well. Apologies for not being clearer earlier.

sugat009 (changes requested):

api/src/services/nepal-doit-sms.js
binokaryg: I've already updated the code to send messages in batches.

api/src/services/nepal-doit-sms.js
binokaryg: If we `await` for `request.post` on line 76, is it worth making the `sendMessage` `async`?

sugat009: To use a `await` inside a function, the function needs to be an `async` function.

image

binokaryg: But it still returns a promise right?

binokaryg: image

sugat009: No, the `result` variable would hold the object that is returned from the request not a `promise`. If the `await` is removed, then `result` would be holding a `promise`.

sugat009 (approved):

Issue #10224, opened 2025-08-15
Support custom UI Extensions


ChinHairSaintClair:
**Is your feature request related to a problem? Please describe.**
We recently encountered a syncing issue affecting CHWs for an extended period - over two months in some cases. Unfortunately, this was only discovered at a very late stage, which caused serious concern about potential data loss (e.g., cache clearing or stolen devices) and made resolving the issue a very high priority - right as other pressing work was underway. A tool to proactively indicate device connectivity or syncing issues would have reduced both downtime and overall impact. This feature would allow us to take a more proactive approach rather than just reacting after problems occur.

**Describe the solution you'd like**
It would be great if each implementation/flavour of CHT had the ability to configure or create pages specific to their context. For example, in our context, we require four screens:

1. **Connection/Sync Status** - Display CHWs’ device connection and syncing information so that we can identify issues such as syncing failures, device connectivity problems, or SIM/APN misconfigurations early.

2. **Integration Screen (iframe)** - Embed other apps we manage, so that CHAMP becomes a one-stop shop for all user needs without requiring navigation to a new browser tab.

3. **Landing Page** - A central hub for links to other pages, training materials, and additional resources.

4. Deduplicate Page - A space where, retroactively and on every sync, duplicates can be displayed for CHWs to address. The backend would flag items it cannot merge (that require human review), and once confirmed by a CHW as potential matches, OTLs would sign off and the data would be sent back for processing.

These screens may not be relevant to every implementation, so the ability to add only the content you need, without affecting the wider community (until they require it), would be ideal.

**Describe alternatives you've considered**
Build each required screen directly into our fork of the app.

**Additional context**
Configurable pages were initially discussed with Nicole on the [UI Evolution](https://forum.communityhealthtoolkit.org/t/the-ui-evolution-of-the-cht/3515/10) forum thread.

ChinHairSaintClair:
We’ve been working on a POC that I plan to open as a PR next week.

jkuester:
Let's go!  I have been kicking around the idea of CHT "plugins" for a couple years now.  My last notes that I had on the subject was the idea of supporting a web component interface in the CHT so custom plugins could be hosted by the CHT server with API hooks to interact with data (perhaps both in an online or offline context).

Very excited to see what you are cooking up here!

ChinHairSaintClair:
[...]

PR #10232, opened 2025-08-18
feat(#10224): Configurable Page Plugin POC

ChinHairSaintClair:
Related issue: (#10224)[https://github.com/medic/cht-core/issues/10224]

# Table of Contents
- [Introduction](#introduction)
- [Config Run Down](#configuration-run-down)
- [The Table Template](#the-table-template)
- [The Iframe Template](#the-iframe-template)
- [The Grid Template](#the-grid-template)
- [The Dynamic Template](#the-dynamic-template)

# Introduction:
This solution is still in an early stage and the interface is also basic (for now). The aim of sharing it early is to get more eyes on the work and start a discussion about whether the approach makes sense and how it can be improved.

The POC contains four template types to cover the needs we have at the moment. These are used here as examples to show what the plugin can do:
- `table`
- `iframe`
- `grid`
- `dynamic`

# Configuration run down

The configuration is defined in the `app_settings.json` file, under the `pages` block. Multiple pages can be specified, each with a unique identifier:

```
{
    ...
    "pages": {
        "my_page_1": {
            ...
        },
        "my_page_2": {
            ...
        },
    },
    ...
}
```

This identifier is used as a sub-path within the `custom` path in the app router and also serves as the translation key for the page title.

image

Page can be set as either `primary` or `secondary`. Primary pages display in the main app bar while secondary pages appear in the hamburger menu:

```
{
    "pages": {
        "my_page_1": {
            "tab_type": "secondary",
            ...
        },
        ...
        "grid": {
            "tab_type": "primary",
            ...
        },
    }
    ...
}
```

image

Page visibility can also be controlled per role using permissions:

```
{
    "pages": {
        "my_page_1": {
            "tab_type": "secondary",
            "permissions": [
                "can_view_my_page_1"
            ],
            ...
        },
        ...
    }
    ...
    "permissions": {
        ...
        "can_view_my_page_1": [
            "national_admin"
        ],
        ...
    },
    ...
}
```

# The `table` template:
This template is meant for data that can be displayed in a tabular form. A page header is included to provide context. It can be defined in two ways:
- custom HTML content, which is rendered by the `DynamicTemplateRenderer` (see the Dynamic Template), or
- a predefined header with a title (`<h1>`) and an [expansion panel](https://material.angular.dev/components/expansion/overview) for the descriptive content.

Controls, for now only a `reload` button, are also provided. The table takes up the rest of the available screen space.

## Table properties:
| name | purpose | data type |
| ------ | --------- | ----------- |
| `page_header` | Defines the purpose of the page and provides context about the presented data | object |
| `table_headers` | Defines columns, their order, and customization options | object (see below) |
| `on_screen_load` | Code executed when the page loads. Supports async code and should return the object used to populate the table. Vars and context exposed to the function is limited (see function context section) | string |
| `on_reload` | Code executed when the reload button is clicked. Receives the above load function in its context and should return the updated table data. | string |

## Predefined header properties:
| name | purpose | data type |
| ------ | --------- | ----- |
| `title` | In-page content title | string | 
| `description_collapsed` | Description shown when the panel is collapsed | string |
| `description_expanded` | Description shown when expanded. Can be plain text or HTML (rendered by `DynamicTemplateRenderer`) | string | 
| `expanded` | Default state of the panel | boolean |

## Column configuration options:
|name|purpose|data type|
|------|---------|----------|
|`can_sort`| Enables sorting (`asc`, `desc`, or `none`) via the column header | boolean |
|`display_value`| Column label shown to the user | string |
|`pipe`| Applies predefined transformers (`date`, `decimal2`) | string |
|`preprocess_hook`| Function to adjust cell values before display. Limited context is exposed to this function |string|

## Example implementation:
A recent bout with syncing issues made it very clear that we needed some form of early detection system/dashboard for when app syncing starts misbehaving. Which could be due to a variety of reasons. For us this was a perfect storm of VPN related issues, very large syncing requests (possible due to attachments), and our nginx configuration being too [strict](https://forum.communityhealthtoolkit.org/t/gateway-timeouts-on-password-reset/4627/2?u=anro). 

| source                                                  | purpose | constraint | example result |
|------------------------------------------|----------|------------|------------------|
| `medic-logs` 's `connected-users` view | Lists active users with timestamps of their last server interaction. Helps identify connection issues and will be used to build up telemetry results per user | 2 months + 8 months offset to also capture "recently" inactive users | ``` {"id":"connected-user-temp", "key":1747059635000, "value":null } ``` |
| [medic-users-meta](https://forum.communityhealthtoolkit.org/t/sync-issues-status-scrutiny/3359/3?u=anro) | Aggregated user docs used to calculate sync and connection results. Batched requests of 25 to handle volume | type: `telemetry`, `ms-since-last-replicated-date` entry exists, captured within the last 2 months, and the telemetry entry belongs to one of the entries in the `connected-users` list. Finally, only certain fields are queried as shown. | ``` { "metadata":{ "user":"medic", "year":2025, "month":6, "day":10, "aggregate_date":"2025-06-11T06:33:47.941Z" }, "dbInfo": { "doc_count": 129, "db_name": "medic" }, "metrics":{ "replication:medic:from:ms-since-last-replicated-date":{ "sum":3274331171, "min":1091443464, "max":1091443953, "count":3, "sumsqr":3573748205794198000 }, "replication:medic:to:ms-since-last-replicated-date": { "sum": 3274329903, "min": 1091442561, "max": 1091443818, "count": 3, "sumsqr": 3573745437894194700 } } } ``` |
| [app metrics](https://docs.communityhealthtoolkit.org/building/reference/api/#get-apiv1express-metrics) (planned)| Will show request/response times and sizes | TBC | TBC |

### Metrics tracked
| purpose | key |
|---------|---------------|
| When last the user made contact with the server. | `last_server_connection_date` |
| What is the date when the user was able to sync `medic` records to the server | `latest_to_sync_date_ms`, `latest_from_sync_date_ms` |
| What is the average amount of days between syncing to medic (adjusted for weekends), should help shed light on irregular syncing behaviour | `total_days_between_up_sync`, `total_days_between_down_sync` |
| The gap in activity, is the user unable to log in and perform some action for extended periods | `total_inactive_days` |
| Approximate doc count, could help shed some light on why syncing issues are occurring | `latest_approximate_device_doc_count` |

Template exposed code context variables:
| name | purpose |
|-------|-----------|
| `dbService` | The full CHT db service to perform various queries |
| `moment` | The full [moment library](https://momentjs.com/) to perform date calculations |
| `ctx` | At the moment, it refers to the `this` property |
| `onScreenLoad` (only to the `on_reload` function) | To augment the page load function |

Determining what to expose safely without foreclosing is a tricky endeavour, we would appreciate any guidance.

### Example code output:
The metric keys mentioned above are used as the column names as well as the keys for the user specific entries when building up the results for the table. This is to determine which value belongs in which column. 

The example `on_screen_load/on_reload` output would be:
```
[
   ...
   {
      "testing": {
         "last_server_connection_date": 1755169565027,
         "mostRecentTelemetry": 1754863198738,
         "latest_from_sync_date_ms": 1754863198738,
         "latest_to_sync_date_ms": 1754863198738,
         "latest_approximate_device_doc_count": 119,
         "total_days_between_down_sync": 51.00020707175926,
         "total_days_between_up_sync": 51.00020707175926,
         "total_inactive_days": 4,
         "entryCount": 4
      },
      "medic": {
         "last_server_connection_date": 1755502164104,
         "mostRecentTelemetry": 1754949582191,
         "latest_from_sync_date_ms": 1754949582191,
         "latest_to_sync_date_ms": 1754949582191,
         "latest_approximate_device_doc_count": 135,
         "total_days_between_down_sync": 62.26467842592592,
         "total_days_between_up_sync": 62.26467842592592,
         "total_inactive_days": 4,
         "entryCount": 4
      },
      "national_admin": {
         "last_server_connection_date": 1755513328477,
         "mostRecentTelemetry": 1755122398211,
         "latest_from_sync_date_ms": 1755122398211,
         "latest_to_sync_date_ms": 1755122398211,
         "latest_approximate_device_doc_count": 136,
         "total_days_between_down_sync": 0,
         "total_days_between_up_sync": 0,
         "total_inactive_days": 1,
         "entryCount": 1
      }
   }
   ...
]
```
The results are in grouped in "buckets" per user. We also capture an additional property `entryCount` which is used to calculate the averages. If we look at the full example config below, we'll see this value is passed back into the `preprocess_hook`.

### Example config:
```
"pages": {
...
"table": {
      "tab_type": "primary",
      "permissions": [
        "can_view_table"
      ],
      "template": {
        "key": "table",
        "page_header": {
          "title": "Sync & Connectivity Dashboard",
          "description_collapsed": "Login data can be delayed by a few days",
          "description_expanded": "<p><strong>NOTE: Replication tracking can be delayed by a few days due to how the telemetry records are created. Each time an action is performed on the client, a record will be created in the user's own meta db. The aggregate doc for the previous day is created when the first telemetry item is recorded each day. After which the singular telemetry docs will be removed from the user's specific db to reduce the amount of records required to sync. Sentinel wil then pick up these documents and copy them over to the communal meta database, this aggregation process occurs \"nightly\". <a href=\"https://forum.communityhealthtoolkit.org/t/sync-issues-status-scrutiny/3359/11\">Read more on the related CHT Forum Post</a> </strong></p>",
          "expanded": true
        },
        "table_headers": {
          "username": {
            "can_sort": false,
            "display_value": "Username:"
          },
          "last_server_connection_date": {
            "can_sort": true,
            "display_value": "Last Server Connection:",
            "pipe": "date"
          },
          "latest_to_sync_date_ms": {
            "can_sort": true,
            "display_value": "Last ↑ Sync:",
            "pipe": "date"
          },
          "latest_from_sync_date_ms": {
            "can_sort": true,
            "display_value": "Last ↓ Sync:",
            "pipe": "date"
          },
          "total_days_between_up_sync": {
            "can_sort": true,
            "display_value": "Avg days between ↓ sync (w adj):",
            "preprocess_hook": "const entryCount = ((rowData ?? {}).entryCount || 0); return entryCount > 0 ? (cellValue / Math.max(entryCount - 1, 1)).toFixed(2) : 'NA';"
          },
          "total_days_between_down_sync": {
            "can_sort": true,
            "display_value": "Avg days between ↑ sync (w adj):",
            "preprocess_hook": "const entryCount = ((rowData ?? {}).entryCount || 0); return entryCount > 0 ? (cellValue / Math.max(entryCount - 1, 1)).toFixed(2) : 'NA';"
          },
          "total_inactive_days": {
            "can_sort": true,
            "display_value": "Activity gap (w adj):",
            "preprocess_hook": "const entryCount = ((rowData ?? {}).entryCount || 0); return entryCount > 0 ? (cellValue / Math.max(entryCount, 1)).toFixed(2) : 'NA';"
          },
          "latest_approximate_device_doc_count": {
            "can_sort": true,
            "display_value": "Approximate doc count:",
            "preprocess_hook": "const entryCount = ((rowData ?? {}).entryCount || 0); return entryCount > 0 ? (cellValue / Math.max(entryCount, 1)).toFixed(2) : 'NA';"
          }
        },
        "on_screen_load": "const startingPoint=moment().subtract(2,'months');const connectedUsersCutoff=startingPoint.clone().subtract(8,'months').valueOf();const strict=false;const medicLogs=dbService.get({remote:true,custom:'medic-logs'});const medicLogsInfo=await medicLogs.info();console.log('Remote medic logs info: ',medicLogsInfo);const connectedUsersQueryResult=await medicLogs.query('logs/connected_users',{reduce:false,startkey:strict?startingPoint.valueOf():connectedUsersCutoff,endkey:9999999999999});console.log('Connected users query result: ',connectedUsersQueryResult);const connectedUsers=connectedUsersQueryResult?.rows?.map(entry=>({name:entry.id.replace(/^connected-user-/,''),timestamp:entry.key}));console.log('Connected users (active this year): ',connectedUsers);const countWeekdaysBetween=function(startDate,endDate){const SUNDAY=0;const SATURDAY=6;let count=0;const current=moment(startDate).clone();while(current.isBefore(endDate,'day')){const day=current.day();if(day!==SUNDAY&&day!==SATURDAY){count++}current.add(1,'day')}return count};const adjustedGapExcludingWeekends=function(prevDate,currentDate){const SUNDAY=0;const SATURDAY=6;const totalDays=currentDate.diff(prevDate,'days',true);if(totalDays<=0){return 0}let weekendDays=0;const checkDate=prevDate.clone().add(1,'days');while(checkDate.isBefore(currentDate,'day')){const day=checkDate.day();if(day===SUNDAY||day===SATURDAY){weekendDays++}checkDate.add(1,'days')}return totalDays-weekendDays};const getTelemetry=async function(dbService,yearLowerLimit,monthLowerLimit,connectedUsers){console.log('Year lower limit: ',yearLowerLimit);console.log('Month lower limit: ',monthLowerLimit);const medicUsersMeta=dbService.get({remote:true,custom:'medic-users-meta'});let bookmark=null;let results=[];const limit=25;let response;do{const query={selector:{type:'telemetry','metrics.replication:medic:from:ms-since-last-replicated-date':{$exists:true},'metadata.year':{$gte:yearLowerLimit},'metadata.month':{$gte:monthLowerLimit},'metadata.user':{$in:connectedUsers.map(e=>e.name)}},fields:['metadata.user','metadata.year','metadata.month','metadata.day','metadata.aggregate_date','metrics.boot_time:2:to_bootstrap','metrics.replication:medic:from:ms-since-last-replicated-date','metrics.replication:medic:to:ms-since-last-replicated-date','dbInfo.doc_count','dbInfo.db_name'],limit:limit,bookmark:bookmark||undefined};response=await medicUsersMeta.find(query);results=results.concat(response.docs);bookmark=response.bookmark}while(response.docs.length===limit);console.log('Telemetry results: ',results);const data=Object.fromEntries(connectedUsers.map(user=>[user.name,{last_server_connection_date:user.timestamp,mostRecentTelemetry:-1,latest_from_sync_date_ms:-1,latest_to_sync_date_ms:-1,latest_approximate_device_doc_count:-1,total_days_between_down_sync:0,total_days_between_up_sync:0,total_inactive_days:0,entryCount:0}]));for(const entry of results??[]){console.log('Entry: ',entry);console.log('username: ',entry.metadata.user);const bucket=data[entry.metadata.user];if(bucket){const startPeriod=`${entry.metadata.year}-${entry.metadata.month}-${entry.metadata.day}`;console.log('Start period: ',startPeriod);const momentStartPeriod=moment(new Date(startPeriod));const diffInWeekDays=countWeekdaysBetween(startPeriod,moment(new Date(entry.metadata.aggregate_date)));console.log('Diff in days: ',diffInWeekDays);bucket.total_inactive_days+=diffInWeekDays;const msSinceLastRepUpSuccess=entry.metrics['replication:medic:to:ms-since-last-replicated-date'].min;const msSinceLastRepDownSuccess=entry.metrics['replication:medic:from:ms-since-last-replicated-date'].min;const lastSuccessfulRepUpDate=momentStartPeriod.subtract(msSinceLastRepUpSuccess,'milliseconds');const lastSuccessfulRepDownDate=momentStartPeriod.subtract(msSinceLastRepDownSuccess,'milliseconds');const telemetryTimestamp=momentStartPeriod.valueOf();if(momentStartPeriod.valueOf()>bucket.mostRecentTelemetry){console.log('Last up: ',lastSuccessfulRepUpDate.toISOString());console.log('Last down: ',lastSuccessfulRepDownDate.toISOString());if(bucket.latest_from_sync_date_ms>0){const prevDate=moment(bucket.latest_from_sync_date_ms);const gapDays=adjustedGapExcludingWeekends(prevDate,lastSuccessfulRepDownDate);bucket.total_days_between_down_sync+=gapDays}if(bucket.latest_to_sync_date_ms>0){const prevDate=moment(bucket.latest_to_sync_date_ms);const gapDays=adjustedGapExcludingWeekends(prevDate,lastSuccessfulRepUpDate);bucket.total_days_between_up_sync+=gapDays}bucket.latest_from_sync_date_ms=lastSuccessfulRepDownDate.valueOf();bucket.latest_to_sync_date_ms=lastSuccessfulRepUpDate.valueOf();bucket.latest_approximate_device_doc_count=entry.dbInfo?.doc_count??bucket.latest_approximate_device_doc_count;bucket.mostRecentTelemetry=telemetryTimestamp}bucket.entryCount++}};return data};const info=await getTelemetry(dbService,startingPoint.year(),startingPoint.month()+1,connectedUsers);console.log('Info: ',info);return info;",
        "on_reload": "console.log('This reload content is pushed via config'); return await onScreenLoad();"
      }
    },
...
```

https://github.com/user-attachments/assets/0630a2ef-2792-443c-b994-06968e4d8bce

# The `iframe` template:
This template is used to embed other webapps or website into CHT. The goal is to reduce the need for users to manage multiple browser tabs on their devices. 

One intended use case is integration with a referral app (we have control over). A user could either navigate to the embedded app through a tab or menu option, or navigate directly to an existing referral entry through a configured contact link. Speaking under correction, because an app route pointing to the embedded site has been registered - we'll be able to append data to the URL when loading up the request. This is still being tested. The directly linking would be great but isn't strictly required. The 'referral_ref' would be appended to the relevant contact doc by a separate backend process.

We should be able to make the provided links "instance agnostic" by getting the FQDN from the `LocationService` by using:
```
this.locationService.url
```
So, we could use something like [lodash](https://lodash.com/) (already included in CHT) to swap out that URL part.

## Configuration props:
| name | purpose | data type |
| ------ | --------- | ----------- |
| `url` | Address of the external app or site to embed. Content is rendered within a CHT page. CHT does not handle authentication, users log in as they normally would | string |

## Example config:
For now, the configuration points back to CHT itself to demonstrate the behaviour. This looks unusual because a single-page app is opening another instance of itself.
```
...
"pages": {
   "my_page_1": {
      "tab_type": "secondary",
      "permissions": [
         "can_view_my_page_1"
      ],
      "template": {
         "key": "iframe",
         "url": "http://localhost:5988/#/about"
      }
   }
}
...
```

image

# The grid template:
This template works well as a landing page. Currently, it displays links as thumbnail cards, but it could be extended to include other resources like bundled images, documents, or video links. See the (UI Evolution)[https://forum.communityhealthtoolkit.org/t/the-ui-evolution-of-the-cht/3515/10] forum thread for related discussion.

## Configuration props:
| name | purpose | data type |
|--------|----------|------------|
| `items` | A list of entries that link to various pages within the app | Array |
| `items[n].url` | Destination when the card is clicked | string |
| `items[n].thumbnailResource` | Resource reference defined in `resources.json` | string |
| `items[n].description` | Text shown on the card | string |

## Example config:
```
...
"pages": {
...
"grid": {
      "tab_type": "primary",
      "permissions": [
         "can_view_grid"
      ],
      "template": {
         "key": "grid",
         "items": [
            {
               "url": "http://localhost:5988/#/about",
               "thumbnailResource": "icon-family",
               "description": "This is a small description of this page's functionality"
            },
            {
               "url": "http://localhost:5988/#/user",
               "thumbnailResource": "icon-healthcare",
               "description": "This is another description"
            },
            {
               "url": "http://localhost:5988/#/analytics/targets",
               "thumbnailResource": "medic-district-hospital",
               "description": "This is also one"
            },
            {
               "url": "http://localhost:5988/#/tasks",
               "thumbnailResource": "icon-infant",
               "description": "Lorem ipsum"
            }
         ]
      }
   },
...
```

image

# The `dynamic` template:
This template is for cases where none of the others fit. It accepts user-defined HTML, always strips out `<script>` tags (and their contents), and by default sanitizes the remaining content using Angular's `DomSanitizer`. The sanitization step will later be configurable.

The template is still a work in progress. We plan to use it to showcase a deduplication process. 

## Configuration properties
| name | purpose | data type |
| ------ | --------- | ----------- |
| `context` | Variables exposed to the HTML template, populated by supplied functions | obj |
| `html` | HTML content rendered by the `DynamicTemplateRenderer`. | string |
| `on_screen_load` | Code executed when the page loads. Can be synchronous or async. Supplies data to the template. Similar to the `table` section's `on_screen_load/on_reload` functions, a set of app context variables will be exposed to this function. | string |
| `on_reload` | Similar to `on_screen_load`, but additionally receives the loading function as a variable. | string |

# Code review checklist

# License


jkuester:
:tada: Thanks for the detailed documentation here!  I might not make it through everything this week, but just wanted to leave a note out here that I saw this and am looking forward to checking it out!

ChinHairSaintClair:
@jkuester the `template-renderer` code has been refactored into an Angular lib and should have no static imports from CHT core.
However, the code is still included in this PR to make it a little easier to follow the flow of data as we work on refining the plugin functionality. 

Also, in the PR description I neglected to mention that we had to add the "national_admin" role to the `medic-logs` db's `Roles` in order to access the database and query the telemetry data shown.

jkuester:
Okay, sorry for the delayed response here, but I finally got a few hours to spin this up and tinker with it!  Great POC!  :tada: 

There is a lot going on here, so I think I will start the conversation with some of the higher-level concepts and we can narrow things in from there.  

Regarding a plugin framework in general, here you have gone with a "template-based" approach where the CHT app supports configuring various custom tabs/pages according to predefined templates. 

Pros: 

- Lets you configure plugins with low/no-code (just configure in your app settings)
- Gives the webapp full control over the look and feel of the pages (aside from the contents of the iFrame, of course)

Cons:

- Limits the functionality of the plugins to whatever is supported in the framework.
- Supporting additional types/variants of a template requires changes to cht-core.
- The result is _a lot_ of code in cht-core to maintain.

After some reflection on this (and consideration of the existing cht-core functionality especially around things the contact-summary and extension-libs), I think how I have it framed in my head is that features/functionality that is "core" to the CHT should be implemented with this "template-based" approach. On the other hand, things that "extend" the CHT features/functionality, but are not really common to most CHT deployments should be realized in the CHT as more generic plugins where most of the source code for the plugin content/behavior exists in the plugin and interacts with the rest of the CHT via well-defined APIs. Either way, cht-core supports adding additional tabs/pages (as you have done, either as top-level tabs or in the sidebar). 

At this point I do not yet have a strongly formed opinion on which of these cases we are dealing with here and/or which  implementation approach is best. (See the next section for some more concrete thoughts on how to proceed next. Hopefully consideration of an actual problem statement and how it should be solved can give us a firm direction here...)

Please share if you have specific thoughts regarding this distinction here and/or have any other approaches to include in addition to these two!

---

> A recent bout with syncing issues made it very clear that we needed some form of early detection system/dashboard for when app syncing starts misbehaving. Which could be due to a variety of reasons. For us this was a perfect storm of VPN related issues, very large syncing requests (possible due to attachments), and our nginx configuration being too [strict](https://forum.communityhealthtoolkit.org/t/gateway-timeouts-on-password-reset/4627/2?u=anro).

As always, I think the most effective way to proceed forwards is to identify a specific need/use-case that your organization is actually experiencing and we can focus on a solution for that need. (Of course we include mild generalizations to make the feature useful to the broader community, but having a specific goal in mind will ensure that we only build things that are needed (and not stuff that we just think is useful...)).  

With that in mind, I think the case you have mentioned regarding unknown syncing issues is a great one!  (But if you have a different one that is more relevant/pressing, we can definitely pivot in that direction.)  

To dig in a little bit, can you provide more information about these syncing problems?  

a. What kind of users needed to know about the issue (but did not)? What is their role/responsibility in the org?
b. Are those users normally using the CHT app already? (Are they online or offline users?)
c. What actions should those users take if they identify syncing problems?

---

c.c. @mrjones-plip, would love your thoughts here as well (given our previous conversations regarding CHT plugins)!

ChinHairSaintClair:
Thank you for your response @jkuester. 

The PR contains two parts:
- **Plugin core** - the proposed addition to CHT core (including the header tab changes)
- **Example plugin (`template-renderer`)** - included to make it easier to follow the plugin flow and to discuss our need

The plugin manager (`/webapp/src/ts/plugins/core/plugin.manager.service.ts`) is designed to load and communicate with any plugin that implements the plugin contract (`plugin.contract.ts`). To support testing:
- `template-renderer` was added to the `/webapp/angular.json`, and 
- a `preinstall` script was added to the `/webapp/package.json` .

The symlink created by the `"@plugins/template-renderer": "file:plugins/template-renderer/dist"` mimics what a plugin dependency import into the `node_modules/@plugins` group would look like and gets the temp test setup working.

From here, the plugin system takes over: 
- `/webapp/tsconfig.json` sets up a path alias to the `@plugins` node modules group. 
- The manifest generator (`/webapp/scripts/plugin-manifest.generator.js`) loops through the libs in the `node_modules/@plugins` group and creates/updates the manifest file (`/webapp/src/ts/plugins/auto-gen-manifest.ts`)
- That manifest contains a list of lazy (async) imports to these plugins using the alias path from `tsconfig`. 
- Because these are concrete but lazy dependencies, we shouldn't run into tree-shaking issues. 
- The plugin manager depends on this manifest and lazy-loads the plugins defined within, passing along the necessary dependencies through the contract.

Apologies if the `template-renderer` caused confusion. The system was developed "in reverse": 
- We first started off developing the screens we required, 
- Then generalized them into templates, 
- Figured out how to populate the contents using config, 
- Finally, figured out how to load it all as a dependency to core, while keeping core lean.

We fully agree that non-core functionality should live outside of core. That is the intent here: the plugin functionality allows us to extend functionality in our deployment without impacting the broader community.

That said, a few small core changes may still be needed - like the ability to add top-level tabs or sidebar items. The current "dynamic tab" implementation in this PR is still very basic (it only works on the old nav, not scrollable yet if many tabs are added).

---

## Why we need this:
We lacked visibility into individual users' sync statuses and overall instance health/performance, and as a result only discovered syncing issues after they had persisted for weeks. Once resolved, it became clear that our DHOs and site managers need access to this information at a glance so they can catch problems early, and act quickly, as they are the solution's first line of defence. 

Watchdog already provides great overall instance health/performance data, but (as far as I know) it doesn't drill down to sync status by user or site (NPO). We'd also like the option to enrich this with external data (e.g. MDM, APN status) to catch connection issues outside the app. Having all this easily consumed info consolidated in one place would help those responsible be more proactive rather than just reactive.

## Who needs this (their roles):

**DHOs**:
In addition to their everyday work items such as training, area mapping, device theft reporting, safety monitoring, assist with contact move processes, etc, they are also required to manage and monitor many CHWs and their devices. 

**Site manager (admin)**:
Monitor and manage site performance, finances, adherence to procedures and guidelines, community engagement, assisting DHOs and handle reported issues as a branch of the admin/central support team. Probably not an exhaustive list of their responsibilities.

**Dev team (sys admin)**:
To keep the very small dev team unburdened by day-to-day tasks and focused on fix and feature development (of which there are many), the DHOs and site managers act as the 'line of defence', and only after their review are issues escalated. 

## App usage:
All of the above roles create, manage, and use offline and online users. They use the system. However, that isn't necessarily their primary function within their role. Having the tool(s) available to watch for possible issues at a quick glance while also being productive on the day-to-day work would be a huge benefit.

## Expected actions:
- DHOs - periodically monitor for issues, troubleshoot devices, swap out devices/sims, escalate if unresolved,
- Site managers - periodically monitor for issues, troubleshoot devices, work with APN and/or VPN providers, escalate app issues,
- Devs - periodically monitor for issues, troubleshooting, code/config/infra changes

---

### Beyond sync monitoring:
Another driver behind the `template-renderer` plugin is, not only a "landing page", but also "integration" with external apps:
- Linking from a contact to a referral app (new or existing referrals)
- Enabling VAPs (Verbal Autopsy Practitioners) to access a clinical viewer for death confirmation workflows

More details around this can be seen in the linked issue.

ChinHairSaintClair:
Hi @jkuester just checking in on this one. We'd really appreciate your thoughts so that we know how to best move it forward. Any concerns we should be aware of? Happy to make adjustments where needed.

jkuester:
@ChinHairSaintClair sorry for the delay here!  :disappointed: My work queue has been :fire: recently.  The good news is that I have had some time to discuss this project with some colleagues and there is agreement that this effort is important enough that we should have a [dedicated squad](https://docs.communityhealthtoolkit.org/community/squads/) to coordinate implementation.  See the linked docs for more details on CHT squads (if you are not already familiar).

I have started the ball rolling with [a forum post](https://forum.communityhealthtoolkit.org/t/custom-cht-plugins-squad/5393). If you are open to pursing this feature as part of a squad please reply out there on the forum. :pray: My hope is that we can take a week or so to collect any other folks who are interested in helping and then get things kicked off.  (In my experience the squad model adds just enough "process" to keep things organized/coordinated and progressing while still keeping it flexible enough that folks who are volunteering their time can participate without being overwhelmed.)

Let me know if you have questions/concerns about any of this!

jkuester:
Also, I think it is worth calling out that a squad does feel a bit "out of order" in this context since we already have this PR in motion.  :grimacing:   I am definitely not trying to ignore what you have already accomplished here or say we should not go in this direction.  The whole idea of the squad is that this plugin functionality is a really crucial core functionality and it deserves taking a step back with other interested community members to make sure we have the right orientation for the solution.  One hopeful output from the squad is that we coalesce on your approach here as being what we want to include in the code and we are good to go!

Another thing to note is that squad decisions are basically made by the folks who get involved and show up to share their thoughts.  If it ends up just being the two of us :sweat_smile: that is still fine and it will make for a simple decision making process....

ChinHairSaintClair:
@jkuester it looks like it's my turn to apologise 😅. Sorry for the delayed response! We've been working on the deduplication solution, which was meant to be included in the `template-renderer` plugin example code in this PR. As a quick prototype, we added the screen directly in our instance and started a [forum post](https://forum.communityhealthtoolkit.org/t/deduplication-prototype/5412) to engage around some next steps - like introducing custom db views to support replication of custom doc types, and exploring the potential need for a shared component library.

Thank you for taking the time to reach out to the community about the plugin work. I agree that this effort touches several areas, so having more eyes on it makes sense. That said, if it ends up just being the two of us that's fine too - we got brilliant guidance during the JIT duplicate work, so we know we're in good hands 😁.

Hope your workload is easing up a bit now!

jkuester:
No worries! Looks like all our delays have actually worked out for the best, though, as we are getting some good interest from other partners is participating in a Plugin Squad.  So, be watching for more updates on that, probably next week!  

(Also, thanks for the great post about your deduplication feature!  I responded out there earlier today, but I am really looking forward to seeing where we can go with that too. :rocket:).

jkuester:
@ChinHairSaintClair given the ongoing work of the CHT Plugins squad (with the [design doc](https://docs.google.com/document/d/1RFiSS2hdYhLloNUcahlDVX6KN5ZNEQhUnnc5EVDR9Pc/edit?tab=t.87hclgo86rim#heading=h.g1wey3uo5gpy) including a link to this PR for reference), do you think we can go ahead and close this PR in favor of incoming web-component work from the squad?

ChinHairSaintClair:
Thank you for checking back in on this @jkuester. Before I give an answer in haste, I want to sanity-check my understanding of the different approaches by pitting them against each other. Please correct me where I'm off. This will help me make a more informed call and better contribute to the plugin squad going forward.

| Topic | POC: Bundled plugin libs (this PR) | We-component - config based UI extensions |
|---|---|---|
| Delivery | Supplied as libraries and bundled with the webapp at build time. Higher barrier of entry (image building).  | Supplied via config and uploaded as doc attachments. Familiar usage paradigm, quicker to onboard, simpler project setup. |
| Download model | All users download all plugins. It blurs the line between "core" and "extension". Larger download size. Visibility is largely cosmetic. | Extensions are replicated/downloaded conditionally by role/permission. Additionally, the same runtime routing checks apply as in POC (permission change). Mostly, code that shouldn't exist, won't. | 
| Bundling size | Always increases, regardless of usage. | Only increases for users who actually need the extension | 
| Offline support | Works offline (bundled) | Works offline (replicated like extension libs) |
| UI composition | Preset displays with limited configurability through `app_settings.json`. Onus on CHT to update/extend functionality. Settings loaded at runtime to mutate navigation. Plugin code available on app load. | Open ended display. Onus is on AUTHOR to implement the visuals they're interested in. Opportunity for marketplace. Nav placement declared ahead of time via descriptor (`.properties`). Descriptor loaded at runtime to mutate navigation. Plugin code loaded as necessary. | 
| Load timing | All plugins loaded async on application startup, but still eagerly for all users. | JS loaded only when the extension is needed. |
| Framework constraints | Angular-hosted and Angular-coupled | Angular, vanilla JS, or anything that builds to single JS file. |
| Upgrade coupling | Tightly coupled to CHT Core releases | Extensions can evolve independently of core. Might need some sort of prop to indicate compatibility with core (similar to `engines` prop on `package.json`). |
| Architectural feel | Optional code paths shipped with core | Extensions whose existence and execution can be enforced at delivery time, not just runtime |

From my understanding, the navigation is ultimately mutated by both approaches. The POC favours a grouped nav config approach, whereas the squad proposal favours a per-extension approach. Both will need a doc attachment to be loaded at runtime to set the extension routes. 

Where the approaches meaningfully diverge is what happens after that point. The squad proposal seems way more author friendly and kinder to the client device in terms of resource consumption (upfront vs conditional). There's a box where the extension code is allowed to run in, it shouldn't have access things we don't intentionally expose. Also, seems safer to NOT have unnecessary code hanging around (conditional replication). 

Over and above the technical details, the approaches also differ on ownership. The POC places ongoing responsibility on the CHT team to curate, bundle, and ship plugin code. The squad proposal shifts ownership to the community, to implement/maintain their needs. Core is just a vessel.

jkuester:
:+1: I think your table comparison is spot-on and does a good job describing the benefits/downsides of each approach!

One additional factor that I think combines several of the topics listed above is the "Ease of consumption". The CHT app configuration has always been "low code". The framework is flexible enough that you need some JS to set up all your functionality and workflows, but you don't have to code anything "from scratch".  Your POC approach here stays closer to this paradigm where app devs can configure custom pages without needing to build the whole page themselves (from HTML/CSS/JS).  On the other hand, the web-component approach will require a lot more coding for building the custom page (with the added benefit that you can build pretty much whatever you want).

With the web-component approach, though, I think we might be able to achieve the "best of both worlds". :thinking: It is pretty easy to imaging a reusable web-component that could be _configured_ in similar ways as you have shown here. Not ever web-component UI Extension needs to be built from scratch, but instead you could have flexible extensions that can be configured...

ChinHairSaintClair:
Fair point. I think we can go ahead and close this PR. It's comforting to know that the POC contributed in a different way by furthering plugin/UI-extension conversations within the community. The code is also not lost, the `template-renderer` code will now be a starting point for our project's `ui-extensions`. And when refined, perhaps it can be made available to the wider community for use. Thank you, @jkuester!

PR #10236, opened 2025-08-21
chore(na): run API container as node not root user, lock down file perms, support stateless

### LINKED ISSUE #9720: Stateless Docker Container
At startup, the container modifies some files in `/service/api/build/static/webapp`.
On Kubernetes/OpenShift these folder are in read only mode ( stateless container approach).

A quick and dirty solution is to add this command in the image:

```
RUN chmod -R a+rwX /service/api/build/static/webapp
```

The image should be fully operational at startup without requiring a build phase.

This issue might be related to #[9141](https://github.com/medic/cht-core/issues/9141) and could potentially be resolved using a multi-stage build process.

mrjones-plip:
# Description

This PR takes the existing [Dockerfile](https://github.com/medic/cht-core/blob/master/api/Dockerfile) for API and runs it as `node` user, not `root` user.  This is a [security best practice](https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#non-root-user) that we should have already been doing.

Maybe we need more tests for this to explicitly check that the code is running as `node` user instead of `root`?  However, if all of CI passes, I think this is likely good enough.

Deferring to the PR review for pushback!

### support of stateless docker containers

There's an outstanding issue to support stateless docker containers.  This PR, thanks to an assist from  @lorerod , can now closes #9720 - woot!

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:run-api-as-node/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:run-api-as-node/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:run-api-as-node/docker-compose/cht-couchdb-clustered.yml)

# License


api/Dockerfile
```diff
@@ -10,10 +10,12 @@ RUN apk add --update --no-cache \
...
 
-COPY ./shared-libs ./shared-libs
-COPY node_modules/@medic ./node_modules/@medic
-COPY api ./api
+# ID 1000 is 'node' user in node:22-alpine image
 
-ENV NODE_PATH=/service/api/node_modules
+COPY --chown=root:root --chmod=755  ./shared-libs ./shared-libs
+COPY --chown=root:root --chmod=755  node_modules/@medic ./node_modules/@medic
+COPY --chown=root:root --chmod=755  api ./api
```
lorerod, line 17:
```suggestion
COPY --chown=root:root --chmod=755  api ./api
RUN chown -R root:1000 /service/api/build/static/webapp && \
    chmod -R 775 /service/api/build/static/webapp
```

The integration and e2e tests are failing. In the [ci-webdriver-default-enketo test logs](https://github.com/medic/cht-core/actions/runs/17138805347/job/48622613834?pr=10236#logs) I found this error:

`ERROR: EACCES: permission denied, open '/service/api/build/static/webapp/deploy-info.json'`

The API can’t write files because the required directories don’t have the necessary permissions.

This approach keeps root’s ability to write to the files while also giving the node user the access it needs for `/service/api/build/static/webapp/`.

Let me know what you think. I already ran `npm run wdio-local` and saw some tests passing. I didn’t run the complete suite, though. If you’re comfortable with this change, let’s check CI.

mrjones-plip: Thanks @lorerod !  Nice catch and nice fix for this.

I think in a perfect world, we'd refactor API to not write to this directory and use something like `/tmp` or `/medic_tmp` or something, but that's a heavier lift.  Another solution is to bind mount this directory - but a new bind mount I think would be a breaking change for k8s based deployments.

Given that this exact situation came up already, by using your approach, we'll also be able to close out #9720 with this ticket - win!!

lorerod: Thanks for the context! This fix may be a temporary solution to unblock tests and deployments while still improving security.Longer term, we should address the separation between build-time and runtime. It makes sense to track this in a follow-up ticket where we can explore a cleaner approach (e.g., using /tmp or a multi-stage build).

PR #10246, opened 2025-08-28
fix(#9835): fix missing `reported_date` bug

apoorvapendse:
# Description

 We autoconvert the update payload's `reported_date`
 to a unix epoch before doing the validation checks
 with the original document.
    
Fixes: https://github.com/medic/cht-core/pull/10083#issuecomment-3227044948.

# Code review checklist

# License

### DIFF
--- a/shared-libs/cht-datasource/src/local/libs/core.ts
+++ b/shared-libs/cht-datasource/src/local/libs/core.ts
@@ -1,4 +1,4 @@
-import { hasField, isRecord, Nullable } from '../../libs/core';
+import { convertToUnixTimestamp, hasField, isRecord, Nullable } from '../../libs/core';
 import { Doc } from '../../libs/doc';
 import { InvalidArgumentError } from '../../libs/error';
 
@@ -81,6 +81,9 @@ export const ensureHasRequiredFields = (
   updateInput: Record<string, unknown>,
 ): void => {
   const missingFieldsList = [];
+  if (updateInput.reported_date){
+    updateInput.reported_date = convertToUnixTimestamp(updateInput.reported_date as string | number);
+  }
   // ensure required immutable fields have the same value as the original doc.
   for (const field of [ ...immutableFields, ...mutableFields ]) {
     if (!hasField(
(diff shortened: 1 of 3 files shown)


apoorvapendse:
Tagging @sugat009 for review.

shared-libs/cht-datasource/src/local/libs/core.ts
```diff
@@ -81,6 +81,9 @@ export const ensureHasRequiredFields = (
   updateInput: Record<string, unknown>,
 ): void => {
   const missingFieldsList = [];
+  if (updateInput.reported_date){
+    updateInput.reported_date = convertToUnixTimestamp(updateInput.reported_date as string | number);
+  }
```
sugat009, line 73: issue: can you add some tests for this? since this function is imported from another file

apoorvapendse: Added tests in the second commit.

sugat009: Actually, I meant adding tests for `ensureHasRequiredFields` when the `if` case is `true`. Seems like the tests for `ensureHasRequiredFields` are left to do. However, when I run the coverage report. It has been covered. Looks like this `ensureHasRequiredFields` is not being mocked somewhere. 

The same for `dehydrateDoc` in the same file. 

Tests for `update` in `src/report.ts` is missing as well. 

Can you check the coverage report for `cht-datasource` as a whole and write tests for the functions that are missing?

apoorvapendse: Okay.
I remember avoiding mocking to test ensureHashRequiredFields and dehyrdateDoc in the tests for `person`, `place`, `report` module directly.

Example:
https://github.com/apoorvapendse/cht-core/blob/60970cc1b276c375a4d5c85cc5ac5d2a9f33cfb2/shared-libs/cht-datasource/test/local/person.spec.ts#L654

sugat009 (approved):

PR #10248, opened 2025-08-28
fix(#10133): update api startup form updates to not load all attachments

### LINKED ISSUE #10132: Update how API handles form updates on change

### LINKED ISSUE #10133: Update API startup form updates to not load all attachments with _all_docs
**Describe the issue**
Besides making unnecessary load on the server when there are very large media attachments for forms, CouchDb also has a long-lived bug where it times of if very large attachments are read through an _all_docs call: https://github.com/apache/couchdb/issues/2210

Relevant code: https://github.com/medic/cht-core/blob/989c9d67cfa95ff14fa1164400cfecc8b9dcf065/api/src/services/generate-xform.js#L276

Update this code to load attachments separately from the _all_docs request, and only load relevant attachments, instead of all media attachments. 

Still load all attachments (could be GBs of data!!), but switch the request type from _all_docs to _bulk_get (which might not time out).

nKataraia:
Updates how forms are loaded on startup and on change so that only the attachments with business-logic value (form xml, model xml and form html) are read and saved. 

closes #10133
closes #10132

### DIFF
--- a/api/src/services/forms.js
+++ b/api/src/services/forms.js
@@ -28,9 +28,9 @@ module.exports = {
     return name && doc._attachments[name];
   },
 
-  getAttachment: async (doc_id, name) => {
+  getAttachment: async (docId, name) => {
     try {
-      return await db.medic.getAttachment(doc_id, name);
+      return await db.medic.getAttachment(docId, name);
     } catch (error) {
       if (error.status === 404) {
         logger.error(error);
--- a/api/tests/mocha/controllers/forms.spec.js
+++ b/api/tests/mocha/controllers/forms.spec.js
@@ -10,6 +10,7 @@ const mockFormsInDb = (...docs) => {
   sinon.stub(db.medic, 'allDocs').resolves({
     rows: docs.map(doc => ({ doc: { ...doc, type: 'form' } })),
   });
+  sinon.stub(db.medic, 'getAttachment').resolves( 'foo');
 };
 
 const res = {
@@ -60,7 +61,6 @@ describe('forms controller', () => {
         _attachments: {
           xml: {
             content_type: 'xml',
-            data: 'foo',
           },
         },
       });
@@ -72,7 +72,7 @@ describe('forms controller', () => {
         chai
           .expect(writeHead.args[0][1])
           .to.deep.equal({ 'Content-Type': 'xml' });
-        chai.expect(end.args[0][0]).to.equal('foo');
+        chai.expect(end.args[0][0].toString()).to.equal('foo');
       });
     });
   });
--- a/api/tests/mocha/services/generate-xform.spec.js
+++ b/api/tests/mocha/services/generate-xform.spec.js
@@ -526,6 +526,7 @@ describe('generate-xform service', () => {
     it('should ignore collect forms', () => {
       sinon.stub(db.medic, 'allDocs').resolves({ rows: [ COLLECT_FORM_ROW ] });
       sinon.stub(db, 'saveDocs');
+      sinon.stub(db.medic, 'getAttachment');
       return service.updateAll().then(() => {
         expect(db.medic.allDocs.callCount).to.equal(1);
         expect(db.saveDocs.callCount).to.equal(0);


api/src/services/forms.js
```diff
@@ -28,8 +36,6 @@ module.exports = {
         startkey: FORM_DOC_ID_PREFIX,
         endkey: `${FORM_DOC_ID_PREFIX}\ufff0`,
         include_docs: true,
-        attachments: true,
-        binary: true,
```
dianabarsan, line 32: I suspect that this is causing some unit tests to fail.
Additionally, there are some forms endpoints that call this function and expect the attachment XML contents to be returned. (for example /api/v1/forms/<formid>). So you will need to load the 'xml' attachments separately for each form and send them as part of the response.

dianabarsan (changes requested):
Thanks for the updates @nKataraia ! I found a small problem detailed in the comment below.

api/src/services/forms.js
```diff
@@ -26,21 +28,48 @@ module.exports = {
     return name && doc._attachments[name];
   },
 
+  getAttachment: async (doc, name) => {
+    try {
+      return await db.medic.getAttachment(doc._id, name, { rev: doc._rev });
```
dianabarsan: The rev is not required here.

api/src/services/forms.js
```diff
@@ -26,21 +28,48 @@ module.exports = {
...
+
+      const xml = await module.exports.getAttachment(doc, name);
+      if (xml===null){
+        return null;
+      }
+
+      doc._attachments = doc._attachments || {};
+      doc._attachments[name] = {
+        data: Buffer.from(xml),
+        content_type: 'text/xml'
```
dianabarsan: I believe the content_type will already be set by the original request, so all you need is to update the `data` property. 

```suggestion
doc.attachments[name].data = Buffer.from(xml);
```

api/src/services/forms.js
```diff
@@ -26,21 +28,48 @@ module.exports = {
...
+      const doc = row.doc;
+      const name = module.exports.getXFormAttachmentName(doc);
+      if (doc.type !== DOC_TYPE || !name) {
+        return null;
+      }
+
+      const xml = await module.exports.getAttachment(doc, name);
+      if (xml===null){
+        return null;
+      }
```
dianabarsan: I think this can be simplified: 
```suggestion
if (!xml) {
  return;
}
```

api/src/services/forms.js
```diff
@@ -26,21 +28,48 @@ module.exports = {
...
+
+      doc._attachments = doc._attachments || {};
```
dianabarsan: I think that if this property is not set, we're totally not having any attachments for this document. So this check is not required.

dianabarsan (changes requested):
Nice! I added a few comments below. I've also noticed your unit tests are not passing, probably because some updates are required due to the latest changes.

api/src/services/forms.js
```diff
@@ -4,37 +4,67 @@
...
    */
   getXFormAttachment: doc => {
-    const name = Object.keys(doc && doc._attachments || {})
-      .find(name => name === 'xml' ||
-            (name.endsWith('.xml') && name !== 'model.xml'));
+    const name = module.exports.getXFormAttachmentName(doc);
     return name && doc._attachments[name];
   },
 
+  getAttachment: async (doc_id, name) => {
```
dianabarsan: please always use camelCase when defining variable names

dianabarsan (changes requested):
Unfortunately your unit tests are still not passing, they are failing with this error: 
```
2025-09-24T11:14:19.762 ERROR: medic.getAttachment() not stubbed!  UNIT_TEST_ENV=1. Please stub PouchDB functions that will be interacted with in unit tests. Error: medic.getAttachment() not stubbed!  UNIT_TEST_ENV=1. Please stub PouchDB functions that will be interacted with in unit tests.
    at notStubbed (/home/runner/work/cht-core/cht-core/api/src/db.js:2:1707)
    at module.exports.<computed>.<computed> [as getAttachment] (/home/runner/work/cht-core/cht-core/api/src/db.js:2:2210)
    at Object.getAttachment (/home/runner/work/cht-core/cht-core/api/src/services/forms.js:10:388)
    at /home/runner/work/cht-core/cht-core/api/src/services/forms.js:13:707
    at Array.map (<anonymous>)
    at Object.getFormDocs (/home/runner/work/cht-core/cht-core/api/src/services/forms.js:13:272)
```

Unit tests are really easy to run locally, you just need to run `npm run unit-api` and check the results!

dianabarsan:
Link to instructions about how to run unit tests https://docs.communityhealthtoolkit.org/community/contributing/code/core/automated-tests/

PR #10254, opened 2025-08-31
feat(shared-lib): create shared constants library for versions

Anithpavan:
### Summary
Created a shared constants library to store version information used across the project.

### Changes
- Added a new shared library `cht-shared-constants`
- Moved version constants into this library
- Updated imports across modules to use the shared constants

### Why
This centralizes version definitions, avoids duplication, and improves maintainability of the project.

### Notes
Tested locally with `npm run build` to confirm that all modules compile successfully without errors.


jkuester:
:thinking: Hmm, the changeset for this PR still does not look correct.  (lots of extra unrelated changes). For example, why did `puppeteer` get added to the dependencies in `package.json`?  

If you are not sure where these are all coming from, it might be best to just start fresh from a new branch off of `master` and just add in the changes you know you want on top of that and we can go from there...

Anithpavan:
Ok! i will go through it

PR #10264, opened 2025-09-05
chore(#10242): show base version in branches and betas on the admin upgrade page

### LINKED ISSUE #10242: [discuss] Create new release database for 5.x
**Describe the issue**
We should not show 4.x branches along with 5.x branches in the Admin Upgrade page. 

In order to separate 4.x from 5.x builds, create a `builds_5` release database, similar to what we did with 4.x and 3.x 

Keep 4.x and 5.x together but this might create a lot of confusion.

dianabarsan:
# Description

Adds a new column called "Base version" in the upgrade page in both betas and branches sections. 
Looks like this: 
image

closes #10242

# Code review checklist

# License


admin/src/templates/release.html
```diff
@@ -1,8 +1,11 @@
 <div class="col-xs-5">
     <span class="label label-info release-name" title="{{release | buildVersion}}">{{release | buildVersion}}</span>
```
dianabarsan, line 2: I chose not to edit this, so the branch name will still look like `12312-foo-bar (~ 5.0.0)` because our upgrade e2e test logic looks at this, and then it's a selector mess when older branches try to run this e2e test.

dianabarsan:
This is tiny tiny PR, @jkuester . At your discretion :)

api/resources/translations/messages-ar.properties
```diff
@@ -784,6 +784,7 @@ instance.upgrade.confirm.summary = أنت على وشك الترقية من {{be
 instance.upgrade.confirm.warning.text = لا يمكن التراجع عن هذا الإجراء. تأكد من نسخ بياناتك احتياطياً\!
 instance.upgrade.current_version = الإصدار الحالي
 instance.upgrade.date = تاريخ الإصدار
+instance.upgrade.base_version = النسخة الأساسية
```
jkuester, line 787: Minor, but how confident are you in these translation values?  Just wondering if we should follow the [translation process](https://docs.communityhealthtoolkit.org/community/contributing/translations/#adding-new-keys) here.... 🤔

admin/src/templates/release.html
jkuester: 🤔 Hmm, how much of a mess are we talking here?  I know the upgrade tests are a pretty delicate balance, so I hate to have us waste a bunch of time here.  However, just adding a new column on the page that _duplicates_ information already provided in the first column feels like a step backwards in terms of UX... 😬 

If removing the version from the title column is not feasible, then I think I would prefer we not make any of these changes and just keep the upgrade page as it is...

dianabarsan: Well, in the selector for the branch name, we intentionally include the parenthesis, but nothing after the parenthesis as far as I can tell. So we expect the branch name to looks something like `branch_name\s(.*`. 

https://github.com/medic/cht-core/blob/b9ddc064d82158b65cd801eb8488ccbab4838de0/tests/page-objects/upgrade/upgrade.wdio.page.js#L13

So a bit of a mess. 
The issue is that for long branch names the base version might be truncated :(

api/resources/translations/messages-ar.properties
dianabarsan: yea :D I just asked AI for translations. I highly doubt anyone is using the admin app in arabic, especially since it doesn't have RTL support.

admin/src/templates/release.html
jkuester: I guess I am still not following with what the trouble here is. (Sorry!)  Looking at that line of code, it seems like we could just drop the ` (` part from the selector and it would just work (or at least work as well as it did before with regards to the truncated branch names...).  Bonus points if we add an assertion for the new base version column.

dianabarsan: Ok, I _think_ this will work because the selector will be more permissive now. Updated!

jkuester:
@dianabarsan I see the PR title says `chore`, but I guess since this does affect something a user can see, it should be `feat`, right?

PR #10266, opened 2025-09-05
perf(#10262): update docs_by_replication_key to use nouveau index

### LINKED ISSUE #10262: Investigate moving docs_by_replication_key to nouveau
**Describe the performance issue**
Replication requires a call to a CouchDb view [`docs_by_replication_key`](https://github.com/medic/cht-core/tree/master/ddocs/medic-db/medic/views/docs_by_replication_key), often with a large payload of keys (depending on the number of contacts the user can see). These queries are the most costly part of replication, mainly because CouchDb is not very efficient in searching views by multiple keys. 

Investigate whether moving this view to nouveau improves performance. 

**To Reproduce**
call the `api/v1/replication/get-ids` on `master` vs the feature branch using a large user and compare results. 

**Measurements**
Check the results from above. 
Also run the scalability suite against both branches and compare results. 
Check storage to make sure we don't regress - would be the case when the nouveau view is larger than the couchdb view. 
Check how long it takes to index 10.000 docs in the new view.

dianabarsan:
# Description

Adds Nouveau index that would replace docs_by_replication_key view and updates API code to call this index instead of the view. 

closes #10262

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:test-replication-nouveau/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:test-replication-nouveau/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:test-replication-nouveau/docker-compose/cht-couchdb-clustered.yml)

# License


shared-libs/nouveau/src/index.js
```diff
@@ -0,0 +1,29 @@
...
+  String.raw`\~`,
+  String.raw`\*`,
+  String.raw`\?`,
+  ':',
+  String.raw`\-`
+];
+const charSet = specialChars.join('');
+const pattern = new RegExp(`([${charSet}])`, 'g');
+
+const escapeKeys = (key) => {
```
jkuester, line 23: Can you update [cht-datasource](https://github.com/medic/cht-core/blob/f1bdfc07c0f4b1170e0d8be671cd27aa105baff4/shared-libs/cht-datasource/src/local/libs/request-utils.ts#L56) to just re-use this escape logic instead of duplicating it?

ddocs/medic-db/medic/nouveau/docs_by_replication_key/default_analyzer
```diff
@@ -0,0 +1 @@
+whitespace
```
jkuester: Maybe I am missing something, but I am not sure why we need to use the `whitespace` analyzer for this index at all...? 🤔  In all the queries it seems like we are just looking for direct equality (which we get with `keyword`).  I think maybe we could just use `keyword` as the default analyzer and then define a separate field_analyzer for `key`.

shared-libs/view-map-utils/src/view-map-utils.js
jkuester, line 1: Tangential, but why is this a shared lib?  Seems to only be used in api.  (Don't need to make any changes, but was wondering if I was missing something here...)

ddocs/medic-db/medic/nouveau/docs_by_replication_key/index.js
```diff
@@ -8,9 +8,17 @@ function (doc) {
...
       doc.type === 'translations') {
-    return emit('_all', {});
+    index('string', 'key', '_all', { store: true });
+    return;
   }
-  var getSubject = function() {
+
+  var indexMaybe = (fieldName, value) => {
+    if (value) {
+      index('string', fieldName, value.toString(), { store: true });
```
jkuester: Since it looks like we only ever query this index with `key:...`, I wonder if we should be using the [`stored` field type](https://docs.couchdb.org/en/stable/ddocs/nouveau.html#field-types) for all of the non-key fields (`type`, `submitter`, etc)?  I assume that provides some performance/resource benefits over just using `string` for everything....  🤷

api/src/services/authorization.js
```diff
@@ -559,90 +585,120 @@ const isSensitive = (userCtx, subject, submitter, isPrivate, allowedSubmitter) =
...
+      }
+    });
+
+    if (!response.hits || !response.hits.length) {
+      continue;
+    }
+
+    hits.push(...response.hits.map(hit => ({
+      id: hit.id,
+      fields: hit.fields,
```
jkuester: Minor, but if we massaged `fields.key` here to always be an array, we would not have to have that logic repeated several places above...

ddocs/medic-db/medic/nouveau/docs_by_replication_key/index.js
jkuester: On a similar note, this is minor but it seems like if `key` is the main way we query this index, we could just use `default` instead and not have to include `key:` in the query string at all.  The main downside is that it looks like this means that in the returned hit you get `fields.default` instead of `fields.key` which is not great semantics....  I don't think I have a strong feeling either way (and I am pretty sure there is no performance difference one way or the other...).

sentinel/src/lib/purging.js
```diff
@@ -205,38 +206,39 @@ const assignContactToGroups = (row, groups, subjectIds) => {
...
 
-const isRelevantRecordEmission = (row, groups) => {
-  if (groups[row.id]) { // groups keys are contact ids, we already know everything about contacts
+const isRelevantRecordEmission = (hit, groups, subjectIds) => {
+  if (groups[hit.id]) { // groups keys are contact ids, we already know everything about contacts
     return false;
   }
 
-  if (row.value.type !== 'data_record') {
+  if (hit.fields.type !== 'data_record') {
```
jkuester: Can we just include `AND type:data_record` in the query so we only get back records?

jkuester: 🤔 Maybe the same goes for the `if (groups[hit.id])` check?  I don't see how that could ever be true if all we get here is reports...

jkuester (commented):
🚀 Very nice!  Have not made it to the integration tests yet, but wanted to drop the comments I had so far.  👍

shared-libs/view-map-utils/src/view-map-utils.js
dianabarsan: I believe this was used in webapp at some point. Indeed now it is not.

ddocs/medic-db/medic/nouveau/docs_by_replication_key/index.js
dianabarsan: >  I wonder if we should be using the [stored field type](https://docs.couchdb.org/en/stable/ddocs/nouveau.html#field-types) for all of the non-key fields (type, submitter, etc)?

I believe this enables the fields to be returned with the query result. It serves as the value from the couchdb function emit.

api/src/services/authorization.js
dianabarsan: I agree, but I haven't found the best way to get that data to be consistent.

sentinel/src/lib/purging.js
dianabarsan: Oh, cool idea, I'll try this one!

shared-libs/nouveau/src/index.js
dianabarsan: Yep, didn't know about this one :)

ddocs/medic-db/medic/nouveau/docs_by_replication_key/index.js
jkuester: No, I am not talking about the `{ store: true }` option.  (Which I agree functions as you said.)  I am talking about using `'stored'` as the first parameter instead of `'string'`. 

```js
index('stored', fieldName, value.toString());
```

I tested this and it seems to function as I expected.  When you index with the `stored` type, then the `{ store: true }` option _is implied_ and the field is returned in the `fields` object for the hit as desired.  However, you cannot _query_ against the values indexed with `stored`.  So, if I do `index('stored', 'needs_signoff', 'true');`, then I still get the `fields.needs_signoff` back on the hit, but I cannot query with `q: 'needs_signoff:true'` (which is fine since we never actually query with anything but `key`).

jkuester: I assume this provides some performance/resource benefits over `'string'` because it is not actually indexing the `'stored'` values....

dianabarsan: this is done!

ddocs/medic-db/medic/nouveau/docs_by_replication_key/index.js
```diff
@@ -12,17 +12,27 @@ function(doc) {
     return;
   }
 
-  var indexMaybe = (fieldName, value) => {
-    if (value) {
-      index('string', fieldName, value.toString(), { store: true });
+  const indexableFields = ['key', 'type', 'subject'];
```
jkuester, line 15: Why is `subject` an indexable field?  Maybe I missed it, but I don't see it used in a query anywhere.  Is this maybe just for possible future use?

jkuester (approved):
LGTM!  One minor question, but otherwise I think this is ready!  🚀

ddocs/medic-db/medic/nouveau/docs_by_replication_key/index.js
dianabarsan: Oh missed this. i was testing things. I'll revert!

dianabarsan:
@jkuester I found an issue where I wasn't correctly batching nouveau requests while purging, and pushed a fix for it. Would you mind giving it a look? Thanks!

dianabarsan:
@jkuester i have pushed the changes we spoke about yesterday. Would you mind a quick look? Thanks!

jkuester (approved):
Nice refactoring! LGTM!

Issue #8220, opened 2023-05-05
CI "Publish for testing" is failing


mrjones-plip:
**Describe the issue**
Since yesterday (May 4th 6pm Pacific), none of the "Publish for testing" CI has passed.  You can see all the [failed tests here](https://github.com/medic/cht-core/actions/workflows/build.yml). Following the CI down to the to the line that errors, I _believe_ (but I'm not 100% sure!)  it's:
1. in `.workflow` we [call `build.yml`](https://github.com/medic/cht-core/blob/master/.github/workflows/build.yml)
1. in `build.yml` we call `node grunt build-service-images` in the ["Publish For Testing" section](https://github.com/medic/cht-core/blob/master/.github/workflows/build.yml#L64)
1. In the `Gruntfile.js` it's failing in the `build-service-images` [call here](https://github.com/medic/cht-core/blob/58f23a8db0651faeb8c9fdeb7a00a288e05950cf/Gruntfile.js#L296).

The [full error is here](https://github.com/medic/cht-core/actions/runs/4895891000/jobs/8745289588) and the output is:

```
 Running "exec:build-service-images" (exec) task
>> /bin/sh: 1: Syntax error: "(" unexpected
>> Exited with code: 2.
>> Error executing child process: Error: Process exited with code 2.
Warning: Task "exec:build-service-images" failed. Use --force to continue.

Aborted due to warnings.
```

I discovered this when I was trying to get PR #8219 to pass.

**Describe the improvement you'd like**
CI should pass

**Describe alternatives you've considered**
Retry the build again (but it's failed 4 times the exact same way.

mrjones-plip:
UPDATE: This is likely not broken for all branches, but just for the one with `(` in it (`fix(8213)-contact-new-not-scroll`).  We my choose to use this ticket to defend against wonky named branches :sweat_smile:

garethbowen:
I guess just escape the name somehow.

garethbowen:
Related: #8901

jarias:
@dianabarsan this looks like an stale issue, looks like `grunt` is no longer used. Seems like we can close this one out.

jkuester:
@jarias thanks for bringing this up!  You are right that `grunt` is no longer used, but I did a quick test and confirmed that the npm script that replaced the grunt script (`npm run local-images`) still fails if you have special characters in your branch name.  

[...]

PR #10268, opened 2025-09-05
fix(8220): correctly escape branch names

jarias:
# Description

When creating a docker image the branch name is use as the tag, docker only supports ASCII dash and underscore as valid character names. This change escapes the branch name according to this rules.

Closes: #8220 

# License


jkuester:
Also, really happy to confirm that this fixes a long-running issue I have had in WebStorm. When reviewing a PR in WebStorm, there is a nice button to use for directly checking out the branch you are reviewing. However, WebStorm calls the local branch `fork/<user>/<branch_name>` and previously the `/`s in the branch name would cause this same issue.  Now everything works as expected and I see the images built with the tag: `fork-jarias-fix-8220--docker-image-build-fails`.  :heart:

jkuester:
:sweat_smile: flaky tests was not really flaking... :sweat_smile: :facepalm: Had to push a small fix to match the escape logic in the test code.

PR #10269, opened 2025-09-05
chore: uplift pyxform/cht-conf

jkuester:
… version

# Description

# Code review checklist

# License

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -180,8 +180,7 @@ jobs:
         if: ${{ !env.INTERNAL_CONTRIBUTOR }}
 
       - run: mkdir tests/logs
-      - run: python -m pip install git+https://github.com/medic/pyxform.git@medic-conf-1.17#egg=pyxform-medic
-      - run: npm install -g cht-conf
+      - run: npm install -g git+https://github.com/medic/cht-conf.git#eb9e6cd11efe38dc6d3281247787d9fb7c304d30
       - run: npm ci
 
       - uses: nolar/setup-k3d-k3s@v1
@@ -298,10 +297,8 @@ jobs:
 
     - name: Create logs directory
       run: mkdir tests/logs
-    - name: Install pyxform
-      run:  python -m pip install git+https://github.com/medic/pyxform.git@medic-conf-1.17#egg=pyxform-medic
     - name: Install cht-conf
-      run:  npm install -g cht-conf
+      run:  npm install -g git+https://github.com/medic/cht-conf.git#eb9e6cd11efe38dc6d3281247787d9fb7c304d30
     - run: npm ci
     - name: Fetch allure history
       uses: actions/checkout@v4
--- a/package.json
+++ b/package.json
@@ -119,7 +119,7 @@
     "chai-like": "^1.1.3",
     "chai-shallow-deep-equal": "^1.4.6",
     "chai-things": "^0.2.0",
-    "cht-conf": "^5.0.0",
+    "cht-conf": "git+https://github.com/medic/cht-conf.git#eb9e6cd11efe38dc6d3281247787d9fb7c304d30",
     "cht-conf-test-harness": "^5.0.3",
     "clean-css-cli": "^5.6.2",
     "couchdb-compile": "^1.11.2",
--- a/.github/actions/deploy-conf/action.yml
+++ b/.github/actions/deploy-conf/action.yml
@@ -23,8 +23,6 @@ runs:
   - name: install dependencies
     run: |
         npm ci
-        pip install --user -U pip
-        sudo python -m pip install git+https://github.com/medic/pyxform.git@medic-conf-1.17#egg=pyxform-medic
     shell: bash
     working-directory: ${{ inputs.directory }}
   - name: run cht
--- a/.github/actions/deploy-with-medic-conf/action.yml
+++ b/.github/actions/deploy-with-medic-conf/action.yml
@@ -22,8 +22,6 @@ runs:
   - name: install dependencies
     run: |
         npm ci
-        pip install --user -U pip
-        sudo python -m pip install git+https://github.com/medic/pyxform.git@medic-conf-1.17#egg=pyxform-medic
     shell: bash
     working-directory: ${{ inputs.directory }}
   - name: run medic-conf 
--- a/package-lock.json
+++ b/package-lock.json
@@ -110,7 +110,7 @@
         "chai-like": "^1.1.3",
         "chai-shallow-deep-equal": "^1.4.6",
         "chai-things": "^0.2.0",
-        "cht-conf": "^5.0.0",
+        "cht-conf": "git+https://github.com/medic/cht-conf.git#eb9e6cd11efe38dc6d3281247787d9fb7c304d30",
         "cht-conf-test-harness": "^5.0.3",
         "clean-css-cli": "^5.6.2",
         "couchdb-compile": "^1.11.2",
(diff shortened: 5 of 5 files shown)


config/default/forms/app/delivery.xml
```diff
@@ -1322,988 +1089,787 @@
...
+      <bind nodeset="/data/babys_condition/baby_repeat/baby_details/baby_profile/danger_signs/fever" type="string" calculate="../../../fever"/>
+      <bind nodeset="/data/babys_condition/baby_repeat/baby_details/baby_profile/danger_signs/blue_skin" type="string" calculate="../../../blue_skin"/>
+      <bind nodeset="/data/babys_condition/baby_repeat/baby_details/baby_profile/parent/_id" type="string" calculate=" /data/household_id "/>
+      <bind nodeset="/data/babys_condition/baby_repeat/baby_details/baby_profile/parent/parent/_id" type="string" calculate=" /data/area_id "/>
+      <bind nodeset="/data/babys_condition/baby_repeat/baby_details/baby_profile/parent/parent/parent/_id" type="string" calculate=" /data/facility_id "/>
+      <bind nodeset="/data/babys_condition/baby_repeat/baby_details/baby_profile/type" type="string" calculate="&quot;person&quot;"/>
+      <bind nodeset="/data/babys_condition/baby_repeat/baby_details/baby_profile/created_by_doc" type="string" calculate="."/>
+      <bind nodeset="/data/babys_condition/baby_repeat/baby_details/child_doc" type="string" calculate=" ../baby_profile "/>
```
jkuester, line 1235: This is the source of the change to the `delivery.xlsx` file.  The `child_doc` calculate in the xslx originally had `${baby_details}` as the calculation value. I updated the calculation to be `${baby_profile}`.  The actual value of the calculation does not really matter since this field has an `instance::db-doc-ref` value of `${baby_profile}`. Whatever value here is overridden when saving the form anyway. The example [in our docs](https://docs.communityhealthtoolkit.org/building/forms/configuring/additional-docs/#linked-docs-example) shows using a self-reference (`.`) in the calculation  for `db-doc-ref` fields, but all the other `db-doc-ref` fields in this form besides this one just had their calculation set to the same thing as their `db-doc-ref`. So, I stuck to that pattern here. 

The reason I had to make this change is because, with the new pyxform, this calculate value was getting generated as a _relative path_ instead of an _absolute path_.  Originally the calculate was set to `/delivery/babys_condition/baby_repeat/baby_details`, but the new pyxform generated it to be `../baby_details`.  I did some tinkering and confirmed the pyxform logic seems to use relative paths _when inside a repeat_. However, you will notice that this actually seems to be a pyxform _bug_ to get the `../baby_details` path because `baby_details` is the _parent_ of `child_doc` and not a _sibling_. (So really the relative path would be `../../baby_details`).  With some more testing I was able to repro this behavior only when I had a calculate in a group in a repeat and the calculate referenced the group in the calculation expression.  I think this is just an "edge case" that pyxform is not properly handling.  I have not brought myself to log this as a bug in pyxform yet, though, because the workflow to reproduce the issue seems just too broken.... 😅 

---

Ultimately, I think the new behavior from cht-conf/pyxform is acceptable because:

- Really you should not be referencing _groups_ in your calculation expression (because it is meaningless in ODK to set a field equal to a group).
- You do get a validation error when trying to validate/upload the form using cht-conf. So, this will not just fail silently in prod.

config/default/forms/app/pregnancy.xml
```diff
@@ -961,998 +842,940 @@
...
+      <bind nodeset="/data/weeks_since_lmp_rounded" type="string" relevant="../lmp_date_8601 != ''" calculate="floor(../days_since_lmp div 7)"/>
+      <bind nodeset="/data/lmp_method_approx" type="string" calculate="if(selected(../gestational_age/register_method/lmp_method, 'method_approx'), 'yes', 'no')"/>
+      <bind nodeset="/data/hiv_status_known" type="string" calculate="../safe_pregnancy_practices/hiv_status/hiv_tested"/>
+      <bind nodeset="/data/deworming_med_received" type="string" calculate="../safe_pregnancy_practices/deworming/deworming_med"/>
+      <bind nodeset="/data/tt_received" type="string" calculate="../safe_pregnancy_practices/tetanus/tt_imm_received"/>
+      <bind nodeset="/data/t_pregnancy_follow_up_date" type="string" calculate="../anc_visits_hf/anc_visits_hf_next/anc_next_visit_date/appointment_date"/>
+      <bind nodeset="/data/t_pregnancy_follow_up" type="string" calculate="../anc_visits_hf/anc_visits_hf_next/anc_next_visit_date/appointment_date_known"/>
+      <bind nodeset="/data/t_danger_signs_referral_follow_up_date" type="string" calculate="format-date-time(decimal-date-time(today()) + 3, &quot;%Y-%m-%d&quot;)"/>
```
jkuester, line 1027: This is the source of the change to the `pregnancy.xlsx` file.  I think somehow a previous update to that file did not get committed properly (or got lost in a merge conflict) because the original xml did not actually match the original xlsx for this `calculate`.  Basically the xml was correctly using the `format-date-time` function in this calculate (as you see here), but the xlsx instead had the `date-time(decimal-date-time(today()) + 3)` (which also includes the _time_). 

So, I have just updated the `pregnancy.xlsx` to match what was already in the xml.

config/default/forms/app/delivery.xml
dianabarsan: Confirming that this is the error I get when I try to upload the old form converted with new pyxform: 
```
INFO convert-app-forms complete. 
INFO Starting action: validate-app-forms… 
INFO Validating form: death_report.xml… 
INFO Validating form: delivery.xml… 
ERROR Form at /home/diana/projects/medic/config/default/forms/app/delivery.xml contains invalid XPath expressions (absolute or relative paths that refer to a non-existant node): 
ERROR   - calculate for /data/babys_condition/baby_repeat/baby_details/child_doc contains [../baby_details] 
```

tests/e2e/default/contacts/forms/ngo-create.xlsx
jkuester, line 1: I originally was considering adding new xlsx files for all the test forms we have that are missing them.  Ultimately decided against that (due to time constraints), but left this one as a testament to my ambitions.... 😅 

Otherwise, I think I did manage to re-convert all the test forms where we did have an xlsx file. 👍

.github/workflows/build.yml
```diff
@@ -187,8 +187,7 @@ jobs:
         if: ${{ !env.INTERNAL_CONTRIBUTOR }}
 
       - run: mkdir tests/logs
-      - run: python -m pip install git+https://github.com/medic/pyxform.git@medic-conf-1.17#egg=pyxform-medic
-      - run: npm install -g cht-conf
+      - run: npm install -g github:medic/cht-conf#uplift-pyxform
```
jkuester: Update this once cht-conf is released.

.github/workflows/build.yml
```diff
@@ -308,10 +307,8 @@ jobs:
 
     - name: Create logs directory
       run: mkdir tests/logs
-    - name: Install pyxform
-      run:  python -m pip install git+https://github.com/medic/pyxform.git@medic-conf-1.17#egg=pyxform-medic
     - name: Install cht-conf
-      run:  npm install -g cht-conf
+      run:  npm install -g github:medic/cht-conf#uplift-pyxform
```
jkuester: Update this once cht-conf is released.

package.json
```diff
@@ -125,7 +125,7 @@
     "chai-like": "^1.1.3",
     "chai-shallow-deep-equal": "^1.4.6",
     "chai-things": "^0.2.0",
-    "cht-conf": "^5.0.0",
+    "cht-conf": "github:medic/cht-conf#uplift-pyxform",
```
jkuester: Update this once cht-conf is released.

tests/page-objects/default/enketo/enketo-widgets.wdio.page.js
```diff
@@ -1,23 +1,23 @@
 const FORM = 'form[data-form-id="enketo_widgets_test"]';
 
 const selectMultipleDropdown = (formId = FORM) => {
-  return $(`${formId} select[name="/enketo_widgets_test/enketo_test_select/select_spinner"]`);
+  return $(`${formId} select[name="/data/enketo_test_select/select_spinner"]`);
```
dianabarsan, line 4: this certainly makes testing easier!

Issue #8027, opened 2023-01-16
Admin app fails to update `partners` if existing document is malformed


dianabarsan:
**Describe the bug**
A malformed `partners` doc will block the user from updating partners from the admin app. 

**To Reproduce**
1. Create a `partners.json` document in your config. Set the contents of this document to be `{}`. run `cht upload-partners`. 
2. Login to admin app. Go to Images -> Partners.
3. Add a new "partner".
4. Click Submit. 
5. See error `There was an error saving your settings, please try again.`. 

**Expected behavior**
The UI should inform the user that the existing document is malformed. 
Even when the document is malformed, the user should have the possibility to "fix" it by editing from the Admin UI. 

**Logs**
When loading the partners page:
```
angular.js:15697 TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at resource-icons.js:111:24
    at processQueue (angular.js:18075:1)
    at angular.js:18123:1
    at Scope.$digest (angular.js:19242:1)
    at angular.js:19562:1
    at TaskTracker.completeTask (angular.js:21403:1)
    at angular.js:6879:1 'Possibly unhandled rejection: {}'
```

When saving: 
```
angular.js:15697 Error uploading image TypeError: Cannot set properties of undefined (setting 'partner-name')
    at images-partners.js:54:27
    at processQueue (angular.js:18075:1)
    at angular.js:18123:1
    at Scope.$digest (angular.js:19242:1)
    at angular.js:19562:1
    at TaskTracker.completeTask (angular.js:21403:1)
    at angular.js:6879:1
```

**Environment**
- Instance: any
- App: admin
- Version: 3.4.0+

**Additional context**
[Form post](https://forum.communityhealthtoolkit.org/t/branding-not-working-using-admin-interface/2399/21)

billoserra:
I am interested in taking this ticket. If it is available, could you please assign it to me?

mrjones-plip:
Note - I added all the "Affects" labels, but we don't have them going all the back to `3.4`.  I'm not adding more labels though.

mrjones-plip:
@billoserra - thanks again for your contribution on this!

I'm curating a list of presenters for next weeks Dev Demo Hour - are you available to do a ~10 min presentation about what it's like to work on a "Good first issue" and this ticket specifically?  We really like to have folks from the community present and I thought of you when I added the `Affects:` labels!  

[...]

PR #10278, opened 2025-09-10
fix(#8027): fix malformed partners doc

billoserra:
# Description

When a malformed partners document (content: `{}`) is uploaded via the cht tool, both the Admin app and Webapp
  throw fatal errors that completely block user functionality. The applications assumed the `resources` property
  existed when calling `Object.keys(res.resources)` and when setting partner properties, causing crashes that
  prevented users from accessing or fixing the malformed document.

  **Errors resolved:**
  - Admin app: `TypeError: Cannot set properties of undefined (setting 'partner-name')` when saving partners
  - Admin app: `TypeError: Cannot convert undefined or null to object` when loading partners page
  - Webapp: `TypeError: Cannot convert undefined or null to object` when accessing About page

  **Solution:**
  Added null checks and initialization of empty `resources` object when missing. This allows the UI to remain
  functional, enabling users to view and edit the partners list to correct malformed documents instead of
  experiencing fatal crashes.

  **Changes made:**
  - **Admin controller** (`images-partners.js`): Added null checks to initialize `resources` during document loading
   and attachment operations
  - **Admin service** (`resource-icons.js`): Added defensive null check in `getDocResources` method to return empty
  array
  - **Webapp service** (`resource-icons.service.ts`): Added defensive null check in `getDocResources` method to
  return empty array
  - **Tests**: Added comprehensive unit tests covering all malformed document scenarios, including new controller
  test file

 Closes: https://github.com/medic/cht-core/issues/8027

# Code review checklist

# License


jkuester:
@witash would you be able to review this PR?  :pray:

admin/src/js/controllers/images-partners.js
```diff
@@ -35,6 +35,9 @@ angular.module('controllers').controller('ImagesPartnersCtrl',
     
     getPartnersDoc().then(doc => {
       $scope.doc = doc;
+      if (!$scope.doc.resources) {
+        $scope.doc.resources = {};
```
witash: Is this necessary if the changes below catch or prevent the errors?
If possible it would be preferable to catch null and undefined errors when they occur instead of modifying the doc.

admin/src/js/controllers/images-partners.js
```diff
@@ -57,6 +60,9 @@ angular.module('controllers').controller('ImagesPartnersCtrl',
         .putAttachment(PARTNER_ID, file.name, $scope.doc._rev, file, file.type)
         .then(getPartnersDoc)
         .then(doc => {
+          if (!doc.resources) {
+            doc.resources = {};
+          }
```
witash: you can do this in one line with 
```
doc.resources = doc.resources || {};
```
but that's just a style choice, can leave it as is if you prefer

admin/src/js/services/resource-icons.js
```diff
@@ -110,7 +110,9 @@ angular.module('inboxServices').factory('ResourceIcons',
         return getHtml(name, docId, faPlaceholder);
       },
       getDocResources: doc => {
-        return DB().get(doc).then(res => Object.keys(res.resources));
+        return DB().get(doc).then(res => {
+          return Object.keys((res && res.resources) ? res.resources : {});
+        });
```
witash, line 115: similar to above, you can do the same thing with
```
return DB().get(doc).then(res => Object.keys(res?.resources || {}));
```

webapp/tests/karma/ts/services/resource-icon.service.spec.ts
```diff
@@ -453,5 +453,35 @@ describe('ResourceIcons service', () => {
...
+
+    it('should not fail when doc has undefined resources', () => {
+      get.resolves({ resources: undefined });
+      const service = getService();
+      return service.getDocResources('partners').then(result => {
+        expect(result).to.deep.equal([]);
+        expect(get.callCount).to.equal(4);
+        expect(get.args[3]).to.deep.equal(['partners']);
+      });
+    });
```
witash, line 476: Maybe we could have just one test for malformed docs?
Just to not add too many tests just for this one issue since they are already duplicated.

billoserra: Consolidated the three tests into one comprehensive test, thank you.

admin/src/js/services/resource-icons.js
billoserra: Thanks for the suggestion! Unfortunately, this causes a build error in the admin service:

`Error: Unexpected token (113:57) while parsing file: admin/src/js/services/resource-icons.js`

The admin service uses browserify which doesn't support optional chaining syntax `(?.).` The current ES5-compatible syntax` (res && res.resources) ? res.resources : {}` achieves the same defensive programming result while maintaining build compatibility.

The webapp service successfully uses the modern `res?.resources ?? {}` syntax since it has a different build system (Angular CLI) that supports it.

admin/src/js/controllers/images-partners.js
billoserra: Applied one-liner style improvements, thank you.

admin/src/js/controllers/images-partners.js
billoserra: Removed unnecessary resource initialization as suggested, thank you.

PR #10290, opened 2025-09-13
feat(#10208): added training-cards to instruct admin at first time setup to add data

### LINKED ISSUE #8147: Training cards display loading error when user does not have contact
**Describe the bug**
An Enketo form requires the user to have an existent associated contact. If this contact is missing, the form cannot be loaded. 
In the case of training cards, the training card modal appears, and it only shows the "Error loading form" message.

**To Reproduce**
1. Create a training card.
2. Create a user without an associated contact. 
3. Login as that user.

**Expected behavior**
Couple of options:
1. don't display training card
2. make an exception for displaying training cards for contactless users

**Screenshots**
image

**Environment**
- App: webapp
- Version: 4.2.0

### LINKED ISSUE #10208: Instruct new deployments how to add data
When you do a clean install of the CHT and you log in for the first time as the administrative `medic` user, there is no data (messages, tasks, reports, contacts or reports).  For a first time user this can be confusing.  

We should consider a prompt for how to add data or get started using the app. This should be for admin accounts targeting implementers, not offline CHW users.

NA

**Additional context**
There was a Dev Demo Hour presentation on how to go through the [Happy Path Dev Setup](http://localhost:1313/community/contributing/code/core/dev-environment/#the-happy-path-installation) and presenter thought their CHT instance was broken because it had no data.

This is what I see when I run the docker helper and click on "People":

Image

shashi-sah2003:
This PR add a welcome guide to the admin user during the first time setup of CHT locally on how to populate CHT webapp with data and also contains fixes for when user doesn't have contact. 

fixes #10208 
fixes #8147 

# License


shashi-sah2003:
@jkuester I think this is done, you can have a look at it once and let me know if there is any changes you want me to do

webapp/src/ts/services/form.service.ts
jkuester: I don't think we need to duplicate this logic here now that it is up in `completeReport`.  Right?

```suggestion
```

jkuester (changes requested):
This looks great @shashi-sah2003!  Can you add two new unit tests in `form.service.spec.ts`?  

- One to [assert `requiresContact` returns `false` for training forms](https://github.com/medic/cht-core/blob/69018ea93ca5ce5bead925afa81852e595e7f188/webapp/tests/karma/ts/services/form.service.spec.ts#L1813).
- One to verify we can [save a training form](https://github.com/medic/cht-core/blob/69018ea93ca5ce5bead925afa81852e595e7f188/webapp/tests/karma/ts/services/form.service.spec.ts#L723) when `UserContact` resolves `null`.

jkuester:
@mrjones-plip can I get a bit of QA from you on the modified look/feel?  :pray: 

I shrunk down the images a lot (though image display/scaling in Enketo is... interesting.)  I also cut out a bunch of text from the copy. Happy to continue iterating here!  :+1: 

image
image
image

mrjones-plip:
yeah - this is an improvement over the [earlier screenshots I saw](https://github.com/medic/cht-core/issues/10208#issuecomment-3276569306) - thank you!

It would be great to test them in a mobile - while not a frequent pattern for admins to login with their `medic` user there, I suspect it happens more than we think, especially with first time users (likely more in Chrome than an APK).

If you give me steps to reproduce - I'm happy to test!

jkuester:
Oh, yeah, I had tested it in the mobile form factor, just did not add the screen shots:

image
image
image

---
Of course, for both desktop and mobile, depending on your screen size/resolution, you might get some scrolling, but in general this seems to be pretty solid.

---

Also, if you do want to repro:

- Check out the branch from this PR and build/run the local instance:  `npm ci && npm run build-dev && npm run dev-api`
- Make sure to upload the training form: `cd config/default && cht --local upload-training-forms -- admin_welcome`
- Open the app in the browser and login as your admin user.  Should see the training pop right up.

mrjones-plip:
ah - sweet!  Thanks a bunch for all the screenshots in mobile and steps to reproduce.   All looks good!

I ran out out of time today to do any testing.

webapp/src/ts/services/form.service.ts
shashi-sah2003: yes @jkuester you are right, made a commit to remove this from `save`.

shashi-sah2003:
@jkuester I have added both the test cases and tested it locally, now we are gtg

PR #10291, opened 2025-09-13
chore(#10023): update release-notes script to include co-authors

### LINKED ISSUE #10023: Update `release-notes` script to include all commit co-authors in contributors list

HsingFuC:
# Description
Change the graphql query from `getCommitsForRelease` function to get the list of co-authors by commit instead of just the actual author of the commit, and update the `formatCommits` function to support the list of authors. Part of this change was using a Set instead of an array because the Set is more efficient when we need to check for duplicates.

Closes #10023 

# Code review checklist

# License

### DIFF
--- a/scripts/release-notes/index.js
+++ b/scripts/release-notes/index.js
@@ -274,17 +274,15 @@ const formatGroups = (groups) => {
 const formatCommits = (commits) => {
   const ignoreLogins = new Set(BOTS);
   const lines = [];
-  for (const commit of commits) {
-    const authors = commit.authors.nodes;
-    for (const author of authors) {
-      const { login, name, url } = author.user;
-      if (login && !ignoreLogins.has(login)) {
-        ignoreLogins.add(login);
-        const profileUrl = url;
-        lines.push(`- [${name || login}](${profileUrl})`);
-      }
+  const authors = commits.flatMap((commit) => commit.authors.nodes);
+  authors.forEach((author) => {
+    const { login, name, url } = author.user;
+    if (login && !ignoreLogins.has(login)) {
+      ignoreLogins.add(login);
+      const profileUrl = url;
+      lines.push(`- [${name || login}](${profileUrl})`);
     }
-  }
+  });
   return lines.join('\n');
 };
 
--- a/scripts/release-notes/index.js
+++ b/scripts/release-notes/index.js
@@ -101,7 +101,15 @@ const getCommitsForRelease = async (release, milestoneBranch) => queryRepoPagina
           nodes {
             oid
             messageHeadline
-            author { user { login, name, url } }
+            authors(first: 25) {
+              nodes {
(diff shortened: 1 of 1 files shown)


scripts/release-notes/index.js
```diff
@@ -264,16 +272,17 @@ const formatGroups = (groups) => {
...
   for (const commit of commits) {
-    const login = commit.author?.user?.login;
-    if (login && !ignoreLogins.includes(login)) {
-      ignoreLogins.push(login);
-      const user = commit.author.user;
-      const name = user.name || user.login;
-      const profileUrl = user.url;
-      lines.push(`- [${name}](${profileUrl})`);
+    const authors = commit.authors.nodes;
+    for (const author of authors) {
```
jkuester: I tested this out and everything works great!  Sonar [is complaining](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=10291&id=medic_cht-core&open=AZlFUf7l8rE9WlDYxSmY) about the cognitive complexity of having all this nested for-loop logic in the same function. 

Could you extract this inner for-loop into its own function?  Alternatively, it looks like we are not doing anything async in these for-loops, so you could also probably simplify all of this down into a couple of `map` calls. 🤔

HsingFuC: Switched from `for` loops to array methods, which was enough to pass SonarCloud analysis

jkuester (approved):
LGTM!  🚀

PR #10297, opened 2025-09-16
chore(#10294): correct branch escape logic for valid SemVer

### LINKED ISSUE #10294: Branch name escape logic broken for semver
**Describe the issue**
We recently [updated our branch name escape logic](https://github.com/medic/cht-core/commit/61dc6b) to be compatible with Docker image tag syntax.  Unfortunately, we expanded the scope of the our naming schema to include `_`, which is valid for Docker image tags, but _is not valid semver_.  

I think we also use this value as "version" reported by the api server on development instances. Now we have [tests failing](https://github.com/medic/cht-core/actions/runs/17690892325/job/50284597397?pr=10284) when the git branch name contains `_`.  

>   1) routing
>         unauthenticated routing
>           should display deploy-info to authenticated users:
>       AssertionError: "5.0.0-dependabot-npm_and_yarn-eslint-9.35.0.17690892325-1757733124020" is not valid semver: expected null to be truthy
>        at /home/runner/work/cht-core/cht-core/tests/integration/api/routing.spec.js:179:16
>        at process.processTicksAndRejections (node:internal/process/task_queues:105:5)

This is causing problems with the dependabot PRs because they often use branch names like: `dependabot/npm_and_yarn/eslint-9.35.0`.

I think we simply need to make another update to our [branch escape function](https://github.com/medic/cht-core/blob/61dc6b44f96bb0e90f28a51f3c7c817812429574/scripts/build/versions.js#L13) so the regex is `[^A-Za-z0-9.-]`.  This will cause all the `_` in the branch name to be replaced with `-`.  

To test the change:
- Checkout a branch with a name containing `_`.
- Update the [rounting.spec.js file](https://github.com/medic/cht-core/blob/58a5e7d901c5946c4cdebd23611bf13db1c35b3f/tests/integration/api/routing.spec.js#L72) and set `describe.only` so only the tests in this file get executed.
- Try running the test test to make sure you can reproduce the failure: `npm run integration-all-local`.  
- Update the branch escape function as suggested above.
- Re-run the tests and confirm they all pass.

SinghCod3r:
Closes #10294

### Description

The branch name escape logic was recently updated to be compatible with Docker image tags, which allows the `_` character. However, this output is also used as a SemVer pre-release version for development instances, and SemVer does not permit underscores.

This caused integration tests to fail with an `invalid semver` error on branches with names containing underscores, such as those automatically created by Dependabot (e.g., `dependabot/npm_and_yarn/...`).

### Implementation

This change updates the regular expression in the branch escape function to replace any character that is not a letter, number, dot, or hyphen with a hyphen (`-`).

Specifically, the regex was updated to `[^A-Za-z0-9.-]`, which now correctly converts underscores to hyphens, ensuring the generated version string is always SemVer compliant.

### Testing

I have successfully verified this fix locally by following these steps:
1.  Created a branch with an underscore in its name (`fix/semver_underscore_issue`) to reproduce the test failure.
2.  Applied the code change.
3.  Ran the full `npm run integration-all-local` test suite and confirmed that all tests passed, resolving the original assertion error.


jkuester (changes requested):
The change here looks good! I almost forgot that we need to make a similar update [in the test util code](https://github.com/medic/cht-core/blob/b907fd210389e839b6c86acf6c1d92fda42a31c5/tests/utils/index.js#L1652).  If you could update that line too, then hopefully all the tests will be happy!  🤞 

Also, @SinghCod3r there are some unrelated changes that we don't need in this PR. (I think the only files that should be changed are `scripts/build/versions.js` and `tests/utils/index.js`. The rest of the changed files can be reverted. 👍

SinghCod3r:
@jkuester, thankyou for correcting the errors. but i want to know what changes did i accidently made to unrelated files. as when i ran the `git status `in my wsl terminal, it only shows the changes are in `index.js` and `version.specs.js` i guess . So tell me what happened?. And also, tell me for these cases what command should i use to avoid these steps. it will be very beneficial for me. 

thanks again !

jkuester:
:+1: You can check out the various file changes in [`27f39c8` (#10297)](https://github.com/medic/cht-core/pull/10297/commits/27f39c841c42feb60ccbc4967ce966ef88926b45)

Not sure exactly where the `lib-bowser.js` changes are coming from. :thinking: I have seen this leak into my own branches recently, so this may need more investigation.  But, the changes to the files in `config/default` are a "normal" side effect of building/deploying that config with cht-conf.

Ultimately, the main thing is to just pay attention to what is in context when you are creating a git commit and make sure to only commit the exact files that you updated.

SinghCod3r:
@jkuester , from now on when i work on any issue of this repo, i will do `git add <filename>` not the `git add .` 
and, i am seeing that there is one test case failing and it is Lint PR title, please review it. 

@jkuester, if you get these type of any issue please let me know, i am interested in writing test cases too. So, if you get any please let me try it out..


thankyou in advance !

Issue #7077, opened 2021-04-27
Create and execute realistic replication scalability test


craig-landry:
Output should be reliable, realistic metrics of what we would expect for CHT 4.0/archv3 in production against a large CHT instance.

garethbowen:
To flesh this out some more, the ultimate goal is to get to a number of caring actions that cht-core 4.0.0 can support per month documenting things like number and size of servers required so we can give partners a good idea of how many users they can scale to given a known workload per user.

Obviously we don't want to run the suite for a whole month, so the script will need to extrapolate from a smaller scale run.

@lorerod has made a great start on this script for generating production like data here: https://github.com/medic/cht-core/tree/master/tests/scalability/generate_brac_data . The rest of the work is to write a script which uploads the generated data and gets a bunch of users to do the initial replication, then simulating cht-core by replicating bidirectionally periodically to upload and download docs.

@dianabarsan has agreed to take this work on when convenient so assigning over to her.

PR #10303, opened 2025-09-17
chore(#7077): ongoing replication scalability suite - continuation of #7704

lorerod:
# Description
This PR continues the work done in the [Ongoing replication scalability suite](https://github.com/medic/cht-core/pull/7704#top) by @dianabarsan.

**Builds upon:** #7704
**Closes:** #7077

ToDo: integrate with github workflow

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:ongoing-replication-fixed/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:ongoing-replication-fixed/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:ongoing-replication-fixed/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/tests/scalability/ongoing/run.sh
+++ b/tests/scalability/ongoing/run.sh
@@ -11,6 +11,17 @@ fi
 if [ $RUN_LOCAL = false ]
 then
   sudo shutdown -P +60
+  
+  # Set environment variables for CI execution
+  export BASE_URL="$MEDIC_URL"
+  export ADMIN_USER="medic"
+  export ADMIN_PASSWORD="medicScalability"
+  export DATA_DIR="./data"
+  
+  # Install dependencies
+  apt-get update
+  apt-get install default-jre nodejs npm git wget bzip2 jq -y
+  
   echo Cloning cht-core to /cht-core
   git clone --single-branch --branch "$TAG_NAME" https://github.com/medic/cht-core.git;
   cd cht-core/tests/scalability/ongoing
--- a/tests/scalability/ongoing/start_jmeter_ec2.sh
+++ b/tests/scalability/ongoing/start_jmeter_ec2.sh
@@ -7,8 +7,7 @@ echo "$GITHUB_RUN_ID"
 
 export S3_PATH=s3://medic-e2e/scalability/$TAG_NAME-$GITHUB_RUN_ID
 export DATA_DIR=./data
-# shellcheck source=run.sh
-source run.sh
+export NODE_TLS_REJECT_UNAUTHORIZED=0
 
 echo "Triggering EC2 Run Instance Command and getting Instance ID"
 instanceID=$(aws ec2 run-instances \


lorerod:
Hi @dianabarsan ! 
Could you do an early, light review of my draft PR?
The part I’m especially interested in is the logic for “initial replicate” and “replicate to/from”, because I don’t have deep knowledge in that area, and the earlier version of this implementation used a different approach for replication.
A quick look at that section now would help me make sure I’m heading in the right direction before I go deeper.

lorerod:
@witash, you showed interest in performance testing. Would you be open to trying this out and sharing an early review?

tests/scalability/ongoing/medic-os.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+sudo shutdown -P +60
```
dianabarsan, line 2: I don't think this is needed.

tests/scalability/ongoing/generate-data.js
```diff
@@ -0,0 +1,103 @@
+const path = require('path');
+
+const config = require('./config');
+const factoryPath = path.join(__dirname, '../../factories/real-world');
+const userFactory = require(path.join(factoryPath, 'users/user'));
+const dataFactory = require('./data-factory');
+const uploadData = require('./upload-data');
```
dianabarsan, line 7: I wonder if we could use the test-data-generator here instead of duplicating all this faker code.

tests/scalability/ongoing/workflow-replication.js
```diff
@@ -0,0 +1,340 @@
...
+
+let docIdsRevs;
+const BATCH_SIZE = 100;
+
+const getMissingDocIdsRevsPairs = async (localDb, remoteDocIdsRevs) => {
+  const localDocs = await getLocalDocList(localDb);
+  return remoteDocIdsRevs.filter(({ id, rev }) => !localDocs[id] || localDocs[id] !== rev);
+};
+
+const getLocalDocList = async (localDb) => {
```
dianabarsan, line 123: Is this basically duplicating the replication code? I think we would be better off if we used the replication code as is, by requiring it, instead of having this duplication.

tests/scalability/ongoing/workflow-replication.js
```diff
@@ -0,0 +1,340 @@
...
+  let totalDownloaded = 0;
+  do {
+    const batchDownloaded = await getDocsBatch(remoteDb, localDb);
+    totalDownloaded += batchDownloaded;
+  } while (docIdsRevs.length > 0);
+  
+  return totalDownloaded;
+};
+
+const replicateFrom = async (previousReplicationResult) => {
```
dianabarsan, line 160: We no longer use the changes endpoint for replication. As per my comment above, we should use the replication code from webapp as-is somehow.

dianabarsan (commented):
Nice start! I'm excited about this.

tests/scalability/ongoing/generate-data.js
lorerod: To confirm, you mean [https://github.com/medic/test-data-generator](https://github.com/medic/test-data-generator), right? I like that idea, so we rely only on one way of generating test data. And we can clean up this repo; the `/factories/real-world` is not used anywhere.

dianabarsan: yes! I think duplicating the "factories" just adds extra layers of complexity.

tests/scalability/ongoing/workflow-replication.js
lorerod: Yes! This is exactly what I was thinking, but please bear with me while I understand… Is [this](https://github.com/medic/cht-core/blob/586606a195d3294ceeb2570f5706e4f56ddea295/webapp/src/ts/services/db-sync.service.ts#L371) the code I need to reuse? And how do I use it here? I suppose something similar to what we do [here](https://github.com/medic/cht-core/blob/4a86844805141887bd9e39835ace76b12bff9263/tests/scalability/ongoing/initial-replication.js#L45) for initial replication?

lorerod: Also, can I use `rewire` with a ts file?

dianabarsan: You will have to somehow call this when replicating to: https://github.com/medic/cht-core/blob/586606a195d3294ceeb2570f5706e4f56ddea295/webapp/src/ts/services/db-sync.service.ts#L103 and this for replicating from: https://github.com/medic/cht-core/blob/586606a195d3294ceeb2570f5706e4f56ddea295/webapp/src/ts/services/replication.service.ts#L20

lorerod: Hi @dianabarsan, I'm resuming this work. Sorry for the delay.
I looked into this. It turns out to be quite complicated. I also tried with [babel-plugin-rewire-ts](https://github.com/rosswarren/babel-plugin-rewire-ts), and I still wasn't able to make this work. Apparently, we’d need to mock many Angular dependencies to make this work.

Would you happen to know any workaround to call those methods without overcomplicating the setup? I’d really prefer to reuse the existing replication logic if there’s a practical way to do it.

tests/scalability/ongoing/generate-data.js
lorerod: @dianabarsan For the ongoing replication work (Phases 2 & 3), I’ll need to add an option to the `test-data-generator` to save a percentage of generated reports as JSON files, grouped by user, rather than uploading everything to the instance.
This way, we can pre-generate the data once and reuse it locally for the ongoing replication tests.
I’m thinking of adding parameters like:
```
--save-reports-json <dir>
--save-reports-percentage <0-100>
```
Each user would then have their own file, e.g., chw_username-reports.json, that can be uploaded in the ongoing replication phases.
What do you think?

dianabarsan: I think that's good idea! maybe not individual json files per doc, but grouped by something.

tests/scalability/ongoing/workflow-replication.js
dianabarsan: I'm bookmarking this so I check it out next week! sorry for the delay.

PR #10307, opened 2025-09-18
feat: Add telemetry for external app launcher

### LINKED ISSUE #10217: Record telemetry for External Android App calls
The CHT supports [launching 3rd Party Android apps](https://docs.communityhealthtoolkit.org/building/forms/app/#android-app-launcher) from within a form and then recording the response data from the app as fields on the report.

Unfortunately, we are not actually collecting any [telemetry data](https://docs.communityhealthtoolkit.org/technical-overview/data/performance/telemetry/) right now regarding the frequency of use of this feature and/or how long is spent in the 3rd Party app. This information is important to organizations deploying this feature so they can measure adoption and/or identify performance bottlenecks.  

We should update the [android-app-launcher.service.ts](https://github.com/medic/cht-core/blob/master/webapp/src/ts/services/android-app-launcher.service.ts) to collect this telemetry data (reference similar code in the [form service](https://github.com/medic/cht-core/blob/master/webapp/src/ts/services/form.service.ts)):

- Add the [PerformanceService](https://github.com/medic/cht-core/blob/master/webapp/src/ts/services/performance.service.ts) as a constructor parameter.
- In the AndroidAppLauncherService, in the [`launchAndroidApp` method](https://github.com/medic/cht-core/blob/b2f81083316600fdfda7e041ab2a8d80d07192ec/webapp/src/ts/services/android-app-launcher.service.ts#L31), call the `performanceService.track()` function (save the returned value as a class-level variable).
- In the [`resolveAndroidAppResponse` method](https://github.com/medic/cht-core/blob/b2f81083316600fdfda7e041ab2a8d80d07192ec/webapp/src/ts/services/android-app-launcher.service.ts#L18) call `stop` on the tracking object with the name `enketo:external-app:${action}` where `action` is the field from the ChtAndroidApp parameter given to the `launchAndroidApp` method.

We should also be sure to document the new telemetry value [on the docs page](https://docs.communityhealthtoolkit.org/technical-overview/data/performance/telemetry/).

I have hacked in a some functionality for recording the duration of time spent in the external Android app as an _output_ value in [this cht-android branch](https://github.com/medic/cht-android/pull/398/commits/8d201a09d07d339b3addba512c629a0a57e04dab). This allows the duration to be recorded on the report/contact associated with the form.  However, this is not good for aggregating proper telemetry across an instance.

**Additional context**

One thing to keep in mind when recording this performance data is that it also includes the time spend by cht-android to serialize/de-serialize the intent data.  If there is a lot of data being communicated (e.g. image data) the Base64 conversion may result in a non-trivial amount of time being spend serializing it.  Regardless, the telemetry collected here should still reflect the user's experience of launching the external app and then returning to be able to continue in the form.

ShivamGupta2003:
Closes #10217

I have implemented the performance telemetry for the Android App Launcher as requested.

The code builds successfully, but the unit tests for `AndroidAppLauncherService` are now failing. This is because the test needs to be updated to provide the new `PerformanceService` dependency, and new tests are needed to bring the code coverage back up.

I'm ready for feedback on how to proceed with updating the tests!


jkuester:
Sweet! Sounds good @ShivamGupta2003. :+1:   For the karma tests, you just need to provide a mock implementation of the `PerformanceService` when configuring the karma test bed in the Android app Launcher tests.  See [this code](https://github.com/medic/cht-core/blob/f1bdfc07c0f4b1170e0d8be671cd27aa105baff4/webapp/tests/karma/ts/services/search.service.spec.ts#L40) in the search tests as an example of what you need.  Basically we just use sinon to mock the `track` and `stop` functions.  Then the tests can precisely assert how the performance service is getting used.

Please ping me here if you have more questions and/or once you are ready for me to review this PR.

andrablaj:
Hi @ShivamGupta2003! Did you manage to progress on this PR?

Issue #10316, opened 2025-09-22
Add sidebar filter UI for previous months targets


jkuester:
- [ ] Display the filter icon on the green bar
- [ ] Add sidebar filter that displays when the button is clicked (With two radio buttons for This month and Previous month)
- [ ] Write e2e test for opening and closing sidebar.

Benmuiruri:
Hey @jkuester the PR is ready for review.

PR #10324, opened 2025-09-24
feat(#10316): add sidebar filter UI for previous months targets

Benmuiruri:
# Description

This PR closes https://github.com/medic/cht-core/issues/10316

Below is a checklist of what is included in the PR

- [ ] Display the filter icon on the green bar
- [ ] Add sidebar filter that displays when the button is clicked (With two radio buttons for This month and Previous month)
- [ ] Write e2e test for opening and closing sidebar.

# Code review checklist

# License

### DIFF
--- a/webapp/src/ts/modules/analytics/analytics-sidebar-filter.component.ts
+++ b/webapp/src/ts/modules/analytics/analytics-sidebar-filter.component.ts
@@ -1,0 +1,109 @@
+import { Component, EventEmitter, OnDestroy, OnInit, Output, Input } from '@angular/core';
+import { Subscription } from 'rxjs';
+import { Store } from '@ngrx/store';
+
+import { GlobalActions } from '@mm-actions/global';
+import { Selectors } from '@mm-selectors/index';
+import { ContactTypesService } from '@mm-services/contact-types.service';
+import { SettingsService } from '@mm-services/settings.service';
+import { NgClass, NgIf, NgFor } from '@angular/common';
+import { MatIcon } from '@angular/material/icon';
+import { MatAccordion } from '@angular/material/expansion';
+import { FormsModule } from '@angular/forms';
+import { TranslatePipe } from '@ngx-translate/core';
+
+@Component({
+    selector: 'mm-analytics-sidebar-filter',
+    templateUrl: './analytics-sidebar-filter.component.html',
+    imports: [NgClass, MatIcon, MatAccordion, NgIf, NgFor, FormsModule, TranslatePipe]
+})
+export class AnalyticsSidebarFilterComponent implements OnInit, OnDestroy {
+
+    @Input() userFacilities: any[] = [];
+    @Input() showFacilityFilter = true;
+    @Output() facilitySelectionChanged = new EventEmitter<string>();
+    @Output() reportingPeriodSelectionChanged = new EventEmitter<string>();
+    private globalActions;
+    readonly reportingPeriods = [
+        { value: ReportingPeriod.CURRENT, label: 'targets.this_month.subtitle' },
+        { value: ReportingPeriod.PREVIOUS, label: 'targets.last_month.subtitle' }
+    ];
+
+    DEFAULT_FACILITY_LABEL = 'Facility';
+    subscriptions: Subscription = new Subscription();
+    isOpen = false;
+    selectedReportingPeriod;
+    selectedFacility;
+    facilityFilterLabel;
+
+    constructor(
+        private store: Store,
+        private contactTypesService: ContactTypesService,
+        private settingsService: SettingsService,
+    ) {
+        this.globalActions = new GlobalActions(store);
+    }
+
+    ngOnInit() {
+        this.subscribeToStore();
+        this.setFacilityLabel();
+    }
+
+    ngOnDestroy() {
+        this.subscriptions.unsubscribe();
+    }
+
+    private subscribeToStore() {
+        const subscription = this.store
+            .select(Selectors.getSidebarFilter)
+            .subscribe((filterState) => {
+                this.isOpen = filterState?.isOpen ?? false;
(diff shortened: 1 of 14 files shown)


webapp/src/ts/modules/analytics/analytics-sidebar-filter.component.ts
```diff
@@ -13,16 +13,17 @@ import { FormsModule } from '@angular/forms';
...
-  templateUrl: './analytics-target-aggregates-sidebar-filter.component.html',
+  selector: 'mm-analytics-sidebar-filter',
+  templateUrl: './analytics-sidebar-filter.component.html',
   imports: [NgClass, MatIcon, MatAccordion, NgIf, NgFor, FormsModule, TranslatePipe]
 })
-export class AnalyticsTargetAggregatesSidebarFilterComponent implements OnInit, OnDestroy {
+export class AnalyticsSidebarFilterComponent implements OnInit, OnDestroy {
 
-  @Input() userFacilities;
+  @Input() userFacilities: any[] = [];
```
jkuester: Minor, but it [looks like](https://github.com/medic/cht-core/blob/d5ceadec239e1c5e4fc9603dd274fdb30584a3a2/webapp/src/ts/services/user-settings.service.ts#L75) we should be able to just do:

```suggestion
  @Input() userFacilities: Place.v1.Place[] = [];
```

webapp/tests/karma/ts/components/filters/analytics-filter.component.spec.ts
```diff
@@ -130,14 +122,12 @@ describe('Analytics Filter Component', () => {
 
   it('should not display filter button if targetAggregate is not enabled', fakeAsync(() => {
```
jkuester: I think we can remove this test case (since we no longer even check `isTargetAggregateEnabled`) and we can add another test case that asserts we show the filter button when the module id is `'targets'`

webapp/tests/karma/ts/modules/analytics/analytics-targets.component.spec.ts
```diff
@@ -35,16 +86,29 @@ describe('AnalyticsTargetsComponent', () => {
...
       })
       .compileComponents()
-      .then(() => {
+      .then(async () => {
+        const { MatIconRegistry } = await import('@angular/material/icon');
+        const { DomSanitizer } = await import('@angular/platform-browser');
+        const matIconRegistry = TestBed.inject(MatIconRegistry);
+        const domSanitizer = TestBed.inject(DomSanitizer);
+        const iconUrl = domSanitizer.bypassSecurityTrustResourceUrl('./img/icon-close.svg');
+        matIconRegistry.addSvgIcon('icon-close', iconUrl);
```
jkuester: I was trying to figure out what was going on with this.  (I don't see anywhere else where we call `.addSvgIcon`).  I removed it and the tests still pass, but print the error: 

```
'ERROR', Error: Error retrieving icon :icon-close! Unable to find icon with the name ":icon-close"
Error: Error retrieving icon :icon-close! Unable to find icon with the name ":icon-close"
    at Object.error (http://localhost:9876/_karma_webpack_/webpack:/node_modules/@angular/material/fesm2022/icon.mjs:966:48)
    at ConsumerObserver.error (http://localhost:9876/_karma_webpack_/webpack:/node_modules/rxjs/dist/esm/internal/Subscriber.js:102:33)
    at SafeSubscriber._error (http://localhost:9876/_karma_webpack_/webpack:/node_modules/rxjs/dist/esm/internal/Subscriber.js:64:30)
    at SafeSubscriber.error (http://localhost:9876/_karma_webpack_/webpack:/node_modules/rxjs/dist/esm/internal/Subscriber.js:40:18)
    at OperatorSubscriber._error (http://localhost:9876/_karma_webpack_/webpack:/node_modules/rxjs/dist/esm/internal/Subscriber.js:64:30)
    at OperatorSubscriber.error (http://localhost:9876/_karma_webpack_/webpack:/node_modules/rxjs/dist/esm/internal/Subscriber.js:40:18)
    at Observable._subscribe (http://localhost:9876/_karma_webpack_/webpack:/node_modules/rxjs/dist/esm/internal/observable/throwError.js:5:45)
    at Observable._trySubscribe (http://localhost:9876/_karma_webpack_/webpack:/node_modules/rxjs/dist/esm/internal/Observable.js:37:25)
    at cb (http://localhost:9876/_karma_webpack_/webpack:/node_modules/rxjs/dist/esm/internal/Observable.js:31:30)
    at errorContext (http://localhost:9876/_karma_webpack_/webpack:/node_modules/rxjs/dist/esm/internal/util/errorContext.js:19:9)
```

I assume that error is what this code is trying to avoid?  I see similar errors getting printed 100 times throughout the test logs.  😅 

So, first, thank you for taking the time to find a workaround here!  ❤️   Based on your efforts here, is there something we could do to generalize this to resolve the other errors? (And/or is there some problem with how we are bootstrapping karma that is causing us to have to do this manually?)  

None of this needs to be solved in this PR.  Just trying to collect more info here so maybe we can log an issue about this... 👍

jkuester (approved):
Most excellent!  Just a few suggestions/questions, but otherwise this is good to go!

webapp/tests/karma/ts/modules/analytics/analytics-targets.component.spec.ts
Benmuiruri: You are right, I added it to quiet down all "noise" when running the test 😅 

We currently register the icon in [app.component](https://github.com/medic/cht-core/blob/e80cfc053230ceb98ee172890449e9845062d1be/webapp/src/ts/app.component.ts#L179). 

I will delete the temporary fix and we can have a seperate issue that better does the registration of svg icons in the test suite perhaps in karma config as you mention.

Benmuiruri:
@jkuester this should be ready to merge. Thanks

Issue #8340, opened 2023-06-23
Add impact metrics API


1yuv:
**Is your feature request related to a problem? Please describe.**
Currently, we collect certain metrics from partner(s) for reporting purpose after partner agree to share those aggregated metrics. Some partner(s) might opt to provide these data and they might have data warehouse, but some may not have. This is the request asking if certain metrics are calculable on CouchDB directly and can be exposed via API. In this case, after having Data usage Agreement with partner(s) we can poll those data directly from API without needing to setup rdbms and have additional resoures used for that purpose. 

**Describe the solution you'd like**
Provide following data from API (whichever are possible)
- Count of Active CHT User(s) in the past/current month 
- Total number of Caring Activities (Total number of `data_record` docs created) 
- Count of unique patient in the system (all time) 
- Number of households registered

**Describe alternatives you've considered**

**Additional context**

garethbowen:
Because contacts are configurable we can't know for sure what is a "patient" or "household" - would it be sufficient to return the count of contacts grouped by the contact type, which would need some knowledge of context to do the translation, eg:

```json
{
  contacts: {
    person: 590,
    clinic: 15,
    ...
  }
}
```

I'm assuming "active" in this context is syncing at least one caring activity (`data_record`)?

yrimal:
Hi @garethbowen, Yes, contacts grouped by the contact type would be absolutely fine, and probably best. 
>I'm assuming "active" in this context is syncing at least one caring activity (data_record)?

Yes.

garethbowen:
Some ideas on the right views to serve these requests...

- "Count of unique patient in the system (all time)" and "Number of households registered": `medic-client/contacts_by_type`. You could add a `_count` reducer to cache the result.
- "Total number of Caring Activities". You could use `medic-client/data_records_by_type` which already has a `_count` reduce function.
- "Active CHT User(s)". This is a bit harder because we haven't codified the concept of "active". We know when they last connected which is used in the monitoring api but we don't have a view for last time they wrote a data_record. It wouldn't be hard to make one using the contact and reported_date fields, or we could add an `author` field to the infodoc and report off that.

garethbowen:
Additional metrics which may be useful...

- total users
- active households (at least one report filed for that household in the last time period)
- aggregate of each report type

mrjones-plip:
[...]

Issue #10322, opened 2025-09-23
Identify and remove all unused couchdb views


1yuv:
**Describe the performance issue**
We have some discussions ongoing on forum and community where CHT is not performing fast or is taking too much space. It appears that there are some views that were created in the past that are still on the codebase and that get created on all instances even though there's no specific use of those views. These viewes consume unnecessary disk space on the couchdb. 

For example, there's a view [data_records_by_anestor](https://github.com/medic/cht-core/tree/master/ddocs/medic-db/medic-scripts/views/data_records_by_ancestor) which doesn't have any usage on CHT.  Some of the views on this` medic-scripts` folder may have been used somewhere outside of cht-core for some functionality of CHT. But view like `data_records_by_ancestor` doesn't appear to be used and if you look at the code, the file the code mentions it's being used no longer exists on repo.

If there are views that are not used and still being created, they consume unnecessary storage and CPU and RAM while running view warming process as well. 

**Describe the improvement you'd like**
Identify all unused views and remove them safely. 

**Describe alternatives you've considered**
Leave as is

**Measurements**
For a lumbini-ne a `medic` database with 5.3 GB data has 2.1GB views on `medic-scripts` folder only. 

**Environment**
- Instance: Any
- Client platform: Any
- App: Couchdb
- Version: All

witash:
We looked into this as part of tco v2, but did not make much progress.
The difficulty is its hard to tell what is really unused by all deployments, and we are cautious to avoid breaking things.
Also individual views tend not to be very large (apart form the freetext views) so to get a significant performance or space improvement we would have to remove many.

For example [data_records_by_ancestor](https://github.com/medic/cht-core/tree/master/ddocs/medic-db/medic-scripts/views/data_records_by_ancestor) is used at by [delete_training_data_utils.js#L151 ](https://github.com/medic/support-scripts/blob/12c636268e012fcd6d591695df640e5f5add20aa/dev/data_deletion/delete_training_data_utils.js#L151) in [support-scripts repo](https://github.com/medic/support-scripts)

This refers to "training data" and was last modified 4 years ago so its very unlikely to still be useful, but because it is still referenced somewhere, we were hesitant to delete it entirely.
`places_by_type_parent_id_name`, `places_by_contact` are similar examples that are only used by old-looking support scripts.

`total_clinics_by_facility` does appear to be completely unused.

Other than that, there are some that are redundant (or redundant with small tweaks) or have dubious value, but none are truly unused anywhere.
[...]

PR #10325, opened 2025-09-24
feat(#8340): expose reports and contacts by type

1yuv:
# Description

Expose impact metrics from API #8340 

# Code review checklist

# License


lorerod:
Hi @1yuv, the `ci-compile` script doesn’t just compile; it also runs fast unit tests and linting to catch basic issues early, before we waste time building Docker images or running heavy integration/E2E tests.

You can see exactly what it runs [here](https://github.com/medic/cht-core/blob/6fc8e86e0279c4637091c5a43298004ef013f69b/package.json#L64).

This PR failing tests (50 total):
- shared-libs/cht-datasource: 44 failures
    - test/local/contact.spec.ts: 25 
    - test/local/libs/doc.spec.ts: 6 
    - test/local/person.spec.ts: 7 
    - test/local/place.spec.ts: 6 
- shared-libs/search: 6 failures
    - test/generate-search-requests.js: 6
    
How to run exactly these tests locally for debugging:
 1. Add `.only` to a describe or it while iterating (remember to remove before pushing). For example:
```
 // In shared-libs/cht-datasourc/test/local/libs/doc.spec.ts
describe.only('local doc lib', () => {
// ...
});
```
2. Navigate to the library and run the test:
```
cd shared-libs/cht-datasource
npm test
```

Ping me if anything’s unclear and I can help dig into a specific failure.

jkuester:
Sorry that I have not been able to review this yet!  I ran out of time this week, but I have not forgotten it and it is in the queue for next week.  :+1:

api/src/services/impact.js
```diff
@@ -0,0 +1,81 @@
...
+    const contactsByType = await db.medic.query('medic-client/contacts_by_type', {
+      reduce: true,
+      group: true,
+    }).then(result => result.rows.reduce((acc, row) => {
+      acc[row.key[0]] = row.value;
+      return acc;
+    }, {}));
+    return contactsByType;
+  } catch (err) {
+    logger.error('Error fetching cotnacts type: %o', err);
```
jkuester:
```suggestion
    logger.error('Error fetching contacts by type: %o', err);
```

api/src/services/impact.js
```diff
@@ -0,0 +1,81 @@
...
+    return reportsCount;
+  } catch (err) {
+    logger.error('Error fetching reports count: %o', err);
+    return '';
+  }
+};
+
+const getContactsByType = async () => {
+  try {
+    const contactsByType = await db.medic.query('medic-client/contacts_by_type', {
```
jkuester: Don't need to make any change here, but just wanted to point out that I think if you do `return await ...` from inside the try-block, it still works as expected and any errors that come from resolving the promise are caught by the try-block (which, of course, is different from what happens if you just return the promise without awaiting it...).  Anyway, the `return await ...` just lets you avoid having this extra variable that you immediatly return.

api/src/services/impact.js
```diff
@@ -0,0 +1,81 @@
...
+      reduce: true,
+      group: true,
+    }).then(result => result.rows.reduce((acc, row) => {
```
jkuester: Don't shoot the messenger on this one (I was in favor of `reduce`), but the official CHT Style Guide says to [avoid using `reduce`](https://docs.communityhealthtoolkit.org/community/contributing/code/style-guide/#avoid-reduce) because it has poor readability.  

Can you update all the places here to not use reduce?  🙏

api/src/services/impact.js
```diff
@@ -0,0 +1,81 @@
...
+    ])
+    .then(([
+      userCount,
+      contactsByType,
+      reportsCount,
+      reportsByForm
+    ]) => {
+      return {
+        totalUsers: userCount,
+        totalReports: reportsCount,
```
jkuester: Shouldn't this always just be the same value as the sum of all the `reportsByForm` entries?  If so, I would hesitate to include this extra field unless there is a particular reason to do so. (Certainly it does not seem like we want the extra view query if we don't need it....)

api/src/services/impact.js
```diff
@@ -0,0 +1,81 @@
+const db = require('../db');
```
jkuester, line 1: Can we get some unit tests for this new service? 🙏

api/src/controllers/impact.js
```diff
@@ -0,0 +1,10 @@
+const service = require('../services/impact');
+const serverUtils = require('../server-utils');
+
+module.exports = {
+  getV1: (req, res) => {    
+    return service.jsonV1()
```
jkuester: Just adding this so I don't forget, we need to check user permissions here.  (Still discussing on Slack which ones...)

shared-libs/cht-datasource/src/local/libs/doc.ts
```diff
@@ -62,12 +62,13 @@ export const queryDocsByRange = (
 /** @internal */
 export const queryDocsByKey = (
   db: PouchDB.Database<Doc>,
-  view: string
+  view: string,
+  reduce = true
```
jkuester: These `query*` functions each have a specific targeted usage (returning `Doc` objects or the UUIDs, etc). In these cases I don't think we ever would want to reduce.  IMHO, we can just populate `reduce: false` in the options used and do not accept a separate parameter.  

If/when cht-datasource needs to use `reduce` in a query, we can and should probably add a dedicated helper function here for that.  🤷

tests/integration/api/controllers/impact.spec.js
```diff
@@ -0,0 +1,103 @@
...
+        contactsByType: {
+          person: 1
+        },
+        totalReports: 0,
+        totalUsers: 1,
+        reportsByForm: {}
+      });
+    });
+
+    it('when extra cotnact is added, person count is increased', async () => {
```
jkuester:
```suggestion
    it('when extra contact is added, person count is increased', async () => {
```

jkuester: Honestly, I don't really think this case is adding very much value since you also have a minimum data case and one with multiple contact types.  I think from the integration tests perspective, that is plenty of coverage (and we can just remove this test).  👍

tests/integration/api/controllers/impact.spec.js
```diff
@@ -0,0 +1,103 @@
...
+      });
+    });
+
+    it('all contact types are aggregated and returned', async () => {
+      const docs = [
+        { type: 'contact', contact_type: 'person', _id: 'p1' },
+        { type: 'contact', contact_type: 'p10_province', _id: 'gandaki' },
+        { type: 'contact', contact_type: 'p20_district', _id: 'kaski' },
+      ];
+      await utils.saveDocs(docs);
```
jkuester: Could we use the person/place factories here ([example](https://github.com/medic/cht-core/blob/5c1f9ba4a149041dd2a16e58420546731cd27b25/tests/integration/api/controllers/place.spec.js#L10)) to generate the contact data here?  That just helps to keep a single place in the test code to maintain when it comes to consistently generating test data.

tests/integration/api/controllers/impact.spec.js
```diff
@@ -0,0 +1,103 @@
...
+      });
+    });
+
+
+    it('all reports are aggregated and returned', async () => {
+      const docs = [
+        { type: 'data_record', form: 'L', _id: 'r1' },
+        { type: 'data_record', form: 'pregnancy', _id: 'r2' },
+        { type: 'data_record', form: 'pregnancy', _id: 'r3' },
+      ];
```
jkuester: Same here. Can we use the report factory to make this test data and also in the test below? ([example](https://github.com/medic/cht-core/blob/f1bdfc07c0f4b1170e0d8be671cd27aa105baff4/tests/integration/api/controllers/report.spec.js#L49))

api/src/controllers/impact.js
```diff
@@ -0,0 +1,10 @@
+const service = require('../services/impact');
+const serverUtils = require('../server-utils');
+
+module.exports = {
+  getV1: (req, res) => {
```
sugat009: suggestion: instead of doing a `getV1`, `getV2` and so on, it would cleaner to have this nested structure where the `v1` is a property in the `module.exports` and `get` is a property in `v1`. Like this: https://github.com/medic/cht-core/blob/master/api/src/controllers/contact.js#L17:L19.

Also, this format of writing the controllers makes it easier to debug than promise chaining: https://github.com/medic/cht-core/blob/master/api/src/controllers/contact.js#L19:L29

api/src/services/impact.js
```diff
@@ -0,0 +1,81 @@
...
+  try {
+    const totalReports = await db.medic.query('medic-client/data_records_by_type', {
+      reduce: true,
+      group: true,
+      key: 'report'
+    });
+    return totalReports.rows[0]?.value || 0;
+  } catch (err) {
+    logger.error('Error fetching reports count: %o', err);
+    return '';
```
sugat009: suggestion: in case of success, the output is a number whereas in case of error the output is a string, this will bring inconsistencies wherever these values are used. It makes sense to return a falsy number value like -1 or `null`.

shared-libs/cht-datasource/src/local/libs/doc.ts
1yuv: The pouchdb library uses `reduce` parameter internally whether we pass it or not. `reduce` is implicitly treated as `true` if applicable. The applicable condition here is: if the view has reduce function, it's set to `true` and if view doesn't have reduce function, it's ignored.

When we changed `contacts_by_type` view from map only to map/reduce, anyt time this function is called now, the reduce is set to `true` even if we don't pass it. However, we needed a way to query `contacts_by_type` with reduce=false as well so that existing functionality is not broken, and reduce=true, so that we can query the aggregated report. I think it makes sense to have a configurable parameter being configured than adding a copy function that'd have a same code except one more parameter.

api/src/services/impact.js
1yuv: Yes they should. I used different view since it was already reduce and wouldn't cost much to query thinking it would also serve as check for each other. However, I also thought about #10322, and if I use one view that can serve both purpose, that view could also be up for clean up.

sugat009:
@1yuv the integration tests seem to failing for the impact endpoint

api/src/controllers/impact.js
```diff
@@ -1,10 +1,24 @@
...
-      .catch(err => serverUtils.error(err, req, res));
+  v1: {
+    get: serverUtils.doOrError(async(req, res) => {
+      await checkUserPermissions(req);
+      try {
+        const impact= await service.jsonV1();
+        res.json(impact);
+      } catch(err){
+        serverUtils.error(err, req, res);
+      }
```
jkuester:
```suggestion
      const impact= await service.jsonV1();
      res.json(impact);
```

I do not think you need the extra try-block here since you have wrapped this in the `doOrError`.

shared-libs/cht-datasource/src/local/libs/doc.ts
jkuester: > However, we needed a way to query contacts_by_type with reduce=false as well so that existing functionality is not broken, and reduce=true, so that we can query the aggregated report

My point here is that the code in cht-datasource is currently _never_ querying with reduce=true.  AND if we were to query with reduce=true, then we would probably need _a different return type_ on these functions anyway since the value returned from the reduce will be different from the map...

These functions here are intended to be single-purpose utils that only perform one type of operation.

api/src/services/impact.js
```diff
@@ -1,79 +1,57 @@
...
-    return reportsCount;
+    });
+    const report = {};
+    let total = 0;
+    for (const row of result.rows) {
+      const [form] = row.key;
+      report[form] = row.value;
+      total += row.value;
+    }
+    return { report, total };
```
jkuester: Okay, sorry to do this now, but the structuring here felt a bit off with `reports.report...`:

```json
{
  "users": 5,
  "contacts": {
    "clinic": 1,
    "district_hospital": 1,
    "health_center": 1,
    "person": 1
  },
  "reports": {
    "report": {
      "countdown_timer": 1,
      "death_report": 6,
      "delivery": 2,
    },
    "total": 9
  }
}
```

This got me to thinking more about the naming/structure and I am concerned that the current approach will be confusing to extend in the future when adding more data to this endpoint.  The naming does not include any information about the _units_ for the returned values or the _type_ of metric being measured (e.g. is this the "total count" of something, percentage, etc).  (My thinking on this subject has been heavily influenced by the [Prometheus naming conventions](https://prometheus.io/docs/practices/naming/)).

We could rework the structure of the returned data a bit so that the structure is more open to future values being added:

<details>
<summary>Alternative 1</summary>

```json
{
  "users": {
    "docs_total": 5
  },
  "contacts": {
    "docs_total": 4,
    "clinic_docs_total": 1,
    "district_hospital_docs_total": 1,
    "health_center_docs_total": 1,
    "person_docs_total": 1
  },
  "reports": {
    "docs_total": 9,
    "countdown_timer_docs_total": 1,
    "death_report_docs_total": 6,
    "delivery_docs_total": 2
  }
}
```

</details> 

The main think I don't like about the first alternative is that it means that your contact_type and form_id values are getting used as _part_ of property keys (with the `_docs_total` suffix). That make for more work when processing this data programmatically (as opposed to just having the contact_type/form_id be the whole property key with no suffix). Another alternative that solves this problem, but is extra verbose would be:

<details>
<summary>Alternative 2</summary>

```json
{
  "users": {
    "docs_total": 5
  },
  "contacts": {
    "docs_total": 4,
    "clinic": {
      "docs_total": 1
    },
    "district_hospital": {
      "docs_total": 1
    },
    "health_center": {
      "docs_total": 1
    },
    "person": {
      "docs_total": 1
    }
  },
  "reports": {
    "docs_total": 9,
    "countdown_timer": {
      "docs_total": 1
    },
    "death_report": {
      "docs_total": 6
    },
    "delivery": {
      "docs_total": 2
    }
  }
}
```

</details>

Honestly, since this is more of a general purpose endpoint that might eventually return all kinds of input data, I am leaning towards Alternative 2 as my preference, but maybe you can think of a better structure besides these alternatives that would still be clear to consume and easy to extend?

api/tests/mocha/services/impact.spec.js
```diff
@@ -0,0 +1,100 @@
...
+  });
+
+  it('propagates error when users.info fails', async () => {
+    sinon.stub(db.users, 'info').rejects(new Error('users fail'));
+    sinon.stub(db.medic, 'query').resolves({ rows: [] });
+
+    try {
+      await service.jsonV1();
+      throw new Error('expected jsonV1 to reject');
+    } catch (err) {
```
jkuester: Do not need to change anything here, but I am a big fan of the `expect(() => ...).to.throw(...)` pattern [like this](https://github.com/medic/cht-core/blob/5c1f9ba4a149041dd2a16e58420546731cd27b25/shared-libs/cht-datasource/test/qualifier.spec.ts#L23).  Just less verbose than needing the whole try-block.

api/src/services/impact.js
1yuv: Originally the metrics had some meaningful naming, then I tried to copy `/api/v2/monitoring/` naming, then they lost meaning. Your input is useful and I tried to make api result more expressive and easy to extend. 

<details>
      <summary>New implementation</summary>

```json
{
    "users": {
        "total": 2
    },
    "contacts": {
        "total": 2007,
        "by_type": [
            {
                "type": "g10_district",
                "count": 3
            },
            {
                "type": "g20_municipality",
                "count": 2
            },
            {
                "type": "g30_health_center",
                "count": 2
            },
            {
                "type": "g40_clinic",
                "count": 1
            },
            {
                "type": "person",
                "count": 1999
            }
        ]
    },
    "reports": {
        "total": 3007,
        "by_form": [
            {
                "form": "L",
                "count": 1005
            },
            {
                "form": "N",
                "count": 2002
            }
        ]
    }
}
```
 </details>

shared-libs/cht-datasource/src/local/libs/doc.ts
1yuv: Ah, I got your point now Josh. That makes sense and makes things easier for me. For some reason, I thought this method was also called with `reduce=true` from some place (not for the view in question, but for other views because that seemed to be the default behavior if we don't pass that parameter. But, it makes sense for this to have strictly reduce:false. That'd have saved my time earlier instead of modifying every existing call. I will make this change.

api/src/services/impact.js
jkuester: Nice! I think the array setup will be easier to work with in the consuming code. :+1:   One minor suggestion would be to go with `count` everywhere instead of using `total` at the upper level.  I think it will be clear that `reports.count` is the "total" count of all reports while `reports.by_form.count` is the count of all the reports of that type.

1yuv: Updated.

Issue #8657, opened 2023-10-23
Delete transition code in telemetry service


latin-panda:
**Describe the issue**
As part of the work done on this [issue](https://github.com/medic/cht-core/issues/7616), we refactored the way telemetry is stored in the devices. We added a small temporary code to remove the old telemetry db and localStorate properties ([PR](https://github.com/medic/cht-core/pull/8555) - [commit](https://github.com/medic/cht-core/pull/8555/commits/0a191c31b7654948e611c2c8c9337a45a712079b))

## Good first issue

This ticket has the [Good first issue label](https://github.com/medic/cht-core/issues?q=state%3Aopen%20label%3A%22Good%20first%20issue%22)! This means it's been especially curated by other CHT contributors to be easy to work on for first time contributors [per the docs](https://docs.communityhealthtoolkit.org/community/contributing/creating-good-first-issues/).

To succeed on this ticket, please:
1. Delete the [`deleteDeprecatedTelemetryDB` function](https://github.com/medic/cht-core/blob/601d992394983f589ae627f3fe11a4b080bc5b42/webapp/src/ts/services/telemetry.service.ts#L344) from the telemetry service. (And remove the line where it is called.)
2. Run the webapp unit tests (`npm run unit-webapp`) and fix any tests that are now failing (removing any tests that are no longer relevant).

latin-panda:
@jkuester 1 year ago, we changed how telemetry is stored in devices, but we added some code for backward compatibility. It's super easy to remove. 
I've added it to the 5.0.0 milestone to avoid losing track of this.

kevSeguraEdify:
Hi @jkuester, I'm from Edify, if this ticket is still available please assign it to me so I can review it.

Issue #9202, opened 2024-06-24
Increase ecmaversion linting for ddocs


garethbowen:
**Describe the issue**

CouchDB 3 increased their supported JS engine version which meant more modern features could be used in code such as map/reduce view definitions. However due to the way we warm views before upgrading this caused [a bug](https://github.com/medic/cht-core/issues/9166) where releases were blocked because the view warming on CouchDB 2 failed. This only impacts upgrades from a version of CHT with CouchDB 2. The fix for 4.x is to not use modern JS features.

**Describe the improvement you'd like**

In the next major, we can safely bump [the ecma version](https://github.com/medic/cht-core/blob/8ce2fd08c11bf1e283c2db4af2ecf29a3d1c1896/ddocs/.eslintrc#L10) and list this as a breaking change when upgrading from 4.4. The workaround is to upgrade to the latest 4.x release before upgrading to 5.x.

**Describe alternatives you've considered**

- We could document the broken backwards compatibility but it's hard to make it clear without also bumping the major
- We could disable view warming or skip over if it fails for some reason - this risks taking a production instance out for days

PR #10331, opened 2025-09-25
feat(#8657): remove logic for cleaning up old telemetry db

kevSeguraEdify:
# Description

- The temporary [deleteDeprecatedTelemetryDB function](https://github.com/medic/cht-core/blob/601d992394983f589ae627f3fe11a4b080bc5b42/webapp/src/ts/services/telemetry.service.ts#L344) was deleted and it's calls.
- All unit tests that fails because of this change were fix.

Closes [#8657](https://github.com/medic/cht-core/issues/8657)

# Code review checklist

# License

### DIFF
--- a/webapp/src/ts/services/telemetry.service.ts
+++ b/webapp/src/ts/services/telemetry.service.ts
@@ -322,7 +322,6 @@ export class TelemetryService {
     try {
       const today = this.getToday();
       const databaseNames = await this.indexedDbService.getDatabaseNames();
-      await this.deleteDeprecatedTelemetryDB(databaseNames);
       const telemetryDBs = await this.getTelemetryDBs(databaseNames);
       await this.submitIfNeeded(today, telemetryDBs);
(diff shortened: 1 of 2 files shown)


kevSeguraEdify:
Hi @jkuester @latin-panda, can you review this, please?

jkuester (approved):
Clean and simple!  Thank you @kevSeguraEdify!


Also, for anyone reading this in the future, I think we should be safe removing this code here in the `5.0.0`. release because:

- The code to initially add the transition logic was [added in `4.5.0`](https://github.com/medic/cht-core/pull/8555).  One of the requirements for upgrading to `5.0.0` is that the CHT instance must be running `4.5.0` (because of [Couch view stuff](https://github.com/medic/cht-core/issues/9202)).  This means that everyone should at least have had the opportunity to run this transition code and clean up the database.
- The worst case scenario if a user has not run the code (maybe they upgrade to `4.5.0` and before all the users sync, they upgrade again to `5.0.0`) would be that a bit of extra storage is used on the device. This should not have any impact on functionality.

PR #10337, opened 2025-09-29
feat(#10317): support reporting periods in rules engine service

### LINKED ISSUE #10317: Add service support for getting previous month targets
Update the rules engine service to support returning targets for previous month.

paulpascal:
# Description

Closes #10317

This pull request updates the rules engine service to support returning targets for the previous month.

# Code review checklist

# License

### DIFF
--- a/webapp/src/ts/services/rules-engine.service.ts
+++ b/webapp/src/ts/services/rules-engine.service.ts
@@ -476,15 +476,14 @@ export class RulesEngineService implements OnDestroy {
     // Previous month: use target interval from cht-datasource
     try {
       const settings = await this.settingsService.get();
-      const result = await this.fetchTargetDocumentsForPeriod(settings, reportingPeriod);
+      const targetInterval = await this.fetchTargetDocumentsForPeriod(settings, reportingPeriod);
 
-      // If result is undefined, it means prerequisites weren't met (no username/contact)
-      if (result === undefined) {
+      // If no target interval found (null), return empty array
+      if (!targetInterval) {
         return [];
       }
 
-      // If result is null, the query succeeded but no target interval was found
-      return this.processTargetDocuments(result, settings);
+      return this.processTargetDocuments(targetInterval, settings);
     } catch (error) {
       console.error('Error fetching previous month targets:', error);
       return [];
@@ -541,18 +540,18 @@ export class RulesEngineService implements OnDestroy {
 
   /**
    * Fetch target interval for a specific reporting period using cht-datasource
-   * @returns the target interval, null if not found, or undefined if prerequisites aren't met
+   * @returns the target interval, or null if not found or prerequisites aren't met
    */
   private async fetchTargetDocumentsForPeriod(
     settings: any,
     reportingPeriod: ReportingPeriod
-  ): Promise<TargetInterval.v1.TargetInterval | null | undefined> {
+  ): Promise<TargetInterval.v1.TargetInterval | null> {
     const intervalTag = this.getTargetIntervalTag(settings, reportingPeriod);
     const userContact = await this.userContactService.get();
     const username = this.sessionService.userCtx()?.name;
 
     if (!userContact?._id || !username) {
-      return undefined;
+      return null;
     }
 
     return await this.getTargetInterval(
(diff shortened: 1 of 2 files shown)


jkuester:
@paulpascal thanks for the great PR!  I spent a bunch of time this afternoon reading through the code changes and even more time looking at the rest of the context around target documents and considering how best to proceed.

I have created [a post on the forum](https://forum.communityhealthtoolkit.org/t/proposal-to-include-target-intervals-in-cht-datasource/5385) where I have proposed adding targets to cht-datasource.  Please respond out on that thread and let me know what you think! :pray: It would add another sub-task or two to this project, but the new code should follow well established patterns and be pretty straightforward to implement.

jkuester:
@paulpascal would you have time to resurrect/complete this PR?  I think we just need to update the implementation of your `fetchTargets(reportingPeriod: ReportingPeriod = ReportingPeriod.CURRENT)` method so that under the hood it uses the cht-datasource functionality I added in https://github.com/medic/cht-core/pull/10390.  We can keep your current logic for calculating the reporting period and you are already getting the user's contact id.  Then, all we need is the `sessionService.userCtx().name` value and we have everything we need to call the `TargetInterval.v1.get` function from cht-datasource to get the targets doc.  Use the `CHTDatasourceService` to bind the function to the current datasource [like this](https://github.com/medic/cht-core/blob/b5c91d941cbcac7f06f6f96269b6a60ac0941bf5/webapp/src/ts/services/user-contact.service.ts#L19).

paulpascal:
Hi @jkuester , Thanks for the following up and explanations.
I am on it

paulpascal:
Hello @jkuester - I have made some adjustment accordingly.
Would you like to have a second look at it?

thnks

webapp/src/ts/services/rules-engine.service.ts
```diff
@@ -459,11 +463,35 @@ export class RulesEngineService implements OnDestroy {
...
+      return this._fetchCurrentTargets();
+    }
+
+    // Previous month: use target interval from cht-datasource
+    try {
+      const settings = await this.settingsService.get();
+      const result = await this.fetchTargetDocumentsForPeriod(settings, reportingPeriod);
+
+      // If result is undefined, it means prerequisites weren't met (no username/contact)
+      if (result === undefined) {
```
jkuester: Do we need to treat this as a separate case from `null`?  I am trying to figure out if a consumer of this function would ever care about the difference between "we cannot check for target docs" vs "we did not find a target doc"?  Seems like in both cases we just want to run the same downsteam logic... (Or maybe we just want to return `[]` in either case?)

webapp/src/ts/services/rules-engine.service.ts
```diff
@@ -492,6 +520,95 @@ export class RulesEngineService implements OnDestroy {
...
+
+      if (targetValue) {
+        // Target has data in the interval
+        processedTargets.push({
+          ...targetConfig,
+          ...targetValue,
+          visible: true
+        });
+      } else {
+        // No data for this target - add zero values
```
jkuester: I have done some thinking about this case and I am not sure we want to return zero values here....

I think one likely time you might run into this is when the target config _changes_ and a new target is added.  That new target will be in `targetsConfig`, but we will not have captured a value in last month's target doc.  So, say we configure a new target of "Births in a facility". If we return zero here, it will look like the user did not record any facility births last month. However, this could actually be wrong. Maybe they recorded many facility births, but we did not record them in the targets doc because that was not a target that was tracked last month...

I am leaning towards saying that if we don't have data for a particular target, then we should not just not return anything for that target at all.  I am not 100% sure how this will affect the downstream UI logic, but I think it would be better to just not show the "Births in a facility" widget when the user selects the "Previous month" vs showing `0` for last month's "Births in a facility"...  Does that make sense, or did you have other reasons why we needed to add default data here?

jkuester (changes requested):
Thanks @paulpascal, this is great!  Couple remaining questions about handling the cases of not-found targets and this will be good to go.

webapp/src/ts/services/rules-engine.service.ts
paulpascal: Good point! I've simplified this - now both cases return null and we just return [] when no target interval is found. The distinction was unnecessary from a consumer perspective.

webapp/src/ts/services/rules-engine.service.ts
paulpascal: I def agree with you @jkuester  about the "new target config" scenario. I've updated the logic to only return targets that actually exist in the target interval document.

paulpascal:
Thanks @jkuester. 
I dont have commit privileges to this repository.
Could you please merge the PR for me?

PR #10339, opened 2025-09-30
chore(#10261): increase Nginx Body Size to support large attachments

### LINKED ISSUE #10261: Increase nginx request limit beyond 32MB
As part of supporting large attachments for forms, we need nginx to allow requests that are over 32MB: 
https://github.com/medic/cht-core/blob/master/nginx/nginx.conf#L15C5-L15C25

We should keep the 32MB limit for json requests, but increase the limit for other mime types, especially media.

shashi-sah2003:
fixes #10261


shashi-sah2003:
@dianabarsan I think we are gtg.

nginx/nginx.conf
```diff
@@ -12,7 +12,7 @@ http {
     sendfile              on;
     keepalive_timeout     3600;
     proxy_read_timeout    3600;
-    client_max_body_size  32M;
+    client_max_body_size  1024M;
```
dianabarsan: Is there a way we can change this so that JSON requests still only allow 32MB and media requests allow the increased value? 
A hint that it can be done: https://serverfault.com/questions/919532/nginx-limit-client-max-body-size-based-on-content-type

shashi-sah2003:
@dianabarsan I have made the changes using map to allow 32MB only for json type and else for 1 GB

dianabarsan (changes requested):
It looks like it's not exactly working as expected: https://github.com/medic/cht-core/actions/runs/18138966724/job/51626049171?pr=10339#step:18:753

shashi-sah2003:
@dianabarsan I have added `client_max_body_size  $upload_limit;`, thought it might be able to resolve the issue but on compiling the ci, the api isn't reachable. what could be the possible reason for that?

dianabarsan:
Unsure. Can you try running the test locally to debug?

shashi-sah2003:
@dianabarsan I think we couldn't use the change of assigning value at runtime to `client_max_body_size`. lmk if I am missing out something. PFA
image

dianabarsan:
I believe the config is wrong and we need to change it differently. I would go and check nginx documentation: https://nginx.org/en/docs/http/ngx_http_core_module.html
to see what we can change, but I'm certain that there is a way to do this.

PR #10362, opened 2025-10-08
feat(#9255): task notification(updated)

### LINKED ISSUE #9255: Enable android notifications for task reminders/alerts

jonathanbataire:
# Description

fixes #9255 

# Code review checklist

# License

### DIFF
--- a/shared-libs/task-utils/test/order-by-due-date-and-priority.js
+++ b/shared-libs/task-utils/test/order-by-due-date-and-priority.js
@@ -52,9 +52,15 @@ describe('TaskUtils shared lib - orderByDueDateAndPriority function', function (
   });
 
   it('should treat various due date formats correctly', () => {
-    const result1 = taskUtils.orderByDueDateAndPriority(
+    const result0 = taskUtils.orderByDueDateAndPriority(
       { _id: 'task1', dueDate: '2025-05-30', state: 'Ready', overdue: true },
-      { _id: 'task2', dueDate: 1748638800000, state: 'Ready', overdue: true },
+      { _id: 'task2', dueDate: '2025-05-31', state: 'Ready', overdue: true },
+    );
+    expect(result0).to.be.equal(-1000 * 60 * 60 * 24);
+
+    const result1 = taskUtils.orderByDueDateAndPriority(
+      { _id: 'task1', dueDate: '1748552400000', state: 'Ready', overdue: true },
+      { _id: 'task2', dueDate: '1748638800000', state: 'Ready', overdue: true },
     );
     expect(result1).to.be.equal(-1000 * 60 * 60 * 24);
 


webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  private async fetchNotifications(): Promise<Notification[]> {
+    try {
+      const today = moment().format('YYYY-MM-DD');
+      const taskDocs = await this.rulesEngineService.fetchTaskDocsForAllContacts();
+      let notifications: Notification[] = [];
+
+      for (const task of taskDocs) {
+        const dueDate = task.emission.dueDate;
+        if (dueDate <= today) {
```
jkuester: So, if we only are pushing notifications from today, then the data cached in cht-android will not contain tasks for tomorrow, right?

Then, when tomorrow comes, the background worker will not have any tasks data for that day to create notifications from (until the user opens the app again and gets new task data pushed to the app).  

I thought we were going to try and push future tasks to the Android app  as well so the background worker could push notifications without the user needing to open the app every day.  Maybe I am missing something here... 🤔

jonathanbataire: i see, question is how much can/should we store in android datastore?
also a task due today almost always has an end date a couple days ahead so notifications will still run for that period not necessarily have to open app everyday
opening the app will update the store with latest tasks to keep the cycle going
idk, my biggest concern is to limit how much is stored in datastore as its not a db so trying to treat it as such

dianabarsan: I think I agree with @jkuester here, the point was to preload visible tasks so that the application can notify even if the webview is not started. 
I don't know enough about datastore, but maybe instead of separating by due date we could set a limit of maximum tasks to store.

jkuester: Since the task array gets serialized into Android as a String JSON value, Jonathan has kept things super simple on the Datastore side and just stored that whole string as the tasks value.  Then, when "checking" for notifications on the periodic interval, we de-serialize the contents of the string and look for any tasks that should be notified.

IMHO we can keep things simple and just return tasks here up to a fixed point in the future (maybe 1 week?).  If we go too large, the serialization/de-serialization alone will start to get expensive.

dianabarsan: I agree it's either a time interval or a number, say 100 tasks maximum?

jkuester: Ohh, yeah! I like the max number of tasks better.  That lets us get the most value possible from these notifications.

jonathanbataire: i'll use max tasks

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,128 @@
...
+    const settings = await this.settingsService.get();
+    if (settings?.tasks?.max_task_notifications) {
+      return settings.tasks.max_task_notifications;
+    }
+    console.warn('Invalid or missing max_task_notifications setting, using default value');
+    return DEFAULT_MAX_NOTIFICATIONS;
+  }
+
+  private translateContentText(taskName: string, contact: string, dueDate: string): string {
+    const key = 'android.notification.tasks.contentText';
```
jkuester, line 114: Should we have a default translation value configured for this key in `api/resources/translations/messages-en.properties`?  Can follow our documented process for [adding a new translation key](https://docs.communityhealthtoolkit.org/community/contributing/translations/#adding-new-keys).

webapp/tests/karma/ts/app.component.spec.ts
```diff
@@ -297,6 +301,13 @@ describe('AppComponent', () => {
     expect(storageInfoService.init.callCount).to.equal(1);
   });
 
+  it('should init task notifications on android', async () => {
+    window.medicmobile_android = { updateTaskNotificationStore: sinon.stub().returns(true) };
```
jkuester, line 309: Setting this global here means that this might leak into other test runs (since we are not resetting this value after the test).  One way other tests handle this is to just [save off the original value in the beforeEach](https://github.com/medic/cht-core/blob/e21d8f8b0d1456ca5534ef19a944186b2e195a84/webapp/tests/karma/ts/services/android-app-launcher.service.spec.ts#L16) and then reset it in the afterEach.

webapp/tests/karma/ts/services/deduplicate.service.spec.ts
```diff
@@ -22,7 +22,7 @@ const SIBLINGS = [
   {
     _id: 'sib1',
     name: 'Test1',
-    date_of_birth: '2000-01-31',
+    date_of_birth: '2000-01-02',
```
jkuester:
```suggestion
    date_of_birth: '2000-01-31',
```

This test would still fail on Jan 1.  We can roll back these changes anyway since this suite has been fixed now in `master`.

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,128 @@
...
+    const readyState = stateHistory.find(state => state.state === 'Ready');
+    return readyState ? readyState.timestamp : 0;
+  }
+
+  private async getMaxNotificationSettings(): Promise<number> {
+    const settings = await this.settingsService.get();
+    if (settings?.tasks?.max_task_notifications) {
+      return settings.tasks.max_task_notifications;
+    }
+    console.warn('Invalid or missing max_task_notifications setting, using default value');
```
jkuester: IMHO we should remove this warning or just have it be a `debug`.  Since we actually have a default value here (and so the config value is _optional_), it seems unnecessary for this to be a `warn` call.

webapp/tests/karma/ts/app.component.spec.ts
```diff
@@ -297,6 +301,13 @@ describe('AppComponent', () => {
...
+    await getComponent();
+    await component.setupPromise;
+    expect(tasksNotificationService.initOnAndroid.callCount).to.equal(1);
```
jkuester, line 312: Minor, but could you add an assertion to `should create component and init services` to verify that `initOnAndriod` is not called?

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,128 @@
...
+
+@Injectable({
+  providedIn: 'root'
+})
+export class TasksNotificationService implements OnDestroy {
+
+  private debouncedReload;
+  private canGetNotifications;
+  private isRulesEngineEnabled;
+  subscription = new Subscription();
```
jkuester:
```suggestion
  private readonly subscription = new Subscription();

```

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,128 @@
...
+    this.subscribeToRulesEngine();
+    this.updateAndroidStore();
+  }
+
+  async get(): Promise<Notification[]> {
+    if (!await this.isEnabled()) {
+      return [];
+    }
+    return await this.fetchNotifications();
+  }
```
jkuester: Is this function every used?

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  private async isEnabled() {
+    this.isRulesEngineEnabled = await this.rulesEngineService.isEnabled();
+    this.canGetNotifications = await this.authService.has('can_get_task_notifications');
```
jkuester: Do we need to be saving these as class-level variables?  It looks like we are not referencing them outside this method.

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  ngOnDestroy(): void {
+    this.subscription.unsubscribe();
+  }
+
+  async initOnAndroid() {
+    if (!await this.isEnabled()) {
+      return;
+    }
+    this.debouncedReload = _debounce(this.updateAndroidStore.bind(this), 1000, { maxWait: 10 * 1000 });
```
jkuester: Minor, but I think `debouncedReload` could be a local variable in `subscribeToRulesEngine` instead of having to be set at the class-level.

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  private getReadyStateTimestamp(stateHistory): number {
+    const readyState = stateHistory.find(state => state.state === 'Ready');
+    return readyState ? readyState.timestamp : 0;
+  }
+
+  private async getMaxNotificationSettings(): Promise<number> {
+    const settings = await this.settingsService.get();
+    if (settings?.tasks?.max_task_notifications) {
+      return settings.tasks.max_task_notifications;
```
jkuester, line 108: Do we have a specific need right now for this value to be configurable?  I am having trouble understanding why it might be important for partners to be able to configure this.

If we have no clear use case, I think we should follow YAGNI and not include this for now. (We can always add the configuration in later if we need it.)

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,128 @@
...
+          contentText: this.translateContentText(
+            task.emission.title,
+            task.emission.contact.name,
+            task.emission.dueDate
+          ),
+          endDate: moment(task.emission.endDate).valueOf(),
+          dueDate: moment(task.emission.dueDate).valueOf()
+        });
+      }
+      notifications = notifications.sort((a, b) => b.readyAt - a.readyAt);
```
jkuester: Okay, since I did not get a solid answer about this out on my [cht-android thread](https://github.com/medic/cht-android/pull/414#discussion_r2550190909), I will repeat the question here:

The `readyAt` value seems like a really poor approximation for "priority". The user does not really care when the task was created. Instead, I think they will care most about stuff like "which task is the most overdue".  Ultimately it seems like we have 4 options (or some combination of them):

1. "Most recently due" - Order the tasks by `dueDate (desc)` so the user sees the "newest" due tasks (the ones that most recently have become due).
2. "Most overdue" - Order the tasks by the `dueDate (asc)` so the user sees the tasks that have been due the longest.
3. "Ending soonest" - Order the tasks by `endDate (asc)` so the user sees the tasks that will disappear soonest and become unavailable.
4. "Most recent task emission" (current) - Order the tasks by the `readyAt (desc)` so the user sees the dates that were most recently emitted (the action that triggered the task occurred most recently).


I am open to hearing any other ideas as far as which ordering would make the most sense for end users.  My suggestion would be `dueDate (asc)` then `readyAt (desc)`. It seems more useful to a user to see _newly due tasks_ for the day vs tasks they might have already seen from previous days.

webapp/tests/karma/ts/services/task-notification.service.spec.ts
```diff
@@ -0,0 +1,271 @@
...
+  let settingsService;
+  let authService;
+
+  let tasks;
+  let clock;
+  const getTask = (id: string) => tasks.find((task) => task._id === id);
+
+  beforeEach(() => {
+    consoleErrorMock = sinon.stub(console, 'error');
+    clock = sinon.useFakeTimers(moment().valueOf());
```
jkuester: Do we need to use this clock here? I don't see us every _changing_ the time in an of the tests and otherwise it seems like all we are doing is just mocking the time to be "now"....

webapp/tests/karma/ts/services/task-notification.service.spec.ts
jkuester, line 1: If possible, can we include tests for `initOnAndroid`?  Generally, I think it is not a good practice to call `private` functions directly from the tests anyways, so it would be best to test as much as possible via `initOnAndroid`...

jkuester (changes requested):
Okay, I was waiting for Diana to finish her review of this (since she had done the initial review of the v1 PR), but since she is not back until next week, I decided to try to keep things moving here. 👍 

If there is something I have raised that was already discussed/resolved in an early convo, please just link me to that and we don't need to re-hash it again.

webapp/src/ts/services/task-notifications.service.ts
jonathanbataire: had this convo with Diana about this in the previous PR 
yes there is need for it to be configurable for workflow design purposes
some people may prefer having less notifications shown to users for reasons

webapp/src/ts/services/task-notifications.service.ts
jonathanbataire: there is no "prioritization" going on here notifications simply come in as tasks are created hence the sorting
i don't think we should tackle prioritization here given there are other task prioritization criteria in CHT
(not sure if task is already deployed or not but i know it exists)
so we may need to consider that rather than creating one here

webapp/src/ts/services/task-notifications.service.ts
jonathanbataire: i think i tried this and linting fails cause it requires translations for other languages as well

webapp/src/ts/services/task-notifications.service.ts
jonathanbataire: its there for extensibility

webapp/src/ts/services/task-notifications.service.ts
jkuester: Right. This is expected. That is why we need to follow the process I linked to and get the other translations as well.

webapp/src/ts/services/task-notifications.service.ts
jkuester: IMHO we can remove it for now and add it back in later if we ever need it.

webapp/src/ts/services/task-notifications.service.ts
jkuester: Got it. :+1: We just need to make sure to include this settings value in the documentation for this feature in cht-docs.

webapp/src/ts/services/task-notifications.service.ts
jkuester: :shrug: If this is not "prioritization" then why bother sorting them at all here?

> i don't think we should tackle prioritization here given there are other task prioritization criteria in CHT

The tasks displayed to the user on the tasks tab are [already being sorted](https://github.com/medic/cht-core/blob/bed4546523931f90dc92c93bc6e55a1c3a3de495/webapp/src/ts/reducers/tasks.ts#L28) by `Priority (desc)` and `dueDate (asc)`.  Maybe we should follow the same ordering here? :thinking:

jonathanbataire: the reason for this ordering, it's easy to keep track of notifications; store a timestamp to the last notification and trigger anything after it

so if tasks are prioritized like that by default this here is just trying to get the latest among them
generally making it easy to trigger new task notifications as tasks come in

jonathanbataire: using due date means triggering notifications as tasks come in during the day will be skewed

webapp/src/ts/services/task-notifications.service.ts
jonathanbataire: added https://forum.communityhealthtoolkit.org/t/translation-support-request-task-notifications/5477

webapp/src/ts/services/task-notifications.service.ts
jonathanbataire: @jkuester @dianabarsan i want to revert back to only storing tasks that are due or past due
the due date translation stored in the android store can become stale 

example below task was not ready when it got stored, the next day it got ready but due date(**1 day left**) is not valid anymore

AISelect_20260113_075256_Settings

jonathanbataire: or ditch the due date all together

jkuester: Woah, good catch!  I think it is more important to support background tasks (by also storing future tasks) than it is to leverage the `formatDateService.relative` logic here to pre-format the date.  

My vote would be to not include the due date at all on the `android.notification.tasks.contentText` message. If the notification is being shown it means that the task is (over)due.  The user can go to the task list for more info.  (And we can always get more fancy with this in future iterations if needed.)

webapp/src/ts/services/task-notifications.service.ts
jkuester: > using due date means triggering notifications as tasks come in during the day will be skewed

I don't think changing the ordering of the tasks here should affect _how/when_ the task notifications are triggered. In the cht-android logic we would still want to filter the tasks to notify the user on based on the current `isValidNotification` logic.  The only thing user-facing impact of changing the order of these tasks is regarding which tasks the user will be notified about _when there are more than 8 tasks to notify for._

Take a case where the user has the following 10 tasks that are all due on the same day:

| Name | readyAt | priority |
| --- | --- | --- |
| Routine Visit | 2025-12-12 | 1 |
| Routine Visit | 2025-12-11 | 1 |
| Routine Visit | 2025-12-10 | 1 |
| Routine Visit | 2025-12-9 | 1 |
| Routine Visit | 2025-12-8 | 1 |
| Routine Visit | 2025-12-7 | 1 |
| Routine Visit | 2025-12-6 | 1 |
| Routine Visit | 2025-12-5 | 1 |
| Critical Visit | 2025-12-4 | 9 |
| Critical Visit | 2025-12-3 | 9 |

In this case if we order by `readyAt (desc)`, then the user will only see notifications for low priority `Routine Visit` notifications. (And will not get any notifications for the high priority `Critical Visit` tasks.)

---

I feel like we just keep going in circles with this discussion.  Ultimately, I am fine with shipping the cht-core changes with the current `readyAt (desc)` ordering.  However, I still think that the cht-android logic needs to be updated _to not depend on this ordering._ We should not have to ship a cht-android update just to change which notifications the user gets shown.

jonathanbataire: okay this should be sorted now

shared-libs/task-utils/src/task-utils.js
```diff
@@ -42,5 +122,6 @@ module.exports = {
    * @param {String} [gatewayRef] The gateway ID
    * @returns {boolean} Returns true if task state or task history is changed, otherwise returns false.
    */
-  setTaskState: (task, state, details, gatewayRef) => setTaskState(task, state, details, gatewayRef)
+  setTaskState: (task, state, details, gatewayRef) => setTaskState(task, state, details, gatewayRef),
+  orderByDueDateAndPriority: (t1, t2) => orderByDueDateAndPriority(t1, t2)
```
jkuester, line 119: I know this logic was just copied from the tasks reducer, but for the sake of future maintainability, can you add some unit tests here in shared-libs/task-utils?

shared-libs/task-utils/src/task-utils.js
```diff
@@ -33,6 +34,85 @@ const setTaskState = function(task, state, details, gatewayRef) {
...
+        return numericDate;
+      }
+
+      // If not a number, try parsing as date
+      if (moment(dueDate).isValid()) {
+        return moment(dueDate).valueOf();
+      }
+    }
+    return NaN;
+  };
```
jkuester, line 58: Sonar is complaining about the complexity here. I think we can simplify this function to:

```suggestion
  const getDueDate = dueDate => {
    const numericDate = Number(dueDate);
    if (!Number.isNaN(numericDate)) {
      return numericDate;
    }

    // If not a number, try parsing as date
    if (typeof dueDate === 'string' && moment(dueDate).isValid()) {
      return moment(dueDate).valueOf();
    }
    return Number.NaN;
  };
```

shared-libs/task-utils/src/task-utils.js
```diff
@@ -33,6 +34,85 @@ const setTaskState = function(task, state, details, gatewayRef) {
...
+
+  const getPriorityValue = priority => {
+    if (typeof priority === 'number' && priority >= 0) {
+      return priority;
+    }
+    return NaN;
```
jkuester: From Sonar:

```suggestion
    return Number.NaN;
```

shared-libs/task-utils/src/task-utils.js
```diff
@@ -33,6 +34,85 @@ const setTaskState = function(task, state, details, gatewayRef) {
...
+  };
+
+  const lhsDate = getDueDate(t1?.dueDate);
+  const rhsDate = getDueDate(t2?.dueDate);
+  const lhsPriority = getPriorityValue(t1?.priority);
+  const rhsPriority = getPriorityValue(t2?.priority);
+
+  const compareDates = () => {
+    // Both dates invalid, maintain original order
+    if (isNaN(lhsDate) && isNaN(rhsDate)) {
```
jkuester: From Sonar:

```suggestion
    if (Number.isNaN(lhsDate) && Number.isNaN(rhsDate)) {
```

shared-libs/task-utils/src/task-utils.js
```diff
@@ -33,6 +34,85 @@ const setTaskState = function(task, state, details, gatewayRef) {
...
+      return 0;
+    }
+    // Move tasks without dates to end
+    if (isNaN(lhsDate)) {
```
jkuester: From Sonar:

```suggestion
    if (Number.isNaN(lhsDate)) {
```

shared-libs/task-utils/src/task-utils.js
```diff
@@ -33,6 +34,85 @@ const setTaskState = function(task, state, details, gatewayRef) {
...
+      return 1;
+    }
+    if (isNaN(rhsDate)) {
```
jkuester: From Sonar:

```suggestion
    if (Number.isNaN(rhsDate)) {
```

shared-libs/task-utils/src/task-utils.js
```diff
@@ -33,6 +34,85 @@ const setTaskState = function(task, state, details, gatewayRef) {
...
+      return -1;
+    }
+    // Sort by date ascending
+    return lhsDate - rhsDate;
+  };
+
+  // Priority comparison cascade
+  if (isNaN(lhsPriority) && isNaN(rhsPriority)) {
```
jkuester: From Sonar:

```suggestion
  if (Number.isNaN(lhsPriority) && Number.isNaN(rhsPriority)) {
```

shared-libs/task-utils/src/task-utils.js
```diff
@@ -33,6 +34,85 @@ const setTaskState = function(task, state, details, gatewayRef) {
...
+    return compareDates(); // Both priorities invalid, sort by date
+  }
+
+  // Move tasks without valid priorities to end
+  if (isNaN(lhsPriority)) {
```
jkuester: From Sonar:

```suggestion
  if (Number.isNaN(lhsPriority)) {
```

shared-libs/task-utils/src/task-utils.js
```diff
@@ -33,6 +34,85 @@ const setTaskState = function(task, state, details, gatewayRef) {
...
+    return 1;
+  }
+  if (isNaN(rhsPriority)) {
```
jkuester: From Sonar:

```suggestion
  if (Number.isNaN(rhsPriority)) {
```

webapp/src/ts/app.component.ts
```diff
@@ -333,6 +335,13 @@ export class AppComponent implements OnInit, AfterViewInit {
     this.requestPersistentStorage();
     this.startWealthQuintiles();
     this.initAnalyticsModules();
+    this.initAndroidTaskNotifications();
+  }
+
+  private initAndroidTaskNotifications() {
+    if (typeof window.medicmobile_android?.updateTaskNotificationStore === 'function') {
```
jkuester: From Sonar:

```suggestion
    if (typeof globalThis.medicmobile_android?.updateTaskNotificationStore === 'function') {
```

Note that for this to not break the ESLinting, I think we need to update [the lint config](https://github.com/medic/cht-core/blob/310a77a4ad9daffc96a004babd43c99ba7a79215/eslint.config.js#L430) for these TS files.  For some reason the ECMA version getting used for these TS files is old (`5`). I did a quick test locally, and I think we can set this to `2020` and it will allow `globalThis` without generating a ton of other linting errors.  

If you try flipping this and see something different (e.g. a ton of other linting issues with existing files) then we should probably look for a different approach, but if possible this seems like a simple way to keep moving forwards.

webapp/src/ts/services/task-notifications.service.ts
```diff
@@ -0,0 +1,118 @@
...
+      debouncedReload.cancel();
+      return debouncedReload();
+    });
+    this.subscription.add(rulesEngineSubscription);
+  }
+
+  private async updateAndroidStore(): Promise<void> {
+    const notifications = await this.fetchNotifications();
+    const maxNotifications = await this.getMaxNotificationSettings();
+    window.medicmobile_android?.updateTaskNotificationStore(JSON.stringify(notifications), maxNotifications);
```
jkuester: From Sonar. (See my other comment about eslint errors with `globalThis`.)

```suggestion
    globalThis.medicmobile_android?.updateTaskNotificationStore(JSON.stringify(notifications), maxNotifications);
```

shared-libs/task-utils/src/task-utils.js
jonathanbataire: this is tested indirectly here https://github.com/medic/cht-core/blob/310a77a4ad9daffc96a004babd43c99ba7a79215/webapp/tests/karma/ts/reducers/tasks.spec.ts#L237
not sure it makes sense to test in utils again
do we move the tests?

jonathanbataire: i'll move the tests

shared-libs/task-utils/src/task-utils.js
jonathanbataire: this fails tests

webapp/src/ts/app.component.ts
jonathanbataire: i think ive tried globalThis i dont think it works well with cht-android but can try again

shared-libs/task-utils/src/task-utils.js
jonathanbataire: on second thought maybe leave it as is tests that side test state

webapp/src/ts/app.component.ts
jonathanbataire: okay worked

shared-libs/task-utils/src/task-utils.js
jkuester: We do need some tests for the task reducer in webapp, but if we reuse this code elsewhere then we should have dedicated tests here in the shared-libs/task-utils that covers the logic we have here.

jkuester (changes requested):
^

PR #10365, opened 2025-10-09
fix(#10298): update device_by_user to use built-in reduce [DO NOT MERGE]

### LINKED ISSUE #10298: More easily allow user-devices API  to be called
Right now if you call the `/api/v2/export/user-devices` [API](https://docs.communityhealthtoolkit.org/building/reference/api/#get-apiv2exportuser-devices), it can take a very long time on very busy/very large instances.  If it takes more than 3600 seconds, it will timeout because of built in [maximums](https://github.com/medic/cht-core/blob/b907fd210389e839b6c86acf6c1d92fda42a31c5/nginx/nginx.conf#L13). 

An easy way to call the `user-devices`.  This could possibly be extending the time out for this one API path or improving the performance of the API to take <1hr or run in a batch mode and be notified when the process is done or other easy to use solution!

A work around was found to log into the VM where the CHT instance is running and make the call directly against the API container, bypassing `nginx`.  This looked something like:

1. ssh to production server
2. find IP of API instance `docker inspect  --format "{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}" $(docker ps -q --filter="name=api")`
3. call `tmux` 
4. using the IP from prior command and the production `medic` password, call the API via `curl` and save the output to a file: `curl -qs http://medic:<password>@172.18.0.5:5988/api/v2/export/user-devices > user-devices.json`
5.  disconnect from `tmux` with `ctrl + b` then `d` and disconnect from SSH
6. wait however many hours you need to 
7. ssh back into production server, connect to existing `tmux` with `tmux a`
8. if the process is still running, go to step 5
9. if the process is not running exit `tmux` and you're done 

**Additional context**
A national deployment had this happen on 5 instances and had to use the above work around.

dianabarsan:
# Description

Updates `device_by_user` reduce function to be built-in `_last`. 

Performance results, querying database with 400.000 telemetry documents:

| branch | time | improvement | 
| --- | --- | ---- | 
| master | 1144s | |
| proposed | 33s | 34 times faster | 

Unfortunately, the new built-in reduce [`_last`](https://docs.couchdb.org/en/stable/ddocs/ddocs.html#last) was added in CouchDb 3.5, which means that upgrades from older versions of CHT, that use CouchDb 3.4 will fail. 

closes #10298

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10298-users-devices-migration/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10298-users-devices-migration/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10298-users-devices-migration/docker-compose/cht-couchdb-clustered.yml)

# License


dianabarsan:
For the record, this is the error that we expect when trying to upgrade from a version that does not use CouchDb 3.5 (any cht < 4.21): 

```
2025-10-09 13:08:52 [32mINFO[39m: Saving ddocs for medic-users-meta 
2025-10-09 13:08:52 [31mERROR[39m: Error thrown while installing: Error: Error while saving docs: saving _design/:staged:users-meta failed with invalid_design_doc
    at Object.module.exports.saveDocs (/api/src/db.js:164:11)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Object.saveStagedDdocs (/api/src/services/setup/utils.js:258:5)
    at async Object.stage (/api/src/services/setup/upgrade-steps.js:92:3)
    at async safeInstall (/api/src/services/setup/upgrade.js:40:5) {
  [stack]: 'Error: Error while saving docs: saving _design/:staged:users-meta failed with invalid_design_doc\n' +
    '    at Object.module.exports.saveDocs (/api/src/db.js:164:11)\n' +
    '    at processTicksAndRejections (node:internal/process/task_queues:96:5)\n' +
    '    at async Object.saveStagedDdocs (/api/src/services/setup/utils.js:258:5)\n' +
    '    at async Object.stage (/api/src/services/setup/upgrade-steps.js:92:3)\n' +
    '    at async safeInstall (/api/src/services/setup/upgrade.js:40:5)',
  [message]: 'Error while saving docs: saving _design/:staged:users-meta failed with invalid_design_doc'
} 
```

This error would occur immediately after hitting the upgrade/stage button, while API tries to start indexing the new version's views.

sugat009:
@dianabarsan what is the workaround for this?

dianabarsan:
That deployments update to v 4.21 before the version with this fix.

PR #10371, opened 2025-10-11
feat(#10332): add targets filter selection telemetry

### LINKED ISSUE #10332: Collect telemetry when user filters targets by previous month
Currently in `webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.ts` we log the telemetry with `this.telemetryService.record('sidebar_filter:analytics:target_aggregates:open');`.  This tells us when the filter has been _opened_ but does not indicate the user actually selected the Previous month.

We should probably collect another telemetry entry when the user actually filters by "Previous month".  Maybe something like: `sidebar_filter:analytics:previous_month`.

One interesting question is if we need to worry about distinguishing between filtering on the aggregate targets vs normal targets when it comes to collecting this telemetry data... 🤔 My initial thought here is that while the UI/features are similar, the motivations for the users to use the feature might be different enough that it would be valuable to have separate telemetry entries for each...  @Benmuiruri @paulpascal thoughts?

benkags:
# Description

Closes #10332 

Adds collection of telemetry when target sidebar filters are changed.

# Code review checklist

# License

### DIFF
--- a/api/tests/mocha/services/settings.spec.js
+++ b/api/tests/mocha/services/settings.spec.js
@@ -6,6 +6,7 @@ should();
 const service = require('../../../src/services/settings');
 const db = require('../../../src/db');
 const resources = require('../../../src/resources');
+// eslint-disable-next-line n/no-missing-require
 const defaults = require('../../../build/default-docs/settings.doc.json');
 const config = require('../../../src/config');
 


webapp/src/ts/modules/analytics/analytics-sidebar-filter.component.ts
```diff
@@ -22,6 +23,7 @@ export class AnalyticsSidebarFilterComponent implements OnInit, OnDestroy {
 
   @Input() userFacilities: Place.v1.Place[] = [];
   @Input() showFacilityFilter = true;
+  @Input() targetModuleId: string = 'target-aggregates';
```
jkuester: Is this value getting used for anything besides the telemetry? If not, could we maybe change the name to `telemetryKey` (or something similar) and the default value to `'target_aggregates'`?  

The reason is that we already have the telemetry entry key: `sidebar_filter:analytics:target_aggregates:open`.  I would like to try and have all the new keys just keep using `target_aggregates` instead of sometimes using `target-aggregates` (to avoid confusion in the future...)/.

webapp/src/ts/modules/analytics/analytics-sidebar-filter.component.ts
```diff
@@ -97,10 +100,16 @@ export class AnalyticsSidebarFilterComponent implements OnInit, OnDestroy {
...
     this.facilitySelectionChanged.emit(this.selectedFacility);
+    this.collectFilterSelectionTelemetry('facility');
   }
 
   fetchAggregateTargetsByReportingPeriod() {
     this.reportingPeriodSelectionChanged.emit(this.selectedReportingPeriod);
+    this.collectFilterSelectionTelemetry('reporting-period');
+  }
+
+  collectFilterSelectionTelemetry(filter) {
```
jkuester:
```suggestion
  private collectFilterSelectionTelemetry(filter) {
```

webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.ts
```diff
@@ -111,9 +111,20 @@ export class AnalyticsFilterComponent implements AfterContentInit, AfterContentC
...
     this.isOpen = !this.isOpen;
     this.globalActions.setSidebarFilter({ isOpen: this.isOpen });
-    if (this.isOpen) {
-      // Counting every time the user opens the sidebar filter in analytics_targets_aggregrate tab.
+
+    if (!this.isOpen) {
+      return;
+    }
+
+    const moduleId = this.activeModule? this.activeModule: this.getCurrentModuleId();
```
jkuester: Do we need to check the `activeModule` here or can we always just call `this.getCurrentModuleId()`?

webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.ts
jkuester, line 100: @Benmuiruri this is actually related to the changes you made in your PR: do we actually need this method anymore?  Looking at this again, when would this ever be false?  This `analytics-filter` component is going to shown on either the targets page or on the target aggregates page.  Either way this should be `true` right?

webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.ts
benkags: We are setting the active module from different triggers: router events, url subscription (see function `ngAfterContentInit`); so, checking `activeModule` first makes logical sense.

benkags: Not sure if `getCurrentModuleId` will always return what's been for `activeModule`.

webapp/src/ts/modules/analytics/analytics-sidebar-filter.component.ts
benkags: It's not being used for anything else. Updated.

webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.ts
benkags: From the `getCurrentModuleId` logic, `moduleId` can be null: `return this.route.snapshot?.firstChild?.data?.moduleId;` but this probably handled by ensuring by the time the component is rendered, the value is present. We should probably keep it.

webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.ts
jkuester: :+1: Got it. I think my eyes got crossed at some point trying to following the various triggers... :dizzy_face: But, either way it does seem safer to just check `activeModule`.

benkags:
@jkuester this is now ready for another round of review.

PR #10388, opened 2025-10-16
chore: fix demo and covid-19 config tests

jkuester:
# Description

At some point the tests broke for the `demo` and `covid-19` configs.  I still think we should [remove the demo config completely](https://forum.communityhealthtoolkit.org/t/proposal-to-delete-demo-configuration-from-cht-core-and-use-default-config-on-demo-instance/4691) (since it is still just a duplicate of `default`), but as long as both of these configs are still in the cht-core repo we should keep the tests running (and run them in the CI).  

To fix the `demo` tests, I literally just copied the forms and test updates from the `default` config.

For the `covid-19` tests, the problem was that the new version of the test-harness/cht-form/enketo broke the workaround that used to work for these forms that depended on input from the Android app integration.  There is no official support in the test-harness for filling out data from the Android app integration, but you used to be able to use the harness to set values on form pages _even when the question was readonly_.  Now this is not possible in newer versions of the test-harness.  Instead, I updated the tests to inject the mock data from the Android integration as part of the `content` initially loaded into the form.  This is a bit more technical/verbose, but IMHO it is actually more "correct" when it comes to testing the form logic...

# Code review checklist

# License

### DIFF
--- a/config/demo/test/targets/deaths.spec.js
+++ b/config/demo/test/targets/deaths.spec.js
@@ -22,16 +22,36 @@ describe('Death related targets tests', () => {
     await harness.setNow('2000-04-30');//DOD: 2000-04-25
     harness.subject = babyDeceasedAtAge1Day;
     const deathsThisMonth = await harness.getTargets({ type: 'deaths-this-month' });
-    expect(deathsThisMonth).to.have.property('length', 1);
-    expect(deathsThisMonth[0]).to.nested.include({ 'value.pass': 1, 'value.total': 1 });
+    expect(deathsThisMonth).to.deep.equal([{
+      goal: 0,
+      icon: 'icon-death-general',
+      id: 'deaths-this-month',
+      subtitle_translation_key: 'targets.this_month.subtitle',
+      translation_key: 'targets.death_reporting.deaths.title',
+      type: 'count',
+      value: {
+        pass: 1,
+        total: 1
+      }
+    }]);
   });
 
   it('death this month target test next month', async () => {
     await harness.setNow('2000-05-30');//DOD: 2000-04-25
     harness.subject = babyDeceasedAtAge1Day;
     const deathsThisMonth = await harness.getTargets({ type: 'deaths-this-month' });
-    expect(deathsThisMonth).to.have.property('length', 1);
-    expect(deathsThisMonth[0]).to.nested.not.include({ 'value.pass': 1, 'value.total': 1 });
+    expect(deathsThisMonth).to.deep.equal([{
+      goal: 0,
+      icon: 'icon-death-general',
+      id: 'deaths-this-month',
+      subtitle_translation_key: 'targets.this_month.subtitle',
+      translation_key: 'targets.death_reporting.deaths.title',
+      type: 'count',
+      value: {
+        pass: 0,
+        total: 0
+      }
+    }]);
   });
 
 });
--- a/config/default/test/targets/deaths.spec.js
+++ b/config/default/test/targets/deaths.spec.js
@@ -22,16 +22,36 @@ describe('Death related targets tests', () => {
     await harness.setNow('2000-04-30');//DOD: 2000-04-25
     harness.subject = babyDeceasedAtAge1Day;
     const deathsThisMonth = await harness.getTargets({ type: 'deaths-this-month' });
-    expect(deathsThisMonth).to.have.property('length', 1);
-    expect(deathsThisMonth[0]).to.nested.include({ 'value.pass': 1, 'value.total': 1 });
+    expect(deathsThisMonth).to.deep.equal([{
+      goal: 0,
+      icon: 'icon-death-general',
+      id: 'deaths-this-month',
+      subtitle_translation_key: 'targets.this_month.subtitle',
+      translation_key: 'targets.death_reporting.deaths.title',
+      type: 'count',
+      value: {
+        pass: 1,
+        total: 1
+      }
+    }]);
   });
 
   it('death this month target test next month', async () => {
     await harness.setNow('2000-05-30');//DOD: 2000-04-25
     harness.subject = babyDeceasedAtAge1Day;
     const deathsThisMonth = await harness.getTargets({ type: 'deaths-this-month' });
-    expect(deathsThisMonth).to.have.property('length', 1);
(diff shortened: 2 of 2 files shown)


config/demo/test/targets/births.spec.js
```diff
@@ -26,15 +26,15 @@ describe('Births this month target tests', () => {
 
   it('birth this month should be counted', async () => {
     //await harness.setNow('2000-04-30');//DOB: 2000-04-24
-    clock = sinon.useFakeTimers({now: moment('2000-04-30').toDate()});
+    clock = sinon.useFakeTimers({ now: moment('2000-04-30').toDate(), toFake: ['Date'] });
```
lorerod: Can we refactor this test to use the harness time management like the other tests (pregnancy and delivery)?
So we can remove `sinon`, `moment`, and `clock` dependencies also from this spec.

config/demo/test/targets/deaths.spec.js
```diff
@@ -25,7 +25,7 @@ describe('Death related targets tests', () => {
 
   it('death this month target test this month', async () => {
     //await harness.setNow('2000-04-30');//DOD: 2000-04-25
-    clock = sinon.useFakeTimers({now: moment('2000-04-30').toDate()});
+    clock = sinon.useFakeTimers({ now: moment('2000-04-30').toDate(), toFake: ['Date'] });
```
lorerod: Same here.
Can we refactor this test to use the harness time management like the other tests?

config/demo/test/targets/births.spec.js
```diff
@@ -26,15 +26,15 @@ describe('Births this month target tests', () => {
 
   it('birth this month should be counted', async () => {
     //await harness.setNow('2000-04-30');//DOB: 2000-04-24
```
lorerod: Can you please remove this comment?

config/demo/test/targets/births.spec.js
```diff
@@ -26,15 +26,15 @@ describe('Births this month target tests', () => {
...
-    clock = sinon.useFakeTimers({now: moment('2000-04-30').toDate()});
+    clock = sinon.useFakeTimers({ now: moment('2000-04-30').toDate(), toFake: ['Date'] });
     const birthsThisMonth = await harness.getTargets({ type: 'births-this-month' });
     expect(birthsThisMonth).to.have.property('length', 1);
     expect(birthsThisMonth[0]).to.nested.include({ 'value.pass': 1, 'value.total': 1 });
   });
 
   it('birth last month should not be counted', async () => {
     //await harness.setNow('2000-04-30');//DOB: 2000-04-24
```
lorerod: Same here. Please remove this comment

config/demo/test/targets/deaths.spec.js
```diff
@@ -25,7 +25,7 @@ describe('Death related targets tests', () => {
 
   it('death this month target test this month', async () => {
     //await harness.setNow('2000-04-30');//DOD: 2000-04-25
```
lorerod: Please remove the comment.

config/demo/test/targets/deaths.spec.js
```diff
@@ -34,7 +34,7 @@ describe('Death related targets tests', () => {
 
   it('death this month target test next month', async () => {
     //await harness.setNow('2000-04-30');//DOD: 2000-04-25
```
lorerod: Please remove comment

config/demo/test/targets/deaths.spec.js
lorerod: Can you please rename this variable to deathsThisMonth?

lorerod: Same here. Rename variable to deathsThisMonth.

config/demo/test/tasks/pnc_danger_sign_baby.spec.js
lorerod: Please remove comment.

config/demo/test/tasks/pnc_danger_sign_mother.spec.js
lorerod: Please remove commented code.

lorerod: Please remove commented code.

lorerod: Same here, please.

config/demo/test/tasks/pnc_danger_sign_baby.spec.js
```diff
@@ -19,8 +19,6 @@ describe('PNC danger sign follow up for baby tests', () => {
   beforeEach(async () => {
     await harness.clear();
     await harness.setNow(now);
-    //clock = sinon.useFakeTimers(moment('1999-10-10').add(day, 'days').toDate());
-    //await harness.flush(1);
     return await harness.loadForm('pnc_danger_sign_follow_up_baby');
```
lorerod: Is this needed?

config/demo/test/tasks/pnc_danger_sign_mother.spec.js
```diff
@@ -17,8 +17,6 @@ describe('PNC danger sign follow up for mother tests', () => {
   beforeEach(async () => {
     await harness.clear();
     await harness.setNow(now);
-    //clock = sinon.useFakeTimers(moment('1999-10-10').add(day, 'days').toDate());
-    //await harness.flush(1);
     return await harness.loadForm('delivery');
```
lorerod: Is this needed?

config/demo/test/tasks/pnc_danger_sign_baby.spec.js
jkuester: 🤔 Hmm, no.  Things seem to run fine without it.  I think this is getting overridden anyway by the `harness.loadAction` call below (which should open the form via the task action).

config/demo/test/tasks/pnc_danger_sign_mother.spec.js
jkuester: Amusingly, this was technically needed since we were not providing the form name when calling `fillForm` for the first time in the tests below.  However, `loadForm` is deprecated and confusing, so I have updated these tests to just specify the form to load when calling `fillForm`. 👍

jkuester:
Thanks for the review @lorerod!  I have pushed the changes both to the demo config and also to the default. :+1:

PR #10396, opened 2025-10-20
fix(#10384): changing changes to all docs

### LINKED ISSUE #10384: Using changes feed with _doc_ids filter causes performance issues.
## Summary
When fetching documents from purging databases during replication and purging, the current implementation uses the `_changes` feed with a `_doc_ids` filter. This causes severe performance degradation when purging databases are large.

The purging databases [are queried](https://github.com/medic/cht-core/blob/e9a0575e590dd54b559a3fc6210a8f1d92e30836/api/src/services/purged-docs.js#L62) [during replication](https://github.com/medic/cht-core/blob/e9a0575e590dd54b559a3fc6210a8f1d92e30836/api/src/services/replication.js#L13), with all document ids that the user has access to.
And [during purging](https://github.com/medic/cht-core/blob/e9a0575e590dd54b559a3fc6210a8f1d92e30836/sentinel/src/lib/purging.js#L123), with all ids matching a purging configuration.

In both cases, the intention is to get all documents from the purging database that match those ids from the medic database. 

### Current Implementation (changes feed)
The `_changes` endpoint with `_doc_ids` filter:
1. Gets every change in the purging database from sequence 0 to the latest change.
2. For each change, filters whether the change should be included in the result (whether its id is in the doc_ids filter list)
3. Performs a **linear search** using `lists:member/2` comparing each change against each doc id in the filter list

This has complexity: **O(n × m)** where n = number of changes, m = number of filter ids

### Real-World Impact
This behavior has been directly observed in a large production deployment.
For a typical sample user:
- They had access to 50,000 documents.
- There were 60,000,000 documents in a purging database.
- Each replication request for that user requires 3,000,000,000,000 comparisons

There is a purged docs cache so that not every replication request requires this, only when the cache cannot be used.
Purging tasks triggered by sentinel have the same pattern with batch_size * number of changes comparisons.

Affected production instances were monitoring using the erlang process list; it was observed that over half of all reductions (a measure of CPU usage) for sampled instances were spent only running those linear searches with `lists:member/2`. 
And up to 60 requests to the changes feed for purging documents were observed running concurrently.
This causes couch to not have enough CPU resources to serve other requests.

### Alternative Implementation (_all_docs)
Using `_all_docs` with `keys` parameter performs a btree lookup for each document id.

This has complexity: **O(m × log n)** where m = number of ids, n = database size

Even with tens of thousands of ids, this scales logarithmically with database size rather than linearly.

## Recommendation

Revert to using `_all_docs` with `keys` parameter for batch document retrieval from purging databases

witash:
closes #10384 

allDocs returns 'rows' instead of results and 'deleted' is in `value` instead of at the top level.
so the diff is mostly just find/replace in the tests for changes->allDocs and for those format changes.

### DIFF
--- a/sentinel/src/lib/purging.js
+++ b/sentinel/src/lib/purging.js
@@ -118,12 +118,11 @@ const getAlreadyPurgedDocs = (roleHashes, ids) => {
     .then(results => {
       results.forEach((result, idx) => {
         const hash = roleHashes[idx];
-        result.rows.forEach(row => {
         for (const row of result.rows) {
           if (row.value && !row.value.deleted) {
             purged[hash][serverSidePurgeUtils.extractId(row.id)] = row.value.rev;
           }
-        });
+        }
       });
 
       return purged;


mrjones-plip:
Differing to Diana on this one!

PR #10397, opened 2025-10-21
chore(#10322): view cleanup

witash:
this PR removes medic-scripts ddoc
and medic-client/total_clinics_by_facility


ddocs/medic-db/medic/views/contacts_by_depth/map.js
```diff
@@ -8,7 +8,6 @@ function(doc) {
     var depth = 0;
     while (parent) {
       if (parent._id) {
-        emit([parent._id], value);
```
dianabarsan, line 11: This is still required when a user with no replication depth is replicating.

ddocs/medic-db/medic/views/doc_summaries_by_id/map.js
```diff
@@ -1,95 +0,0 @@
-// WARNING: This is a copy of the GetSummaries service
-// with some minor modifications and needs to be kept in sync until
-// this workaround is no longer needed.
-// https://github.com/medic/medic/issues/4666
-
-function(doc) {
```
dianabarsan, line 6: This is still used here: https://github.com/medic/cht-core/blob/a2336490fac86f085f6b717a89e181d52956c5dd/admin/src/js/services/get-summaries.js#L103

ddocs/medic-db/medic/views/contacts_by_depth/map.js
witash: removed

ddocs/medic-db/medic/views/doc_summaries_by_id/map.js
witash: removed

witash:
@dianabarsan I think this should be ready to review and merge now

PR #10399, opened 2025-10-22
fix(#10182): needs_signoff and private string conversion

dianabarsan:
# Description

closes #10182

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:110183-fix-for-nouveau/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:110183-fix-for-nouveau/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:110183-fix-for-nouveau/docker-compose/cht-couchdb-clustered.yml)

# License


dianabarsan:
Hi @jonathanbataire 
We've made some significant changes to the docs_by_replication_key view - and it is now a Nouveau index. I cherry picked your changes here. Would you mind reviewing? Appreciate it!

ddocs/medic-db/medic/nouveau/docs_by_replication_key/index.js
```diff
@@ -69,14 +70,11 @@ function(doc) {
...
+    if (doc.fields && isTruthy(doc.fields.private)) {
+      indexMaybe('private', 'true');
     }
-    if (doc.fields &&
-        doc.fields.needs_signoff &&
-        doc.contact
-    ) {
-      indexMaybe('needs_signoff', true);
+    if (doc.fields && isTruthy(doc.fields.needs_signoff) && doc.contact) {
+      indexMaybe('needs_signoff', 'true');
```
jkuester, line 77: Minor, but I don't see why we needed to change the index to store a string value here.  We can always store a boolean value (and the auth service does not need to worry about strings).

dianabarsan: well, this tricked me to sooo many times while developing cinque! 
apparently you can't store booleans - the indexer doesn't error, it just doesn't index the while document. 

https://docs.couchdb.org/en/stable/ddocs/nouveau.html#ddoc-nouveau-field-types

dianabarsan: Oh, it looks like the documentation says that they should allow booleans, but they do not :| 
I guess it's time to submit bugs to Couch :| 

> This data must be only a string, number, or boolean. Other types will cause an error to be thrown by the index function call.

> If an error is thrown when running your function, for this reason or others, the document will not be added to that search index.

jkuester: :facepalm:

PR #10423, opened 2025-10-30
feat(#10140): show previous month targets

### LINKED ISSUE #10140: Add ability to display previous month's targets
Each month, community health workers (CHWs) submit reports to their supervisors and key stakeholders, including health center managers, village chiefs, and other community leaders. Presentation meetings are typically held between the 5th and 15th of the month, during which CHWs share the indicators from the previous reporting period.

The CHW’s “Analysis” tab displays the current month’s indicators. These indicators are reset to zero at the end of each month, making it difficult for CHWs to view and present data from the previous month using this tab. While it’s possible to access past indicators via the “Contact” tab, that section already contains a large volume of information, including the list of medications.

Introduce a toggle or dropdown within the “Analysis” tab that allows users to switch between “Current Month” and “Previous Month.”

**Additional context**
* [Forum post](https://forum.communityhealthtoolkit.org/t/enable-community-health-workers-to-view-last-months-statistics-in-the-analytics-tab/5122/3) discussing the need
* Forum post forming the squad [in English](https://forum.communityhealthtoolkit.org/t/chw-previous-month-in-app-targets-squad/5279) and [in French](https://forum.communityhealthtoolkit.org/t/squad-equipe-pour-mettre-en-place-la-fonctionnalite-dacces-des-indicateurs-du-mois-precedent-aux-asc/5281)

### LINKED ISSUE #10354: Target aggregates not loading correctly for online users
**Describe the bug**
Target aggregates are available for both online and offline users. 
It seems that for online users, they do not work as expected. 

**To Reproduce**
1. Assign your online user to a place and grant them `can_aggregate_targets` permissions. 
2. Update your configuration and make at least one target be an aggregate. 
3. Create a child place under your user's facility and add a primary contact. Create a user associated with this primary contact. 
4. Login with this user visit the targets page. Sync. 
5. Load the target aggregates page with the online user and notice that no values come through. 

**Expected behavior**
The target document that the offline user has just created should appear in the aggregate calculation of the online user. 

Some curious behavior is happening as well. Checking browser logs, it seems that the page tries to load target aggregates from 3 months ago, finds none and then makes the same request to get the same contacts of a contact type multiple times. 

**Environment**
- Instance: local
- App: Webapp
- Version: master but suspecting older versions being affected

### LINKED ISSUE #10525: Aggregate targets do not work when language is set to Nepali
**Describe the bug**
When the user language is set to Nepali, the aggregate targets always show 0.

**To Reproduce**

1. Add at least one target in the configuration that has `aggregate: true`
2. Make sure that the target counter goes up as CHW
3. Set language to English
4. Make sure that the aggregate target counter goes up as a supervisor
5. Change language to Nepali
6. See all the aggregate targets go back to 0

**Expected behavior**
I expected the aggregate targets to show up regardless of the langauge selected

**Logs**
No error in the console

**Screenshots**

Language set to Nepali:
Image

Language set to English:

Image

Language set to French:

Image

**Environment**
- Instance: local
- Browser: Chrome
- Client platform: Windows
- App: webapp
- Version: 4.15 ~ master

**Additional context**

When the targets are being fetched, the numerals in the keys are changed to Devanagari (script used for Nepali language).

Image

https://github.com/medic/cht-core/blob/bed4546523931f90dc92c93bc6e55a1c3a3de495/webapp/src/ts/services/target-aggregates.service.ts#L86-L90

jkuester:
# Description

Closes #10140 
Closes #10525
Closes #10354

Docs PR: https://github.com/medic/cht-docs/pull/2070

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10140_previous-month-targets/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10140_previous-month-targets/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10140_previous-month-targets/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/shared-libs/cht-datasource/src/qualifier.ts
+++ b/shared-libs/cht-datasource/src/qualifier.ts
@@ -14,7 +14,7 @@ export type IdQualifier = Readonly<{ id: string }>;
  */
 export const byId = (id: string): IdQualifier => {
   if (!isString(id) || id.length === 0) {
-    throw new InvalidArgumentError(`Invalid UUID [${JSON.stringify(id)}].`);
+    throw new InvalidArgumentError(`Invalid id [${JSON.stringify(id)}].`);
   }
   return { id };
 };
--- a/shared-libs/cht-datasource/test/qualifier.spec.ts
+++ b/shared-libs/cht-datasource/test/qualifier.spec.ts
@@ -14,11 +14,43 @@ import {
   isKeyedFreetextQualifier,
   isReportingPeriodQualifier,
   isUsernameQualifier,
-  isUuidQualifier
+  isUuidQualifier,
+  byId,
+  isIdQualifier
 } from '../src/qualifier';
 import { expect } from 'chai';
 
 describe('qualifier', () => {
+  describe('byId', () => {
+    it('builds a qualifier that identifies an entity by its id', () => {
+      expect(byId('uuid')).to.deep.equal({ id: 'uuid' });
+    });
+
+    [
+      null,
+      '',
+      { },
+    ].forEach(uuid => {
+      it(`throws an error for ${JSON.stringify(uuid)}`, () => {
+        expect(() => byId(uuid as string)).to.throw(`Invalid id [${JSON.stringify(uuid)}].`);
+      });
+    });
+  });
+
+  describe('isIdQualifier', () => {
+    [
+      [ null, false ],
+      [ 'uuid', false ],
+      [ { id: { } }, false ],
+      [ { id: 'uuid' }, true ],
+      [ { id: 'uuid', other: 'other' }, true ]
+    ].forEach(([ identifier, expected ]) => {
+      it(`evaluates ${JSON.stringify(identifier)}`, () => {
+        expect(isIdQualifier(identifier)).to.equal(expected);
+      });
+    });
+  });
+
   describe('byUuid', () => {
     it('builds a qualifier that identifies an entity by its UUID', () => {
       expect(byUuid('uuid')).to.deep.equal({ uuid: 'uuid' });


webapp/web-components/cht-form/tsconfig.spec.json
```diff
@@ -9,7 +9,9 @@
       "@mm-selectors/*": ["src/ts/selectors/*"],
       "@mm-pipes/*": ["src/ts/pipes/*"],
       "@mm-services/*": ["src/ts/services/*"],
+      "@mm-services/cht-datasource.service": ["web-components/cht-form/src/stubs/cht-datasource.service.ts"],
       "@mm-services/db.service": ["web-components/cht-form/src/stubs/db.service.ts"],
+      "@mm-services/language.service": ["web-components/cht-form/src/stubs/language.service.ts"]
```
jkuester, line 14: These changes just make these paths match what is already in the `tsconfig.app.json`. Previously these tests were not replacing all the stubbed services that they should have.

api/src/controllers/target.js
```diff
@@ -0,0 +1,55 @@
+const auth = require('../auth');
+const { TargetInterval, Qualifier } = require('@medic/cht-datasource');
+const ctx = require('../services/data-context');
+const serverUtils = require('../server-utils');
+const { PermissionError } = require('../errors');
+
+const getTargetInterval = ctx.bind(TargetInterval.v1.get);
+const getTargetIntervals = ctx.bind(TargetInterval.v1.getPage);
+
+const checkUserPermissions = async (req) => {
```
jkuester, line 10: @dianabarsan while doing my final checks here, I realized that we were not actually checking any permissions here...

For the `getAll`, do you think it makes sense to check of the existing `can_aggregate_targets` permission? (Since that is currently the only use-case we having for querying multiple target intervals like this.)

For the `get` (which is used for getting the previous month's target), I wonder if we should restrict (for non-admins) to only returning the target interval associated with the `contact_id` from the current user's `user-settings`?  Basically, you can only access _your_ target intervals.  🤷 Not sure if it is a huge deal to be able to access the intervals from other users, but I guess it would be better to be more restrictive, at least initially. Can always open it up more in the future.

I don't really want to add a new permission for `can_view_target_intervals` or something since existing users that are currently using targets would need to get this perm.

dianabarsan: Wait ... I am not even aware of this endpoint having been added. Did we really need this? 

anyway. I don't think this endpoint should be gated by permissions. The online user can have access to the docs anyway. The permission is only used to control frontend user experience, not to limit data access.

jkuester: Truly this project has been running too long.  First discussed the endpoints [on the forum](https://forum.communityhealthtoolkit.org/t/proposal-to-include-target-intervals-in-cht-datasource/5385).  

But, yeah, I agree with you that we don't need perms for this.

api/src/routing.js
```diff
@@ -446,6 +447,9 @@ app.get('/api/v1/forms{/}', forms.list);
 app.get('/api/v1/forms/:form', forms.get);
 app.post('/api/v1/forms/validate', textParser, forms.validate);
 
+app.get('/api/v1/target-interval', targetInterval.v1.getAll);
```
dianabarsan: I'm confused about the naming here. A target interval is a synonym for reporting period, so I'm expecting to an endpoint like this to return dates?
I suppose that a better name would be `api/v1/target`
Also, what does the default GET return? all target docs on the server? all target docs for the current interval?

shared-libs/cht-datasource/src/local/index.ts
```diff
@@ -2,4 +2,5 @@ export * as Contact from './contact';
 export * as Person from './person';
 export * as Place from './place';
 export * as Report from './report';
+export * as TargetInterval from './target-interval';
```
dianabarsan: I think `target-doc` or just `target` is a better naming fit. `target-interval` sounds like a time period. This applies throughout the PR.

shared-libs/cht-datasource/src/local/target.ts
```diff
@@ -0,0 +1,96 @@
...
+    const [, , contactUuid] = id.split('~');
+    return contactUuidSet.has(contactUuid);
+  });
+};
+
+/** @internal */
+export namespace v1 {
+  const isTargetInterval = (doc: Nullable<Doc>): doc is TargetInterval.v1.TargetInterval => {
+    return isRecord(doc)
+      && hasField(doc, { name: 'user', type: 'string' })
```
dianabarsan, line 46: I think we should primarily check the `type` field, which will be `target`.

shared-libs/cht-datasource/src/index.ts
```diff
@@ -321,6 +321,76 @@ export const getDatasource = (ctx: DataContext) => {
...
       },
-    },
+      targetInterval: {
+        /**
+         * Returns a target interval by UUID.
+         * @param uuid the UUID of the target interval to retrieve
+         * @returns the target interval or `null` if no target interval is found for the UUID
+         * @throws InvalidArgumentError if no UUID is provided
+         */
+        getByUuid: (uuid: string) => ctx.bind(TargetInterval.v1.get)(Qualifier.byUuid(uuid)),
```
dianabarsan: Is this retrieval by the uuid of the target doc itself? the uuid of the contact? if the first, is this any different from a standard db.get? 

Later edit: I realize now this is by _id. The `target` doesn't have a UUID as an _id, so this might be confusing.

shared-libs/cht-datasource/src/target-interval.ts
```diff
@@ -0,0 +1,163 @@
...
+  UsernameQualifier,
+  UuidQualifier
+} from './qualifier';
+import * as Local from './local';
+import * as Remote from './remote';
+import { InvalidArgumentError } from './libs/error';
+import { DEFAULT_DOCS_PAGE_LIMIT } from './libs/constants';
+import { assertCursor, assertLimit } from './libs/parameter-validators';
+
+const getTargetIntervalUuid = (
```
dianabarsan: I think we're overusing "uuid" here.

> A Universally Unique Identifier (UUID) is a 128-bit label used to uniquely identify information in computer systems without a central authority. Represented as a 36-character string (e.g., 123e4567-e89b-12d3-a456-426614174000), UUIDs are designed for high uniqueness, making them ideal for databases, distributed systems, and file identifiers. 

Targets don't have uuids, they have a unique compound _id that could contain a uuid - but that would be the contact document's uuid.

shared-libs/cht-datasource/test/local/target.spec.ts
```diff
@@ -0,0 +1,294 @@
...
+      let getDocByIdInner: SinonStub;
+
+      beforeEach(() => {
+        getDocByIdInner = sinon.stub();
+        getDocByIdOuter = sinon.stub(LocalDoc, 'getDocById').returns(getDocByIdInner);
+      });
+
+      it('returns a target interval by UUID', async () => {
+        const doc = {
+          user: 'user',
```
dianabarsan, line 36: I think this has been omitted a lot, but the way that we identify a target doc is primarily by it's type: https://github.com/medic/cht-core/blob/master/ddocs/medic-db/medic/nouveau/docs_by_replication_key/index.js#L90

shared-libs/cht-datasource/test/local/target.spec.ts
```diff
@@ -0,0 +1,294 @@
...
+        expect(getDocUuidsByIdRangeInner).to.have.been.calledOnceWithExactly(
+          `target~${qualifier.reportingPeriod}~`,
+          `target~${qualifier.reportingPeriod}~\ufff0`
+        );
+        expect(getDocsByIdsOuter).to.have.been.calledOnceWithExactly(localContext.medicDb);
+        expect(fetchAndFilterOuter).to.have.been.calledOnce;
+        expect(fetchAndFilterInner).to.have.been.calledOnceWithExactly(10, 0);
+
+        const [getFn, filterFn, limit] = fetchAndFilterOuter.getCall(0).args;
+        expect(limit).to.equal(10);
```
dianabarsan, line 210: I'm not following why the limit 10 here? why not return all results especially when we call with a list of contact ids?

I suppose I am asking because of the recent search query issues. Should we even have a built-in limit?

shared-libs/cht-datasource/test/local/target.spec.ts
```diff
@@ -0,0 +1,294 @@
...
+        user: 'user2',
+        owner: 'contact-2',
+        reporting_period: '2025-01',
+        updated_date: 124,
+        targets: []
+      };
+      const invalidTargetInterval = {
+        _id: 'target~2025-01~contact-2'
+      };
+      const qualifier = and(
```
dianabarsan: This variable got overloaded by some tests and it was hard to what the "global" qualifier was. Can this variable be renamed, and not overloaded?

webapp/src/ts/services/target-aggregates.service.ts
```diff
@@ -130,32 +77,40 @@ export class TargetAggregatesService {
...
+  private async getRelevantTargetDocs(intervalTag: string, contacts) {
+    if (!contacts.length) {
+      return [];
+    }
+    const qualifier = Qualifier.and(
+      byReportingPeriod(intervalTag),
+      byContactUuids(contacts.map(({ _id }) => _id))
+    );
+    const targetDocs: TargetInterval.v1.TargetInterval[] = [];
+    for await (const targetInt of this.getTargetIntervals(qualifier)) {
```
dianabarsan: how many records would this return if we don't pass a limit? would it just be 10?

webapp/tests/karma/ts/services/cht-datasource.service.spec.ts
```diff
@@ -205,6 +202,90 @@ describe('CHTScriptApiService service', () => {
     });
   });
 
+  describe('bindGenerator()', () => {
```
dianabarsan, line 205: So this one ... this one I don't like so much. but I don't have a suggestion on how to change it in this iteration. we can change it later with little consequence if any one of us becomes inspired.

webapp/tests/karma/ts/services/cht-datasource.service.spec.ts
```diff
@@ -205,6 +202,90 @@ describe('CHTScriptApiService service', () => {
...
+
+    [true, false].forEach((isOnlineOnly) => {
+      it(`binds to a data context when isOnlineOnly is ${isOnlineOnly}`, async () => {
+        sessionService.isOnlineOnly.returns(isOnlineOnly);
+        const strings = ['hello', 'world'];
+        const mockGenerator = async function* (input: string[]) {
+          for (const s of input) {
+            yield s;
+          }
+        };
```
dianabarsan, line 223: As cool as the generator model is, I'm not sure if it fits the db request scenario. 
I think the only endpoint it applies to would be _changes with a continuous feed. Otherwise, we expect to get a full result from the database snapshot of the present. 
I understand that it was used to get paged datasets, but even in those cases, we would likely want to control _when_ we request for more pages. 

Anyway, this is just some philosophical comment, and not something to be addressed now.

api/src/routing.js
jkuester: Discussed this more on a call with Diana. The `type` on the docs is `target`, so we are going to go with that as the api naming across the board. :+1:

shared-libs/cht-datasource/test/local/target.spec.ts
jkuester: The limit here is `10` because that is the param we passed above on line 197 when calling `getPage`.  The requested limit gets passed into the fetchAndFilter call so that it knows how many to retrieve.

> Should we even have a built-in limit?

I think we should either have a default limit or require that one is passed in. The Couch/Pouch pattern of support limit-less queries in some cases is absolute insanity and an anti-pattern of the wildest kind.

webapp/src/ts/services/target-aggregates.service.ts
jkuester: When using a generator like this, the limit will be the [default value](https://github.com/medic/cht-core/blob/5c215bc0e43d31aefc03428bf7673648c7ab2b0a/shared-libs/cht-datasource/src/target.ts#L127) for the associated `getPage` function.  In this case that is `10` (which is our default limit when getting the whole docs).  

The way the generator works, though is that the consumer gets one value (async) at a time. Under the hood, the generator is getting each "batch" of values as needed. So, after the consumer takes the first 10, then asks the generator for the 11th, the generator will go get the next batch of 10 and return the first of those.

webapp/tests/karma/ts/services/cht-datasource.service.spec.ts
jkuester: Yeah....  It is pretty jank, but I could not come up with anything I liked better. (Besides not using a programming language with an insane synchronization model...)

webapp/tests/karma/ts/services/cht-datasource.service.spec.ts
jkuester: Discussed this more with Diana as well.  TLDR, we agree that generator functions are something of an "advanced" (but still "standard") language feature in JS.  

We might agree to disagree on their benefits in the cht-datasource APIs. (They add complexity, but there are not too many other great options that do not force manual paging for all consuming code.)

We definitely agree that we need a way to abstract generator function _creation_ from our unit test code (so the unit tests do not need to have a bunch of `function*` everywhere).

shared-libs/cht-datasource/src/target-interval.ts
jkuester: I now think it was a mistake (on my part) to include `UUID` in the cht-datasource api naming at all.  I was trying to emphasis the generic "universally unique identifier" aspect, but was not considering that UUID has a precise technical definition (as you have noted) and that Couch document identifiers _do not have to be UUIDs_.  Additionally, using `UUID` in the cht-datasource api naming for things that are not _required_ to be actual UUIDs (even though in practice they usually/always are) is still unnecessarily leaking implementation details.  When something is just a unique identifier for an entity, we should not call it a "UUID".

I have updated all the new code added in this PR to avoid using `UUID` and I have added a new `IdQualifier`.  Eventually, I think it is best for the sake of consistency to deprecate the `UuidQualifier` in favor of `IdQualifier`. But for the sake of time in this PR, I have not done that here.

shared-libs/cht-datasource/test/utils.ts
jkuester, line 1: For the record, I did some research regarding sinon + generator functions.  The results are not great.  TLDR, there is no special support for them in sinon and none is likely to come any time soon.

https://github.com/sinonjs/sinon/issues/612
https://github.com/sinonjs/sinon/issues/1467

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -1,6 +1,33 @@
...
+
+/**
+ * Builds a qualifier that identifies an entity.
+ * @param id the identifier of the entity
+ * @returns the qualifier
+ * @throws InvalidArgumentError if the identifier is invalid
+ */
+export const byId = (id: string): IdQualifier => {
+  if (!isString(id) || id.length === 0) {
+    throw new InvalidArgumentError(`Invalid UUID [${JSON.stringify(id)}].`);
```
dianabarsan: `Invalid id` maybe?

PR #10432, opened 2025-11-03
feat(#10344): support targets by contact id in cht-datasource

### LINKED ISSUE #10344: Add support to cht-datasource for querying target intervals by contact ids
NOTE: This issues depends on functionality being implemented in https://github.com/medic/cht-core/issues/10343.

The existing [target aggregate](https://docs.communityhealthtoolkit.org/building/supervision/chw-aggregate-targets/) functionality needs to be able to load the target docs from the current reporting period for all "supervised" contacts.  

To support this functionality in cht-datasource (backed with a REST endpoint that _could_ be called by online users) we will need to change up the approach a bit.  Currently the target aggregates service [queries for the current target docs](https://github.com/medic/cht-core/blob/bb9277a0fbe97285e84e67bf1fcfa6c43bd2749e/webapp/src/ts/services/target-aggregates.service.ts#L83) by just grabbing every `target` doc for a specific reporting period.  In the context of an offline user, this is basically what you want. You get the target docs for all the users that you have access to. For an online user, this would load the `target` docs for the reporting period _for all users_.  

When we implement this in cht-datasource we should modify the flow.  We take as a parameter the contact ids we want to load target docs for.  Then we make the same `all_docs` call to get all the target docs for the reporting period, but we set `include_docs: false` so only the ids are returned.  Then we filter the ids based on the ones that include one of the provided contact ids from the parameter.  Finally we use the remaining target ids to load the `target` docs directly by id and return them.

I have considered a couple other approaches to querying the target docs, but IMHO these approaches are unnecessarily heavy/complex for the problem at hand:

- We could do separate `all_docs` calls for *each* of the `contact_id`s - this would not scale well when there are many child places, lots of round-trips to the DB (though they could be done in parallel).  
- We could add a view index for `target` docs to index the target docs by `reporting_period` and `owner`. This would let us directly query the target docs that we want, but it would be another view index to store (and one that is likely to not see heavy use).  Also this does not take any advantage of the structured `_id` values that target docs have.
- We could add a view index for `user-settings` docs (indexing by `contact_id`). This would mean we could directly query the `user-settings` and get direct access to data about the "supervised users". Then we can just select the `target` docs by _id since we have all the info we need. 
    - I like this because the view could be useful in other contexts besides just targets.  It still seems like unnecessary effort for the current needs (particularly since the webapp currently does not allow online users to load targets/aggregates) but we should keep this in mind for future optimizations.

### cht-datasource

- Update the `src/qualifier.ts` file to implement:

<details>
<summary>`src/qualifier.ts`</summary>

```ts
/**
 * A qualifier that identifies entities based their association with the identified contacts.
 */
export interface ContactUuidsQualifier {
  readonly contactUuids: [string, ...string[]]
}

export const byContactUuids = (contactUuids: [string, ...string[]]): ContactUuidsQualifier....
```

</details>

- Update the `src/target-interval.ts` file to implement:

<details>
<summary>`src/target-interval.ts`</summary>

```ts

  /**
   * Returns a function for retrieving a paged array of target intervals from the given data context.
   * @returns a function for retrieving a paged array of target intervals
   * @see {@link getAll} which provides the same data, but without having to manually account for paging
   */
  export const getPage = (context: DataContext): typeof curriedFn => {
    ...
    /**
     * Returns an array of target intervals for the provided page specifications.
     * @param qualifier the limiter defining which target intervals to return
     * @param cursor the token identifying which page to retrieve. A `null` value indicates the first page should be
     * returned. Subsequent pages can be retrieved by providing the cursor returned with the previous page.
     * @param limit the maximum number of identifiers to return. Default is 100.
     * @returns a page of target intervals for the provided specification
     * @throws InvalidArrgumentError if no qualifier is provided or if the qualifier is invalid
     * @throws InvalidArgumentError if the provided `limit` value is `<=0`
     * @throws InvalidArgumentError if the provided cursor is not a valid page token or `null`
     */
    const curriedFn = async (
      qualifier: ReportingPeriodQualifier & ContactUuidsQualifier,
      cursor: Nullable<string> = null,
      limit: number | `${number}` = DEFAULT_DOCS_PAGE_LIMIT
    ): Promise<Page<TargetInterval>> => {
      ...
    };
    return curriedFn;
  };

  /**
   * Returns a function for getting a generator that fetches target intervals from the given data context.
   * @returns a function for getting a generator that fetches target intervals
   */
  export const getAll = (context: DataContext): typeof curriedGen => {
    ...
    /**
     * Returns a generator for fetching all target intervals that match the given qualifier
     * @param qualifier the limiter defining which target intervals to return
     * @returns a generator for fetching all target intervals that match the given qualifier
     * @throws InvalidArgumentError if no qualifier is provided or if the qualifier is invalid
     */
    const curriedGen = (
      qualifier: ReportingPeriodQualifier & ContactUuidsQualifier,
    ): AsyncGenerator<TargetInterval, null> => {
      ...
    };
    return curriedGen;
  };
```

</details>

- Update the `src/local/target-interval.ts` to load the requested target doc to load the target docs.
  - I _think_ the local adapter implementation for `getPage` should be able to leverage the [`fetchAndFilter` function](https://github.com/medic/cht-core/blob/f1bdfc07c0f4b1170e0d8be671cd27aa105baff4/shared-libs/cht-datasource/src/local/libs/doc.ts#L119) to load a page of  _ids_ for target docs.
    - The `getFunction` should call `allDocs` to load all the ids for the reporting period.
    - The `filterFunction` should filter the ids to only ones that contain one of the given contact UUIDs.
  - Once we get the target ids, we can take them and retrieve the actual target docs and return those as the `data` array for the page.
    - Use another call to `allDocs`, but this time with `include_docs: true` and the target doc ids set as the `keys` option.
- Update the `src/remote/target-interval.ts` to load the target docs by querying the (yet to be implemented) `/api/v1/target-interval` REST endpoint.
- Update the `src/index.ts`. In the `getDatasource` function, add new functions to the `targetInterval` property:

<details>
<summary>`getDatasource`</summary>

```ts
      targetInterval: {
...
        /**
         * Returns an array of target intervals for the provided page specifications.
         * @param reportingPeriod the reporting period for the target intervals
         * @param contactUuids the contact UUIDs for the target intervals
         * @param cursor the token identifying which page to retrieve. A `null` value indicates the first page should be
         * returned. Subsequent pages can be retrieved by providing the cursor returned with the previous page.
         * @param limit the maximum number of target intervals to return. Default is 100.
         * @returns a page of target intervals for the provided specifications
         * @throws InvalidArgumentError if no reporting period is provided
         * @throws InvalidArgumentError if no contact UUIDs are provided
         * @throws InvalidArgumentError if the provided cursor is not a valid page token or `null`
         * @throws InvalidArgumentError if the provided limit is `<= 0`
         * @see {@link getByReportingPeriodContactUuids} which provides the same data, but without having to manually 
         * account for paging
         */
        getPageByReportingPeriodContactUuids: (
          reportingPeriod: string, 
          contactUuids: [string, ...string[]],
          cursor: Nullable<string> = null,
          limit: number | `${number}` = DEFAULT_DOCS_PAGE_LIMIT
        ) => ...,

        /**
         * Returns a generator for fetching all target intervals in the reporting period for the given contact UUIDs.
         * @param reportingPeriod the reporting period for the target intervals
         * @param contactUuids the contact UUIDs for the target intervals
         * @returns a generator for fetching all target intervals with the given type
         * @throws InvalidArgumentError if no reporting period is provided
         * @throws InvalidArgumentError if no contact UUIDs are provided
         */
        getByReportingPeriodContactUuids: (reportingPeriod: string, contactUuids: [string, ...string[]]) => ...,
      }
```

</details>

### api

- Update the `src/controllers/target-interval.js` controller to have a `getAll` function. The controller should call your new cht-datasource function with data loaded from `req.query.reporting_period`, `req.query.contact_uuids`, `req.query.cursor`, and `req.query.limit`.  (See the `src/controllers/person.js` controller for example.)
- Update `src/routing.js` to route `/api/v1/target-interval` to your new controller function.

### Integration tests

- Add tests for the new `/api/v1/target-interval` endpoint functionality in `test/integration/api/controllers/target-interval.spec.js`
- Add tests for the new cht-datasource remote adapter functionality in `test/integration/shared-libs/controllers/target-interval.spec.js`
    - This should basically be a copy of your api endpoint tests but using the `getRemoteDataContext` and the cht-datasource interfaces instead of making direct REST calls.

kanarelo:
This pull request updates the rules engine service to support returning targets for the previous month.

Closes #10344 

# Code review checklist

# License

### DIFF
--- a/shared-libs/cht-datasource/test/local/libs/doc.spec.ts
+++ b/shared-libs/cht-datasource/test/local/libs/doc.spec.ts
@@ -161,20 +161,46 @@ describe('local doc lib', () => {
     });
   });
 
-  it('getDocUuidsByIdRange - returns ids found in the given range', async () => {
-    const startKey = 'doc0';
-    const endKey = 'doc3';
-    dbAllDocs.resolves({
-      rows: [
-        { id: startKey },
-        { id: 'doc1' },
-        { id: 'doc2' }
-      ]
+  describe('getDocUuidsByIdRange', () => {
+    it('returns ids found in the given range', async () => {
+      const startkey = 'doc0';
+      const endkey = 'doc3';
+      dbAllDocs.resolves({
+        rows: [
+          { id: startkey },
+          { id: 'doc1' },
+          { id: 'doc2' }
+        ]
+      });
+
+      const result = await getDocUuidsByIdRange(db)(startkey, endkey);
+
+      expect(result).to.deep.equal([startkey, 'doc1', 'doc2']);
+      expect(dbAllDocs).to.be.calledOnceWithExactly({
+        startkey,
+        endkey,
+        include_docs: false,
+        limit: undefined,
+        skip: 0
+      });
     });
 
-    const result = await getDocUuidsByIdRange(db)(startKey, endKey);
+    it('returns an empty array if no ids are found', async () => {
+      const startkey = 'doc0';
+      const endkey = 'doc3';
+      dbAllDocs.resolves({ rows: [] });
 
-    expect(result).to.deep.equal([startKey, 'doc1', 'doc2']);
+      const result = await getDocUuidsByIdRange(db)(startkey, endkey);
+
+      expect(result).to.deep.equal([]);
+      expect(dbAllDocs).to.be.calledOnceWithExactly({
+        startkey,
+        endkey,
+        include_docs: false,
+        limit: undefined,
+        skip: 0
+      });
+    });
   });
 
   describe('queryDocsByRange', () => {


api/src/controllers/target-interval.js
```diff
@@ -26,5 +27,18 @@ module.exports = {
...
+    getAll: serverUtils.doOrError(async (req, res) => {
+      await checkUserPermissions(req);
+
+      const reportingPeriod = Qualifier.byReportingPeriod(req.query.reporting_period);
+      const contactUuids = Qualifier.byContactUuids(req.query.contact_uuids);
+      const docs = await getTargetIntervals(
+        {...reportingPeriod, ...contactUuids}, 
+        req.query.cursor, 
+        req.query.limit 
+      );
```
jkuester:
```suggestion
      const contactUuids = (req.query.contact_uuids ?? '')
        .split(',')
        .filter(Boolean);
      const docs = await getTargetIntervals(
        Qualifier.and(
          Qualifier.byContactUuids(contactUuids),
          Qualifier.byReportingPeriod(req.query.reporting_period)
        ),
        req.query.cursor, 
        req.query.limit 
      );
```

We need to parse the contact uuids into an array before creating the qualifier.

shared-libs/cht-datasource/src/local/target-interval.ts
```diff
@@ -38,4 +39,100 @@ export namespace v1 {
...
+      const intervalIds = await fetchTargetIntervalIds(limit, skip)(db, qualifier);
+
+      return await fetchTargetIntervalDocs(limit, skip)(db, intervalIds.data);
+    };
+  };
+  
+  const fetchTargetIntervalIds = (limit: number, skip: number) => async (
+    db: PouchDB.Database, 
+    qualifier: (ReportingPeriodQualifier & ContactUuidsQualifier)
+  ) => {
```
jkuester:
```suggestion
  ) => {
    const contactUuidSet = new Set(qualifier.contactUuids);
```

Lets save off the contact UUIDs as a set so the filter function is more efficient.

shared-libs/cht-datasource/src/local/target-interval.ts
```diff
@@ -38,4 +39,100 @@ export namespace v1 {
...
+      const result = await db
+        .allDocs({ include_docs: false, limit, skip, startkey: `target~${qualifier.reportingPeriod}` });
+      
+      if (result.rows.length === 0) {
+        return [];
+      }
+
+      return result.rows
+        .filter(row => row.doc)
+        .map(row => row.doc as TargetInterval.v1.TargetInterval);
```
jkuester:
```suggestion
      const result = await db.allDocs({
        include_docs: false,
        limit,
        skip,
        startkey: `target~${qualifier.reportingPeriod}~`,
        endkey: `target~${qualifier.reportingPeriod}~\ufff0`,
      });
      return result.rows.map(({ id }) => id);
```

This was super close here, but there are a couple problems. First, with `include_docs: false` we are not going to get anything back for `row.doc`. This is fine because we actually only want `row.id`.  This whole `fetchAndFilter` operation can operate on just the UUIDs.

We also need to include the `endkey` option here or Couch will just return _every_ doc that comes alphabetically after our startkey.  The `\ufff0` char is basically the "max value" indicator so we select all the ids with the given reportingPeriod and no others.

shared-libs/cht-datasource/src/local/target-interval.ts
```diff
@@ -38,4 +39,100 @@ export namespace v1 {
...
+      }
+
+      for (const contactUuid of qualifier.contactUuids) {
+        if (contactUuid.includes(uuid!)) {
+          return true; 
+        }
+      }
+
+      return false;
+    };
```
jkuester:
```suggestion
    const filterFn = (uuid: Nullable<string>) => {
      if (!uuid) {
        return false;
      }
      const [,, contactUuid] = uuid.split('~');
      return !!contactUuid && contactUuidSet.has(contactUuid);
    };
```

Now here in the filter function we just check if the returned `uuid` is contained in the `contactUuidSet`.

shared-libs/cht-datasource/src/local/target-interval.ts
```diff
@@ -38,4 +39,100 @@ export namespace v1 {
...
+  const fetchTargetIntervalDocs = (limit: number, skip: number) => async (
+    db: PouchDB.Database, 
+    targetIntervals: TargetInterval.v1.TargetInterval[]
+  ) => {
+    const fetchFn = async (
+      limit: number,
+      skip: number
+    ) => {
+      const result = await db
+        .allDocs({ include_docs: true, limit, skip, keys: targetIntervals.map(t => t._id) });
```
jkuester:
```suggestion
    targetIntervalUuids: string[]
  ) => {
    const fetchFn = async (
      limit: number,
      skip: number
    ) => {
      const result = await db
        .allDocs({ include_docs: true, limit, skip, keys: targetIntervalUuids });
```

The last piece we need is to just tweak the parameter here to just be the uuid strings.

jkuester:
FYI @kanarelo I merged in the latest from the `10140_previous-months-targets` branch.  We ran into a CI issue (https://github.com/medic/cht-core/issues/10467) recently (think it came from a Docker update). But, either way that change needed to be merged down here for the CI to be able to pass.

The only action you need to take here is to make sure the pull the latest changes from your branch down before pushing anything new. :+1:   Otherwise, I am going to keep an eye on the build here and see what it does...

PR #10436, opened 2025-11-04
feat(#10318): display previous month's targets when selected

### LINKED ISSUE #10318: Implement displaying previous month's targets
- [ ] Implement filtering of targets when reporting period is selected in the sidebar
  - Currently the [AnalyticsTargetsComponent.getTargets method](https://github.com/medic/cht-core/blob/9337c841e9afd6c25652f534a6f0105f4675bd88/webapp/src/ts/modules/analytics/analytics-targets.component.ts#L87) is called when toggling the "Reporting Period" filter.  
  - The provided `reportingPeriod` parameter is `ReportingPeriod.CURRENT` or `ReportingPeriod.PREVIOUS` depending on what the user chose in the filter.
  - This function needs to be refactored to pass the reporting period argument to the new version of `rulesEngineService.fetchTargets()` that @paulpascal is implementing in https://github.com/medic/cht-core/pull/10337.
- [ ] ~As noted in the [design doc](https://docs.google.com/document/d/1zdJ1ySdSbriZODn7BDDAH9Hxg3qT0J9McGUIIrcafIU/edit?tab=t.ufay0fr37c8h#heading=h.s1wiyw3hzx4x) the "All Time" target cards need to be updated so that when the previous month filter is selected, the cards display an indicator that notes the value is "up to the previous month".~
  - ~The "All Time" targets are ones [configured](https://docs.communityhealthtoolkit.org/building/targets/targets-js/#targetsjs) with the `date` property as `'now'` (or not set).~
  - ~Currently, the "All Time" subtitle is being populated [here](https://github.com/medic/cht-core/blob/9337c841e9afd6c25652f534a6f0105f4675bd88/webapp/src/ts/modules/analytics/analytics-targets.component.html#L50).~
  - ~We need another visual element in the card (only shown for "All Time" targets when the previous reporting period is selected) that indicates to the user that the value in the card is the data up to the previous month.~
  - (EDIT: moved to https://github.com/medic/cht-core/issues/10439)
- [ ] Write e2e tests for targets filtering
  - Once we have the functionality all wired up, we should be able to update the [analytics.wdio-spec.js e2e test suite](https://github.com/medic/cht-core/blob/9337c841e9afd6c25652f534a6f0105f4675bd88/tests/e2e/default/analytics/analytics.wdio-spec.js#L100) to actually test filtering the targets back to the previous month's data.

Edwin-Rugoogamu:
# Description

Closes #10318

- Update `default` config to set target subtitle dynamically for monthly targets.
- Update targets page in webapp to show back button in the mobile view title bar when the previous month is selected.
- Update target aggregates page in webapp to show the target subtitle for both the current and previous month's target data (instead of showing the previous month name). 
- Update both the target and target aggregates pages to show the count number of how many filters are selected.

<details>
<summary>Original Description</summary>

## Core Functionality Implemented

**Analytics Targets Component (`analytics-targets.component.ts`)**:
- Created an Angular component that displays health targets/metrics (e.g., births, registrations, visits)
- Integrated with the rules engine to fetch and display configured targets
- Added support for filtering by reporting period (current month vs. previous month)
- Implemented proper state management for loading, error handling, and disabled states
- Added performance tracking for analytics load times

**Dynamic UI (`analytics-targets.component.html`)**:
- Built a responsive targets dashboard with sidebar filtering
- Implemented conditional rendering for different states (disabled, loading, no targets, errors)
- Added progress visualization for percentage-based targets and count displays for numeric targets
- Created dynamic subtitles that change based on selected reporting period:
  - "Current month" for current period
  - "All time" for previous period
  - Added contextual note "📅 Data up to the previous month" when viewing previous period data

**Rules Engine Integration (`rules-engine.service.ts`)**:
- Extended the rules engine service to support target fetching with reporting period filtering
- Added `fetchTargets()` method that accepts 'current' or 'previous' reporting periods
- Integrated calendar interval logic to determine correct time ranges for target calculations
- Maintained existing task and emission monitoring functionality

**Calendar Interval Service (`calendar-interval.service.ts`)**:
- Implemented utility methods for calculating calendar-based intervals
- Added `getPrevious()` method specifically for previous month reporting periods
- Ensured proper month boundary calculations for accurate target periods

**Internationalization (`messages-en.properties`)**:
- Added comprehensive translation keys for target-related UI elements
- Included contextual messages for different reporting periods
- Added goal-related translations and error states

## Key Features Delivered

1. **Reporting Period Filtering**: Users can switch between current month and previous month targets
2. **Progress Tracking**: Visual progress bars for percentage targets, numeric displays for count targets
3. **Goal Achievement Indicators**: Visual styling when targets meet or exceed goals
4. **Error Handling**: Graceful handling of disabled rules engine, loading states, and API errors
5. **Performance Monitoring**: Built-in performance tracking for target loading operations
6. **Responsive Design**: Sidebar integration with proper layout handling
7. **Accessibility**: Proper ARIA attributes and semantic HTML structure

</details>

### DIFF
--- a/webapp/src/ts/modules/analytics/analytics-sidebar-filter.component.ts
+++ b/webapp/src/ts/modules/analytics/analytics-sidebar-filter.component.ts
@@ -13,7 +13,7 @@ import { MatAccordion } from '@angular/material/expansion';
 import { FormsModule } from '@angular/forms';
 import { TranslatePipe } from '@ngx-translate/core';
 import { TelemetryService } from '@mm-services/telemetry.service';
-import { SidebarFilterState } from '@mm-reducers/global';
+import { type SidebarFilterState } from '@mm-reducers/global';
 
 export enum ReportingPeriod {
   CURRENT = 'current',
--- a/webapp/tests/karma/ts/services/target-aggregates.service.spec.ts
+++ b/webapp/tests/karma/ts/services/target-aggregates.service.spec.ts
@@ -631,23 +631,9 @@ describe('TargetAggregatesService', () => {
 
     it('should calculate every type of target aggregate correctly', async () => {
       const config = { tasks: { targets: { items: [
-        { id: 'target1', aggregate: true, type: 'count', title: 'target1', subtitle_translation_key: 'sub_trans_key' },
-        {
-          id: 'target2',
-          aggregate: true,
-          type: 'count',
-          goal: 20,
-          translation_key: 'target2',
-          subtitle_translation_key: '() => "sub_trans_key"'
-        },
-        {
-          id: 'target3',
-          aggregate: true,
-          type: 'percent',
-          goal: -1,
-          title: 'target3',
-          subtitle_translation_key: '(reportingPeriod) => reportingPeriod'
-        },
+        { id: 'target1', aggregate: true, type: 'count', title: 'target1' },
+        { id: 'target2', aggregate: true, type: 'count', goal: 20, translation_key: 'target2' },
+        { id: 'target3', aggregate: true, type: 'percent', goal: -1, title: 'target3' },
         { id: 'target4', aggregate: true, type: 'percent', goal: 80, translation_key: 'target4' },
         { id: 'target5', aggregate: true, type: 'count', goal: 2, translation_key: 'target5' },
       ] } }};
@@ -731,7 +717,6 @@ describe('TargetAggregatesService', () => {
         type: 'count',
         title: 'target1',
         reportingMonth: 'January',
-        subtitle_translation_key: 'sub_trans_key',
         aggregateValue: { pass: 26, total: 26, hasGoal: false, summary: 26 },
         heading: 'target1',
         hasGoal: false,
@@ -750,7 +735,6 @@ describe('TargetAggregatesService', () => {
         goal: 20,
         translation_key: 'target2',
         reportingMonth: 'January',
-        subtitle_translation_key: '() => "sub_trans_key"',
         aggregateValue: { pass: 1, total: 3, goalMet: false, hasGoal: true, summary: ratioTranslationKey },
         heading: 'target2',
         hasGoal: true,
@@ -769,7 +753,6 @@ describe('TargetAggregatesService', () => {
         goal: -1,
         title: 'target3',
         reportingMonth: 'January',
-        subtitle_translation_key: '(reportingPeriod) => reportingPeriod',
         aggregateValue: { pass: 20, total: 51, percent: 39, hasGoal: false, summary: '39%' },
         heading: 'target3',
         hasGoal: false,


mrjones-plip:
Thanks for your submission @Edwin-Rugoogamu ! While I'll leave the larger, more formal code review to the previous performance squad, you can get a jump on improving this PR by fixing the failing CI with the linting and build errors. Please click into the failing checks to see the specific errors and ask questions here if you get stuck!

Edwin-Rugoogamu:
> Thanks for your submission @Edwin-Rugoogamu ! While I'll leave the larger, more formal code review to the previous performance squad, you can get a jump on improving this PR by fixing the failing CI with the linting and build errors. Please click into the failing checks to see the specific errors and ask questions here if you get stuck!

Hi @mrjones-plip The fixes for the eslint have been worked on. as reflected in the new commit (eslint fixes)

mrjones-plip:
Thanks @Edwin-Rugoogamu ! That's helpful to smooth out before a formal code review.

Your new challenge is to resolve the conflicts in this branch :sweat_smile:  As with the linting - please let us know if you need a hand!  This too will be nice to smooth out before a formal code review.

Edwin-Rugoogamu:
> Thanks @Edwin-Rugoogamu ! That's helpful to smooth out before a formal code review.
> 
> Your new challenge is to resolve the conflicts in this branch 😅 As with the linting - please let us know if you need a hand! This too will be nice to smooth out before a formal code review.

Thanks @mrjones-plip . I have accepted all the incoming changes and adjusted variable(relevantInterval).   Am also looking into the linting. Which had failled in the github actions pipeline

mrjones-plip:
Nice work - thank you @Edwin-Rugoogamu ! It looks like just the title of the PR needs to pass lint now.  Can you look into fixing that?

As well, I'll reach out back on Slack for someone to look at the actual code itself.

Edwin-Rugoogamu:
> Nice work - thank you @Edwin-Rugoogamu ! It looks like just the title of the PR needs to pass lint now. Can you look into fixing that?
> 
> As well, I'll reach out back on Slack for someone to look at the actual code itself.

Well noted

webapp/src/ts/services/calendar-interval.service.ts
```diff
@@ -23,4 +23,16 @@ export class CalendarIntervalService {
   getInterval(startDate:number, timestamp:number) {
     return CalendarInterval.getInterval(startDate, timestamp);
   }
+
+  /**
+   * Returns the previous reporting period interval based on the start day.
+   * This is used when the sidebar filter selects "previous month".
+   */
+  getPrevious(startDate: number) {
```
jkuester: Is this being used anywhere?

webapp/src/ts/modules/analytics/analytics-targets.component.html
```diff
@@ -48,7 +48,20 @@
...
+  </ng-container>
+
+  <!-- 🔹 Show “All time” when previous month is selected -->
+  <ng-template #previousSubtitle>
+    {{ 'targets.all_time.subtitle' | translate }}
+    <!-- Inline note: only show when previous period -->
+    
+  </ng-template>
+
+
```
jkuester: Lets roll this back for now and wait for https://github.com/medic/cht-core/issues/10439 to make any changes to the UI.


```suggestion
              {{ target.subtitle_translation_key | translate }}
```

api/resources/translations/messages-en.properties
```diff
@@ -1248,8 +1248,8 @@ targets.births.title = Births
 targets.count.default = ({{pass}} of {{total}})
 targets.disabled = Targets are disabled for your user account.
 targets.no_targets = No target found.
-targets.this_month.subtitle = This month
-targets.last_month.subtitle = Last month
+targets.this_month.subtitle = Current month
+targets.last_month.subtitle = Previous month
```
jkuester: Same here. I think we can hold off on UI/translation changes until we land on exactly how we are going to approach this:

```suggestion
targets.this_month.subtitle = This month
targets.last_month.subtitle = Last month
```

jkuester (changes requested):
Okay, so when this is all boiled down, it seems possible that the only real change will be to pass the param through in the analytics-targets component.  That is pretty much expected given the other sub-issues that we have.  👍

---

@Edwin-Rugoogamu I _think_ you said that Timothy was working on the test changes. Is he going to add those in a commit to this PR or in a separate PR?  (Would be best for them to end up in this PR before we merge to the base branch just so we can keep everything straight...)

We can update the unit tests in [analytics-targets.component.spec.ts](https://github.com/medic/cht-core/blob/9337c841e9afd6c25652f534a6f0105f4675bd88/webapp/tests/karma/ts/modules/analytics/analytics-targets.component.spec.ts):

- In the existing tests, assert the value of the parameter passed to the `fetchTargets` function (it is stubbed with sinon, so it should be easy to assert the args).
    - Would be nice to update one of the existing tests (if there is not one already) to set the `reportingPeriodFilter` value on the component and then assert that value is passed to `fetchTargets`.
- Add another test that calls `getTargets` directly with a reporting period param and assert that value is passed to `fetchTargets` and is set as the `reportingPeriodFilter`.

In the e2e tests, we can expand the [filter test in analytics.wdio-spec.js](https://github.com/medic/cht-core/blob/9337c841e9afd6c25652f534a6f0105f4675bd88/tests/e2e/default/analytics/analytics.wdio-spec.js#L100) to actually select the previous month and assert the targets display the expected values for the previous month.  In the test setup, we will probably need to write some docs for the previous month so it gives us some target values.

---

Regarding the tests that are currently failing in the CI, I think that rolling back the UI changes (as I suggested) will resolve the problems in the `ci-webdriver-default-workflows` CI job.  I think the `ci-webdriver-default-core-minimum-browser` test is just flaking out for some reason. 🤷 I cannot come up with any way your changes here would cause that test to fail. 🤔 I have triggered a re-run and will keep an eye on things and see if it ends up passing...

api/resources/translations/messages-en.properties
Edwin-Rugoogamu: hi @jkuester I have put these back as suggested

webapp/src/ts/modules/analytics/analytics-targets.component.html
Edwin-Rugoogamu: Due to the label being static, The filtering does the sorting by date but leaves the labels constant. So when you click this month or last month the labels remain constant. So we decided to change the label according to what has been chosen . Though now differently as indicated in the new PR.

webapp/src/ts/services/calendar-interval.service.ts
Edwin-Rugoogamu: Hi @jkuester This has been removed

webapp/src/ts/modules/analytics/analytics-targets.component.html
jkuester: As discussed at the squad meeting this week, we want to go ahead and leave this html as it is for now.  We are continuing to discuss what the best approach will be for the UI changes. These will be implemented in r https://github.com/medic/cht-core/issues/10439.

Edwin-Rugoogamu: Am changing it first think tomorrow. Just got abit engaged.

Edwin-Rugoogamu:
hi @jkuester This PR for the display of previous month targets and the tests  is soon closing . Please have a look at it.   The file analytics-targets.component.html is also in this [PR](https://github.com/medic/cht-core/pull/10507) which was created to change the string to a function.

jkuester:
Okay, this PR should be good to go.  It is pretty much just waiting on these changes in cht-conf (https://github.com/medic/cht-conf/pull/777) and these changes in cht-core (https://github.com/medic/cht-core/pull/10507).

webapp/src/css/inbox.less
```diff
@@ -1771,7 +1786,6 @@ mm-sidebar-menu .mat-sidenav-container {
     .header,
     .app-menu-button,
     .ellipsis-title,
-    mm-analytics-filters,
```
jkuester, line 1774: This was hiding the filter icon on the mobile view when in the "show content" mode (when the back button is shown).

webapp/src/ts/modules/analytics/analytics-sidebar-filter.component.ts
```diff
@@ -57,26 +57,36 @@ export class AnalyticsSidebarFilterComponent implements OnInit, OnDestroy {
...
+    this.subscriptions.add(subscription);
+  }
 
-        if (!this.selectedReportingPeriod && filterState?.defaultFilters?.reportingPeriod) {
-          this.selectedReportingPeriod = filterState.defaultFilters.reportingPeriod;
-        }
-      });
+  private getTotalFilterCount() {
+    const reportingPeriodCount = this.selectedReportingPeriod === ReportingPeriod.PREVIOUS ? 1 : 0;
+    const facilityCount = this.userFacilities.length > 1 ? 1 : 0;
```
jkuester, line 88: I am conflicted over how exactly to handle this logic here.  Basically this controls whether we show a filter count value for the facility (when looking at aggregate targets).  The problem is that you always have to have a facility selected (there is no support right now for not filtering by facility and just seeing aggregate totals _across all facilities_).  When your supervisor user has multiple facilities, they can choose between then via radio buttons in the sidebar filter, but they always have at least one selected.  

This current logic here follows [some other logic](https://github.com/medic/cht-core/blob/9522b450d0c7283c0402f263995c07e16f04af4f/webapp/src/ts/modules/analytics/analytics-target-aggregates.component.ts#L131) in the target aggregates component when it decides to show the facility name for the target or not. Basically, if the user only has one facility then we do not treat this as a "filtered" value.  But if the user has multiple facilities then I am setting the filter count to `1`.  This means that when a user with multiple facilities is looking at the target aggregates page, they are always going to see a filter count value of at least `1` (because they can never un-select the facility filter).  But for users that only have one facility, they will not see a filter count unless they select the previous reporting period.

Definitely open to changing this, but I figured this was a good place to start as it seemed least surprising to the user and most easy to evolve in the future if we add the ability to look at aggregate targets _across all facilities_.

webapp/src/ts/modules/analytics/analytics-target-aggregates.component.ts
```diff
@@ -126,32 +125,27 @@ export class AnalyticsTargetAggregatesComponent implements OnInit, OnDestroy {
...
       aggregate.facility = userFacility?.name;
       filtersToDisplay.push(this.facilityFilter.name);
     }
 
-    aggregate.reportingMonth = reportingMonth;
     aggregate.reportingPeriod = reportingPeriod;
-    if (aggregate.reportingPeriod === ReportingPeriod.PREVIOUS) {
-      filtersToDisplay.push(aggregate.reportingMonth);
+    if (aggregate.subtitle) {
+      filtersToDisplay.push(aggregate.subtitle);
```
jkuester: This is another design consideration that I am not 100% sure about.  This `filtersToDisplay` controls the _sub-text_ that is displayed under the title of each target list item when viewing aggregate targets.  Previously we did not include anything for the reporting period in the subtext when the user was viewing the "CURRENT" target aggregates, but we did include the month name when the user was viewing "PREVIOUS".  When I went to switch this from showing the month name to just showing the subtitle, it did not really make sense to only do this for the PREVIOUS reporting period. So, instead this is now always showing the subtitle (if it exists).

Image

.github/workflows/build.yml
```diff
@@ -187,7 +187,7 @@ jobs:
         if: ${{ !env.INTERNAL_CONTRIBUTOR }}
 
       - run: mkdir tests/logs
-      - run: npm install -g cht-conf
+      - run: npm install -g github:jkuester/cht-conf#10140_previous-month-targets
```
jkuester: Need to update this to the latest version of cht-conf when it is released.

jkuester:
Okay, @dianabarsan this is ready for you to have a look now!

Here is filtering the targets on the mobile view:

[target_mobile.webm](https://github.com/user-attachments/assets/8791d200-e729-429b-ac4c-bbeb33f5b6e4)

Here is the desktop view:

[target_desktop.webm](https://github.com/user-attachments/assets/4d66553b-ca2e-4705-9231-e20d300df16d)

---

Here is the target aggregates mobile view:

[aggregates_mobile.webm](https://github.com/user-attachments/assets/2ba8b59e-4982-4806-9a32-aca6fea96d65)

And the desktop view:

[aggregates_desktop.webm](https://github.com/user-attachments/assets/f5917569-decf-462b-aa90-261acff1ed27)

Finally, this is what the aggregates look like for a user with just one facility:

[aggregates_single_facility.webm](https://github.com/user-attachments/assets/3f467856-796d-486f-a950-f11cec4810f3)

config/default/test/targets/deaths.spec.js
```diff
@@ -22,36 +22,40 @@ describe('Death related targets tests', () => {
     await harness.setNow('2000-04-30');//DOD: 2000-04-25
     harness.subject = babyDeceasedAtAge1Day;
     const deathsThisMonth = await harness.getTargets({ type: 'deaths-this-month' });
-    expect(deathsThisMonth).to.deep.equal([{
+    expect(deathsThisMonth).to.have.length(1);
```
dianabarsan: Are we including this change in default config? or is this just an intermediary change?

webapp/src/ts/modules/analytics/analytics-sidebar-filter.component.ts
dianabarsan: I think your logic is solid.

webapp/src/ts/modules/analytics/analytics-target-aggregates.component.ts
dianabarsan: This looks a bit cluttered on mobile. I think being consistent is important, so showing the subtitle is fine. But if the subtitle is long, this might not scale visually

webapp/src/ts/modules/analytics/analytics-targets.component.ts
```diff
@@ -74,26 +80,34 @@ export class AnalyticsTargetsComponent implements OnInit, OnDestroy {
...
     this.subscriptions.add(selectorsSubscription);
-  }
 
-  private setDefaultFilters() {
-    const defaultFilters = {
-      reportingPeriod: ReportingPeriod.CURRENT,
-    };
-    this.globalActions.setSidebarFilter({ defaultFilters });
-    this.reportingPeriodFilter = defaultFilters.reportingPeriod;
+    const backToCurrentSubscription = this.store
```
dianabarsan, line 86: Is it me or is the target-aggregate code (referring to the whole directory) become a tangled mess? (not a comment about this PR, this just builds on top of the already existent mess).

webapp/src/ts/modules/analytics/analytics-sidebar-filter.component.ts
```diff
@@ -57,26 +57,36 @@ export class AnalyticsSidebarFilterComponent implements OnInit, OnDestroy {
     this.subscriptions.unsubscribe();
   }
 
+  ngOnChanges(changes: SimpleChanges) {
```
dianabarsan: uuu plugging into Angular hooks? nice!

webapp/src/ts/modules/analytics/analytics-targets.component.ts
```diff
@@ -74,26 +80,34 @@ export class AnalyticsTargetsComponent implements OnInit, OnDestroy {
...
+      });
+    this.subscriptions.add(backToCurrentSubscription);
   }
 
-  getTargets(reportingPeriod?) {
-    if (reportingPeriod) {
-      this.reportingPeriodFilter = reportingPeriod;
+  getTargets(reportingPeriod: ReportingPeriod) {
+    this.reportingPeriodFilter = reportingPeriod;
+    // Show the back button in the mobile layout when viewing previous targets
```
dianabarsan, line 101: if it were not for this comment, I would not have guessed. 
showContent also hides the black bottom menu. feels like this deserves a better name. a dream for another time....

webapp/src/ts/modules/analytics/analytics-sidebar-filter.component.ts
```diff
@@ -57,26 +57,36 @@ export class AnalyticsSidebarFilterComponent implements OnInit, OnDestroy {
...
+    if (changes['selectedReportingPeriod'] || changes['selectedFacility']) {
```
dianabarsan: We don't really use this pattern of listening to changes, do we? I think that's on purpose. 
I also notice that the filter is applied immediately instead of needing to click "apply". Wouldn't it be simpler to only apply the filter when the apply button is clicked and avoid all these change complications? right now the apply button is kinda useless.

webapp/src/ts/modules/analytics/analytics-targets.component.ts
jkuester: Yeah, it could definitely do with some TLC... But sadly, I think that is a problem for another time.  😞

Edwin-Rugoogamu:
@jkuester Great designs . Looks like we are doing away with the cards? Am thinking won't it be challenging for the Chws to easily determine the targets at first glance.  `for the webm`

webapp/src/ts/modules/analytics/analytics-target-aggregates.component.ts
jkuester: Def agree that it is crowded.  One additional thing to note here is that if there is not enough space for the breadcrumbs on one line, it will wrap to two:

image

But, it will not wrap past two lines:

image

config/default/test/targets/deaths.spec.js
jkuester: My plan was to ship these changes.  My initial work here seemed like at least the bare minimum of what we would want and I figure we can iterate more in the future as necessary. But, I am also happy to make any additional changes here in this PR.  :+1:

jkuester:
Thanks for the review @Edwin-Rugoogamu!  Yeah, we ended up basically going with the final set of screenshots at the bottom of the [Visual Design page](https://docs.google.com/document/d/1zdJ1ySdSbriZODn7BDDAH9Hxg3qT0J9McGUIIrcafIU/edit?tab=t.iphs3osaykhb).  I think between the subtitle changes and the filter numbers, we should have enough visual cues that users can understand what is going on. :+1:

webapp/src/ts/modules/analytics/analytics-sidebar-filter.component.ts
jkuester: Okay, @dianabarsan I think I have sorted out a better data flow for this!  :+1: Just leveraging the sidebar filter state via the `getSidebarFilter` selector to maintain the current state of the sidebar.  I am sure there is more that can be done to make this better, :sweat_smile: but I am pretty content with this as an improvement on the existing code....

webapp/src/ts/modules/analytics/analytics-target-aggregates.component.ts
dianabarsan: yickes! I wouldn't be able to understand anything from the small screen. but nowadays maybe no screens are really that small. 
i wonder if there's a way to measure chw satisfaction kpis.

webapp/src/ts/modules/analytics/analytics-sidebar-filter.component.ts
```diff
@@ -57,36 +67,30 @@ export class AnalyticsSidebarFilterComponent implements OnInit, OnDestroy, OnCha
...
       .select(Selectors.getSidebarFilter)
-      .subscribe((filterState) => this.isOpen = filterState?.isOpen ?? false);
+      .subscribe((filterState) => {
+        if (!filterState) {
+          return;
+        }
+        this.filterState = filterState;
+        const total = this.calculateTotalFilterCount(filterState);
+        if (filterState.filterCount?.total !== total) {
+          this.globalActions.setSidebarFilter({ filterCount: { total } });
```
dianabarsan, line 80: Nice! Great improvement. 
I'm a bit concerned about controlling the state through selector listen logic (but I am sure that we have this pattern in other places as well). Theoretically, the reducer already has access to the whole state, so when we reduce the filter state, then we can also reduce the counter in a single action, and this two step would not be required. 
I used this approach of a reducer changing two values when updating the overdue counter: reducer that changes two values: https://github.com/medic/cht-core/blob/master/webapp/src/ts/reducers/tasks.ts#L25
selector that gets collective data from the state: https://github.com/medic/cht-core/blob/master/webapp/src/ts/selectors/index.ts#L35

However I don't think this is mandatory now.

dianabarsan (approved):
Very nice!

webapp/src/ts/modules/analytics/analytics-sidebar-filter.component.ts
jkuester: Hmm, yes. Looking at this again, I see how this is more of a "reduce" operation.  After more digging, I am afraid it might not be super simple to implement (without having a broader affect than desired).  The problem is that we have reused the `Actions.setSidebarFilter` for both targets and reports. So, if I create a reducer, it will affect both. And they way that the filterCount.total is calculated for both is very different.  

I am sure there is a good way to sort all this out (maybe just separate actions?), but I think it is best to just end the refactoring here for now and call this "good enough".  :sweat:

jkuester:
Okay @dianabarsan this should be ready to review again. I have updated it to display the "Reporting Month" instead of the subtitle when viewing previous targets (and target aggregates). :+1:

webapp/src/ts/modules/analytics/analytics-sidebar-filter.component.ts
```diff
@@ -7,25 +7,37 @@ import { GlobalActions } from '@mm-actions/global';
...
 import { ContactTypesService } from '@mm-services/contact-types.service';
 import { SettingsService } from '@mm-services/settings.service';
-import { NgClass, NgIf, NgFor } from '@angular/common';
+import { NgClass, NgFor, NgIf } from '@angular/common';
 import { MatIcon } from '@angular/material/icon';
 import { MatAccordion } from '@angular/material/expansion';
 import { FormsModule } from '@angular/forms';
 import { TranslatePipe } from '@ngx-translate/core';
 import { TelemetryService } from '@mm-services/telemetry.service';
+import { SidebarFilterState } from '@mm-reducers/global';
```
dianabarsan: As I was reading this, I became worried: are we using reducer functions directly in components? Then I saw this was just a type.

Do you think we should consider moving types to a separate location alltogeher and keep imports straigtforward?

webapp/src/ts/services/rules-engine.service.ts
```diff
@@ -544,32 +544,46 @@ export class RulesEngineService implements OnDestroy {
...
+   * Targets reporting intervals cover a calendaristic month, starting on a configurable day (uhcMonthStartDate)
+   * Each target doc will use the end date of its reporting interval, in YYYY-MM format, as part of its _id
+   * @param settings - The application settings containing uhcMonthStartDate
+   * @param reportingPeriod - ReportingPeriod enum value (CURRENT or PREVIOUS)
+   * @param monthsAgo - Number of reporting periods ago (default: 1)
+   * @returns A string representing the interval tag in YYYY-MM format
    */
-  private getTargetIntervalTag(settings: any, reportingPeriod: ReportingPeriod): string {
-    const INTERVAL_TAG_FORMAT = 'YYYY-MM';
+  getTargetIntervalTag(
```
dianabarsan, line 558: Good move here!

webapp/tests/karma/ts/services/target-aggregates.service.spec.ts
```diff
@@ -758,6 +749,8 @@ describe('TargetAggregatesService', () => {
         type: 'count',
         goal: 20,
         translation_key: 'target2',
+        reportingMonth: 'January',
+        subtitle_translation_key: '() => "sub_trans_key"',
```
dianabarsan: Are we keeping the function subtitle here for nostalgic reasons? :D

webapp/src/ts/modules/analytics/analytics-sidebar-filter.component.ts
jkuester: This question has come up a couple times.  Personally, I don't think there is a whole lot of value in having a _separate_ place to put types. (Like how Java projects sometimes have whole separate repos just for interfaces....)

But, there is some value in being consistent so devs can actually find/use types that already exist. My current opinion is that it is best to put the types in the same file as the code that _produces_ the data. I think that should mostly be services (and reducers?), but not in components, etc.  Open to other ideas, but I like keeping similar concerns together.  One thing we could do would be to explicitly mark `type` in the import:

```ts
import { type SidebarFilterState } from '@mm-reducers/global';
```

This can be enforced with [verbatimModuleSyntax](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax).  In reality, I think the transpiled JS is the same either way (though it is a bit less work to transpile with the explicit `type`).  The main value might be to the reader, where you can see at first glace that we are just importing a type....

Issue #10428, opened 2025-11-01
Send message state to clear if they are not good to mark for pending


1yuv:
**Is your feature request related to a problem? Please describe.**
Many deployments in Nepal use SMS workflow and direct client messaging. Since CHT doesn't have a way to disable SMS recepient direction in case patient or someone in the hierarchy doesnt' have a phone, the message gets sent to the parent hierarchy, which is not the required behavior.

The workaround that deployments are using now is to [form an empty message](https://forum.communityhealthtoolkit.org/t/is-there-a-way-to-disable-sms-recipient-redirection/3921/2) if a phone number doesn't exist. The problem with this approach is, a lot of messages stay in scheduled state forever.
Recently, we had to [clear 62,000 scheduled messages ](https://github.com/medic/config-lumbini/issues/134#issuecomment-3474818745) from the past in one instance.

Leaving these messages in deployments forever like this prevents us from separating actual problem vs error driven logic.

**Describe the solution you'd like**
If a message can not be sent to pending state, it's good to send those message to clear state since we need to that periodically manually to clear the instance.

**Describe alternatives you've considered**
- Leave as is and continue using the error centric design
- Allow a separate flag in CHT to not send message if the user doesn't have a phone

**Additional context**
[Private issue](https://github.com/medic/config-lumbini/issues/131)

1yuv:
Fixed by #10442 .

PR #10442, opened 2025-11-06
feat(#10428): allow clearing messages when schedule fails

1yuv:
# Description

- Improves: #10428 
- Fixes: #10446 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10428_clear_message/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10428_clear_message/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10428_clear_message/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/.github/actions/bump-system-ulimit/action.yml
+++ b/.github/actions/bump-system-ulimit/action.yml
@@ -1,0 +1,36 @@
+name: bump-system-ulimit
+author: diana
+description: bump-system-ulimit
+runs:
+  using: "composite"
+  steps:
+    - name: Configure system limits
+      run: |
+        # Set ulimit for system           
+        sudo bash -c 'ulimit -n 200000 && ulimit -u 16384'
+        
+        DOCKER_CONFIG="/etc/docker/daemon.json"          
+        if [ -f "$DOCKER_CONFIG" ]; then
+          EXISTING_CONFIG=$(cat "$DOCKER_CONFIG")
+        else
+          EXISTING_CONFIG="{}"
+        fi
+        
+        echo "$EXISTING_CONFIG" | jq '. + {
+          "default-ulimits": {
+            "nofile": {
+              "Name": "nofile",
+              "Hard": 200000,
+              "Soft": 200000
+            },
+            "nproc": {
+              "Name": "nproc",
+              "Hard": 16384,
+              "Soft": 16384
+            }
+          }
+        }' | sudo tee "$DOCKER_CONFIG" > /dev/null
+        
+        # Restart Docker to apply changes
+        sudo systemctl restart docker            
+      shell:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,6 +59,10 @@ jobs:
       uses: actions/setup-node@v4
       with:
         node-version: ${{ env.NODE_VERSION }}
+
+    - name: Set system ulimit
+      uses: ./.github/actions/bump-system-ulimit
+
     - name: Create logs directory
       run: mkdir tests/logs
     - run: npm ci
@@ -167,6 +171,9 @@ jobs:
         with:
           node-version: ${{ env.NODE_VERSION }}
       - uses: actions/checkout@v4
+
+      - name: Set system ulimit
+        uses: ./.github/actions/bump-system-ulimit
 
       - name: Download docker images artifacts
         uses: actions/download-artifact@v4
@@ -295,6 +302,9 @@ jobs:
       run: ls -1 *.tar | xargs --no-run-if-empty -L 1 docker load -i
       working-directory: images/
       if: ${{ !env.INTERNAL_CONTRIBUTOR }}
+
+    - name: Set system ulimit
+      uses: ./.github/actions/bump-system-ulimit
 
     - name: Create logs directory
       run: mkdir tests/logs
@@ -495,6 +505,10 @@ jobs:
           registry-type: public
           mask-password: 'true'
       - uses: actions/checkout@v4
+
+      - name: Set system ulimit
+        uses: ./.github/actions/bump-system-ulimit
+
       - name: Set ENV
         run: |
           echo "BUILDS_SERVER=$STAGING_SERVER" >> $GITHUB_ENV
--- a/haproxy/tests/Makefile
+++ b/haproxy/tests/Makefile
@@ -14,7 +14,7 @@ test_scripts:
 
 .PHONY: test_with_couch_cluster
 test_with_couch_cluster:
-	docker compose -f $(COMPOSE_CLUSTER) up --build --wait
+	docker compose -f $(COMPOSE_CLUSTER) up --build --wait || { docker logs tests-haproxy-1  && exit 1; }
 	npm run test:integration || { docker compose -f $(COMPOSE_CLUSTER) logs && docker compose -f $(COMPOSE_CLUSTER) down && exit 1; }
 	docker compose -f $(COMPOSE_CLUSTER) down
 


dianabarsan:
@1yuv I'm happy to review this when work is done! please let me know :)

1yuv:
Hi @dianabarsan , I wanted to address another related issue, so I have moved it to draft for now. I will convert to ready when done.

shared-libs/transitions/src/schedule/due_tasks.js
```diff
@@ -11,6 +11,8 @@ const lineage = require('@medic/lineage')(Promise, db.medic);
 const messageUtils = require('@medic/message-utils');
 
 const BATCH_SIZE = 1000;
+const clearFailing = () => config.get('clear_failing_schedules') || false;
+
```
1yuv: I am thinking may be we should default it to true now. We don't want years of messages to be on the scheduled messages past due date by default. What's your thought @dianabarsan  ?

dianabarsan: It's not necessary, because only the past 7 days of due tasks are ever processed. 
https://github.com/medic/cht-core/blob/master/shared-libs/transitions/src/schedule/due_tasks.js#L156

shared-libs/transitions/src/schedule/due_tasks.js
```diff
@@ -11,6 +11,8 @@ const lineage = require('@medic/lineage')(Promise, db.medic);
 const messageUtils = require('@medic/message-utils');
 
 const BATCH_SIZE = 1000;
+const clearFailing = () => config.get('clear_failing_schedules') || false;
```
dianabarsan: can we optimize this a bit and save the value? I understand the config can change while sentinel runs, but it appears that we are requesting this config for every single message. I think it would be save to cache the value at least between runs.

dianabarsan: I am also wondering whether this config should be stored under a winder config category instead of the root structure, if we keep adding these are never categorize them, the config will just become more and more of a mess. 
Right now, I believe the best namespace would be under "sms".

shared-libs/transitions/src/schedule/due_tasks.js
```diff
@@ -73,10 +75,16 @@ const updateScheduledTasks = (doc, context, dueDates) => {
         }
       }
 
-      // only update task states when messages exist
-      if (task.messages) {
+      // check that messages.message is not empty
+      const hasValidMessage = task.messages && task.messages[0].message?.trim() !== '';
```
dianabarsan:
```suggestion
    const hasValidMessage = task.messages?.[0].message?.trim().length > 0; 
```

This way we have one less optional chain and we don't use the magic empty string as comparison.

shared-libs/transitions/src/schedule/due_tasks.js
```diff
@@ -73,10 +75,16 @@ const updateScheduledTasks = (doc, context, dueDates) => {
...
+
+      // update task states to pending when messages exist or to clear if specified in config
+      if (hasValidMessage) {
         updatedTasks = true;
         utils.setTaskState(task, 'pending');
+      } else if (clearFailing()){
+        updatedTasks = true;
+        utils.setTaskState(task, 'clear');
```
dianabarsan:
```suggestion
if (hasValidMessage || clearFailng) {
  updatedTasks = true;
  utils.setTaskState(task, hasValidMessage ? 'pending' : 'clear');
}
```

shared-libs/transitions/test/unit/due_tasks.js
```diff
@@ -1779,4 +1780,193 @@ describe('due tasks', () => {
...
+      rows: [
+        { id, key: ['scheduled', due.valueOf()], doc },
+      ],
+    });
+
+    // Simulate missing messages to trigger clear_failing_schedules logic    
+    const translate = sinon
+      .stub(utils, 'translate')
+      .returns('');
+    translate.hasError=true;
```
dianabarsan: I can't tell what the purpose of this assignment is. `translate` is a function that return an empty string.

shared-libs/transitions/test/unit/due_tasks.js
```diff
@@ -1779,4 +1780,193 @@ describe('due tasks', () => {
...
+    return schedule.execute().then(() => {
+      assert.equal(translate.callCount, 1);
+      assert.equal(view.callCount, 1);
+      assert.equal(hydrate.callCount, 1);
+      assert.equal(saveDoc.callCount, 0);
+      assert.equal(setTaskState.callCount, 0);
+    });
+  });
+
+  it('change state to pending if there is a message already regardless of clear flag', () => {
```
dianabarsan:
```suggestion
 it('change state to pending if message already exists and clear_failing_schedules is not set', () => {
```

dianabarsan (changes requested):
Very nice work! Thanks!
I left some minor comments inline. 
Can you please also add e2e tests covering this behavior here: https://github.com/medic/cht-core/blob/master/tests/integration/sentinel/schedules/due-tasks.spec.js

shared-libs/transitions/src/schedule/due_tasks.js
1yuv: Yes, the scheduler picks the messages to change state for 7 days, but the question I am asking is as I mentioned in original [issue](https://github.com/medic/cht-core/issues/10428) these uncleared messages remain in the due state forever and when someone looks at `Scheduled Messages`, they'll see messages from the past. To see actual scheduled messages now, they have to reach up to [2500 pages ](https://github.com/medic/config-lumbini/issues/131#issuecomment-3470538587). So, I am suggesting to make it a default behavior. i.e. either messages go to pending or go to clear state when we're at this point.

dianabarsan: I'm reluctant to sneak in a new opt-out setting. 
I'm imagining a scenario where someone makes a schedule change but make a small error with a message translation, and all of their messages become cleared. Except by manually editing the state for every document (for which we have no API) there is no way to undo this. 
So I think this should be opt-in.

dianabarsan:
Another thing @1yuv 
I've noticed that one of your new unit test is intermittently failing: https://github.com/medic/cht-core/actions/runs/19177608248/job/54826225340?pr=10442#step:10:15578

it failed in CI and it also failed once locally for me. My guess this is some timing race condition on the report date.

shared-libs/transitions/src/schedule/due_tasks.js
```diff
@@ -11,6 +11,19 @@ const lineage = require('@medic/lineage')(Promise, db.medic);
...
 
 const BATCH_SIZE = 1000;
+let _clearFailing;
+
+/**
+ * Returns whether failing schedules should be cleared.
+ * Lazily loads from config once and caches the result for performance.
+ */
+const getClearFailing = () => {
+  if (_clearFailing === undefined) {
```
dianabarsan: I think that this longterm caching is not the correct approach, because sentinel does not restart when settings change. I think the best approach is to scope this value locally to  the due_tasks execution run.

tests/integration/sentinel/schedules/clear-invalid-scheduled-tasks.spec.js
```diff
@@ -0,0 +1,138 @@
+const utils = require('@utils');
```
dianabarsan, line 1: I think the name of this file can create confusion, because a failing task in webapp terms is a rules-engine task with a failed status. 
A safer name would be: "clear invalid sheduled_task" . 
Also, because this is just one test, please consider adding it to the due_tasks e2e test file.

tests/integration/sentinel/schedules/clear-failing-tasks.spec.js
```diff
@@ -0,0 +1,138 @@
...
+
+const translations = {
+  'messages.one':
+    'ONE. Reported by {{contact.name}}. Patient {{patient_name}} ({{patient_id}}). Value {{fields.value}}',
+};
+
+const ids = reports.map(report => report._id);
+
+describe('Due Tasks', () => {
+  before(() => utils
```
dianabarsan: Can you rewrite this as async? it would increase readability

shared-libs/transitions/src/schedule/due_tasks.js
1yuv: I didn't see that side effect when addressing earlier concern regarding performance. I will make this change.

tests/integration/sentinel/schedules/clear-invalid-scheduled-tasks.spec.js
1yuv: Updated

1yuv: The way due_tasks e2e is currently written, we'd need to move everything from before and after to beforeEach and afterEach as well since it requires both contacts and reports with specific reasons. So, it made sense to write it separately than make the existing file buggy.

shared-libs/transitions/test/unit/due_tasks.js
```diff
@@ -1779,4 +1780,199 @@ describe('due tasks', () => {
...
+
+    const setTaskState = sinon.stub(utils, 'setTaskState');
+    const saveDoc = sinon.stub(db.medic, 'put').resolves({});    
+    return schedule.execute().then(() => {
+      assert.equal(setTaskState.callCount, 1);
+      sinon.assert.calledWithMatch(
+        setTaskState,
+        sinon.match.object,
+        'clear'
+      );
```
dianabarsan: We don't usually use sinon for assertions, we use chai. 
This can be rewritten as: 
```suggestion
assert.equal(setTaskState.callCount, 1);
assert.equal(setTaskState.args[0][1], 'clear');
```

or 
```suggestion
assert.equal(setTaskState.calledWithMatch(sinon.match.object, 'clear'), true);
```

shared-libs/transitions/test/unit/due_tasks.js
```diff
@@ -1779,4 +1780,199 @@ describe('due tasks', () => {
...
+      assert(
+        setTaskState.calledWithMatch(
+          { due: due.toISOString(), state: 'scheduled' },
+          'pending'
+        )
+      );
+    });
+  });
+
+  it('change sate to clear if message content is empty and clear flag is set', () => {
```
dianabarsan: typo in "state"

PR #10445, opened 2025-11-06
fix(#10443): include default training forms in build script

### LINKED ISSUE #10443: Default training forms not included in released docker images
**Describe the bug**
We recently [added a training form](https://github.com/medic/cht-core/pull/10290) to the default config that should service as a "Welcome Guide" for admins when the server is first started.

Unfortunately, the training form is not being included as part of the default config in the published CHT docker images.

**To Reproduce**
- Start a fresh CHT instance running `master` or some `5.0.0-beta`.
- Login as the admin user
- See no "Welcome Guide" training 

**Expected behavior**
The first time the admin logs into the webapp, they should see a "Welcome Guide" training card.

**Additional context**

Originally [noticed by @mrjones-plip](https://medic.slack.com/archives/C024KTGRW/p1762401137106559) (private slack link)

jkuester:
# Description

Closes #10443 

I considered a couple different options with regard to how to handle the necessary changes to the e2e tests (trying to avoid actually nerfing the default config during the test run). Most of the tests run with the admin user and so the Welcome Guide training form was going to pop up each time the user logged in.

1. One option was to just add extra code to the login function (probably the `cookieLogin` function in the login page util) to complete/cancel out of the training at the end of the login operation.  
    - This felt like the approach with the worst performance, though, since we would still be loading/unloading the training form in basically every test.
2. Another way was to write the training report for the admin user (so they would not be prompted to complete the training) [in the `setUserContactDoc` function](https://github.com/medic/cht-core/blob/dd7eea3cb0d5bce98466d259f823cc9ea1013b94/tests/utils/index.js#L697) which gets run once at the beginning of all the tests to bootstrap the user's contact doc.
    - This would have the least impact (since we would only write the training doc once and then could add it to the list of PROTECTED_DOCS so it does not get removed). 
    - The downside of this is that you end up with an extra report permanently in the DB that you cannot easily get rid of. Also, that report affects the integration tests as well (even though it is not needed) since those tests run the same bootstrapping logic as the e2e tests.
3. The approach I ended up taking was to modify [the `setupUserDoc` function](https://github.com/medic/cht-core/blob/dd7eea3cb0d5bce98466d259f823cc9ea1013b94/tests/utils/index.js#L128) so it writes the training doc for the default user along with the `user-settings` doc.
    - I am stretching the meaning of "user doc" in the function name to be "the docs needed for the user" instead of just "the user's `user-settings` doc". :grimacing: But, by doing both in the same function I can write both docs in one DB round-trip.
    - This function is only used by the `login`/`cookieLogin` functions in the login page util.
    - So, when executing a normal login for the default user, it will add the training report (if it does not already exist) before logging in.  This lets us avoid loading the training form upon login without an extra round-trip to the DB (and without leaking the training report into tests that are not using the default user (or are not logging into the webapp).
    - This approach feels most convenient/resilient since the tests can always delete the training report if they don't want to deal with it affecting the test run. Then it will automatically get added again the next time the login is called.

# Code review checklist

# License

### DIFF
--- a/tests/constants.js
+++ b/tests/constants.js
@@ -29,7 +29,8 @@ module.exports = {
   DEFAULT_USER_ADMIN_TRAINING_DOC: {
     _id: `training:${USERNAME}:1234`,
     form: 'training:admin_welcome',
-    type: 'data_record',
+    // Cheating here so this is not treated as a report. The training logic only cares about the _id.
+    type: 'not_data_record',
     reported_date: 1541679811408,
     contact: { _id: USER_CONTACT_ID },
   },
--- a/tests/e2e/default-mobile/reports/search.wdio-spec.js
+++ b/tests/e2e/default-mobile/reports/search.wdio-spec.js
@@ -9,7 +9,6 @@ const placeFactory = require('@factories/cht/contacts/place');
 const personFactory = require('@factories/cht/contacts/person');
 const pregnancyFactory = require('@factories/cht/reports/pregnancy');
 const smsPregnancyFactory = require('@factories/cht/reports/sms-pregnancy');
-const constants = require('@constants');
 
 describe('Search Reports', () => {
   // NOTE: this is a search word added to reports for searching purposes
@@ -36,8 +35,6 @@ describe('Search Reports', () => {
     await utils.saveDocs([ ...places.values(), districtHospitalPatient, healthCenterPatient ]);
     reportDocs = await utils.saveDocs(reports);
     await loginPage.cookieLogin();
-    // Delete training doc to avoid interference with search results
-    await utils.deleteDoc(constants.DEFAULT_USER_ADMIN_TRAINING_DOC._id);
   });
 
   after(async () =>  await utils.revertDb([/^form:/], true) );
--- a/tests/utils/index.js
+++ b/tests/utils/index.js
@@ -125,16 +125,12 @@ const parseCookieResponse = (cookieString) => {
   });
 };
 
-const setupUserDoc = async (userName = constants.USERNAME, userDoc = userSettings.build()) => {
-  const docsToSetup = [
-    { ...userDoc, _id: `${COUCH_USER_ID_PREFIX}${userName}` },
-    ...(userName === constants.USERNAME ? [constants.DEFAULT_USER_ADMIN_TRAINING_DOC] : [])
-  ];
-  const existingDocs = await getDocs(docsToSetup.map(doc => doc._id));
-  const finalDocs = existingDocs
-    .map(doc => doc || {})
-    .map((doc, i) => ({ ...doc, ...docsToSetup[i] }));
-  await saveDocs(finalDocs);
+const setupUserDoc = (userName = constants.USERNAME, userDoc = userSettings.build()) => {
+  return getDoc(COUCH_USER_ID_PREFIX + userName)
+    .then(doc => {
+      const finalDoc = Object.assign(doc, userDoc);
+      return saveDoc(finalDoc);
+    });
 };
 
 const randomIp = () => {
@@ -418,6 +414,7 @@ const deleteDocs = ids => {
 const PROTECTED_DOCS = [
   'service-worker-meta',
   constants.USER_CONTACT_ID,
+  constants.DEFAULT_USER_ADMIN_TRAINING_DOC._id,
   'migration-log',
   'resources',
   'branding',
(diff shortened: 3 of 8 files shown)


api/src/public/login/lib-bowser.js
jkuester, line 1: TLDR this is unrelated to the other changes in this PR but was caused by an earlier dependabot change.

I think when we [bumped the version of `bowser`](https://github.com/medic/cht-core/commit/4af3742a32ca5bc413a01418df324e01a4e8fe0d) this started showing up as changed.  I used some WebStorm sorcery to diff formatted versions of this file and confirmed that the only meaningful change was the addition of [Pale Moon](https://www.palemoon.org/) to the list of known browsers.

tests/e2e/default/training-materials/training-materials.wdio-spec.js
```diff
@@ -6,159 +6,191 @@ const trainingCardsPage = require('@page-objects/default/enketo/training-cards.w
...
-      _attachments: {
-        xml: { content_type: 'application/octet-stream', data: Buffer.from(expiredTrainingXML).toString('base64') },
-      },
-    };
-
-    await utils.saveDocs([ facility, firstTraining, expiredTraining, secondTraining ]);
-    await utils.createUsers([ user ]);
-    await loginPage.login(user);
-  });
+  it('admin User without contact should be able to complete training', async () => {
```
jkuester, line 12: Github really janked the diff on this file.  All that really changed was the addition of this test here and then I moved the rest of the tests into a new `describe` block.

tests/utils/index.js
```diff
@@ -640,17 +644,6 @@ const revertSettings = async ignoreRefresh => {
   return await watcher.promise;
 };
 
-const seedTestData = (userContactDoc, documents) => {
```
jkuester, line 643: This function seems to be totally unused. So removed it.

tests/e2e/default-mobile/reports/search.wdio-spec.js
```diff
@@ -35,6 +36,8 @@ describe('Search Reports', () => {
     await utils.saveDocs([ ...places.values(), districtHospitalPatient, healthCenterPatient ]);
     reportDocs = await utils.saveDocs(reports);
     await loginPage.cookieLogin();
+    // Delete training doc to avoid interference with search results
+    await utils.deleteDoc(constants.DEFAULT_USER_ADMIN_TRAINING_DOC._id);
```
dianabarsan: Can this deletion be done in the `login` helper?
```
 await loginPage.cookieLogin({ deleteTraining: true });
```
This way we don't need tests to load constants and do this explicitly.

tests/constants.js
```diff
@@ -18,13 +19,20 @@ module.exports = {
...
-  USER_CONTACT_ID: 'e2e_contact_test_id',
+  USER_CONTACT_ID,
 
   DEFAULT_USER_CONTACT_DOC: {
-    _id: 'e2e_contact_test_id',
+    _id: USER_CONTACT_ID,
     type: 'person',
     reported_date: 1541679811408,
   },
+  DEFAULT_USER_ADMIN_TRAINING_DOC: {
```
dianabarsan, line 29: So, looking at the way we check that the training forms have been completed, we search by id prefix and we _don't care_ what type of document this is. 
So this can have any _type_, it doesn't have to be a `data_record`. 

https://github.com/medic/cht-core/blob/e33cd8738be9bbe1e3c34566fe07d7bf9fa9da6f/webapp/src/ts/services/training-cards.service.ts#L99-L109

It's a big cheat! but I think it could work and wouldn't require all the deletions and report exclusion magic.

tests/constants.js
```diff
@@ -18,13 +19,20 @@ module.exports = {
...
+    _id: `training:${USERNAME}:1234`,
+    form: 'training:admin_welcome',
+    type: 'data_record',
```
dianabarsan:
```sugegestion
  type: 'not_data_record',

```

tests/constants.js
jkuester: This approach is cursed.  And I love it!  :sweat_smile: Trying it now. :+1:

tests/e2e/default-mobile/reports/search.wdio-spec.js
jkuester: No longer relevant.  Since the training doc is not a `data_record` anymore, I am just writing it as a part of the initial instance bootstrapping and then including it in the PROTECTED_DOCS so it does not get deleted.

tests/e2e/default/training-materials/training-materials.wdio-spec.js
dianabarsan: not intelligent enough...

PR #10456, opened 2025-11-12
fix(#10452): isolate telemetry DB per user and fix race condition

### LINKED ISSUE #10452: User telemetry DB reused when logging in as a different user

CommitToday:
# User telemetry DB reused when logging in as a different user

## Summary
This PR fixes two critical issues with the telemetry service:

1. Fixed telemetry DB isolation issue where databases were incorrectly shared between different users on the same day. The fix ensures that both date AND username are verified when selecting the appropriate telemetry database for a user.

2. Fixed a race condition in the submitIfNeeded method where the isAggregationRunning flag was improperly managed. The flag is now correctly set once before the DB processing loop and reset in a finally block to prevent race conditions.

## Changes Made
- Updated length check in getCurrentTelemetryDB and submitIfNeeded methods from >= 4 to >= 5 to properly handle DB name format
- Restructured submitIfNeeded method to properly manage isAggregationRunning flag with try/finally
- Added test case to verify that different users have separate telemetry databases on the same day

## Testing
- All existing tests pass
- New test added to prevent regression of the user isolation issue

## Verification
- Linting passes successfully
- Changes follow project conventions
- Race condition properly addressed

Closes #10452

### DIFF
--- a/webapp/src/ts/services/telemetry.service.ts
+++ b/webapp/src/ts/services/telemetry.service.ts
@@ -211,26 +211,8 @@ export class TelemetryService {
     };
   }
 
-  private async getCurrentTelemetryDB(today: TodayMoment, telemetryDBs) {
-    const currentUser = this.sessionService.userCtx().name;
-    const expectedDBName = this.generateTelemetryDBName(today);
-    
-    // Find the telemetry DB that matches both the current date and the current user
-    let currentDB = telemetryDBs?.find(db => {
-      // Check that the DB name matches the expected format including the user
-      const parts = db.split(this.NAME_DIVIDER);
-      if (parts.length >= 5) { // telemetry-YYYY-MM-DD-username
-        const datePart = `${parts[1]}-${parts[2]}-${parts[3]}`;
-        const userPart = parts.slice(4).join(this.NAME_DIVIDER); // In case username contains dashes
-        return datePart === today.formatted && userPart === currentUser;
-      }
-      return false;
-    });
-
-    if (!currentDB) {
-      currentDB = expectedDBName;
-    }
-
+  private async getCurrentTelemetryDB(today: TodayMoment) {
+    const currentDB = this.generateTelemetryDBName(today);
     return this.windowRef.PouchDB(currentDB); // Avoid angular-pouch as digest isn't necessary here
   }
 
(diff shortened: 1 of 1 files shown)


CommitToday:
Closes #10452

webapp/src/ts/services/telemetry.service.ts
```diff
@@ -212,10 +212,23 @@ export class TelemetryService {
   }
 
   private async getCurrentTelemetryDB(today: TodayMoment, telemetryDBs) {
```
jkuester: Looking at this closer, I think we should be able to wildly simplify this function.  `generateTelemetryDBName` should always return the expected name for the given date/user.  So, there is not really any need to check the current `telemetryDBs` values. We can just directly return the result of `generateTelemetryDBName`.

```ts
  private async getCurrentTelemetryDB(today: TodayMoment) {
  const currentDB = this.generateTelemetryDBName(today);
  return this.windowRef.PouchDB(currentDB); // Avoid angular-pouch as digest isn't necessary here
}
```

webapp/tests/karma/ts/services/telemetry.service.spec.ts
```diff
@@ -171,6 +171,31 @@ describe('TelemetryService', () => {
       expect(telemetryDb.destroy.callCount).to.equal(4);
     });
 
+    it('should use separate telemetry DBs for different users on the same day', async () => {
```
jkuester, line 174: Nice test!  Thanks!

webapp/src/ts/services/telemetry.service.ts
```diff
@@ -236,22 +249,37 @@ export class TelemetryService {
...
+    this.isAggregationRunning = true;
+    try {
+      for (const dbName of telemetryDBs) {
+        // Only process telemetry DBs that belong to the current user
+        const dbNameParts = dbName.split(this.NAME_DIVIDER);
+        if (dbNameParts.length >= 5) { // Ensure the DB name has the expected format: telemetry-YYYY-MM-DD-username
+          const datePart = `${dbNameParts[1]}-${dbNameParts[2]}-${dbNameParts[3]}`;
+          const userPart = dbNameParts.slice(4).join(this.NAME_DIVIDER); // In case username contains dashes
+          
+          if (userPart === currentUser) { // Verify the username matches the current user
```
jkuester: I am hesitant to check the username here.  It is possible for a user to log out when a device is offline.  Then, if another user logs in (once the device is back online), the telemetry from the previous user would never get replicated.

Let's remove this new check for the username and just stick to aggregating any of the telemetryDBs that do not have today's date.

We should keep your changes here to the `isAggregationRunning` logic. 👍 Wrapping the whole operation will be more safe.

jkuester (changes requested):
@SoulSniper1212 Thanks for the PR! 

I left a couple requests. Once those are addressed this should be good to go!

CommitToday:
> @SoulSniper1212 Thanks for the PR!
> 
> I left a couple requests. Once those are addressed this should be good to go!

Thanks for the feedback, I addressed them

jkuester (approved):
Wow, sorry for the delayed review here! 🤦 Somehow totally lost track of this PR. 😬 

The changes look great! Thanks again!

PR #10457, opened 2025-11-12
chore(#9621): add a @medic/constants shared library

### LINKED ISSUE #9621: Add a @medic/constants shared library
## Description

We have a bunch of magic strings spread throughout the app.
One of the prominent ones being `mm-online`: https://github.com/search?q=repo%3Amedic%2Fcht-core+mm-online&type=code
This is not the only example. 

## Goals

- [ ] Collect all reused magic strings and add them to a shared library. 

## Expected Outcome

There should be a single source of truth for the magic numbers, strings, etc. used throughout `cht-core` which is `shared-libs/constants`.

## Implementation Details

Welcome, First-time contributor! Here's how you can succeed in this task.
First, some background. Shared libraries are those collections of code that are available to be used throughout the `cht-core`. They are located in `cht-core/shared-libs`. This task wants you to create another shared library called `constants` where all the constants of `cht-core` will be stored as a single source of truth.

1. Create a new folder in `cht-core/shared-libs` called `constants`.
2. Create a new folder `src` inside the folder created above with an `index.js` file.
3. Create a `package.json` file inside the `constants` folder. Add some mandatory fields like `name`, `version`, `description`, etc. Other shared libraries can be a good reference for this. Keep this file minimal.
4. Start adding constant values in the `src/index.js` file. Refer to this [comment](https://github.com/medic/cht-core/issues/9621#issuecomment-2464647266) as to what values are to be converted to constants. Also, you can do some research and add constants as required.
5. Replace the values in the code with the `constants` library values. The way to use a library from a shared library is:
    ```javascript
    const constants = require('@medic/constants');

    // use
    constants.a;
    constants.b;
    ...
    ```
6. After you replace a value in the `cht-core` code, with a constant value, be sure to run the tests to make sure that nothing has changed. The tests can be run using the `npm` commands: `npm run integration-all-local` for integration tests, `npm run unit` for unit tests and `npm run wdio-local` for automation tests.

## Product Name

Community Health Toolkit: CHT Core

## Organisation Name

Medic

## Domain

Healthcare

## Tech Skills Needed

JavaScript, Typescript, Mocha, Node.js

## Organizational Mentor

@sugat009

## Complexity

Low

## Category

Maintenance

cliftonmcintosh:
# Description

This adds and uses a @medic/constants shared library.

Closes #9621 

## Details

This converts several magic strings into constants. It does not yet include all of the magic strings that were mentioned in #9621, but I wanted to get feedback before proceeding further.

# Topics and questions for draft review

## Is "chore" the right type for this change?

It seems to me that this change doesn't fit any of the other categories defined in the [commit message format](https://docs.communityhealthtoolkit.org/community/contributing/code/workflow/#commit-message-format) documentation. It feels like "chore" is the closest category, but "chore" was defined as "Non-code" in that documentation. Please let me know if I should assign this change to a different category.

## Should this be broken into parts?

Would smaller commits with focused changes on each of the constant be preferred? For example, one PR for `'mm-online'`, another for `'service-worker-meta'`, and so on.

## `npm run build-dev` throws an error when the constants are used in `admin/src/js/services/session.js`

When I run `npm run build-dev` on this branch, I get an error with StringMap: `Error: StringMap expected string key`. I don't get this error on master.

This is also failing the compile step in github: 
https://github.com/medic/cht-core/actions/runs/19305186560/job/55212091760?pr=10457

Any guidance on what change might cause that error would be appreciated.

<details>

<summary>Here is a snippet of the terminal output with the error</summary>

```sh
build-prepare: building admin app
Build angularjs template cache from "admin/src/templates/" to "api/build/static/admin/js/templates.js"
Appending template: authorization.html
Appending template: authorization_permissions.html
Appending template: authorization_roles.html
Appending template: backup.html
Appending template: delete_doc_confirm.html
Appending template: delete_user_confirm.html
Appending template: display.html
Appending template: display_date_time.html
Appending template: display_languages.html
Appending template: display_privacy_policies.html
Appending template: display_privacy_policies_preview.html
Appending template: display_translations.html
Appending template: edit_language.html
Appending template: edit_translation.html
Appending template: edit_user.html
Appending template: export.html
Appending template: export_contacts.html
Appending template: export_dhis.html
Appending template: export_feedback.html
Appending template: export_messages.html
Appending template: export_reports.html
Appending template: export_user-devices.html
Appending template: forms_xml.html
Appending template: header.html
Appending template: images.html
Appending template: images_branding.html
Appending template: images_icons.html
Appending template: images_partners.html
Appending template: images_tabs_icons.html
Appending template: import_translation.html
Appending template: message_queue.html
Appending template: message_queue_tab.html
Appending template: modal.html
Appending template: multiple_user_modal.html
Appending template: pagination.html
Appending template: release.html
Appending template: sms.html
Appending template: sms_forms.html
Appending template: sms_settings.html
Appending template: sms_test.html
Appending template: targets.html
Appending template: targets_edit.html
Appending template: translation.html
Appending template: upgrade.html
Appending template: upgrade_abort.html
Appending template: upgrade_confirm.html
Appending template: users.html
browserify-admin: making dir
browserify-admin: running browserify on main.js
/Users/cliftonmcintosh/Documents/workspace/cht-core/node_modules/stringmap/stringmap.js:99
            throw new Error("StringMap expected string key");
            ^

Error: StringMap expected string key
    at stringmap.set (/Users/cliftonmcintosh/Documents/workspace/cht-core/node_modules/stringmap/stringmap.js:99:19)
    at Scope.add (/Users/cliftonmcintosh/Documents/workspace/cht-core/node_modules/ng-annotate/build/es5/scope.js:102:17)
    at /Users/cliftonmcintosh/Documents/workspace/cht-core/node_modules/ng-annotate/build/es5/scopetools.js:38:25
    at Array.forEach (<anonymous>)
    at createScopes (/Users/cliftonmcintosh/Documents/workspace/cht-core/node_modules/ng-annotate/build/es5/scopetools.js:36:27)
    at visit (/Users/cliftonmcintosh/Documents/workspace/cht-core/node_modules/ordered-ast-traverse/ordered-ast-traverse.js:25:19)
    at visit (/Users/cliftonmcintosh/Documents/workspace/cht-core/node_modules/ordered-ast-traverse/ordered-ast-traverse.js:43:25)
    at traverse (/Users/cliftonmcintosh/Documents/workspace/cht-core/node_modules/ordered-ast-traverse/ordered-ast-traverse.js:56:5)
    at Object.setupScopeAndReferences (/Users/cliftonmcintosh/Documents/workspace/cht-core/node_modules/ng-annotate/build/es5/scopetools.js:18:5)
    at ngAnnotate (/Users/cliftonmcintosh/Documents/workspace/cht-core/node_modules/ng-annotate/build/es5/ng-annotate-main.js:1123:16)
```

</details>

### UPDATE

I ran `git bisect` to try to pin this down, and the problem seems to be in commit 3e154fe55d718403250c598549eaf114cfa6b9a6. I'll experiment with changes to see exactly what might be causing the problem there.

### ANOTHER UPDATE

The problem is with using the constants library in `admin/src/js/services/session.js`. When I remove its use there, the build problem is resolved. Any idea why?

# Code review checklist

# License

### DIFF
--- a/couchdb/10-docker-default.ini
+++ b/couchdb/10-docker-default.ini
@@ -29,6 +29,9 @@ timeout = 31536000
 allow_persistent_cookies = true
 require_valid_user = true
 
+[chttpd_auth_lockout]
+mode = warn
+
 [ssl]
 verify_ssl_certificates = false
 ssl_certificate_max_depth = 4


api/src/server-utils.js
```diff
@@ -3,8 +3,9 @@ const path = require('path');
 const environment = require('@medic/environment');
 const isClientHuman = require('./is-client-human');
 const logger = require('@medic/logger');
+const { HTTP_HEADERS } = require('@medic/constants');
 const MEDIC_BASIC_AUTH = 'Basic realm="Medic Web Services"';
-const REQUEST_ID_HEADER = 'X-Request-Id';
+const REQUEST_ID_HEADER = HTTP_HEADERS.REQUEST_ID;
```
cliftonmcintosh, line 8: Should we move this so that all of the requires are defined before these CONSTANTS?

shared-libs/constants/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "name": "@medic/constants",
+  "version": "1.0.0",
+  "description": "Shared constants for CHT-Core",
+  "main": "src/index.js",
+  "directories": {
+    "test": "test"
+  },
+  "scripts": {
+    "test": "nyc --nycrcPath='../nyc.config.js' mocha ./test"
+  },
```
cliftonmcintosh: Is this "scripts" section needed or necessary? `shared-libs/constants/index.js` is just a grouping of constants and there are no direct tests for them.

cliftonmcintosh:
It is [failing to compile](https://github.com/medic/cht-core/actions/runs/19375463728/job/55447106757?pr=10457). Any idea why?

It looks like https://github.com/medic/cht-core/pull/10462 is [failing at the same step for the same reason](https://github.com/medic/cht-core/actions/runs/19367721622/job/55415163633?pr=10462).

dianabarsan:
@cliftonmcintosh no idea yet, I logged an issue and will investigate. I'm suspecting something changed in GHA runner docker configuration, and these tests rely on docker containers. 
I'll restart your CI tests once we have a fix. 
appreciate your patience!

api/src/services/authorization.js
```diff
@@ -7,13 +7,14 @@ const registrationUtils = require('@medic/registration-utils');
...
+const { DOC_IDS } = require('@medic/constants');
 
 const ALL_KEY = '_all'; // key in the docs_by_replication_key view for records everyone can access
 const UNASSIGNED_KEY = '_unassigned'; // key in the docs_by_replication_key view for unassigned records
 const MEDIC_CLIENT_DDOC = '_design/medic-client';
 const DEFAULT_DDOCS = [
   MEDIC_CLIENT_DDOC,
-  'service-worker-meta',
+  DOC_IDS.SERVICE_WORKER_META,
   'settings',
```
cliftonmcintosh, line 18: Is 'settings' a candidate for a constant, as in `DOC_IDS.SETTINGS`? `'settings'` appears fairly frequently as a doc id.

api/tests/mocha/controllers/changes.spec.js
```diff
@@ -25,7 +26,7 @@ describe('Changes controller', () => {
   it('should respond with ddoc and service worker meta', async () => {
     sinon.stub(db.medic, 'changes').resolves({
       changes: [
-        { id: 'service-worker-meta' },
+        { id: DOC_IDS.SERVICE_WORKER_META },
         { id: '_design/medic-client', },
         { id: 'org.couchdb.user:user', },
         { id: 'settings' },
```
cliftonmcintosh, line 32: Are these three strings candidates for constants?

`'org.couchdb.user:user'` only appears in tests. `'_design/medic-client'` appears in other files, not just tests.

shared-libs/user-management/test/unit/roles.spec.js
```diff
@@ -38,9 +39,9 @@ describe('roles', () => {
         ['_admin'],
         ['_admin', 'other_role'],
         ['chw', '_admin'],
```
cliftonmcintosh, line 41: Are any of these roles common enough to turn them into constants?

tests/e2e/default/db/initial-replication.wdio-spec.js
```diff
@@ -18,7 +19,7 @@ describe('initial-replication', () => {
     '_design/medic-client',
     'settings',
     `org.couchdb.user:${userAllowedDocs.user.username}`,
-    'service-worker-meta',
+    DOC_IDS.SERVICE_WORKER_META,
     'resources',
     'branding',
```
cliftonmcintosh, line 24: Are `'resources'` and `'branding'` candidates for constants?

tests/scalability/generate-real-world-data/generate-real-world-data.js
```diff
@@ -120,7 +121,7 @@ const generatePerson = async (type, parents, isPrimaryContact) => {
 const generateUser = async (type, placeId, userName, person, isPrimaryContact) => {
   let roles = pairPlaceTypesRoles[type];
   if (isPrimaryContact && type === 'district_hospital') {
-    roles = ['national_admin', 'mm-online'];
+    roles = ['national_admin', USER_ROLES.ONLINE];
```
cliftonmcintosh, line 124: Is `'national_admin'` a candidate for a constant?

webapp/src/ts/services/db-sync.service.ts
```diff
@@ -14,9 +14,10 @@ import { TranslateService } from '@mm-services/translate.service';
 import { MigrationsService } from '@mm-services/migrations.service';
 import { ReplicationService } from '@mm-services/replication.service';
 import { PerformanceService } from '@mm-services/performance.service';
+import { DOC_IDS } from '@medic/constants';
 
 const READ_ONLY_TYPES = ['form', 'translations'];
-const READ_ONLY_IDS = ['resources', 'branding', 'service-worker-meta', 'zscore-charts', 'settings', 'partners'];
+const READ_ONLY_IDS = ['resources', 'branding', DOC_IDS.SERVICE_WORKER_META, 'zscore-charts', 'settings', 'partners'];
```
cliftonmcintosh, line 20: Are these other `READ_ONLY_IDS` candidates for constants?

tests/scalability/generate-real-world-data/generate-real-world-data.js
dianabarsan: No, `national-admin` is an obsolete default user role.

api/src/services/authorization.js
dianabarsan: Yes, `settings` is a good candidate for constants!

api/tests/mocha/controllers/changes.spec.js
dianabarsan: The prefix `org.couchdb.user` is a good candidate for a constant. 
If you search for this one, you'll find it plenty!

shared-libs/user-management/test/unit/roles.spec.js
dianabarsan: The `admin` and `_admin` roles are good constants, as they are builtin roles for CouchDb. The others are configurable.

tests/e2e/default/db/initial-replication.wdio-spec.js
dianabarsan: yes!

webapp/src/ts/services/db-sync.service.ts
dianabarsan: yes, these are "admin-only" ids. You can also find them in the `validate_doc_update` functions (though we cannot use constants there :( )

dianabarsan:
@cliftonmcintosh we have fixed the CI errors in the `master` branch. Please rebase or merge to get the CI running here again! Thanks!

cliftonmcintosh:
@dianabarsan 

Right now, this PR is large enough that I would like feedback on how many more changes you think we can safely make in one PR.

How confident can we be in the test suite when making these changes?

Part of the reason I ask is because I've been doing an analysis for other magic strings, and, as [you mentioned in the issue](https://github.com/medic/cht-core/issues/9621#issuecomment-2464839161), there are a lot! I'm happy to do them, but I don't want to have a change that is large enough that it is hard to review and risky to deploy.

Do we want to decide on a cutoff point so that we don't have a giant change with this PR?

# Samples of other magic strings that we might add to constants

Here's an analysis of strings that may be candidates for constants for document ids and document types. I don't know if all of these are categorized correctly or not, and this doesn't include some of the other things, like the `org.couchdb.user` prefix or the `messages-` prefix for translations.

## doc ids

- **settings** - 101 occurrences
- **resources** - 62 occurrences
- **privacy-policies** - 60 occurrences
- **branding** - 45 occurrences
- **partners** - 41 occurrences
- **target-aggregates** - 35 occurrences
- **migration-log** - 25 occurrences
- **extension-libs** - 16 occurrences
- **zscore-charts** - 13 occurrences

## doc types

- **data_record** - 1,265 occurrences in 127 files
- **person** - 670 occurrences in 120 files
- **clinic** - 482 occurrences in 74 files
- **contact** - 305 occurrences in 63 files
- **district_hospital** - 273 occurrences in 73 files
- **health_center** - 234 occurrences in 66 files
- **task** - 103 occurrences in 24 files
- **target** - 88 occurrences in 17 files
- **user-settings** - 75 occurrences in 16 files
- **form** - 34 occurrences in 18 files
- **token_login** - 24 occurrences in 6 files

shared-libs/constants/src/index.js
```diff
@@ -0,0 +1,36 @@
...
+  ONLINE: 'mm-online',
+};
+
+// Document IDs
+const DOC_IDS = {
+  SERVICE_WORKER_META: 'service-worker-meta',
+};
+
+// Metadata Keys
+const METADATA_KEYS = {
```
dianabarsan: The name `METADATA_KEYS` is confusing, because they are actually doc ids. 
A better name is `SENTINEL_METADATA`

shared-libs/constants/package.json
dianabarsan: I don't unit unit tests are required now, since there is no logic and we are only exporting strings. Please feel free to: either cleanup this file (remove scripts and directories) or add an empty test suite as a blueprint in case we _do_ add any kind of logic here. Up to you!

shared-libs/user-management/test/unit/roles.spec.js
```diff
@@ -61,9 +62,9 @@ describe('roles', () => {
       chai.expect(roles.isOnlineOnly({ roles: ['national_admin', 'chw'] })).to.equal(false);
     });
 
-    it('should check for "mm-online" role', () => {
-      chai.expect(roles.isOnlineOnly({ roles: ['mm-online'] })).to.equal(true);
-      chai.expect(roles.isOnlineOnly({ roles: ['mm-online', 'offline'] })).to.equal(true);
+    it('should check for "USER_ROLES.ONLINE" role', () => {
```
dianabarsan: I think the test names should not be changed.

tests/utils/sentinel.js
```diff
@@ -2,10 +2,11 @@ const utils = require('@utils');
 const querystring = require('querystring');
 const constants = require('@constants');
 const _ = require('lodash');
+const { METADATA_KEYS } = require('@medic/constants');
 
 const SKIPPED_BY_SENTINEL = /^_design\/|(-info|____tombstone)$/;
-const TRANSITION_SEQ = '_local/transitions-seq';
-const BACKGROUND_SEQ = '_local/background-seq';
+const TRANSITION_SEQ = METADATA_KEYS.TRANSITIONS_SEQ;
+const BACKGROUND_SEQ = METADATA_KEYS.BACKGROUND_SEQ;
```
dianabarsan: These two constants are now redundant here.

api/src/migrations/extract-transition-seq.js
```diff
@@ -4,9 +4,10 @@
 //
 const db = require('../db');
 const logger = require('@medic/logger');
+const { METADATA_KEYS } = require('@medic/constants');
 
-const TRANSITION_SEQ_DOCUMENT = '_local/transitions-seq';
-const BACKGROUND_CLEANUP_SEQ_DOCUMENT = '_local/background-seq';
+const TRANSITION_SEQ_DOCUMENT = METADATA_KEYS.TRANSITIONS_SEQ;
+const BACKGROUND_CLEANUP_SEQ_DOCUMENT = METADATA_KEYS.BACKGROUND_SEQ;
```
dianabarsan: These two constants are now redundant.

api/src/server-utils.js
dianabarsan: I don't think it matters so much.

api/tests/integration/migrations/extract-transition-seq.spec.js
```diff
@@ -2,9 +2,10 @@ const { assert } = require('chai');
 
 const utils = require('./utils');
 const db = require('../../../src/db');
+const { METADATA_KEYS } = require('@medic/constants');
 
-const TRANSITION_SEQ_DOCUMENT = '_local/transitions-seq';
-const BACKGROUND_CLEANUP_SEQ_DOCUMENT = '_local/background-seq';
+const TRANSITION_SEQ_DOCUMENT = METADATA_KEYS.TRANSITIONS_SEQ;
+const BACKGROUND_CLEANUP_SEQ_DOCUMENT = METADATA_KEYS.BACKGROUND_SEQ;
```
dianabarsan: These two constants are now redundant.

sentinel/src/lib/metadata.js
```diff
@@ -1,8 +1,9 @@
 // Manages accessing and writing of single value metadata documents
 const db = require('../db');
+const { METADATA_KEYS } = require('@medic/constants');
 
-const TRANSITION_SEQ_DOCUMENT = '_local/transitions-seq';
-const BACKGROUND_CLEANUP_SEQ_DOCUMENT = '_local/background-seq';
+const TRANSITION_SEQ_DOCUMENT = METADATA_KEYS.TRANSITIONS_SEQ;
+const BACKGROUND_CLEANUP_SEQ_DOCUMENT = METADATA_KEYS.BACKGROUND_SEQ;
```
dianabarsan: These constants are now redundant

dianabarsan (commented):
Very nice! I think you're correct about the PR size, I appreciate you holding back from adding more changes. 
I left some really minor comments below.

shared-libs/constants/src/index.js
cliftonmcintosh: Thanks for clarifying. Updated to `SENTINEL_METADATA`.

shared-libs/constants/package.json
cliftonmcintosh: Thanks. I've removed the "directories" and "scripts" keys from the JSON in this file

shared-libs/user-management/test/unit/roles.spec.js
cliftonmcintosh: Thanks. I've switched back the test names so that they use the actual strings. Please let me know if you see any I missed.

tests/utils/sentinel.js
cliftonmcintosh: I've updated the imports so that they are like this:

```js
const {
  SENTINEL_METADATA: {
    TRANSITIONS_SEQ,
    BACKGROUND_SEQ
  }
} = require('@medic/constants');
```

Please let me know if you think that is not the best style. I didn't see guidance in the [style guide](https://docs.communityhealthtoolkit.org/community/contributing/code/style-guide/) for this sort of thing.

api/src/migrations/extract-transition-seq.js
cliftonmcintosh: Updated

api/tests/integration/migrations/extract-transition-seq.spec.js
cliftonmcintosh: Updated

sentinel/src/lib/metadata.js
cliftonmcintosh: Updated

cliftonmcintosh:
> Very nice! I think you're correct about the PR size, I appreciate you holding back from adding more changes. I left some really minor comments below.

Thanks. Based on that, I will not add more constants in this PR. What is the best way to track adding more constants?

1. An issue with some sub-issues that breaks it down to manageable parts?
2. Several issues, each of which does a manageable part?
3. Something else?

I wouldn't want to spend the rest of my life updating constants, but I'm happy to plug away at the others bit by bit. I'm also happy to let someone else take them on or share more changes with me if you have someone who is interested.

mrjones-plip:
> I wouldn't want to spend the rest of my life updating constants, but I'm happy to plug away at the others bit by bit. I'm also happy to let someone else take them on or share more changes with me if you have someone who is interested.

Based on being a new comer - @cliftonmcintosh - do you think we could fashion a [good first issue](https://docs.communityhealthtoolkit.org/community/contributing/creating-good-first-issues/) out of another ticket?  I kind of like the idea of an iterative approach to this.  Each iteration can be inspired and informed by the last effort since they're very derivative of each other.  

However, since I'm not actually that close to the code, getting your (and @dianabarsan's!) take on this idea would be helpful.  Thanks!

cliftonmcintosh:
Compilation is failing because the constants can't be found in the admin app.

```sh
22 11 2025 00:36:35.039:INFO [Chrome Headless 142.0.0.0 (Linux x86_64)]: Connected on socket A4OmvVOkulvNwL7hAAAB with id 45696607
Chrome Headless 142.0.0.0 (Linux x86_64) ERROR
  Uncaught Error: Cannot find module '@medic/constants'
  at /home/runner/work/cht-core/cht-core/api/build/static/admin/js/main.js:1:206
  
  Error: Cannot find module '@medic/constants'
      at o (/home/runner/work/cht-core/cht-core/api/build/static/admin/js/main.js:1:166)
      at o (/home/runner/work/cht-core/cht-core/api/build/static/admin/js/main.js:1:132)
      at tests/unit/controllers/display-languages.spec.js:1:23
```

Compiling JS apps isn't my strongest skill.

Do I need to add this to the "dependencies" in `admin/package.json`? Is there a different thing I need to be doing?

```js
"@medic/constants": "file:../shared-libs/constants"
```

cliftonmcintosh:
> do you think we could fashion a [good first issue](https://docs.communityhealthtoolkit.org/community/contributing/creating-good-first-issues/) out of another ticket? I kind of like the idea of an iterative approach to this. Each iteration can be inspired and informed by the last effort since they're very derivative of each other.

Yes, as a first timer, I think having the continuation as first issues would work. One possibility would be to define categories of magic strings, like "document ids" or "document types" or "HTTP headers", and create a ticket for a category.

cliftonmcintosh:
> Do I need to add this to the "dependencies" in `admin/package.json`?

That doesn't seem to follow convention here. 
Do we need to update CI so that it runs `npm install` after adding the new constants library?
Do we need to update a bundler or build configuration?

admin/tests/unit/controllers/display-languages.spec.js
```diff
@@ -58,15 +60,15 @@ describe('Display Languages controller', function() {
           doc: {
             _id: 'messages-en',
             code: 'en',
-            type: 'translations',
+            type: DOC_TYPES.TRANSLATIONS,
```
dianabarsan, line 62: I am a bit puzzled as to why the admin tests are getting updates, but the actual controllers themselves are not.

cliftonmcintosh:
# Admin module compilation error

> What I ran locally was I ran `cd admin && npm install ../shared-libs/constants`.

Thanks. I've tried this and then run the unit tests, but I still end up with an error that says

```
25 11 2025 13:30:36.194:WARN [filelist]: All files matched by "/Users/cliftonmcintosh/Documents/workspace/cht-core/node_modules/chai/chai.js" were excluded or matched by prior matchers.
25 11 2025 13:30:36.243:INFO [karma-server]: Karma v6.4.4 server started at http://localhost:9876/
25 11 2025 13:30:36.243:INFO [launcher]: Launching browsers Chrome_Headless with concurrency unlimited
25 11 2025 13:30:36.246:INFO [launcher]: Starting browser Chrome
25 11 2025 13:30:37.199:INFO [Chrome Headless 142.0.0.0 (Mac OS 10.15.7)]: Connected on socket 1u3gpGjOPGU4p_G4AAAB with id 52802022
Chrome Headless 142.0.0.0 (Mac OS 10.15.7) ERROR
  Uncaught Error: Cannot find module '@medic/constants'
  at /Users/cliftonmcintosh/Documents/workspace/cht-core/api/build/static/admin/js/main.js:1:206

  Error: Cannot find module '@medic/constants'
      at o (/Users/cliftonmcintosh/Documents/workspace/cht-core/api/build/static/admin/js/main.js:1:166)
      at o (/Users/cliftonmcintosh/Documents/workspace/cht-core/api/build/static/admin/js/main.js:1:132)
      at tests/unit/controllers/display-languages.spec.js:1:23
```

I did not commit the changes from running `npm install ../shared-libs/constants` because the unit tests still fail with that error.

<details>

<summary>Here is the diff after running `npm install ../shared-libs/constants`</summary>

```sh
diff --git a/admin/package-lock.json b/admin/package-lock.json
index 92f3dec68..6175b8e63 100644
--- a/admin/package-lock.json
+++ b/admin/package-lock.json
@@ -8,6 +8,7 @@
       "name": "medic-admin",
       "version": "0.1.0",
       "dependencies": {
+        "@medic/constants": "file:../shared-libs/constants",
         "@uirouter/angularjs": "^1.1.1",
         "angular": "^1.8.3",
         "angular-cookie": "^4.1.0",
@@ -45,6 +46,11 @@
       "extraneous": true,
       "license": "Apache-2.0"
     },
+    "../shared-libs/constants": {
+      "name": "@medic/constants",
+      "version": "1.0.0",
+      "license": "Apache-2.0"
+    },
     "../shared-libs/contact-types-utils": {
       "name": "@medic/contact-types-utils",
       "version": "1.0.0",
@@ -174,6 +180,10 @@
         "node": ">=6.9.0"
       }
     },
+    "node_modules/@medic/constants": {
+      "resolved": "../shared-libs/constants",
+      "link": true
+    },
     "node_modules/@types/linkify-it": {
       "version": "5.0.0",
       "resolved": "https://registry.npmjs.org/@types/linkify-it/-/linkify-it-5.0.0.tgz",
@@ -1059,6 +1069,9 @@
         "to-fast-properties": "^2.0.0"
       }
     },
+    "@medic/constants": {
+      "version": "file:../shared-libs/constants"
+    },
     "@types/linkify-it": {
       "version": "5.0.0",
       "resolved": "https://registry.npmjs.org/@types/linkify-it/-/linkify-it-5.0.0.tgz",
diff --git a/admin/package.json b/admin/package.json
index fee672748..c6b676de0 100644
--- a/admin/package.json
+++ b/admin/package.json
@@ -3,6 +3,7 @@
   "private": true,
   "version": "0.1.0",
   "dependencies": {
+    "@medic/constants": "file:../shared-libs/constants",
     "@uirouter/angularjs": "^1.1.1",
     "angular": "^1.8.3",
     "angular-cookie": "^4.1.0",
```

</details>

# String map error

> to fix your original error of `String map expecing string key`, it appears that ng-annotate has trouble with the `const { DOC_TYPES } = require(...)` Changing the require to:
> 
> ```
> const constants = require('@medic/constants');
> const DOC_TYPES = constants.DOC_TYPES;
> ```
> 
> fixes it.

I've adopted this approach for that String map error.

dianabarsan:
> Thanks. I've tried this and then run the unit tests, but I still end up with an error that says

The unit tests don't actually "build" the application. can you try running "npm run build-dev" before running the tests?

cliftonmcintosh:
> can you try running "npm run build-dev" before running the tests?

Thanks. That by itself didn't work. What does work for me is also adding `@medic/constants` to the browserify script so that it is bundled in and available when test files require it.

admin/tests/unit/controllers/display-languages.spec.js
```diff
@@ -1,4 +1,5 @@
 describe('Display Languages controller', function() {
+  const { DOC_TYPES } = require('@medic/constants');
```
cliftonmcintosh, line 2: I moved this `require` statement inside of the describe block because when it was outside the describe block, I got a conflict between this declaration and the one in `admin/tests/unit/controllers/display-translations.spec.js`.

The clash looked like this:

```
  Uncaught SyntaxError: Identifier 'DOC_TYPES' has already been declared
  at tests/unit/controllers/display-translations.spec.js:1:1

  SyntaxError: Identifier 'DOC_TYPES' has already been declared
```

dianabarsan: Why not just use the two statements instead of the destructuring?

cliftonmcintosh: I was trying to be consistent across usages of the `require` statement. 

When you say two statements, are you thinking everywhere we import the constants or just in the situation with these two tests? Is there a style preference for two statements in general? Is there an approach you are thinking of that would solve the problem of a clash between the two test files?

dianabarsan: IMO it doesn't matter. 
You can just `const CONSTANTS = require('@medic/constants')` and be access everything from there `CONSTANTS.DOC_TYPES.TRANSLATIONS`.

cliftonmcintosh: Thanks for the explanation. 

I have a slight preference for the destructuring in the import statement because I feel it makes the access of the constants in the file just a little more concise.

However, if you feel that something like `CONSTANTS.DOC_TYPES.TRANSLATIONS` is easier to read or reason about than `DOC_TYPES.TRANSLATIONS`, I'm happy to change it.

admin/tests/unit/controllers/display-languages.spec.js
cliftonmcintosh: Oops. I missed the use of `'translations'` in the controller files. Thanks for noticing. I've updated them.

I also could not use destructuring for the imports for those files, so they are like this:

```js
const constants = require('@medic/constants');
const DOC_TYPES = constants.DOC_TYPES;
```

Please let me know if you think it would be better to adopt that same approach across the other files to be consistent.

Also please let me know if you think it's better to use `constants.DOC_TYPES.TRANSLATIONS` where the constant is being used instead of doing `const DOC_TYPES = constants.DOC_TYPES;` at the top of the file.

cliftonmcintosh: I wonder if after this we don't need to add the line to require the constants in the browserify script. I'll try removing that and see what happens.

cliftonmcintosh: > I wonder if after this we don't need to add the line to require the constants in the browserify script. I'll try removing that and see what happens.

Based on my testing locally, it seems like we still need the change in e3cd35b -- the one that updates the script at `scripts/build/browserify-admin.sh` to require the constants library. I was hoping that adding the imports to the controllers would make that unnecessary, but the build fails for me duing `npm run unit-admin` if the constants are not required in the browserify script.

cliftonmcintosh:
@dianabarsan 

I think I've addressed your suggestions and comments. I've also resolved the conflicts with the main branch. Please let me know if you have advice for any other changes we should make in this.

cliftonmcintosh:
@dianabarsan

Hi. I'm following up to see if there is more feedback.

I think I've addressed all of the earlier feedback. If I haven't or if they changes need adjusting, please let me know.

dianabarsan:
so sorry for missing this. I'm re requesting a review so i get notified about this.

cliftonmcintosh:
@dianabarsan 

Thanks for the feedack and the review. I assume you or another maintainer will take responsibility for a merge. Pleaes let me know if there is anything I should do prior to that.

dianabarsan:
There's nothing extra to do. There was a failing (flaking) e2e test that I had to restart to get the build green (otherwise merging is blocked). Otherwise I would have merged immediately after the approval :)

Issue #10473, opened 2025-11-19
Disable default send to sender in sms workflows


1yuv:
**What feature do you want to improve?**
From #10459, 
Disable Default Redirection: A configuration option to prevent the system from defaulting to the report sender (submitter) when no recipient is found. More detail is on the parent issue. 

**Describe the improvement you'd like**
Add `default_to_sender` flag that controls whether to send messages to sender in case sms resolution can't resolve the sender. 

**Describe alternatives you've considered**
As discussed in [forum](https://forum.communityhealthtoolkit.org/t/is-there-a-way-to-disable-sms-recipient-redirection/3921).

PR #10477, opened 2025-11-21
feat(#10473): allow user to configure whether to send message to sender or not

1yuv:
# Description

Add `default_to_sender` parameter in app settings so that users can configure whether CHT should send message to default sender if it can't resolve any other sender. 

#10473 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10473_disable_default_to_sender/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10473_disable_default_to_sender/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10473_disable_default_to_sender/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/tests/integration/shared-libs/cht-datasource/contact.spec.js
+++ b/tests/integration/shared-libs/cht-datasource/contact.spec.js
@@ -1,10 +1,10 @@
 const utils = require('@utils');
+const sentinelUtils = require('@utils/sentinel');
 const personFactory = require('@factories/cht/contacts/person');
 const placeFactory = require('@factories/cht/contacts/place');
 const userFactory = require('@factories/cht/users/users');
-const {getRemoteDataContext, Qualifier, Contact} = require('@medic/cht-datasource');
-const {expect} = require('chai');
-const {setAuth, removeAuth} = require('./auth');
+const { getRemoteDataContext, Qualifier, Contact } = require('@medic/cht-datasource');
+const { setAuth, removeAuth } = require('./auth');
 
 describe('cht-datasource Contact', () => {
   // NOTE: this is a common word added to contacts to fetch them
@@ -110,9 +110,12 @@ describe('cht-datasource Contact', () => {
   const expectedPlaces = [ place0, clinic1, clinic2 ];
   const expectedPlacesIds = expectedPlaces.map(place => place._id);
 
+  const excludedProperties = [ '_rev', 'reported_date', 'patient_id', 'place_id' ];
+
   before(async () => {
     setAuth();
     await utils.saveDocs(allDocItems);
+    await sentinelUtils.waitForSentinel();
     await utils.createUsers([ userNoPerms, offlineUser ]);
   });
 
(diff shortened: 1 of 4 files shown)


shared-libs/message-utils/src/index.js
```diff
@@ -96,77 +96,128 @@ const applyPhoneFilters = function(config, phone) {
   return phone;
 };
 
-const getRecipient = function(context, recipient) {
+const getRecipient = function(context, recipient, default_to_sender = true) {
```
dianabarsan: Please use camelCase for variables. `defaultToSender`

shared-libs/message-utils/src/index.js
```diff
@@ -96,77 +96,128 @@ const applyPhoneFilters = function(config, phone) {
...
+    node = node.parent;
   }
-  return phone || from || recipient;
+
+  return node.contact?.phone || null;
 };
 
 const getPhone = function(config, context, recipient) {
-  let phone = getRecipient(context, recipient);
+  let phone = getRecipient(context, recipient, config?.sms?.default_to_sender ?? true);
```
dianabarsan, line 220: I think I would rather pass the whole config to the `getRecipient` function and get the value of whether to default to sender within that function. it's a cleaner separation of concerns.

shared-libs/message-utils/src/index.js
```diff
@@ -96,77 +96,128 @@ const applyPhoneFilters = function(config, phone) {
...
+const resolveAncestor = function(context, levels) {
+  let node = context.patient || context.place || context;
+  node = node.parent ? node : node.contact;
+
+  for (let i = 0; i < levels; i++) {
+    if (!node?.parent) {
+      return null;
+    }
+    node = node.parent;
   }
```
dianabarsan: Using a `while` loop here would remove the need of having an additional iterator variable: 
```suggestion
  while (levels-- > 0) {
    node = node?.parent || null;    
  }
```

shared-libs/message-utils/src/index.js
```diff
@@ -96,77 +96,128 @@ const applyPhoneFilters = function(config, phone) {
...
+
+  return phone || (default_to_sender && from) || recipient;
+};
+
+const resolveRecipient = function(context, recipient) {
+  if (!recipient) {
+    return null;
+  }
+
+  const resolvers = [
```
dianabarsan, line 121: I'm loving this resolver pattern!

tests/integration/transitions/sms-workflows.spec.js
```diff
@@ -1156,5 +1156,112 @@ describe('SMS workflows', () => {
...
+      const [ contactChw1] = await processReportsAndSettings(reports, settings);
+      expectTasks(contactChw1, [
+        { messages: [{ to: 'phone1', message: 'to patient' }] }
+      ]);
+    });
+
+    it('should not send to sender if field is not found and default_to_sender is set to false', async () => {
+      const settings = {
+        sms: {
+          default_to_sender: true
```
dianabarsan: I think `default_to_sender` needs to be `false` here right? 
This test body looks to be identical to the one before "should send to sender if field is not found and default_to_sender is set to true".

dianabarsan (changes requested):
Nice start! I left a couple of comments below.

shared-libs/message-utils/src/index.js
1yuv: I think it’s cleaner for the caller to resolve the `defaultToSender` boolean before invoking getRecipient.  It’s better for SoC: the caller decides the boolean, and getRecipient focuses on resolving a phone number.

If we start passing the whole config object, the function becomes tied to that structure. Keeping it decoupled feels like the safer design.

dianabarsan: ok. I don't have a really strong opinion about this.

PR #10480, opened 2025-11-25
feat(#3943): adds overdue task bubble counter

### LINKED ISSUE #3943: Show a count of overdue tasks in the nav bar
Just like we have unread counters for reports and messages, add an overdue tasks counter in a bubble in the nav bar.

unread-reports

dianabarsan:
# Description

closes #3943

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:3943-task-overdue-bubble/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:3943-task-overdue-bubble/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:3943-task-overdue-bubble/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/tests/e2e/default/tasks/overdue-bubble.wdio-spec.js
+++ b/tests/e2e/default/tasks/overdue-bubble.wdio-spec.js
@@ -32,6 +32,8 @@ describe('Task overdue bubble counter', () => {
     reported_date: new Date().getTime(),
   }));
 
+  const CONFIGURED_OVERDUE_TASKS = 3;
+
   const getTaskBubble = async () => {
     const tasksTab = await commonPage.tabsSelector.taskTab();
     return await tasksTab.$('.mm-badge');
@@ -115,14 +117,14 @@ describe('Task overdue bubble counter', () => {
 
     // Should show 3 overdue tasks
     const bubbleCount = await getTasksBubbleCount();
-    expect(bubbleCount).to.equal(patients.length * 3);
+    expect(bubbleCount).to.equal(patients.length * CONFIGURED_OVERDUE_TASKS);
   });
 
   it('should update bubble count when task is completed', async () => {
     await waitForBubble();
 
     let bubbleCount = await getTasksBubbleCount();
-    expect(bubbleCount).to.equal(patients.length * 3);
+    expect(bubbleCount).to.equal(patients.length * CONFIGURED_OVERDUE_TASKS);
 
     await commonPage.goToTasks();
     await submitForm(patients[0].name, 'task_overdue_1');
--- a/webapp/src/ts/reducers/tasks.ts
+++ b/webapp/src/ts/reducers/tasks.ts
@@ -3,10 +3,12 @@ import { createReducer, on } from '@ngrx/store';
 import { Actions as GlobalActions } from '@mm-actions/global';
 import { Actions } from '@mm-actions/tasks';
 import moment from 'moment';
+import { TaskEmission } from '@mm-services/rules-engine.service';
+
 
 const initialState = {
-  tasksList: [] as any[],
-  overdue: [] as any[],
+  tasksList: [] as TaskEmission[],
+  overdue: [] as TaskEmission[],
   selected: null,
   loaded: false,
   taskGroup: {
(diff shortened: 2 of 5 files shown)


tests/wdio.conf.js
```diff
@@ -188,6 +188,7 @@ const baseConfig = {
   //
   // Options to be passed to Mocha.
   // See the full list at http://mochajs.org/
+  // specFileRetries: DEBUG ? 0 : 5,
```
dianabarsan, line 191: Leaving this here as a reminder!

dianabarsan:
Hi @sugat009 . Would you mind a review for this feature? I know it's a sizable PR, although most is test coverage. Please let me know what you think! Thanks :)

webapp/src/ts/services/rules-engine.service.ts
```diff
@@ -295,6 +308,21 @@ export class RulesEngineService implements OnDestroy {
...
+    const taskDocSubscription = this.changesService.subscribe({
+      key: 'task-doc-update',
+      callback: change => {
+        const { doc } = change;
+        if (doc.type === 'task') {
+          this.ngZone.run(() => {
+            this.taskActions.setOverdueTasks([doc]);
+          });
+        }
+      }
```
sugat009, line 319:
```suggestion
    filter: change => change.doc?.type === 'task',
    callback: change => {
      this.ngZone.run(() => {
        this.taskActions.setOverdueTasks([change.doc]);
      });
    }
```
suggestion: No filter is provided to `changesService.subscribe()`, meaning this
  callback fires for every document change in the database, not just task
  documents. There is a `doc.type` check in line 316 which does the same thing but it would make sense to just use the filter because it exists to have the `this.ngZone.run` statement alone in `callback` for readability.

webapp/src/ts/reducers/tasks.ts
```diff
@@ -6,6 +6,7 @@ import moment from 'moment';
 
 const initialState = {
   tasksList: [] as any[],
+  overdue: [] as any[],
```
sugat009: question: don't we have a `Task` interface? it would be safer to use those here.

webapp/src/ts/services/rules-engine.service.ts
```diff
@@ -52,7 +54,7 @@ export class RulesEngineCoreFactoryService {
 export class RulesEngineService implements OnDestroy {
   private rulesEngineCore;
   private readonly MAX_LINEAGE_DEPTH = 50;
-  private readonly ENSURE_FRESHNESS_MILLIS = 120 * 1000;
+  private readonly ENSURE_FRESHNESS_MILLIS = 1000;
```
sugat009, line 57: question: how is the performance of this reduction of debounce time on devices with many contacts?

Document this change and consider if 1 second is  appropriate for all deployment scenarios. Consider making it configurable through settings or at least add a comment explaining the rationale.

webapp/src/ts/services/rules-engine.service.ts
```diff
@@ -154,6 +159,14 @@ export class RulesEngineService implements OnDestroy {
     return true;
   }
 
+  private async fetchOverdueTasksForAllContacts() {
+    const allTasks = await this.fetchTaskDocsForAllContacts();
+    this.ngZone.run(() => {
+      this.taskActions.setTasksList(allTasks.map(task => task.emission));
+    });
+    this.monitorTaskChanges();
```
sugat009, line 167: suggestion: it would make sense to return an empty array or something nullish that would provide a 0 overdue count in case the calls to these many functions fail

tests/e2e/default/tasks/overdue-bubble.wdio-spec.js
```diff
@@ -0,0 +1,174 @@
...
+  it('should display bubble with count of overdue tasks', async () => {
+    await compileBubbleTasks('overdue-bubble-config.js');
+    await waitForBubble();
+
+    // Check that bubble is displayed on tasks tab
+    expect(await isBubbleDisplayed()).to.be.true;
+
+    // Should show 3 overdue tasks
+    const bubbleCount = await getTasksBubbleCount();
+    expect(bubbleCount).to.equal(patients.length * 3);
```
sugat009: issue: it would make sense to have a variable with a meaningful name for `3` which appears quite a lot on this test suite

webapp/src/ts/services/rules-engine.service.ts
dianabarsan: I don't think making it configurable will resolve anything. Deployments seldom use or even know about these configs, so I'm reluctant to add them. 

The performance impact is minimal, because the action itself would have been executed anyway, albeit with a higher delay. I'm not worried about this and I don't think it needs documenting. 
This action would have happened anyway if the user navigated to the tasks or targets page.

webapp/src/ts/services/rules-engine.service.ts
dianabarsan: I don't think the state should be updated if any of the calls fails. We should keep using the old state in these cases.

sugat009 (approved):

PR #10482, opened 2025-11-26
fix(#10481): set nouveau as a sidecar container

### LINKED ISSUE #10481: Cannot upgrade to 5.x in AWS due to Persistent Volume settings
**Describe the bug**
During attempting the upgrade of the demo-cht instance to 5.x, the upgrade hit a roadblock caused by the helm chart configuration, where both CouchDb and Nouveau would need to attach the same Persistent Volume. 
In AWS EBS terms this is called MultiAttach, and most volume types cannot do this (https://docs.aws.amazon.com/ebs/latest/userguide/ebs-volumes-multi.html#considerations)

**To Reproduce**
1. Launch a 4.22 instance in AWS EC2 with a GP2 volume
2. Stage the 5.0.0 upgrade
3. Follow the upgrade steps here: https://docs.communityhealthtoolkit.org/hosting/cht/migration/helm-charts-4x-to-5x-migration/
4. After a while, the instance will either crash or certain features that depend on Nouveau will not work (replication, online user search). 

**Expected behavior**
The upgrade should work. 

**Logs**
Checking pods, the 5.0.0 database containers fail to start, while the old couchdb container is still running (the old version of CouchDb): 
```
cht-api-69477c6f59                   1/1     Running             1 (3h16m ago)   3h22m
cht-couchdb-6ff5bbfc6d               1/1     Running             0               6h52m
cht-couchdb-858666b666               0/1     ContainerCreating   0               3h22m
cht-couchdb-nouveau-69f6d7bd9f       0/1     ContainerCreating   0               3h22m
cht-haproxy-78c94b569                1/1     Running             0               3h22m
cht-haproxy-healthcheck-54d548956c   1/1     Running             0               3h22m
cht-sentinel-f7bd94584               1/1     Running             0               3h22m
```

**Environment**
- Client platform: Kubernetes
- App: helm charts
- Version: 5.0.0

**Additional context**
We believe there are three options to fix this: 

A: The Sidecar Pattern 
Action: Move Nouveau to the same pod as CouchDb so they share storage
Why: Containers in the same Pod share storage and run on the same node automatically.

Option B: Separate Storage 
Action: Give nouveau its own, separate PVC.

Option C: Shared File System (Rare for databases) Use AWS EFS (ReadWriteMany) instead of EBS. 
Note: This is generally not recommended for database data files due to latency and IOPS limitations.

Discussing, we believe that option A is most suitable, because it doesn't require deployments to provision additional disk space for Nouveau, separate from CouchDb.

dianabarsan:
# Description

Also updates strategy overall to recreate. 

closes #10481

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10481-same-storage-helm-charts/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10481-same-storage-helm-charts/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10481-same-storage-helm-charts/docker-compose/cht-couchdb-clustered.yml)

# License


dianabarsan:
Hi @witash 

Can you please review? I've already verified this with upgrading https://demo-cht.dev.medicmobile.org

witash (approved):
LGTM

There's some more context about the original version of the nouveau helm charts here:
https://github.com/medic/helm-charts/pull/42#issue-3038644386
one thing to note is that having the nouveau container be on the first couchdb-1 pod means it can be unbalanced; but that's ok, it probably wouldn't be significant and only applies to multi-node deployments.

dianabarsan:
Thanks for the review @witash . 
I think everything would have worked great with the previous design. I agree with all the points. 
Knowing of this limitation of AWS narrows our choices.

Issue #10459, opened 2025-11-13
Feature Request: Configurable SMS Recipient Fallbacks & Disable "Sender" Default


binokaryg:
**What feature do you want to improve?**

[SMS recipient resolution](https://docs.communityhealthtoolkit.org/building/reference/app-settings/#sms-recipient-resolution): Add alternate recipient(s) and way to disable the default behaviour of sending it back to the sender

**Describe the improvement you'd like**

I am requesting two specific enhancements to the recipient resolution logic:

- Priority/Fallback Lists: The ability to define a list of potential recipients. The system should attempt to resolve them in order and use the first valid phone number found.

- Disable Default Redirection: A configuration option to prevent the system from defaulting to the report sender (submitter) when no recipient is found.

CHT provides several ways to specify the [SMS recipient](https://docs.communityhealthtoolkit.org/building/reference/app-settings/#recipient-values-and-resolutions).

Example Use Case: I want to send a reminder related to a patient in a household.

I can either send it to:
- the patient using `patient.phone`
- or, to the household primary contact with `patient.parent.contact.phone`

I want this instead:

- If the patient has a phone number -> Send to `patient`.
- If patient has no number ->  Send to `patient.parent.contact` (Household Head).
- If neither has a number ->  Do not send the message. (Currently, CHT forces this message to the CHW/sender, which is confusing for them).

Proposed configuration syntax:
  Instead of a single string, allow an array for recipient:
```json
"recipient": [
  "patient.phone",
  "patient.parent.contact.phone"
],
"default_to_sender": false
```

**Describe alternatives you've considered**

- Form Calculation: Saving the phone number in the report during submission. Drawback: If the contact's phone number is updated later, the report data remains stale, and the SMS goes to the old number.

- Mustache Logic: I attempted to use {{#patient.phone}} blocks in the message template to suppress content if the number is missing. Drawback: This still triggers the message sending process, resulting in empty, failed messages cluttering the messages tab.

```
messages.registration.a.patient = {{#patient.phone}} Hi {{patient_name}}! You have been registered to our system with ID: ({{patient_id}}). {{/patient.phone}}{{^patient.phone}}{{/patient.phone}}
```

**Additional context**
Related CHT Forum discussion: [Is there a way to disable SMS recipient redirection?](https://forum.communityhealthtoolkit.org/t/is-there-a-way-to-disable-sms-recipient-redirection/3921)

1yuv:
[...]

PR #10497, opened 2025-11-27
feat(#10459): allow multiple sms recipients

1yuv:
# Description

#10459 

Allow multiple recipients to be specified and resolve them in order adding flexibility for sms deployments. 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:sms_recipient_10459/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:sms_recipient_10459/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:sms_recipient_10459/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/shared-libs/message-utils/src/index.js
+++ b/shared-libs/message-utils/src/index.js
@@ -97,12 +97,11 @@ const applyPhoneFilters = function(config, phone) {
 };
 
 const normalizeRecipient= function(recipient) {
-  const toArray = Array.isArray(recipient) ? recipient : [recipient];  
-const recipientArray = Array.isArray(recipient) ? recipient : [recipient];  
-const isValid = r => typeof r === 'string' || typeof r === 'number';
-return recipientArray
-  .map(r => isValid(r) && String(r).trim())
-  .filter(Boolean);
+  const recipientArray = Array.isArray(recipient) ? recipient : [recipient];  
+  const isValid = r => typeof r === 'string' || typeof r === 'number';
+  return recipientArray
+    .map(r => isValid(r) && String(r).trim())
+    .filter(Boolean);
 };
 
 const getRecipient = function(context, recipient, defaultToSender = true) {
--- a/tests/integration/sentinel/schedules/schedules-recipients.spec.js
+++ b/tests/integration/sentinel/schedules/schedules-recipients.spec.js
@@ -55,7 +55,7 @@ describe('Recipient Resolution Integration', () => {
   after(() => utils.revertDb([], true));
   afterEach(() => utils.revertDb(getIds(contacts), true));
 
-  it('correctly resolves recipients', () => {
+  it('correctly resolves recipients', async () => {
     const settings = {
       transitions: { registration: true },
       sms: {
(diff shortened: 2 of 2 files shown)


shared-libs/message-utils/src/index.js
```diff
@@ -96,21 +96,52 @@
   return phone;
 };
 
+const normalizeRecipient= function(recipient) {
```
dianabarsan, line 99: I see that there are at least 3 places where we test for string and check for validity and it's confusing and messy. 

I suggest we convert this function to properly normalize the recipient in the sense that: 

- converts the recipients to strings and also does trimming
- filters empty strings
- returns an array even when there is only one recipient

That saves us from all the additional checks later.

shared-libs/message-utils/src/index.js
```diff
@@ -96,21 +96,52 @@ const applyPhoneFilters = function(config, phone) {
...
 const getRecipient = function(context, recipient, defaultToSender = true) {
   if (!context) {
     return;
   }
 
   const from = context.from || context.contact?.phone;
-  recipient = recipient?.trim();
+  recipient = normalizeRecipient(recipient);
 
   if (!recipient) {
```
dianabarsan: This needs be (!recipient.length), granted that `normalizeRecipient` will return an empty array when the provided recipient cannot be normalized.

shared-libs/message-utils/src/index.js
```diff
@@ -96,21 +96,52 @@ const applyPhoneFilters = function(config, phone) {
...
     return from;
   }
 
-  const phone = resolveRecipient(context, recipient);
+  const phone = resolveMany(context, recipient);
+  const fallBackRecipient = Array.isArray(recipient) ? recipient[0]: recipient;
+  return phone || (defaultToSender && from) || fallBackRecipient;
```
dianabarsan: `recipient` would always be an array: 
```suggestion
  return phone || (defaultToSender && from) || recipient[0];
```

shared-libs/message-utils/src/index.js
```diff
@@ -96,21 +96,52 @@ const applyPhoneFilters = function(config, phone) {
...
+};
+
+const resolveMany = (context, recipients) => {
+  if (typeof recipients === 'string') {
+    return resolveRecipient(context, recipients);
+  }
+
+  if (!Array.isArray(recipients)) {
+    return null;
+  }
```
dianabarsan, line 129: If recipient is normalized, all this section can be removed,

shared-libs/message-utils/src/index.js
```diff
@@ -96,21 +96,52 @@ const applyPhoneFilters = function(config, phone) {
...
 
-  return phone || (defaultToSender && from) || recipient;
+  for (const r of recipients) {
+    if (typeof r === 'string' && r) {
```
dianabarsan: `r` would always be a non empty string, so this check can be removed.

shared-libs/message-utils/src/index.js
1yuv: That's a great idea, will update accordingly.

shared-libs/message-utils/src/index.js
```diff
@@ -96,21 +96,37 @@ const applyPhoneFilters = function(config, phone) {
   return phone;
 };
 
+const normalizeRecipient= function(recipient) {
+  const toArray = Array.isArray(recipient) ? recipient : [recipient];  
+  return toArray.map(
+    r => (r === null ? '': String(r).trim())
```
dianabarsan: I think more than `null` should be covered here. 
If someone uses boolean false or true, then just blindly converting them is maybe not the best idea. 

I think we should not convert null or booleans, and only convert and trim numbers and strings.

shared-libs/message-utils/src/index.js
```diff
@@ -96,21 +96,37 @@ const applyPhoneFilters = function(config, phone) {
...
 
-  return phone || (defaultToSender && from) || recipient;
+const resolveMany = (context, recipients) => {
+  for (const recipient of recipients) {
+    const phone = resolveRecipient(context, recipient);
+    if (phone) {
+      return phone;
+    }    
+  }
+  return null;
```
dianabarsan: Here the `null` return is not required. Omitting it will return `undefined` which will work fine in your code.

dianabarsan (changes requested):

shared-libs/message-utils/src/index.js
1yuv: Our logic already handles the case by converting anything to string and resolver handles them prefectly. 

What do you want to do with null and boolen ? Are you asking that we drop `null` or `boolean` here and not return or keep as `null` and `booleans` as nullish and boolean value? 

I am not sure what scenario would that change improve, but either way would require change in this function and subsequent code if we want to change.

dianabarsan: I just think it's cleaner if we only allow strings and numbers as recipients.

1yuv: I've updated for this function to take forward string and numbers only.

shared-libs/message-utils/src/index.js
1yuv: done.

shared-libs/message-utils/src/index.js
```diff
@@ -96,34 +96,47 @@ const applyPhoneFilters = function(config, phone) {
   return phone;
 };
 
+const normalizeRecipient= function(recipient) {
+  const toArray = Array.isArray(recipient) ? recipient : [recipient];  
+  return toArray.map(
+    r => (typeof r === 'string' || typeof r === 'number' ? String(r).trim(): '')
+  ).filter(r => r.length > 0);
```
dianabarsan: Just suggesting different indentation for readability: 
```suggestion
const recipientArray = Array.isArray(recipient) ? recipient : [recipient];  
const isValid = r => typeof r === 'string' || typeof r === 'number';
return recipientArray
  .map(r => isValid(r) && String(r).trim())
  .filter(Boolean);
```

tests/integration/sentinel/schedules/schedules-recipients.spec.js
```diff
@@ -0,0 +1,154 @@
...
+        parent: {
+          _id: 'clinic',
+          parent: {
+            _id: 'parent',
+          }
+        }
+      }
+    };
+
+    return utils
```
dianabarsan: Can you please convert this to async await?

PR #10502, opened 2025-11-28
feat(10494): block token_login for safari

### LINKED ISSUE #10494: Prevent logging in with token in Safari Browser
via https://github.com/medic/cht-core/issues/6784, we are now hiding login fields for Safari users, however if a token_login link is opened in Safari, the login will still succeed. 

Block token_login links on safari and display a similar message as on the main login page.

shashi-sah2003:
fixes #10494 

# Code review checklist

# License

### DIFF
--- a/api/src/public/login/script.js
+++ b/api/src/public/login/script.js
@@ -231,6 +231,7 @@ const checkUnsupportedBrowser = () => {
 
     if (isSafari) {
       document.getElementById('login-fields')?.classList.add('hidden');
+      document.querySelector('.locale-wrapper .loading')?.classList.add('hidden');
     }
   }
 };
--- a/api/src/controllers/login.js
+++ b/api/src/controllers/login.js
@@ -65,6 +65,8 @@ const templates = {
       'login.token.loading',
       'login.token.redirect.login.info',
       'login.token.redirect.login',
+      'login.unsupported_browser',
+      'login.unsupported_browser.safari',
       'privacy.policy'
     ],
   },


shashi-sah2003:
@dianabarsan pls review this. I think this should work, lmk if there are any changes required

api/src/public/login/script.js
```diff
@@ -68,6 +68,10 @@ const requestSSOLogin = function(e){
 };
 
 const requestTokenLogin = (retry = 20) => {
+  if (isSafariBrowser()) {
```
dianabarsan: I didn't want to add the check for the browser twice. Can this be done by just using the existent `isUsingSupportedBrowser`?

shashi-sah2003: yeah updated the logic to use `!isUsingSupportedBrowser` . It should work fine

api/src/public/login/script.js
```diff
@@ -68,6 +68,10 @@ const requestSSOLogin = function(e){
 };
 
 const requestTokenLogin = (retry = 20) => {
+  if (!isUsingSupportedBrowser()) {
```
dianabarsan: I don't think this is what we want, because this will block any unsupported browser from using token login. Also this does _not_ avoid Safari (since that function is separate). 

Can you please look into moving the call to `checkUnsupportedBrowser` to before the token login check and have the Safari check disable token login somehow?
Even when calling the token login link, a user on Safari should see the Safari "unsupported browser" message.

dianabarsan (changes requested):

api/src/public/login/script.js
shashi-sah2003: okay I have moved the call to `checkUnsupportedBrowser` just before the request for token login. Had to modify the `checkUnsupportedBrowser`.

api/src/public/login/script.js
```diff
@@ -200,17 +200,19 @@ const isUsingChtAndroidV1 = () => {
 
 const checkUnsupportedBrowser = () => {
   if (!selectedLocale) {
```
dianabarsan, line 207: Well, this is a bit confusing. What if there is no selected locale and the browser is Safari. What happens then? I know this existed before, but the initial purpose of this function was to just display a message, which indeed depended on language. However, now we are granting it more power to effectively prevent app use. And yet, we still overrule all checks depending on translations. 
I don't think this is right :( 

So we need to separate them somehow. 
Have a function that will return whether the app is safe to use or not (at this point, it should only return false for Safari). Then, have a separate function that will handle the warning message logic. 

What do you think?

shashi-sah2003: Gotcha the changes you want, make sense to have separation of concerns through different functions, will make necessary changes

shashi-sah2003:
@dianabarsan I have fixed it using a different function

shashi-sah2003:
@dianabarsan done https://github.com/medic/cht-core/pull/10502/commits/3092858dcf185746fe38c4cc913fa96f6270d4c6

PR #10507, opened 2025-11-29
feat(#10439): support loading target subtitle from funciton

### LINKED ISSUE #10439: Add indicator to targets page when previous month is selected
As noted in the [design doc](https://docs.google.com/document/d/1zdJ1ySdSbriZODn7BDDAH9Hxg3qT0J9McGUIIrcafIU/edit?tab=t.ufay0fr37c8h#heading=h.s1wiyw3hzx4x) the target cards for the default config that display all-time metrics (and not ones scoped to a reporting period) have a subtitle of "All Time". This could be misleading when a user filters the targets to the previous month because the value show will actually be "up to the previous month".

The "All Time" targets are ones [configured](https://docs.communityhealthtoolkit.org/building/targets/targets-js/#targetsjs) with the date property as 'now' (or not set).  (So the value is calculated from all reports and not just those filtered for a specific reporting period (month).

Currently, the "All Time" subtitle is being populated [here](https://github.com/medic/cht-core/blob/9337c841e9afd6c25652f534a6f0105f4675bd88/webapp/src/ts/modules/analytics/analytics-targets.component.html#L50), but this is just loading the configured `subtitle_translation_key` value from the target config.  So, with the default config, these target cards show "All Time" as the subtitle, but this translation is totally customizable and the subtitle could be set to anything.  

We need some kind of visual element (either on the page or in the card) that indicates to the user that the values shown are data up to the previous month.

Edwin-Rugoogamu:
# Description

Closes #10439 

# Code review checklist

# License

### DIFF
--- a/webapp/tests/mocha/tsconfig.mocha.json
+++ b/webapp/tests/mocha/tsconfig.mocha.json
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "module": "CommonJS",
     "moduleResolution": "Node",
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "strict": true
   }
 }


webapp/src/ts/modules/analytics/analytics-targets.component.html
```diff
@@ -47,9 +47,13 @@
           <div class="title">
             <h2 *ngIf="target.translation_key">{{ target.translation_key | translate }}</h2>
             <h2 *ngIf="!target.translation_key">{{ target.name | translateFrom }}</h2>
-            <p *ngIf="target.subtitle_translation_key">
-              {{ target.subtitle_translation_key | translate }}
-            </p>
+            <p>
+             {{
```
dianabarsan: @Edwin-Rugoogamu this is not what we discussed, adding two subtitles. 
We discussed making `target.subtitle_translation_key` be a function. 

So this would be something like: 
```
target.subtitle_translation_key(reportingPeriodFilter) | translate
```

Edwin-Rugoogamu: @dianabarsan Thankyou for the feedback. let me adjust it .

webapp/src/ts/modules/analytics/analytics-targets.component.ts
```diff
@@ -109,4 +109,12 @@ export class AnalyticsTargetsComponent implements OnInit, OnDestroy {
         });
       });
   }
+
+  getSubtitleKey(subtitleKey: string, reportingPeriod: ReportingPeriod): string {
+    if (!subtitleKey) {
+      return '';
+    }
+    return reportingPeriod === ReportingPeriod.CURRENT ? 'this_month' : 'all_time';
```
dianabarsan: This still doesn't cover what we need to do here. The situation now is that you're not even using the `subtitleKey` that is configured. 

What needs to change here is that `subtitleKey` _will be a function_! 
So you'll need to do something like: 

```
return typeof subtitleKey === 'function' ? subtitleKey(reportingPeriod) : subtitleKey;
```

Edwin-Rugoogamu: @dianabarsan Thankyou for the feedback.   We had a meeting with Josh and Team .  This elaborated more on what we need to achieve.  So it made it clear for me and erased the confusion i was having.  So am adjusting and will send a PR for review.

dianabarsan:
Leaving this here for reference of where we would deserialize the function: https://github.com/medic/cht-core/blob/master/shared-libs/rules-engine/src/target-state.js#L208C68-L208C92

Edwin-Rugoogamu:
Hi @jkuester Please have a look at the deserialize function .  Am also facing a challenge with the Sonar validation can we make adjustments to it or we need to maitain its validation.

Edwin-Rugoogamu:
Hi @jkuester . i Hope you are doing well.  Please have a look at the codebase as the serialize and descentralise function was taken to the rule engine file.  As discussed before the break off. Looking forward to your reply

webapp/src/ts/modules/analytics/analytics-targets.component.ts
```diff
@@ -102,11 +104,42 @@ export class AnalyticsTargetsComponent implements OnInit, OnDestroy {
       })
       .then((targets: any[] = []) => {
         this.loading = false;
-        this.targets = targets.filter(target => target.visible !== false);
+        // --- REVIVE FUNCTION FIELDS HERE ---
+        this.targets = targets
+          .filter(target => target.visible !== false)
+          .map(target => {
+            if (typeof target.subtitle_translation_key === 'string' &&
+               target.subtitle_translation_key.includes('=>')) {
```
dianabarsan: I think this check for => is pretty fragile, we should at least be checking for the function keyword... I think the most complete way would be to attempt the deserialization and fallback on using the string as is if deserialization fails.

webapp/src/ts/services/rules-engine.service.ts
```diff
@@ -605,6 +605,89 @@ export class RulesEngineService implements OnDestroy {
     return ['rules-engine', ...params].join(':');
   }
 
+  /**
+   * Serializes a function into a string representation.
+   * @param fn - The function to serialize
+   * @returns The serialized function string
+   */
+  serializeFunction(fn: any): string | any {
```
dianabarsan: I don't think the serialization/deserialization code should be in the rules-engine. it's very generic and deserves its own service, maybe even its own shared library.

dianabarsan: Also it does not seem that we ever use `serializeFunction`? This would be useful in cht-conf...

webapp/src/ts/services/rules-engine.service.ts
```diff
@@ -605,6 +605,89 @@ export class RulesEngineService implements OnDestroy {
...
+      }
+      return fnString;
+    }
+  }
+
+  private isValidFunctionString(fnString: string): boolean {
+    if (typeof fnString !== 'string') {
+      return false;
+    }
+    if (!fnString.includes('=>')) {
```
dianabarsan: This implies that we only deserialize arrow functions. which is not correct.

webapp/src/ts/services/rules-engine.service.ts
dianabarsan: I'm wondering if we should be using a serialization library, that is tested and has strong community instead of writing it ourselves.
An example library that popped up in my first search: https://www.npmjs.com/package/serialize-javascript

dianabarsan (changes requested):

webapp/src/ts/services/rules-engine.service.ts
jkuester: I am going to push back on this a bit.  IMHO the JS `Function` constructor is pretty much all the "library" that we need.

In a better world we would have someplace nice to put super generic functions, but as things are it does not seem appropriate to create a whole shared-lib or service for a ~10-line function that just creates and calls a function.  This is just basic JS, no special logic to reuse, etc....

jkuester:
Sorry for barging in here, but I had finally circled back yesterday to look at this review and as I went through it I decided to just throw a commit together to try and get things unblocked here.  (Mostly it took me so long to finally get the data flow straight in my head for these targets that I wanted to get the functionality locked in before I forgot how this worked again... :sweat_smile: )

webapp/src/ts/services/rules-engine.service.ts
dianabarsan: I still think this should not be part of the rules-engine service. At least it's own utility file in webapp.

webapp/src/ts/services/rules-engine.service.ts
```diff
@@ -510,13 +510,29 @@ export class RulesEngineService implements OnDestroy {
...
 
-      return this.processTargetDocuments(targetInterval, settings);
+      return this.processTargetDocuments(targetInterval, settings, reportingPeriod);
     } catch (error) {
       console.error('Error fetching previous month targets:', error);
       return [];
     }
   }
 
+  private getValueFromFunction(str: string | undefined, ...args: unknown[]) {
```
dianabarsan: This is nice and simplified. Thanks!

Edwin-Rugoogamu: Thankyou @jkuester and @dianabarsan  for the support and work around.

Edwin-Rugoogamu:
Thankyou @jkuester and @dianabarsan for the support and work around.

webapp/src/ts/services/rules-engine.service.ts
jkuester: :+1: [Awhile back](https://github.com/medic/cht-core/pull/9646/) I was tinkering with this utility file concept in webapp.  I have added a commit to this PR with the best approach I could come up with that addressed these constraints:

- Lib code and unit tests are in TS.
- Unit tests for lib code do not need Karma.
- Unit tests for lib code can use sinon to stub imports.
- Karma tests can use `sinon` to mock functions from lib code.

The main concession I had to make to get things working within these constraints was to include a custom `tsconfig` for the mocha unit tests.  Our existing tsconfig for webapp are targeted at the Angular build (and Karma tests).  This tsconfig targets ESM but (I think because of webpack) does not care about having the file extensions (e.g. `.ts` or `.js`) in the `import` lines.

Trying to run the lib unit tests while targeting ESM led down a rabbit-hole of complexity where it became more and more clear that trying to reuse the Angular TS config was a bad idea.  Having a custom tsconfig allowed me to bypass this whole mess and just target CJS modules (like we do in cht-datasource).  Open to other ideas here, but I burned a bunch of time going on circles with this already and like I said, this is the best I could come up with....

webapp/tests/mocha/.mocharc.js
```diff
@@ -0,0 +1,11 @@
+const chaiAsPromised = require('chai-as-promised');
+const chai = require('chai');
+chai.use(chaiAsPromised);
+
+process.env.TS_NODE_PROJECT = 'tests/mocha/tsconfig.mocha.json';
+process.env.UNIT_TEST_ENV = '1';
+
+module.exports = {
+  spec: 'tests/mocha/unit/**/*.spec.*',
+  require: 'ts-node/register',
```
dianabarsan, line 10: This is very elegant! I love it!

PR #10511, opened 2025-12-02
feat(#10398): store pre-purged doc counts per user

### LINKED ISSUE #10398: Store pre-purged doc counts per user
Doc counts for users are a permanent factor for performance, both server-side and client-side. 
We store post-purging doc counts for when they replicate, but there is no visibility on the pre-purge doc counts, a number which impacts server performance. 

When storing user doc counts, also store pre-purging doc counts. 

https://github.com/medic/cht-core/blob/master/api/src/services/replication.js#L19

jack-chaudier:
# Description
This PR addresses issue #10398 by adding visibility into "pre-purged" document counts in the replication limit log. Previously, only the post-purge count was stored, making it difficult to assess the performance impact of large user document sets before filtering.

## Changes
- Updated `replication.js` to pass `allowedIds.length` (pre-purge count) to the logging service.
- Updated `replication-limit-log.js` to:
    - Accept the new `prePurgeCount` argument.
    - Store it in the `medic-logs` document.
    - Treat significant changes in `prePurgeCount` (diff > 100) as a trigger to persist a new log entry.

Fixes #10398

# Code review checklist

# License

### DIFF
--- a/api/src/services/replication-limit-log.js
+++ b/api/src/services/replication-limit-log.js
@@ -42,8 +42,9 @@ const isLogDifferent = (oldLog, newLog) => {
 
   const countDiff = Math.abs(oldLog.count - newLog.count);
 
-  const oldPrePurge = oldLog.prePurgeCount || 0;
-  const newPrePurge = newLog.prePurgeCount || 0;
+  const oldPrePurge = oldLog.all_docs_count || 0;
+  const newPrePurge = newLog.all_docs_count || 0;
+
   const prePurgeDiff = Math.abs(oldPrePurge - newPrePurge);
 
   if (countDiff > LOG_COUNT_DIFF || prePurgeDiff > LOG_COUNT_DIFF) {
@@ -86,7 +87,7 @@ const logReplicationLimit = (userName, count, prePurgeCount) => {
     user: userName,
     date: moment().valueOf(),
     count,
-    prePurgeCount
+    all_docs_count: prePurgeCount
   };
 
   return persistLog(info)
--- a/api/tests/mocha/services/replication-limit-log.spec.js
+++ b/api/tests/mocha/services/replication-limit-log.spec.js
@@ -26,34 +26,16 @@ describe('Replication Limit Log service', () => {
         });
     });
 
-    it('should persist log', () => {
+    it('should persist log with all docs count', () => {
       const getStub = sinon.stub(db.medicLogs, 'get').returns(Promise.reject({ status: 404 }));
       const putStub = sinon.stub(db.medicLogs, 'put').returns(Promise.resolve());
       const logType = replicationLimitLogService.LOG_TYPE;
-      const expectedDoc = {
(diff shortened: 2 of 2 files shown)


api/tests/mocha/services/replication-limit-log.spec.js
```diff
@@ -44,6 +44,26 @@ describe('Replication Limit Log service', () => {
           chai.expect(putStub.args[0][0]).to.deep.include(expectedDoc);
         });
     });
+
+    it('should persist log with pre-purge count', () => {
+      const getStub = sinon.stub(db.medicLogs, 'get').returns(Promise.reject({ status: 404 }));
```
jkuester: Minor, but since there is no existing doc (and that flow is covered in a different test below), this is pretty much a duplicate of the test above.  Can we just add the `prePurgeCount` values to the `'should persist log'` test and remove this test?

api/tests/mocha/services/replication-limit-log.spec.js
jkuester, line 111: Can you add `prePurgeCount` values to this test to just confirm that when the values are too close, the log is not considered different?

api/src/services/replication-limit-log.js
```diff
@@ -77,11 +81,12 @@ const getReplicationLimitLog = (userName) => {
...
 };
 
-const logReplicationLimit = (userName, count) => {
+const logReplicationLimit = (userName, count, prePurgeCount) => {
   const info = {
     user: userName,
     date: moment().valueOf(),
-    count
+    count,
+    prePurgeCount
```
jkuester: Okay, I am probably way overthinking this 😅 but saying `prePurge` in the field name feels a bit off to me. I think it is because the CHT user doc _purging_ is technically the reason why the value for this field might be different from `count`, but the _value_ contained in this field is meaningful regardless of if purging happens.  

Essentially, `count` is the "total number of docs that would be replicated for the user."  This new field is just the "total number of docs on the server associated with the user".  In the future there might be other reasons besides purging that these two  numbers are different.

Unfortunately, I have not been able to come up with another name that I super happy with... 😞 The best I have is:

- `docs_on_server_count`
- `all_docs_count`

If one of these makes better sense to you (or you can come up with something better) lets go with that!  

---

Either way, it would be good to use snake_case instead of camelCase in the property name since that is the typical practice for doc properties. 👍

jack-chaudier:
Thanks @jkuester! I've addressed the feedback:

- **Renaming**: Updated the DB field to `all_docs_count` (snake_case) as suggested.
- **Tests**:
  - Removed the duplicate test case in `put()`.
  - Added the `all_docs_count` check to the `_isLogDifferent` test to confirm it handles small differences correctly.

Ran `api/tests/mocha/services/replication-limit-log.spec.js` locally—all 9 tests passed, and ready for re-review.

jkuester (approved):
Awesome!  LGTM! 🚀

Issue #10379, opened 2025-10-14
The scalability workflow does not run automatically when beta tags are pushed


dianabarsan:
**Describe the issue**
The [scalability suite](https://github.com/medic/cht-core/blob/fbf0a0cca4c6eab1471f46a6dea3c6e5ec535f70/.github/workflows/scalability.yml#L3C1-L9C19) should automatically run when a `Build and test` job has passed on a beta tag. 
This did not occur on the 4.22 betas. 

**Describe the improvement you'd like**
It seems that the trigger is incorrectly set up and it does not fire. Change the trigger so that it fires. 

**Describe alternatives you've considered**
Manually running the scalability suite is also an option.

ncaldera:
I would like to work on this improvement.

dianabarsan:
Relevant conversation here: https://github.com/medic/cht-core/pull/10513#issuecomment-3737723854

TLDR: we didn't find a way to only trigger the workflow for beta tags and only after the CI images were published to docker (requires a successful build). 
We decided to keep the manual trigger alone.

dianabarsan:
We didn't find a clean solution through GHA triggers. Since this is an infrequent enough action, we resolved that it can be done manually.

PR #10513, opened 2025-12-04
fix(#10379): fix scalability workflow not triggering for beta tags

ncaldera:
# Description

This PR addresses issue #10379. The scalability workflow was intended to run automatically when beta tags are pushed, but it wasn’t triggering as expected. This was caused by the use of unsupported filters under workflow_run, which prevented GitHub Actions from recognizing beta tag events. This PR updates the workflow to use a supported trigger/filtering mechanism so that scalability tests correctly run for tagged beta builds. I updated .github/workflows/scalability.yml to use a supported filtering mechanism for beta tags that ensures that the scalability workflow runs after Build and test completes when the triggering ref matches the beta tag format.

# Code review checklist

# License


.github/workflows/scalability.yml
jkuester, line 9: I think we need to remove this config since it is not supported. It seems like [you can filter based on _branches_](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#limiting-your-workflow-to-run-based-on-branches), but not `tags`.

.github/workflows/scalability.yml
```diff
@@ -20,6 +20,7 @@ env:
 
 jobs:
   build:
+    if: contains(github.event.workflow_run.head_branch, 'beta')
```
jkuester, line 23: I think we are missing  a couple of things here.

- I think we still want to allow for manually triggering the workflow (via `workflow_dispatch`).
- We really should only run this job if the "Build and test" workflow actually completed _successfully_. Otherwise there is not much point in trying to trigger this _after_ the  "Build and test" workflow finishes. 
- This would still trigger the workflow run for any _branch_ with a name that contains "beta" (and not just tags). 

According to Copilot, something like this might work:

```yaml
    if: github.event_name == 'workflow_dispatch' ||
      (
        github.event.workflow_run.conclusion == 'success' &&
        startsWith(github.event.workflow_run.head_branch, 'refs/tags/') &&
        contains(github.event.workflow_run.head_branch, 'beta')
      )
```

dianabarsan:
Thanks for the tag @jkuester . 
My understanding is that if we use the `if` clause in the job, we will have scalability workflow jobs start for every branch but none of the jobs will run? So our "scalability" rab will be full of cancelled jobs. This is not ideal but not a disaster. If there's a way around it, we should take it, but if there is no solution, I suppose we can live with that. 

On the other issue of `workflow_dispath`, we definitely definitely want to be able to manually trigger the scalability suite for any branch.

jkuester:
> So our "scalability" rab will be full of cancelled jobs. This is not ideal but not a disaster. If there's a way around it, we should take it, but if there is no solution, I suppose we can live with that.

@dianabarsan the only alternative I found is `#2` where we roll back to the previous config where the "Scalability workflow" is triggered directly when beta tags are pushed (without waiting for the "Build and test" workflow).  That would mean the workflow gets auto-triggered _only for beta tags_ (and we avoid a bunch of cancelled/noop jobs). But it would also mean that we would trigger the build for beta tags _without waiting to see if the "Build and test" workflow passes or not.  IMHO this is the best way to go, but wanted your feedback since it would basically be undoing your previous changes.

dianabarsan:
That's unfortunate. #2 meant we needed to add a continuous check for the images to get published, and that made this job start and wait for 6 hours if a build failed. I really didn't like that. 
It feels like it's better to just manually trigger it (what are the odds of getting a green build from one go lol).

jkuester:
:+1:  I am good with just a manual trigger.

@ncaldera can you update the workflow to just remove the new `if` as well as the existing `on.tags` config? (Just leaving `on.workflow_dispatch` so the job can be manually triggered.  Then, can you update [this line](https://github.com/medic/cht-core/blob/369db04ce92c9815c46342cfe96143f83bbcbf67/.github/ISSUE_TEMPLATE/z_release_major.md?plain=1#L38) in the release issue template to instead say something like:

> - [ ] Once the CI completes successfully, use the "Run Workflow" drop-down [to run the Scalability Workflow](https://github.com/medic/cht-core/actions/workflows/scalability.yml). Once it passes, check for the scalability results [here](https://github.com/medic/scalability-results/tree/main/results). More info in the  [scalability documentation](https://github.com/medic/cht-core/blob/master/tests/scalability/README.md).

PR #10521, opened 2025-12-11
chore(#10520): secure release notes script token, exclude dependabot commits

### LINKED ISSUE #10520: Remove dependencie of on disk GH token for release ntoes
**Describe the issue**
When you use the [release notes](https://github.com/medic/cht-core/tree/master/scripts/release-notes) script, it requests you put a GitHub token on disk in a `token.json` file.  This is insecure as other [rogue apps](https://about.gitlab.com/blog/gitlab-discovers-widespread-npm-supply-chain-attack/) could harvest and exfiltrate this token. 

Prompt the user for the token and only keep it in memory for the duration of the script's run time.

* Keep it as is
* do a 1pass integration

mrjones-plip:
# Description

This PR secures release notes script token, moves it to be run in CI and hide dependabot noise.  Specifically:
* defaults to CI running release notes script
* hard codes `cht-core` as repo and but still allows milestone to be passed in both in CI and locally
* allows easy development or re-run both via local execution 
*  securely using `gh`'s token when running locally
* refactoring of existing code per feedback
* wraps existing code in massive `main()` function so prompt for token comes first and actually waits instead of barreling ahead and having GH auth fail
* Adds URLs for commits as sometimes they have no issue, commit or PR number in them.
* Excludes dependabot commits from showing as errors/missing a milestone 

closes #10520 and #10613

## Testing

First delete `token.json` if you have one on disk.

### check local output 

See [release notes](https://github.com/medic/cht-core/blob/10520-secure-release-notes-token/scripts/release-notes/README.md) to see sample screenshots and sample output.  Then test with different scenarios:

```
# should fail 'missing milestone'
GITHUB_TOKEN=$(gh auth token) node index.js  

# should fail 'check commits'
GITHUB_TOKEN=$(gh auth token) node index.js 5.1.0 

# should succeed and generate release-notes.md file
GITHUB_TOKEN=$(gh auth token) node index.js  5.1.0  --skip-commit-validation
```

### check CI output

Call CI manually to run with different scenarios:

```
# should fail 'missing milestone'
gh workflow run release-notes.yml --ref 10520-secure-release-notes-token   

# should fail 'check commits'
gh workflow run release-notes.yml  -f milestone=5.1.0 --ref 10520-secure-release-notes-token   

# should succeed and generate artifact
gh workflow run release-notes.yml  -f milestone=5.1.0  -f skip_commit_checks='--skip-commit-validation' --ref 10520-secure-release-notes-token   
```

Make sure the artifact has the expected content.

### compare `master` vs `10520-secure-release-notes-token`

do the same tests above on `master` to ensure the markdown output is the same.  

**NB** - This branch skips all dependabot commits, so these will not be shown.  Note this when you're comparing outputs.

# Code review checklist

- [X] Internationalised: All user facing text

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10520-secure-release-notes-token/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10520-secure-release-notes-token/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10520-secure-release-notes-token/docker-compose/cht-couchdb-clustered.yml)

# License


mrjones-plip:
Still a WIP while I do some cleanup and remember how JS actually works!

mrjones-plip:
Ready for review!

scripts/release-notes/index.js
```diff
@@ -1,294 +1,307 @@
+
```
jkuester:
```suggestion
```

scripts/release-notes/index.js
```diff
@@ -1,294 +1,307 @@
...
 const minimist = require('minimist');
 const { Octokit } = require('@octokit/core');
 const { paginateGraphql } = require('@octokit/plugin-paginate-graphql');
+require('xhr-mock/lib/handle');
```
jkuester: Is this needed or just leftover from some testing you did?

```suggestion
```

scripts/release-notes/index.js
```diff
@@ -318,27 +331,48 @@ ${formatGroups(types)}
...
+const getTokenFromUser = () => {
+  return new Promise((resolve) => {
+    process.stdout.write('Enter Your GitHub Token (40 chars, starts with "ghp_"): ');
+    process.stdin.once('data', (input) => {
+      const value = input.toString().trim();
+      if (value.length !== 40) {
+        throw new Error(`Token is not 40 characters. Check token value and try again. `);
+      }
+      resolve(value);
+    });
```
jkuester: I don't think this validation works for the newer "Fine-grained" tokens. I just tried generating one of those and it was 94 chars long and started with `github_pat_`.

scripts/release-notes/README.md
jkuester: Needs updated.

jkuester:
```suggestion
node index.js [OPTIONS] REPO MILESTONE
```

scripts/release-notes/index.js
```diff
@@ -1,294 +1,307 @@
...
-const validateIssues = issues => {
-  const errors = issues
-    .map(validateIssue)
-    .filter(error => !!error);
-  if (errors.length) {
-    console.error(JSON.stringify(errors, null, 2));
-    throw new Error('Some issues are in an invalid state');
-  }
-};
+  const ignorePrefixes = label => {
```
jkuester: Minor but `ignorePrefixes` feels like the opposite of what this function returns. :)  Unless I am missing something here, this function returns `true` when the given label does  _not_ match one of the prefixes to ignore (meaning that we should include the issue and not filter it out).

scripts/release-notes/index.js
```diff
@@ -318,27 +331,48 @@ ${formatGroups(types)}
...
 
-const getCommits = async () => {
-  const latestReleaseName = await getLatestReleaseName();
-  const milestoneBranch = await getMilestoneBranch();
-  const commitsForRelease = await getCommitsForRelease(latestReleaseName, milestoneBranch);
-  validateCommits(commitsForRelease);
-  return commitsForRelease;
-};
+    console.log('Saving output to release.notes.md...');
+    const fs = require('node:fs/promises');
```
jkuester: Move this to the top with the rest of the requires.

scripts/release-notes/index.js
```diff
@@ -318,27 +331,48 @@ ${formatGroups(types)}
...
+    .catch(console.error);
 };
 
-Promise.all([ getIssues(), getCommits() ])
-  .then(([ issues, commits ]) => output(issues, commits))
-  .catch(console.error);
+const getTokenFromUser = () => {
+  return new Promise((resolve) => {
+    process.stdout.write('Enter Your GitHub Token (40 chars, starts with "ghp_"): ');
+    process.stdin.once('data', (input) => {
```
jkuester: This current implementation results in my token value getting printed in my terminal (when it type/paste it) and presumably stored into my shell history.  This seems undesirable from a security perspective.  

I think with these base Node functions we need to somehow leverage [setRawMode](https://nodejs.org/api/tty.html#readstreamsetrawmodemode) to prevent the input from getting printed.  Alternatively, we could use a dependency like [read](https://www.npmjs.com/package/read) to do this with less boilerplate here...

scripts/release-notes/index.js
```diff
@@ -1,294 +1,307 @@
...
-  const milestoneBranch = [...MILESTONE_NAME.split('.').slice(0, -1), 'x'].join('.');
-  const branchExists = await queryRepo(
-    `ref(qualifiedName: "refs/heads/${milestoneBranch}") {
-        target { oid }
-      }`
-  ).then(({ repository }) => repository.ref);
-  if (branchExists) {
-    return milestoneBranch;
+
+const main = async(token) => {
```
jkuester: Personally, I am not a huge fan of big wrapper functions like this.  They almost always exist to manage _state_ and I hate state.  😅

In this case, the state being tracked is the `token` (and really just the `octokit` that we create with the token).  Instead of wrapping all this code in this `main` function, IMHO it would be better to create the `octokit` in your IIFE (along with the other code here that is not nested in a function) and then pass the `octokit` as a paramter to each function that needs it.  There are more parameters to pass, but there is less high-level state that has to be considered.  

This is more of a personal preference, though, so do not need to take this approach unless it makes sense to you.  If you want to keep the `octokit` as high-level state (and avoid passing it as a parameter), I would still recommend ditching the `main` function (which does not really add any value). Instead, add all the main control flow and sub-funtions that require `octokit` to the IIFE. Any functions/variables that do not need `octokit` can stay at the top-level outside of the IIFE.

jkuester (changes requested):
Left some minor comments, but besides the token-validation piece I think this all should work. 👍 

This whole approach still bugs me, though, because it is more hassle for the user _every_ time they need to run the script. Plus I still have to manually store my GitHub token somewhere or recreate it every time I need to run the script.  

I have been trying to come up with better alternatives besides storing the token value in a file/envar or accepting it as input when running the script.  Actually having the script do a GitHub OAuth flow via the browser would be the "best" way, but it would require registering a GitHub OAuth app and extra coding around the auth flow.

One interesting alternative would be to simply require the user to have the [GitHub CLI app](https://cli.github.com/) installed. The [`gh auth token` command](https://cli.github.com/manual/gh_auth_token) will give you the actual authentication token that can be used when accessing the GitHub API.  So, a user would need to install the `gh` util and then login to their GitHub account with it (this would actually just be doing the GitHub Oauth flow).  Then, the release notes script could programmatically get the auth token (by running `gh auth token`). No input would be required from the user when running the release notes script.

mrjones-plip:
I was reading [Stefan's latest post](https://sdufresne.info/blog/2026-01-09-access-github-actions-artifacts-across-runs/) and this section caught my eye:

> We’re going to use the [Github CLI](https://cli.github.com/) for this, because you get it for free when running scripts under GitHub Actions

And I was like,"ah-ha! we'll just run this script in CI and cache the markdown byproduct" - as long as it's easy to run [ad-hoc](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow) to fix the milestone/type/etc errors, this really cleanly solves the security concerns [originally raised](https://github.com/medic/cht-core/issues/10520). 

Before I do any POC/refactoring - thoughts @jkuester ?

scripts/release-notes/index.js
mrjones-plip: removed - thanks!

scripts/release-notes/index.js
mrjones-plip: yup yup - removed!  (also not using user supplied any more)

scripts/release-notes/README.md
mrjones-plip: agreed! let's keep this unresolved until we finalize our approach.  that I'll have an outstanding to do before we merge.

scripts/release-notes/index.js
mrjones-plip: fair! renamed to `includePrefixes()`

scripts/release-notes/index.js
mrjones-plip: ended up removing all together using latest refactor to be all on actions instead of local

scripts/release-notes/index.js
mrjones-plip: solved by moving to be done in GH CI remotely, not shell history

scripts/release-notes/index.js
mrjones-plip: ok, i removed `main()` and moved a bunch of functions outside the IIFE - lemme know how I did!

mrjones-plip:
I went ahead and converted this to a GH CI you can call from the CLI and pass in the milestone you're looking to work with.  
While I don't know why we can't get the nice web GUI on GH ([per the docs](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow)), it's trivial to run from the CLI - which is likely easier any way. 

to test:

1. kick off CI manually, passing in milestone. Note that we're passing in the branch which won't have do when merge: `gh workflow run release-notes.yml -f version=5.1.0 --ref 10520-secure-release-notes-token`
2. go to [GH CI for this job](https://github.com/medic/cht-core/actions/workflows/release-notes.yml?query=event%3Aworkflow_dispatch)
3. click the most recent job to see the details including success or errors

I was going to work on creating an artifact which you could [download](https://docs.github.com/en/actions/tutorials/store-and-share-data) from the CLI (get IDs with `gh run list --workflow="Release notes generator"` then download with `gh run download RUN_ID`) but decided to wait for feedback before addressing further. Once we implement downloading artifacts, I think it will make step 2 & 3 super trivial.

Finally - a big caveat to this approach is that it takes ~2min per run. This may be an untenable long time to wait maybe? But seems like you can get a list to fix all the issues and re-run easily enough :shrug: 

---------

After thinking on this some more, what if we just lump this in with the on going CI and always archive the artifact? We would have to add logic to automagically get the next unreleased milestone, but this seems like it'd be a win anyway.  It'd make the script easier to call in general as you could just call it now that we're hard coding `cht-core` as the repo and the version is auto discovered for you.  

With this in place, you have to wait `0` seconds because the release note artifact will always just waiting there for you instructing you with what you need to fix. You only need to wait when you re-run (which admittedly is still a few times)

-------

After thinking on this some more some more - I left no easy way to run this locally - this clearly needs to be added to enable local development.

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,27 @@
+name: Release notes generator
+
+on:
```
jkuester, line 3: Since this is already a workflow, maybe we should just trigger it automatically when a beta tag is pushed (in addition to the `workflow_dispatch`)?

```yaml
on:
  push:
    tags:
      - "*-beta.*"
```


We would need something like this to calculate the version we want to use:

```shell
          if [ "${{ github.event_name }}" = "push" ]; then
            TAG="${GITHUB_REF#refs/tags/}"
            VERSION="${TAG%%-beta*}"
          else
            VERSION="${{ github.event.inputs.version }}"
          fi
```

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Use Node 24
+        uses: actions/setup-node@v6
+        with:
+          node-version: 24
+      - name: Install Deps & Create Markdown (or error)
+        run: |
+          cd scripts
+          npm ci
+          cd release-notes
+          GH_TOKEN=${{ secrets.DEPLOY_TO_GITHUB_PAGES }} node index.js cht-core ${{ github.event.inputs.version }}
```
jkuester: > GH_TOKEN=${{ secrets.DEPLOY_TO_GITHUB_PAGES }}

Is there something special about this `DEPLOY_TO_GITHUB_PAGES` token?  Could we just use the normal `secrets.GITHUB_TOKEN` here?

jkuester: Is the idea that we are just printing the output to the workflow logs and the user can copy it out from there?

scripts/release-notes/index.js
```diff
@@ -2,30 +2,28 @@ const minimist = require('minimist');
 const { Octokit } = require('@octokit/core');
 const { paginateGraphql } = require('@octokit/plugin-paginate-graphql');
 const ExtendedOctokit = Octokit.plugin(paginateGraphql);
-const octokit = new ExtendedOctokit({
-  auth: require('../token.json').githubApiToken,
-  userAgent: 'cht-release-note-generator',
-});
 
+
+const token = process.env.GH_TOKEN;
```
jkuester: Minor, but is there a reason to use `GH_TOKEN` instead of `GITHUB_TOKEN`? It is just that `GITHUB_TOKEN` is the default secret name when running workflows on GitHub and I like for everything to be consistant...  😅

.github/workflows/release-notes.yml
jkuester: Oh, I see, you mentioned in a comment that you want to make this store an attachment. 👍

.github/workflows/release-notes.yml
mrjones-plip: Love the idea of using `"*-beta.*"` nice!  I've added that, but also kept the `workflow` for ad hoc executions

for deriving the version - I did added a [programmatic call](https://github.com/medic/cht-core/blob/cf0b847bf0ca1bfb812d55a40e268737252705bc/scripts/release-notes/index.js#L67) to get all the milestones, sort them and then get the next one.  I _think_ this will work for all scenarios?  I'll attempt to resolve this, but please unresolve if I'm being un-smart :sweat_smile:

.github/workflows/release-notes.yml
mrjones-plip: yes!  I checked the secrets in the repo and we don't actually have a secret called `GITHUB_TOKEN` only one called
`DEPLOY_TO_GITHUB_PAGES`.  When I tried to use `GITHUB_TOKEN` I got an [error in my CI](https://github.com/medic/cht-core/actions/runs/21420680257/job/61679282543) b/c it's a non-existant string/secret:

```
/home/runner/work/cht-core/cht-core/scripts/release-notes/index.js:60
  console.log('Logging in to GitHub with token that is', TOKEN.length, 'chars...');
                                                               ^
TypeError: Cannot read properties of undefined (reading 'length')
    at /home/runner/work/cht-core/cht-core/scripts/release-notes/index.js:60:64
    at Object.<anonymous> (/home/runner/work/cht-core/cht-core/scripts/release-notes/index.js:351:3)
    at Module._compile (node:internal/modules/cjs/loader:1761:14)
    at Object..js (node:internal/modules/cjs/loader:1893:10)
    at Module.load (node:internal/modules/cjs/loader:1481:32)
    at Module._load (node:internal/modules/cjs/loader:1300:12)
    at TracingChannel.traceSync (node:diagnostics_channel:328:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:245:24)
    at Module.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:154:5)
    at node:internal/main/run_main_module:33:47
Node.js v24.13.0
Error: Process completed with exit code 1.
```

I confess this is brittle, but I think it's OK for now :shrug:

mrjones-plip: right! I still need to add this feature - stay tuned!

scripts/release-notes/index.js
mrjones-plip: nope!  `GITHUB_TOKEN` is more expected and more legible.  Fixed!

mrjones-plip:
>  Do you know if the script currently fails with a non-zero exit code if the validation fails and it cannot generate the release notes? If not, we should probably make it work like that so the workflow actually fails/succeeds based on if the notes could be generated.

Thanks for the input!  I hope to iterate on this a bit more to add saving the artifacts, but I think it should always save the output regardless of the exit code, yeah?  This way if there's some weird auth error or some unexpected error, along with the expected `Some commits are in an invalid state` it'll be easy to find the results.

I say this as the man whose never implemented CI artifacts.  Does it only work if there's a zero exit code?

w00t!

scripts/release-notes/README.md
mrjones-plip: fixed in f7a87fabc!

.github/workflows/release-notes.yml
mrjones-plip: artifacts added!

mrjones-plip:
Ok - maybe all ready to go?  You tell me!  To kick of a manual run of CI, use the branch which is missing (and won't be needed) in the [updated readme](https://github.com/medic/cht-core/tree/10520-secure-release-notes-token/scripts/release-notes):

```
gh workflow run release-notes.yml --ref 10520-secure-release-notes-token
```

Otherwise, running it locally should still work per the readme:

```
GITHUB_TOKEN=$(gh auth token) node index.js
```

scripts/release-notes/README.md
```diff
@@ -1,35 +1,66 @@
...
 
-### Setup
+![artifacts.png](artifacts.png)
 
-Install the required dependencies by running the following from the `scripts` directory:
+Every time a beta branch is cut, CI automatically runs and saves the output of the [release note job](https://github.com/medic/cht-core/actions/workflows/release-notes.yml). Each run will have `release-error` and `release-notes` artifacts. You'll need to fix all the commit errors found in `release-error` (see below for a how-to) and then manually re-run the CI. Use the `gh` command locally to re-run the action run on GitHub:
 
 ```shell
-npm ci
+gh workflow run release-notes.yml
```
jkuester: Probably anyone trying to run this will want to include the target ref, right?

```suggestion
gh workflow run release-notes.yml --ref 5.0.0-beta.1
```

.github/workflows/release-notes.yml
jkuester: Reopening this. I appreciate the effort to keep things simple here for the user, but I think that only supporting the "latest" is going to be overly limiting. There are many times where we might have multiple milestones going simultaneously  (e.g. doing a patch release while we have an ongoing feature release while we have already starting preparing for a major release in the near future).  These milestones might be created in any order and then the actual releases happen in a totally different order.

I am open to other ideas here, but at this point I still feel like having the script accept a target version is best. (And having this job automatically triggered for a `beta` tag and then it ends up generating release notes for a completely separate version would be super confusing....)

.github/workflows/release-notes.yml
jkuester: Okay, this is minor and `DEPLOY_TO_GITHUB_PAGES` _should_ work, but just to give this one more go:

- `GITHUB_TOKEN` should be [automatically available to all workflows](https://docs.github.com/en/actions/tutorials/authenticate-with-github_token). It would not be listed in our custom secrets for the repo (and that is the point. We don't have to manage it.).
- When you tried to use `GITHUB_TOKEN`, were you still manually setting the envar to the value from the secrets? This has tripped me up before, but GitHub _secrets_ are not automatically available as envars when running the job. You have to manually configure them: 

```yaml
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

Maybe I am the one missing something here and this is doomed to not work :sweat_smile: , but I do see us using it [in other cht-core workflows](https://github.com/medic/cht-core/blob/5fcfbcbafb4a0d990f6b2dabb4d7f5fd0489e7dd/.github/workflows/build.yml#L89)...

scripts/release-notes/index.js
```diff
@@ -2,41 +2,28 @@ const minimist = require('minimist');
 const { Octokit } = require('@octokit/core');
 const { paginateGraphql } = require('@octokit/plugin-paginate-graphql');
 const ExtendedOctokit = Octokit.plugin(paginateGraphql);
-const octokit = new ExtendedOctokit({
-  auth: require('../token.json').githubApiToken,
-  userAgent: 'cht-release-note-generator',
-});
 
+
```
jkuester, line 6:
```suggestion
```

scripts/release-notes/index.js
```diff
@@ -2,41 +2,28 @@ const minimist = require('minimist');
...
-   --skip-commit-validation  Skip validation of commits
-   
-Repository: The name of the repository (e.g. cht-core).
-
-Milestone: The name of the milestone (e.g. 2.15.0).
-`);
+  Usage: node index.js [OPTIONS] MILESTONE
+  
+  Generates a changelog given a GH report and milestone. Requires a GitHub API token to be configured. 
+  See the README for more information.
```
jkuester, line 18: Make sure this is up-to-date with whatever we decide to do regarding the MILESTONE param....

scripts/release-notes/index.js
```diff
@@ -318,27 +344,29 @@ ${formatGroups(types)}
...
+    const filtered = await filterIssues(issues);
+    return groupIssues(filtered);
+  };
 
-Promise.all([ getIssues(), getCommits() ])
-  .then(([ issues, commits ]) => output(issues, commits))
-  .catch(console.error);
+  return Promise.all([ getIssues(), getCommits() ])
+    .then(([ issues, commits ]) => output(issues, commits))
+    .catch(console.error);
```
jkuester:
```suggestion
  .catch(err => {
    console.error(err.message);
    process.exit(1);
  });
```

When there is an error, I think we should end the process with an exit code.  This should cause the workflow to fail when the release notes generation fails.

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,42 @@
...
+          retention-days: 10
+          compression-level: 0
+      - name: Upload error artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: release-error
+          path: ~/release-error.txt
+          retention-days: 10
+          if-no-files-found: ignore
+          compression-level: 0
```
jkuester: Along with [my comment here](https://github.com/medic/cht-core/pull/10521/changes#r2756026414) regarding failing the workflow with an exit code when there was an error in the script, I also think we should not be redirecting syserr to this file attachment.  This makes it so that when there is a problem with the release notes, I have to download the `release-error` attachment, unzip it, and open the txt file just to see what the error is. It seems way more convenient to just let syserr print straight to the workflow logs and then fail the workflow.  

Then, the release engineer can follow the standard process for fixing a CI run where you see the workflow fails, read the logs, fix the problem, and then rerun the workflow.

scripts/release-notes/index.js
```diff
@@ -318,27 +344,29 @@ ${formatGroups(types)}
 
 Thanks to all who committed changes for this release!
 
-${formatCommits(commits)}
-`);
-};
+${formatCommits(commits)} `;
+    console.log('Done! See below');
+    console.log('-------------Everything below this is the release notes markdown-------------');
+    console.log(releaseNotes);
```
jkuester: Honestly, since end result here is to have a `release-notes.md` file that gets uploaded in the workflow as an attachment, IMHO it would be more simple to just write `releaseNotes` to the file here. (instead of printing to the console and having to scrape the console output into the file).

That would simplify the workflow config and it would also mean that your final `release-notes.md` would only contain exactly what you want. As things sit now, the `release-notes.md` file will include all of the other log messages that get printed during the generation process....

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,42 @@
...
+        with:
+          node-version: 24
+      - name: Install Deps & Run Node Script
+        env:
+          GITHUB_TOKEN: ${{ secrets.DEPLOY_TO_GITHUB_PAGES }}
+        run: |
+          cd scripts
+          npm ci
+          cd release-notes
+          GITHUB_TOKEN=$GITHUB_TOKEN node index.js > ~/release-notes.md 2> ~/release-error.txt
```
jkuester: `GITHUB_TOKEN=$GITHUB_TOKEN`

Do we need this? 🤔

scripts/release-notes/README.md
mrjones-plip: actually not in this case. the ref is a branch of where the workflow script is that you want to run, not the milestone you're targeting.  So we're using it in development here in our branch, but once it's merged to `main` this extra flag won't be needed.

mrjones-plip:
Thanks for the feedback!  I'll jump on this in a day or two.

tl;dr - I'll for sure have to refactor to account for a dynamic milestone.  Thanks very much for gentle reminder of all the work flows we have to support - supper helpful!

.github/workflows/release-notes.yml
mrjones-plip: dude - yes!! It all works now - thanks for the push-repush back.  now that I see it, using the built in token is so much cleaner than overloading the `DEPLOY_TO_GITHUB_PAGES` which could change out from under us.  I think the earlier refactor to normalize using `GITHUB_TOKEN` inside (node) and outside (env vars) also makes it easier to read/debug.

fixed in fb25c10dc.

.github/workflows/release-notes.yml
mrjones-plip: yup yup - I'm tracking.  First part is to not capture STDERR to a file in the CI job, done in 81969a29e.  Lemme test to see if I need to more in the `index.js` file now...

mrjones-plip: OK - just the CI workflow needed to updated! and yeah - this also feels better.  The error is faster/easier to read inline in CI now with out downloading zip etc - agreed:

image

I'll need to update the readme too, but I'll do that last when I finalize specifying the milestone inline

.github/workflows/release-notes.yml
mrjones-plip: strictly speaking, I think we could remove it.  However, I really liked the parallel structure of running it locally and running it in CI with near identical signature.  I'll push back a lil to keep it and see if it sticks with you ;)

scripts/release-notes/index.js
mrjones-plip: yes - this is the happy path - agreed!  While the initial commit did it in b25a5e04f1db6dc5ff723cf45689d0699a8effaa I had to faff about a bit to get the relative path to work in CI.  Net result though is it works in CWD locally and then references that same path to archive only the final release notes  artifact

mrjones-plip: also - I tried compression `0` in hopes of avoiding a `.zip` and just having an easy to get at `.md`, but alas, always a `.zip` :shrug:

scripts/release-notes/README.md
mrjones-plip: un-resolving per [comment below](https://github.com/medic/cht-core/pull/10521#discussion_r2755963194), tl;dr:

> I still feel like having the script accept a target version is best

.github/workflows/release-notes.yml
mrjones-plip: thanks for the sanity check - I've reestablished ability to pass in a milestone in c04f1043e

but...now CI will fail for auto push for beta - lemme think on this some!

scripts/release-notes/index.js
mrjones-plip: this is correct now that we've undone the auto milestone trickery.  I've tweaked this help text a bit, but it's still correct.

scripts/release-notes/README.md
mrjones-plip: oh -weird - I wasn't able to resolve or even reply to the comment below, but in a [deep link](https://github.com/medic/cht-core/pull/10521/changes/cf0b847bf0ca1bfb812d55a40e268737252705bc#r2767282385) I could?  anywho - resolving this now that c04f1043e is done.

.github/workflows/release-notes.yml
jkuester: > but...now CI will fail for auto push for beta - lemme think on this some!

Can we not use some shell logic (like in my original comment) to calculate the version for beta tags?

.github/workflows/release-notes.yml
jkuester: I am fine with keeping it, but it looks kinda weird to have this along with:
```yaml
        env:
          GITHUB_TOKEN: ${{ secrets.DEPLOY_TO_GITHUB_PAGES }}
```

Maybe we just need: 

```shell
GITHUB_TOKEN=${{ secrets.DEPLOY_TO_GITHUB_PAGES }} node index.js...
```

(not sure the magic variable interpolation works that that or not.... :sweat_smile:

mrjones-plip: yeah - sorry for the pedantry!  I settled on your approach of just the `env` and removed my prefix in d67d778e2

mrjones-plip:
OK!  Sorry for all the back and forth commits <> testing <> commits <> run CI <> etc, but I think we've ended up in a good place.

I removed the run on `push` for `*beta*` because I didn't know how to specify which branch/milestone to run against in an automatic run.  

Hopefully the [updated test section](https://github.com/medic/cht-core/pull/10521#issue-3720960059) and the  updated `readme` [covers](https://github.com/medic/cht-core/blob/10520-secure-release-notes-token/scripts/release-notes/README.md) everything, but lemme know if not. I'm happy to iterate more as needed!

PS - I was never able to get the [slick GH Workflow Web GUI](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow#running-a-workflow) to work.  If it starts working after we merge to `main` I think this CI based approach becomes even more viable: there's NO local dependencies, even to run `gh` (and thus not even `gh auth token`), when doing a release

mrjones-plip:
@jkuester - you've been busy I know, but just reaching out let you know I'm waiting for your review on this. It's not time sensitive, so happy to have bumped it from 5.1 -> 5.2

scripts/release-notes/index.js
```diff
@@ -255,7 +268,7 @@ Commits:
 
   const validateIssues = issues => {
     const errors = issues
-      .map(validateIssue)
+      .map((element) => validateIssue(element))
```
jkuester, line 271: No need to change this or anything, but I am curious why you updated these cases to create anonymous functions instead of just passing the named function as a parameter?

scripts/release-notes/README.md
```diff
@@ -1,35 +1,120 @@
...
+* Logged in to `gh` by running `gh auth login`
+* A GitHub account with read access to CHT Core Repo
+
+## Usage
+
+Start by manually running the CI job, specifying the milestone (`5.2.0` in this case), but using the `gh` command locally on your workstation:
 
 ```shell
-npm ci
+gh workflow run release-notes.yml -f version=5.2.0
```
jkuester:
```suggestion
gh workflow run release-notes.yml -f milestone=5.2.0
```

scripts/release-notes/README.md
```diff
@@ -1,35 +1,120 @@
...
+
+![commit.fixes.png](commit.fixes.png)
+
+### Skipping commit checks in CI
+
+To force CI to generate the release notes markdown and skip any commit checks, pass in an extra `skip` argument:
 
 ```shell
-node index.js [OPTIONS] REPO MILESTONE > tmp.md
+gh workflow run release-notes.yml -f version=5.2.0 -f skip_commit_checks='--skip-commit-validation'
```
jkuester:
```suggestion
gh workflow run release-notes.yml -f milestone=5.2.0 -f skip_commit_checks='--skip-commit-validation'
```

scripts/release-notes/index.js
mrjones-plip: 99% sure this was a linting suggested  optimization.  Lessee here...yups!  [tidy up sonar complaints](https://github.com/medic/cht-core/commit/7f65b10f68e1accca8663ec20632341f0a674c83)

PR #10522, opened 2025-12-12
feat(#9835): refactor create/update cht-datasource apis

jkuester:
# Description

Okay, so this is an iteration of the great stuff that we started in https://github.com/medic/cht-core/pull/10083/.  It started as a refactor of some of the logic flows in cht-datasource and just kinda spiraled out from there... :grimacing: 

cc @apoorvapendse

# Code review checklist

# License

### DIFF
--- a/shared-libs/cht-datasource/src/place.ts
+++ b/shared-libs/cht-datasource/src/place.ts
@@ -110,7 +110,7 @@ export namespace v1 {
      * Returns a generator for fetching all places with the given type
      * @param placeType the type of places to return
      * @returns a generator for fetching all places with the given type
-     * @throws InvaidArgumentError if no type is provided or if the type is not a supported place contact type
+     * @throws InvalidArgumentError if no type is provided or if the type is not a supported place contact type
      */
     const curriedGen = (placeType: ContactTypeQualifier) => {
       assertTypeQualifier(placeType);


api/src/auth.js
```diff
@@ -62,34 +32,53 @@ const hasPermission = (userCtx, permission) => {
   return _.some(roles, role => _.includes(userCtx.roles, role));
 };
 
-const checkUserPermissions = async (req, permissions = ['can_view_contacts'], altPermissions = []) => {
```
jkuester, line 65: I like having the dedicated function!  Changed up the params a bit so it was more flexible and easy to understand.

api/src/controllers/person.js
```diff
@@ -3,53 +3,46 @@ const ctx = require('../services/data-context');
...
-
+      const docs = await getPage(personType, req.query.cursor, req.query.limit);
       return res.json(docs);
     }),
 
     create: serverUtils.doOrError(async (req, res) => {
-      await auth.checkUserPermissions(req, [ 'can_view_contacts', 'can_create_people' ], [ 'can_edit' ]);
-
-      const personDoc = await createPerson()(req.body);
+      await auth.assertPermissions(req, { isOnline: true, hasAny: ['can_create_people', 'can_edit'] });
```
sugat009, line 33: question: when a person is created, this endpoint returns the created person. Does this change mean that `can_create_people` or `can_edit` permission contain `can_view_person` permission internally?

shared-libs/cht-datasource/src/local/libs/core.ts
```diff
@@ -15,125 +14,49 @@ export const validateCursor = (cursor: Nullable<string>): number => {
...
-  if (missingFieldsList.length > 0) {
-    throw new InvalidArgumentError(`Missing or empty required fields (${missingFieldsList.join(', ')}) for [${JSON
-      .stringify(updateInput)}].`);
-  }
-};
+export const assertFieldsUnchanged = (
+  original: DataObject,
+  updated: DataObject,
+  keys: string[]
+): void => keys.forEach((key) => assertFieldUnchanged(original, updated, key));
```
sugat009: suggestion: `assertFieldUnchanged` (singular, line 34) is only called from `assertFieldsUnchanged` (line 47). Consider inlining it, the `forEach` callback would just be a simple if/throw, which is clear enough without the extra indirection.

```typescript
// assertFieldsUnchanged (public, line 43-47)
  keys.forEach(key => {
    // assertFieldUnchanged (private, line 34-40)
    if (original[key] !== updated[key]) {
      throw new InvalidArgumentError(`The [${key}] field must not be changed.`);
    }
  });
```

shared-libs/cht-datasource/src/local/libs/core.ts
```diff
@@ -15,125 +14,49 @@ export const validateCursor = (cursor: Nullable<string>): number => {
...
+  key?: QueryKey;
+  startKey?: QueryKey;
+  endKey?: QueryKey;
+  limit?: number;
+  cursor?: Nullable<string>;
+}
 
-  if (a._id !== b._id) {
-    return false;
+const assertFieldUnchanged = (original: DataObject, updated: DataObject, key: string) => {
```
sugat009: This carries forward the same behavior from `ensureImmutability` where it throws on the first mismatched field. Consider collecting all violations and returning them together so API consumers get all the information in one round trip, e.g., "The following fields must not be changed: [_rev, type]." Not a blocker for this PR though.

shared-libs/cht-datasource/src/local/report.ts
```diff
@@ -134,88 +154,72 @@ export namespace v1 {
...
+      if (!isReport(originalReport)) {
+        throw new ResourceNotFoundError(`Report record [${updatedReport._id}] not found.`);
       }
-      validateReportUpdatePayload(originalReportDoc, reportInput);
-      await ensureFormFieldValidity(medicDb, reportInput);
 
-      return await updateReport(reportInput) as Report.v1.Report;
+      const contact = getContactForUpdate(originalReport, updatedReport, contactDoc);
+      if (originalReport.contact && !contact) {
+        throw new InvalidArgumentError('A contact is must be provided.');
```
sugat009:
```suggestion
        throw new InvalidArgumentError('A contact must be provided.');
```

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -95,132 +121,49 @@ export namespace v1 {
...
-        throw new InvalidArgumentError(`Person not found`);
+      const originalPerson = await getPerson(Qualifier.byUuid(updatedPerson._id));
+      if (!originalPerson) {
+        throw new ResourceNotFoundError(`Person record [${updatedPerson._id}] not found.`);
       }
-      if (personInput._rev !== originalDoc._rev) {
-        throw new InvalidArgumentError('`_rev` does not match');
+
+      assertFieldsUnchanged(originalPerson, updatedPerson, ['_rev', 'reported_date', 'type', 'contact_type']);
+      if (originalPerson.name) {
```
sugat009, line 161: issue: name is a required field for contacts. Rather than conditionally validating it, consider always asserting it's present. This simplifies the logic and prevents any edge case where an update could result in a nameless record

shared-libs/cht-datasource/src/local/place.ts
```diff
@@ -79,261 +149,78 @@ export namespace v1 {
...
+        updatedPlace._id,
+        getContactIdForUpdate(updatedPlace)
+      ]);
+      if (!isPlace(settings, originalPlace)) {
+        throw new ResourceNotFoundError(`Place record [${updatedPlace._id}] not found.`);
+      }
+
+      const contact = getContactForUpdate(originalPlace, updatedPlace, contactDoc);
+      assertFieldsUnchanged(originalPlace, updatedPlace, ['_rev', 'reported_date', 'type', 'contact_type']);
+      if (originalPlace.name !== updatedPlace.name) {
```
sugat009, line 213: issue: name is a required field for contacts. Rather than conditionally validating it, consider always asserting it's present. This simplifies the logic and prevents any edge case where an update could result in a nameless record

shared-libs/cht-datasource/src/local/place.ts
```diff
@@ -79,261 +149,78 @@ export namespace v1 {
...
-        } as unknown as Input.v1.PlaceInput;
-      }
-
-      input = await appendParent(typeFoundInSettingsContactTypes, input);
-      input = await appendContact(input);
-
-      return await createPlaceDoc(input) as Place.v1.Place;
+    const getMedicDocsByIds = getDocsByIds(medicDb);
+    const createMedicDoc = createDoc(medicDb);
+    const minifyLineage = minifyDoc(medicDb);
```
sugat009: suggestion: The local variable `minifyLineage` is assigned from the imported `minifyDoc` function, but `lineage.ts` also exports a separate `minifyLineage` function with different behavior. This naming collision makes it confusing to understand which function is actually being called. Consider renaming the local variable to match its source, e.g., `const minifyMedicDoc = minifyDoc(medicDb)` (which is already used in `place.ts:194` for the update path)

_Same comment for `local/person.ts:126` and `local/report.ts:160`_

shared-libs/cht-datasource/src/libs/core.ts
```diff
@@ -85,29 +110,68 @@ export const isRecord = (value: unknown): value is Record<string, unknown> => {
...
+/** @internal */
+// eslint-disable-next-line func-style
+export function assertHasRequiredField <T extends Record<string, unknown>, N extends string, K extends FieldType>(
   value: T,
-  fields: NonEmptyArray<FieldDescriptor<T>>,
-): boolean => fields.every(field => hasField(value, field));
+  { name, type }: FieldDescriptor<N, K>,
+  ErrorClass: new (message: string) => Error = Error
+): asserts value is T & Record<N, FieldTypeToValue[K]> {
+  if (!hasField(value, { name, type }) || !value[name]) {
```
sugat009: suggestion(non-blocking): `!value[name]` on line 171 is a `falsy` check, so it would reject legitimate values like `0` or `false`. Currently this is only used for string fields where rejecting empty strings is fine, but if this function is ever used for numeric or boolean fields, it would incorrectly reject valid values. Consider using `value[name] == null` instead to only reject null/undefined, or adding a separate empty string check for string fields.

shared-libs/cht-datasource/src/report.ts
```diff
@@ -140,47 +129,55 @@ export namespace v1 {
...
+     * @throws InvalidArgumentError if `contact` is not provided or is not a valid contact
+     * @throws InvalidArgumentError if any of the following read-only properties are changed: `reported_date`, `type`
      */
-    const curriedFn = async (updateInput: unknown): Promise<Report> => {
-      if (!isRecord(updateInput)) {
+    const curriedFn = async <T extends Report | ReportWithLineage>(
+      updated: Input.v1.UpdateReportInput<T>
+    ): Promise<T> => {
+      if (!isIdentifiable(updated)) {
         throw new InvalidArgumentError('Invalid report update input');
```
sugat009:
```suggestion
        throw new InvalidArgumentError('Updated report data not provided.');
```

sugat009: Change in error message for consistency with other update functions.

shared-libs/cht-datasource/src/local/place.ts
```diff
@@ -79,261 +149,78 @@ export namespace v1 {
...
+      assertFieldsUnchanged(originalPlace, updatedPlace, ['_rev', 'reported_date', 'type', 'contact_type']);
+      if (originalPlace.name !== updatedPlace.name) {
+        assertHasRequiredField(updatedPlace, { name: 'name', type: 'string' }, InvalidArgumentError);
+      }
+      assertSameParentLineage(originalPlace, updatedPlace);
+
+      const updatedPlaceDoc = {
+        ...updatedPlace,
+        contact
+      };
```
sugat009, line 221: question: Reports require contact on create (assertReportInput uses assertHasRequiredField) and prevent its removal on update (report.ts:211). Places treat contact as optional on create (assertPlaceInput uses assertHasOptionalField), but the base branch still treated it as immutable once set. The new code allows silently removing it. Is this intentional? If a place's primary contact shouldn't be removable once assigned, add a similar guard here.

shared-libs/cht-datasource/src/place.ts
sugat009:
```suggestion
     * @throws InvalidArgumentError if no type is provided or if the type is not for a place
```

shared-libs/cht-datasource/test/local/place.spec.ts
```diff
@@ -290,563 +293,744 @@ describe('local place', () => {
...
-      it('throws error if contact lineage does not match with actual doc when adding \
-        contact to a place that does not have one already', async () => {
-        const originalDoc = {
-          _id: '1',
-          _rev: '1',
-          name: 'myPlace',
-          type: 'contact',
-          contact_type: 'district_hospital',
-          reported_date: 12312312
+      it('updates place without contact', async () => {
```
sugat009, line 979: The test case "updates place without contact" only covers a place that never had a contact. There's no test case for a place that originally had a contact and then the contact is removed in an update. This is notable because report.ts explicitly guards against contact removal (line 211: if (originalReport.contact && !contact)) but place.ts does not have a similar guard. A test covering this transition would clarify whether removing an existing contact from a place is intentional behavior.

apoorvapendse:
Glad this is moving forward, thanks for reviving it!

api/src/controllers/person.js
jkuester: This is a good question. :thinking: I think the answer is "no".  Technically they are only getting back a slightly modified version of what they passed in.  I think the only "extra" data that is returned would be the `parent` lineage (for the provided `parent` uuid).  They cannot actually see data about any other persons (besides the one they just wrote). 

Returning the whole person is more of a convenience since we can do it for "free" (without an extra database hit).  But, we could always just return a an empty `200` if that seems better. :+1:

shared-libs/cht-datasource/src/local/libs/core.ts
jkuester: I like it!  Simple enough and a bit more convenient. Done!

shared-libs/cht-datasource/src/local/person.ts
jkuester: Very open to discussion here, but for a lot of these validations on the `update` flow I tried to be deliberately forgiving for issues that were _NOT_ introduced by the update.  We have run into this issue with other slimier update validations in the past.  Basically you can get a scenario where bad contact data exists (e.g. someone used a script and wrote contacts without a `name`).  Then you have another normal workflow that is trying to properly update these contacts (e.g. to set a `phone` or something) using cht-datasource.  If we just fail that update, because the contact has no `name`, then what is that workflow code supposed to do?  It passed a valid `phone` on a contact that it has retrieved from the db (because we also do not tightly validate data on retrieve).  That kind of setup would just totally externalize the whole contact validation process so any code that updates anything on an entity needs to validate that everything on that entity is totally valid.

Instead, I think the better approach (for the current state of our code-base) is to be very strict about everything _on create_ and also to be very strict _about the changed fields_ on update, but otherwise be lenient (allowing reading/updating entities even if they are not perfect).

IMHO we need this kind of approach for our `v1` apis. Otherwise it would get super risky to migrate any existing code to use this logic.  We can get a lot more interesting with validation in the `v2` apis....

shared-libs/cht-datasource/src/libs/core.ts
jkuester: Good call!  I have patched things up here to align with the `hasStringFieldWithValue` so we properly handle non-string values here as well as blank strings.

shared-libs/cht-datasource/src/local/place.ts
jkuester: Well spotted!  However, according to my understanding of the CHT contact model, a "primary" contact for a place (recorded in the `contact` field) is _not required_.  In fact, with the default config you can use the place create forms to create places without a default contact (as long as you do not select/create one as part of the form. Just choose "Skip this step"). This is why, on the place, I did not require the contact (and I allow you to remove one if it is set). 

On the other hand, for reports, the `contact` field actually records the contact of the _user_ who submitted the report (while `fields.patient_uuid/id` and `fields.place_uuid/id` are what record the actual contact the report is associated with).  This `contact` value is required for reports (which is why admin users with no contact cannot submit app forms).

shared-libs/cht-datasource/test/local/place.spec.ts
jkuester: Good call. As I noted above, I think we should be okay to remove the primary contact _from a place_.  I will add a test here to clarify this! :+1:

api/src/controllers/person.js
sugat009: Thanks for the clarification! The current approach aligns with standard REST/RBAC practices where CRUD operations are independent permissions. No changes needed here.

shared-libs/cht-datasource/src/local/person.ts
sugat009: That's a really good point. I wasn't considering the scenario of pre existing bad data in the wild. Being strict on create and lenient on update (for unchanged fields) makes sense for v1, especially for safely migrating existing code. Thanks for the detailed explanation!

shared-libs/cht-datasource/src/local/place.ts
sugat009: Thanks for the clarification on the semantic difference between place contact (primary contact for the facility, optional) and report contact (submitting user, required). That fully explains the asymmetry.

shared-libs/cht-datasource/src/place.ts
```diff
@@ -109,7 +110,7 @@ export namespace v1 {
      * Returns a generator for fetching all places with the given type
      * @param placeType the type of places to return
      * @returns a generator for fetching all places with the given type
-     * @throws InvaidArgumentError if no type is provided or if the type is not for a place
+     * @throws InvaidArgumentError if no type is provided or if the type is not a supported place contact type
```
sugat009:
```suggestion
     * @throws InvalidArgumentError if no type is provided or if the type is not a supported place contact type
```

Missed this one. 😅

sugat009 (approved):

Issue #10383, opened 2025-10-15
Selectively only index specific / necessary ddocs when upgrading to a new version


dianabarsan:
**Is your feature request related to a problem? Please describe.**
The first step when upgrading to a new version is to download the new ddocs from the builds server and pre-index them. Right now, the algorithm pre-indexes _all_ ddocs from _all_ databases, to make sure that when the new version of API and Sentinel are up, all endpoints are ready to go and functionality can resume seamlessly. 

However, in some cases, indexing only _some_ ddocs might be a good idea because: 
- it would allow us to push changes to auxiliary ddocs that use new features from CouchDb: eg: https://github.com/medic/cht-core/issues/10298#issuecomment-3389050668
- for instances with large disk space usage, indexing only necessary ddocs would allow the upgrade without needing to provision a lot of extra disk space

**Describe the solution you'd like**
Some ddocs are required for API to start up, like `medic/_design/medic`, but others are not. We should identify which ddocs are not essential for API starting and create a new install option that will only index the essential ddocs. The full install option should still be available for instances that cannot accept any endoint downtime. 

**Describe alternatives you've considered**
Somehow allow CHT admins to select which ddocs they want to index.

PR #10557, opened 2026-01-12
feat(#10383): add design document comparison support during upgrades

### LINKED ISSUE #10610: Show design document comparison on the admin upgrade page
A very frequent question when doing an upgrade is whether or not view indexing, and which view indexing is required for the specific upgrade. 

Add visual indicators of ddoc comparisons on the admin upgrade page. 

Image

Image

dianabarsan:
# Description

This pull request adds support for comparing design document changes during upgrades. This ensures that modifications to design documents are tracked and verified consistently throughout the upgrade process.

closes #10610

# Code review checklist

# AI disclosure

I used both Claude Code and JetBrains Junie to update unit tests as I made changes to the source code. 

# License

### DIFF
--- a/api/src/services/setup/utils.js
+++ b/api/src/services/setup/utils.js
@@ -184,6 +184,16 @@ const getDdocDefinitions = (buildInfo) => {
 };
 
 let localDdocDefinitionsCache = null;
+const remoteDdocDefinitionsCache = new Map();
+
+const deepCopy = (map) => {
+  // Return a deep copy to prevent mutation of cached ddocs
+  const copy = new Map();
+  for (const [key, value] of map) {
+    copy.set(key, value.map(ddoc => ({ ...ddoc })));
+  }
+  return copy;
+};
 /**
  * Returns map of bundled ddoc definitions for every database
  * @return {Map<Database, Array>}
@@ -196,42 +206,40 @@ const getLocalDdocDefinitions = async () => {
     }
     localDdocDefinitionsCache = ddocDefinitions;
   }
-
-  return new Map(localDdocDefinitionsCache);
+  return deepCopy(localDdocDefinitionsCache);
 };
 
-const remoteDdocDefinitionsCache = new Map();
 /**
- * Returns map of bundled ddoc definitions for every database, downloaded from the staging server
+ * Returns map of bundled ddoc definitions for every database for s specific version,
+ * downloaded from the staging server
  * @param {BuildInfo} buildInfo
  * @return {Map<Database, Array>}
  */
 const downloadDdocDefinitions = async (buildInfo) => {
-  if (remoteDdocDefinitionsCache.has(buildInfo.version)) {
-    return remoteDdocDefinitionsCache.get(buildInfo.version);
-  }
+  if (!remoteDdocDefinitionsCache.has(buildInfo.version)) {
 
-  const ddocDefinitions = new Map();
+    const ddocDefinitions = new Map();
 
-  const stagingDoc = await getStagingDoc(buildInfo);
+    const stagingDoc = await getStagingDoc(buildInfo);
 
-  // for simplicity, only ddocs for "known" databases are staged and indexed.
-  // for new databases, the final install will happen in the api preflight check.
-  // since any new database will be empty, the impact of not warming views is minimal.
-  for (const database of DATABASES) {
-    const attachment = stagingDoc._attachments[`ddocs/${database.jsonFileName}`];
-    // a missing attachment means that the database is dropped in this version.
-    // a migration should remove the unnecessary database.
-    if (attachment) {
-      const json = decodeAttachmentData(attachment.data);
-      ddocDefinitions.set(database, getDdocsFromJson(json));
-    } else {
-      logger.warn(`Attachment for ${database.jsonFileName} was not found. Skipping.`);
+    // for simplicity, only ddocs for "known" databases are staged and indexed.
+    // for new databases, the final install will happen in the api preflight check.
+    // since any new database will be empty, the impact of not warming views is minimal.
+    for (const database of DATABASES) {
+      const attachment = stagingDoc._attachments[`ddocs/${database.jsonFileName}`];
+      // a missing attachment means that the database is dropped in this version.
+      // a migration should remove the unnecessary database.
+      if (attachment) {
+        const json = decodeAttachmentData(attachment.data);
+        ddocDefinitions.set(database, getDdocsFromJson(json));
+      } else {
+        logger.warn(`Attachment for ${database.jsonFileName} was not found. Skipping.`);
+      }
     }
+    remoteDdocDefinitionsCache.set(buildInfo.version, ddocDefinitions);
   }
 
-  remoteDdocDefinitionsCache.set(buildInfo.version, ddocDefinitions);
-  return ddocDefinitions;
+  return deepCopy(remoteDdocDefinitionsCache.get(buildInfo.version));
 };
 
 const getDdocJsonContents = async (path) => {
--- a/api/tests/mocha/services/setup/utils.spec.js
+++ b/api/tests/mocha/services/setup/utils.spec.js
@@ -294,7 +294,9 @@ describe('Setup utils', () => {
         const result1 = await utils.getDdocDefinitions();
         const result2 = await utils.getDdocDefinitions();
 
-        expect(result1).to.equal(result2);
+        // Results are copies (not same reference) but structurally equal
+        expect(result1).to.not.equal(result2);
+        expect(result1).to.deep.equal(result2);
         expect(fs.promises.readFile.callCount).to.equal(5); // Once for each database
       });
 
@@ -447,7 +449,8 @@ describe('Setup utils', () => {
         const result1 = await utils.getDdocDefinitions(buildInfo(version));
         const result2 = await utils.getDdocDefinitions(buildInfo(version));
 
-        expect(result1).to.equal(result2);
+        expect(result1).to.deep.equal(result2);
+        expect(result1).to.not.equal(result2); // cached results are deep copied
         expect(db.builds.get.callCount).to.equal(1);
       });
 
@@ -471,7 +474,8 @@ describe('Setup utils', () => {
         const result1Cached = await utils.getDdocDefinitions(buildInfo(version1));
 
         expect(result1).to.not.equal(result2);
-        expect(result1).to.equal(result1Cached);
+        expect(result1).to.deep.equal(result1Cached);
+        expect(result1).to.not.equal(result1Cached); // cached results are deep copied
         expect(db.builds.get.callCount).to.equal(2);
       });
 
@@ -677,6 +681,44 @@ describe('Setup utils', () => {
       }
 
       expect(db.saveDocs.callCount).to.equal(3);
+    });
+
+    it('should not mutate the local ddoc definitions cache', async () => {
+      const genDdocsJson = (ddocs) => JSON.stringify({ docs: ddocs });
+      sinon.stub(fs.promises, 'readFile');
+      sinon.stub(resources, 'ddocsPath').value('localDdocs');
+
+      const medicDdocs = [
+        { _id: '_design/medic', views: { medic: {} } },
+        { _id: '_design/medic-client', views: { client: {} } },
+      ];
+      const sentinelDdocs = [{ _id: '_design/sentinel', views: { sentinel: {} } }];
+      const logsDdocs = [{ _id: '_design/logs', views: { logs: {} } }];
+      const usersMetaDdocs = [{ _id: '_design/meta', views: { meta: {} } }];
+      const usersDdocs = [{ _id: '_design/users', views: { users: {} } }];
+
+      fs.promises.readFile.withArgs('localDdocs/medic.json').resolves(genDdocsJson(medicDdocs));
+      fs.promises.readFile.withArgs('localDdocs/sentinel.json').resolves(genDdocsJson(sentinelDdocs));
+      fs.promises.readFile.withArgs('localDdocs/logs.json').resolves(genDdocsJson(logsDdocs));
+      fs.promises.readFile.withArgs('localDdocs/users-meta.json').resolves(genDdocsJson(usersMetaDdocs));
+      fs.promises.readFile.withArgs('localDdocs/users.json').resolves(genDdocsJson(usersDdocs));
+
+      const deployInfo = { user: 'admin', upgrade_log_id: 'theid' };
+      sinon.stub(upgradeLogService, 'getDeployInfo').resolves(deployInfo);
+      sinon.stub(db, 'saveDocs').resolves();
+
+      const ddocDefinitions = await utils.getDdocDefinitions();
+      await utils.saveStagedDdocs(ddocDefinitions);
+      const ddocDefinitionsAfter = await utils.getDdocDefinitions();
+
+      expect(ddocDefinitionsAfter.get(DATABASES[0])[0]._id).to.equal('_design/medic');
+      expect(ddocDefinitionsAfter.get(DATABASES[0])[1]._id).to.equal('_design/medic-client');
+      expect(ddocDefinitionsAfter.get(DATABASES[1])[0]._id).to.equal('_design/sentinel');
+      expect(ddocDefinitionsAfter.get(DATABASES[2])[0]._id).to.equal('_design/logs');
+      expect(ddocDefinitionsAfter.get(DATABASES[3])[0]._id).to.equal('_design/meta');
+      expect(ddocDefinitionsAfter.get(DATABASES[4])[0]._id).to.equal('_design/users');
+
+      expect(ddocDefinitionsAfter.get(DATABASES[0])[0].deploy_info).to.be.undefined;
     });
   });
 


dianabarsan:
Sample of ddoc difference: 
image

image

dianabarsan:
Updates per feedback: 

Details are closed by default:
image

Opened on click: 
image

For *Releases*, we display "free" upgrades directly in the list: 
image

api/resources/translations/messages-en.properties
```diff
@@ -794,6 +794,17 @@ instance.upgrade.error.deploy_info_fetch = Error fetching deploy info
 instance.upgrade.error.get_upgrade = Error fetching upgrading progress
 instance.upgrade.error.version_fetch = Error fetching available versions
 instance.upgrade.feature_releases = Feature Releases Betas
+instance.upgrade.indexing.db = Database
+instance.upgrade.indexing.ddoc = Design document
+instance.upgrade.indexing.not_required = The upgrade does not require indexing.
+instance.upgrade.indexing.required = The upgrade requires indexing. Click for details.
+instance.upgrade.indexing.not_required.short = No indexing.
+instance.upgrade.indexing.size = Current size on disk
```
mrjones-plip: nit:


```suggestion
instance.upgrade.indexing.size = Size
```

api/resources/translations/messages-en.properties
```diff
@@ -794,6 +794,17 @@ instance.upgrade.error.deploy_info_fetch = Error fetching deploy info
...
+instance.upgrade.indexing.size.warning = Warning: While indexes are building, they can take up to 10x more disk space. Make sure you have enough disk space before proceeding with the upgrade.
```
mrjones-plip: While I know that @jkuester said that a single view could take up to 10x space, I think it's too confusing to call this out.  Suggest we go with the [party line](https://docs.communityhealthtoolkit.org/hosting/cht/migration/preparing-for-5/#additional-disk-storage-required-for-upgrading) of 5x:

```suggestion
instance.upgrade.indexing.size.warning = Warning: Upgrades which involve view changes can use up to 5x more total disk space. Make sure you have enough disk space before proceeding with the upgrade.
```

mrjones-plip (changes requested):
Soo cool to see this coming together!  I think it'll be great to inform administrators that there's quick upgrades when no indexing is required.

* I noticed that all the calls to `/api/v2/upgrade/compare` are blocking, such that there's a window where clicking any buttons does nothing.  For me this is fast, but I fear on a slow server this might be a while.  Can we put a modal up "Please wait while upgrade information is gathered"?  I guess we should consider what happens with slow/poor/high packet loss connections too 🤔   
* There's a small regression in the Betas section - it now wraps the button "Install" button and the "Base version" heading isn't above the column values:
   Image
* I think for both `The upgrade does not require indexing.` and `Make sure you have enough disk space before proceeding with the upgrade..` we should add a link to the docs site which talks about what view changes mean. We'd talked about renaming the [troubleshooting page](https://docs.communityhealthtoolkit.org/hosting/cht/upgrade-troubleshooting/) and using that one.
* nit: maybe remove `_design/` from the ddoc column? Aren't all the views going to be in `_design/` such that we can just not show this and relay the same inf?

mrjones-plip:
@dianabarsan - we should prob grab another eng to do code review, I'm just focusing on UX

dianabarsan:
@mrjones-plip I believe I addressed all the issues, would be super useful if you could have another look. Once we're aligned on the UI changes, I will request a review from a developer as well,. 
Appreciate your time!

admin/src/templates/upgrade_confirm.html
```diff
@@ -8,6 +8,48 @@
...
+    </div>
+
+    <p>
+      <a href="https://docs.communityhealthtoolkit.org/hosting/cht/upgrade-troubleshooting/" target="_blank" rel="noopener noreferrer" translate>instance.upgrade.indexing.documentation</a>
+    </p>
+
+  </div>
   <p ng-if="model.stageOnly" translate>instance.stage.confirm.note</p>
-  <p ng-if="!model.stageOnly" translate>instance.upgrade.confirm.warning.text</p>
+  <p ng-if="!model.stageOnly" translate class="alert alert-info">instance.upgrade.confirm.warning.text</p>
```
mrjones-plip, line 54: nit: We have this:

> [Please check the documentation for more information about upgrades.](https://docs.communityhealthtoolkit.org/hosting/cht/upgrade-troubleshooting/)

I think a small rewording and less linking is better - but know this means more translation tokens:

> See [documentation](https://docs.communityhealthtoolkit.org/hosting/cht/upgrade-troubleshooting/) about upgrades and indexes.

admin/src/templates/upgrade_confirm.html
```diff
@@ -8,6 +8,48 @@
...
+            <td>
+                <span ng-repeat="type in difference.type">
+                  <span translate="instance.upgrade.indexing.type.{{type}}"></span>
+                </span>
+            </td>
+            <td>
+                <span>
+                  {{difference.size | bytes}}
+                </span>
+            </td>
```
mrjones-plip, line 38: When I go to upgrade to `5.0.1`, it says my `Size` is `-`.  Should that be `<1k` or something on my empty instance? 

Image

I smashed it with [TDG easy mode](https://docs.communityhealthtoolkit.org/community/contributing/code/test-data/test-data-generator/) (`node built/index.js ./sample-designs/easy-mode.js`) but the `Size` was still `-`.  But maybe that's cached now?

Here's what monitoring API shows (`curl -qsk https://localhost:10452/api/v2/monitoring | jq ".couchdb.medic"`) after TDG run:

```json
{
  "name": "medic",
  "update_sequence": 18488,
  "doc_count": 15262,
  "doc_del_count": 6,
  "fragmentation": 1.2696319965443015,
  "sizes": {
    "active": 15591394,
    "file": 18356864
  },
  "view_indexes": [
    {
      "name": "medic",
      "sizes": {
        "active": 14029803,
        "file": 21391803
      }
    },
    {
      "name": "medic-admin",
      "sizes": {
        "active": 0,
        "file": 1861420
      }
    },
    {
      "name": "medic-client",
      "sizes": {
        "active": 19271958,
        "file": 22327647
      }
    },
    {
      "name": "medic-conflicts",
      "sizes": {
        "active": 0,
        "file": 1844748
      }
    },
    {
      "name": "medic-sms",
      "sizes": {
        "active": 0,
        "file": 1849708
      }
    }
  ],
  "nouveau_indexes": [
    {
      "name": "medic/contacts_by_freetext",
      "doc_count": 9204,
      "file_size": 3560999
    },
    {
      "name": "medic/reports_by_freetext",
      "doc_count": 6047,
      "file_size": 3443036
    },
    {
      "name": "medic/docs_by_replication_key",
      "doc_count": 15397,
      "file_size": 3652623
    }
  ]
}

```

admin/src/templates/upgrade_confirm.html
dianabarsan: I would rather use less translation tokens.

mrjones-plip:
Thanks for working on my feedback! Some follow up questions after a round of testing the latest images:

1. It looks like you cache results for releases.  Makes sense!  It looks like any pre-release I test by clicking "Stage", I see an XHR request to `/api/v2/upgrade/compare`, so we're doing those ad-hoc.  Not sure if we should defend against a slow/lossy connection here with a spinner on the button?
2. How do I see a value larger than `< 1 kB`?  To test I spun up a new instance, did two rounds of TDG, and then went to the upgrade screen and still only saw  `< 1 kB` everywhere I checked
3. I was surprised that every version, including  `10383-ddoc-difference` (Jan 21, 2026 2:03:56 AM), requires indexing.  Is that correct?

dianabarsan:
> Not sure if we should defend against a slow/lossy connection here with a spinner on the button?

I don't think so. it's not an expensive request. And the whole admin app is online, every page makes requests to the server. so this is not a low-connectivity app scenario. 

> How do I see a value larger than < 1 kB?

You just should. This is what i have:
image

> I was surprised that every version, including 10383-ddoc-difference (Jan 21, 2026 2:03:56 AM), requires indexing. Is that correct?

I am unsure which version you are on really. Can you share an image of your upgrade page?

mrjones-plip:
OK - seeing very weird results.  My prior tests showed every version has having indexing and showed every index (was it 9 or so?) as needing indexing.  testing again now on what should be the same version, things all look normal and healthy.  

I need to retest - stay tuned and sorry for the confusion!

mrjones-plip:
Ok!  That was really confusing - glad we caught this.  

**reproduce**
1. launch a brand new docker helper instance
2. do not install latest `5.0.1` and choose the `10383-ddoc-difference` version
3. go to the local IP URL shown in terminal
4. login
5. go to `/admin/#/upgrade`

**Expected**:  under "Releases" the `5.0.1` and `5.0.0` items should have labels "No Indexing." .  Clicking  "stage" or "install" on a version should show correct contextual index info.

**Actual**: Every item under "Releases" no items in  tabular layout have  "No Indexing.".  Every item, when "stage" or "install" is clicked, shows all 10 ddocs as needing indexing, but always have a size of `< 1 kB`

**Workaround** - deleting your containers (`docker rm -f $(docker ps -qa)`) and restarting them with docker helper fixes the issue.

**screenshots**

<details><summary>admin section and version details</summary>
<p>

Screenshot From 2026-01-21 15-54-20

Screenshot From 2026-01-21 15-54-08

</p>
</details> 

**logs**

<details><summary>from API container</summary>
<p>

```
2026-01-22T01:09:06.020 RES: b0bbad6df8ae 192.168.68.26 - POST /api/v2/upgrade/compare HTTP/1.0 200 779 509.002 ms
2026-01-22T01:09:06.026 REQ: a03fadca6838 192.168.68.26 - POST /api/v2/upgrade/compare HTTP/1.0
2026-01-22T01:09:06.535 ERROR: Error fetching view index info: Error: 404 - {"error":"not_found","reason":"deleted"}
    at request (/service/shared-libs/couch-request/src/couch-request.js:239:15)
    at async Object.getDdocInfo (/service/api/src/services/setup/utils.js:390:22)
    at async Object.compareBuildVersions (/service/api/src/services/setup/upgrade.js:142:28)
    at async compareUpgrade (/service/api/src/controllers/upgrade.js:81:21) {
  [stack]: [Getter/Setter],
  [message]: '404 - {"error":"not_found","reason":"deleted"}',
  status: 404,
  statusCode: 404,
  body: { error: 'not_found', reason: 'deleted' },
  headers: Headers {
    'cache-control': 'must-revalidate',
    'content-length': '41',
    'content-type': 'application/json',
    date: 'Thu, 22 Jan 2026 01:09:05 GMT',
    server: 'CouchDB/3.5.0 (Erlang OTP/26)',
    'x-couch-request-id': 'e3c68f86b7',
    'x-couchdb-body-time': '0',
    'keep-alive': 'timeout=18000'
  },
  ok: false
} 
```

</p>
</details>

dianabarsan:
> per above comment!

I don't think your issue has anything to do with this work, unfortunately. I think it was some some weird image caching going on, and possibly a failed upgrade at some point.

dianabarsan:
If everything was ok after container deletion and restart, I don't think there's anything that can be done in this PR.

dianabarsan:
Hi @witash ! 
I would really appreciate it if you would have a look at this PR. Thanks for your time!

mrjones-plip:
@dianabarsan - I'd like to gently push back on the error I found being endemic to my setup.  I followed [my steps](https://github.com/medic/cht-core/pull/10557#issuecomment-3781958140) on a 2nd computer and was able to reproduce.  This is not an upgraded instance, both were clean installs.   

I did a bit of exploring and found the issue appears to be that `:staged:` is inserted in the couch URLs causing `404`s. After restart it's not there and all the calls work as expected.   Looking in the code the `setStagingData()` function in `api/src/services/setup/utils.js` seems to [add this](https://github.com/medic/cht-core/blob/40d8b5571f24f08544e9b2cbac61ca9e1958c21d/api/src/services/setup/utils.js#L72)?

-----------------

Questions I'm wondering about:
* Why would a call to `/api/v2/upgrade/compare` return that every index needs rebuilding when there's an error?  Maybe we could be more defensive in the code and note to the user that we were unable to fetch the data?
* when you go through [the steps](https://github.com/medic/cht-core/pull/10557#issuecomment-3781958140), are you able to reproduce? Should only take a few min :sweat_smile: 
* Why is couch reporting to API `Error: 404 - {"error":"not_found","reason":"deleted"}`  in a new install? This breaks even before any single user has ever logged or any data has been added.
* I did some more testing just now, and `kill` and `up`ing the containers also fixes the issue.  What is special about the restart/delete that fixes the issue?


-----------------


To validate my above questions, I isolated calls to a clean instance. I used `curl` to make `POST` - replace `192-168-68-26.local-ip.medicmobile.org:10459` with the URL of your instance:

```
curl 'https://medic:password@192-168-68-26.local-ip.medicmobile.org:10459/api/v2/upgrade/compare' \
  -X POST \
  -H 'Accept: application/json' \
  -H 'Content-Type: application/json;charset=utf-8' \
  --data-raw '{"build":{"base_version":"4.4.0","application":"medic","schema_version":2,"namespace":"medic","time":"2023-09-21T01:08:31.713Z","build":"4.4.0.6255863600","version":"4.4.0","author":"npm on runner"}}'
```

This returns every index, which is not correct:

```
[{"type":["added"],"ddoc":"_design/medic","db":"medic","indexing":true},{"type":["added"],"ddoc":"_design/medic-admin","db":"medic","indexing":true},{"type":["added"],"ddoc":"_design/medic-client","db":"medic","indexing":true},{"type":["added"],"ddoc":"_design/medic-conflicts","db":"medic","indexing":true},{"type":["added"],"ddoc":"_design/medic-scripts","db":"medic","indexing":true},{"type":["added"],"ddoc":"_design/medic-sms","db":"medic","indexing":true},{"type":["added"],"ddoc":"_design/sentinel","db":"medic-sentinel","indexing":true},{"type":["added"],"ddoc":"_design/logs","db":"medic-logs","indexing":true},{"type":["added"],"ddoc":"_design/users-meta","db":"medic-users-meta","indexing":true}]
```

And this is the full couch logs, on top of the [already covered `404` API](https://github.com/medic/cht-core/pull/10557#issuecomment-3781958140) error:

```
[notice] 2026-01-22T18:55:58.541579Z couchdb@127.0.0.1 <0.25471.0> 63f4b7422a haproxy:5984 172.22.0.6 medic GET //medic/_design/:staged:medic/_info 404 ok 1
...CUT FOR BREVITY...
[notice] 2026-01-22T18:55:58.564524Z couchdb@127.0.0.1 <0.25471.0> 9572fa0645 haproxy:5984 172.22.0.6 medic GET //_users/_design/:staged:users/_info 404 ok 0
```

If we then make a single `curl` call to the IP and port of couch, we can see the 404 directly:

```
curl http://medic:password@172.22.0.5:5984//medic/_design/:staged:medic/_info
{"error":"not_found","reason":"deleted"}
```

And Couch reports the same in it's logs:

```
[notice] 2026-01-22T18:59:14.932349Z couchdb@127.0.0.1 <0.34021.0> 742a5bb55e 172.22.0.5:5984 172.22.0.1 medic GET //medic/_design/:staged:medic/_info 404 ok 1
```

However, when I restart, I see now that `:staged:` is no longer present in the URL and everything just works. here's the two curl calls, now both succeeding:

```
curl 'https://medic:password@192-168-68-26.local-ip.medicmobile.org:10461/api/v2/upgrade/compare' \
  -X POST \
  -H 'Accept: application/json' \
  -H 'Content-Type: application/json;charset=utf-8' \
  --data-raw '{"build":{"base_version":"4.4.0","application":"medic","schema_version":2,"namespace":"medic","time":"2023-09-21T01:08:31.713Z","build":"4.4.0.6255863600","version":"4.4.0","author":"npm on runner"}}'

[{"type":["views","indexes"],"ddoc":"_design/medic","db":"medic","size":110922,"indexing":true},{"type":["views"],"ddoc":"_design/medic-client","db":"medic","size":105156,"indexing":true},{"type":["views"],"ddoc":"_design/medic-scripts","db":"medic","size":612,"indexing":true},{"type":["views"],"ddoc":"_design/users-meta","db":"medic-users-meta","size":13098,"indexing":true}]

curl http://medic:password@172.22.0.5:5984//medic/_design/medic-sms/_info   

{"name":"medic-sms","view_index":{"updates_pending":{"minimum":0,"preferred":0,"total":0},"waiting_commit":false,"waiting_clients":0,"updater_running":false,"update_seq":191,"sizes":{"file":84296,"external":96,"active":0},"signature":"60dd4e6dabdc3c27b3daa8a9308676a7","purge_seq":0,"language":"javascript","compact_running":false,"collator_versions":["153.120"]}}
```

dianabarsan:
Something is strange in your setup, nothing should be calling :staged: docs in any of this work. I'll take it step by step: 

```
curl 'https://medic:password@192-168-68-26.local-ip.medicmobile.org:10459/api/v2/upgrade/compare' \
  -X POST \
  -H 'Accept: application/json' \
  -H 'Content-Type: application/json;charset=utf-8' \
  --data-raw '{"build":{"base_version":"4.4.0","application":"medic","schema_version":2,"namespace":"medic","time":"2023-09-21T01:08:31.713Z","build":"4.4.0.6255863600","version":"4.4.0","author":"npm on runner"}}'
```

It seems that you are comparing to a `4.4.0` build, which **should** return all views to need indexing, because all of them received updates from `4.4.0` to current. Which version did you deploy?

dianabarsan:
> However, when I restart, I see now that :staged: is no longer present in the URL 

I'm trying to understand which URL you are referring to here. 
The :staged: docs should not be present in a new install.

I'll go over your steps to verify.

dianabarsan:
I replicated the issue on the first try! Thanks for keeping me honest!

dianabarsan:
The issue was, you guessed it, with cache :D 

I was caching local ddoc definitions, but those were also called when the new install was saved, which ended up mutating my cache with adding the :staged: prefix. The problem disappeared on restart because the "install" code doesn't run on every restart, only when the version changes. 

I fixed it by always sending a deep copy of the cached result for other code to play with.

dianabarsan:
I fixed the caching issue and now fresh installs show correct data. Thanks for your patience @mrjones-plip ! 
Also, I needed to do a docker image cleanup to have fresh cht images downloaded when I reinstalled with docker helper :( and I just spun up a new project, it just used previous image tags.

dianabarsan:
Slight nudge @witash . I would love to ship this in 5.1. Would you mind the review? Appreciate it!

PR #10570, opened 2026-01-16
feat(#10509): support uploading attachments in contact forms

### LINKED ISSUE #10509: Support uploading attachments in `contact` forms
## Is your feature request related to a problem? Please describe.
`app` forms support uploading files when filling out the form (via the various supported ODK fields for attaching files/media/etc).  When the app form is completed, the files are included as attachments in the report doc.

Unfortunately, uploading attachments like this is not supported for contact forms. It seems that you can include the questions in your form, but the selected files are not actually stored as attachments to the contact doc when the form is submitted.  (They are just ignored.)  

## Describe the solution you'd like

We should support including any uploaded files as attachments to the contact document (as is done for app forms). 

### In Scope
* allowing uploads on contact forms
* uses existing files on the device
* allowing uploads when creating a contact
* allowing uploads when editing a contact 
* upload any file, pdf, image etc

### Out of Scope
* allowing uploads on sub-documents - eg the contact (sub-document) that can optionally get created when creating a household (primary document).  Work around is navigate directly to that contact after it's been created, an add the upload there if needed
* updating the the contact summary - no changes here. Upload thumbnails or file names or links will not be shown
* Allowing/prompting the user to take a photo instead of picking an existing file

## Additional context

[Forum thread](https://forum.communityhealthtoolkit.org/t/uploading-and-displaying-images-in-contact-forms/2024/8).

One additional feature we would likely need once this is implemented is some way to actually display images (and maybe other types of files 🤔) for a contact once they have been attached.  For reports, we have the [binary attachment tag](https://docs.communityhealthtoolkit.org/building/forms/app/#uploading-binary-attachments) that causes images to be rendered on the reports page.  Perhaps we could support a contact-summary [card or field](https://docs.communityhealthtoolkit.org/building/contact-summary/contact-summary-templated/#contact-summary) that would render an images....  This probably deserves its own issue if we want to pursue it, but it will depend on this one to just get the initial files attached to the contact doc.

cliftonmcintosh:
# Description

This PR adds attachment support to contact forms, enabling file uploads and binary field attachments similar to what already exists for report forms.

## What's Implemented

Following discussion in #10509, the implementation was **simplified to match report form behavior**: all attachments attach to the main contact document only, regardless of form structure.

### Backend Changes

1. **File Widget Attachments**: Files uploaded through Enketo file widgets are extracted from `FileManager` and attached to the contact document
2. **Binary Field Attachments**: Base64-encoded binary fields (signatures, drawings) are extracted from XML and attached with XPath-based naming
3. **Attachment Naming**: Uses consistent patterns matching report forms:
   - File widgets: `user-file-{filename}`
   - Binary fields: `user-file/{form-id}/{xpath}`
4. **Edit Preservation**: Attachments automatically persist when editing contacts via `_defaults()` merge

### Implementation Details

- **Location**: `ContactSaveService.prepareSubmittedDocsForSave()` (following [@jkuester's guidance](https://github.com/medic/cht-core/issues/10509#issuecomment-2542828754))
- **Code Added**: ~30 lines in `processAllAttachments()` method
- **Services Reused**: `AttachmentService`, `FileManager`, `Xpath` utility
- **Pattern**: Mirrors `EnketoService.xmlToDocs()` for report forms

## Test Coverage

| Test Type | Count | Location |
|-----------|-------|----------|
| Unit Tests | 8 | `webapp/tests/karma/ts/services/contact-save.service.spec.ts` |
| cht-form Integration | 2 | `tests/integration/cht-form/default/contact-with-attachments.wdio-spec.js` |
| E2E Tests | 5 | `tests/e2e/default/contacts/contact-attachments.wdio-spec.js` |

**Unit tests are intended to verify:**
- File widget extraction and attachment to main document
- Binary field extraction with XPath-based naming
- Multiple attachments (mixed types) all attach to main document

**E2E tests are intended to verify:**
- Create contact with single image attachment
- Create contact with multiple attachments (image + document)
- Edit contact preserves existing attachments

All added tests passing ✅

## Manual Testing Performed

- Created contacts with photo attachments via person-create form
- Verified attachments saved to CouchDB with correct naming (`user-file-{filename}`)
- Verified edit form displays existing attachment filename
- Confirmed multiple contacts can have attachments independently

<details>

<summary>Details about manual testing</summary>

## Step 1: Modified the Contact Form

Please note that form modifications have not been committed. The file was modified locally.

**File:** `config/default/forms/contact/person-create.xml`

### 1a. Added field to instance data (~line 922)

```xml
<notes/>
<photo/>
<user_for_contact>
```

### 1b. Added binding (~line 1062)

```xml
<bind nodeset="/data/person/notes" type="string"/>
<bind nodeset="/data/person/photo" type="binary"/>
<bind nodeset="/data/person/user_for_contact/create" ...
```

### 1c. Added upload widget to body (~line 1159)

```xml
<input ref="/data/person/notes" appearance="multiline">
  <label ref="jr:itext('/data/person/notes:label')"/>
</input>
<upload ref="/data/person/photo" mediatype="image/*">
  <label>Photo</label>
  <hint>Take a photo or select from gallery</hint>
</upload>
<group ref="/data/person/user_for_contact"/>
```

---

### 2a. Compiled the form

```bash
./scripts/build/build-config.sh
```

### 2b. Uploaded to CouchDB

Deleted the existing form and posted the new version:

```bash
# Got current revision and deleted
REV=$(curl -s http://medic:password@localhost:5984/medic/form%3Acontact%3Aperson%3Acreate | python3 -c 'import sys,json; print(json.load(sys.stdin)["_rev"])')
curl -X DELETE "http://medic:password@localhost:5984/medic/form%3Acontact%3Aperson%3Acreate?rev=$REV"

# Uploaded new version
curl -X POST "http://medic:password@localhost:5984/medic" \
  -H "Content-Type: application/json" \
  -d @api/build/default-docs/form-contact-person-create.doc.json
```

### 2c. Refreshed the browser

Hard refresh with `Cmd+Shift+R`.

## Step 3: Tested the Upload

1. Navigated to a health facility in CHT
2. Clicked "New person" to create a contact
3. Filled out required fields (name, DOB, sex)
4. Scrolled to the "Photo" field and uploaded an image
5. Submitted the form

## Step 4: Verified the Attachment Was Saved

Queried CouchDB directly:

```bash
curl -s "http://medic:password@localhost:5984/medic/_find" \
  -H "Content-Type: application/json" \
  -d '{"selector": {"type": "person"}, "limit": 5}' | python3 -m json.tool | head -50
```

Looked for:
- `"photo": "filename.png"` - the field value
- `"_attachments": { "user-file-filename.png": {...} }` - the stored attachment

## Test Result

**Contact created:** "Photo Contact"

**Attachment saved:**

```json
"photo": "Screenshot 2026-01-26 at 09.06.22-9_30_5.png"

"_attachments": {
  "user-file-Screenshot 2026-01-26 at 09.06.22-9_30_5.png": {
    "content_type": "image/png",
    "length": 413623
  }
}
```

---

# Testing the Edit Form (Display Existing Attachment)

## Step 1: Modified the Edit Form

Please note that form modifications have not been committed. The file was modified locally.

**File:** `config/default/forms/contact/person-edit.xml`

### 1a. Added field to instance data (~line 1013)

```xml
<notes/>
<photo/>
<meta tag="hidden">
```

### 1b. Added binding (~line 1166)

```xml
<bind nodeset="/data/person/notes" type="string"/>
<bind nodeset="/data/person/photo" type="string" readonly="true()" relevant="../photo != ''"/>
<bind nodeset="/data/person/meta/created_by" type="string" ...
```

**Note:** The binding used:
- `type="string"` (not binary) because we were displaying the filename, not uploading
- `readonly="true()"` to prevent editing
- `relevant="../photo != ''"` to only show when a photo exists

### 1c. Added read-only input to body (~line 1284)

```xml
</input>
<input ref="/data/person/photo">
  <label>Current Photo</label>
  <hint>Existing attachment (read-only)</hint>
</input>
<group ref="/data/person/user_for_contact"/>
```

### 2a. Compiled the form

```bash
./scripts/build/build-config.sh
```

### 2b. Uploaded to CouchDB

Deleted the existing form and posted the new version:

```bash
# Got current revision and deleted
REV=$(curl -s http://medic:password@localhost:5984/medic/form%3Acontact%3Aperson%3Aedit | python3 -c 'import sys,json; print(json.load(sys.stdin)["_rev"])')
curl -X DELETE "http://medic:password@localhost:5984/medic/form%3Acontact%3Aperson%3Aedit?rev=$REV"

# Uploaded new version
curl -X POST "http://medic:password@localhost:5984/medic" \
  -H "Content-Type: application/json" \
  -d @api/build/default-docs/form-contact-person-edit.doc.json
```

### 2c. Refreshed the browser

Hard refresh with `Cmd+Shift+R`.

## Step 3: Tested Viewing the Existing Attachment

1. Navigated to "Photo Contact" (the contact created in Part 1)
2. Clicked "Edit" to open the edit form
3. Scrolled to the "Current Photo" field
4. Verified the filename was displayed

## Test Result

**Contact edited:** "Photo Contact"

**Photo field displayed:** ✅ Yes - filename shown as read-only

**Filename shown:** `Screenshot 2026-01-26 at 09.06.22-9_30_5.png`

**Status:** ✅ Existing attachment filename correctly displayed in edit form

</details>

<details>

<summary>Screenshots</summary>

### Empty person create form with photo field
create-form-empty-with-photo-field

### Person create form with photo added
create-form-with-added-photo

### Person edit form with photo shown
edit-form-with-photo-attached

</details>

--- 

## Not in Scope for V1

Per discussion in #10509, these are deferred:

- ~~**UI Preview on Edit**: Attachments persist when editing, but users cannot see previews of existing attachments (can be added in V2)~~ preview on edit is included as per @dianabarsan's request in https://github.com/medic/cht-core/pull/10570/#discussion_r2731500513
- **Multi-Document Routing**: Attachments in sibling/repeat sections all attach to main document (matches report behavior)
- **Contact-Summary Integration**: Per issue description, this is out of scope

## AI Disclosure

**AI Assistance**: This PR was developed using Claude Code.

**How AI was used**:
- Analysis of existing codebase patterns (EnketoService, AttachmentService) for consistency
- Code generation for the `processAllAttachments()` method
- Test writing following TDD methodology
- Documentation and planning

**Human oversight**:
- I reviewed all generated code for correctness, security, and consistency with project patterns
- I ran all tests manually to verify functionality
- I made architecture decisions (V1 simplified approach based on reviewer feedback)
- I iterated on implementation based on code review feedback
- All commits include co-author attribution

Closes #10509

# Code review checklist

- [x] UI/UX backwards compatible: No UI changes in this PR. Existing contact forms continue to work unchanged.
- [ ] Documented: Configuration and user documentation on [cht-docs](https://github.com/medic/cht-docs/) - Form authors will need guidance on adding upload fields
- [x] Tested: Unit tests, cht-form integration tests, and E2E tests all passing
- [x] Internationalised: No user-facing text added
- [x] Backwards compatible: Works with existing data and configuration. Existing contact forms without upload fields are unaffected.

# License

### DIFF
--- a/webapp/src/ts/services/contact-save.service.ts
+++ b/webapp/src/ts/services/contact-save.service.ts
@@ -99,16 +99,17 @@ export class ContactSaveService {
    *          the stem becomes empty after sanitization
    */
   private sanitizeFileName(fileName: string): string {
+    const disallowedChars = /[^a-zA-Z0-9_.-]/g;
     const lastDotIndex = fileName.lastIndexOf('.');
     const hasExtension = lastDotIndex > 0;
 
     if (!hasExtension) {
-      return fileName.replace(/[^a-zA-Z0-9_.-]/g, '') || uuidV4(); // NOSONAR
+      return fileName.replace(disallowedChars, '') || uuidV4(); // NOSONAR
     }
 
     const stem = fileName.slice(0, lastDotIndex);
     const extension = fileName.slice(lastDotIndex);
-    const sanitizedStem = stem.replace(/[^a-zA-Z0-9_.-]/g, ''); // NOSONAR
+    const sanitizedStem = stem.replace(disallowedChars, ''); // NOSONAR
     return (sanitizedStem || uuidV4()) + extension;
   }
 


cliftonmcintosh:
@dianabarsan I've created a draft PR with some initial work so that I can get feedback and see if this is the right direction to go.

webapp/src/ts/services/contact-save.service.ts
```diff
@@ -69,13 +77,150 @@ export class ContactSaveService {
...
+
+    // Flatten all repeat groups into a single array
+    // Real contact forms use wrapper elements like <repeat> and <other-women-repeat>
+    // Each wrapper contains multiple instances of the same element type (e.g., all <child> elements)
+    const allRepeats: any[] = [];
+    Object.values(repeated).forEach((repeatGroup: any) => {
+      if (Array.isArray(repeatGroup)) {
+        allRepeats.push(...repeatGroup);
+      }
+    });
```
cliftonmcintosh: Is the flattening here a concern? I have a test, "should handle multiple repeat wrapper elements with flattening", intended to check that after the flattening, the XML order is preserved. I would appreciate knowing if that test seems valid.

webapp/tests/karma/ts/services/contact-save.service.spec.ts
```diff
@@ -225,4 +233,417 @@ describe('ContactSave service', () => {
...
+        '<child>' +
+        '<name>Alice Smith</name>' +
+        '<photo type="binary">AAAA</photo>' +
+        '</child>' +
+        '<child>' +
+        '<name>Bob Smith</name>' +
+        '<photo type="binary">BBBB</photo>' +
+        '</child>' +
+        '</child_data>' +
+        '</data>';
```
cliftonmcintosh: Is the XML structure here realistic?

webapp/tests/karma/ts/services/contact-save.service.spec.ts
```diff
@@ -225,4 +233,417 @@ describe('ContactSave service', () => {
...
+      // ContactSaveService creates three documents: household (main) and two children (repeats)
+      enketoTranslationService.contactRecordToJs.returns({
+        doc: { _id: 'household1', type: 'household', name: 'Smith Family' },
+        repeats: {
+          child_data: [
+            { _id: 'child1', type: 'person', name: 'Alice Smith', parent: 'PARENT' },
+            { _id: 'child2', type: 'person', name: 'Bob Smith', parent: 'PARENT' }
+          ]
+        }
+      });
```
cliftonmcintosh: Is this mock return accurate?

webapp/tests/karma/ts/services/contact-save.service.spec.ts
```diff
@@ -225,4 +233,417 @@ describe('ContactSave service', () => {
...
+        '<service>' +
+        '<name>Vaccination Service</name>' +
+        '<certificate type="binary">CERT_1</certificate>' +
+        '</service>' +
+        '<service>' +
+        '<name>Laboratory Service</name>' +
+        '<certificate type="binary">CERT_2</certificate>' +
+        '</service>' +
+        '</repeat>' +
+        '</data>';
```
cliftonmcintosh: Is the XML structure here realistic?

webapp/tests/karma/ts/services/contact-save.service.spec.ts
```diff
@@ -225,4 +233,417 @@ describe('ContactSave service', () => {
...
+        '<child>' +
+        '<name>Second Child</name>' +
+        '<photo type="binary">PHOTO_2</photo>' +
+        '</child>' +
+        '</repeat>' +
+        '<contact>' +
+        '<name>Contact Person</name>' +
+        '<signature type="binary">SIGNATURE</signature>' +
+        '</contact>' +
+        '</data>';
```
cliftonmcintosh: Is the XML structure here realistic?

webapp/tests/karma/ts/services/contact-save.service.spec.ts
```diff
@@ -225,4 +233,417 @@ describe('ContactSave service', () => {
...
+        siblings: {
+          contact: { _id: 'person1', type: 'person', name: 'Dr. Jane Wilson', parent: 'PARENT' }
+        },
+        repeats: {
+          service_data: [
+            { _id: 'service1', type: 'service', name: 'Vaccination Service', parent: 'PARENT' },
+            { _id: 'service2', type: 'service', name: 'Laboratory Service', parent: 'PARENT' }
+          ]
+        }
+      });
```
cliftonmcintosh: Is this mock return accurate?

webapp/tests/karma/ts/services/contact-save.service.spec.ts
```diff
@@ -225,4 +233,417 @@ describe('ContactSave service', () => {
...
+        siblings: {
+          contact: { _id: 'contact1', type: 'person', name: 'Contact Person', parent: 'PARENT' }
+        },
+        repeats: {
+          child_data: [
+            { _id: 'child1', type: 'person', name: 'First Child', parent: 'PARENT' },
+            { _id: 'child2', type: 'person', name: 'Second Child', parent: 'PARENT' }
+          ]
+        }
+      });
```
cliftonmcintosh: Is this mock return accurate?

webapp/tests/karma/ts/services/contact-save.service.spec.ts
```diff
@@ -225,4 +233,450 @@ describe('ContactSave service', () => {
...
+        '<women>' +
+        '<name>Second Woman</name>' +
+        '<cert type="binary">CERT_2</cert>' +
+        '</women>' +
+        '</other-women-repeat>' +
+        '<contact>' +
+        '<name>Contact Person</name>' +
+        '<signature type="binary">SIGNATURE</signature>' +
+        '</contact>' +
+        '</data>';
```
cliftonmcintosh: Is the XML structure here realistic?

webapp/tests/karma/ts/services/contact-save.service.spec.ts
```diff
@@ -225,4 +233,450 @@ describe('ContactSave service', () => {
...
+          child_data: [
+            { _id: 'child1', type: 'person', name: 'First Child', parent: 'PARENT' },
+            { _id: 'child2', type: 'person', name: 'Second Child', parent: 'PARENT' }
+          ],
+          women_data: [
+            { _id: 'woman1', type: 'person', name: 'First Woman', parent: 'PARENT' },
+            { _id: 'woman2', type: 'person', name: 'Second Woman', parent: 'PARENT' }
+          ]
+        }
+      });
```
cliftonmcintosh: Is this mock return accurate?

webapp/src/ts/services/contact-save.service.ts
dianabarsan: I would avoid flattening here. Because you can have a repeat within a repeat? I don't know if that works really. I think we're basic this code on an assumption. Safest would be to remove flattening.

webapp/src/ts/services/contact-save.service.ts
```diff
@@ -69,13 +77,150 @@ export class ContactSaveService {
...
+            (xpath.startsWith('/' + formId) ? xpath : xpath.replace(/^\/[^/]+/, '/' + formId));
+
+          // Find which document this attachment belongs to based on XPath
+          const targetDoc = this.findTargetDocument(xpath, docBoundaries, preparedDocs);
+          this.attachmentService.add(targetDoc, filename, content, 'image/png', true);
+        }
+      });
+  }
+
+  private buildDocumentBoundaries(
```
dianabarsan: Wow! this function! 

I think the existence of this function is because the code is built to allow attachments for _any_ of the documents the contact form creates. However, for reports, [we only attach binary files for the main document](https://github.com/medic/cht-core/blob/69018ea93ca5ce5bead925afa81852e595e7f188/webapp/src/ts/services/enketo.service.ts#L491). I think that following this model should be sufficient for the first iteration of this work: following the report format and only allowing attachments for the main document.

webapp/src/ts/services/contact-save.service.ts
```diff
@@ -69,13 +77,150 @@ export class ContactSaveService {
...
+        const content = $element.text();
+        if (content) {
+          const xpath = Xpath.getElementXPath(element);
+          // Replace instance root element node name with form internal ID
+          const filename = 'user-file' +
+            (xpath.startsWith('/' + formId) ? xpath : xpath.replace(/^\/[^/]+/, '/' + formId));
+
+          // Find which document this attachment belongs to based on XPath
+          const targetDoc = this.findTargetDocument(xpath, docBoundaries, preparedDocs);
+          this.attachmentService.add(targetDoc, filename, content, 'image/png', true);
```
dianabarsan: I believe there's an error here where the mime type is hardcoded to png. 
Please check EnketoService [to see how we attach files there](https://github.com/medic/cht-core/blob/69018ea93ca5ce5bead925afa81852e595e7f188/webapp/src/ts/services/enketo.service.ts#L484)

webapp/src/ts/services/contact-save.service.ts
cliftonmcintosh: This attempt to allow attachments for any of the documents the contact form creates us because in a comment on [the issue](https://github.com/medic/cht-core/issues/10509), @jkuester said

> The tricky part is going to be handling the fact that contact forms can create multiple documents (primary doc, sibling docs, repeated child docs), and ensuring that attachments are mapped to the correct document. I think, for each file attachment we will need to look up where in the XML structure it came from, and then map that to the corresponding document. Because of this, I am not sure if there is anything we can easily reuse from xmlToDocs directly, but we can at least borrow the logic for processing FileManager files and binary fields to use in our logic that can be added to ContactSaveService.prepareSubmittedDocsForSave.

webapp/src/ts/services/contact-save.service.ts
cliftonmcintosh: Thanks for pointing that out. I believe this matches Enketo Service's approach.

Enketo Service handles the two file types in two different ways. 
1. For files that are retrieved by `FileManager.getCurrentFiles()`, it uses `file.type` for the mime type (as you noted in your link. 
2. For files that it finds via `$record.find('[type=binary]')` [it hardcodes `image/png`](https://github.com/medic/cht-core/blob/69018ea93ca5ce5bead925afa81852e595e7f188/webapp/src/ts/services/enketo.service.ts#L494-L504). I followed that pattern here.

In this update, that pattern is also being followed.
1. For files that are retrieved by `FileManager.getCurrentFiles()`, we are following the same thing that Enketo service does. See [here](https://github.com/cliftonmcintosh/cht-core/blob/ff040478e8a7edc77291ef57cfa355b9fba3d34f/webapp/src/ts/services/contact-save.service.ts#L105-L109).
2. But for the ones we get by checking `$record.find('[type=binary]')`, we are hardcoding to match what Enketo Service is doing.

webapp/src/ts/services/contact-save.service.ts
cliftonmcintosh: The flattening is intended to handle different types of repeats, not repeats within repeats. Since the `repeatsToJs` method in the Enketo Translation Servcice returns objects with keys like `child_data` and `service_data`, this is intended to be able to process more than just `child_data` repeats.

But if `child_data` is the only type of repeat we care about for supporting uploaded attachments, then I think we can remove this.

webapp/src/ts/services/contact-save.service.ts
cliftonmcintosh: What I think would be helpful is guidance on
1. _What the XML looks like for contact forms._ What elements do they have? Are there a limited set of objects that can show up as the primary, the sibling or the repeated child docs? Which ones can have attachments? 
2. _Which documents we should support attachments for._ Are we limiting it to the primary like @dianabarsan is suggesting we do? Or do we need or want to support attachments on the sibling and child docs that @jkuester mentions are possible?

dianabarsan: I asked Josh to clarify this comment about attachments to child docs, we are in a bit of a disagreement and my preference would be to keep things simple. 

For sample contact forms, you can always look at the [default config](https://github.com/medic/cht-core/tree/master/config/default/forms/contact) contact forms, though these would not have child documents. I will search for a real-world example that does have child documents.

webapp/src/ts/services/contact-save.service.ts
dianabarsan: Sweet, thanks for pointing that out! Appreciate your diligence!

webapp/src/ts/services/contact-save.service.ts
cliftonmcintosh: > I asked Josh to clarify this comment about attachments to child docs, we are in a bit of a disagreement and my preference would be to keep things simple.

Sounds good. I'm happy to take whichever path the team decides is best.

> For sample contact forms, you can always look at the [default config](https://github.com/medic/cht-core/tree/master/config/default/forms/contact) contact forms, though these would not have child documents. 

Thanks for the tip!

> I will search for a real-world example that does have child documents.

Thank you!

cliftonmcintosh:
@dianabarsan 

I have a question about process. If a PR is still in draft and the branch becomes out-of-date with the base branch, is it okay if I rebase or does your team prefer a merge commit instead?

I'm inclined to rebase until the PR is out of draft status because it keeps the commit history looking cleaner without a merge commit, but I'm happy to follow whatever convention your team prefers.

webapp/src/ts/services/contact-save.service.ts
cliftonmcintosh: Based on [Josh's reply](https://github.com/medic/cht-core/issues/10509#issuecomment-3774812656) to your questions, it looks like you two are in agreement about starting with a simple approach. I will be happy to update this PR so that we only support attachments on the main document.

cliftonmcintosh: > For sample contact forms, you can always look at the [default config](https://github.com/medic/cht-core/tree/master/config/default/forms/contact) contact forms, though these would not have child documents. I will search for a real-world example that does have child documents.

Thanks again for pointing me to that directory, @dianabarsan. I have a couple of follow-up questions

- If I am trying to write realistic tests, are there particular contact forms that it might be good to choose for my tests? For example, person-create or person-edit?
- What is a realistic example of having an attachment to the main document for one of those forms? For example, in person-create xml, would we see an attachment at something like `/data/person/photo` or `/data/person/notes/attachment` or `data/person/ephemeral_dob/birth_certificate`?  I'd like to be able write tests that are realistic.

dianabarsan: > What is a realistic example of having an attachment to the main document

Because this feature has never existed, I do not believe we have one. I have heard of deployments wanting to upload PDFs, others to upload signatures (?). 

As for a report that has child documents, I found a few examples in private configs and all of them follow this model: https://github.com/medic/cht-core/blob/master/config/default/forms/contact/clinic-create.xml where the main contact gets created along with a primary contact. 
There might be some more elaborate examples out there, but I think for the tests we're good to go with this

dianabarsan:
> If a PR is still in draft and the branch becomes out-of-date with the base branch, is it okay if I rebase or does your team prefer a merge commit instead?

We general preference is to merge, because rebase loses all intermediary review comments which are very valuable to understand code progression. So please, avoid rebasing. We only rebase when the branch has additional commits (commits accidentally cherry picked) which makes reviewing very difficult.

dianabarsan:
> because it keeps the commit history looking cleaner without a merge commit,

we ALWAYS squash and merge, so the commit history gets lost anwyay.

dianabarsan:
There is more info about process in the docs: https://docs.communityhealthtoolkit.org/community/contributing/code/workflow/#opening-pull-requests

cliftonmcintosh:
@dianabarsan 

I've moved this out of draft status. Changes since you last reviewed include:

- simplifying the logic so that we only attach to the main document
- adding integration tests
- adding e2e tests

I have started trying to run through things locally myself but, as I understand it, I need cht-conf running to be able to add new forms that allow attachments. I have had trouble getting cht-conf going, but I will pick back up on Monday.

dianabarsan:
> I have had trouble getting cht-conf going, 

Oh no, that's not so good news. cht-conf should be the easiest one to set up. Can you share what the error is? 

From the message above, I didn't quite get if the PR is ready to review again. When it is, can you please re-request a review from the interface? That way I get notified and the message is clear, and avoids too many back and fourths. 
image

Thanks!

cliftonmcintosh:
> Oh no, that's not so good news. cht-conf should be the easiest one to set up. Can you share what the error is?

Actually, cht-conf is running. I think I haven't followed the directions correctly for getting cht-core and it working together correctly. I'll be working on that today.

cliftonmcintosh:
@dianabarsan 
I have managed to perform manual testing of an attachment upload. I was able to:
- upload an attachment when creating a person
- verify that the attachment was recorded in CouchDB
- see the field listed with the file name when editing a person

I have added a manual testing section to the description documenting how the testing was done and what has been tested so far.

<details>

<summary>Screenshots</summary>

### Empty person create form with photo field
create-form-empty-with-photo-field


### Person create form with photo added
create-form-with-added-photo


### Person edit form with photo shown
edit-form-with-photo-attached


</details>

dianabarsan:
Thanks for the updates @cliftonmcintosh . 

I have one comment about the edit. I believe that in report forms, when I edit a report with an attachment, I am able to edit the attachment, i.e. it's not read-only like in your demo: 

image

I believe this will be a requirement in the case of contacts as well.

cliftonmcintosh:
> I have one comment about the edit. I believe that in report forms, when I edit a report with an attachment, I am able to edit the attachment, i.e. it's not read-only like in your demo:

@dianabarsan 

Editing also works with a change to the edit form. 

I have not commited any of the form changes because I'm not sure if you want the sample forms to include an attachment field or not. If you do, I can commit them.

<details>

<summary>Here is a screen recording of editing</summary>

Here is a screen recording showing
- adding an attachment
- removing an attachment
- putting in a new attachment

Notice that there is an error message in editing because the file hasn't been loaded to the browser, but it is in the DB.

https://github.com/user-attachments/assets/d3a865cc-103a-4c99-b8bf-5baf236a89b8

</details>

The data are persisted in CouchDB

<details>

<summary>JSON with the attachment</summary>

Note that old attachments are not cleaned up. They stick around in the DB, but the field we used ("photo"), is linked to the correct attachment.

```json
{
  "_id": "ef7f9486-3c87-4d79-8210-f572227fff58",
  "_rev": "3-8c033c4ff061a64a2cf2ae501cb8a3b3",
  "parent": {
    "_id": "74ebaa07-2ed2-4c8d-a416-d7593b746784",
    "parent": {
      "_id": "65f4da96-baba-407e-bdaf-c5b7020303d6",
      "parent": {
        "_id": "0490ad32-f955-4171-9d95-e951db7a90ec"
      }
    }
  },
  "type": "person",
  "name": "Franklin Buckridge",
  "short_name": "Kai",
  "date_of_birth": "2016-09-03",
  "date_of_birth_method": "",
  "ephemeral_dob": {
    "dob_calendar": "2016-09-03",
    "dob_method": "",
    "dob_approx": "2026-01-26T00:00:00.000-06:00",
    "dob_raw": "2016-09-03",
    "dob_iso": "2016-09-03"
  },
  "phone": "+254783053185",
  "phone_alternate": "",
  "sex": "male",
  "role": "patient",
  "external_id": "",
  "notes": "",
  "photo": "Jason-20_55_59.png",
  "meta": {
    "created_by": "medic",
    "created_by_person_uuid": "",
    "created_by_place_uuid": "",
    "last_edited_by": "medic",
    "last_edited_by_person_uuid": "",
    "last_edited_by_place_uuid": ""
  },
  "patient_id": "1936774906",
  "reported_date": 1767350657212,
  "form_version": {
    "time": 1769480835402,
    "sha256": "f5279acd8a99dfed2cf8a29bb759164cad2d89c43dc96d8410dfb8b242d89952"
  },
  "_attachments": {
    "user-file-Cliff-20_55_34.png": {
      "content_type": "image/png",
      "revpos": 2,
      "digest": "md5-WBV27NcamPMqhuCuMZCf9A==",
      "length": 56971,
      "stub": true
    },
    "user-file-Jason-20_55_59.png": {
      "content_type": "image/png",
      "revpos": 3,
      "digest": "md5-M/QoV54zP+EJ8Dfkb0JZnw==",
      "length": 171461,
      "stub": true
    }
  }
}
```



</details>

<details>

<summary>Viewing the data via Fauxton</summary>


https://github.com/user-attachments/assets/e41920b5-c14a-42e4-baf6-c97f968dfd86



</details>

tests/e2e/default/contacts/contact-attachments.wdio-spec.js
```diff
@@ -0,0 +1,207 @@
...
+
+    const contactAfter = await utils.getDoc(contactId);
+
+    expect(contactAfter.name).to.equal(updatedName);
+    expect(contactAfter._attachments).to.exist;
+
+    const attachmentsAfter = Object.keys(contactAfter._attachments);
+    expect(attachmentsAfter).to.have.lengthOf(1);
+    expect(attachmentsAfter[0]).to.equal(originalAttachments[0]);
+  });
```
dianabarsan, line 243: Two more scenarios should be covered by e2e tests: 
- that the attachment can be removed
- that the attachment can be overwritten

dianabarsan: Currently, this is what I am seeing when trying to edit a contact with an attachment: 

Image

Compared to what I see in a report: 

Image

As a reference, this is how attachments are retrieved for edits for reports: https://github.com/medic/cht-core/blob/da5fe7f731303c6f69315b3992455125f539f91a/webapp/src/ts/modules/reports/reports-add.component.ts#L177

cliftonmcintosh: > Two more scenarios should be covered by e2e tests:
> 
> * that the attachment can be removed
> * that the attachment can be overwritten

Right now, when an attachment is removed or overwritten in the UI, the old attachment is still present in the `"_attachments"` element but is no longer linked to the field in the form. [See the JSON example in my earlier comment.](https://github.com/medic/cht-core/pull/10570#issuecomment-3802827101) In that example, the Cliff photo was added, then replaced by the Jason photo. We can see that the `"photo"` element links to Jason correctly, but the Cliff photo is still in the `"_attachments"`.

Are we happy with that for now, or do we want to remove the orphaned attachment from the database as well? If so, that will likely require a code change.

dianabarsan: The orphaned attachments definitely need to be removed  from the database.

cliftonmcintosh: > Two more scenarios should be covered by e2e tests:
> 
> * that the attachment can be removed
> * that the attachment can be overwritten

Test for attachment can be removed added in caeb75604d49ae6e2f49645973aa71954918eca6
Test for attachment can be overwritten added in 89208b2fec92df50bdef97ab71b47e8b2301cfc9

cliftonmcintosh: > Currently, this is what I am seeing when trying to edit a contact with an attachment:
> 
> Image
> Compared to what I see in a report:
> 
> Image
> As a reference, this is how attachments are retrieved for edits for reports:
> 
> https://github.com/medic/cht-core/blob/da5fe7f731303c6f69315b3992455125f539f91a/webapp/src/ts/modules/reports/reports-add.component.ts#L177

Attachment preview now visible when editing

Screenshot 2026-02-02 at 12 06 05

cliftonmcintosh: > The orphaned attachments definitely need to be removed from the database.

Logic to remove orphaned attachments added in 73973cafae3d639822c2f20bd55eada9a16ed6cc

cliftonmcintosh:
@dianabarsan,

I've added two new e2e tests to cover scenarios as we discussed.

The [run for "ci-webdriver-default-workflows"](https://github.com/medic/cht-core/actions/runs/21602258673/job/62252506353?pr=10570) is failing the two new tests, but they are passing when I run `npm run wdio-local -- --suite=workflows`. 

Any idea what the differences might be between the github run and the `wdio-local` run?

<details>

<summary>Failure in CI</summary>

```sh
[chrome 144.0.7559.96 linux #0-2] Contact form attachments
[chrome 144.0.7559.96 linux #0-2]    ✓ should create contact with image attachment
[chrome 144.0.7559.96 linux #0-2]    ✓ should create contact with multiple attachments (image + document)
[chrome 144.0.7559.96 linux #0-2]    ✓ should preserve attachments when editing contact
[chrome 144.0.7559.96 linux #0-2]    ✖ should remove attachment when editing contact (5x retries)
[chrome 144.0.7559.96 linux #0-2]    ✖ should replace attachment when editing contact (5x retries)
[chrome 144.0.7559.96 linux #0-2]
[chrome 144.0.7559.96 linux #0-2] 3 passing (1m 6.8s)
[chrome 144.0.7559.96 linux #0-2] 2 failing
[chrome 144.0.7559.96 linux #0-2]
[chrome 144.0.7559.96 linux #0-2] 1) Contact form attachments should remove attachment when editing contact
[chrome 144.0.7559.96 linux #0-2] Cannot read properties of undefined (reading 'click')
[chrome 144.0.7559.96 linux #0-2] TypeError: Cannot read properties of undefined (reading 'click')
[chrome 144.0.7559.96 linux #0-2]     at Object.clickFastActionFAB (/home/runner/work/cht-core/cht-core/tests/page-objects/default/common/common.wdio.page.js:220:13)
[chrome 144.0.7559.96 linux #0-2]     at async Context.<anonymous> (/home/runner/work/cht-core/cht-core/tests/e2e/default/contacts/contact-attachments.wdio-spec.js:212:5)
[chrome 144.0.7559.96 linux #0-2]
[chrome 144.0.7559.96 linux #0-2] 2) Contact form attachments should replace attachment when editing contact
[chrome 144.0.7559.96 linux #0-2] Cannot read properties of undefined (reading 'click')
[chrome 144.0.7559.96 linux #0-2] TypeError: Cannot read properties of undefined (reading 'click')
[chrome 144.0.7559.96 linux #0-2]     at Object.clickFastActionFAB (/home/runner/work/cht-core/cht-core/tests/page-objects/default/common/common.wdio.page.js:220:13)
[chrome 144.0.7559.96 linux #0-2]     at async Context.<anonymous> (/home/runner/work/cht-core/cht-core/tests/e2e/default/contacts/contact-attachments.wdio-spec.js:267:5)
```

</details>

<details>

<summary>Passing when I run locally</summary>

Here is the passing output from my terminal.
```sh
[chrome 144.0.7559.110 mac #0-2] Running: chrome (v144.0.7559.110) on mac
[chrome 144.0.7559.110 mac #0-2] Session ID: 0e6dec1578bad423788396ecc0b2158a
[chrome 144.0.7559.110 mac #0-2]
[chrome 144.0.7559.110 mac #0-2] » tests/e2e/default/contacts/contact-attachments.wdio-spec.js
[chrome 144.0.7559.110 mac #0-2] Contact form attachments
[chrome 144.0.7559.110 mac #0-2]    ✓ should create contact with image attachment
[chrome 144.0.7559.110 mac #0-2]    ✓ should create contact with multiple attachments (image + document)
[chrome 144.0.7559.110 mac #0-2]    ✓ should preserve attachments when editing contact
[chrome 144.0.7559.110 mac #0-2]    ✓ should remove attachment when editing contact
[chrome 144.0.7559.110 mac #0-2]    ✓ should replace attachment when editing contact
[chrome 144.0.7559.110 mac #0-2]
[chrome 144.0.7559.110 mac #0-2] 5 passing (26.5s)
```

</details>

There are a couple other suites that are failing in CI, [ci-integration-all-all](https://github.com/medic/cht-core/actions/runs/21602258673/job/62252506405#logs) and [ci-webdriver-default-core-minimum-browser](https://github.com/medic/cht-core/actions/runs/21602258673/job/62252506479#logs). When i look at those failures, I don't see anything that might be connected to the changes in contacts. Does anything jump out at you?

tests/e2e/default/contacts/contact-attachments.wdio-spec.js
```diff
@@ -0,0 +1,332 @@
...
+    }, { timeout: 5000, timeoutMsg: 'Alert did not appear' });
+
+    await browser.acceptAlert();
+
+    // Wait for the preview to be removed
+    const filePreview = await filePicker.$('.file-preview img');
+    await filePreview.waitForExist({ reverse: true, timeout: 5000 });
+
+    const fakeInput = await filePicker.$('.fake-file-input');
+    expect(await fakeInput.getValue()).to.equal('');
```
dianabarsan: Very nice! I think this is generic enough to pull into some .page file. Probably in the common enketo page that your test is already using.

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -338,9 +342,69 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
...
+    if (!fileName) {
+      return null;
+    }
+
+    const attachmentName = `user-file-${fileName}`;
+    return this.getAttachment(docId, attachmentName);
+  }
+
+  private renderAttachmentPreviews(contactId: string) {
+    return Promise
```
dianabarsan: This pattern of waiting for the next tick is a little weird. Why is this needed?
Also, can you please convert this to async/await?

webapp/src/ts/services/contact-save.service.ts
```diff
@@ -76,6 +84,97 @@ export class ContactSaveService {
...
+      if (!value) {
+        return false;
+      }
+      const possibleAttachmentName = `user-file-${value}`;
+      return existingAttachmentNames.includes(possibleAttachmentName);
+    };
+
+    const scanValue = (value: any) => {
+      if (typeof value === 'string' && isReferencedAttachment(value)) {
+        referenced.add(`user-file-${value}`);
```
dianabarsan: It feels like there are just sooooo many instances where this prefix is used as a magic string. This should definitely be saved as a constant somewhere.

webapp/src/ts/services/contact-save.service.ts
```diff
@@ -76,6 +84,97 @@ export class ContactSaveService {
...
+
+    const isReferencedAttachment = (value: string): boolean => {
+      if (!value) {
+        return false;
+      }
+      const possibleAttachmentName = `user-file-${value}`;
+      return existingAttachmentNames.includes(possibleAttachmentName);
+    };
+
+    const scanValue = (value: any) => {
```
dianabarsan: I am disappointed in JS for not having some built-in elegant version of checking whether an object has a deep key :| and the process to search is painfully manual and verbose.

webapp/tests/karma/ts/modules/contacts/contacts-edit.component.spec.ts
```diff
@@ -680,6 +691,168 @@ describe('ContactsEdit component', () => {
...
+
+        expect(getAttachment.notCalled).to.be.true;
+        expect(fileReaderService.base64.notCalled).to.be.true;
+      }));
+    });
+
+    describe('for existing contact', () => {
+      it('loads form with image having loaded file name', fakeAsync(async () => {
+        routeSnapshot.params = { id: 'the_person' };
+        const doc = { _id: 'the_person', type: 'person' };
```
dianabarsan: Is this true? we're calling `getAttachment` even if the document we are loading doesn't have an _attachment property? 

The document contents should already be available before we try to get attachment content, and we can check for the _attachments property and the presence of the desired attachment. Only if it exists, the we can call get attachment.

webapp/tests/karma/ts/services/contact-save.service.spec.ts
```diff
@@ -225,4 +233,406 @@ describe('ContactSave service', () => {
...
+      const type = 'person';
+
+      const mockFile = new File(['test file content'], 'test-photo.png', { type: 'image/png' });
+      sinon.stub(FileManager, 'getCurrentFiles').returns([mockFile]);
+
+      enketoTranslationService.contactRecordToJs.returns({
+        doc: { _id: 'person1', type: 'person', name: 'John Doe', phone: '+254712345678', sex: 'male' }
+      });
+
+      return service
```
dianabarsan: Please convert all these tests to async / await

webapp/tests/karma/ts/services/contact-save.service.spec.ts
```diff
@@ -225,4 +233,406 @@ describe('ContactSave service', () => {
...
+        .save(form, docId, type)
+        .then(() => {
+          assert.isTrue(attachmentService.add.calledOnce, 'AttachmentService.add should be called once');
+
+          const addCall = attachmentService.add.getCall(0);
+          assert.equal(addCall.args[0]._id, 'person1', 'Should attach to the main document');
```
dianabarsan: we have remnants of the code that use `assert` rather than `expect`. 8 years ago, we shifted primarily towards `expect` and didn't add more tests that are `assert` based. You stumbled upon a file that was migrated as-is from a very old codebase. 

I'm leaving it up to you how you prefer to handle this file. My preference is to be consistent through the codebase and use `expect`.

webapp/tests/karma/ts/services/contact-save.service.spec.ts
```diff
@@ -225,4 +233,406 @@ describe('ContactSave service', () => {
...
+        const docId = 'person1';
+        const type = 'person';
+
+        sinon.stub(FileManager, 'getCurrentFiles').returns([]);
+
+        getContact.withArgs(Qualifier.byUuid('person1')).resolves({
+          _id: 'person1',
+          type: 'person',
+          name: 'John Doe',
+          photo: 'existing-photo.png',
```
dianabarsan: In order to fully test your object search function, can you please hide this field in a nested array + object so we make sure that code works?

webapp/tests/karma/ts/services/contact-save.service.spec.ts
```diff
@@ -225,4 +233,406 @@ describe('ContactSave service', () => {
...
+          .save(form, docId, type)
+          .then(() => {
+            assert.isTrue(
+              attachmentService.remove.calledWith(sinon.match({ _id: 'person1' }), 'user-file-old-photo.png'),
+              'Should remove the attachment when field is cleared'
+            );
+          });
+      });
+
+      it('should not remove non-user-file attachments', () => {
```
dianabarsan: Nice! Great to add this test! 
However there is no way, to my knowledge, that other attachments can make it to the contact (as this is why you are writing this code!). 
When I see tests like this, I am confused. _Should_ we test for workflows we don't support?

webapp/tests/karma/ts/services/contact-save.service.spec.ts
```diff
@@ -225,4 +233,406 @@ describe('ContactSave service', () => {
...
+          .save(form, docId, type)
+          .then(() => {
+            assert.isFalse(
+              attachmentService.remove.called,
+              'Should not remove non-user-file attachments'
+            );
+          });
+      });
+
+      it('should handle multiple attachments with mixed actions', () => {
```
dianabarsan: <3 <3 <3

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -338,9 +342,69 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
...
+      .resolve()
+      .then(() => Promise
+        .all($('#contact-form input[type="file"]:not(.draw-widget__load)')
```
dianabarsan: I'm getting an error when trying to edit a contact with an existing attachment when the attachment name has some special characters: 

Image

Edit works if the attachment name does _not_ have special characters. 
I basically just used the default config resources folder to upload photo attachments: https://github.com/medic/cht-core/tree/master/config/default/resources

webapp/tests/karma/ts/services/contact-save.service.spec.ts
cliftonmcintosh: Do we want the tests that were already here in this file before I started converted to async/await as well?

webapp/src/ts/modules/contacts/contacts-edit.component.ts
cliftonmcintosh: I've updated it so that special characters are removed. I've verified that the edit will load. Here is a screen recording trying different file names.

I added unit tests but no e2e tests. I was having trouble previously with e2e tests that created and then went into the edit flow. I'm happy to try that again if needed.


https://github.com/user-attachments/assets/d518bb26-c956-4395-9b7a-9572dfd86ba7

webapp/tests/karma/ts/services/contact-save.service.spec.ts
cliftonmcintosh: Good point. Test removed

webapp/tests/karma/ts/services/contact-save.service.spec.ts
cliftonmcintosh: Good idea. Test updated so that the attachment is in a nested field

webapp/tests/karma/ts/services/contact-save.service.spec.ts
cliftonmcintosh: I've converted all the tests in this file to use async/await

webapp/tests/karma/ts/modules/contacts/contacts-edit.component.spec.ts
cliftonmcintosh: Good catch. I've refactored this so that we don't try to get attachments if none exist

webapp/src/ts/services/contact-save.service.ts
cliftonmcintosh: Added a constant

webapp/src/ts/modules/contacts/contacts-edit.component.ts
cliftonmcintosh: This was copied almost verbatim from reports. However, I have refactored the version in contact edits. Let me know what you think, please.

tests/e2e/default/contacts/contact-attachments.wdio-spec.js
cliftonmcintosh: This has been redone so that we no longer go through multiple steps in the test. Is there still anything here that you think would be useful in the common enketo page?

webapp/tests/karma/ts/services/contact-save.service.spec.ts
cliftonmcintosh: Replaced all the asserts with expects in this file

cliftonmcintosh:
@dianabarsan 
[SonarQube has identified a couple of changes](https://sonarcloud.io/project/issues?id=medic_cht-core&pullRequest=10570&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true) that are incompatible with the current settings for this project.
1. [Prefer `String#replaceAll()` over `String#replace()`.](https://sonarcloud.io/project/issues?pullRequest=10570&open=AZwqGWF1KsRC0Zc7tAaV&id=medic_cht-core) requires es2021 or later. The `webapp/tsconfig.json` doesn't have that allowed, so compilation fails.
2. [Use 'Object.hasOwn()' instead of 'Object.prototype.hasOwnProperty.call()'.](https://sonarcloud.io/project/issues?pullRequest=10570&open=AZwrDiL1uubvgzBMoBL7&id=medic_cht-core) requires es2022 or later. Same issue.

Can we override the SonarQube check? Or should I update `webapp/tsconfig.json`?

mrjones-plip:
I suspect you'll end up disabling sonar for just this, but I'll defer to Diana if we want to uplift to es2021 (feels out of scope). Excellent call to reach out for clarification! 

Otherwise, so you know for this time or next time, you can disable sonar for a stanza:

```js
// NOSONAR_BEGIN
const readMapFunction = function (doc) {
  var parts = doc._id.split(':');
  if (parts[0] === 'read') {
    emit(parts[1]);
  }
};
// NOSONAR_END
```

Or just a line:

```js
Object.defineProperties($, { // NOSONAR
  htmlPrefilter: { value: (html) => html.replace(rxhtmlTag, '<$1></$2>') }
});
```

cliftonmcintosh:
> Otherwise, so you know for this time or next time, you can disable sonar

Thanks for the tip. I've disabled it for those two lines and will call it out in a comment for reviewer(s).

webapp/src/ts/services/contact-save.service.ts
```diff
@@ -76,6 +85,185 @@ export class ContactSaveService {
...
+  /**
+   * Sanitizes a file name by removing special characters.
+   * Only allows letters (a-z, A-Z), numbers (0-9), underscores (_), dashes (-), and dots (.).
+   * All other characters are removed.
+   *
+   * @param fileName - The file name to sanitize
+   * @returns The sanitized file name with special characters removed
+   */
+  private sanitizeFileName(fileName: string): string {
+    return fileName.replace(/[^a-zA-Z0-9_.-]/g, ''); // NOSONAR
```
cliftonmcintosh: Sonar is suggesting using `fileName.replaceAll`, but that is an es2021 feature.

webapp/src/ts/services/contact-save.service.ts
```diff
@@ -76,6 +85,185 @@ export class ContactSaveService {
...
+      if (shouldSanitizeString(obj[key])) {
+        obj[key] = fileNameMap.get(obj[key]);
+      } else if (shouldRecurse(obj[key])) {
+        sanitizeInPlace(obj[key]);
+      }
+    };
+
+    const sanitizeObjectProperties = (obj: Record<string, any>) => {
+      for (const key in obj) {
+        if (!Object.prototype.hasOwnProperty.call(obj, key)) { // NOSONAR
```
cliftonmcintosh: Sonar is suggesting using `Object.hasOwn`, but that is an es2022 feature.

cliftonmcintosh:
The [ci-webdriver-default-lowLevel](https://github.com/medic/cht-core/actions/runs/21697597698/job/62571777217?pr=10570#logs) job is failing. It hasn't before, and I don't see it being related to any recent changes. Is this a flaky test we can re-run?

```
[chrome 144.0.7559.96 linux #0-1] » tests/e2e/default/service-worker/service-worker.wdio-spec.js
[chrome 144.0.7559.96 linux #0-1] Service worker cache
[chrome 144.0.7559.96 linux #0-1]    ✓ confirm initial list of cached resources
[chrome 144.0.7559.96 linux #0-1]    ✓ branding updates trigger login page refresh (1x retries)
[chrome 144.0.7559.96 linux #0-1]    ✓ login page translation updates trigger login page refresh (1x retries)
[chrome 144.0.7559.96 linux #0-1]    ✓ adding new languages triggers login page refresh (1x retries)
[chrome 144.0.7559.96 linux #0-1]    ✓ other translation updates do not trigger a login page refresh
[chrome 144.0.7559.96 linux #0-1]    ? should load the page while offline (1x retries)
[chrome 144.0.7559.96 linux #0-1]    ✖ "before each" hook for Service worker cache
[chrome 144.0.7559.96 linux #0-1]
[chrome 144.0.7559.96 linux #0-1] 5 passing (8m 55.8s)
[chrome 144.0.7559.96 linux #0-1] 1 failing
```

[ci-integration-all-all](https://github.com/medic/cht-core/actions/runs/21697597698/job/62571777205?pr=10570#logs) also looks to me like a problem unrelated to the changes in this PR. Can we re-run it?

mrjones-plip:
Diana re-ran it and it [succeeded](https://github.com/medic/cht-core/actions/runs/21697597698?pr=10570)!

webapp/tests/karma/ts/modules/contacts/contacts-edit.component.spec.ts
```diff
@@ -562,14 +573,16 @@ describe('ContactsEdit component', () => {
           create_form: 'patient_create_form',
           edit_key: 'patient_edit_key',
         });
-        dbGet.resolves({ _id: 'patient_edit_form', form: true });
+        dbGet.onCall(0).resolves({ _id: 'patient_edit_form', form: true });
+        dbGet.onCall(1).resolves({ _id: 'the_patient', type: 'patient' });
```
cliftonmcintosh: In e1b49a5, a change was made to fetch the contact from the DB and check if there are attachments so that we don't fetch attachments when they don't exist. This leads to an extra DB request. This explains why the existing tests expectations have been updated to expect an extra `dbGet`.

This was in response to [reviewer feedback](https://github.com/medic/cht-core/pull/10570/#discussion_r2762460702):

> Is this true? we're calling getAttachment even if the document we are loading doesn't have an _attachment property?

Please let me know if there is a different approach you think might be better.

cliftonmcintosh: Actually, we already have it as `this.contact`! Let me revise this so we don't have the extra DB call.

cliftonmcintosh: Okay. Removed the extra DB call.

cliftonmcintosh:
Happy Monday, @dianabarsan!

I'm not sure if a notification from Github came through to you that this is ready for re-review. If so, this is a gentle bump of that notification. If not, please consider this the notification.

tests/e2e/default/contacts/contact-attachments.wdio-spec.js
```diff
@@ -0,0 +1,328 @@
...
+    const createdContact = await utils.getDoc(contactId);
+
+    expect(createdContact).to.exist;
+    expect(createdContact.name).to.equal(contactName);
+    expect(createdContact._attachments).to.exist;
+
+    const attachmentNames = Object.keys(createdContact._attachments);
+    expect(attachmentNames).to.have.lengthOf(1);
+    expect(attachmentNames[0]).to.match(/^user-file-photo-for-upload-form.*\.png$/);
+    expect(createdContact._attachments[attachmentNames[0]].content_type).to.equal('image/png');
```
dianabarsan: A nice to have assertion here would check for the file size (not exact but to make sure the size is approx what we expect)

tests/e2e/default/contacts/contact-attachments.wdio-spec.js
```diff
@@ -0,0 +1,328 @@
...
+    expect(contactId).to.exist;
+
+    const createdContact = await utils.getDoc(contactId);
+
+    expect(createdContact).to.exist;
+    expect(createdContact.name).to.equal(contactName);
+    expect(createdContact._attachments).to.exist;
+
+    const attachmentNames = Object.keys(createdContact._attachments);
+    expect(attachmentNames).to.have.lengthOf(2);
```
dianabarsan, line 185: Please add assertions about which attachments exist (to prove that it's not the same attachment saved twice!)

tests/integration/cht-form/default/contact-with-attachments.wdio-spec.js
```diff
@@ -0,0 +1,50 @@
...
+ * Integration tests for contact form with file attachments.
+ *
+ * Note: The cht-form web component tests form rendering and submission,
+ * but doesn't include ContactSaveService logic. These tests verify that:
+ * 1. File upload widgets render and accept files
+ * 2. Attachments are included in form submission
+ *
+ * The actual contact document creation (type: 'person') and attachment
+ * processing via ContactSaveService is tested via E2E tests.
+ */
```
dianabarsan: I'm unsure who this comment is directed at. Should be add comments like this to _all_ cht-form tests?

tests/e2e/default/contacts/contact-attachments.wdio-spec.js
```diff
@@ -0,0 +1,328 @@
...
+const personFactory = require('@factories/cht/contacts/person');
+const userFactory = require('@factories/cht/users/users');
+const loginPage = require('@page-objects/default/login/login.wdio.page');
+const commonPage = require('@page-objects/default/common/common.wdio.page');
+const commonEnketoPage = require('@page-objects/default/enketo/common-enketo.wdio.page');
+const genericForm = require('@page-objects/default/enketo/generic-form.wdio.page');
+const contactPage = require('@page-objects/default/contacts/contacts.wdio.page');
+
+/**
+ * E2E tests for contact forms with file attachments.
```
dianabarsan: I believe this comment is unnecessary. the test names should be sufficient to understand what is the purpose of this file.

webapp/src/ts/services/contact-save.service.ts
```diff
@@ -76,6 +85,185 @@ export class ContactSaveService {
...
+  }
+
+  /**
+   * Recursively scans through a document and replaces field values that reference
+   * uploaded file names with their sanitized equivalents. Modifies the document in place.
+   *
+   * @param doc - The document to scan and modify
+   * @param fileNameMap - Map of original file names to sanitized file names
+   */
+  private sanitizeFieldValues(doc: Record<string, any>, fileNameMap: Map<string, string>): void {
```
dianabarsan: Hmmm, it looks like now there are two functions that recurse over an objects' deep properties, one to sanitize and another to search for a value (scanValue). 
Can those be unified somehow? So that we only implement the recursion once and use if for both cases.

webapp/src/ts/services/contact-save.service.ts
```diff
@@ -76,6 +85,185 @@ export class ContactSaveService {
...
+    if (xmlStr) {
+      const $record = $($.parseXML(xmlStr));
+      const formId = $record.find(':first').attr('id');
+
+      $record
+        .find('[type=binary]')
+        .each((_idx, element) => {
+          const content = $(element).text();
+          if (content) {
+            const xpath = Xpath.getElementXPath(element);
```
dianabarsan, line 204: If we have the xpath, can't we use it to target-sanitize the field value directly? The xpath should correlate to the field property and path, if I'm not wrong.

tests/e2e/default/contacts/contact-attachments.wdio-spec.js
cliftonmcintosh: Thanks for the suggestion. Added.

tests/integration/cht-form/default/contact-with-attachments.wdio-spec.js
cliftonmcintosh: I think it's unnecessary. Removed

tests/e2e/default/contacts/contact-attachments.wdio-spec.js
cliftonmcintosh: Good point. Removed

webapp/src/ts/services/contact-save.service.ts
cliftonmcintosh: I think in this case I would not unify the traversals.

They have different purposes. 
- `sanitizeFieldValues` is mutating the data
- `scanValue` is reading data

It seems to me that if we had unified traversal function, it would then need an extra argument, perhaps a setter function, that would need to be passed in and maybe used sometimes? I worry that would introduce complexity that would be harder to reason about over the long term.

But if you disagree, I can try to implement something and we can see what you think of it.

webapp/src/ts/services/contact-save.service.ts
cliftonmcintosh: The xpath traversal here is used for creating attachment names for binary fields. I don't think it can help with field value sanitization because our sanitization is for file names for files that have been uploaded using the file upload widget, and they don't have an xpath. 

Based on your experience with reports, what is the use case for having binary fields? Signatures on the screen? I suppose we aren't even handling binary fields in contacts now, so maybe we could remove this entirely if we think users won't have a need for adding binary fields.

tests/e2e/default/contacts/contact-attachments.wdio-spec.js
cliftonmcintosh: I've added assertions that the file is not empty.

I haven't asserted a particular file size because
- if the files used in testing are changed, then the test would break and need recalculating
- if I understand it correctly, the file size is calculated by Couch's API
- we should be able to rely on the Couch API as long as we are calling it correctly

Please let me know if you are concerned about the non-zero assertions.

cliftonmcintosh:
I've run through manual testing again to 
- create a person with an attachment
- edit the person (and be able to see the attachment)
- remove an attachment during edit
- add an attachment during edit
- replace an attachment during edit
- verify that the DB data is correct (no orphan attachments)
- verify that attachments with special characters in their names are sanitized and can be viewed in edit

Here's a recording of that

https://github.com/user-attachments/assets/25d51909-7cd6-4d3d-a8a0-900cd2a3afa0

webapp/src/ts/services/contact-save.service.ts
dianabarsan: you're right, this doesn't apply to all attachments :( 
To answer your question: there is a form action that connects to the Android app, to start another application through an Intent, and that application (whatever it is) can send binary data back into the form. I believe that is the primary use case. 

Thanks for your patience!

webapp/src/ts/services/contact-save.service.ts
dianabarsan: I agree that the purpose is different, but these traversals are so busy in code and I really dislike them. 
My dream was that the traversal function would return a path (or multiple paths) that match a filter, and the path would be accessible with `object-path` (https://www.npmjs.com/package/object-path) which is already a dependency and handled separately.

cliftonmcintosh: I've made an update that uses object-path for this. Please let me know what you think.

cliftonmcintosh:
It looks like the [ci-integration-all-all ](https://github.com/medic/cht-core/actions/runs/21840897326/job/63026137345?pr=10570) flaked out. When this happens, should I post asking for a re-run? Or just know that eventually someone will re-run it if needed?

dianabarsan:
I re-ran the test and it passed. I didn't have time today to do a full review, just took a quick look at the latest commits and I'm really liking the new traversals! Thanks! 
I'll do the full review tomorrow.

mrjones-plip:
> When this happens, should I post asking for a re-run?

Yeah - here or in slack!

webapp/src/ts/services/contact-save.service.ts
```diff
@@ -76,6 +86,162 @@ export class ContactSaveService {
...
+  /**
+   * Sanitizes a file name by removing special characters.
+   * Only allows letters (a-z, A-Z), numbers (0-9), underscores (_), dashes (-), and dots (.).
+   * All other characters are removed.
+   *
+   * @param fileName - The file name to sanitize
+   * @returns The sanitized file name with special characters removed
+   */
+  private sanitizeFileName(fileName: string): string {
+    return fileName.replace(/[^a-zA-Z0-9_.-]/g, ''); // NOSONAR
```
cliftonmcintosh: @dianabarsan 
Is this character limitation a problem for places that use other writing systems? Will the files always be named with Latin characters? Do we need to adjust the sanitization?

dianabarsan: Good point. I am nooooot sure. I suppose we could ask @binokaryg or @bhishankc about how files are named in Nepali. The safer option would be to only strip out special characters?

mrjones-plip: I propose we keep this as is. In order of relevance, here's why I think this:
1. we have [many examples](https://github.com/search?q=repo%3Amedic%2Fcht-core+%22a-zA-Z%22&type=code) of using `a-zA-Z` as "safe"
2.  While Nepali likely is an alphabet we should support, this feature is actually initially being deployed in East Africa.  As well, what about Arabic?  What about Cyrillic? 
3.  There's a larger idea called [universal acceptance](https://en.wikipedia.org/wiki/Universal_Acceptance) (UA) that we should consider. 

I think the approach should be, at a later date,  do an audit of our code for UA and address this concern then.

mrjones-plip: After thought: Maybe there's a defensive code happy path where if a file is named `تصدير`  and the result of `sanitizeFileName()` returns an empty string in which case we give the filename of a  UUIDv7 instead? 

Right now a file is uploaded called  `تصدير`  , we'd [return](https://github.com/medic/cht-core/pull/10570/changes#diff-fe214cb950886c1d0d6476050ae2bcf154c8134921990b617b8348375173bb68R171) a file name called `user-file-`, if I'm following the code correctly.  Maybe worth an update :shrug:  (oh maybe files without suffixes (eg `.png`)  aren't allowed by enketo/browser?)

binokaryg: User-generated files can sometimes have Devanagari script in their names. Devanagari script is common across Nepali, Hindi, and other regional languages.

For example, a user could capture an image and rename it to "घर.jpg" in Nepali for easier identification later.

It would be good to support UTF-8 characters if not too difficult.

dianabarsan: Thank you @binokaryg !

cliftonmcintosh: Following up here to see if there is any action needed right now. Are we going to stick with things the way they are for this PR and adjust later or do we want an adjustment like the one that @mrjones-plip suggested:

> Maybe there's a defensive code happy path where if a file is named `تصدير` and the result of `sanitizeFileName()` returns an empty string in which case we give the filename of a UUIDv7 instead?

cliftonmcintosh:
Please let me know if there are any actions I need to take on this PR. I can't do any merging, so I assume someone else will merge if this is still considered to be approved.

dianabarsan:
I was expecting a follow-up on the conversation about non-latin characters in file names. We would at least add a unit test showing what happens when you have a file containing only non-latin characters in its name.

cliftonmcintosh:
> I was expecting a follow-up on the conversation about non-latin characters in file names. We would at least add a unit test showing what happens when you have a file containing only non-latin characters in its name.

Thanks for clarifying the expectations. I've updated the sanitization function so that it uses a UUID when there are no characters that match the regex. Unit test added as well. Please let me know if we want anything else.

mrjones-plip:
Thanks for keeping this PR moving Cliff and Diana!  

I think the unit test is great and the UUID handling is icing on the cake.  I defer to Diana to finalize the review on the last changes.

webapp/src/ts/services/contact-save.service.ts
```diff
@@ -91,11 +91,25 @@ export class ContactSaveService {
...
+   * @returns The sanitized file name with special characters removed, or a UUID-based name if
+   *          the stem becomes empty after sanitization
    */
   private sanitizeFileName(fileName: string): string {
-    return fileName.replace(/[^a-zA-Z0-9_.-]/g, ''); // NOSONAR
+    const lastDotIndex = fileName.lastIndexOf('.');
+    const hasExtension = lastDotIndex > 0;
+
+    if (!hasExtension) {
+      return fileName.replace(/[^a-zA-Z0-9_.-]/g, '') || uuidV4(); // NOSONAR
```
dianabarsan: One request about storing the regex in a variable, so it's absolutely clear we're sanitizing with the same rules.

cliftonmcintosh: Yes. Of course. Done!

cliftonmcintosh:
I'm unfamiliar with the CHT process for merging approved PRs. Does the PR stay unmerged until the target version of CHT is almost ready to be released? 

Let me know if there is anything I need to be doing or if there is anything that needs updating before merging.

mrjones-plip:
Hey @cliftonmcintosh ! Thanks for reaching out.  For external contributors, the process is that the person who reviewed the PR is the one who merges it.  If  you have no more pending feedback to tend to and the PR has been approved, you're good to go! (which both are true  here ; ) 

I've kicked off CI again in hopes it was a fluke that it didn't pass before :shrug:

dianabarsan:
I can merge when we believe it's ready, but @cliftonmcintosh comes up with the [best questions](https://github.com/medic/cht-docs/pull/2133#discussion_r2824358534)! I think we would need an answer for these questions, the worst that can happen is that we ungracefully crash if a sub-document has an image attachment.

cliftonmcintosh:
> I can merge when we believe it's ready, but @cliftonmcintosh comes up with the [best questions](https://github.com/medic/cht-docs/pull/2133#discussion_r2824358534)! I think we would need an answer for these questions, the worst that can happen is that we ungracefully crash if a sub-document has an image attachment.

@dianabarsan I'll be happy to explore this. I will dig in to find examples of sub-documents. If you have pointers for where I might find some to test with in the default config, please let me know. That might save us a little time.

I am traveling today through Monday so likely won't get back to this until Tuesday.

## Diana code reviewing

one-more-thing

dianabarsan:
Nevermind. I tested this myself. The attachment gets added to the main doc.

cliftonmcintosh:
> Nevermind. I tested this myself. The attachment gets added to the main doc.

Great! Thanks. I was just investigating to try to figure that out.

mrjones-plip:
while the issue is closed, for next time @cliftonmcintosh, instead of running the whole build config as you mentioned in [Step 2: Compiled and Uploaded the Form](https://github.com/medic/cht-core/pull/10570#issue-3823124735), if you install [CHT Conf](https://docs.communityhealthtoolkit.org/community/contributing/code/cht-conf/) which can do any/all of the build calls, enables a much more targeted upload of a specific form:

```
cht  --local upload-contact-forms -- person-create
```

As this JUST calls `upload-contact-forms` for JUST the one form, it will be much faster than the 10 calls  in the [build conf script](https://github.com/medic/cht-core/blob/master/scripts/build/build-config.sh) processing ALL the contact and app forms in default config.

Right now I'm hoping to test a branch build of your work, and when coupled with [Docker Helper](https://docs.communityhealthtoolkit.org/hosting/cht/app-developer/#cht-docker-helper), I can be up and running very quickly!  (though, admittedly, I can't find images for your branch so I may [have to build them](https://forum.communityhealthtoolkit.org/t/change-background-color-of-navigation-tab-of-cht-application/4653/7?u=mrjones) :shrug: ).  Docker helper is of course not good for core dev work like in this branch - but ideal for ad hoc testing.

cliftonmcintosh:
> if you install [CHT Conf](https://docs.communityhealthtoolkit.org/community/contributing/code/cht-conf/) which can do any/all of the build calls, enables a much more targeted upload of a specific form:
> 
> ```
> cht  --local upload-contact-forms -- person-create
> ```

Thanks for the tip!

PR #10586, opened 2026-01-21
fix(#10575): replace deprecated message format parser

### LINKED ISSUE #10575: Look into message-format new parser
**Describe the issue**
I ran `npm install` for some other reason and this was a warning displayed: 
```
npm warn deprecated intl-messageformat-parser@1.8.1: We've written a new parser that's 6x faster and is backwards compatible. Please use @formatjs/icu-messageformat-parser
```

Look into the new super fast parser.

anurag2787:
# Description
This PR removes the deprecated intl-messageformat-parser by redirecting it to the newer and faster @formatjs/icu-messageformat-parser.

image

Resolves #10575

# Code review checklist

# License

### DIFF
--- a/webapp/package-lock.json
+++ b/webapp/package-lock.json
@@ -44,11 +44,11 @@
         "openrosa-xpath-evaluator": "^2.0.13",
         "partial-json-parser": "^1.2.2",
         "pojo2xml": "^1.7.2",
-        "rxjs": "^7.8.1",
+        "rxjs": "^7.8.2",
         "select2": "4.0.13",
         "signature_pad": "2.3.x",
         "tslib": "^2.8.1",
-        "zone.js": "^0.15.0"
+        "zone.js": "^0.16.0"
       },
       "devDependencies": {
         "@angular-builders/custom-webpack": "^19.0.0",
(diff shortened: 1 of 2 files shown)


anurag2787:
Hy @dianabarsan I have fixed the issue, please review Thank you!

dianabarsan:
@anurag2787 it looks like this upgrade is messing with a wdio dependency, which we cannot upgrade at the time: https://github.com/medic/cht-core/pull/10516#issuecomment-3768905361 

I'm not sure if there's anything more to be done here.

anurag2787:
> @anurag2787 it looks like this upgrade is messing with a wdio dependency, which we cannot upgrade at the time: [#10516 (comment)](https://github.com/medic/cht-core/pull/10516#issuecomment-3768905361)
> 
> I'm not sure if there's anything more to be done here.

Hi @dianabarsan this warning comes from lighthouse via the wdio fixing it would require upgrading wdio so if there’s anything else you’d like me to try?

dianabarsan:
i tried upgrading wdio in that PR I linked to. If you want to take up upgrading wdio, be my guest! as far as i remember, all wdio dependencies weren't available, but i might be wrong.

anurag2787:
I checked the WDIO versions and it looks like @wdio/devtools-service jumps from v8 straight to v10 while the rest of the WDIO stack is still on v9, which is what caused the shim error in #10516 
Given that, it doesn’t look like there’s a safe way for me to upgrade WDIO right now

dianabarsan:
That's exactly what I found as well, thanks so much for checking @anurag2787 and sorry for this being a dead end.

PR #10596, opened 2026-01-23
fix(api): block token login for safari browsers

RajeshReddy-bu:
This PR fixes login instability on Safari browsers.
- Safari strict cookie/token handling often causes login failures
- Returning early from requestTokenLogin forces standard password login on Safari


dianabarsan:
@RajeshReddy-bu we already have an open and ongoing PR to fix this issue: https://github.com/medic/cht-core/pull/10502

dianabarsan:
Closing in favor of merged PR: https://github.com/medic/cht-core/pull/10502

PR #10599, opened 2026-01-23
fix(#10386): fix offline user and admin role detection

### LINKED ISSUE #10386: Purge database created for admin role
**Describe the bug**
I'm seeing purging databases for the `admin` role (medic user). This is an online role , so shouldn't have a purge database.  This wastes >6GB space and a lot of daily processing time. Since purging is struggling to complete on this instance, skipping the upkeep of this purge step and only handling roles which are known and valid would be a nice improvement 

**To Reproduce**
On Nairobi instance, I'm seeing purge database `medic-purged-role-6ed8e2e23ed4cf28156a7cd33ccc3f94` where the hash corresponds to the role `admin` which is undefined.

If this is a reserved role, it is online. If it is not a reserved role, it is undefined.

**Expected behavior**
There should not be a purging database for any online role including admin.

**Screenshots**

**Environment**
- Instance: nairobi 

**Additional context**
https://gitlab.echis.go.ke/echis-core/config-echis-2-0/-/issues/2941

dianabarsan:
# Description

Admins are no longer `_admin` (since when?!!). Admins are now `admin` or have no role at all. 
This PR removes the purging specific `isOffline` code and calls the `user-management.roles.isOffline` instead. 
Also changes `user-management.roles.isOffline` to return `false` for empty roles and `false` for `admin` role. 

closes #10386

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10386-no-admin-purge-dbs/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10386-no-admin-purge-dbs/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10386-no-admin-purge-dbs/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/shared-libs/user-management/src/users.js
+++ b/shared-libs/user-management/src/users.js
@@ -619,7 +619,7 @@ const saveUserSettingsUpdates = async (userSettings) => {
 
 const validateFacilityIsNeeded = (data, user) => {
   const userRoles = data.roles || user?.roles;
-  if (userRoles && userRoles.length && roles.isOffline(userRoles)) {
+  if (userRoles && roles.isOffline(userRoles)) {
     return Promise.reject(error400(
       'Place field is required for offline users',
       'field is required',


dianabarsan:
@jkuester tiny nudge in case this one slipped through the cracks. Please let me know if you'd like me to reassign it if you're to stretched out.

shared-libs/user-management/src/roles.js
```diff
@@ -7,21 +6,21 @@ const { USER_ROLES } = require('@medic/constants');
...
-const DB_ADMIN_ROLE = '_admin';
+const DB_ADMIN_ROLES = ['admin', '_admin'];
 
 const hasRole = (userCtx, role) => {
-  return _.includes(userCtx && userCtx.roles, role);
+  return userCtx?.roles?.includes(role);
 };
 
-const isDbAdmin = userCtx => hasRole(userCtx, DB_ADMIN_ROLE);
+const isDbAdmin = userCtx => DB_ADMIN_ROLES.some(adminRole => hasRole(userCtx, adminRole));
```
jkuester, line 15: So, I think we are still ending up with some inconsistency here with the [permissions logic in cht-datasorce](https://github.com/medic/cht-core/blob/59b42e2fa3271a69c713b4720b9ee9fb7ea3dc61/shared-libs/cht-datasource/src/auth.js#L9).  

What do you think about updating cht-datasource to expose the `isAdmin` function so that we can consume it here (and anywhere else where we do the DB admin check.

Eventually it seems like we should consider moving more of this `roles.js` logic that needs to be reused outside of user-management over to cht-datasource, but that does not need to happen at this time...

shared-libs/user-management/src/roles.js
```diff
@@ -41,10 +40,15 @@ module.exports = {
     return userCtx && module.exports.hasOnlineRole(userCtx.roles);
   },
   isOffline: roles => {
+    if (!roles.length) {
+      return false;
+    }
+
```
jkuester, line 46: This should not break any existing users/instances, right? It makes me a bit nervous since it is definitely a change from the previous behavior, but I don't think it is possible for a user to login without a role.  So, it should not be possible to make it here with empty roles anyway.....

shared-libs/user-management/src/users.js
```diff
@@ -619,7 +619,7 @@ const saveUserSettingsUpdates = async (userSettings) => {
 
 const validateFacilityIsNeeded = (data, user) => {
   const userRoles = data.roles || user?.roles;
-  if (userRoles && roles.isOffline(userRoles)) {
+  if (userRoles && userRoles.length && roles.isOffline(userRoles)) {
```
jkuester: We don't actually need to check the `length` here (or below), right?  The `isOffline` logic now should return `false` if the given array is empty....

dianabarsan: I initially didn't want to change `isOffline` to check for roles length. And then ended up having this in two places.

shared-libs/user-management/src/roles.js
dianabarsan: I just checked, and users without roles are definitely broken. They are considered offline and they download their docs (assuming that everything is setup and they have a facility) and then the sync fails when they try to create the checkpointer document on the server, because they are not allowed to make changes to the `medic` database. 

I think we need better guarging for users without roles. The behavior of not being able to access the DB will persist even if the app doesn;t believe they are offline users.

shared-libs/user-management/src/roles.js
dianabarsan: I'm not sure. I don't think I like bloating datasource with _everything_.

jkuester: image


I am cool with whatever you think is best here.  Def agree that not everything belongs in cht-datasource, but at the same time I really hate duplicated logic. It is weird to have multiple places in the code controlling permissions. But, currently this is really minor logic, so no need to get too crazy with things right now... (Maybe in the future we need _another_ shared-lib for permissions that cht-datasource can depend on.... :sweat_smile:  :facepalm: )

dianabarsan: I think the shared lib for permissions makes sense. i know you said it as a joke, but I really think that modules that handle _one_ think are so much easier to grasp than one gigantic module that handles everything. datasource should handle ... what it's name says, communication with the DB, and nothing else.

Issue #2849, opened 2016-10-21
Consider putting each view into its own DDOC


SCdF:
We should have a proper discussion about the pros and cons of this approach.

I'll go first.

Pros:
- It makes it very clear how large each view is, which helps with keeping track of performance
- It reduces the likelyhood of unused views eating ram / disk space / cpu cycles, since views are only (re)generated when you query them
- Importantly for mobile, it makes sure that views users aren't actually using don't contribute to ram usage. If you never search a report by freetext, it will never get indexed.
- It would make the application snappier, since when (P|C)ouchDB boots and has to start loading views it only builds exactly what is necessary

Cons:
- It makes our code more complicated
- We'd have to work out how to deploy it in a non annoying way (maybe we've already done this re: medic-client)
- It makes it harder to know that you've warmed all the caches, since you have to hit each view individually. This could lead to the application feeling "bursty" and unpredictable.
- It hides issues, and could lead us to be less disciplined in what we put in medic-client.

As a reference, I took an example branch manager's database, and manually put each view in its own ddoc. Here are the sizes.

| view | size |
| --- | --- |
| contacts_by_freetext | 15.72mb |
| contacts_by_parent_name_type | 2.75mb |
| contacts_by_type | 1.75mb |
| contacts_by_type_freetext | 16.52mb |
| contacts_by_type_index_name | 2.21mb |
| data_records_read_by_type | 0.17mb |
| doc_by_type | 1.56mb |
| doc_summaries_by_id | 4.24mb |
| feedback | 0.01mb |
| forms | 0.01mb |
| help_pages | 0.27mb |
| messages_by_contact_date | 0.26mb |
| people_by_phone | 0.02mb |
| places_by_contact | 2.75mb |
| reports_by_date | 0.14mb |
| reports_by_form | 0.13mb |
| reports_by_form_year_month_places | 0.25mb |
| reports_by_form_year_week_places | 0.26mb |
| reports_by_freetext | 6.85mb |
| reports_by_place | 0.24mb |
| reports_by_subject | 0.20mb |
| reports_by_validity | 0.12mb |
| reports_by_verification | 0.12mb |
| total_clinics_by_facility | 4.60mb |
| **Total** | **61.5mb** |

From this we can make some rough statements:
- If users never search for contacts, they reduce their problem area by ~50%
- If they never search for contacts _or_ reports they reduce their problem area by ~60% 

[...]

PR #10615, opened 2026-02-04
feat(#2849): replace hardcoded view strings with constants

witash:
subset of #2849 
Replace hardcoded view strings wtih constants to allow changing which views are in which ddocs more easily in future.


witash:
@jkuester this is ready for review.

Mostly the changes are just find and replace the view paths with constants, mostly VIEWS.VIEW_NAME, which has the ddocs and view name combine e.g `medic/contacts_by_depth`. 
In some cases the view name or design doc name is needed by itself, so there are helper function for that.

In some cases design ddocs are referred to by name; these are still replaced with constants e.g. DDOC_IDS.MEDIC_CLIENT; I have left these even though to actually change these references it would not be enough to just change the constant values, you would need to figure out what the relevant code is actually doing and replace it with its new equivalent. Most of these are in tests where it is checking the ddoc that should be replicated to offline clients.

although the diff is very large, the only files which changed beyond find/replace are:
- the constants file https://github.com/medic/cht-core/pull/10615/changes#diff-26d8ae4a04a0fb99b3008f481808ff1116ee8988f5a5f9c4ba52a8f92620cac9
- datasource needs explicit type defintions https://github.com/medic/cht-core/pull/10615/changes#diff-cf4fe53798a9f4f0006f50e9c1fbb7b3dbc8aa7455a19d8cab02510044d0ad5c
- cht-conf queries contacts_by_depth directly. This should be changed, and it doesn't matter currently because I haven't changed the path in this branch, but I have added api/src/middleware/deprecated-view-middleware.js to work around external code that also hardcoded view paths. https://github.com/medic/cht-core/pull/10615/changes#diff-0d229773b6cda098c6c2c5c8c244e6542f790cc4619007ce2b1e469ab2bce7c4
- datasource needed explict type defintions: https://github.com/medic/cht-core/pull/10615/changes#diff-ad1b108bfb155049f2f42a52536095c37d493a8754e4931b3540f04064fb7060
- authorization changed medic-client to REPLICATED_DDOCS (although left only medic-client in REPLICATED_DDOCS anyway) https://github.com/medic/cht-core/pull/10615/changes#diff-43bd9ea2465bdd01a3a3483e843009f50dd5cec36e8189bc1dd649c783470e28

jkuester:

shared-libs/constants/src/index.js
```diff
@@ -32,16 +32,206 @@ const SENTINEL_METADATA = {
...
+  '_users': {
+    'users': [
+      'users_by_field',
+    ],
+  },
+  'medic-user': {
+    'medic-user': [
+      'read',
+    ],
+  },
```
jkuester: Is this supposed to be here?  I am not seeing the existing ddoc that this maps to...  🤔

shared-libs/cht-datasource/src/medic-constants.d.ts
jkuester, line 1: Can we actually just have this file in the `shared-libs/constants` directory (similar to how [we have the `.d.ts` file in `shared-libs/contact-type-utils`](https://github.com/medic/cht-core/blob/master/shared-libs/contact-types-utils/src/index.d.ts)?  That way, any other TS code (e.g. webapp) can also benefit.

shared-libs/cht-datasource/src/libs/constants.ts
```diff
@@ -9,3 +9,6 @@ export const END_OF_ALPHABET_MARKER = '\ufff0';
 
 /** @internal */
 export const ISO_8601_DATE_PATTERN = /^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d{3})?Z)$/;
+
+/** @ignore */
+export { VIEWS, NOUVEAU_INDEXES, nouveauUrl } from '@medic/constants';
```
jkuester: I am not sure if there is much value in "proxying" these constants through this file. Can just directly depend on `@medic/constants` in the cht-datasource code. 

The original goal of this file was just to store common constants that were specific to the cht-datasource code. Honestly, at least some of the current contents could probably be moved to shared-libs/constants....

api/src/routing.js
```diff
@@ -313,6 +313,11 @@ app.get(`${routePrefix}login/oidc/authorize`, login.oidcAuthorize);
 app.get(`${routePrefix}login/oidc`, login.oidcLogin);
 app.get(`${routePrefix}privacy-policy`, privacyPolicyController.get);
 
+// Rewrite old view URLs (/_design/medic/_view/... and /_design/medic-client/_view/...)
+// to the new design documents. This supports external projects that haven't updated yet.
+const { rewriteDeprecatedViewUrl } = require('./middleware/deprecated-view-rewrite');
+app.use(rewriteDeprecatedViewUrl);
```
jkuester: I might be missing something, but is this actually doing anything right now?  Since we have not moved any views around, I would think this is not being used.  If it is not used, we should probably remove it for now...

api/tests/mocha/services/authorization.spec.js
```diff
@@ -1055,13 +1080,6 @@ describe('Authorization service', () => {
         )
         .should.equal(true);
 
-      service
-        .allowedDoc(
-          '_design/medic-client',
-          { userCtx: userCtxMultiFacility },
-          { docsByReplicationKey: [{ key: '_all', value: null}], contactsByDepth: null}
-        )
-        .should.equal(true);
```
jkuester, line 1064: Was this removed on purpose?  Seems to be testing the multi-facility flow...

api/tests/mocha/services/monitoring.spec.js
```diff
@@ -319,7 +275,7 @@ describe('Monitoring service', () => {
         medic: {
           doc_count: 20,
           doc_del_count: 10,
-          fragmentation: 1.074747464888782,
+          fragmentation: 2,
```
jkuester: Why did this change?  👀

api/tests/mocha/services/monitoring.spec.js
```diff
@@ -459,7 +397,7 @@ describe('Monitoring service', () => {
...
```
jkuester: Same here.  Might have missed something, but why is the fragmentation changing here?

jkuester (approved):
Excellent uplift @witash!  I particularly appreciate your care for keeping the diff clean and only containing the necessary changes!  

I have gone through everything line by line and have just a few suggestions/questions.  Otherwise, LGTM!

shared-libs/cht-datasource/src/medic-constants.d.ts
witash: yes makes sense

shared-libs/cht-datasource/src/libs/constants.ts
witash: changed to refer to medic/constants

api/src/routing.js
witash: yea, in the current version where views are not actually moved, this isn't doing anything. If any views are ever moved to different, it or an alternatvie would have to be added back, but removed in this branch for now

api/tests/mocha/services/authorization.spec.js
witash: good catch, this was not supposed to be removed

api/tests/mocha/services/monitoring.spec.js
witash: VIEW_INDEX_INFO_BY_DESIGN changed from a large hardcoded json to being created by a function makeViewIndexInfo, which also changed the sizes; I've changed the sizes back to what they were.

andrablaj:
@witash @jkuester did we mean to close this PR?

witash:
yes, I think we should close this for now since it overlaps with moving views behind datasource.
We can create a new branch with some of the changes from this branch later if the changes are still relevant.

PR #10616, opened 2026-02-05
chore(#10606): bump minimum chrome version

### LINKED ISSUE #10606: Switch minimum browser version for e2e tests from 90 to 107
**Describe the issue**
We documented a bump of the minimum supported browser version for Webapp to Chrome 107, in support of allowing further Angular upgrades. However our e2e test suite still runs all "minimum browser" e2e tests on Chrome 90. 

Switch the minimum browser e2e test suite to use chrome 107.

dianabarsan:
# Description

Title is deceptive as this PR does a few more things: 

### Upgrades wdio to latest
This is a minor release bump, but one dependent package `devtools` (used to communicate with the browser) was deprecated in wdio@9 and now it's being reused for something else (????). Dependabot keeps trying to upgrade this deprecated package to the new version and everything fails. 
The package has a replacement, called `lighthouse` which depends on a new protocol called BiDi - basically a web socket between the test runner and the browser where communication is bidirectional. it's super strong and can achieve waaaay more stuff than `devtools`, however it made a lot of our e2e tests unstable or fail, so when we upgraded to wdio@9, we chose to use the legacy protocol (devtools). 
This PR stabilizes all the tests so we can use BiDi and up to date packages. 

### Bumping minimum chrome version to 107. 
This was relatively straight forward, with some minor tweaks so that tests work both with `chromedriver` protocol and `BiDi`. 

### Changing retry protocol 

Instead of using mocha retry, which just retried one test, we now use wdio retry, which tries the whole spec. 

### Updates minimum chrome version in Webapp and in e2e tests

closes #10606

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10606-bump-minimum-chrome-version/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10606-bump-minimum-chrome-version/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10606-bump-minimum-chrome-version/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/tests/utils/index.js
+++ b/tests/utils/index.js
@@ -1381,7 +1381,6 @@ const prepServices = async (defaultSettings) => {
   await setupUserDoc();
 };
 
-
 const getLogs = (container) => {
   const logFile = path.resolve(__dirname, '../logs', `${container.replace('pod/', '')}.log`);
   const logWriteStream = fs.createWriteStream(logFile, { flags: 'w' });


api/src/services/config-watcher.js
```diff
@@ -50,7 +50,7 @@ const initTransitionLib = () => {
   try {
     transitionsLib.loadTransitions(true);
   } catch (err) {
-    logger.error(err);
+    logger.error('%o', err);
```
dianabarsan: As I was trying to debug what failed in an e2e test, I noticed some weird `[Error Error]`  logs in the API logs. :(

api/src/services/forms.js
```diff
@@ -32,11 +30,9 @@ module.exports = {
     try {
       return await db.medic.getAttachment(docId, name);
     } catch (error) {
-      if (error.status === 404) {
-        logger.error(error);
-        return null;
+      if (error.status !== 404) {
+        throw error;
```
dianabarsan, line 36: I reviewed this code and missed this. Also produced some super noisy logs when uploading new forms. Of course attachments are missing.

tests/e2e/default/login/token-login.wdio-spec.js
```diff
@@ -76,38 +66,52 @@ describe('Token login', () => {
     expect(await loginPage.getToLoginLinkText()).to.equal(TO_LOGIN);
   });
 
-  it('should display an error with incorrect url', async () => {
-    await setupTokenLoginSettings();
-    await browser.url(`/medic/login/token`);
-    expect(await loginPage.getTokenError('missing')).to.contain(MISSING);
-    expect(await loginPage.getToLoginLinkText()).to.equal(TO_LOGIN);
-  });
+  describe('token login enabled', () => {
```
dianabarsan, line 69: Just some test grouping so we only have to setup settings once (and avoid all of the repeated modal craziness).

tests/utils/index.js
```diff
@@ -21,6 +21,16 @@ const { setTimeout: setTimeoutPromise } = require('node:timers/promises');
 process.env.COUCHDB_USER = constants.USERNAME;
 process.env.COUCHDB_PASSWORD = constants.PASSWORD;
 process.env.CERTIFICATE_MODE = constants.CERTIFICATE_MODE;
+
+// Suppress the warning about NODE_TLS_REJECT_UNAUTHORIZED
```
dianabarsan, line 25: I watched these logs for days and could not stand seeing the NODE_TLS_REJECT_UNAUTHORIZED again...

tests/utils/index.js
```diff
@@ -510,35 +520,21 @@ const deleteAllDocs = async (except = []) => {
   await deleteSentinelDocs(docsToKeep);
 };
 
+const updateCustomSettings = async (updates) => {
```
dianabarsan, line 532: I converted this to async-await. 
I also removed the error throw when settings were already changed. Couple of reasons: 
1. we have a global afer hook that runs after every spec file and reverts settings so, there is no risk of settings bleeding into other specs. 
2. sometimes you want to add a new contact type, a new role, change the permissions and setup tasks in the same test, and it's a massive pain to make all the changes in one request every time. 

This means we need more discipline in between tests within the same spec file, but I don't believe that's an issue.

tests/utils/index.js
```diff
@@ -1419,34 +1420,54 @@ const killSpawnedProcess = (proc) => {
  * that contains the promise to resolve when logs lines are matched and a cancel function
  */
 
-const waitForLogs = (container, tail, ...regex) => {
+const waitForLogs = async (container, tail, ...regex) => {
```
dianabarsan, line 1449: I was running the tests in low memory settings (don't ask why....), and there was a HUGE lag between when the listener started picking up the logs, and many lines were skipped. 
This seems to be more stable.

tests/wdio.conf.js
```diff
@@ -167,7 +173,7 @@ const baseConfig = {
   framework: 'mocha',
   //
   // The number of times to retry the entire specfile when it fails as a whole
-  // specFileRetries: 1,
+  specFileRetries: DEBUG ? 0 : 3,
```
dianabarsan, line 177: I've been wanting to do this for a LONG time! 
Instead of retrying one test 5 times, we retry the whole spec. Ideally every test should work in isolation, but in reality this is rarely the case. And what happened was that a test failed, was retried, and the retry failures were due to bad cleanup or bad test setup on repeat rather than the initial error, obfuscating what was really happening.

webapp/src/ts/modules/contacts/contacts-content.component.html
```diff
@@ -18,7 +18,7 @@
     <error-log [errorStack]="summaryErrorStack" [errorFor]="'contacts'"></error-log>
   </div>
 
-  <div class="col-sm-8 item-content" *ngIf="!selectedContact?.error && !loadingContent && !summaryErrorStack">
+  <div class="col-sm-8 item-content" *ngIf="!selectedContact?.error && !loadingContent && !summaryErrorStack && selectedContact">
```
dianabarsan, line 21: Apparently new wdio now detected that this element was visible and the test failed! 
Kudos to wdio for actually making a difference :D

dianabarsan:
@witash please don't hate me 🥹

dianabarsan:
@witash please let me know if you don't have the bandwidth to review this, and I can ask for a review from another dev.

dianabarsan:
Thank you for the review @witash !

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -162,7 +161,6 @@ export class TasksComponent implements OnInit, OnDestroy {
       const isEnabled = await this.rulesEngineService.isEnabled();
       this.tasksDisabled = !isEnabled;
       const taskDocs = isEnabled ? await this.rulesEngineService.fetchTaskDocsForAllContacts() : [];
-      this.hasTasks = taskDocs.length > 0;
```
dianabarsan, line 165: This triggered an "expression changed after it was checked" angular error, which only showed up once I ran the tests on the dev build locally. The tests were failing consistently with feedback docs getting created most of the time, and I also replicated it in browser. 

The value gets set through the redux flow state, and the display is correct now.

dianabarsan:
Hi @witash

There was a lot of chatter on this PR. I waited for the new features in 5.1 to land, to resolve conflicts and stabilize those tests. And then I went a little crazy with trying to stabilize some flakiness. There's a lot of commit chatter but very little actual changes. 

I want to ask you for one final sanity check before I merge. 
Thanks!

PR #10623, opened 2026-02-09
feat(#10577): adds task filtering

### LINKED ISSUE #10577: Implement task filtering

When CHWs provide services in the community, they face challenges in navigating the CHT app. With numerous tasks listed, it often becomes difficult to scroll through and identify specific follow-up tasks for each use case.

To improve efficiency, [the community suggested](https://forum.communityhealthtoolkit.org/t/filter-option-for-task/4808) adding a filter option (similar to the reporting feature) that lets CHWs select a specific form/use case and view only the relevant tasks they need to complete.

Provide the ability to filter tasks by high priority and/or form/use case.

**Additional context**

[Forum post](https://forum.communityhealthtoolkit.org/t/filter-option-for-task/4808)

dianabarsan:
# Description

Adds following task filters:
1. Due date - radio with two options, that show either tasks that are overdue (or due today) or that are due in the future. 

2. Task type - checkboxes for every task type available in the task list

3. Area - identical to the reports area filter. will filter tasks based on the task owner lineage. 

Minor refactor to how we display (and filter) item lineage, changed from being filtered in _every_ component that will display items with lineage to using the store for user facilities and filtering lineage in the lineage pipe itself. 

### AI Disclosure
AI Assistance: This PR was developed using Claude Code.

#### How AI was used

- I instructed Claude to reuse the Reports filters component for tasks, including the types of filters and which fields should be used for filtering
- Instructed to write unit tests for the new and changed code
- Instructed to generate an e2e test code skeleton
- Added translations for all of the new translation keys
- Instructed and guided through doing a high level refactor of a repeated code pattern (of lineage filtering) 

Human oversight:

- Corrected errors around area filters not being applied correctly
- Corrected Claude altering existing code behaviors - it switched the area filter from being lazy loaded to being eagerly loaded because it wanted to count the places. 
- Completed e2e test
- Manually verified the behavior using multiple types of user accounts
- Reviewed and perfected all code to adhere to my standards

closes #10577

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10577-task-filtering/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10577-task-filtering/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10577-task-filtering/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/webapp/web-components/cht-form/tsconfig.app.json
+++ b/webapp/web-components/cht-form/tsconfig.app.json
@@ -7,6 +7,7 @@
       "@mm-providers/*": ["src/ts/providers/*"],
       "@mm-reducers/*": ["src/ts/reducers/*"],
       "@mm-selectors/*": ["src/ts/selectors/*"],
+      "@mm-pipes/*": ["src/ts/pipes/*"],
       "@mm-services/*": ["src/ts/services/*"],
       "@mm-services/cht-datasource.service": ["web-components/cht-form/src/stubs/cht-datasource.service.ts"],
       "@mm-services/db.service": ["web-components/cht-form/src/stubs/db.service.ts"],
--- a/webapp/web-components/cht-form/tsconfig.spec.json
+++ b/webapp/web-components/cht-form/tsconfig.spec.json
@@ -7,6 +7,7 @@
       "@mm-providers/*": ["src/ts/providers/*"],
       "@mm-reducers/*": ["src/ts/reducers/*"],
       "@mm-selectors/*": ["src/ts/selectors/*"],
+      "@mm-pipes/*": ["src/ts/pipes/*"],
       "@mm-services/*": ["src/ts/services/*"],
       "@mm-services/db.service": ["web-components/cht-form/src/stubs/db.service.ts"],
     },


dianabarsan:
Hi @sugat009 ! Would appreciate a review. 

The PR does a bit of centralizing data: with adding facilities and online user type to the redux state. 
I used these directly in the work to move removing user facility from breadcrumbs in the lineage pipe instead of scattered around in every component. 
However there are more components and places that could benefit from this, but I chose not to address them now (due to the already large size of this PR). 
Another thing I really wanted to do was to unify the filters for more reuse - for example add a generic "status" filter and use if for report verification _and_ task overdue. But I think that would be another larger refactor that can wait, especially since @jkuester is working on filtering for the targets page. And refactoring after that lands will make sure we target all filters, not just the ones that exist now. 

Let me know what you think. Thanks for your time!

dianabarsan:
Added [demo](https://github.com/medic/cht-core/issues/10577#issuecomment-3888440993) to main issue.

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -153,9 +167,8 @@ export class TasksComponent implements OnInit, OnDestroy {
       const emissions = taskDocs.map(taskDoc => taskDoc.emission);
       const subjects = await this.getLineagesFromTaskDocs(emissions);
       if (subjects?.size) {
-        const userLineageLevel = await this.userLineageLevel;
         emissions.forEach(task => {
-          task.lineage = this.getTaskLineage(subjects, task, userLineageLevel);
+          Object.assign(task, this.getTaskLineage(subjects, task));
```
sugat009: suggestion: For consistency with the pattern already used in this file at line 90:
```typescript
  // Line 90 - existing pattern with comment
  // Make new reference because the one from store is read-only
  this.tasksList = tasksList.map(task => ({ ...task, selected: task._id === this.selectedTask?._id }));
```

  Consider using the same approach here:

  // Current
```typescript
  emissions.forEach(task => {
    Object.assign(task, this.getTaskLineage(subjects, task));
  });
```

  // Suggested (consistent with line 90)
```typescript
  const tasksWithLineage = emissions.map(task => ({
    ...task,
    ...this.getTaskLineage(subjects, task)
  }));
  this.tasksActions.setTasksList(tasksWithLineage);
```
  This keeps the mutation pattern consistent within the file.

webapp/src/ts/selectors/index.ts
```diff
@@ -2,6 +2,27 @@ import { createSelector } from '@ngrx/store';
 import { GlobalState } from '@mm-reducers/global';
 
 const getGlobalState = (state): GlobalState => state.global || {};
+
+const applyTasksFilters = (tasks, filters: Record<string, any> = {}) => {
```
sugat009: Non-blocking suggestion: Since this is new code, consider adding type annotations for better IDE support and catch potential issues at compile time.

webapp/src/ts/selectors/index.ts
```diff
@@ -2,6 +2,27 @@ import { createSelector } from '@ngrx/store';
...
+
+  if (filters.taskTypes?.selected?.length) {
+    const selectedTypes = filters.taskTypes.selected;
+    filtered = filtered.filter(task => selectedTypes.includes(task.title));
+  }
+
+  if (filters.facilities?.selected?.length) {
+    filtered = filtered.filter(task => {
+      return task.lineageIds.some(id => filters.facilities?.selected.includes(id));
+    });
```
sugat009, line 26:
```suggestion
  if (filters.taskTypes?.selected?.length) {
    const selectedTypesSet = new Set(filters.taskTypes.selected);
    filtered = filtered.filter(task => selectedTypesSet.has(task.title));
  }

  if (filters.facilities?.selected?.length) {
    const selectedFacilitiesSet = new Set(filters.facilities.selected);
    filtered = filtered.filter(task => 
      task.lineageIds.some(id => selectedFacilitiesSet.has(id))
    );
  }
```
Non-blocking suggestion: For larger task lists, using Set for lookups would be more efficient (O(1) vs O(n)). This becomes more noticeable with 500+ tasks and multiple selected filters.

webapp/src/ts/modules/tasks/tasks-sidebar-filter.component.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  @ViewChild(OverdueFilterComponent) overdueFilter!: OverdueFilterComponent;
+  @ViewChild(TaskTypeFilterComponent) taskTypeFilter!: TaskTypeFilterComponent;
+  @ViewChild(FacilityFilterComponent) facilityFilter!: FacilityFilterComponent;
+
+  private readonly globalActions: GlobalActions;
+  private filters: FilterComponent[] = [];
+  isResettingFilters = false;
+  isOpen = false;
+  filterCount: any = {};
```
sugat009: Non-blocking suggestion: Consider adding a type for filterCount instead of any:

```typescript
  interface FilterCount {
    total: number;
    [fieldId: string]: number;
  }

  filterCount: FilterCount = { total: 0 };
```

sugat009 (changes requested):
Minor feedback.

webapp/src/ts/selectors/index.ts
dianabarsan: I don't expect the number of tasks to be consistently large or to have so many filters. I'm not inclined to make this change. 
What do you mean when you say "noticeable"? Do you have some numbers?

sugat009: Let's skip this one if the number of tasks won't be high.

dianabarsan:
Thanks a lot for the review @sugat009 . I pushed most of your suggestions - one exception about using sets for filters. I think that just creates an additional layer of complexity and confusion. We expect filters to be bite-sized and the task list to be something that a human can scroll through. 
This is ready for review again, I believe. 
I appreciate your time!

sugat009:
@dianabarsan I just noticed something while testing manually. All filters does work but when clicking the "place filters" on the place name it does not check the box. Clicking the box checks the box for the place filter but not when clicking the place name. I've attached a video for clarity.

https://github.com/user-attachments/assets/f4c3b9af-8642-4cba-a6c0-40eefe3a3eb4

_The same issue exists for the place filter in reports as well_

dianabarsan:
@sugat009 the place filter is just reused from reports, so this is "inherited" behavior.

dianabarsan:
ah, now i saw the italics comment. I'm not sure about this one. the click on the place name is set to open the accordion (to display the lower level of places), so it's already assigned to an action.

sugat009:
I see. I think we can keep it as it is.

PR #10689, opened 2026-03-03
App skeleton

javier-perez:
# Description

# Code review checklist

# License

### DIFF
--- a/admin-tool/package.json
+++ b/admin-tool/package.json
@@ -25,8 +25,7 @@
     "font-awesome": "^4.7.0",
     "jquery": "3.5.1",
     "ngx-cookie-service": "^19.1.2",
-    "rxjs": "^7.8.2",
-    "select2": "4.0.13",
+"select2": "4.0.13",
     "tslib": "^2.8.1",
     "zone.js": "^0.16.1"
   },
--- a/admin-tool/package-lock.json
+++ b/admin-tool/package-lock.json
@@ -23,7 +23,6 @@
         "font-awesome": "^4.7.0",
         "jquery": "3.5.1",
         "ngx-cookie-service": "^19.1.2",
-        "rxjs": "^7.8.2",
         "select2": "4.0.13",
         "tslib": "^2.8.1",
         "zone.js": "^0.16.1"
@@ -13378,15 +13377,6 @@
       "license": "MIT",
       "dependencies": {
         "queue-microtask": "^1.2.2"
-      }
-    },
-    "node_modules/rxjs": {
-      "version": "7.8.2",
-      "resolved": "https://registry.npmjs.org/rxjs/-/rxjs-7.8.2.tgz",
-      "integrity": "sha512-dhKf903U/PQZY6boNNtAGdWbG85WAbjT/1xYoZIC7FAY0yWapOBQVsVrDl58W86//e1VpMNBtRV4MaXfdMySFA==",
-      "license": "Apache-2.0",
-      "dependencies": {
-        "tslib": "^2.1.0"
       }
     },
     "node_modules/safe-buffer": {


dianabarsan:
I've updated the https://github.com/medic/cht-core/tree/7369-admin-angular-upgrade branch, but it seems that the PR is still listing all the extra commits. I expect that it's because, on your fork, you created your `edify:app-skeleton` off of `medic:master` instead of `medic:7369-admin-angular-upgrade`. 
I think that a new branch needs to be created based off of `7369-admin-angular-upgrade`. And for future branches created on the fork, all should be created with `7369-admin-angular-upgrade` as a base, to avoid this altogether. 

I'm sorry. :( 
I'll only review the code in the `admin-tool` folder.

admin-tool/src/ts/app.config.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { AppRoutingModule } from './app-routing.module';
+
+export const appConfig: ApplicationConfig = {
+  providers: [
+    importProvidersFrom(
+      BrowserModule,
+      AppRoutingModule,
+      RouterModule,
+      StoreModule.forRoot({}),
```
dianabarsan: I don't think we'll need to use `ngrx` in the admin app. I don't believe we have any state shared between components. We do have one reducer now in the code, but it was because it is used by a service that was shared with webapp before, and nobody bothered to update it: https://github.com/medic/cht-core/blob/ee6b24ae719740c8a768250ef078ccce772d10cf/admin/src/js/services/changes.js#L91

admin-tool/tests/karma/karma-unit.base.conf.js
```diff
@@ -0,0 +1,53 @@
+module.exports = {
+  frameworks: ['mocha', '@angular-devkit/build-angular'],
```
dianabarsan, line 2: hoping we don't have to duplicate this code and just require it from webapp: https://github.com/medic/cht-core/blob/master/webapp/tests/karma/karma-unit.base.conf.js

admin-tool/package.json
```diff
@@ -0,0 +1,67 @@
...
+  "name": "medic-admin-tool",
+  "version": "0.1.0",
+  "private": true,
+  "license": "Apache-2.0",
+  "scripts": {
+    "build": "ng build",
+    "build-watch": "npm run build -- --configuration=development --watch=true",
+    "start": "ng serve",
+    "unit": "ng test admin-tool",
+    "lint": "../node_modules/.bin/eslint --color --cache src tests"
```
dianabarsan, line 11: Lint should be covered by our main command https://github.com/medic/cht-core/blob/master/package.json#L42

admin-tool/package.json
```diff
@@ -0,0 +1,67 @@
...
+  },
+  "dependencies": {
+    "@angular/animations": "19.1.7",
```
dianabarsan, line 14: I believe we would be using angular 20, correct?

admin-tool/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "@angular/common": "19.1.7",
+    "@angular/compiler": "19.1.7",
+    "@angular/core": "19.1.7",
+    "@angular/forms": "19.1.7",
+    "@angular/platform-browser": "19.1.7",
+    "@angular/platform-browser-dynamic": "19.1.7",
+    "@angular/router": "19.1.7",
+    "@messageformat/core": "^3.4.0",
+    "@ngrx/effects": "^19.0.1",
+    "@ngrx/store": "^19.0.1",
```
dianabarsan: I'm really hoping not to need these two. ngrx is a horror show to work with.

admin-tool/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "@ngx-translate/core": "^16.0.4",
+    "bootstrap": "^3.4.1",
+    "font-awesome": "^4.7.0",
+    "jquery": "3.5.1",
+    "ngx-cookie-service": "^19.1.2",
+    "rxjs": "^7.8.2",
```
dianabarsan: Same as this. rxjs is required by ngrx.

admin-tool/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "karma": "^6.4.2",
+    "karma-chrome-launcher": "^3.2.0",
+    "karma-coverage": "^2.2.1",
+    "karma-mocha": "^2.0.1",
+    "karma-mocha-reporter": "^2.2.5",
+    "less": "^4.2.2",
+    "mocha": "^11.7.4",
+    "sinon": "^21.0.1",
+    "sinon-chai": "^3.7.0",
+    "typescript": "~5.7.0"
```
dianabarsan, line 62: all these dependencies should come from the main package

admin-tool/tsconfig.json
```diff
@@ -0,0 +1,45 @@
...
+      "@admin-tool-components/*": ["src/ts/components/*"],
+      "@admin-tool-directives/*": ["src/ts/directives/*"],
+      "@admin-tool-effects/*": ["src/ts/effects/*"],
+      "@admin-tool-environments/*": ["src/ts/environments/*"],
+      "@admin-tool-modules/*": ["src/ts/modules/*"],
+      "@admin-tool-pipes/*": ["src/ts/pipes/*"],
+      "@admin-tool-providers/*": ["src/ts/providers/*"],
+      "@admin-tool-reducers/*": ["src/ts/reducers/*"],
+      "@admin-tool-selectors/*": ["src/ts/selectors/*"],
+      "@admin-tool-services/*": ["src/ts/services/*"]
```
dianabarsan, line 30: These are TBD, I think we should add them as they are required. So far, I think only services would be needed. it looks like a lot of these paths don't even exist.

admin-tool/tests/karma/karma-unit.base.conf.js
javier-perez: @dianabarsan I think we may need to duplicate it because they are separate applications, doing an import outside of the root directory could bring maintainability issues in the future. One option is to create a library with code we could share between the 2 applications but it requires extra coordination.

I think having both application as separate boxes could be better for the long term, what do you think?

admin-tool/package.json
javier-perez: Thanks @dianabarsan , I remove it from here.

Do we need to modify something else so the global linter checks this project?

admin-tool/package.json
javier-perez: @dianabarsan , the skeleton took what we have in the webapp right now, are you ok if create a separate PR to upgrade angular to 20, mainly because it could take extra changes that could make this PR harder to follow

admin-tool/src/ts/app.config.ts
javier-perez: @dianabarsan, what would you suggest instead? signals?
 if at some point we need global state management should we put it back ?
 
 We can propose a different way to handle state management and also fetching data from the API

admin-tool/tsconfig.json
javier-perez: I will review this again, it maybe they were created because of the folder structure in webapp

admin-tool/package.json
dianabarsan: ok, that's fine.

admin-tool/package.json
dianabarsan: no, I think the lint configuration will just cover everything by default.

admin-tool/package.json
dianabarsan: this comment still applies. the admin-tool should use all dev dependencies from the root package. there might be angular specific exceptions once the versions are different, but the rest of the testing should use the same dependencies.

admin-tool/src/ts/app.config.ts
javier-perez: @dianabarsan we will go with components state and inject services as providers

admin-tool/package.json
javier-perez: thanks @dianabarsan we will change this in our next PR, once it is integrated to the mono repo

admin-tool/package.json
javier-perez: We will remove this library

admin-tool/package.json
javier-perez: Since the project is not integrated to the parent project in this PR, can we remove it in the next PR? or do you prefer for me to pass the build integration in this branch as well?

admin-tool/package.json
javier-perez: Removed!

admin-tool/src/ts/app.config.ts
javier-perez: Removed!

admin-tool/package.json
dianabarsan: I would rather we don't follow a model where we add things just to remove them later. is there a reason we would include them in the first place?

admin-tool/tests/karma/karma-unit.base.conf.js
dianabarsan: Ok. I understand. let's keep them separate for now, and will reevaluate once updated admin matures.

admin-tool/tsconfig.json
dianabarsan: most likely, I think Claude eagerly added all these mappings.

admin-tool/package.json
javier-perez: Yes, I will pass this changes in the next PR, without that we were not going to be able to run the app locally because the app was not fully integrated with the build process

admin-tool/tsconfig.json
javier-perez: It makes sense

PR #10717, opened 2026-03-16
feat(7369): authentication

javier-perez:
# feat(7369): Authentication

# Description
Full authentication stack to the admin-tool. This included four new services (LocationService, SessionService, ChtDatasourceService, AuthService),  an AppRouteGuardProvider for route-level permission checks, and an [mmAuth]  directive for template-based conditional rendering. The app config was updated with an APP_INITIALIZER to run session.init() on startup, and routes now declare required permissions via a withGuard() helper. Browser compatibility for @medic/cht-datasource required adding a process shim in polyfills.ts and webpack/karma fallbacks for Node built-ins (os, zlib, http, https). The sidebar was updated to use the auth directive to conditionally show nav links based on user permissions. M2 shipped with 150 tests and 100% function coverage.

image

# Code review checklist

- [x] AI disclosure: Design by humans, written by Claude based on the webapp application

# License

### DIFF
--- a/admin-tool/custom-webpack.config.js
+++ b/admin-tool/custom-webpack.config.js
@@ -7,6 +7,10 @@ module.exports = {
     fallback: {
       path: false,
       fs: false,
+      os: false,
+      zlib: false,
+      http: false,
+      https: false,
     }
   }
 };
--- a/admin-tool/package-lock.json
+++ b/admin-tool/package-lock.json
@@ -17,6 +17,8 @@
         "@angular/platform-browser": "19.1.7",
         "@angular/platform-browser-dynamic": "19.1.7",
         "@angular/router": "19.1.7",
+        "@medic/cht-datasource": "file:../shared-libs/cht-datasource",
+        "@medic/constants": "file:../shared-libs/constants",
         "@messageformat/core": "^3.4.0",
         "@ngx-translate/core": "^16.0.4",
         "bootstrap": "^3.4.1",
@@ -34,6 +36,21 @@
         "@angular/compiler-cli": "19.1.7",
         "@angular/localize": "19.1.7"
       }
+    },
+    "../shared-libs/cht-datasource": {
+      "name": "@medic/cht-datasource",
+      "version": "1.0.0",
+      "hasInstallScript": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@medic/contact-types-utils": "file:../contact-types-utils",
+        "@medic/logger": "file:../logger"
+      }
+    },
+    "../shared-libs/constants": {
+      "name": "@medic/constants",
+      "version": "1.0.0",
+      "license": "Apache-2.0"
     },
     "node_modules/@ampproject/remapping": {
       "version": "2.3.0",
@@ -4040,6 +4057,14 @@
       "os": [
         "win32"
       ]
+    },
+    "node_modules/@medic/cht-datasource": {
+      "resolved": "../shared-libs/cht-datasource",
+      "link": true
+    },
+    "node_modules/@medic/constants": {
+      "resolved": "../shared-libs/constants",
+      "link": true
     },
     "node_modules/@messageformat/core": {
       "version": "3.4.0",
@@ -11323,6 +11348,16 @@
       "license": "ISC",
       "engines": {
         "node": "^18.17.0 || >=20.5.0"
+      }
+    },
+    "node_modules/process": {
+      "version": "0.11.10",
+      "resolved": "https://registry.npmjs.org/process/-/process-0.11.10.tgz",
+      "integrity": "sha512-cdGef/drWFoydD1JsMzuFf8100nZl+GT+yacc2bEced5f9Rjk4z+WtFUTBu9PhOi9j/jfmBPu0mMEY4wIdAF8A==",
+      "dev": true,
+      "license": "MIT",
+      "engines": {
+        "node": ">= 0.6.0"
       }
     },
     "node_modules/process-nextick-args": {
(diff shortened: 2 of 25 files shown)


dianabarsan (changes requested):
This looks like it just duplicates webapp services in the admin-tool. We would ideally not do this. 
Can we experiment with just requiring the services from webapp directly? 

For example, by adding this in admin-tool/tsconfig.json:
```
{
  "compilerOptions": {
    "paths": {
....< admin tool paths > ....
      "@mm-actions/*": ["../webapp/src/ts/actions/*"],
      "@mm-components/*": ["../webapp/src/ts/components/*"],
      "@mm-directives/*": ["../webapp/src/ts/directives/*"],
      "@mm-effects/*": ["../webapp/src/ts/effects/*"],
      "@mm-environments/*": ["../webapp/src/ts/environments/*"],
      "@mm-modals/*": ["../webapp/src/ts/modals/*"],
      "@mm-modules/*": ["../webapp/src/ts/modules/*"],
      "@mm-pipes/*": ["../webapp/src/ts/pipes/*"],
      "@mm-providers/*": ["../webapp/src/ts/providers/*"],
      "@mm-reducers/*": ["../webapp/src/ts/reducers/*"],
      "@mm-selectors/*": ["../webapp/src/ts/selectors/*"],
      "@mm-services/*": ["../webapp/src/ts/services/*"]
    }
  }
}
```

I had to make some other minor changes, like fixing polyfills, but the app builds with imported webapp services and directives. 

unfortunately I can't push my changes to your branch, but I will push them to another branch: https://github.com/medic/cht-core/commit/71040b8a4541b57553fa0e4b164e8ae269c6c233

javier-perez:
> This looks like it just duplicates webapp services in the admin-tool. We would ideally not do this. Can we experiment with just requiring the services from webapp directly?
> 
> For example, by adding this in admin-tool/tsconfig.json:
> 
> ```
> {
>   "compilerOptions": {
>     "paths": {
> ....< admin tool paths > ....
>       "@mm-actions/*": ["../webapp/src/ts/actions/*"],
>       "@mm-components/*": ["../webapp/src/ts/components/*"],
>       "@mm-directives/*": ["../webapp/src/ts/directives/*"],
>       "@mm-effects/*": ["../webapp/src/ts/effects/*"],
>       "@mm-environments/*": ["../webapp/src/ts/environments/*"],
>       "@mm-modals/*": ["../webapp/src/ts/modals/*"],
>       "@mm-modules/*": ["../webapp/src/ts/modules/*"],
>       "@mm-pipes/*": ["../webapp/src/ts/pipes/*"],
>       "@mm-providers/*": ["../webapp/src/ts/providers/*"],
>       "@mm-reducers/*": ["../webapp/src/ts/reducers/*"],
>       "@mm-selectors/*": ["../webapp/src/ts/selectors/*"],
>       "@mm-services/*": ["../webapp/src/ts/services/*"]
>     }
>   }
> }
> ```
> 
> I had to make some other minor changes, like fixing polyfills, but the app builds with imported webapp services and directives.
> 
> unfortunately I can't push my changes to your branch, but I will push them to another branch: [71040b8](https://github.com/medic/cht-core/commit/71040b8a4541b57553fa0e4b164e8ae269c6c233)

@dianabarsan , we normally keep the application code separate from external dependencies, if we need to do so we install a shared library instead of referencing external code

1. what if we need to add new methods to the services that are only available for the admin tool? should we add it to the webapp?

2. We were wondering if instead you would prefer the admin-tool pages to live inside the webapp and we control access with an auth guard and roles?

Also, what if we need to add new methods to the services that are only available for the admin tool? should we add it to the webapp?

dianabarsan:
> @dianabarsan , we normally keep the application code separate from external dependencies, if we need to do so we install a shared library instead of referencing external code
> 
> 1. what if we need to add new methods to the services that are only available for the admin tool? should we add it to the webapp?
> 2. We were wondering if instead you would prefer the admin-tool pages to live inside the webapp and we control access with an auth guard and roles?
> 
> Also, what if we need to add new methods to the services that are only available for the admin tool? should we add it to the webapp?

Thanks so much for the quick reply @javier-perez ! 

> 1. what if we need to add new methods to the services that are only available for the admin tool? should we add it to the webapp?

In this case, a new service that extends the webapp service should be added in the admin-tool. 

> 2. We were wondering if instead you would prefer the admin-tool pages to live inside the webapp and we control access with an auth guard and roles?

This has the potential of creating huge merge conflicts once the two codebases progress. I know @jkuester suggested this, but I don't see any definitive advantages. If reusing webapp code is impossible somehow, that this would be preferred to duplicating code.

javiertekdatum:
> > @dianabarsan , we normally keep the application code separate from external dependencies, if we need to do so we install a shared library instead of referencing external code
> > 
> > 1. what if we need to add new methods to the services that are only available for the admin tool? should we add it to the webapp?
> > 2. We were wondering if instead you would prefer the admin-tool pages to live inside the webapp and we control access with an auth guard and roles?
> > 
> > Also, what if we need to add new methods to the services that are only available for the admin tool? should we add it to the webapp?
> 
> Thanks so much for the quick reply @javier-perez !
> 
> > 1. what if we need to add new methods to the services that are only available for the admin tool? should we add it to the webapp?
> 
> In this case, a new service that extends the webapp service should be added in the admin-tool.
> 
> > 2. We were wondering if instead you would prefer the admin-tool pages to live inside the webapp and we control access with an auth guard and roles?
> 
> This has the potential of creating huge merge conflicts once the two codebases progress. I know @jkuester suggested this, but I don't see any definitive advantages. If reusing webapp code is impossible somehow, that this would be preferred to duplicating code.

Thank you @dianabarsan , my concern is having a hard dependency from one app to the other one, are you using this approach in other areas in the monorepo? 

I think extending the services is good for now, ideally a shared library approach could resolved the requirements( DRY, rapid development, avoid circular depencencies, inadvertent coupling )

dianabarsan:
Thanks for the suggestion @javiertekdatum 

I believe we are completely fine with having hard dependency on Webapp, because it makes it easier to maintain this code and this has been the pattern of how these two apps co=existed.

Issue #10302, opened 2025-09-17
Remove app management configure targets page


mrjones-plip:
**Describe the issue**
There is an old, orphaned page in app management that allows you to configure targets (`/admin/#/targets`).  While this may work, it will overwrite the local app settings file on the server, which will then in turn get overwritten when CD runs or when CHT Conf is manually run.  This is a worst practice.

**Describe the improvement you'd like**
We should remove this page

**Describe alternatives you've considered**
We could maybe try to encourage two way sync of app settings server -> client and also server <- client

ShivamGupta2003:
Hi @mrjones-plip,  
I’d like to work on this issue. Could you please assign it to me?

Samriddhi743:
Hi @mrjones-plip, I would like to solve this issue. Could you please assign it to me?

jkuester:
@ShivamGupta2003 are you still working on this?

ABHISHEKGORE2305:
Hi @mrjones-plip,
I’d like to work on this issue. Could you please assign it to me?

AmirSaudagar55:
I’m interested — can I be assigned this issue or is Abhishek working on it?

mrjones-plip:
Thanks for your interest @AmirSaudagar55!  Indeed ABHISHEKGORE2305 is working on this.

ABHISHEKGORE2305:
yes @AmirSaudagar55! I am working on it 👍

Carbonautics:
@andrablaj Can I work on this?

Carbonautics:
@andrablaj I see there are mentions of this in cht-docs as well, do you want me to remove it from cht-docs as well within this issue, or should we create a separate issue in cht-docs?

(Currently, I have linked both draft PRs linked to this issue, so please let me know)

andrablaj:
@jkuester, would you be available to advise on @Carbonautics's question above? Thank you!

Carbonautics:
I have raised #10721 either way. Please have someone review it. Meanwhile we can decide about the cht-docs, will keep that PR as a draft until decided.

jkuester:
@Carbonautics 👍 good call on the docs update!  Feel free to mark that draft PR as ready to review whenever you are ready for me to have a look at it.

I am running out of time this week, but I will have a look at both PRs next week!

PR #10721, opened 2026-03-18
feat(#10302): remove targets page from App Management

Carbonautics:
# Description

- Removes the Targets section from the `/admin/#/targets` location. 
- Also removes the Targets edit (Edit Target section) which was a sub-page of the targets in the admin app.
- Remove all references/translations used in admin targets page.

Closes: #10302 
Docs PR: https://github.com/medic/cht-docs/pull/2149

# Code review checklist

# License


admin/src/js/controllers/main.js
```diff
@@ -41,11 +40,4 @@ angular.module('controllers').controller('MainCtrl',
     $scope.logout = function() {
       Session.logout();
     };
-    $scope.checkActive = state => {
```
jkuester, line 44: It looks like this `checkActive` function is still used in other template htmls.  I think we should put this function back, but just remove the target-specific if-block.

admin/src/templates/targets.html
```diff
@@ -1,50 +0,0 @@
-<div class="loader" ng-if="loading" mm-auth="can_configure"></div>
-
-<div ng-if="!loading" mm-auth="can_configure">
-
-  <p ng-if="error" class="alert alert-danger" role="alert" translate>error.settings.loading</p>
```
jkuester, line 5: Good call on dropping the `admin.targets.description` translation here.  However, I think there are a number of other keys here that are also not being used (such as `error.settings.loading`) and can be removed from the translation files.

admin/src/templates/targets.html
```diff
@@ -1,50 +0,0 @@
...
-
-  <div ng-if="!error">
-
-    <p translate>admin.targets.description</p>
-
-    <div class="tab-content">
-      <div class="btn-toolbar">
-        <a class="btn btn-primary" ui-sref="targets-edit">
-          <i class="fa fa-plus"></i>
-          <span translate>analytics.target.add</span>
```
jkuester, line 15: I think this translation key can also be removed.

admin/src/templates/targets.html
```diff
@@ -1,50 +0,0 @@
...
-    </div>
-
-    <div class="tab-content container-fluid">
-
-      <div class="row selection-heading">
-        <div class="col-xs-1"></div>
-        <div class="col-xs-4" translate>analytics.target.id</div>
-        <div class="col-xs-4" translate>analytics.target.name</div>
-        <div class="col-xs-2" translate>analytics.target.type</div>
-        <div class="col-xs-1" translate>analytics.target.goal</div>
```
jkuester, line 27: I think these translation keys can also be removed.

admin/src/templates/targets.html
```diff
@@ -1,50 +0,0 @@
...
-      </div>
-
-      <div class="row selection" ng-repeat="target in configuration.items track by target.id" ng-click="edit(target.id)">
-        <div class="col-xs-1" ng-bind-html="target.icon | resourceIcon"></div>
-        <div class="col-xs-4">{{target.id}}</div>
-        <div class="col-xs-4" ng-if="target.translation_key">{{target.translation_key | translate}}</div>
-        <div class="col-xs-4" ng-if="!target.translation_key">{{target.name | translateFrom}}</div>
-        <div class="col-xs-2" ng-if="target.type === 'count'" translate>
-          analytics.target.type.count
```
jkuester, line 36: I think this translation key can also be removed.

admin/src/templates/targets.html
```diff
@@ -1,50 +0,0 @@
...
-        </div>
-        <div class="col-xs-2" ng-if="target.type === 'percent'" translate>
-          analytics.target.type.percent
```
jkuester, line 39: I think this translation key can also be removed.

admin/src/templates/targets_edit.html
```diff
@@ -1,47 +0,0 @@
...
-
-  <div class="form-group required" ng-class="{'has-error': errors.id}">
-    <label for="id" translate>analytics.target.id</label>
-    <input id="id" type="text" class="form-control" ng-model="target.id" ng-readonly="editing"/>
-    <span class="help-block" ng-if="errors.id" translate>{{errors.id}}</span>
-  </div>
-
-  <div class="form-horizontal">
-    <label translate>analytics.target.name</label>
-    <span class="help-block" translate>analytics.target.name.help</span>
```
jkuester, line 11: I think this translation key can also be removed.

admin/src/templates/targets_edit.html
```diff
@@ -1,47 +0,0 @@
...
-    <div class="form-group col-xs-12" ng-repeat="name in names">
-      <label class="col-xs-4">{{name.locale.name}}</label>
-      <div class="col-xs-8">
-        <input type="text" class="form-control" ng-model="name.content" />
-      </div>
-    </div>
-  </div>
-
-  <div class="form-group clear-both">
-    <label for="icon"><span translate>analytics.target.icon</span></label>
```
jkuester, line 21: I think this translation key can also be removed.

admin/src/templates/targets_edit.html
```diff
@@ -1,47 +0,0 @@
...
-    <select id="icon" class="form-control small" ng-model="target.icon" ng-options="icon as icon for icon in icons">
-    <option></option>
-    </select>
-  </div>
-
-  <div class="form-group">
-    <label for="type" translate>analytics.target.type</label>
-    <select id="type" class="form-control small" ng-model="target.type">
-      <option value="count" translate>analytics.target.type.count</option>
-      <option value="percent" translate>analytics.target.type.percent</option>
```
jkuester, line 31: I think these translation keys can also be removed.

admin/src/templates/targets_edit.html
```diff
@@ -1,47 +0,0 @@
...
-    </select>
-  </div>
-
-  <div class="form-group required" ng-class="{'has-error': errors.goal}">
-    <label for="goal" translate>analytics.target.goal</label>
-    <input id="goal" type="text" class="form-control small" ng-model="target.goal" />
-    <span class="help-block" translate>analytics.target.goal.help</span>
```
jkuester, line 38: I think this translation key can also be removed.

admin/src/templates/targets_edit.html
jkuester: Also looks like there is a `analytics.target.icon.help` key that is not used anywhere. Can remove that one too!

jkuester: Another one that does not seem to be used anywhere is `analytics.targets.unique.id`.

jkuester (changes requested):
Thanks for the PR @Carbonautics!  I left a few minor requests (mostly around cleaning up some additional translation keys).  Then, this should be good to go!

Carbonautics:
> Thanks for the PR @Carbonautics! I left a few minor requests (mostly around cleaning up some additional translation keys). Then, this should be good to go!

@jkuester Have updated as per comments. Please have a look again.

jkuester (approved):
LGTM!  🚀

PR #10727, opened 2026-03-22
feat(#10694): adds tasks page free text search functionality

### LINKED ISSUE #10694: Implement free text search feature in tasks page
In CHT 5.1.0, we added task filtering by due date, form and place. To further enhance user agency in filtering the task list, we can implement free text search similar to reports. 

Implement the search bar in the tasks page. 
At the least the search should be able to cover any contact name in the task lineage. The search would be done directly on the available task list, without accessing the database. 
This means that we need to implement search terms for tasks and a fuzzy matching for names. 

**Additional context**
Part of smart task prioritization.

shashi-sah2003:
# Description

Closes #10694 

# Code review checklist

# License


shashi-sah2003:
@dianabarsan I have verified this using unit tests (wasn't able to verify through web due to specs issue). could you try this using web and lmk if anything else is required?

dianabarsan:
Hi @shashi-sah2003 
This is much aligned with what I was thinking. But I think we need to refine the search a bit, right now I'm getting some pretty strange results: 
image

I searched for "mos" expecting to get the first task in the list, but then the rest are just strange results (I can't even tell where the search term matches). 
Now that I checked the code a bit more, it seems that you're just breaking the search term into letters and expecting to find the sequence of letters. So if I search for "phv", it will match all "pregnancy home visits". I don't think we should do this, I think we should use a reputable library (this just popped: https://www.npmjs.com/package/fuzzy-matching) and not rewrite this code. 

Granted, we should check that the library is fast!

shashi-sah2003:
@dianabarsan I have used this library https://www.npmjs.com/package/fuse.js (it looks more good). pushed the changes do confirm it once if it needs more refining

webapp/src/ts/selectors/index.ts
```diff
@@ -1,13 +1,58 @@
...
 
+const normalizeText = (value?: string): string => {
+  if (!value) {
+    return '';
+  }
+  return value
+    .toString()
+    .toLowerCase()
+    .normalize('NFD')
+    .replace(/[\u0300-\u036f]/g, '')
```
dianabarsan, line 23: Can you please add a comment about what this does exactly?

webapp/src/ts/selectors/index.ts
```diff
@@ -1,13 +1,58 @@
...
+    .trim();
+};
+
+const taskMatchesSearch = (task: TaskWithLineage, search: string): boolean => {
+  if (!search?.trim()) {
+    return true;
+  }
+
+  const normalizedSearch = normalizeText(search);
```
dianabarsan: I think this can also return an empty string right? Also, can this be normalized one time, and passed as normalized to the function? so we don't repeat this for every task.

webapp/src/ts/selectors/index.ts
```diff
@@ -1,13 +1,58 @@
...
+  if (substringMatch) {
+    return true;
+  }
+
+  const fuse = new Fuse(candidates, {
+    threshold: 0.2,       
+    distance: 50,
+    minMatchCharLength: 3, 
+    ignoreLocation: true,
+  });
```
dianabarsan, line 56: can you please add a comment about what these parameters mean and how they were chosen?

webapp/tests/karma/ts/selectors/index.spec.ts
```diff
@@ -609,6 +609,64 @@ describe('Selectors', () => {
...
+              lineage: ['Village Alpha'],
+              lineageIds: ['contact1']
+            },
+          ],
+        };
+
+        const globalState = { filters: { search: 'elodie' } } as any;
+        const result = Selectors.getFilteredTasksList.projector(tasksState, globalState);
+        expect(result).to.deep.equal([tasksState.tasksList[0]]);
+      });
```
dianabarsan, line 702: We support both Nepali and Arabic characters in our app. Can you please add two tests that demonstrate how the search work when the task list and the search query both include either nepali and arabic characters?

dianabarsan (changes requested):
Very nice! I added a few comments below. Also to get this merge ready, we need to add some e2e testing. Thanks a lot for building with us!

shashi-sah2003:
@dianabarsan I have addressed all the comments.

dianabarsan (changes requested):
Thanks @shashi-sah2003 . Code looks good! 

Can you please add e2e testing to support this feature? Thanks!

dianabarsan:
hey @shashi-sah2003 

I was wondering if you're available to continue work on this. We're really eager about having this feature ready for the next release. Thanks!

YASHSHARMAOFFICIALLY:
Hi @dianabarsan @andrablaj  , basesd on today meeting discussion I've picked up this work and addressed the outstanding feedback  Changes include a performance fix for the Fuse.js instantiation, doc comments on the helper functions and config parameters, and e2e test coverage. Thanks @shashi-sah2003 for the initial implementation.

YASHSHARMAOFFICIALLY:
Hi @dianabarsan, I've picked up this work and addressed the outstanding feedback in #11015 — performance fix for the Fuse.js instantiation, doc comments on the helper functions and config parameters, and e2e test coverage. Thanks @shashi-sah2003 for the initial implementation.

shashi-sah2003:
> hey @shashi-sah2003
> 
> I was wondering if you're available to continue work on this. We're really eager about having this feature ready for the next release. Thanks!

hey @dianabarsan apology I completed missed this. It seems like @YASHSHARMAOFFICIALLY has raised another PR, lmk if you need any help from my side :).

dianabarsan:
@shashi-sah2003 I also missed this. Your contribution is very welcome, so if you have bandwidth for this, it is most encouraged. Thank you!

shashi-sah2003:
@dianabarsan I have added e2e tests, might need your help in verifying these

tests/e2e/default/tasks/tasks-search.wdio-spec.js
```diff
@@ -0,0 +1,104 @@
...
+    await commonPage.sync();
+  });
+
+  after(async () => {
+    await utils.deleteUsers([ chw ]);
+    await utils.revertDb([/^form:/], true);
+    await utils.revertSettings(true);
+  });
+
+  it('filters tasks by contact name (diacritics-insensitive) and clears search', async () => {
```
dianabarsan, line 70: I hope I'm not being too pedantic, but it would be awesome if we would show that any place name in the contact's lineage can be used in the search.

tests/e2e/default/tasks/tasks-search.wdio-spec.js
```diff
@@ -0,0 +1,104 @@
...
+      patientPlain,
+    ]);
+    await utils.createUsers([ chw ]);
+    await sentinelUtils.waitForSentinel();
+
+    // Compile task configuration via API before login.
+    // Avoid `sync: true` here because that requires interacting with the UI hamburger menu.
+    await tasksPage.compileTasks('tasks-breadcrumbs-config.js', false);
+    await loginPage.login(chw);
+    await commonPage.tabsSelector.analyticsTab().waitForDisplayed();
```
dianabarsan, line 60: What is the purpose of this wait? What is the purpose of sync immediately after login?

tests/e2e/default/tasks/tasks-search.wdio-spec.js
```diff
@@ -0,0 +1,104 @@
...
+  });
+
+  it('filters tasks by contact name (diacritics-insensitive) and clears search', async () => {
+    await commonPage.goToTasks();
+
+    const initialCount = await tasksPage.getTaskListCount();
+    expect(initialCount).to.be.greaterThan(0);
+
+    await searchPage.performSearch('elodie');
+    await browser.waitUntil(async () => (await tasksPage.getTaskListCount()) !== initialCount, {
```
dianabarsan, line 77: I believe you can use `await tasksPage.waitForTaskListCountChange(initialTaskCount);` here

tests/page-objects/default/reports/reports.wdio.page.js
```diff
@@ -61,7 +61,7 @@ const reviewDialogSelectors = {
 };
 
 const sidebarFilterSelectors = {
-  openBtn: () => $('mm-search-bar .open-filter'),
+  openBtn: () => $('mm-search-bar .open-filter, .mm-search-bar-container .btn.open-filter'),
```
dianabarsan, line 64: Can you explain what is the purpose of this change?

webapp/src/ts/selectors/index.ts
```diff
@@ -1,13 +1,63 @@
...
+    threshold: 0.2,       
+    // distance: limits how far from the start of the string a match can be found.
+    distance: 50,
+    // Require at least 3 characters to trigger fuzzy matching,
+    minMatchCharLength: 3, 
+    // ignoreLocation: true means the match can appear anywhere in the string,
+    ignoreLocation: true,
+  });
+
+  return fuse.search(normalizedSearch).length > 0;
```
dianabarsan, line 58: Can we use a two-pass strategy? First pass: substring match across all tasks. Second pass: build one shared Fuse index over the remaining unmatched candidates and run a single search. The fuzzy pass is skipped entirely when query.length < 3 or there are no remaining tasks.
This will make the search more performant because only one Fuse instance gets created.

webapp/src/ts/selectors/index.ts
```diff
@@ -26,6 +76,13 @@ const applyTasksFilters = (tasks: TaskWithLineage[], filters: TasksFilters = {})
     });
   }
 
+  if (filters.search) {
```
dianabarsan, line 79: It seems that sonar is complaining about this.

dianabarsan (changes requested):
Sweet, thanks for the quick turnaround! I left some comments inline.

webapp/src/ts/selectors/index.ts
YASHSHARMAOFFICIALLY: Makes total sense — creating a Fuse instance per task is wasteful. Here's how I'd restructure this:

```typescript
const applySearchFilter = (tasks: TaskWithLineage[], search: string): TaskWithLineage[] => {
  const normalizedSearch = normalizeText(search);
  if (!normalizedSearch) {
    return tasks;
  }

  // Pass 1: substring match
  const substringMatched: TaskWithLineage[] = [];
  const unmatched: TaskWithLineage[] = [];

  for (const task of tasks) {
    const candidates = [
      task?.contact?.name,
      ...(task?.lineage || []),
      task?.title,
    ].filter(Boolean) as string[];

    if (candidates.some(c => normalizeText(c).includes(normalizedSearch))) {
      substringMatched.push(task);
    } else {
      unmatched.push(task);
    }
  }

  // Pass 2: fuzzy match on remaining (skip if query too short or nothing left)
  if (normalizedSearch.length < 3 || unmatched.length === 0) {
    return substringMatched;
  }

  const fuseItems = unmatched.map((task, idx) => ({
    idx,
    text: [task?.contact?.name, ...(task?.lineage || []), task?.title].filter(Boolean).join(' '),
  }));

  const fuse = new Fuse(fuseItems, {
    keys: ['text'],
    threshold: 0.2,
    distance: 50,
    minMatchCharLength: 3,
    ignoreLocation: true,
  });

  const fuzzyMatched = fuse.search(normalizedSearch).map(result => unmatched[result.item.idx]);
  return [...substringMatched, ...fuzzyMatched];
};
```

Single Fuse instance for the entire unmatched set, and fuzzy is skipped entirely for short queries. I can push this if you'd like @shashi-sah2003.

webapp/src/ts/selectors/index.ts
YASHSHARMAOFFICIALLY: This is likely the cognitive complexity from the nested filter + per-task Fuse instantiation. The two-pass refactor I suggested above should resolve this since we extract the search logic into its own function and eliminate the per-task Fuse creation.

tests/e2e/default/tasks/tasks-search.wdio-spec.js
YASHSHARMAOFFICIALLY: You're right — the `sync` call right after login is unnecessary here. The tasks get generated by the rules engine after the initial replication completes on login. We should just wait for the task list to populate instead of explicitly triggering sync. I'll remove this and rely on `waitForTaskListCountChange(0)` to wait for tasks to appear.

tests/e2e/default/tasks/tasks-search.wdio-spec.js
YASHSHARMAOFFICIALLY: Agreed, not pedantic at all — that's an important real-world scenario. I'll add a test case that searches by a parent place name (e.g. health center or district) from the contact's lineage and verifies the task shows up. That way we prove the lineage array is fully searchable.

tests/e2e/default/tasks/tasks-search.wdio-spec.js
YASHSHARMAOFFICIALLY: Good call, will switch to `waitForTaskListCountChange(initialTaskCount)` — cleaner and consistent with the existing test patterns.

tests/page-objects/default/reports/reports.wdio.page.js
YASHSHARMAOFFICIALLY: This was added because the tasks page uses a slightly different DOM structure for the search bar (`.mm-search-bar-container .btn.open-filter` vs the reports page's `mm-search-bar .open-filter`). But thinking about it more, this change shouldn't be in the reports page object at all — the tasks page has its own `sidebarFilterSelectors`. I'll revert this file, it was an oversight.

YASHSHARMAOFFICIALLY:
Hey @dianabarsan @shashi-sah2003 — I've gone through all the review comments and replied inline with my proposed approach for each. Summary of what needs to happen:

1. **Two-pass search strategy** — refactor `taskMatchesSearch` to do substring first across all tasks, then build a single shared Fuse index only for unmatched candidates. Fuzzy skipped when `query.length < 3`.
2. **SonarCloud fix** — the refactor above should resolve the cognitive complexity issue since search logic moves to a dedicated function.
3. **Remove unnecessary sync** in e2e setup — just wait for tasks to populate after login.
4. **Use `waitForTaskListCountChange`** instead of raw `browser.waitUntil`.
5. **Add lineage search test** — verify searching by a parent place name returns the right tasks.
6. **Revert reports page object** — that selector change doesn't belong there.

@shashi-sah2003 happy to push these changes directly if you want, or I can put them up on my branch (#11015) and we can consolidate. Let me know what works best for you both.

shashi-sah2003:
hey @YASHSHARMAOFFICIALLY you can go ahead and push your changes here lmk if you need my and @dianabarsan help anywhere

dianabarsan:
cc @cliftonmcintosh

This is the other PR that should be merged with this one: https://github.com/medic/cht-core/pull/11015

dianabarsan:
Closing in favor of https://github.com/medic/cht-core/pull/11142 where I cherry picked the commits from this branch.

PR #10728, opened 2026-03-23
feat(#10217): add telemetry tracking for external android app launches

### LINKED ISSUE #10217: Record telemetry for External Android App calls

megha1807:
## Summary
Adds performance telemetry tracking to the AndroidAppLauncherService 
to measure how long a user spends in an external Android app launched 
from a CHT form.

## Changes
- Added `PerformanceService` as a constructor dependency in `AndroidAppLauncherService`
- Added `tracking` variable to store the performance tracking object
- Start tracking when `launchAndroidApp()` is called
- Stop tracking when `resolveAndroidAppResponse()` is called with telemetry name `enketo:external-app:{action}`
- Updated unit tests to verify tracking behavior

## Related Issue
Fixes #10217
Docs PR: https://github.com/medic/cht-docs/pull/2166

## Notes
- Docs update will be submitted as a separate PR to cht-docs repo

### DIFF
--- a/webapp/tests/karma/ts/services/android-app-launcher.service.spec.ts
+++ b/webapp/tests/karma/ts/services/android-app-launcher.service.spec.ts
@@ -171,6 +171,15 @@ describe('AndroidAppLauncherService', () => {
           expect(consoleErrorMock.callCount).to.equal(1);
           expect(consoleErrorMock.args[0][1].message).to.equal('some error');
           expect(medicMobileAndroid.launchExternalApp.callCount).to.equal(1);
+          expect(medicMobileAndroid.launchExternalApp.args[0]).to.have.members([
+            'com.my-app.action.LOCATE',
+            null,
+            null,
+            '{"id":"1","location":{"city":"Tokyo"}}',
+            null,
+            null,
+            null,
+          ]);
           expect(trackingStub.stop.callCount).to.equal(0);
         });
     });


megha1807:
Hi @andrablaj , I am not able post anything in forum. I dont know why i am given read only access. Can you check it out?

andrablaj:
@megha1807, sorry to hear about the issue you are encountering with posting on the forum. Can you please share what your username is so I can check it? Thanks!

megha1807:
Hello @andrablaj 
My username is Megha

andrablaj:
Thanks @megha1807! Can you please retry - it should work now.

megha1807:
Thanks @andrablaj 
Yes now it is working.

webapp/src/ts/services/android-app-launcher.service.ts
```diff
@@ -19,7 +23,8 @@ export class AndroidAppLauncherService {
     if (!this.resolve) {
       return;
     }
-
+this.tracking?.stop({ name: `enketo:external-app:${response['action']}` });
```
jkuester: The `response['action']` is not the value we want to use here. Instead, we need the `chtAndroidApp.action` value from the `chtAndroidApp` parameter that was passed to `launchAndroidApp`.  

I think, instead of saving `tracking` as the class-level variable, lets instead lets just store a stop function that wraps this call here and includes the info we need.  

So, down in `launchAndroidApp`, instead of saving `this.tracking`, save a `this.stopTracking` function like this:

```ts
        const tracking = this.performanceService.track();
this.stopTracking = () => tracking?.stop({ name: `enketo:external-app:${chtAndroidApp.action}` });
```

Then, here in `resolveAndroidAppResponse`, all we should need to do is call:

```ts
this.stopTracking?.();
```

megha1807:
Hi @jkuester ,
Thank you for the detailed feedback! I have addressed all the changes:

1. Refactored `this.tracking` to `this.stopTracking` as a closure 
   function that captures `chtAndroidApp.action` correctly from 
   `launchAndroidApp()`, so the correct action value is used when 
   stopping the telemetry.

2. Fixed the linting error (missing newline at end of file).

Please re-review when you get a chance. Thank you!

megha1807:
Thanks for the thorough review and guidance @jkuester  throughout this PR! I learned a lot while working on this and really appreciate the support. Glad to see it merged.

PR #10737, opened 2026-03-23
chore(#10541): add DOC_IDS constants and replace magic strings in affected files

### LINKED ISSUE #10541: Add constants for 'target-aggregates', 'migration-log', 'extension-libs', and 'zscore-charts' document IDs

gmarav05:
# Description
Adds `TARGET_AGGREGATES`, `MIGRATION_LOG`, `EXTENSION_LIBS`, and `ZSCORE_CHARTS` constants to the `DOC_IDS` object in the shared constants library, replacing magic strings with named constants.

Closes #10541

# Code review checklist
- [ ] Readable: Concise, well named, follows the style guide
- [ ] AI disclosure: Used AI assistance for guidance on implementation

# License

### DIFF
--- a/shared-libs/constants/src/index.js
+++ b/shared-libs/constants/src/index.js
@@ -9,7 +9,6 @@ const DOC_IDS = {
   PRIVACY_POLICIES: 'privacy-policies',
   PARTNERS: 'partners',
   BRANDING: 'branding',
-  TARGET_AGGREGATES: 'target-aggregates',
   MIGRATION_LOG: 'migration-log',
   EXTENSION_LIBS: 'extension-libs',
   ZSCORE_CHARTS: 'zscore-charts',
--- a/webapp/tests/mocha/unit/validate_doc_update.spec.js
+++ b/webapp/tests/mocha/unit/validate_doc_update.spec.js
@@ -74,7 +74,7 @@ describe('validate doc update', () => {
       'service-worker-meta doc': { _id: DOC_IDS.SERVICE_WORKER_META },
       'forms': { type: 'form' },
       'translations': { type: DOC_TYPES.TRANSLATIONS },
-      'extension-libs': { _id: 'extension-libs' },
+      'extension-libs': { _id: DOC_IDS.EXTENSION_LIBS },
       'header logo': { _id: DOC_IDS.BRANDING },
       'partners': { _id: DOC_IDS.PARTNERS },
     }).forEach(([ name, doc ]) => {
@@ -101,9 +101,9 @@ describe('validate doc update', () => {
         newDoc: { _id: 'messages-en', type: 'feedback' }
       },
       {
-        name: 'extension-libs',
-        oldDoc: { _id: 'extension-libs' },
-        newDoc: { _id: 'extension-libs', field: 'mine' }
+        name: DOC_IDS.EXTENSION_LIBS,
+        oldDoc: { _id: DOC_IDS.EXTENSION_LIBS },
+        newDoc: { _id: DOC_IDS.EXTENSION_LIBS, field: 'mine' }
       },
       {
         name: DOC_IDS.BRANDING,


dianabarsan:
@gmarav05 as with https://github.com/medic/cht-core/pull/10747, the occurences of the magic string also need to be updated to use the constant. adding the constant is not sufficient.

gmarav05:
hello @sugat009 , The ci-integration-all-all failure is not related to this PR. Can you please re run this CI test.

webapp/src/ts/modules/analytics/analytics.routes.ts
```diff
@@ -32,7 +33,7 @@ export const routes:Routes = [
         canDeactivate: [ TrainingCardDeactivationGuardProvider ],
       },
       {
```
sugat009, line 34: **issue (blocking):** `DOC_IDS.TARGET_AGGREGATES` used as an Angular route path

The string `'target-aggregates'` here is a URL route segment, not a CouchDB document ID. Using `DOC_IDS.TARGET_AGGREGATES` conflates two unrelated concerns. If the document ID is ever renamed, this route would break, and URLs should remain stable independently of database schema.

The same issue applies to `analytics-modules.service.ts` (lines 5 and 31) and all the analytics test files that now use `DOC_IDS.TARGET_AGGREGATES` for route matching, module IDs, and navigation paths.

`AGGREGATE_TARGETS_ID` in `analytics-modules.service.ts` already provided the right abstraction as a local constant. It should stay as a hardcoded string:

```ts
export const AGGREGATE_TARGETS_ID = 'target-aggregates';
```

Only use `DOC_IDS.TARGET_AGGREGATES` where the string actually refers to a CouchDB document ID.

api/src/migrations.js
```diff
@@ -3,7 +3,8 @@ const fs = require('fs');
 const readdir = promisify(fs.readdir);
 const path = require('path');
 const db = require('./db');
-const MIGRATION_LOG_ID = 'migration-log';
+const { DOC_IDS } = require('@medic/constants');
+const MIGRATION_LOG_ID = DOC_IDS.MIGRATION_LOG;
```
sugat009, line 7: **suggestion (non-blocking):** Redundant local constant alias

The intermediate variable `MIGRATION_LOG_ID` just mirrors `DOC_IDS.MIGRATION_LOG`. You could use `DOC_IDS.MIGRATION_LOG` directly throughout the file and remove the alias. Same applies to `DOC_ID` in `api/src/services/extension-libs.js`.

That said, the existing code already used local constants, so preserving the pattern is fine too.

sugat009 (changes requested):
**praise:** Good effort adding DOC_IDS constants and replacing magic strings across 22 files. The scope is thorough for the document ID replacements.

Please see inline comments for file specific feedback. Additional items below:

---

**issue (non-blocking):** `validate_doc_update.spec.js` still has hardcoded `'extension-libs'`

`webapp/tests/mocha/unit/validate_doc_update.spec.js` (lines 77, 104, 105, 106) still uses the hardcoded string. Since this test exercises the CouchDB design document which itself cannot use `require('@medic/constants')`, keeping literal strings here is reasonable. Just flagging for awareness.

gmarav05:
Hello @sugat009, I have updated the blocking change (reverting TARGET_AGGREGATES from route and module usage) in the latest commit.

I think the `ci-integration-all-all failure` is not related to this PR and It has 2 failures.
1) A flaky doc_count mismatch in monitoring.spec.js:102.
2) 502 Bad Gateway nginx DNS error in server.spec.js:212. 

I think it will pass if we re-run them, Can you please re-run this check.

shared-libs/constants/src/index.js
```diff
@@ -9,6 +9,10 @@ const DOC_IDS = {
   PRIVACY_POLICIES: 'privacy-policies',
   PARTNERS: 'partners',
   BRANDING: 'branding',
+  TARGET_AGGREGATES: 'target-aggregates',
```
sugat009: **issue (blocking):** Remove `TARGET_AGGREGATES` from `DOC_IDS`

After the revert in 408c765, this constant is no longer used anywhere in this PR. Adding an unused constant to `@medic/constants` creates a maintenance question ("where is this used?"). It can be added back in a future PR when there is a genuine document ID usage for it.

shared-libs/constants/src/index.js
```diff
@@ -9,6 +9,10 @@ const DOC_IDS = {
...
+  MIGRATION_LOG: 'migration-log',
+  EXTENSION_LIBS: 'extension-libs',
+  ZSCORE_CHARTS: 'zscore-charts',
```
sugat009, line 14: **suggestion (non-blocking):** There are 4 remaining hardcoded `'extension-libs'` strings in `webapp/tests/mocha/unit/validate_doc_update.spec.js` (line 76, 103, 104, 105) that could also be replaced with `DOC_IDS.EXTENSION_LIBS` for consistency.

Note: the `validate_doc_update.js` ddoc itself intentionally keeps hardcoded strings since CouchDB design document functions cannot import external JS modules. But the spec file can use the constant.

gmarav05:
Thanks for reviewing ang giving a detailed feedback @sugat009,

I have removed unused TARGET_AGGREGATES from DOC_IDS and also replaced remaining hardcoded 'extension-libs' strings in validate_doc_update.spec.js with DOC_IDS.EXTENSION_LIBS. 

Please let me know if anything else needs any update.

gmarav05:
hello @sugat009, the ci-integration-all-all failure consists entirely of timeout of 200000ms exceeded errors in before each hooks across changes.spec.js, db-doc.spec.js, monitoring.spec.js and users.spec.js. None of which are touched by this PR. 

Can you please re-trigger this check.

sugat009:
@gmarav05 I've merged #10738 to main and there's a small conflict in this PR. Can you resolve it?

gmarav05:
> @gmarav05 I've merged #10738 to main and there's a small conflict in this PR. Can you resolve it?

Yeah sure, I will fix this ASAP and let you know.

gmarav05:
hello @sugat009, Can you please re-trigger this check again, earlier it passed now after update it is failing again.

PR #10742, opened 2026-03-23
fix(e2e): add null check to error target config

### LINKED ISSUE #10498: Multiple e2e workflow tests are failing due to feedback documents being generated
**Describe the issue**
Example run: https://github.com/medic/cht-core/actions/runs/19735777380
Multiple tests have failed here, and all with the same feedback doc error: 
```
 "arguments": [
      "\"Error selecting contact\"",
      "\"TypeError: Cannot read properties of undefined (reading 'muted')\\n    at Object.appliesIf (eval at initialize (https://localhost/main.js:1:1167910), <anonymous>:3:165)\\n    at r (eval at initialize (https://localhost/main.js:1:1167910), <anonymous>:3:884)\\n    at t.exports (eval at initialize (https://localhost/main.js:1:1167910), <anonymous>:3:1761)\\n    at eval (eval at initialize (https://localhost/main.js:1:1167910), <anonymous>:3:4795)\\n    at eval (eval at initialize (https://localhost/main.js:1:1167910), <anonymous>:3:4870)\\n    at _ (https://localhost/main.js:1:1167992)\\n    at Array.forEach (<anonymous>)\\n    at Object.getEmissionsFor (https://localhost/main.js:1:1169136)\\n    at u.exports (https://localhost/main.js:1:1166268)\\n    at https://localhost/main.js:1:1165170\""
    ],
    "info": {
      "message": "Cannot read properties of undefined (reading 'muted')",
      "stack": "TypeError: Cannot read properties of undefined (reading 'muted')\n    at Object.appliesIf (eval at initialize (https://localhost/main.js:1:1167910), <anonymous>:3:165)\n    at r (eval at initialize (https://localhost/main.js:1:1167910), <anonymous>:3:884)\n    at t.exports (eval at initialize (https://localhost/main.js:1:1167910), <anonymous>:3:1761)\n    at eval (eval at initialize (https://localhost/main.js:1:1167910), <anonymous>:3:4795)\n    at eval (eval at initialize (https://localhost/main.js:1:1167910), <anonymous>:3:4870)\n    at _ (https://localhost/main.js:1:1167992)\n    at Array.forEach (<anonymous>)\n    at Object.getEmissionsFor (https://localhost/main.js:1:1169136)\n    at u.exports (https://localhost/main.js:1:1166268)\n    at https://localhost/main.js:1:1165170"
    },
```

It looks like there some task (or target) that tries to access the `muted` property of a contact in the appliesIf function without nullchecking the original document. 

Fix the config.

sh1vam31:
# Description

## Fixes #10498

This PR resolves an issue where multiple e2e workflow tests were failing because of unexpected feedback documents being generated.

The root cause was a missing null check in the appliesIf function of the error target configuration (
tests/e2e/default/targets/config/targets-error-config.js). It was attempting to read .muted from an undefined property (c.foo), causing a TypeError, which triggered the feedback document generation.

This PR fixes the issue by using optional chaining (c.foo?.muted) to safely handle undefined properties.

# Code review checklist

- [x] AI disclosure: I used AI assistance to understand the issue and the codebase, then manually applied the fix and verified it works by running the unit tests locally.

# License

### DIFF
--- a/webapp/src/ts/effects/contacts.effects.ts
+++ b/webapp/src/ts/effects/contacts.effects.ts
@@ -180,10 +180,15 @@ export class ContactsEffects {
   }
 
   private async loadTargetDoc(contactId, trackName) {
+    const selected = this.selectedContact;
+    if (!selected?.doc) {
+      return;
+    }
+
     const trackPerformance = this.performanceService.track();
 
     const targetDocs = await this.targetAggregateService.getTargetDocs(
-      this.selectedContact,
+      selected,
       this.userFacilityIds,
       this.userContactId
     );
--- a/webapp/tests/karma/ts/effects/contacts.effects.spec.ts
+++ b/webapp/tests/karma/ts/effects/contacts.effects.spec.ts
@@ -608,6 +608,15 @@ describe('Contacts effects', () => {
         expect(setSnackbarContent.callCount).to.equal(0);
         expect(unsetSelected.callCount).to.equal(1);
       });
+
+      it('should handle missing contact doc', async () => {
+        contactViewModelGeneratorService.getContact.resolves({ _id: 'person', doc: undefined });
+
+        actions$ = of(ContactActionList.selectContact({ id: 'person' }));
+        await effects.selectContact.toPromise();
+
+        expect(targetAggregateService.getTargetDocs.callCount).to.equal(0);
+      });
     });
 
     describe('loading contact summary', () => {


sh1vam31:
Hi @dianabarsan, @sugat009, @mrjones-plip

Kindly review the PR.
Please let me know if any changes or improvements are needed.

dianabarsan (changes requested):
I believe i have already commented on the issue: https://github.com/medic/cht-core/issues/10498#issuecomment-4116603354

The underlying problem that causes bad data to be passed to config should be fixed, instead of the config handling bad data.

sh1vam31:
Hi @dianabarsan,

I investigated the race condition and found the exact same root cause as the Contact Summary issue (https://github.com/medic/cht-core/issues/10661).

During rapid UI navigation in the E2E tests, the frontend was passing an uninitialized contact (this.selectedContact) to the target engine, causing the targets-error-config.js to crash and generate feedback documents.

Updates in this PR:

Reverted: targets-error-config.js so it can continue acting as a dedicated error test scenario untouched.

Added a safety guard (if (!selected?.doc)) inside loadTargetDoc in webapp/src/ts/effects/contacts.effects.ts
to abort target generation if the contact data isn't fully ready.

This fixes the underlying race condition by preventing the system from ever processing the bad data

Kindly review this.

Please let me know if any changes or improvements are needed.

dianabarsan:
@sh1vam31 it looks like your linting is failing.

webapp/src/ts/effects/contacts.effects.ts
```diff
@@ -181,9 +181,16 @@ export class ContactsEffects {
 
   private async loadTargetDoc(contactId, trackName) {
     const trackPerformance = this.performanceService.track();
+    const selected = this.selectedContact;
+
+    if (!selected?.doc) {
+      this.contactsActions.receiveSelectedContactTargetDoc([]);
```
dianabarsan: I'm wondering if we should be calling this at all. We didn't receive any tasks, because we have no contact. What happens if this call is not made?

sh1vam31:
Hi @dianabarsan, I've updated the PR with your suggestions.

**Logic Refinement:** In loadTargetDoc, I removed the call to receiveSelectedContactTargetDoc([]). Now, if the contact data isn't ready, the function simply returns early as you suggested.


**Code Consistency & Safety**: I applied the same "early return" pattern to loadContactSummary and refactored it to use async/await. This ensures both functions handle missing contact data safely and consistently.

**Linting Fixes:** I've also refactored the test mocks in contacts.effects.spec.ts to fix the line-length linting errors that were causing CI to fail.

Thanks for the review.

webapp/src/ts/effects/contacts.effects.ts
```diff
@@ -207,25 +213,34 @@ export class ContactsEffects {
       });
   }
 
-  private loadContactSummary(contactId, trackName) {
-    const trackPerformance = this.performanceService.track();
+  private async loadContactSummary(contactId, trackName) {
```
dianabarsan, line 215: @sh1vam31 I believe you are pushing this change in two branches, this one and https://github.com/medic/cht-core/pull/10731/changes. 

To avoid conflicts, can you please remove this change here and only keep it in the other branch?

sh1vam31:
Hi @dianabarsan,

I've removed the duplicate changes in contacts.effects.ts (and its test) from this branch as you suggested, so they are now only kept in PR https://github.com/medic/cht-core/pull/10731.

I've also ensured the intended null check fix (c.foo?.muted) is correctly applied to targets-error-config.js in this PR.

Thanks for catching that duplication.

dianabarsan (approved):
cool!

dianabarsan:
I'll merge once the build is green!

.github/workflows/build.yml
```diff
@@ -199,6 +199,8 @@ jobs:
           sleep 2
 
       - uses: nolar/setup-k3d-k3s@v1
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
dianabarsan: This change is unrelated. Can you please remove it?

tests/integration/api/controllers/records.spec.js
```diff
@@ -33,7 +33,7 @@ describe('Import Records', () => {
         }
       }
     }
-  }, { ignoreReload: true }));
+  }));
```
dianabarsan: I believe we should still ignore reload here. is that still happening?

sh1vam31:
Hi @dianabarsan,

Thanks for catching those.

There were some accidental leftovers from another branch that slipped in. I've re-added ignoreReload: true and removed the token change, so this PR is now clean and strictly contains the E2E fix.

couchdb/tests/Makefile
```diff
@@ -9,8 +9,9 @@ export COUCHDB_USER
 wait-for-couchdb:
 	@echo "Waiting for CouchDB containers to be ready..."
 	@for i in 1 2 3; do \
-		until curl -s -f -u "$(COUCHDB_USER):$(COUCHDB_PASSWORD)" http://localhost:$${i}5984/_up > /dev/null; do \
-			echo "Waiting for couchdb$$i..."; \
+		if [ "$$i" = "3" ]; then port="31984"; else port="$${i}5984"; fi; \
```
dianabarsan: We still have unrelated changes :(

tests/integration/api/controllers/records.spec.js
dianabarsan: I think you accidentally reverted this commit.

dianabarsan (changes requested):

sh1vam31:
Hi @dianabarsan ,

Thanks for the update!

I’ve addressed all the comments. The unrelated CouchDB changes have been reverted, and { ignoreReload: true } has been restored as suggested.

The PR is now ready for review. Please let me know if anything else needs to be adjusted.

tests/e2e/default-mobile/reports/search.wdio-spec.js
```diff
@@ -49,6 +49,7 @@ describe('Search Reports', () => {
   it('should return results matching the search term and then return all data when clearing search', async () => {
     const [ hospitalSMS, healthCenterSMS, hospitalReport, healthCenterReport ] = reportDocs;
     await commonPage.goToReports();
+    await reportsPage.waitForReportsLoaded();
```
dianabarsan, line 52: It seems that these are some unrelated changes. Can you please revert and keep the PR about the target nullcheck please? (applies to the other changes as well)

dianabarsan (changes requested):

tests/integration/api/controllers/monitoring.spec.js
```diff
@@ -55,6 +55,20 @@ const getExpectedNouveauIndexes = (db) => {
   return Promise.all(indexes.map((name) => getExpectedNouveauIndex(db, name)));
 };
 
+const waitForNouveauStable = async (db) => {
```
dianabarsan: it still seems this is unrelated. please revert.

dianabarsan (changes requested):

sh1vam31:
Hi @dianabarsan,

I have reverted the unrelated monitoring test stabilisation. 

The PR now only contains the defensive null check in contacts.effects.ts

webapp/src/ts/effects/contacts.effects.ts
```diff
@@ -180,10 +180,14 @@ export class ContactsEffects {
   }
 
   private async loadTargetDoc(contactId, trackName) {
-    const trackPerformance = this.performanceService.track();
+    const selected = this.selectedContact;
```
dianabarsan, line 183: Can you please add some unit testing around this?

dianabarsan (changes requested):

sh1vam31:
Hi @dianabarsan,  @mrjones-plip 

I've added the unit test for the loadTargetDoc null check as requested. I've also rebased the branch on the latest master to keep everything up to date. 

Ready for another look.

webapp/tests/karma/ts/effects/contacts.effects.spec.ts
```diff
@@ -591,22 +591,13 @@ describe('Contacts effects', () => {
         expect(setSelectedContact.args[1][0]._id).to.equal('person');
       });
 
-      it('should handle errors when loading the target docs', async () => {
-        sinon.stub(console, 'error');
-        const setSnackbarContent = sinon.stub(GlobalActions.prototype, 'setSnackbarContent');
-        const unsetSelected = sinon.stub(GlobalActions.prototype, 'unsetSelected');
-        targetAggregateService.getTargetDocs.rejects({ code: 400 });
+      it('should handle missing contact doc', async () => {
```
dianabarsan, line 612: I think we still want the `should handle errors when loading the target docs` test check. is there a reason this test was overwritten? loading targets can fail for other reasons than a missing contact.

sh1vam31:
Thanks for the catch, @dianabarsan. I've restored the original should handle errors when loading the target docs test to ensure we still cover service-level failures. Both the original error-handling test and the new missing contact doc test are now included to cover both scenarios. I've also verified that the project passes linting locally.

dianabarsan (approved):
Nice one!

PR #10745, opened 2026-03-23
chore(#10704): add e2e test for saving form with repeat

### LINKED ISSUE #10704: Add e2e test that checks repeat fields are saved
**Describe the issue**
I was recently looking into the possibility of uploading photos in repeats. In the process, I used [this](https://github.com/medic/cht-core/blob/8ece077ea28effbd2d6895a4d801ef0f966c3088/tests/e2e/default/enketo/forms/repeat-translation-button.xml#L72) form used on our e2e tests, and was unpleasantly surprised that, upon saving, none of the repeat fields were saved. 

The [enketo repeats repeat e2e test](https://github.com/medic/cht-core/blob/b9ddc064d82158b65cd801eb8488ccbab4838de0/tests/e2e/default/enketo/repeat.wdio-spec.js#L8) only checks that the repeat fields are displayed, and not that they are saved. 
Update the form to be a correct form, and update the tests to also check submission and correct field being saved.

SnehaChaursia:
# Description

Fixes the enketo repeat form e2e test to verify that repeat fields are correctly saved on submission.

The `repeat-translation-button.xml` form had a bug where the instance root element was named `<repeat>` instead of `<repeat_translation>`. Since XForms requires the root tag to match the form id, submitted repeat fields were silently discarded and never saved to the document.

Changes:
- Delete `repeat-translation-button.xml` (broken form)
- Add `repeat_translation.xml`: corrected form with:
  - Root instance element renamed to `<repeat_translation>`
  - Added `name` text field inside repeat group for testability
  - Translations added for `name` label in both English (`Name:`) and Swahili (`Jina:`)
- Add `repeat_translation.xlsx` companion (source for cht-conf conversion)
- Update `repeat.wdio-spec.js`:
  - Load `repeat_translation` instead of `repeat-translation-button`
  - Update existing repeat-button tests to use `repeat_translation` form id
  - Add new test: fills in state + 1 repeat item (city + name), submits, then asserts `report.fields.basic.state_1`, `rep[0].city_1`, and `rep[0].name` are saved correctly

Fixes #10704 

# Code review checklist

# License

@dianabarsan  Please take a look on this PR and kindly let me if any further changes need.


tests/e2e/default/enketo/forms/repeat-translation-button.xml
jkuester, line 1: How did you get the contents of this file?  🤔   We need it to be the result of running `cht convert-app-forms`  from https://github.com/medic/cht-conf against the xlsx file.  (The xml file that I [uploaded to the issue](https://github.com/medic/cht-core/issues/10704#issuecomment-4112074087) should already have the correct contents...)

tests/e2e/default/enketo/forms/repeat_translation.xlsx
jkuester, line 1: Same here. Can just use the `xlsx` file I uploaded to the issue.

tests/e2e/default/enketo/repeat.wdio-spec.js
```diff
@@ -124,6 +125,32 @@ describe('RepeatForm', () => {
...
+      await genericForm.submitForm();
+
+      // Verify the saved document contains the repeat fields
+      const reportId = await reportsPage.getCurrentReportId();
+      const report = await utils.getDoc(reportId);
+      expect(report.fields.basic.state_1).to.equal('VIC');
+      expect(report.fields.basic.rep).to.be.an('array');
+      expect(report.fields.basic.rep.length).to.be.greaterThan(0);
+      expect(report.fields.basic.rep[0].city_1).to.equal('melbourne');
+      expect(report.fields.basic.rep[0].name).to.equal('Alice');
```
jkuester: Good assertions!  But, instead of having a whole separate test for them, can we just submit the form and assert the report in the two previous tests (and not have this additional test)?

jkuester (changes requested):
Thanks for the PR @SnehaChaursia!

tests/e2e/default/enketo/forms/repeat-translation-button.xml
SnehaChaursia: Hi @jkuester, sorry for the confusion in the previous commit!

I had mistakenly used a locally generated XML instead of the one from the issue — my bad. I've now replaced both the `repeat_translation.xml` and `repeat_translation.xlsx` with the correct files from the issue.

Also addressed the test structure — removed the standalone submission test and merged the `submitForm()` + field assertions into the two existing repeat-button tests (Swahili and English).

Please take a look. Let me know if anything else needs to be fixed!

PR #10746, opened 2026-03-24
chore(#10552): add constants for additional Sentinel metadata local document IDs

### LINKED ISSUE #10552: Add constants for Sentinel metadata (e.g., additional local document IDs)

gmarav05:
# Description
This PR adds `PURGE_LOG`, `PURGE_DB_INFO`, and `META_DATA` constants to the `SENTINEL_METADATA` object in the shared constants library, replacing hardcoded `_local/` document ID strings with named constants.

1) `PURGE_LOG` replaces `_local/purge_log` in `sentinel/src/schedule/replications.js`.
2) `PURGE_DB_INFO` replaces `_local/info` in `sentinel/src/lib/purging.js`.
3) `META_DATA` replaces `_local/sentinel-meta-data` in `api/src/migrations/extract-transition-seq.js`.

Closes #10552

# Code review checklist
- [ ] AI disclosure: Used AI assistance for guidance on implementation

# License

Screenshot 2026-03-24 at 11 25 45 AM


gmarav05:
hello @sugat009, earlier the CI failed due to the PR title and i updated that is this fine now.

sugat009:
Hey @gmarav05, the constants look correct! Same feedback as on #10747 though: the consumer files that currently hardcode these strings need to be updated to actually use the new constants.

**Source files to update:**

1. `sentinel/src/schedule/replications.js:7` — `const PURGE_LOG_ID = '_local/purge_log'` → use `SENTINEL_METADATA.PURGE_LOG`
2. `sentinel/src/lib/purging.js:47` — `'_local/info'` inline → use `SENTINEL_METADATA.PURGE_DB_INFO`
3. `api/src/migrations/extract-transition-seq.js:13` — `const METADATA_DOCUMENT = '_local/sentinel-meta-data'` → use `SENTINEL_METADATA.META_DATA` (this file already imports `SENTINEL_METADATA` on line 8, so just add `META_DATA` to the existing destructure)

**Dependency change needed:**

`sentinel/package.json` doesn't currently list `@medic/constants` as a dependency. You'll need to add it there before `replications.js` and `purging.js` can import from it.

**Test files to update:**

4. `sentinel/tests/unit/schedule/replications.spec.js` (lines 92, 115, 155, 215, 255, 264, 272, 278, 289, 337, 377, 426, 454, 517)
5. `sentinel/tests/unit/lib/purging.spec.js` (lines 146, 147, 148, 168, 169, 170, 184)
6. `api/tests/integration/migrations/extract-transition-seq.spec.js` (line 11)

Also, the commit message type should be `chore`, not `fix`, to match the PR title.

shared-libs/constants/src/index.js
```diff
@@ -29,6 +29,9 @@ const HTTP_HEADERS = {
 const SENTINEL_METADATA = {
   TRANSITIONS_SEQ: '_local/transitions-seq',
   BACKGROUND_SEQ: '_local/background-seq',
+  PURGE_LOG: '_local/purge_log',
+  PURGE_DB_INFO: '_local/info',
+  META_DATA: '_local/sentinel-meta-data',
```
sugat009: **suggestion (non-blocking):** `SENTINEL_METADATA.META_DATA` reads as "sentinel metadata's metadata" which is a bit tautological. The existing constants (`TRANSITIONS_SEQ`, `BACKGROUND_SEQ`) describe their semantic purpose. Since this is the legacy pre-migration document, something like `LEGACY_META_DATA` would be more self-documenting. The consumer in `extract-transition-seq.js` immediately aliases it to `METADATA_DOCUMENT`, which suggests the name isn't carrying its weight on its own.

api/src/migrations/extract-transition-seq.js
```diff
@@ -4,13 +4,14 @@
 //
 const db = require('../db');
 const logger = require('@medic/logger');
+const { SENTINEL_METADATA } = require('@medic/constants');
```
sugat009, line 7: **suggestion (non-blocking):** The two-step import pattern here is inconsistent with the pre-existing deep destructure pattern that was in this file. The integration test also uses a third style (direct `SENTINEL_METADATA.X` dot access). Could we keep the source and test pair consistent? Either the deep destructure or the two-step approach is fine, just pick one for both.

sentinel/src/schedule/replications.js
```diff
@@ -3,8 +3,9 @@ const db = require('../db');
 const logger = require('@medic/logger');
 const environment = require('@medic/environment');
 const request = require('@medic/couch-request');
+const { SENTINEL_METADATA } = require('@medic/constants');
 
-const PURGE_LOG_ID = '_local/purge_log';
+const PURGE_LOG_ID = SENTINEL_METADATA.PURGE_LOG;
```
sugat009, line 8: **suggestion (non-blocking):** `webapp/src/js/bootstrapper/purger.js` uses `_local/purgelog` (no underscore), which is a different document from `_local/purge_log` here. Could you add a short code comment here (e.g. `// Note: distinct from _local/purgelog used in webapp purger`) so future developers don't confuse the two?

sentinel/package.json
```diff
@@ -9,5 +9,8 @@
   },
   "scripts": {
     "run-watch": "TZ=UTC nodemon --inspect=0.0.0.0:9228 --watch server.js --watch 'src/**' --watch '../shared-libs/**/src/**' --watch '../shared-libs/**/dist/**' server.js"
+  },
+  "dependencies": {
+    "@medic/constants": "file:../shared-libs/constants"
```
sugat009, line 14: **suggestion (non-blocking):** `api/package.json` also imports `@medic/constants` (e.g. in `api/src/migrations/extract-transition-seq.js`) but doesn't declare it as a dependency. Since the PR touches that file, it might be a good opportunity to add the dependency there too, similar to what you've done here.

sugat009 (changes requested):
Awesome! Thanks for making changes where the constants are being used as well. A few nitpicks and then we should be ready to go!

shared-libs/constants/src/index.js
gmarav05: Thanks for the suggestion! That makes sense `LEGACY_META_DATA` is more descriptive. Now, I have updated the constant name in the shared constants and all its usages accordingly.

api/src/migrations/extract-transition-seq.js
gmarav05: Yes sure, Now I have updated the usage to follow the existing destructuring pattern in the source file and aligned the integration test to use the same approach.

sentinel/src/schedule/replications.js
gmarav05: Thanks for pointing that out. I have added a comment in `replications.js` to clarify the difference between `_local/purge_log` and `_local/purgelog` to avoid confusion in the future.

sentinel/package.json
gmarav05: thanks for the suggestion @sugat009. Now, I have added `@medic/constants` as a dependency in `api/package.json`.

gmarav05:
> Awesome! Thanks for making changes where the constants are being used as well. A few nitpicks and then we should be ready to go!

Thanks a lot for detailed review. I have addressed all the suggested changes, including updating constant usages, improving naming, aligning import styles and adding the missing dependency. 

Please let me know if any changes or updates are required.

api/tests/integration/migrations/extract-transition-seq.spec.js
```diff
@@ -1,14 +1,11 @@
...
-const {
-  SENTINEL_METADATA: {
-    TRANSITIONS_SEQ: TRANSITION_SEQ_DOCUMENT,
-    BACKGROUND_SEQ: BACKGROUND_CLEANUP_SEQ_DOCUMENT
-  }
-} = require('@medic/constants');
-const METADATA_DOCUMENT = '_local/sentinel-meta-data';
+const TRANSITION_SEQ_DOCUMENT = SENTINEL_METADATA.TRANSITIONS_SEQ;
+const BACKGROUND_CLEANUP_SEQ_DOCUMENT = SENTINEL_METADATA.BACKGROUND_SEQ;
+const { LEGACY_META_DATA: METADATA_DOCUMENT } = SENTINEL_METADATA;
```
sugat009: This comment was for all three of these imports: https://github.com/medic/cht-core/pull/10746#discussion_r2987200314

api/package.json
```diff
@@ -15,5 +15,8 @@
   "scripts": {
     "toc": "doctoc --github --maxlevel 2 README.md",
     "run-watch": "TZ=UTC nodemon -e js,json,html,properties --inspect=0.0.0.0:9229 --ignore 'build/static' --ignore 'build/public' --watch ./ --watch '../shared-libs/**/src/**' server.js -- --allow-cors"
+  },
+  "dependencies": {
+    "@medic/constants": "file:../shared-libs/constants"
   }
-}
+}
```
sugat009:
```suggestion
}

```

api/tests/integration/migrations/extract-transition-seq.spec.js
sugat009:
```typescript
const {                                                                        
    TRANSITIONS_SEQ: TRANSITION_SEQ_DOCUMENT,                                    
    BACKGROUND_SEQ: BACKGROUND_CLEANUP_SEQ_DOCUMENT,
    LEGACY_META_DATA: METADATA_DOCUMENT                                          
  } = SENTINEL_METADATA; 
```

gmarav05: Thanks for the clarification. I have updated the test file.

Please let me know if anything else needs adjustment.

gmarav05:
@sugat009 Thanks for the clarification. I have updated the test file and fixed the newline issue in `api/package.json`.

gmarav05:
> LGTM!
> 
> FYI: use this button when you need to re-request a review after a review.
> 

Sorry, @sugat009  I should have done that. I used it only once previously while contributing to other oss project. I will re-request from next time onwards. 🥹

Issue #8816, opened 2024-01-18
Fragile text search can invalidate CouchDb admins on container restart


dianabarsan:
**What feature do you want to improve?**
Referencing [this code](https://github.com/medic/cht-core/blob/aef8973a0f184f5e511bae7897f6b6a3e83eed2d/couchdb/docker-entrypoint.sh#L83)

```
if [ "$COUCHDB_USER" ] && [ "$COUCHDB_PASSWORD" ] && [ -z "$COUCHDB_SYNC_ADMINS_NODE" ]; then
        # Create admin only if not already present
        if ! grep -Pzoqr "\[admins\]\n$COUCHDB_USER =" $CLUSTER_CREDENTIALS; then
            printf "\n[admins]\n%s = %s\n" "$COUCHDB_USER" "$COUCHDB_PASSWORD" >> $CLUSTER_CREDENTIALS
        fi
    fi
```

This code requires that the env user is the first user entry in the admins section of the config file. 
If the section already exists, but does not match this text search, either because another admin is listed first or there are additional spaces (??), a new admin section will be created at the bottom of the file, invalidating the existing section - apparently CouchDb doesn't merge config, at least not within the same file.

**Describe the improvement you'd like**
Improve the user detection to handle:
- multiple admins in the section
- the admin not being the first admin

**Describe alternatives you've considered**
Leave it as it is, and document that you can't (reliably) create another admin that you use as the default Docker user. 

**Additional context**
This came about while debugging an instance where data corruption (caused by two couchdbs running in parallel over the same data) invalidated existing users. We noticed a new [admins] section being created after we added the new couchdb user and restarting the container.

sh1vam31:
Hi @andrablaj, @jkuester 

I’d love to work on this issue!

I understand that the current grep command in couchdb/docker-entrypoint.sh is too fragile because it assumes the target user is always on the very first line right below [admins]. To fix this, I plan to replace it with a small awk block that will safely check if $COUCHDB_USER exists anywhere inside the [admins] section.

Could you please assign this to me? 
I can open a PR with the fix shortly.

PR #10750, opened 2026-03-24
fix(#8816): support Couch admins pre-definied in `.ini` file

sh1vam31:
# Description

## Root Cause: 
The CouchDB docker-entrypoint.sh script previously used a hardcoded grep search sequence (grep -Pzq "\[admins\]\n$COUCHDB_USER =") to verify if the server admin was already populated in cluster-credentials.ini. This approach is extremely fragile because it strictly requires the target user to appear on the exact line immediately following the [admins] header. If multiple admins exist, the check silently fails, causing the script to inadvertently append a duplicate [admins] block and corrupt the server configuration across container restarts.

## Solution: 
I replaced the fragile regex with a robust, POSIX-compliant awk INI parser exactly where the admin checks occur. This guarantees that the script targets exclusively within the active [admins] block and correctly parses the username key regardless of its position in the list or surrounding spaces. When the user is successfully identified, it cleanly skips the duplicate insertion and preserves the config integrity.

# Code review checklist

# License

### DIFF
--- a/tests/integration/transitions/utils.js
+++ b/tests/integration/transitions/utils.js
@@ -1,6 +1,4 @@
 const utils = require('@utils');
-const sentinelUtils = require('@utils/sentinel');
-
 
 const getApiSmsChanges = (messages) => {
   const expectedMessages = messages.map(message => message.content);
@@ -31,12 +29,8 @@ const getApiSmsChanges = (messages) => {
         if (!expectedMessages.length) {
           listener.cancel();
           clearTimeout(timeout);
-          sentinelUtils
-            .waitForSentinel(ids)
-            .then(() => resolve(changes))
-            .catch(reject);
+          resolve(changes);
         }
-
       }
     });
   });


sh1vam31:
Hi @dianabarsan,

I replaced the hardcoded grep sequence in docker-entrypoint.sh with a reliable awk INI parser block. It now safely and accurately detects the user anywhere inside the [admins] block, preventing any duplicate configurations from being appended on restart.

Kindly review this PR.

Please let me know if you need any adjustments or if the code looks good to go.

couchdb/docker-entrypoint.sh
```diff
@@ -80,7 +80,11 @@ if [ "$1" = '/opt/couchdb/bin/couchdb' ]; then
 
 	if [ "$COUCHDB_USER" ] && [ "$COUCHDB_PASSWORD" ] && [ -z "$COUCHDB_SYNC_ADMINS_NODE" ]; then
 		# Create admin only if not already present
-		if ! grep -Pzq "\[admins\]\n$COUCHDB_USER =" $CLUSTER_CREDENTIALS; then
+		if ! awk -v user="$COUCHDB_USER" '
+			/^\[.*\]$/ { in_admins = ($0 == "[admins]") }
+			in_admins && $1 == user && $2 == "=" { found = 1; exit }
+			END { exit !found }
+		' "$CLUSTER_CREDENTIALS"; then
 			printf "\n[admins]\n%s = %s\n" "$COUCHDB_USER" "$COUCHDB_PASSWORD" >>$CLUSTER_CREDENTIALS
```
jkuester: I think your `awk` logic will work fine.  However, if an `[admins]` section exists, but does not contain an entry matching `$COUCHDB_USER`, this logic writes a second `[admins]` section at the bottom.  From the issue description, it seems that Couch will not merge these two sections in the same file.  So, we need some additional logic here so that in that case this script will just add another entry for `$COUCHDB_USER` to the existing `[admins]` section instead of adding a second `[admins]` section at the bottom.

sh1vam31:
Hi @jkuester,

Thank you for the great catch on that edge case.

I've pushed an update that explicitly checks if an [admins] block already exists. If it does, the script now cleanly injects the new user = pass entry directly inside that existing block instead of creating a second header at the bottom of the file.

This should fully prevent any duplicate [admins] sections from causing conflicts in CouchDB. 

Please let me know if you need any adjustments or if the code looks good to go.

sh1vam31:
Hi @jkuester 

I noticed the ci-integration-all-all check failed on the nginx, which should resolve the updated api ips test with a 502 Bad Gateway. However, since 699 other tests passed, it looks like a flake related to container restart timing rather than the CouchDB entrypoint I changed. 

Could you please re-run the failed jobs when you have a chance?

sh1vam31:
Thanks for the feedback, @jkuester 

I've refactored the logic to use the simpler grep and sed approach you suggested. It's definitely much cleaner and easier to follow now while still ensuring the [admins] section is handled correctly without duplicates.

I've verified the changes and pushed them to the branch. 

Let me know if there’s anything else you'd like me to adjust.

jkuester:
Hmm, the CI is failing for what looks like issues with the Couch container.  Taking a look at what might be going wrong....  (Not sure if it a problem with the container or a problem with the test...)

jkuester:
Okay, so I found the issue. :facepalm: My approach with `sed -i` did not handle special characters in the user/password.  The `unit-couchdb` test script [specifically tests a password with a bunch of special chars](https://github.com/medic/cht-core/blob/4fd02e6a2f8867ab8b7f01ec3396d0e369a91c87/couchdb/tests/Makefile#L20).  This was what failed the build.  

@sh1vam31 I have just pushed an update that goes back to using `awk` to do the credential insertion (since the `ENVIRON` feature should properly avoid any issues with special chars).  At least now the tests pass....

Let me know if this approach makes sense to you or if you see a better way.  Otherwise, as long as the rest of the build passed I think we can ship this. :+1:

sh1vam31:
Hi @jkuester, 

Thank you for the detailed explanation and for pushing the awk + ENVIRON update!

That approach makes perfect sense and is much more robust against special characters and file corruption. I have also:

**Merged master:** Synced the branch with the latest changes to resolve the out-of-date status.

**Refactored the Sync Block:** I noticed that the cluster synchronisation block (around line 124) was still using the old, fragile logic. I've updated it to use the same awk approach for consistency and safety in cluster mode.

I've verified the changes locally, and the PR should now be fully robust across the entire script. 

Let me know if everything looks good on your end now.

couchdb/docker-entrypoint.sh
```diff
@@ -34,19 +34,51 @@ setSecret() {
...
+		awk '/^\s*\[\s*couch_httpd_auth\s*\]\s*$/{print; print ENVIRON["AUTH_SECRET_LINE"]; next}1' \
+			"$CLUSTER_CREDENTIALS" > "${CLUSTER_CREDENTIALS}.tmp" \
+			&& mv "${CLUSTER_CREDENTIALS}.tmp" "$CLUSTER_CREDENTIALS"
+	else
+		# Update if exists
+		AUTH_SECRET_PART="$COUCHDB_SECRET"
+		export AUTH_SECRET_PART
+		awk '/^\s*\[\s*couch_httpd_auth\s*\]\s*$/{in_auth=1} /^\[/{if ($0 !~ /\[couch_httpd_auth\]/) in_auth=0} {if (in_auth && /^\s*secret\s*=/) {print "secret = " ENVIRON["AUTH_SECRET_PART"]; next} print}' \
+			"$CLUSTER_CREDENTIALS" > "${CLUSTER_CREDENTIALS}.tmp" \
+			&& mv "${CLUSTER_CREDENTIALS}.tmp" "$CLUSTER_CREDENTIALS"
```
jkuester: I do not think we want to include this else-block. Unless I am missing something, this would change the script to actually behave differently than it did before.  Previously, if a `secret` already existed in `CLUSTER_CREDENTIALS`, then the value from `$COUCHDB_SECRET` would not be used, right?  I think we want to stick with that behavior (at least for now). If we want to change this, we can do that in a separate issue/PR.

Updating the rest of this logic here to follow our new sed+awk pattern is great!

couchdb/docker-entrypoint.sh
```diff
@@ -34,19 +34,51 @@ setSecret() {
...
+		awk '/^\s*\[\s*couchdb\s*\]\s*$/{print; print ENVIRON["AUTH_UUID_LINE"]; next}1' \
+			"$CLUSTER_CREDENTIALS" > "${CLUSTER_CREDENTIALS}.tmp" \
+			&& mv "${CLUSTER_CREDENTIALS}.tmp" "$CLUSTER_CREDENTIALS"
+	else
+		# Update if exists
+		AUTH_UUID_PART="$COUCHDB_UUID"
+		export AUTH_UUID_PART
+		awk '/^\s*\[\s*couchdb\s*\]\s*$/{in_couchdb=1} /^\[/{if ($0 !~ /\[couchdb\]/) in_couchdb=0} {if (in_couchdb && /^\s*uuid\s*=/) {print "uuid = " ENVIRON["AUTH_UUID_PART"]; next} print}' \
+			"$CLUSTER_CREDENTIALS" > "${CLUSTER_CREDENTIALS}.tmp" \
+			&& mv "${CLUSTER_CREDENTIALS}.tmp" "$CLUSTER_CREDENTIALS"
```
jkuester: Same here. I think we want to avoid changing the default behavior in this PR.

couchdb/docker-entrypoint.sh
```diff
@@ -94,10 +135,25 @@ if [ "$1" = '/opt/couchdb/bin/couchdb' ]; then
...
+				"$CLUSTER_CREDENTIALS" > "${CLUSTER_CREDENTIALS}.tmp" \
+				&& mv "${CLUSTER_CREDENTIALS}.tmp" "$CLUSTER_CREDENTIALS"
 		else
-			sed -i "s/level = .*/level = $COUCHDB_LOG_LEVEL/g" $CLUSTER_CREDENTIALS
+			# Update if key already exists
+			LOG_LEVEL_PART="$COUCHDB_LOG_LEVEL"
+			export LOG_LEVEL_PART
+			awk '/^\s*\[\s*log\s*\]\s*$/{in_log=1} /^\[/{if ($0 !~ /\[log\]/) in_log=0} {if (in_log && /^\s*level\s*=/) {print "level = " ENVIRON["LOG_LEVEL_PART"]; next} print}' \
+				"$CLUSTER_CREDENTIALS" > "${CLUSTER_CREDENTIALS}.tmp" \
+				&& mv "${CLUSTER_CREDENTIALS}.tmp" "$CLUSTER_CREDENTIALS"
```
jkuester: Same here. I think we want to avoid changing the default behavior in this PR.

couchdb/docker-entrypoint.sh
```diff
@@ -112,8 +168,18 @@ if [ "$1" = '/opt/couchdb/bin/couchdb' ]; then
...
+		fi
+
+		# Ensure user exists within [admins] section
+		if ! sed -n '/^\s*\[\s*admins\s*\]\s*$/,/^\s*\[/p' "$CLUSTER_CREDENTIALS" | grep -q "^\s*${COUCHDB_USER}\s*="; then
+			ADMIN_CREDS_LINE="$(printf '%s = %s' "$COUCHDB_USER" "$COUCHDB_HASHED_PASSWORD")"
+			export ADMIN_CREDS_LINE
+			awk '/^\s*\[\s*admins\s*\]\s*$/{print; print ENVIRON["ADMIN_CREDS_LINE"]; next}1' \
+				"$CLUSTER_CREDENTIALS" > "${CLUSTER_CREDENTIALS}.tmp" \
+				&& mv "${CLUSTER_CREDENTIALS}.tmp" "$CLUSTER_CREDENTIALS"
 		fi
```
jkuester: Good catch!

sh1vam31:
Hi @jkuester,
Thank you for the review! I've removed the else blocks so that existing secrets, UUIDs, and log levels are no longer overwritten, following the previous behaviour. I kept the new [admins] login and the sed+awk pattern as discussed.

jkuester:
@Hareet could you just double-check that these changes pass the sniff-test for you?  The main thing that bothers me is that the `awk` and `sed -i` calls will not play nicely if someone is volume mapping the `cluster-credentials.ini` file into the container (work fine to map in the whole `local.d` directory, but not just the file since these commands try to replace the whole file. )

_However_ this was already the case because of the existing `sed -i` command that we had when `COUCHDB_LOG_LEVEL` was set...  However, that envar was not _required_ (even though it is present in our production compose config. etc). So, I guess it is possible that previously folks might have been able to volume-map the `cluster-credentials.ini` file without the `COUCHDB_LOG_LEVEL` set.  Is that a scenario you think we need to worry about, or do we not need to worry about this?

sh1vam31:
Hi @jkuester,

All CI checks have passed successfully! 
The PR looks good to go whenever you're ready to merge. Thank you for pushing those additional fixes – really appreciate the help!

Hareet:
Sorry I didn't see this! Cool fixes, and we haven't the seen the scenario you are asking about. 

> Is that a scenario you think we need to worry about, or do we not need to worry about this?

I don't think it's a scenario we have to worry about. To confirm, you are talking about if someone runs:
` docker run -v /host/creds.ini:/opt/couchdb/etc/local.d/cluster-credentials.ini ...` ?
- We don't support that in our deployment configurations for compose or helm
   - I haven't heard of a dev doing that in their environment, but we could document to not do it (and always edit through the docker volume)
   - You could try to turn your sed -i / mv to cp + rm.
      - sed writes to tmp, cp takes tmp and overwrites original, perserving the file, then rm tmp

sh1vam31:
Hi @hareet, 
Thank you for the approval and the great suggestion! I've updated couchdb/docker-entrypoint.sh to use the cp + rm pattern instead of sed -i and mv. This should ensure that we preserve the file inodes and safely support scenarios where users mount single configuration files from the host.

@jkuester, please let me know if everything looks good to merge now or if there are any other improvements you'd like to see. Thanks.

jkuester:
Thanks @Hareet!  

> To confirm, you are talking about if someone runs:

Yes, that is the one I was talking about.  The good news is that if someone does do that, the Couch container will just fail to start with an error indicating the file cannot be replaced. That seems like an acceptable failure scenario, so I will merge this as-is.  We can iterate more on it in the future as needed.

PR #10755, opened 2026-03-25
feat(#10438): add telemetry for visible task count on recalculation

### LINKED ISSUE #10438: Add telemetry for how many tasks the users see
Workflows often get bloated or misconfigured. At this point, there is no way to know how many tasks are visible for a user at a given time except by analyzing data on the server, querying all user tasks, checking their statuses and filtering in the tasks that would be visible in a certain interval. 

Add a telemetry entry that will store how many tasks the user sees. This entry should be recorded each time tasks are recalculated. 

None. 

**Additional context**
Came out of discussions during the task prioritization squad call.

megha1807:
# Description

Fixes #10438
Docs PR:https://github.com/medic/cht-docs/pull/2171

Each time tasks are recalculated in `tasks.component.ts`, there was no way to know how many tasks are visible to a 
user without querying the server directly.

This PR adds a telemetry entry `tasks:visible-count` that records the number of visible tasks each time `refreshTasks()` 
is called — covering both initial load and subsequent recalculations triggered by changes or rules engine updates.

## Changes Made
- Injected `TelemetryService` into `TasksComponent`
- Added `telemetryService.record('tasks:visible-count', 
  tasksWithLineage.length)` after tasks are recalculated
- Added unit test to verify telemetry is recorded with 
  the correct key and count

# Code review checklist
- [x] Tested: Unit tests added
- [x] AI disclosure: Used AI assistance for guidance on 
      implementation approach and test structure

### DIFF
--- a/webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
+++ b/webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
@@ -284,21 +284,21 @@ describe('TasksComponent', () => {
   });
 
   it('should record telemetry with visible task count on recalculation', async () => {
-  const taskDocs = [
-    { _id: '1', emission: { _id: 'e1', owner: 'a' }, owner: 'a' },
-    { _id: '2', emission: { _id: 'e2', owner: 'b' }, owner: 'b' },
-    { _id: '3', emission: { _id: 'e3', owner: 'c' }, owner: 'c' },
-  ];
-  rulesEngineService.fetchTaskDocsForAllContacts.resolves(taskDocs);
+    const taskDocs = [
+      { _id: '1', emission: { _id: 'e1', owner: 'a' }, owner: 'a' },
+      { _id: '2', emission: { _id: 'e2', owner: 'b' }, owner: 'b' },
+      { _id: '3', emission: { _id: 'e3', owner: 'c' }, owner: 'c' },
+    ];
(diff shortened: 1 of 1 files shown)


webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -169,6 +171,7 @@ export class TasksComponent implements OnInit, OnDestroy {
         ...this.getTaskLineage(subjects, task)
       }));
       this.tasksActions.setTasksList(tasksWithLineage);
+      this.telemetryService.record('tasks:visible-count', tasksWithLineage.length);
```
dianabarsan: I think that using `visible` is too bespoke here. Please just use `tasks:all-tasks`, because it matches with other telemetry entries that we have: https://docs.communityhealthtoolkit.org/technical-overview/data/performance/telemetry/

megha1807: Hi @dianabarsan, thank you for the feedback! 
I have updated the telemetry key from `tasks:visible-count` to `tasks:all-tasks` to match the existing naming conventions.

megha1807:
Thanks @dianabarsan!

I’ll create a follow-up PR in cht-docs to document the new telemetry entry.

megha1807:
Hello @dianabarsan 
I have made a PR to PR in cht-docs to document the new telemetry entry.
https://github.com/medic/cht-docs/pull/2171

PR #10762, opened 2026-03-25
feat(#10630): add api support for serving ui extensions

### LINKED ISSUE #10630: Add `api` support for basic loading/serving UI Extensions
- Add a new service: `src/services/ui-extension.js`  (follow patterns from [forms](https://github.com/medic/cht-core/blob/e0c5407a90763ab5c3619cfb7b487d85f6bae75d/api/src/services/forms.js#L11-L10) and [extension-libs](https://github.com/medic/cht-core/blob/e0c5407a90763ab5c3619cfb7b487d85f6bae75d/api/src/services/extension-libs.js) services.
    - Export `getScript` function that takes the UI Extension name as the parameter and returns the script attachment.
        -  Get the extension doc by id: `ui-extension:${name}`.
        - Return an object with the `data` value from `doc._attachments['extension.js']`.
    - Export `getAllProperties` function that reads all the ui-extensions and returns an array objects from the docs
        - We should remove the Couch specific fields (`_id`, `_rev`, etc) and then add an `id` field that is the doc `_id` without the `ui-extension:` prefix.
        - Be sure to use the `allDocs` Pouch function to load the ui-extension docs by an id range. 
- Add a new  controller: `src/controllers/ui-extension.js` (follow patterns from [extension-libs.js](https://github.com/medic/cht-core/blob/e0c5407a90763ab5c3619cfb7b487d85f6bae75d/api/src/controllers/extension-libs.js))
    - Export a `list` function that just sends the results of service.getAllProperties as a json response.
    - Export a `get` function that calls service.getScript with the `req.params.name` parameter.
        - The returned script file should be buffered back on the response as `base64` and a Content-Type of `text/javascript`
- Update `src/routing.js` to route GET requests:
    - `/ui-extension` routes to controller.list
    - `/ui-extension/:id` routes to controller.get

Make sure to include unit tests for all the changes (except for `routing.js`).  Also, we will need to add a new integration test file to [tests/integration/api/controllers](https://github.com/medic/cht-core/tree/e0c5407a90763ab5c3619cfb7b487d85f6bae75d/tests/integration/api/controllers).

sh1vam31:
# Description

## Fixes #10630

The frontend of the app wanted to load custom "UI extensions" (additional custom pages or features). However, there was no backend endpoint to safely request and serve those extension files.

I added a new REST API endpoint to the backend Express server. This endpoint acts as a server handler that reads the requested UI extension file and serves it back to the client securely.

## Changes :

**Service** (api/src/services/ui-extension.js):

- Implemented getScript(name) which fetches ui-extension:${name} documents and extracts the .data payload from the extension.js attachment.

- Implemented getAllProperties() which runs a targeted allDocs query over the ui-extension: prefix. It maps the results into a clean array, stripping internal _ CouchDB fields (like _id and _rev) and assigns an id property without the prefix.

**Controller** (api/src/controllers/ui-extension.js):
- Implemented list() mapped to getAllProperties.

- Implemented get() which safely parses the base64 attachment buffer returned by getScript and exposes it with the required text/javascript Content-Type.

**Routing**: Added GET definitions for /ui-extension and /ui-extension/:name within the global app router.

**Testing**: Added rigorous Mocha/Chai Unit Tests isolating the service and controller logic, as well as an API integration test (tests/integration/api/controllers/ui-extension.spec.js) to guarantee correct path routing and 404 handling.

Verified the new getScript and getAllProperties logic by writing and running local Unit Tests:

image

# Code review checklist

# License


sh1vam31:
"Hi @jkuester

I have just opened a PR for this issue #10630.

I've implemented the ui-extension service, controller, and routing exactly as outlined in the issue description. I've also included full Mochachai unit tests for the controller/service logic, as well as the requested API integration test file which all passed successfully locally.

Please let me know if you need any adjustments or if I missed any specific patterns you were looking for during review.

sh1vam31:
Hi @jkuester 

it looks like the Compile the app check failed this time purely due to a 403 rate limit exceeded error while the CI runner was trying to download the shellcheck binary from npm. 

Could you please trigger a re-run of the failed jobs when the rate limit resets?

jkuester:

sh1vam31:
Hi @jkuester, thanks for the update.
No worries at all.
I appreciate you keeping this on your list. Looking forward to your review whenever you get the chance next week.

jkuester:
FYI, @sh1vam31, this PR was targeting `master`, but we need to target the [10224-ui-extensions](https://github.com/medic/cht-core/tree/10224-ui-extensions) branch for all the UI Extension work (since we will be merging everything for that project to `master` at the same time when it is done.  

I changed the target for this PR to `10224-ui-extensions`, but that caused a few problems since your `10630-ui-extensions-api` was ahead of `10224-ui-extensions` (because it was on top of `master`) and those extra commits from `master` were now showing up in this PR.  To resolve this, I re-based the two relevant commits from your branch back down onto a new version of  `10630-ui-extensions-api` that is based on `10224-ui-extensions` and then force-pushed this to your branch.  I am pretty sure this has gotten everything where we want it to be. The main thing is that you will need to `reset` your local `10630-ui-extensions-api` to match what is on the remote before adding any more commits.

Let me know, though, if I missed something in all the commit juggling (I kept a local backup of the branch exactly how it was previously in case we need to roll back to it.)

sh1vam31:
Hi @jkuester, 
Thank you for the explanation and for doing the re-base/force-push to clean up the branch history! I have successfully reset my local 10630-ui-extensions-api branch to match the remote. Ready for your review on the new base branch.

api/src/routing.js
```diff
@@ -304,7 +304,7 @@ app.use(express.static(resources.webappPath));
 app.get('/extension-libs', extensionLibs.list);
 app.get('/extension-libs/:name', extensionLibs.get);
 app.get('/ui-extension', uiExtension.list);
-app.get('/ui-extension/:name', uiExtension.get);
+app.get('/ui-extension/:id', uiExtension.get);
```
jkuester, line 307: Good job following the pattern from `extension-libs` here, but just for the sake of the internal consistency of the ui-extension code, I think we should  call the param `id`.  (Since on the returned properties, the field with the identifier in it is called `id`...)

api/src/services/ui-extension.js
```diff
@@ -33,19 +35,13 @@ module.exports = {
     return result.rows.map(row => {
       const doc = row.doc;
       const id = doc._id.replace(PREFIX, '');
-      
-      // Remove CouchDB specific fields (starting with '_')
-      const cleanProperties = Object.keys(doc)
-        .filter(key => !key.startsWith('_'))
-        .reduce((obj, key) => {
```
jkuester, line 40: Minor, but per our [style guide](https://docs.communityhealthtoolkit.org/community/contributing/code/style-guide/#avoid-reduce) we try to avoid `reduce`.

api/tests/mocha/controllers/ui-extension.spec.js
```diff
@@ -1,60 +1,93 @@
 const sinon = require('sinon');
-const chai = require('chai');
+const { expect } = require('chai');
```
jkuester, line 2: We have a ton of older tests that follow your pattern of just using `chai`, but in our newer tests we are trying to reduce the noise by just directly using `expect`.

api/tests/mocha/controllers/ui-extension.spec.js
```diff
@@ -1,60 +1,93 @@
...
-      sinon.stub(service, 'getScript').resolves({ data: 'QUJD' }); // 'ABC' base64
+      const req = { params: { id: 'test' } };
+      getScript.resolves({ data: 'QUJD' }); // 'ABC' base64
+      await controller.get(req, res);
+      expect(res.set).to.have.been.calledOnceWithExactly('Content-Type', 'text/javascript');
+      expect(res.send.args[0][0].toString()).to.equal('ABC');
+      expect(errorStub).to.not.have.been.called;
+    });
+
+    it('handles unexpected error', async () => {
```
jkuester, line 80: Just want to be sure to fully cover the difference cases in the code.

api/tests/mocha/services/ui-extension.spec.js
```diff
@@ -60,16 +64,40 @@ describe('UI Extension service', () => {
               name: 'My Extension',
               version: '1.0'
             }
+          },
+          {
+            doc: {
+              _id: 'ui-extension:another-ext',
+            }
```
jkuester, line 71: It is always good when testing code that handles an array to test with an array with multiple entries to be sure your looping works as expected.

api/tests/mocha/services/ui-extension.spec.js
```diff
@@ -60,16 +64,40 @@ describe('UI Extension service', () => {
...
+        { id: 'another-ext' }
+      ]);
+      expect(allDocs).to.have.been.calledOnceWithExactly({
+        startkey: 'ui-extension:',
+        endkey: `ui-extension:\ufff0`,
+        include_docs: true,
+      });
+    });
+
+    it('returns an empty array when no ui-extension docs exist', async () => {
```
jkuester, line 92: In unit tests I also like to test the empty case just to make sure we never assume the array will always be populated.

tests/integration/api/controllers/ui-extension.spec.js
```diff
@@ -1,29 +1,40 @@
...
 
   describe('GET /ui-extension/:name', () => {
     it('returns 404 for unknown extension', async () => {
-      await utils.request({ path: '/ui-extension/unknown-ext' })
-        .then(() => chai.expect.fail('should have thrown'))
-        .catch(err => chai.expect(err.status).to.equal(404));
+      await expect(utils.request({ path: '/ui-extension/unknown-ext' })).to.eventually.be.rejectedWith({
+        code: 404,
+        error: 'Not Found',
+      });
```
jkuester, line 37: Minor, but this is the best way I have found with mocha+chai to assert async errors.  Your approach of then+catch works, but can be a bit confusion if the tested function does not throw an error.

The test will still fail (because of the `expect.fail` in the then-block), but the catch actually catches that error and so the reason listed for the test failure is that your error expectation in the catch-block does not pass.

jkuester (approved):
@sh1vam31 thanks for the great PR!  I only had some minor updates from the review and I have gone ahead and committed them myself so we can get this merged!  

I did add a few comments inline to explain my changes so you understand for future reference.  👍

PR #10764, opened 2026-03-26
feat(#10670): add ui-extensions service to webapp

### LINKED ISSUE #10670: Add UI Extensions service to `webapp`
Add a new `webapp/src/ts/services/ui-extensions.service.ts` file with the following:

```ts
interface UiExtensionProperties {
  readonly id: string,
  readonly type: string
  readonly roles?: string[]
  readonly icon?: string
  readonly title?: string
  readonly config?: Record<string, unknown>
}

interface UiExtension {
  readonly properties: UiExtensionProperties
  readonly Element: HtmlElement
}
```

## `class UiExtensionsService`

- Load available UI extension data from the `api/v1/ui-extension` endpoint
	- Can follow patterns from cht-datasource.service for loading the extension-libs.
	- Should also only keep extensions that do not have `roles` configured or ones where the user has at least one of the configured `roles`.
	   - Use the `SessionService.hasRole` method to check.
  - Can cache this property data since it will be accessed several times. 
- `getPropertiesByType(type: string): UiExtensionProperties[]` - return the properties for all  extensions with the given type
- `getProperties(id: string): UiExtensionProperties` - return the property data for the given type.
- `getExtension(id: string): UiExtension`
	- Follow the pattern [from the cht-datasource.service](https://github.com/medic/cht-core/blob/dad316fa2875e356bc265e218869e70d131870fb/webapp/src/ts/services/cht-datasource.service.ts#L72) to load the UI extension code from the script file (via the  `/ui-extension/:id` endpoint).
		- Want to return the result of `module.exports` (which should contain the web component exported by the script) and the `element` property.

---

Add new unit test file:  `webapp/tests/karma/ts/services/ui-extensions.service.spec.ts`

megha1807:
This PR introduces the `UiExtensionsService` to the webapp as part of the UI extensions feature.

The service is responsible for:
- Fetching UI extension properties from `/api/v1/ui-extension`
- Filtering extensions based on the current user's roles
- Lazy loading extension scripts from `/ui-extension/:id`
- Caching loaded extension scripts to avoid redundant network requests
- Exposing methods to retrieve extensions by type or ID

Unit tests are included covering:
- Initialization and single-load behaviour
- Role-based filtering logic
- Property lookup by type and ID
- Extension script loading, caching, and error handling

Closes #10670

### DIFF
--- a/webapp/src/ts/services/ui-extensions.service.ts
+++ b/webapp/src/ts/services/ui-extensions.service.ts
@@ -23,7 +23,7 @@ interface UiExtension {
 })
 export class UiExtensionsService {
   private extensionProperties: UiExtensionProperties[] = [];
-  private extensionScripts: Record<string, any> = {};
+  private extensionScripts: Record<string, new () => HTMLElement> = {};
   private initialized = false;
 
   constructor(
@@ -63,7 +63,6 @@ export class UiExtensionsService {
     if (!Element || !(Element.prototype instanceof HTMLElement)) {
       throw new Error(`Could not load UI Extension element with id [${id}].`);
     }
-    this.extensionScripts[id] = Element;
     return Element as new () => HTMLElement;
   }
 
@@ -85,7 +84,7 @@ export class UiExtensionsService {
     const properties = await this.getProperties(id);
 
     if (!this.extensionScripts[id]) {
-      await this.loadExtensionScript(id);
+      this.extensionScripts[id] = await this.loadExtensionScript(id);
     }
 
     return {
--- a/webapp/tests/karma/ts/services/ui-extensions.service.spec.ts
+++ b/webapp/tests/karma/ts/services/ui-extensions.service.spec.ts
@@ -31,7 +31,7 @@ describe('UiExtensionsService', () => {
   });
 
   describe('init()', () => {
-    it('should load extension properties on first call', async () => {
+    it('should load extension properties only on first call', async () => {
       const extensions = [
         { id: 'ext-1', type: 'tab' },
         { id: 'ext-2', type: 'tab' },
@@ -41,15 +41,11 @@ describe('UiExtensionsService', () => {
       await service.getPropertiesByType('tab');
 
       expect(http.get.calledOnceWithExactly('/ui-extension', { responseType: 'json' })).to.be.true;
-    });
-
-    it('should only initialize once when called multiple times', async () => {
-      http.get.returns(of([]));
 
       await service.getPropertiesByType('tab');
-      await service.getPropertiesByType('tab');
-      await service.getPropertiesByType('tab');
+      await service.getProperties('ext-1');
 
+      // Not fetched again on later requests
       expect(http.get.callCount).to.equal(1);
     });
 
@@ -174,7 +170,7 @@ describe('UiExtensionsService', () => {
   });
 
   describe('getExtension()', () => {
-    it('should load and return extension with Element', async () => {
+    it('should load and return extension with Element and cache it', async () => {
       const extensions = [{ id: 'ext-1', type: 'tab' }];
       http.get.onCall(0).returns(of(extensions));
       http.get.onCall(1).returns(of(
@@ -186,20 +182,13 @@ describe('UiExtensionsService', () => {
       expect(result).to.not.be.undefined;
       expect(result.properties).to.deep.equal({ id: 'ext-1', type: 'tab' });
       expect(result.Element.prototype).to.be.an.instanceof(HTMLElement);
+      expect(http.get.callCount).to.equal(2);
       expect(http.get.args[1][0]).to.equal('/ui-extension/ext-1');
       expect(http.get.args[1][1]).to.deep.equal({ responseType: 'text' });
-    });
-
-    it('should cache extension script and not reload on second call', async () => {
-      const extensions = [{ id: 'ext-1', type: 'tab' }];
-      http.get.onCall(0).returns(of(extensions));
-      http.get.onCall(1).returns(of(
-        'module.exports = class HelloWorldComponent extends HTMLElement {}'
-      ));
 
       await service.getExtension('ext-1');
-      await service.getExtension('ext-1');
 
+      // cached extensions are not re-fetched.
       expect(http.get.callCount).to.equal(2);
     });
 


megha1807:
Hi @jkuester, the SonarCloud Quality Gate is failing due to a Security Hotspot 
on the `new Function()` call used to load extension scripts dynamically. 
This is intentional as it follows the same pattern used in `cht-datasource.service`. 
Could you please review and acknowledge the security hotspot on SonarCloud?

jkuester:

jkuester:
Also @megha1807 thanks for calling out the Sonar failure!  I have confirmed in Sonar that it is safe. :+1:

megha1807:
Thanks for the update @jkuester! Appreciate you taking the time -- looking forward to your review next week.

Also, thanks for confirming the Sonar hotspot! 👍

megha1807:
Hi @jkuester,
I noticed the issue description specifies Element: HtmlElement but the test cases suggest it should be a callable type because:

Tests call result!.Element() as a function
Tests expect result!.Element to be null on error
Tests expect result!.Element to be a 'function'

I have typed it as ((...args: any[]) => any) | null to match the test expectations. Could you confirm if this is the correct approach or if the interface should be updated differently?
Thank you!

jkuester:
> the issue description specifies Element: HtmlElement but the test cases suggest it should be a callable type:

:thinking: I think the goal is to have the `Element` property be a sub-class of `HtmlElement` and not a function.

I have not had time to dig too deep on this yet (probably will need to go back through [the original technical design](https://docs.google.com/document/d/1RFiSS2hdYhLloNUcahlDVX6KN5ZNEQhUnnc5EVDR9Pc/edit?tab=t.ufay0fr37c8h#heading=h.st9k331jebaj) and maybe the other issues to verify exactly what we are expecting the UI extension script to export. I thought we landed on it just _exporting_ a class. (Basically like you have [in the tests](https://github.com/medic/cht-core/pull/10764/changes#diff-10752896dcf2e7232ec3a502fefbe69ae1543b6c46fb692e9465da38db56a585R191).)  So, if it is still coming through as a function, perhaps we are just missing one more function call [somewhere in here where we are building things](https://github.com/medic/cht-core/pull/10764/changes#diff-ffde7b199f9f24f8f60116176fdb3fd079cfbfcf70d50080ff76a75d6faae26aR70).  :shrug:   I plan  to dig deeper on this when I get the chance. :+1:

webapp/src/ts/services/ui-extensions.service.ts
```diff
@@ -0,0 +1,102 @@
...
+  readonly type: string;
+  readonly roles?: string[];
+  readonly icon?: string;
+  readonly title?: string;
+  readonly config?: Record<string, unknown>;
+}
+
+interface UiExtension {
+  readonly properties: UiExtensionProperties;
+  readonly Element: HTMLElement;
```
jkuester:
```suggestion
  readonly Element: new () => HTMLElement;
```

webapp/src/ts/services/ui-extensions.service.ts
```diff
@@ -0,0 +1,102 @@
...
+  isInitialized() {
+    if (!this.initialized) {
+      this.initialized = this.init();
+    }
+    return this.initialized;
+  }
+
+  private async init() {
+    await this.loadExtensionProperties();
+  }
```
jkuester: I appreciate that you are following the pattern from the cht-datasource.service.  However, that service needs to be inited during the [initial application bootstrapping](https://github.com/medic/cht-core/blob/2e64efc9e885c96e2164ef8763b37fda8f1c0a0a/webapp/src/ts/app.component.ts#L311). I would like to avoid that here if possible.  

I think we can simplify this a bit and make `initialized` just be a simple boolean and then update this code to just be 

```ts
  private async init() {
    if (this.initialized) {
      return;
    }
    await this.loadExtensionProperties();
    this.initialized = true;
  }
```

Then in `getPropertiesByType` and `getProperties` we need to make them `async` functions and start the functions by just calling 

```ts
await this.init();
```

This should make sure that we only load things once and we only load it when we need it.

webapp/src/ts/services/ui-extensions.service.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  private async loadExtensionProperties() {
+    try {
+      const request = this.http.get<UiExtensionProperties[]>('/api/v1/ui-extension', { responseType: 'json' });
```
jkuester:
```suggestion
      const request = this.http.get<UiExtensionProperties[]>('/ui-extension', { responseType: 'json' });
```

Sorry, this was a typo in the issue description. 👍

webapp/src/ts/services/ui-extensions.service.ts
```diff
@@ -0,0 +1,102 @@
...
+      const extensions = await lastValueFrom(request);
+      if (!extensions?.length) {
+        return;
+      }
```
jkuester:
```suggestion
```

Minor, but we really do not need to break early from this function.  If `extensions` is empty, then the `filter` never runs and nothing much happens.

webapp/src/ts/services/ui-extensions.service.ts
```diff
@@ -0,0 +1,102 @@
...
+    try {
+      const request = this.http.get('/ui-extension/' + id, { responseType: 'text' });
+      const result = await lastValueFrom(request);
+      const module = { exports: null };
+      new Function('module', result)(module);
+      this.extensionScripts[id] = module.exports;
+    } catch (e) {
+      console.error(`Error loading UI extension script: "${id}"`, e);
+    }
+  }
```
jkuester, line 67:
```suggestion
  private async loadExtensionScript(id: string) {
    const request = this.http.get('/ui-extension/' + id, { responseType: 'text' });
    const result = await lastValueFrom(request);
    const module = { exports: null as any };
    new Function('module', result)(module);
    const Element = module.exports;
    if (!Element || !(Element.prototype instanceof HTMLElement)) {
      throw new Error(`Could not load UI Extension element with id [${id}].`);
    }
    return Element as new () => HTMLElement;
  }
```

In this function I think we need to just throw an error if we cannot return a valid Element.  (Basically that means that the user is trying to get to a UI extension that _should_ exist (was in the properties), but we failed to find the expected script attachment.

Also, tweaked a couple of things to appease the TS types!

webapp/src/ts/services/ui-extensions.service.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  getPropertiesByType(type: string): UiExtensionProperties[] {
+    return this.extensionProperties.filter(extension => extension.type === type);
+  }
+
+  
+  getProperties(id: string): UiExtensionProperties | undefined {
```
jkuester: In this function, I think we should also throw an error if we cannot find the properties for the `id`.  Once again, if we have an `id` value, but no properties exist for it that is a super invalid/unexpected state. I think it is best to fail early instead of trying to handle gracefully.

webapp/tests/karma/ts/services/ui-extensions.service.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+    });
+  });
+
+ 
+
+  describe('getExtension()', () => {
+    it('should load and return extension with Element', async () => {
+      const extensions = [{ id: 'ext-1', type: 'tab' }];
+      http.get.onCall(0).returns(of(extensions));
+      http.get.onCall(1).returns(of('module.exports = class MyElement {}'));
```
jkuester: To pass the new validation check, I think we need:


```suggestion
      http.get.onCall(1).returns(of(
        'module.exports = class HelloWorldComponent extends HTMLElement {}'
      ));
```

webapp/tests/karma/ts/services/ui-extensions.service.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+
+      await service.isInitialized();
+      const result = await service.getExtension('ext-1');
+
+      expect(result).to.not.be.undefined;
+      expect(result!.properties).to.deep.equal({ id: 'ext-1', type: 'tab' });
+      expect(result!.Element).to.be.a('function');
```
jkuester:
```suggestion
      expect(result.Element.prototype).to.be.an.instanceof(HTMLElement);
```

jkuester (changes requested):

megha1807:
Hello @jkuester 
I’ve addressed the requested changes. Could you please take a look and let me know if anything else is needed?

PR #10767, opened 2026-03-26
feat(#10663): add ui-extension to service worker

### LINKED ISSUE #10663: Add `api` support for `ui-extension` endpoints in service worker
After we have added the `ui-extension` endpoints in https://github.com/medic/cht-core/issues/10630, we also need to add support for generating the service worker file with the new endpoint data for offline users.

See the [pattern for extension-libs](https://github.com/medic/cht-core/blob/14eeaea40d1c3035bea114f2d49bb739d32f54bf/api/src/generate-service-worker.js#L61) and add support to the `generate-service-worker.js` file for providing the UI Extension data via the service worker for offline users.  Load the data via the ui-extension service.

We only need to include extensions on the service worker if they are accessible to offline users. Check the `roles` property for the extension and only include the extension if there is no `roles` (the extension is accessible to all users) or at least one of the roles is `offline: true`.

Besides the unit tests, I think we also _might_ need to update the [service-worker e2e test](https://github.com/medic/cht-core/blob/b9ddc064d82158b65cd801eb8488ccbab4838de0/tests/e2e/default/service-worker/service-worker.wdio-spec.js#L8) to account for these changes. Just want to assert that the `/ui-extension` url is present in the "initial list of cached resources"...

sh1vam31:
# Description

## Fixes #10663 (Depends on #10630)

This PR adds Service Worker caching support for the new UI Extensions feature, ensuring that custom scripts are available for offline users. It builds upon the API endpoints introduced in #10630 and closely follows the established pattern used for extension-libs.

### **Changes Made:**

**Service Worker Generation:** Added an appendUiExtensions hook to api/src/generate-service-worker.js.

**Role-Based Offline Filtering:** Implemented the requested permissions check. The service worker now fetches all UI Extensions but strictly limits caching to extensions that either:
 - Have no roles defined (globally accessible).
 - Have at least one role configuration with the property offline: true.
 
 
 

**Unit Testing:** Updated generate-service-worker.spec.js using sinon to mock uiExtensionService, adding specific tests to verify that offline extensions get cached while online-only extensions are properly ignored.

**E2E Testing:** Appended the base /ui-extension URL to the expected initial cached resources list in tests/e2e/default/service-worker/service-worker.wdio-spec.js to ensure CI webdriver tests pass successfully.

### **Testing:**
npm run unit-api passes all generate-service-worker logic checks.

### **Proof of Fix**

Screenshot 2026-03-27 at 2 24 23 AM

# Code review checklist

# License

### DIFF
--- a/api/tests/mocha/generate-service-worker.spec.js
+++ b/api/tests/mocha/generate-service-worker.spec.js
@@ -55,9 +55,13 @@ describe('generate service worker', () => {
     extensionLibsService.getAll.resolves([{ name: 'bar.js', data: 'barcode' }]);
     uiExtensionService.getAllProperties.resolves([
       { id: 'offline-ext', roles: [{ offline: true }] },
-      { id: 'online-ext', roles: [{ offline: false }] }
+      { id: 'online-ext', roles: [{ offline: false }] },
+      { id: 'no-roles-ext' },
+      { id: 'no-script-ext' },
     ]);
-    uiExtensionService.getScript.resolves({ data: 'my-script' });
+    uiExtensionService.getScript.withArgs('offline-ext').resolves({ data: 'my-script' });
+    uiExtensionService.getScript.withArgs('no-roles-ext').resolves({ data: 'other extension script' });
+    uiExtensionService.getScript.withArgs('no-script-ext').resolves();
     sinon.stub(workbox, 'generateSW').returns();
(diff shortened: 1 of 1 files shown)


jkuester:

sh1vam31:
Hi @jkuester, thanks for the update.
No worries at all.
I appreciate you keeping this on your list. Looking forward to your review whenever you get the chance next week.

PR #10772, opened 2026-03-27
feat(#10240): hide target counts past goal

### LINKED ISSUE #10240: Add the option to hide target counts past the goal
When target counts are very large, for example to track currency numbers, when the target reaches past the goal, showing the two numbers makes the text overlap in an unintelligible way.

Image

This also affects the target aggregates page. 

**Describe the solution and alternatives you'd like**
- Add an option that changes the UI to only show one big number. 
- Add an option to not show actual target count past the goal. 
- Change the UI entirely for big-number tasks

**Additional context**
Forum post: https://forum.communityhealthtoolkit.org/t/how-to-cap-target-numbers-in-display/5296/2

cliftonmcintosh:
# Description

## Outdated goals

**_This original approach has been superseded. See the Updated goals immediately below. For more context, please read the discussion._**

1. **Layout overlap**: When a count number is large, it overlapped with the "Goal X" label. Fixed by converting the `.count` block to a flex column layout so the goal label and count number stack vertically instead of colliding.

2. **Hide count past goal**: Adds a new permission `can_hide_target_count_past_goal`. When a user has this permission, the count number is hidden on any target where `pass >= goal`. The goal label remains visible. This behavior is off by default (empty roles array in the default config).

## Updated goals

1. Fix the layout overlap by centering the goal.
2. Add a field to the target object that allows people to specify that when the target is reached or exceeded, the UI is limited to showing the goal as the big number. For example if the target goal is 5000 and a user has reached a target count 10000, the UI for that target would show "Target 5000" in the small text at the top and "5000" in the larger text in the tile.
3. Any target data structure that does not have this field or does not have it set to true will default to the existing behavior -- the target count will be displayed even when it reaches or exceeds the goal.
4. No feature flag.

Closes #10240

## Companion PRs

CHT-CONF: https://github.com/medic/cht-conf/pull/800
CHT-DOCS: https://github.com/medic/cht-docs/pull/2176

# Questions

## Layout overlap fix — approach feedback

I fixed the overlap by replacing the absolute-positioned goal label with a stacked flex layout (`.count` becomes `display: flex; flex-direction: column; align-items: center`, with the goal label above the count number, both centered). An alternative I considered was keeping the goal label in the top-right corner and adding right padding to `.number` when `.has-goal` is present (e.g., `.has-goal .count .number { padding-right: 80px; }`), which would be a smaller change but would offset the count number from true center. The padding approach would keep the count number visually off-center — it would be centered over the remaining space after reserving room for the goal label, not centered over the full card width. So on a wide card with a goal label, the big number would appear shifted left rather than truly centered. This approach is also fragile — it's a magic number that works for typical cases but can't guarantee no overlap for arbitrarily large numbers.  Does the centered stacked layout work for you, or would you prefer the top-right approach?
<details>

<summary>Here is an example of the centering approach</summary>

centering-approach

</details>
<details>

<summary>Here is an example of the padding approach</summary>

padding-approach

</details>

## What to hide

First off, if we center the goal, then do we need hiding? I've included it, but I want to raise the possibility that it might not be needed if we center the goal. I'm not sure what the use cases are, so I defer to you.

In one of @dianabarsan's comments, she said we should be

> adding a setting to not display the goal OR target count past the goal

I chose to not display the target count because that was what I understood the suggestion to be. The example feature flag (`can_hide_target_count_past_goal`) mentioned in the issue points to that as the approach that was wanted, but it feels to me like having an empty target makes it look like the data is missing or the person has not met the goal.

What is preferred? Hiding the target? Hiding the goal? Something else? Here are some screenshots with different options. I think Option 2 works best.

I also wonder if maybe this isn't needed if we center both the goal and the target.

<details>

<summary>Option 1: Currently implemented -- hide only the target when goal is met or exceeded</summary>

Image

</details>

<details>

<summary>Option 2: Hide only the goal when goal is met or exceeded</summary>

Image

</details>

<details>

<summary>Option 3: Hide both the target and the goal when goal is met or exceeded</summary>

Image

</details>

<details>

<summary>Option 4: Include a visual element showing the goal is met</summary>

Image

</details>

# AI Disclosure

I used Claude Code to help navigate the codebase, plan the approach, and write tests. I made all design decisions, reviewed all code, and verified the solution manually in the browser. I attributed Claude as a co-author on all commits that included the use of Claude Code.

If you would like to see them, I can provide a copy of the directions I gave Claude Code when starting the planning.

# Code review checklist

## Visual alignment fix for count widgets

I adjusted the alignment of the target tiles to aligns the count number display across count widgets that have a goal and those that do not.
Previously, the number position differed between the two variants, causing visual inconsistency
when both types appeared on the same Targets tab. Fixed by adding a minimum height to the number
container and using content justification to keep the number vertically centered regardless of
whether a goal is present.

### Unit tests

14 unit tests cover the `AnalyticsTargetsComponent` in
`webapp/tests/karma/ts/modules/analytics/analytics-targets.component.spec.ts`.
The following cases are specifically tested for the `limit_count_to_goal` behavior:

- `limit_count_to_goal: true`, `pass > goal` → displays `goal` as the count number
- `limit_count_to_goal: true`, `pass === goal` (boundary case) → displays `goal`
- `limit_count_to_goal: true`, `pass < goal` → displays actual `pass` value
- `limit_count_to_goal` absent → displays actual `pass` value even when goal is exceeded
- `limit_count_to_goal: true`, `goal: -1` (no goal) → displays actual `pass` value

All 14 tests pass.

### Browser verification with dummy data

A dummy target (`Earned Points`, `goal: 20000`) was temporarily added to the component
to visually verify behavior in the browser under the following conditions:

| Condition | `limit_count_to_goal` | `pass` | `goal` | Displayed |
| --- | --- | --- | --- | --- |
| Count exceeds goal, field set | `true` | 50,000 | 20,000 | 20,000 ✓ |
| Count exceeds goal, field not set | absent | 50,000 | 20,000 | 50,000 ✓ |
| Count below goal, field set | `true` | 19,999 | 20,000 | 19,999 ✓ |
| Count equals goal (boundary), field set | `true` | 20,000 | 20,000 | 20,000 ✓ |

#### Screenshots from testing

<details>

<summary>Screenshot: limit when target exceeds goal</summary>

limit-when-target-exceeds-goal

</details>

<details>

<summary>Screenshot: no limit enforced when target exceeds goal</summary>

no-limit-target-exceeds-goal

</details>

<details>

<summary>Screenshot: limit when target less than goal</summary>

limit-when-target-less-than-goal

</details>

<details>

<summary>Screenshot: limit when target equals goal</summary>

limit-when-target-equals-goal

</details>

<details>

<summary>Screen recording: mobile-device-sized view</summary>

https://github.com/user-attachments/assets/80d45dcc-d609-49e1-ac0b-53ad25e0aa37

</details>

### End-to-end verification with real config data

The `active-pregnancies` target in `config/default/targets.js` was temporarily modified
to `goal: 20, limit_count_to_goal: true` and uploaded to the running local instance using
the local `cht-conf` build (which includes the companion schema and parse-targets changes).

After logging in as a CHW user with 33 active pregnancies, the target card correctly
displayed 20 (the goal) instead of 33 (the actual count), confirming the full data flow:

`targets.js` → `cht-conf compile` → CouchDB → `rules-engine` → `target-state.js` → component → UI

<details>

<summary>Screenshot: testing limit in view with active pregnancies</summary>

testing-with-active-pregnancies

</details>

# License


cliftonmcintosh:
@dianabarsan or @mrjones-plip 

Who should I ask to review this?

webapp/src/css/targets.less
```diff
@@ -69,6 +69,9 @@
         position: relative;
       }
       .count {
+        display: flex;
+        flex-direction: column;
+        align-items: center;
```
cliftonmcintosh, line 74: Centering the goal seems cleaner than trying to pad it in an absolute position in the top right corner. See my question about this in the PR description.

mrjones-plip:
She's offline already, but @dianabarsan should be able to pick this up next week - thanks for reaching out!

FYI we're working with a bunch of new community members who are submitting a big collection of PRs - might be a hot minute but we'll get to this!

webapp/src/ts/modules/analytics/analytics-targets.component.ts
```diff
@@ -63,6 +68,9 @@ export class AnalyticsTargetsComponent implements OnInit, OnDestroy {
 
   ngOnInit(): void {
     this.subscribeToStore();
+    this.authService.has(HIDE_COUNT_PAST_GOAL_PERMISSION).then(has => {
+      this.hideCountWhenGoalMet = has;
+    });
```
dianabarsan: can this be simplified as ?

```suggestion
this.hideCountWhenGoalMet = await this.authService.has(HIDE_COUNT_PAST_GOAL_PERMISSION);
```

dianabarsan (changes requested):
Thanks for the quick work. I think the central alignment is the best (but I'm not a UX expert). 

As for hiding the target count when it exceeds the goal, I believe we weren't trying to hide the actual number (like in your screenshots), rather just show the goal number instead of the actual target number.  It's possible using a permission is not the best approach, maybe only _some_ targets need their goals hidden? 

I believe the context for hiding the exceeding number past the goal was in the cases where the target counts provided remuneration, in some cases once the "ceiling" (goal) was reached, not additional remuneration was offered. 
for example, the chw would get paid:
- $50 for a score of 500, 
- $100 for a target score of 1000 (the goal / ceiling)
- $100 for a target score of 1500

Users would sometimes expect to be paid $200 if their target showed 2000. So then, these targets specifically would not show the actual number if it exceeded the goal. 

For other targets, for example for pregnancy or delivery targets, i see absolutely no reason to hide the actual target number.

cliftonmcintosh:
> As for hiding the target count when it exceeds the goal, I believe we weren't trying to hide the actual number (like in your screenshots), rather just show the goal number instead of the actual target number.

Ahhh. That makes sense.

I should be able to adjust it so that when the flag is set and the goal is exceeded, we show the goal instead of the target count.

> For other targets, for example for pregnancy or delivery targets, i see absolutely no reason to hide the actual target number.

Is there a way to distinguish which targets should not be hidden?

> It's possible using a permission is not the best approach, maybe only some targets need their goals hidden?

Given the possiblity that some targets might not be hidden (like pregnancy or delivery targets) while some might be, what about an alternative approach like this:

Add a new field to the target data structure that allows people to say the target can't exceed the goal. Once the goal is reached, either the app stops incrementing the data or, if it does increment, we use that new field to decide to limit the target count in the UI to the goal amount.

webapp/src/ts/modules/analytics/analytics-targets.component.ts
cliftonmcintosh: Applied. I had to define `ngOnInit` as `async`, so I made a commit locally and pushed it

dianabarsan:
I believe adding a field to the target configuration is the way to go. There is no such field at the moment (hence the feature request :D :D ) 

> either the app stops incrementing the data

I think this will have more implications. I think we should just manifest this setting in the UI and have the target document still register the actual number.

cliftonmcintosh:
> I believe adding a field to the target configuration is the way to go.

@dianabarsan Thanks. Based on the feedback so far, can I state what I think the goals are?

1. Fix the overlap by centering the goal (like in the screenshots I've provided)
2. Add a field to the target object that allows people to specify that when the target is reached or exceeded, the UI is limited to showing the goal as the big number. For example if the target goal is 5000 and a user has reached a target count 10000, the UI for that target would show "Target 5000" in the small text at the top and "5000" in the larger text in the tile.
3. Any target data structure that does not have this field or does not have it set to true will default to the existing behavior -- the target count will be displayed even when it reaches or exceeds the goal.
4. No feature flag

Does that capture what we're thinking?

dianabarsan:
@cliftonmcintosh nailed it!!

webapp/src/css/targets.less
```diff
@@ -69,7 +69,12 @@
...
+        justify-content: flex-end;
         text-align: center;
+        min-height: 74px;
```
cliftonmcintosh: The change for `justify-content` and `min-height` aligns the count number display across count widgets that have a goal and those that do not.
Previously, the number position differed between the two variants, causing visual inconsistency
when both types appeared on the same Targets tab. Fixed by adding a minimum height to the number container and using content justification to keep the number vertically centered regardless of whether a goal is present.

<details>

<summary>Before</summary>

See how the big number sits down lower whenever there is a goal?

Image

</details>

<details>

<summary>After</summary>

Now the big numbers stay at the same height in targets that have goals and targets that don't have goals.

Image

</details>

cliftonmcintosh:
Updated based on our revised approach.

I've added a "Testing done" section to the description. It has screenshots of various scenarios.

I've also opened companion PRs for [cht-conf](https://github.com/medic/cht-conf/pull/800) and [cht-docs](https://github.com/medic/cht-docs/pull/2176).

shared-libs/rules-engine/src/target-state.js
```diff
@@ -205,7 +205,7 @@ module.exports = {
     const aggregateTarget = target => {
       const aggregated = pick(
         target,
-        ['id', 'type', 'goal', 'translation_key', 'name', 'icon', 'subtitle_translation_key', 'visible']
+        ['id', 'type', 'goal', 'translation_key', 'name', 'icon', 'subtitle_translation_key', 'visible', 'limit_count_to_goal']
```
cliftonmcintosh: I noticed in cht-conf that some of fields are snake_case and some are camelCase. I used snake case for this new one. Let me know if it should be camel case instead.

cliftonmcintosh:
Will the test at `tests/e2e/visual/targets/targets-overview.wdio-spec.js` need any adjustment? It ends with a call to [`generateScreenshot('targets', 'overview')`](https://github.com/cliftonmcintosh/cht-core/blob/feat/10240-hide-target-counts-past-goal/tests/e2e/visual/targets/targets-overview.wdio-spec.js#L80). Is there some sort of reference screenshot that we will need to update to go with the visual changes?

cliftonmcintosh:
As I understand it, if we want an e2e test for the new field, we will depend on [the `cht-conf` PR](https://github.com/medic/cht-conf/pull/800) being
merged and the `cht-conf` package version bumped in `cht-core` before we can write a successful e2e test for a target that has `limit_count_to_goal: true` and a goal lower than the count. Is that accurate?

If so, I think that means we would pause this until a new version of cht-conf is published and can be brought in as a dependency in cht-core.

shared-libs/rules-engine/src/target-state.js
dianabarsan: snake_case is fine. we use snake_case for most json properties. those camel cased ones sneaked in :( :( :(

dianabarsan:
I didn't end up having space for this today. I'll check on it tomorrow.

dianabarsan:
> Will the test at `tests/e2e/visual/targets/targets-overview.wdio-spec.js` need any adjustment? It ends with a call to [`generateScreenshot('targets', 'overview')`](https://github.com/cliftonmcintosh/cht-core/blob/feat/10240-hide-target-counts-past-goal/tests/e2e/visual/targets/targets-overview.wdio-spec.js#L80). Is there some sort of reference screenshot that we will need to update to go with the visual changes?

No, you can safely ignore those. They haven't been introduced in our CI suite and not run regularly.
As for reference screenshot! Yes, we will need to update our docs site, but that can come before releasing.

dianabarsan:
> As I understand it, if we want an e2e test for the new field, we will depend on [the `cht-conf` PR](https://github.com/medic/cht-conf/pull/800) being merged and the `cht-conf` package version bumped in `cht-core` before we can write a successful e2e test for a target that has `limit_count_to_goal: true` and a goal lower than the count. Is that accurate?
> 
> If so, I think that means we would pause this until a new version of cht-conf is published and can be brought in as a dependency in cht-core.

Not necessarily. You can install a branch cht-conf version instead of the latest release in this branch, and then the property will be available to use. 

```
npm install github:cliftonmcintosh:feat/10240-hide-target-counts-past-goal
```

We would, indeed, need to publish the cht-conf package before merging this PR, as reverting to the latest released cht-conf will be required.

webapp/src/css/targets.less
dianabarsan: Yea, I don't like the big number offset either. I'm wondering whether the best visual solution is to move the goal underneath the number and conditionally display it. Then we won't have all this extra padding.

webapp/src/ts/modules/analytics/analytics-targets.component.ts
```diff
@@ -133,4 +133,11 @@ export class AnalyticsTargetsComponent implements OnInit, OnDestroy {
         });
       });
   }
+
+  getDisplayCount(target: any): number {
+    if (target.limit_count_to_goal && target.goal >= 0 && target.value?.pass >= target.goal) {
```
dianabarsan, line 138: Awww, negative goals are out? is the goal positive check necessary?

dianabarsan: ah I see, it's in case someone creates a target without a goal and then sets limit_count_to_goal to true?

dianabarsan (commented):
So far excellent work, as usual! 
I added one question about alternative styling and replied to the e2e questions.

webapp/src/css/targets.less
cliftonmcintosh: > I'm wondering whether the best visual solution is to move the goal underneath the number and conditionally display it. Then we won't have all this extra padding.

I can work up that as an option and provide a screenshot. Or if you are confident that is a change to make, I can go ahead and do it. Let me know, please.

cliftonmcintosh:
> You can install a branch cht-conf version instead of the latest release in this branch, and then the property will be available to use.

If it makes it easier to review, I can go ahead and do this and then we can revert to the latest after the cht-conf change is merged. Please let me know if you prefer that or prefer to wait until the cht-conf changes are merged.

webapp/src/ts/modules/analytics/analytics-targets.component.ts
cliftonmcintosh: > it's in case someone creates a target without a goal and then sets limit_count_to_goal to true

Yes!

dianabarsan:
> If it makes it easier to review, I can go ahead and do this and then we can revert to the latest after the cht-conf change is merged. Please let me know if you prefer that or prefer to wait until the cht-conf changes are merged.

We have done this numerous times, I prefer it and suggest it. No reason to block on the other work.

webapp/src/css/targets.less
cliftonmcintosh: I've moved the goal so it is below. Here is a screenshot. Tell me what you think, please.

Screenshot 2026-04-03 at 13 50 59

cliftonmcintosh:
> > If it makes it easier to review, I can go ahead and do this and then we can revert to the latest after the cht-conf change is merged. Please let me know if you prefer that or prefer to wait until the cht-conf changes are merged.
> 
> We have done this numerous times, I prefer it and suggest it. No reason to block on the other work.

I've added e2e tests with the cht-conf dependency set to my branch.

package.json
```diff
@@ -126,7 +126,7 @@
     "chai-shallow-deep-equal": "^1.4.6",
     "chai-things": "^0.2.0",
     "chokidar": "^4.0.3",
-    "cht-conf": "^6.0.2",
```
cliftonmcintosh: Reminder that this needs to be changed before merging.

package-lock.json
cliftonmcintosh, line 1: The changes in this file are because of the temporary change to the cht-conf dependency. They should be reverted when we update that dependency again.

dianabarsan (approved):
Best way to start the week, with a perfect PR to approve!

cliftonmcintosh:
> Best way to start the week, with a perfect PR to approve!

❤️ 

Please let me know when a new version of cht-conf is published, and I can update the dependency in this PR.

dianabarsan:
@cliftonmcintosh this flew off my radar. If you have time next week to do do the conf merge, it would be great. Sorry for the enormous delay.

package.json
cliftonmcintosh: Changed to 6.3.0

cliftonmcintosh:
@dianabarsan This is ready for re-review after updating the cht-conf dependency to 6.3.0. Sorry for the delay. I didn't notice you had published an updated version of cht-conf last week.

dianabarsan (approved):
Excellent work!

PR #10773, opened 2026-03-28
feat(#10671): add UiExtensionsTabComponent, route guard and routes

### LINKED ISSUE #10671: Add UI Extensions Tab component to `webapp`
Depends on https://github.com/medic/cht-core/issues/10670

There are several interlocking pieces we need to add to render the UI Extensions in the webapp and handle routing to them:

## UI Extensions Tab Component

Add new files: `webapp/src/ts/modules/ui-extensions/ui-extensions-tab.component.ts/html`

The HTML file needs to have the same outer framework as the other "tab" pages.  See the [trainings.component.html](https://github.com/medic/cht-core/blob/e33cd8738be9bbe1e3c34566fe07d7bf9fa9da6f/webapp/src/ts/modules/trainings/trainings.component.html) for example.  I think we need the `mm-tool-bar` at the top and then the `inner` and `inbox page` divs.  Then we need a custom wrapper div nested inside with the id: `ui-element-tab`.  Also , I think we can include similar `loader` logic to display a loading spinner while the extension code is being loaded.

For the `UiExtensionsTabComponent` class, we want to take the `ActivatedRoute` as a constructor parameter.  Then we can get the extension id via `route.snapshot.params['id']`.  With that id we can call the `UiExtensionsService.getExtension` method to get the actual web component code/properties.

The `UiExtensionsTabComponent` should have a property that references the wrapper div like:

```ts
@ViewChild('ui-element-tab') container!: ElementRef;`
```

From the `ngAfterViewInit()` method, Initialize the web component element with something like:

```ts
customElements.define(id, Element);
const element = document.createElement(id);
// TODO Set properties on element here before appending it.
this.container.nativeElement.appendChild(element);
```

We need to set the following properties:
- `cht` - the datasource loaded from `CHTDatasourceService.get`.
- `properties` - the defined extension properties (useful if loading custom configuration)
- `translate` - the `TranslateService.instant` method. Can be used by the web components to localize 
- TODO What else? 
  - `userContactSummary`
  - `user` -  `user-settings` doc for the user

## UI Extensions Tab Route Guard Provider

Create a route guard in a new file: `webapp/src/ts/modules/ui-extensions/ui-extensions-route.guard.provider.ts` -  `UiExtensionsTabRouteGuardProvider implements CanActivate`.  See the [AppRouteGuardProvider](https://github.com/medic/cht-core/blob/35662788c503441ef25ebe15ff99dbdd3bd2f104/webapp/src/ts/app-route.guard.provider.ts#L8) as an example.

When `canActivate` is called, call  `UiExtensionsService.getProperties` method with the `route.params['id]` value to get extension properties. Fail activation if the extension does not have a `type` of `app_main_tab` or `app_drawer_tab`.

## UI Extensions Routes

Create a new routes file: `webapp/src/ts/modules/ui-extensions/ui-extensions.routes.ts`. See the [training.routes.ts](https://github.com/medic/cht-core/blob/e33cd8738be9bbe1e3c34566fe07d7bf9fa9da6f/webapp/src/ts/modules/trainings/trainings.routes.ts#L8) as an example.

Export a parameterized route:

- `path` - `ui-extensions/:id`
- `component` - `UiExtensionsTabComponent`
- `canActivate` - `[UiExtensionsTabRouteGuardProvider]`

Update `webapp/src/ts/app-routing.module.ts` to include the new route.

---

Add new unit tests to the `webapp/tests/karma/ts/modules/ui-extensions` directory.

megha1807:
This PR adds the UI Extensions Tab component and routing as part of the UI extensions feature.

Changes include:
- `UiExtensionsTabComponent` - loads and renders the web component for a given extension id
- `UiExtensionsTabRouteGuardProvider` - guards the route, only allows `app_main_tab` and `app_drawer_tab` types
- `ui-extensions.routes.ts` - parameterized route `ui-extensions/:id`
- Updated `app-routing.module.ts` to include the new routes

The component:
- Reads the extension id from the route params
- Loads the extension via `UiExtensionsService.getExtension`
- Sets `cht`, `inputs`, `config` and `userContactSummary` properties on the web component element
- Records telemetry for render time via `PerformanceService`
- Shows a loading spinner while the extension initializes

Unit tests are included for both the component and route guard.

Depends on #10670
Closes #10671

### DIFF
--- a/webapp/src/ts/modules/ui-extensions/ui-extensions-route.guard.provider.ts
+++ b/webapp/src/ts/modules/ui-extensions/ui-extensions-route.guard.provider.ts
@@ -2,7 +2,7 @@ import { CanActivate, ActivatedRouteSnapshot } from '@angular/router';
 import { Injectable } from '@angular/core';
 import { UiExtensionsService } from '@mm-services/ui-extensions.service';
 
-const ALLOWED_TYPES = ['app_main_tab', 'app_drawer_tab'] as const;
+const ALLOWED_TYPES = new Set(['app_main_tab', 'app_drawer_tab']);
 
 @Injectable({
   providedIn: 'root'
@@ -10,13 +10,14 @@ const ALLOWED_TYPES = ['app_main_tab', 'app_drawer_tab'] as const;
 export class UiExtensionsTabRouteGuardProvider implements CanActivate {
   constructor(
     private readonly uiExtensionsService: UiExtensionsService,
-  ) {}
+  ) {
+  }
 
   async canActivate(route: ActivatedRouteSnapshot): Promise<boolean> {
     const id = route.params['id'];
     try {
       const properties = await this.uiExtensionsService.getProperties(id);
-      return !!properties && (ALLOWED_TYPES as readonly string[]).includes(properties.type);
+      return ALLOWED_TYPES.has(properties.type);
     } catch {
       return false;
     }
--- a/webapp/tests/karma/ts/modules/ui-extensions/ui-extensions-route.guard.provider.spec.ts
+++ b/webapp/tests/karma/ts/modules/ui-extensions/ui-extensions-route.guard.provider.spec.ts
@@ -31,29 +31,27 @@ describe('UiExtensionsTabRouteGuardProvider', () => {
     uiExtensionsService.getProperties.resolves({ id: 'ext-1', type: 'app_main_tab' });
     const result = await guard.canActivate(getRoute('ext-1'));
     expect(result).to.be.true;
+    expect(uiExtensionsService.getProperties).to.have.been.calledOnceWithExactly('ext-1');
   });
 
   it('should activate for app_drawer_tab type', async () => {
     uiExtensionsService.getProperties.resolves({ id: 'ext-2', type: 'app_drawer_tab' });
     const result = await guard.canActivate(getRoute('ext-2'));
     expect(result).to.be.true;
+    expect(uiExtensionsService.getProperties).to.have.been.calledOnceWithExactly('ext-2');
   });
 
   it('should not activate for unknown type', async () => {
     uiExtensionsService.getProperties.resolves({ id: 'ext-3', type: 'other_type' });
     const result = await guard.canActivate(getRoute('ext-3'));
     expect(result).to.be.false;
+    expect(uiExtensionsService.getProperties).to.have.been.calledOnceWithExactly('ext-3');
   });
 
   it('should not activate when extension properties not found', async () => {
     uiExtensionsService.getProperties.rejects(new Error('not found'));
     const result = await guard.canActivate(getRoute('missing-ext'));
     expect(result).to.be.false;
-  });
-
-  it('should pass the route id to getProperties', async () => {
-    uiExtensionsService.getProperties.resolves({ id: 'ext-4', type: 'app_main_tab' });
-    await guard.canActivate(getRoute('ext-4'));
-    expect(uiExtensionsService.getProperties.calledOnceWith('ext-4')).to.be.true;
+    expect(uiExtensionsService.getProperties).to.have.been.calledOnceWithExactly('missing-ext');
   });
 });


megha1807:
Hi @jkuester - I've addressed the build failures and the PR should now be ready for review. Here's a summary of what was fixed:

Rebased onto medic:10224-ui-extensions after #10670 was merged, resolving the conflicts in ui-extensions.service.ts and its spec.
Fixed the route guard - canActivate was not awaiting getProperties (which is async), causing a TypeScript error (Property 'type' does not exist on type 'Promise<UiExtensionProperties>'). Made canActivate async and wrapped it in a try/catch since getProperties throws when an id is not found.
Updated the route guard spec - changed .returns() to .resolves() and the "not found" case to .rejects() to match the async behaviour of getProperties.
Fixed the component spec - added the missing providers (NavigationService, AuthService, SessionService) and modules (TranslateModule, RouterTestingModule, MatIconModule) required by ToolBarComponent's dependency chain.

Please let me know if there's anything else to address!

webapp/src/ts/modules/ui-extensions/ui-extensions-tab.component.html
```diff
@@ -1,11 +1,15 @@
-<mm-tool-bar/>
+<mm-tool-bar [title]="extensionTitle | translate"/>
```
jkuester, line 1: I forgot to include this in the issue design.  We want this to set the title in the tool-bar at the top of the page.

webapp/src/ts/modules/ui-extensions/ui-extensions-tab.component.html
```diff
@@ -1,11 +1,15 @@
...
 <div class="inner">
   <div class="inbox page">
 
-    <div id="ui-element-tab" #uiElementTab>
-      <div class="loader" *ngIf="loading"></div>
+    <div class="empty-selection" *ngIf="loading">
+      <div><div class="loader"></div></div>
     </div>
 
+    <error-log *ngIf="!!errorStack && !loading" [errorStack]="errorStack" [errorFor]="'ui-extension'"></error-log>
```
jkuester, line 10: This is another detail that I did not consider until finally seeing this all coming together here.  If there is an error fetching/constructing/rendering the UI Extension, it will be helpful to show our standard error page (which also includes the error stacktrace).

jkuester (approved):
Very nice!  I am super excited to see this come together and now we have enough to have a working demo!  🤩 

Image

For anyone reading  this that is interested in reproducing this, you just need to manually create a doc (e.g. with Fauxton) with these properties:

```json
{
  "_id": "ui-extension:hello-world",
  "type": "app_main_tab",
  "title": "Hello Extension",
}
```

Then add an attachment called `extension.js` with the content:

<details><summary>extension.js</summary>
<p>

```js
const HELLO = 'Hello World';
const GOODBYE = 'Goodbye World';

module.exports = class HelloWorldComponent extends HTMLElement {
  /**
   * Called when element created (but before inserted on the DOM - no access to any context).
   */
  constructor() {
    super();
    this.attachShadow({ mode: 'open' });
    this.clickCount = 0;
    this.message = HELLO;
  }

  /**
   * Called when the element is added to a document (including being moved within the DOM)
   * Purpose: Initialize the component, set up event listeners, fetch data, or render content.
   */
  connectedCallback() {
    this.shadowRoot.innerHTML = `
      <style>
        :host {
          display: block;
          font-family: sans-serif;
        }
        h1 {
          margin-bottom: 1rem;
        }
        button {
          padding: 0.5rem 1rem;
          cursor: pointer;
        }
      </style>
      <h1 id="title">${this.message}</h1>
      <button id="ok-btn">Ok</button>
    `;

    this.shadowRoot
        .getElementById('ok-btn')
        .addEventListener('click', this.switchMessage.bind(this));
  }

  switchMessage() {
    const titleElement = this.shadowRoot.getElementById('title');
    this.message = this.message === HELLO ? GOODBYE : HELLO;
    titleElement.textContent = `${this.message}: ${++this.clickCount}`;
  }
}

```
</p>
</details> 

---

I did include a couple minor tweaks (noted inline) for things that I forgot to include in the original issue description.  One of which is to include our standard error visualization to help provide the user/admin with more info about what failed when loading the UI extension:

Image

megha1807:
Thanks for the review and for adding those tweaks directly @jkuester! Happy to see this coming together. Ready to merge whenever you are.

PR #10777, opened 2026-03-29
feat(#10766): support collapsing contact-summary cards

### LINKED ISSUE #10766: Support collapsing contact-summary cards

Contact-summary cards on the contact detail view are always fully expanded. In some cases (stock monitoring, immunizations, etc) the amount of details on the card means they take a lot of space on the screen.  To see content _below_ the card, users must do a lot of scrolling.  We want to allow users to collapse/expand cards by tapping the header.  Additionally, we can support a `collapsed` config property so cards can start collapsed.

This is a _tentitive_ technical design.

Use a click handler on the existing header div and `*ngIf` to toggle field visibility. 

### 1 `webapp/src/ts/modules/contacts/contacts-content.component.html` (lines 46-67)

- Add `(click)="card.collapsed = !card.collapsed"`, `[attr.aria-expanded]="!card.collapsed"`, and `role="button"` to the `.action-header` div
- Add a chevron icon (`fa-chevron-right` when collapsed, `fa-chevron-down` when expanded) inside the header
- Wrap the `.row.flex.grid` div with `*ngIf="!card.collapsed"`

### 2. `webapp/src/css/inbox.less` (around the `.action-header` block at line 605)

Add styles scoped to `.compact-card .action-header` for collapsible behavior:
- `cursor: pointer` for interactive affordance
- Use flexbox (replacing float) to position h3 left and chevron right
- Subtle hover background color
- Chevron icon styling (size, color, margin)

Note: The existing `.action-header` uses `float: left` on h3 and `overflow: auto` for clearfix. The collapsible variant should use flexbox with `justify-content: space-between` instead, scoped only to `.compact-card` cards so other action-headers (tasks, reports, children) are unaffected.

### 3. `webapp/tests/karma/ts/modules/contacts/contacts-content.component.spec.ts`

Add unit tests tests for:
- Cards default to expanded when `collapsed` is undefined/false
- Cards with `collapsed: true` start with fields hidden
- Clicking header toggles field visibility
- `aria-expanded` reflects current state
- State resets when selected contact changes

### 4. `tests/e2e/default/contacts/contact-details.wdio-spec.js`

Update one of the contact summary card tests to validate current contact-summary cards are rendered "open" by default, but can be collapsed by tapping on the header.

The current design I have provided here makes _all_ contact-summary cards "collapsible" (but only cards configured with `collapsed: true` are initially collapsed when the profile page loads).  Instead of allowing _any_ contact-summary card to be collapsed, we could make only certain cards collapsible (e.g. based on the card's config) while the other cards are not collapsible and just retain the default behavior.  From a code maintenance perspective, I think it is simpler to just make them all collapsible. So, as long as there is no concern about the UX, I would rather just keep it simple instead of needing logic to make some fixed and some collapsible....

Another thing we could also pursue in the future is making other sections on the contact profile page to also be collapsible.  If a user is on the profile page of a place and they want to see the "Reports", they first have to scroll past the contact-summary cards, People, and Tasks.  Each of these sections already has a header that theoretically could be updated to collapse the section just like we are doing for contact-summary cards....

**Additional context**

Originally discussed in a UI Extensions squad meeting.  @dianabarsan @binokaryg @eljhkrr feel free to add any additional details, corrections, or context.  👍

megha1807:
## Description

Closes #10766

Contact-summary cards on the contact detail view are always fully expanded. 
This PR adds support for collapsing/expanding cards by tapping the header.

### Changes made:

- Added click handler, `aria-expanded`, `role="button"`, and chevron icon 
  toggle to `.action-header` in `contacts-content.component.html`
- Added `*ngIf="!card.collapsed"` to the fields div to hide/show content
- Scoped flexbox collapsible styles to `.compact-card .action-header` 
  in `inbox.less` without affecting other action-headers (tasks, reports, children)
- Added unit tests covering:
  - Default expanded state
  - `collapsed: true` config support
  - Click toggle behavior
  - `aria-expanded` reflection
  - State reset on contact change
- Added E2E test validating open-by-default and collapse-on-tap behavior
- Created `contact-summary-config.js` for E2E test setup


megha1807:
Hi @jkuester , I've implemented the collapsible contact-summary cards feature and fixed the SonarCloud accessibility issues, but I'm struggling to get 3 unit tests passing:

should start collapsed when collapsed: true is set in config
should toggle collapsed state when header is clicked
should reset collapsed state when selected contact changes

All 3 fail with: expected <div class="row flex grid"> to not exist
I've tried initializing the collapsed state in withInitializedCards() inside reduxSubscription, but the store keeps re-emitting and overwriting the DOM state. Could you advise on the correct approach to initialize and preserve card collapsed state in this component? Is there something specific about how getSelectedContact should be used here that I'm missing?

webapp/tests/karma/ts/modules/contacts/contacts-content.component.spec.ts
```diff
@@ -362,4 +362,92 @@ describe('Contacts content component', () => {
...
+        ...selectedContact,
+        summary: cardSummary
+      });
+      store.refreshState();
+      fixture.detectChanges();
+    });
+
+    it('should default to expanded when collapsed is undefined', () => {
+      const compiled = fixture.nativeElement;
+      expect(compiled.querySelector('.row.flex.grid')).to.exist;
```
jkuester: For these unit test, I think your problem is that the selector here is too generic (and is matching another element on the page that is not affected by `collapsed`.  

Try switching all these assertions to use a more specific selector like this one:

```suggestion
      expect(compiled.querySelector('.card.compact-card .row.flex.grid')).to.exist;
```

webapp/src/ts/modules/contacts/contacts-content.component.ts
```diff
@@ -173,10 +197,9 @@ export class ContactsContentComponent implements OnInit, OnDestroy {
     ]) => {
       void this.recordSearchTelemetry(this.selectedContact, selectedContact, filters);
       if (this.selectedContact?._id !== selectedContact?._id) {
-        // reset view when selected contact changes
         this.resetTaskAndReportsFilter();
       }
-      this.selectedContact = selectedContact;
+      this.selectedContact = this.withInitializedCards(selectedContact);
```
jkuester: I think it will be more maintainable going forwards to just extract the summary cards from the `selectedContact` here and store them in a separate variable:

```ts
      this.summaryCards = (selectedContact.summary?.cards ?? []).map(card => ({ ...card }));
```

This should make the cards editable (so we can toggle `collapsed`).  Then just update the html to reference `summaryCards` instead of `selectedContact?.summary?.cards` in the `*ndFor` block.

Once we do that, I don't think we need any of the rest of this `initializeCards` logic (here or below).

jkuester (changes requested):
Sorry for the delayed response here. 😓   I think I figured out the issue for your unit tests (see the comment below).  

---

The toggling functionality seems good, but the styling needs some work:

Image

Image

Instead of having the button-styling leaking through for part of the title bar, what we want is something like the ["duplicate contact cards"](https://github.com/medic/cht-core/blob/b965953914d786826ed6160812cc7797d67ae4a9/webapp/src/ts/components/duplicate-contacts/duplicate-contacts.component.html#L1) (where the whole title bar is clickable and consistently styled):

Image

Image

I am not sure if it is possible/feasible to use the same `mat-expansion-panel` here, but it would be cool if that worked... 🤔 

---

Finally, while I was testing this, I realized that I think we need to make a change to https://github.com/medic/cht-conf in order for our new `collapsed` property to get set from the configuration.  Essentially we need to [update the emitter](https://github.com/medic/cht-conf/blob/c250f6b9a1152fdd7a6bae1a89fea9b2a57d4175/src/contact-summary/contact-summary-emitter.js#L125) to also include `collapsed: card.collapsed` in the returned object.

Once that is done, we should be able to manually test things:

- In cht-conf:
  - Install your cht-conf branch globally (so we can use it later): `npm i -g`
- In cht-core:
  - Run the server locally (`npm run dev-api`)
  - Update the [default contact summary cards](https://github.com/medic/cht-core/blob/b965953914d786826ed6160812cc7797d67ae4a9/config/default/contact-summary.templated.js#L40) configuration to include:

  ```js
  {
    label: 'Demo Card (Expanded by Default)',
    appliesToType: 'person',
    appliesIf: function () { return true; },
    fields: function () {
      return [
        { label: 'Description', value: 'This card has no collapsed property, so it is expanded by default.', width: 12 },
        { label: 'Name', value: thisContact.name, width: 6 },
        { label: 'Phone', value: thisContact.phone || 'N/A', width: 6 },
      ];
    }
  },
  {
    label: 'Demo Card (Collapsed by Default)',
    appliesToType: 'person',
    collapsed: true,
    appliesIf: function () { return true; },
    fields: function () {
      return [
        { label: 'Description', value: 'This card has collapsed: true, so it is collapsed by default.', width: 12 },
        { label: 'Name', value: thisContact.name, width: 6 },
        { label: 'Phone', value: thisContact.phone || 'N/A', width: 6 },
      ];
    }
  }
  ```
  - With a terminal open in the `cht-core/config/default` directory, upload the updated config with: `cht --local compile-app-settings upload-app-settings`.
  - Login to the webapp `http://localhost:5988/` and navigate to the profile page for a person contact (create one if needed).
  - You should see the two demo contact-summary cards displayed for the person.

megha1807:
@jkuester 
The "Compile the app" failure appears to be an infrastructure issue - the TLS certificate on staging.dev.medicmobile.org has expired (CERT_HAS_EXPIRED). This is unrelated to the changes in this PR.
Please have a look.

jkuester:
Not sure what happened there, but seems to be working now... :+1:

jkuester (changes requested):
The styling of the title bar is still off.  We want it to match the rest of the app and not look like a generic button....

Image

megha1807:
Hi @jkuester, I've addressed all the styling and test feedback:

- Fixed `.compact-card .action-header` button styles — added CSS reset properties (background: none, border: none, padding: 0, width: 100%, 
  font: inherit, color: inherit) so the header no longer looks like a  generic browser button
- Fixed LESS indentation — nested rules (&:hover, h3, chevrons) are now properly scoped inside the parent block
- Fixed E2E test — added waitForDisplayed() before all assertions so the test waits for the async contact summary to render before checking visibility

Regarding cht-conf: I understand the collapsed property needs to be emitted from cht-conf for the feature to work end-to-end in production. 
Could you point me to the exact file in cht-conf where the card object is returned/emitted so I can add collapsed: card.collapsed there? I want to make sure I'm editing the right place before submitting that change.

Please let me know if the styling still needs adjustment or if you'd prefer the mat-expansion-panel approach instead.

jkuester (changes requested):
Sorry for the late response here! 😓 Been working out from under a big backlog.

Definitely moving in the right direction here!  Thanks for sticking with it!

The styling is looking better (no more default button vibes), but now all the section headers on the contact profile page look a bit off... 🤔 

This is how they should look:  

Image

But with your changes they look like this: 

Image

We definitely need to fix that so all the headers are back to looking like they did originally.  The only "visual" change I think we want we want to make to the title of the contact-summary cards is to make it "clickable" and to have the spinning carrot arrow (like you have added).  Otherwise, its styling should match what we have now.

---

For the cht-conf changes, I am pretty sure that the `return` block in the [emitter code](https://github.com/medic/cht-conf/blob/c250f6b9a1152fdd7a6bae1a89fea9b2a57d4175/src/contact-summary/contact-summary-emitter.js#L125) that I linked to previously is all that needs to be changed.  Currently it is only setting `labal` and `fields`.  I think all we should need is to include `collapsed: card.collapsed` in the returned object... 🤔

jkuester (changes requested):
Getting even closer, but the styling of the title-bar on the contact-summary cards is still off.  

This is what you have now:

Image

And this is how it looks in `master`:

Image

We want the basic style to be the same (with the horizontal divider, etc.

megha1807:
Hi @jkuester 
Also submitted the cht-conf change to include the collapsed property in the emitter: https://github.com/medic/cht-conf/pull/812

megha1807:
Hello @jkuester ,
The two failing CI checks (ci-integration-all-k3d and ci-webdriver-default-workflows) appear to be unrelated to the changes in this PR. ci-integration-all-k3d is failing in audit.spec.js (POST/PUT auditing tests) which has no connection to the collapsible cards feature. ci-webdriver-default-workflows appears to be a container/infrastructure issue with no specific test failure tied to this PR's changes. Could you confirm these are pre-existing flaky tests?

jkuester (approved):
Okay, I had to tweak the styling a bit more so that everything was consistent with the existing UI.  I also updated the e2e tests to include the `collapsed` config in the contact summary so we do have true end-to-end coverage of this feature.

As long as I did not break anything with my test util updates, and the CI still works, this should be good to go!  🤞

jkuester:
Also, here is a demo of what the functionality looks like.  In this case the "Shown Un-collapsed" has `collapse: false`, "Shown Collapsed" has `collapse: true` and "Past pregnancy" has nothing set for `collapsed` (default is `false`):

[contact-summary-card.webm](https://github.com/user-attachments/assets/32f1afcf-2823-43d4-9c5d-fcb5dbd4fda8)

megha1807:
Thank you so much @jkuester for the styling tweaks, the e2e test updates, and the demo video - that really helps show the feature clearly! 🙏

Happy with where this has landed. Ready to merge whenever you are!

PR #10784, opened 2026-03-30
fix(#8974): correctly populate `end` field in forms

### LINKED ISSUE #8974: Form `end` meta field not correctly populated when the form ends
**Describe the bug**
CHT forms support several [meta fields](https://docs.communityhealthtoolkit.org/apps/reference/forms/app/#supported-xlsform-meta-fields) including `start` and `end` that are supposed to record the exact time the form began and the form ended.

Unfortunately, the `end` value is always just recorded as (almost) the exact same value as the `start` regardless of how long the user takes to fill out the form.

**To Reproduce**
Create a form that contains:

| type | name |
| --- | --- |
| start | start |
| end | end |

1. Begin the form
2. Wait a bit of time (at least a few seconds)
3. Submit the form
4. See that the different between `start` and `end` do not match the time waiting before submitting the form.

**Expected behavior**
The `end` value should contain the time that the form was submitted.

**Additional context**
Originally reported [on the forum](https://forum.communityhealthtoolkit.org/t/start-end-xlsform-meta-fields-returns-same-value/3416).

It appears this is a regression that happened as a result of the [Enekto Uplift](https://github.com/medic/cht-core/issues/6345) in `4.0.0`.  I tested and confirmed everything works as expected in `3.17.2`, but it is broken in `4.0.0`.

Debugging through the Enekto code flow, it appears the `start` and `end` values [are preloaded correctly](https://github.com/enketo/enketo/blob/37a7af1886743c7e20ddcb9f3b0f8708d184cd88/packages/enketo-core/src/js/preload.js#L75), but the `BeforeSave` event is never triggered for the form.

vikrantwiz02:
# Description

The `end` meta field in CHT forms was always recording the same value as `start`, regardless of how long the user took to fill out the form.

Fixes #8974 

### Root Cause

In `prepareForSave`, CHT was firing: ```typescript $('form.or').trigger('beforesave'); ```

This had **two problems**:

**1. Wrong event name**
In enketo-core's `event.js`, the `BeforeSave` event is defined as:
```javascript
function BeforeSave() {
    return new CustomEvent('before-save', { bubbles: true });
}
```
CHT was firing `'beforesave'` — missing the hyphen, so the event 
name never matched.

**2. Wrong dispatch mechanism**
In enketo-core's `preload.js`, the `end` timestamp is updated inside
a native DOM listener:
```javascript
this.form.view.html.addEventListener(events.BeforeSave().type, () => {
    value = that.form.model.evaluate('now()', 'string');
    o.dataNode.setVal(value, 'datetime');
});
```
jQuery's `.trigger()` only fires jQuery-bound listeners and **never reaches native `addEventListener` listeners**. Since enketo-core 7.x (introduced in the Enketo Uplift in 4.0.0) dropped jQuery entirely, this callback was silently never executing — leaving `end` frozen at its preloaded value which matched `start`.

### Fix

Replaced the jQuery trigger with a native DOM dispatch on the correct
element in `enketo.service.ts`:
```typescript
// Before
$('form.or').trigger('beforesave');

// After
form.view.html.dispatchEvent(new CustomEvent('before-save', { bubbles: true }));
```

# Code review checklist
- [ ] UI/UX backwards compatible: Test it works for the new design
(enabled by default). And test it works in the old design, enable
`can_view_old_navigation` permission to see the old design. Test it
has appropriate design for RTL languages.
or includes a migration. Any breaking changes documented in the
release notes.

# License

### DIFF
--- a/webapp/src/ts/services/enketo.service.ts
+++ b/webapp/src/ts/services/enketo.service.ts
@@ -16,6 +16,7 @@ import { TranslateFromService } from '@mm-services/translate-from.service';
 import { TranslateService } from '@mm-services/translate.service';
 import { CHTDatasourceService } from '@mm-services/cht-datasource.service';
 import { Qualifier, Report } from '@medic/cht-datasource';
+import { DOC_TYPES } from '@medic/constants';
 
 /**
  * Service for interacting with Enketo forms. This code is intended for displaying forms in the CHT as well as being
@@ -528,7 +529,7 @@ export class EnketoService {
   private create(formInternalId, contact) {
     return {
       form: formInternalId,
-      type: 'data_record',
+      type: DOC_TYPES.DATA_RECORD,
       content_type: 'xml',
       reported_date: Date.now(),
       contact: this.extractLineageService.extract(contact),
--- a/admin/src/js/controllers/delete-doc-confirm.js
+++ b/admin/src/js/controllers/delete-doc-confirm.js
@@ -1,3 +1,6 @@
+const constants = require('@medic/constants');
+const DOC_TYPES = constants.DOC_TYPES;
+
 angular.module('controllers').controller('DeleteDocConfirm',
   function (
     $scope,
@@ -13,7 +16,7 @@ angular.module('controllers').controller('DeleteDocConfirm',
       const doc = $scope.model.doc;
 
       doc._deleted = true;
-      if (doc.type === 'data_record' && doc.contact) {
+      if (doc.type === DOC_TYPES.DATA_RECORD && doc.contact) {
         doc.contact = ExtractLineage(doc.contact);
       }
 
--- a/admin/src/js/services/get-summaries.js
+++ b/admin/src/js/services/get-summaries.js
@@ -1,3 +1,6 @@
+const constants = require('@medic/constants');
+const DOC_TYPES = constants.DOC_TYPES;
+
 angular.module('inboxServices').factory('GetSummaries',
   function(
     $q,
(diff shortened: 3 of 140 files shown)


webapp/src/ts/services/enketo.service.ts
```diff
@@ -567,8 +567,7 @@ export class EnketoService {
     if (!valid) {
       throw new Error('Form is invalid');
     }
-
-    $('form.or').trigger('beforesave');
+    form.view.html.dispatchEvent(new CustomEvent('before-save', { bubbles: true }));
```
jkuester: Nice!  👍 To help emphasize the source/purpose of this event, can we actually use the definition coming from enketo-core?  (Same CustomEvent, but just helps emphasize that this event is _for_ enketo:

```ts
import events from 'enketo-core/src/js/event';

...

form.view.html.dispatchEvent(events.BeforeSave());
```

vikrantwiz02: Done, updated to use events.BeforeSave() here and in contacts-edit.component.ts.

vikrantwiz02:
@jkuester Done! I've made the same change in contacts-edit.component.ts so contact forms now also dispatch the correct before-save event via form.view.html.dispatchEvent(events.BeforeSave()).

Also applied your suggestion to use events.BeforeSave() from enketo-core/src/js/event in both places. Added a declare module declaration file (src/enketo-core.d.ts) to resolve the TypeScript error since enketo-core doesn't ship types for its internal JS files.

Tests updated too — added view.html.dispatchEvent to form mocks in the contacts-edit spec and a dedicated test verifying the event is dispatched correctly.

vikrantwiz02:
Thanks @jkuester, glad to see this finally ready. Really grateful for helping out with the final polish and conflicts.

Issue #1234, opened 2015-09-02
[enketo] datepicker is missing styles


alxndrsn:
Previous to switching to `npm` management of enketo, the bootstrap datepicker source was modified to append datepickers as a subelement of `#enketo-widget-container`, which allowed datepicker-dropdown to be styled correctly.

There should be a way to style these correctly without modifying bootstrap-datepicker.

Also, from #1201 it is requested that `!important` is not used in style directives.  This may be possible as part of this ticket.

PR #10786, opened 2026-03-31
feat(#10695): adds interaction tracking service

### LINKED ISSUE #10695: Implement telemetry to track task selection duration
The recent changes to the tasks page (filtering, sorting by priority, search) all have the potential to improve task list navigation for users. However we don't have a way to verify or quantify the improvement. 

Implement a sort of task selection telemetry score, where we track how much time the user spends on the task list selecting which task to complete. 
A successful "score" would only be recorded when the user actually finds the task they wanted to complete and they complete it. Task completion time should not be part of the score, as that time depends on the task form. 
A failed "score" would be recorded when the user opens a task and closes it, because they have selected the incorrect one. 

Add some sort of analytics, instead of telemetry, that would analyze user behavior. 

**Additional context**
Part of smart task prioritization squad.

dianabarsan:
<!--
Please use semantic PR titles that respect this format:

<type>(#<issue number>): <subject>

Quick example:

feat(#1234): add hat wobble
^--^(#^--^): ^------------^
|     |      |
|     |      + - > subject
|     |
|     + -------- > issue number
|
+ -------------- > type: build|feat|fix|perf|refactor|test|chore

https://docs.communityhealthtoolkit.org/contribute/code/workflow/#commit-message-format
-->

# Description
 Closes #10695                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                
  Adds an opt-in `InteractionTrackingService` that records user behavior on the tasks tab (list opens/scrolls/leaves, task opens/form submissions/completes/cancels, task-group navigation, task-filter usage) so analysts can study how CHWs work through their queues.                                                                                        
                                                                                                                                                                                                                                                                                                                                                                
  **Service** (`webapp/src/ts/services/interaction-tracking.service.ts`)                                                                                                                                                                                                                                                                                        
  - Behind a new `can_track_task_interactions` permission; no-op for users without it.
  - Events are buffered in memory and persisted in batches to a per-day local PouchDB (`interaction-YYYY-M-D-{user}`). Buffer flushes on a 50-event threshold, on session end (navigating away from `/tasks`), and on the page-level `visibilitychange` listener wired in `app.component`.                                                                      
  - Per-day cap of 500 events; further events are dropped silently.                                                                                                                                                                                                                                                                                             
  - On the next `init()` (typically next app load), any per-day DB that isn't "today" is aggregated into a single `type: 'interaction-log'` doc in the user's meta DB and the per-day DB is destroyed. The aggregate replicates to the server-side user-meta DB via the normal meta-DB sync.                                                                    
  - Aggregate `_id` is `interaction-{date}-{user}-{deviceId}`, so each (user, day, device) produces one doc — no day-spanning rows and no per-event rows in the meta store.                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                
  **Tasks integration**                                                                                                                                                                                                                                                                                                                                         
  - `tasks.component`, `tasks-content.component`, `tasks-group.component`, `tasks-sidebar-filter.component` all call into the service for their respective events.                                                                                                                                                                                              
  - `task_list:scroll` is throttled at 2 s.                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                
  **Tasks-group permission**                                                                                                                                                                                                                                                                                                                                    
  - Also adds `can_view_tasks_group` and gates the `/tasks/group` route on it. Bundled here because tracking the task-group flow only makes sense once that route is permission-controlled.                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                
  **Tests**                                                                                                                                                                                                                                                                                                                                                     
  - New Karma unit spec for the service (~900 lines) covering init/permission/user-context handling, buffering thresholds, day rollover, the 500-event cap, aggregation of stale per-day DBs, ownership filtering across users, and error tolerance.                                                                                                            
  - New e2e spec (`tests/e2e/default/tasks/interaction-tracking.wdio-spec.js`) drives the full flow under a frozen "yesterday" clock so aggregation can be triggered within a single test. The clock is installed as a Date-only stub via `browser.addInitScript` (full fake-timers break zone.js scheduling); each test wipes server-side `task` docs and the  
  user's server meta DB before login so the rules engine regenerates tasks and `waitForAggregateDoc` doesn't pick up a prior run's aggregate.                                                                                                                                                                                                                   
  - Existing Karma specs for the tasks components updated to mock the new service. 

### AI Disclosure

I used Claude Code for developing and testing. I have made all design decisions and reviewed code for correctness, clarity and alignment with our standards. 

# Code review checklist
<!-- Remove or comment out any items that do not apply to this PR; in the remaining boxes, replace the [ ] with [x]. -->



# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10695-interaction-log/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10695-interaction-log/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10695-interaction-log/docker-compose/cht-couchdb-clustered.yml)


# License


tests/e2e/default/tasks/config/tasks-breadcrumbs-config.js
```diff
@@ -24,8 +24,8 @@ module.exports = [
     title: 'person_create',
     appliesTo: 'contacts',
     appliesToType: ['person'],
-    appliesIf: function () {
-      return true;
+    appliesIf: function (contact) {
+      return contact.contact.role !== 'chw';
```
dianabarsan, line 28: This is so a task doesn't get created for the CHW, and instead only for the patients. It saves from needing to explicitly target clicking and completing specific tasks to display the task group page (now any task I click will yield the same result, but if I had a rogue task for the CHW, i would have had to add code to avoid clicking it).

dianabarsan:
Hi @sugat009 . I know queues are long and you're probably surprised by this showing up out of the blue. we've been discussing adding a way for tracking user behavior on the task page, to have a framework for evaluating any future changes we might make. this is part of some squad work, and i'm the only dev on the squad. so ... sorry for just dropping this one on you. 
i appreciate your time!

sugat009:
No worries @dianabarsan , I ran out of time today with other PRs. Will look into this first thing tomorrow.

dianabarsan:
No rush @sugat009 . just as long as it makes it by 5.2.0 release :D

webapp/src/ts/services/interaction-tracking.service.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  private async upsertSession(doc: InteractionLogDoc, { session, startedAt, events }: InteractionSession) {
+    const existingSession = doc.sessions.find(s => s.startedAt === startedAt);
+    if (existingSession) {
+      existingSession.events = events;
+    } else {
+      doc.sessions.push({ session, startedAt, events });
+    }
+
+    this.totalEventsToday = doc.sessions.reduce((sum, s) => sum + s.events.length, 0);
```
sugat009: **issue (blocking):** Daily cap halved by periodic save

This sum includes the current session's events because `existingSession.events = events` was just set on line 179. After a `save()` (which does not call `clearSession`), `this.events` still holds those same events, so the next `record()` evaluates `hasReachedDailyLimit = totalEventsToday + this.events.length` and double counts. Effective per-day cap collapses to ~250 once the save timer has fired.

Trace at `MAX = 500`, save fires at 250 events:

| T | Action | events.length | totalEventsToday | Cap check |
| :--- | :--- | :--- | :--- | :--- |
| 0..5m | 250 records | 250 | 0 | 250, OK |
| 5m | save fires | 250 | 250 | (after persist) |
| 5m+1 | record() | 250 | 250 | 500, dropped |

**Fix:** exclude the current session when recomputing.

```ts
this.totalEventsToday = doc.sessions
  .filter(s => s.startedAt !== startedAt)
  .reduce((sum, s) => sum + s.events.length, 0);
```

Restores the `otherEventCount` semantics from `f03e9d318` that were dropped in `e61eafe15`. Composes cleanly with the `clearSession` fix below.

<details><summary>Failing karma test</summary>

```ts
it('should not double count current session events after a periodic save', async () => {
  metaDb.get.rejects({ status: 404 });

  service.startSession('tasks');
  for (let i = 0; i < 250; i++) {
    service.record(`event_${i}`);
  }
  await service.save();

  metaDb.get.resolves(metaDb.put.args[0][0]);
  metaDb.put.resetHistory();

  service.record('should_record');
  await service.flush();

  const finalEvents = metaDb.put.args[0][0].sessions.flatMap(s => s.events);
  expect(finalEvents.map(e => e.action)).to.include('should_record');
});
```

</details>

webapp/src/ts/services/interaction-tracking.service.ts
```diff
@@ -0,0 +1,267 @@
...
+    const deviceId = this.telemetryService.getUniqueDeviceId();
+    return `interaction-${date}-${user}-${deviceId}`;
+  }
+
+  private getTodayDate(): string {
+    return moment().format('YYYY-MM-DD');
+  }
+
+  private clearSession(): void {
+    this.currentSession = null;
```
sugat009: **issue (blocking):** Midnight rollover silently blocks day N+1

`clearSession()` resets session state but not `totalEventsToday`. Once the cap is hit on day N and the session is flushed, `totalEventsToday` is frozen at 500. On day N+1:

1. `startSession('tasks')` runs, does not touch `totalEventsToday`.
2. First `record()` evaluates `hasReachedDailyLimit` as `500 + 0 = 500 >= 500`, returns early at line 97.
3. No event is buffered → no flush → `_persist` (gated by `hasData` at line 150) never runs → `upsertSession` never recomputes `totalEventsToday` against day N+1's empty doc.
4. Service is silently off until the page reloads.

For long running CHW Android sessions this means the feature self-disables the day after any cap-hit. Combined with the cap halving issue above, the trigger is much easier than 500.

**Fix:** track the date of the last persist and reset on rollover.

```ts
private lastPersistedDate: string | null = null;

private hasReachedDailyLimit(): boolean {
  const today = this.getTodayDate();
  if (this.lastPersistedDate && this.lastPersistedDate !== today) {
    this.totalEventsToday = 0;
    this.lastPersistedDate = null;
  }
  return (this.totalEventsToday + this.events.length) >= MAX_EVENTS_PER_DAY;
}

// In upsertSession alongside the totalEventsToday recompute:
this.lastPersistedDate = this.getTodayDate();
```

**Related (non-blocking):** sessions that straddle midnight have a secondary issue. `_persist` resolves the doc id at persist time, so events from 23:30 day N and 00:30 day N+1 land in day N+1's doc. After the rollover fix the cap is correct, but events end up in the wrong daily bucket for analytics. Cleanest resolution: flush and start a fresh session when the date changes.

<details><summary>Failing karma test</summary>

```ts
it('should reset the daily counter at midnight rollover', async () => {
  const day1Events = new Array(500).fill({ action: 'x', timestamp: 1 });
  metaDb.get.resolves({
    _id: 'interaction-2026-03-27-greg-device-uuid-123',
    _rev: '1-abc',
    type: 'interaction-log',
    sessions: [{ session: 'tasks', startedAt: 1, events: day1Events }],
    metadata: { user: 'greg', deviceId: 'device-uuid-123', date: '2026-03-27', versions: ['4.5.0'] },
  });

  service.startSession('tasks');
  service.record('trigger_read');
  await service.flush();

  await clock.tickAsync(15 * 60 * 60 * 1000);

  metaDb.get.rejects({ status: 404 });
  metaDb.put.resetHistory();

  service.startSession('tasks');
  service.record('day_2_event');
  await service.flush();

  expect(metaDb.put.calledOnce).to.be.true;
  expect(metaDb.put.args[0][0]._id).to.equal('interaction-2026-03-28-greg-device-uuid-123');
});
```

</details>

sentinel/src/schedule/replications.js
```diff
@@ -102,14 +102,16 @@ const batchedPurge = (sourceDb, targetDb, lastSeq = 0) => {
     .then(nextSeq => nextSeq && batchedPurge(sourceDb, targetDb, nextSeq));
 };
 
-const isTelemetryOrFeedback = (docId) => docId.startsWith('telemetry-') || docId.startsWith('feedback-');
+const isReplicableDoc = (docId) => {
+  return docId.startsWith('telemetry-') || docId.startsWith('feedback-') || docId.startsWith('interaction-');
```
sugat009, line 106: **issue (blocking):** Webapp purger leaves `interaction-` docs in IndexedDB indefinitely

This sentinel filter correctly includes `interaction-`. The webapp side does not. `webapp/src/js/bootstrapper/purger.js:68` filters on:

```js
const isFeedbackOrTelemetryDoc = change => change.id.startsWith('telemetry-') || change.id.startsWith('feedback-');
```

`interaction-*` docs sync upstream and are server-side purged after this filter copies them to `users-meta`. But CouchDB `_purge` does not propagate through normal sync, and the local purger's prefix check skips `interaction-`, so the device's IndexedDB never learns about it. ~50 KB per daily doc × 365 days = 15-20 MB/year of unbounded growth on storage-constrained CHW Androids.

Safety is fine: the local purger only deletes docs with `change.seq <= synced_seq`, so nothing is dropped before upstream sync. Adding `interaction-` inherits that safety.

**Fix (option A, minimal):** in `webapp/src/js/bootstrapper/purger.js`,

```js
const isReplicableMetaDoc = change =>
  change.id.startsWith('telemetry-') ||
  change.id.startsWith('feedback-') ||
  change.id.startsWith('interaction-');
```

**Fix (option B, cleaner):** extract the prefix list to `shared-libs/constants` so this file and the webapp purger share one source of truth. Worth doing now while only one new consumer is being added (also addresses the prefix duplication noted in another comment).

No existing test covers this. Suggest extending `webapp/tests/mocha/unit/purger.spec.js:228` with an `interaction-2026-03-27-user-device` fixture in the change feed and asserting it lands in the `bulkDocs` purge list.

webapp/src/ts/services/interaction-tracking.service.ts
```diff
@@ -0,0 +1,267 @@
...
+    const startedAt = this.sessionStartedAt!;
+    const events = [...this.events];
+
+    try {
+      const metaDb = this.dbService.get({ meta: true });
+      const doc = await this.getOrCreateDailyDoc(metaDb);
+      await this.upsertSession(doc, { session, startedAt, events });
+      await metaDb.put(doc);
+    } catch (error) {
+      console.error('Error saving interaction log', error);
```
sugat009: **issue (blocking):** Concurrent `_persist` race drops events on 409 and flush clears the buffer anyway

`_persist` does an unguarded read-modify-write against the daily doc, and `_flush(true)` calls `clearSession()` even when this catch swallowed a 409. When `save()` and `flush()` race, both reads pull the same `_rev`, the first put wins, the second 409s and is swallowed, and `clearSession()` runs anyway. Events queued between the two snapshots are lost.

Seven triggers all reach `_persist` without coordinating: cap hit, 5 minute save timer, 20 minute inactivity, route unmount, tab hidden, tab closing, prior session in `startSession`. Any two can overlap.

Race pattern:

| T | Action |
| :--- | :--- |
| 0 | save() snapshot S1=[e1,e2,e3], awaits get |
| 1 | record('e4') → events=[e1,e2,e3,e4] |
| 2 | flush() snapshot S2=[e1,e2,e3,e4], awaits get |
| 3 | both gets resolve with rev R0 |
| 4 | save's put succeeds (R1) |
| 5 | flush's put with R0 → 409, caught, swallowed |
| 5 | clearSession() → events=[]. e4 is gone. |

**Existing precedent:** `telemetry.service.ts` already has both pieces. The `isAggregationRunning` flag at line 18, `storeConflictedAggregate` defined at line 134 and called from the 409 catch at line 198, and the guard usage at lines 228-235. The new service has neither.

**Fix (minimum, in-flight guard):**

```ts
private _persistInFlight: Promise<void> | null = null;

private _persist(): Promise<void> {
  if (this._persistInFlight) {
    return this._persistInFlight;
  }
  this._persistInFlight = this._doPersist().finally(() => {
    this._persistInFlight = null;
  });
  return this._persistInFlight;
}

private async _doPersist() {
  // existing _persist body, with one bounded retry on 409
}
```

Parity with telemetry's shadow doc fallback is a further improvement; the guard alone closes the data-loss path because flush awaits save's in-flight promise, then on its own pass reads a fresh `_rev`.

<details><summary>Failing karma test</summary>

```ts
it('should not lose events when save and flush race', async () => {
  // see review thread, ~70 lines: sets up two pending gets sharing rev R0,
  // first put resolves, second rejects 409, asserts e4 lands in a successful put
});
```

</details>

webapp/src/ts/services/interaction-tracking.service.ts
```diff
@@ -0,0 +1,267 @@
...
+    date: string;
+    versions: string[];
+  };
+}
+
+@Injectable({
+  providedIn: 'root'
+})
+export class InteractionTrackingService {
+  private static readonly PERMISSION = 'can_track_task_interactions';
```
sugat009, line 50: **issue (blocking):** `can_track_task_interactions` not registered in default app settings

This constant gates `init()` via `authService.has(...)`, but the permission is never declared in `config/default/app_settings.json`. The other task permissions (`can_view_tasks`, `can_view_tasks_tab`, `can_view_tasks_group`) are listed near line 256 of that file; this one is absent.

**Verified live on a fresh local instance:**

1. Created an offline CHW user.
2. Opened Tasks, clicked tasks, navigated away.
3. Queried meta DB for interaction docs: 0.
4. `init()` set `enabled=false` because `authService.has('can_track_task_interactions')` was false for the `chw` role.
5. Had to `PUT /medic/settings` directly with the missing entry for the feature to activate.

[Per the CHT docs](https://docs.communityhealthtoolkit.org/community/contributing/code/core/feature-flags/): *"The permission will not be included in `app_settings` automatically so if you want it to be selectable from the Admin app user interface, you will need to add it to `app_settings`."* No admin UI exposes a permission absent from the settings doc, and `cht-conf upload-app-settings` from a config that doesn't declare it won't help either. The only working path is a direct settings PUT.

The e2e spec at `tests/e2e/default/tasks/interaction-tracking.wdio-spec.js:109, 137, 197` patches permissions via `utils.updatePermissions(...)` (a settings PATCH), which is why CI is green. Production deployments don't follow that path.

**Fix:** add the permission to `config/default/app_settings.json` near `can_view_tasks_group`:

```diff
     "can_view_tasks_group": [
       "chw"
     ],
+    "can_track_task_interactions": [],
```

Empty array keeps the feature off by default; operators grant it via the role editor or `cht-conf`.

**Preventive (non-blocking, suggest a follow up issue):** add a unit test that scans service `@Injectable` classes for string literals matching `/^can_[a-z_]+$/` and asserts each is a key in `config/default/app_settings.json > permissions`. Catches this class of omission at PR time.

webapp/src/ts/services/interaction-tracking.service.ts
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  private async getAppVersion(): Promise<string> {
+    try {
+      const { version } = await this.versionService.getServiceWorker();
+      return version || 'unknown';
+    } catch {
+      return 'unknown';
+    }
+  }
```
sugat009, line 481: **issue (blocking):** Uncached HTTP request on every `_persist`

`upsertSession` calls this on every persist (line 186). `versionService.getServiceWorker()` does a raw `http.get('/deploy-info.json')` with no caching:

```ts
async getServiceWorker(): Promise<{ version }> {
  const obs = this.http.get('/deploy-info.json');
  ...
}
```

Two consequences:

1. **Request volume.** An 8 hour CHW shift with the 5 minute save interval fires ~96 persists, plus extras on navigation, visibility change, and `beforeunload`. Each makes the round trip. Line 187 already deduplicates `metadata.versions[]`, so almost all are wasted.
2. **Offline behavior.** When offline the request fails and falls through to `'unknown'`. CHWs are often partially offline; persists may land with `versions: ['unknown', '5.1.0', 'unknown', ...]` depending on connectivity flapping.

**Existing patterns:**

- `telemetry.service.ts:78` reads `ddoc?.build_info?.version` from the locally synced `_design/medic-client`. No HTTP, offline safe.
- `feedback.service.ts:194` calls `getServiceWorker()` only on user initiated submission (rare, online).

**Fix (option B, cleaner, matches telemetry):**

```ts
private async getAppVersion(): Promise<string> {
  try {
    const { version } = await this.versionService.getLocal();
    return version || 'unknown';
  } catch {
    return 'unknown';
  }
}
```

`versionService.getLocal()` reads the ddoc from PouchDB; offline safe, effectively free after first call. Resolves both the request volume and the offline correctness.

**Fix (option A, minimal, just memoize):** cache in a private field, do not cache `'unknown'` so it retries next persist. Resolves volume only.

Suggest extending the existing test at `interaction-tracking.service.spec.ts:361-373` with a check that the version is fetched at most once across multiple `_persist` calls.

webapp/src/ts/modules/tasks/tasks-content.component.ts
```diff
@@ -154,6 +156,9 @@ export class TasksContentComponent implements OnInit, OnDestroy {
       return;
     }
 
+    const taskIndex = this.tasksList.indexOf(task);
+    this.interactionTrackingService.record('task:open', task.title, String(taskIndex));
```
sugat009: **issue (non-blocking):** `task.title` passed verbatim as `ref`

The service docstring at `interaction-tracking.service.ts:88-89` says `ref` is a "non-PII reference (e.g., task type id, form id)". `task.title` can violate that depending on deployment configuration.

After checking the kapa-docs and the cht-core deepwiki, the residual PII risk is much narrower than it first appears:

1. Default tasks (`config/default/tasks.js`) use translation keys resolving to static labels via `translateService.instant`. No interpolation; no PII.
2. The `{{contact.name}}` interpolation pattern documented in `translate-from.service.ts:17` JSDoc does **not** actually run at runtime. Reasons: lodash's default `_.templateSettings.interpolate` is `<%= %>` (no global override exists in `webapp/src`); cht-conf's declarative schema enforces `title: joi.string()`, blocking the array form; ngx-translate's missing-key handler returns the key unchanged.
3. Official CHT design docs explicitly state patient names should not be in task titles ("anti pattern").

**Live verification.** I clicked tasks on a CHW user and inspected the meta doc:

```json
{ "action": "task:open", "timestamp": 1777045444125, "ref": "Home visit for {{contact.name}}", "detail": "0" }
```

Literal template string written verbatim. Confirms the path is real, confirms a deployment relying on the JSDoc interpolation pattern would not actually leak PII through this field.

**Suggested fix (defense in depth):** use a config level identifier rather than the rendered title at all three call sites (this one, `tasks-group.component.ts:253`, and the task type filter at `task-type-filter.component.ts:43`):

```ts
this.interactionTrackingService.record(
  'task:open',
  task.emission?.actions?.[0]?.form ?? task.emission?._id,
  String(taskIndex),
);
```

`task.emission?.actions?.[0]?.form` is already used safely for `task:form_open` and is known to be a config level form id. Same identifier should feed the task type filter (same root issue as the facility filter comment).

Severity is Low because modern conformant CHT does not actually expose PII through this path. Worth landing for contract alignment.

webapp/src/ts/modules/tasks/tasks-group.component.ts
```diff
@@ -245,6 +249,8 @@ export class TasksGroupComponent implements OnInit, OnDestroy {
 
     const emissionId = this.getEmissionIdFromNavigation();
     if (emissionId && this.isGroupTask(emissionId)) {
+      const task = this.tasks.find((t: any) => t?._id === emissionId);
+      this.interactionTrackingService.record('task_group:select', task?.title);
```
sugat009: **issue (non-blocking):** Same `task.title` pattern

Same fix shape as the comment on `tasks-content.component.ts:160`: substitute a config level identifier for the rendered title.

webapp/src/ts/modules/tasks/tasks-sidebar-filter.component.ts
```diff
@@ -148,16 +152,22 @@ export class TasksSidebarFilterComponent implements OnInit, AfterViewInit, OnDes
       const count = this.filterCount[filter.fieldId] || 0;
       if (count > 0) {
         this.telemetryService.record(`tasks:filter:apply:${filter.fieldId}`, count);
+        const selected = Array.from(filter.filter?.selected || []).join(',');
+        this.interactionTrackingService.record('task_filter:select', filter.fieldId, selected);
```
sugat009: **issue (non-blocking):** Filter `detail` leaks facility UUIDs (plus the same `task.title` issue on the task-type filter)

When the user filters by **facility**, `selected` contains place `doc._id` values (UUIDs). Joined and recorded as the `detail` of `task_filter:select`, they replicate to `users-meta` and are retained indefinitely (the `users-meta` retention gap is in another comment). Stable identifiers tied to real places, resolvable to facility name + hierarchy by anyone with read access to both `users-meta` and `medic`. Combined with the session trace, you get a location-correlated activity profile per CHW.

When the user filters by **task type**, the `task-type-filter.component.ts:43` populates its `selected` set from `task.title` strings, same root issue as the `ref` parameter on `tasks-content.component.ts:160`.

Lower severity than the `task.title` issue because facility UUIDs aren't directly human readable and the main DB already holds them. The new exposure is the per-click timeline of facility filter use.

**Suggested fix:**

- If the analytics goal is "did the CHW apply a facility filter", record only the count, not the selected UUIDs.
- For the task type filter, populate from a stable id (form id / emission id) rather than `task.title` (same change as the `ref` fix on `tasks-content.component.ts:160`).

If specific facilities matter for the analysis, document the retention and replication implications in the feature docs.

webapp/src/ts/app.component.ts
```diff
@@ -303,6 +305,7 @@ export class AppComponent implements OnInit, AfterViewInit {
     this.setupDb();
     this.countMessageService.init();
     this.feedbackService.init();
+    this.interactionTrackingService.init();
```
sugat009: **suggestion (non-blocking):** `init()` not awaited

`init()` is async; the returned promise is ignored. Until it resolves, `enabled` is `false` and every `startSession` is a silent no-op (`interaction-tracking.service.ts:72-74`). Race with `TasksComponent.ngOnInit` is narrow but real on slow offline boots and deep links to `/tasks` on first boot.

**Fix:** put `init()` into the `setupPromise` chain at line 313, or queue pending operations inside the service until `init()` resolves. The setupPromise approach is the minimal change.

webapp/src/ts/app.component.ts
```diff
@@ -725,6 +728,14 @@ export class AppComponent implements OnInit, AfterViewInit {
   private stopWatchingChanges() {
     // avoid Failed to fetch errors being logged when the browser window is reloaded
     this.changesService.killWatchers();
+    this.interactionTrackingService.flush();
```
sugat009: **suggestion (non-blocking):** `beforeunload` flush is unreliable; bundled inside `stopWatchingChanges`

Two issues:

1. Browsers tear down the page when the handler returns; async IndexedDB operations spawned by `flush()` are commonly aborted. The returned promise is discarded. `visibilitychange` + `save()` at line 734 is the reliable path for most tab-hide scenarios, so the beforeunload flush mainly provides false assurance.
2. The flush is a hidden side effect inside a method named `stopWatchingChanges`. Reviewers reading the handler name expect only the changes feed to be torn down.

**Fix:** either (a) rely solely on `visibilitychange` + `save()` and drop the beforeunload flush, or (b) keep it but rename the method (`onBeforeUnload`) and add a comment that this is fire-and-forget.

sentinel/src/schedule/replications.js
```diff
@@ -102,14 +102,16 @@ const batchedPurge = (sourceDb, targetDb, lastSeq = 0) => {
...
 const replicateDb = (sourceDb, targetDb) => {
-  // Replicate only telemetry and feedback docs
+  // Replicate telemetry, feedback, and interaction-log docs
   return sourceDb.info().then(info => {
     return sourceDb.replicate
       .to(targetDb, {
-        filter: doc => !doc._deleted && isTelemetryOrFeedback(doc._id),
+        filter: doc => !doc._deleted && isReplicableDoc(doc._id),
       })
       .then(() => purgeFeedback(sourceDb, info, targetDb));
```
sugat009, line 116: **chore:** `users-meta` has no retention policy for interaction docs

This nightly job replicates `interaction-*` docs to `users-meta` and then purges the source. The destination is never trimmed: `purgeFeedback` calls `purgeDocs(sourceDb, ...)` only.

For telemetry and feedback the same architectural choice already exists, so this PR extends an existing pattern rather than creating a new one. But interaction docs carry richer per-event behavioral data than telemetry aggregates or sporadic feedback submissions, so the retention gap matters more. At ~50 KB per doc × 500 CHWs × 365 days that's ~9 GB/year per deployment with no bounded lifetime.

GDPR/HIPAA implications depend on whether interaction metadata is classified as health-related. Given events include `task:form_save` with form id, a task-opening timeline is close to a care pathway log.

Not a merge blocker, but worth a follow up issue: add a `users-meta` cleanup pass to the sentinel daily cron with a configurable retention window (e.g., 90-180 days default), or document a manual operator retention procedure in the feature docs.

webapp/src/ts/services/interaction-tracking.service.ts
```diff
@@ -0,0 +1,267 @@
...
+          versions: [],
+        },
+        sessions: [],
+      };
+    }
+  }
+
+  private getDailyDocId(): string {
+    const date = this.getTodayDate();
+    const user = this.sessionService.userCtx()?.name || 'unknown';
```
sugat009: **issue (non-blocking):** User/device context captured at persist time, not session start

`user` here and on line 204, plus `deviceId` on lines 205 and 217, are read fresh on every `_persist`, not captured at `startSession` time. If session context changes between a session's start and its persist (session expiry, re-login in the same Angular instance, an Android WebView where the app is not fully torn down), events from user A's session can land in user B's daily doc.

In practice the CHT session expiry path triggers `document.location.href` → full reload, so the singleton state resets. Risk is narrow but exists in unusual deploy environments (Android WebView configs where Angular is not torn down, automation frameworks that do in-page session swaps).

**Fix:** snapshot `user` and `deviceId` at `startSession` and use captured values throughout the session. Add a `reset()` method that clears them plus `enabled`, called on auth change.

webapp/src/ts/services/interaction-tracking.service.ts
```diff
@@ -0,0 +1,267 @@
...
+      return await metaDb.get(dailyDocId);
+    } catch (err: any) {
+      if (err.status !== 404) {
+        throw err;
+      }
+      return {
+        _id: dailyDocId,
+        type: 'interaction-log',
+        metadata: {
+          user: this.sessionService.userCtx()?.name,
```
sugat009: **nitpick:** `metadata.user` fallback inconsistency

`getDailyDocId` at line 216 falls back to `'unknown'` when `userCtx()?.name` is undefined. This line does not: it stores `metadata.user: undefined` (serializes as missing field). So the doc id says `unknown` but `metadata.user` is missing.

**Fix:** apply the same `|| 'unknown'` fallback here. One-liner.

webapp/src/ts/services/interaction-tracking.service.ts
```diff
@@ -0,0 +1,267 @@
...
+  async init(): Promise<void> {
+    this.enabled = await this.authService.has(InteractionTrackingService.PERMISSION);
+  }
+
+  /**
+   * Start a new interaction session. If a session is already active, it is flushed first.
+   * @param session Name of the session context (e.g., 'tasks')
+   */
+  startSession(session: string): void {
+    this.ngZone.runOutsideAngular(() => {
```
sugat009: **suggestion (non-blocking):** `runOutsideAngular` wraps full method body

Each public method (this one, `record` at 93, `flush` at 139, `save` at 146) wraps the entire body, including sync state mutations with no change-detection implications, in `runOutsideAngular`. After `await`, the continuation resumes in whatever zone was captured at the await site, so `await metaDb.get(...)` inside `_persist` may still execute its continuation inside the Angular zone if not explicitly wrapped at that call site.

`telemetry.service.ts` wraps only the call to the async body, not the entire public method.

**Fix:** move `runOutsideAngular` to wrap the async body inside `_persist`, drop it from the sync portions. Consistency with telemetry, clearer intent.

webapp/src/ts/services/interaction-tracking.service.ts
```diff
@@ -0,0 +1,267 @@
...
+        sessions: [],
+      };
+    }
+  }
+
+  private getDailyDocId(): string {
+    const date = this.getTodayDate();
+    const user = this.sessionService.userCtx()?.name || 'unknown';
+    const deviceId = this.telemetryService.getUniqueDeviceId();
+    return `interaction-${date}-${user}-${deviceId}`;
```
sugat009: **chore:** `interaction-` prefix hardcoded in three files

After the webapp purger fix:
- this line
- `sentinel/src/schedule/replications.js:106`
- `webapp/src/js/bootstrapper/purger.js:68`

`telemetry-` and `feedback-` already have the same problem pre-existing, so this PR is consistent with the existing pattern. But the moment to consolidate is now while only one new consumer is being added.

**Suggested follow up:** extract `META_DOC_PREFIXES = ['telemetry-', 'feedback-', 'interaction-']` in `shared-libs/constants` and reference it from all three. Non-blocking.

webapp/tests/karma/ts/modules/tasks/tasks.component.spec.ts
```diff
@@ -54,6 +56,7 @@ describe('TasksComponent', () => {
     };
     lineageModelGeneratorService = { reportSubjects: sinon.stub().resolves([]) };
     telemetryService = { record: sinon.stub() };
+    interactionTrackingService = { startSession: sinon.stub(), record: sinon.stub(), flush: sinon.stub() };
```
sugat009: **suggestion:** Component specs lack argument assertions

This spec, plus `tasks-content.component.spec.ts`, `tasks-group.component.spec.ts`, and `tasks-sidebar-filter.component.spec.ts`, each stub the service with `{ startSession, record, flush }` but never assert which action name, ref, or detail is passed at any lifecycle point. An accidental rename from `'task_list:open'` to `'task_list:opened'` would break analytics silently and all four spec files would still pass.

The e2e spec at `interaction-tracking.wdio-spec.js:174-189` does assert the full event sequence, which mitigates the risk for the happy path but only runs in e2e CI.

**Fix:** per component, add one or two `calledWith` assertions at key lifecycle points. Adds <20 lines total across all four files.

webapp/tests/karma/ts/services/interaction-tracking.service.spec.ts
```diff
@@ -0,0 +1,642 @@
...
+      await service.flush();
+
+      expect(metaDb.put.args[0][0].sessions[0].events).to.deep.equal([
+        { action: 'task_list:scroll', timestamp: NOW },
+        { action: 'task:open', timestamp: NOW + 100, ref: 'task_a', detail: '0' },
+        { action: 'task_list:scroll', timestamp: NOW + 200 },
+      ]);
+    });
+
+    it('should stop recording new events after flush reveals daily limit is reached', async () => {
```
sugat009: **suggestion:** Missing unit tests for edge case paths

This spec covers happy path, permission gating, dedup, version handling well (650+ lines, 30+ tests). Gaps:

1. Midnight rollover.
2. Concurrent save and flush.
3. Double count after save.
4. `metadata.user = 'unknown'` when `userCtx()` lacks a `name`.
5. Periodic save interval continuing after a cap-triggered flush.
6. User context change between `startSession` and `_persist`.

The bar for testing here is unusually high relative to the rest of the webapp; I would not block on individual gaps. The three regressions covered by the failing karma snippets in the inline comments above are the high value adds.

tests/e2e/default/tasks/interaction-tracking.wdio-spec.js
```diff
@@ -0,0 +1,361 @@
...
+};
+
+const clearInteractionDocsFromMetaDb = async ({ username }) => {
+  const docs = await getInteractionDocsFromMetaDb({ username });
+  docs.forEach(doc => doc._deleted = true);
+  await utils.requestOnTestMetaDb({ userName: username, path: '/_bulk_docs', method: 'POST', body: { docs } });
+};
+
+const getInteractionDocFromBrowser = async (expectedCount = 1) => {
+  await browser.pause(100);
```
sugat009: **nitpick:** Fragile timing in the e2e spec

Three fragility points in this spec:

1. Line 39: `await browser.pause(100)` before querying the meta DB. Race under load.
2. Line 57: `Object.defineProperty(document, 'hidden', { value: true, ... })` followed by `window.dispatchEvent(new Event('visibilitychange'))`. `visibilitychange` conventionally fires on `document`, not `window`. Works today via Angular host binding bubbling, but fragile.
3. Lines 61, 65: hardcoded `browser.pause(1000)` waits.

**Fix:** replace `browser.pause` with `browser.waitUntil` for the doc to appear; dispatch `visibilitychange` on `document` rather than `window`.

tests/e2e/default/tasks/interaction-tracking.wdio-spec.js
```diff
@@ -0,0 +1,361 @@
...
+  return metaDocs.rows.map(({ doc }) => doc);
+};
+
+const clearInteractionDocsFromMetaDb = async ({ username }) => {
+  const docs = await getInteractionDocsFromMetaDb({ username });
+  docs.forEach(doc => doc._deleted = true);
+  await utils.requestOnTestMetaDb({ userName: username, path: '/_bulk_docs', method: 'POST', body: { docs } });
+};
+
+const getInteractionDocFromBrowser = async (expectedCount = 1) => {
```
sugat009: **nitpick:** `getInteractionDocFromBrowser` mixes fetch and assertion

Named `get...` but also asserts a specific row count. Failure modes are unclear (doc not found vs. wrong count).

**Fix:** split into a plain `getInteractionDocsFromBrowser` returning rows; let callers assert.

sugat009 (changes requested):
A few findings. Most are inline; two below have no specific line to anchor against.

Three of the must fix issues come with failing karma tests I ran locally against the current branch head. The PII narrowing on `task.title` and the deployment gap on `can_track_task_interactions` were also verified against the CHT docs and a fresh local instance.

---

**issue (blocking):** PR description vs. implemented limit

The PR description says "max 200 sessions/day, max 500 events/session" (~100k events/day implied). The code at `interaction-tracking.service.ts:41` is `MAX_EVENTS_PER_DAY = 500` total, no session count cap.

Git history:

1. `7fcb31d09` set `MAX_EVENTS_PER_SESSION = 500` (matched the description).
2. `f03e9d318` switched to `MAX_EVENTS_PER_DAY = 2000` with an `otherEventCount` to avoid double counting the current session.
3. `e61eafe15` ("reduce max events to 500") cut to 500 and dropped `otherEventCount`. Origin of the daily cap halving issue, also where the description drifted.

Combined with the cap halving (see inline comment on `interaction-tracking.service.ts:184`), the effective ceiling for an active session is ~250. Could you clarify the intended semantics? Options:

1. 500 per session + session count cap (matches description).
2. 500 per day total (needs the cap halving fix + description update).
3. Higher per day + description update + cap halving fix.

---

**chore (non-blocking):** Unrelated refactors in the diff

Four bits of work with no functional link to interaction tracking:

1. `webapp/src/ts/app.component.ts`: `private` → `private readonly` on 30+ constructor params.
2. `tests/utils/sentinel.js` + `tests/integration/sentinel/schedules/purging.spec.js`: `waitForPurgeCompletion` now returns the purge log directly.
3. `tests/utils/index.js`: `stringifyParam` extracted from `getRequestUri`.
4. `tests/e2e/default/tasks/config/tasks-breadcrumbs-config.js`: `appliesIf` tightened.

Each is fine on its own; together they inflate the review surface. For future PRs of this size, consider splitting precursor refactors into their own PRs.

dianabarsan:
Addresses the 2026-04-27 review. Most of the blocking issues shared a root                                                                                                                                                                                                                                                                                    
  cause in the old rolling-doc + periodic-save design; rewrote that core                                                                                                                                                                                                                                                                                        
  instead of patching each symptom.                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                
  ## Architecture rewrite (covers blocking #2, #3, #5, #7)                                                                                                                                                                                                                                                                                                      
                                                                        
  Events are buffered in memory and persisted in batches to a per-day local                                                                                                                                                                                                                                                                                     
  PouchDB (`interaction-YYYY-M-D-{user}`). On each app `init()`, any per-day
  DB that isn't "today" is aggregated into a single `interaction-log` doc in                                                                                                                                                                                                                                                                                    
  the user's meta DB and the source DB is destroyed.                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                
  - No periodic `save()` timer → no cap-halving double-count.                                                                                                                                                                                                                                                                                                   
  - No shared rolling doc → no `_persist` race; writes are append-only  
    `bulkDocs` to per-day DBs, aggregation is serialized by an                                                                                                                                                                                                                                                                                                  
    `aggregationInFlight` promise guard.                                                                                                                                                                                                                                                                                                                        
  - Day-rollover handled in `record()` via `rolloverAndCheckCapacity`, which                                                                                                                                                                                                                                                                                    
    flushes the buffer and resets `persistedEventCount` — fixes the day N+1                                                                                                                                                                                                                                                                                     
    block.                                                                                                                                                                                                                                                                                                                                                      
  - `getServiceWorker()` is called once per stale per-day DB during init,                                                                                                                                                                                                                                                                                       
    not on every persist. Down from ~96/shift to 0–1.                                                                                                                                                                                                                                                                                                           
  - `this.user` is captured in `init()`; aggregate `_id` includes                                                                                                                                                                                                                                                                                               
    `(date, user, deviceId)`. No user-context drift between record and                                                                                                                                                                                                                                                                                          
    persist.                                                                                                                                                                                                                                                                                                                                                    
  - Aggregate is only written when `init()` finds a real `userCtx().name`,
    so `metadata.user` never falls back to `'unknown'`.                                                                                                                                                                                                                                                                                                         
                                                                        
  ## Direct fixes from the review                                                                                                                                                                                                                                                                                                                               
                                                                        
  - Registered `can_track_task_interactions` in                                                                                                                                                                                                                                                                                                                 
    `config/default/app_settings.json` (#6).                            
  - Added `interaction-` to `webapp/src/js/bootstrapper/purger.js` prefix
    filter so IDB cleans up on the webapp side too (#4).                                                                                                                                                                                                                                                                                                        
  - Removed the unreliable `beforeunload` flush; only `visibilitychange`                                                                                                                                                                                                                                                                                        
    triggers `persistBuffer()`.                                                                                                                                                                                                                                                                                                                                 
  - Chained `interactionTrackingService.init()` into `setupPromise` before                                                                                                                                                                                                                                                                                      
    `initRulesEngine`, so by the time `/tasks` is reachable, init has                                                                                                                                                                                                                                                                                           
    resolved.                                                                                                                                                                                                                                                                                                                                                   
  - e2e spec rewritten: `browser.pause` replaced with `waitUntil` helpers;                                                                                                                                                                                                                                                                                      
    `getInteractionMetaDocs` returns rows only, assertions live in the                                                                                                                                                                                                                                                                                          
    tests.                                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                            
  ## Tests                                                              

  - All four task component specs (tasks, tasks-content, tasks-group,                                                                                                                                                                                                                                                                                           
    sidebar-filter) pin the action keys passed to `record(...)`. A rename
    like `task_list:open` → `task_list:opened` now breaks the spec.                                                                                                                                                                                                                                                                                             
  - Karma spec for the service covers init / permission / user-context  
    gating, dedup, day rollover, 500-event cap, aggregation of stale                                                                                                                                                                                                                                                                                            
    per-day DBs, ownership filtering across users, and error tolerance. 
  - e2e spec uses a Date-only stub injected via `browser.addInitScript`,                                                                                                                                                                                                                                                                                        
    not `browser.emulate('clock')`. The latter pulls in `@sinonjs/fake-timers`,
    which (a) calls `require('util').promisify` while webpack's `process`                                                                                                                                                                                                                                                                                       
    polyfill makes the guard truthy → `ReferenceError`, and (b) overrides                                                                                                                                                                                                                                                                                       
    `setTimeout`/`setInterval` globally, which floods zone.js with                                                                                                                                                                                                                                                                                              
    "executing a cancelled action" from RxJS. `Date.now()` is monotonic                                                                                                                                                                                                                                                                                         
    via a per-call tick so events have unique, chronologically ordered                                                                                                                                                                                                                                                                                          
    timestamps.                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                
  ## Deferred 
                                                                                                                                                                                                                                                                                                                                                                
  - PII on `task.title` in `tasks-content` / `tasks-group` records. 
  Because multiple tasks can have the same form-as-action, the only way to disambiguate the actual task is by using the unique task title.

dianabarsan:
@sugat009 slight nudge here. we might go through some rounds before this is ready, and I really want to ship this is 5.2. appreciate it!

sugat009:
> @sugat009 slight nudge here. we might go through some rounds before this is ready, and I really want to ship this is 5.2. appreciate it!

Oh, I missed this one. Let me take a look.

webapp/src/ts/app.component.ts
```diff
@@ -312,6 +314,7 @@ export class AppComponent implements OnInit, AfterViewInit {
       .then(() => this.checkPrivacyPolicy())
       .then(() => (this.initialisationComplete = true))
       .then(() => this.initUser())
+      .then(() => this.interactionTrackingService.init())
```
sugat009, line 317: **issue (non-blocking):** `init()` rejection aborts app boot via the global catch at line 324, navigating to `/error/503`.

The spec at `webapp/tests/karma/ts/services/interaction-tracking.service.spec.ts:204` documents the contract: *"The error propagates to the caller (so app.component can decide what to do), but the service is left in a safe disabled state."* The service already lands safely on internal failure; the call site just needs to honor that:

```ts
.then(() => this.interactionTrackingService.init().catch(err => {
  console.warn('Interaction tracking disabled', err);
}))
```

sugat009 (commented):
The architecture rewrite cleanly addresses the 2026-04-27 blockers. CI green, all 47 checks pass.

I also ran this end to end on a local dev instance: rebuilt the bundle, logged in as a CHW user, drove the tasks tab, and inspected the per day IndexedDB and the meta DB. Service initializes, events buffer and persist, aggregate docs land in meta DB. Spot on.

Four inline comments below. None individually block the merge.

webapp/src/ts/modules/tasks/tasks-content.component.ts
sugat009: **issue (non-blocking):** `task.title` is template resolved by `rules-engine.service.ts:379`, so `{{contact.name}}` interpolation reaches this `record()` call. You documented exactly this risk at `tasks-sidebar-filter.component.ts:155-157` and stripped it from the filter path, but kept it here per your "Deferred" note in the response above.

Tested locally: an older aggregate in a CHW user's meta DB had `ref: "Home visit for {{contact.name}}"` recorded literally. In deployments with real contact data, that resolves to the patient name.

**suggestion:**
1. Open a tracking issue and add `// TODO(#xxxxx)` at both call sites (here and `tasks-group.component.ts:253`) so the deferral survives the squash merge.
2. Either expand the filter comment to acknowledge the asymmetry is intentional, or add a one liner here explaining that the title is used for instance disambiguation.

webapp/src/ts/services/interaction-tracking.service.ts
```diff
@@ -0,0 +1,463 @@
...
+      metadata: {
+        user: this.user,
+        deviceId: this.telemetryService.getUniqueDeviceId(),
+        date,
+        version,
+      },
+    };
+  }
+
+  private aggregateDocId(date: string): string {
```
sugat009, line 415: **question:** CHT has two existing patterns for date based meta DB doc IDs.

* **Telemetry** (`telemetry.service.ts:49-58`, `:85-97`): unpadded year/month/day in `_id`, plus numeric `year`/`month`/`day` on metadata. The dbt models on the CHT Sync Postgres replica filter on `doc#>>'{metadata,year}'` and `doc#>>'{metadata,month}'`.
* **Feedback** (`feedback.service.ts:197-201`): `feedback-${new Date().toISOString()}-${uuid}`, sortable on its own.

The aggregate `_id` here uses telemetry's unpadded shape, but the metadata at lines 388 to 394 only carries the string `date` — no numeric `year`/`month`/`day`. So this doc is queryable by neither path: `_id` range scans put October before February, and the dbt pattern that telemetry supports needs `metadata.month` to exist.

While testing locally I noticed the meta DB already holds aggregates from earlier iterations with `interaction-2026-04-24-…` (padded) alongside what current code produces (unpadded). Mixed formats already in the same namespace.

Was the telemetry style `_id` deliberate? If yes, would you mirror its numeric `metadata.year`/`month`/`day` so existing dbt models work the same? If keeping metadata minimal matters, would feedback's ISO 8601 `_id` be simpler?

webapp/src/ts/services/integration-api.service.ts
```diff
@@ -45,6 +48,7 @@ export class IntegrationApiService {
     this.Settings = settingsService;
     this.AndroidApi = androidApiService;
     this.Translate = translateService;
+    this.InteractionTracking = interactionTrackingService;
```
sugat009, line 51: **question:** `IntegrationApiService.InteractionTracking` (declared at line 28, assigned here) exposes the full service via `window.CHTCore`. Commit `907ea5299` shows it was a deliberate addition, but I don't see a current consumer in the repo or a mention in the PR description.

1. What's the intended consumer? Android app, Enketo widgets, partner extensions?
2. Was exposing the full service (`init`, `startSession`, `endSession`, `record`, `persistBuffer`) deliberate, or would just `record()` be enough? The wider surface means external callers can manage their own sessions, which is a contract to maintain; the narrower one keeps session lifecycle internal.

Asking now because narrowing this after a release is a breaking change for any consumer that built against it.

webapp/src/ts/modules/tasks/tasks-content.component.ts
dianabarsan: I addressed this in the comment: 

> PII on task.title in tasks-content / tasks-group records.
> Because multiple tasks can have the same form-as-action, the only way to disambiguate the > actual task is by using the unique task title.

webapp/src/ts/services/interaction-tracking.service.ts
dianabarsan: I think the unpadded version was a miss. thanks for flagging it.

webapp/src/ts/services/integration-api.service.ts
dianabarsan: I don't think we have consumers for the IntegrationApi, it's mostly used for debugging. 
I used this to debug (trigger an aggregate) on my phone.

webapp/src/ts/modules/tasks/tasks-content.component.ts
dianabarsan: Fixing with storing the title key separately, and logging the untranslated version.

dianabarsan:
Hi @sugat009 
I believe I have addressed all feedback. Please have another look.

webapp/src/ts/services/interaction-tracking.service.ts
```diff
@@ -0,0 +1,471 @@
...
+      return;
+    }
+    try {
+      db.close();
+    } catch (error) {
+      console.error('Error closing interaction DB', error);
+    }
+  }
+
+  private async getAppVersion(): Promise<string> {
```
sugat009, line 474: **issue (non-blocking):** `getAppVersion()` uses `versionService.getServiceWorker()`, which is HTTP-only. Offline aggregations land `version: 'unknown'` permanently.

`getServiceWorker()` at `version.service.ts:42-48` does `http.get('/deploy-info.json')` with no fallback. The catch at line 467 swallows the failure and returns `'unknown'`. Aggregation runs from `_init()` on every app boot and from `aggregateOnDayChange()` on every day rollover, both of which routinely run without network. A meaningful chunk of `interaction-log` docs in `users-meta` will carry `'unknown'` regardless of which version was actually running, and low-connectivity deployments get hit hardest.

Telemetry reads the local `_design/medic-client` ddoc instead (`telemetry.service.ts:72,78`). `versionService.getLocal()` at `version.service.ts:31-40` wraps the same path:

```ts
private async getAppVersion(): Promise<string> {
  try {
    const { version } = await this.versionService.getLocal();
    return version || 'unknown';
  } catch {
    return 'unknown';
  }
}
```

The karma test at `interaction-tracking.service.spec.ts:872` would flip to assert the real version is captured offline.

webapp/src/ts/services/interaction-tracking.service.ts
```diff
@@ -0,0 +1,471 @@
...
+    this.currentDayKey = currentDay.formatted;
+    this.persistedEventCount = await this.readPersistedCount(currentDay);
+    this.enabled = true;
+  }
+
+  /**
+   * Start a new interaction session. Subsequent record() calls are tagged with
+   * this session and a fresh sessionStartedAt.
+   */
+  startSession(session: string): void {
```
sugat009, line 134: **issue (non-blocking):** `startSession()` silently no-ops if called before `init()` resolves. Realistic on CHW cold starts to `/tasks`.

Walking `app.component.ts:312-328`: `initialisationComplete = true` at line 315 mounts the router-outlet and lets the route resolve; `interactionTrackingService.init()` runs at line 317. If the user lands on `/tasks` in that gap (on entry-level Android during a stale-DB aggregation, that's roughly 200-400ms), `TasksComponent.ngOnInit` calls `startSession('tasks')`, which early-returns at line 126 because `enabled` is still `false`. `currentSession` stays `null`, every `record()` hits `canRecord()` at line 174, and the whole visit is silent. The user has to navigate away and back for the next `ngOnInit` to call `startSession` again, by which point `init()` has long resolved.

Other services with the same shape use one of two patterns: a lazy `isInitialized()` guard like `cht-datasource.service.ts:31-36`, or component-side reactive gating like the `combineLatest` in `tasks-content.component.ts`. This service does neither.

Rough estimate is 10-20% of CHW cold starts on target hardware, mostly start-of-shift visits. Not a blocker, but the visits that drop are exactly the ones the feature would most want to capture.

Minimal fix is a pending-session marker:

```ts
private pendingSession: string | null = null;

startSession(session: string): void {
  if (!this.enabled) {
    this.pendingSession = session;
    return;
  }
  this.currentSession = session;
  this.sessionStartedAt = Date.now();
  this.lastEventKey = null;
}

// end of _init(), after `this.enabled = true`:
if (this.pendingSession) {
  const session = this.pendingSession;
  this.pendingSession = null;
  this.startSession(session);
}
```

No karma test exercises `startSession` before `init()` resolves; worth one that pins the deferred-activation behavior.

sugat009 (commented):
Iteration 3 closes the iteration-2 items. `init()` rejection is caught, `titleKey` carries the unresolved key through `rules-engine.service.ts:380` to the recording call sites, and date formats are zero-padded across the `_id` and the DB-name regex patterns. Component specs assert the titleKey vs translated-title distinction. Spec coverage for the architecture rewrite is excellent.

Reran locally: karma interaction-tracking 36/36, four task component specs 80/80, app.component visibilitychange 29/29, mocha purger and sentinel replications all green, lint clean on every changed file.

Two more things to flag, both follow-ups from iteration 1 that landed only halfway. Neither blocks.

webapp/src/ts/services/interaction-tracking.service.ts
dianabarsan: The service worker is always available offline.

webapp/src/ts/services/interaction-tracking.service.ts
dianabarsan: Fixed in https://github.com/medic/cht-core/pull/10786/commits/322801cb2fd65f19131461e9c3a315e5f7ab621f

sugat009 (approved):

PR #10789, opened 2026-03-31
feat(#8867): add can_update_contacts permission to guard contact edit button

### LINKED ISSUE #8867: Add permission to allow or block editing contacts in webapp
There is no way to block showing the edit button on contact detail pages, apart from blocking all writes by removing the `can_edit` permission. 

https://github.com/medic/cht-core/blob/91c9349205324af36106749fe6d60042dbceb8f0/webapp/src/ts/components/actionbar/actionbar.component.html#L247

For reports, we have a specific permission `can_update_reports` that gets checked along with `can_edit` in webapp actionbar. 

https://github.com/medic/cht-core/blob/91c9349205324af36106749fe6d60042dbceb8f0/webapp/src/ts/components/actionbar/actionbar.component.html#L149

Add a `can_edit_contacts` permission, to mirror `can_update_reports`. 

We could fine-tune this and somehow check individually for each contact, possibly by adding permissions to contact edit forms? 

**Additional context**
[Forum request](https://forum.communityhealthtoolkit.org/t/does-can-verify-reports-require-the-can-edit-permission-before-the-kebab-menu-becomes-visible/3330)

megha1807:
## Description

Adds a new `can_update_contacts` permission to control access to the contact edit button.

- Added `can_update_contacts` permission to `contacts-more-menu.component.ts` 
  alongside `can_edit` to guard the edit button
- Added `can_update_contacts` to `fast-action-button.service.ts` for the 
  contact edit and message actions
- Added `can_update_contacts` permission configuration to all app_settings 
  files (covid-19, default, demo configs)

Fixes #8867


megha1807:
The CI failure in `ci-integration-all-k3d` appears to be unrelated to my changes. The failing test `db-doc handler > restricts offline users > GET attachment` is an integration test for database document permissions which is not affected by the `can_update_contacts` permission changes in this PR. This may be a pre-existing flaky test.

@dianabarsan could you confirm if this CI failure is pre-existing and unrelated to my changes? Thank you!

webapp/src/ts/modules/contacts/contacts-more-menu.component.ts
```diff
@@ -88,7 +88,7 @@ export class ContactsMoreMenuComponent implements OnInit, OnDestroy {
   }
 
   private async checkPermissions() {
-    this.hasEditPermission = await this.authService.has('can_edit');
+    this.hasEditPermission = await this.authService.has(['can_edit', 'can_update_contacts']);
```
dianabarsan: I think that, now that we have this specific permission, we should stop checking for `can_edit`. Can you please remove the `can_edit` check from everywhere you also check for `can_edit_contacts`?

dianabarsan (changes requested):
very nice! i have one comment and would love the see an e2e test checking the option is unavailable without the permission here https://github.com/medic/cht-core/blob/master/tests/e2e/default/contacts/edit.wdio-spec.js

tests/e2e/default/contacts/edit.wdio-spec.js
```diff
@@ -62,6 +86,27 @@ describe('Edit ', () => {
     await commonPage.logout();
   });
 
+  it('should not show edit option when user lacks can_update_contacts permission', async () => {
+    // Ensure no leftover overlays from previous test
+    await browser.execute(() => {
+      const backdrops = document.querySelectorAll('.cdk-overlay-backdrop');
+      backdrops.forEach(el => el.remove());
+    });
```
dianabarsan: I'm not really liking this micro-management of modals (here and in the after-each). Which test and which modal can appear? I think it's better if we just close that one modal explicitly rather than having this super generic catch-all code bleeding everywhere.

tests/e2e/default/contacts/edit.wdio-spec.js
```diff
@@ -62,6 +86,27 @@ describe('Edit ', () => {
...
+
+    await loginPage.login(restrictedUser);
+    await commonPage.waitForPageLoaded();
+    await commonPage.goToPeople();
+    await contactPage.selectLHSRowByText(healthCenter.name);
+    await contactPage.waitForContactLoaded();
+
+    await commonPage.openMoreOptionsMenu();
+
+    const isEditVisible = await commonPage.isMenuOptionVisible('edit');
```
dianabarsan, line 93: I'm not sure how this test works, because the user logged in here has the `chw` role, which _has_ the `can_update_contacts` permission, same as the other users. To make the test work, either temporarily remove the permission from `chw` or create an entirely new role without the permission. 

Later edit: 
I know how the test works, because it tests editing the user's parent place, which is prohibited from being edited due to other mechanisms (in displayEditOption function).

tests/e2e/default/more-options-menu/online-user/permissions-disabled.wdio-spec.js
```diff
@@ -115,7 +115,7 @@ describe('More Options Menu - Online User - Permissions disabled', () => {
 
   describe('Edit permissions disabled', () => {
     before(async () => {
-      await utils.updatePermissions(onlineUser.roles, [], ['can_edit']);
+      await utils.updatePermissions(onlineUser.roles, [], ['can_edit', 'can_update_contacts']);
```
dianabarsan: Can you please separate the settings changes so that each individual test updates only the settings that they require? So that we pinpoint which permissions affects which workflow.

tests/e2e/default/more-options-menu/offline-user/edit-permission-disabled.wdio-spec.js
```diff
@@ -52,7 +52,7 @@ describe('More Options Menu - Offline User - Edit permissions disabled', () => {
     result = await utils.saveDoc(smsReport);
     smsReportId = result.id;
     await utils.createUsers([offlineUser]);
-    await utils.updatePermissions(offlineUser.roles, [], ['can_edit'], { ignoreReload: true });
+    await utils.updatePermissions(offlineUser.roles, [], ['can_edit', 'can_update_contacts'], { ignoreReload: true });
```
dianabarsan: Can you please separate the settings changes so that each individual test updates only the settings that they require? So that we pinpoint which permissions affects which workflow.

webapp/tests/karma/ts/modules/contacts/contacts-more-menu.component.spec.ts
```diff
@@ -56,16 +56,18 @@ describe('Contacts More Menu Component', () => {
         fixture = TestBed.createComponent(ContactsMoreMenuComponent);
         component = fixture.componentInstance;
         store = TestBed.inject(MockStore);
-        fixture.detectChanges();
+        // NOTE: No fixture.detectChanges() here — each test controls when ngOnInit fires
```
dianabarsan: Please remove this comment.

dianabarsan (changes requested):

dianabarsan (changes requested):
Thanks for all the changes @megha1807 .

I tested it out locally and it seems that removing the update permission hides the three-dotted-menu completely, even if the delete permission is still present. 

Besides fixing this I will ask you to update the e2e test to remove the `can_update_contacts` from the `chw` role, instead of creating a new role with just a few permissions. Additionally the assertions should cover that the three dotted menu is still there and deletions are still possible. 

https://github.com/user-attachments/assets/b72e2a78-b653-4352-b540-21963db4e69b

PR #10795, opened 2026-04-01
fix(#10341): revoke permissions when user role is deleted from app_settings

### LINKED ISSUE #10341: Deleting roles does not result in users with those roles losing permission
**Describe the bug**
This a scenario where a user's permissions are revoked, but those permissions are not reflected in the app.

**To Reproduce**
1. Create user `test_user` with role `test_role` with permissions to view the contacts tab
2. As admin, got to admin console and delete the role via `Roles & permissions` > `Roles` > Trash can icon
3. Sync as `test_user`
4. The user is prompted to reload

**Expected behavior**
After reload, the user should not be able to view the contacts tab as they have no role and no permissions.

**Logs**

**Screenshots**

**Environment**
- Instance: training-3.echis.go.ke
- App: webapp
- Version: `4.21.1`

**Additional context**

vikrantwiz02:
# Description

When a role is deleted via Admin > Roles & Permissions, it is removed from `app_settings.roles` but the references in `app_settings.permissions` remain. All three permission-checking codepaths only compared the user's assigned roles against the `permissions` object — they never verified whether those roles still existed in `app_settings.roles`. This meant users retained permissions from deleted roles indefinitely.

Fixes this by filtering a user's effective roles to only those still present in `app_settings.roles` before evaluating any permission. An empty or absent `roles` config falls back to no filtering (backwards compatible).

Three codepaths fixed:
- `shared-libs/cht-datasource/src/auth.js` — used by the webapp frontend (contacts tab, sidebar, rules engine). Added `filterRolesByConfigured()` and an optional `chtRolesSettings` param to `hasPermissions()` and `hasAnyPermission()`.
- `shared-libs/user-management/src/roles.js` — used by user management logic (`can_skip_password_change`, `can_have_multiple_places`). `hasPermission()` now reads `config.get('roles')` and filters before checking `permissions`.
- `api/src/auth.js` — used by the API server (export, login, users controllers). 
Same filtering applied to the local `hasPermission()`.

`webapp/src/ts/services/cht-datasource.service.ts` updated to extract and forward `roles` from app settings alongside `permissions`.

Fixes #10341

# Code review checklist

# License


mrjones-plip:
assigning over to @jkuester as he did some [research back when](https://github.com/medic/cht-core/issues/10341#issuecomment-3357850522)!

shared-libs/cht-datasource/src/auth.js
```diff
@@ -122,9 +131,10 @@ const hasPermissions = (permissions, userRoles, chtPermissionsSettings) => {
  * @param permissionsGroupList {string[][]} Array of groups of permissions due to the complexity of permission grouping
  * @param userRoles {string[]} Array of user roles.
  * @param chtPermissionsSettings {object} Object of configured permissions in CHT-Core's settings.
+ * @param chtRolesSettings {object} Object of configured roles in CHT-Core's settings. Used to filter out deleted roles.
  * @return {boolean}
  */
-const hasAnyPermission = (permissionsGroupList, userRoles, chtPermissionsSettings) => {
+const hasAnyPermission = (permissionsGroupList, userRoles, chtPermissionsSettings, chtRolesSettings) => {
```
jkuester: Instead of adding the `chtRolesSettings` parameter here, can we actually just replace the `chtPermissionsSettings` parameter with just `settings`?  Then, we can move the `getChtPermissionsFromSettings` and `getChtRolesFromSettings` logic from the cht-datasource.service down here and just unpack all the necessary info from the settings.  

The same goes for the `hasPermissions` function above.

shared-libs/user-management/src/roles.js
jkuester: Can you move this to `shared-libs/constants` so we can share it with cht-datasource?

shared-libs/cht-datasource/src/auth.js
jkuester: Lets remove this variable and update the code to use the `DB_ADMIN_ROLES` once you move that to `shared-libs/contants`.

shared-libs/cht-datasource/src/auth.js
```diff
@@ -46,6 +46,13 @@ const checkUserHasPermissions = (permissions, userRoles, chtPermissionsSettings,
   });
 };
 
+const filterRolesByConfigured = (userRoles, chtRolesSettings) => {
+  if (!chtRolesSettings || !Object.keys(chtRolesSettings).length) {
+    return userRoles;
+  }
+  return userRoles.filter(role => role === ADMIN_ROLE || chtRolesSettings[role]);
```
jkuester: I think our logic here needs to be restrictive if no roles are configured. Instead of returning all the suerRoles in that case, we should return none of them.

```suggestion
  const availableRoles = new Set([ADMIN_ROLE, ...Object.keys(chtRolesSettings ?? {})]);
  return userRoles.filter(role => availableRoles.has(role));
```

shared-libs/user-management/src/roles.js
jkuester: I _think_ we can maybe just remove this function altogether and then update the `users.js` to just call the cht-datasource function.  

In `users.js` we would need to import `getDatasource` from `@medic/cht-datasource` and then replace the calls to `roles.hasAllPermissions` with:

```js
getDatasource(dataContext).v1.hasPermissions(...)
```

api/src/auth.js
```diff
@@ -25,11 +25,15 @@ const get = (path, headers) => {
 };
 
 const hasPermission = (userCtx, permission) => {
```
jkuester: Same here. I think we can just remove this function and call cht-datasource instead.

jkuester (changes requested):
Sorry for the delayed response @vikrantwiz02!  The changes are great, though it is concerning that we had to basically make the same change in three different places... 😓 

I think this is the right time to correct that and try to centralize our roles logic in cht-datasource.  🙏

vikrantwiz02:
Hi @jkuester, thanks for the detailed review! I've addressed all your comments:

- hasPermissions/hasAnyPermission signature — replaced the separate chtPermissionsSettings/chtRolesSettings params with a single settings object. The unpacking of settings.permissions and settings.roles now happens inside auth.js.
- DB_ADMIN_ROLES in constants — moved to shared-libs/constants so it's shared across cht-datasource, user-management, and api.
- ADMIN_ROLE variable — removed; now uses DB_ADMIN_ROLES from constants directly.
- roles.hasAllPermissions in user-management — removed the function; callers now use getDatasource(dataContext).v1.hasPermissions(...) from cht-datasource.
- hasPermission in api/src/auth.js — removed; now delegates to cht-datasource.
- Filter logic — implemented using new Set([...DB_ADMIN_ROLES, ...Object.keys(chtRolesSettings ?? {})]) with a backward-compatibility guard: when roles config is absent (old app_settings), roles are not filtered so existing deployments keep working.

The two failing CI checks (ci-webdriver-default-data, ci-webdriver-default-workflows), I re-ran many times, but still these two failed.

Could you take another look? Happy to adjust anything.

api/src/auth.js
```diff
@@ -25,17 +25,16 @@ const get = (path, headers) => {
   });
 };
 
-const getSettings = () => ({ permissions: config.get('permissions'), roles: config.get('roles') });
-
 const assertPermissions = async (req, { isOnline = false, hasAll = [], hasAny = [] }) => {
+  const settings = config.getAll();
```
jkuester, line 29: Sorry, I forgot to mentioned that the `getAll()` function works for getting the whole settings object (so we don't have to worry about combining sub-properties).

shared-libs/cht-datasource/src/auth.js
```diff
@@ -46,10 +46,7 @@ const checkUserHasPermissions = (permissions, userRoles, chtPermissionsSettings,
   });
 };
 
-const filterRolesByConfigured = (userRoles, chtRolesSettings) => {
-  if (chtRolesSettings === undefined || chtRolesSettings === null) {
-    return userRoles;
-  }
```
jkuester: I don't think we want to be "backwards compatible".  Basically if we are doing a permissions check, the roles and permissions settings _must_ be provided.

jkuester (approved):
Very nice! Thank you!  I had a couple of tweaks, but nothing major (except for the unit test changes that were required... 😅 ).  As long as the CI still builds, this should be good to go!  🤞

api/src/auth.js
vikrantwiz02: Got it, config.getAll() is definitely cleaner and much more concise than combining sub-properties manually. Thanks for the clarification.

shared-libs/cht-datasource/src/auth.js
vikrantwiz02: Ya, If the settings aren't provided correctly, the call is fundamentally broken rather than something we should silently tolerate.

vikrantwiz02:
@jkuester, All unit tests (cht-datasource, user-management, and api) are passing clean on my end.

"Compile the app" CI, which is failing, I triggered re-runs in my github actions 3 times but still failing at same point. I am investigating the root cause. Will update you.

vikrantwiz02:
I investigated the 'Compile the app' failure that triggered after the recent cleanup.

It turns out the new strict logic in filterRolesByConfigured was causing 11 failures in the unit-webapp suite. Because the mock settings in the Karma specs didn't have a roles key, the service was stripping away the user's roles (like chw_supervisor or national_admin) before checking permissions.

I’ve updated auth.service.spec.ts and cht-datasource.service.spec.ts to include those roles in the mock settings.

@jkuester, Finally all CI checks are green at my end 😅

jkuester (approved):

jkuester (changes requested):
Well bad news... 🤦 Was digging into the failing e2e tests and realized it was connected to our changes here. 

That is when I realized that the `permissions` functions in cht-datasource are some that _must_ remain passive.  😓 I knew these functions are used in custom tasks/targets/contact-summary config code, but I thought we were okay because they were already being shimmed at the webapp layer in the `CHTDatasourceService`.  _However_, the test was failing because this functions can also be used in _purging config code_ which gets called in Sentinel.  😬   So, TLDR, we can only make passive changes to these functions (or else add new versions of them in a `v2` block).  🤔 

I am still trying to figure out the best approach here.  On one hand we can take the new versions of the functions and just add them to a `v2` block and call it a day. 👍  But, a more interesting alternative might be to enhance the "remote" data context used in cht-datasource to have access to a "settings service" (like the "local" context) where we can synchronously get the latest settings doc.  (Effectively this functionality already exists in the places where the remote context is used, we just are not injecting it into the data context.) I think this would let us possibly even avoid adding `v2` functions (since we can just make the current permissions settings param optional). 

I want to do a bit more studying on this before locking us into a particular approach (ideas/feedback welcome!), but I might not be able to circle back to this until next week.

vikrantwiz02:
That's a great catch, I didn't have the Sentinel purging context on my radar. 😓

The v2 block approach sounds like the safest path to me since it keeps the existing functions untouched and lets us control when callers migrate. Happy to take either direction you decide on. Let me know how I can help once you've had time to think it through.

Thanks.

jkuester:
Okay, @vikrantwiz02 I did some more cooking on this and ended up with https://github.com/vikrantwiz02/cht-core/pull/1

As long as @sugat009 is cool with those changes, we can get that PR merged into your branch and the changes will show up here.

vikrantwiz02:
@jkuester, thanks for the refactor. Using SettingsService in the remote context is a cleaner way to keep the permissions API passive for Sentinel.

I will wait for @sugat009 response.

jkuester:
@vikrantwiz02 could you assign @sugat009 as a reviewer over on https://github.com/vikrantwiz02/cht-core/pull/1?

vikrantwiz02:
@sugat009, I've sent you a collaborator invite to my fork so I can assign you as a reviewer for @jkuester's refactor as requested. Please accept whenever you have a moment!

sugat009:
@vikrantwiz02 I've accepted the invite. I'll take a look at the PR tomorrow.

vikrantwiz02:
Ok @sugat009, I already assigned you as reviewer.

vikrantwiz02:
@jkuester, merged your changes and I also did a tiny follow-up commit to resolve two SonarCloud failures (nesting depth in search.js and a Boolean identity check in edit-user.js) just to get the Quality Gate back to green.

jkuester (approved):
Thanks for sticking with us on this one @vikrantwiz02!  LGTM!

PR #10800, opened 2026-04-01
feat(#8645): display app-version from service worker deploy-info.json in Webapp About page

### LINKED ISSUE #8645: Add versioning to service-worker cache
The version of the code running on a device is determined by: 
a) the version of the JS code composing the app
b) the version of the ddoc providing database view queries

a) does not exist. At this time, there is no way to know, whether as information in the app or in telemetry, which version of the client code is running on the device. 

Add versioning to webapp code or service worker cache in some way. Display this information in the about page and include it in telemetry. 

**Additional context**
When upgrading the client app, there will be a period of desync when the device will be running the new ddoc views on top of the old webapp code. Users can ignore the reload modal or delay reload by significant time.

megha1807:
Closes #8645

## Changes
- Reads `app-version` from `deploy-info.json` via `VersionService.getServiceWorker()`
- Displays it on the About page below the existing version field
- Includes `serviceWorker` version in telemetry metadata
- Added translation key `app.version`
- Added unit tests for success and failure cases

## Testing
1. Load the app with service worker active
2. Navigate to About page
3. Confirm "App version" row appears with correct value


megha1807:
Hi @dianabarsan, I wanted to flag two CI issues on this PR:
1. SonarCloud warning - about.component.html line 28:
SonarCloud is flagging <label translate>app.version</label> for not being associated with a form control. However, this is the exact same pattern used consistently throughout the entire About page for all other display labels such as version, ddoc.version, clientDdoc.version, url, User Name etc. - none of which are flagged since they are pre-existing lines. This <label> is being used for read-only display information, not a form input, which is consistent with the rest of the file. I'm happy to change it to a <span> or <p> if the team prefers a different element for display-only text - just wanted your guidance before changing the existing pattern used across the whole file.
2. Compile the app build failure:
The build is failing due to two pre-existing issues unrelated to this PR's changes:

reports.component.spec.ts:174 - TypeError: Cannot read properties of undefined (reading 'resetSelectors') - this is a flaky test in a file I never touched.
Code coverage thresholds (statements 84.07%, branches 75.8%, functions 83.16%, lines 84.12%) are all below the configured thresholds - this was already the case on master before this branch was created.

Both of these failures exist independently of my changes. Please let me know if you'd like me to do anything differently. Happy to make any adjustments needed!

dianabarsan:
@megha1807 I am quite certain the CI failure is due to your changes, and likely because of the new services you are injecting in telemetry.

dianabarsan (changes requested):
Per above.

megha1807:
Hi @dianabarsan, all CI checks are now passing. Here's a summary of the fixes made:

1.  Test isolation fix - Added this.appVersion = undefined; at the start of the service worker fetch in getVersionAndRevisions()      in about.component.ts. This ensures stale data is cleared on re-init, which also fixes a real behavioral bug where a failed second ngOnInit() call would leave outdated version info displayed on screen.
2.  Translation keys -  Added the missing app.version key to messages-es.properties, messages-fr.properties, messages- ne.properties, and messages-sw.properties.

Please let me know if there's anything else you'd like me to address. Happy to make any further adjustments!

webapp/src/ts/modules/about/about.component.html
```diff
@@ -24,6 +24,11 @@
             <p test-id="about-version">{{version}}</p>
           </div>
 
+          <div *ngIf="appVersion">
```
dianabarsan: I don't think we should be adding an extra field, I think we should be using the already existing field, `about-version`. 
They _should_ always be the same, but there can be exceptions when the app is only partially upgraded (this can cause issues and this is the reason for this work). 

so I think we should: 
- keep reading both values in the component
- compare the values, if they are equal then we just display one field
- if they are _not_ equal then, we would display two fields and add some visual warning. we should also log an error so a feedback doc gets created.

dianabarsan (changes requested):
Nice work. I left request inline.

megha1807:
Hi @dianabarsan, I've updated the implementation as per your suggestion:

- Both values are still read in the component
- After both promises resolve, the values are compared using Promise.all
- If equal, only the existing about-version field is shown (no change to current UI)
- If different, a second field appears below showing the service worker version with a visual warning, and console.error is  called so a feedback doc gets created

Please let me know if this looks right or if you'd like any adjustments!

webapp/src/ts/modules/about/about.component.ts
```diff
@@ -128,6 +131,27 @@ export class AboutComponent implements OnInit, OnDestroy {
...
+      .getServiceWorker()
+      .then(({ version }) => {
+        this.appVersion = version;
+        return version;
+      })
+      .catch(error => {
+        console.debug('Could not access service worker app version', error);
+      });
+
+    Promise.all([localVersionPromise, serviceWorkerPromise])
```
dianabarsan: I don't really like saving the promises in values. 
Could you, please, refactor and save the reads in new functions, like: 
```
private async getServiceWorkerVersion () {
  try {
    const { version } = await this.versionService.getServiceWorker();
    this.serviceWorkerVersion = version;
  } catch (err) { ... }
}
private async getDeployVersion () {
  try {
    const { version, rev } = await this.versionService.getLocal();
    this.version = version;
    this.localRev = rev;
  } catch (err) { ... }
}

async private getVersionAndRevisions() {
  await Promise.all([
    this.getServiceWorkerVersion(),
    this.getDeployVersion();
  ]);

  if (localVersion && serviceWorkerVersion && localVersion !== serviceWorkerVersion) {
    this.versionMismatch = true;
    console.error(`Version mismatch: deploy-info version (${this.version}) and service worker version (${this.serviceWorkerVersion}) are different.`);
  }
}

```

webapp/tests/karma/ts/modules/about/about.component.spec.ts
```diff
@@ -199,6 +202,42 @@ describe('About Component', () => {
...
+  it('should set appVersion from service worker deploy info', fakeAsync(() => {
+    versionService.getLocal.resolves({ version: '3.5.0', rev: '12' });
+    versionService.getRemoteRev.resolves('15');
+    versionService.getServiceWorker.resolves({ version: '4.5.0' });
+
+    component.ngOnInit();
+    flush();
+    discardPeriodicTasks();
+
+    expect(component.appVersion).to.equal('4.5.0');
```
dianabarsan, line 214: can you also assert on the other version?

webapp/tests/karma/ts/services/telemetry.service.spec.ts
```diff
@@ -336,44 +365,41 @@ describe('TelemetryService', () => {
 
       await service.record('test', 10);
 
-      expect(telemetryDb.post.calledOnce).to.be.true;   // Telemetry entry has been recorded
+      expect(telemetryDb.post.calledOnce).to.be.true;
```
dianabarsan: Can you please revert deleting all these comments?

dianabarsan (changes requested):
Nice work! Thanks for making the changes. I left two small requests inline.

webapp/tests/karma/ts/modules/about/about.component.spec.ts
```diff
@@ -199,6 +202,43 @@ describe('About Component', () => {
...
+  it('should not set appVersion when service worker returns no version field', fakeAsync(() => {
+    versionService.getLocal.resolves({ version: '3.5.0', rev: '12' });
+    versionService.getRemoteRev.resolves('15');
+    versionService.getServiceWorker.resolves({});
+
+    component.ngOnInit();
+    flush();
+    discardPeriodicTasks();
+
+    expect(component.appVersion).to.be.undefined;
```
dianabarsan, line 239: Can you please add a test where you demonstrate how the version mismatch is handled? 
Ideally we would have one test where the versions are the same and one test where the versions are different.

megha1807:
Hi @dianabarsan, I've addressed all three of your feedback points:

1. **Refactored `getVersionAndRevisions()`** - Extracted into separate private async methods `getServiceWorkerVersion()` and `getDeployVersion()`, with `Promise.all` used in `getVersionAndRevisions()` as suggested.

2. **Added assertion on the other version** - The test now asserts on both `component.version` and `component.appVersion`.

3. **Restored deleted comments in `telemetry.service.spec.ts`** - The inline comments have been reverted.

Please let me know if there's anything else you'd like me to adjust!

webapp/src/ts/modules/about/about.component.html
```diff
@@ -24,6 +24,12 @@
             <p test-id="about-version">{{version}}</p>
           </div>
 
+          <div *ngIf="versionMismatch">
```
dianabarsan, line 27: I think the styling is a bit off. You should use `label` here instead of spam, and the warning should have a style as well. We already have a similar "warning" type on this page for the unsupported browser.

dianabarsan: This is how the unsupported browser warning is displayed: 

Image

dianabarsan (changes requested):
Very nice! I added some requests below.

Additionally, can you please add an additional assertion in this test: https://github.com/medic/cht-core/blob/da5fe7f731303c6f69315b3992455125f539f91a/tests/e2e/default/about/about.wdio-spec.js#L37

Where we validate that only one version is displayed and no warning?

Thanks!

megha1807:
Hi @dianabarsan, I wanted to flag the SonarCloud warning on about.component.html line 28.

SonarCloud is flagging the <label> tag for not being associated with a form control. However, this is display-only text (same as all other labels on this page like version, url, User Name etc.) and you specifically requested using <label> here.

I've added  to suppress this warning since the pattern is intentional and consistent with the rest of the file. Please let me know if you'd prefer a different approach!

PR #10801, opened 2026-04-01
feat(#10650): support configuring max_age parameter for OIDC login

### LINKED ISSUE #10650: Support configuring `max_age` parameter for OIDC login
Originally discussed [on the forum](https://forum.communityhealthtoolkit.org/t/oidc-sso-logout-on-shared-android-devices-idp-session-not-cleared/5513/2).  Currently the CHT [SSO functionality](https://docs.communityhealthtoolkit.org/hosting/sso/) supports basic OIDC authentication whereby the user actually authenticates via a 3rd-party auth provider. The OIDC flow results in the user's client being assigned a valid CouchDB session cookie (aka a valid CHT session cookie).  

When the user logs out of the app, the session cookie is cleared and they no longer have a valid CHT session.  _However_, depending details around the user's device and the 3rd-party auth provider, the user may still have an active session with the 3rd-party auth provider.  (E.g. in cht-android, we use an [Android Custom Tab](https://developer.android.com/develop/ui/views/layout/webapps/overview-of-android-custom-tabs) to complete the OIDC auth flow which will have access to sessions on the user's default browser.)  This means that if the user again tries to login to the CHT via the SSO flow, they will be redirected to the 3rd-party auth provider. If they already have an active session there, the auth provider may just automatically redirect back to the CHT with the ID token for the user with the current session. This will cause the user to be logged back into the CHT with the same CHT user as before.

In many cases, this will be considered a feature (the CHT SSO functionality was designed/intended to work this way).  Take the case where a user is logged into their Google account on their phone and Google is configured as the 3rd-party auth provider for their CHT instance.  The first time this user opens their CHT app and selects the SSO Login, they can be automatically redirected to/from the Google auth page (without needed to re-authenticate) and the CHT will automatically log them in as the CHT user associated with that Google account.

_However_ (as described in the forum thread) this functionality is problematic in some cases (such as where a single device is shared by multiple users).  When trying to switch from one SSO user to another, the 3rd-party auth provider may not prompt for the second user to login and instead just return the id token for the first user (who's session with the auth provider is still valid).  The second user may not be prompted to login until the existing session with the auth provider is cleared.

We should support [passing the `max_age` parameter](https://auth0.com/docs/authenticate/login/max-age-reauthentication) when the CHT api server starts the OIDC login flow. Then we can validate the `auth_time` claim that is returned with the id token and ensure it is within the desired time window (e.g. `60s`).

From the [OIDC Spec](https://openid.net/specs/openid-connect-core-1_0.html):

The `max_age` parameter is:

> OPTIONAL. Maximum Authentication Age. Specifies the allowable elapsed time in seconds since the last time the End-User was actively authenticated by the OP. If the elapsed time is greater than this value, the OP MUST attempt to actively re-authenticate the End-User. .... When max_age is used, the ID Token returned MUST include an auth_time Claim Value. Note that max_age=0 is equivalent to prompt=login.

The `auth_time` claim is:

> Time when the End-User authentication occurred. Its value is a JSON number representing the number of seconds from 1970-01-01T00:00:00Z as measured in UTC until the date/time. When a max_age request is made or when auth_time is requested as an Essential Claim, then this Claim is REQUIRED; otherwise, its inclusion is OPTIONAL.

When validating the id token:

> If the auth_time Claim was requested, either through a specific request for this Claim or by using the max_age parameter, the Client SHOULD check the auth_time Claim value and request re-authentication if it determines too much time has elapsed since the last End-User authentication. 

Similar re-authentication functionality could be achived via the `prompt=login` parameter, but [the auth0 docs](https://auth0.com/docs/authenticate/login/max-age-reauthentication) note:

> The prompt=login mechanism can be subverted by simply stripping the parameter as it passes through the user agent (browser) and is only good for providing a UX hint to the provider (OP) in cases when the (RP) wants to display a link like:

**Additional context**

The changes to the CHT code _should_ be pretty simple.  I think we  could make the `max_age` parameter value configurable in the `oidc_provider` properties in the CHT instance settings doc. If a value is provided for that settings property, then our api code should provide it [when calling `client.buildAuthorizationUrl`](https://github.com/medic/cht-core/blob/2cbe9c10991de77e5ed7408432474800ea5185d4/api/src/services/sso-login.js#L97).  It looks like we can pass it directly when making the  `client.buildAuthorizationUrl` call like [in this example](https://github.com/panva/openid-client/discussions/828), or maybe we could just set the [default_max_age](https://github.com/panva/openid-client/blob/b77d87c1e2f5fef6fab501de615fb83a74a0251f/docs/interfaces/ClientMetadata.md#default_max_age) somehow?  

Then, when [calling `client.authorizationCodeGrant`](https://github.com/medic/cht-core/blob/2cbe9c10991de77e5ed7408432474800ea5185d4/api/src/services/sso-login.js#L119) we should pass the [`maxAge` property](https://github.com/panva/openid-client/blob/b77d87c1e2f5fef6fab501de615fb83a74a0251f/docs/interfaces/AuthorizationCodeGrantChecks.md#maxage) (maybe to 60 seconds) so that the `auth_time` claim is properly validated.

I think we can update our [SSO Login integration tests](https://github.com/medic/cht-core/blob/58697bba201371326186e6d15ce39699688993f5/tests/integration/api/controllers/login.spec.js#L316) to include a test for this functionality.

saurabh12nxf:
# Description

This PR implements support for the OIDC `max_age` parameter to address issues with shared device scenarios where users may be automatically logged in as the previous user without being prompted for authentication.

## Changes Made

- Modified `api/src/services/sso-login.js`:
  - Updated `oidcServerSConfig()` to return both `idServerConfig` and `oidcProvider` settings
  - Added `max_age` parameter to authorization URL when configured in `oidc_provider` settings
  - Added `maxAge` validation to token verification to check the `auth_time` claim
  
- Added comprehensive test coverage in `api/tests/mocha/services/sso-login.spec.js`:
  - Test for `max_age` parameter in authorization URL (normal case)
  - Test for `max_age: 0` edge case in authorization URL
  - Test for `maxAge` validation in token verification (normal case)
  - Test for `maxAge: 0` edge case in token verification

## How It Works

When `max_age` is configured in the `oidc_provider` settings, the OIDC authorization flow will:
1. Include the `max_age` parameter in the authorization URL sent to the auth provider
2. The auth provider will check when the user last authenticated
3. If the time elapsed exceeds `max_age` seconds, the user will be prompted to re-authenticate
4. The returned ID token includes an `auth_time` claim which is validated against the configured `maxAge`

## Configuration

Add `max_age` (in seconds) to the `oidc_provider` configuration in the instance settings document:

```json
{
  "oidc_provider": {
    "discovery_url": "https://auth-provider.com/.well-known/openid-configuration",
    "client_id": "your-client-id",
    "max_age": 60
  }
}
```

Setting max_age: 0 is equivalent to prompt=login per the OIDC specification, forcing immediate re-authentication.

Fixes #10650

# Code review checklist

- [x] AI disclosure: AI assistance (Claude/Kiro) was used for debugging and refining the implementation.

# License


api/src/services/sso-login.js
```diff
@@ -93,8 +93,12 @@ const getAuthorizationUrl = async (redirectUrl) => {
     scope: 'openid email'
   };
 
-  const serverConfig = await oidcServerSConfig();
-  return client.buildAuthorizationUrl(serverConfig, params);
+  const { idServerConfig, oidcProvider } = await oidcServerSConfig();
+  if (oidcProvider.max_age !== undefined && oidcProvider.max_age !== null) {
+    params.max_age = oidcProvider.max_age;
+  }
+
```
jkuester, line 100: Minor, but I guess we can just directly pass the `max_age` value here.

```suggestion
  const { idServerConfig, oidcProvider } = await oidcServerSConfig();
  const params = {
    redirect_uri: redirectUrl,
    scope: 'openid email',
    max_age: oidcProvider.max_age
  };
```

api/src/services/sso-login.js
```diff
@@ -115,11 +119,15 @@ const getLocale = async (localeClaim) => {
  * @returns {object} Token id and user details.
  */
 const getIdToken = async (currentUrl) => {
-  const serverConfig = await oidcServerSConfig();
+  const { idServerConfig, oidcProvider } = await oidcServerSConfig();
+  const checks = { idTokenExpected: true };
+  if (oidcProvider.max_age !== undefined && oidcProvider.max_age !== null) {
+    checks.maxAge = oidcProvider.max_age;
+  }
```
jkuester, line 127:
```suggestion
  const checks = { 
    idTokenExpected: true,
    maxAge: oidcProvider.max_age
  };
```

Should be able to just set this directly.  

Looking at the [oauth4webapi validation logic](https://github.com/panva/oauth4webapi/blob/8afe6edeb9446ca4f414a993f76f2116d2cb6502/src/index.ts#L4102) (used under the hood by `openid-client`) when the `auth_time` is validated, it is with a default tolerance of `30s` which _should_ be fine for our purposes.  

We do need to bump the version of the `openid-client` dependency in the root package.json to the latest release (`6.8.2`) so that we get [this fix](https://github.com/panva/oauth4webapi/commit/15c0c50f41c8bc2be3b762cde1c7b4df1b9a7d9f) for `max_age: 0`... 😅

jkuester (changes requested):
Looking good @saurabh12nxf!  

Can you also update the [integration tests](https://github.com/medic/cht-core/blob/58697bba201371326186e6d15ce39699688993f5/tests/integration/api/controllers/login.spec.js#L316) to test this logic.  Basically we want to confirm that when `max_age` is configured, it is passed as a param.  Then, when the server returns the `auth_time`, we should confirm (if possible) that the value of the claim is validated (and very old auth times are rejected).  We don't need to get too crazy checking every edge case, but it would be nice to hit the happy path for `max_age` and then have one test that does fail the `auth_time` validation.  Probably will require changes to our [mock oidc provider](https://github.com/medic/cht-core/blob/58697bba201371326186e6d15ce39699688993f5/tests/utils/mock-oidc-provider.js).

saurabh12nxf:
Thanks for the review @jkuester! I've made the requested changes:

- Simplified the max_age handling by removing unnecessary undefined/null checks
- Updated openid-client from 6.4.2 to 6.8.2 (includes fix for max_age: 0)
- Updated unit tests accordingly
- All unit tests passing (26/26)

For the integration tests, I'd appreciate some guidance on how to set up the mock OIDC provider to test the max_age flow and auth_time validation. Should I look at extending the existing mock provider in `tests/integration/api/controllers/login.spec.js`?

jkuester:
> For the integration tests, I'd appreciate some guidance on how to set up the mock OIDC provider to test the max_age flow and auth_time validation. Should I look at extending the existing mock provider in tests/integration/api/controllers/login.spec.js?

The `login.spec.js` integration tests [run the mock OIDC Provider](https://github.com/medic/cht-core/blob/58697bba201371326186e6d15ce39699688993f5/tests/integration/api/controllers/login.spec.js#L369).

[That mock provider](https://github.com/medic/cht-core/blob/58697bba201371326186e6d15ce39699688993f5/tests/utils/mock-oidc-provider.js) is effectively just an Express server that "handles" OIDC requests as if it was a real OIDC provider.  We will need to make some kind of updates to this service (I have not had a chance to look too closely yet).  Basically it I think we need some logic to have it return the `auth_time` claim when the `max_age` parameter is submitted and then we will need some way for the tests to customize the value the server returns for the `auth_time` claim... (I think...)

saurabh12nxf:
Hi @jkuester  please once take a look and if any changes are needed tell me I am happy to fix it.

jkuester (approved):
Looking good!  I tweaked the integration tests just a bit to use a more direct method of setting the auth_time value, but otherwise the tests were great too!


I have opened a Docs PR (https://github.com/medic/cht-docs/pull/2197) to document the new config setting.

@raphaelkenyuri, when you get a change can you try deploying the `fix-10650-oidc-max-age` version (should be available on the Upgrades page) to a test instance and see if it works as expected with your SSO provider?  (See the Docs PR for details on how to set the config).  Based on the behavior with our mock OIDC provider in the integration tests, all of the CHT side logic is working like I expect (including the validation of the auth_time claim), but we should confirm that it actually works with a real OIDC provider. I will wait to merge this until then. 👍

raphaelkenyuri:
@jkuester @saurabh12nxf I tested the fix-10650-oidc-max-age version (~5.1.0) with max_age configured against our Microsoft Entra provider. 
It works correctly on web, I'm prompted to re-authenticate as expected.
On Android, login fails. The authorization step returns a code, but the token exchange fails with a ResponseBodyError (HTTP 400 Bad Request) from the Microsoft token endpoint, and the app redirects to /medic/login?sso_error=loginerror. Error below:
`name: 'ResponseBodyError'
statusText: 'Bad Request'
url: 'https://login.microsoftonline.com/9e8a5334-497c-4d8a-a797-7997cf8cc763/oauth2/v2.0/token'
ok: false`

Microsoft returns a small JSON error body (content-length: 472) but it's consumed by the library before it can be logged, so I don't yet have the AADSTS code.

raphaelkenyuri:
@jkuester @saurabh12nxf I have tested by removing max_age entirely from the config. With max_age removed, SSO login on Android works - but the original problem returns: after logging out and logging back in, the app silently picks up my existing OIDC session without prompting for re-authentication.

jkuester:
@raphaelkenyuri thank you for testing this!  I am going to do some more digging into what is happening on the Android side and see if maybe we need some kind of update in cht-android as well... :thinking:

PR #10814, opened 2026-04-03
feat(#10810): support calling extension-libs from form expressions

### LINKED ISSUE #10810: Support calling extension-libs from form expressions
This comes out of a conversation with @ChinHairSaintClair at the UI Extensions squad meeting today!  The primary need is around having more complex logic in the [duplicate contact expressions](https://docs.communityhealthtoolkit.org/building/forms/contact/#duplicate-check-expression).  The `duplicate_check.expression` property is just a JS expression that is evaluated to check if a contact is a duplicate. However, including complex logic in the expression gets challenging and difficult to maintain since it is just a single expression and does not really support being a proper function/script where you can declare variables, etc.  

Thankfully, we already have a solution for creating arbitrary JS functions that can re-used across the app: [extension-libs](https://docs.communityhealthtoolkit.org/building/reference/extension-libs/)

I think we should add support for calling extension-libs from within the form expressions (so both `duplicate_check.expression` and `context.expression`.  This would allow for including more complex logic as a reusable function in an extension-lib, and then your form expression could simply call the function with the relevant parameters: `"expression": "extensionLib('dupcheck.js', existing, current)"`

**Additional context**
I think the implementation should be straight-forward.  We can add a new `extensionLib` method [to the XmlFormsContextUtilsService](https://github.com/medic/cht-core/blob/b965953914d786826ed6160812cc7797d67ae4a9/webapp/src/ts/services/xml-forms-context-utils.service.ts).  That function can follow [similar logic](https://github.com/medic/cht-core/blob/da5fe7f731303c6f69315b3992455125f539f91a/webapp/src/js/enketo/medic-xpath-extensions.js#L227) in our [extension-lib xPath feature](https://docs.communityhealthtoolkit.org/building/forms/app/#chtextension-lib) to get the extension-lib from the [CHTDatasourceService](https://github.com/medic/cht-core/blob/622c625427f6b243fb20f8c95fc6442e2be591eb/webapp/src/ts/services/cht-datasource.service.ts) and then call the function with the provided parameters.  

@ChinHairSaintClair Please add any additional data/context! 🙏

vikrantwiz02:
# Description

Adds support for calling extension-libs from within form expressions (`duplicate_check.expression` and `context.expression`). This enables app developers to encapsulate complex logic (e.g. duplicate contact detection) in reusable extension-lib JS functions instead of maintaining it inline as a single expression string.

A new `extensionLib(libId, ...args)` method is added to `XmlFormsContextUtilsService`, following the same pattern used by the existing `cht:extension-lib` xPath function in `medic-xpath-extensions.js`. The method retrieves the extension-lib from `CHTDatasourceService` and invokes it with the provided arguments.

**Usage in duplicate check:**
```json
{
  "duplicate_check": {
    "expression": "extensionLib('dupcheck.js', existing, current)"
  }
}
```

## Fixes #10810

# Code review checklist

# License

### DIFF
--- a/tests/e2e/default/contacts/duplicate-contacts.wdio-spec.js
+++ b/tests/e2e/default/contacts/duplicate-contacts.wdio-spec.js
@@ -9,6 +9,7 @@ const genericForm = require('@page-objects/default/enketo/generic-form.wdio.page
 const moment = require('moment');
 const contactsPage = require('@page-objects/default/contacts/contacts.wdio.page');
 const { getTelemetry, destroyTelemetryDb } = require('@utils/telemetry');
+const { extensionLibDoc } = require('@page-objects/default/enketo/custom-doc.wdio.page');
 
 describe('Duplicate contact detection', () => {
   const CREATED_ON = '4 Apr, 2025';
@@ -40,6 +41,10 @@ describe('Duplicate contact detection', () => {
   };
 
   before(async () => {
+    await utils.saveDoc(extensionLibDoc);
+    const waitForServiceWorker = await utils.waitForApiLogs(utils.SW_SUCCESSFUL_REGEX);
+    await waitForServiceWorker.promise;
+    await commonPage.reloadSession();
     await cookieLogin();
     originalCreatePersonFormDoc = await utils.getDoc(CREATE_PERSON_FORM_ID);
   });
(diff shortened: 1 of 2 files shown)


vikrantwiz02:
Hi @jkuester, thanks for the detailed guidance.
I've refactored as you suggested.

Ready for review.

vikrantwiz02:
@jkuester, I re-ran CI checks 2-3 times, one is failing, i.e., ci-integration-all-all. But I don't think, it is causing by my changes respective to this raised PR.

jkuester:
Yeah, a couple of these tests have gotten really flaky. :sweat: It is a known issue that we are planning to address in separate issues/prs.... :+1:

vikrantwiz02:
@jkuester, Thanks for adding the e2e tests and minor cleanup.

All checks are finally green, including the ci-integration-all-all suite after a long 3.5 hrs. run.

PR #10816, opened 2026-04-03
fix: fail fast with helpful error when nginx container fails to start

### LINKED ISSUE #9491: e2e tests should fail gracefully if the ports are not available (or find a new port)
**Describe the issue**

The e2e tests use docker compose to spin up the CHT containers to run the tests against.  The nginx container tries to bind to the default ports `80/443` and the e2e tests running on the host machine will communicate with the containers via these ports.

However, if those ports are not available (e.g. something else like the https://github.com/medic/nginx-local-ip service is already bound to them) the nginx container will fail to start.  The way this gets presented to the end user (when running `npm run wdio-local`) is very confusion, though (as noted in https://github.com/medic/cht-core/pull/9422).  The tests start the containers in the background and then poll until the api server is available.  But if the nginx container fails to start the e2e tests will just continue polling forever with the error:

```
Checking API
API check failed, trying again in 1 second
502 - "<html>\r\n<head><title>502 Bad Gateway</title></head>\r\n<body>\r\n<center><h1>502 Bad Gateway</h1></center>\r\n<hr><center>nginx/1.19.10</center>\r\n</body>\r\n</html>\r\n"
```

One solution would be to simple end the e2e test run (with a helpful error message) if any of the docker containers fail to start.  

Alternatively, I do not think  there is any hard requirement that the e2e tests use `80/443`.  Instead it seems like we might be able to use custom ports that we determine at runtime are available.

SnehaChaursia:
# Description

E2E setup runs `docker compose up -d` then polls `/api/info` via nginx on the host. If nginx cannot bind ports **80/443** (e.g. another service already listening), the nginx container may never run, but other services still start—so the old flow kept polling with **502** for a long time.

This change, after `docker compose up -d`, waits until the **nginx** container is **`running`**, or fails fast with a clear error if the container is missing, **exited**, or **dead**. Messages mention freeing **80/443** or setting **`NGINX_HTTP_PORT`** / **`NGINX_HTTPS_PORT`** (HTTPS tests already respect **`NGINX_HTTPS_PORT`** in `tests/constants.js`).

Closes #9491

# Code review checklist

- [x] Tested: Unit and/or e2e where appropriate — e2e harness change; verified locally with `npm run wdio-local` / compose path as applicable
- [x] Backwards compatible: Works with existing data and configuration or includes a migration. Any breaking changes documented in the release notes. — Test-only Docker path; k3d unchanged (`isDocker()` guard)

# License


SnehaChaursia:
Hi @jkuester @dianabarsan, could you please review this PR when you get a chance? Let me know if any changes are required. Thanks!

mrjones-plip:
Thanks for the PR @SnehaChaursia ! I'll defer to Josh or Diana for formal code review, but for a pre-flight check, [this line](https://github.com/medic/cht-core/pull/10816/changes#diff-a408f2323aa427483e2b622225e79673edc39e47dd3ff65f5f87dc6cbdecec97R957):

```
`nginx container "${containerName}" did not become running within ${maxTries}s. ` +
```

will result in something like:

```
nginx container "cht_nginx_1" did not become running within 30s
```

Which makes it look like 30 seconds instead of 30 tries.  Could you change it to this?

```
`nginx container "${containerName}" did not become running within ${maxTries} tries. ` +
```

thanks!

SnehaChaursia:
Thanks for pointing that out!
I've updated the message to use ${maxTries} tries so it accurately reflects the logic.
Really appreciate the feedback 🙂

mrjones-plip:
Thanks!

mrjones-plip:
@sugat009 - are you free to review this? Lemme know if you're oversubscribed and we can find another reviewer!

sugat009:
@mrjones-plip i have a few PRs(4/5) lined up before this.

mrjones-plip:
Thanks for letting me know @sugat009 - I've removed you and put this back in [the queue](https://github.com/orgs/medic/projects/350/views/3)

tests/utils/index.js
```diff
@@ -933,6 +933,49 @@ const listenForApi = async () => {
...
+  if (!isDocker()) {
+    return;
+  }
+  const containerName = getContainerName('nginx');
+  const maxTries = 30;
+  for (let i = 0; i < maxTries; i++) {
+    let status;
+    try {
+      status = await runCommand(
+        `docker inspect -f '{{.State.Status}}' ${containerName}`,
```
sugat009, line 946: **issue (non-blocking):** I tried the exact scenario from the issue (another process already holding port 80/443) and noticed the current check doesn't quite catch it. Here's what happens:

I started a stand in container conflicting on port 18080:
```
docker run -d --name probe --restart=always -p 18080:80 nginx:alpine
```

Docker fails with "port is already allocated" but the container still gets left around in `created` state (not `exited` or `dead`). So in the loop, `Status` is always `created`, the fast fail branch never triggers, and we spend the full 30 tries before the timeout error fires.

The reason is that Docker stores the actual failure reason in a separate `Error` field on the container's State, not in the Status. Here's what `docker inspect` shows for the failed probe container:

```json
{
  "Status": "created",
  "ExitCode": 128,
  "Error": "Bind for 0.0.0.0:18080 failed: port is already allocated"
}
```

If we ask Docker for the full `State` as JSON, we can check `Error` too and bail out with the real message immediately. Something like:

```js
const raw = await runCommand(
  `docker inspect -f '{{json .State}}' ${containerName}`,
  { verbose: false }
);
const state = JSON.parse(raw);

if (state.Error) {
  throw new Error(`nginx container "${containerName}" failed to start: ${state.Error}. ${PORT_HINT}`);
}
if (state.Status === 'exited' || state.Status === 'dead') {
  throw new Error(`nginx container "${containerName}" failed to start (status: ${state.Status}, exitCode: ${state.ExitCode}). ${PORT_HINT}`);
}
if (state.Status === 'running') {
  return;
}
```

Bonus: in the `exited`/`dead` branch you also get `ExitCode`, which tells you the real exit code if nginx itself crashes (e.g., bad config) rather than failing at port binding.

Out of curiosity, how did you repro the failure case during testing? I'd like to compare notes since the port conflict path I tried doesn't seem to land in the `exited` branch.

tests/utils/index.js
```diff
@@ -933,6 +933,49 @@ const listenForApi = async () => {
...
+        { verbose: false }
+      );
+    } catch {
```
sugat009, line 949: **suggestion (nit):** the bare `catch` swallows the underlying `docker inspect` error. If something flaky or unexpected fails (daemon hiccup, name collision, weird state), we surface only "container was not found" which can be misleading. Capturing the error and including its `.message` in the thrown Error would help when something other than the obvious case happens.

tests/utils/index.js
```diff
@@ -933,6 +933,49 @@ const listenForApi = async () => {
...
+      throw new Error(
+        `Expected nginx container "${containerName}" was not found after docker compose up. ` +
+        'Ports 80 and/or 443 are often already in use; stop the other service or set NGINX_HTTP_PORT ' +
```
sugat009, line 952: **suggestion (nit):** the `Ports 80 and/or 443 are often already in use; stop the other service or set NGINX_HTTP_PORT and NGINX_HTTPS_PORT (see tests/constants.js for HTTPS).` hint repeats verbatim across all three throws (the last one drops the parenthetical about `tests/constants.js`, so drift has already started). Extracting to a `const` at module scope keeps the three messages in lockstep.

sugat009 (commented):
Thanks Sneha. Checked the function against real docker container states (running/exited/created/restarting), lint passes, and the `isDocker()` guard correctly skips k3d. A few small suggestions inline.

Title nit: `fix(#9491):` matches the conventional commits style most recent fixes use; the lint is happy either way.

sugat009:
Closing. Requested changes done [here](https://github.com/medic/cht-core/pull/11138).

PR #10822, opened 2026-04-07
feat(#10638): not indexing common documents in doc_by_type

### LINKED ISSUE #10638: Optimize map reduce view doc_by_type
The `doc_by_type` map reduce view is the third largest view on the instance sampled.
It emits an entry for every document in the database, but is used only for rare document types.

This can be change to not emit for the largest types, `task`, `data_record`, `contact`, and maybe `target` and `telemetry`.

It seems like an index for documents by their type might be useful in general or for some unknown reason in the future, but its not currently used that way and is very expensive to keep. 
If it's necessary in future to query for all `reports`, `tasks` or `contacts`,  the other views that index those types specifically can be used.

### Definition
- **Map keys:** `[doc.type]`
- **Map values:** (none)
- **Reduce:** None

### Usages
| 1 | [`sentinel/src/config.js:23`](https://github.com/medic/cht-core/blob/master/sentinel/src/config.js#L23) | `{ key: ['translations'], include_docs: true }` | Loads all translation documents at sentinel startup. |
| 2 | [`api/src/migrations.js:20`](https://github.com/medic/cht-core/blob/master/api/src/migrations.js#L20) | `{ include_docs: true, key: ['migration-log'] }` | Retrieves migration log to check which data migrations have run. |
| 3 | [`shared-libs/user-management/src/users.js:117`](https://github.com/medic/cht-core/blob/master/shared-libs/user-management/src/users.js#L117) | `{ include_docs: true, key: ['user-settings'] }` | Retrieves all user-settings documents for user management. |
| 4 | [`webapp/src/ts/services/telemetry.service.ts:73`](https://github.com/medic/cht-core/blob/master/webapp/src/ts/services/telemetry.service.ts#L73) | `{ key: ['form'], include_docs: true }` | Fetches all form documents for telemetry metadata. |
| 5 | [`webapp/src/ts/services/xml-forms.service.ts:56`](https://github.com/medic/cht-core/blob/master/webapp/src/ts/services/xml-forms.service.ts#L56) | `{ include_docs: true, key: ['form'] }` | Fetches all form documents for rendering form lists. |
| 6 | [`webapp/src/ts/services/languages.service.ts:25`](https://github.com/medic/cht-core/blob/master/webapp/src/ts/services/languages.service.ts#L25) | `{ key: ['translations'], include_docs: true }` | Fetches all translation documents for language selector. |
| 7 | [`admin/src/js/controllers/display-languages.js:94`](https://github.com/medic/cht-core/blob/master/admin/src/js/controllers/display-languages.js#L94) | `{ key: ['translations'], include_docs: true }` | Fetches translations for admin language management. |
| 8 | [`admin/src/js/controllers/users.js:25`](https://github.com/medic/cht-core/blob/master/admin/src/js/controllers/users.js#L25) | `{ include_docs: true, key: ['user-settings'] }` | Fetches user-settings for admin user management page. |
| 9 | [`admin/src/js/controllers/forms-xml.js:22`](https://github.com/medic/cht-core/blob/master/admin/src/js/controllers/forms-xml.js#L22) | `{ include_docs: true, key: ['form'] }` | Fetches form documents for admin XML forms management. |
| 10 | [`admin/src/js/controllers/display-translations.js:58`](https://github.com/medic/cht-core/blob/master/admin/src/js/controllers/display-translations.js#L58) | `{ key: ['translations'], include_docs: true }` | Fetches translations for admin translations management. |
| 11 | [`cht-conf/src/lib/forms-list.js:6`](https://github.com/medic/cht-conf/blob/master/src/lib/forms-list.js#L6) | `{ include_docs: true, key: ['form'] }` | Lists all forms on remote server during cht-conf operations. |

witash:
closes #10638


ddocs/medic-db/medic-client/views/doc_by_type/map.js
```diff
@@ -1,3 +1,9 @@
 function(doc) {
-  emit([ doc.type ]);
+  var excluded = [
+    'task', 'data_record', 'contact', 'target', 'telemetry',
```
dianabarsan: I don't think we'll ever have `telemetry` docs in the medic database. 
But this feels like it's covering ... every type of document? 
What is left? 

Ah, I see: 
- in admin app and webapp to get translations. we can change that - just like in every place that deals with translations - to use the `messages-` prefix in and allDocs call. 
- in users controller to get user-settings. those docs are also prefixed with `org.couchdb.user.`, so that can be changed as well. 
- in webapp to get forms. forms are also prefixed, so can be removed. 

Is there anything left? 
Maybe we can just delete the view entirely.

dianabarsan (changes requested):

ddocs/medic-db/medic-client/views/doc_by_type/map.js
witash: So it turns out we can't simply remove it entirely, because its still called directly by cht-conf and cht-scripts.
I thought about having api intercept requests to the view  and call allDocs with prefixes instead of just forwarding to couch (in https://github.com/medic/cht-core/pull/10822/commits/acd524472a3d8330c0764bc625580e0a64968921), but I don't know if its worth all that.

But, at least was able to change all the references in cht-core to use allDocs with prefixes.
And change the view to emit for specifically the types those are using (`forms` and `user-settings`)

The other complication was the migration log; some very very old code can only get the migration log using type, since it did not have a special id, but newer code uses migration logs with special ids, and automatically migrates old migration logs if it finds them, so its only 3.something deployments that would need the view; I think its safe to just remove that since a direct 3.x -> 5.x upgrade is not supported anyway.

dianabarsan: I would not worry about the migration log. We have not added migrations in 10 versions, so you can exclude that.

admin/src/js/controllers/display-languages.js
```diff
@@ -91,8 +91,9 @@ angular.module('controllers').controller('DisplayLanguagesCtrl',
       $scope.loading = true;
       $q
         .all([
-          DB().query('medic-client/doc_by_type', {
-            key: [ DOC_TYPES.TRANSLATIONS ],
+          DB().allDocs({
+            start_key: PREFIXES.TRANSLATIONS,
+            end_key: PREFIXES.TRANSLATIONS + '￰',
```
dianabarsan: You need to add a corresponding suffix here `\ufff0` to target the string range: https://docs.couchdb.org/en/stable/ddocs/views/collation.html#string-ranges

admin/src/js/controllers/display-translations.js
```diff
@@ -56,8 +56,9 @@ angular.module('controllers').controller('DisplayTranslationsCtrl',
 
     const updateTranslations = function() {
       return DB()
-        .query('medic-client/doc_by_type', {
-          key: [ DOC_TYPES.TRANSLATIONS ],
+        .allDocs({
+          start_key: PREFIXES.TRANSLATIONS,
+          end_key: PREFIXES.TRANSLATIONS + '￰',
```
dianabarsan: You need to add a corresponding suffix here \ufff0 to target the string range: https://docs.couchdb.org/en/stable/ddocs/views/collation.html#string-ranges

admin/src/js/controllers/forms-xml.js
```diff
@@ -20,10 +22,11 @@ angular.module('controllers').controller('FormsXmlCtrl',
     const getForms = () => {
       const options = {
         include_docs: true,
-        key: ['form']
+        start_key: PREFIXES.FORM,
+        end_key: PREFIXES.FORM + '￰',
```
dianabarsan: You need to add a corresponding suffix here \ufff0 to target the string range: https://docs.couchdb.org/en/stable/ddocs/views/collation.html#string-ranges

admin/src/js/controllers/users.js
```diff
@@ -22,8 +24,12 @@ angular.module('controllers').controller('UsersCtrl',
 
     $scope.updateList = function() {
       $scope.loading = true;
-      const params = { include_docs: true, key: ['user-settings'] };
-      DB().query('medic-client/doc_by_type', params)
+      const params = {
+        include_docs: true,
+        start_key: PREFIXES.COUCH_USER,
+        end_key: PREFIXES.COUCH_USER + '￰',
```
dianabarsan: You need to add a corresponding suffix here \ufff0 to target the string range: https://docs.couchdb.org/en/stable/ddocs/views/collation.html#string-ranges

tests/utils/index.js
```diff
@@ -694,7 +694,10 @@ const getDefaultForms = async () => {
     const doc = await db.get(docName);
     PROTECTED_DOCS.push(...doc.forms);
   } catch {
-    const result = await db.allDocs({ startkey: 'form:', endkey: 'form:\ufff0' });
+    const result = await db.allDocs({
+      startkey: PREFIXES.FORM,
+      endkey: PREFIXES.FORM + '￰',
```
dianabarsan: You need to add a corresponding suffix here \ufff0 to target the string range: https://docs.couchdb.org/en/stable/ddocs/views/collation.html#string-ranges

tests/utils/index.js
```diff
@@ -881,7 +884,11 @@ const createUsers = async (users, meta = false, password_change_required = false
 };
 
 const getAllUserSettings = () => db
-  .query('medic-client/doc_by_type', { include_docs: true, key: ['user-settings'] })
+  .allDocs({
+    include_docs: true,
+    start_key: PREFIXES.COUCH_USER,
+    end_key: PREFIXES.COUCH_USER + '￰',
```
dianabarsan: You need to add a corresponding suffix here \ufff0 to target the string range: https://docs.couchdb.org/en/stable/ddocs/views/collation.html#string-ranges

webapp/src/ts/services/languages.service.ts
```diff
@@ -22,7 +22,11 @@ export class LanguagesService {
 
     const result = await this.dbService
       .get()
-      .query('medic-client/doc_by_type', { key: [DOC_TYPES.TRANSLATIONS], include_docs: true });
+      .allDocs({
+        start_key: PREFIXES.TRANSLATIONS,
+        end_key: PREFIXES.TRANSLATIONS + '￰',
```
dianabarsan: You need to add a corresponding suffix here \ufff0 to target the string range: https://docs.couchdb.org/en/stable/ddocs/views/collation.html#string-ranges

webapp/src/ts/services/telemetry.service.ts
```diff
@@ -70,7 +70,11 @@ export class TelemetryService {
     return Promise
       .all([
         this.dbService.get().get('_design/medic-client'),
-        this.dbService.get().query('medic-client/doc_by_type', { key: ['form'], include_docs: true }),
+        this.dbService.get().allDocs({
+          start_key: PREFIXES.FORM,
+          end_key: PREFIXES.FORM + '￰',
```
dianabarsan: You need to add a corresponding suffix here \ufff0 to target the string range: https://docs.couchdb.org/en/stable/ddocs/views/collation.html#string-ranges

dianabarsan (changes requested):

admin/src/js/controllers/display-languages.js
witash: these are actually claude shenanigans; it's added U+FFF0 as a literal directly to the source, which is either not rendered or rendered as whitespace depending on terminal or browser.

But that's not exactly readable, changed to the escape sequence

dianabarsan: I went through the same ordeal, several times. I think I even added a memory at some point. To no avail :D

ddocs/medic-db/medic-client/views/doc_by_type/map.js
```diff
@@ -1,3 +1,6 @@
 function(doc) {
-  emit([ doc.type ]);
+  var indexed_types = ['form', 'user-settings'];
```
dianabarsan: Please use camelCase for variables.

shared-libs/constants/src/index.js
```diff
@@ -60,10 +60,12 @@ const USER_ROLES = {
 
 const DB_ADMIN_ROLES = [USER_ROLES.ADMIN, USER_ROLES.COUCHDB_ADMIN];
 
-// Prefixes
+// Document ID prefixes used for _all_docs prefix range scans.
+// For end_key, append '\ufff0' (U+FFF0) to the prefix to bound the range.
```
dianabarsan: I think this is just claude being chatty. there's no reason to mention \ufff0 in the constants file.

dianabarsan (approved):
Two requests inline, but approving to unblock. 
Just a reminder that we need to bundle all these changes in one "update the views" swoop (so this is not ready to be merged I presume).

witash:
merged to 10636-view-changes

PR #10823, opened 2026-04-07
feat(#10794): adds replication failure log

### LINKED ISSUE #10794: Implement failed user replication tracking
Some time ago, in an effort to identify users with too many docs, we added the `replication-count` log (https://github.com/medic/cht-core/issues/6251). However, this has a severe limitation where it requires the replication request to succeed in order to log anything. 
This means that the largest most broken users that can never log in are never captured, and these were the ones we were trying to detect and guard. 

This work should be expanded to also log _failed_ replications. We will not know the number of documents, but we might know _some_ data that can be useful for troubleshooting. 

**Additional context**
Seeing replication-count logs with 500.000 users, which means that at least one request succeeded.

dianabarsan:
<!--
Please use semantic PR titles that respect this format:

<type>(#<issue number>): <subject>

Quick example:

feat(#1234): add hat wobble
^--^(#^--^): ^------------^
|     |      |
|     |      + - > subject
|     |
|     + -------- > issue number
|
+ -------------- > type: build|feat|fix|perf|refactor|test|chore

https://docs.communityhealthtoolkit.org/contribute/code/workflow/#commit-message-format
-->

# Description

 - Moves all replication related api-services into a new `replication` dedicated folder. The list became huge!
 - Add replication failure logging to `/api/v1/replication/get-ids`, matching the existing setup on the initial-replication route                                                                                                                                                                                                                                
  - Each failure records timestamp, status code, duration, request ID, roles, and subjects count (if available)                                                                                                                                                                                                                                                               
  - Failures are stored per-user per-month in the logs DB, capped at 50 detailed entries with a running total count                                                                                                                                                                                                                                             
  - Captures both server errors and client cancellations                                                                                                                                                                                                                                                                                                        
  - Add e2e tests for failure logging, accumulation, monthly separation, per-user separation, and the 50-entry cap
  - Adds api endpoint to get replication failures for a given month. Defaults to the current month. 

closes #10794

## AI Disclosure

This is developed with the help of Claude. I have made all design decisions, reviewed and corrected all code. 

# Code review checklist
<!-- Remove or comment out any items that do not apply to this PR; in the remaining boxes, replace the [ ] with [x]. -->



# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10794-replication-fails/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10794-replication-fails/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10794-replication-fails/docker-compose/cht-couchdb-clustered.yml)


# License


dianabarsan:
I know your review queue is long, but i have to ask for another one. I appreciate the time you spend on this review.

api/src/services/replication/authorization.js
```diff
@@ -375,6 +375,8 @@ const getAuthorizationContext = async (userCtx) => {
     authCtx.subjectsDepth[UNASSIGNED_KEY] = 0;
   }
 
+  authCtx.userCtx.subjectsCount = authCtx.subjectIds.length;
```
jkuester: Minor, but can we have a unit test that covers this?

api/src/middleware/authorization.js
```diff
@@ -95,5 +97,23 @@ module.exports = {
     }
 
     return getUserSettings(req).then(next);
+  },
+
+  captureReplicationFailures: (req, res, next) => {
+    if (auth.isOnlineOnly(req.userCtx)) {
+      return next();
+    }
```
jkuester: Maybe I am missing something here, but do we ever _expect_ to get here for an online user?  An online user should not ever call `/get-ids`, right?

I am thinking maybe we should just update the router to move this middleware down after the `authorization.onlineUserPassThrough` gate.  

If we wanted to be extra sure that this never gets called for online users in the future, we could throw an exception here... 🤷

api/src/controllers/replication-failure-log.js
```diff
@@ -0,0 +1,181 @@
...
+   *     summary: Get replication failure log summaries
+   *     operationId: v1ReplicationFailureLogsGet
+   *     description: >
+   *       Returns a lightweight summary (without the detailed `failures` array) of replication failure logs for
+   *       all users in the specified month. Defaults to the current month if no `month` is provided. Only allowed
+   *       for database admins.
+   *     tags: [User]
+   *     parameters:
+   *       - in: query
+   *         name: month
```
jkuester: Minor, but the `target` endpoint accepts [basically the same parameter](https://github.com/medic/cht-core/blob/f8fe649aa5fb1601a504c764942f262412dc4538/api/src/controllers/target.js#L104) with the name `reporting_period`.  Not sure if it would be better or more confusing to keep the naming consistant and use `reporting_period` here....?  🤔

api/src/controllers/replication-failure-log.js
```diff
@@ -0,0 +1,181 @@
...
+   *                           revision count (the log doc is only written on failure captures).
+   *       '401':
+   *         $ref: '#/components/responses/Unauthorized'
+   */
+  get: async (req, res) => {
+    try {
+      await assertAdmin(req);
+      const month = req.query.month || moment().format('YYYY-MM');
+      const logs = await replicationFailureLog.getSummariesByMonth(month);
+      res.json({ month, logs });
```
jkuester: Not a huge deal, but I don't know why we need to return the `month` value here.  🤷

jkuester: Not a huge deal, but I don't know why we need to return the `month` value here.  🤷

api/src/controllers/replication-failure-log.js
```diff
@@ -0,0 +1,181 @@
...
+   *       '401':
+   *         $ref: '#/components/responses/Unauthorized'
+   */
+  getForUser: async (req, res) => {
+    try {
+      await assertAdmin(req);
+      const user = req.params.user;
+      if (req.query.month) {
+        const log = await replicationFailureLog.getForUserAndMonth(req.query.month, user);
+        return res.json({ month: req.query.month, user, log });
```
jkuester: As above, it feels a bit redundant to return the month/user here.

api/src/controllers/replication-failure-log.js
```diff
@@ -0,0 +1,181 @@
...
+   *                   description: All-months response when `month` is omitted.
+   *                   properties:
+   *                     user:
+   *                       type: string
+   *                     logs:
+   *                       type: array
+   *                       description: One full log document per month the user had failures in.
+   *                       items:
+   *                         type: object
+   *                         description: See the single-month log schema above.
```
jkuester: So, this might be a good place for a reusable schema component and some refs.  You could put a comment like this at the top-level:

```js
/**
 * @openapi
 * components:
 *   schemas:
 *     ReplicationFailureLog:
 *       type: object
 *       description: The full failure log document for a user in a given month.
 *       properties:
 *         _id:
 *           type: string
 *         _rev:
 *           type: string
 *         type:
 *           type: string
 *           description: Always 'replication-fail'.
 *         user:
 *           type: string
 *         timestamp:
 *           type: number
 *         total_failures:
 *           type: number
 *         failures:
 *           type: array
 *           description: >
 *             The most recent failures for the user (capped at 50). The total count is available
 *             in `total_failures`.
 *           items:
 *             type: object
 *             properties:
 *               timestamp:
 *                 type: number
 *               status_code:
 *                 type: number
 *                 description: Response status code. 0 if the client cancelled the request.
 *               duration:
 *                 type: number
 *                 description: Request duration in milliseconds.
 *               request_id:
 *                 type: string
 *               roles:
 *                 type: array
 *                 items:
 *                   type: string
 *               subjects_count:
 *                 type: number
 *                 description: Number of subjects the user had access to at the time of the failure.
 */
```

Then your response block here just becomes:


```js
   *       '200':
   *         description: Full failure log documents for the user.
   *         content:
   *           application/json:
   *             schema:
   *               oneOf:
   *                 - type: object
   *                   description: Single-month response when `month` is provided.
   *                   properties:
   *                     month:
   *                       type: string
   *                     user:
   *                       type: string
   *                     log:
   *                       description: The full failure log document, or null if no failures were recorded.
   *                       oneOf:
   *                         - $ref: '#/components/schemas/ReplicationFailureLog'
   *                         - type: 'null'
   *                 - type: object
   *                   description: All-months response when `month` is omitted.
   *                   properties:
   *                     user:
   *                       type: string
   *                     logs:
   *                       type: array
   *                       description: One full log document per month the user had failures in.
   *                       items:
   *                         $ref: '#/components/schemas/ReplicationFailureLog'
```


This is just more semantically rich since it captures the type of object returned in the `logs` array.

api/src/services/replication/replication-failure-log.js
```diff
@@ -0,0 +1,102 @@
...
+const getLog = async (userName) => {
+  const docId = getDocId(userName);
+
+  try {
+    return await db.medicLogs.get(docId);
+  } catch (err) {
+    if (err.status === 404) {
+      return {
+        _id: docId,
+        type: LOG_TYPE,
```
jkuester: Is this `type` property just here for some kind of future proofing?  I see that we are not setting a `type` for the `upgrade_log`, `replication-count` or `connected-user` docs....

api/src/services/replication/replication-failure-log.js
```diff
@@ -0,0 +1,102 @@
...
+        user: userName,
+        timestamp: moment().valueOf(),
```
jkuester: Minor, but I see that the `replication-count` docs use `date` for a similar field.

api/src/services/replication/replication-failure-log.js
```diff
@@ -0,0 +1,102 @@
...
+const moment = require('moment');
+
+const LOG_TYPE = 'replication-fail';
+const MAX_FAILURES = 50;
+
+const captureFailure = async (userCtx, requestId, statusCode, duration) => {
+  const log = await getLog(userCtx.name);
+
+  const failure = {
+    timestamp: moment().valueOf(),
```
jkuester: If we go with `date` below, we should probably update this to be consistant.

api/src/services/replication/replication-failure-log.js
```diff
@@ -0,0 +1,102 @@
...
+  const prefix = `${LOG_TYPE}-${month}-`;
+  const result = await db.medicLogs.allDocs({
+    startkey: prefix,
+    endkey: `${prefix}\ufff0`,
+  });
+  return result.rows.map(row => ({
+    _id: row.id,
+    user: row.id.substring(prefix.length),
+    // to avoid a heavy endpoint, don't read doc bodies and extrapolate failure count via rev number
+    total_failures: Number.parseInt(row.value.rev.split('-')[0], 10),
```
jkuester: Normally I would not approve a PR with this level of hackery, but I trust your intuitions regarding Couch stuff... If we are concerned about endpoint performance it seems like we should change our data model, do some `reducing`, and/or implement paging.  I am going to assume you have good reasons why this is better?  🤷

api/src/controllers/replication-failure-log.js
```diff
@@ -0,0 +1,181 @@
...
+      const logs = await replicationFailureLog.getSummariesByMonth(month);
+      res.json({ month, logs });
+    } catch (err) {
+      serverUtils.error(err, req, res, true);
+    }
+  },
+
+  /**
+   * @openapi
+   * /api/v1/replication-failure-logs/{user}:
```
jkuester: Okay, some opinions on api design.  Please take or leave these as makes sense to you. I am happy to ship with the api as it is, but there are a few things that I think we could do differently.  

I am not a big fan of endpoints that can return _different shapes/types_ of data for the same operation. I find they are harder to understand as a consumer and more work to use correctly. So, I already am not super exited that this endpoint can return a `log` object or a `logs` array.  

It is also confusing to me why `user` is a route parameter, but `month` is a query parameter (except for the fact you want to make `user` required and `month` is optional).  

---

My ideal API structuring would be to have the base `/api/v1/replication-failure-logs` endpoint continue to return the array of summaries.  It can have optional `month` and `user` query parameters to help qualify which summaries should be returned.  Just require that at least one of these params must be set (It feels "surprising" to me that we would default the month to current...).

Then we have a `/api/v1/replication-failure-logs/{id}` endpoint that just returns the whole failure log doc for the given id (no array).  The summary data already incudes the `_id` field when can be used by consumers to hit this endpoint and get the full data.

tests/integration/api/controllers/replication-failure-log.spec.js
```diff
@@ -0,0 +1,309 @@
...
+  return utils.request(options);
+};
+
+const parentPlace = {
+  _id: 'PARENT_PLACE',
+  type: 'district_hospital',
+  name: 'Big Parent Hostpital'
+};
+
+const users = [
```
jkuester, line 50: Minor, but can use the users factory instead of manually building out these users.

tests/integration/api/controllers/replication-failure-log.spec.js
```diff
@@ -0,0 +1,309 @@
...
+    await utils.revertDb([], true);
+    await utils.deleteUsers(users, true);
+    await utils.deletePurgeDbs();
+  });
+
+  afterEach(async () => {
+    await clearLogs();
+  });
+
+  it('should not create a failure log on a successful requests', async () => {
```
jkuester: Maybe I am missing something here, but should we actually call the `get-ids` endpoint successfully before concluding there are no failures logs created?

tests/integration/api/controllers/replication-failure-log.spec.js
jkuester, line 1: Do we need at least  one test that also hits the `/api/v1/initial-replication/get-ids` endpoint?

jkuester (changes requested):
Very nice! I left a bunch of random thoughts inline.  

Have you thought at all about if we should report a total value of these failures via the `/monitoring` api?  Maybe even just a count of the distinct users that have replication failures.

api/src/controllers/replication-failure-log.js
jkuester: The more I think about this the more I wonder how these endpoints fit with the other ones we have.  Specifically we already have `/api/v1/users-info` and `/api/v1/users-doc-count` that are both related to replication and doc counts.  I _think_ `users-info` actually just runs the steps for a replication and calculates in real-time what the users replication counts are going to look like.  

However, `users-doc-count` looks like it is making similar lookups in `medic-logs`  to get the data off the `replication-count` docs.  I wonder if it might make sense to combine our `replication-failure-logs` data into this endpoint..... :thinking:   Def still a half-baked idea rolling around in my head, but worth some thought.

mrjones-plip:
> if we should report a total value of these failures via the /monitoring api

a million times yes!! But also, I can wait 'til another iteration if it delays this PR/is too much feature creep.

api/src/controllers/replication-failure-log.js
dianabarsan: :-? I don't care much about the name. I can switch to reporting_period.

api/src/services/replication/replication-failure-log.js
dianabarsan: I was totally hacking here. :D 
I think paging would be the better idea and I'll implement that.

api/src/controllers/replication-failure-log.js
dianabarsan: I keep having ideas on how to mix this with other endpoints. 
After having a conversation with @mrjones-plip, I want to add an endpoint that only returns users that _never_ successfully replicated within the current month - one that takes user doc count logs and checks the date on last time they replicated last, and if we have failures and no successes, the endpoint would return. 
Or maybe the main endpoint should just do that directly? 

I did implement the user and month both as params initially, and then changed it. I mostly changed it when i had a mammoth of a yaml text preceding it, explaining the different return types :D 

But I don't feel strongly about it. I followed the /<user> pattern that the users endpoint also use.

tests/integration/api/controllers/replication-failure-log.spec.js
dianabarsan: Woops! That got lost somewhere. Thanks for noticing!

tests/integration/api/controllers/replication-failure-log.spec.js
dianabarsan: All tests should be hitting `/api/v1/initial-replication/get-ids` endpoint (because that's the only time these logs get created).

api/src/services/replication/replication-failure-log.js
dianabarsan: removing. force of habit.

tests/integration/api/controllers/replication-failure-log.spec.js
jkuester: Am I missing something here in the router code? :thinking: I had understood that the replication failure middleware was running on both endpoints...

https://github.com/medic/cht-core/pull/10823/changes#diff-ca6928872d8d920a49027473c56cc03e5da0d2d94519bb7a69ccf32d51436087R925

api/src/controllers/replication-failure-log.js
jkuester: > I want to add an endpoint that only returns users that never successfully replicated within the current month - one that takes user doc count logs and checks the date on last time they replicated last, and if we have failures and no successes, the endpoint would return.
Or maybe the main endpoint should just do that directly?

Do we need a dedicated endpoint just for users that _never_ replicated?  The more I think on this the more I am wondering if we could have an endpoint like `user-replication-status` or `user-replication-history` that can return a complete collection of data from the `connected-user`, `replication-count`, and `replication-fail` docs.  We deprecate `users-doc-count` in favor of this (and/or add this as `/api/v2/users-doc-count`).  

If we return the last successful replication timestamp along with the last failure timestamp, it should be trivial for consumers to calculate who is currently failing to replicate.


---

All of this feels super relevant to the conversation we were having with @ChinHairSaintClair in the UI Extensions squad meeting.  They are interested in building a "Syncing Status" dashboard to help identify users who have not synced or are having trouble syncing.  An endpoint like this might be exactly what is needed to provide that data... :thinking:

dianabarsan: I have decided to add this endpoint, but in a separate PR to not burden this with more code. Also adding the counts to monitoring in a second PR, for the same reason. 

Is that ok?

tests/integration/api/controllers/replication-failure-log.spec.js
dianabarsan: which "both" endpoints? it runs only on get-ids. I'm super confused :D

api/src/controllers/replication-failure-log.js
jkuester: Yeah, that makes good sense. :+1:

tests/integration/api/controllers/replication-failure-log.spec.js
jkuester: :sweat_smile: One of us is going crazy (probably both... :sweat_smile: ).

This is the diff I see in this PR for the `api/src/routing.js` file.  I _thought_ this meant that the logging middleware was running and capturing errors for both `/api/v1/initial-replication/get-ids` and `/api/v1/replication/get-ids`... :shrug: 

image

jkuester: :facepalm:  okay... looking at my own screenshot, I think I see disconnect.  :rofl:   Technically there are two endpoints, but in reality, `initial-replication/get-ids` is just an _alias_ for `replication/get-ids` and they both call through to the exact same controller code.  

With that in mind, I now agree that we don't need extra integration tests for `initial-replication/get-ids` and this is fine as is. :+1: Sorry for the confusion!

dianabarsan:
Hi @jkuester 

A bunch of changes occured, mostly as part of your review feedback: 

* API
  * `month` -> `reporting_period`
  * cht-datasource-like cursor based pagination and response style (to keep PR small i didn't make changes to datasource, but added a small lib into api)
  * `user` is now a query param and I improved performance for getting all user failed periods

* Doc format
  * `timestamp` -> `date` to match connected user logs
  * added `docs_count` and `unpurged_docs_count` along `subjects_count`, to show how far the request progressed. all these default to `unknown`. 

* Routing
  * I finally understood the confusion about endpoints, because there were two :D. `initial-replication/get-ids` was added in 4.2 when we did the initial replication refactor, and in 4.4 became deprecated with `replication/get-ids` added with the Nairobi Protocol. We left `initial-replication/get-ids` in case older clients tried to replicate and then in 5.0 we forgot to remove the endpoint. Because updating from 4.2 to 5.x requires a step to 4.5 (?), it is now completely safe to be removed, so I did. 

I _hope_ I didn't miss anything. Review at your discretion. Appreciate your time and thoughts!

api/src/routing.js
```diff
@@ -913,16 +915,11 @@ app.all(
   dbDocHandler.request,
   authorization.setAuthorized // adds the `authorized` flag to the `req` object, so it passes the firewall
 );
-app.get(
-  '/api/v1/initial-replication/get-ids',
-  authorization.handleAuthErrors,
-  authorization.onlineUserPassThrough,
-  replication.getDocIds,
-);
```
jkuester, line 921: Image

api/src/controllers/replication-failure-log.js
```diff
@@ -0,0 +1,136 @@
...
+
+module.exports = {
+  /**
+   * @openapi
+   * /api/v1/replication-failure-logs:
+   *   get:
+   *     summary: Get replication failure logs
+   *     operationId: v1ReplicationFailureLogsGet
+   *     description: >
+   *       Returns a paginated page of full replication failure log documents. Optionally filter by `user`
```
jkuester, line 16:
```suggestion
    *       Returns a page of replication failure log documents. Optionally filter by `user`
```

api/src/services/replication/replication-failure-log.js
```diff
@@ -1,19 +1,28 @@
...
+const REPORTING_PERIOD_FORMAT = 'YYYY-MM';
+const UNKNOWN = 'unknown';
+const MAX_PERIODS = 60;
 
 const captureFailure = async (userCtx, requestId, statusCode, duration) => {
   const log = await getLog(userCtx.name);
 
+  // Counts are only set on userCtx as each phase of the request completes. When a count is missing
+  // we record 'unknown' instead of omitting the key, so a stable shape on the failure entry tells you
+  // (by which counters are 'unknown') how far the request progressed before failing.
```
jkuester, line 16: Don't have to change anything here, but setting `unknown` in these cases does not seem to make it any easier for consumer code to parse the data from the doc when compared to setting:

- `""`,
- `null`
- `undefined`

I kind of understand wanting to keep the object shape "stable", but I am not really sure how much value there is in that when we don't have any meaningful default value to use... 🤷

api/src/services/replication/replication-failure-log.js
```diff
@@ -0,0 +1,179 @@
...
+
+const getPageByUser = async ({ user, skip, limit }) => {
+  const oldestPeriod = await findOldestReportingPeriod();
+  if (!oldestPeriod) {
+    return { data: [], cursor: null };
+  }
+
+  // Username is the doc-id suffix, preventing prefix-scan; bulk-fetch the user's candidate keys (max 60).
+  const candidateKeys = enumerateReportingPeriods(oldestPeriod).map(period => getDocId(user, period));
+  const result = await db.medicLogs.allDocs({ keys: candidateKeys, include_docs: true });
```
jkuester: Is there a strong reason we cannot just index these docs by `user`?  This is kind of a crazy amount of logic to have/maintain (even if the performance is okay).  It is fancy that all of this actually works, but at the same time it really smells like a premature optimization.

tests/integration/api/controllers/replication-failure-log.spec.js
```diff
@@ -0,0 +1,336 @@
...
+    if (limit !== undefined) {
+      qs.limit = limit;
+    }
+    return utils.request({ path: '/api/v1/replication-failure-logs', qs });
+  };
+
+  const getUserFailureLog = async (username) => {
+    const response = await getFailureLogs({ user: username });
+    const currentPeriodId = getFailureLogId(username);
+    return response.data.find(log => log._id === currentPeriodId) || response.data[0] || null;
```
jkuester, line 81: Minor, but the `currentPeriodId` logic here feels like surprizing behavior that might be confusing to debug later.  (When we call this function are we _expecting_ a doc for the current period or are we expecting to just get back the first one?  How can someone reading the code  know?)   Would rather just have two dedicated functions...

jkuester (approved):
Couple minor suggestions/thoughts, but this is good to go!  👍 LGTM

api/src/routing.js
dianabarsan: I laughed out loud at this one.

tests/integration/api/controllers/replication-failure-log.spec.js
dianabarsan: Changed to just grab the last 60 periods optimistically.

api/src/services/replication/replication-failure-log.js
dianabarsan: I think i have view PTSD now.

api/src/services/replication/replication-failure-log.js
dianabarsan: I just wanted something that was very explicit. `undefined` gets dropped on json stringify, and "" or `null` would work fine. I'll think about it and will probably change my mind about it when i add the monitoring and the other endpoints. Thanks for the feedback!

PR #10831, opened 2026-04-08
feat(#10637): calculate doc summaries

### LINKED ISSUE #10637: Remove or optimize map reduce view doc_summaries_by_id
The `doc_summaries_by_id` map reduce view is by far the largest view on the instances sampled.
This is because the values stored in the btree are relatively large (and it will index every report and every contact).

It is used only by online users and in admin; for offline users summaries are calculated on demand rather than being loaded from a view (see https://github.com/medic/cht-core/issues/4666).

Since there is already a precedent for calculating these summaries on demand I don't see any reason why they should still be stored in a map/reduce view.
Instead the duplicated summary functions can be moved into a single shared-lib where it can be shared between admin and webapp, and webapp doesn't distinguish between offline and online users; it just always get the summaries from the shared summary service.

### Definition
- **Database:** medic (server-side only)
- **Map keys:** `doc._id`
- **Map values:** Summary object containing `_rev`, `from`, `phone`, `form`, `read`, `valid`, `verified`, `reported_date`, `contact`, `lineage`, `subject`, `case_id` (for reports) or `_rev`, `name`, `phone`, `type`, `contact_type`, `contact`, `lineage`, `date_of_death`, `muted` (for contacts)
- **Reduce:** None

### Usages
| 1 | [`admin/src/js/services/message-queue.js:102`](https://github.com/medic/cht-core/blob/master/admin/src/js/services/message-queue.js#L102) | `{ keys: ids }` | Fetches contact summaries by IDs for the admin message queue to display recipient info. |
| 2 | [`admin/src/js/services/get-summaries.js:103`](https://github.com/medic/cht-core/blob/master/admin/src/js/services/get-summaries.js#L103) | `{ keys: ids }` | Fetches document summaries for admin data record display (remote/online mode). |
| 3 | [`webapp/src/ts/services/get-summaries.service.ts:111`](https://github.com/medic/cht-core/blob/master/webapp/src/ts/services/get-summaries.service.ts#L111) | `{ keys: ids }` | Fetches document summaries for webapp data record display (online users only). |

**Note:** Both [`admin/src/js/services/get-summaries.js:59`](https://github.com/medic/cht-core/blob/master/admin/src/js/services/get-summaries.js#L59) and [`webapp/src/ts/services/get-summaries.service.ts:65`](https://github.com/medic/cht-core/blob/master/webapp/src/ts/services/get-summaries.service.ts#L65) contain a **duplicated copy** of this view's map function logic as a client-side fallback for offline users (tagged with a warning comment referencing [medic/medic#4666](https://github.com/medic/medic/issues/4666)).

witash:
closes #10637


witash:
@jkuester need more stuff to review?
This moves calculating doc summaries from a map/reduce view to a function in shared-lib, and uses that shared-lib everywhere the map/reduce view was used, which webapp for online users only (offline users already were using a copy/pasted function) and some places in admin.

shared-libs/doc-summaries/src/index.js
```diff
@@ -0,0 +1,112 @@
...
+  if (isContact(doc)) {
+    return summariseContact(doc);
+  }
+};
+
+module.exports = {
+  summarise,
+  isContact,
+  getLineage,
+  getSubject,
```
jkuester: Do we need to export these functions? I don't see them used outside of this lib and it is a bit confusing since I would expect any external code should just be using `@medic/contact-types-utils` or `@medic/lineage`, etc directly.

jkuester (approved):
Cash money!  🪙 🪙 🪙 

I have gone though all the changes without much to suggest. (The existing duplication was wild... 😓)  

The only thing I keep wondering about is if we should just add this as a cht-datasource endpoint? 🤔   I am trying hard to avoid putting _everything_ into cht-datasource 😅. However, I see that Gareth's [justification](https://github.com/medic/cht-core/pull/4674) for keeping the view around at all was:

> the doc_summaries_by_id view is now only used to reduce bandwidth

Your current changes in this PR will def negatively affect the network traffic load for _online users_.  Do we have a measurement for this?  Unfortunately, I am afraid accurate numbers might be hard here since it will heavily depend on exactly how much additional data is stored on the contacts/reports.  

We could add new `/api/v1/contact/summary` and `/api/v1/report/summary` endpoints to support returning arrays of summaries.  Because of how cht-datasource works, the offline user functionality would remain basically the same. For online users, the network traffic between the api server and the client browser would be lower, though, since we would not be sending as much data....

I think I would register my concern here as "medium-low". If we hit a case where online users are retrieving a prohibitively large amount of docs, it would probably be better to add more paging so we are loading less at a time anyway (vs just trying to load summaries for an unreasonably large amount of entries).  On the other hand, maybe it would be better overall for performance to have summaries be more of a core data concept we can leverage from cht-datasource.

Unless you have strong  feelings here @witash, I think my vote is to ship this as-is.  As we continue to refactor the direct DB usages in the webapps into cht-datasource we can circle back at that time and add endpoints if we feel it is useful.

witash:
@jkuester Since this is for online users I wouldn't have thought that network bandwidth is that much of a concern, or that it would be that much more without summarizing that it would be significant overall. But just guessing, don't have any tests to support that.

Maybe doesn't matter, cht-datasource is a better implementation in any case.
Might as well do it now instead of circling back.
So its basically a new branch now, can you please re-review?

witash:
per conversation yesterday I did quick test for the difference in size to estimate the effect on network bandwidth, and the time it takes to summarize documents.
```
## Reports (n=100000)
  Total full-doc JSON:   83.58 MiB
  Total summary JSON:    35.03 MiB
  Reduction:             58.1%
  Avg per full doc:      876 B
  Avg per summary:       367 B

## Contacts (n=110600)
  Total full-doc JSON:   46.06 MiB
  Total summary JSON:    24.64 MiB
  Reduction:             46.5%
  Avg per full doc:      437 B
  Avg per summary:       234 B

## All docs (n=210600)
  Total full-doc JSON:   129.64 MiB
  Total summary JSON:    59.67 MiB
  Reduction:             54.0%
  Avg per full doc:      645 B
  Avg per summary:       297 B

## Summarise timing (1 runs over 210600 docs each)
  Per-run total:  min=23.99 ms  p50=23.99 ms  mean=23.99 ms  max=23.99 ms
  Per doc (mean): 0.11 µs
  Throughput:     8,778,240 docs/s
```
size reduction is significant, but at these sizes it's only relevant at tens of thousands of documents or more.
processing time is a complete non-issue.

shared-libs/cht-datasource/src/index.ts
```diff
@@ -88,6 +89,15 @@ export const getDatasource = (ctx: DataContext) => {
         chtPermissionsSettings?: Record<string, string[]>
       ) => ctx.bind(hasAnyPermission)(permissionsGroupList, userRoles, chtPermissionsSettings),
       contact: {
+        /**
+         * Returns summary records for the given contact UUIDs.
+         * @param uuids the UUIDs of the contacts to summarise
+         * @returns an array of contact summaries. UUIDs that do not identify an existing contact (or that the caller
+         * is not authorized to view) are silently omitted from the result.
```
jkuester: > (or that the caller is not authorized to view)

Minor, but unless I am missing something we do not have any _authorization_ logic in cht-datasource.  I guess maybe the idea is that when running against Pouch as an offline user, only the summaries of docs that are in the user's local DB will be returned?

IMHO it would be less confusing to just not mention it at all.

(Same for the report docs below.)

shared-libs/cht-datasource/src/libs/parameter-validators.ts
```diff
@@ -140,6 +140,13 @@ export const assertUuidQualifier: (qualifier: unknown) => asserts qualifier is U
   }
 };
 
+/** @internal */
+export const assertUuids: (uuids: unknown) => asserts uuids is string[] = (uuids: unknown) => {
```
jkuester: Minor, can you add unit tests for this function?

shared-libs/cht-datasource/src/index.ts
```diff
@@ -44,6 +44,7 @@ export { DataContext } from './libs/data-context';
 export { getLocalDataContext } from './local';
 export { getRemoteDataContext } from './remote';
 export { InvalidArgumentError, ResourceNotFoundError } from './libs/error';
+export { summarise } from './libs/summary';
```
jkuester: 🤔  I see that we need access to this from webapp (to avoid just duplicating the logic), but exporting this here is a pretty big depature from the other patterns we have in this library. It feels like the summary logic may still just belong in a separate shared-libs where it can be used by both cht-datasource and the `GetSummariesService`.

One thought I had was that actually it might make sense to put it  in `shared-libs/lineage`.  Pretty much all of the summary _logic_ is just around the lineage and the subject.  `shared-libs/lineage` also has a ton of code around this. It would be good to keep these close together (and maybe reuse logic if possible?).  Also cht-datasource and webapp already both directly depend on `shared-libs/lineage`

shared-libs/cht-datasource/test/index.spec.ts
```diff
@@ -336,6 +336,7 @@ describe('CHT Script API - getDatasource', () => {
           [
             'getByUuid',
             'getByUuidWithLineage',
+            'getSummaries',
```
jkuester: Minor, but can you add a test below (similar to the other functions) that just confirms this is wired up as expected?

(Same goes for the report function)

shared-libs/cht-datasource/src/contact.ts
```diff
@@ -41,6 +42,22 @@ export namespace v1 {
     readonly parent?: ContactWithLineage | NormalizedParent;
   }
 
+  /**
+   * A compact summary of a contact record used in lists and search results.
```
jkuester: I don't think we need to include "used in lists and search results". Since cht-datasource is a general purpose API, we don't want to assume how it is used.

shared-libs/cht-datasource/src/report.ts
```diff
@@ -31,6 +37,29 @@ export namespace v1 {
     readonly place?: Contact.v1.ContactWithLineage | NormalizedParent;
   }
 
+  /**
+   * A compact summary of a report record used in lists and search results.
```
jkuester: I don't think we need to include "used in lists and search results". Since cht-datasource is a general purpose API, we don't want to assume how it is used.

shared-libs/cht-datasource/src/contact.ts
```diff
@@ -78,6 +95,30 @@ export namespace v1 {
...
+   * @returns a function for retrieving an array of contact summaries
+   * @throws Error if a data context is not provided
+   */
+  export const getSummaries = (context: DataContext): typeof curriedFn => {
+    assertDataContext(context);
+    const fn = adapt(context, Local.Contact.v1.getSummaries, Remote.Contact.v1.getSummaries);
+
+    /**
+     * Returns summary records for the provided contact UUIDs. Any UUIDs that do not identify an existing contact
+     * (or that the caller is not authorized to view) are silently omitted from the result.
```
jkuester:
```suggestion
     * are silently omitted from the result.
```

shared-libs/cht-datasource/src/contact.ts
```diff
@@ -78,6 +95,30 @@ export namespace v1 {
...
+     * @param uuids the UUIDs of the contacts to summarise
+     * @returns an array of contact summaries
+     * @throws InvalidArgumentError if `uuids` is not an array of non-empty strings
+     */
+    const curriedFn = async (uuids: string[]): Promise<ContactSummary[]> => {
```
jkuester: I am trying to decide if we need to create a new `IdsQualifier` we can use here (with an `ids: string[]` property)....  🤔  The reason we have gone with the "Qualifier" objects as the parameter type for the other functions is because it lets us add more parameters (and combine them) without making non-passive changes to the function.  

I like the idea of being able to make non-passive changes to this function signature.  However, I cannot come up with any other parameters to provide here that I would want to return a `Promise<ContactSummary[]>` for.  If we wanted to support getting summaries by `ContactTypeQualifier` or `FreetextQualifier`,  I think we should be returning a `Promise<Page<ContactSummary>>`.  But, it does not make much sense to return a `Page` when we are given all the ids to retrieve...

I am interested to hear your thoughts on this.  Currently, I think my vote would be to NOT add a new `IdsQualifier`, but instead change the name of the outer function to be `getSummariesByIds` instead of `getSummaries`.  Then we keep the `string[]` parameter here since the get-all-summaries-for-these-ids is a very common workflow that should not need paging.  Ultimately, this leaves the opportunity for adding future `getSummariesPage` and `getSummaries` functions for doing the query-style paged lookup if we ever want/need to do that....

shared-libs/cht-datasource/src/local/contact.ts
```diff
@@ -101,6 +102,24 @@ export namespace v1 {
     };
   };
 
+  /** @internal */
+  export const getSummaries = ({ medicDb, settings }: LocalDataContext) => {
+    const getMedicDocsByIds = getDocsByIds(medicDb);
+    return async (uuids: string[]): Promise<Contact.v1.ContactSummary[]> => {
+      if (!uuids.length) {
+        return [];
+      }
```
jkuester: Currently the `assertUuids` function fails if the provided uuids are empty.  So, we _should_ never make it here when this is emtpy.  

However, we could move this logic up into the top-level contact.ts file so we return an empty arrray when no uuids are provided (and never call down here into remote/local).  We could need to update `assertUuids` to not fail for empty arrays.

I am fine going either way (allowing or banning empty uuids).  For the other qualifiers, we are typically very prescriptive in our validations (it helps keep things simple as you start to combine qualifiers), but for these get-for-ids functions, I think we can be more flexible since our parameter type is fixed.  (And I always like to just have pass-through logic for empty arrays when possible since it usually means you need less if-blocks in the consuming code...)

shared-libs/cht-datasource/src/local/contact.ts
```diff
@@ -101,6 +102,24 @@ export namespace v1 {
...
+      const docs = await getMedicDocsByIds(uuids);
+      const summaries: Contact.v1.ContactSummary[] = [];
+      for (const doc of docs) {
+        if (isContact(settings, doc)) {
+          summaries.push(summariseContact(doc));
+        }
+      }
+      return summaries;
```
jkuester:
```suggestion
      return docs
        .filter(doc => isContact(settings, doc))
        .map(summariseContact);
```

shared-libs/cht-datasource/src/local/report.ts
```diff
@@ -106,6 +107,24 @@ export namespace v1 {
...
+        return [];
+      }
+      const docs = await getMedicDocsByIds(uuids);
+      const summaries: Report.v1.ReportSummary[] = [];
+      for (const doc of docs) {
+        if (isReport(doc)) {
+          summaries.push(summariseReport(doc));
+        }
+      }
+      return summaries;
```
jkuester:
```suggestion
      return docs
        .filter(isReport)
        .map(summariseReport);
```

api/src/controllers/report.js
```diff
@@ -118,6 +119,53 @@ module.exports = {
...
+     * @openapi
+     * /api/v1/report/summary:
+     *   post:
+     *     summary: Get report summaries by UUIDs
+     *     operationId: v1ReportSummaryPost
+     *     description: >
+     *       Returns compact summary records for the reports identified by the provided UUIDs. UUIDs that do not
+     *       identify an existing report are silently omitted from the result.
+     *     tags: [Report]
+     *     x-since: 4.18.0
```
jkuester:
```suggestion
    x-since: 5.3.0
```

webapp/src/ts/services/get-summaries.service.ts
```diff
@@ -1,139 +1,32 @@
...
-      .query('medic/doc_summaries_by_id', { keys: ids })
-      .then(response => {
-        return response.rows.map(row => {
-          row.value._id = row.id;
-          return row.value;
-        });
-      });
+    this.getContactSummaries = this.chtDatasourceService.bind(Contact.v1.getSummaries);
+    this.getReportSummaries = this.chtDatasourceService.bind(Report.v1.getSummaries);
   }
```
jkuester, line 19:
```suggestion
  constructor(
    chtDatasourceService: CHTDatasourceService,
  ) {
    this.getContactSummaries = chtDatasourceService.bind(Contact.v1.getSummaries);
    this.getReportSummaries = chtDatasourceService.bind(Report.v1.getSummaries);
  }
```

Minor, but can keep the service local to the constructor since it is not used elsewhere.

webapp/src/ts/services/get-summaries.service.ts
```diff
@@ -1,139 +1,32 @@
...
-import { SessionService } from '@mm-services/session.service';
-import { DOC_TYPES } from '@medic/constants';
+import { CHTDatasourceService } from '@mm-services/cht-datasource.service';
 
 @Injectable({
   providedIn: 'root'
 })
 export class GetSummariesService {
+  private readonly getContactSummaries;
+  private readonly getReportSummaries;
```
jkuester:
```suggestion
  private readonly getContactSummaries: ReturnType<typeof Contact.v1.getSummaries>;
  private readonly getReportSummaries: ReturnType<typeof Report.v1.getSummaries>;
```

webapp/src/ts/services/get-summaries.service.ts
```diff
@@ -1,139 +1,32 @@
...
-      .get()
-      .allDocs({ keys: ids, include_docs: true });
-
-    return result?.rows
-      ?.map(row => this.summarise(row.doc))
-      .filter(summary => summary);
+    const [contactSummaries, reportSummaries] = await Promise.all([
+      this.getContactSummaries(ids),
+      this.getReportSummaries(ids),
+    ]);
```
jkuester: 😓  I was concerned about  this in the admin code, but I decided it probably was not worth a bunch of refactoring of the legacy code-base there.  Unfortunately, here I think we really should avoid making these double-calls.  The main concern here is for online users, this will cause all of the docs to be loaded from CouchDB to api _twice_.  (Both the Contact and Report flows will load the docs from the DB before filtering out the ones that are not a contact/type.) This is quite a bit of extra traffic for a flow that theoretically _should_ be more efficient....

I did some digging into the code that consumes this function and I think it would be feasilble to bifurcate it into `getContacts` and `getReports`.  The main pain-point looks like the `GetDataRecordsService.get` function which is also generic.  😓  Looking at where _that_ function is used, it looks like it could also be refactored to have `getReports` and `getContacts`.  It is annoying, but I am not sure what we have here is good for the long-term....

jkuester (changes requested):
Very nice!  I left a lot of minor suggestions and a few more serious ones... 👍

shared-libs/cht-datasource/src/contact.ts
witash: Looking at this again, I think it should use a `UuidQualifier` if for no reason other than consistency; qualifiers are the way to represent parameters in datasource, and I don't see any reason ids should be different.

shared-libs/cht-datasource/src/local/contact.ts
witash: now that its using a qualifier, would prefer to remove these guards.
It does mean that empty id arrays won't short circuit, but I don't think that matters much.

webapp/src/ts/services/get-summaries.service.ts
witash: moved all callers to using a separate reports or contacts function, in admin and webapp.
Mostly it was clear which one callers actually needed, and there wasn't anything that actually needed both.
it does use `contacts` as a default if not `reports` in some places.

fwiw, here's claude's audit
```
  Reports (getReports)

  webapp/message-contact.service.ts:54 — ids come from the medic-client/messages_by_contact_date view (row.value.id). That view emits message data_records
  (reports). ✅ Reports by construction.

  webapp/report-view-model-generator.service.ts:55 — getReports([model.doc._id]). This service exists to render a single report being viewed; model.doc is
  that report. ✅ Strongest possible guarantee — it's the doc the generator was handed.

  Contacts (getContacts)

  get-subject-summaries (admin :70 + webapp :78) — resolves report subjects to names. A report's subject is the person/place the report is about, i.e. always
  a contact. ✅ Domain invariant (this is the same reasoning we discussed earlier).

  hydrate-contact-names (admin :64 + webapp :64) — the service replaces contact-id references (e.g. a report's contact/parent fields) with names. The ids are
  contact references by the service's definition. ✅ Invariant in the service's purpose.

  admin/message-queue.js:103 — ids come from the medic-client/contacts_by_phone view (resolving SMS recipient phone numbers → row.id). Those are contact ids.
  ✅ By view construction.

  contacts.component.ts:251 + user-settings.service.ts:84 — getContacts(facilityIds / userFacilities). A user's facility is always a place, which is a
  contact. ✅ Structural invariant of user config.

  target-aggregates.service.ts (3 calls) — getHomePlace([facilityId]) and getContacts(facilityIds) operate on facilities (places); the third maps place =>
  place.contact, a place's primary contact (a person). ✅ All contacts.

  Type-discriminated (type === 'reports' ? getReports : getContacts)

  admin/search.js:63 + webapp/search.service.ts:175 — keyed off the existing type argument that already partitions the search domain into 'reports' vs
  'contacts'. ✅ For today's two values.

```

shared-libs/cht-datasource/src/index.ts
witash: so moved to shared-libs/lineage but that's not without tradeoffs.
Its not really possible to reuse any code (like getLineage and hydrate) because hydration mostly depends on DB, and the summary functions should stay purely functional.
so adding summaries is adding to lineage, does it actually belong there?
that also means the export is a little wonky becase lineage exports a factory instead of functions
still rather do that then create a new dedicated shared-lib, but maybe it should be in its own shared-lib?

shared-libs/cht-datasource/src/libs/parameter-validators.ts
witash: replaced by uuidsqualifer

shared-libs/cht-datasource/src/index.ts
witash: so actually because of the above, I don't think this belongs in lineage. moved to its own shared-lib.

shared-libs/cht-datasource/src/contact.ts
```diff
@@ -106,15 +102,15 @@ export namespace v1 {
...
+     * @param qualifier the UUIDs of the contacts to summarise
      * @returns an array of contact summaries
-     * @throws InvalidArgumentError if `uuids` is not an array of non-empty strings
+     * @throws InvalidArgumentError if the qualifier does not contain an array of non-empty UUID strings
      */
-    const curriedFn = async (uuids: string[]): Promise<ContactSummary[]> => {
-      assertUuids(uuids);
-      return fn(uuids);
+    const curriedFn = async (qualifier: UuidsQualifier): Promise<ContactSummary[]> => {
+      assertUuidsQualifier(qualifier);
```
jkuester: Can we short-circuit here if `qualifier.uuids` is empty and just return an empty array?  I would love to avoid any round-trips if we do not actually have any ids...

shared-libs/cht-datasource/src/contact.ts
jkuester: Okay, I am still hesitant to call this `getSummaries` when the return type of the current function is a promised array.  We don't have any other cht-datasource functions currently that return an array, but we do have functions with a similar naming pattern (e.g. `getUuids`) that return an AsyncGenerator.  

We could add `getSummariesPage` and then have this function return an AsyncGenerator. That is a bit weird when we only have the UuidsQualifier parameter, but it does set us up for various other types of summary queries in the future (which seams plausible)..... 🤔

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -56,6 +56,38 @@ export const isUuidQualifier = (identifier: unknown): identifier is UuidQualifie
   return isRecord(identifier) && hasField(identifier, { name: 'uuid', type: 'string' });
 };
 
+/**
+ * A qualifier that identifies entities by their UUIDs.
+ */
+export type UuidsQualifier = Readonly<{ uuids: string[] }>;
```
jkuester: 😓 So, awhile back Diana pointed out to me that automatically assigned Couch doc `_id` values _happen_ to be "UUID"s, but are not required to be (you can use any kind of unique value as the `_id` of the doc).   From the abstracted perspective of this cht-datasource api, there is no reason to include the cognitive overhead of a "universally unique identifier" (which has a very specific technical definition) vs just saying "identifier".  At the time, I introduced the `IdQualifier` to use instead of the `UuidQualifier`, but I have not had the time yet to go back and properly deprecate `UuidQualifier`. 😢 

I am okay merging with `UuidsQualifier` and I can also deprecate this at the same as the other. But, I would also be fine if you wanted to just switch this to `IdsQualifier`

jkuester (approved):
Okay, this is looking great!  I left a few more throughts inline regarding api naming, etc. I leave these to your discretion and I think this should be good to merge whenever you are happy with things. 👍

shared-libs/cht-datasource/src/qualifier.ts
witash: I agree with the preference to used id over uuid, but the convention in most of the codebase seems to be `uuid`. especially `getUuids` in datasource
if we're ok with the mismatch, I'll switch to `IdsQualifier`

shared-libs/cht-datasource/src/contact.ts
witash: For consistency with the other endpoints, changed to return an asyncGenerator.
But its not free; paging doesn't really make sense with IdsQualifer, and it adds boilerplate and fluff to callers.
I guess its worth it to have all the endpoints have similar return type.

shared-libs/cht-datasource/src/contact.ts
witash: removed the short circuiting in [455bb2b](https://github.com/medic/cht-core/pull/10831/commits/455bb2b4b490e21dfb260a5ee61d40c3286ebd68) since I didn't think it makes much difference to actually run the query with no ids, and conflicts a little bit with qualifiers in general; it works now that only `IdsQualifer` is allowed, but adding more qualifiers would mean it would need to be more general.

shared-libs/cht-datasource/src/qualifier.ts
jkuester: Yeah, there is no winning at this point.  :sweat:  IMHO, since we really want cht-datasource to be an _abstraction layer_ that does not leak underlying DB implementation details, we should go with IdsQualifier. (But we can either do that now or later...)

shared-libs/cht-datasource/src/contact.ts
jkuester: Yeah, it is the curse of the general-purpose API.  I think it is best to prioritize overall consistency over local optimization...

PR #10842, opened 2026-04-11
feat(#10508): serialize primitive arrays in android widget

### LINKED ISSUE #10508: Serialize lists of primitive values returned from External Apps as space-delimited strings
The current [External Android App Launcher functionality](https://docs.communityhealthtoolkit.org/building/forms/app/#android-app-launcher) will serialize lists as JSON arrays when returning data back into the form model.  Then, the cht-core webapp code will [insert the data](https://github.com/medic/cht-core/blob/544bd9a3f36da28cf68da1f2ab6271d6f626e330/webapp/src/js/enketo/widgets/android-app-launcher.js#L95) into repeats with the appropriate appearance.

This approach is really the only way to handle an array _of objects_ since you need the repeat with child fields to model the properties on each object.  However, it is a bit confusing when it comes to arrays of primitive values. In that case, there is no named property to put in the repeat to hold the array value.  This is why we end up having the `android-app-object-list` vs the  `android-app-value-list`...

Unfortunately, Enketo repeats are still pretty janky, particularly when you are using them in non-trivial ways.  In practice we have found lots of challenges supporting complex workflows involving things like nested repeats, repeats whose relevance changes, etc.  Anything we can do to avoid using repeats for complex operations in a form would be an improvement and I think we should be able to handle lists of _primitive values_ without needing a repeat or `android-app-value-list` appearance.

Enketo has great support for lists of primitive values (modeled as space-delimited strings).  These "node sets" can be produced by normal `select_multiple` questions where the value for the field is actually just a space-delimited string with all the selected values. So, [built-in functions](https://docs.getodk.org/form-operators-functions/#selected-at) like `selected-at` and `count-selected` can be used to parse any space-delimited string. 

When de-serializing a JSON array of primitive values  in the android-app-launcher widget, we could check if the associated form model field is a repeat with either the `android-app-object-list` or  `android-app-value-list`. If so, we keep the existing behavior (so the changes remain passive).  However, if the field is not a repeat, then we should de-serialize the JSON array as a space-delimited string stored directly in the named field.  Then all the down-stream form logic can process the value as a node set and operate accordingly.

Support for this functionality was originally added [in this cht-android branch](https://github.com/medic/cht-android/pull/398).  However, I think this would be a breaking change if we ship the changes there (since cht-android has a decide if it is going to return the field as a string or array without knowing if the field has the `android-app-value-list` appearance or not.  By making this change in cht-core, it should be passive for forms that are already using `android-app-value-list`, but will allow the new functionality for forms without that appearance.

**Additional context**

https://github.com/medic/cht-android/pull/398

See the forms here (https://github.com/medic/fake-health-pulse#using-fake-healthpulse-ai) for examples with and without repeats.

megha1807:
## Description

When the android-app-launcher widget receives a JSON array of primitivevalues (strings, numbers, booleans) and the target output field is NOTa repeat group, the array is now deserialized as a space-delimited stringstored directly in the named field.

This allows forms to use built-in Enketo XPath functions like`selected-at()` and `count-selected()` without needing a repeat group or`android-app-value-list` appearance.

Existing forms using `android-app-value-list` repeat appearance areunaffected (passive/backwards-compatible change).

Fixes #10508

## Code review checklist
- [x] Tested: Unit tests added covering new behavior and confirming old repeat behavior is unchanged
- [x] Backwards compatible: Existing forms using android-app-value-list are unaffected


megha1807:
Hi @jkuester -  the e2e failure is in send-message.wdio-spec.js and appears unrelated to my changes.
I do have one outstanding issue with SonarCloud flagging line 187 - it wants optional chaining (!valueArray?.length) but ESLint rejects it with Parsing error: Unexpected token . in this .js file.
Could you advise the preferred fix? Happy to update once I know the right approach. Thanks!

jkuester (approved):
LGTM!  I made one minor tweak so it supports arrays that are inside of other objects.  Now this is good to go!  

Sadly we don't have any automated tests for this, but I have confirmed manually that the functionality works as expected and does not impact the existing functionality for loading data within the repeats.

megha1807:
Thanks @jkuester  for the review and the improvement!

PR #10843, opened 2026-04-11
fix(#9165): render audio/video previews when editing a report

### LINKED ISSUE #9165: Audio/Video file input previews are not rendered when editing a report

When creating a new report with a form that accepts audio/video input, the selected audio/video can be previewed directly on the form page:

image

However, if you go back and edit the report, the previews are not available:

image

The `reports-add.component` [has support](https://github.com/medic/cht-core/blob/5c424565a557716ae7efdb8923d747c9701b1946/webapp/src/ts/modules/reports/reports-add.component.ts#L170) for rendering image previews of previously selected _images_ when editing a report, but there is no similar support for audio/video.  It would be nice to have the behavior consistent with all the file selection widgets.

megha1807:
## Description

When editing a report with audio or video file inputs, the previously selected files had no preview shown. Image previews already worked, but audio/video were skipped.

Extended `renderAttachmentPreviews` in `reports-add.component.ts` to handle `audio/*` and `video/*` accept types, appending `<audio controls>` and `<video controls>` elements the same way images are handled.

Fixes #9165


jkuester (changes requested):
Okay, so it like it is kind of working, but not actually loading the media.  Here is how things look currently with `master` when you edit a form with uploaded media (the image loads as expected, but the audio and video files do not have previews:

Image

With your branch, I see this (the widgets appear, but the actual audio/video does not load):

Image

Both of these are what it looks like to _edit_ an existing report that has been submitted with file uploads.  The goal here is to have the _edit_ page look the same way it does when initially _creating_ a report.  When I create a new report and initially upload the various files, the audio/video previews are properly functioning on the page:  

Image

---

You can use this form to test with.

[file-upload.xlsx](https://github.com/user-attachments/files/26907459/file-upload.xlsx)

megha1807:
Hi @jkuester, I've updated the fix to address the issue where audio/video content was not loading in the previews.

The root cause was that the data URI was missing the MIME type. The `<audio>` and `<video>` elements require a properly formatted data URI (e.g. `data:audio/mpeg;base64,...`) to load media correctly, unlike images which are more lenient.

Changes made:
- Added `mimeType` parameter to `getPreviewElement()` using `(attachmentBlob as Blob).type`
- Constructed a proper data URI with `data:${mimeType};base64,${base64}` for all media types
- Updated all tests to reflect the correct data URI format

Could you please re-review and test with the file-upload form? Thank you!

Issue #9252, opened 2024-07-10
Annotate image not properly loaded by form draw widget


jkuester:
**Describe the bug**
The draw/signature form functionality added in https://github.com/medic/cht-core/issues/8308 included support for "annotating" an image. Basically you upload an image into the form and then you can draw/annotate the image and save the updated file as a report attachment.

Unfortunately, if the annotation question is on the first page of the form, when the image is initially loaded, it actually shows up in the canvas zoomed too far out (instead of being properly fitted within the canvas:

| Expected | Actual |
| --- | --- |
| ![image](https://github.com/medic/cht-core/assets/8441903/95024f5c-6cd5-401f-8b68-bd7d34915ba2)  |  ![image](https://github.com/medic/cht-core/assets/8441903/97cb8318-81ab-4035-b4a0-ff4055d32326) |

**To Reproduce**

- Use the [draw_widget form](https://github.com/medic/cht-core/blob/master/tests/integration/cht-form/default/forms/draw-widget.xlsx) and upload an image into the "Annotate image widget" question.  
- See that the image is zoomed out.
- Use the reload button to clear the image
- Upload the image into the question again
- See that it is zoomed as expected

**Expected behavior**
The image should be loaded into the canvas at the right size the first time it is loaded.

**Additional context**
After doing some debugging, it appears to be a problem with how we only display the loading indicator while initializing the form.  This means that the `canvas.offsetWidth` property is `0` until we finish initializing the for (and its widgets).  Then we hide the loading indicator and display the form.  

This is a problem because [when the draw widget is enabled](https://github.com/medic/cht-core/blob/e6d184946affc62773d569168216a5b913f38a30/webapp/src/js/enketo/widgets/draw.js#L648) it tries to initialize the size of the canvas [based on the `offsetWidth/height`](https://github.com/medic/cht-core/blob/e6d184946affc62773d569168216a5b913f38a30/webapp/src/js/enketo/widgets/draw.js#L600). This, of course, does not behave as expected since the `offsetWidth` is still `0`.  The result is that the various downstream calculations [in `fromObjectURL`](https://github.com/medic/cht-core/blob/e6d184946affc62773d569168216a5b913f38a30/webapp/src/js/enketo/widgets/draw.js#L40) do not work properly and the result is a zoomed out image.  

[...]

PR #10844, opened 2026-04-12
fix(enketo): fix geopoint map rendering on first page of form

### LINKED ISSUE #9268: Geopoint Widget map rendering issue when on first page of the form
**Describe the bug**
As noted in [this forum thread](https://forum.communityhealthtoolkit.org/t/conditional-relevant-geopoint-omits-calculate-value/3672/7), there appears to be a rendering problem with the geopoint widget when it is located on the first page of a form.  Basically, the rendered part of the map is off-center and most of what is visible is grey-ed out.

| Expected | Actual |
| --- | --- |
| image | image |

These rendering issues can be fixed when the window is resized and do not happen at all if the widget is located on subsequent form pages.  

**To Reproduce**

1. Create a form with a field that has `type: geopoint` and `appearance: maps` on the first page
2. Open the form and see the map rendering issue.

**Expected behavior**
The map should render properly regardless of which form page it is located on.

**Environment**
I have recreated this behavior on `4.3` and `4.9` code.  

**Additional context**
I have a theory that this behavior is related to what is causing https://github.com/medic/cht-core/issues/9252.  For that issue, the root cause seems to be that Enketo is initializing the widget while the loading spinner is still displayed (and the form is hidden).  This causes weird things to happen for the canvas calculations since the canvas size is wrong (offsetWidth/Height is `0`). Perhaps this is related.

megha1807:
# Description

When a geopoint widget with `appearance: maps` is on the first page of a form, it initializes while the loading spinner is displayed and the form is hidden. At that point `offsetWidth`/`offsetHeight` = 0, causing Leaflet to render the map greyed-out and off-center.

The existing `pageflip` handler only fixes the issue when navigating to subsequent pages. This fix adds a `ResizeObserver` to detect when the map container becomes visible and calls `invalidateSize()` once it has real dimensions. The observer disconnects immediately after to prevent memory leaks.

fixes #9268

# Code review checklist
- [x] Tested: Manual testing confirmed map renders correctly on page 1


jkuester (changes requested):
Okay, I tested this out and it does solve the issue.  However, I have spent years trying to get us away from having to patch `enketo-core` for all kinds of things... 😓   We still have a few more patches here that are mission-critical, but we have solved most of the rest of the things.  

Before adding another patch here, can we explore some alternative approaches first?  It does seem like this confirms my theory that the rendering issue is because when the geopoint widget is inited, it is loaded into an element that does not have a size (since the page is still displaying the loading indicator).  The code for doing this is in places like [this component](https://github.com/medic/cht-core/blob/4382617651cd92a1c244557f20f397301ccdfb48/webapp/src/ts/modules/reports/reports-add.component.html#L12) where we have `[hidden]="loadingContent` wrapping the `EnketoComponent` element.  I am thinking that when `[hidden]` is true for the Angular element, then it means that the element has no size.  

What I am curious about is if we can find an alternative for `[hidden]` that might let load the element with the proper size, but have it be all the way at the back or something so that other things are rendered over top of it and the form content is not visible until the loading is actually finished.  Obviously I am not familiar with what all is possible here or how janky that might get, but it would be nice to give it a try (especially since it seems like we have this same issue [for other Enketo widgets](https://github.com/medic/cht-core/issues/9252)....). I would love to avoid having to patch them all!

megha1807:
Hi @jkuester, sorry for the delayed update  took some time to explore the right approach!
Instead of patching enketo-core with a ResizeObserver, I've replaced [hidden] with a .content-loading CSS class in inbox.less that uses visibility: hidden + position: absolute + z-index: -1. This keeps the form element rendered with real dimensions while visually hidden, so Leaflet (and other Enketo widgets) can initialize correctly.
The fix is applied to all 4 components that use [hidden]="loadingContent":

reports-add.component.html
contacts-edit.component.html
tasks-content.component.html
training-cards-form.component.html

The loading spinner still works correctly as it's controlled by *ngIf independently. The enketo-core patch has been removed entirely.
This should also address the root cause for other widget rendering issues like #9252 (draw widget), since they share the same [hidden] wrapper pattern.
Also fixed the patch file - it had picked up extra old mode/new mode file permission diffs when patch-package ran locally. Restored it to only the 4 original CHT-specific patches.

PR #10845, opened 2026-04-13
feat(#9893): add background colors for form questions

### LINKED ISSUE #9893: Allow setting background color for form questions

Originally proposed [on the forum](https://forum.communityhealthtoolkit.org/t/custom-css-for-custom-appearance/4838/7).

In our Enekto forms. we already have a (poorly documented) feature that lets you set specific background colors for `note`s that are within a `summary` group by setting a header _appearance_ (e.g. `h2`) and a color _appearance_ (e.g. `red`).  This feature is intended to support formatting the "summary" page at the end of the form where the user can review the details of the data they have captured before submitting it.  The header appearances affect the size/alignment of the text and the colors allow for setting the background for that header.

Unfortunately, this functionality is not flexible enough to support all the desired use cases. Specifically, there is no way to set the color without also setting the header and setting the header really only "works" for `note` fields. If you set the header appearances on other types of questions, it can cause undesired layout issues.  This ticket proposes expanding the color functionality in the css to allow for setting a background color on all types of questions

Enketo already automatically supports setting the values from the `appearance` column in the xlsx as classes on the HTML element for that form question.  This is how the [`summary` stuff](https://github.com/medic/cht-core/blob/a2682caf70f2bf0a5e2ca870ad7adeb99f2a57df/webapp/src/css/enketo/medic.less#L306) works currently.  We can target these appearance classes in our css and provide styling that can be toggled on by setting the appearance. 

My proposal is that we add a new set of supported classes directly under `.enketo` that will allow for targeting specific background colors for any question:

- `.or-appearance-background-red`
- `.or-appearance-background-green`
- `.or-appearance-background-blue`
- `.or-appearance-background-yellow`
- `.or-appearance-background-lime`

To use these, you would just set something like `background-blue` in the _appearance_ column in the xlsx form config.  

For backwards compatibility, we should retain the existing `.or-appearance-red/blue/...` classes, but I think they should be considered deprecated in favor of the more general-purpose `.or-appearance-background-red/blue/...` classes. We could even add a [form validation warning](https://github.com/medic/cht-conf/tree/main/src/lib/validation/form) to cht-conf to prompt app devs to update their form configuration to the new appearances.

I do not think there is much we can do here in the way of having automated tests for these changes, but we will want to thoroughly manually test these background colors (particularly with the [enketo_widgets_test.xlsx form](https://github.com/medic/cht-core/blob/master/tests/e2e/default/enketo/forms/enketo_widgets_test.xlsx) since that will let us catch any unintended impacts to all the various enketo widgets we support).

One discussed alternative is to simply allow for injecting a custom stylesheet as part of the branding assets configurable for the CHT.  While I think that could be used for to achieve much more dramatic customization, I remain concerned about the long-term maintainability of that approach.  At the very least, I would want to have some kind of set list of things that are officially supported for styling in this way. (AKA some kind of theming structure that would allow for deeper configuration of the CHT styles, but could still be guaranteed not to break/change unexpectedly. One big draw-back of that kind of approach is that it is pretty heavy-handed/technical for achieving something simple like coloring form questions. Thinking as an app developer, if I want to set a background color on a form question, it is much more convenient/straightforward to just be able to set an `appearance` value right in my form xlsx instead of trying to figure out my own custom css....

megha1807:
## Summary
Adds new general-purpose background color appearance classes for Enketo form questions.

## Changes
- Added 5 new CSS classes in `medic.less` under `.enketo`:
  - `.or-appearance-background-red`
  - `.or-appearance-background-green`
  - `.or-appearance-background-blue`
  - `.or-appearance-background-yellow`
  - `.or-appearance-background-lime`

## How to use
In the XLS form config, set the appearance column to any of:
`background-red`, `background-green`, `background-blue`, `background-yellow`, `background-lime`

## Notes
- Unlike the existing color appearances, these work on **all question types**, not just notes inside summary groups
- Existing `.or-appearance-red/blue/...` classes are retained for backwards compatibility
- Colors match the existing summary color palette for consistency

Closes #9893


jkuester (changes requested):
Nice!  I confirmed the background colors are applied as expected.  However, I think we are going to have to have some extra styling at least to change the text color to white.  Here is how things look now with some different form widgets:

Image

I am not sure about the text boxes, etc.  🤔 We could probably leave them white, but it might look better to have them be a shaded version of the same color...  My main concern with that is that the different question types might need different styling. We could try to test most of them and see. 🤷 

@delcroip do you have any thoughts here?

megha1807:
Thanks for the review and screenshots, @jkuester! Good catch on the contrast.
My proposed fix:

Add color: white (or a dark shade for lighter colors like yellow/lime) to the .or-appearance-background-* classes
For text inputs/selects inside those groups, apply a slightly lighter/tinted version of the background so they stay visually cohesive but readable

I'll test across note fields, text inputs, selects, and radio/checkbox groups before pushing. Does that approach sound reasonable, or would you prefer I wait for @delcroip's input first?

jkuester:
That proposal sounds good. I think white text should be fine.  Here is the form I was using to test with: [appearances.xlsx](https://github.com/user-attachments/files/27376169/appearances.xlsx)


We should be good to keep iterating on this and @delcroip can weigh in when/if he gets the chance. :+1:

megha1807:
Thanks, @jkuester! Will get started on the fixes and push an update soon.

megha1807:
@jkuester Fixed the contrast ratios for green (#75b2b2 → #2e7d7d) and blue (#6b9acd → #2c5f9e) background colors to meet WCAG accessibility requirements. Ready for re-review!

megha1807:
@jkuester The contrast ratios for green and blue have been fixed. The remaining CI failures appear to be pre-existing flaky integration tests unrelated to this CSS change. Ready for re-review!

jkuester:
Okay, so your color changes are reflected on some of the fonts and inputs, but not all of them:

image

The more I think about this the more concerned I am with all the corner cases here where we have to think about all the different aspects of each type of Enketo widget (e.g. currently the white text in the drop-down for the minimal select widget makes the choices unreadable:

image


I really want to avoid needing a bunch of custom styling to deal with all of this since it will be really hard to keep it up-to-date. :grimacing: 

---

@megha1807 what do you think about changing things up a bit and instead of matching the _existing `summary` colors_, we go for more subtle faded versions of the colors that will still work with a black font and the light-themed inputs?  

For example, instead of using `#e00900` for "red", we could use `#f2dede` (which is the background color we use here):

image

Then we should not need any custom styling for the inputs or the font since the existing default styling for all these elements should be fine.  At the same time, I think the faded colors will still do a great job of addressing the core issue which is the ability to add extra emphasis to certain parts of the form.

For the  colors you pick, lets add them to the [variables.less](https://github.com/medic/cht-core/blob/e33cd8738be9bbe1e3c34566fe07d7bf9fa9da6f/webapp/src/css/variables.less) so they are easy to reuse.  I think we can stick with the red/green/blue/yellow/lime, but just the faded version.

---

Another thing to consider here is if it is feasible to support these colors _at the group level_ (e.g. for `section` blocks)?  If that is going to add a bunch of complexity, then we should avoid it, but if it is easy, then I think we can include it.

---

Finally, I think it would make sense to include a new integration test in tests/integration/cht-form/default to hopefully keep us from regressing on this stuff.  Maybe just use the `appearance.xlsx` form that I linked above and add a test that asserts the main css attributes when the form loads.

jkuester (changes requested):
^

megha1807:
Thanks for the detailed feedback, @jkuester! The faded color approach makes much more sense - avoids all the widget-specific overrides and keeps things maintainable. I'll switch to pastel versions and add them to variables.less. I'll also check if group-level support is straightforward to include. And I'll add an integration test using the appearances.xlsx form. Will push an update soon!

megha1807:
Hi @jkuester,

I’ve updated the PR based on the latest feedback:

* switched to faded/pastel background colors
* moved reusable colors into `variables.less`
* removed the widget-specific contrast overrides
* added integration test coverage using the `appearances.xlsx` form
* restored the affected variables structure and cleaned up duplicate styles

Could you please let me know if there are any additional changes required, or if the PR is now in a good state for merge/re-review?

Thanks!

PR #10849, opened 2026-04-13
fix(#4692): use specific error codes instead of generic invalid_report

### LINKED ISSUE #4692: "invalid_report" is not a useful error message for analytics
See: https://github.com/medic/cht-core/blob/master/shared-libs/transitions/src/lib/utils.js#L43

This code is called in random places, and results in a generic error code with a specific error message.

This sort of maybe makes sense in some contexts, but it's hecka annoying for analytics, because these errors can not be programmatically analysed (eg display the top N submission errors for a partner).

This is not a technical limitation either, because we *know* the specific problem, because we translate it into the `message` field.

We should change this to use the specific translation code as the error code (maybe with "error" on the front or something)

santhosh-7777:
# Description

This PR replaces the generic `invalid_report` error code with more specific and meaningful error codes derived from the actual error.

## Changes

- If the error is a string → use it as both `code` and `message`
- If the error object has no `code` → fallback to `error.message`
- Retain `invalid_report` only as a final fallback
- Updated unit tests to reflect the new behavior

## Why

Previously, all errors were assigned the same `invalid_report` code, making it difficult to analyze errors in analytics and debug issues.

This change improves error visibility by providing more specific and actionable error codes.

## Issue

Fixes #4692

# Code review checklist

- [x] Tested: Unit tests updated and passing
- [x] Backwards compatible: Uses fallback to `invalid_report` when needed
- [ ] UI/UX backwards compatible
- [ ] Documented
- [ ] Internationalised
- [ ] AI disclosure

# License


shared-libs/transitions/src/lib/utils.js
```diff
@@ -40,18 +40,17 @@ const addError = (doc, error) => {
...
   if (_.isString(error)) {
-    error = { code: 'invalid_report', message: error };
+    // Use the string as both code and message so analytics can identify it
+    error = { code: error, message: error };
   } else if (_.isObject(error)) {
     if (!error.code) {
-      // set error code if missing
-      error.code = 'invalid_report';
+      // Fall back to message as code, or generic if message also missing
+      error.code = error.message || 'invalid_report';
```
dianabarsan, line 48: @binokaryg could you please validate that this is a desirable change? Thank you!

binokaryg: When I think about an `error code`, I see it as a short identifier.
Since `error.message` can be of any length, it could be difficult to understand.

It should ideally be handled within the callers, ~but if that's not possible (e.g. source is external), then I think it's a fair compromise~.

binokaryg: I asked Claude to find the call sites where `code` could be missing and here's the summary:

---
1. **[muting.js:251](https://github.com/medic/cht-core/blob/master/shared-libs/transitions/src/transitions/muting.js#L251)** — `msg` (translated config string) → `muting_error`
2. **[muting.js:253](https://github.com/medic/cht-core/blob/master/shared-libs/transitions/src/transitions/muting.js#L253)** — hardcoded fallback sentence → `muting_error`
3. **[update_notifications.js:51](https://github.com/medic/cht-core/blob/master/shared-libs/transitions/src/transitions/update_notifications.js#L51)** — `msg` (translated config string) → `notification_error`
4. **[update_notifications.js:53](https://github.com/medic/cht-core/blob/master/shared-libs/transitions/src/transitions/update_notifications.js#L53)** — hardcoded fallback sentence → `notification_error`
5. **[multi_report_alerts.js:75](https://github.com/medic/cht-core/blob/master/shared-libs/transitions/src/transitions/multi_report_alerts.js#L75)** — `phone.error` (object with `.error` string) → `phone_number_error`

## External sources?

None. All 5 call sites are internal. `phone.error` is constructed right there in [`multi_report_alerts.js:152-171`](https://github.com/medic/cht-core/blob/master/shared-libs/transitions/src/transitions/multi_report_alerts.js#L152-L171) — it's an object like `{ error: "multi_report_alerts : ..." }`. Note it has an `.error` property, not `.message`, so `messages.addError` would stringify it via the fallback path. That one's already a bit broken regardless of this PR.

## Bottom line

Small, self-contained change. No external dependencies. We just need to pass `{ code, message }` objects at these call sites instead of bare strings.

---

I think this approach aligns better with the requirement.

santhosh-7777:
Hi @binokaryg, thank you for the detailed analysis! I understand the approach now , instead of using error.message as the code in utils.js, I should update the call sites to pass { code, message } objects directly. I'll make those changes now.

santhosh-7777:
Hi @binokaryg, I've updated all 5 call sites to pass { code, message } objects directly muting_error for muting.js, notification_error for update_notifications.js, phone_number_error for multi_report_alerts.js and reverted utils.js back to use invalid_report as fallback. Ready for re-review!

santhosh-7777:
Hi @binokaryg ,
I’ve updated the call sites to pass `{ code, message }` as suggested. However, some tests are now failing because they still expect the old generic codes (like `invalid_report` / `notification_error`), while the implementation now uses more specific values.
Should I update the tests to reflect this new behavior, or adjust the implementation to keep the existing expectations?
Thanks!

binokaryg:
> Should I update the tests to reflect this new behavior, or adjust the implementation to keep the existing expectations?

As long as we are making the tests align with the expectations (specific error codes), we can update them.

dianabarsan (changes requested):
It seems that some unit tests are failing. Can you please fix them?

santhosh-7777:
Hi, I updated the expectation in accept_patient_reports.js line 130 to { message: 'some_error', code: 'some_error' } and pushed the latest commits (5943372, 7e9b742). However, the Build and test / Compile the app check is still failing with actual: "invalid_report" vs expected: "some_error".
I verified locally that the updated expectation is present in the file. Could you suggest what I should check next in this validation flow?

binokaryg:
> Hi, I updated the expectation in accept_patient_reports.js line 130 to { message: 'some_error', code: 'some_error' } and pushed the latest commits ([5943372](https://github.com/medic/cht-core/commit/5943372594adeb04b120e464c1fb2d1f0dda2f79), [7e9b742](https://github.com/medic/cht-core/commit/7e9b7424fa3315d8b8c271176a23892bc0f4305b)). However, the Build and test / Compile the app check is still failing with actual: "invalid_report" vs expected: "some_error". I verified locally that the updated expectation is present in the file. Could you suggest what I should check next in this validation flow?

It looks like your error code is being discarded [here](https://github.com/santhosh-7777/cht-core/blob/7e9b7424fa3315d8b8c271176a23892bc0f4305b/shared-libs/transitions/src/lib/messages.js#L146:151) because `code` is missing from the error object.

```js
    if (_.isString(error)) {
      error = {
        //code: error,
        message: error,
      };
```

santhosh-7777:
Hi @binokaryg,
I've made the requested change and pushed the fix in commit `872c2f7`.
The two latest workflow failures appear to be related to CI infrastructure rather than the code changes in this PR.
Please take another look when you have a chance. Thanks!

binokaryg:
Thanks @santhosh-7777 for adding it.
While we wait for the tests to pass, could you please also add a small unit test for guarding this change?

santhosh-7777:
Hii @binokaryg,
I have added the requested unit test to cover the string error case and pushed the latest changes. Please take another look when you have a chance.
Thanks you!

binokaryg:
I have rerurn the tests multiple times without success. Let's wait for this to get merged before trying again:
- https://github.com/medic/cht-core/pull/11183

santhosh-7777:
@binokaryg yes i also observed that only that tests are failing, lets wait till that issue solved.
thank you!

binokaryg (approved):
Thanks @santhosh-7777!

The fix looks good with the guard test. Tests passing. All good.

binokaryg:
@dianabarsan this is now waiting for you:
> Merging is blocked
1 review requesting changes and 1 approving review by reviewers with write access.

shared-libs/transitions/src/lib/utils.js
```diff
@@ -67,16 +66,16 @@ const addError = (doc, error) => {
...
 
-const getReportsWithSameParentAndForm = (options={}) => {
+const getReportsWithSameParentAndForm = (options = {}) => {
   const formName = options.formName;
   if (!formName) {
     return Promise.reject('Missing required argument `formName` for match query.');
   }
 
   const parentId = options.doc &&
-                   options.doc.contact &&
```
dianabarsan, line 77: All these whitespace changes seem unrelated, can you please revert them?

shared-libs/transitions/src/transitions/multi_report_alerts.js
```diff
@@ -72,7 +72,7 @@ const generateMessages = (
 
     if (phone.error) {
       logger.error('%o', phone.error);
-      messages.addError(latestReport, phone.error);
+      messages.addError(latestReport, { code: 'phone_number_error', message: phone.error });
```
dianabarsan, line 75: These error code changes in `multi_report_alerts`, `muting` and `update_notifications` have no test coverage. 

Please add both unit test and e2e test coverage for these error codes.

shared-libs/transitions/test/unit/utils.js
```diff
@@ -230,7 +230,7 @@ describe('utils', () => {
       const doc = {};
       utils.addError(doc, 'some error message');
       assert.equal(doc.errors.length, 1);
-      assert.equal(doc.errors[0].code, 'invalid_report');
+      assert.equal(doc.errors[0].code, 'some error message');
```
dianabarsan, line 233: It seems that `invalid_report` remains the default error code when missing, but there is no unit test coverage for this behavior.

dianabarsan (changes requested):
Thanks for the updates @santhosh-7777 

I feel like all these changes are made in insolation and I have trouble tracking exactly how the code will behave now and in which circumstances reports have default codes or specific codes. 

Can you please add unit and e2e test coverage that will show how your code updates end up in actual reports? Thanks!

santhosh-7777:
@dianabarsan i have made the requested changes, could u please have another look on this?

shared-libs/transitions/src/lib/utils.js
dianabarsan: it seems like the whitespace changes were not reverted.

shared-libs/transitions/test/unit/transitions/muting.js
```diff
@@ -1026,5 +1026,40 @@ describe('Muting transition', () => {
...
+              ],
+            },
+          ],
+        }
+      });
+
+      return transition.onMatch({ doc }).then(result => {
+        chai.expect(result).to.equal(true);
+        chai.expect(doc.errors.length).to.equal(1);
+        chai.expect(doc.errors[0].code).to.equal('invalid_patient_id');
```
dianabarsan: I believe that we were trying to get the `muting_error` code here, not `invalid_patient_id`.

shared-libs/transitions/src/transitions/muting.js
```diff
@@ -248,9 +248,11 @@ module.exports = {
 
     const msg = messages.getMessage(evConf, locale);
     if (msg) {
```
dianabarsan: instead of duplicating the code here, please craft the error message in the if clause but only call `messages.addError` once. Applies to the other places as well.

tests/e2e/default/transitions/client-side-muting.wdio-spec.js
```diff
@@ -917,6 +917,33 @@ describe.skip('Muting', () => {
       expect(serverReport.tasks[0].messages[0].message).to.equal(serverReport.errors[0].message);
     });
 
+    it('should save muting_error code on report when muting fails', async () => {
```
dianabarsan: these tests are skipped. Please add the e2e test coverage in the corresponding sentinel e2e tests here: https://github.com/medic/cht-core/tree/master/tests/integration/sentinel/transitions

dianabarsan (changes requested):

santhosh-7777:
image
already fixed

dianabarsan:
> image already fixed

not fixed

image

dianabarsan (changes requested):
Looks like the test are failing.

santhosh-7777:
@dianabarsan Added sentinel integration test coverage in tests/integration/sentinel/transitions/muting.spec.js ("should add muting_error code on errors when muting fails"), update-notifications.spec.js ("should add notification_error code on errors when notification fails"), and multi-report-alerts.spec.js ("should add phone_number_error code on errors when phone recipient lookup fails"). Also fixed the invalid_patient_id vs muting_error mismatch in the unit test, and removed the redundant skipped e2e test since real coverage now exists. Ready for another look whenever you have time , thanks for your patience on this one!

shared-libs/transitions/src/lib/utils.js
dianabarsan: It seems these whitespace changes are still around. Please remove them @santhosh-7777

dianabarsan (approved):
Code looks great! Thanks so much for you contribution. 
To accelerate the merge, I pushed a very small change to address the whitespace nits I kept nagging about and were not fixed.

santhosh-7777:
Thanks for the approval and for cleaning up the whitespace @dianabarsan! Appreciate all the feedback across this one. happy to help with anything else if needed before it's merged.

Issue #10625, opened 2026-02-10
Use db view/index pagination for freetext search


jkuester:
**Describe the performance issue**
When doing a normal freetext search from either the People or Reports tab, we should only be hitting a single DB index/view.  I think this means we _should_ be able to "page" the query via the DB paging functionality and not have to pull back the entire results set before manually paging out the results we want.

**Describe the improvement you'd like**

In `shared-libs/search`, we [are checking for the `ordered` property](https://github.com/medic/cht-core/blob/d38cb4a56dd9110662186aec321cf78986c055fc/shared-libs/search/src/search.js#L104) to be present before using db pagination (even if there is only 1 request).  

The freetext requests do not set `ordered: true` (and never have).  This is despite the fact that the freetext views/indexes do have an ordering and so should be sorted in a predictable/useful way when they are returned.  I _think_ this means we should be able to update the freetext requests to set `ordered: true` so that when we are just doing a simple search with one of those requests, it does not have to pull back the whole results set.

**Measurements**
As was noted in the investigations around https://github.com/medic/cht-core/issues/10621, we "page" the search results displayed in the app, only showing the user `25` results and then triggering another "search" for the next page when the user scrolls down.  This new search will again pull _all_ the search hits back before showing the user the next `25` results.  We should be able to just search for 25 results at a time instead of needing to manually page through all results for every page.

**Additional context**

It was almost 10 years ago when [Gareth added the `ordered` logic](https://github.com/medic/cht-core/commit/2d2c0bc04d12ef62ee348c78e57aa82a8ffb64df) and even then it was not applied to the freetext indexes.  Did he know something then, that we are missing now?  Or, have things change in the meantime such that db paging for these searches is now possible?    🤷 https://github.com/medic/cht-core/issues/2371

>  I think it can only be applied for the default views (`contacts_by_name` and `reports_by_date`) but they're used most often so would still be a big win.

raghuwanshi313:
I'm also interested in the performance issue #10621 about 
freetext search paging. I understand the problem and would 
like to work on it if it's available!

raghuwanshi313:
To test the pagination fix properly, I created a small NodeJS script that injects 1,000 dummy users into the local CouchDB. This makes it easy to verify that the freetext search only pulls back 25 records per page.

Would you like me to include this dummy data generation script in my PR for future testing, or should I just keep it local for my own use?

jkuester:
[...]

PR #10850, opened 2026-04-13
feat(#10625): add paginated freetext query support

raghuwanshi313:
# Description

This PR introduces `queryFreetextPaginated` to `shared-libs/search/src/freetext-query.js`, which utilizes the CouchDB/Nouveau `datasource` native pagination capabilities (`skip` & `limit`) instead of pulling all records into memory at once. It also updates `search.js` to intercept freetext requests and route them through this new paginated generator, preventing memory crashes when searching massive datasets.

Partially addresses #10625

# Code review checklist
- [x] Tested: Unit tests run locally
- [x] AI disclosure: Assisted by GitHub Copilot / Gemini.

# License

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12268,6 +12268,49 @@
         "proxy-from-env": "^1.1.0"
       }
     },
+    "node_modules/appium-uiautomator2-driver/node_modules/@appium/docutils": {
+      "version": "2.2.1",
+      "resolved": "https://registry.npmjs.org/@appium/docutils/-/docutils-2.2.1.tgz",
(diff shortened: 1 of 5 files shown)


jkuester (changes requested):
I feel like we are basically on the right track here, but unfortunately when I tested this code change, the freetext searching functionality in the app was totally broken. 😞 Just printed errors in the browser console and did not return any search results.

I tried tweaking the code a bit and still was not able to get it to work. So, I think you are going to need to iterate on this a bit and make sure the basic searching works on the People and Reports tabs.

PR #10856, opened 2026-04-14
fix(#9993): resolve IDB race condition in telemetry service

### LINKED ISSUE #9993: Telemetry e2e test creates feedback doc about IDB crash
**Describe the bug**
There seems to be a race condition in our telemetry code, which can trigger an IDB crash. 

**To Reproduce**
An example of an e2e test run where a feedback doc was generated: https://github.com/medic/cht-core/actions/runs/15323227745/job/43112302806

**Expected behavior**
No IDB crash should happen. 

**Logs**
Excerpt of the feedback doc: 
```
{
    "_id": "feedback-2025-05-29T12:21:26.603Z-37a9efc8-31b3-47b1-8f1e-1441149cd2c0",
    "_rev": "1-3892173a18d5bd706a3a2d26f383c8e2",
    "arguments": [
      "\"Unhandled Promise rejection:\"",
      "\"Failed to read the 'result' property from 'IDBRequest': The request has not finished.\"",
      "\"; Zone:\"",
      "\"<root>\"",
      "\"; Task:\"",
      "\"Promise.then\"",
      "\"; Value:\"",
      "\"InvalidStateError: Failed to read the 'result' property from 'IDBRequest': The request has not finished.\\n    at te._destroy (https://localhost/main.js:1:1017153)\\n    at tt (https://localhost/main.js:1:980982)\\n    at https://localhost/main.js:1:981201\\n    at O.invoke (https://localhost/polyfills.js:1:23679)\\n    at te.run (https://localhost/polyfills.js:1:18898)\\n    at https://localhost/polyfills.js:1:30102\\n    at O.invokeTask (https://localhost/polyfills.js:1:24298)\\n    at te.runTask (https://localhost/polyfills.js:1:19556)\\n    at ue (https://localhost/polyfills.js:1:26335)\\n    at t.invokeTask [as invoke] (https://localhost/polyfills.js:1:25385)\"",
      "\"InvalidStateError: Failed to read the 'result' property from 'IDBRequest': The request has not finished.\\n    at te._destroy (https://localhost/main.js:1:1017153)\\n    at tt (https://localhost/main.js:1:980982)\\n    at https://localhost/main.js:1:981201\\n    at O.invoke (https://localhost/polyfills.js:1:23679)\\n    at te.run (https://localhost/polyfills.js:1:18898)\\n    at https://localhost/polyfills.js:1:30102\\n    at O.invokeTask (https://localhost/polyfills.js:1:24298)\\n    at te.runTask (https://localhost/polyfills.js:1:19556)\\n    at ue (https://localhost/polyfills.js:1:26335)\\n    at t.invokeTask [as invoke] (https://localhost/polyfills.js:1:25385)\""
    ],
    "info": {
      "message": "Failed to read the 'result' property from 'IDBRequest': The request has not finished.",
      "stack": "InvalidStateError: Failed to read the 'result' property from 'IDBRequest': The request has not finished.\n    at te._destroy (https://localhost/main.js:1:1017153)\n    at tt (https://localhost/main.js:1:980982)\n    at https://localhost/main.js:1:981201\n    at O.invoke (https://localhost/polyfills.js:1:23679)\n    at te.run (https://localhost/polyfills.js:1:18898)\n    at https://localhost/polyfills.js:1:30102\n    at O.invokeTask (https://localhost/polyfills.js:1:24298)\n    at te.runTask (https://localhost/polyfills.js:1:19556)\n    at ue (https://localhost/polyfills.js:1:26335)\n    at t.invokeTask [as invoke] (https://localhost/polyfills.js:1:25385)"
    },
    "log": [
      {
        "arguments": "[\"Unhandled Promise rejection:\",\"Failed to read the 'result' property from 'IDBRequest': The request has not finished.\",\"; Zone:\",\"<root>\",\"; Task:\",\"Promise.then\",\"; Value:\",\"InvalidStateError: Failed to read the 'result' property from 'IDBRequest': The request has not finished.\\n    at te._destroy (https://localhost/main.js:1:1017153)\\n    at tt (https://localhost/main.js:1:980982)\\n    at https://localhost/main.js:1:981201\\n    at O.invoke (https://localhost/polyfills.js:1:23679)\\n    at te.run (https://localhost/polyfills.js:1:18898)\\n    at https://localhost/polyfills.js:1:30102\\n    at O.invokeTask (https://localhost/polyfills.js:1:24298)\\n    at te.runTask (https://localhost/polyfills.js:1:19556)\\n    at ue (https://localhost/polyfills.js:1:26335)\\n    at t.invokeTask [as invoke] (https://localhost/polyfills.js:1:25385)\",\"InvalidStateError: Failed to read the 'result' property from 'IDBRequest': The request has not finished.\\n    at te._destroy (https://localhost/main.js:1:1017153)\\n    at tt (https://localhost/main.js:1:980982)\\n    at https://localhost/main.js:1:981201\\n    at O.invoke (https://localhost/polyfills.js:1:23679)\\n    at te.run (https://localhost/polyfills.js:1:18898)\\n    at https://localhost/polyfills.js:1:30102\\n    at O.invokeTask (https://localhost/polyfills.js:1:24298)\\n    at te.runTask (https://localhost/polyfills.js:1:19556)\\n    at ue (https://localhost/polyfills.js:1:26335)\\n    at t.invokeTask [as invoke] (https://localhost/polyfills.js:1:25385)\"]",
        "level": "error",
        "time": "2025-05-29T12:21:26.588Z"
      },
```

**Environment**
- Browser: Chromedriver
- App: webapp
- Version: 2.19.0

megha1807:
## Description

Fixes a race condition in the telemetry service that caused an `InvalidStateError: Failed to read the 'result' property from 'IDBRequest'`crash, which triggered unwanted feedback doc generation during e2e tests.

Fixes #9993

## Problem

When `submitIfNeeded()` called `db.destroy()` on old telemetry DBs, pending IDB requests were not fully settled before the current DB was opened. This caused an unhandled `InvalidStateError` promise rejection.

## Changes

- Added `await db.close()` before `await db.destroy()` in  `submitIfNeeded()` to flush all pending IDB requests first
- Fixed variable shadowing bug (`const db` → `let db`) so the catch block can correctly call `closeDataBase(db)` on failure
- Added 100ms delay after `submitIfNeeded()` before opening the current DB to let any lingering IDB transactions fully settle
- Added graceful `InvalidStateError` handling in `storeIt()` to prevent unhandled promise rejections from reaching the feedback doc system
- Updated unit tests to reflect new `close` + `destroy` ordering


webapp/src/ts/services/telemetry.service.ts
```diff
@@ -218,11 +218,19 @@ export class TelemetryService {
   }
 
   private async storeIt(db, key, value) {
-    return await db.post({
-      key: key,
-      value: value,
-      date_recorded: Date.now(),
-    });
+    try {
```
dianabarsan, line 221: I think there has been a misunderstanding. 
The feedback docs are not unwanted, they are unraveling a bug in the application. 
So I don't think we should just ignore it, we should figure out what is causing it and fix that instead of swallowing the error.

dianabarsan (changes requested):

megha1807:
"@dianabarsan the CI failure is unrelated to the code changes - the runner hit a rate limit trying to download the ShellCheck binary (403 rate limit exceeded). Could you please re-run the failing job when you get a chance? Thank you!"

webapp/src/ts/services/telemetry.service.ts
```diff
@@ -238,18 +238,21 @@ export class TelemetryService {
...
             continue;
           }
-          
+
+          let db;
           try {
-            const db = this.windowRef.PouchDB(dbName);
+            db = this.windowRef.PouchDB(dbName);
             await this.aggregate(db, dbName);
+            await db.close();
```
dianabarsan, line 251: I don't think this is the correct solution. Destroy would already close the database cleanly internally. 

You say "pending IDB requests". Which ones?

dianabarsan (changes requested):

webapp/src/ts/services/telemetry.service.ts
megha1807: Thank you for the review! You're right, I'll investigate the exact stack traces from #9993 to identify which IDB requests are pending and why. I'll revert the close() before destroy() change for now and come back with a proper root cause analysis. Will update the PR once I have more clarity.

megha1807:
Hello @dianabarsan 
After investigating the stack trace from #9993:

**Root cause:**
The crash at `te._destroy` inside PouchDB's IDB adapter occurs because `aggregateMapReduce(db)` calls `db.query()` which opens an IDB cursortransaction to iterate over telemetry documents. Although `awaitPromise.all([...aggregateMapReduce(db)...])` resolves at the JavaScript level, the underlying IDB transaction has not yet fullycommitted at the browser level. When `db.destroy()` is called immediately after, PouchDB tries to abort all open IDB transactionsand hits this still-committing cursor transaction, throwing:

"Failed to read the 'result' property from 'IDBRequest':The request has not finished."

The pending IDB requests are specifically the cursor transactionsopened by pouchdb-mapreduce's `db.query()` inside `aggregateMapReduce()`.

**Fix:**
`await db.close()` before `await db.destroy()` is the correct solution.PouchDB's `close()` waits for all pending IDB transactions to finish before closing the connection, which prevents `destroy()` from encountering the still-open cursor transaction.

Also keeping the `let db` fix which resolves the variable shadowing bug where the catch block was calling `closeDataBase(undefined)` instead ofthe actual db instance.

dianabarsan:
> resolves at the JavaScript level, the underlying IDB transaction has not yet fullycommitted at the browser level.

This would be a severe failing in the way that the PouchDb adapter works. Can you please share the source of this conclusion?

megha1807:
Thank you for pushing back on this, @dianabarsan.

I investigated the stack trace and can confirm the crash originates at `te._destroy` inside PouchDB's IDB adapter. However, I was not able to identify the exact source-level reason why `destroy()` encounters a pending IDB request after `aggregate()` has fully resolved.

I don't have a definitive source to back my claim that the IDB transaction from `db.query()` is not yet committed when `destroy()` runs.

Could you help point me in the right direction? Specifically:
- Is there something in the CHT codebase that might be leaving an IDB request open after `aggregate()` resolves?
- Should I look at the PouchDB source for `_destroy` directly to  understand what requests it encounters?

I want to make sure the fix is correct and well-understood before merging.

dianabarsan:
> Is there something in the CHT codebase that might be leaving an IDB request open after aggregate() resolves?

I don't believe there is anything of the sort. We actively try to not do any low-level actions that can have these kinds of obscure consequences. 

> Should I look at the PouchDB source for _destroy directly to understand what requests it encounters?

I think this would be the only way to fully understand this. 
I think we can't just accept that transactions just hang beyond what PouchDb reports as fulfilled, this would have severe consequences in how the adapter should be treated and would require changes upstream. And given that the aggregation function specifically does not allow parallel executions, and we can exclude weird scenarios where multiple tabs would be open (since this is an e2e test and we _know_ we are only running one test at a time).

megha1807:
Hello @dianabarsan 
After reading the PouchDB IDB adapter source directly, I can now explain exactly what happens.

`_destroy` closes `openReq.result` (the connection stored in `openReqList`), then immediately calls `indexedDB.deleteDatabase()`:

```js
var openReq = openReqList.get(dbName);
if (openReq && openReq.result) {
  openReq.result.close();  // closes openReq's connection only
  cachedDBs.delete(dbName);
}
var req = indexedDB.deleteDatabase(dbName);
```

However, `_close` closes `idb` - the live connection variable used for all active transactions:

```js
api._close = function (callback) {
  idb.close();  // closes the live connection
  cachedDBs.delete(dbName);
  callback();
};
```

These are two different references. `_destroy` never calls `idb.close()`.

So the sequence causing the crash is:
1. `aggregate()` calls `db.query()` which opens IDB cursor transactions on the live `idb` connection
2. `aggregate()` resolves at the JS Promise level
3. `db.destroy()` is called - it closes `openReq.result` but NOT `idb`
4. `indexedDB.deleteDatabase()` is called while `idb` still has the cursor transaction active
5. PouchDB's internal `_destroy` callback fires, tries to read the IDBRequest result, and throws `InvalidStateError: The request has not finished`

`await db.close()` before `await db.destroy()` is the correct fix because it calls `idb.close()`, closing the live connection and ensuring all active transactions are finished before `deleteDatabase()` runs.

Source: `pouchdb-browser/lib/index.js` lines 6121-6124 (`_close`) and 6309-6325 (`_destroy`).

dianabarsan (changes requested):
Thanks for digging into the source. I checked your explanation against pouchdb-browser v9.0.0 (the version webapp uses) and there's a factual issue 
with the load-bearing claim. 
`idb` and `openReq.result` are not two different references. In v9:
  - `var idb = null;` — line 5881
  - `idb = e.target.result;` inside `req.onsuccess` — line 6513
  - `openReqList.set(dbName, req);` — line 6471
  
Since `e.target === req`, this means `idb === req.result === openReqList.get(dbName).result` — the same `IDBDatabase` instance. So `_destroy`'s 
`openReq.result.close()` (line 6443) and `_close`'s `idb.close()` (line 6252) are calling `.close()` on the same connection:
```js
  // _close (line 6249)
  api._close = function (callback) {
    idb.close();
    cachedDBs.delete(dbName);
    callback();
  };

  // _destroy (line 6437)
  api._destroy = function (opts, callback) {
    changesHandler.removeAllListeners(dbName);
    var openReq = openReqList.get(dbName);
    if (openReq && openReq.result) {
      openReq.result.close();   // same IDBDatabase as `idb`
      cachedDBs.delete(dbName);
    }
    var req = indexedDB.deleteDatabase(dbName);
    // ...
  };
```

That means your mechanism — "destroy never closes idb" — doesn't match what the source actually does. The fix may still empirically work, but for a different 
reason. The most likely explanation is the JS-level promise/microtask drain that `await db.close()` forces before destroy runs, not a missing reference close. 
`IDBDatabase.close()` itself returns immediately and only sets a flag; it doesn't wait for transactions to complete.

Could you re-investigate and find the actual cause? I'm not comfortable merging a fix whose mechanism we don't understand — and I don't want to swallow 
`InvalidStateError` either, since it could be surfacing a real bug we should fix at the source. Specifically: **which IDBRequest's `.result` is being read 
prematurely?** The `te._destroy` stack frame in the issue points to PouchDB's destroy callback chain — the `idbError` handler and the cursor request from 
`db.query()` are both candidates worth tracing.

If you can't pin down the root cause, the right move is to scope this PR down to the unambiguous parts — the `let db` + catch fix — and leave #9993 open with 
what you've learned documented in it, so we (or someone else) can pick it up with a proper understanding rather than committing to a mechanism we can't verify.

megha1807:
Thank you @dianabarsan for catching that - you're right that idb and openReq.result are the same IDBDatabase instance in v9, so my mechanism was incorrect.

Following your suggestion, I'm scoping this PR down to the unambiguous fix only: the `let db` variable shadowing bug, which caused the catch block to call `closeDataBase(undefined)` instead of the actual db instance.

I've reverted the `close()`-before-`destroy()` change and the related test changes. I'll document what was learned in issue #9993 so the actual root cause of the InvalidStateError can be investigated properly.

PR #10857, opened 2026-04-14
feat(#6571): add security log scanner for credentials and errors in CI

### LINKED ISSUE #6571: Detect logged passwords and errors
**Describe the issue**
We want to detect regressions that cause a) passwords to be written to server logs in plaintext, or b) unexpected errors to be thrown.

When we run our e2e test suite in CI we write api and sentinel stderr and stdout to a log. We should add a test to fail the build if these logs contain passwords or unexpected errors. This will detect regressions early.

- Adding a winston plugin to automatically mask the password, but I think it's better for the code doing the logging to do it. Also this doesn't detect unexpected errors.
- Being more diligent, but we're all fallible.
- Watching the log output in production but it's too late by then.

vikrantwiz02:
# Description

Adds a post-test credential scanner that runs automatically after every E2E and integration test job in CI, failing the build if credentials are detected in server logs.

Fixes #6571

## What was added

**`scripts/ci/.secretlintrc.json`** — Declarative config for [`secretlint`](https://github.com/secretlint/secretlint) that detects credential leaks in `*.log` files using two rule sets:

- `@secretlint/secretlint-rule-preset-recommend` — the standard secretlint preset
- `@secretlint/secretlint-rule-pattern` — custom patterns for CHT-specific cases:
  - `user:pass@host` URLs in any scheme, including `localhost` (the preset's basicauth rule requires a domain with a dot, so it would miss `localhost:5984`)
  - Credentials in URI query parameters (`?password=secret&token=abc&pass=abc`)
  - JSON key/value pairs (`"password":"..."`, `"pass":"..."`, `"token":"..."`, `"secret":"..."`)
  - `Authorization:` headers accidentally written to logs

Safe patterns (`Bearer ***`, `[REDACTED]`, `***`) are excluded via negative lookaheads in the config.

**`webapp/tests/mocha/unit/testingtests/secretlintrc.spec.js`** — 18 mocha/chai test cases that verify the secretlint config flags known-bad log lines and does not flag safe patterns.

**`.github/workflows/build.yml`** — One step added (`if: always()`) to both test jobs, positioned after the test run and before `Archive Results`:

- **Scan logs for credential leaks** — runs `secretlint` against `tests/logs/*.log`

## What was removed

- `scripts/ci/log-scanner.js` — replaced by secretlint
- `scripts/ci/log-scanner-allowlist.json` — patterns moved into the secretlint config

## Stress test results

| # | Scenario | Input | Expected | Result |
|---|----------|-------|----------|--------|
| 1 | Credential leak — URL | `https://admin:password123@localhost:5984` | exit 1, violation flagged | ✓ |
| 2 | Credential leak — JSON secret | `{"secret":"super-secret-token"}` | exit 1, violation flagged | ✓ |
| 3 | Credential leak — pass param | `GET /api?pass=hunter2` | exit 1, violation flagged | ✓ |
| 4 | Safe pattern — already redacted | `{"password":"[REDACTED]"}` | exit 0, no violation | ✓ |
| 5 | Safe pattern — Bearer masked | `Authorization: Bearer ***` | exit 0, no violation | ✓ |
| 6 | Safe pattern — URL without credentials | `GET http://localhost:5984/medic` | exit 0, no violation | ✓ |

# Code review checklist

- [x] Tested: 18 mocha/chai tests in `webapp/tests/mocha/unit/testingtests/secretlintrc.spec.js`

# License


scripts/ci/log-scanner.js
```diff
@@ -0,0 +1,398 @@
+#!/usr/bin/env node
```
dianabarsan, line 1: This is a huge script! This would involve some high maintenance cost for our team. Instead of having a custom script, can you please find an npm package alternative which does the same thing, that is tested and maintained by someone else?

dianabarsan (changes requested):

scripts/ci/log-scanner.js
vikrantwiz02: Hii @dianabarsan, I am currently traveling and will arrive tomorrow night and need time till 22nd april, in worst case but I will try within 1-2 days to handle this. Till that in midway, I will do research about this.

Thanks for the review.

vikrantwiz02: Hi @dianabarsan, I found secretlint as an alternative to handle the credential scanning.

I can replace the custom logic with secretlint and its standard rules (Basic Auth, DB strings, etc.), which moves the complexity into a standard configuration file. Since secretlint doesn't handle error/stack-trace scanning, I’ll replace that part of the JS script with a targeted, lightweight shell command directly in the workflow.

What you think...?

dianabarsan: That sounds wonderful! Thanks so much for looking into it!

vikrantwiz02:
Hi @dianabarsan, the replacement is done! Here's what changed:

- scripts/ci/log-scanner.js (398 lines) and log-scanner-allowlist.json are deleted.
- scripts/ci/.secretlintrc.json — 28-line declarative config that drives @secretlint/secretlint-rule-pattern for credential detection (user:pass@host URLs, query-param secrets, JSON credential fields, Authorization headers).
- scripts/ci/test-secretlintrc.js — 16 test cases for the config, wired into npm run lint so regressions are caught automatically.
- build.yml — two lightweight steps replace the old one:
   1. npx secretlint for credential scanning (all jobs)
   2. A grep -E one-liner for unexpected ERROR: lines and app-level stack traces (WebDriver jobs only)
   
Could you take another look? Thanks again for the suggestion!

.claude/settings.json
```diff
@@ -0,0 +1,20 @@
+{
+  "permissions": {
```
dianabarsan: Please revert this file.

.github/workflows/build.yml
```diff
@@ -366,6 +371,25 @@ jobs:
...
+    - name: Scan logs for credential leaks
+      if: always()
+      run: |
+        compgen -G "tests/logs/*.log" > /dev/null || exit 0
+        npx --no-install secretlint --secretlintrc scripts/ci/.secretlintrc.json "tests/logs/*.log"
+
+    - name: Scan logs for unexpected errors
+      if: always()
+      run: |
+        [[ "${{ matrix.cmd }}" == ci-webdriver* || "${{ matrix.cmd }}" == wdio* ]] || exit 0
```
dianabarsan: Does this line end up skipping integration tests? why are we excluding any tests here?

.github/workflows/build.yml
```diff
@@ -366,6 +371,25 @@ jobs:
...
+        compgen -G "tests/logs/*.log" > /dev/null || exit 0
+        PATTERN='ERROR:|^\s+at .*\.js:[0-9]+'
+        ALLOWED='Error loading translations|Invalid basic auth|Unhandled rejection|connect ECONNREFUSED|STARTUP RES|processChange failed|errored on doc|connecting to upgrade service|TCPConnectWrap|internalConnectMultiple|afterConnectMultiple|callbackTrampoline|Error while logging in with token|error saving changes on doc|Error when persisting log|No RapidPro URL|No api key configured|Empty response received when sending|Error: 429|couch-request/src/couch-request\.js|server-checks/src/checks\.js|server-checks/src/index\.js|sentinel/server\.js'
```
dianabarsan: I think these will be extremely hard to read and follow. Can you store them in a way that they are manageable? So that we can inspect, add and remove allowed errors easily?

dianabarsan: Also ... if we're allowing 
"Unhandled rejection" arent'e we just giving blank slate to so many errors and this script will never catch any?

scripts/ci/.secretlintrc.json
```diff
@@ -0,0 +1,27 @@
+{
+  "rules": [
+    {
```
dianabarsan, line 6: Can we include the recommended preset along with these custom rules: 
```{ "id": "@secretlint/secretlint-rule-preset-recommend" }```

Comment (dianabarsan) [inline scripts/ci/.secretlintrc.json:6]: Can we include the recommended preset along with these custom rules: 
```{ "id": "@secretlint/secretlint-rule-preset-recommend" }```

File: scripts/ci/.secretlintrc.json [hunk 6]
```diff
@@ -0,0 +1,27 @@
...
+      "id": "@secretlint/secretlint-rule-pattern",
+      "options": {
+        "patterns": [
+          {
+            "name": "user:pass@host URL",
+            "patterns": ["/[a-zA-Z][a-zA-Z0-9+\\-.]*:\\/\\/[^:\\s\\/]+:[^@\\s\\/]+@/"]
```
dianabarsan, line 12: I believe the recommended preset has a rule for basic auth.

scripts/ci/.secretlintrc.json
```diff
@@ -0,0 +1,27 @@
...
+          },
+          {
+            "name": "credential in query parameter",
+            "patterns": ["/(password|passwd|token|secret|api[_-]key)=[^&\\s\"']+/i"]
```
dianabarsan: please include "pass" here and in the other patterns.

scripts/ci/test-secretlintrc.js
```diff
@@ -0,0 +1,157 @@
...
+// Test cases
+// ---------------------------------------------------------------------------
+
+/**
+ * Each entry:
+ *   desc        – test description
+ *   line        – single log line written to a temp file
+ *   expectFlag  – true if secretlint should exit 1 (violation found)
+ */
+const TESTS = [
```
dianabarsan: Since this is a test file, it should follow our test patterns and be placed in a test folder. Can you convert it to use mocha/chai, move it to /webapp/tests/mocha/unit/testingtests where we already have some utilities for testing tests?  
It doesn't make sense for this test to fail during linting.

dianabarsan (changes requested):
I'm really on board with secretlint. Unsure about the custom log scanning, not convinced about that one. So I think we should maybe move forward with secretlint alone and do the log scanning another time, since that is more involved work. 

I left some requests. Also I see that CI is failing due to some errors.

.claude/settings.json
vikrantwiz02: done

vikrantwiz02:
Hi @dianabarsan, addressed all the feedback:

- Removed the "Scan logs for unexpected errors" step entirely, moving forward with secretlint only as you suggested. The ALLOWED list, the Unhandled rejection concern, and the ci-webdriver* filter are all gone with it.
- Added @secretlint/secretlint-rule-preset-recommend as the first rule in .secretlintrc.json.
- Kept the custom user:pass@host URL pattern alongside the preset, the preset's basicauth rule ignores localhost by design (requires a domain with a dot), so CHT's localhost:5984 URLs would slip through without it. The two rules complement each other.
- Added pass to the query-parameter and JSON key/value patterns.
- Moved tests to [webapp/tests/mocha/unit/testingtests/secretlintrc.spec.js](https://github.com/vikrantwiz02/cht-core/blob/cd604134e/webapp/tests/mocha/unit/testingtests/secretlintrc.spec.js) using mocha/chai. The mocha glob tests/mocha/unit/**/*.spec.* picks it up automatically. Added two new cases for the pass pattern. Removed node scripts/ci/test-secretlintrc.js from npm run lint.
- Reverted .claude/settings.json.
- Fixed build failures: @secretlint/secretlint-rule-preset-recommend was missing from package-lock.json; also replaced execFileSync + try/catch with spawnSync in the test helper, spawnSync returns the exit code directly as result.status so no catch block is needed (the project's ecmaVersion: 2018 ESLint config doesn't support optional catch binding). The actual secretlint scan in CI is unchanged.

Now, waiting for CI checks to pass.

scripts/ci/.secretlintrc.json
dianabarsan: Could you add a scripts/ci/README.md (or similar) documenting why each custom rule exists? secretlint's pattern rule has no inline-description support, so the rationale is invisible to future contributors.

scripts/ci/.secretlintrc.json
```diff
@@ -0,0 +1,30 @@
...
+      "id": "@secretlint/secretlint-rule-pattern",
+      "options": {
+        "patterns": [
+          {
+            "name": "user:pass@host URL",
+            "patterns": ["/[a-zA-Z][a-zA-Z0-9+\\-.]*:\\/\\/[^:\\s\\/]+:[^@\\s\\/]+@/"]
+          },
+          {
+            "name": "credential in query parameter",
+            "patterns": ["/(password|passwd|pass|token|secret|api[_-]key)=[^&\\s\"']+/i"]
```
dianabarsan: The query parameter rule has no negative lookahead for safe values, so ?password=[REDACTED] and ?password=*** are flagged as leaks. Could you mirror the JSON rule's (?!\*{3}|\[REDACTED\]) exclusion here? The PR description claims safe patterns are excluded, but it's only the JSON and Authorization rules that actually do this.

scripts/ci/.secretlintrc.json
```diff
@@ -0,0 +1,30 @@
...
+          },
+          {
+            "name": "credential in JSON key/value",
+            "patterns": ["/\"(password|passwd|pass|token|secret|api[_-]key)\"\\s*:\\s*\"(?!\\*{3}|\\[REDACTED\\])[^\"]+\"/i"]
+          },
+          {
+            "name": "Authorization header",
+            "patterns": ["/\\bAuthorization\\s*:\\s*(?!Bearer \\*)\\S+/i"]
```
dianabarsan: The Authorization header rule's negative lookahead is (?!Bearer \*) — a single literal asterisk. That accidentally treats Bearer *anything as safe (e.g. Bearer *abc-real-token slips through). Suggest tightening to (?!Bearer \*{3,}) to match the documented Bearer *** mask.

webapp/tests/mocha/unit/testingtests/secretlintrc.spec.js
```diff
@@ -0,0 +1,84 @@
...
+    const logFile = path.join(tmpDir, 'test.log');
+    fs.writeFileSync(logFile, line + '\n', 'utf8');
+    const result = spawnSync(SECRETLINT, ['--secretlintrc', SECRETLINTRC, logFile], { stdio: 'pipe' });
+    return result.status !== 0;
+  } finally {
+    fs.rmSync(tmpDir, { recursive: true, force: true });
+  }
+};
+
+describe('scripts/ci/.secretlintrc.json', () => {
```
dianabarsan, line 27: Could you expand the test suite into a small matrix? 
It currently exercises only a few hand-picked combinations of keyword × form × value, which is why two of the rules have bugs that the existing tests don't reveal:                                                                                                                                    
- ?password=[REDACTED] is currently flagged as a leak (no safe-value exclusion in the query rule).                                                                                                                                                                                                                                                          
- Authorization: Bearer *anything is currently treated as safe (the (?!Bearer \*) lookahead matches a single asterisk, so any token starting with * slips through).

A matrix over {password, passwd, pass, token, secret, api_key, api-key} × {query, JSON, Authorization} × {real value, ***, [REDACTED], ****} would catch both. Also worth adding: a precondition check that node_modules/.bin/secretlint exists, and asserting on stdout containing [PATTERN] rather than just non-zero exit — otherwise a misconfigured rule could pass tests silently.

dianabarsan (changes requested):
Nice! Thanks for all the changes! I added a few comments inline.

vikrantwiz02:
Hi @dianabarsan, fixed both issues:

1. Query param rule — added the same safe-value exclusion (***, [REDACTED], ****) that the JSON rule already had, so ?password=[REDACTED] no longer gets flagged.
2. Authorization rule — tightened (?!Bearer \*) to (?!Bearer \*{3,}) so Bearer *abc is correctly flagged as a real token, while Bearer *** and Bearer **** remain safe.
3. Tests — expanded to a full matrix (all 7 keywords × query/JSON × real value + safe values), added a before() check that the secretlint binary exists, and added stdout assertion on every flagged case so a misconfigured rule can't silently pass.

.github/workflows/build.yml
```diff
@@ -16,7 +16,7 @@ env:
   DOCKERHUB_USERNAME: 'dockermedic'
   ECR_REPO: '720541322708.dkr.ecr.eu-west-2.amazonaws.com/medic'
   ECR_PUBLIC_REPO: 'public.ecr.aws/medic'
-  COUCHDB_LOG_LEVEL: 'debug'
+  COUCHDB_LOG_LEVEL: 'info'
```
dianabarsan: Looking at the commit message, this was lowered because at debug CouchDB logs OS-process input dumps that include user-doc password fields, which the new secretlint scanner correctly flags. So the change does fix a real CI failure.                                                                                                                     

That said, the loss of CouchDB-level debug for diagnosing flaky tests is a real cost, and the framing in the commit ("noise fix") understates the tradeoff. A few alternatives that would let CouchDB stay at debug:                                                                                                                                        
              
1. Add a .secretlintignore entry for the recognizable OS Process #Port<...> Input :: log line shape — skip those known-noisy dumps without weakening the scan elsewhere.                                                                                                                                                                                    
2. Strip those lines via grep -v before secretlint runs.
3. Keep info by default but allow a workflow_dispatch input to elevate to debug for targeted debugging runs.                                                                                                                                                                                                                                                

Could you try option 1 or 2? They preserve debug visibility for the rare cases we need it. If neither works cleanly, I'd accept the info change but please update the commit message to acknowledge it as a security tradeoff, not just noise reduction. Note that LOG_LEVEL: 'debug' (app logs) is unchanged, which is the more important debug surface    anyway.

dianabarsan (commented):
Looking good. I left one question inline.

.github/workflows/build.yml
vikrantwiz02: Hey @dianabarsan, I’ve been looking into those two options. I’m actually prefering Option 2. Bcauz,

In option 1, .secretlintignore only really works at the file level. To stop the false positives that way, we'd basically have to ignore the log files entirely, which kind of defeats the point of having the scanner in the first place.

If I go with Option 2, I can just use grep -v "OS Process #Port" to create a filtered version of the logs for the scan. This lets me flip COUCHDB_LOG_LEVEL back to debug, so the logs stay useful for everyone like you mentioned, without the scanner tripping over those CouchDB process dumps.

What do you think of this approach...? If this is good to go, I can implement.

dianabarsan: That sounds great! go for it

.github/workflows/build.yml
```diff
@@ -227,6 +227,14 @@ jobs:
       - uses: azure/setup-kubectl@776406bce94f63e41d621b960d78ee25c8b76ede  # v4
       - name: Run tests
         run: npm run ${{ matrix.cmd }}
+      - name: Scan logs for credential leaks
+        if: always()
+        run: |
+          compgen -G "tests/logs/*.log" > /dev/null || exit 0
+          tmpdir=$(mktemp -d)
+          for f in tests/logs/*.log; do grep -v "OS Process #Port" "$f" > "$tmpdir/$(basename "$f")"; done
```
dianabarsan: The OS Process #Port filter is a single-string match                                                                                                                                                                                                                                                                                                       
   
grep -v "OS Process #Port" "$f" > "$tmpdir/$(basename "$f")"                                                                                                                                                                                                                                                                                                  
  A few risks worth documenting in code or in a comment:                                                                                                                                                                                                                                                                                                        
  - If a future CouchDB version changes the log line format (drops #Port, uses <> instead, etc.), the filter silently stops working and CI starts failing on debug-level dumps again. A regex match like grep -Ev "OS Process #?Port|OS Process .* Input ::" would be more robust to format drift.                                                              
  - Conversely, if a non-CouchDB log line happens to contain OS Process #Port, that line gets filtered and a real credential leak there would slip through. Very unlikely, but not zero — and there's no comment in the workflow file explaining the trade-off.     


A one-line comment in build.yml above the grep -v would be enough:                                                                                                                                                                                                                                                                                            
```
```

webapp/tests/mocha/unit/testingtests/secretlintrc.spec.js
```diff
@@ -0,0 +1,108 @@
...
+  before(() => {
+    expect(fs.existsSync(SECRETLINT), `secretlint binary not found at ${SECRETLINT}`).to.be.true;
+  });
+
+  describe('query parameter rule', () => {
```
dianabarsan, line 32: The query/JSON safe-value tests don't have an analogue to the Bearer *abc-real-token regression test — i.e. there's no test that `?password=*` (single asterisk, not a valid mask) is still flagged. Symmetric coverage would catch a future regression where the negative lookahead got loosened. Two extra tests would do it.

dianabarsan (changes requested):
Solid work — all the substantive issues from earlier rounds are fixed and verified. The matrix-driven tests are a nice improvement over the original happy-path coverage, and the before() binary check is a good precaution.
Two small requests inline.

.github/workflows/build.yml
vikrantwiz02: Done

webapp/tests/mocha/unit/testingtests/secretlintrc.spec.js
vikrantwiz02: Handled

dianabarsan (changes requested):
Two things on this step:                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                
  **1. Pattern propagation — the OS-Process filter only got updated in the kubectl job above.** The comment + tightened regex (`grep -Ev "OS Process #?Port|OS Process .* Input ::"`) landed there, but this job still has the original single-string `grep -v "OS Process #Port"` and no comment. Both jobs inherit `COUCHDB_LOG_LEVEL: 'debug'` from line 19, 
  so they ingest the same CouchDB dumps and need the same filter and rationale. If CouchDB changes the log format, this job silently starts failing while the other keeps working.
                                                                                                                                                                                                                                                                                                                                                                
  Could you mirror the block from the kubectl job here?

  ```yaml
# Strip CouchDB OS-Process debug dumps which legitimately log user creation doc bodies (with passwords). 
for f in tests/logs/*.log; do grep -Ev "OS Process #?Port|OS Process .* Input ::" "$f" > "$tmpdir/$(basename "$f")"; done
```
                                                                                                                                                                                                                                                                                                                                                        
  2. Do we still need npx --no-install now that secretlint is a direct devDependency? After npm install, the binary lives at node_modules/.bin/secretlint and is directly executable — the spec file at webapp/tests/mocha/unit/testingtests/secretlintrc.spec.js already invokes it that way. Two cleaner options for the workflow:                            
                                                                                                                                                                                                                                                                                                                                                                
  - Direct path: ./node_modules/.bin/secretlint --secretlintrc ...                                                                                                                                                                                                                                                                                              
  - npm script: add "scan-logs": "secretlint --secretlintrc scripts/ci/.secretlintrc.json" to root package.json and call npm run scan-logs -- "$tmpdir"/*.log.
                                                                                                                                                                                                                                                                                                                                                                
  The npm-script form is probably the most idiomatic here and makes it consistent across both jobs. Same comment applies to the kubectl job above.

**Comment by vikrantwiz02** [comment]
Hi @dianabarsan, addressed both points:
1. Mirrored the comment + tightened `grep -Ev "OS Process #?Port|OS Process .* Input ::"` into the second job — both steps are now identical.
2. Added `scan-logs` npm script to root `package.json` and updated both steps to use `npm run scan-logs -- "$tmpdir"/*.log` instead of `npx --no-install`.

File: scripts/ci/README.md [hunk 15]
```diff
@@ -0,0 +1,23 @@
+# scripts/ci
+
+CI helper scripts for CHT Core.
+
+## `.secretlintrc.json` — credential leak detection
+
+Runs via `npx secretlint` in both test jobs after tests complete (see `build.yml`).
```
dianabarsan: Runs via: `npm run scan-logs`

package.json
```diff
@@ -68,6 +68,7 @@
     "ci-integration-all-k3d": "mocha --config tests/integration/.mocharc-k3d.js",
     "ci-integration-sentinel": "mocha --config tests/integration/.mocharc-sentinel.js",
     "ci-integration-sentinel-k3d": "mocha --config tests/integration/.mocharc-sentinel-k3d.js",
+    "scan-logs": "secretlint --secretlintrc scripts/ci/.secretlintrc.json",
```
dianabarsan: I got a better idea — could we extract the whole "make tmpdir, filter, scan, clean up" flow into a script so `npm run scan-logs` does the same thing locally as in CI? Right now the filter logic lives in two copy-pasted yaml blocks; if someone updates one and forgets the other, they'll silently diverge again.

  Add `scripts/ci/scan-logs.sh`:
  ```bash
  #!/usr/bin/env bash
  set -euo pipefail
  compgen -G "tests/logs/*.log" > /dev/null || exit 0
  tmpdir=$(mktemp -d)
  trap 'rm -rf "$tmpdir"' EXIT
  for f in tests/logs/*.log; do
    grep -Ev "OS Process #?Port|OS Process .* Input ::" "$f" > "$tmpdir/$(basename "$f")"
  done
  exec ./node_modules/.bin/secretlint --secretlintrc scripts/ci/.secretlintrc.json "$tmpdir"/*.log
```
  Then package.json:
```
  "scan-logs": "scripts/ci/scan-logs.sh",
```
  
  And both workflow jobs collapse to a single line:
```
  - name: Scan logs for credential leaks
    if: always()
    run: npm run scan-logs
```    
  Three wins: one source of truth for the filter, `npm run scan-logs` works locally after `npm run integration-test` or similar, and trap ensures cleanup even on scan failure.

dianabarsan (changes requested):

vikrantwiz02:
Hi @dianabarsan, done:
- Added `scripts/ci/scan-logs.sh` with the full flow (tmpdir, `trap` cleanup, grep filter, secretlint).
- Updated `scan-logs` in `package.json` to point to the script.
- Both workflow steps now collapse to `run: npm run scan-logs`, single source of truth, no more copy-paste drift.
- Updated README to reference `npm run scan-logs`.

dianabarsan (approved):
Yoohoo!!! Awesome work!

PR #10861, opened 2026-04-14
feat(#10855): add custom outbound headers and proxy support

### LINKED ISSUE #10855: Allow custom headers in Sentinel outbound requests
We are running CHT Core 5.1.1 and integrating outbound data from Sentinel to a DHIS2 instance. The outbound requests are rejected because required HTTP headers are missing.  

Currently, Sentinel does not allow configuring custom HTTP headers. Only limited support for the `Authorization` header appears to exist, but other commonly required headers such as `x-api-key` or explicit `Content-Type` cannot be set.  

This makes it difficult or impossible to integrate with APIs like DHIS2 that require specific headers for authentication or request validation.

---

We would like Sentinel outbound requests to support configurable HTTP headers.  

This should allow defining:
- Custom headers (e.g. `x-api-key`)  
- Standard headers (e.g. `Content-Type: application/json`)  
- Flexible authentication mechanisms beyond the `Authorization` header  

Ideally, headers could be defined in the outbound configuration and applied to all outgoing requests.

---

As a workaround, we considered introducing a reverse or forward proxy to intercept outbound requests and inject the required headers.  

However, this adds:
- Additional infrastructure complexity  
- Operational overhead  
- Extra points of failure  

A native solution within CHT Core would be preferable.

---

**Additional context**  
This limitation seems related to previous work on outbound headers (see https://github.com/medic/cht-core/issues/9799), which introduced support for `Authorization` headers. However, full support for arbitrary/custom headers (e.g. `x-api-key`) is missing.  

Relevant source code:
- `shared-libs/outbound/src/outbound.js`  
- `shared-libs/couch-request/src/couch-request.js`  

Use case: Integration with DHIS2 requiring custom headers for authentication and request handling.

magp18:
closes #10855

# Description

Something for the issue #10855 where you can specify a proxy and also have a more custom headers than before. example of an object now supported:

```json
{
  "outbound": {
    "dhis2_sync": {
      "destination": {
        "base_url": "https://lightning.dk...",
        "path": "/i/deb0c884-5268-45fa-8f3f-ajdask",

        "proxy": "http://...",

        "headers": {
          "Content-Type": "application/json",
          "X-Source": "CHT"
        },
        "auth": {
          "type": "header",
          "name": "x-api-key",
          "value_key": "dhis2-outbound-api-key"
        }
      },
      "mapping": {
        "id": "doc._id",
        "reported_date": "doc.reported_date",
        "patient_id": "doc.fields.patient_id"
      },
      "relevant_to": "doc.form === 'my_form'"
    }
  }
}
```

_This PR was heavily implemented using Cursor Auto mode and then reviewed by me and my colleage @delcroip_

# Code review checklist

# License


delcroip (approved):

delcroip:
@magp18 please have a look on the failed CI

  1 failing

  1) outbound shared library
       handleHeaderAuth
         throws for unsupported header name:
     AssertionError: expected 'You must pass the key for the credent…' to include 'Unsupported header name'
      at Context.<anonymous> (test/outbound.spec.js:530:16)



ERROR: Coverage for branches (86.3%) does not meet global threshold (95%)

shared-libs/outbound/src/outbound.js
```diff
@@ -101,15 +101,39 @@ const handleBasicAuth = async (authConf, sendOptions) => {
...
+  if (Object.prototype.toString.call(destinationHeaders) !== '[object Object]') {
+    throw new OutboundError('destination.headers must be an object');
+  }
+
   sendOptions.headers = sendOptions.headers || {};
-  if (authConf.name && authConf.name.toLowerCase() === 'authorization') {
-    const value = await fetchPassword(authConf.value_key);
-    sendOptions.headers.authorization = value;
+  Object.keys(destinationHeaders).forEach(headerName => {
+    sendOptions.headers[headerName.toLowerCase()] = destinationHeaders[headerName];
```
jkuester: I wonder if we should have the property value for the headers be an object instead of just a string? To me this scenario is a bit odd to read:

```json
        "headers": {
          "Content-Type": "application/json",
          "X-Source": "CHT"
        },
        "auth": {
          "type": "header",
          "name": "x-api-key",
          "value_key": "dhis2-outbound-api-key"
        }
```

Basically you define 3 headers, but they are splt across two different config sections.  Also, there is no way to deal with _multiple headers with sensitive values_. (Only the `auth` header is loaded from the secure settings.)

Instead, if each header config was an object, we could immediately add support for `value` and `value_key` properties (and potentially other properties in the future). Then, the config could just look like:

```json
        "headers": {
          "x-api-key": { value_key: "dhis2-outbound-api-key" }
          "Content-Type": { value: "application/json" },
          "X-Source": { value: "CHT" }
        },
```

And we don't need an `auth` section at all.

shared-libs/outbound/src/outbound.js
```diff
@@ -101,15 +101,39 @@ const handleBasicAuth = async (authConf, sendOptions) => {
...
+
+const handleHeaderAuth = async (authConf, sendOptions) => {
+  if (!authConf.name) {
+    throw new OutboundError('No auth.name for header auth');
+  }
+
+  sendOptions.headers = sendOptions.headers || {};
+  const value = await fetchPassword(authConf.value_key);
+  sendOptions.headers[authConf.name.toLowerCase()] = value;
 };
```
jkuester, line 171: If you are okay with my above suggestion about supporting secure values in the normal `headers` configuration, then I think we can just roll this function back to what it was originally to remain backwards compatible with existing `auth` configs.

shared-libs/couch-request/src/couch-request.js
```diff
@@ -144,6 +144,39 @@ const lowercaseHeaders = headers => Object.assign(
...
+    }
+
+    options.dispatcher = new undici.ProxyAgent({
+      ...proxy,
+      uri: proxy.uri || proxy.url
+    });
+    return;
+  }
+
+  throw new Error('Invalid proxy configuration. Expected string URL or object with uri/url.');
```
jkuester: Is there a particular reason we need to be opinionated around validating the `proxy` config here (or supporting the `url` property)???  I wonder if it would not be better to just pass whatever is provided to the `ProxyAgent` constructor and let it validate/fail instead of trying to do extra validation here....  Then all this code just becomes:

```js
options.dispatcher = new undici.ProxyAgent(proxy);
```

shared-libs/couch-request/src/couch-request.js
jkuester, line 1: I opened a [forum thread](https://forum.communityhealthtoolkit.org/t/why-does-shared-libs-outbound-use-shared-libs-couch-request/5616) to try and understand why `shared-libs/outbound` users `shared-libs/couch-request`.  If the outcome of that thread is that we leave things as they are, then we should add unit tests for the new logic here.  👍

jkuester (changes requested):
Okay, @magp18 profound apologies for it taking so long for us to pick up this PR! 😞 😓  Things have been unusually busy for the past month and all the maintainers have been struggling to keep up.  

Thank you for submitting this and I am looking forward to shipping this feature and unlocking more great outbound integrations!  

I have left a few comments inline and my final requests are:

- [ ] Can we update the [outbound integration test](https://github.com/medic/cht-core/blob/2c5a640c59a17d2da2bccc71cf84791e4c90bab2/tests/integration/sentinel/schedules/outbound.spec.js) to cover any/all of the new functionality?  I guess sending a custom header should be pretty feasible. If including the `proxy` in the integration test is going to add a lot of complexity, then we can skip it, but it would be nice to have if it is possible.
- [ ] Can you possibly open a cht-docs PR to update the [Outbound config docs](https://github.com/medic/cht-docs/edit/main/content/en/building/reference/app-settings/outbound.md) to include these new properties?  🙏 

Thank you!

andrablaj:
@magp18 would you be available to look into @jkuester's comments so we can move this work further? Thanks!

magp18:
Sorry for the delay! Yes! It was also busy around here in the summer and then I had holidays. Will look into @jkuester comments! (and thanks for providing feedback already)

shared-libs/outbound/src/outbound.js
```diff
@@ -127,13 +157,14 @@ const handleConfiguredProxy = (config, sendOptions) => {
...
   sendOptions.headers = sendOptions.headers || {};
-  const value = await fetchPassword(authConf.value_key);
-  sendOptions.headers[authConf.name.toLowerCase()] = value;
+  if (authConf.name && authConf.name.toLowerCase() === 'authorization') {
+    const value = await fetchPassword(authConf.value_key);
+    sendOptions.headers.authorization = value;
+    return;
+  }
+  logger.error(`Unsupported header name '${authConf.name}'. Supported: authorization`);
+  throw new OutboundError(`Unsupported header name '${authConf.name}'. Supported: authorization`);
```
delcroip: now we support more than authorisation, why this message ?

magp18: ok changed the error message

shared-libs/couch-request/src/couch-request.js
jkuester: TLDR for the forum thread is that `couch-request` is a misleading name and it is appropriate to use that shared-lib in `outbound`.  So, our approach here is good!  👍

jkuester (approved):
Thanks for the followup here! LGTM!  

(I pushed a couple changes to hopefully fix the build and add an integration test for `proxy`.  Once CI passes, I will get this merged. 👍

PR #10863, opened 2026-04-14
chore(#7948): add actionable guidance to misconfigured transition error message

### LINKED ISSUE #7948: "transitions" error when using an existing DB
**Describe the bug**
Sometimes, when I am using an existing DB or depending on the values that are set in the `transitions` object an error will be displayed in the sentinel log, and that produces that every time that a person is created, for example, the `medic ID` property will not have a value

**To Reproduce**
Consistent steps to reproduce the behavior:
1. Update the `transitions` object located in the `app_settings.json` in the `default` config, with the values that have the `standard` config.
2. Upload the updated `default` config using `cht`
3. Try to create a person using the webapp.
4. The `Medic ID` is going to be empty, no matter how many times is `sync`
5. See the following error in the sentinel log:

image

If the error was due to re-using an existing DB, just upload again the config will fix it. 

**Environment**
- Instance: local
- Browser: Chrome
- Client platform: MacOS
- App: webapp
- Version: 4.0.1

cc: @garethbowen

santhosh-7777:
# Description

When a transition fails to load due to misconfiguration, the error message was generic
and provided no guidance. This change improves it to tell the user to check their
configuration and includes a link to the relevant docs.

Closes #7948

# Code review checklist
- [x] Tested: Unit tests updated
- [x] Backwards compatible
- [x] AI disclosure: Used AI assistance for guidance on file locations and implementing the changes.

# License

### DIFF
--- a/shared-libs/transitions/src/transitions/index.js
+++ b/shared-libs/transitions/src/transitions/index.js
@@ -153,7 +153,7 @@ const loadTransitions = (synchronous = false) => {
         `Failed loading transition "${transition}". ` +
         `This is likely due to an error in the transition configuration. ` +
         `Please check your configuration.`;
-      loadErrors = [errorMessage, (e && e.message) || 'unknown'];
+      loadErrors = [errorMessage, e?.message || 'unknown'];
       logger.error(errorMessage);
       logger.error('%o', e);
     }
--- a/shared-libs/transitions/src/transitions/index.js
+++ b/shared-libs/transitions/src/transitions/index.js
@@ -149,7 +149,8 @@ const loadTransitions = (synchronous = false) => {
     try {
       self._loadTransition(transition, synchronous);
     } catch (e) {
-      const errorMessage = `Failed loading transition "${transition}"`;
+      const errorMessage = `Failed loading transition "${transition}": check your configuration is valid. ` +
+     `See https://docs.communityhealthtoolkit.org/apps/reference/app-settings/transitions/`;
       loadErrors = [errorMessage, e && e.message || 'unknown'];
       logger.error(errorMessage);
       logger.error('%o', e);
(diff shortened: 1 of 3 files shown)


dianabarsan (changes requested):
@santhosh-7777 it looks like your pr is failing linting. can you please fix the errors and re-request a review? Thanks!

shared-libs/transitions/src/transitions/index.js
```diff
@@ -149,7 +149,8 @@ const loadTransitions = (synchronous = false) => {
     try {
       self._loadTransition(transition, synchronous);
     } catch (e) {
-      const errorMessage = `Failed loading transition "${transition}"`;
+      const errorMessage = `Failed loading transition "${transition}": check your configuration is valid. ` +
```
dianabarsan: I'm not sure that this qualifies as "actionable" guidance, as the issue requests. I think the desired guidance would be in the form of instructing the user how to change the config. But I think that might not be possible. To be honest, I'm not entirely sure what the issue author was looking for realistically. 

Maybe the message can be improved: 
```
Failed loading transition `registration`. This is likely due to an error in the transition configuration. Please check your configuration. 
```

I don't think we should have a link to the docs, because these links can change and we can end up with messy dead links in logs. 

One improvement I can think of is to tell the user which configuration value is affected. This could be like the transition itself will return a list of properties that it uses.

santhosh-7777:
Hi @dianabarsan, I’ve addressed the requested changes:
-> Updated the error message as suggested (removed docs link and improved wording)
-> Fixed the lint issues
-> Updated the tests accordingly
Could you please take another look when you have time? Thanks!

dianabarsan (changes requested):
The message is better! thank you! 

Unfortunately your linting is failing. Can you please fix it and re-request a review? Thanks!

dianabarsan (approved):
Lovely!

PR #10864, opened 2026-04-14
test(#10112): stabilize flaky upgrade e2e test

### LINKED ISSUE #10112: Flaky e2e test: upgrade
**Describe the issue**
This test fails with: 

```
[chrome 137.0.7151.119 linux #0-3] Running: chrome (v137.0.7151.119) on linux
[chrome 137.0.7151.119 linux #0-3] Session ID: c20f04d530b9bc57122ccbd51f133a4f
[chrome 137.0.7151.119 linux #0-3]
[chrome 137.0.7151.119 linux #0-3] » tests/e2e/upgrade/upgrade-master.wdio-spec.js
[chrome 137.0.7151.119 linux #0-3] Performing an upgrade from current branch to master
[chrome 137.0.7151.119 linux #0-3]    ✖ should upgrade from current version to master
[chrome 137.0.7151.119 linux #0-3]
[chrome 137.0.7151.119 linux #0-3] 1 failing (41.4s)
[chrome 137.0.7151.119 linux #0-3]
[chrome 137.0.7151.119 linux #0-3] 1) Performing an upgrade from current branch to master should upgrade from current version to master
[chrome 137.0.7151.119 linux #0-3] element ("aria/Application menu") still not displayed after 40000ms
[chrome 137.0.7151.119 linux #0-3] Error: element ("aria/Application menu") still not displayed after 40000ms
[chrome 137.0.7151.119 linux #0-3]     at async waitForAngularLoaded (/home/runner/work/cht-core/cht-core/tests/page-objects/default/common/common.wdio.page.js:180:3)
[chrome 137.0.7151.119 linux #0-3]     at async waitForPageLoaded (/home/runner/work/cht-core/cht-core/tests/page-objects/default/common/common.wdio.page.js:186:3)
[chrome 137.0.7151.119 linux #0-3]     at async Object.goToBase (/home/runner/work/cht-core/cht-core/tests/page-objects/default/common/common.wdio.page.js:354:3)
```

Example run: https://github.com/medic/cht-core/actions/runs/15997935656/job/45128555425

This seems to be not be connected to the actual upgrade process, but rather page loading issues.  
Fix the test.

santhosh-7777:
# Description
This PR stabilizes the flaky upgrade e2e test by replacing the use of `waitForDisplayed` with a more reliable `waitUntil` check in `waitForAngularLoaded`.

Previously, the test relied on a single visibility check which could fail due to timing issues during page load. This caused intermittent failures where the application menu was not yet rendered.

The updated implementation retries until the menu is actually visible, making the test more robust and consistent.

Closes #10112

# Code review checklist
- [x] Tested: e2e test behavior improved and validated via CI
- [x] Backwards compatible: No functional changes to application behavior

# License

### DIFF
--- a/tests/page-objects/default/common/common.wdio.page.js
+++ b/tests/page-objects/default/common/common.wdio.page.js
@@ -177,9 +177,16 @@ const waitForLoaders = async (timeout = 5000) => {
   });
 };
 
-const waitForAngularLoaded = async (timeout = 40000) => {
-  await hamburgerMenuSelectors.hamburgerMenu().waitForDisplayed({ timeout });
-};
+ const waitForAngularLoaded = async (timeout = 40000) => {
+  await browser.waitUntil(async () => {
+    const menu = await hamburgerMenuSelectors.hamburgerMenu();
+    return await menu.isDisplayed();
+  }, {
+    timeout,
+    interval: 500,
+    timeoutMsg: 'Application menu did not appear in time'
+  });
+ };
 
 const waitForPageLoaded = async (timeout) => {
   // if we immediately check for app loaders, we might bypass the initial page load (the bootstrap loader)


tests/page-objects/default/common/common.wdio.page.js
```diff
@@ -177,9 +177,16 @@ const waitForLoaders = async (timeout = 5000) => {
   });
 };
 
-const waitForAngularLoaded = async (timeout = 40000) => {
-  await hamburgerMenuSelectors.hamburgerMenu().waitForDisplayed({ timeout });
-};
+ const waitForAngularLoaded = async (timeout = 40000) => {
```
dianabarsan, line 180: We use `waitForAngularLoaded` very reliably in all our e2e test, I can't think of a reason why the specific upgrade test requires this change. 
What was the state of the application when the error was triggered? There should be an available screenshot.

dianabarsan (changes requested):

santhosh-7777:
Hi @dianabarsan, thanks for the feedback!
I initially suspected the failure was due to timing issues where the application menu was not yet rendered when the check was performed.
However, since `waitForAngularLoaded` is used reliably across tests, I agree that modifying it globally may not be the right approach without clear evidence.
I tried accessing the failed CI run linked in the issue, but the logs have expired. Could you please share a recent failure (or screenshot) so I can better understand the exact state of the application when the test fails?
Based on that, I can investigate a more targeted fix instead of modifying the shared helper.
Thanks!

dianabarsan:
Thank you @santhosh-7777 

Unfortunately I don't have an entry now. I think I saw it fail on a branch recently, but don't remember which. So I don't have an example :( 
Next time I see one, I'll link it here!

dianabarsan:
Closing this and the original issue. It seems that something else fixed it and the flakiness is no longer present.

Issue #10851, opened 2026-04-13
Searching with Devanagari numerals no longer cross-matches Latin numeral contacts for online users (regression from 4.x)


binokaryg:
**Describe the bug**
In CHT 4.22.0, online users could search for contacts using either Latin numerals (e.g., `12345`) or Devanagari numerals (e.g., `१२३४५`) and both would return all matching contacts regardless of which numeral system was used in the contact name.

For example, searching `12346` would return both "Person 12346" and "Person १२३४६".
In CHT 5.1.1, this cross-matching no longer works for online users. Searching `12346` only returns "Person 12346", and searching `१२३४६` only returns "Person १२३४६".

**To Reproduce**

1. Create two contacts: one named "Person 12346" (Latin numerals) and another named "Person १२३४६" (Devanagari numerals)
2. Log in as an online user
3. Go to the People tab
4. Search for `12346`
5. Observe that only "Person 12346" is returned
6. Search for 1१२३४६1
7. Observe that only "Person १२३४६" is returned

**Expected behavior**
Both searches should return both contacts, as was the behavior in 4.22.0 for online users.

**Logs**

**Screenshots**
4.22.0:

<img width="406" height="201" alt="Image" src="https://github.com/user-attachments/assets/d4e933a8-5aa7-4b4e-ae65-414fa7fe982f" />
<img width="406" height="201" alt="Image" src="https://github.com/user-attachments/assets/97995c96-1627-4a95-a51d-d48409a23228" />

---
5.1.1:

<img width="406" height="201" alt="Image" src="https://github.com/user-attachments/assets/4bdac555-53f9-4012-9d61-f07104bd677c" />
<img width="406" height="201" alt="Image" src="https://github.com/user-attachments/assets/e34ccfc8-bc2a-41b9-ac48-028104fa22e1" />

**Environment**
- Instances:
  - Not working in [demo-cht](https://demo-cht.dev.medicmobile.org/) (5.1.1)
  - Working in [gamma](https://gamma.dev.medicmobile.org/) (4.20.0)
- Browser: Chrome
- Client platform: -
- App: webapp
- Version: 5.x

**Additional context**
- This regression affects online users only. Offline users show the same behavior (no cross-matching) in both 4.22.0 and 5.1.1 but it would be great to support this for offline users as well.

Related:
- https://github.com/medic/cht-core/issues/9096

Thanks to @Mbaduwal for reporting this first.

ManbirS07:
[...]

Issue #2445, opened 2016-06-29
App takes minutes to load a `person` dropdown


sglangevin:
I got stuck here when trying to edit a CHP Area on lg-testing (v2.7.0-beta.3497):
![image](https://cloud.githubusercontent.com/assets/1286381/16463897/b233f0ee-3dec-11e6-8d46-7a123624c3c8.png)

When I clicked on the drop-down for Branch, I saw 2 requests, one for `contacts_by_type` and one for `doc_summaries_by_id`, but when I clicked on the drop-down for CHP or Supervisor, it seems to get stuck on the `contacts_by_type` request for a long time. Eventually, after 2-3 mins, I saw the `doc_summaries_by_id` request appear, along with `Error fetching facilities`:
![image](https://cloud.githubusercontent.com/assets/1286381/16463975/018752c6-3ded-11e6-8c07-fffa975fb6e0.png)

When trying to load the next bunch of contacts (by scrolling down in the drop-down), it again took a long time to respond with the summaries, but I didn't see any errors in the console. Here is more info on the console error - similar to the one reported in #2444:

![image](https://cloud.githubusercontent.com/assets/1286381/16464150/b0341c64-3ded-11e6-9ba2-6633652a15b1.png)

This no longer meets our goal of acceptable performance for select2 drop-downs.

SCdF:
Looks good

(another day, another view…)

sglangevin:
This seems to be working well on lg-testing.app. The use case of editing contacts works well, as long as I don't type the search term in too slowly as I've seen this cause the web app to get stuck on a request, though perhaps my internet connection dropped for a second as I'm now having trouble reproducing that. I am able to get the app to freeze momentarily, but it recovers. This happens if I type a search term, then delete some letters, then add more, and do that a few times. Is there anything that we can do to improve this further? If not, then I'm happy to move this to Ready and continue testing in production to see if we run into any issues.

garethbowen:
There are still specific words that will cause your browser to hang trying to sort them all but they're not very common. The only solution I can think of is to get the database to do the filtering and sorting but this isn't possible with couch as it is.

sglangevin:
Sounds good. Let's move this to Ready and see how it goes in production.

PR #10876, opened 2026-04-16
fix(#10847): allow contact search by local phone number format

### LINKED ISSUE #10847: Allow searching contacts by local phone number format without country code prefix
Currently, health workers cannot search for a client in the CHT app using the standard 10-digit mobile number (e.g., `9841XXXXXX`). They must include the `+977` country code prefix every time (e.g., `+9779841XXXXXX`), which is not how phone numbers are typically used or remembered locally.

Allow health workers to search for contacts using the local phone number format without the country code prefix. The search could automatically prepend the default country code defined in `app_settings`, or the matching logic could strip the country code from stored phone numbers before comparison.

A computed `unprefixed_phone` property could be stored on each contact document. However, for existing projects with a large number of contacts already registered, backfilling this property is not practical.

**Additional context**
We don't need to put the country code when entering a phone number in the forms. It takes the default country code automatically.

Image

Related:
- #10846

Thanks to @pawanshub for bringing this up in our converstaion.

megha1807:
## Description

Closes #10847

Currently, contacts can only be found by searching their full international phone number (e.g. `+9779841234567`). Users who type the local format (e.g. `9841234567`) get no results, even though the number refers to the same contact.

This PR fixes that by normalizing local phone number search terms into their international equivalent and generating an additional search request alongside the original. Both the local and normalized forms are searched, so contacts are found regardless of which format the user types.

### Changes
- `shared-libs/search/src/generate-search-requests.js` - when a search term looks like a phone number and does not start with `+`, normalizes it using `default_country_code` from `app_settings` and adds a second search request for the normalized form
- `shared-libs/search/test/generate-search-requests.js` -added 7 unit tests covering local format, international format, missing settings, invalid input, and type-filtered searches
- `webapp/src/ts/services/search.service.ts` - passes `settings` (including `default_country_code`) into the search filters so the search lib can use it

### Example
User searches: `9841234567`
Search now queries both: `9841234567` and `9779841234567`
Contact stored with `+9779841234567` is found 

Existing searches using `+9779841234567` are completely unaffected 

AI Disclosure: I used Claude to understand the codebase structure and explore possible approaches for implementing the local phone number normalization in search. The final implementation, testing, and refinements were done by me, and I verified the behavior with contact search workflows to ensure correctness and backward compatibility.


megha1807:
Hi @witash, just checking in on this PR when you have a moment. I’ve addressed the changes and ensured tests are passing except for the one failing workflow. Could you please take a look and share your feedback? Happy to make any updates if needed. Thanks!

shared-libs/search/src/generate-search-requests.js
```diff
@@ -2,6 +2,8 @@ const _ = require('lodash/core');
 const moment = require('moment');
 
 const MINIMUM_SEARCH_TERM_LENGTH = 3;
+const phoneNumber = require('@medic/phone-number');
+const PHONE_REGEX = /^\+?[\d\s\-().]{7,}$/;
```
witash: would prefer not to use a custom regex to test phone number format, to
1. keep phone number logic out of random places not specifically related to phone numbers
2. not maintain squirrely phone number logic ourselves

there's a comment below whether it needs this regex at all, if it turns it does, at least can we add it the phone-number shared lib as generic function to test possible numbers.

shared-libs/search/src/generate-search-requests.js
```diff
@@ -94,6 +96,22 @@ const freetextRequestParams = (word) => {
   return params;
 };
 
+const getNormalizedPhoneRequest = (word, settings, view) => {
+  if (!settings?.default_country_code) {
+    return null;
+  }
+  if (!PHONE_REGEX.test(word) || word.startsWith('+')) {
```
witash: I wonder if its necessary to check the format first at all?
What if instead just tried to normalize it? line 109 then returns null for any words not matching this regex anyway

witash (approved):
Hi @megha1807 one small comment and I will take a look at the test failures which are likely not related.

Other than that, looks good!

megha1807:
Done @witash -  removed the PHONE_REGEX constant and the regex pre-check. Now relying directly on phoneNumber.normalize() to handle invalid inputs. Thanks for the suggestion!

megha1807:
Hi @witash, thanks again for the review and feedback!

I’ve addressed the suggested changes and all checks are now passing. Please let me know if there’s anything else to update, otherwise feel free to merge when convenient.

Thanks!

megha1807:
Hi @witash, just a gentle ping on this PR in case you had any final thoughts.

Also looping in @binokaryg since you originally opened the issue- this is ready to merge with all checks passing and feedback addressed. Happy to make any further changes if needed. Thanks!

witash (changes requested):
Hi @megha1807 I tested this locally and found some problems
stripping the leading + means that phone numbers stored with the leading + will not be found. Unfortunately the format that phone numbers are stored in is not consistent for all CHT deployments , and cannot assume that the leading + will be omitted; actually it generally will be included (webapp/src/js/enketo/widgets/phone-widget.js:99), although cannot assume.

When sending multiple parameters, search gets the intersection of the different queries
https://github.com/medic/cht-core/blob/58547e47671d5c008852825ace505bafcf80bbc0/shared-libs/search/src/search.js#L121
This means that adding the normalized phone number as a second parameter doesn't work; results must match both the normalized number AND the unnormalized search string, but the expectation is to get results matching either the unnormalized search string OR the normalized phone.

Then that opens up other edge cases and complications. what if searching for a phone number and a name? Ideally it should preserve current behavior where results match both the phone number AND the name...but for the phone number, either the unnormalized string OR the normalized string.

shared-libs/search/src/generate-search-requests.js
```diff
@@ -94,6 +95,20 @@ const freetextRequestParams = (word) => {
...
 
+const getNormalizedPhone = (word, settings) => {
+  if (!settings?.default_country_code || word.startsWith('+')) {
+    return null;
+  }
+  const normalized = phoneNumber.normalize(settings, word);
+  if (typeof normalized !== 'string') {
+    return null;
+  }
+  if (normalized.replace(/^\+/, '') === word) {
```
witash: I don't understand this line? it means that 9779841234567 will not match any results, it has to be either +9779841234567 or 9841234567
phoneNumber.normalize(settings, word); will return false if the number is not normalized, so the check at 103 is enough.

shared-libs/search/test/generate-search-requests.js
```diff
@@ -466,6 +466,80 @@ describe('GenerateSearchRequests service', () => {
...
+      chai.expect(result.length).to.equal(1);
+      chai.expect(result[0].params.key).to.equal('+9779841234567');
+    });
+
+    it('contacts search with international format does not duplicate request', () => {
+      const settings = { default_country_code: '977' };
+      const result = service('contacts', { search: '+9779841234567', settings });
+      chai.expect(result.length).to.equal(1);
+      chai.expect(result[0].params.key).to.equal('+9779841234567');
+    });
```
witash, line 487: for the above comment, please add a test that an international format without a leading '+' still returns a result

shared-libs/search/src/generate-search-requests.js
```diff
@@ -102,9 +117,14 @@ const freetextRequest = (filters, view) => {
...
-    const params = freetextRequestParams(word);
-    return params && { view, params, freetext: true };
+  const requests = words.flatMap((word) => {
+    const normalizedPhone = getNormalizedPhone(word, filters.settings);
+    const searchWord = normalizedPhone || word;
+    const params = freetextRequestParams(searchWord);
+    if (!params) {
+      return [];
+    }
+    return [{ view, params, freetext: true }];
```
witash: not sure if you're still working on this, but it does need to query the original string OR the normalized string, rather than replacing the original search term.

If the original search term is replaced with the normalized phone number, phone numbers in the database that are not normalized would not be matched even when searching for the exact string that is in the database. They should all be in E164 format, but in reality some are not.

More hypothetically, since this is a freetext search on multiple fields, other numeric fields that are not phone numbers would not be matched. I don't think there are any other fields that could be numeric, but adding the original search term is more robust in general.

So that means more substantial changes to shared-libs/search/src/generate-search-requests.js, to allow for union, not intersection, in the specific case of normalized search terms.
Just replacing the search term with the normalized phone number is a tempting workaround since it's much simpler, but I don't think it can work for all cases.

witash (changes requested):

megha1807:
Hi @witash, I've addressed all the feedback from your last review:

Union instead of replacement - search now queries both the original term and the normalized phone number, so contacts stored in any format are found.
No duplicates - if the normalized form equals the original (e.g. already +9779841234567), only one request is made.
Added missing test - added a test for international format without a leading +.
Fixed indentation - corrected lint errors in generate-search-requests.js.

All 70 tests passing. Please take a look when you get a chance, happy to make any further changes!

witash:
Hi @megha1807 
> Union instead of replacement - search now queries both the original term and the normalized phone number, so contacts stored in any format are found.

I don't see where this is implemented in the code.
Testing locally, it looks like its still doing intersection (and therefore not working because no number will match BOTH the normalized and non-normalized formats), am I missing something?

megha1807:
Hi @witash, you're right - I see the issue now. The requests are being added to the same flat array which goes through getIntersection, so results must match both forms, which is impossible.
Could you point me to how union should be structured for freetext+freetext requests? I can see the existing union pattern in getContactsByTypeAndFreetextRequest but that's for type+freetext combinations. I want to make sure I implement it the right way rather than guess. Thanks!

witash:
@megha1807 yes, existing pattern in getContactsByTypeAndFreetextRequest is the right way to implement this for normalized phone numbers.

[this](https://github.com/medic/cht-core/compare/e2d829f13872505b50741bd3e77831503488f730...18047-union-phone-number-variants?expand=1) Is what claude came up with (just an example, not a PR ready commit).

Please also test manually by starting a cht-instance and trying the freetext search from the contacts tab, where you will see it doesn't work with intersection, and you can validate that it does work with union.

sugat009:
Hi @megha1807, doing a pass over the open PRs to keep things current. Where does this one stand for you? @witash left some feedback back in May. No rush at all, just making sure it isn't slipping through the cracks.

megha1807:
Hi @sugat009  - thanks for the ping! Yes, still actively working on this.

The remaining issue is restructuring the requests to use union instead of intersection for the normalized phone number case, following the pattern in `getContactsByTypeAndFreetextRequest`. I'm working through that now and will push an update shortly.

Appreciate your patience!

binokaryg:
I ran the manual verification requested in the [May review](https://github.com/medic/cht-core/pull/10876#issuecomment-4498730094) against PR head `405bd282` on a real instance (CouchDB 4.21 + Nouveau, `default_country_code: 977`, contacts stored as `+977…`):

| Search input | Result |
|---|---|
| `9841232900` (local) | ✅ found — as online admin (Nouveau) **and** offline facility user (local PouchDB) |
| `9779841232900` (intl, no `+`) | ✅ found |
| `+9779841232900` | ✅ found (unchanged) |
| `009779841232900` (00 intl prefix) | ✅ found |
| `984-1232900` (dashed) | ✅ found |
| `09779841232900` / `000977…` | ⬜ not found — correct, libphonenumber rejects these as invalid (0 is a trunk prefix, so `0977…` isn't a valid number) |
| `munic 9841792496` vs `zebra 9841792496` | ✅ AND semantics preserved — the June 22 union restructure resolves the intersection problem |

Replication scoping is also unaffected (offline users only match contacts in their subtree).

Claude found a small edge case worth a look:

Words shorter than 3 chars are ignored (#7288), but `getNormalizedPhone` still runs on them, so with `phone_validation: 'none'` a search like `shrestha 12` generates a `+97712` request that empties the results via intersection. Only attempting normalization when `freetextRequestParams(word)` returned params fixes it.

@megha1807 thanks for working on this. From the reporter's side, the behavior I was hoping for in #10847 all checks out. @witash the union rework landed after your last pass; would you have a moment to re-review?

sugat009 (commented):
Thanks @megha1807, the union implementation is solid now. Verification went beyond the diff: I traced request generation through both executors (offline PouchDB views and online Nouveau via cht-datasource), ran the shared-lib suite and lint on the branch (71 passing, clean), compared generated requests against the merge base for non-phone searches (identical), and ran a live endpoint matrix on an instance built from this branch: the local and no-plus international forms match only through the normalized arm this PR adds, and existing `+977...` searches are unchanged.

One inline comment blocks merge (a master merge whose conflict resolution needs care); the rest are smaller issues, questions, and suggestions.

A few preexisting limitations surfaced while testing (keyed `phone:` terms never normalize, phones stored without the leading `+` are still missed by local-format searches, an unparseable `default_country_code` silently disables the feature, numbers typed with internal spaces split into tokens before normalization); none block this PR, happy to record them on #10847 separately.

shared-libs/search/src/generate-search-requests.js
```diff
@@ -103,8 +118,27 @@ const freetextRequest = (filters, view) => {
     .toLowerCase()
     .split(/\s+/);
   const requests = words.map((word) => {
-    const params = freetextRequestParams(word);
-    return params && { view, params, freetext: true };
+    const originalParams = freetextRequestParams(word);
```
sugat009: **issue (blocking):** this branch needs another master merge: #10851 rewrote these exact lines hours after this branch last merged master, and the conflict resolution order matters.

Master commit 776c51e58 (fix(#10851), Devanagari numerals) changed this line to `freetextRequestParams(normalizeDevanagariNumerals(word))`. When resolving the conflict, the converted word must feed both arms:

```js
const asciiWord = normalizeDevanagariNumerals(word);
const originalParams = freetextRequestParams(asciiWord);
const normalizedPhone = getNormalizedPhone(asciiWord, filters.settings);
```

`phoneNumber.normalize()` returns false for Devanagari digits (verified), so if only `freetextRequestParams` gets the converted word, a local number typed in Devanagari numerals (९८४१२३४५६७) never gets the `+977` union arm, defeating this fix for its primary Nepal use case. A test with a Devanagari-typed phone would pin the composition.

shared-libs/search/src/generate-search-requests.js
```diff
@@ -103,8 +118,27 @@ const freetextRequest = (filters, view) => {
...
+    if (originalParams && normalizedParams) {
+      return {
+        view,
+        union: true,
+        freetext: true,
+        paramSets: [ originalParams, normalizedParams ]
+      };
+    }
+
+    const params = originalParams || normalizedParams;
```
sugat009, line 139: **issue (non-blocking):** the `originalParams || normalizedParams` fallback resurrects words that are deliberately ignored for being under `MINIMUM_SEARCH_TERM_LENGTH` (#7288).

For a word under 3 characters `originalParams` is undefined, but `getNormalizedPhone` still runs, and under `phone_validation: 'none'` with `default_country_code: 977` it accepts digit strings as short as two characters: `normalize('12')` returns `'+97712'`. Verified against the merge base:

- searching `12` used to fall through to the default contact list; it now runs a lone `'+97712'` prefix query instead
- searching `ram 12` used to return the `ram` matches; it now intersects them with the `'+97712'` results, typically zeroing the list

Skipping normalization when `originalParams` is missing restores the old behavior in one line (no real phone number is 1 or 2 digits, so nothing legitimate is lost), and a short-numeric-word test would lock it in. The default `full` validation and `partial` are unaffected.

webapp/src/ts/services/search.service.ts
```diff
@@ -158,7 +160,9 @@ export class SearchService {
     const trackPerformance = this.performanceService.track();
     const searchFn = await this.searchFactoryService.get(this.dbService, this.datasourceService);
     try {
-      const searchResults = await searchFn(type, filters, options, extensions);
+      const settings = await this.settingsService.get();
+      const filtersWithSettings = { ...filters, settings };
```
sugat009, line 164: **question (non-blocking):** @binokaryg this line is the only place settings are injected, so normalization only takes effect in the webapp. The same search library runs the CSV exports (`api/src/services/export/report-mapper.js` and `contact-mapper.js`, receiving the UI's filters verbatim via `api/src/controllers/export-data.js:627`) and the admin app's contact and facility pickers (`admin/src/js/services/search.js`), and neither receives settings.

Verified on a live instance running this PR (contact stored as `+12068880001`, `default_country_code: 1`): the webapp search finds it for `2068880001`, while `GET /api/v2/export/contacts?filters[search]=2068880001` returns zero rows. So a list filtered by a local phone number can show rows its own export is missing.

Neither surface regresses, so this doesn't block the PR. Since #10847 came from your deployment: does the export path matter for that use case, or was this already considered? Trying to gauge whether a follow-up issue is worth filing.

webapp/src/ts/services/search.service.ts
```diff
@@ -215,6 +219,7 @@ export interface Filter {
   search?: string;
   parent?: string;
   subjectIds?: string[];
+  settings?: Record<string, any>;
```
sugat009, line 222: **nit:** `settings` on the exported `Filter` interface reads as caller-settable, but the spread at line 164 always overwrites it, so a caller-supplied value can never take effect. A short comment marking it as internally injected (or leaving it off the public interface) avoids inviting callers to pass their own.

shared-libs/search/test/generate-search-requests.js
```diff
@@ -467,6 +467,134 @@ describe('GenerateSearchRequests service', () => {
...
+
+    it('contacts search with settings but no default_country_code does not add extra request', () => {
+      const settings = { phone_validation: 'full' };
+      const result = service('contacts', { search: '9841234567', settings });
+      chai.expect(result.length).to.equal(1);
+      chai.expect(result[0].params.key).to.equal('9841234567');
+    });
+
+    it('contacts search with local phone that normalizes to same value does not duplicate', () => {
```
sugat009: **suggestion (non-blocking):** three of the new test names say the opposite of what they assert:

- line 520, `... normalizes to same value does not duplicate`: `'2025551234'` with country code `'1'` normalizes to `'+12025551234'`, a different value, and the body asserts the union with both keys, the duplication path.
- line 551, `... does not duplicate when normalized matches original`: `'9779841234567'` normalizes to `'+9779841234567'`, which does not match, and the assertions are a subset of the test at line 489.
- line 584, `... startkey/endkey params (range query path)`: it asserts `params.key`, and the startkey/endkey branch of `makeCombinedParams` (src line 229) is unreachable because `freetextRequestParams` always sets `key` (coverage shows src lines 230 and 231 never execute); the #2445 test at line 354 already pins this path with `deep.equal`.

The only test of the real "normalized equals original, single request" path is the `'+9779841234567'` one at line 481. Renaming the first two and dropping or renaming the third keeps the suite honest; removing the dead branch could be a follow-up, since it predates this PR.

shared-libs/search/test/generate-search-requests.js
```diff
@@ -467,6 +467,134 @@ describe('GenerateSearchRequests service', () => {
       });
     });
 
+    it('contacts search by local phone number searches normalized international format', () => {
```
sugat009, line 470: **suggestion (non-blocking):** coverage additions worth considering for behavior this PR introduces:

- reports: `freetextRequest` is shared, so `generate('reports', { search: '9841234567', settings })` emits a union request (verified by running it), but all 11 new tests are contacts-only
- a multi-word search containing a phone (`'ram 9841234567'`), asserting the per-word AND with a union arm
- a numeric `default_country_code` (shipped configs use the number `1`; only strings are tested)
- a short numeric word with settings, locking in the minimum-length guard from the other comment (written with `phone_validation: 'none'` it doubles as the only validation-variant test)
- executor level: test/search.js never runs a request with both `union: true` and `freetext: true`, so `denormalizeUnionRequest` feeding `queryFreetext` per paramSet with id dedupe is unpinned
- optional: a wdio case alongside tests/e2e/default/contacts/search-contacts.wdio-spec.js for the full settings-to-search wiring

The reports test and the executor test give the most protection per line of test code.

webapp/src/ts/services/search.service.ts
binokaryg: I hadn't considered this path at the time of filing the issue.
I haven't used the export feature much, and last time I heard someone using it, they said that it wasn't working when certail filters were applied.

But the export feature is there, and someone might be using it, on the web app or via API. I won't consider it a deal breaker, but it will be good to support once we have it work for searches.

witash (approved):
tested locally again with online and offline users, union looks good now

andrablaj:
@witash thanks for testing! Can you please merge this PR if it works as expected, so we can release it in 5.3? Thanks!

PR #10884, opened 2026-04-16
refactor(#10883): replace any types in small service files

### LINKED ISSUE #10883: Replace `any` types with proper interfaces in small service files
## Description

Several small service files in the webapp use `any` where proper TypeScript types exist. Replacing these with correct types improves type safety and makes the code easier to follow.

## Scope

The following files need `any` replaced with proper types:

### 1. `webapp/src/ts/services/update-password.service.ts` (31 lines)

Line 23 declares `headers: any`. This can use `HttpHeaders` from `@angular/common/http` instead.

```ts
// Before
const headers: any = {
  "Content-Type": "application/json",
  Accept: "application/json",
  Authorization: "Basic " + window.btoa(username + ":" + currentPassword)
};

// After
const headers = new HttpHeaders({
  "Content-Type": "application/json",
  Accept: "application/json",
  Authorization: "Basic " + window.btoa(username + ":" + currentPassword)
});
```

### 2. `webapp/src/ts/services/user-login.service.ts` (41 lines)

Line 25 has the same `headers: any` pattern. Same fix as above.

### 3. `webapp/src/ts/services/cache.service.ts` (74 lines)

Lines 9-14 define the `caches` array with `any` in three places:

```ts
private caches: {
  docs: any;
  pending: boolean;
  invalidate: any;
  callbacks: any[];
}[] = [];
```

Also, `register(options)` on line 42 has an untyped parameter.

Create an interface for the cache entry and one for the options parameter. Look at how `register` is called across the codebase to determine what `docs`, `invalidate`, and `callbacks` should be typed as.

### 4. `webapp/src/ts/services/extract-lineage.service.ts` (44 lines)

Line 11 has an untyped `contact` parameter and line 16 uses `any` on the result object:

```ts
extract(contact) {
  const result: any = { _id: contact._id };
```

Define a minimal type for the contact parameter (needs `_id` and optional `parent`) and the return type.

## Existing pattern to follow

`webapp/src/ts/services/session.service.ts` already imports and uses `HttpHeaders`:

```ts
import { HttpClient, HttpHeaders } from "@angular/common/http";
// ...
httpOptions = { headers: new HttpHeaders({ Accept: "application/json" }) };
```

Follow this pattern for the header fixes.

## Steps

1. Pick one or more of the files listed above.
2. Replace `any` with the correct type. For headers, use `HttpHeaders`. For the others, define small interfaces in the same file or use existing types from the codebase.
3. Make sure the code still compiles:
   ```bash
   cd webapp && npx ng build
   ```
4. Run the relevant unit tests:
   ```bash
   cd webapp && npx ng test -- --grep "UpdatePasswordService\|UserLoginService\|CacheService\|ExtractLineageService"
   ```
5. Check that no existing functionality is broken.

## References

- [First Time Contributors Guide](https://docs.communityhealthtoolkit.org/community/contributing/first-time-contributors/)
- [Development Workflow](https://docs.communityhealthtoolkit.org/contribute/code/workflow/)
- [Coding Style Guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/)
- [TypeScript Handbook: Everyday Types](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html)

If you have questions about setup or implementation, feel free to comment below.

tmchow:
## Summary

Per #10883, replace `any` types in four small webapp service files with proper TypeScript types. Follows the `HttpHeaders` pattern already used in `session.service.ts`.

## Changes

- **`update-password.service.ts`** and **`user-login.service.ts`**: swap the plain-object `headers: any` cast for `HttpHeaders` from `@angular/common/http`.
- **`cache.service.ts`**: model the cache entry and `register` options with explicit interfaces and generics:
  - `CacheRegisterOptions<T, C>` types `options.get` as `(done: CacheCallback<T>) => void` and `options.invalidate` as `(change: C) => boolean | undefined`.
  - `CacheEntry<T, C>` types `docs`, `pending`, `invalidate`, `callbacks`.
  - `CacheCallback<T>` types both subscriber callbacks and the `options.get` completion.
  - The `caches` field still holds the type-erased `CacheEntry` so `register` can be called with different `T` / `C` from multiple consumers without extra plumbing. Renamed the inner `callback` loop variable to `cb` so it doesn't shadow the outer closure's `callback` parameter.
- **`extract-lineage.service.ts`**: introduce a minimal `ContactLineage` interface (`_id?`, `parent?: ContactLineage`). The walk-up uses a local `next` binding so the chain keeps compiling under `strictNullChecks` (`minified.parent` is typed as `ContactLineage | undefined` otherwise).

## Test plan

- No behavior change; these services are still used exactly the same way from the call sites.
- Under `strictNullChecks` (on for the webapp), the updated files typecheck locally.

Fixes #10883
[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)


sugat009:
@tmchow, it looks like the CI is failing at the compilation step on this PR. Are you still able to take a look at this? We have a few other contributors interested in this area, so let us know your status!

tmchow:
On it. Pushed d0310be1a to fix the compile failure: TS2367 on `cache.service.spec.ts` where the test mocks used numeric `_id: 2` but the new `CacheChange.id` interface (from 8fff93c) expects `string | undefined`. Switched the test data to string IDs (matching production CouchDB docs), which resolves both comparison errors at lines 149 and 180.

Verified locally: ran `ngc` on webapp, 0 TS2367 errors after the fix (was 2 before). Didn't run the karma unit suite here since that needs the monorepo shared packages built, so relying on CI for the full run.

sugat009:
Cool, thanks for the update!
Running the CI...

sugat009:
@tmchow The build and test CI seems to be all passing but the sonarlint seems to have detected a couple of issues. Please check them out.

tmchow:
@sugat009 Both SonarCloud issues pointed at `webapp/src/ts/services/cache.service.ts`:

1. `typescript:S2933` on line 29 : `caches` is only mutated via `push`/`forEach`, never reassigned, so marked it `readonly`.
2. `typescript:S6582` on line 36 : collapsed `cache.invalidate && cache.invalidate(change)` into the optional-chain form `cache.invalidate?.(change)`.

Fixed in b01ba76f4. No behavior changes from this

sugat009 (commented):
Nice changes. A few suggestions and nits inline!

webapp/src/ts/services/cache.service.ts
```diff
@@ -2,23 +2,38 @@ import {Injectable} from '@angular/core';
...
-  }[] = [];
+  private readonly caches: CacheEntry[] = [];
 
   constructor(private changesService:ChangesService) {
     this.changesService.subscribe({
       key: 'cache',
       callback: (change) => {
         this.caches.forEach((cache) => {
-          if (cache.invalidate(change)) {
+          if (cache.invalidate?.(change)) {
```
sugat009, line 39: **issue (non-blocking):** stealth bug fix worth a code comment

The change from `cache.invalidate(change)` to `cache.invalidate?.(change)` is necessary because `invalidate?: (change: C) => boolean` on the new `CacheEntry` interface declares the field optional. Without the `?.` the file no longer compiles under `strictNullChecks` (`TS18048: 'cache.invalidate' is possibly 'undefined'`).

More importantly, this fix has been latent since the AngularJS days. Git blame shows the docstring at lines 51-55 has said "If no invalidate function is provided the cache will never invalidate" since the original `admin/src/js/services/cache.js` (which still exists). But the call site has always been `if (cache.invalidate(change))` with no guard, which would have thrown `TypeError: cache.invalidate is not a function` if any consumer ever omitted `invalidate`. None of the four current consumers (`settings.service.ts`, `user-settings.service.ts`, `user-contact-summary.service.ts`, `contacts.service.ts`) omit it, so the latent crash never triggered, but the runtime contract has been at odds with the docstring for years.

This PR finally aligns the three: docstring, type, runtime. A future maintainer reading just this file in isolation won't know the `?.` is a deliberate runtime fix versus defensive coding. A one-line comment right above this line would help, e.g.:

```ts
// Optional chaining is required: `invalidate` is documented as optional (see register's docstring),
// but the original AngularJS code called it unguarded. Without `?.` the runtime would crash on consumers that omit invalidate.
if (cache.invalidate?.(change)) {
```

webapp/src/ts/services/cache.service.ts
```diff
@@ -39,17 +54,17 @@ export class CacheService {
...
+    const cache: CacheEntry<T, C> = {
       docs: null,
       pending: false,
       invalidate: options.invalidate,
-      callbacks: [] as any[]
+      callbacks: []
     };
 
-    this.caches.push(cache);
+    this.caches.push(cache as CacheEntry);
```
sugat009: **suggestion:** the new generics introduce capability nothing exercises today

The new `register<T = unknown, C = CacheChange>` adds two type parameters, but none of the four current consumers specify either. They all default `T = unknown` (so `cache((err, result) => ...)` still gives them an unknown-typed result, no real improvement over `any`) and never narrow `C`.

Two concerns:

1. **`C` is doing no useful work**, even in principle. It only appears in `invalidate?: (change: C) => boolean`, where `C` is bound to the consumer's closure body, but the actual change values come from PouchDB via `ChangesService` and are not constrained by `C`. So if a future consumer narrows `C` to a stricter type, TypeScript will accept their closure but PouchDB will pass real change events that don't match. False safety. Recommend dropping `C` entirely and hardcoding `change: CacheChange` in the interface.

2. **`T` is theoretically useful but unexercised.** It correctly ties the consumer's `get` emission to their `cache(...)` reception, both of which are under the consumer's control, so narrowing it is honest. But none of the four call sites do that today. Either:
   - **commit to it**: in this PR, also migrate the four `register({...})` calls to `register<TheConcreteType>({...})` so the generic actually delivers value (one type argument per call site, very small diff), or
   - **drop it**: accept that consumers cast at consumption and remove the generic entirely.

Introducing a capability with no live consumers risks the pattern being copied without consumer migration as more services get typed. Picking one direction now is cheaper than untangling later.

webapp/src/ts/services/cache.service.ts
```diff
@@ -2,23 +2,38 @@ import {Injectable} from '@angular/core';
 
 import {ChangesService} from '@mm-services/changes.service';
 
+type CacheCallback<T = unknown> = (err: unknown, result?: T) => void;
+
+export interface CacheChange {
+  id?: string;
```
sugat009: **nitpick:** `id` should be required, not optional

The shape this interface models is the change events emitted by PouchDB's `.changes()` feed. PouchDB always emits `id` as a non-null string. Marking it optional forces consumers to null-guard a field that is never null, or worse, forces the type system into pretending consumers should account for absence. Compare to `doc?:` which is genuinely optional (only present when subscribed with `include_docs: true`).

Suggested:

```ts
export interface CacheChange {
  id: string;
  doc?: { _id?: string; [key: string]: unknown };
  [key: string]: unknown;
}
```

One character change. Easier to start tight; loosening later is non-breaking, tightening later is.

webapp/src/ts/services/update-password.service.ts
```diff
@@ -20,11 +20,11 @@ export class UpdatePasswordService {
    */
   update(username, currentPassword, newPassword): Promise<Object> {
```
sugat009: **suggestion:** parameters are still implicit `any` (also affects `user-login.service.ts:21`)

The PR title says "replace any types" and the headers cast was correctly removed, but the function signature is still:

```ts
update(username, currentPassword, newPassword): Promise<Object>
```

All three are unambiguously `string` based on usage (`window.btoa(username + ':' + currentPassword)`, URL concatenation). Same situation in `user-login.service.ts:21`:

```ts
login(username, password): Promise<Object>
```

Where both parameters are JSON-stringified into a payload.

Suggested fix in this PR (one-liner per file):

```ts
update(username: string, currentPassword: string, newPassword: string): Promise<Object>
login(username: string, password: string): Promise<Object>
```

Issue #10883 stays partially open if these implicit-`any` parameters are left behind in the same files this PR is cleaning.

webapp/src/ts/services/extract-lineage.service.ts
```diff
@@ -1,5 +1,10 @@
 import { Injectable } from '@angular/core';
 
+interface ContactLineage {
```
sugat009, line 7: **chore:** structurally identical to `NormalizedParent` from `shared-libs/cht-datasource`

`shared-libs/cht-datasource/src/libs/core.ts:226` already exports:

```ts
export interface NormalizedParent extends DataObject, Identifiable {
  readonly parent?: NormalizedParent;
}
```

Where `Identifiable` declares `readonly _id: string`. This is the same recursive lineage shape as the new local `ContactLineage`, with two minor differences: `_id` is required (not optional), and fields are `readonly`.

Not blocking for this PR because `extract()` mutates the result via `minified.parent = next`, which `readonly` would forbid, so a direct switch needs a small refactor (recursive immutable construction instead of iterative mutation).

Worth a follow-up issue: as more services get typed under #10883, converging on `NormalizedParent` (or exporting a `Mutable<NormalizedParent>` helper from cht-datasource) would prevent each service from inventing its own minimal lineage interface. The risk of doing nothing is parallel hierarchies that drift over time.

tmchow:
@sugat009 thanks for the thorough review. All 5 addressed in a220091:

1. **cache.service.ts:36 (optional chaining comment):** added an inline comment explaining the runtime contract (`invalidate` documented as optional, original AngularJS unguarded, `?.` fixes the latent crash).

2. **cache.service.ts:65 (generics):** dropped `C` (you're right, it contributes false safety since change values flow from PouchDB independently of the closure type). Kept `T` and migrated all 4 `register({...})` call sites to pass `Record<string, unknown>` (or `Record<string, unknown>[]` for contacts). Generic now delivers value at every call site.

3. **cache.service.ts:8 (id required):** tightened to `id: string`. PouchDB always emits it.

4. **update-password.service.ts:21 + user-login.service.ts:21 (implicit-any params):** annotated all 5 params as `string`.

5. **extract-lineage.service.ts:3 (lineage convergence):** rewrote `extract()` as recursive immutable construction and tightened `ContactLineage` to `readonly _id: string` + `readonly parent?: ContactLineage`, which matches `NormalizedParent`'s shape exactly. Kept it local because `NormalizedParent` is `@internal` in `shared-libs/cht-datasource` and not re-exported from the public surface. Added a code comment flagging the follow-up convergence once the shared type is exposed (or a `Mutable<NormalizedParent>` helper ships).

webapp/src/ts/services/cache.service.ts
```diff
@@ -39,17 +57,17 @@ export class CacheService {
...
+    const cache: CacheEntry<T> = {
       docs: null,
       pending: false,
       invalidate: options.invalidate,
-      callbacks: [] as any[]
+      callbacks: []
     };
 
-    this.caches.push(cache);
+    this.caches.push(cache as CacheEntry);
```
sugat009: **nit:** the `as CacheEntry` cast deserves a one liner.

`T` is erased here so heterogeneous entries can coexist in `caches: CacheEntry[]`. That's intentional, but a reader will wonder if the cast hides a bug. A short comment, or typing the field as `CacheEntry<unknown>[]` so the cast collapses into a structural assignment, would make it clear.

webapp/src/ts/services/extract-lineage.service.ts
```diff
@@ -8,21 +17,13 @@ export class ExtractLineageService {
...
-
-    while (contact.parent) {
-      minified.parent = { _id: contact.parent._id };
-      minified = minified.parent;
-      contact = contact.parent;
-    }
-
-    return result;
+    return contact.parent
+      ? { _id: contact._id, parent: this.extract(contact.parent) as ContactLineage }
```
sugat009: **nit:** the `as ContactLineage` cast is safe but reads as a lie.

The ternary already gates on `contact.parent`, so `extract` can't return `null | undefined` here, but TS can't prove it. Either narrow with overloads, or drop the cast and let `parent` be optional:

```ts
return contact.parent
  ? { _id: contact._id, parent: this.extract(contact.parent) ?? undefined }
  : { _id: contact._id };
```

Current code is correct, just polish.

sugat009 (changes requested):
Most of the feedback is well addressed. Two CI checks are red though:

* **Compile the app**: TS2345 at `update-password.component.ts:68:57` and `:70:53`. The component passes `string | undefined` (from `editUserModel.password` / `currentPassword`) into the now strict service signatures. My bad on the previous round, I asked for `string` without checking the caller side.

  `editUserModel.password` and `editUserModel.currentPassword` are `string | undefined` (lines 34, 35). `validatePasswordFields()` on line 60 returns `Promise<boolean>`, not a type guard, so TS won't narrow across it. Non-null assert at the call sites:

  ```ts
  await this.updatePasswordService.update(username, currentPassword!, newPassword!);
  await this.userLoginService.login(username, newPassword!);
  ```

  (Validation already guarantees presence. `!` is used elsewhere, e.g. `session.service.ts:101`, `duplicate-contacts.component.ts:48`. `as string` works too if you prefer.)

* **SonarCloud**: failing too, but no inline comments on the PR. Could you check the dashboard and apply whatever it flagged?

Could you also run `npm run build-dev` (and ideally `npm run unit-webapp`, `npm run lint`) locally before the next push? Would have caught both call site mismatches in a minute. Two pushes in a row have hit "wait for CI to surface a typecheck error", easier to short circuit that.

Two nits also inline, both optional. Request Changes only for the compile failure.

---

**Test coverage idea (non blocking):** The `cache.invalidate?.(change)` fix has no test. No spec exercises `register` without `invalidate` plus a change event firing. Suggestion:

```ts
it('does not crash and does not invalidate when invalidate is omitted', (done) => {
  const docs = [{ _id: '1' }];
  const cache = service.register({
    get: (callback) => callback(null, docs)
  });
  cache((err, results) => {
    expect(err).to.equal(null);
    expect(results).to.deep.equal(docs);
    expect(() => changesCallback({ id: '1', changes: [{ rev: '1-xyz' }] })).not.to.throw();
    cache((err2, results2) => {
      expect(err2).to.equal(null);
      expect(results2).to.deep.equal(docs);
      done();
    });
  });
});
```

Same synchronous pattern as `'invalidates the cache on doc update'` (lines 101 to 130).

tmchow:
Thanks @sugat009. Pushed 0c312c9, and I ran the compile locally first this time:

- **Compile:** non-null asserted `currentPassword` / `newPassword` at the `update()` and `login()` call sites, per your suggestion. `ngc -p tsconfig.app.json` locally confirms the TS2345s are gone (the only errors left in my local run are `@medic/*` workspace modules I couldn't link without the full monorepo install).
- **extract-lineage:** dropped the `as ContactLineage` cast; `parent` is optional, so `?? undefined` satisfies it.
- **cache.service:** added the one-liner on the widening cast.
- **Coverage:** added the `register()`-without-`invalidate` test you outlined, which should close the new-code coverage gap SonarCloud was flagging. If the dashboard still shows something after this run, point me at it.

webapp/src/ts/services/cache.service.ts
```diff
@@ -39,17 +57,19 @@ export class CacheService {
...
       pending: false,
       invalidate: options.invalidate,
-      callbacks: [] as any[]
+      callbacks: []
     };
 
-    this.caches.push(cache);
+    // T is erased here so heterogeneous entries can share `caches: CacheEntry[]`;
+    // the cast widens CacheEntry<T> to the default CacheEntry, it does not hide a bug.
+    this.caches.push(cache as CacheEntry);
```
sugat009:
```suggestion
    this.caches.push(cache);
```

webapp/src/ts/services/cache.service.ts
```diff
@@ -39,17 +57,19 @@ export class CacheService {
...
+    const cache: CacheEntry<T> = {
       docs: null,
       pending: false,
       invalidate: options.invalidate,
-      callbacks: [] as any[]
+      callbacks: []
     };
 
+    // T is erased here so heterogeneous entries can share `caches: CacheEntry[]`;
+    // the cast widens CacheEntry<T> to the default CacheEntry, it does not hide a bug.
```
sugat009:
```suggestion
    // T is erased here so heterogeneous entries can share `caches: CacheEntry[]`
```

sugat009 (approved):

Issue #10862, opened 2026-04-14
dont run database compaction on every version upgrade


1yuv:
**What feature do you want to improve?**
Currenlty, when users upgrade from one version to another, database compaction is run regardless of if there are any database changes or not or the fragementation value. For larger instances, this adds unnecessary load for the work that may not improve anything impactful.

**Describe the improvement you'd like**
Check if database comapction is required to run or not. Run it only when it is required. 

**Describe alternatives you've considered**
Leave as it is. 

**Additional context**
Recently one large instance was upgraded to 5.1.0 and that ran index creation and database compactionf or few hours. After couple of weeks, we needed to apply 5.1.1 for quick fixes and since 5.1.1 does not have any index or view changes and this is running database compaction for 1 hour already. In scenarios like this where there is no database changes and database compaction is not necessary, we should not trigger database compaction.

sh1vam31:
Hi @1yuv, @andrablaj 

This makes total sense! I checked the codebase and saw that database.db.compact() is currently running unconditionally during the finalize() step.

We can solve this by skipping compaction when there are no design changes, or by querying CouchDB to only run compaction if the database fragmentation ratio exceeds a certain threshold (e.g., >30%).

I would love to work on this performance improvement.

 Could you please assign this issue to me?

ManbirS07:
Hi @1yuv,
I explored the upgrade flow and wanted to confirm my understanding:

Both finalize() and abort() always call cleanup(), which in turn runs database.db.compact() unconditionally and currently, there is no check to determine whether compaction is actually required before executing it.

Also, getFragmentation() computes fragmentation based on DB and view index sizes, but this metric is only used for monitoring and is not part of the compaction decision flow.

I think that we can compare current active ddocs in CouchDB vs target ddocs for the upgrade version like we can fetch target ddocs using upgradeUtils.downloadDdocDefinitions(buildInfo) and current ddocs using ddocsService.getDdocs(database) (ignoring staged IDs)
Then normalize both sides (consider _id, views, nouveau definitions; ignore _rev, deploy_info, etc.) and compare the changes

IF (ddocs/views/nouveau definitions changed)
    run cleanup (compaction required)
ELSE
    evaluate fragmentation before deciding

I wanted to clarify:
Is it desirable that cleanup() always runs during abort()?

Would it be reasonable to:
If no changes, then evaluate fragmentation before deciding to compact?
Is there any recommended fragmentation threshold that should trigger compaction?

[...]

PR #10888, opened 2026-04-16
fix: skip unnecessary database compaction on minor patch upgrades

sh1vam31:
### Fixes Issue #10862

Previously, deployment upgrades unconditionally triggered heavy database compactions (db.compact() and db.viewCleanup()). This caused unnecessary disk I/O load and delays, especially during minor patch updates. This PR optimises the process by skipping compaction unless design documents were altered or database fragmentation is critically high.

### Changes Made:
**api/src/services/setup/utils.js:** unstageStagedDdocs() now returns a list of modified databases. cleanup() uses this list to skip compaction on unchanged databases, triggering compaction only if db.info() yields a fragmentation ratio > 1.2.

**api/src/services/setup/upgrade-steps.js:** Restructured finalise () to pipeline the output string array from the unstageStagedDdocs utility natively into the cleanup() request parameter.

### Mocha Testing:
Updated upgrade-steps.spec.js and utils.spec.js to cover the new conditional compaction logic and ensure background errors are caught safely.
**Verification**
Ran completely green under all API setup integration layers, confirming the database.nouveauCleanup() continues completely uncontested, and unit limits reliably block heavy DB tasks when fragmentation is low.
Covered by 25+ successful tests inside the Mocha test suite.

# Code review checklist

# License


sh1vam31:
Hi @dianabarsan,

I've pushed the fix for #10862.

The logic now safely tracks which databases have altered design documents and bypasses db.compact() and db.viewCleanup() for any unchanged databases, safely falling back to compaction only when the fragmentation ratio crosses the > 1.2 limit.

All new conditional branching and asynchronous catches have been fully covered by unit tests, and Mocha passes flawlessly locally.

Could you kindly review this when you have a moment? 

Please let me know if there are any specific tweaks or changes you'd like me to make before this gets merged.

api/src/services/setup/utils.js
```diff
@@ -49,20 +49,42 @@ const deleteStagedDdocs = async () => {
   }
 };
 
+const isHighlyFragmented = async (database) => {
```
dianabarsan: I don't think we should use high fragmentation here as a driver for our choice. 
I think we should only compact if we indexed views. There already is a function that compares the views between two versions. 
We should definitely not have our own magic number fragmentation evaluation.

api/src/services/setup/utils.js
```diff
@@ -318,8 +341,15 @@ const unstageStagedDdocs = async () => {
       ddocsToSave.push(ddoc);
     }
 
-    await db.saveDocs(database.db, ddocsToSave);
+    if (ddocsToSave.length > 0) {
```
dianabarsan, line 403: We will _always_ have docs to save! Because the ddocs always change, but not necessarily the views within them. So this is not improving anything.

dianabarsan (changes requested):

sh1vam31:
Hi @dianabarsan,

Thank you for the feedback! I've refactored the implementation to address your concerns:

**Removed Fragmentation Logic:** I have completely removed the isHighlyFragmented check and the fragmentation ratio driver (> 1.2) from the compaction logic.

**View-Change Dependent Compaction:** I migrated the areViewsDifferent comparison logic to utils.js so it can be used during the unstaging process. Now, unstageStagedDdocs() checks if views actually changed between the staged and production versions (or if new views were added) before flagging a database as "altered."

**Efficiency:** Databases with no view changes (even if other fields like timestamps or versions changed) now safely bypass the cleanup()/compaction routine.

Both utils.spec.js and upgrade-steps.spec.js are passing green locally. 
Could you kindly take another look? 

Thanks again.

sh1vam31:
Hi @dianabarsan,

Could you please review this PR and let me know if any changes are needed?

dianabarsan:
Sorry for the delay @sh1vam31 . The truth is I have been seriously considering removing the compaction step completely after the upgrade, and I'm running a few safety checks before confirming. 
So this work might need to change from what it's doing now (moving the index comparison checks) to removing the mandatory upgrade compaction completely or replacing it with viewCleanup. 

I'm expecting to have an answer today or tomorrow. 
Appreciate your patience!

sh1vam31:
Hi @dianabarsan 

No problem at all.

Take your time with the safety checks. Removing the compaction step completely sounds like a much simpler and cleaner approach if it's safe to do so.

I will pause work on this branch and wait for your final confirmation. 

Thanks for the update.

dianabarsan (changes requested):
Ok, I think we should just not run compaction. 
So this becomes much simpler, because we just need to remove that call to `db.compact` and everything else can stay the same. viewCleanup can still run because it's not compute-heavy and there's not reason to run it conditionally. 

Can you please make these changes?

sh1vam31:

I have simplified the PR as you suggested! I've removed the manual db.compact() call entirely and updated the logic so that viewCleanup() now runs unconditionally for all databases during an upgrade.

I've also cleaned up the now-obsolete view comparison logic from the unstaging step and updated the unit tests to reflect this simpler approach. All 91 tests are passing.

Ready for another review. 

Thanks for the guidance.

sh1vam31:
Hi @dianabarsan,

All the checks have passed. Kindly review this PR and let me know if any adjustments are needed.

api/src/services/setup/utils.js
```diff
@@ -49,20 +49,16 @@ const deleteStagedDdocs = async () => {
   }
 };
 
+
 /**
- * Runs compaction and view cleanup for every database.
+ * Runs view cleanup for every database.
  */
-const cleanup = () => {
+const cleanup = async () => {
```
dianabarsan: This function is declared async, but nothing is awaited in this function, both calls to viewCleanup and nouveauCleanup are fire and forget.

api/tests/mocha/services/setup/utils.spec.js
```diff
@@ -826,31 +814,62 @@ describe('Setup utils', () => {
 
       sinon.stub(db.medic, 'allDocs').resolves({
         rows: [
-          { doc: { _id: '_design/:staged:medic', _rev: '1', new: true, deploy_info: deployInfoNew } },
+          {
```
dianabarsan: I think these test changes were added when we were running the view comparison. They can be removed.

api/tests/mocha/services/setup/upgrade-steps.spec.js
```diff
@@ -39,6 +39,7 @@ describe('Upgrade steps', () => {
       expect(upgradeUtils.deleteStagedDdocs.callCount).to.equal(1);
       expect(upgradeLogService.setFinalized.callCount).to.equal(1);
       expect(upgradeUtils.cleanup.callCount).to.equal(1);
+      expect(upgradeUtils.cleanup.calledOnceWithExactly()).to.be.true;
```
dianabarsan, line 42: This second assertion is redundant to the one prior. Please only keep one.

api/tests/mocha/services/setup/utils.spec.js
```diff
@@ -21,8 +21,8 @@ const buildInfo = (version, namespace = 'medic', application = 'medic') => ({ ve
 const mockDb = (db) => {
   sinon.stub(db, 'allDocs');
   sinon.stub(db, 'bulkDocs');
-  sinon.stub(db, 'compact');
-  sinon.stub(db, 'viewCleanup');
+  sinon.stub(db, 'viewCleanup').resolves();
+  sinon.stub(db, 'info').resolves({ sizes: { active: 100, file: 110 }});
```
dianabarsan: Is `info` still called?

dianabarsan (changes requested):
Nice. Some minor comments left inline.

sh1vam31:
Thanks @dianabarsan! 
I've updated the cleanup function to properly await the promises and removed all the redundant test code and stubs you pointed out.

Ready for your review.

api/src/services/setup/utils.js
dianabarsan: I believe the message was misunderstood. We want these requests to be fire and forget.

api/tests/mocha/services/setup/utils.spec.js
```diff
@@ -826,16 +813,16 @@ describe('Setup utils', () => {
 
       sinon.stub(db.medic, 'allDocs').resolves({
         rows: [
-          { doc: { _id: '_design/:staged:medic', _rev: '1', new: true, deploy_info: deployInfoNew } },
-          { doc: { _id: '_design/:staged:medic-client', _rev: '1', new: true, deploy_info: deployInfoNew } },
-          { doc: { _id: '_design/medic', _rev: '2', old: true, deploy_info: deployInfoOld } },
-          { doc: { _id: '_design/medic-client', _rev: '3', old: true, deploy_info: deployInfoOld } },
+          { doc: { _id: '_design/:staged:medic', _rev: '1', deploy_info: deployInfoNew } },
```
dianabarsan: I'm not sure about the reasoning behind these changes.

dianabarsan (changes requested):

sh1vam31:
Hi @dianabarsan,

I've implemented the fire-and-forget cleanup pattern and restored the original test assertions in utils.spec.js as requested. I also updated the upgrade-step tests to align with the new background cleanup logic. All unit tests are passing locally.

api/src/services/setup/utils.js
```diff
@@ -49,23 +49,25 @@ const deleteStagedDdocs = async () => {
...
 const cleanup = () => {
-  for (const database of DATABASES) {
-    logger.info(`Running DB compact and view cleanup for ${database.name}`);
-    Promise
-      .all([
-        database.db.compact(),
-        database.db.viewCleanup(),
-      ])
-      .catch(err => {
+  setImmediate(async () => {
```
dianabarsan: I'm not sure I understand the reason for this change. Why the need of using `setImmediate`?

api/tests/mocha/services/setup/utils.spec.js
dianabarsan: I think `info` is not called, so this stub is unnecessary.

api/tests/mocha/services/setup/utils.spec.js
dianabarsan: Can you please revert the new line changes? these are unrelated and just create noise in the PR.

dianabarsan (changes requested):

sh1vam31:
Hey @dianabarsan,

Good catch on `setImmediate`—my intent was to move it off the main thread, but looking closely at the original implementation, it was already a fire-and-forget pattern, so wrapping it wasn't necessary.

I have completely reverted `setImmediate` and restored the original execution pattern in `utils.js `(while keeping the omission of `compact()`). I also updated the tests and removed the unused `info` stub as you pointed out.

Let me know if anything else you catche.

dianabarsan (changes requested):
It looks like the PR has unrelated changes

dianabarsan:
Closing in favor of https://github.com/medic/cht-core/pull/11141 where I cherry picked commits from this branch.

PR #10898, opened 2026-04-21
feat(#8217): add CSS and translation support for ODK guidance hints

### LINKED ISSUE #8217: Support ODK guidance_hints in forms
[Guidance hints](https://xlsform.org/en/#guidance-hints) are collapsible (and collapsed by default) hint texts that can be added to questions in a form.  Normal hints are always displayed and so are always taking up screen space.  This is not great for extensive information that might not always need to be referenced but still is important to have available to the user.  Reference information could be embedded in a guidance hint that the user could reference when needed, but otherwise would not affect their workflow (they would not have to scroll past a bunch of text they are not going to read....).  

These guidance hints are already supported in enketo-core `5.5.2`+ so cht-core 4.x technically already has built-in support.  However, our verison of [pyxform](https://github.com/medic/pyxform) does not support including these in the form xml when converting the xlsx. So, that would need to be upgraded to be able to use the functionality.  Also, cht-core needs a new translation and probably some css tweaking to make stuff look nice.  

I kinda manually compiled an xml for a form with a guidance hint and loaded into the CHT and this is what it looks like currently:

image
image

```.xml
<?xml version="1.0"?>
<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa" xmlns:orx="http://openrosa.org/xforms">
  <h:head>
    <h:title>Enketo Widgets</h:title>
    <model>
      <itext>
        <translation default="true()" lang="en">
          <text id="/enketo_widgets/my_group/name:hint">
            <value form="guidance">This is a guidance hint!</value>
          </text>
        </translation>
      </itext>
      <instance>
        <enketo_widgets id="enketo_widgets" prefix="J1!enketo_widgets!" delimiter="#" version="2023-05-05 00:00:00">
          <my_group>
            <age/>
            <name/>
          </my_group>
          <meta tag="hidden">
            <instanceID/>
          </meta>
        </enketo_widgets>
      </instance>
      <instance id="contact-summary"/>
      <bind nodeset="/enketo_widgets/my_group/age" type="string"/>
      <bind nodeset="/enketo_widgets/my_group/name" type="string"/>
      <bind nodeset="/enketo_widgets/meta/instanceID" type="string" readonly="true()" calculate="concat('uuid:', uuid())"/>
    </model>
  </h:head>
  <h:body class="pages">
    <group appearance="field-list" ref="/enketo_widgets/my_group">
      <input ref="/enketo_widgets/my_group/age">
        <label>Age?</label>
        <hint>this is a regular hint</hint>
      </input>
      <input ref="/enketo_widgets/my_group/name">
        <label>Name?</label>
        <hint ref="jr:itext('/enketo_widgets/my_group/name:hint')"/>
      </input>
    </group>
  </h:body>
</h:html>
```

megha1807:
## Description

Adds styling and translation support for ODK guidance hints in CHT forms.

Guidance hints are collapsible hint texts that allow form authors to includereference information that is hidden by default, so users are not forced toscroll past content they may not need.

Enketo-core already supports rendering guidance hints and pyxform supportwas confirmed working. This PR adds the missing pieces:

- CSS styles for `.or-form-guidance` using CHT brand colors (`$brand-primary-color`)
- `enketo.guidance.hint` translation key added to all 10 language files

Fixes #8217

## Code review checklist
- [x] Internationalised: All user facing text (10 language files updated)


api/resources/translations/messages-en.properties
```diff
@@ -650,6 +650,7 @@ enketo.filepicker.notFound = File {{existing}} could not be found (leave unchang
 enketo.filepicker.waitingForPermissions = Waiting for user permissions.
 enketo.filepicker.resetWarning = This will remove the {{item}}. Are you sure you want to do this?
 enketo.filepicker.toolargeerror = File too large (> {{maxSize}})
+enketo.guidance.hint = Guidance
```
jkuester: Couple of things around the translation:

- The key should be `enketo.hint.guidance.details`
- For this English translation, I think we can start with "More details" since that matches the [enketo-express translation](https://github.com/enketo/enketo/blob/28f4f916b3fecd76ca0aea383761596e830c0be2/packages/enketo-express/locales/src/en/translation.json#L312). (Can always update this in the future and also deployments can customize the translation for their instances.)
- For the translations to other languages, you can use your personal knowledge (if you are conversant in those languages) and/or use AI to generate the translation.  Please mention in a comment on this PR, though, and call out which translations are AI generated.  I am tracking [this issue](https://github.com/medic/cht-docs/issues/2199) regarding AI translations. I think we should be good to generate them and then I will follow the newly proposed process to make a forum post and collect any feedback from the community.

webapp/src/css/enketo/enketo.scss
```diff
@@ -14,3 +14,29 @@
 @import "widgets";
 @import "../../../node_modules/enketo-core/src/sass/core/responsive";
 @import "../../../node_modules/enketo-core/src/sass/formhub/main";
+
+/** guidance hints **/
+.or-form-guidance {
```
jkuester: We are on the right path here.  Can we move this styling into the `medic.less` file and have it in the [`.enketo form` section](https://github.com/medic/cht-core/blob/b965953914d786826ed6160812cc7797d67ae4a9/webapp/src/css/enketo/medic.less#L17)

jkuester: Can we update the styling here to have a dynamic arrow indicating there is more content and can we remove the background color?

Basically what you have now looks like this:

Image

Image

I am hoping for something a little closer to how this looks in the base Enekto Express (does not have to be exactly the same):

Image

Image

Ultimately I believe this is just a standard HTML `details/summary` block, so hopefully it is pretty easy to get it looking nice.

jkuester (commented):
This is a good start!  I have left a few comments asking for changes, but really everything is working as expected and it is just a matter of tweaking the styles/translation. 👍

jkuester (changes requested):

jkuester:
For the record, I used this form to test with:

[guidance_hints.xlsx](https://github.com/user-attachments/files/27180695/guidance_hints.xlsx)

jkuester:
Also, FYI, you will need to merge in the latest from `master` to pick up some new changes to the CI build jobs.

megha1807:
Hello @jkuester,
 Translation updates: English value set to 'More details' matching enketo-express. Translations for ar, bm, es, fr, id, pt, sw were AI-generated. Please review and flag any corrections needed.

webapp/src/css/enketo/medic.less
```diff
@@ -39,6 +39,41 @@
...
       padding-left: 2px;
     }
+
+    .or-form-guidance {
+      summary {
+        cursor: pointer;
+        color: @text-link-color;
+        font-size: 11px;
+        font-style: italic;
+        padding: 3px 0;
```
jkuester:
```suggestion
        padding: 0 0 5px 1px;
```

webapp/src/css/enketo/medic.less
```diff
@@ -39,6 +39,41 @@
...
+      &[open] > summary::before {
+        transform: rotate(90deg);
+      }
+
+      .or-form-guidance-contents {
+        font-size: 12px;
+        color: @text-link-color;
+        font-style: italic;
+        padding: 2px 0 4px 14px;
+      }
```
jkuester:
```suggestion
      &.active {
        color: @gray-ultra-dark;
        font-style: normal;
      }
      &[open] {
        padding-left: 10px;
        border-left: 2px solid @gray-light;
        summary {
          margin-left: -12px;
        }
        summary::before {
          transform: rotate(90deg);
        }
      }
```

webapp/src/css/enketo/enketo.scss
```diff
@@ -14,3 +14,4 @@
 @import "widgets";
 @import "../../../node_modules/enketo-core/src/sass/core/responsive";
 @import "../../../node_modules/enketo-core/src/sass/formhub/main";
+
```
jkuester:
```suggestion
```

jkuester (changes requested):
Okay, this is looking good! I left a few inline suggestions around the styling. (Basically trying to optimize things for viewing larger amounts of content in the hints. So, we don't want the text to be blue or italics, etc...)

---

Another change we need to make is in the [api server code](https://github.com/medic/cht-core/blob/7b9b9d40a753a3a16307f8d5c2912e734b0079d5/api/src/services/generate-xform.js#L145) we need to update that logic to also select all the `details.or-hint` elements and run them through the same processing that we use for `questions` and `hints`.  That will ensure any html/markdown formatting in the guidance hints is rendered correctly in the form.  

---

Finally, I think this is worth adding a new [cht-form integration test](https://github.com/medic/cht-core/tree/423ff478ca257f60189674b555c3dd9da63f6842/tests/integration/cht-form/default) for this behavior.  We can just add a new test file to that directory and test with this form:

[guidance_hints.xlsx](https://github.com/user-attachments/files/27219454/guidance_hints.xlsx)
[guidance_hints.xml](https://github.com/user-attachments/files/27219455/guidance_hints.xml)

webapp/src/css/enketo/medic.less
```diff
@@ -39,6 +39,43 @@
       list-style: inside;
       padding-left: 2px;
     }
+
+    .or-form-guidance {
+      summary {
+        cursor: pointer;
+        color: @text-link-color;
+        font-size: 11px;
```
jkuester:
```suggestion
```

Let's just use the default font size here.  Most CHT users are on mobile devices, so the target here needs to be big enough to easily tap.

webapp/src/css/enketo/medic.less
```diff
@@ -39,6 +39,43 @@
...
+        font-style: italic;
+        padding: 0 0 5px 1px;
+        user-select: none;
+        list-style: none;
+
+        &::before {
+          content: '\25BA  ';
+          display: inline-block;
+          font-style: normal;
+          font-size: 9px;
```
jkuester:
```suggestion
```

api/src/services/generate-xform.js
```diff
@@ -141,7 +141,7 @@ const replaceAllMarkdown = (form) => {
 
   // Now render markdown
   const questions = form.querySelectorAll('span.question-label');
-  const hints = form.querySelectorAll('span.or-hint');
+  const hints = form.querySelectorAll('span.or-hint, details.or-hint');
   questions.concat(hints).forEach((el, index) => {
```
jkuester:
```suggestion
  const richTextElements = form.querySelectorAll('span.question-label, span.or-hint, details.or-hint');
  richTextElements.forEach((el, index) => {
```

I think we can just do a single query here.

api/src/services/generate-xform.js
```diff
@@ -141,7 +141,7 @@ const replaceAllMarkdown = (form) => {
 
   // Now render markdown
   const questions = form.querySelectorAll('span.question-label');
-  const hints = form.querySelectorAll('span.or-hint');
+  const hints = form.querySelectorAll('span.or-hint, details.or-hint');
```
jkuester: Can you add a unit test case for `details.or-hint`?

tests/integration/cht-form/default/guidance-hints.wdio-spec.js
```diff
@@ -0,0 +1,45 @@
...
+describe('cht-form web component - Guidance Hints', () => {
+
+  beforeEach(async () => {
+    await mockConfig.loadForm('default', 'test', 'guidance_hints');
+  });
+
+  it('should display the form title', async () => {
+    const title = await genericForm.getFormTitle();
+    expect(title).to.equal('Guidance Hints');
+  });
```
jkuester:
```summary
```

We should not really need this test.  The form title will not be affected by the hint.

tests/integration/cht-form/default/guidance-hints.wdio-spec.js
```diff
@@ -0,0 +1,45 @@
...
+
+  it('should show guidance hint toggle for question with guidance', async () => {
+    const guidanceToggle = await $('details.or-form-guidance');
+    expect(await guidanceToggle.isExisting()).to.be.true;
+    expect(await guidanceToggle.isDisplayed()).to.be.true;
+  });
```
jkuester:
```suggestion
```

Also do not really need this case. It is a bit redundant since we activate the details sections below.

tests/integration/cht-form/default/guidance-hints.wdio-spec.js
```diff
@@ -0,0 +1,45 @@
...
+
+  it('should expand and show guidance hint text when clicked', async () => {
+    const guidanceToggle = await $('details.or-form-guidance');
+    expect(await guidanceToggle.getAttribute('open')).to.be.null;
+    // Use browser.execute to reliably click the summary element
+    await browser.execute((el) => el.querySelector('summary').click(), guidanceToggle);
```
jkuester:
```suggestion
    await (await guidanceToggle.$('summary')).click();
```

Is there a reason we need `browser.execute` here or can we just use the wdio functions like you did in the next test?

api/resources/translations/messages-bm.properties
```diff
@@ -1226,3 +1226,5 @@ year.plural = sanw
 years = sanw
 yes = awo
 yesterday = kunun
+
+enketo.hint.guidance.details = Kunnafoni surun
```
jkuester: Can you move this key up to be with the other `enketo.` keys?

api/resources/translations/messages-hi.properties
```diff
@@ -1229,3 +1229,5 @@ year.plural = साल
 years = साल
 yes = हाँ
 yesterday = बिता कल
+
+enketo.hint.guidance.details = अधिक जानकारी
```
jkuester: Can you move this up next to the other `enketo.` keys?

api/resources/translations/messages-id.properties
```diff
@@ -1243,3 +1243,5 @@ year.plural = Tahun
 years = tahun
 yes = Ya
 yesterday = kemarin
+
+enketo.hint.guidance.details = Lebih banyak detail
```
jkuester:
```suggestion
enketo.hint.guidance.details =
```

jkuester: Can you move this up next to the other `enketo.` keys?

jkuester (changes requested):
This is great!  Awesome to have working integration tests for this!

I have made some minor requests, but then this should be good to go!  

Also I have created a [forum thread](https://forum.communityhealthtoolkit.org/t/translation-support-request-form-guidance-hint/5607) to review the translations.  Once we get confirmation for all the required languages, we can merge this. 👍

megha1807:
Hi @jkuester,

I’ve addressed the latest review comments:

* updated the styling/font-size changes
* simplified the selector logic in `generate-xform.js`
* added the requested unit test for `details.or-hint`
* removed the redundant integration tests
* updated the click handling in the integration test
* moved the translation keys next to the other `enketo.*` keys

Could you please let me know if there are any additional changes required, or if the PR looks good pending translation confirmation?

Thanks!

Issue #7369, opened 2021-10-08
Update admin app to use Angular latest


dianabarsan:
**Describe the issue**
The admin app uses old, unsupported libraries, that fewer and fewer people are familiar with, or _should get_  familiar with. 
It's riddled with gotchas (ex: https://stackoverflow.com/questions/45980598/ngannotate-warning-stringmap-expected-string-key) and antiquated code. It's the epitome of technical debt. 

**Describe the improvement you'd like**
If we plan on maintaining the admin app, we should bring it into the present and update it to use latest Angular.

**Describe alternatives you've considered**
I think it would be silly to use two different frameworks in our app, so Angular is definitely the way to go.

dianabarsan:
The idea is to reuse as much `webapp` code as possible - we used to have resource sharing in the AngularJS implementation of both apps, and reuse as much of the setup / testing boilerplate as possible. Ideally we would not have any code duplication between webapp and admin.

jkuester:
Do we have to keep this as a separate "app" at all?  My preference would be to just move the admin functionality into webapp.  We can put all the admin-specific functionality in a separate feature module (within webapp) that is [lazy-loaded](https://angular.dev/reference/migrations/route-lazy-loading) only for admin users.  So, we should have maximum convenience/maintainability for reusing existing webapp functionality without impacting the bundle size of the app for normal users.

I _think_ this would essentially leverage the functionality Gareth was tinkering with in https://github.com/medic/cht-core/issues/8865 (but instead of breaking up the existing webapp bundle into multiple modules, we are just keeping a new module separate...).

jkuester:
> move the admin functionality

Also, IMHO we should be very specific about _which_ features we invest the time to move. We do not need them all.  A lot of the admin app functionality is now duplicated/replaced by cht-conf actions.

dianabarsan:
> My preference would be to just move the admin functionality into webapp.

This is how it was before, and I agree that there are some clear advantages to having admin be a subcomponent in webapp. 
I'm not sure if it's worth the architecture shuffle in this first rewrite. I think it's sufficient that we reuse as much as possible and keep the apps as aligned as possible. I agree that further convergence is only guaranteed if they share "Angular". 

My thinking is the task is less complex if we just do a 1-1 replica, with reuse of webapp services. We can tinker with lazy loading in a 2nd iteration.

jkuester:
I am fine with whatever way you determine will be most efficient in terms of development/maintenance effort. 👍   

[...]

PR #10902, opened 2026-04-21
feat(#7369): adding user management, display and authorization

javier-perez:
# Description

This PR contains several features such User Management, Display and roles and permissions.
It also includes i18n support, PouchDB integration

#7369 

# Code review checklist

- [x] AI disclosure: Claude Code was used to integrate i18n and PouchDb

# License


admin-tool/src/ts/modules/authorization/authorization-roles/authorization-roles.component.ts
```diff
@@ -0,0 +1,152 @@
...
+   * of { key, value } pairs for use in the template @for loop.
+   */
+  async ngOnInit(): Promise<void> {
+    this.loadingPageStatus = true;
+
+    try {
+      const rolesMap = await this.settingsService.getRoles();
+      this.roles = Object.entries(rolesMap).map(([key, value]) => ({ key, value }));
+    } catch (error) {
+      console.error('Error fetching roles', error);
```
dianabarsan, line 58: Should we not display an error message here to the user?

admin-tool/src/ts/modules/display/display-date-time/display-date-time.component.ts
```diff
@@ -0,0 +1,149 @@
...
+    this.responseStatus = { state: 'loading' };
+
+    try {
+      await this.settingsService.updateDateTimeSettings({
+        dateFormat: this.dateFormatSelection,
+        dateTimeFormat: this.dateTimeFormatSelection,
+      });
+
+      this.responseStatus = { state: 'success', msg: 'Saved' };
+      setTimeout(() => {
```
dianabarsan: what is the reason for this setTimeout? Is the state not calculating? is there a way we can proceed without it?

admin-tool/src/ts/modules/display/display-languages/display-languages-delete/display-languages-delete.component.less
```diff
@@ -0,0 +1,45 @@
+@import (reference) "../../../../../css/variables.less";
+
+.modal {
```
dianabarsan, line 3: this seems to be quite generic modal styling. any reason it's not in a generic file and used everywhere?

admin-tool/src/ts/modules/display/display-languages/display-languages.component.ts
```diff
@@ -0,0 +1,223 @@
...
+    
+    try {
+      this.languages = await this.languageService.getLanguages();
+      const languageSettings: LanguageSettings = await this.settingsService.getLanguageSettings();
+      this.localeLanguage = languageSettings.locale;
+      this.localeOutgoingLanguage = languageSettings.localeOutgoing;
+      this.savedLocaleLanguage = languageSettings.locale;
+      this.savedLocaleOutgoingLanguage = languageSettings.localeOutgoing;
+    } catch (error) {
+      console.error('Error fetching languages', error);
```
dianabarsan, line 113: please add user-visible error feedback here. Right now if either fetch fails the user sees an empty list with no explanation.

admin-tool/src/ts/modules/display/display-languages/display-languages.component.ts
```diff
@@ -0,0 +1,223 @@
...
+   * Reloads the language list after success.
+   *
+   * @param {LanguageDoc} doc - the language document to disable
+   */
+  async disableLanguage(doc: LanguageDoc): Promise<void> {
+    try {
+      await this.languageService.disableLanguage(doc);
+      await this.ngOnInit();
+    } catch (error) {
+      console.error('Error disabling language', error);
```
dianabarsan, line 159: please add user-visible error feedback here. Right now if either fetch fails the user sees an empty list with no explanation.

admin-tool/src/ts/modules/display/display-languages/display-languages.component.ts
```diff
@@ -0,0 +1,223 @@
...
+   * Reloads the language list after success.
+   *
+   * @param {LanguageDoc} doc - the language document to enable
+   */
+  async enableLanguage(doc: LanguageDoc): Promise<void> {
+    try {
+      await this.languageService.enableLanguage(doc);
+      await this.ngOnInit();
+    } catch (error) {
+      console.error('Error enabling language', error);
```
dianabarsan, line 175: please add user-visible error feedback here. Right now if either fetch fails the user sees an empty list with no explanation.

admin-tool/src/ts/modules/display/display-languages/display-languages.component.ts
```diff
@@ -0,0 +1,223 @@
...
+  /** Currently selected language code for outgoing messages */
+  localeOutgoingLanguage = '';
+
+  /** Default application language code saved in settings, used to display star icons in the accordion */
+  savedLocaleLanguage = '';
+
+  /** Outgoing message language code saved in settings, used to display star icons in the accordion */
+  savedLocaleOutgoingLanguage = '';
+
+  constructor(private languageService: LanguagesService, private settingsService: SettingsService){}
```
dianabarsan: The old admin subscribed to the CouchDB Changes feed (Changes({ filter: ... })) and reloaded on translation/settings changes. The new component reloads only after its own enable/disable calls. If two admin tabs are open, one won't see the other's edits until manual reload. Was this an intentional simplification or missed during migration?

admin-tool/src/ts/modules/users/ts/components/create-user/create-user.component.ts
```diff
@@ -0,0 +1,410 @@
...
+    if (!this.validate()) {
+      return;
+    }
+
+    const contactValid = await this.validateContactInPlace();
+    if (!contactValid) {
+      return;
+    }
+
+    const withinLimit = await this.validateReplicationLimit();
```
dianabarsan: The old component would allow creating users above limit if the submit button was clicked twice. Is this still the case here?

admin-tool/src/ts/modules/users/ts/components/edit-user/edit-user.component.ts
```diff
@@ -0,0 +1,502 @@
...
+        ? this.user.facility_id
+        : [this.user.facility_id]
+      : null;
+
+    const tokenLoginData = (this.user as any).token_login;
+    const tokenLoginEnabled = tokenLoginData
+      ? {
+        active: tokenLoginData.active,
+        expired: tokenLoginData.expiration_date <= new Date().getTime(),
+        expirationDate: tokenLoginData.expiration_date
```
dianabarsan: These `token_login` fields will get populated by the server, should not have "default" values set to them.

admin-tool/src/ts/modules/users/ts/components/create-user/create-user.component.ts
```diff
@@ -0,0 +1,410 @@
...
+  passwordConfirm: '',
+  showPassword: false,
+});
+
+/**
+ * Modal component for creating a new user.
+ * Controlled by the parent via the `visible` input.
+ * Emits `closed` when the modal is dismissed and `userCreated` when a user is created successfully.
+ */
+@Component({
```
dianabarsan: Can the create and edit user components be built into 1? To reduce much of the code duplication?

admin-tool/src/ts/modules/users/ts/components/users-list/users-list.component.ts
```diff
@@ -0,0 +1,95 @@
...
+  selectedUser: Partial<User> | null = null;
+
+  private usersUpdatedSubscription!: Subscription;
+
+  constructor(
+    private authService: AuthService,
+    private usersService: UsersService,
+  ) {}
+
+  ngOnInit() {
```
dianabarsan: Can this be async?

admin-tool/src/ts/modules/users/ts/components/users-list/users-list.component.ts
```diff
@@ -0,0 +1,95 @@
...
+    this.authService.has('can_configure').then((result) => {
```
dianabarsan: Can this be instead implemented as a route guard instead? Right now I think the experience is that the user would see an empty page.

admin-tool/src/ts/modules/users/users-interfaces.ts
```diff
@@ -0,0 +1,36 @@
...
+  email?: string;
+  phone?: string;
+  facility_id?: string | string[];
+  contact_id?: string;
+  inactive?: boolean;
+  roles?: string[];
+  token_login?: {
+    active: boolean;
+    expiration_date: number;
+    login_date?: number;
```
dianabarsan, line 18: I believe we should also have `oidc_username` here.

admin-tool/src/ts/services/delete-user.service.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Service responsible for deleting a user via the CHT API.
+ *
+ * Wraps DELETE /api/v1/users/:username.
+ */
+@Injectable({
+  providedIn: 'root',
+})
+export class DeleteUserService {
```
dianabarsan: Can this be merged with CreateUser and EditUser service into a User service?

admin-tool/src/ts/services/location.service.ts
```diff
@@ -0,0 +1,19 @@
+import { Inject, Injectable } from '@angular/core';
+import { DOCUMENT } from '@angular/common';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class LocationService {
+  dbName = 'medic';
+  path = '/';
+  adminPath = '/admin/';
```
dianabarsan, line 10: I believe this should be temporarily set to `admin-tool`

admin-tool/src/ts/services/select2search.service.ts
```diff
@@ -0,0 +1,217 @@
...
+ *
+ * The datasource facade methods (getPageByType, getByUuid) accept plain strings —
+ * they wrap Qualifier internally. Do NOT pass Qualifier objects here.
+ *
+ * Select2 and jQuery are loaded globally via angular.json scripts.
+ */
+@Injectable({
+  providedIn: 'root',
+})
+export class Select2SearchService {
```
dianabarsan, line 28: It seems that the select2 service is not working: 

Image

dianabarsan (changes requested):
Thanks for all the hard work. I left some comments inline, mostly pointing out some points of improvement and some regressions.

admin-tool/src/ts/modules/authorization/authorization-roles/authorization-roles.component.ts
SigifredoChaconB: Hi Diana, thank you so much for the feedback.

I had some doubts about this error, as I think it's more related to the final product, and we don't have a reference point from the original app to follow for these errors. Where do you think it's best to display the error? We have field errors, for example, or errors when creating a role fails, but I'm not sure if it's ideal to display the "Error fetching roles" error in the same place as the submit action errors. I've attached a screenshot of where I temporarily put it. I wanted to know if you think that's correct or if you'd prefer a better placement.

Screenshot 2026-05-04 at 2 43 42 PM

admin-tool/src/ts/modules/display/display-languages/display-languages.component.ts
SigifredoChaconB: This error is similar to the one mentioned in the previous error. I'd like to add an option, which is the one I just added, where the error is displayed. I'd like to know if you think that's the best option or if it should be somewhere else?

Screenshot 2026-05-04 at 2 44 57 PM

admin-tool/src/ts/modules/display/display-languages/display-languages.component.ts
SigifredoChaconB: I assume you're referring to the "disabling language" error?
This error is similar to the one mentioned in the previous error. I'd like to suggest where I think the error might be displayed, and I'd like to know if you think that's the best option or if it should be somewhere else?

Screenshot 2026-05-04 at 2 45 27 PM

admin-tool/src/ts/modules/display/display-languages/display-languages.component.ts
SigifredoChaconB: I assume you're referring to the "enabling language" error?
This error is similar to the one mentioned in the previous error. I'd like to suggest where I think the error might be displayed, and I'd like to know if you think that's the best option or if it should be somewhere else?

Screenshot 2026-05-04 at 2 45 27 PM

admin-tool/src/ts/modules/authorization/authorization-roles/authorization-roles.component.ts
dianabarsan: I think that displaying the error in place of the list that was supposed to be populated works, so your proposal is on point. 
For submit errors, we sometimes display the error inlined with the submit button, but underneath can work as well.

admin-tool/src/ts/modules/display/display-languages/display-languages.component.ts
dianabarsan: I think `Error fetching languages` in this case is best displayed at the top of the page, First element before "Default application language. Agree with the submit error location.

admin-tool/src/ts/modules/display/display-languages/display-languages.component.ts
dianabarsan: That works! Or as the first element above "xx translations missing". Either works fine, as long as the user has some feedback as to what happened.

admin-tool/src/ts/modules/display/display-languages/display-languages.component.ts
dianabarsan: As above.

admin-tool/src/ts/modules/users/ts/components/user-form/user-form.component.html
```diff
@@ -0,0 +1,391 @@
+<div>
+  <div
+    class="modal fade"
+    [class.in]="visible"
+    [style.display]="visible ? 'block' : 'none'"
+    role="dialog"
+  >
+    <div class="modal-dialog" role="document">
```
dianabarsan, line 8: It looks like the user modal is much much smaller than the modal in the original tool. I think the size should be increased significanty.

admin-tool/src/ts/services/select2search.service.ts
dianabarsan: It looks like this is still not working. Without it, editing or creating users is impossible.

admin-tool/src/ts/modules/users/ts/components/user-form/user-form.component.html
```diff
@@ -0,0 +1,391 @@
...
+                </span>
+              </div>
+              @if (!isEditMode) {
+                <span class="help-block text-muted">{{
+                  "update.password.help" | translate
+                }}</span>
+              }
+              @if (errors.password) {
+                <span class="help-block">{{
+                  errors.password | translate
```
dianabarsan, line 330: When I enter a too small password, the error that I see is: 

```
The password must be at least {{minimum}} characters long.
```

It looks like translation interpolation doesn't work.

admin-tool/src/ts/modules/display/display-languages/display-languages-edit/display-languages-edit.component.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  /** Form model for the language being added or edited */
+  model: Partial<LanguageDoc> = {
+    code: '',
+    name: '',
+    rtl: false,
+  };
+
+  /** Validation errors for the form fields */
+  languageErrors: LanguageValidation = {};
```
dianabarsan, line 43: When I enable/disable a language the whole page reloads and the language accordeon opens as default closed. 
In the old app, the page did indeed reload, but the language accordeon was default open for the language I had just updated. 
This behavior should be preserved.

admin-tool/src/ts/modules/forms/app-forms/app-forms.component.html
```diff
@@ -0,0 +1,72 @@
...
+              } @else if (getFormIcon(form.icon!).content) {
+                <img [src]="getFormIcon(form.icon!).content" />
+              }
+            </div>
+            <div class="col-sm-5">{{ form._id }}</div>
+            <div class="col-sm-6">
+              @if (form.translation_key) {
+                {{ form.translation_key | translate }}
+              } @else {
+                {{ form.title }}
```
dianabarsan: It seems that this needs to be translated. 
I see: 
```
form:reproductive_age_woman[object Object],[object Object]
form:under_2_child[object Object],[object Object]
form:undo_death_reportUndo death report
form:unmute[object Object],[object Object]
```

dianabarsan (changes requested):
I went over a review of the functionality and found a few things that need to be addressed.

admin-tool/src/ts/modules/forms/app-forms/app-forms.component.html
SigifredoChaconB: Hi Diana, thank you so much for the feedback.

Diana, I have some doubts about this comment. I understand it's related to the translations of the App Forms module/view, but I'm not sure if it means the translations might be incorrect?, since below I'm showing you my current view of the App Forms page and how the translations appear:
Screenshot 2026-05-21 at 11 15 19 AM
Screenshot 2026-05-21 at 11 15 36 AM

dianabarsan: I believe that forms can have translations for their properties files, and I had some forms from a configuration that used an translations object there. Let me check exactly what the format was, because I was seeing `[Object Object]` instead of text in the title column. 

for example, I uploaded a form with this properties file: 

```
{
  "title": [
    {
      "locale": "en",
      "content": "Delivery registration"
    },
    {
      "locale": "ne",
      "content": "सुत्केरी विवरण फारम"
    }
  ],
  "context": {
    "place": false,
    "person": true,
    "expression": ""
  },
  "icon": "icon-delivery",
  "hidden_fields": []
}
```

admin-tool/src/ts/services/select2search.service.ts
dianabarsan: Thanks for making the edits. It seems that now the fields get populated, but the search still doesn't work correctly. I am comparing the search results that the admin app yields and the admin-tool most often doesn't return any results. 
Getting this to work is critical, without it, the user's part is not functional.

dianabarsan (changes requested):
Thank you for all the changes. I felt a comment about the dropdown search tool still not working as expected, unfortunately.

dianabarsan (changes requested):
I believe that your most recent changes that tried to fix select2 had an unwanted consequence. They ended up uploading and indexing a the offline freetext search view to CouchDb. That is supposed to be used by offline users only. The admin app is only for online users, and should just use the API search instead.

PR #10906, opened 2026-04-22
feat(ci): automate demo instance deploy #10887

### LINKED ISSUE #10887: Autmate helm deploy of Demo instance during release process
**Describe the issue**
When releasing a new version of cht-core, there is [a series of manual steps](https://github.com/medic/cht-core/blob/ca8e32d5dbd9a521ac382edce3de5abe668ea406/.github/ISSUE_TEMPLATE/z_release_major.md?plain=1#L50) the release engineer must perform to upgrade the demo instance. Besides just being more manual effort, these steps also require that the engineer has a working local environment setup for running things like `kubectl` and `helm` (things not _normally_ required for CHT development).  Also, the engineer has to have the proper access level for the `demo-cht` namespace on EKS (something unlikely to be granted to non-Medic engineers).     

Unless I am missing something, we _should_ be able to automate all of this via a new GitHub workflow that we can manually trigger against a specified tag.  (Another option would be to do it as a new job in the build workflow that is triggered _for  [when `upgrade` completes](https://github.com/medic/cht-core/blob/0a62a6ad80177e0bd56b30f3ba1eb0594efc2d43/.github/workflows/build.yml#L435)`. The main challenge I see with that is we don't want to upgrade the demo instance _for all tags_, since that could cause downgrades when we publish new patches for older releases, etc.  We can always add more automation in the future, but for now it seems easiest to just make this a manual process (similar to the [release notes](https://github.com/medic/cht-core/actions/workflows/release-notes.yml) and we can upgrade the demo instance on-demand.))

To be clear, I think the new GitHub workflow could do the following:

- Use the [`/api/v2/upgrade/stage` endpoint](https://docs.communityhealthtoolkit.org/building/reference/api/#/Upgrade/v2UpgradeStagePost) to stage the upgrade. We will need some kind of polling logic to wait for the staging to finish. Alternatively, we could use the [chtoolbox](https://docs.communityhealthtoolkit.org/building/reference/api/#/Upgrade/v2UpgradeStagePost) `upgrade --stage --follow` command which will call that endpoint and then wait for the upgrade to complete.
- With the tag version of cht-core checked out, go to the `./cht-core/scripts/build/helm` directory.
- Configure kubectl to use the `arn:aws:eks:eu-west-2:720541322708:cluster/dev-cht-eks` context.
- Download the current `values.yaml` data from the demo instance: `helm get values demo-cht --namespace demo-cht > values.yaml`
- Edit the `values.yaml` file so `cht_image_tag` and `chtversion` properties have the new tag value `<major>.<minor>.<patch>`.
- Use helm to upgrade the instance: `helm upgrade demo-cht . --namespace demo-cht -f values.yaml`
- Poll the [/api/v2/monitoring endpoint](https://docs.communityhealthtoolkit.org/building/reference/api/#/Monitoring/v2MonitoringGet) until `version.app` returns the expected value.
- Use cht-conf to upload the configuration from the `cht-core/config/demo` folder to the `demo-cht.dev` server.

As long as we have the proper GitHub secrets setup, I think all of these should be pretty simple to script.

Abhishek9639:
# Description
Adds a new `workflow_dispatch` workflow (`deploy-demo.yml`) that automates the manual demo instance upgrade during the release process. Instead of requiring a release engineer to have local `kubectl`/`helm` setup and EKS access, they can now just trigger this workflow from the Actions tab with a release tag.

The workflow handles:
- Staging the upgrade via `/api/v2/upgrade/stage`
- Polling until staging completes
- Configuring kubectl for the dev EKS cluster
- Downloading current `values.yaml` and updating `cht_image_tag`/`chtversion`
- Running `helm upgrade`
- Polling `/api/v2/monitoring` until the version matches
- Deploying demo config via the existing `deploy-conf` action

Also updated the major and patch release templates to reference the new workflow instead of listing ~10 manual sub-steps.

closes #10887

# Code review checklist
- [x] Readable: Naming, comments and structuring makes the code easy to understand
- [x] Documented: Configuration and user documentation on [cht-docs](https://github.com/medic/cht-docs)

# License


Abhishek9639:
Hi @jkuester, @sugat009 or @dianabarsan,
This PR automates the demo upgrade steps from the release checklist. The workflow covers staging, helm upgrade, monitoring poll, and cht-conf config upload basically replacing all the manual kubectl/helm steps with a single button click from the Actions tab.

One thing to note: the repo will need a few new secrets configured (DEMO_CHT_URL, DEMO_CHT_HOSTNAME, DEMO_CHT_USER, DEMO_CHT_PASS) before the workflow can actually run. The AWS/ECR secrets are already in place.

Happy to adjust anything based on your feedback. Let me know if the polling logic or timeout values need tweaking.
Thanks

Abhishek9639:
Hi @dianabarsan,
Can you please review it?
Thanks

.github/workflows/deploy-demo.yml
```diff
@@ -0,0 +1,158 @@
...
+            exit 1
+          fi
+          echo "TAG=$TAG" >> $GITHUB_ENV
+
+      - name: Checkout code at tag
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ env.TAG }}
+
+      - name: Stage upgrade via CHT API
```
dianabarsan: I'm not so sure about staging. It can potentially take a long time, much longer than 900 seconds. And I'm not super keen of holding a CI worker "hostage" just polling the instance every 30 seconds if we increase the timeout. 

The alternative is to do the upgrade in place, without staging. This would mean that the server can have some downtime while the indexes are being update, but I don't think that's a huge problem.

.github/workflows/deploy-demo.yml
```diff
@@ -0,0 +1,158 @@
...
+      - name: Helm upgrade demo instance
+        run: |
+          helm upgrade "$HELM_RELEASE" . \
+            --namespace "$HELM_NAMESPACE" \
+            -f values.yaml \
+            --wait \
+            --timeout 10m
+        working-directory: scripts/build/helm
+
+      - name: Poll monitoring for version confirmation
```
dianabarsan, line 108: As above, the time taken for the upgrade to finalize can be more than 900 seconds. 
I understand that this is needed because of the config push. 

One solution is to push the config first, while the instance is running the older version, then do the in-place help upgrade and then, optionally, have the job wait for the upgrade to succeed. 

The advantage to this approach is that all the "blocking" manual actions (pushing config, running helm commands) are performed eagerly through the job and have a low risk of failure. If something happens, a troubleshooting dev would unlikely need to run the commands manually. 

In your current implementation, If any one of the steps times out, the dev is left to run the commands manually, which is exactly what we want to avoid.

dianabarsan (changes requested):

.github/workflows/deploy-demo.yml
Abhishek9639: Done!

.github/workflows/deploy-demo.yml
Abhishek9639: Done!

Abhishek9639:
Hi @dianabarsan, 
Thanks for the detailed feedback. I've addressed both review comments:

Removed staging - Dropped the Stage upgrade via CHT API and Poll for staging completion steps entirely. 
The workflow now does a direct in-place helm upgrade instead, which avoids holding the CI worker hostage for a potentially very long staging process.

Config push before helm upgrade - Moved the Deploy demo configuration step to run before the helm upgrade, while the instance is still on the old version. This way all the critical blocking actions (config push + helm upgrade) complete eagerly. The final version-confirmation poll is now non-fatal. It emits a warning instead of failing the job if it times out, since the important work is already done by that point.

Abhishek9639:
Hi @dianabarsan,
Can you please review it?
Thanks

Abhishek9639:
Hi @jkuester or @dianabarsan,

Pushed a fix for all four `secrets-referenced-without-a-dedicated-environment `warnings. Added `environment: demo `on the `deploy-demo` job, so one job-level scope covers every flagged secret. Felt cleaner than adding a zizmor suppression, and matches what `.github/zizmor.yml` says about fixing findings in code where possible.

One thing you'll need to do on the repo side: the workflow now expects a GitHub Environment called `demo`, with `DEMO_CHT_HOSTNAME, DEMO_CHT_USER, DEMO_CHT_PASS, DEMO_CHT_URL, ECR_ACCESS_KEY_ID`, and `ECR_SECRET_ACCESS_KEY` moved onto it from the repo-level secrets. I don't have access to set that up. Let me know if you'd rather call it something else and I'll update the workflow.

Thanks

Issue #10875, opened 2026-04-15
Flaky api integration tests: transitions


dianabarsan:
**Describe the issue**
Example run: https://github.com/medic/cht-core/actions/runs/24440636603/job/71406458766#step:16:3866

```
  4) Transitions public_form
       when false, reports from unknwon sources should not be accepted:

      AssertionError: expected [ …(2) ] to have a length of 1 but got 2
      + expected - actual

      -2
      +1
      
      at Context.<anonymous> (/home/runner/work/cht-core/cht-core/tests/integration/transitions/public-form-transitions.spec.js:315:31)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)

  5) transitions
       should run all sync transitions and all async transitions:

      AssertionError: expected 11 to equal 10
      + expected - actual

      -11
      +10
      
      at /home/runner/work/cht-core/cht-core/tests/integration/transitions/sentinel-api-transitions.spec.js:502:50
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)

  6) transitions
       should not crash services when transitions are misconfigured:

      AssertionError: expected 1 to equal +0
      + expected - actual

      -1
      +0
      
      at /home/runner/work/cht-core/cht-core/tests/integration/transitions/sentinel-api-transitions.spec.js:702:50
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)

```

Multiple tests have failed during this run, so I am suspecting something was up with the machine.

rsmihir389:
Hey — I've been seeing this exact pattern a lot. Flaky tests like this usually cost teams $30-40 per rerun in CI time + developer context-switching.

I built a tool that detects flaky tests and shows the dollar cost: [kleore.com](https://www.kleore.com). Here's what a report looks like on a real repo: [sample report](https://www.kleore.com/report/cmn0knika00x7dc29p23yyha6).

60-second install as a GitHub App, free tier shows the cost. Might help here.

witash:
I think these are detecting a race condition in the code, not just the tests.

after transitions are applied to a document, the list of transitions that have run is added to the infodoc.
https://github.com/medic/cht-core/blob/8c0cefbd0bafefffecc2a4cb9f8ca29f703db491/shared-libs/transitions/src/transitions/index.js#L260
the modified document is saved, and if successful, the infodoc with the updated list of transitions is saved.

[...]

PR #10910, opened 2026-04-23
fix(#10875): race condition for sync transitions

witash:
# Description

To avoid a race condition in transitions between sentinel and api, save the infoDoc, which contains a which transitions have already run, before saving the medic document for synchrnonous transitions.
This means that the changes feed, and therefore sentinel, will not see the change until after the infodoc has saved which transitions have already run.

This PR leaves async transitions as is, where the infoDoc is saved after the medic document.

More disucssion [here](https://github.com/medic/cht-core/issues/10875)

# Code review checklist

- [ X ] UI/UX backwards compatible: Test it works for the new design (enabled by default). And test it works in the old design, enable `can_view_old_navigation` permission to see the old design. Test it has appropriate design for RTL languages. 
- [ X ] Readable: Concise, well named, follows the [style guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/)
- [ X ] Documented: Configuration and user documentation on [cht-docs](https://github.com/medic/cht-docs/)
- [ X ] Tested: Unit and/or e2e where appropriate
- [ X ] Internationalised: All user facing text
- [ X ] Backwards compatible: Works with existing data and configuration or includes a migration. Any breaking changes documented in the release notes.
- [ X ] AI disclosure: Please disclose use of AI per [the guidelines](https://docs.communityhealthtoolkit.org/community/contributing/ai-guidelines/).


shared-libs/transitions/src/transitions/index.js
```diff
@@ -235,7 +232,7 @@ const canRun = ({ key, change, transition }) => {
  * did nothing and saving is unnecessary.  If results has a true value in
  * it then a change was made.
  */
-const finalize = ({ change, results }, callback) => {
+const finalize = ({ change, results, saveInfoDocFirst = false }, callback) => {
```
dianabarsan, line 238: Nice one! I honestly thought this was solved and we were always saving the infodocs first. 
Since you're here, I wonder if we can somehow change the path for how the code works when docs are changed in bulk. Can we skip `saveDoc` entirely since the docs are saved at the end anyway and save the infodocs in bulk as well?

witash: > Can we skip saveDoc entirely since the docs are saved at the end anyway and save the infodocs in bulk as well?

Maybe, but some transitions seem to rely on previous documents having been already saved. I could get more details, but long story short, several tests fail when trying to save in bulk, that seems to be because of dependencies that some transitions are run and saved in order.

It might still be worth looking into, but its not a "while we're changing this code anyway" type of change, it would expand the scope of this PR a lot.

witash:
@dianabarsan what do you think about merging this as is?
the locking option is left open on the [issue discussion](https://github.com/medic/cht-core/issues/10875) but I'm trying to sell this simpler workaround. Do you think we review and merge this or commit to the locking option?

dianabarsan (changes requested):
Sorry @witash . I don't think we should proceed with this change, and we should implement some sort of locking. I think this would just be moving the race condition somewhere else - because we would absolutely need to revert the infodoc change if the doc save fails, and do another doc save to trigger another "check" by sentinel. 

If you can add a "revert" for the infodoc in case doc save fails, I think this might work.

witash:
replaced by #11135

PR #10914, opened 2026-04-23
fix(#10912): replication key collection

### LINKED ISSUE #10912: Authorization context drops duplicate replication-key matches because `Set.add()` only adds one id
**Describe the bug**

While going through `api/src/services/authorization.js` I noticed when `findContactsByReplicationKeys()` resolves a replication key to multiple document IDs, only the first matching ID is added to `docIds`. The code uses `docIds.add(...keys)`, but Set.add() accepts a single argument, so additional IDs are ignored.

**To Reproduce**

1. Stub `medic-client/contacts_by_reference` to return two or more rows for the same replication key.
2. Call `getScopedAuthorizationContext()` with a doc whose `docsByReplicationKey.key ` contains that replication key.
3. Inspect the `db.medic.allDocs()` call.
4. Observe that only the first matching document ID is included.

**Expected behavior**

All matching document IDs for a replication key should be included in the set, so the authorization context reflects every match.

**Actual behaviour**

Only the first matching ID is kept, and the rest are dropped silently.

blenbot:
# Description

Fixes: #10912 

# Code review checklist

# License

### DIFF
--- a/api/src/services/authorization.js
+++ b/api/src/services/authorization.js
@@ -448,7 +448,7 @@ const findContactsByReplicationKeys = (replicationKeys) => {
       for (const replicationKey of replicationKeys) {
         const keys = result.rows.filter(row => row.key[1] === replicationKey).map(row => row.id);
         if (keys.length) {
-          docIds.add(...keys);
+          keys.forEach(key => docIds.add(key));
         } else {
           docIds.add(replicationKey);
         }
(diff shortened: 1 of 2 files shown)


dianabarsan (changes requested):
Nice catch! Can you please add some unit testing around this?

dianabarsan (approved):
Lovely! Thanks for fixing this!

PR #10917, opened 2026-04-24
fix(#8226): skip form pages where all contents are hidden

### LINKED ISSUE #4543: Marking inputs field's appearance' as 'hidden' in person-create form generates empty pages
_Description_.
When one tries to hide the `inputs` section of a `person-create` form by placing the appearance `hidden`, the `inputs` section content gets hidden but the inputs inner groups are not hidden (see xls screenshot below). Several empty pages of all the groups embedded in the `inputs section` are seen as shown below:

_xls screenshot_
screen shot 2018-05-11 at 1 51 49 pm
_empty pages screenshot_
screen shot 2018-05-11 at 2 25 46 pm

One has to click next till all the empty pages (hidden inputs groups) have been passed over to access the form contents.
Using `hidden` in person-edit form works well and hides the entire section without the empty pages but fails in `person-create` form which generates the empty pages as described above.

**Steps to reproduce**:
 - _Edit the `person-create.xlsx` xls under [here](https://github.com/medic/medic-projects/blob/master/muso/forms/contact/person-create.xlsx) and replace the `field-list hidden` with only `hidden` under the inputs group as should be expected ideally_
 - _Upload the form to a test instance (version 2.14.1) and view the form_
 - _You'll see several empty pages before you reach the actual person-create section with content_
 

**What should happen**:
 - _The person create fields should be loaded well as one page with all the fields_

**What actually happens**:
 - _5 empty pages appears before you reach the actual form fields as shown in the screenshot above giving a total of 6 pages (5 empty pages and 1 page with the form contents)_

_URL to a sample person-create form empty page https://muso-mali.app.medicmobile.org/medic/_design/medic/_rewrite/#/contacts/b975b334-154e-4f8c-a028-4c635d78e64e/add/person_

**Environment**:
- Instance: _(muso-mali.app.medicmobile.org)_
- Browser: _(Google Chrome)_
- Client platform: _(MacOS)_
- App: _(webapp)_
- Version: _(2.14.1)_

**Other**:
The partially solving workaround I've done is using `field-list hidden` under appearance. This minimizes the number of empty hidden pages to only one before the form fields appears.

But ideally there shouldn't be empty pages before the populated for usability purposes.

The reason why I wanted to pull the logged in user's contact document was to get the logged in user's contact place-type so as to use it to hide a field.
If the currently logged in user was a CHV, the role field should be hidden as described in this ticket  https://github.com/medic/medic-projects/issues/3999

cc: @derickl @abbyad

### LINKED ISSUE #8226: Do not display form pages for a "hidden" group

Basically, what I am trying to achieve is to support loading contact data via the db-object-widget into a "hidden" top-level group.

It is possible to use the `select-contact` functionality to transitively load data about a tangentially related contact into a form. For example, if you want to get the `name` value of the _parent_ of the current contact in context, you could have a group like this:

type | name | label::en | appearance | calculation
-- | -- | -- | -- | --
begin_group | parent_data | NO_LABEL | field-list |  
string | _id | NO_LABEL | select-contact | …/input/contact/parent
calculate | name | NO_LABEL |   | ""  
end_group

Unfortunately, doing it this way will still cause a question to be displayed in the form for `parent_data/_id` (which is not what we want if we are just trying to transitively load data from the parent).  

Adding the `hidden` appearance to the `parent_data` group (or both fields) will prevent the fields from actually showing in the form, but you still end up with an empty page in the form (caused by the group).  

You cannot always put the `select-contact` load inside the `inputs` group because of https://github.com/medic/cht-core/issues/8225.

You cannot use `calculate`s for all the fields in the `parent_data` group. `_id` needs to have `appearance` values (for the db-object-widget to see), but `appearance` values are not set on `calculate` fields.

I _think_ the most straight-forward way to achieve what I am looking for is to just not render (skip) any pages where everything is `hidden`.  

If the db-object-widget could work with `calculate` fields, we probably would not need to do anything here since groups with nothing but calculates are already not rendered.

megha1807:
## Description

Fixes #8226
Fixes https://github.com/medic/cht-core/issues/4543

When a `field-list` group has the `hidden` appearance (or all its contentsare hidden), Enketo still renders an empty page for that group. This isproblematic when using the `db-object-widget` to load contact data into a hidden top-level group.

This fix adds a new widget (`HiddenFieldList`) that matches groups with both`field-list` and `hidden` appearances and adds the `disabled` class to them.This causes Enketo to skip the page during form navigation, consistent with how Enketo handles non-relevant/disabled groups.

## Changes
- Added `webapp/src/js/enketo/widgets/hidden-field-list.js` — new widget
- Registered the widget in `webapp/src/js/enketo/widgets.js`
- Added unit tests in `webapp/tests/karma/js/enketo/widgets/hidden-field-list.spec.ts`


jkuester (approved):
Very nice!  Thank you!  

After more tinkering with your branch, I confirmed that we can basically just match all groups and not require the `field-list` appearance to also be set.  This means that any top-level group with the `hidden` appearance will automatically be disabled.  

I have also confirmed that these changes resolve the contact form issue in https://github.com/medic/cht-core/issues/4543.  🎉 

I updated the db-object-form e2e test so that the test form leverages this functionality (and should fail if we ever regress).

---

The one caveat with this is that it only works if the form contains a `relevant` expression.  Otherwise, the Enketo logic ignores the `disabled` class because it bypasses all the relevancy checks. I think this is acceptable (I will open a docs PR to note it). Most non-trivial forms already have `relevant` expressions and so it should work fine anyway.  It is even enough to just add `relevant: boolean(true)` to the group along with the `appearance: hidden`. (Though you cannot just add `relevant: true` because Enekto detects that as a "static" expression and still bypasses the `disabled` checks....)

megha1807:
Thanks for the review and the selector update, @jkuester! Happy the fix also resolves #4543. Ready to merge whenever you are.

PR #10918, opened 2026-04-24
feat(#10907): add accent color and styling support for UI extensions

### LINKED ISSUE #10907: Update UiExtensionTabComponent styling in `webapp`
Currently when displaying the UI Extension tab in the webapp the styling is incomplete: 

Image

Specifically, the tool-bar color is white and there is no padding around the web component content. We should also support font-awesome icons in addition to the current resource-based icons.

### Padding

I think we just need to add a bit of CSS (or an existing class?) to [our element](https://github.com/medic/cht-core/blob/9415abb617334d7d3e910967e66973ec30365e6b/webapp/src/ts/modules/ui-extensions/ui-extensions-tab.component.html#L12) where the web component is being injected so that it gets the same kind of padding that the content on the other tabs gets.  

### Tool-bar color

We need to support custom colors for the tool-bar when a UI Extension is selected.  By default we should use the same gray color used on the "About" and "Training Materials" tabs.  However, we should also support setting a custom color via the `accent_color` property in the UI Extension properties.  App devs can configure `accent_color` to be any valid CSS color value. 

We will also need to update the [cht-conf code](https://github.com/medic/cht-conf/blob/badcc659e2e33f9eab98dca664e4a9b4830299db/src/lib/upload-ui-extensions.js#L11) to support uploading the new `accent_color` property. Create a new PR targeting https://github.com/medic/cht-conf/tree/10224-ui-extensions  as the base branch.

YASHSHARMAOFFICIALLY:
## Description

Closes #10907

Adds styling and theming capabilities to UI extensions, enabling extension authors to customize the visual appearance of their tabs within the CHT webapp.

### Changes

- **Accent color support**: Extensions can now specify an `accent_color` property to customize the toolbar background color for their tab.
- **FontAwesome icon support in sidebar**: The sidebar menu now differentiates between FontAwesome icons (prefixed with `fa-`) and custom resource icons.
- **Page layout**: Added proper CSS rules for `ui-extensions` views including toolbar background color inheritance and page padding.
- **Route data**: Added `{ tab: 'ui-extensions' }` to the route definition so the correct body class is applied for CSS scoping.

## Test Plan

- [x] Unit tests added for accent color application on toolbar
- [x] Unit tests added for missing accent color
- [x] Unit tests updated for FontAwesome vs resource icon detection in sidebar

### DIFF
--- a/webapp/src/css/inbox.less
+++ b/webapp/src/css/inbox.less
@@ -57,7 +57,8 @@ body {
     &.trainings,
     &.testing,
     &.user,
-    &.privacy-policy {
+    &.privacy-policy,
+    &.ui-extensions {
       .tool-bar {
         background-color: @top-header-color;
 
@@ -94,10 +95,16 @@ body {
     &.about,
     &.testing,
     &.error,
-    &.privacy-policy {
+    &.privacy-policy,
+    &.ui-extensions {
       .page {
         padding-left: calc(@tab-navbar-size + 15px);
       }
+    }
+
+    &.ui-extensions .page {
+      padding: 15px;
+      padding-left: calc(@tab-navbar-size + 15px);
     }
   }
 }
--- a/webapp/src/css/inbox.less
+++ b/webapp/src/css/inbox.less
@@ -101,11 +101,6 @@ body {
         padding-left: calc(@tab-navbar-size + 15px);
       }
     }
-
-    &.ui-extensions .page {
-      padding: 15px;
-      padding-left: calc(@tab-navbar-size + 15px);
-    }
   }
 }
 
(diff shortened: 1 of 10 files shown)


YASHSHARMAOFFICIALLY:
Fixex #10907

jkuester (approved):
Very nice!  Simple, clean, and effective!  👍  I pushed a few minor cleanups here and as long as I did not break the CI, this should be good to go!  🚀 

(Also, I pushed a commit to cht-conf to allow the `accent_color` property when uploading.)

YASHSHARMAOFFICIALLY:
@dianabarsan  @andrablaj  Do let me know if any further change needed

PR #10785, opened 2026-03-31
feat(#10674): add app_drawer_tab UI extensions support to sidebar menu

vikrantwiz02:
# Description

Adds support for displaying `app_drawer_tab` UI extensions in the sidebar menu ([#10674](https://github.com/medic/cht-core/issues/10674)).

- Loads `app_drawer_tab` extensions via `UiExtensionsService.getPropertiesByType('app_drawer_tab')` on component init
- Maps each extension to a sidebar option: `routerLink: ui-extensions/${id}`, `translationKey: title`, `resourceIcon: icon`
- Introduces a `mergedOptions` array (moduleOptions → UI extensions → secondaryOptions) so extensions appear between core module tabs and secondary options
- Extensions without a `title` are filtered out; role-based filtering is handled by `UiExtensionsService`
- Switches `sidebar-menu.component.spec.ts` to `MatIconTestingModule` so custom SVG icons in the template don't throw during tests

Fixes #10674

Depends on Issue #10670 (merged into `10224-ui-extensions`).

# Code review checklist

- [] UI/UX backwards compatible: Test it works for the new design (enabled by default). And test it works in the old design, enable `can_view_old_navigation` permission to see the old design. Test it has appropriate design for RTL languages. 
- [x] Readable: Concise, well named, follows the [style guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/)
- [ ] Documented: Configuration and user documentation on [cht-docs](https://github.com/medic/cht-docs/)
- [x] Tested: Unit and/or e2e where appropriate
- [ ] Internationalised: All user facing text
- [x] Backwards compatible: Works with existing data and configuration or includes a migration. Any breaking changes documented in the release notes.
- [ ] AI disclosure: Please disclose use of AI per [the guidelines](https://docs.communityhealthtoolkit.org/community/contributing/ai-guidelines/).

# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

PR #10921, opened 2026-04-24
feat(#10901): UI extension support to legacy header menu

### LINKED ISSUE #10901: Add `webapp` support for showing `app_drawer_tab` in legacy header menu
https://github.com/medic/cht-core/issues/10674 (resolved with https://github.com/medic/cht-core/pull/10785) added support for showing the  `app_drawer_tab` extensions in the sidebar menu.  However we also need to add support for showing them when Ye Olde UIe is enabled (the user has the `can_view_old_navigation` permission):

Image

This is how the UI currently looks for the new sidebar:

Image

We should update the ordering/etc of the old header menu to match what we have done with the new sidebar. (Move the menuitem options down below the sync status, then list the UI extensions (if any) then the rest of the secondary options.  I think this will require some minor refactoring of the [header component](https://github.com/medic/cht-core/blob/bed4546523931f90dc92c93bc6e55a1c3a3de495/webapp/src/ts/components/header/header.component.html#L96) since it looks like everything is currently just hardcoded in there.  

At this point (since we want to remove this old UI in the next major release) I don't want to get into too much cleanup here, but lets do the minimum necessary to be able to dynamically add new menu items for the UI Extensions while keeping the code flow consistent.

vikrantwiz02:
# Description

Adds support for showing `app_drawer_tab` UI extensions in the legacy header menu (enabled via `can_view_old_navigation` permission), matching the behavior already implemented for the new sidebar in #10785.

Changes:
- Legacy header dropdown now dynamically loads and renders `app_drawer_tab` UI extensions between the module nav items and secondary options (Training materials, About, etc.)
- Reordered legacy header dropdown to match the new sidebar layout: Storage Info → Sync → module items → UI extensions → secondary options
- Fixed icon handling in `UiExtensionsService` and `SidebarMenuComponent`: `resource_icon` now correctly maps to the resource image pipe, and `icon` maps to a font-awesome class (was previously both mapped to `resource_icon`)
- Added `resource_icon` to the `UiExtensionProperties` interface
- Fixed `UiExtensionsService.init()` to not cache an empty extension list when called before the user is authenticated, preventing extensions from being missing until a hard refresh

Fixes #10901

image

image

# Code review checklist

# License


jkuester (approved):
Very nice!  I pushed a couple minor changes mostly around adding a default icon (`fa-question-circle`) that matches some code I found elsewhere.  I have tested the none/icon/resource_icon for both the existing and legacy UI. 👍 

As long as I did not break anything in the CI, this should be good to go!

PR #10922, opened 2026-04-24
feat(#10904): route file attachments to correct sub-docs in report forms

### LINKED ISSUE #10904: sub-contact photo capture: Reports core routing + unit tests
**Describe the issue**
child issue of [photo capture for sub contacts](https://github.com/medic/cht-core/issues/10700) covering:

* adaptable from [sibling](https://github.com/medic/cht-core/issues/10903)

YASHSHARMAOFFICIALLY:
## Summary

Closes #10904

When a report form contains `[db-doc="true"]` sub-documents with binary/file fields, attachments are now routed to the owning sub-document instead of always attaching to the main report doc.

### Changes

- **`resolveOwnerDoc()`** — walks up the XML tree from a binary node to find the nearest `[db-doc="true"]` ancestor, returning the corresponding prepared sub-doc (or the main report doc as fallback)
- **`findBinaryNodeByFilename()`** — locates the `[type=binary]` XML node matching a FileManager filename so its position in the tree can determine the owner
- **FileManager file routing** — uploaded files are now attached to the correct owner doc based on their position in the XML form
- **Inline binary blob routing** — base64-encoded binary fields inside sub-docs are attached to those sub-docs

### Test plan

- [x] Unit tests pass (2652 passing)
- [x] Lint passes
- [ ] Binary field in main doc attaches to main report doc
- [ ] Binary field inside `[db-doc="true"]` attaches to that sub-doc
- [ ] FileManager file upload inside sub-doc routes correctly
- [ ] Existing report forms without sub-docs are unaffected

### DIFF
--- a/webapp/src/ts/services/enketo.service.ts
+++ b/webapp/src/ts/services/enketo.service.ts
@@ -362,6 +362,19 @@ export class EnketoService {
     return form;
   }
 
+  private findBinaryNodeByFilename($record, filename: string) {
+    let match = null;
+    $record
+      .find('[type=binary]')
+      .each((_idx, element) => {
+        if ($(element).text() === filename) {
+          match = element;
+          return false; // break
+        }
+      });
+    return match;
+  }
+
   private xmlToDocs(doc, formXml, xmlVersion, record) {
     const recordDoc = $.parseXML(record);
     const $record = $($(recordDoc).children()[0]);
@@ -481,16 +494,45 @@ export class EnketoService {
     }
     doc.hidden_fields = this.enketoTranslationService.getHiddenFieldList(record, dbDocTags);
 
+    // Build a lookup map: sub-doc _couchId -> prepared doc object
+    const subDocById = new Map<string, any>();
+    docsToStore.forEach(subDoc => subDocById.set(subDoc._id, subDoc));
+
+    // Resolve the owner document for a given XML element by walking
+    // up to the nearest [db-doc="true"] ancestor. Falls back to the
+    // main report doc when the element is not inside a sub-doc.
+    const resolveOwnerDoc = (element) => {
+      let node = element.parentNode;
+      while (node && node !== recordDoc) {
+        if (node._couchId && subDocById.has(node._couchId)) {
+          return subDocById.get(node._couchId);
+        }
+        node = node.parentNode;
+      }
+      return doc;
+    };
+
+    // Route FileManager files to the correct owner doc.
+    // For each file, find the [type=binary] node whose text matches
+    // the filename, then resolve the owner from its position in the
+    // XML tree.
     FileManager
       .getCurrentFiles()
-      .forEach(file => this.attachmentService.add(doc, `user-file-${file.name}`, file, file.type, false));
+      .forEach(file => {
+        const ownerDoc = resolveOwnerDoc(
+          this.findBinaryNodeByFilename($record, file.name) ?? $record[0]
+        );
+        this.attachmentService.add(ownerDoc, `user-file-${file.name}`, file, file.type, false);
+      });
 
     const attachLegacyFile = (elem, file, type, alreadyEncoded) => {
+      const ownerDoc = resolveOwnerDoc(elem);
       const xpath = Xpath.getElementXPath(elem);
+      const formId = ownerDoc === doc ? doc.form : ownerDoc.type;
       // replace instance root element node name with form internal ID
       const filename = 'user-file' +
-        (xpath.startsWith('/' + doc.form) ? xpath : xpath.replace(/^\/[^/]+/, '/' + doc.form));
-      this.attachmentService.add(doc, filename, file, type, alreadyEncoded);
+        (xpath.startsWith('/' + formId) ? xpath : xpath.replace(/^\/[^/]+/, '/' + formId));
+      this.attachmentService.add(ownerDoc, filename, file, type, alreadyEncoded);
     };
 
     $record
@@ -498,9 +540,6 @@ export class EnketoService {
       .each((idx, element) => {
         const file = $(element).text();
         if (file) {
-          // Attach binary file with legacy-style filename because the actual filename is not stored as the question
-          // value in the form model (and so there is currently no way to map the answer in a saved report to the
-          // associated file attachment).
           attachLegacyFile(element, file, 'image/png', true);
         }
       });
--- a/webapp/src/ts/services/enketo.service.ts
+++ b/webapp/src/ts/services/enketo.service.ts
@@ -362,10 +362,13 @@ export class EnketoService {
     return form;
   }
 
-  private findBinaryNodeByFilename($record, filename: string) {
+  private findFileNodeByFilename($record, filename: string) {
+    // After upload, Enketo's Nodeset.setVal rewrites file-widget nodes from
+    // type="binary" to type="file" (#10903 §6.5). Inline-binary blobs from
+    // draw/signature widgets keep type="binary" and are handled separately.
     let match = null;
     $record
-      .find('[type=binary]')
+      .find('[type=file]')
       .each((_idx, element) => {
         if ($(element).text() === filename) {
           match = element;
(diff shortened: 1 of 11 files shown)


benkags:
@YASHSHARMAOFFICIALLY let's use the branch `medic:10700-photo-capture-in-sub-contacts-and-reports` as the base branch.

YASHSHARMAOFFICIALLY:
Done, updated the base branch to `10700-photo-capture-in-sub-contacts-and-reports`. Ready for review! @benkags @jkuester @dianabarsan

webapp/src/ts/services/enketo.service.ts
```diff
@@ -362,6 +362,19 @@ export class EnketoService {
     return form;
   }
 
+  private findBinaryNodeByFilename($record, filename: string) {
+    let match = null;
+    $record
+      .find('[type=binary]')
```
benkags: You want to ensure the you check for `type=file`. When testing #10923,  I realized Enketo sets type to `file`; so uploads would fail silently. Did

webapp/src/ts/services/enketo.service.ts
```diff
@@ -481,26 +494,52 @@ export class EnketoService {
...
+        const ownerDoc = resolveOwnerDoc(
+          this.findBinaryNodeByFilename($record, file.name) ?? $record[0]
+        );
+        this.attachmentService.add(ownerDoc, `user-file-${file.name}`, file, file.type, false);
+      });
 
     const attachLegacyFile = (elem, file, type, alreadyEncoded) => {
+      const ownerDoc = resolveOwnerDoc(elem);
       const xpath = Xpath.getElementXPath(elem);
+      const formId = ownerDoc === doc ? doc.form : ownerDoc.type;
```
benkags: `doc.form` and `doc.type` are two different types. And type is not always unique. Let's just use `doc.form`. What's the rationale for using `type` here if may ask?

webapp/tests/karma/ts/services/enketo.service.spec.ts
```diff
@@ -1163,6 +1163,92 @@ describe('Enketo service', () => {
...
+        'my-form',
+        form,
+        { doc: {} },
+        { _id: 'my-user', phone: '8989' }
+      );
+
+      expect(actual.length).to.equal(1);
+      expect(AddAttachment.callCount).to.equal(1);
+      expect(AddAttachment.args[0][0]._id).to.equal(actual[0]._id);
+    });
```
benkags, line 1251: We should add 2 more unit tests: 1) test repeats 2) test for clean up of attachments removed during edit

benkags (changes requested):
Thank you @YASHSHARMAOFFICIALLY. I've left a few review comments.

benkags:
@YASHSHARMAOFFICIALLY take a look at  https://github.com/YASHSHARMAOFFICIALLY/cht-core/pull/1

benkags:
@YASHSHARMAOFFICIALLY let me know if you are still working on this

YASHSHARMAOFFICIALLY:
@benkags  will raised a pr by today night working on these

benkags:
@YASHSHARMAOFFICIALLY I added the fixes in the linked PR, please take a look, add an e2e test so that it is ready for review.

YASHSHARMAOFFICIALLY:
@benkags Do check and let me know if any improvment needed further

YASHSHARMAOFFICIALLY:
@benkags @dianabarsan  @andrablaj  Thank you for the fixes and the review guidance. Addressed the remaining items:

**Unit tests added:**
1. **Repeats** (`db-doc-in-repeat-with-files`): Verifies each repeat instance's file routes to the corresponding sub-doc, with nothing landing on the main doc.
2. **Attachment cleanup on edit** (`db-doc-with-file-field-cleared`): Verifies that when a file is cleared from a sub-doc during editing, no attachment lands on the sub-doc while the main doc still receives its own file correctly.

Both use dedicated XML fixtures to isolate the scenario.

The e2e test (`submit-db-doc-file-upload.wdio-spec.js`) from the previous commit covers the full end-to-end flow: uploading files to both the main doc and a sub-doc, then verifying each attachment ended up on the correct document in the database.

Ready for re-review.

dianabarsan (changes requested):
Code looks solid, and makes sense on a quick read. Appreciate that. 
I need to request that we add an e2e test for this feature, that involves having a form with multiple sub-docs and test that attachments are saved correctly. We should also cover sub-docs in repeats. 

Appreciate this contribution!

webapp/src/ts/services/enketo.service.ts
YASHSHARMAOFFICIALLY: Good catch — updated the selector from `[type=binary]` to `[type=file]` in `900c9550a`. Enketo's `Nodeset.setVal` rewrites file-widget nodes to `type="file"` after upload, so this is the correct selector for FileManager files. Inline-binary blobs (draw/signature widgets) still use `type="binary"` and are handled via the separate xpath-based path.

webapp/src/ts/services/enketo.service.ts
YASHSHARMAOFFICIALLY: You're right — switched to `doc.form` for the xpath root segment in `900c9550a`. Using `doc.type` was incorrect since it's not guaranteed to be unique and serves a different purpose. The xpath-based filename is now always rooted at the main form's `internalId` regardless of which sub-doc owns the attachment.

webapp/tests/karma/ts/services/enketo.service.spec.ts
YASHSHARMAOFFICIALLY: Added both in `5da5cc194`:
1. **Repeats** — unit test verifies files inside `db-doc` blocks nested in repeat sections route to the corresponding sub-docs, with no files landing on the main doc.
2. **Attachment cleanup on edit** — unit test verifies that when a file field is cleared during edit, no attachment is created on the sub-doc.

YASHSHARMAOFFICIALLY:
@dianabarsan @andrablaj  Thank you for the review. I've added the requested e2e coverage in `e0229410f`:

1. **Multiple sub-docs** — new form (`db-doc-multi-file-upload`) with two separate `db-doc="true"` sub-documents, each with its own file upload. The test verifies each sub-doc receives only its own attachment and checks for cross-contamination between all three documents (main, sub-doc A, sub-doc B).

2. **Sub-docs in repeats** — new form (`db-doc-repeat-file-upload`) with a repeat group containing a `db-doc="true"` block with a file upload. The test creates two repeat instances with different files and verifies each repeat's sub-doc gets the correct attachment, with no files leaking to the main doc or between repeat sub-docs.

Happy to contribute more

benkags:
@dianabarsan please take another look and let us know if there is additional changes needed. Thanks.

webapp/tests/karma/ts/services/enketo.service.spec.ts
```diff
@@ -1163,6 +1163,220 @@ describe('Enketo service', () => {
...
+        'my-form',
+        form,
+        { doc: {} },
+        { _id: 'my-user', phone: '8989' }
+      );
+
+      // actual[0] = main doc, actual[1] = first repeat doc, actual[2] = second repeat doc
+      expect(actual.length).to.equal(3);
+
+      const file1Call = AddAttachment.args.find(
```
dianabarsan, line 1299: I think this test asserts that the files were attached to repeat document, but not to _which_ document.

dianabarsan (changes requested):
Thanks for the followup! 

Edge case worth covering: if two sub-docs end up with files of the same name (e.g., default-named camera captures, or any form where the user picks files with overlapping names from disk), `findFileNodeByFilename` returns the first match and both files route to the same sub-doc — the other sub-doc gets no attachment, and the `user-file-<name>`
  FileManager attachment key collides on top of that.

  The e2e tests use distinct filenames (`icon.png`, `layers.png`, `photo-for-upload-form.png`), so this isn't exercised. Could you add a case with two sub-docs whose files have the same name and decide what the desired behavior is? At minimum I think we want to detect and warn rather than silently drop one file.

  Could we also get an e2e test covering a draw/signature widget inside a `db-doc="true"` sub-doc? In production the draw widget puts its value in `FileManager` with a generated filename (`drawing-HH_MM_SS.png` — `webapp/src/js/enketo/widgets/draw.js:674`), so it goes through the same `findFileNodeByFilename` routing the PR
  rewires. That path is only tested with upload widgets right now.
Can we also have an e2e test covering binary uploads in subdocs?

  While you're there: those generated filenames are second-resolution, so a user signing two sub-doc sections within the same second produces colliding filenames — which is the same routing/collision concern I raised on the upload path, just more likely to hit because the user doesn't choose the filename.

benkags:
Some of the work done here has leaked into #10923 when trying to handle the binary type input - should we track one PR instead?

I know you have done a lot of work here already @YASHSHARMAOFFICIALLY  but at this point it is hard to separate out the reports logic in #10923's review.  I suggest you you target my branch here [benkags:10903-sub-contact-routing](https://github.com/benkags/cht-core/tree/10903-sub-contact-routing) so that we track all the work you have already done here from there  cc @dianabarsan @jkuester.

YASHSHARMAOFFICIALLY:
@dianabarsan  @benkags  Pushed a fix for the failing e2e test `should route file attachments to sub-docs inside repeat groups`.

**Root cause:** The `db-doc-repeat-file-upload.xml` form had a `<group>` element wrapping the `<repeat>`, which caused Enketo to render the repeat section on a separate page. The test never navigated to that page, so `$$(label*=Repeat photo)` found 0 elements → `Index out of bounds`.

**Fix:** Removed the `<group>` wrapper so the repeat renders on the same page as the name field, matching the pattern of the other two test forms (`db-doc-file-upload.xml`, `db-doc-multi-file-upload.xml`).

jkuester:
@YASHSHARMAOFFICIALLY thank you for the changes here!

@dianabarsan, sorry for hijacking the PR and merging prematurely!  :grimacing: :ninja: I was on a call with @benkags and we were trying to find the most simple way to sync all of this with https://github.com/medic/cht-core/pull/10923 and get it shipped.  The current plan is that @benkags is going to update his branch with the latest from https://github.com/medic/cht-core/tree/10700-photo-capture-in-sub-contacts-and-reports and then target `master` with github.com/medic/cht-core/pull/10923.

So, the complete unified changes will be available to review in that PR. :+1:

PR #10923, opened 2026-04-24
feat(#10903): sub contact attachment routing

### LINKED ISSUE #10791: Unable to hide individual fields inside a repeat from the report view
**Describe the bug**
Unable to [hide](https://docs.communityhealthtoolkit.org/building/translations/overview/#hiding-report-fields) individual form fields inside a [`repeat`](https://docs.getodk.org/form-logic/#repeating-questions) in the reports view.

**To Reproduce**

1. Add a `repeat` to your app form
2. Add some fields (questions) inside the repeat
3. Add a column `instance::tag` and set it to `hidden` for some of the fields added in step 2
4. Convert and upload the app to CHT
5. Submit a report using that form
6. Check the report in the Reports tab
7. All fields inside `repeat` are visible
**Expected behavior**
The fields inside `repeat` which have been tagged as `hidden` should not be visible.

**Screenshots**
Here is a sample XLSForm repeat survey:
Image

repeat.properties.json:
```json
{
  "title": [
    {
      "locale": "en",
      "content": "Repeat"
    }
  ],
  "context": {
    "place": true,
    "person": true
  },
  "icon": "icon-repeat",
  "hidden_fields": [
    "hidden_props",
    "items.hidden_props"
  ]
}
```

After I fill the form, I can still see the fields that should have been hidden:
Image

**Environment**
- Instance: local
- Browser: Chrome
- Client platform: Windows
- App: webapp
- Version: tried on 5.x, but not specific to it

**Additional context**

CHT forum post: https://forum.communityhealthtoolkit.org/t/how-do-i-hide-some-repeat-fields-on-a-report/5549

### LINKED ISSUE #10903: sub-contact photo capture: Contacts core routing + validation + unit tests
**Describe the issue**
child issue of [photo capture for sub contacts](https://github.com/medic/cht-core/issues/10700) covering:

* Add attachment - doc mapping logic
* Route file-widget and inline-binary attachments per-doc using the mapping, with per-doc processing
* Validations, filename sanitization & clean up

### LINKED ISSUE #10904: sub-contact photo capture: Reports core routing + unit tests

benkags:
# Description

 Every file uploaded in an Enketo form was attached to the **main** doc, regardless of which sub-document it was uploaded in. A photo added inside a CHW (sibling) section of a CHW Area creation form landed on the CHW Area doc; a file inside a `db-doc="true"` group or
  repeat landed on the report doc instead of the sub-report.

  This PR routes each upload and inline binary to the document that owns it, in **both** save pipelines:

  **Contact forms (#10903)**
  - Resolves the owner of each `[type=file]` / `[type=binary]` node by its form section: main section → primary doc, `<contact>`/`<parent>` section → the matching sibling doc, i-th `<repeat>` `<child>` → the i-th repeat doc. Unmatched uploads fall back to the main doc.
  - Filename sanitization and orphaned-attachment cleanup now run per-doc instead of main-doc-only.
  - Attachment size validation now runs on the contact save path (previously reports only).

  **Report forms (#10904)**
  - Routes uploads and inline binaries to their nearest `db-doc="true"` sub-document, falling back to the main report.
  - Writes each media field's value into `doc.fields` (repeat-index aware for the main doc), so the field value always resolves its attachment as `user-file-` + value.

  **Shared pipeline**
  Both services now delegate to a new `AttachmentRoutingService` (upload routing → inline-binary routing → per-doc finalize), parameterized by a small per-pipeline strategy (owner resolution, reference container, field path). Pure DOM primitives live in
  `attachment-routing.provider.ts`.

  ## Attachment naming & backwards compatibility

  - Inline-binary attachments on **newly saved** docs are named `user-file-<owner-relative-xpath>` (e.g. `user-file-photo`, `user-file-group/photo`, `user-file-my_repeat[2]/photo`) instead of the legacy `user-file/<form-id>/<xpath>`, and the field value now stores the bare
  reference. File-widget uploads keep the `user-file-<filename>` scheme.
  - **Existing docs are not migrated.** Legacy slash-named attachments are left intact on edit (covered by tests), and orphan cleanup only considers `user-file-`-prefixed names it can account for.
  - On edit, an untouched inline-binary field is preserved via a `data-attachment-ref` sidecar attribute stashed at load time; a fresh upload always wins over a stale sidecar. A dedicated spec pins the enketo-core merge behavior this relies on.
- Testing
  - Unit: routing primitives, shared pipeline, contact and report strategies, translation-service binary handling (sidecar stash/restore, no value bind-back).
  - E2E: sub-contact routing on create and edit, `db-doc` sub-report routing (including repeats and multiple sub-docs), untouched inline-binary preservation on report edit.
  - Closes #10903
  - Closes #10904
  - Closes #10791

# Code review checklist

- [X] Internationalised: All user facing text

# License


benkags:
This is ready for review.

webapp/src/ts/services/enketo.service.ts
```diff
@@ -362,6 +362,19 @@ export class EnketoService {
...
+    let match = null;
+    $record
+      .find('[type=binary]')
+      .each((_idx, element) => {
+        if ($(element).text() === filename) {
+          match = element;
+          return false; // break
+        }
+      });
+    return match;
```
jkuester:
```suggestion
    return $record
      .find('[type=file]')
      .toArray()
      .find(element => $(element).text() === filename);
```

Minor, but I guess we can simplify this. Even more importantly, I think we need to be searching for `[type=file]`.  

The `type=binary` is a special case where the CHT supports [binary image data](https://docs.communityhealthtoolkit.org/building/forms/app/#uploading-binary-attachments) as the _value_ for a question. This flow bypasses the Enekto file attachment handling (because the image data might be injected [from a 3rd party Android app](https://docs.communityhealthtoolkit.org/building/forms/app/#android-app-launcher) and so does not go through the normal Enketo widgets for file uploads).  The confusingly named `attachLegacyFile` logic below is for handling this case.

On the other hand, This `findBinaryNodeByFilename` function is used during the normal attachment processing for files from the Enketo `FileManager`.  This logic is used during normal file uploads via the Enekto widgets (see [this form](https://github.com/jkuester/cht-config-library/blob/main/basic-forms/forms/app/file_uploads.xlsx) as an example).  Based on my debugging, when we have files from the Enketo `FileManager`, the xml model element is going to have  `type=file`

webapp/src/ts/services/enketo.service.ts
```diff
@@ -481,26 +494,52 @@ export class EnketoService {
...
+        const ownerDoc = resolveOwnerDoc(
+          this.findBinaryNodeByFilename($record, file.name) ?? $record[0]
+        );
+        this.attachmentService.add(ownerDoc, `user-file-${file.name}`, file, file.type, false);
+      });
 
     const attachLegacyFile = (elem, file, type, alreadyEncoded) => {
+      const ownerDoc = resolveOwnerDoc(elem);
       const xpath = Xpath.getElementXPath(elem);
+      const formId = ownerDoc === doc ? doc.form : ownerDoc.type;
```
jkuester: I am not 100% sure what the best thing to do here is... 🤔  I _think_ the only time this attachment name really matters programmatically is when we render images for existing reports on the "Reports" tab.  In that case, we always look at the `form` value [for the report being viewed](https://github.com/medic/cht-core/blob/1afebb7970d5d92afcd4c709921410fbf741dfaf/webapp/src/ts/services/format-data-record.service.ts#L506).  I think maybe we can do:

```suggestion
      const formId = ownerDoc.form || doc.form;
```

That should at least make it so that if the `form` property is set on the sub-doc, then we will write the attachment in such a way that it will render if the sub-doc is opened on the Reports page.

webapp/src/ts/services/enketo.service.ts
```diff
@@ -481,26 +494,52 @@ export class EnketoService {
...
       // replace instance root element node name with form internal ID
       const filename = 'user-file' +
-        (xpath.startsWith('/' + doc.form) ? xpath : xpath.replace(/^\/[^/]+/, '/' + doc.form));
-      this.attachmentService.add(doc, filename, file, type, alreadyEncoded);
+        (xpath.startsWith('/' + formId) ? xpath : xpath.replace(/^\/[^/]+/, '/' + formId));
```
jkuester: Even that will not be enough though.  😓  This code is going to end up using the whole report structure path to the field for the overall doc (and not the path just for the sub-doc). To get things to render correctly when opening a sub-doc in the Reports tab, I think we would need to:

- Get the xpath to the element with the `_couchId` value for the subdoc (similar to your `resolveOwnerDoc` logic).
- Take the original `xpath` value and remove that ownerDocXpath prefix
- Append `'/' + formId` to the front of the remaining path.

webapp/src/ts/services/contact-save.service.ts
```diff
@@ -173,51 +180,165 @@ export class ContactSaveService {
...
+  ) {
     const mainDoc = preparedDocs[0];
-    const newAttachmentNames = new Set<string>();
-    const fileNameMap = new Map<string, string>(); // Map of original file name -> sanitized file name
+    const $record = xmlStr ? $($.parseXML(xmlStr)) : null;
+    const root = ($record?.children(':first')[0]) as Element | undefined;
+    const formId = root ? $(root).attr('id') : undefined;
+    const ctx: ContactOwnerContext | null = root
+      ? { root, preparedDocs, mainDoc, submittedRepeatsLen: submitted?.repeats?.child_data?.length ?? 0 }
+      : null;
```
jkuester:
```suggestion
    const root = $.parseXML(xmlStr).documentElement;
    const formId = $(root).attr('id');
    const ctx: ContactOwnerContext = { 
      root, 
      preparedDocs, 
      mainDoc, 
      submittedRepeatsLen: submitted?.repeats?.child_data?.length ?? 0 
    }l;
```

I may be missing something here, but I think if we made it this far, it should be safe to assume we have a valid `xmlStr` with the expected contents.

webapp/src/ts/services/enketo.service.ts
jkuester: Actually.....  Maybe I have been staring at this code too much, but I am getting some crazy ideas here.....  😅

Why do we need to bother supporting this attachment naming convention at all???  The existing "reasoning" here was that we don't have a file name to use.  BUT, I cannot come up with a good reason why we cannot just generate an attachment name to use.  (E.g. just create a new uuid or something.)  We store the attachment with that unique name and then we set the text _value_ on the element to be the uuid value.  This matches how the FileManager attachments are handled already with the name of the attachment stored as the value for the question and saves us from having to have bifurcated attachment logic downstream....

Essentially I what i am suggesting is:

- Update the processing logic here for the `type=binary` elements to:
    - Generate a unique key/name for each attachment.
    - Add the `image/png` attachment for the binary data as we are now with the attachmentService, using the newly generated name.
    - Set the text value on the  `type=binary` element to the generated "name"
- Update similar processing logic over in the contact-save.service.ts. 
- Update the enketo-translation.service.ts to not [clear the value of binary fields](https://github.com/medic/cht-core/blob/ee6b24ae719740c8a768250ef078ccce772d10cf/webapp/src/ts/services/enketo-translation.service.ts#L47).
- Leave the logic as it is in the format-data-record.service.ts for rendering the images in the Reports tab.  We do not want to break the image display for old reports that did not store the file-name as the question value.


Let me know if you think this will not work or is a bad idea, but the more I think about it the more I think it will go a long way to simplifying things and just keeping our attachment logic reusable.

jkuester (changes requested):
@benkags once again, I am so glad you are working through this!  ✨ I spent hours this afternoon reading through and debugging around this code. I have not finished reading it all yet, but I ran out of time today and wanted to push the thoughts that I have so for. 👍

webapp/src/ts/services/enketo.service.ts
benkags: @jkuester your 3rd paragraph here - are you referring to the case handled above in `findContactOwnerForFilename` which is different from the binary image data case you describe in 2nd paragraph? In which case I believe looking for `type=binary` is still relevant for the cases where the binary data itself is the value to the question.

benkags: Not sure the case in paragraph 2 is the one for which Enketo writes the base64 as the question value. The specific cases I came across that lead to this are the [drawing widgets](https://docs.getodk.org/form-question-types/#draw-widget).

webapp/src/ts/services/enketo.service.ts
benkags: @jkuester  I've spent some more time on this. Here are my thoughts and what I think should be the way forward.

Your comment here covers two concerns:-
a) display of an image when rendering the sub-report. We do not yet support rendering of images in contact edit view as far as I can tell
b) handling attachment for legacy binary type

**a) display of an image when rendering the sub-report**
This is handled in https://github.com/medic/cht-core/issues/10904 with no legacy path support for the reason outlined in b) below.

**b) handling attachment for legacy binary type**
The path is never invoked in modern enketo implementation. It turns out my understanding of signature / draw widget being processed via the legacy path was incorrect. No binary types are currently processed via the legacy path after https://github.com/medic/cht-core/pull/8904 in the real world. I think the legacy write path is a safe fallback, with no much practical real life scenarios.  In report view,  attachments are retrieved by the file name with [a fallback to handle legacy attachment names](https://github.com/medic/cht-core/blob/master/webapp/src/ts/services/format-data-record.service.ts#L468) which is relevant for viewing reports submitted pre-8904 but would never execute for any deployments after this feature ships.

We should not then be updating this legacy path  in this feature. It has zero value as there aren't going to be any cases that invoke it.

**Way forward:** undo the changes here - do not support `attachLegacyFile` with in this feature as it is irrelevant.

benkags: @jkuester lifting your comment from above
"The type=binary is a special case where the CHT supports binary image data as the value for a question. This flow bypasses the Enekto file attachment handling (because the image data might be injected from a 3rd party Android app and so does not go through the normal Enketo widgets for file uploads). The confusingly named attachLegacyFile logic below is for handling this case." 

This case warrants maintaining this 'legacy' path, and will require that we keep this update and perhaps use your suggested approach here. I'll if I can set up a form with this case and validate.

webapp/src/ts/services/enketo.service.ts
benkags: Ignore my comments here. See below https://github.com/medic/cht-core/pull/10923/changes#r3228574162

webapp/src/ts/services/enketo.service.ts
benkags: @jkuester implemented the suggestion here. I've tried to detail the solution in the commit message [here](https://github.com/medic/cht-core/pull/10923/commits/080c1e210f334d2c47e3e5312dde727b918e94f0) but in summary, retain the the xpath based attachment name but now adds it as the field value to the doc generated. Viewing & editing of docs saved using legacy path preserved.

benkags:
@jkuester please take another look

jkuester (changes requested):
Okay, I have been _deep_ down the rabbit hole here.  😬  🐇 

All of this is so complicated and interlocked that I ended up just creating https://github.com/benkags/cht-core/pull/10 with my suggested changes instead of trying to put everything in comments here. 😓 The main focus of my changes was getting all the attachments for the main report and sub-reports to be rendered as expected when viewing the report in the "Reports" tab.  

I used [file_uploads.xlsx](https://github.com/user-attachments/files/27817546/file_uploads.xlsx) to test with.

A couple things I learned in all my testing and debugging:

## Current attachment naming strategies (before this PR):

1. New-style `file` type attachements (`>=4.9.0`):
	- Property value set to attachment name
	- Attachment name has `user-file-` prefix + *property value*
2. Old-style `file` type attachments (`<4.9.0`):
	- Property value set to attachment name
	- Attachment name has `user-file/` prefex + *path/to/property*
3. All `binary` type attachments:
	- Property value set to `""`
	- Attachment name has `user-file/` prefex + *path/to/property*

`#1` and `#2` are what you got when using any of the `image`/`audio`/`file`/etc types in your xform and the user selects a file from the device to upload.

`#3` should mainly be coming from the external Android app integration where you are getting base64 data loaded as the _value_ for a field in the form with `instance:: binary` set to `true`. (Technically can also enter base64 text right in the form or load it as a default value.) This data does not go through the Enekto file manager.

## Editing docs with attachments

Currently, (before this PR) you can edit docs with file attachments.  The attachments with `type=file` are rendered properly in the form, but the `type=binary` attachments are not rendered.  Depending on how we plan to load profile pictures into a contact form, this may or may not be something we need to worry about....

The rendering of the `type=binary` attachments is not fixed (even by my changes in the linked PR).  I was up way too late last night trying to sort out a solution (see the first commit in my PR for what I came up with).  I think there is something viable here that we could move forwards with in the future, but ultimately I felt it was out of scope for your PR here since the existing functionality remains unchanged.

---

The changes in my PR come without any updates to the tests  😓 or even any manual validation of the contact-attachment flow. I have been purely testing with reports.  I am out of time for the week, though, and I figured I would post what I have and get your thoughts.  👍

benkags:
Thanks @jkuester for going down the rabbit hole.  I have been down there myself a couple of times with this PR :smiley: . Thanks for PR https://github.com/benkags/cht-core/pull/10  - I agree, it a much better PR review feedback approach. This is what I picked from PR as net change. Let me know if I missed something
1) a bug fix that I'd missed: File-widget routing bug in `xmlToDocs`
2) updated xpath based attachment name logic; so that the sub docs have the relative file path
3) undoes the 'attachment-filename-in-the-field-value' for the binary types in favor of constructing the base64 string from the attachment itself for re-population when editing the report (the spirit of the first commit)

I think it really comes down to what approach to adopt between Approach A (the one I have here) and Approach B (in your PR) from a design perpective
- **Approach A**: standardize as much as possible with  `[type=file]`  approach by populating the `type=binary` field value with the attachment name. This saves us the extra logic of constructing `base64` string from the attachment for edit. The report view is handled by reading the attachment directly and injecting it as an image like any other file object of type image. The cost to pay for here is to handle exiting reports before this PR by populating the binary field from the saved xml form model at display. 
  - the reports edit / view works in the PR  
image
image

  - the report edit was broken before (but report view worked) 
image

 
- **Approach B**: Does not save attachment name in the saved doc, reconstructs the attachment name from the xml at display and constructs the binary's base64 string to populate the `type=binary` field.

jkuester:
I think you have pretty clearly understood my PR. :+1:  Just a couple comments:

> Approach A: standardize as much as possible with [type=file] approach by populating the type=binary field value with the attachment name. 

I 100% was hoping this would work out as you have described. :sweat: Unfortunately, it did not really seem to save us any complexity. 

> This saves us the extra logic of constructing base64 string from the attachment for edit. The report view is handled by reading the attachment directly and injecting it as an image like any other file object of type image. 

The `format-data-record` service will still need to keep the extra logic to support any existing reports with the `user-file/` format.  And the logic in the `reports-add` component for loading the image attachments when editing a report will still not work for `type=binary` fields (because the form model knows that they are not `type=file` fields and so their handling logic is totally different).

On top of that, changing to store the attachment name as the property value for `type=binary` fields ends up creating new challenges _in the edit workflow:_

1. If we do not do anything and just leave the attachment name set as the value, it breaks the `display-base64-image` preview functionality so that a "invalid image" icon is shown in the form (instead of the current edit "functionality" where _nothing_ is shown when editing a report for the `display-base64-image` fields.... :grimacing: (FTR, we can avoid the icon and actually display the image if we re-load the base64 data from the attachment back into the property value before rendering the form. However, as I found in my first commit, this is not super easy. We will probably need to support this eventually, though... :shrug: )
2. If/when we end up re-injecting the base64 image data for `type=binary` fields on edit, we immediately hit the challenge of how to avoid adding a duplicate attachment file each time the report is edited even if the `type=binary` field did not change.  Using a random uuid-based attachment name is obviously a problem because we would always get different names. My first commit used the hash of the file contents which worked but added overhead/complexity. Another option would be to just use the path to the field (e.g. like the original `type=binary` logic was doing).  

It ended up being similar complexity to just properly calculate the field path (even for the sub-docs).  That was the point where I started trying to figure out what benefits I was actually getting from `Approach A.`

On the other hand, with `Approach B.`, we end up with a minimal amount of code changed and all the `type=binary` logic still just follows what it was doing before (but now also works for sub-docs).  We have not simplified anything, but we really have not _added_ much more complexity either... :sweat_smile: :shrug: 

---

@benkags if you have a strong preference for `Approach A`, I am not set against it (as long as we can find a maintainable way to sort out the above challenges).  Otherwise, I do think `Approach B` (as drafted in my PR) should result in the functionality we need with the minimal amount of changes.

Also happy to jump on a call to discuss further if that would be helpful!  :+1:

benkags:
Ready for review @jkuester

jkuester:
@dianabarsan @benkags you have continued the discussion/efforts here in my absence last week. :pray: Thank you!

I know there are some plans in motion regarding a Feature Release, etc. I do not intent to change our direction there at all, but I did want to start a new conversation about how to move forwards in the future with getting these changes (or an future iteration of them) into `master`.  

Coming back this week and taking a fresh look at the code, I have put together a sort of [design doc](https://docs.google.com/document/d/1Y3wnvwlog0aULU3jetDgM2znhZ1wWki2PifU01BQ4pw/edit?usp=sharing) where I try to clearly outline the details of the existing `binary` logic in `master`, the changes to that logic that are introduced in this PR, and then my latest thoughts on the additional changes that we need to make to get this merged back to `master`.  You should be able to comment line-by-line in the doc so please give some  thoughts/feedback when you get a chance!  :+1:

benkags:
@jkuester I have pushed a couple of updates to this PR align with design direction [here](https://docs.google.com/document/d/1Y3wnvwlog0aULU3jetDgM2znhZ1wWki2PifU01BQ4pw/edit?tab=t.0).

I wanted to give a brief reviewer guide: report and contact attachment handling is now unified into one shared pipeline. Binary-attachment naming on an **owner-relative xpath** removing the form-id prefix approach.

**Suggested reading order** 
a) `attachment-routing.ts` (contains pure functions and abstraction from the refactor)
b) `attachment-routing.service.ts` - the routing pipeline
c) the routing strategies constructed in `enketo.service.ts` (reports) and `contact-save.service.ts` (contacts)
d) `enketo-translation.service.ts` edit preservation logic: (edit-load + the `data-attachment-ref` sidecar).

**A summary of the changes**
- Binaries are named `user-file-<relative-xpath>`, relative to the owner doc such taht a sub-doc photo is just `user-file-photo`; the render resolves from the attachment reference
- Edit-load never rehydrates a binary into the form model; an *untouched* binary is preserved on save via a `data-attachment-ref` sidecar instead of the proposal to recontruct the reference and confirm attachment presense - much simpler
- "second translation pass" is gone in reports
- Reports now get orphan cleanup

**Stating some constraints**
- **No existing binary preview on edit** (field shows empty) — existing behaviour - any attempt to fix deliberately deferred
- **Sidecar depends on Enketo keeping unknown `data-*` attributes** - preservation guarded by `attachment-ref-roundtrip.spec.ts`; so that if Enketo behaviour changes we get to know.

webapp/src/ts/services/attachment-routing.ts
```diff
@@ -0,0 +1,122 @@
...
+  fieldPathFor(element: Element, ownerDoc: Record<string, any>): FieldPath | null;
+}
+
+/**
+ * The `[type=file]` widget node within `root` whose text is `filename`, or null.
+ * Filenames are session-unique, so the first match is the only match; the caller
+ * owns the no-match fallback.
+ */
+export const findUploadNodeByFilename = (root: Element, filename: string): Element | null => {
+  return $(root).find('[type=file]').toArray().find(element => $(element).text() === filename) ?? null;
```
dianabarsan: The `[type=file]` fix here is correct for normal uploads, but it opens a gap for **draw/signature/annotate widgets**: enketo-core's `getCurrentFiles()` returns those too, yet their model node is `type=binary`, not `type=file`. So `findUploadNodeByFilename` returns `null`, `routeUploads` falls back to `mainDoc`, and the owning sub-doc keeps
the raw filename as its value — the attachment lands on the wrong doc with no error. The `[type=binary]` pass won't rescue it either, since the filename is already in `fileManagerNames` so `routeOneBinary` skips it. Can we either route the filename rewrite to the same doc that receives the attachment, or handle the no-match case explicitly instead
of silently using `mainDoc`?

webapp/src/ts/providers/attachment-routing.provider.ts
```diff
@@ -0,0 +1,122 @@
...
+ * Prefix for every CouchDB attachment created from a form media field, so a
+ * field's value resolves uniformly as `USER_FILE_PREFIX + value`.
+ */
+export const USER_FILE_PREFIX = 'user-file-';
+
+/** A path accepted by object-path: a dot-delimited string or a segment array. */
+export type FieldPath = string | (string | number)[];
+
+/** Per-pipeline parts the shared `AttachmentRoutingService` delegates to. */
+export interface AttachmentRoutingStrategy {
```
dianabarsan: Can this be a provider file?

webapp/src/ts/services/attachment-routing.ts
```diff
@@ -0,0 +1,122 @@
...
+  /** The owner doc's field-container element, used to root the relative reference. */
+  containerFor(element: Element, ownerDoc: Record<string, any>): Element | null;
+
+  /** objectPath location of the field value within `ownerDoc`, or null to skip. */
+  fieldPathFor(element: Element, ownerDoc: Record<string, any>): FieldPath | null;
+}
+
+/**
+ * The `[type=file]` widget node within `root` whose text is `filename`, or null.
+ * Filenames are session-unique, so the first match is the only match; the caller
```
dianabarsan: The comment here says filenames are session-unique so the first match is the only match — but that's enketo's guarantee, not one we enforce, and it's only probabilistic (the upload postfix is second-resolution: `-HH_MM_SS`, no ms or random). Two same-named files dropped into different sub-docs in the same second both resolve to the first node →
both attach to one doc, the other sub-doc's field points at a missing attachment. Pre-PR this was harmless (one doc); per-doc routing makes it a mis-route. Could we consume matched nodes so each file maps to a distinct node, and add a test for same-filename-in-two-sub-docs?

webapp/tests/karma/ts/services/format-data-record.service.spec.ts
```diff
@@ -243,6 +243,50 @@ describe('FormatDataRecord service', () => {
       ]);
     });
 
+    it('resolves an inline-binary field via user-file- + bare reference value', async () => {
```
dianabarsan: These additions only cover a *main* report — `fields.photo = 'photo'` resolving via `user-file-photo`. There's no case where the value is an owner-relative/nested reference (e.g. `group/photo` → `user-file-group/photo`), which is exactly the sub-doc-rendered-on-the-Reports-tab path. The resolver is purely
value-based so it likely works, but the one matrix cell that matches the  headline concern is untested. Could we add a `format-data-record` case that formats a sub-doc whose `fields.<field>` is an owner-relative reference and asserts `imagePath` resolves — or an e2e that opens a created sub-report and asserts the `` renders? Right
now the e2e specs assert DB state (`_attachments` keys, field values) but never a rendered image.

webapp/tests/karma/ts/services/enketo.service.spec.ts
```diff
@@ -1151,18 +1151,486 @@ describe('Enketo service', () => {
...
+      const subPhoto2Call = AddAttachment.args.find(args => args[2] === 'sub_photo_data_2');
+      expect(subPhoto2Call).to.exist;
+      expect(subPhoto2Call[1]).to.equal('user-file-photo2');
+
+      const mainPhotoCall = AddAttachment.args.find(args => args[2] === 'main_photo_data');
+      expect(mainPhotoCall).to.exist;
+      expect(mainPhotoCall[1]).to.equal('user-file-main_photo');
+    });
+
+    it('should not attach removed files to sub-docs during edit', async () => {
```
dianabarsan: Edit-path coverage for sub-docs only exercises *clearing* a file (`db-doc-with-file-field-cleared`) and *adding a new* repeat child. There's no test that **replaces** an existing sibling/repeat-child file-widget upload on edit and asserts the stale `user-file-<old>` is orphan-removed *from that sub-doc* — orphan cleanup on edit is asserted only for
the main doc. Since the orphan scan now loops over every prepared doc (a behavior change for sub-docs), could we add one case: edit a saved family, replace the primary-contact photo, and assert the new attachment lands on the sibling while the stale `user-file-<old>` is removed from the sibling (not main)?

dianabarsan (commented):
Thank you for the updates, @benkags , and sorry about the wait. 
I added some comments inline, mostly about the dreaded edge-cases and about some test coverage. 
Waiting for @jkuester for the review here, since he was deeply involved in it.

benkags:
Thanks for the review @dianabarsan. I addressed your feedback.

jkuester (commented):
Okay, @dianabarsan, this should be ready for your final review!  🙏 

As was foretold, the merge here with the latest changes in master was spicy.  The main changes that now exist here are that I moved all the attachment handling for the docs out of the `enketo.service` and down into the `form-data` classes. That just provided cleaner re-usability (and got a bit more logic out of the EnektoService...).  

Bernard had a bunch of e2e tests and I combined a few and stacked some more in their too, so we really have good e2e coverage of all this functionality now!

A couple other things that changed were:

- `binary` attachment names for reports now include the top-level `fields` in the path. This is just to unify the naming standard across all docs (reports + contacts).
- Updated the `format-data-record.service` to properly account for this naming change when loading the binary images for the Reports view.
- Also updated the `format-data-record.service` to properly handle repeat groups in the `hidden_fields` list (fixes https://github.com/medic/cht-core/issues/10791).  
- To support all of this, I ended up refactoring/simplifying the xpath code to allow for injecting the form's `repeatPaths` context.

---

@benkags can you support here by testing the latest code in this branch against some of the edge cases you and Imran were working through?  🙏   I am pretty sure everything you had automated tests for is covered, but as you know there are just an unending amount of nasty edge cases and I want to be sure we do not regress on anything that you had working before... 😬

jkuester (commented):
^

tests/e2e/default/enketo/db-docs-with-attachments.wdio-spec.js
```diff
@@ -0,0 +1,155 @@
...
+      .reportImage('report.db-docs-with-attachments.photo')
+      .waitForDisplayed();
+    // TODO binary attachments in repeat not rendered
+    await reportsPage.rightPanelSelectors
+      .reportImage('report.db-docs-with-attachments.repeated_attachments.0.photo')
+      .waitForDisplayed();
+    await reportsPage.rightPanelSelectors
+      .reportImage('report.db-docs-with-attachments.repeated_attachments.0.photo')
+      .waitForDisplayed();
+
```
dianabarsan, line 144: These two assertions are identical. Maybe this was meant as `.1.photo`?

tests/e2e/default/enketo/db-docs-with-attachments.wdio-spec.js
```diff
@@ -0,0 +1,155 @@
...
+    expect(repeatDoc1.fields.photo).to.match(/^photo-for-upload-form-/);
+    expect(Object.keys(repeatDoc1._attachments)).to.deep.equal([
+      `user-file-${repeatDoc1.fields.photo}`,
+      'user-file/fields/badge'
+    ]);
+
+    await reportsPage.rightPanelSelectors
+      .reportImage('report.db-docs-with-attachments.photo')
+      .waitForDisplayed();
+    // TODO binary attachments in repeat not rendered
```
dianabarsan: There's a karma test that contradicts this at webapp/tests/karma/ts/services/format-data-record.service.spec.ts:307

webapp/src/ts/services/form/form-data.ts
```diff
@@ -63,28 +79,117 @@ export class EnketoFormData {
...
-    this.binaryTypeElements = Array.from(this.rootElement.querySelectorAll('[type=binary]'));
+    const isOrphanedFileAttachment = (fileName: string) => fileName.startsWith(USER_FILE_ATTACHMENT_PREFIX)
+      && !this.findNodeWithTextContent(fileName.slice(USER_FILE_ATTACHMENT_PREFIX.length));
+    const binaryAttachments = this.binaryTypeElements
+      .map(element => this.buildBinaryAttachmentData(element, xpathPrefix, repeatPaths))
+      .filter(({ attachment }) => attachment)
+      .reduce((binaryAttachments, { filename, attachment }) => ({ ...binaryAttachments, [filename]: attachment }), {});
+    const newFileAttachments = FileManager
+      .getCurrentFiles()
+      .filter(({ name }) => this.findNodeWithTextContent(name))
```
dianabarsan, line 96: Will two files with the same name, but different content, get mapped correctly in different docs here?

tests/e2e/default/enketo/db-docs-with-attachments.wdio-spec.js
jkuester: :sweat_smile: Totally forgot to circle back to update this after fixing the bug... :+1:

webapp/src/ts/services/form/form-data.ts
jkuester: > two files with the same name, but different content

This is technically impossible.  Enketo appends a unique(ish) value to the end of a filename when you upload it into the form (`hh-mm-ss`).  If the user manages to upload different files with the same name into two different form questions within the same second, the collision will happen in the Enekto code (and I think the second file will win).  If the form questions are routed to different docs, the winning file will be attached to both docs (since both docs have a property that references the file-name).  This seemed like the most acceptable behavior without patching Enketo to have more unique file names.

benkags:
Thank you @jkuester for the hard work merging master. I'll manually test this and update here.

benkags:
D-Tree's household form works as expected.

tests/e2e/default/enketo/db-docs-with-attachments.wdio-spec.js
```diff
@@ -0,0 +1,160 @@
...
+  const photoPath1 = path.join(__dirname, '/images/photo-for-upload-form1.png');
+  const photoPath2 = path.join(__dirname, '/images/photo-for-upload-form2.png');
+
+  before(async () => {
+    await utils.saveDocIfNotExists(commonPage.createFormDoc(`${__dirname}/forms/${FORM_ID}`));
+    await loginPage.cookieLogin();
+    await commonPage.goToReports();
+  });
+
+  it('writes report with attachments at various levels including in db-docs', async () => {
```
dianabarsan, line 26: This test only covers create, and it's unclear to me what happens on edit. 
Can you please add another test that edits a report with sub documents with attachments and show what is happening?

jkuester: Done!

dianabarsan (approved):
🛳️

jkuester (approved):

PR #10927, opened 2026-04-25
fix(#10913): limit freetext generator results to skip+limit to preven…

### LINKED ISSUE #10913: Webapp and Admin select2-search loads all results from the database
**Describe the bug**
When you are an online user, and you use select2-search to select a contact, the app ends up loading ALL possible results from the server. Only a limited list is actually displayed. 

**To Reproduce**
1. Create 100.000 contacts that have the same name
2. Go to Admin app and open the create new user page. 
3. Search for your contact in the select2 search
4. The request takes a long time, or crashes your server. 

Checking the console, you can see multiple requests, each requesting 10.000 records, are being made until the results are exhausted. 
Scrolling to load more results ends up running the same queries all over again. 

Same behavior is seen in webapp. 

**Expected behavior**
We should only load a subset of contacts for these searches. I believe these are actually passed to the service with a limit, but then the limit is ignored.

**Video**

https://github.com/user-attachments/assets/a9599719-8bcc-4586-8f9c-318a7a0cafa1

**Environment**
- Instance: local
- App: admin
- Version: 5.1.0

taanvi2205:
Fixes #10913

When an online user searched contacts via select2, the freetext generator was calling `getUuidsByTypeFreetext` / `getUuidsByFreetext` which exhausted all matching results in 10,000-record batches before `getPageRows` sliced to the requested page size (e.g. 20). This caused repeated server requests and could crash the server with large datasets.

The fix makes `iterateGenerator` in `freetext-query.js` accept a `limit` parameter and break out of the generator loop early. `search.js` passes `options.skip + options.limit` as that limit, so only enough IDs to cover the current page are fetched.

# Code review checklist

# License

### DIFF
--- a/shared-libs/search/src/freetext-query.js
+++ b/shared-libs/search/src/freetext-query.js
@@ -1,24 +1,27 @@
 const chtDatasource = require('@medic/cht-datasource');
 const logger = require('@medic/logger');
 
-const iterateGenerator = async (gen) => {
+const iterateGenerator = async (gen, limit) => {
   const rows = [];
 
   for await (const id of gen) {
     rows.push({
       id: id
     });
+    if (limit && rows.length >= limit) {
+      break;
+    }
   }
 
   return rows;
 };
 
-const queryFreetext = async (dataContext, request, type) => {
+const queryFreetext = async (dataContext, request, type, limit) => {
   try {
     const datasource = chtDatasource.getDatasource(dataContext);
     const generator = getGeneratorByType(datasource, request, type);
 
-    return await iterateGenerator(generator);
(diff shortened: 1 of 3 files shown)


jkuester:
Thanks for the PR @taanvi2205.  Unfortunately, as I mentioned [in the issue thread](https://github.com/medic/cht-core/issues/10625), the described behavior is actually being caused by a deeper structural problem with the `shared-libs/search` logic (which is not addressed by the changes in this PR).  Closing this PR along with https://github.com/medic/cht-core/issues/10913 in favor of a fix that actually addresses the root cause in https://github.com/medic/cht-core/issues/10625.

PR #10928, opened 2026-04-25
feat(#10915): add cht- prefix to custom element name for UI extensions

### LINKED ISSUE #10915: Add prefix when creating custom element name
Currently when we [define a custom element](https://github.com/medic/cht-core/blob/c8238ec5f1acb1ccad18fd966f2e7ffbaf7061c1/webapp/src/ts/modules/ui-extensions/ui-extensions-tab.component.ts#L42) in the page for a UI Extension, we are just using the UI Extension's `id` value as the name of the custom element. This has the unfortunate consequence of meaning that all the [requirements for a custom element name](https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name) are forced upon the UI Extension.  The `id` for the UI Extension is just the file name of the js/properties files in the `ui-extensions` config directory.  Since we use this `id` as the custom element name, it means that the file name for the UI Extension js/properties files must be a valid custom element name.

The biggest annoyances are the name must:

- Start with a lowercase ASCII letter (a-z)
- Contain at least one hyphen (-)

Instead of leaking this requirement from custom elements all the way down to the file names of the UI Extensions config, we should update [the webapp code](https://github.com/medic/cht-core/blob/c8238ec5f1acb1ccad18fd966f2e7ffbaf7061c1/webapp/src/ts/modules/ui-extensions/ui-extensions-tab.component.ts#L42) so that we just prepend the `cht-` prefix to the UI Extension name when creating the custom element.

```ts
const extensionId = this.route.snapshot.params['id'];
const elementName = `cht-${extensionId}`;
```

---

The other half of this issue is that we should then relax the [validation logic in cht-conf](https://github.com/medic/cht-conf/blob/badcc659e2e33f9eab98dca664e4a9b4830299db/src/lib/upload-ui-extensions.js#L20) to not have specific requirements about the first character and to not require a hyphen.  

---

Both PRs should target the `10224-ui-extensions` in their respective repositories.

YASHSHARMAOFFICIALLY:
## Summary
- Prepends `cht-` to the UI Extension `id` when defining and creating custom elements, so that [custom element naming requirements](https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name) are not leaked to the UI Extension config file names.
- Updates tests to match the new prefixed element name.

## Related Issue
Closes #10915

## Test plan
- [ ] Verify UI extensions still load correctly with the `cht-` prefixed custom element name
- [ ] Unit tests pass for `UiExtensionsTabComponent`


jkuester (approved):
Simple and effective!  I tested this and it works!

I also pushed an [update to cht-conf](https://github.com/medic/cht-conf/commit/3bf50812788756dba81f2ae6170d6780ea45de19) to relax the validation requirements accordingly. 👍

PR #10932, opened 2026-04-25
fix(#10931): fix broken aria-label binding and translate storage text

### LINKED ISSUE #10931: Storage progress bar has broken aria-label binding and hardcoded English text
**Describe the bug**

The storage pressure progress bar has two problems:

1. The `aria-label` uses plain string syntax instead of Angular property binding, so screen readers read the literal expression text `'Used storage: ' + storageUsagePercentage.toFixed(2) + '%'` instead of the actual value.

2. The "Available space:" label next to the progress bar is hardcoded in English and not using the translation pipeline.

**To Reproduce**

1. Open the webapp
2. Open the sidebar menu or hamburger menu where storage info is shown
3. Inspect the storage progress bar element
4. Notice `aria-label` contains the raw expression string, not the computed value
5. Switch app language to non-English — "Available space:" stays in English

**Expected behavior**

- The `aria-label` should use `[attr.aria-label]` binding so screen readers announce the actual storage percentage (e.g. "Used storage: 45.23%").
- "Available space:" should be a translated string.

**Screenshots**

Not applicable.

**Environment**

- App: webapp
- Files affected:
  - `webapp/src/ts/components/sidebar-menu/sidebar-menu.component.html` (lines 40, 43)
  - `webapp/src/ts/components/header/header.component.html` (lines 67, 70)

**Additional context**

The `[title]` binding on the same element works correctly because it uses `[title]` (property binding). The `aria-label` just needs the same treatment: `[attr.aria-label]`.

For the "Available space:" text, it should use the translate pipe like other strings in the app.

YASHSHARMAOFFICIALLY:
## Summary
- Fixes storage progress bar `aria-label` to use `[attr.aria-label]` property binding so screen readers announce the actual percentage value instead of the raw expression string.
- Translates hardcoded "Available space" and "Used storage" strings using the translate pipe.
- Adds `storage.available.space` and `storage.used` translation keys to `messages-en.properties`.

## Related Issue
Closes #10931

## Test plan
- [ ] Open sidebar menu / hamburger menu where storage info is shown
- [ ] Inspect the progress bar element — `aria-label` should show actual value like "Used storage: 45.23%"
- [ ] Switch to a non-English language and confirm "Available space" label is translated
- [ ] Run webapp unit tests


mrjones-plip:
Looks good @YASHSHARMAOFFICIALLY !  Once CI passes I'll approve to merge.

mrjones-plip (approved):

Issue #10871, opened 2026-04-15
Flaky api integration test: Import Records JSON parses and stores the passed JSON:


dianabarsan:
**Describe the issue**
Example run: 
https://github.com/medic/cht-core/actions/runs/24439489732/job/71402580926

Test fails with: 
```
 1) Import Records
       JSON
         parses and stores the passed JSON:
     Error: 500 - {"code":500,"error":"Server error","details":"Form not found: TEST"}
      at Object.request (/home/runner/work/cht-core/cht-core/tests/utils/index.js:266:15)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)

```

These are the api logs when the error happened: 
```
2026-04-15T06:39:27.551 RES: c6db58bcdb17 43.34.82.111 admin POST /medic-test/ HTTP/1.0 201 94 42.061 ms
2026-04-15T06:39:27.555 REQ: a6fe8c19f5a6 222.151.221.96 admin POST /api/v2/records HTTP/1.0
2026-04-15T06:39:27.571 ERROR: Server error: PublicError: Form not found: TEST
    at Object.createRecordByJSON (/service/api/src/services/records.js:227:11)
    at generate (/service/api/src/controllers/records.js:18:20)
    at /service/api/src/controllers/records.js:26:17
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5) {
  [stack]: [Getter/Setter],
  [message]: 'Form not found: TEST',
  publicMessage: 'Form not found: TEST'
} 
2026-04-15T06:39:27.572 RES: a6fe8c19f5a6 222.151.221.96 admin POST /api/v2/records HTTP/1.0 500 - 16.512 ms
```

**Describe the improvement you'd like**
Stabilize the test

sh1vam31:
Hi @dianabarsan,

I looked into this flaky test and found the root cause! **The Form not found:** TEST error happens because the test immediately tries to submit an API request before the API has finished loading the new TEST form into its memory.

In tests/utils/index.js, the updateSettings() function skips waiting for the API reload in integration tests unless we pass { ignoreReload: true } (which tells it to intelligently tail Docker logs until the API prints "Settings updated").

We can completely stabilize this test by simply updating the before hook in records.spec.js to use { ignoreReload: true } so it explicitly waits for the API cache to refresh before running the tests.

Could you please assign this issue to me?

vaibhav45sktech:
[...]

PR #10933, opened 2026-04-25
fix(#10871): stabilize records.spec.js by skipping --tail=20 historical lines in waitForLogs

taanvi2205:
# Description

Fixes the intermittent failure in `Import Records > JSON > parses and stores the passed JSON` (#10871).

**Root cause:** `waitForLogs` spawns `docker logs --tail=20 --timestamps` so the process starts outputting immediately (resolving the internal `ready` promise quickly). However, the same `data` handler that resolves `ready` also attempted regex matches against those 20 historical lines. If a previous settings change had logged `"Settings updated"` within the 50 ms `startTime` buffer, the watcher resolved `watcher.promise` immediately — before `updateCustomSettings` was even called. The `before()` hook returned with the `TEST` form not yet loaded into the API's in-memory config, causing the first test to fail with `Form not found: TEST`.

**Fix:** Skip regex matching entirely on the first data batch (which contains all the `--tail=20` historical lines). `startTime` is also moved into the first-batch handler so it is set after historical lines arrive, guaranteeing they all pre-date `startTime`. Subsequent batches contain only new log output and are matched normally.

Relates to: medic/cht-core#10871

# Code review checklist

# License

### DIFF
--- a/shared-libs/search/test/freetext-query.js
+++ b/shared-libs/search/test/freetext-query.js
@@ -95,4 +95,25 @@ describe('freetext-query', () => {
 
     chai.expect(result).to.deep.equal([]);
   });
+
+  it('should stop iterating once limit is reached', async () => {
+    const getUuidsByFreetext = sinon.stub().returns(createAsyncGenerator(['id1', 'id2', 'id3', 'id4', 'id5']));
+    sinon.stub(chtDatasource, 'getDatasource').returns({ v1: { report: { getUuidsByFreetext } } });
+
+    const request = { params: { key: 'term' } };
+    const result = await queryFreetext({}, request, 'reports', 3);
+
+    chai.expect(result).to.deep.equal([{ id: 'id1' }, { id: 'id2' }, { id: 'id3' }]);
+  });
+
+  it('should return all results when limit is not set', async () => {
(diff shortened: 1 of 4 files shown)


tests/utils/index.js
```diff
@@ -1489,7 +1489,9 @@ const waitForLogs = async (container, tail, ...regex) => {
   let timeout;
   let logs = '';
   let isReady = false;
-  const startTime = Date.now() - 50; // Subtract a small buffer to account for any clock skew
+  // startTime is set after the watcher is ready (after receiving first log output) to ensure
+  // historical lines from --tail=20 are not mistakenly matched as new logs.
+  let startTime;
```
dianabarsan, line 1494: I don't think this is correct, as this can create other race conditions - this code was added specifically because some older log lines were missed. 
Another approach is required.

shared-libs/search/test/freetext-query.js
```diff
@@ -95,4 +95,25 @@ describe('freetext-query', () => {
 
     chai.expect(result).to.deep.equal([]);
   });
+
+  it('should stop iterating once limit is reached', async () => {
```
dianabarsan, line 99: I believe these changes are unrelated.

dianabarsan (changes requested):

andrablaj:
Closing this PR as the [related issue](https://github.com/medic/cht-core/issues/10871) was closed.

PR #10935, opened 2026-04-26
feat: migrate uuid v4 to v7 for chronologically sortable ids (#10378)

### LINKED ISSUE #10378: Switch to UUID v7 everywhere
**Describe the issue**
UUID v7 is as random as UUID v4, but has an additional prefix nugget based on the unix timestamp. This enables uuids to be sortable based on the time when they were generated. I'm sure we will find some cool applications for this! 

CouchDb is already switching to using UUID v7 by default: https://github.com/apache/couchdb/pull/5699

Switch to UUID v7 everywhere - including PouchDb (somehow?). V7 is already available in the latest version of the [`uuid` library](https://www.npmjs.com/package/uuid#uuidv7options-buffer-offset)

Keep using uuid v4

sh1vam31:
# Description

### Fixes: #10378 

This PR implements a project-wide migration from UUID v4 to UUID v7 as requested in issue #10378. Moving to v7 ensures that all generated IDs (contacts, reports, tasks, etc.) are time-prefixed and chronologically sortable, which will significantly improve B-tree indexing performance in CouchDB.

### Changes:

**Source Code Migration:** Updated all instances of uuid.v4() to uuid.v7() (aliased as uuidv7() for clarity) in webapp, api, sentinel, and shared-libs.
**API Enhancements:** Modified api/src/routing.js to use UUID v7 for request IDs and increased shortUuid entropy from 12 to 16 characters to prevent collisions within the same millisecond.
**Test Suite Updates:** Performed a bulk update of all rosie factories and unit tests to align with the new UUID format.
Dependency Management: Standardized the use of the uuid package (v10+) and added @types/uuid to resolve TypeScript resolution issues in the webapp.
**IDE & Linting:** Fixed 100+ linting errors and shadowed variable issues introduced during the migration.
Verification Performed:

Verified chronological sortability of 100 generated IDs using a custom script.
Verified that all document IDs are explicitly assigned, maintaining compatibility with PouchDB's internal ID management.
Ran npm run lint project-wide (zero errors).
Verified core unit tests for api and webapp services.

# Code review checklist

# License


sh1vam31:

I have completed the migration from UUID v4 to UUID v7 across the codebase to enable chronologically sortable IDs for improved CouchDB performance.

Replaced uuid.v4 with uuid.v7 across the entire project (Source + Tests).
Verified chronological sortability of IDs for CouchDB performance.
Updated package-lock.json with necessary type definitions.

Kindly review this PR.

api/package.json
```diff
@@ -17,6 +17,7 @@
     "run-watch": "TZ=UTC nodemon -e js,json,html,properties --inspect=0.0.0.0:9229 --ignore 'build/static' --ignore 'build/public' --watch ./ --watch '../shared-libs/**/src/**' server.js -- --allow-cors"
   },
   "dependencies": {
-    "@medic/constants": "file:../shared-libs/constants"
+    "@medic/constants": "file:../shared-libs/constants",
+    "uuid": "^10.0.0"
```
dianabarsan: We do not add additional dependencies to individual apps. please remove this.

shared-libs/memdown/package.json
```diff
@@ -7,5 +7,8 @@
     "test": ""
   },
   "author": "",
-  "license": "Apache-2.0"
+  "license": "Apache-2.0",
+  "dependencies": {
```
dianabarsan: We do not add additional dependencies to individual apps. please remove this.

shared-libs/message-utils/package.json
```diff
@@ -9,6 +9,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@medic/logger": "file:../logger",
-    "@medic/phone-number": "file:../phone-number"
+    "@medic/phone-number": "file:../phone-number",
+    "uuid": "^10.0.0"
```
dianabarsan: We do not add additional dependencies to individual apps. please remove this.

shared-libs/transitions/src/transitions/index.js
```diff
@@ -6,7 +6,7 @@ const utils = require('../lib/utils');
 const logger = require('@medic/logger');
 const config = require('../config');
 const infodoc = require('@medic/infodoc');
-const uuid = require('uuid').v4;
+const uuid = require('uuid');
```
dianabarsan: Please use the format you use in other spaces: 
```
const { v7: uuid } = require('uuid');
``` 

so if we switch to a v8 in the future, there are less changes required.

shared-libs/transitions/package.json
```diff
@@ -26,6 +26,7 @@
     "@medic/settings": "file:../settings",
     "@medic/task-utils": "file:../task-utils",
     "@medic/user-management": "file:../user-management",
-    "@medic/validation": "file:../validation"
+    "@medic/validation": "file:../validation",
+    "uuid": "^10.0.0"
```
dianabarsan: We do not add additional dependencies to individual apps. please remove this.

tests/integration/transitions/message-duplicates.spec.js
```diff
@@ -1,5 +1,5 @@
 const chai = require('chai');
-const uuid = require('uuid').v4;
+const uuid = require('uuid');
```
dianabarsan: Please use the format you use in other spaces:
```
const { v7: uuid } = require('uuid');
```
so if we switch to a v8 in the future, there are less changes required.

tests/utils/gateway-api.js
```diff
@@ -1,5 +1,5 @@
 const testUtils = require('@utils');
-const uuid = require('uuid').v4;
+const uuid = require('uuid');
```
dianabarsan: Please use the format you use in other spaces:

```const { v7: uuid } = require('uuid');```
so if we switch to a v8 in the future, there are less changes required.

File: webapp/src/ts/services/contact-save.service.ts [hunk 8]
```diff
@@ -1,4 +1,4 @@
-import { v4 as uuidV4 } from 'uuid';
+import { v7 as uuidv7 } from 'uuid';
```
dianabarsan: Can you rename this to just `uuid` so we keep future changes smaller?

webapp/src/ts/services/enketo.service.ts
```diff
@@ -1,5 +1,5 @@
 import { Injectable, NgZone } from '@angular/core';
-import { v4 as uuid } from 'uuid';
+import { v7 as uuidv7 } from 'uuid';
```
dianabarsan: Please rename to `uuid`.

webapp/src/ts/services/feedback.service.ts
```diff
@@ -1,5 +1,5 @@
 import { Injectable } from '@angular/core';
-import { v4 as uuidv4 } from 'uuid';
+import { v7 as uuidv7 } from 'uuid';
```
dianabarsan: Please rename to uuid.

webapp/src/ts/services/send-message.service.ts
```diff
@@ -1,7 +1,7 @@
 import { Injectable } from '@angular/core';
 import { filter as _filter, flattenDeep as _flattenDeep, groupBy as _groupBy, uniqBy as _uniqBy } from 'lodash-es';
 import { Dictionary } from 'lodash';
-import { v4 as uuidv4 } from 'uuid';
+import { v7 as uuidv7 } from 'uuid';
```
dianabarsan: Please rename to uuid.

webapp/src/ts/services/telemetry.service.ts
```diff
@@ -1,6 +1,6 @@
 import { Inject, Injectable, NgZone } from '@angular/core';
 import { DOCUMENT } from '@angular/common';
-import { v4 as uuidv4 } from 'uuid';
+import { v7 as uuidv7 } from 'uuid';
```
dianabarsan: Please rename to uuid.

webapp/src/ts/services/training-cards.service.ts
```diff
@@ -1,7 +1,7 @@
 import { Injectable } from '@angular/core';
 import { Store } from '@ngrx/store';
 import * as moment from 'moment';
-import { v4 as uuid } from 'uuid';
+import { v7 as uuidv7 } from 'uuid';
```
dianabarsan: Please rename to uuid.

webapp/tests/karma/ts/services/db.service.spec.ts
```diff
@@ -6,7 +6,7 @@ import * as chaiExclude from 'chai-exclude';
 chai.use(chaiExclude);
 import { expect, assert } from 'chai';
 import { NgZone } from '@angular/core';
-import { v4 as uuidv4 } from 'uuid';
+import { v7 as uuidv7 } from 'uuid';
```
dianabarsan: Please rename to uuid.

webapp/tests/karma/ts/services/enketo-translation.service.spec.ts
```diff
@@ -369,7 +369,8 @@ describe('EnketoTranslation service', () => {
         </treatments>
       </instance>
       <bind nodeset="/treatments/group_test/chp/other_chp" required="true()" type="db:person"/>
-      <bind calculate="concat('uuid:', uuid())" nodeset="/treatments/meta/instanceID" readonly="true()" type="string"/>
+      <bind calculate="concat('uuid:', uuidv7())" nodeset="/treatments/meta/instanceID"
```
dianabarsan: is `uuidv7` defined in this context?

webapp/package-lock.json
```diff
@@ -48,14 +48,16 @@
         "select2": "4.0.13",
         "signature_pad": "2.3.x",
         "tslib": "^2.8.1",
+        "uuid": "^10.0.0",
```
dianabarsan: We do not add global packages to individual apps. Please remove this.

dianabarsan (changes requested):
Nice! I left some small comments inline

sh1vam31:
Hi @dianabarsan,
I've updated the PR with the following changes based on your feedback:
* Centralised Dependencies: Removed uuid from all sub-apps; it is now managed only at the root.
* Standardised Naming: Renamed all uuidv7 aliases and usages to just uuid for consistency.
* Enketo Context: Reverted form calculations to uuid() and updated the Enketo extensions to generate v7 IDs.
* Quality Check: All changes have passed local npm run lint.

Ready for another look! 
Thanks!

webapp/tests/karma/ts/services/enketo-translation.service.spec.ts
```diff
@@ -369,7 +369,8 @@ describe('EnketoTranslation service', () => {
         </treatments>
       </instance>
       <bind nodeset="/treatments/group_test/chp/other_chp" required="true()" type="db:person"/>
-      <bind calculate="concat('uuid:', uuid())" nodeset="/treatments/meta/instanceID" readonly="true()" type="string"/>
+      <bind calculate="concat('uuid:', uuid())" nodeset="/treatments/meta/instanceID"
```
dianabarsan: This change can be entirely reverted, correct? it's just empty space.

webapp/package-lock.json
```diff
@@ -12277,6 +12277,16 @@
         "websocket-driver": "^0.7.4"
       }
     },
+    "node_modules/sockjs/node_modules/uuid": {
```
dianabarsan, line 12280: Is this change still necessary?

webapp/src/ts/modules/testing/testing.component.ts
```diff
@@ -94,7 +94,7 @@ export class TestingComponent implements OnInit {
     let promise = Promise.resolve();
 
     for (let i = 0; i < parseInt(this.amountFeedbackDocs); i++) {
-      promise = promise.then(() => this.feedbackService.submit(uuid()));
+      promise = promise.then(() => this.feedbackService.submit({ message: uuid() }));
```
dianabarsan: Can you explain the reason for this change?

webapp/src/js/enketo/medic-xpath-extensions.js
```diff
@@ -205,6 +206,7 @@ module.exports = {
   },
   func: {
     'add-date': addDate,
+    'uuid': () => XPR.string(uuid()),
```
dianabarsan: I believe this change is not required at the moment. Let's default to the uuid implementation from Enketo (like we do for PouchDb).

dianabarsan (changes requested):
Nice turnaround! And super quick. A couple of questions inline.

webapp/src/ts/modules/testing/testing.component.ts
sh1vam31: The FeedbackService.submit method expects an object with a message property. Passing a string directly (as it was before) resulted in the feedback document having an undefined message field. 
So I updated this to pass an object { message: uuid() } so the generated ID is correctly stored in the message field.

sh1vam31:
Hi @dianabarsan,

Thanks for the review! I've addressed your feedback:

Reverted the accidental whitespace in the Enketo spec and the uuid() override in medic-xpath-extensions.
Regenerated the lockfile to remove redundant nested dependencies.
Kept the TestingComponent change as it fixes a bug where feedback messages were being saved as undefined (since submit expects an object).

dianabarsan:
@sh1vam31 you have a small linting error now

sh1vam31:
> @sh1vam31 you have a small linting error now

Thanks @dianabarsan,
Fixed! I accidentally left the unused uuid import after reverting the Enketo override. It's removed now.

webapp/package.json
```diff
@@ -62,6 +62,7 @@
     "select2": "4.0.13",
     "signature_pad": "2.3.x",
     "tslib": "^2.8.1",
+    "uuid": "^10.0.0",
```
dianabarsan: I don't believe we need to install this local uuid.

dianabarsan (changes requested):
one minor thing

webapp/package.json
```diff
@@ -66,7 +66,8 @@
   },
   "overrides": {
     "minimist": ">=1.2.6",
-    "esbuild": "0.25.10"
+    "esbuild": "0.25.10",
+    "uuid": "^10.0.0"
```
dianabarsan: Root already has "uuid": "^10.0.0" as a direct dependency (line 250), so webapp imports resolve fine without anything in webapp/package.json. What does this overrides entry actually fix? My read of the commit history is that some transitive in webapp's tree was resolving to an older uuid (probably <9, where v7() doesn't exist) and breaking the   build, and this pins it to v10.                                                                                                                                                                                                                                                                                                                             

If that's right, could you:                                                                                                                                                                                                                                                                                                                                 
1. Confirm by running npm ls uuid in webapp/ without the override and pasting which package brings in the older version, and
2. Add a one-line comment in webapp/package.json explaining what the override is for?                                                                                                                                                                                                                                                                       
If the override turns out to not be necessary (root's v10 already wins for everything), please drop it. Otherwise leave the comment so the next reader doesn't wonder. 

Locally, I tried removing the transitive dependency, i ran `npm install` in webapp and everything went fine. Please confirm.

dianabarsan (changes requested):
The PR description says "Replaced all uuid.v4() calls with uuid.v7()", but api/src/routing.js:187 (shortUuid) is still on v4. I see this got reverted mid-PR — the v7 timestamp prefix would collide for requests in the same millisecond when sliced to 12 hex chars. That's a legitimate reason to keep v4 here, but please either:                       
- leave a comment at the call site explaining why this one stays on v4 (and update the PR description), or                                                                                                                                                                                                                                                  
- replace it with crypto.randomBytes(6).toString('hex') since this is just a short correlation token, not a true UUID. That cleanly opts out of the version question for good.

Either is fine — but as-is, a future contributor will read the PR and assume the migration is total.

sh1vam31:
Thanks for the catch, @dianabarsan,

You were right about api/src/routing.js. I've updated shortUuid to use crypto.randomBytes(6).toString('hex') to completely sidestep the v4/v7 debate and cleanly generate a random string.

You were also absolutely correct about webapp/package.json. The webapp naturally inherits uuid@^10.0.0 from the root workspace, so the override wasn't necessary. I have removed it entirely.

The PR has been updated! Let me know if everything looks good to you.

webapp/src/ts/modules/testing/testing.component.ts
dianabarsan: thanks for the explanation. That is a real fix (the submit signature does expect an object), but it's unrelated to the UUID migration. Could you split it into a separate PR/commit? Easier to track and revert independently if needed.

dianabarsan (commented):
Great work! Thanks for all the changes. I left one small comment.

sh1vam31:
Hi @dianabarsan,
Thanks for the review! You are totally right, this fix is unrelated to the UUID migration. I have removed it from this PR and I am moving it to a separate, dedicated PR so it can be tracked independently.

sh1vam31:
Hi @dianabarsan,

I've reverted the change from this branch so the UUID migration stays clean, and I've moved the fix into its own dedicated PR here: #10991

dianabarsan (approved):
Sweet!

PR #10938, opened 2026-04-26
Fix RTL layout issues in Enketo drawing/signature widget

### LINKED ISSUE #10924: draw.scss: Make drawing widget controls RTL-aware
Description:
The Enketo drawing/signature/annotation widget in CHT uses hard-coded left positioning in draw.scss, which breaks layout in RTL languages (Arabic, Urdu, Hebrew).

Several instances (including file picker, file input, and hide-canvas button) are anchored using physical properties like left, causing controls to appear on the wrong side in RTL mode.

Fix:
Replaced physical positioning with CSS logical properties to support bidirectional layouts:

left: 0 → inset-inline-start: 0
left: 80px → inset-inline-start: 80px
left: 90px → inset-inline-start: 90px

Also updated one additional instance without an existing RTL TODO.

Impact:
Correct control alignment in RTL languages
No visual change in LTR
Scoped to webapp/src/css/enketo/draw.scss only

Why this version works better
Removes repetition (no need to list every code block)
Keeps problem → fix → impact structure
Signals low risk clearly (important for maintainers)
Keeps enough detail for reviewers to validate quickly

dhruveshmishra:
####  Problem

The Enketo drawing/signature/annotation widget used hard-coded physical positioning (`left`) in `draw.scss`, which caused layout issues in RTL languages such as Arabic, Urdu, and Hebrew.

Several UI elements (file picker, file input, hide-canvas button) were incorrectly aligned because they relied on `left` instead of direction-aware properties.

#### Solution

Replaced physical CSS positioning with logical properties to ensure proper support for bidirectional layouts:

*  left: 0 → inset-inline-start: 0
*  left: 80px → inset-inline-start: 80px
*  left: 90px → inset-inline-start: 90px

Also updated an additional instance that did not previously include an RTL TODO.

#### Impact

* Fixes control alignment issues in RTL languages
* No visual changes in LTR layouts
* Change is scoped to `webapp/src/css/enketo/draw.scss` only

####  Testing

* Verified layout in both LTR and RTL modes
* Confirmed correct alignment of controls in RTL environments

#### Risk

closes #10924


jkuester:
As I mentioned in the issue, I think we need to fix this upstream because we intend to completely remove this draw widget from cht-core in the future.  https://github.com/medic/cht-core/issues/10924#issuecomment-4328003081

PR #10939, opened 2026-04-27
fix(#7780): remove deprecated POST /api/v1/people endpoint

### LINKED ISSUE #7780: A person can be edited through `/api/v1/people`, if the correct body is passed
**Describe the bug**
[`/api/v1/people`](https://docs.communityhealthtoolkit.org/apps/reference/api/#post-apiv1people) is described as an endpoint that can create new people, and has an associated permission `can_create_people`. 
This endpoint _can_ be used to update an existing person if the correct body is sent, that contains the last `_rev` of the existent doc. Since there is no field validation, the endpoint can be used destructively, to delete an existent person document. 

Similarly, for places, we have two separate endpoints for create and edit, with the edit endpoint only allowing certain fields to be updated and using a different permission from the create endpoint.

**To Reproduce**
1. GET the contents of an existent person doc in the DB
2. Edit something, editing and adding new fields
3. POST this to `/api/v1/people`
4. Check the database and observe edits are saved

To delete, add a `_deleted: true` property to the document body.

**Expected behavior**
The endpoint should fail with some error if attempting to edit an existent document. 

**Environment**
- App: api
- Version: any?

vikrantwiz02:
# Description

The legacy `POST /api/v1/people` endpoint could be abused by any caller with `can_create_people` permission to silently overwrite or delete an existing person document by including `_rev` (and optionally `_deleted: true`) in the request body — bypassing the intended permission boundary entirely.

Following maintainer discussion (#7780), the agreed fix is to **remove the endpoint completely** rather than patch it. The modern replacement `POST /api/v1/person` (with `can_create_people`) and `PUT /api/v1/person/:uuid` (with `can_update_people`) already exist and cover all legitimate use cases.

Changes:
- Removed the `POST /api/v1/people` route handler and its OpenAPI JSDoc from `api/src/routing.js`
- Dropped the now-unused `people` import from `@medic/contacts` in `routing.js`
- `shared-libs/contacts/src/people.js` is unchanged — `createPerson()` is still used internally by `getOrCreatePerson()`, which is called by `user-management` and the `create_user_for_contacts` transition

Fixes #7780

# Code review checklist

- [x] Tested: No new tests required — endpoint is removed. Existing unit tests for `createPerson()` and `getOrCreatePerson()` are unaffected and still passing.

# License


witash (approved):
If this is intended for 6.0.0, cannot merge it yet, but can approve it now to mark it ready when the base branch is.

PR #10940, opened 2026-04-28
feat: add server-check to verify CHT CouchDB version (#9860)

### LINKED ISSUE #9860: Add server-check that verifies that the system is using the CHT version of CouchDb
**Describe the issue**
The CHT API will start even when it runs with an incompatible version of CouchDb, for example the default CouchDb image. There are a number of CouchDb configs that we know are incompatible with the CHT (multi replica, other js engines, etc - this is a non-exhaustive list). 

Add a server-check that verifies the version of CouchDb that is running, to make sure that we're on the correct version. I believe that we also need to add some sort of endpoint to the CouchDb image that returns the actual version, and compare that to the expected version. 

Leave it as it is. 

A recent forum support request revealed that a partner was using a default CouchDb image instead of the CHT one and they had issues with users getting logged out: https://forum.communityhealthtoolkit.org/t/extending-session-timeout-duration/4794/3

sh1vam31:
# Description

### Fixes: #9860

This PR implements a server-check during the CHT API startup sequence to verify that the system is running against a compatible CHT-customized version of CouchDB.

## Problem:
Currently, the CHT API starts silently even when connected to a standard "vanilla" CouchDB image. Standard images lack specific CHT optimisations and configurations (such as session handling and cluster settings), which leads to:
- Users being randomly logged out due to incorrect session timeouts.
- Potential data corruption or loss in multi-node clusters.

## Changes:

**CouchDB Fingerprinting:** Added a custom [cht] configuration section to the official CHT CouchDB image (10-docker-default.ini).

**API Verification:** Added a new check (checkCouchDbFlavor) in the server-checks library. This check attempts to read the CHT fingerprint from CouchDB at startup.

**Fail-fast:** If the fingerprint is missing, the API will fail to start and provide a clear error message directing the user to the official hosting documentation.

 ## Testing:

Added unit tests in shared-libs/server-checks/test/checks.js to verify:
- Success: API starts correctly when fingerprint is present.
- Failure: API refuses to start when fingerprint is missing (e.g., standard CouchDB).

# Code review checklist

# License

### DIFF
--- a/shared-libs/server-checks/src/checks.js
+++ b/shared-libs/server-checks/src/checks.js
@@ -107,6 +107,19 @@ const checkCouchDbVersion = async (couchUrl) => {
   console.log(`CouchDB Version: ${version}`);
 };
 
+const checkCouchDbFlavor = async (couchUrl) => {
+  try {
+    const response = await request.get({ url: `${couchUrl}_node/_local/_config/cht/version`, json: true });
+    if (!response) {
+      throw new Error('Empty response');
+    }
+  } catch {
+    throw new Error('This version of the CHT API is incompatible with the version of CouchDB you are using. ' +
+                    'Please make sure you are using the official CHT CouchDB image. ' +
+                    'See: https://docs.communityhealthtoolkit.org/core/install/hosting/docker/#couchdb');
+  }
+};
+
 module.exports = {
   checkNodeVersion,
   checkServerUrl,
@@ -114,4 +127,5 @@ module.exports = {
   checkCouchDbNoAdminPartyMode,
   checkCouchDbCluster,
   checkCouchDbSystemDbs,
+  checkCouchDbFlavor,
 };
--- a/couchdb/10-docker-default.ini
+++ b/couchdb/10-docker-default.ini
@@ -46,3 +46,6 @@ compression_level = 8
 [nouveau]
 enable = true
 url = http://nouveau:5987
+
+[cht]
+version = 5.1.0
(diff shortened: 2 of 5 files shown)


shared-libs/server-checks/src/checks.js
```diff
@@ -107,11 +107,25 @@ const checkCouchDbVersion = async (couchUrl) => {
   console.log(`CouchDB Version: ${version}`);
 };
 
+const checkCouchDbFlavor = async (couchUrl) => {
```
dianabarsan, line 110: I'm sorry, I don't think this is the right approach. Configuration can be added and removed easily. So someone could use vanilla CouchDb and add this config and get their services running, or someone can delete this config by mistake and stop their services completely. I think it's too fragile. 

I don't think I have any other better ideas right now. I'll come back to this at a later time.

dianabarsan (changes requested):

shared-libs/server-checks/src/checks.js
sh1vam31: Hi @dianabarsan,
Thank you for the feedback! I see your point—relying on a configuration value is indeed quite fragile as it can be easily spoofed or accidentally deleted.

I'll put this PR on hold for now. Please let me know if you or the team come up with a more robust way to distinguish the CHT CouchDB image.

dianabarsan: Awesome, thank you! Sorry I don't have more ideas to provide here. I'll convert the PR as draft so it's not confused with a PR that needs review.

PR #10943, opened 2026-04-28
test: add automated validation for shared-libs test runner script

### LINKED ISSUE #10870: Add automated test to validate shared-libs testing script
**Describe the issue**
As part of https://github.com/medic/cht-core/issues/8799, a new script runs the shared-libs unit tests, but it itself is not tested. 

We should add automation to validate the output of this script and validate that it will indeed result in an error if any of the shared libs will error.

sh1vam31:
# Description

### Fixes: #10870

This PR adds automated validation for the test-shared-libs.sh script to ensure it reliably detects and reports failures in the CI pipeline.

Previously, the script was not tested, which posed a risk that logic bugs could lead to unreported failures in the shared-libs unit tests.

### Changes
**Runner Script Enhancement:** Updated scripts/ci/test-shared-libs.sh to allow the library directory (LIBS_DIR) to be overridden via an environment variable. This enables testing the script against controlled mock environments.

**New Integration Test:** Added tests/integration/scripts/test-shared-libs-runner.spec.js which uses Mocha to validate:
- Successful exit code (0) when all sub-libraries pass.
- Non-zero exit code when any sub-library fails.
- Correct aggregation of failures in the final summary.
- Graceful skipping of non-project directories.

# Code review checklist

# License


sh1vam31:
Hi @dianabarsan, 
I've added the automated validation for the shared-libs test runner (#10870).

The new test suite in tests/integration/scripts/test-shared-libs-runner.spec.js verifies that the script correctly detects failures, aggregates multiple errors in the summary, and skips non-project folders. All tests and linting are passing.

Ready for review

dianabarsan (changes requested):
Solid scope — the env-var override is the right escape hatch, and the four cases match the PR description. A few things before merging:                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                
  **1. Move the test to `webapp/tests/mocha/unit/testingtests/`.** The test file currently lives in `tests/integration/scripts/`, which is picked up by `tests/integration/.mocharc-all.js` via the `all` glob in `specs.js` (`tests/integration/!(cht-conf|sentinel)/**/*.spec.js`). That means `npm run ci-integration-all` does run it — but in the          
  integration lane, which spins up Docker, CouchDB, and sentinel, runs serially, and uses a 200-second default timeout. None of that is needed for a test that spawns bash and asserts on stdout. It'd run in ~50ms in the webapp unit-mocha lane instead of holding up the integration pipeline.
                                                                                                                                                                                                                                                                                                                                                                
  This follows the precedent set by the secretlint config test in #10857 — same situation (a CI utility script, not webapp source). The webapp unit:mocha runner picks up `tests/mocha/unit/**/*.spec.*` per `webapp/tests/mocha/.mocharc.js`, so the file just needs to land in that folder.                                                                   
   
  You'll need to update the relative paths from 3-up to 5-up:                                                                                                                                                                                                                                                                                                   
                  
  - `path.resolve(__dirname, '../../../../../scripts/ci/test-shared-libs.sh')`                                                                                                                                                                                                                                                                                  
  - `cwd: path.resolve(__dirname, '../../../../../')`
  - And the mock-libs location — see point 3 below; better to put it under `os.tmpdir()` while you're at it.                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                
  **2. The most complex logic in the script is uncovered.** Lines 62–87 of `test-shared-libs.sh` are a sed+awk pipeline that extracts failing-test names from Mocha output, plus a similar block at 91–97 for failure details with a 15-line `tail` fallback. All four mock libs here use `exit 0` / `exit 1` — they produce no Mocha output, so the awk parsing
   always finds nothing and the fallback path wins every time. A regression in the parsing would never be caught.                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                
  Could you add a test that uses a mock lib whose `test` script echoes realistic Mocha output (a `1) failing test name:` block followed by an Error stack and `1 failing` summary), then asserts both the short summary and the details section in the final output? That's the part of the script most likely to drift when Mocha versions change.             
            
  **3. Mock directory leak risk.** `tests/integration/scripts/mock-shared-libs/` (or wherever it lands after the move) is created at runtime. If a test crashes between `beforeEach` and `after`, it's left in the working tree — easy to accidentally commit. Either add it to `.gitignore`, or — cleaner — move the runtime fixtures under `os.tmpdir()`. That
   avoids the issue entirely and also removes the need for the test to know the project-relative path of the mock dir.                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                
  **4. Minor: missing edge case.** What happens when `LIBS_DIR` is set to a non-existent path? The script silently exits 0 with no work done (the glob expands literally, the `[[ -d ]]` skip handles it). Worth a test that documents this behavior so a future change doesn't accidentally turn it into a hard error or the reverse.

sh1vam31:
Hi @dianabarsan

Thanks for the detailed review! I've addressed all four points:

Moved the test to webapp/tests/mocha/unit/testingtests/ and updated paths to 5-up.
Added a test with realistic Mocha output to exercise the sed+awk parsing pipeline.
Switched mock dirs to os.tmpdir() to prevent working tree pollution.
Added an edge case test documenting the silent exit 0 behaviour when LIBS_DIR doesn't exist.

dianabarsan (changes requested):
Two things on this step:                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                
  **1. Pattern propagation — the OS-Process filter only got updated in the kubectl job above.** The comment + tightened regex (`grep -Ev "OS Process #?Port|OS Process .* Input ::"`) landed there, but this job still has the original single-string `grep -v "OS Process #Port"` and no comment. Both jobs inherit `COUCHDB_LOG_LEVEL: 'debug'` from line 19, 
  so they ingest the same CouchDB dumps and need the same filter and rationale. If CouchDB changes the log format, this job silently starts failing while the other keeps working.
                                                                                                                                                                                                                                                                                                                                                                
  Could you mirror the block from the kubectl job here?

  ```yaml
# Strip CouchDB OS-Process debug dumps which legitimately log user creation doc bodies (with passwords). 
for f in tests/logs/*.log; do grep -Ev "OS Process #?Port|OS Process .* Input ::" "$f" > "$tmpdir/$(basename "$f")"; done
```
                                                                                                                                                                                                                                                                                                                                                        
  2. Do we still need npx --no-install now that secretlint is a direct devDependency? After npm install, the binary lives at node_modules/.bin/secretlint and is directly executable — the spec file at webapp/tests/mocha/unit/testingtests/secretlintrc.spec.js already invokes it that way. Two cleaner options for the workflow:                            
                                                                                                                                                                                                                                                                                                                                                                
  - Direct path: ./node_modules/.bin/secretlint --secretlintrc ...                                                                                                                                                                                                                                                                                              
  - npm script: add "scan-logs": "secretlint --secretlintrc scripts/ci/.secretlintrc.json" to root package.json and call npm run scan-logs -- "$tmpdir"/*.log.
                                                                                                                                                                                                                                                                                                                                                                
  The npm-script form is probably the most idiomatic here and makes it consistent across both jobs. Same comment applies to the kubectl job above.

dianabarsan:
Apologies, I added the comment in the wrong PR.

dianabarsan (approved):
Awesome!

PR #10950, opened 2026-04-29
fix(#10705): stop consuming rate limit on 429 responses

### LINKED ISSUE #10705: Rate limiter can block requests with valid credentials that are unrelated to auth failures
**Describe the bug**

The API rate limiter (`api/src/services/rate-limit.js`) tracks failed login attempts by multiple keys: client IP (`req.ip`), username, and password. It allows 10 failed attempts per 10-second window per key. When any key is exhausted, all requests matching that key are rejected with `429 Too Many Requests`.

Unrelated services with valid credentials (notably couch2pg) have been observed receiving `429 Too Many Requests` errors in a production deployment. 

This could happen when a reverse proxy forwards requests to the API without preserving the real client IP. When `req.ip` resolves to the proxy's internal IP for all requests, the IP-based rate limit key becomes shared across all users and services. Then failed login attempts from one client can trigger rate limiting for every client on the instance.

Also `429` responses also trigger `consume()`, so continued retries from any client continue the lockout.

**Impact**

If the proxy IP is shared across all requests, then 10 failed authentication attempts from any combination of users or services within a 10-second window will block every login request or request using basic authentication instance-wide.

Affected clients include:
- **couch2pg** and other automated scripts using basic auth for every request
- **Any user logging in** via the login page or API during the lockout window; although this hasn't been reported directly.
- **Other services** (e.g., interoperability tools, custom integrations) using basic auth

**To Reproduce**

The exact proxy configuration that triggers this behavior has not yet been confirmed. Using docker compose on an EC2 instance does **not** reproduce the issue, even though it does use nginx; the IPs that the api container gets in that case are the actual IPs, not proxy IPs. I still suspect this is happening on k3s deployments with an upstream nginx reverse proxy.

To confirm, we could check the API logs for the IP addresses associated with requests.

** Proposed fix **

It may be possible for the rate-limiter to use the X-Forwarded-For header instead of the proxy IP.

** Proposed workaround for couch2pg (no release required) **

couch2pg currently uses basic authentication on every request. It can be changed to use session-based authentication instead. Although couch2pg is supposed to be deprecated, this small change could bypass the issue for deployments that are unable or unwilling to upgrade to the version with a fix in cht-core.

officialasishkumar:
# Description

The api rate-limit middleware tracks failed login attempts by client IP, basic-auth username, basic-auth password, and the body `user`/`password` fields, allowing 10 failures per key per 10 second window. Until now the finish handler also called `rateLimitService.consume` whenever a response went out with status code 429.

That double-counted the rate limiter's own response. When several clients share a single key (most importantly a single `req.ip`, because a reverse proxy is not forwarding the real client address), one badly behaved client can exhaust that shared key, after which every other caller gets 429 from the controller. Each of those 429 responses then re-consumed the victim's other keys (their valid username, their valid password), which could lock those credentials out instance-wide even though they had never sent a failing request. The reporter observed exactly this affecting `couch2pg` and other basic-auth integrations on production.

Limiting the finish handler to only authentication failures (401) keeps the protective behaviour against real brute-force attempts while letting unrelated callers recover as soon as the existing 10 second window elapses. The middleware still rejects requests upfront with 429 when the caller's keys are already over the limit; this PR only changes whether *that* rejection should re-punish the caller.

## Verification

`npm run unit-api` for the affected suite (`api/tests/mocha/middleware/rate-limiter.spec.js`):

- The pre-existing `consumes on 401` and `does nothing if not limited` cases still pass.
- The previous `consumes on 429` test, which asserted the buggy behaviour, has been replaced with `does not consume on 429 (#10705)` that asserts no consume call happens.
- New `does not consume on non-401 status codes` test guards the implicit contract that 200/4xx-other-than-401/5xx responses also do not trigger consumption.

`npm run lint` is clean for both modified files.

Fixes #10705

# Code review checklist

- [x] Tested: Unit tests cover the previous and new behaviour
- [x] Backwards compatible: The middleware still issues 429 for callers already over the limit and still consumes on real 401s; only the redundant double-consume on the rate limiter's own 429 has been removed.

# License

### DIFF
--- a/api/src/middleware/rate-limiter.js
+++ b/api/src/middleware/rate-limiter.js
@@ -4,8 +4,12 @@ const serverUtils = require('../server-utils');
 
 const registerReqFinishHandler = (req, res) => {
   res.on('finish', () => {
-    if (res.statusCode === 401 || res.statusCode === 429) {
-      // log in failed - punish user
+    // Only authentication failures (401) increment the failed-login counter.
+    // 429 is the rate-limiter's own response, so consuming on it would
+    // re-punish callers who happen to share a key (e.g. a proxy IP) with a
+    // genuinely failing client and lock out their otherwise-valid
+    // credentials. See #10705.
+    if (res.statusCode === 401) {
       rateLimitService.consume(req);
     }
   });
--- a/api/tests/mocha/middleware/rate-limiter.spec.js
+++ b/api/tests/mocha/middleware/rate-limiter.spec.js
@@ -80,15 +80,29 @@ describe('Rate limiter middleware', () => {
     expect(rateLimitService.consume.args[0][0]).to.equal(req);
   });
 
-  it('consumes on 429', async () => {
+  it('does not consume on 429 (#10705)', async () => {
+    // 429 is the rate limiter's own response. Consuming on it would
+    // double-punish callers that happen to share a key (e.g. a proxy IP)
+    // with a genuinely-failing client and would lock out their valid
+    // credentials, so the finish handler must ignore it.
     rateLimitService.isLimited.resolves(false);
     res.statusCode = 429;
     await middleware(req, res, next);
     expect(next.callCount).to.equal(1);
     expect(rateLimitService.consume.callCount).to.equal(0);
     finish();
-    expect(rateLimitService.consume.callCount).to.equal(1);
-    expect(rateLimitService.consume.args[0][0]).to.equal(req);
+    expect(rateLimitService.consume.callCount).to.equal(0);
+  });
+
+  it('does not consume on non-401 status codes', async () => {
+    rateLimitService.isLimited.resolves(false);
+    for (const statusCode of [200, 400, 403, 404, 500]) {
+      res = { on: sinon.stub(), statusCode };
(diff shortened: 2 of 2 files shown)


witash:
> That double-counted the rate limiter's own response.

Hi @officialasishkumar consuming on 429 is not really double counting; the intention is to keep rate limiting until the number of failed requests actually reduces; otherwise, rate limited clients will stop being rate limited because they are receiving 429 instead of 401, even if the number of requests doesn't actually reduce.

Which is not necessarily wrong either, they would get rate limited again if they kept sending requests in the next 10 second window. 

> When several clients share a single key (most importantly a single req.ip, because a reverse proxy is not forwarding the real client address), one badly behaved client can exhaust that shared key, after which every other caller gets 429 from the controller

The root cause is that clients should not share a single key if they are not actually the same client. I think would rather leave the existing behavior around 429s and fix the root cause (which could be [allowing opting out or using a forwarded IP](https://github.com/medic/cht-core/issues/10705#issuecomment-4333862957))

witash:
closing because of above comment

PR #10954, opened 2026-04-30
fix(#10953): fix memory leak from unmanaged subscription in FastActionButtonComponent

### LINKED ISSUE #10953: Memory leak: unmanaged store subscription in FastActionButtonComponent constructor
## Description

In `webapp/src/ts/components/fast-action-button/fast-action-button.component.ts`, there is a store subscription created in the constructor (line 59) that is **not added** to the component's `subscriptions` collection:

```typescript
constructor(...) {
  this.store.select(Selectors.getDirection).subscribe(direction => {
    this.direction = direction;
  });
}
```

The component correctly manages other subscriptions in `ngOnInit()` by adding them to `this.subscriptions` (a `Subscription` instance), which is then cleaned up in `ngOnDestroy()`:

```typescript
ngOnDestroy() {
  this.subscriptions.unsubscribe(); // does NOT unsubscribe the direction subscription
}
```

However, the `direction` subscription is never tracked, so it **leaks** every time the component is created and destroyed (e.g., when navigating between tabs).

## Steps to Reproduce

1. Open the webapp
2. Navigate between different tabs (messages, tasks, contacts, reports) — each navigation creates and destroys `FastActionButtonComponent`
3. Each time, a new leaked subscription to the store's `direction` selector accumulates

## Expected Behavior

All store subscriptions should be properly managed and cleaned up when the component is destroyed.

## Suggested Fix

Move the subscription from the constructor into `ngOnInit()` and add it to the `subscriptions` collection:

```typescript
ngOnInit() {
  const directionSubscription = this.store
    .select(Selectors.getDirection)
    .subscribe(direction => this.direction = direction);
  this.subscriptions.add(directionSubscription);
  
  this.subscribeToStore();
  this.subscribeToRouter();
}
```

## Impact

- Memory leak on repeated navigation
- Accumulated stale callbacks in long-running sessions
- Particularly impactful on low-resource devices used by CHWs

YASHSHARMAOFFICIALLY:
## Description

Fixes #10953

The `FastActionButtonComponent` had a store subscription for `Selectors.getDirection` created in the constructor that was **not added** to the component's `subscriptions` collection. This meant it was never unsubscribed when the component was destroyed via `ngOnDestroy()`, causing a memory leak on every navigation between tabs.

## Changes

- Moved the `direction` subscription from the constructor into a new `subscribeToDirection()` method
- Called `subscribeToDirection()` in `ngOnInit()` alongside existing subscription setup methods
- Added the subscription to `this.subscriptions` so it is properly cleaned up in `ngOnDestroy()`

## Before (leaking)
```typescript
constructor(...) {
  // This subscription is never unsubscribed
  this.store.select(Selectors.getDirection).subscribe(direction => {
    this.direction = direction;
  });
}
```

## After (properly managed)
```typescript
private subscribeToDirection() {
  const directionSubscription = this.store
    .select(Selectors.getDirection)
    .subscribe(direction => this.direction = direction);
  this.subscriptions.add(directionSubscription);
}
```

## Test plan

- [x] Existing unit tests pass (subscription is already mocked via `provideMockStore` in specs)
- [ ] Verify no regression in FAB button behavior across tabs (messages, tasks, contacts, reports)
- [ ] Confirm RTL direction is still applied correctly to dialog


YASHSHARMAOFFICIALLY:
Hey team, I found this memory leak while going through the codebase looking for contribution opportunities.

The subscription to Selectors.getDirection was created in the constructor but never added to the component's subscription collection, so it leaked on every mount/unmount cycle. I moved it into ngOnInit() following the same pattern the component already uses for its other subscriptions.

Existing unit tests cover the direction selector mock through provideMockStore, so no new tests were needed. Happy to make any changes if needed!

I'm also applying for GSoC 2026 with CHT and would love to keep contributing. Let me know if there's anything else I can help with.

@andrablaj @dianabarsan @jkuester would appreciate your review when you get a chance, thanks!

dianabarsan (changes requested):
Nice find — the leak is real and the fix is correct.                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                
  Two things before merging:                                                                                                                                                                                                                                                                                                                                    
   
  **1. Same leak in a sibling component.** `webapp/src/ts/components/filters/date-filter/date-filter.component.ts:45` has an identical pattern — constructor-level `this.store.select(Selectors.getDirection).subscribe(...)` with no tracking and no cleanup. Same selector, same shape. I swept `webapp/src/ts/components/` and `modules/` for                
  constructor-body `.subscribe()` calls — these were the only two hits in the codebase. Could you fix that one in this PR too? Same change, very small.
                                                                                                                                                                                                                                                                                                                                                                
  **2. Please add a unit test that asserts subscription cleanup.** The existing tests pass whether or not the leak is fixed, so they prove nothing about this PR's value. A meaningful test would dispatch a state change after `component.ngOnDestroy()` and verify `component.direction` doesn't update — or simply spy on the `Subscription` and assert      
  `.unsubscribe()` was called once on destroy. That locks in the behavior and prevents regressions.
                                                                                                                                                                                                                                                                                                                                                                
  Also — your PR description's test plan has two unchecked items (cross-tab regression, RTL direction). Could you confirm you've verified those manually in a browser? Moving the subscribe from constructor to `ngOnInit` shifts when `this.direction` is first set; in practice it should be fine (NgRx selectors emit synchronously on subscribe) but I'd    
  like confirmation rather than inference.

YASHSHARMAOFFICIALLY:
@dianabarsan @andrablaj  Thank you for the detailed review. Addressed both items in the follow-up commit (2ee95d77):

1. **DateFilterComponent fix**: Moved the `Selectors.getDirection` subscription from the constructor to `ngOnInit` and added it to the tracked `this.subscription` collection, same pattern as the FastActionButtonComponent fix.

2. **Unit tests**: Added cleanup assertions for both components. Each test sets the initial direction, calls `ngOnDestroy()`, then dispatches a state change via `MockStore` and verifies the component property does not update.

Regarding manual verification: Confirmed both in a browser session.
- Cross-tab regression: Navigated between Reports, Contacts, and Messages tabs with the date filter active. Filter state and direction applied correctly on each tab without stale values.
- RTL direction: Tested with an RTL locale. The datepicker `opens` and dialog `direction` properties reflected the correct direction after the subscription was moved to `ngOnInit`. NgRx selectors emit synchronously on subscribe, so the value is available before `ngAfterViewInit` runs the daterangepicker initialization.

Ready for re-review when you get a chance.

dianabarsan (approved):
⭐

PR #10961, opened 2026-04-30
perf(#10960): optimize O(n*m) lookups in contact report loading

### LINKED ISSUE #10960: Optimize O(n²) lookups in contact-view-model-generator report loading
## Description

The `ContactViewModelGeneratorService` in `webapp/src/ts/services/contact-view-model-generator.service.ts` uses repeated linear Array.find() calls inside loops when loading reports for the contact detail page. This creates O(n * m) time complexity that degrades noticeably as the number of reports and contacts grows.

With the default limit of 500 reports (`LIMIT_SELECT_ALL_REPORTS = 500`) and potentially hundreds of child contacts under a facility, these nested scans can cause visible lag on low end devices used by health workers.

### Location 1: \`addHeading()\` (lines 279 to 288)

\`\`\`typescript
summaries?.forEach(summary => {
  const report = reports.find(report => report._id === summary._id);
  // ...
});
\`\`\`

For each summary, a full linear scan of the reports array is performed. With 500 reports, this is up to 250,000 comparisons.

### Location 2: \`getReports()\` (lines 296 to 310)

\`\`\`typescript
reports.forEach(report => {
  const patientId = report.fields.patient_id || report.patient_id;
  const patient = contactDocs.find(contact => contact?.patient_id === patientId);
  // ...
});
\`\`\`

For each report, a linear scan of all contact docs is performed to resolve the patient name.

## Suggested Fix

Replace the linear scans with Map based lookups for O(1) access:

1. In \`addHeading()\`: Build a \`Map<string, report>\` from the reports array keyed by \`_id\` before iterating summaries
2. In \`getReports()\`: Build a \`Map<string, contact>\` from contactDocs keyed by \`patient_id\` before iterating reports

This reduces both operations from O(n * m) to O(n + m).

## Impact

This optimization is on the critical rendering path for the contact detail page, the most frequently used page in the CHT webapp. Health workers navigating to a patient or household with many reports will see faster load times, especially on low resource Android devices common in CHT deployments.

YASHSHARMAOFFICIALLY:
## Description

Fixes #10960

The contact detail page is the most frequently visited page in the CHT webapp. When loading reports for a contact, `ContactViewModelGeneratorService` was performing repeated `Array.find()` calls inside loops, resulting in O(n * m) time complexity that scales poorly as report and contact counts grow.

With `LIMIT_SELECT_ALL_REPORTS = 500` and potentially hundreds of child contacts under a facility, the worst case was ~250,000 comparisons per page load on the critical rendering path.

### 1. `addHeading()` — report and form lookups

**Before:** For each summary, linearly scans the full reports array and uses lodash `_find` to scan the forms array.
```typescript
summaries?.forEach(summary => {
  const report = reports.find(report => report._id === summary._id);
  // getHeading() internally does: _find(forms, { code: report.form })
});
```

**After:** Pre-builds `Map<id, report>` and `Map<code, form>` for O(1) lookups.
```typescript
const formsMap = new Map(forms.map(form => [form.code, form]));
const reportsById = new Map(reports.map(report => [report._id, report]));
summaries?.forEach(summary => {
  const report = reportsById.get(summary._id);
  // getHeading() now does: formsMap.get(report.form)
});
```

### 2. `getReports()` — patient name resolution

**Before:** For each report, linearly scans all contact docs to find the matching patient.
```typescript
reports.forEach(report => {
  const patient = contactDocs.find(contact => contact?.patient_id === patientId);
});
```

**After:** Pre-builds `Map<patientId, contact>` for O(1) lookups.
```typescript
const contactsByPatientId = new Map(
  contactDocs.filter(doc => doc?.patient_id).map(doc => [doc.patient_id, doc])
);
reports.forEach(report => {
  const patient = contactsByPatientId.get(patientId);
});
```

### 3. Removed unused lodash `_find` import

The `find` import from lodash-es is no longer needed after replacing it with native Map lookups, slightly reducing the bundle dependency surface.

## Complexity analysis

| Method | Before | After |
|--------|--------|-------|
| `addHeading()` report lookup | O(summaries * reports) | O(summaries + reports) |
| `getHeading()` form lookup | O(forms) per call | O(1) per call |
| `getReports()` patient lookup | O(reports * contacts) | O(reports + contacts) |

## Test plan

- [x] Existing unit tests in `contact-view-model-generator.service.spec.ts` cover all modified code paths (heading resolution, patient name assignment, report sorting)
- [ ] Verify contact detail page loads correctly with reports showing proper headings and patient names
- [ ] Spot check a contact with many reports to confirm no regressions

### DIFF
--- a/webapp/src/ts/services/contact-view-model-generator.service.ts
+++ b/webapp/src/ts/services/contact-view-model-generator.service.ts
@@ -277,8 +277,8 @@ export class ContactViewModelGeneratorService {
   }
 
   private async addHeading(reports, forms) {
-    const formsMap = new Map(forms.map(form => [form.code, form]));
-    const reportsById = new Map(reports.map(report => [report._id, report]));
+    const formsMap: Map<string, any> = new Map(forms.map(form => [form.code, form]));
+    const reportsById: Map<string, any> = new Map(reports.map(report => [report._id, report]));
     const summaries = await this.getDataRecordsService.getDocsSummaries(reports);
     summaries?.forEach(summary => {
       const report = reportsById.get(summary._id);
@@ -293,7 +293,7 @@ export class ContactViewModelGeneratorService {
     const subjectIds: string[] = [];
     contactDocs.forEach(doc => subjectIds.push(...registrationUtils.getSubjectIds(doc)));
 
-    const contactsByPatientId = new Map(
+    const contactsByPatientId: Map<string, any> = new Map(
       contactDocs
         .filter(doc => doc?.patient_id)
         .map(doc => [doc.patient_id, doc])
--- a/webapp/src/ts/services/contact-view-model-generator.service.ts
+++ b/webapp/src/ts/services/contact-view-model-generator.service.ts
@@ -1,5 +1,5 @@
 import { Injectable, NgZone } from '@angular/core';
-import { find as _find, groupBy as _groupBy, partial as _partial, } from 'lodash-es';
+import { groupBy as _groupBy, partial as _partial, } from 'lodash-es';
 
 import registrationUtils from '@medic/registration-utils';
 
@@ -262,8 +262,8 @@ export class ContactViewModelGeneratorService {
       });
   }
 
-  private getHeading(report, forms) {
-    const form = _find(forms, { code: report.form });
+  private getHeading(report, formsMap: Map<string, any>) {
+    const form = formsMap.get(report.form);
     if (form && form.subjectKey) {
       return this.translateService.instant(form.subjectKey, report);
     }
@@ -277,11 +277,13 @@ export class ContactViewModelGeneratorService {
   }
 
   private async addHeading(reports, forms) {
+    const formsMap = new Map(forms.map(form => [form.code, form]));
+    const reportsById = new Map(reports.map(report => [report._id, report]));
     const summaries = await this.getDataRecordsService.getDocsSummaries(reports);
     summaries?.forEach(summary => {
-      const report = reports.find(report => report._id === summary._id);
+      const report = reportsById.get(summary._id);
       if (report) {
-        report.heading = this.getHeading(summary, forms);
+        report.heading = this.getHeading(summary, formsMap);
       }
     });
     return reports;
@@ -290,6 +292,12 @@ export class ContactViewModelGeneratorService {
   private getReports(contactDocs) {
     const subjectIds: string[] = [];
     contactDocs.forEach(doc => subjectIds.push(...registrationUtils.getSubjectIds(doc)));
+
+    const contactsByPatientId = new Map(
+      contactDocs
+        .filter(doc => doc?.patient_id)
+        .map(doc => [doc.patient_id, doc])
+    );
 
     return this.searchService
       .search('reports', { subjectIds }, { include_docs: true, limit: this.LIMIT_SELECT_ALL_REPORTS })
@@ -301,7 +309,7 @@ export class ContactViewModelGeneratorService {
             return;
           }
           const patientId = report.fields.patient_id || report.patient_id;
-          const patient = contactDocs.find(contact => contact?.patient_id === patientId);
+          const patient = contactsByPatientId.get(patientId);
           if (patient) {
             report.fields.patient_name = patient.name;
           }


YASHSHARMAOFFICIALLY:
@andrablaj @dianabarsan @jkuester would love your review on this when you get a chance, thanks!

dianabarsan (changes requested):
@YASHSHARMAOFFICIALLY can you please share a performance comparison for both implementations?
We do not expect contact pages to have 500 reports consistently. So please run the comparison when there are 10, 50, 100, 250 and 500 reports.

Ideally the performance comparison would come from telemetry documents showing how much faster the contact page loads.

YASHSHARMAOFFICIALLY:
@dianabarsan Here is the performance comparison across the requested data sizes.

### Benchmark setup

Isolated microbenchmark simulating the two hot paths changed in this PR:
1. **`addHeading`**: matching summaries to reports (by `_id`) and reports to forms (by `code`)
2. **`getReports`**: matching reports to contacts (by `patient_id`)

Each measurement: 10,000 iterations at the given report count.

### Results: `addHeading` — Array.find vs Map.get

| Reports | Array.find (ms) | Map.get (ms) | Speedup |
|---------|-----------------|--------------|---------|
| 10      | 54.58           | 23.92        | 2.3x    |
| 50      | 240.57          | 62.16        | 3.9x    |
| 100     | 356.84          | 104.28       | 3.4x    |
| 250     | 1,610.13        | 252.53       | 6.4x    |
| 500     | 8,316.57        | 533.44       | 15.6x   |

### Results: `getReports` — Array.find vs Map.get

| Reports | Array.find (ms) | Map.get (ms) | Speedup |
|---------|-----------------|--------------|---------|
| 10      | 15.63           | 14.42        | 1.1x    |
| 50      | 99.27           | 43.48        | 2.3x    |
| 100     | 211.12          | 70.34        | 3.0x    |
| 250     | 296.22          | 124.22       | 2.4x    |
| 500     | 497.73          | 213.32       | 2.3x    |

### Key observations

- At **realistic page sizes (10-50 reports)**, the improvement is modest but consistent (2-4x for `addHeading`, 1-2x for `getReports`).
- At **250+ reports**, `addHeading` benefits significantly because of the double lookup (report-by-id + form-by-code), where the O(n*m) cost compounds.
- `getReports` shows a steadier 2-3x improvement because the contact list is smaller (bounded by distinct `patient_id` values on the page).
- The Map construction cost is negligible at all tested sizes.

I don't have a running CHT instance with enough seeded data to capture telemetry documents for real page load times. If you'd prefer that measurement, I can set one up, but wanted to share these isolated numbers first since they directly measure the changed code paths.

dianabarsan (changes requested):
Thanks for putting these together, @YASHSHARMAOFFICIALLY — the numbers are helpful, but I want to flag how they read once we translate them into per-page-load terms.

  Each measurement is 10,000 iterations, so dividing back out to a single page load:

  | Reports | addHeading saved/load | getReports saved/load | Total saved / load |
  |--------|----------------------|----------------------|--------------------|
  | 10  | ~3 µs    | ~0 µs   | ~0.003 ms |
  | 50  | ~18 µs   | ~6 µs   | ~0.02 ms  |
  | 100 | ~25 µs   | ~14 µs  | ~0.04 ms  |
  | 250 | ~136 µs  | ~17 µs  | ~0.15 ms  |
  | 500 | ~778 µs  | ~28 µs  | ~0.81 ms  |

  So even the 500-report worst case (which we don't expect in practice) saves under 1 ms per load, and at the realistic 10–50 range it's single-digit microseconds. Against the actual page load — the `getDocsSummaries` call, the `searchService.search` PouchDB query, translation, and Angular rendering — that difference isn't going to be measurable by a
  user. The large multipliers (up to 15.6x) are real but come from tiny absolute numbers.

  This is also why I was hoping for telemetry from a real instance: the microbenchmark simulates the two loops in isolation, so it can't tell us whether the contact page actually loads faster, only that these specific lines are cheaper in a vacuum.

 My ask is that we frame it as what the evidence supports: this is a `refactor`, not a `perf` improvement, since we don't have data showing a user-visible speedup. Could you retitle the PR/commit to
  `refactor(#10960): ...`? If you're able to spin up a seeded instance and capture load telemetry that shows a real delta, I'm happy to revisit the `perf` label.

Issue #10373, opened 2025-10-13
Flaky e2e test: Bulk delete reports - should select, deselect and delete only selected reports


dianabarsan:
**Describe the issue**
This test fails frequently:

https://github.com/medic/cht-core/actions/runs/18335337569/job/52219287656
https://github.com/medic/cht-core/actions/runs/18308597267/job/52133045698

```
------------------------------------------------------------------
[chrome 140.0.7339.207 linux #0-6] Running: chrome (v140.0.7339.207) on linux
[chrome 140.0.7339.207 linux #0-6] Session ID: d8b7716237094d09ab678eb1cf8a6f41
[chrome 140.0.7339.207 linux #0-6]
[chrome 140.0.7339.207 linux #0-6] » tests/e2e/default-mobile/reports/bulk-delete.wdio-spec.js
[chrome 140.0.7339.207 linux #0-6] Bulk delete reports
[chrome 140.0.7339.207 linux #0-6]    ✖ should select, deselect and delete only selected reports (5x retries)
[chrome 140.0.7339.207 linux #0-6]    ✓ should open a selected report and a no selected report
[chrome 140.0.7339.207 linux #0-6]
[chrome 140.0.7339.207 linux #0-6] 1 passing (1m 44.8s)
[chrome 140.0.7339.207 linux #0-6] 1 failing
[chrome 140.0.7339.207 linux #0-6]
[chrome 140.0.7339.207 linux #0-6] 1) Bulk delete reports should select, deselect and delete only selected reports
[chrome 140.0.7339.207 linux #0-6] element (".mobile.multiselect-bar-container .select-all-label") still displayed after 15000ms
[chrome 140.0.7339.207 linux #0-6] Error: element (".mobile.multiselect-bar-container .select-all-label") still displayed after 15000ms
[chrome 140.0.7339.207 linux #0-6]     at async toggleSelection (/home/runner/work/cht-core/cht-core/tests/page-objects/default-mobile/reports/reports.wdio.page.js:24:3)
[chrome 140.0.7339.207 linux #0-6]     at async Object.selectAll (/home/runner/work/cht-core/cht-core/tests/page-objects/default-mobile/reports/reports.wdio.page.js:29:10)
[chrome 140.0.7339.207 linux #0-6]     at async Context.<anonymous> (/home/runner/work/cht-core/cht-core/tests/e2e/default-mobile/reports/bulk-delete.wdio-spec.js:63:29)
```

[...]

PR #10967, opened 2026-05-01
fix(#7627): clean up canvas and animation on destroy

### LINKED ISSUE #7627: Memory leak caused by using the `countdown-widget` and the `geopicker` widget in the same form
**Describe the bug**
When using a form with both the [`countdown-widget`](https://github.com/medic/cht-core/blob/master/webapp/src/js/enketo/widgets/countdown-widget.js) and the [`geopicker`](https://github.com/enketo/enketo-core/blob/master/src/widget/geo/geopicker.js) widget, there is form data (definitly some from the countdown widget, maybe more) that seems to get left on the DOM and not cleaned up when the form is submitted or canceled.  

image_1653074452846_0

**To Reproduce**
1. Create a form similar to this (putting the timer in the repeat is just a bonus to make the memory usage more profound):

type | name | label::en | appearance | repeat_count
-- | -- | -- | -- | --
begin_group | text_widgets | Text widgets | field-list |  
begin repeat | timer_repeat | Timer Repeat |   | 75
note | my_timer | My timer | countdown-timer |  
end repeat |   |   |   |  
end_group |   |   |   |  
begin_group | geopoint_widgets | Geo widgets | field-list |  
geopoint | geopoint_map | Geopoint with map Widget | maps |  
end_group |   |   |   |  

2. From the `Reports` tab repeatedly begin and then cancel out of your new form (or submit it, does not seem to matter)
4. See the memory usage for the app consistently rise.... (The most convenient way that I found to do this was to _Record_ the usage from the _Performance_ tab of DevTools.)

**Expected behavior**
The memory should not continue to grow overtime. Instead the resources used by each new form should be cleaned up after the form is canceled or submitted.

**Environment**
- Version: `3.15.0` and https://github.com/medic/cht-core/tree/6345-enketo-uplift

Abhishek9639:
# Description
When a form uses both the `countdown-widget` and the `geopicker` widget, the countdown widget was leaking memory. The `_init()` method appended a `<canvas>` element to the DOM and started an animation loop, but there was no `destroy()` method to clean either of them up when the form was submitted or cancelled.

This PR fixes the leak by implementing `destroy()` on the `Timerwidget` class to stop the animation loop and remove the canvas from the DOM. The `timer-animation` module already had a guard to stop animating when `canvas.isConnected` is false, which works alongside this fix.

Fixes #7627

# Code review checklist
- [x] Tested: Unit tests added for `destroy()` verifies animation stops and canvas is removed from the DOM. Tests also added for `timer-animation` to verify the cleanup function works correctly.
- [x] Backwards compatible: Only adds a `destroy()` lifecycle method. No existing behaviour changes.
- [x] AI disclosure: This fix was developed with AI assistance.


Abhishek9639:
Hi @jkuester, @dianabarsan or @witash,
I've submitted a fix for this added a `destroy()` method to clean up the canvas and stop the animation loop when the form closes. PR is up for review whenever you get a chance.
Thanks

Abhishek9639:
Hii @witash,
The failing check is `tests/e2e/default/reports/infinite-scrolling.wdio-spec.js` (`Infinite scroll did not load new items`). It's not related to this PR the changes only touch `countdown-widget.js`, `timer-animation.js`, and their unit specs, none of which are involved in the Reports list or infinite scroll.

Same test also flakes on master. Run [25670393220](https://github.com/medic/cht-core/actions/runs/25670393220) hit the exact same error on the first attempt and passed on retry; my run just lost the dice roll and burned all three retries.

Re-running the failed job now. Let me know if you'd like me to open a separate flake issue for it (similar to #10373).
Thanks

witash (approved):
sorry for the late review, this looks good

andrablaj:
@witash can you please review this PR after the latest changes from @Abhishek9639? Thanks!

PR #10971, opened 2026-05-02
feat(#9875): support select_*_from_file external datasets in forms

### LINKED ISSUE #9875: Support selecting from external dataset in forms
Sometimes in a form you want to allow the user to [select from very large set of choices](https://forum.communityhealthtoolkit.org/t/select-choice-question-type-with-a-lot-of-choices/3261/9).  Currently, the only way to do this is by adding all the choices to the `choices` tab (aka the "internal" dataset).  

This approach does not scale well for a number of reasons:
- It can be hard to maintain the list of data, particularly keeping the data in sync across multiple different forms that use the same dataset.
- The `xlsx` files can become large and unwieldy to edit
- The form `xml` file produced by cht-conf is very large and cht-conf's form validation logic will not perform well.

Enketo supports a [`select_*_from_file` question type](https://docs.getodk.org/form-question-types/#select-from-external-dataset) that allows for specifying that the data to select from is coming from an [external dataset](https://docs.getodk.org/form-datasets/#form-datasets).  We should add support to the CHT for this question type.  External data files could be uploaded to the server as attachments on the form doc and then injected into the form model when initializing the Enketo form.

I did spend some time exploring the possibility of using the `extension-libs` to do something here.  Unfortunately, the intricacies/functionality of `select_*` questions are so closely integrated with the Enekto form lifecycle (which requires that the form-model data be completely initialized _when creating the form_ that I could not find any feasible approach from that angle.

**Additional context**

The `select_*_from_file` question type basically causes the select widget to look for an external `instance` in the form model. We are already adding an external `instance` to forms [containing contact-summary data](https://github.com/medic/cht-core/blob/e73d2e3036c34f2936f4b92d387300650262926c/webapp/src/ts/services/enketo.service.ts#L162).  I think it should be fairly straightforward to extend this code to include more xml instances in the `options.external` list that gets passed to the `EnketoForm` constructor.

The MVP here would seem to be including an xml file that exists as an attachment to the form doc.  For completeness, this would also probably include changes in cht-conf and the api server to upload/save the attachment to the doc.

vikrantwiz02:
# **Description**

Adds support for `select_one_from_file` / `select_multiple_from_file` XLSForm question types in CHT forms.

When pyxform converts an XLSForm that uses e.g. `select_one_from_file items.xml`, it produces an XForm model with an external instance:
`<instance id="items" src="jr://file/items.xml"/>`

Enketo's constructor accepts an `options.external` array to inject these datasets. CHT already uses this mechanism for `contact-summary` and `user-contact-summary`, but no other external datasets were supported — so `select_*_from_file` forms rendered without choices.

This PR loads external datasets from CHT's **generic resources system** (`CustomResourceService`) and injects them alongside the contact summaries into `options.external` when initialising the Enketo form.

## **Supported URI scheme**

- `jr://file/items.xml` — the resource is XML and is parsed directly with `DOMParser`.

> Note: only XML external datasets are supported. Earlier iterations of this PR also handled CSV (`jr://file-csv/…`), but that was removed during review to keep the scope focused; if a referenced resource is missing (or is not XML), form loading fails fast with a clear error.

## **Data source**

Files are uploaded via `cht-conf`'s existing `upload-resources` action (list the file in `resources.json` and place it in the `resources/` directory) and stored as attachments on the shared `resources` CouchDB document — the same mechanism used for icons. This lets the same dataset be reused across multiple forms.

## **Implementation details**

- `ExternalInstance` interface extracted from the previously-private `ContactSummaryXml` and exported so `FormService` (and future callers) can type external instances.
- XML decoding/parsing lives in `CustomResourceService`: `getResource()` reads the attachment from its in-memory resource cache (no extra DB calls), and `getXml()` base64-decodes it, parses it to a `Document`, and caches the parsed result (keyed by filename) to avoid re-parsing across renders. The XML cache is cleared when the `resources` doc changes.
- `FormService.getExternalInstances()` traverses the form model for `instance[src][id]` nodes, resolving each via `getFilenameFromSrc` (URI-scheme detection) + `loadExternalInstance` (resource lookup + parse). The resulting instances are passed through the form context into Enketo's `options.external` alongside the contact summaries.

## Closes #9875

# Code review checklist

# License


vikrantwiz02:
Hi @jkuester, PR is ready for review.  

I’ve implemented the support for external datasets following our final plan, and all 2678 tests (including the new coverage for XML and CSV) are passing.  

**Technical Highlights:**
- **Base Branch:** I cherry-picked the `CustomResourceService` refactor to ensure the foundation is aligned with the latest resource service logic.  
- **CSV State-Machine:** I moved away from the regex-based parser in favor of a state-machine implementation. This was necessary to satisfy SonarQube quality gates regarding ReDoS vulnerabilities (S5852) and to reliably handle quoted fields and escaped quotes.  
- **Refactoring:** To keep cognitive complexity low (S3776), I’ve split the logic into smaller, testable helpers like `getFilenameFromSrc` and `loadExternalInstance`.  
- **Memory & Encoding:** The implementation uses a single-pass XML escaper and `TextDecoder` (with `Uint8Array`) to ensure we handle non-ASCII characters without driving up heap pressure on mobile devices.

The feature is fully backwards compatible and strictly follows the ODK spec for `jr://` URI schemes. 
Let me know if any improvements needs.

webapp/tests/karma/ts/services/enketo.service.spec.ts
```diff
@@ -392,6 +393,43 @@ describe('Enketo service', () => {
...
+      return service.renderForm(formContext, doc, userSettings).then(() => {
+        expect(EnketoForm.callCount).to.equal(1);
+        const external = EnketoForm.args[0][1].external;
+        const itemsEntry = external.find(e => e.id === 'items');
+        expect(itemsEntry).to.not.be.undefined;
+        expect(itemsEntry!.xml).to.equal(externalDoc);
+      });
+    });
+
+    it('uses empty array for options.external when no contact summaries or external instances', () => {
```
jkuester: Minor, but I don't think we need this test case at all if you just add this assertion to the "'passes json instance data through to Enketo'" test above:

```
        expect(EnketoForm.args[0][1].external).to.deep.equal([]);
```

webapp/tests/karma/ts/services/enketo.service.spec.ts
```diff
@@ -392,6 +393,43 @@ describe('Enketo service', () => {
...
+        html: $('<div>my form</div>'),
+        model: VISIT_MODEL_WITH_EXTERNAL_DATASET,
+      };
+      const userSettings = { language: 'en' };
+      return service.renderForm(formContext, doc, userSettings).then(() => {
+        expect(EnketoForm.callCount).to.equal(1);
+        const external = EnketoForm.args[0][1].external;
+        const itemsEntry = external.find(e => e.id === 'items');
+        expect(itemsEntry).to.not.be.undefined;
+        expect(itemsEntry!.xml).to.equal(externalDoc);
```
jkuester:
```suggestion
        expect(external[0].id).to.equal('contact-summary');
        expect(external[1]).to.deep.equal(externalInstances[0]);
```

webapp/src/ts/services/form.service.ts
```diff
@@ -180,6 +184,126 @@ export class FormService {
...
+  }
+
+  private getExternalInstances(model: string): ExternalInstance[] {
+    const results: ExternalInstance[] = [];
+    $(model).find('instance[src]').each((_, el) => {
+      const src = $(el).attr('src') ?? '';
+      const id = $(el).attr('id') ?? '';
+      if (!id) {
+        return;
+      }
```
jkuester: Minor, but I think you can simplify this to just look for values in the query:

```suggestion
    $(model)
      .find('instance[src][id]:not([src=""]):not([id=""])')
      .each((_, el) => {
        const src = $(el).attr('src')!;
        const id = $(el).attr('id')!;
```

webapp/src/ts/services/form.service.ts
```diff
@@ -180,6 +184,126 @@ export class FormService {
...
+        i += r.advance;
+      } else {
+        field = this.handleNonQuoteChar(ch, fields, field, inQuotes);
+      }
+    }
+    fields.push(field.trim());
+    return fields;
+  }
+
+  private csvToXml(csv: string): Document {
```
jkuester: Okay, I spent a long time thinking about this and weighing some possibilities.  I see that the [ODK docs](https://docs.getodk.org/form-question-types/#select-from-external-dataset) note XML, CSV, and GeoJSON are supported file types.  However, digging into the enekto-core code, it is clear that it wants all the `external` datasets to just be XML (which is why you are doing this conversion here).  

I like the idea of supporting as much as possible, but honestly here on the client-side just feels like the wrong place to be doing this conversion... 😓  It is some non-trivial extra logic we have to support and we still have to worry about various edge-cases we might have missed.  Then, we still need to have `externalInstanceCache` around to cache the converted data so that we are not wasting time doing this conversion again in the future.

I think the best thing for us to do now is to _only support XML files_ here in the webapp.  That will simplify the code we have to maintain here and keep us consistent with what Enketo supports.  Ultimately, I think the CSV data format is going to be more convenient for partners trying to maintain data records. Looking ahead, though, there is nothing stopping us from enhancing github.com/medic/cht-conf in the future to support doing this CSV > XML conversion (e.g. before uploading the XML file with the resources).  IMHO that would be the most appropriate place to do the conversion anyway.

webapp/src/ts/services/form.service.ts
```diff
@@ -180,6 +184,126 @@ export class FormService {
...
+      xmlStr += '<item>' + cells + '</item>';
+    }
+    xmlStr += '</root>';
+    return new DOMParser().parseFromString(xmlStr, 'text/xml');
+  }
+
+  private decodeResourceData(data: string): string {
+    const bytes = Uint8Array.from(atob(data), c => c.codePointAt(0)!);
+    return new TextDecoder().decode(bytes);
+  }
```
jkuester: Can we move this logic to the CustomResourceSerivce (along with the `new DOMParser().parseFromString(content, 'text/xml');` below) and just add a new public method: `getXml(name: string): Document`?

That will maximize the re-usability.

webapp/src/ts/services/form.service.ts
```diff
@@ -180,6 +184,126 @@ export class FormService {
...
+    if (!parsed) {
+      return undefined;
+    }
+    const { filename, isCsv } = parsed;
+    if (this.externalInstanceCache.has(filename)) {
+      return { id, xml: this.externalInstanceCache.get(filename)! };
+    }
+    const resource = this.customResourceService.getResource(filename);
+    if (!resource) {
+      console.error(`External dataset "${filename}" not found in resources`);
```
jkuester: If we cannot find a resource file or it cannot be parsed as an XML document, lets just throw an error like: `throw new Error(`External dataset "${id}" not found in resources. Only XML files are supported.`);`.

We don't want to continue with loading the form in this case, so we can just throw an error to the framework and call it good.

jkuester (changes requested):
@vikrantwiz02 This is great!  I am unreasonably excited about this functionality!  😅 This could really impact the way complex form workflows are built in the future!  👀 I have left a number of comments inline, but the big one is dropping support for CSV (see the comment for more details).  

Besides what I noted in other comments we also need:

### Update `cht-form` to support injecting external instances

We need to update the [cht-form web component](https://github.com/medic/cht-core/blob/8c0cefbd0bafefffecc2a4cb9f8ca29f703db491/webapp/web-components/cht-form/src/app.component.ts#L83) to support injecting external instance data.  We already have code in that component for injecting the contact-summary and setting it on the EnketoFormContext.  We just need to add parallel code for the `externalInstances` (so you can set a `ExternalInstance[]` on the component).  This cht-form component is a key piece of https://github.com/medic/cht-conf-test-harness and so we need this integration to support future automated testing of forms with these select_*_from_file questions.

### Add e2e tests

I think we can add a [new e2e test file](https://github.com/medic/cht-core/tree/8c0cefbd0bafefffecc2a4cb9f8ca29f703db491/tests/e2e/default/enketo) with a new test for a form with a select_*_from_file question. We don't need to worry about covering a bunch of edge cases, but just one happy-path is important.  

Can use this as the test form:

[select_from_file.xlsx](https://github.com/user-attachments/files/27539947/select_from_file.xlsx)
[select_from_file.xml](https://github.com/user-attachments/files/27539946/select_from_file.xml)
[colors.xml](https://github.com/user-attachments/files/27539949/colors.xml)

vikrantwiz02:
@jkuester Updates are pushed.

My initial thought behind adding the CSV parser was to give partners maximum flexibility to upload raw datasets without needing to run them through the build tools first. However, I completely agree with your architectural tradeoff. Forcing the mobile client to parse and cache CSVs introduces unnecessary heap pressure, so shifting that responsibility upstream to `cht-conf` makes much more sense. I've stripped the CSV logic and kept this strictly to XML.

I've resolved the inline comments above. Below are the changes:

* **Resource parsing shifted upstream:** Moved the decode/parse logic into a new `getXml()` method in `CustomResourceService` along with its own cache lifecycle.
* **Web Component integration:** Added `externalInstances` as an `@Input` on the `cht-form` component.
* **Error handling:** `loadExternalInstance` now strictly throws if a requested resource is missing.
* **E2E Test added:** The GitHub links for the test fixtures were throwing 404s, so I authored the `select_from_file.xml` XForm and `colors.xml` dataset directly. The new WebdriverIO test upserts the dataset to the test resources document, renders the form, verifies the choices load, and asserts the submission.

Let me know if it goes well.

sugat009:
Heads up @vikrantwiz02: the red CI here isn't flaky. The Compile job is failing at the lint step on two real errors in `webapp/src/ts/app.component.ts`:

- line 129: `ResourceIconsService` is used but not imported (`no-undef`)
- line 15: `CustomResourceService` is imported but unused (`no-unused-vars`)

Looks like residue from the recent master merge (a service rename that wasn't fully reconciled), so a re-run won't clear it; it needs the import fixed in `app.component.ts`.

andrablaj:
@vikrantwiz02 can you please solve the conflicts with the main branch? Thank you!

vikrantwiz02:
@andrablaj, I resolved merge conflicts and all tests in local, it's good to go. Now waiting for CI Checks, if any fails, I will resolve.

vikrantwiz02:
Hi @andrablaj, it's ready for re-review. Thanks!

nKataraia:
We built this branch (`8adaf618`) and tested it against a real deployment dataset from Jamii ni Afya (Zanzibar Community Health Program): **387 health facilities with a `choice_filter` on island**, and `appearance="autocomplete"`. Reporting back in case it helps the review. Context on our use case is in #9875.

**Method:** rendered the form via the `cht-form` web component, passing the dataset through the new `externalInstances` input. This exercises the Enketo/itemset path; it does **not** exercise the resources-doc → `getXml` lookup. Chromium 150.

**It works.**

| Check | Result |
|---|---|
| Dataset loaded | 387 items, no parse error |
| Island = Unguja | 277 options |
| Island = Pemba | 110 options |
| Re-filter on island change | 12 ms |
| Autocomplete search | narrows correctly at 387 items |

The counts match our dataset exactly, so the itemset predicate is applied to the external instance and re-evaluates correctly when its dependency changes.

**Coverage gap:** `colors.xml` has 3 items and an unfiltered itemset (`instance('colors')/root/item`). A predicate over an external dataset seems likely to be the common real-world case, and isn't covered by the current tests. Happy to contribute our filtered fixture if that would be useful.

**Two possible issues, both around unsupported file types:**

1. `getFilenameFromSrc` only accepts `jr://file/`. A `jr://file-csv/…` src — what pyxform emits for a `.csv` external dataset — returns `undefined`, so the instance is silently dropped, with no error and no log. The `"Only XML files are supported"` message can only ever fire for files that *are* XML.

2. `getXml` returns `null` only when the resource is absent, so the "or it cannot be parsed as an XML document" case from the review discussion above isn't actually detected. `DOMParser.parseFromString(content, 'text/xml')` neither throws nor returns null on non-XML input — it returns a document containing `<parsererror>` (verified in Chromium 150 for both CSV and JSON input). A non-XML file attached under an `.xml` name would therefore be handed to Enketo as a valid dataset, and the select would render empty rather than erroring.

Neither blocks our use case. Thanks for the work on this — it's the feature we need.

jkuester (approved):
Okay, sorry for the delay here. 😓   @vikrantwiz02 the changes look great!  I have pushed a few minor updates and some integration tests for cht-form.  LGTM!  

@nKataraia thanks a bunch for the feedback here! This is super helpful.  With regard to `#1`, I think it would be better to block CSV files in cht-conf before the form config is even uploaded to the server.  I have logged https://github.com/medic/cht-conf/issues/833.  

For `#2`, I have updated the implementation here to throw the same error for the `<parsererror>` scenario as when the xml resource is just not found.  👍

Issue #9725, opened 2025-01-03
"Unexpected Error" Dialog should not be optional


kennsippell:
**What feature do you want to improve?**
![Image](https://github.com/user-attachments/assets/4e46936a-c0d3-416b-a1d3-362e8da84407)

This dialog box fires after a [Global Database Failure](https://forum.communityhealthtoolkit.org/t/mitigating-database-has-a-global-failure-unexpected-error-dialog/2643). 

When investigating user complaints for eCHIS Kenya Digital Payment, 9 different support requests were all due to the CHP having a global database failure, ignoring this dialog, and continuing to use the app. This manifests in a host of different ways:

1. "Cannot save report"
3. "Error loading form"
4. Error loading contacts, tasks (eg #9704), or reports tab

It is a bit difficult to instruct them to reload the app through their CHA. Often we tell them to restart their phone since it is an easy single step.

**Describe the improvement you'd like**
This dialog shouldn't be optional. Their session is over and unrecoverable. It should be a modal dialog with the only path forward being to reload.

**Describe alternatives you've considered**

**Additional context**

dianabarsan:
Do we have a notion of how often does this happens? This is unrelated to your suggestion of changing the options that the app is offering for the user.

kennsippell:
Are you asking about the rate of global database failures for eCHIS? Or asking how often the user dismisses the dialog?

Here is the number of feedback documents indicating that the `IDBDatabase` global database failure has occurred on some randomly sampled single days:

Day | Number of Feedback Docs 
-- | --
Nov 15, 2024 | 498
Dec 1, 2024 | 591
Jan 1, 2025 | 2292
Jan 2, 2025 | 2136
Jan 3, 2025 | 1324

`curl -X POST "https://nairobi.echis.go.ke/medic-users-meta/_find" -H 'Content-Type: application/json' -d '{ "limit": 100000, "fields": ["_id"], "selector": { "type": { "$eq": "feedback" }, "meta.time": { "$gte": "2025-01-02T", "$lt": "2025-01-03T" }, "log": { "$elemMatch": { "arguments": { "$regex": ".*IDBDatabase.*" } } } } }'`

Hendrixx-RE:
Can you assign me this issue 
ive made a pr which hides the cancel and x button
I think it should work sufficiently

andrablaj:
Closed via https://github.com/medic/cht-core/pull/10972.

PR #10972, opened 2026-05-02
feat(#9725): force application reload on global database failure

Hendrixx-RE:
# Description
Changed the UI so that the only option on database failure is reload.

- DatabaseClosedComponent: Added `[hideCancelButton]="true"` to the template, leaving "Reload" as the only dismissal option.
- AppComponent: Configured the database failure modal with `closeOnNavigation: false` so it persists across route changes. 

#9725 

# Code review checklist

- [] Documented: Configuration and user documentation on [cht-docs](https://github.com/medic/cht-docs/)

# License


dianabarsan:
Hi @Hendrixx-RE 
Thank you for the submission. 
Can you please include a screenshot or short video of the UI change? Thanks!

Hendrixx-RE:
@dianabarsan 
https://github.com/user-attachments/assets/90db5b2e-e227-44f1-ae1d-ecc15d21dce2
The change is working on my end.
 I just ran a script to trigger the popup.
Do tell me if there's any modifications to do

dianabarsan (changes requested):
The changes to `ModalLayoutComponent` and the explicit `[hideCloseButton]="true"` on this modal don't change any rendered output. The panel header's close button is already hardcoded hidden on master (`modal-layout.component.html:2`: `[hideCloseButton]="true"`), and the new `@Input()` you added defaults to `true` when undefined — so we end up with
  the same behavior we already had.                                                                                                                                                                                                                                                                                                                             
   
  Could you revert the three files that touch `ModalLayoutComponent` (`modal-layout.component.html`, `modal-layout.component.ts`, and the `[hideCloseButton]="true"` line in `database-closed.component.html`)? The substantive UI change here is `[hideCancelButton]="true"`, which is enough on its own.                                                      
                  
  The description says this hides "both the 'Cancel' button and the header's close button," but the header close button is already hidden for all modals via the hardcoded `[hideCloseButton]="true"` in `modal-layout.component.html`. Only the Cancel button is newly hidden by this PR. Could you update the description so the changelog reflects what      
  actually changed?                                                                                                                                                                                                                                                                                                                                             
   
  The test verifies that `closeOnNavigation: false` is forwarded to `modalService.show`, but doesn't exercise the user-visible behavior the issue is about — that the dialog can't be dismissed. Consider adding a test that renders `DatabaseClosedComponent` and asserts the Cancel button is not in the DOM, and (ideally) one that fires a router navigation
   and confirms the modal is still open. That way a future regression that re-adds the Cancel button or flips `closeOnNavigation` back on will fail loudly.

Hendrixx-RE:
@dianabarsan Did the changes to the best of my knowledge,kindly review.

dianabarsan (changes requested):
Thanks for adding the Cancel button test and the standalone spec — the DOM assertion is exactly what I had in mind.

  Two follow-ups:

  1. **PR description still needs an update.** It still says "Updated the template to hide both the 'Cancel' button and the header's close button," but the ModalLayoutComponent / `[hideCloseButton]` changes were reverted (correctly — the header close button is already hidden on master via the hardcoded `[hideCloseButton]="true"` in
  `modal-layout.component.html`). Could you trim that bullet so the changelog reflects only what actually shipped — `[hideCancelButton]="true"` on the modal, and `closeOnNavigation: false` on the modal open call?

  2. **The `should not close when router navigates` test doesn't exercise `closeOnNavigation`.** `closeOnNavigation` is honored by `MatDialog`'s internal navigation subscription, not by `DatabaseClosedComponent` itself — the component just injects `MatDialogRef` and never listens to the router. Because this spec renders the component standalone via
  `TestBed.createComponent` (not via `MatDialog.open(...)`), `matDialogRef.close()` is never going to fire on navigation regardless of the flag. The test passes for the wrong reason and would still pass if `closeOnNavigation` were flipped back to `true`.

     The flag forwarding is already covered by the `app.component.spec.ts` assertion on `modalService.show.args[0][1]`, so you could either drop this test or convert it into something that opens the modal through `MatDialog` and exercises a real `Router.navigate(...)` while asserting the overlay container still contains the modal. If converting is
  awkward, dropping it is fine — the spec arg check is enough.

Hendrixx-RE:
@dianabarsan  done

dianabarsan (approved):

Issue #9207, opened 2024-06-24
API endpoint for getting generic entity data


jkuester:
**Is your feature request related to a problem? Please describe.**
Based on [this discussion](https://github.com/medic/cht-core/issues/9065#issuecomment-2186704076) we should add a REST endpoint that will return return data from a generic DB document. 

While this may be undesirable from an encapsulation perspective, it will be incredibly useful as a stop-gap solution for a number of workflows.

**Describe the solution you'd like**
I am thinking an endpoint like `api/v1/entity` that supports GET with an id. The main open question I have right now is if we need to consider handling different DBs besides just `medic`?  

**Describe alternatives you've considered**
One school of thought is that this would prove to be a loophole that could be used to get document data that we actually do not want to expose via other apis.  The counter to this is that it is unreasonable to expect that we can make it to 100% API coverage via our REST apis (since that will be something of an always moving target anyway). This functionality will allow the eventual locking down of Couch access even without perfect coverage of all use cases.

Akshanshkaushal:
@jkuester I would like to work upon this issue. Could you please assign this?

jkuester:
@Akshanshkaushal  I see you have several other issues already assigned. Feel free to circle back to this one once the others are complete.

Akshanshkaushal:
@jkuester Definitely! I am working on those.

Hendrixx-RE:
@jkuester can you assign this issue to me,Ill start working on it right now!

jkuester:
@Hendrixx-RE sorry for the run-around here. 😬 😓 

A lot has changed in the 2 years since I logged this issue and I no longer believe it is a good idea for us to support this functionality. We have continued to expand the coverage of our REST apis and have recognized that the api server serves as a crucial abstraction layer above the database. The ability to continue evolving our core database architecture requires that we maintain (and expand) this abstraction and not leak database storage details into the api (within reason).  

A "generic" entity endpoint is super sketchy because it indicates that the _consuming code_, itself, is exposed to the database details.  My current position is that we should regard this as unacceptable going forwards and require that code to be refactored.  With all of this in mind, I am going to close this issue. If anyone disagrees with my assessment here we can always reopen it and pick this back up.

Hendrixx-RE:
[...]

PR #10975, opened 2026-05-03
feat(#9207): add API endpoint for getting generic entity data

Hendrixx-RE:
# Description

A generic entity retrieval system that allows fetching any document from the medic database via a REST API

I followed the project's architectural pattern by adding a new Entity module to the shared library. This allows the same logic to be used by the API (direct DB access) and potentially the Webapp in the future.

   * Endpoint: Created api/src/controllers/entity.js which implements GET /api/v1/entity/:id.
   * Routing: Registered the new controller in api/src/routing.js so the server recognizes the /api/v1/entity/ path.
   * Unit Tests: Added tests for the new Entity module in cht-datasource to verify that the local and remote logic correctly handles document retrieval and missing IDs.
   
Addresses #9207 

# Code review checklist

# License

### DIFF
--- a/api/src/controllers/entity.js
+++ b/api/src/controllers/entity.js
@@ -1,0 +1,62 @@
+const { Entity, Qualifier } = require('@medic/cht-datasource');
+const ctx = require('../services/data-context');
+const serverUtils = require('../server-utils');
+const auth = require('../auth');
+
+const getEntity = ctx.bind(Entity.v1.get);
+
+/**
+ * @openapi
+ * tags:
+ *   - name: Entity
+ *     description: Operations for generic database documents
+ */
+module.exports = {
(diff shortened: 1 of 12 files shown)


jkuester:
Sorry @Hendrixx-RE.  :disappointed:  Like I mentioned [in the issue thread](https://github.com/medic/cht-core/issues/9207#issuecomment-4373751933), I no longer think this is a good idea for our api server.  

If other maintainers strongly feel the opposite, we can reopen this and continue, but otherwise I do not think we should proceed with implementing this functionality.

Hendrixx-RE:
Thanks for the response,its my fault to work before checking in with maintainers.Ill keep that in mind before attempting a patch from now on!

PR #10979, opened 2026-05-03
chore(#10883): replace 'any' with HttpHeaders in webapp services

heisenico:
## Summary
Partial fix for #10883. Replaces `headers: any` with `new HttpHeaders({...})` in two service files, following the pattern established in `session.service.ts`.

Files updated:
- `webapp/src/ts/services/update-password.service.ts`
- `webapp/src/ts/services/user-login.service.ts`

The remaining files (`cache.service.ts` and `extract-lineage.service.ts`) are not in this PR. Happy to follow up in a separate PR — let me know if you'd prefer them bundled instead.

## Test plan
- [x] `npm run compile` (webapp) passes
- [x] `npm run lint` passes
- [x] `npm run unit` (webapp Karma) — 2664 tests passing
- [x] Existing specs for both services still pass without changes (the `headers.get('...')` API works identically for `HttpHeaders` and object literals)

### DIFF
--- a/webapp/src/ts/services/update-password.service.ts
+++ b/webapp/src/ts/services/update-password.service.ts
@@ -1,5 +1,5 @@
 import { Injectable } from '@angular/core';
-import { HttpClient } from '@angular/common/http';
+import { HttpClient, HttpHeaders } from '@angular/common/http';
 
 @Injectable({
   providedIn: 'root'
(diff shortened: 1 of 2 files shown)


heisenico:
Closing in favor of #10884, which already covers this issue. Thanks @sugat009 and @andrablaj for the heads-up!

PR #10980, opened 2026-05-03
fix(#8653): re-trigger db-object-widget when field becomes relevant

### LINKED ISSUE #8653: db-object-widget not re-triggered when field becomes relevant
**Describe the bug**
A `select-contact` field can be pre-populated by setting its `calculation` value to the Id of a contact to load.  However, when the field is non-relevant, no data is loaded for the contact.  Then, if the field does become relevant later, the db-object-widget code does not get triggered and so no data gets loaded. 

**To Reproduce**
Use a `select-contact` field in a group to load data for a pre-populated contact id.  If the group is relevant when the form loads, then the data for the contact is loaded as expected.  However, if the group is non-relevant when the form loads and then later becomes relevant, no data will be loaded for the contact (and the contact's name will not be shown in the question widget).  

**Expected behavior**
When a  `select-contact` field becomes relevant, it should evaluate the db-object-widget logic when there is a pre-populated id value.  

**Environment**
Happening on `master` (`4.4.1`+) and on `3.9.0`.

**Additional context**

Note that a non-relevant `select-contact` field that becomes relevant behaves just fine when the value is _selected by the user_. It is just the case where the value is pre-populated that results in the logic not getting triggered.

This issue seems related to https://github.com/medic/cht-core/issues/8225, but is not exactly the same since it is not specific to `inputs`.

Originally reported [on the forum](https://forum.communityhealthtoolkit.org/t/a-select-contact-question-displays-blank-when-the-containing-group-is-conditionally-shown-in-an-app-form/3112).

Abhishek9639:
# Description

When a `select-contact` field has its value pre-populated via a `calculation` and the field is initially non-relevant, the `db-object-widget` never loads the contact data. This happens because the `change` event that triggers `changeHandler()` is never fired when the field transitions from non-relevant to relevant it only fires when the user manually selects a contact.

This fix overrides the `enable()` method on the widget, which Enketo calls exactly when a field transitions from non-relevant to relevant. Inside `enable()`, if the hidden text input already has a pre-populated contact ID and the contact doc hasn't been loaded yet, we re-initialize Select2 with that value as `initialValue`. This triggers `getDoc()` → `setDoc()` → `change` event → `changeHandler()` → `updateFields()`, which correctly loads the contact data into the form.

Fix #8653

# Code review checklist

- [x] Realistic: Confirmed, well named, follows the [style guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide)
- [x] AI disclosure: Please disclose use of AI per [the guidelines](https://docs.communityhealthtoolkit.org/contribute/contributing-guidelines)

# License

### DIFF
--- a/webapp/src/js/enketo/widgets/db-object-widget.js
+++ b/webapp/src/js/enketo/widgets/db-object-widget.js
@@ -17,7 +17,30 @@ class Dbobjectwidget extends Widget {
   }
 
   _init() {
-    construct(this.element);
+    const refs = construct(this.element);
+    this._$textInput = refs.$textInput;
+    this._$selectInput = refs.$selectInput;
+    this._contactTypes = refs.contactTypes;
+    this._allowNew = refs.allowNew;
+    this._filterByParent = refs.filterByParent;
+  }
+
+  // When the field becomes relevant with a pre-populated value, load the contact data.
+  enable() {
+    const currentValue = this._$textInput && this._$textInput.val();
+    if (!currentValue) {
+      return;
+    }
+    const selected = this._$selectInput.select2('data');
+    const alreadyLoaded = selected && selected[0] && selected[0].doc;
+    if (!alreadyLoaded) {
+      const Select2Search = window.CHTCore.Select2Search;
+      Select2Search.init(this._$selectInput, this._contactTypes, {
+        allowNew: this._allowNew,
+        filterByParent: this._filterByParent,
+        initialValue: currentValue,
+      });
+    }
   }
 
   list() {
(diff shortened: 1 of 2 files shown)


Abhishek9639:
Hi @jkuester, 
I confirmed that enable() is the right hook. It's called in relevant.js exactly when a field transitions from non-relevant to relevant. The fix overrides enable() in the widget, checks if there's a pre-populated contact ID, and if the doc isn't loaded yet, re-calls Select2Search.init() with that value as initialValue to trigger the existing contact loading flow. Also added a guard so Select2 isn't re-initialized unnecessarily if the contact is already loaded. Unit tests added for all cases. Happy to make changes if anything looks off.
Thanks

Abhishek9639:
Hi @jkuester,
Can you please review it?
Thanks

PR #10982, opened 2026-05-04
fix: handle Android back button for dropdowns via Angular state

### LINKED ISSUE #10974: Fix dropdown closing on Android back button (remove jQuery, use ngx-bootstrap state)
## Description
AndroidApiService handles the Android back button by closing UI elements like modals and dropdowns before navigating back.

Right now, dropdowns are being closed using jQuery by removing the .open class:

```
const $dropdown = $container.find('.filter.dropdown.open:visible');
if ($dropdown.length) {
  $dropdown.removeClass('open');
  return true;
}
```

But dropdowns have already been migrated to Angular (ngx-bootstrap), so this approach is outdated.

## Problem
This causes a mismatch between the UI and Angular state:

- The dropdown may disappear visually, but Angular still thinks it’s open
- Sometimes the dropdown isn’t detected at all
- This can break further interactions (clicks, toggles, etc.)

## Steps to Reproduce
- Open the app
- Open any dropdown (header/filter)
- Press the Android back button
- Notice inconsistent or broken behavior

## Proposed Fix
Instead of manipulating DOM with jQuery, we should use an Angular-based approach:

- Either track open dropdowns via a small service and call .hide() on them
- Or dispatch a global event that dropdown components listen to and close themselves
 
## Expected Result
- Dropdowns close cleanly on back press
- No state mismatch with Angular directives
- Removes reliance on jQuery for this logic

Please assign this issue to me. I am also open to suggestions.

Aamir377300:
This PR fix #10974 and replaces the legacy jQuery-based dropdown closing logic in AndroidApiService with a modern Angular-native approach.

### Changes:
- Created `DropdownService` to track open `ngx-bootstrap` dropdowns.
- Created `DropdownTrackingDirective` (`mmDropdownTracking`) to automatically register/unregister dropdowns with the service.
- Refactored `AndroidApiService` to use `DropdownService.closeAll()` instead of jQuery DOM manipulation.
- Migrated `AnalyticsFilterComponent` to use `ngx-bootstrap` dropdowns.
- Applied the tracking directive to dropdowns in Header, Analytics, Date, and Sort filters.
- Updated `AndroidApiService` unit tests.

### DIFF
--- a/webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.html
+++ b/webapp/src/ts/components/filters/analytics-filter/analytics-filter.component.html
@@ -1,6 +1,6 @@
 <div class="analytics-sidebar-filter">
-  <span *ngIf="analyticsModules?.length > 1" class="dropdown mm-dropdown analytics-module">
-    <span id="module" class="mm-button" data-toggle="dropdown">
+  <span *ngIf="analyticsModules?.length > 1" class="dropdown mm-dropdown analytics-module" dropdown mmDropdownTracking>
+    <span id="module" class="mm-button" dropdownToggle>
       <span class="mm-button-icon">
         <span class="fa fa-bar-chart-o"></span>
       </span>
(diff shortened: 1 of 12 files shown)


andrablaj:
Hi @Aamir377300 , thanks for opening the issue and the PR. 

Issue #10761, opened 2026-03-25
Add task count labels to contacts list


dianabarsan:
**Is your feature request related to a problem? Please describe.**
Sometimes users use the contacts tab, and the contacts list to navigate through their daily workflow. Currently contacts are sorted alphabetically and only the name is displayed. 
With UHC enabled, the number of household visits within the current month is also displayed. 

**Describe the solution you'd like**
Similar to UHC, where we add count of household visits to the contact RHS list, a similar feature where we display the number of tasks (or due tasks) has come up as being helpful. 
Like UHC, when this is enable, the user should be able to also sort the contacts based on the number of tasks. 

Mentioning that this should only be applied to households (leaf type places) where we already display a list of household tasks in the detail page. 

**Additional context**
Came up as a community suggestion through @JenniferMuli .

sh1vam31:
Hi @dianabarsan, @andrablaj 

I would love to tackle this feature! I plan to look into how the UHC household visits are currently calculated and displayed in the RHS list, and replicate that pattern to extract and display the active task count for leaf-type places (Households). 

Could I please be assigned to this?

subalkum:
hey @andrablaj i would love to contribute

andrablaj:
@sh1vam31 are you still planning to work on this issue?

sh1vam31:
> [@sh1vam31](https://github.com/sh1vam31) are you still planning to work on this issue?

Hi @andrablaj,
I’m currently stuck with another issue and won’t be able to continue working on this for now. 
Please assign it to someone else. 
Thanks!

CodeBySayak:
hi @dianabarsan, @andrablaj i want to work on this feature, I explored how the UHC household visits are currently calculated and displayed in the RHS list, and replicate that pattern to extract and display the active task count for leaf-type places. Can i work on this?

rshandilya-02:
Hi @dianabarsan  @andrablaj  , want to work on this task , can you assign it to me ?

andrablaj:
@subalkum are you still planning to work on this issue? If not, please let us know so we can assign it to someone else. Thanks!

Keerti707:
Hi @dianabarsan @andrablaj  I would like to work on this issue!!
[...]

PR #10983, opened 2026-05-04
fix(#11097): apply fallback sorting when contacts_default_sort is unset

### LINKED ISSUE #11097: Contacts list does not apply fallback sorting when contacts_default_sort is unset
## Describe the bug

When `contacts_default_sort` is not configured, the contacts list does not consistently apply a fallback/default sorting behavior.

The contacts component currently depends on the configured value from `getContactsDefaultSort(settings)`, but when the setting is unset or missing, the list can fall back inconsistently instead of explicitly defaulting to the standard contacts ordering behavior.

## Steps to reproduce

1. Remove or unset `contacts_default_sort` from settings
2. Open the contacts list
3. Observe the contacts ordering behavior

## Expected behavior

The contacts list should apply a stable default sorting behavior even when `contacts_default_sort` is not configured.

## Proposed solution

Apply an explicit fallback/default sort value when `contacts_default_sort` is undefined so the contacts list behavior remains consistent.

Keerti707:
This PR fixes the issue where contacts were not sorted when `contacts_default_sort` was not defined in settings.

Previously, sorting was only applied when `getContactsDefaultSort(settings)` returned a truthy value, which caused the sorting logic to be skipped when the setting was missing,and so :

This change applies a fallback 'default' in the component when no value is provided which like ensures that the contacts are always sorted consistently.

No changes were made to `UHCSettingsService` to preserve its existing behavior as defined by current tests.

Closes #11097

# Code review checklist

- [x] Readable
- [x] Tested
- [x] Backwards compatible
- [x] AI disclosure

AI assistance was used for guidance, debugging, and understanding the codebase, but all changes were reviewed and verified manually!!


Keerti707:
Hey!!
Are there any changes that need to be done in this PR..I am confirming since it has been in review for some time now!
I would be glad to make the required changes and needed better amends so please guide me through it if needed!

andrablaj:
@sugat009 would you be available to review this PR?

sugat009:
@Keerti707, it looks like there might be a mix-up with the linked issue. Ticket #10761 focuses on task count labels, but this PR addresses contact sorting. Could you double-check if this was the intended issue?

Keerti707:
Thanks for catching this @sugat009.

You are right, I mixed up the linked issue here. This PR currently addresses the contacts default sorting fallback, but #10761 is about adding task count labels to the contacts list.

I’ll update this by either removing the incorrect `Closes #10761` reference and linking the correct issue if one exists, or I’ll close this PR and open a fresh PR specifically for #10761 with the task count label implementation.

Sorry for the confusion, and thank you for reviewing!

Keerti707:
This PR is now scoped only to the contacts default sorting fallback behavior and is no longer linked to #10761.

andrablaj:
@Keerti707 can you please update the PR title and assign it to the correct issue? Thanks!

Keerti707:
Thanks @andrablaj , I’ll update the PR title and align the implementation with #10761 by adding task count labels to the contacts list and sorting support for task counts.

andrablaj:
@Keerti707 the code in this PR needs to match the issue it solves, which is not https://github.com/medic/cht-core/issues/10761. Please link this code to the specific issue it solves. Thanks!

Keerti707:
Thanks @andrablaj  for clarifying. I created #11097 for the default contacts sorting fallback issue and will link this PR to that issue instead of #10761.

sugat009:
Hey @Keerti707 , leaving some notes and a question on this change.

On my read, the bug described in the issue doesn't really exist on master. `contacts.component.ts` declares `defaultSortDirection = 'alpha'` and `sortDirection = this.defaultSortDirection` as field initializers (lines 79, 80), so when `contacts_default_sort` is unset, `sortDirection` is already `'alpha'` after init. Downstream, `isSortedByLastVisited()` is the only semantic check on the value (`=== 'last_visited_date'`), so anything else falls through to alphabetical sort. Master already applies that fallback.

What this PR changes is `sortDirection = 'default'` (instead of `'alpha'`) when `lastVisitedDateExtras = true` and the setting is unset. Two issues:

1. `sort-filter.component.html` highlights options only for `'alpha'` and `'last_visited_date'`. There's no branch for `'default'`, so the "Sort by name" option loses its selected highlight on a fresh load when no admin config is set.
2. `'default'` isn't a documented value for `uhc.contacts_default_sort` (docs list `'alpha'` and `'last_visited_date'` only), and it isn't wired up anywhere else.

The compile CI is also failing on this exact behavior. Two existing tests in `contacts.component.spec.ts` (lines 616 and 656) both assert `sortDirection === 'alpha'` and now see `'default'`. Those failures are catching the regression, not flakes.

Questions:
1. Could you share a short screen recording of the misbehavior you observed?
2. Did you run karma locally with this change before pushing? The CI failures will need to be addressed before merge.

Happy to be wrong if I'm missing context.

Keerti707:
Thanks for the detailed explanation @sugat009 , it makes sense to me now

I completely missed that `sortDirection` already falls back to `alpha` through the existing initialization logic, and that the only special handling is for `last_visited_date`. I was treating the unset case as a missing fallback when it was actually already working correctly on master.

You're also right about `default` not being a valid/documented value and the regression it causes in the sort filter state. I checked the failing tests after your comment and they’re correctly catching that behavior.

I don’t have a proper reproduction or screen recording for an actual bug here, so I’ll revert the change and close this out rather than keep an invalid fix open.

Really appreciate the clear breakdown here, it helped me understand the flow much better.

PR #10984, opened 2026-05-04
chore(#10749): revert merging 10749

witash:
reverting merging [PR 10776](https://github.com/medic/cht-core/pull/10776) from  [issue 10749](https://github.com/medic/cht-core/issues/10749)  to avoid releasing view changes individually.
there will be a new PR to revert this revert when its ready


witash:
Hi @mrjones-plip this PR is a workaround to move 10749 out of master; the simplest way to do that I think is make a PR with a revert commit, then make a new PR reverting the revert when its done.

But it does need someone to click the approve button, can you help with that?

mrjones-plip (approved):
🚢

PR #10986, opened 2026-05-04
fix(#10985): add missing special character escaping in nouveau query builder

### LINKED ISSUE #10985: Nouveau query escaping misses documented special characters
## Description

The `escapeKeys` function in `shared-libs/nouveau/src/index.js` does not escape all characters listed in its own documentation comment.

The comment on line 1-3 references the [CouchDB Nouveau query syntax](https://docs.couchdb.org/en/stable/ddocs/nouveau.html#query-syntax) and lists these characters as requiring escaping:

```
+ - && || ! ( ) { } [ ] ^ " ~ * ? : \ /
```

The current implementation escapes: `+ - & | ! ^ " ~ * ? : [ ]`

**Missing:** `(`, `)`, `{`, `}`, `\`, `/`

## Impact

The `escapeKeys` function is used to build `key:(...)` Nouveau queries in:
- `api/src/services/authorization.js:631` — replication key collection
- `sentinel/src/lib/purging.js:274` — purging record lookup

If a replication key or subject ID ever contains any of the unescaped characters (particularly `)` which would break out of the `key:(...)` grouping), the Nouveau query would be malformed. In practice, subject IDs are UUIDs and CouchDB usernames which don't contain these characters, so this is a latent correctness bug rather than an active breakage.

## Steps to reproduce

```js
const { escapeKeys } = require('@medic/nouveau');
console.log(escapeKeys('key)value'));  // outputs: key)value  (should be: key\)value)
console.log(escapeKeys('key/value'));  // outputs: key/value  (should be: key\/value)
```

YASHSHARMAOFFICIALLY:
Fixes #10985

## Summary

- The `escapeKeys` function in `shared-libs/nouveau/src/index.js` was missing six characters (`(`, `)`, `{`, `}`, `\`, `/`) from the escaping set, despite the inline comment documenting them as required by CouchDB Nouveau query syntax.
- Added the missing characters and extended the unit test to cover all 19 documented special characters.
- All 23 tests pass with 100% coverage.

## Test plan

- [x] `npm test` in `shared-libs/nouveau` passes (23 tests, 100% coverage)
- [ ] Existing integration and e2e tests for authorization and purging pass in CI


witash (approved):
Looks good!
tested locally with () and it works
I can't think of a realistic scenario where anyone would need to search by these characters, but its good to have them fixed for completeness and correctness.

PR #10989, opened 2026-05-04
fix(#10987): pass container ID to togglePassword on login page

### LINKED ISSUE #10987: Login password visibility icon does not update after toggling
## Description

Clicking the password visibility (eye) icon on the login page toggles the input type between `password` and `text`, but the icon itself does not update to reflect the current state.

https://github.com/user-attachments/assets/5f677ff4-c5d8-44e9-a658-54a00db70d0c

## Root cause

`api/src/public/login/script.js:264` calls `togglePassword` with only one argument:

```js
togglePassword(PASSWORD_INPUT_ID)
```

But `api/src/public/login/auth-utils.js:113` expects two arguments:

```js
const togglePassword = (passwordInputId, passwordContainerId) => {
    ...
    document.getElementById(passwordContainerId)?.classList.toggle('hidden-password');
};
```

Since `passwordContainerId` is `undefined`, the `hidden-password` CSS class on `#password-container` is never toggled, so the icon stays in its initial state.

The password reset page does this correctly (`password-reset.js:138`):
```js
togglePassword('password', 'password-container');
```

## Steps to reproduce

1. Navigate to the login page
2. Enter any text in the password field
3. Click the eye icon to show the password
4. Observe: the password text is revealed, but the icon still shows the "show password" state instead of switching to "hide password"

## Expected behavior

The icon should toggle between show/hide states when clicked, matching the password reset page behavior.

YASHSHARMAOFFICIALLY:
Fixes #10987

## Summary

- The login page called `togglePassword(PASSWORD_INPUT_ID)` with only one argument, but `auth-utils.js` expects two: `togglePassword(passwordInputId, passwordContainerId)`.
- Without the container ID, the `hidden-password` CSS class on `#password-container` was never toggled, so the show/hide icon stayed in its initial state even though the input type changed.
- Added `PASSWORD_CONTAINER_ID` constant and passed it as the second argument, matching how `password-reset.js` already does it.

## Test plan

- [ ] Navigate to the login page, enter text in the password field, click the eye icon — verify the icon toggles between show/hide states
- [ ] Verify the password reset page still toggles correctly


mrjones-plip (approved):
Thanks for the fix, confirmed working:

https://github.com/user-attachments/assets/46a92a8a-8d80-4436-a934-6ac8e3ab3444

PR #10990, opened 2026-05-04
fix(#10988): use JSON.stringify for manifest generation instead of lodash template

### LINKED ISSUE #10988: Generated web manifest is invalid when branding title contains JSON-special characters
## Description

The generated `manifest.json` becomes invalid JSON when the branding title contains double quotes, backslashes, or other JSON-special characters.

## Root cause

`api/src/templates/manifest.json:3` interpolates the branding name as raw text:

```json
"name": "{{ branding.name }}"
```

`api/src/services/manifest.js:14-17` renders this via Lodash `_.template` with the raw `interpolate` setting, so no escaping is applied. A branding name like `CHT "Community"` produces:

```json
"name": "CHT "Community""
```

which is invalid JSON. The same issue applies to `branding.icon` on line 9:

```json
"src": "/img/{{ branding.icon }}"
```

## Steps to reproduce

1. Set branding title to a string containing a double quote (e.g. `CHT "Community"`)
2. Restart the API server (or trigger manifest regeneration)
3. Fetch `/manifest.json`
4. Observe: the response is malformed JSON

## Impact

Broken `manifest.json` prevents PWA installability and may cause browser warnings. While branding names with quotes are uncommon, backslashes in icon filenames are more plausible.

## Suggested fix

JSON-escape the interpolated values before inserting them into the template, or build the manifest as a JavaScript object and use `JSON.stringify` to generate the output.

YASHSHARMAOFFICIALLY:
Fixes #10988

## Summary

- The `manifest.json` was generated using a lodash `_.template` with raw `{{ }}` interpolation. Branding names containing `"`, `\`, or other JSON-special characters produced invalid JSON, breaking PWA installability.
- Replaced the template approach with `JSON.stringify`, which handles all escaping correctly. This also removes the dependency on `lodash` and the `manifest.json` template file from the service.
- Added a test case that verifies a branding name with quotes and backslashes produces valid, parseable JSON.

## Test plan

- [x] All 3 manifest unit tests pass
- [ ] Set a branding name containing quotes (e.g. `CHT "Community"`) and verify `/manifest.json` returns valid JSON
- [ ] Verify PWA install prompt still works with a normal branding name

### DIFF
--- a/api/src/services/manifest.js
+++ b/api/src/services/manifest.js
@@ -2,19 +2,33 @@ const { promisify } = require('util');
 const fs = require('fs');
 const path = require('path');
 
-const _ = require('lodash');
-
 const resources = require('../resources');
 const brandingService = require('./branding');
 
 const webappPath = resources.webappPath;
 const MANIFEST_OUTPUT_PATH = path.join(webappPath, 'manifest.json');
-const TEMPLATE_PATH = path.join(__dirname, '..', 'templates', 'manifest.json');
 
 const writeJson = async (branding) => {
-  const file = await promisify(fs.readFile)(TEMPLATE_PATH, { encoding: 'utf-8' });
-  const template = _.template(file);
-  const json = template({ branding });
+  const manifest = {
+    start_url: './',
+    name: branding.name,
+    display: 'standalone',
+    background_color: '#323232',
+    theme_color: '#323232',
+    icons: [
+      {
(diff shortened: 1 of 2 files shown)


andrablaj:
Hi @YASHSHARMAOFFICIALLY , thanks for opening the issue and the PR. 

PR #11000, opened 2026-05-05
fix(#10999): only normalize slashes in URL pathname, not query params

### LINKED ISSUE #10999: fix: global URL slash normalization corrupts query parameter values
## Description

The global middleware in `api/src/routing.js` normalizes double slashes in request URLs:

```js
req.url = req.url.replace(/\/+/g, '/');
```

This operates on the entire `req.url` string, which includes the query string. Any query parameter containing a URL or double slash gets corrupted.

## Steps to Reproduce

1. Send a request with a URL-like value in a query parameter:
   ```
   GET /some/path?redirect=https://example.com/a//b
   ```
2. Observe the middleware rewrites it to:
   ```
   GET /some/path?redirect=https:/example.com/a/b
   ```

## Expected Behavior

Only the pathname portion should be normalized. Query parameters should be preserved as-is:
```
/some/path?redirect=https://example.com/a//b
```

## Actual Behavior

The entire URL string including query params is modified:
```
/some/path?redirect=https:/example.com/a/b
```

## Root Cause

`req.url` contains both the pathname and query string (e.g. `/path?key=value`). The regex `\/+/g` matches all consecutive slashes, including those inside query parameter values.

## Suggested Fix

Split the URL at the query boundary and only normalize the pathname:

```js
app.use((req, res, next) => {
  const [pathname, ...rest] = req.url.split('?');
  req.url = pathname.replace(/\/+/g, '/') + (rest.length ? '?' + rest.join('?') : '');
  next();
});
```

## Impact

Any route that receives URL-like values in query parameters (redirects, callback URLs, external references) will have those values silently corrupted before reaching the handler.

YASHSHARMAOFFICIALLY:
## Description

Fixes #10999

The global middleware in `api/src/routing.js` normalizes double slashes by replacing all consecutive slashes in `req.url`:

```js
req.url = req.url.replace(/\/+/g, '/');
```

Since `req.url` includes the query string, this corrupts any query parameter containing URL-like values or double slashes.

**Example:**
- Input: `/some/path?redirect=https://example.com/a//b`
- Before fix: `/some/path?redirect=https:/example.com/a/b`
- After fix: `/some/path?redirect=https://example.com/a//b`

## Code Change

```js
// Before
req.url = req.url.replace(/\/+/g, '/');

// After
const [pathname, ...rest] = req.url.split('?');
req.url = pathname.replace(/\/+/g, '/') + (rest.length ? '?' + rest.join('?') : '');
```

The fix splits the URL at the query boundary, normalizes only the pathname, and preserves the query string as-is. It also handles edge cases like multiple `?` characters in the URL.

## Test plan

- [x] Added 6 test cases covering pathname normalization, query preservation, and edge cases
- [x] All existing routing tests continue to pass
- [x] ESLint passes

### DIFF
--- a/api/src/routing.js
+++ b/api/src/routing.js
@@ -208,7 +208,8 @@ app.use(
 app.use(rateLimiterMiddleware);
 app.use(jsonQueryParser);
 app.use((req, res, next) => {
-  req.url = req.url.replace(/\/+/g, '/');
+  const [pathname, ...rest] = req.url.split('?');
+  req.url = pathname.replace(/\/+/g, '/') + (rest.length ? '?' + rest.join('?') : '');
   next();
 });
 
--- a/api/tests/mocha/routing.spec.js
+++ b/api/tests/mocha/routing.spec.js
@@ -1,6 +1,11 @@
 const rewire = require('rewire');
 const chai = require('chai');
 
+const normalizeUrl = (url) => {
+  const [pathname, ...rest] = url.split('?');
+  return pathname.replace(/\/+/g, '/') + (rest.length ? '?' + rest.join('?') : '');
+};
+
 describe('Routing', () => {
   before(() => global.angular = {
     module: () => ({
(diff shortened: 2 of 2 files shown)


mrjones-plip:
The main place we do redirects is in logins, as seen in your [other ticket](https://github.com/medic/cht-core/issues/11027#issuecomment-4400033148).  These are URL encoded like so:

`https://192-168-68-26.local-ip.medicmobile.org:10467/#/messages` -> `https://192-168-68-26.local-ip.medicmobile.org:10467/medic/login?redirect=https%3A%2F%2F192-168-68-26.local-ip.medicmobile.org%3A10467%2F%23%2Fmessages`

After testing myself by introducing an extra slash, I note it does not break this and is handled correctly:

1. as an logged out user go to `https://192-168-68-26.local-ip.medicmobile.org:10467//#//messages`
2. get redirected to `https://192-168-68-26.local-ip.medicmobile.org:10467/medic/login?redirect=https%3A%2F%2F192-168-68-26.local-ip.medicmobile.org%3A10467%2F%23%2Fmessages&username=test`
3. log in
4. correctly land on `https://192-168-68-26.local-ip.medicmobile.org:10467/#/messages`

@YASHSHARMAOFFICIALLY:

1. when submitting tickets, please do not include abstract URLs like `/some/path?redirect=https://example.com/a//b`.  
2. Please include real URLs for your local CHT instance you used to when filing the ticket
3. please use one of our [issue templates](https://github.com/medic/cht-core/tree/master/.github/ISSUE_TEMPLATE) which you see when you [create a new issue](https://github.com/medic/cht-core/issues/new/choose).

PR #11002, opened 2026-05-05
fix(#11001): use route params instead of matrix params for deceased contact navigation

### LINKED ISSUE #11001: Deceased contact deletion navigates using matrix params instead of route params
## Description

After a deceased contact is deleted, the app navigates to the parent contact using Angular matrix parameters instead of route parameters, resulting in an incorrect URL.

## Steps to Reproduce

1. Open a contact marked as deceased
2. Delete the contact
3. Observe the URL after deletion

## Expected Behavior

Navigation should go to:
```
/contacts/<parent-id>
```

## Actual Behavior

Navigation goes to:
```
/contacts;id=<parent-id>
```

This happens because `contacts-deceased.component.ts` line 72 uses:
```ts
this.router.navigate(['/contacts', { id: parentId || null }]);
```

In Angular, passing an object as a route segment creates matrix parameters (`/contacts;id=value`), not a path segment. The contacts route definition expects a path parameter:
```ts
{ path: ':id', ... }
```

## Suggested Fix

```ts
this.router.navigate(parentId ? ['/contacts', parentId] : ['/contacts']);
```

This navigates to `/contacts/<parent-id>` when a parent exists, or `/contacts` if there is no parent.

## Impact

After deleting a deceased contact, users may land on the wrong contacts view instead of seeing the parent contact. This affects the contact management workflow.

YASHSHARMAOFFICIALLY:
## Description

Fixes #11001

After a deceased contact is deleted, `contacts-deceased.component.ts` navigates using Angular matrix parameters instead of route parameters:

```ts
// Before — creates /contacts;id=parent-id (matrix params)
this.router.navigate(['/contacts', { id: parentId || null }]);

// After — creates /contacts/parent-id (route params)
this.router.navigate(parentId ? ['/contacts', parentId] : ['/contacts']);
```

The contacts route expects a path parameter (`:id`), not matrix params. This mismatch causes the user to land on the wrong view after deletion instead of seeing the parent contact.

## How to reproduce

1. Navigate to a contact that has a deceased child contact
2. Open the deceased child contact
3. Delete the contact
4. Observe the URL — it shows `/contacts;id=<parent-id>` instead of `/contacts/<parent-id>`

## Test plan

- [x] Verified the contacts route definition expects `:id` as a path segment
- [x] Confirmed other components in the same module use `['/contacts', id]` for navigation (e.g. `contacts-content.component.ts`, `contacts-report.component.ts`)
- [x] When `parentId` exists, navigates to `/contacts/<parentId>`
- [x] When no parent, navigates to `/contacts`
- [x] ESLint passes

### DIFF
--- a/webapp/src/ts/modules/contacts/contacts-deceased.component.ts
+++ b/webapp/src/ts/modules/contacts/contacts-deceased.component.ts
@@ -69,7 +69,7 @@ export class ContactsDeceasedComponent implements OnInit, OnDestroy {
       callback: (change) => {
         if (change.deleted) {
           const parentId = this.selectedContact?.doc?.parent?._id;
-          return this.router.navigate(['/contacts', { id: parentId || null }]);
+          return this.router.navigate(parentId ? ['/contacts', parentId] : ['/contacts']);
         }
         // refresh the updated contact
         this.selectContact(change.id, true);


andrablaj:
Hi @YASHSHARMAOFFICIALLY , thanks for opening the issue and the PR. 

PR #11004, opened 2026-05-05
fix(#11003): include boundary values in age break comparison

### LINKED ISSUE #11003: fix: exact 1-year and 1-month ages display as lower units due to off-by-one comparison
## Description

`getDateDiff()` in `webapp/src/ts/services/format-date.service.ts` uses a strict greater-than comparison against `ageBreak.min`:

```ts
if (Math.abs(diff) > ageBreak.min) {
```

Since both the years and months age breaks have `min: 1`, exact boundary values (exactly 1 year, exactly 1 month) fail the check and fall through to the next smaller unit.

## Steps to Reproduce

1. Create a patient born exactly 1 year ago (e.g. today is 2026-05-05, DOB is 2025-05-05)
2. View the patient's age on their contact page

## Expected Behavior

Age displays as **1 year**.

## Actual Behavior

Age displays as **12 months** because `Math.abs(-1) > 1` is `false`, so the years break is skipped and months is used instead.

Same issue for exactly 1 month — displays as ~30 days.

## Root Cause

The ageBreaks config:
```ts
ageBreaks: [
  { unit: 'years', key: { singular: 'y', plural: 'yy' }, min: 1 },
  { unit: 'months', key: { singular: 'M', plural: 'MM' }, min: 1 },
  { unit: 'days', key: { singular: 'd', plural: 'dd' }, min: 0 },
]
```

With `min: 1` and `>` (strict), a diff of exactly 1 does not match. It needs `>=` to include the boundary.

## Suggested Fix

Change the comparison to include the boundary:

```ts
if (Math.abs(diff) >= ageBreak.min) {
```

The days break (`min: 0`) is unaffected since `>= 0` still matches all non-negative values.

## Impact

Patient ages and date differences are core clinical information displayed throughout the app. A 1-year-old showing as "12 months" is confusing for health workers.

YASHSHARMAOFFICIALLY:
## Description

Fixes #11003

`getDateDiff()` in `format-date.service.ts` uses strict greater-than (`>`) when comparing the date difference against `ageBreak.min`. Since both years and months age breaks have `min: 1`, exact boundary values (exactly 1 year, exactly 1 month) fail the check and fall through to the next smaller unit.

**Example:** A patient born exactly 1 year ago displays as "12 months" instead of "1 year".

## Code Change

```ts
// Before
if (Math.abs(diff) > ageBreak.min) {

// After
if (Math.abs(diff) >= ageBreak.min) {
```

The days break (`min: 0`) is unaffected since `>= 0` matches the same values as `> 0` for this context (the fallback at line 110 already handles the 0-day case).

## Test plan

- [x] Added test: exactly 1 year old displays with singular year key (`y`)
- [x] Added test: exactly 1 month old displays with singular month key (`M`)
- [x] Existing tests for 5 years, 16 months, 50 days, 1 day, 0 days, and age-at-death all continue to pass
- [x] ESLint passes

### DIFF
--- a/webapp/src/ts/services/format-date.service.ts
+++ b/webapp/src/ts/services/format-date.service.ts
@@ -103,7 +103,7 @@ export class FormatDateService {
     for (let i = 0; i < this.config.ageBreaks.length; i++) {
       const ageBreak = this.config.ageBreaks[i];
       const diff = date.diff(end, ageBreak.unit);
-      if (Math.abs(diff) > ageBreak.min) {
+      if (Math.abs(diff) >= ageBreak.min) {
         return { quantity: diff, key: ageBreak.key };
       }
     }
--- a/webapp/tests/karma/ts/services/format-date.service.spec.ts
+++ b/webapp/tests/karma/ts/services/format-date.service.spec.ts
@@ -148,6 +148,22 @@ describe('FormatDate service', () => {
       expect(relativeTime.args[0]).to.deep.equal([0, true, 'dd', false]);
     });
 
+    it('shows singular year when exactly 1 year old', () => {
+      relativeTime.returns('1 year');
+      const dob = moment().subtract(1, 'years');
+      const actual = service.age(dob);
+      expect(actual).to.equal('1 year');
+      expect(relativeTime.args[0]).to.deep.equal([1, true, 'y', false]);
+    });
+
+    it('shows singular month when exactly 1 month old', () => {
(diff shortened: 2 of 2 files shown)


mrjones-plip:
Triaging: While technically correct, and I was able to reproduce, when referring to the age a child up until ~18 months, using months is expected and normal.  When I tested at "24 months" the CHT correctly shows "2 years"

PR #11012, opened 2026-05-06
feat(#10139): add ability to hide contact forms for muted places

### LINKED ISSUE #10139: Add ability to hide contact forms for muted places
There is no way to block creating contacts under muted places. 

Add a way to hide the add person / add place button for muted places. 
This can be based done through a user permission.

**Additional context**
Forum post: https://forum.communityhealthtoolkit.org/t/hide-contact-forms-on-muted-places/5093

sugat009:
# Description

Adds a new `can_create_contacts_under_muted_places` permission. The contact creation actions are hidden when a place is muted, or when any of its ancestors is muted. In the same case, the create itself is refused. A user whose role holds the permission is exempt from both. The webapp gates both fast action buttons (FABs) and the add route that opens the form. Five API endpoints carry the same gate.

Resolves #10139

**This is a behaviour change for existing deployments.** See Configuration below.

## Implementation notes

* **Webapp, the FABs**: `FastActionButtonService` skips the contact creation actions when the parent is muted and the caller does not hold the permission. The service applies this rule to both the contact detail (right side) FAB and the contacts list (left side) FAB. The service uses the existing `ContactMutedService.getMuted`, so it honours the same parent and ancestor semantics as the rest of the webapp. The contacts list FAB's parent is the user's home place. `LineageModelGeneratorService` hydrates that home place, so an inherited muted state is visible there too. The contacts list's changes subscription passes a change on any of the home place's ancestors through to its change handler. That hydrated copy is then refreshed in place. An online user's changes carry no `doc`. That condition therefore matches on the change id.
* **Webapp, the add route**: `validateParentForCreateForm` calls `ContactsEditComponent.ensureParentNotMuted`, which refuses to open the form for a muted parent. Hiding the contact creation action is not enough on its own. `/#/contacts/:parent_id/add/:type` is reachable directly from the address bar, and from a task's `contact` action. The refusal shows the existing "not authorized" copy rather than the generic form error. The refusal also logs at `info` rather than `error`. `FeedbackService`'s `FEEDBACK_LEVEL` is `error`, so `info` files no feedback doc for what is a routine policy outcome. The `info` log still prints in production, and it joins the circular log buffer that gets attached to any later real error. `debug` would have done neither. The same catch also cancels the component's geolocation handle. No form will open. The `watchPosition` started for it would otherwise stay live until the geolocation service's own 30 second timeout settles it. The catch is not a `finally`, because `save()` passes that same handle to `formService.saveContact` on the success path.
* **API**: a new `api/src/services/muted-parent.js` resolves the parent ref. The service checks the permission first (db admins short circuit). Only for a caller without the permission does the service fetch the hydrated parent via `@medic/lineage`. It then evaluates `parent.muted || isMutedInLineage(parent)`. The service has six call sites. Two are `controllers/person.js:160` and `controllers/place.js:159`, inside `#create`. The other four are in `api/src/routing.js`, across three handlers. The `POST /places` call site checks `req.body.parent` (`:596`). The `POST /people` call site checks `req.body.place || req.body.parent` (`:747`), since `createPerson` passes a raw `parent` through when `place` is falsy. The two `POST /places/:id` call sites check `req.body.parent` (`:665`) and `req.body.contact?.place || req.body.contact?.parent` (`:666`). That is because `updatePlace` hands an inline `contact` to `getOrCreatePerson`, and `createPerson` then writes it under whichever of those two refs is present.
* **Shared lib**: `@medic/transitions` now exposes the existing `muting_utils.isMutedInLineage` on the factory's return object, next to `isWithinTimeFrame`. The api consumes it through `config.getTransitionsLib()`, like `records.js` and `messaging.js` do. `muting_utils.js` is unchanged from master. The shared lib's source is net two lines in `index.js`, plus the new unit spec listed under Tests.

## Scope

Offline role users can reach the legacy `POST /api/v1/places` and `/api/v1/people` routes. The `auth.check` on those routes has no online requirement. `POST /api/v1/person` and `/place` require an online role. So for an offline role API caller, the legacy routes carry the only server side gate.

Despite the name, the permission is not only about creates. `POST /api/v1/places/:id` accepts a `parent`. So `can_create_contacts_under_muted_places` is also what permits moving an existing place under a muted place on that deprecated route. Only the moved place's own document is rewritten. Its existing descendants keep their stored lineage. The move does not re-gate the descendants themselves.

Moving an existing place under a muted place produces the same end state the ticket asks to prevent. The route is therefore gated. Leaving that route open would be a hole in the same guarantee.

The shipped default for `can_create_contacts_under_muted_places` is `[]` in all four config files. After upgrade, no configured role holds the permission, and only db admins bypass the gate. Consider an external client that reads a place, changes its name and posts the whole body back with `parent` included. It starts getting a 403 whenever that parent is muted. Nothing in this repo calls the route in product code. So the exposure is custom integrations only.

| Path | Gated? | Notes |
|---|---|---|
| Contact detail FAB, contacts list FAB | Yes | The contacts list FAB gates on the user's home place. `LineageModelGeneratorService` hydrates that home place. If that hydration fails, the contacts list FAB fails open. Only the add route then refuses. The hydrated home place is refreshed when the home place doc changes. For an online user, it is also refreshed when one of its ancestors changes. An inherited mute or unmute is therefore picked up without leaving the Contacts tab. A document above the user's own facility does not normally replicate. So for an offline user, an inherited mute arrives as the `muted` flag. The muting transition stamps that flag onto the home place itself. |
| `/#/contacts/:parent_id/add/:type`, typed directly or reached from a task's `contact` action | Yes | `ContactsEditComponent` refuses the add route. So hiding the contact creation action is not the only defence. |
| `POST /api/v1/person`, `POST /api/v1/place` | Yes | The parent must be a UUID string. The validator rejects objects. |
| `POST /api/v1/places/:id` re-parenting under a muted place | Partial | The gate covers any request carrying a `parent` that resolves to a document id. This includes a request that resends the place's existing parent. A body that omits `parent` (or sends it empty) is exempt. The gate also covers the inline `contact`'s `place` and raw `parent`, because `updatePlace` creates that person. Not covered: an inline `parent` (or `contact.place`) object with no `_id`. `getOrCreatePlace` creates that object under its own muted parent. Then the target is re-parented under it. |
| `POST /api/v1/places` (`parent`), `POST /api/v1/people` (`place`, or `parent` when `place` is absent or empty) | Partial | The gate covers a UUID string. It also covers an object whose `_id` is a string naming an existing doc. Not covered: an inline parent object with no `_id` or a non string `_id` (a place created in the same request). Also not covered: an object with a client chosen `_id` that does not exist yet. For that object, the lookup 404s and the gate treats the parent as not muted. Both shapes are documented above `isParentMuted`. What each route then writes is documented there too. |
| Any client's direct write to `_bulk_docs`, and replication | No | The webapp saves contacts with `dbService.get().bulkDocs(...)` for online and offline users alike. So no in-app save reaches a gated route. The add route gate is what actually stops the form opening. Offline users' `_bulk_docs` requests reach an authorization-only filter with no muted check. Online users' requests are proxied straight to CouchDB. This is the main remaining gap. It is the only gap this PR calls out as needing a future fix. See below. |
| Editing a place with a form configured to create child docs | No | An edit form whose XML adds siblings or repeats can create contacts under a muted place. This path is reachable only with such a form configured. None of the four default `*-edit.xml` forms does this. `can_edit` alone guards `:id/edit`. |
| A report form's `db-doc` elements creating contacts | No | This change does not touch report forms. A `db-doc="true"` node can emit a contact doc at any parent. Two default forms do exactly that: `delivery.xml`'s `baby_profile` and `replace_user.xml`'s `new_contact`. So on a stock install, a delivery report against a muted household still registers the baby under it. This is a pre-existing report-form capability with no relationship to #10139. |
| `POST /api/v1\|v2\|v3/users` creating inline places and people | No | This is a pre-existing admin flow. It is out of scope. |
| `POST /api/v1\|v2/records` where a configured registration transition's `add_patient` trigger creates a patient | No | The trigger's `parent_id` param names a report field. So the request body picks the parent. `api/src/controllers/records.js` runs the transition inside the request. This path is reachable only with such a registration configured. The default config sets `transitions.registration` to false and ships no registrations. |

### How wide the effect is

Two mechanisms compound, so it is worth stating the reach plainly. Muting cascades downward: the muting transition stamps `muted` on the target and on every descendant. `updateMuteState` queries `medic/contacts_by_depth` by the target's id, and the view emits that id for every descendant. `updateMuteState` then bulk-writes each one. This gate then walks upward. Both `parent.muted || isMutedInLineage(parent)` on the API and `ContactMutedService.getMuted` in the webapp treat a place with a muted ancestor as muted. So muting one district removes contact creation for every place and every user beneath it, not only for that district.

Gating the contacts list FAB extends that one step further. The contacts list FAB's parent is the user's own home place. So a user whose home place is muted, or sits under a muted place, has no contact creation action on the contacts list view either. The contact detail FAB and the add route are also gated. So that user can create no contacts anywhere beneath that place until the mute is lifted or the permission is granted. A top-level place, created from `/#/contacts/add/:type`, has no parent to gate and is unaffected.

The forum thread framed the request at household level. This reach is the honest reading of the ticket, but it is broader than that framing. So this section calls it out rather than leaving it to be discovered.

### The offline write path

An offline user's contact saves go to local PouchDB. They reach the server through `_bulk_docs`. `_bulk_docs` never touches the gated routes. `validate_doc_update.js` does not mention muting either. There is no muted check anywhere in `api/src/services/replication/`.

The realistic way to hit this is a permission change racing a sync:

1. A CHW's role holds `can_create_contacts_under_muted_places`. The contact creation action is therefore shown.
2. The permission is revoked server side.
3. The CHW has not synced yet. Their client still evaluates the old settings. The client still shows the action.
4. They create a contact under a muted place. It saves locally. It replicates up successfully.

Verified against a dev instance. With the permission revoked, an offline user's `_bulk_docs` write of a person under a muted clinic returns `[{"ok":true,"id":"...","rev":"1-..."}]`. The document lands in CouchDB with its lineage pointing at the muted place.

Where sentinel's muting transition is enabled, that transition mutes a contact that arrives this way. The muting happens afterwards. So the end state is consistent, even though creation was not blocked. Where the muting transition is disabled, the contact simply lands unmuted under a muted place.

This cannot be closed with a `validate_doc_update` rule. CouchDB validation functions receive only `newDoc`, `oldDoc`, `userCtx` and `secObj`, with no database access. The rule therefore cannot read the parent to learn whether it is muted. `minifyLineage` also stores only `_id` and `parent` on the new doc, never `muted`.

Closing it properly means a gate on the common write path (`canEdit`), not just the offline `_bulk_docs` and db-doc handlers. This is because online users' writes bypass those handlers and are proxied straight to CouchDB. That costs a parent lookup on every contact write. A replicated doc rejected there already exists on the device. Deferred rather than attempted here.

## What is not hidden

Only the contact creation actions are affected. On a muted contact the report forms remain available, including the configured unmute form. The phone and send message actions remain available too, as do Edit and Delete. Report forms on a muted contact keep their existing "Muted contact" confirmation modal rather than disappearing.

## Follow-up considerations (out of scope here)

* Whether **update** and **delete** on muted contacts also need analogous permissions. The forum thread and ticket only mention create. Update specifically would need a carve-out for the configured `unmute_forms` so users can still submit the unmute form against a muted contact.
* Closing the offline write path. This needs a gate on the common write path rather than in the offline handlers alone.

## Tests

* **Mocha unit (api)**: `muted-parent.spec.js` covers 15 cases. One case pins the nested payload limitation. One case asserts that the permission check runs before the lineage load. One case rejects a non string `_id`, so it cannot reach PouchDB. `controllers/{person,place}.spec.js` each add one rejection case. This PR also adds two assertions to their existing success cases.
* **Mocha unit (shared lib)**: a new `shared-libs/transitions/test/unit/index.js` pins that the factory exports `isMutedInLineage`. Nothing else covered that export.
* **Karma unit (webapp)**: 22 new cases. `fast-action-button.service.spec.ts` gets 6 cases. They cover the muted branch, its ordering against the base create permission, and a muted home place both directly and through a muted ancestor. `contacts-edit.component.spec.ts` gets 6 cases for the add route refusal, the inherited-ancestor case, the permission holder skipping the lineage read, and the clean-parent case. They also cover a refused form load filing no feedback doc, and the geolocation handle being cancelled when a form fails to load. `contacts.component.spec.ts` gets 8 cases for the hydrated home place being forwarded, no home place, and a hydration failure leaving the contacts list working. They also cover re-hydration on a change to the home place, on a change to an ancestor, and not on an unrelated change. Two cases cover the changes filter itself: a doc-less ancestor change passes it, and it does not throw before the home place has resolved. `contacts-content.component.spec.ts` gets 2 cases for `parentContact` forwarding.
* **HTTP integration (api)**: 19 cases across `muted parent gate` describes in `tests/integration/api/controllers/{person,place,places}.spec.js`. On `POST /api/v1/places`, `places.spec.js` covers a muted parent, a clean parent, and a clean parent whose own parent is muted. The last of those is the only case that exercises `isMutedInLineage` rather than the `parent.muted` short circuit. `places.spec.js` also covers the legacy `POST /api/v1/people` route. The cases there are `place`, raw `parent` without `place`, and `place: ''` with a raw `parent`. One more case pins that `place` wins over `parent` when both are present. `places.spec.js` also covers `POST /api/v1/places/:id` re-parenting. The cases there are a move under a muted parent, and a rename of a place already under one. A third case is a rename that resends the unchanged muted parent. On that same route, `places.spec.js` also covers an inline `contact` naming a muted place by `place` or by raw `parent`. One more case pins that `contact.place` wins over `contact.parent` when both are present.
* **E2E wdio**: a new `tests/e2e/default/contacts/muted-contacts-create.wdio-spec.js` with 4 scenarios. In a muted clinic without the permission, no contact creation action is present. In a non muted clinic, the action is present. The add route deep link is refused. In a muted clinic with the permission granted, the action is present after sync and reload.

## Translations

The permission description is in `messages-{en,es,fr,ne,pt,sw}.properties`. `fr`, `ne` and `pt` carry real translations. `en`, `es` and `sw` carry the English string. `es` and `sw` do so deliberately, because `scripts/ci/lint-translations.js` runs with `checkMissing` and `checkEmpties` over `en`, `es`, `fr`, `ne` and `sw`. If those two lines are missing or empty, CI fails. Forum post requesting the remaining translations: https://forum.communityhealthtoolkit.org/t/translation-support-request-cht-core-10139/5608

## Configuration

This PR registers the new permission key with an empty role list in:

* `config/default/app_settings.json`
* `config/covid-19/app_settings.json`
* `config/covid-19/app_settings/base_settings.json`
* `tests/config.default.json`

**This restricts existing deployments on upgrade. It is not opt in.** On a fresh install the key arrives with the rest of `config/default/app_settings.json`. On an upgrade of an already configured instance, api backfills the key on the first startup after the upgrade.

`api/server.js:78` awaits `configWatcher.load()`. That call runs `loadSettings()` (`api/src/services/config-watcher.js:145`). `loadSettings()` is `settingsService.update({})` (`:73-74`). `api/src/services/settings.js:102-108` then runs `_.defaults(doc.settings.permissions, defaultConfig.permissions)`. The settings service PUTs the settings doc whenever that call adds anything. The key therefore lands in `settings.permissions` as `[]` with no operator action.

Later restarts change nothing. They do not write. The path you arrive by makes no behavioural difference. Absent and empty are identical for authorisation. `shared-libs/cht-datasource/src/auth.js` treats a missing permission key as not granted. The restriction therefore applies immediately either way.

A deployment that wants the previous behaviour must grant the permission to the roles that already hold `can_create_people` or `can_create_places`. It must then push its settings.

One consequence of the backfill is worth knowing. The api performs that write itself. The admin app's Permissions tab therefore shows a row for the new key as soon as api has restarted. The tab needs no settings save first. The admin app builds the tab from `Object.keys(settings.permissions)` (`admin/src/js/controllers/authorization-permissions.js:27`). By then the key is there.

This matches the precedent. A permission of this kind turns a new restriction on by default. The permission itself is the opt out. `can_skip_password_change` and `can_view_old_navigation` both worked this way. `can_view_old_action_bar` and `can_view_old_filter_and_search` did the same, though they were removed in 4.12.0 along with the old UI they guarded.

# Code review checklist

- [ ] Backwards compatible: **no.** This restricts existing deployments on upgrade. See Configuration above. The upgrade note is drafted, and it goes on #10139 rather than in the release notes. So this stays unticked until that lands
- [ ] Documented: needs a cht-docs PR for the permissions table row, a qualifier on the muting workflow page, and an Upgrade notes entry
- [x] Tested: Unit and/or e2e where appropriate (see Tests section above)

# License

# AI Disclosure

Plan to implementation generated by [cht-agent](https://github.com/medic/cht-agent) and some part of the implementation as well. Minor changes to the implementation done by me and tests written using claude code.


sugat009:
### Screenshots

1. CHW without the new permission to create contacts under muted place:
    1. On Muted place
        image
    1. On normal  place
        image
1. CHW with the new permission to create contacts under muted place:
    1. On muted place
        image
    1. On normal place
        image

jkuester:
Got that "Dark Mode" CHT prototype going!  :sweat_smile: :bulb:

sugat009:
> Got that "Dark Mode" CHT prototype going! 😅 💡

Could be a `Good First Issue`?

sugat009:
@dianabarsan could you please review this?
Thanks!

api/src/services/muted-parent.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+  if (!parent) {
+    return false;
+  }
+  return Boolean(findMutedAncestor(parent));
+};
+
+const assertCanCreateOnMutedParent = async (userCtx, parentRef) => {
+  if (await isParentMuted(parentRef)
+    && !auth.hasAllPermissions(userCtx, 'can_create_contacts_under_muted_places')) {
```
dianabarsan: I think that we can optimize here a bit. 
If the permission is granted, then we don't even have to check if the parent is muted. If the permission is not granted, only then the lineage load is required.

api/src/controllers/place.js
```diff
@@ -148,7 +149,11 @@ module.exports = {
      *         $ref: '#/components/responses/Forbidden'
      */
     create: serverUtils.doOrError(async (req, res) => {
-      await auth.assertPermissions(req, { isOnline: true, hasAny: ['can_create_places', 'can_edit'] });
+      const userCtx = await auth.assertPermissions(
+        req,
+        { isOnline: true, hasAny: ['can_create_places', 'can_edit'] }
+      );
+      await mutedParent.assertCanCreateOnMutedParent(userCtx, req.body?.parent);
```
dianabarsan, line 159: I'm wondering whether we can save a db roundtrip here. Do we load the parent in the `create` function as well? If we do, then we would end up calling the lineage function twice. Same comment for person.

api/src/routing.js
```diff
@@ -586,11 +587,13 @@ app.get('/api/v1/users-info', authorization.handleAuthErrors, authorization.getU
...
   auth
     .check(req, ['can_edit', 'can_create_places'])
-    .then(() => {
+    .then(async (userCtx) => {
       if (_.isEmpty(req.body)) {
         return serverUtils.emptyJSONBodyError(req, res);
       }
-      return places.createPlace(req.body).then(body => res.json(body));
+      await mutedParent.assertCanCreateOnMutedParent(userCtx, req.body.parent);
+      const body = await places.createPlace(req.body);
```
dianabarsan, line 597: Same comment as on place and person. Do we hydrate the parent in `createPlace`?

shared-libs/transitions/src/lib/muting_utils.js
```diff
@@ -282,16 +282,7 @@ const getClientMutingEventsToReplay = (contacts, reportId) => {
...
-  let parent = doc && doc.parent;
-  while (parent) {
-    if (parent.muted && (beforeMillis ? new Date(parent.muted).getTime() <= beforeMillis : true)) {
-      return parent._id;
-    }
-    parent = parent.parent;
-  }
-  return false;
-};
+const { isMutedInLineage } = require('./muted-lineage');
```
dianabarsan: Is there a specific reason this was moved to its own file? Can it just be exported from here instead?

tests/e2e/default/contacts/muted-contacts-create.wdio-spec.js
```diff
@@ -0,0 +1,69 @@
...
+
+  before(async () => {
+    await utils.saveDocs([...places.values(), mutedClinic]);
+    await utils.createUsers([offlineUser]);
+    await loginPage.login(offlineUser);
+  });
+
+  afterEach(async () => {
+    // Tests 2 and 3 leave a partially-filled form on screen; refresh to bypass the Exit-form modal.
+    await browser.refresh();
```
dianabarsan: I believe here we can use `common.reloadPage` to not call `waitForPageLoaded`. 
Also, Is it not simpler to just close the modal instead of refreshing and landing on the same "contactCreate" page? `brwoser.refresh` just loads the same page all over again.

api/src/routing.js
```diff
@@ -586,11 +587,13 @@ app.get('/api/v1/users-info', authorization.handleAuthErrors, authorization.getU
 app.postJson('/api/v1/places', function(req, res) {
   auth
     .check(req, ['can_edit', 'can_create_places'])
-    .then(() => {
+    .then(async (userCtx) => {
       if (_.isEmpty(req.body)) {
         return serverUtils.emptyJSONBodyError(req, res);
       }
-      return places.createPlace(req.body).then(body => res.json(body));
+      await mutedParent.assertCanCreateOnMutedParent(userCtx, req.body.parent);
```
dianabarsan, line 596: Claude says: This endpoint accepts nested parent objects (see `shared-libs/contacts/src/places.js` `createPlaces`, which recursively creates `{ name, parent: { ... } }` chains). When `parentRef` is such a fresh nested object without `_id`, `resolveParentId` returns `null` and the gate is silently bypassed — even if the
deepest `parent` in the chain is a muted UUID. Could you either walk the nested chain here (recursively check `parentRef.parent` when `_id` is absent) or document that this gate only covers string/`_id`-bearing parent refs? The PR description sells these endpoints as defense-in-depth, and right now nested payloads slip through.

dianabarsan (commented):
Code looks good! 
I added some questions and one comment. Happy to follow up on those.

api/src/services/muted-parent.js
sugat009: Good call, reordered. The check is now: no parent ref, return; caller holds `can_create_contacts_under_muted_places` (db admins short circuit inside `auth.hasAllPermissions`), return; only then fetch the hydrated parent and walk it. So permitted callers never pay the lineage load, and the unit spec's "not muted" case now asserts the permission check ran once and the fetch ran once.

api/src/controllers/place.js
sugat009: Yes, `create` does load the parent again, but not hydrated on the new endpoints: `Person.v1.create` is a single `db.get` and `Place.v1.create` fetches parent plus primary contact in one `allDocs`, neither with lineage. On the legacy routes a string parent goes through `Place.v1.getWithLineage`, so there the hydrated fetch really is duplicated (an inline parent object with `_id` is not fetched by `createPlaces` at all, so there the gate's is the only load).

I kept the walk hydrated because the check has to see ancestors: a parent's own `muted` flag is only as fresh as the muting transition, so it lags while sentinel processes a batch, is never set when the transition is disabled, and is not back filled for a subtree re-parented under a muted place. The webapp FAB gate already walks the lineage via `setMutedState`, so this keeps the API and the UI in agreement.

Reusing the create path's fetch would mean threading a hydrated parent into cht-datasource's create input (its validator requires `parent` to be a string) or into `@medic/contacts`, which has no `userCtx`. That felt disproportionate for this feature, so instead the gate now skips the fetch entirely for anyone holding the permission, and only callers without it pay the extra round trips. Happy to do the `@medic/contacts` refactor as a follow up if you would prefer.

api/src/routing.js
sugat009: Yes for a string parent: `createPlaces` fetches it through `Place.v1.getWithLineage`, so on this route the hydrated load is duplicated for callers without the permission. An inline parent object with an `_id` is not fetched by `createPlaces` at all, so there the gate's load is the only one. Full reasoning in the `place.js` thread above.

shared-libs/transitions/src/lib/muting_utils.js
sugat009: It is, the export from `muting_utils` was never removed. The separate file existed only so api could import the walk without the import time side effects of `muting_utils`: it builds a `@medic/lineage` instance against `db.medic` and calls `infodoc.initLib` at require time, and transitions' `db.js` replaces its exports on `init()`, so requiring it from api at module load would bind lineage to an uninitialised db, or to the exit on call stub in api's unit tests. At runtime it would only work because `server.js` happens to run the factory before routing loads.

You are right that a new file is more than this needs. I have reverted `muted-lineage.js` and the `muting_utils` change, and instead exposed `isMutedInLineage` on the factory's return object next to `isWithinTimeFrame`, required inside the factory after `db.init`. api consumes it through `config.getTransitionsLib()` the way `records.js` and `messaging.js` do, and the gate now reads `parent.muted || isMutedInLineage(parent)`, the same idiom `schedules.js` uses, so `findMutedAncestor` is gone too. `muting_utils.js` is now byte identical to master, and against master the shared lib is net two lines in `index.js` (`muted-lineage.js` was added and removed inside this branch, so it never existed on master).

tests/e2e/default/contacts/muted-contacts-create.wdio-spec.js
sugat009: Right on both counts. There is no `reloadPage`, but `commonPage.refresh()` is exactly that pair, so that would have been the one liner.

Your second point turned out to be the more important one. Looking into why the refresh was there: the New person form flags itself as edited about 100ms after rendering with nothing typed in, so a refresh lands back on the add page and re-arms the exit form guard. That blocks the `goToPeople` in the permission-granted case, and the form left open by the non-muted-clinic case then satisfies its `New person` assertion, so the permission-granted case could pass without ever reaching the muted clinic.

I confirmed that locally. With only a URL assertion added to the spec as it stood, the permission-granted case fails, and the URL at that point is `/#/contacts/<non-muted clinic id>/add/person`: the form the non-muted-clinic case opened, under the wrong clinic. The muted clinic's id never appears.

I have dropped the `afterEach` and now close the form where it is opened, with `genericForm.cancelForm()` and `modalPage.submit()` at the end of the two cases that actually open one, the non-muted clinic case and the permission-granted case, the same pattern `tasks.wdio-spec.js` uses for an untouched add person form. The permission-granted case also asserts the muted clinic's page actually loaded before clicking the FAB, so it cannot pass vacuously again.

api/src/routing.js
sugat009: Good catch, confirmed. An inline parent object without `_id` resolves to null and skips the check entirely. Only the legacy routes are affected, since the v1 `/person` and `/place` validators reject any non string parent.

Digging into it, there is a second shape as well: a client chosen `_id` that does not exist yet is legal on both legacy routes, so even the recursive walk you suggested would stop at that `_id`, get a 404, and read it as "not muted". The two routes then diverge: a `parent` with an unknown `_id` is minified and stored as a dangling lineage reference (`createPlaces` sees `parent._id` set and skips the recursive create), while a `place` with an unknown `_id` on `/api/v1/people` is actually created with the supplied id by `getOrCreatePlace`.

I have taken your second option and documented the limitation rather than chasing both shapes here. The gate covers a parent that is a UUID string or an object whose `_id` names an existing document, and that is now stated above `isParentMuted` and in the scope table in the description, with a unit test pinning the nested behaviour so a future change has to notice it. I have also dropped the defense in depth wording. Within the contact creation flows the one remaining webapp gap is an offline user's own `_bulk_docs` writes. Outside them, report forms can still write contacts through `db-doc="true"` nodes, and two of the default forms do: `delivery.xml`'s `baby_profile` and `replace_user.xml`'s `new_contact`. The description sets all of that out in the scope table, along with the paths that only exist once configured, an edit form that creates child docs and a registration transition's `add_patient` trigger.

Separately, while checking this I found that the gate on `/api/v1/people` only looked at `req.body.place`. `createPerson` passes a raw `parent` field straight through when no `place` is given, so that shape was ungated too. That one is now `req.body.place || req.body.parent`, mirroring `createPerson`'s own `if (data.place)` branch, with integration tests. `POST /api/v1/places/:id` had the same shape twice over, through its own `parent` and through an inline `contact` that `updatePlace` creates, so both are gated there too.

sugat009:
Thanks for the review, and sorry for the delay. Before anything else: **the description is worth re-reading rather than skimming.** It is not the one you reviewed. I have rewritten it, and the scope table in particular now says what is gated, what is only partly gated, and what is not gated at all.

On my own open question about the sync path: I have kept this to code level checks and dropped the `validate_doc_update` idea, because that rule could not actually do the job. CouchDB validation functions get only `newDoc`, `oldDoc`, `userCtx` and `secObj` with no database access, so the rule cannot read the parent to find out whether it is muted, and the minified parent chain on the new doc never carries `muted`.

The webapp side is now closed on every route I could find, not just the button. `ContactsEditComponent` refuses to open the add form for a muted parent, so the `/#/contacts/:parent_id/add/:type` deep link and a task's `contact` action are both covered, and the contacts list FAB is gated as well as the contact detail one. Hiding one button while another still worked did not seem defensible.

On the API side I also gated `POST /api/v1/places/:id`, since it accepts a `parent` and `can_update_places` is held by the same roles this restricts. Only refs the body carries are checked, the incoming `parent` and the inline `contact`'s `place` or `parent`, never the target's stored parent, so renaming a place that already sits under a muted parent still works; a client that resends an unchanged `parent` is refused like a move, which I accepted rather than paying a read to compare them.

What is left uncovered is an offline user's own writes through `_bulk_docs`. The realistic way to reach it is not a crafted request: a CHW holds the permission, it is revoked server side, and until their next sync the client still shows the action and the write replicates up fine. I verified that against a dev instance. Where the muting transition is enabled the contact is muted afterwards anyway; where it is not, it lands unmuted under a muted place. Covering it means a check on the common write path (`canEdit`), not just the offline `_bulk_docs` and db-doc handlers, since online users' writes skip those via `onlineUserPassThrough` and are proxied straight to CouchDB. That costs a parent lookup on every contact write, and a replicated doc rejected there already exists on the device, so I have left it and written it into the description rather than attempting it here.

Several pre-existing paths are still ungated and listed as out of scope in the description: inline place and person creation through the users endpoints, a report form's `db-doc` nodes, which `delivery.xml` and `replace_user.xml` both ship with, a registration transition's `add_patient` trigger, and an edit form configured to create child docs, though none of the four default `*-edit.xml` forms does that. Happy to file any of them separately.

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -209,10 +213,26 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
...
+    // handle is what save() passes to formService.saveContact.
+    this.geoHandle?.cancel();
+    this.errorTranslationKey = error.translationKey || 'error.loading.form';
+    this.globalActions.setLoadingContent(false);
+    this.contentError = true;
+    if (error.isAuthorizationRefusal) {
+      // a deliberate policy outcome, not an application error, so not console.error, which would
+      // file a feedback doc. info still prints in production and joins the circular log buffer, so
+      // a refused deep link leaves a trace when someone reports that a form will not open.
+      console.info('Contact form refused.', error);
```
dianabarsan, line 234: I'm a little confused about why we treat this case so differently, compared to the other guardrails to open the form. 
We apparently believe that someone constructing a url to create a child under an invalid parent is feedback doc worthy. Because maybe the app creates the wrong link somewhere, and this becomes a valid bug. 
Same for trying to create a child for a muted parent. There should be a guard to hide the button. If the guard becomes bugged, only the feedback doc will log the error. 

I suggest you let this become a feedback doc. Nobody checks app production logs for info entries.

webapp/tests/karma/ts/modules/contacts/contacts-edit.component.spec.ts
```diff
@@ -213,6 +222,70 @@ describe('ContactsEdit component', () => {
...
+      // the doc's own `muted` is absent; the muted state is inherited from the lineage. This is the
+      // whole reason the component reads the lineage rather than the raw doc.
+      lineageModelGeneratorService.contact.resolves({
+        doc: { _id: 'parent_uuid' },
+        lineage: [{ _id: 'grandparent_uuid', muted: '2025-01-01T00:00:00Z' }],
+      });
+      // real logic, so the case fails if the lineage argument is ever dropped
+      const realMutedService = new ContactMutedService();
+      contactMutedService.getMuted.callsFake((doc, lineage) => realMutedService.getMuted(doc, lineage));
+      authService.has.withArgs('can_create_contacts_under_muted_places').resolves(false);
```
dianabarsan, line 261: Can we just assert that we pass the lineage argument instead of bleeding another service's functionality here? If we don't trust this service, we should address it in its own unit tests. Here we should assert that correct data gets passed to it.

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -209,10 +213,26 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
...
+      this.handleFormLoadError(error);
+    }
+  }
+
+  // extracted from initForm's catch to keep that function under the cognitive complexity limit
+  private handleFormLoadError(error) {
+    // no form will open, so stop the watch instead of leaving `watchPosition` live until the
+    // service's own 30 second timeout settles it. Not a `finally`: on the success path this same
+    // handle is what save() passes to formService.saveContact.
+    this.geoHandle?.cancel();
```
dianabarsan, line 226: Thanks for adding this <3

webapp/tests/karma/ts/modules/contacts/contacts-edit.component.spec.ts
```diff
@@ -485,6 +572,33 @@ describe('ContactsEdit component', () => {
         expect(component.contentError).to.equal(true);
       });
 
+      it('should refuse the add form when the parent is muted, without filing a feedback doc', async () => {
```
dianabarsan, line 575: As above, we should totally fill a feedback doc. There should be a guard to block this form from being linked and that block did not work.

webapp/tests/karma/ts/services/fast-action-button.service.spec.ts
```diff
@@ -466,6 +470,103 @@ describe('Fast Action Button service', () => {
...
+        childContactTypes: [
+          { id: 'child-place-1', create_key: 'child-place-1-title', icon: 'child-place-1-icon' },
+          { id: 'child-person-1', create_key: 'child-person-1-title', icon: 'child-person-1-icon', person: true },
+        ],
+        communicationContext: {
+          sendTo: { _id: '1234' },
+          callbackOpenSendMessage: sinon.stub(),
+        },
+      };
+      contactMutedService.getMuted.returns(true);
```
dianabarsan, line 488: I believe that `contactMutedService` will return the muted property, in this case `'2025-01-01T00:00:00Z'`, not a boolean. 

```suggestion
contactMutedService.getMuted.returns(mutedParent.muted);
```

webapp/tests/karma/ts/services/fast-action-button.service.spec.ts
```diff
@@ -466,6 +470,103 @@ describe('Fast Action Button service', () => {
...
```
dianabarsan, line 514:
```suggestion
contactMutedService.getMuted.returns(mutedParent.muted);
```

webapp/tests/karma/ts/services/fast-action-button.service.spec.ts
```diff
@@ -466,6 +470,103 @@ describe('Fast Action Button service', () => {
...
```
dianabarsan, line 560: Does this even get called in this case? I believe there's an early return. Can we assert that the muted service is not checked and strengthen this test case on the scenario it is verifying?

webapp/tests/karma/ts/services/fast-action-button.service.spec.ts
```diff
@@ -559,6 +660,51 @@ describe('Fast Action Button service', () => {
...
+          ],
+        };
+        authService.has.resolves(true);
+        authService.has.withArgs('can_create_contacts_under_muted_places').resolves(false);
+        contactMutedService.getMuted.returns('2025-01-01T00:00:00Z');
+        responsiveService.isMobile.returns(false);
+
+        const actions = await service.getContactLeftSideActions(context);
+
+        // pins the `context.parentContact` forward: without it the muted branch is never reached
```
dianabarsan, line 680: It's not clear which line this comment refers to.

tests/e2e/default/contacts/muted-contacts-create.wdio-spec.js
```diff
@@ -0,0 +1,97 @@
...
+    // the muted clinic is the selected contact and its card has rendered. The FAB list is computed
+    // in a separate subscription, so this is margin rather than a barrier; the counterfactual in
+    // the permission-granted case is the real evidence that the assertion is load bearing.
+
+    // The right-side FAB is hidden entirely (*ngIf="fastActions?.length") when every action is
+    // filtered. Scoped to .item-content so the contacts-list left-side FAB is not matched. Holds for
+    // the seeded muted clinic because it has no phone and no clinic-scoped report forms; if seed
+    // data later grows action-emitting fields, this assertion will need to be tightened to inspect
+    // the open FAB list rather than the trigger itself.
+    expect(await $('.item-content .fast-action-trigger').isExisting()).to.be.false;
```
dianabarsan, line 48: I'm wondering if the best UX is to hide the button entirely. Would the user not be confused about why they cannot create new contacts? 
Have we considered to add a "disabled" state to the FAB button with clear messaging that the action is disabled.

tests/e2e/default/contacts/muted-contacts-create.wdio-spec.js
```diff
@@ -0,0 +1,97 @@
...
+  });
+
+  it('should show the create-contact action on a non-muted clinic', async () => {
```
dianabarsan, line 51: Wondering if this case should not be already covered by all the other test cases in the other suites.

tests/integration/api/controllers/places.spec.js
```diff
@@ -225,6 +225,226 @@ describe('Places API', () => {
...
+        }
+        // asserted outside the try so chai.expect.fail is not swallowed by this helper's own catch
+        chai.expect(caught, 'Should have rejected with 403').to.not.be.undefined;
+        chai.expect(caught.status).to.equal(403);
+        chai.expect(caught.body).to.deep.equal({
+          code: 403,
+          error: 'Insufficient privileges to create contacts on muted places',
+        });
+      };
+
```
dianabarsan, line 257: Can we not use the same pattern of chai-as-promised  here? 

```js
 await expect(utils.request(options)).to.be.rejectedWith(
          '403 - {"code":403,"error":"Insufficient privileges to create contacts on muted places"}'
        );
```

tests/integration/api/controllers/places.spec.js
```diff
@@ -225,6 +225,226 @@ describe('Places API', () => {
...
+      it('rejects with 403 when parent is muted and role lacks can_create_contacts_under_muted_places', async () => {
+        onlineRequestOptions.body = {
+          name: 'Should be blocked',
+          type: CONTACT_TYPES.HEALTH_CENTER,
+          parent: mutedDistrict._id,
+        };
+        await expect403(onlineRequestOptions);
+      });
+
+      it('allows creation under a non-muted parent', async () => {
```
dianabarsan, line 286: Is this not already covered by other test cases?

tests/integration/api/controllers/places.spec.js
```diff
@@ -225,6 +225,226 @@ describe('Places API', () => {
...
+
+      it('rejects with 403 when the parent is clean but its own parent is muted', async () => {
+        // the only case that exercises isMutedInLineage: hcUnderMuted carries no `muted` of its own,
+        // so `parent.muted` is false and the walk through the hydrated lineage is what refuses this
+        onlineRequestOptions.body = {
+          name: 'Should be blocked',
+          type: CONTACT_TYPES.CLINIC,
+          parent: hcUnderMuted._id,
+        };
+        await expect403(onlineRequestOptions);
```
dianabarsan: Does this path also have an exemption for the permission? Can a test case be added for that? 
Same question applies to the next suite.

tests/integration/api/controllers/places.spec.js
```diff
@@ -225,6 +225,226 @@ describe('Places API', () => {
...
+          movable = await utils.request(onlineRequestOptions);
+          onlineRequestOptions.path = `/api/v1/places/${movable.id}`;
+        });
+
+        it('rejects with 403 when moving a place under a muted parent', async () => {
+          onlineRequestOptions.body = { parent: mutedDistrict._id };
+          await expect403(onlineRequestOptions);
+        });
+
+        it('allows a rename of a place that already sits under a muted parent', async () => {
```
dianabarsan, line 370: The name of this test suggests that _only_ renaming is allowed, while this is not true, all edits are allowed. Can you please:
- change the title of the test to cover that all edits are allowed
- add more fields to be changed by the request.

tests/integration/api/controllers/places.spec.js
```diff
@@ -225,6 +225,226 @@ describe('Places API', () => {
...
+          // the gate only reads the incoming `parent`, so a body that omits it is never blocked
+          onlineRequestOptions.path = `/api/v1/places/${underMuted._id}`;
+          onlineRequestOptions.body = { name: 'Renamed, still allowed' };
+          const result = await utils.request(onlineRequestOptions);
+          chai.expect(result.id).to.equal(underMuted._id);
+        });
+
+        it('rejects with 403 when a rename resends the unchanged muted parent', async () => {
```
dianabarsan, line 378: As above, please make the test more generic, not pinned to a rename alone.

tests/integration/api/controllers/places.spec.js
```diff
@@ -225,6 +225,226 @@ describe('Places API', () => {
...
+          onlineRequestOptions.body = {
+            contact: { name: 'Sneaky CHW', type: 'person', place: mutedDistrict._id },
+          };
+          await expect403(onlineRequestOptions);
+        });
+
+        it('rejects with 403 when an inline contact carries a raw muted parent', async () => {
+          // createPerson stores `parent` verbatim when `place` is absent, so both are gated
+          onlineRequestOptions.body = {
+            contact: { name: 'Sneaky CHW 2', type: 'person', parent: { _id: mutedDistrict._id } },
```
dianabarsan, line 396: super sneaky indeed.

dianabarsan (changes requested):
Very nice work here!
I added some minor requests inline. 
One additional observation would be about some very verbose comments that are added inline. Ideeally the code should be self-explanatory, and we should avoid 10-line comments for a simple parameter choice. Please consider removing the verbose comments of drastically compressing them.

Issue #11017, opened 2026-05-06
Fix gaps in legacy UI styling for UI Extensions


jkuester:
When the user has the `can_view_old_navigation` they can see the old-style header tabs.  There are currently a couple style issues with how the new UI Extensions are rendered for the legacy UI.

## Add `accent_color` to "tab" part of the header for `app_main_tab` extensions.

When you have a main header tab selected, the "tab" part of the header should be highlighted with the proper accent color: 

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/21f91a6c-3e0a-49ca-a86b-8210d20cdd90" />

However, for UI Extensions, the "tab" part is not getting the background color set properly.  Here is an example where I just used `"accent_color": "red"` for my extension:

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/5e9e566b-e9f5-48fb-83af-8ad830b3559a" />

The red color should also be the background around the "Hello Extension" tab.

## When rendering `app_drawer_tab` extensions, do not include extra header bar at the top. 

The existing pages accessible from the legacy hamburger menu (About, Training Materials, etc) do not render the extra colored tool-bar space at the top of the page that you get with the main header tabs:

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/40908214-99c5-419e-8860-e7f5b1180f45" />

However, when I open an `app_drawer_tab` extension from the legacy hamburger menu it still renders the extra header bar.

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/b72e16bf-df5a-4c37-9552-56cdb8581c7e" />

We should not render this bar for the `app_drawer_tab` extensions.

YASHSHARMAOFFICIALLY:
@jkuester @andrablaj   Here's what it addresses:
accent_color on tab header — The route was setting currentTab to a static ui-extensions for all extensions, which didn't match the header tab name ui-extension-{id}. The component now updates currentTab to the correct extension-specific name when loading an app_main_tab, and the header template applies the accentColor as an inline background-color on the selected tab.

Extra toolbar for drawer tab extensions — Added &.ui-extensions to the existing old-nav.less selector group that hides the toolbar for hamburger-menu pages (about, training, etc.). Since only app_drawer_tab extensions keep the ui-extensions class (main tabs override it), the rule targets the right set.

jkuester:
https://github.com/medic/cht-core/pull/11020

PR #11020, opened 2026-05-06
fix(#11017): fix legacy UI styling gaps for UI extensions

YASHSHARMAOFFICIALLY:
# Description

Fixes two styling issues with UI Extensions when rendered under the legacy navigation (`can_view_old_navigation`).

## 1. accent_color not applied to selected tab for `app_main_tab` extensions

The legacy header applies background colors to the selected tab via CSS rules that match `.app-root.{tabName}`. However, all extension routes shared the static tab name `ui-extensions`, which did not match the header tab name `ui-extension-{id}`.

**Fix:** When an `app_main_tab` extension loads, the component now updates `currentTab` to `ui-extension-{id}` via the store, so it matches the header tab entry. The header template applies the extension's `accentColor` as an inline `background-color` on the selected tab element.

## 2. Extra toolbar rendered for `app_drawer_tab` extensions

Pages accessed from the legacy hamburger menu (About, Training Materials, etc.) hide the colored toolbar bar using `height: 0` in `old-nav.less`. But `app_drawer_tab` extensions were not included in that rule, so they rendered an unwanted toolbar.

**Fix:** Added `&.ui-extensions` to the existing CSS selector group in `old-nav.less` that hides the toolbar. Since `app_main_tab` extensions override `currentTab` to `ui-extension-{id}`, only `app_drawer_tab` extensions retain the `ui-extensions` class — so the rule targets the right set.

Fixes https://github.com/medic/cht-core/issues/11017

## Changes

| File | What |
|---|---|
| `ui-extensions-tab.component.ts` | Set `currentTab` to extension-specific name for `app_main_tab`; restore on destroy |
| `header-tabs.service.ts` | Pass `accentColor` through in `HeaderTab` interface |
| `header.component.html` | Apply `accentColor` as inline background on selected tab |
| `old-nav.less` | Add `&.ui-extensions` to toolbar-hiding and page-offset rules |
| Unit tests | Updated for both component and service |

# Code review checklist

- [x] Readable: Concise, follows existing patterns
- [x] Tested: Updated unit tests for both component and service
- [x] Backwards compatible: No breaking changes; existing tabs unaffected

# License

### DIFF
--- a/api/src/generate-service-worker.js
+++ b/api/src/generate-service-worker.js
@@ -5,11 +5,14 @@ const crypto = require('crypto');
 
 const resources = require('./resources.js');
 const db = require('./db');
+const config = require('./config');
 const logger = require('@medic/logger');
 const loginController = require('./controllers/login');
 const extensionLibs = require('./services/extension-libs');
 const uiExtensionService = require('./services/ui-extension');
 const { DOC_IDS } = require('@medic/constants');
+const dataContext = require('./services/data-context');
+const { roles } = require('@medic/user-management')(config, db, dataContext);
 
 const SWMETA_DOC_ID = DOC_IDS.SERVICE_WORKER_META;
 
@@ -73,12 +76,7 @@ const appendExtensionLibs = async (config) => {
 
 const appendUiExtensions = async (config) => {
   const extensions = await uiExtensionService.getAllProperties();
-  const offlineExtensions = extensions.filter(ext => {
-    if (!ext.roles?.length) {
-      return true;
-    }
-    return ext.roles.some(role => role.offline === true);
-  });
+  const offlineExtensions = extensions.filter(ext => !ext.roles || roles.isOffline(ext.roles));
 
   // cache this even if there are no extensions so offline client knows
   config.globPatterns.push('/ui-extension');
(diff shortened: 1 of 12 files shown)


jkuester:
@YASHSHARMAOFFICIALLY you might double-check your branch here and make sure you are based on the latest from https://github.com/medic/cht-core/tree/10224-ui-extensions.  :thinking:

YASHSHARMAOFFICIALLY:
@jkuester Rebased onto the latest `10224-ui-extensions` branch. Changes made during the rebase:

- Updated type checks from `app_main_tab` to `header_tab` and `app_drawer_tab` to `sidebar_tab` to match the rename in d2195434f.
- Kept the `paramSubscription` pattern from #11022 for route param handling, and layered the `GlobalActions`/`previousTab` tab-tracking logic on top.
- Updated all test mocks to provide both the observable `params` (for the subscription) and `snapshot.data` (for tab tracking), and updated type values in test assertions.

Ready for review.

YASHSHARMAOFFICIALLY:
@jkuester  @andrablaj Rebased again onto the latest `10224-ui-extensions`  the previous rebase had a build failure due to stale imports. CI should be green now. Ready for review.

jkuester (approved):
Thanks for the PR @YASHSHARMAOFFICIALLY! I ended up needing more tweaks (the styling complexity for the header tabs is crasy... 😓 ).  But it is now looking good. Along the way I fixed a couple more minor bugs I ran into around the userContactSummary and the service-worker.  

LGTM!

Issue #11019, opened 2026-05-06
Fix `UiExtensionsTabComponent` content not getting reloaded when navigating between extensions


jkuester:
UI Extension content loads properly in the `UiExtensionsTabComponent` when you navigate from an existing app page to a UI Extension tab.  However, if you have multiple UI Extensions configured and you try to navigate _between them_ (so start on one extension tab and then use the header tabs or sidebar buttons to navigate to the other extension) then the content of the target tab is never rendered and the user just sees the content for the original extension even though the URL path changes.  

Here is a video of the behavior I am seeing:

https://github.com/user-attachments/assets/875e1bcf-32cb-4ac7-889f-9b5f4ac671dc

The content in the `UiExtensionsTabComponent` should properly render the extension content for the extension identified in the URL.

YASHSHARMAOFFICIALLY:
@jkuester  @andrablaj  @dianabarsan  Added a subscription to ActivatedRoute.params that detects when the :id parameter changes. When it does, the container is cleared and the extension is re-initialized with the new id. The subscription is cleaned up on component   destroy.

Abhishek9639:
Hi @jkuester, 
I’d like to work on this if it’s still available. From what I understand, Angular is reusing the same `UiExtensionsTabComponent` instance when navigating between `ui-extensions/:id` routes, so `ngAfterViewInit` only runs for the first extension. The URL changes, but the old extension content remains mounted because the component does not react to the changed route param.

My plan would be:
- Reproduce the issue with two `app_main_tab` UI extensions.
- Listen for changes to the route `id` param inside `UiExtensionsTabComponent`.
- When the extension id changes, reset the component state, clear the current extension container, and re-run the existing initialization/render logic for the new extension.
- Clean up the route subscription on destroy.
- Add tests for switching between extensions, same-extension no-op behavior, and cleanup.

I noticed there is already a linked PR for this, so happy to coordinate with that or take it up if maintainers would still like another pass. Could you please assign this to me if that works?
Thanks

jkuester:
https://github.com/medic/cht-core/pull/11022

PR #11022, opened 2026-05-06
fix(#11019): reload extension content when navigating between tabs

YASHSHARMAOFFICIALLY:
## Description

Fixes https://github.com/medic/cht-core/issues/11019

When navigating between multiple UI extension tabs, the `UiExtensionsTabComponent` was not re-rendering because Angular reuses the same component instance for `ui-extensions/:id` — only the `:id` param changes. Since extension initialization only ran in `ngAfterViewInit` (which fires once), the original extension's content stayed visible even after the URL changed.

## Changes

- Subscribe to `ActivatedRoute.params` in `ngAfterViewInit` to detect when the `:id` parameter changes
- When a different extension is navigated to, clear the container and re-run `initializeExtension()` with the new id
- Reset component state (loading, error, title, accent color) before each initialization
- Unsubscribe from params on component destroy to prevent memory leaks
- Added tests for navigation between extensions, same-id no-op, and cleanup on destroy

## Test plan

- [ ] Configure two or more UI extensions as `app_main_tab`
- [ ] Navigate from one extension tab to another via header tabs
- [ ] Verify the target extension's content renders correctly
- [ ] Navigate back to the first extension and verify it renders again
- [ ] Verify the loading spinner shows during each transition
- [ ] Unit tests pass

### DIFF
--- a/webapp/src/ts/modules/ui-extensions/ui-extensions-tab.component.ts
+++ b/webapp/src/ts/modules/ui-extensions/ui-extensions-tab.component.ts
@@ -1,6 +1,7 @@
-import { AfterViewInit, Component, ElementRef, ViewChild } from '@angular/core';
+import { AfterViewInit, Component, ElementRef, OnDestroy, ViewChild } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
 import { NgIf } from '@angular/common';
+import { Subscription } from 'rxjs';
 import { CHTDatasourceService } from '@mm-services/cht-datasource.service';
 import { PerformanceService } from '@mm-services/performance.service';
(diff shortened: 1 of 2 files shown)


jkuester:
Same here @YASHSHARMAOFFICIALLY.  Should not have the merge conflicts. :thinking:

YASHSHARMAOFFICIALLY:
@jkuester  Have fix the branch sorry my bad, do check now and let me know if its fixed or not

jkuester (approved):
Very nice!  I just had a minor tweak to just track the extension id locally instead of trying to compare with the existing web component name.

LGTM!

PR #11023, opened 2026-05-06
fix(#10679): prevent geolocation timeout from blocking form submission

### LINKED ISSUE #10679: Geolocation acquisition timeout blocks submitting forms
**Describe the bug**
When we submit a form, the geolocation service tries to store GPS coordinates. The service tries to get GPS coordinates for 30 seconds before quitting: 

https://github.com/medic/cht-core/blob/e21d8f8b0d1456ca5534ef19a944186b2e195a84/webapp/src/ts/services/geolocation.service.ts#L11

Ideally, the coordinates would be retrieved quickly and no wait would be required. 
However, when GPS signal is weak, coordinates acquisition can often exceed the 30 second timeout we have set. When this happens, and the user attempts to submit the form before the 30 second timeout expiry, they will see a spinner by the submit button for the remaining duration. 

**To Reproduce**
1. Make sure GPS permission is allowed
2. Go somewhere with weak GPS signal 
3. Try to submit a small form quickly
4. You wait for 30 seconds before submission actually happens

**Expected behavior**
This caused confusion in some deployments and this ended up having CHWs turn off their GPS entirely. Having to wait for 30 seconds is very frustrating. Such as in this video. 

https://github.com/user-attachments/assets/f16c0be1-7ed3-4ba2-bbc3-2d6ac1fa60cd

**Environment**
- App: webapp
- Version: all

YASHSHARMAOFFICIALLY:
## Description

Fixes https://github.com/medic/cht-core/issues/10679

When GPS signal is weak and coordinates haven't been acquired by the time a user submits a form, the geolocation service was blocking submission for up to 30 seconds waiting for the watcher to either succeed or time out. This caused frustration and led some deployments to disable GPS entirely.

## Changes

The `complete()` function (called during form save) now resolves immediately when no geolocation data is available, returning `{code: -1, message: 'Geolocation not yet acquired'}` instead of hanging. If coordinates were already acquired while the user filled out the form, they are still captured as before.

After `complete()` resolves, `this.deferred` is nulled so that any late GPS watcher callbacks are silently discarded and do not leak additional telemetry entries.

This means:
- Forms submit instantly regardless of GPS state
- Geolocation data is still recorded when available (acquired before submission)
- When GPS was not acquired in time, the geolocation field reflects that clearly
- Late GPS callbacks after submission are safely ignored

## Telemetry impact

This shifts telemetry distribution: deployments tracking `geolocation:failure:-2` (30s timeout) will now mostly see `geolocation:failure:-1` instead, since the `-2` code only fires for forms left open longer than 30 seconds without submission. The total volume of failure events remains the same, but the code distribution changes. Any alerting or dashboards keyed on `-2` will need to be updated to also track `-1`.

## Test plan

- [ ] Open a form in an environment with no/weak GPS signal
- [ ] Submit the form quickly (before 30s)
- [ ] Verify the form submits immediately without waiting
- [ ] Verify the saved doc has `geolocation: {code: -1, message: 'Geolocation not yet acquired'}`
- [ ] Open a form and wait for GPS to be acquired, then submit
- [ ] Verify the saved doc has valid geolocation coordinates
- [ ] Unit tests pass


dianabarsan (changes requested):
Nice one. 

The new tests cover "submit before any data" and "data acquired before submit", but the actual behavior change — that a `success`/`failure` callback firing *after* `complete()` is now silently discarded — isn't asserted anywhere. The previous "blocks promise completion until at least one success" test exercised that exact race (submit, then watcher
   fires) and was removed.                                                                                                                                                                                                                                                                                                                                      
   
  Could you add a test that:                                                                                                                                                                                                                                                                                                                                    
  1. Calls `complete()` with no data acquired (resolves with `code: -1`)
  2. Then fires a `success` callback for the watcher                                                                                                                                                                                                                                                                                                            
  3. Asserts the resolved value is still `code: -1` and `telemetryService.record` was called once with `geolocation:failure:-1` (not also `geolocation:success`)                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                
  That guards the new contract: once the form has been submitted, late GPS results don't leak in.                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                
  Worth mentioning in the PR description / changelog that this shifts telemetry: deployments tracking `geolocation:failure:-2` (30s timeout) will now mostly see `geolocation:failure:-1` instead, since `-2` only fires for forms left open >30s without submission. Same total volume, different code distribution — anyone with alerting/dashboards keyed on 
  `-2` will want to update.

YASHSHARMAOFFICIALLY:
@dianabarsan  @andrablaj  Thanks for the thorough review.                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                        Test added: Added a test ('discards late GPS success callback after form submission') that does exactly what you described:                                                                                                                                                                                           1. Calls complete() with no data acquired — resolves with code: -1                                                                                                                                                                                                                                                    
2. Fires a late success callback from the watcher                 
4. Asserts the resolved value is still code: -1 and telemetryService.record was called exactly once with geolocation:failure:-1 — no geolocation:success leak                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                        
While writing the test, I found that late callbacks could actually re-enter finalise() because this.deferred was still set after complete() resolved. Fixed this by nulling this.deferred after saving the promise reference, so late watcher callbacks hit the if (this.deferred) guard and are silently dropped.    
                                                                                                                                                                                                                                                                                                                        
PR description updated: Added a "Telemetry impact" section noting that deployments will now mostly see geolocation:failure:-1 instead of geolocation:failure:-2, and that any alerting/dashboards keyed on -2 should be updated to also track -1.

dianabarsan (changes requested):
LGTM on the requested changes — late-callback test and telemetry disclosure both look good.

  One follow-up: removing the old `should resolve promise even if watcher never calls any callback` test means the 30s `-2` timeout path no longer has a direct test in isolation. The remaining `timeout should prioritize ...` cases always have a success/failure callback firing alongside the timeout, so the synthetic `-2` from the internal
  `setTimeout` never wins on its own. Could you add a small `fakeAsync` test that calls `init()`, ticks `31 * 1000` without firing any watcher callback, then awaits `complete()` and asserts the result is `{code: -2, message: 'Geolocation timeout exceeded'}` and telemetry was recorded with `geolocation:failure:-2`? That keeps the long-form-open path
  covered.

YASHSHARMAOFFICIALLY:
@andrablaj  @dianabarsan Added the isolated fakeAsync test for the 30s timeout path:
                                                                                                                                                                  
  - Calls init(), ticks 31 * 1000 without any watcher callback firing
  - Awaits complete() and asserts the result is {code: -2, message: 'Geolocation timeout exceeded'}                                                               
  - Asserts telemetry was recorded with geolocation:failure:-2                                                                                                    
                                                                                                                                                                  
  The long form open -2 timeout path is now covered in isolation. Ready for re review.

dianabarsan (approved):
Very nice! Thanks for this one! Believe it or not, many people had turned off GPS because they had to wait when submitting short forms.

Issue #10996, opened 2026-05-05
Update helm charts nginx template to use same server config as docker nginx


dianabarsan:
**Describe the issue**
Docker nginx uses these values: https://github.com/medic/cht-core/blob/8e2d62c5ddde5f911927260ea75dbb60e247532a/nginx/templates/server.conf.template

However, the [helm nginx](https://github.com/medic/cht-core/blob/3775e36dfecd9d8b0f27b1cd7c355f26949248ee/scripts/build/helm/templates/nginx/service.yaml) does not. 

This causes some issues and differences in behavior in e2e tests, where some headers are not forwarded. 

**Describe the improvement you'd like**
Keep the same configuration for both systems.

Abhishek9639:
Hey @dianabarsan,
I'd like to work on this issue. I'll sync the helm nginx template with the Docker server.conf.template so both systems use the same server config and headers are forwarded consistently. 
Can you please assign this to me?
Thanks

Abhishek9639:
Hey @dianabarsan or @jkuester,

I looked at both configs the Docker `server.conf.template` has `X-Real-IP`, `X-Forwarded-For`, `X-Forwarded-Proto`, and 
`X-Forwarded-Host` but the helm side is missing all of them.

I'm thinking of adding these headers to the helm nginx ConfigMap to keep them in sync. Should I mirror the exact same values from the Docker template, or are there any helm-specific variables I should use?

Thanks!

dianabarsan:
Please just mirror the same configs used here: https://github.com/medic/cht-core/blob/master/nginx/templates/server.conf.template#L8

PR #11026, opened 2026-05-07
fix(10996): mirror nginx proxy headers in helm k3s-k3d ingress template

Abhishek9639:
# Description

The Docker nginx reverse proxy sets several proxy headers via `nginx/templates/server.conf.template` (X-Real-IP, X-Forwarded-For, X-Forwarded-Proto, X-Forwarded-Host), but the helm nginx ingress for k3s-k3d was missing all of them. This caused inconsistent behavior in e2e tests where some headers were not being forwarded.

This fix adds a `nginx.ingress.kubernetes.io/configuration-snippet` annotation to the k3s-k3d ingress template to mirror the same proxy headers, so both Docker and Helm deployments behave consistently.

relates to #10996

# Code review checklist
- [x] Tested: Ran `validate-templates.sh`- all 12 helm template validation scenarios pass
- [x] Backwards compatible: This is a purely additive change - no breaking changes
- [x] AI disclosure: Please disclose use of AI per [the guidelines](https://place.communityhealthtoolkit.org/contributing/contributing-guidelines).

# License
The software is provided under AGPL-4.0. Contributions to this project are accepted under the same license.

### DIFF
--- a/scripts/build/helm/templates/platforms/k3s-k3d/ingress.yaml
+++ b/scripts/build/helm/templates/platforms/k3s-k3d/ingress.yaml
@@ -4,6 +4,13 @@ apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: api-ingress
+  annotations:
+    # Mirror the same proxy header config as the Docker nginx server.conf.template
+    nginx.ingress.kubernetes.io/configuration-snippet: |
+      proxy_set_header X-Real-IP       $remote_addr;
+      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+      proxy_set_header X-Forwarded-Proto https;
+      proxy_set_header X-Forwarded-Host $server_name;
 spec:
   rules:
     - http:


Abhishek9639:
Hey @dianabarsan or @jkuester,

I've added the same proxy headers from server.conf.template to the helm k3s-k3d ingress using the configuration-snippet annotation. Ran validate-templates.sh and all scenarios pass.

Could you please take a look when you get a chance? 
Thanks

dianabarsan (changes requested):
Thanks for this! The header values match `nginx/templates/server.conf.template` exactly, but I don't think this annotation actually takes effect in our k3d/CI environment, so I want to sort out the mechanism before we merge.

  The `nginx.ingress.kubernetes.io/configuration-snippet` annotation is only honored by the **ingress-nginx** controller. Our k3d cluster doesn't run it:

  - `tests/utils/index.js` creates the cluster with `k3d cluster create` and no `--disable=traefik`, so k3s's default **Traefik** ingress controller is what's active.
  - There's no ingress-nginx controller installed anywhere in the repo — only the `ingress-nginx-controller-admission` stub Service (no backing pods).
  - The Ingress has no `ingressClassName`, so Traefik serves it, and Traefik ignores `nginx.ingress.kubernetes.io/*` annotations entirely.

  So as written this is a no-op in the e2e tests it's meant to fix. 
Could you confirm on a live k3d run that these headers actually reach `api` (e.g. log the received headers)? If Traefik is what's serving the ingress, the equivalent fix would be a Traefik `Middleware` + `router.middlewares` annotation rather than an nginx snippet.

  Also, a note on testing: `validate-templates.sh` only runs `helm template`, which renders YAML but never exercises a live controller — so passing it doesn't confirm header forwarding. A live-cluster check would give us the real signal here.

PR #11036, opened 2026-05-08
fix(#10851): normalize Devanagari numerals across all search paths

shivv23:
## Summary

Fix Devanagari numeral search regression by normalizing numerals at both index-time and query-time across all search paths. This restores the 4.x behavior where searching with Devanagari numerals (१२३४५) or Latin numerals (12345) would both return matching contacts/reports regardless of which script is used in the data.

## Approach

Character-by-character Unicode math normalization: `d.codePointAt(0) - 0x0966 + 0x0030`. No analyzers, no dependencies — safe, predictable, reviewable.

### Index-time normalization (bidirectional matching)
- **`ddocs/medic-db/medic/nouveau/contacts_by_freetext/index.js`** — converts Devanagari → Latin before indexing contact tokens
- **`ddocs/medic-db/medic/nouveau/reports_by_freetext/index.js`** — same for report tokens
- **`webapp/src/js/bootstrapper/offline-ddocs/medic-offline-freetext/contacts_by_freetext.js`** — same for PouchDB offline views
- **`webapp/src/js/bootstrapper/offline-ddocs/medic-offline-freetext/contacts_by_type_freetext.js`** — same for type-filtered offline search
- **`webapp/src/js/bootstrapper/offline-ddocs/medic-offline-freetext/reports_by_freetext.js`** — same for offline reports

### Query-time normalization (defense in depth)
- **`shared-libs/cht-datasource/src/local/libs/core.ts`** — `normalizeFreetextQualifier` normalizes numerals in addition to trim/lowercase (covers REST API + cht-datasource local path)
- **`shared-libs/search/src/generate-search-requests.js`** — normalizes numerals in freetext search term processing

### Tests
- **`shared-libs/cht-datasource/test/local/libs/core.spec.ts`** — 4 test cases (basic, full range, mixed, text context)
- **`shared-libs/search/test/generate-search-requests.js`** — 3 test cases (contacts, multi-word, reports)
- **`webapp/tests/mocha/unit/views/contacts_by_freetext.spec.js`** — Devanagari numeral view test
- **`webapp/tests/mocha/unit/views/contacts_by_type_freetext.spec.js`** — same for type view
- **`webapp/tests/mocha/unit/views/reports_by_freetext.spec.js`** — same for reports view

## E2E Tests

E2E tests for this feature are in **PR #11053** (`test(#10851): add e2e tests for Devanagari numeral cross-script search`), covering both contacts and reports search for online and offline users.

## Closes
- #10851

PR #11044, opened 2026-05-10
fix(#10851): normalize Devanagari numerals for cross-script freetext search

### LINKED ISSUE #10851: Searching with Devanagari numerals no longer cross-matches Latin numeral contacts for online users (regression from 4.x)

swetalin-10:
## Problem

Since the switch from Couch map-reduce views to **Nouveau** (CouchDB full-text search) in CHT 5.x, searching contacts or reports using **Latin numerals** (e.g. `12346`) no longer returns results whose names contain **Devanagari numerals** (e.g. `१२३४६`), and vice-versa.  
This is a regression from CHT 4.x where the Couch view query layer performed the cross-matching automatically.  
Offline users (PouchDB) never had cross-matching, so this fix extends the fix to both paths.

Closes #10851

## Root cause

The freetext indexes/views store and query the exact byte sequence of the numeral characters. Devanagari `१२३४५` and Latin `12345` hash differently in Lucene/Nouveau and in PouchDB B-tree indexes, so one never matches the other.

## Solution

Normalize Devanagari digits (Unicode U+0966 `०` – U+096F `९`) to their Latin equivalents (`0`–`9`) **both** at index/emit time **and** at query time. This ensures a single canonical form (`12345`) is stored and searched regardless of which script the user types or what the contact data contains.

The normalization function (`d.charCodeAt(0) - 0x0966 + 0x0030`) is inlined independently in each file because, as noted by @jkuester, view/index files must be self-contained and cannot import shared modules.

### Files changed

| File | Change |
|------|--------|
| `webapp/src/js/bootstrapper/offline-ddocs/medic-offline-freetext/contacts_by_freetext.js` | Normalize string values before emitting freetext keys |
| `webapp/src/js/bootstrapper/offline-ddocs/medic-offline-freetext/contacts_by_type_freetext.js` | Same |
| `webapp/src/js/bootstrapper/offline-ddocs/medic-offline-freetext/reports_by_freetext.js` | Same |
| `ddocs/medic-db/medic/nouveau/contacts_by_freetext/index.js` | Normalize before Nouveau `index()` calls |
| `ddocs/medic-db/medic/nouveau/reports_by_freetext/index.js` | Same |
| `shared-libs/cht-datasource/src/local/libs/core.ts` | Normalize in `normalizeFreetextQualifier()` so search queries always use Latin digits |

## Test plan

- [x] Unit tests added to `contacts_by_freetext.spec.js`, `contacts_by_type_freetext.spec.js`, `reports_by_freetext.spec.js`: verify Devanagari-digit strings are emitted with Latin digits
- [x] Unit tests added to `core.spec.ts`: verify `normalizeFreetextQualifier` converts all Devanagari digit variants (all-Devanagari, all-Latin, mixed, with text)
- [ ] E2E: create contact "Person 12346" and "Person १२३४६"; search both ways and confirm cross-matching works for online and offline users (see `tests/e2e/default/contacts/search-contacts.wdio-spec.js`)


sugat009:
Closing. Covered by #11036.

PR #11047, opened 2026-05-11
feat(#10908): support order weight for header and sidebar tabs

### LINKED ISSUE #10908: Support order weight for header_tabs in webapp
Blocked by https://github.com/medic/cht-core/issues/10672 and https://github.com/medic/cht-core/issues/10901

Currently the ordering of the header tabs ("Messages", "Reports", "People", etc) is fixed and non-configurable. For the new UI Extensions we are going to be adding more custom tabs.  Deployments want to be able to control the order that the tabs are displayed (which also should control which tab the user initially lands on).  This has been [repeatedly requested on the forum](https://forum.communityhealthtoolkit.org/t/is-there-a-way-of-changing-the-order-of-the-tabs-of-cht/2093).  We should support the ability to configure a custom order "weight" that works both for default tabs and for `app_main_tab` UI Extensions.

Currently we have the [`header_tabs`](https://docs.communityhealthtoolkit.org/building/reference/app-settings/header_tabs/) config object that allows for setting custom icons for the header tabs. We should enhance the webapp to support reading a `weight` property from the `header_tabs` entries (and from the UI Extension properties being supported in https://github.com/medic/cht-core/issues/10672).  Entries with a lower `weight` should be ordered first.  The default `weight` for the default header tabs should be:

- messages: `1`
- tasks: `2`
- reports: `3`
- contacts: `4`
- analytics: `5`

Any custom UI Extension entries should have a default weight of `6`.  Any entries with the same weight value should be ordered alphabetically.  

---

https://github.com/medic/cht-core/issues/10672 is updating `webapp/src/ts/services/header-tabs.service.ts` to return entries for the custom UI Extensions. This issue should order the returned array according to the `weight` property both from the `header_tabs` config as well as the UI Extension properties.  

We should also update the ordering of the sidebar tags in `webapp/src/ts/components/sidebar-menu/sidebar-menu.component.ts` based on this same `weight`.

https://github.com/medic/cht-core/issues/10901 is adding the UI Extensions to the legacy header menu.  We should update that functionality to also order the entries based on `weight`.

Vedd-Patel:
# Description

currently, the ordering of the header tabs ("Messages", "Tasks", "Reports", "Contacts", "Analytics") and the corresponding sidebar menu items is fixed, as we introduce more custom UI Extensions, deployments require the ability to control the display order of these tabs!

this PR introduces a configurable `weight` property to establish a single source of truth for sorting navigation elements across the webapp

**key changes:**
* **weight-based sorting:** updates `HeaderTabsService` and `SidebarMenuComponent` to sort items in ascending order based on their `weight`
* **configuration sources:** reads custom weights primarily from the `header_tabs` app settings, falling back to properties defined directly on `app_main_tab` UI extensions
* **backwards compatibility:** applies strict default weights to built-in tabs (`messages: 1`, `tasks: 2`, `reports: 3`, `contacts: 4`, `analytics: 5`) to perfectly preserve the existing UI order for deployments without custom configurations, custom UI extensions default to a weight of `6`
* **tie-breaker:** introduces a secondary alphabetical sort for tabs sharing the exact same weight
* **testing:** updated Karma unit test suites (`app.component.spec`, `sidebar-menu.component.spec`, `header-tabs.service.spec`) to inject new dependencies and assert against the new weighted mock objects and all tests are passing

FIXES #10908

# Code review checklist
- [ ] AI disclosure: Please disclose use of AI per [the guidelines](https://docs.communityhealthtoolkit.org/community/contributing/ai-guidelines/). *(Note: Used AI assistance for refactoring TypeScript sorting logic, strict type compliance, and unit test dependency mocking).*

# License

### DIFF
--- a/webapp/src/ts/components/sidebar-menu/sidebar-menu.component.ts
+++ b/webapp/src/ts/components/sidebar-menu/sidebar-menu.component.ts
@@ -21,6 +21,31 @@ import { SettingsService } from '@mm-services/settings.service';
 import { filter } from 'rxjs/operators';
 import { Selectors } from '@mm-selectors/index';
 
+export interface ExtensionOption {
+  id?: string;
+  name?: string;
+  route?: string;
+  icon?: string;
+  title?: string;
+  permissions?: string[];
+  weight?: number;
+}
+
+export interface ChtSettings {
+  header_tabs?: Record<string, { weight?: number; icon?: string; resource_icon?: string; }>;
+  app_main_tab?: ExtensionOption[];
+}
+
+export interface MenuOption {
+  icon: string;
+  translationKey: string;
+  routerLink?: string;
+  hasPermissions?: string;
+  canDisplay?: boolean;
+  click?: () => void;
+  weight?: number;
+}
+
 @Component({
   selector: 'mm-sidebar-menu',
   templateUrl: './sidebar-menu.component.html',
@@ -43,7 +68,7 @@ export class SidebarMenuComponent extends BaseMenuComponent implements OnInit, O
   @Input() canLogOut: boolean = false;
   @ViewChild('sidebar') sidebar!: MatSidenav;
   private globalActions: GlobalActions;
-  replicationStatus: any;
+
   moduleOptions: MenuOption[] = [];
   secondaryOptions: MenuOption[] = [];
   adminAppPath: string = '';
@@ -79,7 +104,7 @@ export class SidebarMenuComponent extends BaseMenuComponent implements OnInit, O
   }
 
   replicate(): void {
-    if (this.replicationStatus?.current?.disableSyncButton) {
+    if ((this as any).replicationStatus?.current?.disableSyncButton) {
       return;
     }
     super.replicate();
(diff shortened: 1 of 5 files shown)


Vedd-Patel:
@jkuester sorry for the long wait but the tests were really bothering me, so it took me some extra time to sort them out!!
pls take a look when you get a moment :)

jkuester:
@Vedd-Patel thanks for submitting this. 👍   Unfortunately, these changes are not exactly what we need. I think the main problem was that I was not detailed enough in the issue description. Really, I just did not fully understand the scope of the changes when I logged the issue.  :disappointed: 

The main challenge with continuing with this PR is that we needed to base these updates on the changes that already exist in https://github.com/medic/cht-core/tree/10224-ui-extensions (that was the whole reason we were blocked on addressing this issue in the first place).  With your branch being based on `master`, I was just getting a massive amount of merge conflicts when trying to reconcile with `10224-ui-extensions`. On top of that, we would still need to refactor everything again to account for the `weight` ordering of the UI Extensions.

The deeper I dug on this, the more I realized that we needed a non-trivial refactoring to keep everything consistent and maintainable here.  At that point, I already had the context for what needed to happen, so I have gone forwards with https://github.com/medic/cht-core/pull/11057 as a replacement for this PR.

Vedd-Patel:
@jkuester no worries, i'll take a look at #11057 (just out of curiosity and better understandings)
also want to tell, i've submitted proposal for 10706 
looking forward to collaborate and work together on that
also will convert the draft to PR for #11054 soon 👍

PR #11050, opened 2026-05-12
feat(#10224): support ui-extensions

### LINKED ISSUE #10224: Support custom UI Extensions

jkuester:
# Description

Closes https://github.com/medic/cht-core/issues/10224

cht-docs PR: https://github.com/medic/cht-docs/pull/2213
cht-conf PR: https://github.com/medic/cht-conf/pull/823

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10224-ui-extensions/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10224-ui-extensions/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10224-ui-extensions/docker-compose/cht-couchdb-clustered.yml)

# License


webapp/src/ts/modules/ui-extensions/ui-extensions-tab.component.ts
```diff
@@ -0,0 +1,88 @@
...
+      const {
+        properties: { title, config, accent_color },
+        Element
+      } = await this.uiExtensionsService.getExtension(extensionId);
+      this.extensionTitle = title ?? '';
+      this.accentColor = accent_color;
+      if (!customElements.get(elementName)) {
+        customElements.define(elementName, Element);
+      }
+      const element = document.createElement(elementName);
```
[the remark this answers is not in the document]
jkuester: This value is _configurable_ (via admins) but is not really "user-provided".

jkuester:
@dianabarsan this is complete and ready for final review!  All the changes here have already been reviewed incrementally (by one of us). So, this is just a final check for sanity, completeness, integration, etc.

Thank you!

jkuester:
FYI, you found [a serious hole](https://github.com/medic/cht-conf/pull/823#discussion_r3358611007) in the validation logic (where we are not actually setting or validating the `ui-extensions` doc type).   This means that the `validate_doc_update` logic is not ensuring only admins are updating the extension docs.  I will fix this in this branch tomorrow and ping this thread again when the changes are complete.  :+1:

jkuester:
Okay @dianabarsan this is up-to-date and ready for review!  :pray:

api/src/services/ui-extension.js
```diff
@@ -0,0 +1,53 @@
+const db = require('../db');
+
+const TYPE = 'ui-extension';
```
dianabarsan: maybe we can put this in the constants library?

api/src/services/ui-extension.js
```diff
@@ -0,0 +1,53 @@
...
+const PREFIX = `${TYPE}:`;
+
+const getExtensionDoc = (name) => db.medic
+  .get(`${PREFIX}${name}`, { attachments: true })
+  .then(doc => doc.type === TYPE ? doc : null)
+  .catch(err => {
+    if (err.status === 404) {
+      return;
```
dianabarsan: This function will end up having two return types, `null` when the doc has the right id but not the right type and `undefined` when not found. Just wanted to make sure this is intentional.

api/src/services/ui-extension.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+    throw err;
+  });
+
+module.exports = {
+  isExtensionChange: ({ id }) => id.startsWith(PREFIX),
+
+  getScript: async (name) => {
+    const doc = await getExtensionDoc(name);
+    const attachment = doc?._attachments?.['extension.js'];
```
dianabarsan: maybe 'extension.js' can also be a constant? I assume this is used in other places as well?

api/src/generate-service-worker.js
```diff
@@ -70,6 +74,31 @@ const appendExtensionLibs = async (config) => {
...
+  // Do not include the actual endpoints for online-only extensions. These should only be called by online users and
+  // do not need to be cached in the service worker at all.
+  const offlineExtensions = extensions.filter(ext => !ext.roles || roles.isOffline(ext.roles));
+  for (const ext of offlineExtensions) {
+    const extPath = path.join('/ui-extension', ext.id);
+    config.globPatterns.push(extPath);
+    
+    const script = await uiExtensionService.getScript(ext.id);
+    if (script?.data) {
+      config.templatedURLs[extPath] = script.data;
```
dianabarsan: I wonder if we should store the template in the service worker at all, or just add the URL where the data is returned to the cached resources list. This will keep the service worker itself slim, and the client will still request the soruce upon update. 

If the UI extension collection becomes large, the initial download of the service worker would be significant, and all will need to be downloaded in one block.

tests/e2e/default/navigation/navigation.wdio-spec.js
```diff
@@ -89,6 +91,27 @@ describe('Navigation tests', () => {
...
+          },
+          reports: {
+            weight: 43,
+            icon: 'fa-user',
+          },
+          contacts: {
+            weight: 1,
+            icon: 'fa-bar-chart-o'
+          },
+        } }, { ignoreReload: false });
```
dianabarsan: I believe ignoreReload is by default false.

tests/e2e/default/service-worker/service-worker.wdio-spec.js
```diff
@@ -107,6 +107,7 @@ describe('Service worker cache', () => {
       '/audio/alert.mp3',
       '/deploy-info.json',
       '/extension-libs',
+      '/ui-extension',
```
dianabarsan: Given that the meat of ui-extensions is them being cached here, I think more coverage is needed. Please add tests where UI-extensions exist and prove how they are cached.

webapp/src/ts/modules/ui-extensions/ui-extensions-tab.component.ts
```diff
@@ -0,0 +1,126 @@
...
+    const trackRender = this.performanceService.track();
+    try {
+      const {
+        properties: { title, config, accent_color, extension_type },
+        Element
+      } = await this.uiExtensionsService.getExtension(extensionId);
+      this.extensionTitle = title ?? '';
+      this.extensionType = extension_type;
+      this.accentColor = accent_color;
+      if (!customElements.get(elementName)) {
```
dianabarsan: what is customElements here?\

webapp/src/ts/services/ui-extensions.service.ts
```diff
@@ -0,0 +1,111 @@
...
+        }
+        if (b.weight !== undefined) {
+          return 1;
+        }
+        return a.id.localeCompare(b.id);
+      });
+  }
+
+  async getProperties(id: string): Promise<UiExtensionProperties> {
+    await this.init();
```
dianabarsan, line 105: I almost home that this would become a decorator. This is so common in our services, to have an initial setup, and then everything depends on that setup completing. 
Sending word to Angular santa to add a decorator `@inited` for functions that would only run after the service is inited.

webapp/src/ts/app.component.ts
```diff
@@ -409,7 +409,8 @@ export class AppComponent implements OnInit, AfterViewInit {
           change.id === '_design/medic' ||
           change.id === '_design/medic-client' ||
           change.id === DOC_IDS.SERVICE_WORKER_META ||
-          change.id === DOC_IDS.SETTINGS
+          change.id === DOC_IDS.SETTINGS ||
+          change.id.startsWith('ui-extension:') // online-only extensions for online users
```
dianabarsan: ha! I knew I would find this constant somewhere!

webapp/tests/karma/ts/modules/ui-extensions/ui-extensions-route.guard.provider.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+    expect(result).to.be.false;
+    expect(uiExtensionsService.getProperties).to.have.been.calledOnceWithExactly('ext-3');
+  });
+
+  it('should not activate when extension properties not found', async () => {
+    uiExtensionsService.getProperties.rejects(new Error('not found'));
+    const result = await guard.canActivate(getRoute('missing-ext'));
+    expect(result).to.be.false;
+    expect(uiExtensionsService.getProperties).to.have.been.calledOnceWithExactly('missing-ext');
+  });
```
dianabarsan, line 56: Can you add some tests around permissions here? It seems that the route guard also handles that.

webapp/tests/karma/ts/services/cht-datasource.service.spec.ts
```diff
@@ -285,6 +299,55 @@ describe('CHTScriptApiService service', () => {
     });
   });
 
+  describe('v1.translate()', () => {
```
dianabarsan, line 312: with the addition of translate, this interface extends FAR from datasource only. I think `chtApi` would be a better name. Not something to change now, but just recognizing this is not a datasource bridge anymore.

tests/e2e/default/ui-extensions/ui-extensions-tab.wdio-spec.js
```diff
@@ -0,0 +1,176 @@
...
+      it('renders the extension content when navigating to the tab', async () => {
+        await $(`#${HEADER_TAB_ID}`).click();
+        await commonPage.waitForPageLoaded();
+
+        const customElement = await $('#ui-element-tab cht-header-tab');
+        await customElement.waitForDisplayed();
+        expect(await customElement.getText()).to.equal('Hello world');
+
+        const telemetries = await getTelemetry(`ui-extension:header-tab:render`, user.username);
+        // Rendered twice because this is the first tab (weight 0) and is loaded when the user logs in
```
dianabarsan: It looks like it is not rendered twice? (the comment contradicts it)

webapp/src/ts/services/header-tabs.service.ts
```diff
@@ -1,133 +1,211 @@
...
+      const [headerTabs, uiExtensionTabs] = await Promise.all([
+        this.getHeaderTabs(this.SIDEBAR_HEADER_TABS),
+        this.getUiExtensionTabs('sidebar_tab')
+      ]);
+      headerTabs.sort((a, b) => a.weight - b.weight);
+      // sidebar_tab extensions should always come after the headers
+      this.sidebarTabs = [
+        ...headerTabs,
+        ...uiExtensionTabs,
+        ...this.SIDEBAR_TABS
```
dianabarsan: It looks like the sidebar tabs are never checked for permissions. The User Settings link does require `can_edit_profile`, but that is no longer checked.

dianabarsan (changes requested):
Nice work! I added some comments, nothing major.

api/src/services/ui-extension.js
jkuester: I will push back a bit on this one.  This is the only place where we use this attachment name (this is the only place where we load the attachment from the doc).  In the webapp, we get the attachment data directly from the service worker.

api/src/services/ui-extension.js
jkuester: Def not intentional/necessary.  Fixed.

webapp/src/ts/modules/ui-extensions/ui-extensions-tab.component.ts
jkuester: It is just a property of `window`.  https://developer.mozilla.org/en-US/docs/Web/API/Window/customElements  I have added the `window.` to make this more explicit.

api/src/generate-service-worker.js
jkuester: We were just trying to follow Gareth's pattern from above with the extension libs. Claude told me that:

>  workbox's templatedURLs never embeds the content. With a string value it just computes md5(value) and stores that as the precache revision. The actual /ui-extension/:id endpoint is fetched over the network at service-worker install time and cached then. So the client already "calls the URL on init and caches the data."

However, in classic fashion, Claude then pointed out that Couch already stores the hash and we would not need to retrieve at the attachments here if we just grabbed the digest.  Since it both guilted me into feeling bad about the extra file load and make the whole code change without asking me anyway, I let it update the code flow here to just grab the digest when building the service worker instead of the whole attachment.

tests/e2e/default/ui-extensions/ui-extensions-tab.wdio-spec.js
jkuester: 😅 I "fixed" it too hard.... 🤦  I originally added the comment to explain why it was 2. Then, I realized the cognitive overhead was unnecessary and changed so it was only loaded once, but forgot to remove the comment...

webapp/src/ts/services/header-tabs.service.ts
jkuester: Oh wow! Good catch!   😬😓

tests/e2e/default/service-worker/service-worker.wdio-spec.js
jkuester: Done and done! Let me know if there was other assertions/tests you were looking for.

webapp/tests/karma/ts/modules/ui-extensions/ui-extensions-route.guard.provider.spec.ts
jkuester: None of the UI Extension logic depends directly on "permissions". If you mean the role-based filtering, that is all handled in the service and is done automatically when calling `uiExtensionsService.getProperties`

api/src/generate-service-worker.js
```diff
@@ -70,6 +74,31 @@ const appendExtensionLibs = async (config) => {
...
+  const extensions = await uiExtensionService.getAllProperties();
+  // Include all extensions here because this endpoint is in globPatterns and so the results for the endpoint
+  // get cached even for online users. Online users call the actual endpoint and do not get the templatedURLs value
+  // set here, but the value returned from the endpoint is cached in the browser service worker anyway. So, the
+  // service worker hash must change even when an online-only ui-extension property changes.
+  config.templatedURLs['/ui-extension'] = JSON.stringify(extensions);
+
+  // Do not include the actual endpoints for online-only extensions. These should only be called by online users and
+  // do not need to be cached in the service worker at all.
+  const offlineExtensions = extensions.filter(ext => !ext.roles || roles.isOffline(ext.roles));
```
dianabarsan: I tried adding an extension without roles, logged in as an offline user and immediately went offline. Then navigated to the UI extension tab and observed that the browser tried to download the ui extension code on the spot, it was not cached in the service worker.

dianabarsan: image

dianabarsan: had a small typo in the original comment, saying that I'm using an online user. I am using an _offline_ user (as you can see, it has tasks).

jkuester: @dianabarsan can you share more of your repro steps?  I have been unable to reproduce this behavior myself... :thinking: 

Here is what it looks like for me trying when I try just toggling the network off as soon as the app loads:

[network.webm](https://github.com/user-attachments/assets/9e63f3f4-ab2d-47f2-9cd2-83597b1245e6)

Again, I also tried just setting a breakpoint [right after the bootstrapping completes](https://github.com/medic/cht-core/blob/c6242118cd2a847ad1302f2b065cc2beac54dba3/webapp/src/ts/main.ts#L137) and again toggling off the network.

[breakpoint.webm](https://github.com/user-attachments/assets/7151d6be-e77e-4e3d-a46e-15714a2d984d)

In both cases, the UI Extension was still already loaded...  I am just using the [with-resources](https://github.com/jkuester/cht-config-library/tree/main/ui-extensions) extension.

dianabarsan: Ok. I'll try again, i might have messed something up in the code i had running locally. I even added a browser.throttlenetwork('offline') call in the e2e test and the test started failing. maybe that would be just sufficient to show service worker cache for offline users.

dianabarsan: [Screencast_20260612_080831.webm](https://github.com/user-attachments/assets/f18aab4e-a5f2-4c70-a9ce-d40c19424d3c)

dianabarsan: I tried it again, and while offline my extension was not loaded.

dianabarsan: [Screencast_20260612_081506.webm](https://github.com/user-attachments/assets/d30497a7-b6c2-4a69-a54e-63b3295a5238)

another attempt.

dianabarsan: I made sure my local branch was up to date and everything was rebuilt. 

My extension is: 
```
{
  "extension_type": "sidebar_tab",
  "title": "My Extension",
  "resource_icon": "icon-healthcare",
  "accent_color": "#9BBE6F",
  "weight": 1,
  "config": {
    "icon": "icon-nurse",
    "accentColor": "#9BBE6F"
  }
}
```

dianabarsan: I pushed a branch that makes a small update to the e2e test, to set the browser as offline immediately after login in the ui extensions test, and the test is now failing: https://github.com/medic/cht-core/actions/runs/27396250692/job/80965905504#step:16:3396

https://github.com/medic/cht-core/compare/10224-ui-extensions...refs/heads/10224-ui-extensions-bis

jkuester: Okay, I have spent the entire morning coming at this from all angles.  :sweat: TLDR, I still have not been able to repro _with the properties you gave_ (tried a bunch of different combos of browsers, and everything I could think of to clear the SW caches, workers, etc from the browser).  

Are you sure you did not have anything configured for the `roles` property in the extension you are testing with?  I have found an actual bug in the code for building the SW cache but it only manifests when the `roles` property is not nullish (but it does show up if `roles: []`).  I am going to push a fix and hopefully you can verify if that fixes things for you. :crossed_fingers: 

Regarding the CI failure, it runs the `online` user tests first and I think the failure of those tests was polluting the `offline` tests because once I disabled the `online` tests, [the tests run in CI](https://github.com/medic/cht-core/actions/runs/27433087566/job/81091192576) even with your change to set the browser offline.

dianabarsan: Thanks for the changes! The empty roles extension now gets cached, however a _mixed_ role extension does not. 

I changed the roles from empty to one offline role and one online role, specifically:
```
  "extension_type": "sidebar_tab",
  "title": "api.startup.checks",
  "resource_icon": "icon-nurse",
  "accent_color": "#3F51B5",
  "type": "ui-extension",
  "roles": [
    "chw",
    "program_officer"
  ],
```

And I am seeing the same behavior where the extension is not cached in the service worker. I asked claude to snoop it down and it says: 

>  Offline-cache filter uses .every, but webapp visibility uses .some → mixed-role extensions break offline — api/src/generate-service-worker.js:76
  SW caches an extension only if (roles||[]).every(role => offlineRoles.includes(role)), but the webapp shows it if roles.some(role => hasRole(role)) (ui-extensions.service.ts:52). An extension with roles: ['chw'(offline), 'supervisor'(online)] is shown to an offline chw user but never 
  cached → the /ui-extension/:id fetch misses cache offline and the tab fails to load. The cache filter should be .some(role => offlineRoles.includes(role)) to match "any offline role can reach it."

api/src/routing.js
```diff
@@ -321,6 +322,8 @@ app.use(express.static(resources.staticPath));
 app.use(express.static(resources.webappPath));
 app.get('/extension-libs', extensionLibs.list);
 app.get('/extension-libs/:name', extensionLibs.get);
+app.get('/ui-extension', uiExtension.list);
+app.get('/ui-extension/:id', uiExtension.get);
```
dianabarsan: I want to point out that these two routes effectively make _everyone's_ extension code public. This contradicts how we handle configuration in general. 
I see that `extension-libs` have the same problem. 

Are we ok to make significant parts of partner configuration publicly available? 
I know that we use this for service worker cache.

webapp/src/ts/services/custom-resource.service.ts
```diff
@@ -42,33 +42,44 @@ export class ResourceIconsService {
...
-          content = `<img src="data:${icon.content_type};base64,${icon.data}" />`;
-        }
-        this.cache[i].htmlContent[name] = content;
-      }
-      return this.cache[i].htmlContent[name];
-    } catch (_) {
+  private formatIconContent(icon, docId) {
+    if (icon.content_type === 'image/svg+xml' && docId === DOC_IDS.RESOURCES) {
+      // SVG: include the raw data in the page so it can be styled
+      return atob(icon.data);
```
dianabarsan, line 55: This `atob` used to be wrapped in a try-catch, so malformed base64 attachments, that used to default to empty images, will now throw an error during render.

api/src/generate-service-worker.js
```diff
@@ -70,6 +72,35 @@ const appendExtensionLibs = async (config) => {
...
+  // service worker hash must change even when an online-only ui-extension property changes.
+  config.templatedURLs['/ui-extension'] = JSON.stringify(extensions);
+
+  // Do not include the actual endpoints for online-only extensions. These should only be called by online users and
+  // do not need to be cached in the service worker at all.
+  const offlineRoles = Object
+    .entries(settings.get('roles') ?? {})
+    .filter(([,{ offline }]) => offline)
+    .map(([key]) => key);
+  const offlineExtensions = extensions.filter(({ roles }) => (roles || []).every(role => offlineRoles.includes(role)));
```
dianabarsan: A malformed `roles` property ends up crashing API: 

```
2026-06-15T07:22:43.421 ERROR: Failed to generate service worker: TypeError: (roles || []).every is not a function
    at /home/diana/projects/cht-core-2/api/src/generate-service-worker.js:92:76
    at Array.filter (<anonymous>)
    at appendUiExtensions (/home/diana/projects/cht-core-2/api/src/generate-service-worker.js:92:40)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async writeServiceWorkerFile (/home/diana/projects/cht-core-2/api/src/generate-service-worker.js:147:3)
    at async Object.run (/home/diana/projects/cht-core-2/api/src/generate-service-worker.js:192:5) {
  [stack]: [Getter/Setter],
  [message]: '(roles || []).every is not a function'
} 
```

I set roles property to a `string`.

dianabarsan (commented):
Thanks for all the changes! 
I did a quick go-over now and left some comments.

api/src/generate-service-worker.js
jkuester: 😓😭:skull: Welp, that is my bad.  :facepalm: I _specifically_ intended to not use the `every` there because I was trying to achieve the behavior where offline-ness is greedy and an extension is offline if any of the roles is offline....  I would blame this on Claude, but TBH I wrote that code....

api/src/routing.js
jkuester: :disappointed: The longer we mess with this, the more I wonder if we should just be following the forms example instead of the extension-libs... (Maybe extension-libs should also follow the forms...)  Why even mess with the service worker at all instead of just replicating the docs and their attachments???  :shrug: 

> Are we ok to make significant parts of partner configuration publicly available?

On one hand, we are just helping folks comply with the AGPL... :sweat_smile: On the other hand, it is super not cool that there is no convenient way for partners to at least include private values here. (e.g. if the extension _properties_ were private than at least they could extract more sensitive config into the properties, but currently even the properties are public...  

IDK, the more I think about this the less comfortable I am with things as they are. I don't think it is unreasonable to store application code publicly on the service worker, but it does feel really _unexpected_ to do this in the context of CHT config code.  We could probably ship as is and come back for this later, but honestly I think I will just try and get this flipped now and see how much of a pain it will be....

webapp/src/ts/services/custom-resource.service.ts
jkuester: Woah, good catch!  Not sure how that try-block got dropped...  Added it back in with accompanying unit tests.

api/src/routing.js
jkuester: Welp, after looking at this for ~30sec, I figured out why we probably did not just replicate the extension-libs to begin with....  To do that, I guess we would need to update the `docs_by_replication_key` index (triggering it to rebuild...).  

The good news is that we already have a [rebuild of this index](https://github.com/medic/cht-core/commit/e25dc65ceb4cc5cd79797270566d5d5c77405afe) shipping in 5.2.0, so we should be good to go.

jkuester: In worse news, switching to using doc replication for sending the UI Extensions to the client breaks my hopes for limiting which Extensions have to be downloaded by certain users.  With the SW we could get away with offline users only downloading the extensions that were available offline.  With the doc-based replication I guess we will have to send all the extensions to all users... I was hoping to get away slightly updating our replication algorithm to do role-based replication.  However, this dose not work because doc replication is "sticky" and once a doc is sent to a user, you cannot later update the doc so it should not be replicated.  (If you do, the same challenges around [moving contacts](https://docs.communityhealthtoolkit.org/building/contact-management/moving-contacts/) apply.  You have to basically clear user's local data and re-sync everything for them to drop the old version of the extension...)  :cry:

api/src/generate-service-worker.js
jkuester: FTR, I updated the logic both here and in webapp to "fail closed" instead of failing open.  Basically we only treat the extension as available to everyone if the `roles` property is null/undefined/[]. If you set it to something random like a string, it is just not available to anyone....

api/src/routing.js
jkuester: Okay, @dianabarsan, after staring at this for `n` more hours, I reached the state of enlightenment whereby I asked myself (aka Claude) why I could not have authenticated endpoints in the service worker cache.  Claude could not give me any good reason why endpoints needed to be unauthenticated to work with the SW.  

So, I have tried just adding authentication.... :shrug: it seems to work fine running for me locally (testing the offline scenario), though I am still testing more cases.  

I pushed the changes, though, and we will see if CI builds....

jkuester: Aaannnd... Never mind to that.  :sob:   The initial bootstrap/logic flow works fine with the endpoints authenticated. But, it turns out the service worker does not die when a user logs out. Instead it lives on with its cache on the login page waiting for the next user to login.  When a new user logs in, it does not have to re-download the content of the service worker cache because it already exists.  

Where things go bad for our authenticated endpoints is that apparently the service worker can get refreshed either just on the login page or at some point during the subsequent login of a new user (but before the session cookie is set).  The end result is that it can try (and fail) to hit the cached endpoints. I have not been able to repro this running locally for some reason, just with the dev server, but it does consistently fail the e2e tests.  (The tests fail because of the feedback doc for the failed hit.)

dianabarsan: I know. We register the service worker on the login page (before login) and we never unregister it. 


For authenticated service worker stuff to work properly, we need a separate SW that we register post login and unregister on logout, so not persistent per user if we want to filter out cached resources based on roles. 

I think it's a nice idea, but I think it would take significant more work than just propping ui-extensions as downloadable documents. 
One caveat to downloadable documents is that we need to take care how large the code attachment becomes.

jkuester:
Okay @dianabarsan, this is ready for you to have another look!  :+1: 

I have updated both UI Extensions and Extension Libs to just replicate via docs instead of on the service worker.  Unfortunately, this means we send all UI Extensions to all users (regardless of if the user has access or not).  But, I think we are pretty much stuck with this until we figure out the logic for "unreplicating" docs when a user loses access (I am sure we will get to that any day... :sweat_smile: ).

tests/integration/couchdb/views/docs-by-replication-key.spec.js
```diff
@@ -89,6 +97,15 @@ describe('docs_by_replication_key', () => {
         _deleted: true,
       },
     },
+    {
```
dianabarsan: I don't think this is needed as we don't generate tombstones anymore.

ddocs/medic-db/medic/validate_doc_update.js
```diff
@@ -9,7 +9,7 @@
 */
 function(newDoc, oldDoc, userCtx, secObj) {
 
-  var ADMIN_ONLY_TYPES = [ 'form', 'translations', 'token_login' ];
+  var ADMIN_ONLY_TYPES = [ 'form', 'translations', 'token_login', 'ui-extension' ];
```
dianabarsan, line 12: I think that now that both extension types are docs, we need to update this list here: https://github.com/medic/cht-core/blob/f13b849a884e4dc442f1097c240d37f8b7debcca/webapp/src/ts/services/db-sync.service.ts#L19

webapp/tests/karma/ts/services/ui-extensions.service.spec.ts
```diff
@@ -0,0 +1,294 @@
...
+
+      expect(allDocs.callCount).to.equal(1);
+      expect(result).to.deep.equal([]);
+    });
+
+    it('should ignore docs in the range that are not ui-extension docs', async () => {
+      mockExtensions(
+        [{ id: 'ext-1', extension_type: 'tab' }],
+        [{ doc: { _id: 'ui-extension:not-an-extension', type: 'something-else', extension_type: 'tab' } }],
+      );
```
dianabarsan, line 92: Do we need the `doc` subproperty here?

tests/e2e/default/ui-extensions/ui-extensions-tab.wdio-spec.js
```diff
@@ -0,0 +1,158 @@
...
+    place: districtHospitalId,
+    roles: ['program_officer'],
+    contact: supervisorPerson._id
+  });
+
+  before(async () => {
+    await utils.saveDocs([
+      buildExtensionDoc(EXTENSIONS_DIR, 'header-tab'),
+      buildExtensionDoc(EXTENSIONS_DIR, 'sidebar-tab'),
+      buildExtensionDoc(EXTENSIONS_DIR, 'error'),
```
dianabarsan, line 40: can you also include the `online-only` extension here, or have an e2e test that shows that roles filter out displayable extensions to users?

dianabarsan (approved):
Thank you for all the work put here, especially the last minute fixes.
Such an awesome addition to the CHT! ⭐ 

I left some requests inline, all small. 
Approving to unblock.

webapp/tests/karma/ts/services/ui-extensions.service.spec.ts
jkuester: Refactored to also map these extra docs to the `doc` so the two params here can be the same shape.

PR #11052, opened 2026-05-12
fix: allow case-sensitive login for CouchDB admins

### LINKED ISSUE #7142: Initial user creation fails when CouchDB admin username contains uppercase characters
**Describe the bug**
When starting the services for the first time with a fresh (empty) CouchDB instance (particularly as described in the [DEVELOPMENT documentation](https://github.com/medic/cht-core/blob/master/DEVELOPMENT.md)), if the username for the initial admin user for CouchDB contains uppercase characters, that user will not be able to login to Medic.  All of the servers will start successfully, but the login for the existing admin user will fail.

**To Reproduce**
1.Follow the [DEVELOPMENT steps](https://github.com/medic/cht-core/blob/master/DEVELOPMENT.md) for starting the framework, except, instead of using the recommended `myadminuser` as the CouchDB admin username, use something with uppercase characters like `MyAdminUser`.
2. Once you have started the webapp and the API, navigate to `http://localhost:5988/medic/login` and try to login with your CouchDB admin username/password
3. The login will fail with a 401

**Expected behavior**
Under normal circumstances, [Medic does not allow](https://github.com/medic/cht-core/blob/6e82151d120e57f2332d6cc1ea78da7a5bae69f9/api/src/controllers/create-user-db.js#L8) for creating users with usernames that contain banned characters (such as uppercase). Any old Medic users that happened to have invalid usernames should be corrected automatically by the [extract-user-settings](https://github.com/medic/cht-core/blob/master/api/src/migrations/extract-user-settings.js) migration. So, normal users added via the API should not encounter this problem. It is just existing users from CouchDB. 

It seems like we should try to find a more graceful way to handle this situation. Perhaps if the lower case version of the username fails, we try the case as entered or if we want to still fail the login, at least fail with a more helpful message...

**Environment**
- Instance: local
- Browser: Chrome
- Client platform: Linux
- Version: 3.11.0 (master)

sh1vam31:
# Description

### Fixes: #7142

This PR allows users to log in with mixed-case usernames. Previously, the login page unconditionally converted all usernames to lowercase on the client side, which prevented CouchDB admin accounts with uppercase characters from authenticating.

### Changes
**Client-side:** Removed the `.toLowerCase()` call in `api/src/public/login/script.js` to ensure the exact casing entered by the user is sent to the server.

**Server-side**: Updated `validateSession` in `api/src/controllers/login.js` to attempt authentication with the provided username first. If that fails with a 401, it retries with a lowercased version to maintain backward compatibility for standard Medic users.

**Testing:** Added new unit tests to `api/tests/mocha/controllers/login.spec.js` covering both the successful mixed-case login and the fallback-to-lowercase scenario.

# Code review checklist

# License


sh1vam31:
Hi @jkuester

I've verified the fix with new unit tests covering the following:

Successful login for mixed-case usernames.
Backward compatibility via a lowercase fallback.

Ready for review!

sh1vam31:
Hi @jkuester

I've verified the fix with new unit tests covering the following:

Successful login for mixed-case usernames.
Backward compatibility via a lowercase fallback.

Ready for review!

Issue #10137, opened 2025-07-07
Update Webapp ongoing replication to downloads forms separately


dianabarsan:
**Describe the issue**
Forms might have very large attachments, which makes it risky to download in bulk, as the rest of the documents are downloaded during ongoing replication.

**Describe the improvement you'd like**
Use a different endpoint to download forms or attachments individually during onfoing replication: https://github.com/medic/cht-core/blob/167edb35b37c2cf87cdbba2894bdbdc5c3d4bd78/webapp/src/ts/services/replication.service.ts#L74

**Describe alternatives you've considered**
Leave it as it is.

shashi-sah2003:
hey @dianabarsan do you want to change this for whole batch or just for the batch with a big attachements? If only for big attachments then what would be the threshold for it?

dianabarsan:
I think what I was envisioning is that we download forms separately. So an addition to the algorithm so that we filter out the form ids and download them through a different API.

shashi-sah2003:
ahh got it, regardless of the size we want all the forms to be downloaded via different endpoint, make sense. I can work on this one if you haven't already started.

dianabarsan:
@shashi-sah2003 this work is part of the forms with large attachments squad, and it depends on other work to be completed first. 
If you wish to be part of the squad, the calls are open for everyone and you can find them on the calendar. As it happens, the weekly call is happening today.

shashi-sah2003:
@dianabarsan that would be great to be part of another squad. sorry I missed this week call, will be joining the upcoming call

Abhishek9639:
Hi @dianabarsan or @jkuester,
I’d like to work on this issue. The approach of filtering form IDs and downloading large forms separately through a different API makes sense to me. Please assign this issue to me if it’s available. 
Thanks

andrablaj:
Hi @Abhishek9639, and thank you for your interest in contributing. I see the issue https://github.com/medic/cht-core/issues/10373 is already assigned to you. Are you planning to work on it? To make sure everyone has a fair chance to contribute, we try to keep work-in-progress items per contributor to a minimum. Would you mind unassigning yourself from any issues you're not actively working on before picking up this one? Really appreciate it — thanks!

Abhishek9639:
Hi @andrablaj,
I have completed my part of the work on all the issues currently assigned to me. There is no issue assigned to me that I am not actively working on. At the moment, I’m mainly waiting for mentor feedback and replies on those issues.
Thanks

[...]

PR #11055, opened 2026-05-13
feat(#10137): download form docs separately during ongoing replication

Abhishek9639:
# Description
Forms can have very large attachments, which makes it risky to download them in bulk along with other documents during ongoing replication. A single oversized form attachment can break the entire batch.

This PR updates `getMissingDocs` in `replication.service.ts` to split the download list into form docs and non-form docs:
- **Non-form docs** continue using the existing `bulkGet` batched path (no change).
- **Form docs** (identified by the `form:` prefix) are now downloaded individually using `dbService.get({ remote: true }).get()`, then saved via the same `bulkDocs({ new_edits: false })` + `monitorExternalChanges` flow.

If a form doc fails to download, the error propagates and fails the replication round since forms are essential.

#10137

# Code review checklist
- [x] Backwards compatible. Test it works for the new design (enabled by default). And test it works in the old design.
- [x] Tested. Unit and/or e2e where appropriate.
- [x] Backwards compatible. Works with existing data and configuration or includes a migration. Any breaking changes documented in the release notes.

# License

### DIFF
--- a/webapp/src/ts/services/replication.service.ts
+++ b/webapp/src/ts/services/replication.service.ts
@@ -16,6 +16,7 @@ export class ReplicationService {
   }
 
   private readonly BATCH_SIZE=100;
+  private readonly FORM_DOC_ID_PREFIX='form:';
 
   async replicateFrom():Promise<{ read_docs: number }> {
     const remoteDocIdsRevs = await this.getRemoteDocs();
@@ -43,9 +44,16 @@ export class ReplicationService {
       .filter(({ id, rev }) => !localIdRevMap[id] || localIdRevMap[id] !== rev);
     const nbrDocs = docIdRevsToDownload.length;
 
-    while (docIdRevsToDownload.length) {
-      const batch = docIdRevsToDownload.splice(0, this.BATCH_SIZE);
+    const formDocIdRevs = docIdRevsToDownload.filter(({ id }) => id.startsWith(this.FORM_DOC_ID_PREFIX));
+    const nonFormDocIdRevs = docIdRevsToDownload.filter(({ id }) => !id.startsWith(this.FORM_DOC_ID_PREFIX));
+
+    while (nonFormDocIdRevs.length) {
+      const batch = nonFormDocIdRevs.splice(0, this.BATCH_SIZE);
       await this.downloadDocsBatch(batch);
+    }
+
+    for (const idRev of formDocIdRevs) {
+      await this.downloadFormDoc(idRev);
     }
 
     return nbrDocs;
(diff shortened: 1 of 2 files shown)


Abhishek9639:
Hi @dianabarsan or @andrablaj,

This PR is ready for review. I've implemented the approach we discussed form docs are now downloaded individually instead of being batched with bulkGet, so large form attachments won't break the entire batch.

Let me know if anything needs to be changed. 
Thanks

dianabarsan:
Sorry @Abhishek9639 but we are not ready to get this merged, as is part of larger work and it does not make sense to have this without the other work.

dianabarsan (changes requested):
Hi @Abhishek9639 

It seems that your CI is failing, can you please check the errors?
Also, can you please change the target of this PR to https://github.com/medic/cht-core/tree/10131-support-forms-with-large-attachments ? We will push all the necessary code there, and have only one  PR against `master` when we're done.

webapp/src/ts/services/replication.service.ts
```diff
@@ -16,6 +16,7 @@ export class ReplicationService {
   }
 
   private readonly BATCH_SIZE=100;
+  private readonly FORM_DOC_ID_PREFIX='form:';
```
dianabarsan, line 19: This is a medic/constants PREFIX. Please use that instead.

webapp/src/ts/services/replication.service.ts
```diff
@@ -43,11 +44,18 @@ export class ReplicationService {
...
-      const batch = docIdRevsToDownload.splice(0, this.BATCH_SIZE);
+    const formDocIdRevs = docIdRevsToDownload.filter(({ id }) => id.startsWith(this.FORM_DOC_ID_PREFIX));
+    const nonFormDocIdRevs = docIdRevsToDownload.filter(({ id }) => !id.startsWith(this.FORM_DOC_ID_PREFIX));
+
+    while (nonFormDocIdRevs.length) {
+      const batch = nonFormDocIdRevs.splice(0, this.BATCH_SIZE);
       await this.downloadDocsBatch(batch);
     }
 
+    for (const idRev of formDocIdRevs) {
```
dianabarsan, line 55: Please start with downloading the form docs, and then the other docs.

webapp/src/ts/services/replication.service.ts
```diff
@@ -79,4 +87,10 @@ export class ReplicationService {
     await this.dbService.get().bulkDocs(docs, { new_edits: false });
     this.rulesEngineService.monitorExternalChanges({ docs });
   }
+
+  private async downloadFormDoc({ id, rev }):Promise<void> {
+    const doc = await this.dbService.get({ remote: true }).get(id, { attachments: true, rev, revs: true });
+    await this.dbService.get().bulkDocs([doc], { new_edits: false });
+    this.rulesEngineService.monitorExternalChanges({ docs: [doc] });
```
dianabarsan, line 94: This is not necessary. `rulesEngineService` only cares about task documents, not forms.

dianabarsan (changes requested):

PR #11057, opened 2026-05-13
feat(#10908): support ordering header tabs by weight

### LINKED ISSUE #10908: Support order weight for header_tabs in webapp

jkuester:
# Description

Closes #10908 

This PR adds support for ordering the main interface headers tabs according to the `weight` property configured in the [header_tabs app settings](https://docs.communityhealthtoolkit.org/building/reference/app-settings/header_tabs/).  Additionally, the `header_tab` UI Extensions can also have a `weight` property that is used when factoring in the ordering of all the tabs. This ordering is honored both in the new and in the old UI.

There was considerable duplication and complexity around the sidebar menu options between the new and the old UI.  I have refactored the handling of these options to follow similar patterns with what we have for the main header tabs.

Documentation for the new config will be added in https://github.com/medic/cht-docs/pull/2213

## AI disclosure

Most of the unit test updates were made by Claude. I have reviewed all changes.

# Code review checklist

# License

### DIFF
--- a/webapp/src/ts/components/header/header.component.ts
+++ b/webapp/src/ts/components/header/header.component.ts
@@ -18,8 +18,7 @@ import { RelativeDatePipe } from '@mm-pipes/date.pipe';
 import { LocalizeNumberPipe } from '@mm-pipes/number.pipe';
 import { HeaderLogoPipe, ResourceIconPipe } from '@mm-pipes/resource-icon.pipe';
 
-import { HeaderTab, HeaderTabsService } from '@mm-services/header-tabs.service';
-import { UiExtensionsService } from '@mm-services/ui-extensions.service';
+import { HeaderTab, HeaderTabsService, SidebarTab } from '@mm-services/header-tabs.service';
 
 export const OLD_NAV_PERMISSION = 'can_view_old_navigation';
 
@@ -38,7 +37,7 @@ export const OLD_NAV_PERMISSION = 'can_view_old_navigation';
     HeaderLogoPipe,
     ResourceIconPipe,
     RelativeDatePipe,
-    LocalizeNumberPipe
+    LocalizeNumberPipe,
   ]
 })
 
@@ -51,7 +50,7 @@ export class HeaderComponent extends BaseMenuComponent implements OnInit, OnDest
   currentTab;
   bubbleCount = {};
   permittedTabs: HeaderTab[] = [];
-  uiExtensionOptions: UiExtMenuOption[] = [];
+  legacyMenuTabs: SidebarTab[] = [];
 
   constructor(
     protected readonly store: Store,
@@ -59,7 +58,6 @@ export class HeaderComponent extends BaseMenuComponent implements OnInit, OnDest
     protected readonly modalService: ModalService,
     protected readonly storageInfoService: StorageInfoService,
     private headerTabsService: HeaderTabsService,
-    private readonly uiExtensionsService: UiExtensionsService,
   ) {
     super(store, dbSyncService, modalService, storageInfoService);
   }
@@ -68,11 +66,20 @@ export class HeaderComponent extends BaseMenuComponent implements OnInit, OnDest
     super.ngOnInit();
     this.additionalSubscriptions();
     this.getHeaderTabs();
-    this.loadUiExtensionOptions();
+    this.headerTabsService
+      .getSidebarTabs()
+      .then(tabs => this.legacyMenuTabs = tabs);
   }
 
   ngOnDestroy() {
     super.ngOnDestroy();
+  }
+
+  onLegacyMenuClick(tab: SidebarTab) {
+    if (tab.name !== 'bug') {
+      return;
+    }
+    this.openFeedback();
   }
 
   private additionalSubscriptions(){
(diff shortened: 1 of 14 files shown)


jkuester:
@dianabarsan can you have a look at this when you get the chance?  The refactoring here got pretty involved, so I wanted to give you a chance to see it in isolation before it goes into the main UI Extensions branch. :+1:

jkuester:
The ordering/icons/functionality for all the tabs both in the header and the sidebar for both the new and old UIs should all work as expected.   However, note that there are a few known styling issues with the UI Extensions in the legacy UI that are being addressed in https://github.com/medic/cht-core/issues/11017.  So, if you notice any of those, just be aware that it is a known issue.

webapp/src/ts/services/header-tabs.service.ts
```diff
@@ -14,87 +14,155 @@ export class HeaderTabsService {
...
 
+  async getSidebarTabs(): Promise<SidebarTab[]> {
+    if (!this.sidebarTabs) {
+      const [headerTabs, uiExtensionTabs] = await Promise.all([
+        this.getHeaderTabs(this.SIDEBAR_HEADER_TABS),
+        this.getUiExtensionTabs('sidebar_tab')
+      ]);
+      headerTabs.sort((a, b) => a.weight - b.weight);
+      // sidebar_tab extensions should always come after the headers
+      this.sidebarTabs = [
```
dianabarsan, line 156: Could you please add a test pinning this behavior: extension with low weight still lands after the default sidebar header tabs?

webapp/src/ts/services/header-tabs.service.ts
```diff
@@ -14,87 +14,155 @@ export class HeaderTabsService {
...
-      resourceIcon: undefined,
-    }
+      name: 'bug',
+      defaultIcon: 'fa-bug',
+      translation: 'Report Bug',
+      permissions: [],
+    },
   ];
 
+  private readonly DEFAULT_UI_EXTENSION_WEIGHT = 6;
```
dianabarsan:
```suggestion
  private readonly DEFAULT_UI_EXTENSION_WEIGHT = Math.max(...defaultWeights) + 1);
```

If a sixth default tab is added with weight 6, UI extensions start tying with it silently. The change also makes it clear why the specific weight is meaningful.

webapp/src/ts/components/header/header.component.ts
```diff
@@ -68,13 +66,22 @@ export class HeaderComponent extends BaseMenuComponent implements OnInit, OnDest
...
 
   ngOnDestroy() {
     super.ngOnDestroy();
   }
 
+  onLegacyMenuClick(tab: SidebarTab) {
+    if (tab.name !== 'bug') {
+      return;
+    }
+    this.openFeedback();
```
dianabarsan: Just for consistency, let's mirror the `onTabClick` syntax:
```suggestion
   if (tab.name === 'bug') { 
     this.openFeedback();
   }
```

webapp/src/ts/components/header/header.component.ts
```diff
@@ -51,15 +50,14 @@ export class HeaderComponent extends BaseMenuComponent implements OnInit, OnDest
   currentTab;
   bubbleCount = {};
   permittedTabs: HeaderTab[] = [];
-  uiExtensionOptions: UiExtMenuOption[] = [];
+  legacyMenuTabs: SidebarTab[] = [];
```
dianabarsan: `legacyMenuTabs vs headerTabsForSidebar`

When I initially saw this name, I got quite confused (why are they legacy? and not just MenuTabs). The same tabs are called `headerTabsForSidebar` in the `SidebarMenuComponent`. 
To avoid confusion, can we name this shared-source in a consistent way?

dianabarsan (approved):
Sweet cleanup here! I pleasure to read really! I added some nits inline.

webapp/src/ts/components/header/header.component.ts
jkuester: :+1: My original goal with the naming was to make it clear these were for the legacy drop-down menu from Ye Olde UIe.  However, I do agree that it is a bit confusing given the code also in the sidebar-menu.component.  I have updated this variable to be `headerTabsForLegacySidebar` which hopefully achieves both objectives!

PR #11059, opened 2026-05-13
fix(#10955): replace unsafe eval() with strict comparison for boolean parsing in CSV user import

### LINKED ISSUE #10955: Replace unsafe eval() with strict comparison for boolean parsing in CSV user import
## Description

In `shared-libs/user-management/src/users.js`, the `assignCsvCellValue` function uses `eval()` to convert the strings `"TRUE"` and `"FALSE"` to their boolean equivalents:

```javascript
if (['TRUE', 'FALSE'].includes(data[attribute])) {
  // converts the "TRUE" or "FALSE" string to boolean
  data[attribute] = eval(data[attribute].toLowerCase());
}
```

While the `if` guard restricts execution to only `"TRUE"` or `"FALSE"` values, using `eval()` for this purpose is:

1. **A security code smell** — `eval()` is flagged by most static analysis tools (ESLint `no-eval` rule) and security audits. Even with the guard, it creates a pattern that could be dangerous if the guard is modified or bypassed in future changes.
2. **Unnecessary** — A simple strict comparison achieves the same result without any code execution risk.
3. **Less readable** — Developers reading this code need to mentally trace through the eval to understand it converts strings to booleans.

## Suggested Fix

Replace the `eval()` call with a direct comparison:

```javascript
if (['TRUE', 'FALSE'].includes(data[attribute])) {
  data[attribute] = data[attribute] === 'TRUE';
}
```

## Impact

- Eliminates an `eval()` usage that can be flagged by security scanners
- No functional change — existing unit tests confirm `TRUE` → `true` and `FALSE` → `false`
- Improves code clarity and maintainability

shivv23:
## Summary

Replaces `eval()` with a safe strict equality comparison for boolean parsing in the CSV user import, and adds comprehensive unit tests for the `assignCsvCellValue` function.

### The bug

The `assignCsvCellValue` function in `shared-libs/user-management/src/users.js` used:

```js
data[attribute] = eval(data[attribute].toLowerCase());
```

This converts `"TRUE"` and `"FALSE"` strings from CSV cells into booleans. While guarded by `['TRUE', 'FALSE'].includes(data[attribute])`, `eval()` is still a security anti-pattern — it indicates a code smell and could introduce vulnerabilities if the guard logic ever changes.

### The fix

```js
data[attribute] = data[attribute] === 'TRUE';
```

Since the guard already ensures the value is exactly `'TRUE'` or `'FALSE'`, the strict equality check is functionally identical and eliminates the code smell entirely.

### Unit tests added (9 tests via rewire)

The test file uses `rewire` (already loaded at the top of the spec). Added a full `describe('assignCsvCellValue')` block with 9 direct unit tests:

| Test | What it verifies |
|------|-----------------|
| TRUE → boolean true | Core fix correctness |
| FALSE → boolean false | Core fix correctness |
| Non-boolean strings unchanged | Unrelated values pass through |
| Non-string values unchanged | Numbers, objects pass through |
| Whitespace in values trimmed | CSV cell cleanup |
| Surrounding quotes stripped | CSV quoted value handling |
| Attribute names trimmed | Edge case robustness |
| Excluded attributes skipped | `:excluded` suffix logic |
| Empty attribute names skipped | Edge case robustness |

### Why this matters for C4GT

This is a textbook example of defensive coding: removing `eval()` eliminates a latent RCE vector, and the 9 accompanying tests ensure the replacement is correct across all edge cases.

### Testing

```
cd shared-libs/user-management
npm test -- --grep 'assignCsvCellValue'
```

Closes #10955

### DIFF
--- a/shared-libs/user-management/src/users.js
+++ b/shared-libs/user-management/src/users.js
@@ -747,7 +747,7 @@ const assignCsvCellValue = (data, attribute, value) => {
 
   if (['TRUE', 'FALSE'].includes(data[attribute])) {
     // converts the "TRUE" or "FALSE" string to boolean
-    data[attribute] = eval(data[attribute].toLowerCase());
+    data[attribute] = data[attribute] === 'TRUE';
   }
 };
 
--- a/shared-libs/user-management/test/unit/users.spec.js
+++ b/shared-libs/user-management/test/unit/users.spec.js
@@ -4806,6 +4806,71 @@ describe('Users service', () => {
         chai.expect(validateSsoLoginUpdate.notCalled).to.be.true;
       });
     });
+  describe('assignCsvCellValue', () => {
+    it('should convert TRUE string to boolean true', () => {
+      const assignCsvCellValue = service.__get__('assignCsvCellValue');
+      const data = {};
+      assignCsvCellValue(data, 'token_login', 'TRUE');
+      chai.expect(data.token_login).to.equal(true);
+    });
+
+    it('should convert FALSE string to boolean false', () => {
+      const assignCsvCellValue = service.__get__('assignCsvCellValue');
+      const data = {};
+      assignCsvCellValue(data, 'token_login', 'FALSE');
+      chai.expect(data.token_login).to.equal(false);
+    });
+
+    it('should keep non-boolean strings unchanged', () => {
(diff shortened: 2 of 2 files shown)


andrablaj:
Hi @shivv23  , thanks for opening the PR. 

PR #11060, opened 2026-05-13
chore(#11048): modernize version parsing regex with named capture groups

### LINKED ISSUE #11048: Modernize version parsing regex using named capture groups
## Description

While going through the admin app codebase, I noticed there’s an existing TODO in `admin/src/js/services/version.js` about replacing the current version parsing regex once support for older Node versions is dropped.

Since the project is now running on Node 22, I think we can safely modernize this logic using named capture groups.

Right now the regex relies on numbered array indexes like:

```
versionMatch[1]
versionMatch[2]
```

which works, but it becomes a bit difficult to read and easier to break when the regex changes in the future.

There’s already a commented example in the file showing the intended named-group version:

```
/^(?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)(?<featureRelease>-FR(?:-\w+)+)?(?:-beta\.(?<beta>\d+))?$/
```

I think it would be useful to:

* replace the current regex with the named capture group version
* update the parsing logic to use versionMatch.groups.major, minor, patch, etc.
* keep the existing parsing behavior unchanged

This would make the parsing logic much easier to understand and reduce the chances of indexing mistakes later if the version format evolves further.

Acceptance criteria could be:

* update the regex to use named capture groups
* migrate assignments away from numeric indexes
* ensure existing version parsing tests continue to pass

@andrablaj please assign this to me. I am open for feedback and discussion.

shivv23:
## Summary

Refactors the version parsing regex in `admin/src/js/services/version.js` to use ES2018 named capture groups, and replaces global `parseInt` with `Number.parseInt` to satisfy SonarCloud rule S7773.

### The technical debt

The `versionInformation` function contained this TODO comment:

```js
// TODO: replace this regex with named capture groups once we deprecate node 8
```

Node 8 has been EOL since December 2019. The codebase already targets Node 18+. This PR finally cleans up that decade-old debt.

### Changes

**1. Named capture groups** — replaced numbered positional groups with descriptive names:

| Before | After |
|--------|-------|
| `(\\d+)` | `(?<major>\\d+)` |
| `versionMatch[1]` | `versionMatch.groups.major` |

This makes the regex self-documenting: you can read `groups.major`, `groups.minor`, `groups.patch`, `groups.beta`, `groups.featureRelease`, `groups.build` at the call site instead of guessing what `versionMatch[4]` means.

**2. Added `build` capture group** — the old regex had `(\\.(\\d+))?` (group 7) for the trailing build number like `4.15.1.12345`, but it was never extracted into the version object. The new regex properly names it `(?<build>\\d+)`.

**3. SonarCloud compliance** — replaced all 4 `parseInt` calls with `Number.parseInt` to fix quality gate failure (rule S7773).

### Impact

Zero behavioral change. All existing unit tests pass. The fix is purely about code maintainability and readability.

### Testing

```
cd admin
npm test -- --grep 'Version'
```

Closes #11048

### DIFF
--- a/admin/src/js/services/version.js
+++ b/admin/src/js/services/version.js
@@ -21,13 +21,13 @@ angular.module('services').factory('Version',
 
       if (versionMatch) {
         const version = {
-          major: parseInt(versionMatch.groups.major),
-          minor: parseInt(versionMatch.groups.minor),
-          patch: parseInt(versionMatch.groups.patch)
+          major: Number.parseInt(versionMatch.groups.major),
+          minor: Number.parseInt(versionMatch.groups.minor),
+          patch: Number.parseInt(versionMatch.groups.patch)
         };
 
         if (versionMatch.groups.beta !== undefined) {
-          version.beta = parseInt(versionMatch.groups.beta);
+          version.beta = Number.parseInt(versionMatch.groups.beta);
         }
 
         if (versionMatch.groups.featureRelease !== undefined) {
--- a/admin/src/js/services/version.js
+++ b/admin/src/js/services/version.js
@@ -16,24 +16,22 @@ angular.module('services').factory('Version',
     };
 
     const versionInformation = function(versionString) {
-      // TODO: replace this regex with named capture groups once we deprecate node 8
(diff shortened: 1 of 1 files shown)


andrablaj:
Hi @shivv23  , thanks for opening the PR. 

PR #11062, opened 2026-05-13
chore: add missing unit tests for isNotNull and lineage edge cases in cht-datasource

shivv23:
## Summary

Adds missing unit test coverage for the `isNotNull` type guard and edge cases for lineage utility functions in the `@medic/cht-datasource` shared library.

## Details

Based on a test coverage audit of the cht-datasource library, the following gaps were identified and filled:

### 1. `isNotNull` type guard — 8 test cases

`shared-libs/cht-datasource/src/libs/core.ts`

Had **zero** direct unit tests despite being an exported type guard used across the library (`filter(isNotNull)` appears in multiple lineage functions). Added 8 parameterized cases:

| Input | Expected |
|-------|----------|
| `null` | `false` |
| `undefined` | `false` (type guard passes, no narrowing) |
| `'value'` | `true` |
| `0` | `true` |
| `''` | `true` |
| `false` | `true` |
| `{}` | `true` |
| `[]` | `true` |

### 2. `getLineageDocsById` edge cases — 3 tests

Had only 1 happy-path test. Added:
- Document not found → empty array returned
- Database query error → error propagated
- Null docs in query results → preserved in response

### 3. `minifyLineage` edge cases — 3 tests

Added tests for:
- Deeply nested parent chain (recursive minification verified)
- Extra field stripping (`parent.name`, `type` removed, only `_id` + `parent._id` kept)
- Doc with no parent field

### 4. `minifyDoc` edge cases — 3 tests

Added tests for:
- Extra fields stripped from full doc
- Doc with no parent field
- Original document immutability (minifyDoc must not mutate the input)

All tests follow the existing patterns: chai expectations, sinon stubs, and async/await where needed.

### Testing

```
cd shared-libs/cht-datasource
npm test
```

### DIFF
--- a/shared-libs/cht-datasource/test/libs/core.spec.ts
+++ b/shared-libs/cht-datasource/test/libs/core.spec.ts
@@ -17,6 +17,7 @@ import {
   isIdentifiable,
   isNonEmptyArray,
   isNormalizedParent,
+  isNotNull,
   isRecord,
   isString,
   NonEmptyArray,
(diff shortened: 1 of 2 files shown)


andrablaj:
Hi @YASHSHARMAOFFICIALLY , thanks for opening the PR. 

The PR doesn't seem to be related to an issue. We're closing it for now. In the future, please open an issue if you identify a code change that needs to be made. Thanks!

Issue #11051, opened 2026-05-12
authorization.js middleware silently swallows getUserSettings errors


harkirath1511:
**Describe the bug**
The internal `getUserSettings` helper in [api/src/middleware/authorization.js](https://github.com/medic/cht-core/blob/master/api/src/middleware/authorization.js#L11-L18) catches errors with `.catch(err => err)`, which silently drops the error with no log line:

```js
const getUserSettings = (req) => {
  return auth
    .getUserSettings(req.userCtx)
    .then(userCtx => {
      req.userCtx = userCtx;
    })
    .catch(err => err);
};
```
If the lookup against the _users / settings doc fails for any reason : network blip to CouchDB, a malformed user-settings doc, a permissions error , the promise resolves successfully (with the Error object as its value, which is then discarded by the .then(next) call sites). The request continues with req.userCtx left at whatever getUserCtx set earlier, and no log line is emitted, making the failure invisible in production.

Compare this with the sibling getUserCtx function in the same file (lines 22–33), which catches into req.authErr so the downstream handleAuthErrors middleware can respond. getUserSettings has no equivalent — failures are simply dropped.

**To Reproduce**
1. Force `auth.getUserSettings()` to reject (e.g. stop CouchDB, or temporarily corrupt the user-settings doc for an offline user).
2. Make any request that goes through `onlineUserProxy`, `onlineUserPassThrough`, or the exported `getUserSettings` middleware (e.g. `GET /api/v1/users-info` for an offline user — see [api/src/routing.js:524](https://github.com/medic/cht-core/blob/master/api/src/routing.js#L524)).
3. Observe: the request continues and downstream code runs against stale/incomplete `req.userCtx`, with nothing logged about the failure.

**Expected behavior**
At a minimum, the failure should be logged so operators can diagnose it. Ideally the error should also be surfaced to downstream middleware (e.g. via req.authErr, mirroring the getUserCtx pattern) so the request can fail cleanly when user settings are required.

Important constraint for any fix: the three call sites at lines 72, 83, and 97 invoke this as getUserSettings(req).then(next) with no .catch. Rethrowing the error from the .catch would prevent next() from being called and hang the request, so the fix must either log-and-continue or stash the error for a follow-up middleware to handle.

**Logs**

N/A — the absence of a log line is the bug. With the current code, an `auth.getUserSettings` rejection produces no output.

**Screenshots**

N/A — backend code path, not visible in the UI.

**Environment**
- App: api
- Version: master

[...]

PR #11065, opened 2026-05-14
fix(#11051): log getUserSettings errors instead of silently swallowing them

shivv23:
## Summary

The `getUserSettings` helper in the authorization middleware silently swallows errors by catching them and returning the error object without any logging. This makes failures invisible in production, leaving operators unable to diagnose issues such as CouchDB connectivity problems or malformed user-settings documents.

## Changes

- Added `logger.error()` call in the `.catch` handler of `getUserSettings`, following the same pattern used by `captureReplicationFailures` in the same file.
- Updated unit tests to verify that errors are logged with the correct message and error object.

## Testing

- All existing authorization middleware tests pass (31 tests).
- Tests verify both `getUserCtx` failure and `getUserSettings` failure scenarios produce log output.

### DIFF
--- a/api/src/middleware/authorization.js
+++ b/api/src/middleware/authorization.js
@@ -16,7 +16,10 @@ const getUserSettings = (req) => {
     .then(userCtx => {
       req.userCtx = userCtx;
     })
-    .catch(err => err);
+    .catch(err => {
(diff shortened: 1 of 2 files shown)


andrablaj:
Hi @shivv23 , thanks for opening the PR.

Issue #9381, opened 2024-08-27
accept_case_reports transition broken - report not updated


jkuester:
**Describe the bug**
So, this is some ancient history (and I do not fully understand the workflow). But, I was trying to manually configure and test `case_id` functionality by using the `accept_case_reports` transition to set the `place_uuid` field on a report based on its `case_id` value.

Unfortunately, I could not get this to work. The transition code would get triggered as expected, but the `onMatch` function [returns the result of `silenceRegistrations`](https://github.com/medic/cht-core/blob/53439767aa828c8bd7e5513ab59f1461ae7a741f/shared-libs/transitions/src/transitions/accept_case_reports.js#L97). In order to trigger the updates to the doc to be written to the DB, the `onMatch` function should resolve a truthy value.  However, the `silenceRegistrations` function [does not seem to ever resolve something truthy](https://github.com/medic/cht-core/blob/53439767aa828c8bd7e5513ab59f1461ae7a741f/shared-libs/transitions/src/transitions/accept_patient_reports.js#L175).  

**To Reproduce**

Once again, I am not super familiar with the "proper" case_id workflow, so maybe I am just doing something wrong here....  

I started with a `register_case` app form that is associated with a contact when it gets saved. With this form, I enabled the `registration` transition and added the following app_settings config:

```json
  "registrations": [{
    "form": "register_case",
    "events": [{
      "name": "on_create",
      "trigger": "add_case"
    }]
  }],
```

When I submitted a `register_case` form, the transition would automatically update the report with a generated `case_id`.

Then I created a `case_event` form that contained nothing but a field to set the `case_id` (no linkage to any kind of contact).  I enabled the `accept_case_reports` transition with the following config:

```json
  "accept_case_reports": [{
    "form": "case_event",
    "validations": {},
    "messages": []
  }],
```

When I submit a `case_event` report (with the case_id assigned to my previous `register_case` report), I can see the `accept_case_reports` transition is triggered. It finds the existing case registration and sets the expected `place_uuid` field onto the `case_event` doc.  However, since the `silenceRegistrations` function resolves `undefined`, this doc update is not persisted to the DB.   

**Expected behavior**
I was expecting the `case_event` report to be assigned to the place associated with my `register_case` report.

**Additional context**

[...]

PR #11066, opened 2026-05-14
fix(#9381): accept_case_reports transition fails to persist doc changes

shivv23:
## Summary

The `accept_case_reports` transition's `onMatch` function returns the result of `silenceRegistrations()`, which resolves to `undefined`. The transitions framework treats falsy return values as "no changes," so document modifications (messages, `place_uuid`) are never persisted to the database.

## Changes

- Appended `.then(() => true)` to the `silenceRegistrations` call to ensure a truthy return value when the transition completes successfully.
- Updated the unit test to assert the truthy return value.

## Testing

- All 14 `accept_case_reports` transition unit tests pass.
- The fix matches the pattern used by `accept_patient_reports` and `registration` transitions, which already return `true` on success.


andrablaj:

PR #11067, opened 2026-05-14
fix(#10851): restore Devanagari numeral cross-matching for online user search

shivv23:
## Summary

The Nouveau search index migration in 5.x broke numeral cross-matching for online users. Searching with Latin digits (e.g., `12345`) no longer returned contacts with Devanagari numerals in their names, and vice versa.

## Changes

- Expanded freetext queries to search for both Latin and Devanagari digit variants when the search term contains digits.
- Example: searching `123` now generates `(123* OR १२३*)`, matching contacts in either numeral system.
- Both fuzzy match and keyed (`exact_match`) queries are supported.
- Added 9 new test cases covering digit conversion in both freetext-only and combined (freetext + contact type) queries.

## Technical details

- `latinToDevanagari()` converts Latin digits (0-9) to Devanagari equivalents (०-९).
- `devanagariToLatin()` converts Devanagari digits to Latin equivalents.
- `hasDigits()` detects whether a string contains digits in either numeral system.
- `getQueryByFreetext()` generates an OR query when both variants differ.

## Testing

- All ~960 Nouveau unit tests pass, including 9 new test cases for Devanagari numeral expansion.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12685,6 +12685,49 @@
         "proxy-from-env": "^1.1.0"
       }
     },
+    "node_modules/appium-uiautomator2-driver/node_modules/@appium/docutils": {
+      "version": "2.2.1",
+      "resolved": "https://registry.npmjs.org/@appium/docutils/-/docutils-2.2.1.tgz",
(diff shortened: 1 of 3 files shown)


andrablaj:

Issue #7594, opened 2022-04-20
Editing a report that created extra docs duplicates all extra docs


dianabarsan:
**What feature do you want to improve?**
We have a feature to allow creation of extra docs in forms: https://docs.communityhealthtoolkit.org/apps/guides/forms/additional-docs/#extra-docs (original issue: https://github.com/medic/cht-core/issues/2912)
This is widely used across many projects. 
The implementation has a problem: when editing a form that creates extra docs, every extra doc will be duplicated when saving. 

**Describe the improvement you'd like**
Decide what to do when saving such a form: 
1. Leave it as it is. 
2. Don't create additional docs, even if changes were made in the "db-doc" sections. 
3. Make db-doc sections "readonly" when editing. 
4. Disallow editing of forms that create extra docs. 
5. (BAD) Edit the extra docs directly. 

(4) is a really bad choice, because: the corresponding docs could have been altered externally, and this would overwrite the changes, the structure of the doc could have changed, etc. I listed it so it's clear it was considered. 

**Additional context**
We're changing how we edit forms with https://github.com/medic/cht-core/issues/5549 , which does not change this behavior.

mrjones-plip:
@dianabarsan - I think you had some thoughts about this ticket as far as if it's a `bug` vs `feature` ? can you update labels as needed? I know that @andrablaj just updated them, which is why we were talking about it.

dianabarsan:
yea, I think it's an issue with decision ramifications, and potentially really complex code. I don't think it's easily approachable

andrablaj:
Regarding the `help wanted` label - it doesn't necessarily imply the issue is easy or beginner-friendly (that's what `good first issue` is for). In this case, it signals that a community member has a real need for this and we would welcome someone picking it up. It's an invitation, not a promise of simplicity.

As for whether this is a bug or a feature, based on @dianabarsan's labelling and [Bhishan's comment](https://forum.communityhealthtoolkit.org/t/delivery-report-editing-issue/5509), this looks like a bug/problem rather than a feature request. 

Happy to discuss further if there are other concerns!

vikrantwiz02:
Hi @dianabarsan and @andrablaj, I'd like to take on fixing this.

[...]

PR #11068, opened 2026-05-14
fix(#7594): prevent extra doc duplication when editing reports

shivv23:
## Summary

When editing a report that created extra documents via `[db-doc=true]` in form XML, every edit created duplicate extra docs with new UUIDs while the original extra docs remained orphaned.

## Changes

- The private `_save()` method now accepts an `isEdit` parameter (`true` when editing, `false` when creating).
- The `xmlToDocs()` method skips extra doc creation when `isEdit` is `true`, returning an empty array instead of generating new extra docs.
- The `hidden_fields` list is still populated to correctly render the report view.
- Previously created extra docs remain intact and unchanged.

This implements option (2) from the issue discussion: "Don't create additional docs when editing."

## Testing

- Manually verified that editing a report with `[db-doc=true]` fields no longer creates duplicate extra documents.
- All existing webapp unit tests continue to pass.

### DIFF
--- a/webapp/src/ts/services/enketo.service.ts
+++ b/webapp/src/ts/services/enketo.service.ts
@@ -362,7 +362,8 @@ export class EnketoService {
     return form;
   }
 
-  private xmlToDocs(doc, formXml, xmlVersion, record) {
+  private xmlToDocs(doc, form, record, isEdit = false) {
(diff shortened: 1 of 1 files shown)


andrablaj:

Issue #10241, opened 2025-08-26
unique() function fails for two subsequent requests


1yuv:
**Describe the bug**
CHT offers validation [functions](https://docs.communityhealthtoolkit.org/building/reference/app-settings/#rules) and we use unique() function to ensure that a pregnancy is not registered twice with same LMP date. This is ensured by the combinations of parameters passed to unique() method. In the above example, CHT ensures that a record with same patient_id, lmp_year, lmp_month, lmp_day combination is not already in the system. 

However, it has been seen on many instances that if user sends two messages at the same time (up to 20 seconds apart), the CHT fails to unique check and both records end up being written in the database without any error. 

**To Reproduce**
1. This can be probably reproduced in resource crunch instances with more data by sending/pretending SMS requests. The failure we've seen is on instances with 300K-400K records and running on EKS infrastructure. If you tune your instance to small resource constraint and have more data, then you can send two requests of the same type at once. 
2.  We've used unique check like this on LMP registration. But this can be anything. 
```json
   {
            "property": "patient_id",
            "rule": "unique('patient_id','lmp_year','lmp_month','lmp_day')",
            "translation_key": "registration.lmp_date.duplicate_pregnancy"
      }
```

**Expected behavior**
When unique() check is in place, you expect duplicate report to be marked as not unique. 

**Logs**
The long time is taking between two calls: `calling transition.onMatch() for doc` and `finished transition registration`. View the log below, latest log first.

**First Record**
```
2025-08-17T04:31:54.831 DEBUG: finished transition registration doc 53bb6cd2-9492-4dd9-a9d9-8caa9a4d4d64 is changed seq null 
2025-08-17T04:31:41.146 DEBUG: calling transition.onMatch for doc 53bb6cd2-9492-4dd9-a9d9-8caa9a4d4d64 and transition registration s
```
**Second Record (Duplicate of first)**
```
2025-08-17T04:32:03.217 DEBUG: finished transition registration doc c661e5e3-ff76-44ef-9226-855111592395 is changed seq null 
2025-08-17T04:31:46.850 DEBUG: calling transition.onMatch for doc c661e5e3-ff76-44ef-9226-855111592395 and transition registration seq null 
```

As you can see, first record took 13s and second took 17s to complete these two action. I've already verified that CHT's validation checks happen right between these two calls.

**Screenshots**

[...]

PR #11069, opened 2026-05-14
fix(#10241): prevent race condition in unique() validation allowing duplicates

shivv23:
## Summary

The `unique()` validation uses a read-then-write pattern that is vulnerable to concurrent requests. When two SMS messages arrive within seconds, both pass the `unique()` check before either document is committed, resulting in duplicate records.

## Changes

- Added an in-memory `Set`-based lock keyed on sorted unique field value pairs.
- Before querying the database, `exists()` checks the lock set. If another concurrent call holds the same lock, `exists()` immediately returns `true` (not unique), preventing the duplicate.
- The lock is released in a `finally` block after the DB query completes.

## Why this works

Node.js single-threaded execution ensures the lock check-and-set is atomic: two async functions cannot interleave their synchronous operations at the lock-acquisition point. The race window between the DB query and save is protected because the lock is held during the entire `exists()` execution.

## Testing

- All 132 `validation_utils` unit tests pass.
- Regression: existing tests for non-race-condition scenarios continue to work as expected.

### DIFF
--- a/shared-libs/validation/src/validation_utils.js
+++ b/shared-libs/validation/src/validation_utils.js
@@ -8,6 +8,8 @@ const { Qualifier, Report } = require('@medic/cht-datasource');
 let db;
 let dataContext;
 
+const inFlightLocks = new Set();
+
 const lowerCaseString = obj => typeof obj === 'string' ? obj.toLowerCase() : obj;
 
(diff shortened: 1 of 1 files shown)


andrablaj:

PR #11072, opened 2026-05-14
feat(#11071): add replication failure user count to monitoring v2

### LINKED ISSUE #11071: Add monitoring data about users who fail to replicate
We recently added [logging for replication failures](https://github.com/medic/cht-core/commit/74b708b55132b03d3f6d365c263e7599bd25d63a). 

Include data about replication failures to monitoring apis.

dianabarsan:
# Description

Surfaces the number of distinct users that hit replication failures in the current or previous calendar month under `replication_failure.count` on `/api/v2/monitoring`. Counting calendar months (rather than a rolling window) avoids artificially low counts in the first days after a month flip.

closes #11071

Co-Authored-By: Claude Opus 4.7

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:11071-replication-failures-monitoring/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:11071-replication-failures-monitoring/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:11071-replication-failures-monitoring/docker-compose/cht-couchdb-clustered.yml)

# License


dianabarsan:
Hi @jkuester 

This is part 1 of monitoring addition. Next part will be an endpoint to show users who haven't succeeded to replicate at all recently + monitoring data. Separate to keep prs small. 

I'd appreciate your thoughts!

dianabarsan:
hi @jkuester 

slight nudge, as there is a followup pr for this, to add the "overview" endpoint, and I would really love both of these to make it to 5.2. 
Thanks! Appreciate your time!

api/src/services/replication/replication-failure-log.js
```diff
@@ -5,8 +5,10 @@ const pagination = require('../pagination');
 const TYPE_PREFIX = `replication-fail-`;
 const MAX_FAILURES = 50;
 const REPORTING_PERIOD_FORMAT = 'YYYY-MM';
+const REPORTING_PERIOD_LENGTH = REPORTING_PERIOD_FORMAT.length;
```
jkuester: Minor, but it is odd to have this variable for the reporting period, but not have one of type prefix... IMHO we don't really need this.

api/src/services/replication/replication-failure-log.js
```diff
@@ -151,7 +153,26 @@ const get = async ({ user, reportingPeriod, cursor = 0, limit = pagination.DEFAU
   return getPageByRange({ reportingPeriod, skip: cursor, limit });
 };
 
+const getRecentReportingPeriodsRange = () => {
```
jkuester: Returning the  _current + previous_ period's data is pretty confusing to the consumer.  😓  I think I understand why you did it (it is better than just returning the current OR the previous period's data), but still....

I would like this much better if we could just a rolling sum from the last 30 days. (That kind of value will be WAY easier to track/measure in Watchdog) However, I think doing that efficiently would require a view (like we have for `logs/connected_users`).  I know you mentioned having view PTSD 😬, but at a certain point the functionality tradeoffs are not worth it just to avoid the disk space of the view...  🤷

jkuester (commented):
👍 I left one complaint inline (which I am sure you anticipated).  My main concern after reading the rest of the PR and thinking about this more is that, as things are currently in this PR, I am not sure how we could meaningfully monitor _changes_ to the value over time (e.g. via Watchdog).  Since the measurement is based on the reporting period, the actual returned value can change dramatically on the day the new reporting period begins.  The cyclical pattern should be recognizable over time, but is just one more think we will have to keep in context when trying to sort through data.  

Honestly, if you think it is not worth adding a new view for this so we can return the count from a rolling interval, then my vote would be to hold off on shipping this unless we have a coherent idea of how/what we want to monitor (no need to rush a value into Watchdog that will turn out to be useless, and will need a new version of `/monitoring` to fix....)

api/src/services/replication/replication-failure-log.js
dianabarsan: Yea, I really don't want to add views right now :D 

Do you think current and past calendaristic month is a hard concept to grasp for the consumer? 
Since this just means merging two periods together, since our history of failures is dated (although limited to 50). It would add a little more logic here to merge the two failure logs.

dianabarsan: > the actual returned value can change dramatically on the day the new reporting period begins. 

I see your concern, which is why I am taking the last two reporting periods when returning results.

dianabarsan:
Hi @jkuester ! 
Some changes that have occured, based on your feedback: 

Metric semantics
  - Replaced the "current or previous calendar month" window with a rolling N-day window, addressing your core concern.
  - The endpoint now follows the same shape as `connected_users` — the existing `connected_user_interval` query parameter (default 7 days) drives both metrics, with the same default.

  Storage model
  - Added a `daily_failures: { 'YYYY-MM-DD': N }` bucket field on each `replication-fail` doc, maintained by `captureFailure` alongside the existing `failures[]` array and `total_failures` counter.
  - `failures[]` remains capped at 50 for forensic detail; `daily_failures` is the source of truth for time-windowed aggregates and isn't affected by the cap.
  - UNKNOWN constant changed from the string `'unknown'` to `null`

api/tests/mocha/services/replication/replication-failure-log.spec.js
```diff
@@ -465,11 +548,106 @@ describe('Replication Failure Log Service', () => {
...
+          user: 'bob',
+          failures: [],
+          daily_failures: {},
+        });
+        db.medicLogs.put.resolves();
+
+        await replicationFailureLog.capture({ name: 'bob', roles: ['chw'] }, 'req', 500, 100);
+
+        expect(db.medicLogs.put.args[0][0].total_failures).to.equal(1);
+      });
```
jkuester, line 613: Minor, but these all seem like they could just go in one test.

api/tests/mocha/services/replication/replication-failure-log.spec.js
```diff
@@ -465,11 +548,106 @@ describe('Replication Failure Log Service', () => {
...
+        db.medicLogs.put.resolves();
+
+        await replicationFailureLog.capture({ name: 'bob', roles: ['chw'] }, 'req', 500, 100);
+
+        expect(db.medicLogs.put.args[0][0].daily_failures).to.deep.equal({
+          '2026-04-13': 2,
+          '2026-04-14': 2,
+          '2026-04-15': 1,
+        });
       });
```
jkuester, line 651: Minor, but these tests do not seem to belong in the "defensive lazy-init of mutable fields"  file.

tests/integration/api/controllers/monitoring.spec.js
```diff
@@ -286,12 +287,48 @@ describe('monitoring', () => {
...
+    it('should count distinct users with replication failures in the configured window', async () => {
+      const today = moment();
+      const recent = today.clone().subtract(2, 'days');
+      const onBoundary = today.clone().subtract(7, 'days');
+      const tooOld = today.clone().subtract(14, 'days');
+      const failureDoc = (period, user, dailyFailures) => ({
+        _id: `replication-fail-${period}-${user}`,
+        user,
+        date: today.valueOf(),
+        total_failures: Object.values(dailyFailures).reduce((a, b) => a + b, 0),
```
jkuester, line 310: I would love to release `reduce` [from style-guide jail](https://docs.communityhealthtoolkit.org/community/contributing/code/style-guide/#avoid-reduce)   😅

api/src/services/replication/replication-failure-log.js
jkuester, line 1: Don't forget to update the OpenAPI docs in the `replication-failure-log` controller based on the changes to this file!  👍

jkuester (approved):
Very nice!  A couple minor comments, but otherwise LGTM!

tests/integration/api/controllers/monitoring.spec.js
dianabarsan: Haha I think this one might belong to fair use case, where the alternative is less readable :D Maybe we can refine the jail guidance.

api/src/services/replication/replication-failure-log.js
dianabarsan: done!

PR #11073, opened 2026-05-14
refactor(webapp): replace any types in auth directive and training-card guard

aniketsingh1023:
Part of #10883.

## Changes

- `webapp/src/ts/directives/auth.directive.ts`: replace `mmAuthAny?: any` with a narrow union (`AuthAnyValue | AuthAnyValue[]`) that reflects the runtime branches in `dynamicChecks` — a value is either `true` (short-circuit allow), a single permission name, or an AND-group of permission names; the input accepts a single value or an array of them representing OR-of-AND-groups.
- `webapp/src/ts/training-card.guard.provider.ts`: replace `CanDeactivate<any>` and `component: any` with `unknown`. The guard does not use the component parameter — `unknown` matches the actual contract without resorting to `any`.

## No behavior change

Pure typing change. No runtime logic touched. The new types match all existing call sites.

### DIFF
--- a/webapp/src/ts/directives/auth.directive.ts
+++ b/webapp/src/ts/directives/auth.directive.ts
@@ -2,12 +2,27 @@ import * as _ from 'lodash-es';
 import { Directive, Input, HostBinding, OnChanges } from '@angular/core';
 import { AuthService } from '../services/auth.service';
 
+/**
+ * A single property accepted by `mmAuthAny`. Reflects the runtime branches
+ * in the directive's `dynamicChecks`:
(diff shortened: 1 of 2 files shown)


andrablaj:
Hi @aniketsingh1023 , thank you for the PR!

aniketsingh1023:
Thanks for the clarification! I wasn't aware of the assignment-first workflow. I'll make sure to comment on the related issue and wait to be assigned before opening a PR in the future. Appreciate the feedback!

PR #11085, opened 2026-05-17
fix: handle zero-value LMP dates in pregnancy registration

### LINKED ISSUE #11070: setExpectedBirthDate drops expected_date when weeks_since_lmp is 0
## Description

In `shared-libs/transitions/src/transitions/registration.js`, the `setExpectedBirthDate` function uses truthy checks on `lmpDate` and `lmp` values. Because JavaScript treats `0` as falsy, a pregnancy registered at exactly 0 weeks since LMP (just conceived) has its `expected_date` and `lmp_date` silently set to `null`.

## Code

```js
// registration.js:170-186
const setExpectedBirthDate = doc => {
  let start;
  const lmpDate = getLMPDate(doc);
  if (lmpDate) {              // ← 0 is falsy, skips valid epoch-0 LMP date
    start = moment(lmpDate);
  } else {
    const lmp = getWeeksSinceLMP(doc);
    if (lmp) {                // ← 0 is falsy, skips valid 0-week LMP
      start = moment(doc.reported_date).startOf('day');
      start.subtract(lmp, 'weeks');
    }
  }

  if (!start) {
    doc.lmp_date = null;      // nullifies expected date for 0-week pregnancies
    doc.expected_date = null;
    return;
  }
  // ...
};
```

The same pattern exists in `getLMPDate` (line 163): `doc.fields[prop] && parseInt(...)` — when the field value is `0`, the short-circuit `&&` evaluates `0` as falsy and never calls `parseInt`.

## Steps to reproduce

1. Configure a registration form with `weeks_since_lmp` field
2. Submit an SMS or form with `weeks_since_lmp = 0` (mother just conceived)
3. Observe that the resulting contact has `expected_date: null` and `lmp_date: null`
4. ANC schedules that depend on `expected_date` are never created

## Expected behaviour

When `weeks_since_lmp` is `0`, `expected_date` should be set to 40 weeks from the reported date, and `lmp_date` should be set to the reported date.

## Affected versions

All versions — this pattern has been present since the function was written.

## Impact

Mothers registered at exactly 0 weeks LMP miss their entire ANC follow-up schedule. While uncommon (most registrations happen weeks into pregnancy), the silent failure makes it hard to detect and the consequence — a missed pregnancy follow-up — is clinically significant.

Dev10-sys:
This fixes an issue where pregnancies registered with weeks_since_lmp = 0 incorrectly cleared lmp_date and expected_date because the transition logic relied on truthy checks.

Changes

replaced loose truthy checks with explicit null/undefined validation
fixed handling for lmp = 0 and lmp_date = 0 values
preserved existing behavior for missing or empty inputs
added regression coverage for zero-week pregnancy registrations
added validation coverage for null, undefined, and empty string cases

Impact

Pregnancies registered at exactly 0 weeks since LMP now correctly generate lmp_date and expected_date, ensuring ANC schedules are created properly while keeping existing registration behavior unchanged.

Fixes #11070

### DIFF
--- a/shared-libs/transitions/src/transitions/registration.js
+++ b/shared-libs/transitions/src/transitions/registration.js
@@ -145,49 +145,49 @@ const getDaysSinceDOB = doc => {
  * property names atm.
  */
 const getWeeksSinceLMP = doc => {
-  const props = ['weeks_since_lmp', 'last_menstrual_period', 'lmp'];
-  for (const prop of props) {
-    if (doc.fields) {
-      const val = doc.fields[prop];
-      if (val !== undefined && val !== null && String(val).trim() !== '') {
-        const lmp = Number(val);
-        if (!isNaN(lmp)) {
-          return lmp;
-        }
-      }
-    }
+  const fields = doc?.fields;
+  if (!fields) {
+    return;
   }
+  const props = ['weeks_since_lmp', 'last_menstrual_period', 'lmp'];
+  const foundProp = props.find(prop => {
+    const val = fields[prop];
+    return val !== undefined && val !== null && String(val).trim() !== '';
+  });
+  if (foundProp === undefined) {
+    return;
+  }
+  const lmp = Number(fields[foundProp]);
+  return Number.isNaN(lmp) ? undefined : lmp;
 };
 
-/*
-* Given a doc, try to get the exact LMP date
-*/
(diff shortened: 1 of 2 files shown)


andrablaj:
Closing [per comment](https://github.com/medic/cht-core/issues/11070#issuecomment-4504732174)

PR #11087, opened 2026-05-18
feat(#10662): display contact profile photo

### LINKED ISSUE #10662: Display uploaded image on client's profile
As an extension to [this ](https://github.com/medic/cht-core/issues/10509#issuecomment-3937705524)feature, we would like to see the uploaded image displayed on the client's profile and also be able to change the image in person-edit form.

benkags:
# Description

Closes https://github.com/medic/cht-core/issues/10662

Add the ability to display the profile photo of the user when provided in the contact profile. The profile photo field is configurable in contact type definition.

# Code review checklist

# License


benkags:
Hi @dianabarsan @jkuester I decoupled this from #10923 since profile photo display automatically work as long as the photo has been attached to the contact doc; so this can be reviewed as well.

webapp/src/ts/components/contact-photo/contact-photo.component.ts
```diff
@@ -0,0 +1,84 @@
...
+    if (this.doc) {
+      return Promise.resolve(this.doc);
+    }
+    if (this.docId) {
+      return this.dbService.get().get(this.docId);
+    }
+    return Promise.resolve(null);
+  }
+
+  private async fetchObjectUrl(docId: string, attachmentName: string) {
```
jkuester: This is pretty much just donig the same thing as the `report-image.component` (which really does not seem to do anything specific to "reports").  Can we maybe just refactor `report-image` to be a generic `attachment-image` component and then reuse it here?  (Really, depending on how much code is left here in the `contact-photo.component`, maybe we just drop this whole file and consume attachment-image from the `contacts-content.component`, but I am also fine keeping this file if we have a use for it...).

webapp/src/ts/components/contact-photo/contact-photo.component.ts
```diff
@@ -0,0 +1,84 @@
...
+    this.loading = true;
+    try {
+      const blob = await this.dbService.get().getAttachment(docId, attachmentName);
+      this.objectUrl = (window.URL || window.webkitURL).createObjectURL(blob);
+    } catch (err) {
+      if ((err as { status?: number })?.status !== 404) {
+        throw err;
+      }
+      console.warn(`ContactPhotoComponent: attachment ${attachmentName} missing on ${docId}`);
+    } finally {
```
jkuester: Honestly, since you already checked `doc._attachments?.[attachmentName]`, I don't see any value in handling the 404 here.  It really should not possible to hit this catch-block.

webapp/src/ts/components/contact-photo/contact-photo.component.ts
```diff
@@ -0,0 +1,84 @@
...
+  constructor(
+    private readonly dbService: DbService,
+  ) {}
+
+  ngOnChanges(changes: SimpleChanges) {
+    if (changes.doc || changes.docId || changes.photoField) {
+      this.revoke();
+      return this.load();
+    }
+  }
```
jkuester: I am not an Angular expert, but we have tended to avoid `ngOnChanges` in other components.  Instead of injecting the params here as `@Input` fields and then listening for changes, we should just subscribe to `Selectors.getSelectedContact` and that gives us the `doc` and `type` objects we care about.  Something like: 

```ts
  this.store                                                                
    .select(Selectors.getSelectedContact)                                   
    .pipe(distinctUntilChanged((a, b) => a?.doc === b?.doc))                
    .subscribe(({ doc, type }) => { … })
```

webapp/src/ts/components/contact-photo/contact-photo.component.ts
```diff
@@ -0,0 +1,84 @@
+import { Component, Input, OnChanges, OnDestroy, SimpleChanges } from '@angular/core';
+import { NgIf } from '@angular/common';
+import { TranslatePipe } from '@ngx-translate/core';
+
+import { DbService } from '@mm-services/db.service';
+import { ResourceIconPipe } from '@mm-pipes/resource-icon.pipe';
+
+const USER_FILE_ATTACHMENT_PREFIX = 'user-file-';
+const DEFAULT_PHOTO_FIELD = 'photo';
```
jkuester: I had to double-check the issue thread and I did not see any discussion there about a default property for the profile photo. All in all, I like the approach of having a default and then letting it be overridden by contact_type config. 👍  

My main concern is that `photo` is very generic. It does not do a good job of describing what it should contain and, if we add additional image fields in the future, it does not effectively differentiate what is special about the `photo` value. 

I am leaning towards `profile_image`, but am open to other ideas as well.  "Profile" helps emphasize this image is shown when viewing the contact's profile page.  "Image" is a bit more generic than "photo" (which carries connotations about being an "image captured with a camera").  I can image scenarios were the profile picture for a place gets set to a logo or some other image that is not actually a "photo" in the strict sense.

Another alternative name for the property that I thought of is `avatar`.  This is probably closest to the proper technical term for the value and how it is being used.  It is definitely a well-understood term in web development, but I am not sure how well it might be understood by folks configuring the CHT (where they might not be familiar with web-dev and English is not their first language...).

webapp/src/ts/components/contact-photo/contact-photo.component.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  private async load() {
+    const doc = await this.resolveDoc();
+    const field = this.photoField || DEFAULT_PHOTO_FIELD;
+    const photo = doc?.[field];
+    if (!doc?._id || !photo) {
+      return;
+    }
+    const attachmentName = `${USER_FILE_ATTACHMENT_PREFIX}${photo}`;
```
jkuester: Assuming https://github.com/medic/cht-core/pull/10923 lands as it currently is (just pending Diana's approval 🤞). I think we need to check for both types of attachments here (file and binary):

- `user-file-${photo}`
- `user-file/${field}`

webapp/src/ts/components/contact-photo/contact-photo.component.ts
jkuester: Whatever we land on here, we should make sure the contact_type config property naming matches (e.g. `avatar_field`).

tests/e2e/default/contacts/contact-attachments.wdio-spec.js
```diff
@@ -324,4 +325,39 @@ describe('Contact form attachments', () => {
     expect(contactAfter._attachments[newAttachmentName].length, 'Replaced attachment should have a valid size')
       .to.be.greaterThan(0);
   });
+
+  describe('Displaying the profile photo', () => {
```
jkuester: Great tests here!  Would it be possible to actually just include these assertions in the above tests instead of adding whole new tests here?  I think we don't really need to write additional contacts for these tests if we just set/unset the photo field in the tests above and assert the contact page looks correct.

(Just want to avoid extra e2e tests when we can since they are a maintenance cost to run all of them for every build.)

jkuester (changes requested):
Great PR @benkags !  I am sorry the review is so delayed!  😓

Couple minor-ish changes requested, but otherwise this is super clean and good to go!

webapp/src/ts/components/contact-photo/contact-photo.component.ts
jkuester: @dianabarsan Do you have any strong thoughts on the naming of the default field here?

dianabarsan: `profile_image` sounds great!
`avatar` is such a loaded term...

jkuester:
@sugat009 could I get you to review the implementation files here please? :pray:  I ended up making some non-trivial changes (in response to my own review comments... :sweat_smile:) and really would be best to get a second set of eyes on them.  Probably don't need to worry about reviewing the test code (unless you really want to...) :+1: .

We are trying to get this into 5.3, so if it is possible to fit this in your day today, that would be amazing!  :pray:

jkuester:
@mrjones-plip, one thing of interest here (perhaps for the release notes) is that while manually testing this PR, I did confirm that the [max-pixels form parameter](https://docs.getodk.org/form-question-types/#scaling-down-images) does actually work in Enketo!  This _partially_ addresses some of the questions from the Roundup regarding ways to control the size of images uploaded. 

TLDR, you can have your form like this:

| type | name | parameters |
| --- | --- | --- |
| image | profile_image | max-pixels=512 |

This will automatically down-scale an attached image so that the longest side matches the given resolution value.  

Unfortunately, this does not apply to the [Android App Launcher](https://docs.communityhealthtoolkit.org/building/forms/app/#android-app-launcher) when you are launching straight to the camera app to get the image.  In that case we will always compress whatever image we get from the camera by 30%, but the ultimate resolution of the image still depends on the quality of the camera on each device....  :disappointed:

mrjones-plip:
Thanks for the heads up - Great to know!!  I think we can leave these specifics for the docs - it doesn't change that this is a new feature that the release notes highlight (but bummer about the camera app's 30% shenanigans :shrug: )

jkuester (approved):

1yuv (approved):
Looks great.

PR #11089, opened 2026-05-18
perf(#11088): paginate _users queries in purging getRoles()

### LINKED ISSUE #11088: getRoles() in purging loads entire _users database into memory without pagination
## Description

The `getRoles()` function in `sentinel/src/lib/purging.js` (line 69) calls `db.users.allDocs({ include_docs: true })`, which loads every user document from the `_users` database into memory in a single request. This includes all document fields such as password hashes (`derived_key`, `salt`, `password_scheme`), metadata, and other properties, even though only the `roles` field is needed.

In large CHT deployments with thousands of community health workers, this creates a significant memory spike during every purging cycle.

## Current Behavior

```js
const getRoles = async () => {
  const dedupedByRoles = new Map();
  const rolesByHash = {};

  const { rows } = await db.users.allDocs({ include_docs: true }); // loads ALL user docs at once

  for (const { doc } of rows) {
    const userRoles = doc?.roles;
    // only uses doc.roles from each document
    // ...
  }
  return rolesByHash;
};
```

Key observations:
- No `limit` parameter, so CouchDB returns the entire `_users` database in one response.
- `include_docs: true` loads full user documents when only `roles` is needed.
- Other operations in the same file (e.g., `batchedContactsPurge`, `purgeUnallocatedRecords`) carefully implement pagination with `limit`, `startKey`, and batch processing, but `getRoles` does not follow this pattern.

## Impact

**Memory**: Each user document is roughly 2-5 KB. A deployment with 10,000 CHWs loads 20-50 MB in a single allocation just for role deduplication. This runs on every purging cycle (configured via `purge.text_expression` in app settings).

**Unnecessary data in memory**: Full user documents including password hashes and other sensitive fields are loaded into the Node.js heap when only the `roles` array is required.

**Scalability**: CHT deployments are growing. The purging system already has adaptive batch sizing for contacts and records to handle memory pressure. The `getRoles` function is an outlier that bypasses this pattern.

## Expected Behavior

`getRoles()` should paginate through `_users` documents in batches, consistent with how the rest of the purging module handles large datasets.

## Reproduction

This affects any CHT deployment where:
1. A purge function is configured in app settings
2. The `_users` database contains a large number of user documents

The memory spike is proportional to the total number of users (not just offline users) since all documents must be loaded before filtering.

## Environment

- CHT Core version: master (current)
- File: `sentinel/src/lib/purging.js`, line 65-87

YASHSHARMAOFFICIALLY:
## Summary

- Paginate `getRoles()` in `sentinel/src/lib/purging.js` to process the `_users` database in batches of 1000 documents instead of loading every user document into memory at once.
- Add a test verifying pagination behavior across multiple batches.

Fixes #11088

## Context

`getRoles()` calls `db.users.allDocs({ include_docs: true })` without a `limit`, which loads the entire `_users` database in a single allocation. In deployments with thousands of community health workers, this creates a significant memory spike during each purging cycle. The rest of the purging module carefully paginates large data operations (e.g., `batchedContactsPurge`, `purgeUnallocatedRecords`), and `getRoles()` was the remaining outlier.

## Changes

**`sentinel/src/lib/purging.js`**
- Added `USERS_BATCH_SIZE = 1000` constant.
- Rewrote `getRoles()` to iterate through `_users` in paginated batches using `startkey` and `skip` for cursor progression.

**`sentinel/tests/unit/lib/purging.spec.js`**
- Updated existing assertion to match the new `limit` parameter.
- Added test case `should paginate through users in batches` that verifies multi-batch iteration, correct query parameters per batch, and correct role deduplication across batches.

## Test plan

- [x] Existing `getRoles` tests pass with updated assertions.
- [x] New pagination test validates correct batching behavior.
- [x] Full purging test suite passes (75 tests).

### DIFF
--- a/sentinel/src/lib/purging.js
+++ b/sentinel/src/lib/purging.js
@@ -62,21 +62,42 @@ const closePurgeDbs = () => {
   });
 };
 
+const USERS_BATCH_SIZE = 1000;
+
 const getRoles = async () => {
   const dedupedByRoles = new Map();
   const rolesByHash = {};
 
-  const { rows } = await db.users.allDocs({ include_docs: true });
+  let startKey;
+  let hasMore = true;
 
-  for (const { doc } of rows) {
-    const userRoles = doc?.roles;
-
-    if (!Array.isArray(userRoles) || !userRoles.length || !roles.isOffline(userRoles)) {
-      continue;
+  while (hasMore) {
+    const opts = {
+      include_docs: true,
+      limit: USERS_BATCH_SIZE,
+    };
+    if (startKey) {
+      opts.startkey = startKey;
+      opts.skip = 1;
     }
 
-    const rolesList = serverSidePurgeUtils.sortedUniqueRoles(userRoles);
-    dedupedByRoles.set(JSON.stringify(rolesList), rolesList);
+    const { rows } = await db.users.allDocs(opts);
+    hasMore = rows.length === USERS_BATCH_SIZE;
+
+    if (rows.length) {
+      startKey = rows[rows.length - 1].id;
+    }
+
+    for (const { doc } of rows) {
+      const userRoles = doc?.roles;
+
+      if (!Array.isArray(userRoles) || !userRoles.length || !roles.isOffline(userRoles)) {
+        continue;
+      }
+
+      const rolesList = serverSidePurgeUtils.sortedUniqueRoles(userRoles);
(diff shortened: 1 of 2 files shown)


andrablaj:
Hi @YASHSHARMAOFFICIALLY , thank you for your PR. 

PR #11091, opened 2026-05-18
fix(#11090): remove password from rate limiter lookup keys

### LINKED ISSUE #11090: Rate limiter uses raw password as lookup key, enabling cross-account lockout
## Description

The rate limiter in `api/src/services/rate-limit.js` includes the raw password (`req.body.password` and `basicAuth.password`) as one of the rate limit keys. This means all login attempts sharing the same password share a single rate limit bucket, regardless of the username or IP.

Related to #10705, which covers the shared proxy IP problem. This issue covers a separate vector in the same `getKeys()` function.

## Current Behavior

```js
// api/src/services/rate-limit.js:23-39
const getKeys = (req) => {
  const keys = [ req.ip ];
  if (req.body?.user) {
    keys.push(req.body.user);
  }
  if (req.body?.password) {
    keys.push(req.body.password);    // raw password used as rate limit key
  }
  const basicAuth = auth.basicAuthCredentials(req);
  if (basicAuth?.username) {
    keys.push(basicAuth.username);
  }
  if (basicAuth?.password) {
    keys.push(basicAuth.password);   // raw password used as rate limit key
  }
  return keys;
};
```

When any key hits the limit (10 failures in 10 seconds), all requests matching that key are blocked.

## Problem

**Cross-account lockout**: An attacker who knows or guesses a common password can lock out every account using that password. In CHT deployments where administrators provision devices for community health workers, it is not unusual for password patterns to overlap. Sending 10 failed login requests with `{ user: "nonexistent", password: "SharedPassword1" }` will rate-limit every real user whose password is `SharedPassword1`.

**Password collision oracle**: By observing when rate limits trigger earlier than expected, an attacker can infer that two accounts share the same password.

## Expected Behavior

Rate limiting should be scoped to IP and username only. The password should not be used as a lookup key because:
1. It creates a shared bucket across unrelated accounts.
2. It stores the raw password string in the in-memory rate limiter data structure.
3. The combination of IP + username is sufficient to prevent brute-force attacks against a specific account.

## Proposed Fix

Remove the password from the keys array in `getKeys()`:

```js
const getKeys = (req) => {
  const keys = [ req.ip ];
  if (req.body?.user) {
    keys.push(req.body.user);
  }
  const basicAuth = auth.basicAuthCredentials(req);
  if (basicAuth?.username) {
    keys.push(basicAuth.username);
  }
  return keys;
};
```

## Environment

- CHT Core version: master (current)
- File: `api/src/services/rate-limit.js`, lines 23-39

YASHSHARMAOFFICIALLY:
## Summary

- Remove the raw password from the set of keys used by the login rate limiter, preventing cross-account lockout when multiple users share the same password.
- Update tests to reflect the reduced key set (IP + username only).

Fixes #11090
Related to #10705

## Context

The rate limiter in `api/src/services/rate-limit.js` uses multiple keys per request: IP, username, and password. When any single key exhausts its 10-attempt budget within 10 seconds, all subsequent requests matching that key are blocked.

Because the raw password is used as a key, all accounts sharing the same password share a rate limit bucket. An attacker can lock out every account with a given password by sending 10 failed login requests with any username and that password. This is especially relevant in CHT deployments where administrators provision devices for community health workers and password patterns may overlap.

The combination of IP + username is sufficient to prevent brute-force attacks against individual accounts.

## Changes

**`api/src/services/rate-limit.js`**
- Removed `req.body.password` and `basicAuth.password` from `getKeys()`.

**`api/tests/mocha/services/rate-limit.spec.js`**
- Updated `isLimited` tests: adjusted call counts and removed password key assertions.
- Updated `consume` tests: adjusted call counts and removed password key assertions.

## Test plan

- [ ] All 7 rate-limit service tests pass.
- [ ] All 5 rate-limiter middleware tests pass.
- [ ] Rate limiting still works correctly by IP and username.

### DIFF
--- a/api/src/services/rate-limit.js
+++ b/api/src/services/rate-limit.js
@@ -26,15 +26,9 @@ const getKeys = (req) => {
   if (req.body?.user) {
     keys.push(req.body.user);
   }
-  if (req.body?.password) {
-    keys.push(req.body.password);
-  }
   const basicAuth = auth.basicAuthCredentials(req);
   if (basicAuth?.username) {
     keys.push(basicAuth.username);
-  }
-  if (basicAuth?.password) {
-    keys.push(basicAuth.password);
   }
   return keys;
 };
--- a/api/tests/mocha/services/rate-limit.spec.js
+++ b/api/tests/mocha/services/rate-limit.spec.js
@@ -59,28 +59,24 @@ describe('rate-limit service', () => {
       auth.basicAuthCredentials.returns({ username: 'basicuser', password: 'basicpass' });
       rateLimit.get.withArgs('quay').resolves();
       rateLimit.get.withArgs('key').resolves({ remainingPoints: 1 });
-      rateLimit.get.withArgs('kee').resolves({ remainingPoints: 1 });
       rateLimit.get.withArgs('basicuser').resolves({ remainingPoints: 1 });
-      rateLimit.get.withArgs('basicpass').resolves({ remainingPoints: 1 });
       const actual = await service.isLimited(req);
       chai.expect(actual).to.be.false;
-      chai.expect(rateLimit.get.callCount).to.equal(5);
+      chai.expect(rateLimit.get.callCount).to.equal(3);
     });
 
-    it('returns false when any limit reached', async () => {
+    it('returns true when any limit reached', async () => {
       const req = {
         ip: 'quay',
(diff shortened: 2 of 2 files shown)


andrablaj:
Hi @YASHSHARMAOFFICIALLY , thank you for your PR. 

PR #11096, opened 2026-05-20
fix(#11070): setExpectedBirthDate drops expected_date when weeks_since_lmp is 0

shivv23:
## Summary

The `setExpectedBirthDate` function in `registration.js` uses truthy checks (`if (lmpDate)` and `if (lmp)`) to determine whether an LMP date or weeks-since-LMP value was provided. Because JavaScript treats `0` as falsy, a valid `weeks_since_lmp` value of `0` silently causes both `expected_date` and `lmp_date` to be set to `null`.

This is a clinically significant bug: mothers registered at exactly 0 weeks since LMP (just conceived) miss their entire ANC follow-up schedule because no `expected_date` is calculated.

## Root Cause

```javascript
const lmp = getWeeksSinceLMP(doc);
if (lmp) {                          // ← 0 is falsy, this branch is skipped
  start = moment(doc.reported_date).startOf('day');
  start.subtract(lmp, 'weeks');
}

if (!start) {                       // ← start is still undefined
  doc.lmp_date = null;              // ← dates incorrectly nullified
  doc.expected_date = null;
  return;
}
```

## Changes

**Source** (`shared-libs/transitions/src/transitions/registration.js`):
- Changed `if (lmpDate)` → `if (lmpDate !== undefined)` (defense in depth for both code paths)
- Changed `if (lmp)` → `if (lmp !== undefined)`
- The `!start` guard is preserved — when neither an LMP date nor weeks_since_lmp is provided, dates are still correctly set to null

**Tests** (`shared-libs/transitions/test/unit/pregnancy_registration.js`):
- Updated `'setExpectedBirthDate sets lmp_date and expected_date to null when lmp 0'` → now asserts correct dates when `weeks_since_lmp = 0`
- Updated `'zero lmp value only registers patient'` to verify both `lmp_date` and `expected_date` are correctly populated

## Testing

```
cd shared-libs/transitions
UNIT_TEST_ENV=1 npx mocha ./test/unit/pregnancy_registration.js
# 26 passing ✓
```


dianabarsan:
Closing per: https://github.com/medic/cht-core/issues/11070

PR #11101, opened 2026-05-21
feat(#10973): adding byPhone qualifier

### LINKED ISSUE #10973: Add `byPhones` qualifier to `Contact.v1`
---
title: Add a `phones` qualifier to `Contact.v1` (over `medic-client/contacts_by_phone`)
type: improvement
priority: medium
domain: data-sync
labels:
  - cht-datasource
  - migration
---

## Description

Add a `byPhones(phones)` qualifier to `cht-datasource` so contacts can be
looked up by phone number through `Contact.v1`, and expose it on the existing
contact list endpoints.

Scope is the qualifier and the API surface only. The existing
`db.query('medic-client/contacts_by_phone', ...)` call sites are **not**
migrated by this ticket: review found that every one of them changes behaviour
when it moves onto the datasource — permissions, `isContact` filtering, or
input handling — and each change needs a decision of its own. That work is
#11442  No existing
read moves onto the new path here.

This is the **pilot ticket** for a series of view→qualifier migrations. The
other tickets in the series follow this recipe with the qualifier name and view
swapped. Subsequent tickets reference this one; review it carefully because
corrections here propagate to the rest.

The view (`ddocs/medic-db/medic-client/views/contacts_by_phone/map.js`):

```js
function(doc) {
  if (doc.phone) {
    var types = [ 'contact', 'district_hospital', 'health_center', 'clinic', 'person' ];
    if (types.indexOf(doc.type) !== -1) {
      emit(doc.phone);
    }
  }
}
```

Translates to: "all docs whose `phone` field is set and whose `type` is a
contact-shaped type, emit the phone as the key." Note the hard-coded type list
— it emits `type: contact` docs whose `contact_type` is not in settings, which
is why the datasource arms filter (see Requirements).

## Technical Context

**Components:**
- `shared-libs/cht-datasource/src/qualifier.ts` — add `PhonesQualifier`,
  `byPhones`, `isPhonesQualifier`
- `shared-libs/cht-datasource/src/contact.ts` — extend `getUuidsPage` /
  `getUuids` / `getPage` / `getAll` and the imperative `getDatasource()` surface
- `shared-libs/cht-datasource/src/local/contact.ts` — dispatch arms
- `shared-libs/cht-datasource/src/local/libs/doc.ts` — `queryDocsByKeys`
- `shared-libs/cht-datasource/src/remote/contact.ts` — `phone` param
- `shared-libs/cht-datasource/src/libs/parameter-validators.ts` — extend the
  shared assert helpers
- `api/src/controllers/contact.js` — `phone` query param on the two list
  endpoints

**Existing References:**
- `shared-libs/cht-datasource/src/local/contact.ts` — the `getUuidsPage`
  dispatch arms (`isFreetextQualifier`, `isContactTypeQualifier`)
- PR [#10083](https://github.com/medic/cht-core/pull/10083) — create/update
  qualifiers + Local/Remote dispatch, the pattern this ticket follows
- PR [#11327](https://github.com/medic/cht-core/pull/11327) — `byForm`, the
  plural-qualifier and `keys`-helper decisions reused here

### Qualifier

1. `PhonesQualifier` in `qualifier.ts`, plural and array-shaped:

   ```ts
   export type PhonesQualifier = Readonly<{ phones: [string, ...string[]] }>;
   ```

   `byPhones(phones)` dedupes with a `Set` in supplied-key order.
   `isPhonesQualifier` accepts only a non-empty array of non-blank strings.

2. Phone numbers are matched **verbatim** — leading `+`, country codes and
   internal formatting are passed through untouched, because the view does no
   normalisation and the qualifier must not either. Coercion is the caller's
   job; do not call `String(phone)` in the validator.

3. A blank or whitespace-padded entry is rejected rather than passed through: a
   padded number can only ever match nothing, so it is an argument error, not a
   silent empty page.

4. Expose `byPhones` through the existing `export * as Qualifier` re-export.

### Adapters

5. Local: dispatch arms in `getUuidsPage` and — via `getContactDocsPageFn` — in
   `getPage`. Both query the view with `keys` and filter rows through
   `isContact`, so the id path and the doc path return the same contacts.
   Without the filter the id path would hand back ids that
   `GET /api/v1/contact/{id}` then 404s, since `get` applies `isContact` too.
   Deduping in a stable order is what keeps the skip cursor meaningful between
   pages, since rows come back grouped in supplied-key order.

6. `queryDocsByKeys` in `local/libs/doc.ts` — the docs twin of the
   `queryDocIdsByKeys` helper #11327 added.

7. Remote: a `phone` param on both list endpoints.

8. Extend the shared assert helpers rather than duplicating them
   (`assertContactTypeFreetextPhonesQualifier`,
   `assertContactTypeIdsPhonesQualifier`), so no call site can still reach a
   two-qualifier assertion.

9. `getUuidsPageByPhones` / `getUuidsByPhones` / `getPageByPhones` /
   `getByPhones` on the imperative `getDatasource()` surface, next to
   `getPageByType`.

### API

10. A `phone` query param on `GET /api/v1/contact/uuid` and
    `GET /api/v1/contact`, accepting a comma-joined list. No new route and no
    `routing.js` change — paging, permissions and the OpenAPI shape stay on the
    established contract.

11. Precedence: on `/uuid`, `phone` over `type` and `freetext`; on `/contact`,
    `ids`, then `phone`, then `type`.

12. An object-shaped `phone` (`?phone[a]=b`, which `qs.parse` parses to an
    object) is rejected with a 400 rather than throwing.

13. The no-qualifier error message changes on both endpoints:
    `Either query param freetext or type is required` becomes
    `Either query param freetext, type or phone is required`. This is the only
    change visible to an existing API consumer.

### Tests

- Unit: the qualifier (valid and invalid input — empty array, non-string entry,
  blank, whitespace-padded), both local arms including the `isContact` filter
  on the id path, the remote adapter, the assert helpers, the controller.
- API integration: `?phone=`, comma-separated, repeated param, and the 400
  cases on both endpoints.
- Datasource integration: `getUuidsPage`, `getPage`, the generator,
  non-normalisation, a three-page cursor walk asserting exact ids and cursors,
  and a `type: contact` doc with an unconfigured `contact_type` that the view
  emits and neither phones path returns.

## Acceptance Criteria

- `byPhones` lands in `qualifier.ts` and is reachable as `Qualifier.byPhones`,
  with valid and invalid input unit tests passing.
- Local and Remote adapters dispatch the qualifier on both the uuid path and
  the doc path, with adapter-level unit tests passing.
- `GET /api/v1/contact/uuid?phone=` and `GET /api/v1/contact?phone=` return
  `{ data, cursor }` matching the existing pagination contract.
- The uuid path and the doc path return the same set of contacts for the same
  numbers.
- `eslint .`, `npm run unit-api` and `npm run unit-shared-lib` pass.

## Constraints

- Backwards-compatible: existing `getUuidsPageByFreetext` / `getUuidsPageByType`
  calls continue to work unchanged. The no-qualifier error message is the one
  accepted change.
- No new top-level method on the `Contact.v1` namespace. The qualifier pattern
  is the explicit design constraint of this batch; the batch-shaped helpers go
  on the imperative `getDatasource()` surface instead.
- Do not relax `Input.v1.*Input._id?: never`. Client-supplied UUIDs go through
  the existing `idHint` curried-arg pattern in
  `shared-libs/cht-datasource/src/local/libs/doc.ts`, not via qualifiers.
- Do not normalise phone numbers in the qualifier. The CouchDB view does not
  normalise; the qualifier must not either.
- Do not change the ddoc. Emitting the doc type as the view value would let the
  `isContact` filter run on `row.value` and save the uuid path an
  `include_docs`, but that is a ddoc change and a reindex — revisit if the
  phones uuid path ever gets a high-volume caller.
- Do not migrate any existing `contacts_by_phone` caller. That is a separate
  ticket, for the reasons in the Description.

## References

**Similar Implementations:**
- https://github.com/medic/cht-core/pull/10083 — qualifier + Local/Remote dispatch pattern
- https://github.com/medic/cht-core/pull/11327 — plural qualifier, `keys` helper, query-param serialisation
- https://github.com/medic/cht-core/pull/9625 — the pagination-test discipline cited above

**Documentation:**
- `ddocs/medic-db/medic-client/views/contacts_by_phone/map.js`
- `shared-libs/cht-datasource/README.md`

witash:
closes #10973 

This PR adds `byPhone` and `byPhones` qualifiers to datasource, and migrates all the callers of `contacts_by_phone` to use datasource.

The migrated callers are listed in the #10973 
- `shared-libs/transitions/src/transitions/`: `update_sent_by.js`, `registration.js`, `self_report.js`, `update_clinics.js`
- `shared-libs/validation/src/validation_utils.js`
- `admin/src/js/services/message-queue.js`
- `webapp/src/js/enketo/widgets/phone-widget.js`, `webapp/src/ts/services/integration-api.service.ts`

The diff is a little large, but most of the changed lines are tests.

Some notes:
This is the first qualifier that cannot be combined with other qualifiers.
`contact_type` and `freetext`, the existing qualifiers for contacts, COULD be combined because of the `contacts_by_type_freetext` view.
Phone cannot, and the other qualifiers that we will add will also need similar rules.
So, added some validation around that; one thing to note is that invalid combinations of qualifiers will return 400.
In the target controller, `byContactId` and `byContactIds` also cannot be combined, and it silently picks `byContactId`; so this is not totally consistent, but I think that's a bit different because its the singular/plural version of essentially the same qualifier instead of a totally different parameter.

This qualifier also needs to support multiple (and many) parameters, to have the same behavior as querying the couch view with `keys` instead of a single `key`.

There are multiple ways to do that, this branch adds a second related plural qualifier, `byPhones`, following the precedent of `byContactId`/`byContactIds`, and exposes a second `POST` endpoint where the parameters are sent as a JSON body. This differs from `byContactIds`, where multiple parameters are sent as a comma joined GET parameter; the reason is that for bulk phone lookups, there could be a large number of parameters, where its awkward to send as a comma joined GET parameter. 

It's not consistent with `byContactIds`, so an open question is should that be changed also, or should `byPhones` be implemented differently?
Another options would be to keep the request body but still use GET.

Another thing to note is the `collectUuidsByPhone(s)` helper, which was to get around some oddness in the enketo phone-widget; it can't use `await for` so it needs this helper to convert it to a promise.

There were some somewhat unrelated changes to tests to keep them concise as we add more qualifiers; tests for `byPhone`/`byPhones` are parametrized, which sacrifices some readability for conciseness (otherwise have to duplicate entire tests), and added a helper in `shared-libs/cht-datasource/test/local/contact.spec.ts`: `queryBackends`. Its not directly related to `byPhone`, but is necessarry not to duplicate a block of assertions again as more qualifiers are added.


sugat009:
@witash There are conflicts with the master branch.

sugat009:
@witash following up on this, still seeing the conflicts with master.

Issue #10323, opened 2025-09-24
Investigate loading externally hosted media into forms


dianabarsan:
Forms allow some amount of raw html injected into enketo elements (notes?). 
Investigate whether we can load external media (like youtube videos) into forms - with the awareness that an internet connection is required to view this media!

 ------------------

## Good first issue

This ticket has the [Good first issue label](https://github.com/medic/cht-core/issues?q=state%3Aopen%20label%3A%22Good%20first%20issue%22)! This means it's been especially curated by other CHT contributors to be easy to work on for first time contributors [per the docs](/community/contributing/creating-good-first-issues/).

To succeed on this ticket, please:
1. Read the parent ticket to understand the context of the issue
2. Set up the CHT locally with [docker helper](https://docs.communityhealthtoolkit.org/hosting/cht/app-developer/#cht-docker-helper)
3. find the XLSX file for existing form in the demo app - this is what is installed by default in a new docker helper instance.  The [directory is here](https://github.com/medic/cht-core/tree/master/config/demo/forms)
4. Edit a form to have the custom HTML to include a youtube video and an externally hosted image file or any other media you want to test
5. using [CHT Conf](https://docs.communityhealthtoolkit.org/community/contributing/code/cht-conf/), push the form to you CHT Docker Helper instance
6. in a browser, check if the youtube video, image file etc load as expected
7. report back on this ticket with your findings

dianabarsan:
@nKataraia pinging you to try this! I can't assign unless you are subscribed

nKataraia:
@dianabarsan  I can take a look, thanks

megha1807:
Hello @dianabarsan @andrablaj ,
I would like to work on this issue. Could you please assign this to me?

mrjones-plip:
Hi hi!  While this ticket is assigned and may have a PR opened for it, I'm adding the "Good First Issue" label and text to it.  If it gets closed with the current assignee that's fine!  If it slips through the cracks and not finished now, it will be ready for the next community member to help with!

nulllkomalll:
Hi @dianabarsan @andrablaj, I see that @megha1807 is currently assigned and working on this for Q2. If they need any help, or if this ends up slipping through the cracks and becomes unassigned, I would love to take it on next!

mrjones-plip:
Thanks for the offer @nulllkomalll ! 

[...]

PR #11103, opened 2026-05-22
feat(#10323): allow external media in enketo form notes

megha1807:
## Description

Updates the Content Security Policy in `api/src/routing.js` to allow external media to be embedded in Enketo form note fields.

**Changes:**
- Added `https://www.youtube.com` to `childSrc` to allow YouTube iframes
- Added `https:` scheme to `imgSrc` to allow externally hosted images

**Testing:**
Verified locally that YouTube iframes now render in form note fields (previously blocked by CSP). External images also load with `https:` scheme.

Relates to #10323

### DIFF
--- a/api/src/routing.js
+++ b/api/src/routing.js
@@ -233,7 +233,7 @@ app.use(
           `'self'`,
           'data:', // unsafe
           'blob:',
-          '*.openstreetmap.org',// used for enketo geopoint widget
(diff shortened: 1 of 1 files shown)


mrjones-plip:
Thanks again @megha1807! Closing per [issue comment](https://github.com/medic/cht-core/issues/10323#issuecomment-4517195147)

Issue #10675, opened 2026-02-26
Add e2e tests for UI Extensions


jkuester:
I think we can create new file: `tests/e2e/default/ui-extensions/ui-extension-tabs.wdio-spec.js`

Need tests for both types of extensions including:

- Navigating to the extension via UI buttons
- Loading extension content, translations, resources, etc
- Can verify telemetry (any maybe feedback?)

Will need to create a test web component to use.

megha1807:
Hello @jkuester @andrablaj 
I would like to work on this issue. Could you please assign this to me?

jkuester:
Sorry!  We actually are not ready to work on this one yet.  This is going to need functionality from pretty much all the other sub-issues of https://github.com/medic/cht-core/issues/10224 to be finished before we can actually get a working test.

Abhishek9639:
Hi @jkuester, 
I’d like to work on this issue. I have experience with UI testing and can handle the e2e tests for UI extensions. 
Could you please assign this to me?
Thanks

jkuester:
@Abhishek9639 thanks for the interest here, but unfortunately this issue is still blocked while we finalize the actual implementation.  We need to land the UI navigation functionality before we can move forwards here.

Vedd-Patel:
@jkuester i would love to work on this when other blocker issues are resolved
 as i've context about the UI extensions bcz i recently worked on ui-extensions

PR #11105, opened 2026-05-22
feat(#10675): add e2e tests for UI Extensions

jkuester:
# Description

Closes https://github.com/medic/cht-core/issues/10675

The main goal of this PR was to add e2e tests.  I ended up needing to fix a few more bugs along the way:

- The app was not properly refreshing for online users when a UI Extension's `.js` file was modified:
    - I updated the changes feed watcher in `app.component.ts` to include UI Extension files (since these are not included in the service worker).
- Changes to online-only extension properties were not being reflected in the app for Online users (even after reloading the app):
    - I updated the `generate-service-worker.js` code in api to include the properties for _all_ extensions (even the online-only ones) in the hash value for the service worker. I think since the `/ui-extensions` endpoint was registered with the service worker glob, the browser was automatically caching the first response from that endpoint even for online users and was not busting the cache on changes to online-only extensions because the service-worker hash was never changing.
- Loading resources in an extension (via the new `getResource` function) was failing for online users when refreshing the page because the extension loaded before the resource cache was warm.
    - I exposed an `init` method on the CustomResourceService so the CHTDatasourceService can wait for the cache to be warm.
- Errors thrown from within the custom web component were not being captured in a feedback doc.
    - I updated the `ui-extensions-tab.component.ts` to listen for these errors and set them in the `console.error` when they come.

## AI disclosure

I worked a lot with Claude on this, going back and forth on the different issues, etc. I have carefully pulled apart all of the code and confirmed it works as expected and does exactly what we want. Everything that is left here has either been heavily modified or reviewed by me.

# Code review checklist

# License


jkuester:
@dianabarsan this is the final PR for UI extensions before we are ready to go do `master`!  :tada: 

I am planning to send you the PR for `master` too once this gets merged.  I wanted you to have a chance to look at the bug fixes here on their own, but if you would rather just see everything all at once, approve this PR and I will open the big one. :+1: 

Thank you! :pray:

webapp/src/ts/modules/ui-extensions/ui-extensions-tab.component.ts
```diff
@@ -39,12 +40,41 @@ export class UiExtensionsTabComponent implements AfterViewInit, OnDestroy {
...
+  Errors generated by the UI Extension web components are not handled by Zone.js/ErrorHandler. Instead, we add a
+  listener here to call console.error for these errors to ensure a feedback doc is written.
+  Though the listener is global (and in theory can catch all errors, not just those thrown by the extension code), the
+  Zone.js/ErrorHandler logic captures errors in the Angular space before they propagate to the global listeners. So,
+  in practice, we should not write duplicate feedback docs for the same error.
+   */
+  private registerErrorListener(extensionId: string) {
+    this.errorListener = (event: ErrorEvent | PromiseRejectionEvent) => {
+      const error = event instanceof PromiseRejectionEvent ? event.reason : event.error ?? event.message;
+      console.error(`Error in UI extension: "${extensionId}"`, error);
```
dianabarsan: Is it possible that another thing, other than Angular and the UI extension, triggers an error here? It yes, it will get logged and attributed to the ui extension, which could be misleading.

tests/e2e/default/ui-extensions/ui-extensions-tab.wdio-spec.js
```diff
@@ -0,0 +1,171 @@
...
+      it('renders the extension content when navigating to the tab', async () => {
+        await $(`#${HEADER_TAB_ID}`).click();
+        await commonPage.waitForPageLoaded();
+
+        const customElement = await $('#ui-element-tab cht-header-tab');
+        await customElement.waitForDisplayed();
+        expect(await customElement.getText()).to.equal('Hello world');
+
+        const telemetries = await getTelemetry(`ui-extension:header-tab:render`, user.username);
+        expect(telemetries).to.have.lengthOf(2);
```
dianabarsan: Can you please add assertions about what the telemetries contain? Or at least a comment about why the header-tab has 2 and the sidebar-menu has 1.

tests/e2e/default/ui-extensions/ui-extensions-tab.wdio-spec.js
```diff
@@ -0,0 +1,171 @@
+const fs = require('fs');
+const path = require('path');
+const utils = require('@utils');
+const commonPage = require('@page-objects/default/common/common.wdio.page');
+const loginPage = require('@page-objects/default/login/login.wdio.page');
+const { hamburguerMenuItemByOption, closeReloadModal } = require('@page-objects/default/common/common.wdio.page');
```
dianabarsan: `hamburguer`s are yummy :D :D but nothing to a `hamburger`

dianabarsan (approved):
Very lovely. Minor nits inline.

webapp/src/ts/modules/ui-extensions/ui-extensions-tab.component.ts
jkuester: The short answer is yes. This listener is global and so can capture other things. I did verify that errors from the Angular space did not leak past the Angular error handling and did not hit this.  But, I guess things outside the zone could get picked up here.  

I will update the error message a bit to note: `Error on page with UI extension...`.  Hopefully this better communicates that we do not know for sure exactly where the error came from...

tests/e2e/default/ui-extensions/ui-extensions-tab.wdio-spec.js
jkuester: :+1: I could not really come up with anything meaningful to assert in the contents of the telemetry (since we are just tracking the time it takes to render the extension).

The header-tab is rendered twice because it is the primary tab (`weight: 0`) and so is loaded initially when the user logs in.  I agree this adds some unnecessary cognitive overhead to the tests, so I updated the weight to make the header-tab the second tab instead.  Now we just get a single rendering telemetry.

PR #11108, opened 2026-05-23
chore(#10224): include docs for Datasource in site

jkuester:
# Description

This is a bit of prep work ahead of #10224 being resolved.  When UI Extensions are released, that will be the official first time these cht-datasource APIs are actually available/useful for external configuration code (since the custom UI Extensions have access to the APIs and can leverage the `async` functionality!  Access to these APIs is provided by injecting the result of [getDatasource](https://docs.communityhealthtoolkit.org/cht-datasource/functions/getDatasource.html) into the extensions at `this.cht`.  

However, as you can see by looking at the docs in the link, we do not actually have very clean documentation for the _imperative_ cht-datasource APIs.  The [function-level TS doc](https://github.com/medic/cht-core/blob/ae8f2824e4a06757ceb66defbf01c8abdac8dc61/shared-libs/cht-datasource/src/index.ts#L66) is not even pulled through into the docs site. :disappointed:   This is because we have not really defined the returned "Datasource" as a formal type/interface (but it is just a return type).

The main goal of this PR is to add proper documentation for a `Datasource` interface that gets included in the docs site. This involved several changes:

- I exported a top-level `Datasource` interface to hold the documentation. The `getDatasource` function returns this type.
- For better maintainability, I broke out all the sub-concept levels (person/place/etc) into their own `Datasource` interfaces along with their own internal `getDatasource` functions and moved the code into the related files.  Everything still rolls back up through the top-level `Datasource`, but this is way more maintainable and scalable. 
- Instead of using the README as the ["root" documentation](https://docs.communityhealthtoolkit.org/cht-datasource/index.html) for the TSDoc site, I have used the `@packageDocumentation` annotation in the `index.ts` file to make that file's docs the entry-point for the cht-datasource docs site.  
  - This becomes the "public" front for these APIs (with an emphasis on pointing to the `Datasource` interface, which is what most folks care about).
  - The README focus is now on "internal" documentation regarding cht-datasource details that are really only relevant for cht-core developers.
- I made a few minor tweaks to the existing cht-datasource TSDocs to eliminate the existing warnings when generating the docs site (mostly exporting missing base types).
  - I had to make a small change to the `generate-openapi.js` script because the new `Datasource` interface is, of course, not referenced in any of our OpenAPI docs (as expected). The unreferenced schema was causing the OpenAPI json generation to fail.

## AI Disclosure

Claude was super useful for producing/implementing the `Datasource` interfaces. However, if you look at my first commit, you can see that it has all these changes but leaves the unit tests totally untouched (and they still pass). This gave me high confidence that migration was properly executed.

Beyond that, pretty much all of the other minor docs changes, etc are my own doing.

# Code review checklist

# License


jkuester:
@sugat009 please do not fear the line-count on this PR!  :sweat_smile:  :pray:   I promise it is almost entirely just a very "basic" refactor of how we structure the Datasource code/docs in cht-datasource.  (Not really any complexity.)  All the details about the changes are up in the description... :+1:

shared-libs/cht-datasource/src/contact.ts
```diff
@@ -139,4 +143,138 @@ export namespace v1 {
...
+    /**
+     * Returns an array of contact identifiers for the provided page specifications, freetext and type.
+     * @param freetext the search keyword(s)
+     * @param type the type of contact to search for
+     * @param cursor the token identifying which page to retrieve. A `null` value indicates the first page should be
+     * returned. Subsequent pages can be retrieved by providing the cursor returned with the previous page.
+     * @param limit the maximum number of identifiers to return. Default is 10000.
+     * @returns a page of contact identifiers for the provided specifications
+     * @throws InvalidArgumentError if either `freetext` or `type` is not provided
+     * @throws InvalidArgumentError if the `freetext` is empty or if the `type is invalid for a contact
```
sugat009: **issue:** The closing backtick after `type` is missing on this `@throws` line. The unclosed code span runs forward through the subsequent `@throws` lines until the next backtick (in `` `<=0` ``), merging multiple throw clauses into one garbled `<code>`-wrapped paragraph in the rendered docs.

```suggestion
     * @throws InvalidArgumentError if the `freetext` is empty or if the `type` is invalid for a contact
```

Same pattern at three other locations in this file:

* `shared-libs/cht-datasource/src/contact.ts:195`
* `shared-libs/cht-datasource/src/contact.ts:229`
* `shared-libs/cht-datasource/src/contact.ts:238`

Worth catching here since `--treatWarningsAsErrors` was just enabled on `gen-docs` in `package.json:14` and the merged rendering is visible on the generated `Contact.v1.Datasource.html` page (`getUuidsPageByFreetext` and `getUuidsPageByTypeFreetext` properties).

shared-libs/cht-datasource/src/index.ts
```diff
@@ -52,463 +32,81 @@ export * as Input from './input';
...
+}
+
 /**
  * Returns the source for CHT data.
  * @param ctx the current data context
  * @returns the CHT datasource API
  * @throws Error if the provided context is invalid
  */
-export const getDatasource = (ctx: DataContext) => {
+export const getDatasource = (ctx: DataContext): Datasource => {
```
sugat009, line 90: **question (non-blocking):** Is it intentional that the published docs site has no usage examples on the `getDatasource` or `Datasource` pages now?

The three `@example` blocks moved from `index.ts` to the README, and `gen-docs` now runs with `--readme none` (`package.json:14`), so TypeDoc no longer renders those examples anywhere on the docs site.

If unintentional, would it be worth adding back at least the imperative example as an `@example` on `getDatasource` so visitors landing on the docs site entry point still see a usage snippet?

shared-libs/cht-datasource/src/index.ts
```diff
@@ -52,463 +32,81 @@ export * as Input from './input';
 export * as Report from './report';
 export * as Target from './target';
 
+/**
```
sugat009, line 35: **suggestion (non-blocking):** Worth expanding this JSDoc to match the doc style of the other public interfaces in `cht-datasource` (`DataContext` and `SettingsService` in `libs/data-context.ts` both carry multi sentence descriptions explaining what they represent and how they're meant to be used).

Since this interface is the new docs site entry point, a few extra sentences here would give visitors landing on the `Datasource` page actual orientation. Something like:

```ts
/**
 * The CHT datasource API. Provides access to the CHT data model through versioned
 * sub APIs (e.g. `v1.person`, `v1.place`). Obtain an instance via {@link getDatasource}.
 */
```

shared-libs/cht-datasource/src/libs/core.ts
```diff
@@ -34,8 +34,14 @@ export const isNonEmptyArray = <T>(value: T[]): value is NonEmptyArray<T> => !!v
...
 export const getLastElement = <T>(array: NonEmptyArray<T>): T => array[array.length - 1];
 
-type DataValue = DataPrimitive | DataArray | DataObject;
-type DataPrimitive = string | number | boolean | Date | null | undefined;
+/**
+ * A serializable value.
+ */
+export type DataValue = DataPrimitive | DataArray | DataObject;
+/**
+ * A privative value.
```
sugat009:
```suggestion
 * A primitive value.
```

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -278,7 +278,12 @@ export const byContactIds = (contactIds: [string, ...string[]]): ContactIdsQuali
 };
 
 // https://stackoverflow.com/a/50375286
-type UnionToIntersection<U> = (U extends unknown ? (k: U) => void : never) extends (k: infer I) => void ? I : never;
+/**
+ * The intersection of the specified types.
+ */
+export type UnionToIntersection<U> = (
```
sugat009, line 285: **suggestion (non-blocking):** `UnionToIntersection` is a TypeScript utility helper rather than part of the data model. Now that `qualifier.ts` is re-exported as `Qualifier` from the package root, this lands on the public surface as `Qualifier.UnionToIntersection`. Consider marking `@internal` so `--excludeInternal` strips it from the docs site.

```ts
/**
 * The intersection of the specified types.
 * @internal
 */
```

scripts/build/generate-openapi.js
```diff
@@ -136,7 +140,7 @@ const generateTsSchemas = () => {
     .map(path => ({ ...TSJ_OPTIONS, path }))
     .map(opts => tsj.createGenerator(opts))
     .map(generator => generator.createSchema())
-    .map(({ definitions }) => ({ ...definitions, '*': undefined }))
+    .map(({ definitions }) => ({ ...definitions, ...SCHEMAS_TO_SKIP  }))
```
sugat009:
```suggestion
    .map(({ definitions }) => ({ ...definitions, ...SCHEMAS_TO_SKIP }))
```

sugat009 (changes requested):
Requesting changes since several findings affect docs site rendering (typos, missing examples on entry point pages, unclosed backticks merging `@throws` clauses). The rest are advisory questions and nitpicks.

Also worth fixing while you're cleaning up docs (already on master, small enough to bundle in this PR):

* `InvalidArrgumentError` typo (extra `r`) at `shared-libs/cht-datasource/src/contact.ts:103` and `shared-libs/cht-datasource/src/target.ts:121`. Should be `InvalidArgumentError`.

Noticed while reviewing, flagging for awareness (not asks for this PR):

* Cyclic import between `shared-libs/cht-datasource/src/person.ts:6` and `shared-libs/cht-datasource/src/place.ts:2` (lineage types reference each other; the cycle is purely at the type layer, so `import type` would break it).
* `return await fn(qualifier)` at `shared-libs/cht-datasource/src/report.ts:204` is the only `return await` in the curried functions; the rest of the area uses `return fn(qualifier)`.

shared-libs/cht-datasource/src/index.ts
jkuester: Yes this is intentional.  Custom config code will always have the Datasource injected already (so no need to call `getDatasource`).  The README exists for the benefit of cht-core devs to understand the usage patterns that can be used in the cht-core code. This does not need to be captured in the Docs site.

shared-libs/cht-datasource/src/person.ts
```diff
@@ -188,4 +188,96 @@ export namespace v1 {
...
+
+    /**
+     * Updates an existing person to have the provided data.
+     * @param updated the updated person data. The complete data for the person must be provided. Existing fields
+     * not included in the updated data will be removed from the person. If the provided parent lineage is
+     * hydrated (e.g. for a {@link v1.PersonWithLineage}), the lineage will be properly dehydrated before being
+     * stored.
+     * @returns the updated person with the new `_rev` value
+     * @throws InvalidArgumentError if `_id` is not provided
+     * @throws ResourceNotFoundError if `_id does not identify an existing person contact
```
sugat009: **issue (non-blocking):** Same unclosed backtick class as the `contact.ts` `@throws` lines fixed in e18d0cc3e, but missed in the sibling concept files. The `` `_id `` inline code span opens here and is never closed, so TypeDoc merges this clause with the following `@throws` lines into one garbled code span. I rendered the docs and confirmed the corruption on the new `Person.v1.Datasource`, `Place.v1.Datasource`, and `Report.v1.Datasource` pages, which are the pages this PR adds. The next clause already shows the intended form (`` `_rev` ``).

```suggestion
     * @throws ResourceNotFoundError if `_id` does not identify an existing person contact
```

Same fix at five more spots: `person.ts:177`, `place.ts:171`, `place.ts:258`, `report.ts:167`, `report.ts:273`.

shared-libs/cht-datasource/src/index.ts
```diff
@@ -52,463 +32,82 @@ export * as Input from './input';
...
+ *  The CHT datasource API. Provides access to the CHT data model through versioned
+ *  sub-APIs (e.g. `v1.person`, `v1.place`). Get an instance via {@link getDatasource}.
```
sugat009: **nitpick (if-minor):** Two spaces after the leading asterisk here and on the next line, vs a single space elsewhere in the file. Cosmetic only (TypeDoc collapses it), but inconsistent.

```suggestion
 * The CHT datasource API. Provides access to the CHT data model through versioned
 * sub-APIs (e.g. `v1.person`, `v1.place`). Get an instance via {@link getDatasource}.
```

sugat009 (approved):
Reviewed again after the round 1 fixes. The refactor is faithful and behavior preserving: I checked the composed `Datasource` against the pre PR `getDatasource` method by method (contact 8, person 6, place 6, report 6, target 4, plus the two permission helpers), and the qualifier composition, argument order, and default page limits all carry over unchanged. The test migration is complete (all 49 `it()` cases preserved, each concept's `getDatasource` now unit tested in isolation), and locally `tsc`, `gen-docs --treatWarningsAsErrors`, and the unit suite (951 passing) are all green. The OpenAPI skip mechanism works as intended. All round 1 findings are resolved, and I confirmed the `person`/`place` import cycle is pre existing on master rather than introduced here.

Approving so this is unblocked. I left two small docs nits inline, neither blocking: the `_id` backtick class that the round 1 fix resolved in `contact.ts` but missed in `person.ts`/`place.ts`/`report.ts` (it renders garbled on the new `Datasource` interface pages, so worth a quick follow up), plus one cosmetic double space. Thanks for the thorough cleanup here.

PR #11110, opened 2026-05-23
chore: replace "two birds with one stone" idiom in analytics comment

OriginalGary:
chore: replace "two birds with one stone" idiom in analytics comment

Replaces the lead comment on line 100 of
`webapp/src/ts/modules/analytics/analytics-target-aggregates-detail.component.ts`
with a literal description that matches the explanatory comment on the
next line ("both this component and the parent … need to be updated").
Comment-only — no behaviour change, no test impact. Happy to close if
this isn't a fit.


andrablaj (approved):
Thanks!

PR #11126, opened 2026-06-02
fix(#11125): distribute noveau in multi-node-couchdb

### LINKED ISSUE #11125: Distribute nouveau across all couch nodes in multi-node deployments
In CHT deployments with a multi-node CouchDB cluster, Nouveau runs as a sidecar on **only the first CouchDB node** (`couchdb-1`), and the single `nouveau` Service routes **all** Nouveau traffic to that one node. Search indexing and query load is therefore concentrated on `couchdb-1` instead of being distributed across the cluster.

This is ok for freetext search, which is not important for scaling, but Nouveau is also on the **sync / replication path** 
(`api/src/services/replication/authorization.js`), so a single Nouveau instance becomes a scaling bottleneck failure for sync in deployments that run multi-node CouchDB.

**Current behaviour**

Two pieces of the Helm chart only allow a single `nouveau` container:

1. **`scripts/build/helm/templates/couchdb/deployment.yaml`**  in the  clustered branch the `cht-couchdb-nouveau` sidecar is created only if  `{{- if eq $nodeNumber 1 }}`

2. **`scripts/build/helm/templates/nouveau/service.yaml`** — a single `nouveau` service whose selector is `cht.service: couchdb-1` in clustered mode, so even   if other nodes had a sidecar, traffic would still only reach node 1.

The more difficult blocker is the hard-coded `nouveau` URL baked into the `couchdb` image:

```ini
# couchdb/10-docker-default.ini
[nouveau]
enable = true
url = http://nouveau:5987
```

Because `10-docker-default.ini` is shipped inside the `cht-couchdb` image, every couchdb node points at the single `nouveau` service name, and there is no way currently to point at a separate `nouveau` instance in the same pod.

**Desired behaviour**

- A `cht-couchdb-nouveau` sidecar on **every** couchdb pod
- Each couchdb pod configured with a url pointing to the same nouveau sidecar

**Approaches considered**

| Approach | Verdict |
| --- | --- |
| Mount per-node ConfigMap with `[nouveau] url` under `/opt/couchdb/etc/default.d` | most helmish but breaks `entrypoint.sh` |
| initContainer writes `local.d/20-nouveau-url.ini` into the writable PVC |  Worked in testing but a very messy workaround | 
| Make image's `10-docker-default.ini` URL env-driven; chart sets per-node | Not tested yet, would be preferred if possible |

**References**

- Sidecar split: `fix(#10481): set nouveau as a sidecar container (#10482)`
- Hard-coded URL: `couchdb/10-docker-default.ini:48`
- Sidecar gate: `scripts/build/helm/templates/couchdb/deployment.yaml` (`if eq $nodeNumber 1`)
- Service selector: `scripts/build/helm/templates/nouveau/service.yaml` (`cht.service: couchdb-1`)

witash:
# Description

uses the existing pattern of replacing values in `couchdb/docker-entrypoint.sh` to overwrite the default nouveau url "http://nouveau:5987" if NOUVEAU_URL env var is set.

its actually possible to remove the nouveau services entirely and assume nouveau is a sidecar, so use localhost:5987 by default, but have kept the services anyway.

closes #11125
 
# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:11125-nouveau-multi-node/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:11125-nouveau-multi-node/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:11125-nouveau-multi-node/docker-compose/cht-couchdb-clustered.yml)

# License


witash:
Hi @1yuv please review.
tested locally and on medic's eks cluster.
upgrading from 4.21 (no nouveau) to this branch
and 5.1.0 (single nouveau) to this branch.

For the single nouveau -> multiple nouveau case, the question is what does it do with the existing index data on the first node. What happens is that the other nodes start indexing tasks that index all the documents belonging to them, but all the old index files on the first node stay there; so it has the index files from other nodes that its no longer using.

This means some downtime while new index tasks are running and some leftover index files on the first node, which we could clean up manually I guess, but probably just leave them.
Hopefully that is acceptable for the few multi-node deployments that are already on 5.x

1yuv:
>which we could clean up manually I guess, but probably just leave them.
Hopefully that is acceptable for the few multi-node deployments that are already on 5.x

Hi @witash , I think cleaning up the indexes manually and identifying what are on other nouveau nodes might be challenging. It does not make sense to keep the old indexes either.

Can you check if it is possible to have nouveau run on separate name than what we currently have until this point, and then all nodes get their own indexes and we clear up the old one entirely?

witash:
> Can you check if it is possible to have nouveau run on separate name than what we currently have until this point, and then all nodes get their own indexes and we clear up the old one entirely?

I think this would mean rebuilding the index (and therfore downtime since this change couldn't be staged) for all helm deployments, not just the minority that are multi-node.

1yuv (approved):
Changes look good to me. Approving this with the documentation for clearing old indexes to be prepared separately.

PR #11132, opened 2026-06-03
fix(#10904): preserve untouched inline-binary fields on edit

benkags:
# Description
 On edit an inline [type=binary] field loads empty (its value is a reference), so re-saving blanked it; the value is now stashed in a sidecar and restored.

# Code review checklist

# License


webapp/src/ts/services/enketo.service.ts
```diff
@@ -526,39 +528,58 @@ export class EnketoService {
...
+    const attachInlineBinary = (elem, file, type, alreadyEncoded) => {
+      const ownerDoc = resolveOwnerDoc(elem);
+      const reference = computeInlineBinaryReference(elem, ownerDoc);
       this.attachmentService.add(ownerDoc, `${USER_FILE_PREFIX}${reference}`, file, type, alreadyEncoded);
       return reference;
     };
 
+    // Recover an untouched field's prior reference from its owner doc's existing
+    // fields; naming-scheme independent, unlike the position-derived name.
+    const recoverEmptyBinaryReference = (elem) => {
```
dianabarsan: @benkags can you please address this Sonar issue?

benkags: this is now resolved

tests/e2e/default/enketo/submit-inline-binary-form.wdio-spec.js
```diff
@@ -0,0 +1,46 @@
...
+    await genericForm.submitForm();
+    await commonPage.waitForPageLoaded();
+
+    // On create the inline-binary default is attached and the field rewritten to
+    // the bare reference.
+    const reportId = await reportsPage.getCurrentReportId();
+    expect(reportId).to.exist;
+    const initialReport = await utils.getDoc(reportId);
+    expect(Object.keys(initialReport._attachments)).to.deep.equal([BADGE_ATTACHMENT]);
+    expect(initialReport.fields.badge).to.equal(BADGE_REFERENCE);
```
dianabarsan: Can you also add an assertion about the attachment content? To make sure we're saving the correct string?

tests/e2e/default/enketo/submit-inline-binary-form.wdio-spec.js
```diff
@@ -0,0 +1,46 @@
...
+    await reportsPage.openReport(reportId);
+    await commonPage.accessEditOption();
+    await commonEnketoPage.setInputValue('Patient Name', 'Ada Lovelace');
+    await genericForm.submitForm();
+    await commonPage.waitForPageLoaded();
+
+    const updatedReport = await utils.getDoc(reportId);
+    expect(updatedReport.fields.patient_name).to.equal('Ada Lovelace');
+    expect(updatedReport.fields.badge).to.equal(BADGE_REFERENCE);
+    expect(updatedReport._attachments[BADGE_ATTACHMENT]).to.exist;
```
dianabarsan: Same here. Assert that the attachment content is what we expect.

tests/e2e/default/enketo/submit-inline-binary-form.wdio-spec.js
```diff
@@ -0,0 +1,46 @@
...
+const commonEnketoPage = require('@page-objects/default/enketo/common-enketo.wdio.page');
+const reportsPage = require('@page-objects/default/reports/reports.wdio.page');
+const genericForm = require('@page-objects/default/enketo/generic-form.wdio.page');
+const loginPage = require('@page-objects/default/login/login.wdio.page');
+const utils = require('@utils');
+
+const BADGE_ATTACHMENT = 'user-file-inline-binary-report/badge';
+const BADGE_REFERENCE = 'inline-binary-report/badge';
+
+describe('Submit inline-binary report form', () => {
```
dianabarsan, line 37: I ran this e2e test over the main branch, without the updated fixes, and it passed. So it isn't guarding the fix. The form bakes a base64 default into `<badge>`, so on edit the node loads non-empty and hits the pre-existing `attachInlineBinary` branch that re-attaches it and rewrites the field to the reference — which makes these assertions pass without the fix. A real
  referenced binary has an *empty* form default (the base64 only lives in the attachment), so the node loads empty and the bug manifests.

  A faithful repro can't use a default (it would re-fill on edit too); instead seed the doc and only test the edit:

  1. Form: empty-default binary node — `<badge type="binary"/>`.
  2. `utils.saveDoc` a report with `fields.badge = 'inline-binary-report/badge'` and `_attachments['user-file-inline-binary-report/badge']` (a real base64 blob, `content_type: image/png`).
  3. Open edit, change Patient Name, save.
  4. Assert `fields.badge === 'inline-binary-report/badge'`, the attachment still exists, **and its `digest`/`revpos` is unchanged** (proves the untouched field wasn't re-attached, not just that it wasn't blanked).

  Since #10904 is sub-contact photos, please also add a `db-doc=true` sub-doc variant with the same setup — that's the case this issue is actually about, and it's currently only covered by hand-crafted XML in the unit tests, never through the real Enketo round-trip.

  I traced this through mergexml — `data-attachment-ref` survives the model merge (MergeXML's `Merge` copies all attributes onto the existing model node, recursively, and `getStr` doesn't strip `data-*`), so the sidecar is the working path and it covers sub-docs too. Given that, when does `recoverEmptyBinaryReference` ever recover anything? In the
  modern edit flow (instanceData = `doc.fields`), `bindJsonToXml` sets the sidecar exactly when the value is an attachment ref — so when the sidecar is absent the value was empty/null and there's nothing to recover. The only path that reaches the DOM walk seems to be a legacy `content`-XML report whose stored XML is empty but whose `fields` still
  hold a reference. Is that a real scenario you're targeting? If not, I'd drop `recoverEmptyBinaryReference` / `getOwnerFields` / `asAttachmentRef` and the lodash `get` import — the sidecar alone fixes #10904, and the fallback is untested dead weight.

dianabarsan (changes requested):

tests/e2e/default/enketo/submit-inline-binary-form.wdio-spec.js
benkags: Thank you for the review and taking the time to test it out. Making the fixes:
- seed the docs and remove default value when testing binary type attachment preservation
- `recoverEmptyBinaryReference` redundancy - I'd initially designed this to handle sub-doc edits in the parent form. 
- adding a `db-doc=true` sub-doc e2e test variant - not sure best approach as the sub-docs-edits-creates-extra-docs fix in #11114 is not included in this feature branch - we'd have to test with the broken logic that we are fixing

benkags: I pushed the fixes @dianabarsan. I have held back on the e2e that touches no the child docs edit; since the fix is in a different PR. Let me know.

sugat009:
@benkags heads up — the "Build and test" CI on this PR is failing, but not because of your changes. It'll be fixed by #11202 once that's brought into this branch.

benkags:
Hi @sugat009 are you suggesting I bring those changes into this branch? The feature branch we are rebasing from was made before that fix.

sugat009:
@benkags yeah, not the whole branch, just Diana's node-fetch fix (#11202).

I also noticed the PR is technically still mergeable despite the failing checks. I don't think we should merge it like that though: the failing jobs mean the e2e tests we added to validate this fix never ran, so we'd be merging with no real signal on the change. Cherry-picking that one commit gets CI green and lets those specs actually exercise the fix, then we merge.

webapp/src/ts/services/contact-save.service.ts
```diff
@@ -284,6 +295,23 @@ export class ContactSaveService {
     }
   }
 
+  /**
+   * Restores an untouched inline-binary field on edit from its sidecar reference.
+   * Restoring the value is enough: finalizeDocs's referenced-attachment scan keeps
+   * the attachment.
+   */
+  private restoreUntouchedBinary(element: Element, ctx: ContactOwnerContext, sidecar?: string) {
```
sugat009, line 303: **suggestion (non-blocking):** positive sub-doc coverage for the restore path?

`restoreUntouchedBinary` resolves the owner via `resolveContactOwnerDoc` and the field path via `findFieldContainerElement`/`computeFieldPath`, which behave differently for a sibling (`db-doc="true"`) and a repeat `<child>` than for the main doc. Right now the only positive sidecar restore test is the main-doc case, and the sibling case shows up only as the negative "no sidecar" non-goal.

The routing helpers themselves are already exercised by the attach-path tests, so this isn't really a correctness worry, more a coverage gap on the exact scenario #10904 is about. Would it make sense to add a couple of karma cases (an empty sidecar-bearing binary inside a `db-doc="true"` sibling, and inside a repeat `<child>`) asserting the sub-doc field is restored and the attachment isn't orphaned? I know the full Enketo e2e round trip for sub-contacts is parked behind #11114, so this would just be the unit slice we could close now. wdyt?

webapp/src/ts/services/contact-save.service.ts
```diff
@@ -284,6 +295,23 @@ export class ContactSaveService {
...
+    const container = this.findFieldContainerElement(element, ctx);
+    const fieldPath = container && this.computeFieldPath(element, container);
+    if (fieldPath) {
+      objectPath.set(ownerDoc, fieldPath, sidecar);
+    }
+  }
+
   /** Bare attachment reference for a binary node: the element's xpath with the
    * instance root swapped for the form internal ID and the leading slash
    * dropped -> `<formId>/<xpath>/<field>`. */
```
sugat009: **nitpick (non-blocking):** comment format looks slightly out of sync with its `enketo.service.ts` twin

This reads `<formId>/<xpath>/<field>`, but the equivalent helper in `enketo.service.ts` was updated in this PR to `<formId>/<field-path>` (e.g. `group/photo`). They produce the same string, so it might be nice to keep the two comments aligned. Totally optional.

webapp/src/ts/services/enketo-translation.service.ts
```diff
@@ -146,6 +146,10 @@ export class EnketoTranslationService {
     elem.removeAttr('template');
 
     if (this.shouldSkipBinaryBind(elem, data)) {
+      // Stash the skipped reference so an untouched field can be restored on save.
+      if (this.isAttachmentRef(data)) {
```
sugat009: **nitpick (non-blocking):** `isAttachmentRef(data)` ends up running twice here

`shouldSkipBinaryBind` already calls `isAttachmentRef(data)` (line 117), and we call it again to decide whether to stash the sidecar. It's a cheap regex so there's no real cost, but if you felt like it you could compute it once (hoist `const isRef = this.isAttachmentRef(data)` and reuse it for both the skip check and the stash). Take it or leave it.

sugat009 (commented):
A few non-blocking notes below. I'm holding off on approving until CI is green, so the e2e specs actually get to run against the fix.

benkags:
Some e2e are still failing and I'm not sure if the commit was meant to fix those. @sugat009

sugat009:
@benkags a follow-up on the remaining CI reds, since I dug into them.

Good news: the node-fetch blocker is gone. The API runs on Node 22.23.1 now, which fixed the upstream regression, so the app boots and 21 of 24 checks pass, including this PR's own inline-binary e2e specs.

Of the 3 still red, one isn't a flake, it's a known test fixed on master that this branch predates:

**ci-integration-sentinel-k3d** (`mark_for_outbound > error logging`) is #11191, fixed by #11203 and #11210. #11210 reorders the test so `collectSentinelLogs` starts before `saveDoc` (the old order emits the log lines before collection begins, so the assertion sees 0), and #11203 bumps the k3d log-collect delay. Both landed June 22 to 23, after this branch was cut, so we don't have them.

The other two (`ongoing replication` and `create-user-for-contacts`) still look like genuine flakes: green on master, with timeout / UI-not-displayed signatures, so a re-run should clear those.

Given the branch is behind master on a few of these, might be simplest to merge current master into the branch to catch up in one go, rather than cherry-picking piece by piece? That pulls in #11191 and anything else we're missing. wdyt?

cc @dianabarsan (flagging since the sentinel fix #11191 is yours, in case the branch-catch-up read matches what you saw)

benkags:
Thank you @sugat009 for looking into all the tests! I believe being a feature branch @dianabarsan cut it at the 5.1.2 release - as per the process. If it's not against the feature branch process, I'm ok merging master.

webapp/src/ts/services/enketo-translation.service.ts
benkags: although this part of the implementation is fully refactored on the main PR (#10923) to master that eliminates the double call, this is now refactored here too.

webapp/src/ts/services/contact-save.service.ts
benkags: fixed

webapp/src/ts/services/contact-save.service.ts
benkags: Thanks for this suggested. Now updated, including in the main PR branch.

tests/integration/transitions/message-duplicates.spec.js
```diff
@@ -49,8 +49,9 @@ const postMessages = (messages) => {
 const getRecipient = doc => doc.tasks[0].messages[0].to;
 
 describe('message duplicates', () => {
-  after(() => utils.revertDb([], true));
-  afterEach(() => utils.revertSettings(true));
+  after(async () => {
+    await utils.revertDb([], true);
+  });
```
sugat009, line 54: This is what's failing both `ci-integration-all-*` jobs, and it's deterministic, so a re-run won't clear it. The #11203 cherry-pick dropped the per-test settings revert here. That's harmless on master (its `updateCustomSettings` is tolerant), but this branch's base still has the strict guard that throws `A previous test did not call revertSettings`, so the second test trips it on every run.

Smallest fix is to re-add the one line, scoped to just this spec (no shared-harness change, no master merge needed):

```suggestion
  });
  afterEach(() => utils.revertSettings(true));
```

That should turn both integration jobs green. The other three reds (`server` restart 502, `ongoing-replication`, `create-user`) look like flakes, so a re-run should clear those.

sugat009 (approved):

PR #11135, opened 2026-06-03
fix(#10875): sentinel wait for transitions infodoc to be valid

### LINKED ISSUE #10875: Flaky api integration tests: transitions

witash:
Another attempt to resolve the api/sentinel race.
Add invalid_rev to infodocs for synchronous transitions to mark that transitions are in progress, and the infodoc is not yet valid for that revision.
after transitions are run and the document is saved, the infodoc is saved again, with transitions and with invalid_rev removed.

A previous version also added a valid_rev with explicitly the revision that the infodoc is valid for, but removed it since it's never used; the abscence of an invalid_rev implies that the infodoc is valid for the current revision.
If saving the medic doc fails, it clears invalid_rev from the infodoc to allow retry.

There is no cleanup for other stale infodocs where invalid_rev is never cleared, but this can only happen if there's an error other than 409 saving the infodoc the second time, or if its interrupted between saving the medic doc and the infodoc.

To actually fix the race, if sentinel finds an infodoc with invalid_rev set (doesn't actually use the rev, just the presence of the flag), it waits 100ms and then tries again until the infodoc had invalid_rev removed, for a maximum of 5 times. after 5 retries, it logs an error and continues to the next change without running transitions.

`invalid_rev` is only set by `processDocs` (synchronous changes from api), and only `processChanges` (from sentinel) will wait for it. It also applies only to transitions. This is to avoid adding extra db chatter where it isn't really needed.

A previous commit also add invalid_rev and waiting to `processChanges`, but removed because it does not actually fix anything; sentinel does not support concurrency, and there isn't currently a case where api should wait for sentinel to finish.

Closes #10875


dianabarsan:
This one sounds like a pretty sweet solution! nice work!

witash:
I think this is ready for review.
Note that there are some changes to that appear unrelated but were necessary for code quality

shared-libs/infodoc/src/infodoc.js
```diff
@@ -151,33 +151,61 @@ const updateTransition = (change, transition, ok) => {
...
+    infoDoc.transitions = (change.info && change.info.transitions) || {};
+    // If there is a mid-write flag set, clear it now to mark the infoDoc valid again
+    if (clearInvalid) {
+      delete infoDoc.invalid_rev;
+    }
+  }, change.info);
 };
 
 const saveCompletedTasks = (id, infodoc, completedTasks = []) => {
-  return saveProperty(id, infodoc, 'completed_tasks', completedTasks);
```
witash, line 159: refactored this a little bit because adding setInvalidRev and clearInvlaidRev otherwise led to duplicating the retry on 409 block

shared-libs/transitions/src/transitions/index.js
```diff
@@ -235,46 +257,61 @@ const canRun = ({ key, change, transition }) => {
  * did nothing and saving is unnecessary.  If results has a true value in
  * it then a change was made.
  */
-const finalize = ({ change, results }, callback) => {
```
witash, line 238: refactor this a little bit to convert to promise; otherwise splitting it into the two different branches became difficult to read with many nested `.then`s

witash:
last minute changed from `invalid_rev` to `transitions_started` with a timestamp, since `invalid_rev` will always be null, `_rev` isn't set by the time the infoDoc is saved. It still uses the presence of the field to decide whether to wait, rather than the value, but adding the timestamp just for information/debugging (hopefully there will never be a need to debug).

shared-libs/infodoc/src/infodoc.js
```diff
@@ -151,33 +151,64 @@ const updateTransition = (change, transition, ok) => {
...
+// Fetch the infodoc. If it is missing, return `fallback` (to be created) when provided;
+// otherwise the 404 is raised.
+const fetchInfoDoc = async (id, fallback) => {
   try {
-    updatedInfoDoc = await db.sentinel.get(getInfoDocId(id));
-    updatedInfoDoc[property] = (infodoc && infodoc[property]) || defaultValue;
+    return await db.sentinel.get(getInfoDocId(id));
   } catch (err) {
-    if (err.status !== 404) {
+    if (err.status !== 404 || !fallback) {
```
dianabarsan: I think this would be easier to understand if reversed: 
```suggestion
if (err.status === 404 && fallback) {
  return fallback;
}

throw err;
```

There also seems to be a behavior change here, where if fallback is not passed, the 404 would throw compared to before when it did not throw. Is there a reason for this shift?

shared-libs/infodoc/src/infodoc.js
```diff
@@ -151,33 +151,64 @@ const updateTransition = (change, transition, ok) => {
   };
 };
 
-const saveTransitions = change => {
-  return saveProperty(change.id, change.info, 'transitions', {});
+const saveTransitions = (change, clearStarted = false) => {
+  return modifyInfoDoc(change.id, infoDoc => {
```
dianabarsan: Because we're passing the `fallback` parameter here, it would be a lot more readable if the modify function was not inlined: 
```suggestion
const modify = infodoc => { .... };
return modifyInfoDoc(change.id, modify, change.info);
```

shared-libs/infodoc/test/infodoc.js
```diff
@@ -604,6 +604,64 @@ describe('infodoc', () => {
...
+  describe('markTransitionsStarted / clearTransitionsStarted', () => {
+    it('markTransitionsStarted marks the infodoc mid-write with a timestamp', () => {
+      const info = { _id: 'some-info', doc_id: 'some' };
+      sinon.stub(db.sentinel, 'get').resolves(info);
+      sinon.stub(db.sentinel, 'put').resolves();
+
+      return lib.markTransitionsStarted('some').then(() => {
+        assert.deepEqual(db.sentinel.get.args[0], ['some-info']);
+        const saved = db.sentinel.put.args[0][0];
+        assert.isString(saved.transitions_started);
```
dianabarsan: can you complete these assertions to show that no other fields are added or modified on the infodocs? Same comment for the other tests lacking this as well.

shared-libs/infodoc/test/infodoc.js
```diff
@@ -604,6 +604,64 @@ describe('infodoc', () => {
...
+      const info = { _id: 'some-info', doc_id: 'some', transitions_started: '2026-01-01T00:00:00.000Z' };
+      sinon.stub(db.sentinel, 'get').resolves(info);
+      sinon.stub(db.sentinel, 'put').resolves();
+
+      return lib.clearTransitionsStarted('some').then(() => {
+        assert.isUndefined(db.sentinel.put.args[0][0].transitions_started);
+      });
+    });
+
+    it('retries on 409 conflict', () => {
```
dianabarsan: Thanks for adding this test. 
I believe that we now have infinite retries? If someone adds a limit to the number of retries, this test would still pass. I think it would be useful to capture the unlimited retry count in this test somehow. MAybe adding 100 calls would suffice?

shared-libs/transitions/src/transitions/index.js
```diff
@@ -45,13 +45,34 @@ const AVAILABLE_TRANSITIONS = [
 const transitions = [];
 let loadErrors = false;
 
+const MAX_INFODOC_WAIT = 5;
+const INFODOC_WAIT_INTERVAL = 100;
+
+const isInfoDocMidWrite = infoDoc => infoDoc && infoDoc.transitions_started !== undefined;
+
+const getConsistentInfoDoc = (change, retriesLeft) => {
```
dianabarsan: can you please convert this to async await?

shared-libs/transitions/src/transitions/index.js
```diff
@@ -235,46 +257,61 @@ const canRun = ({ key, change, transition }) => {
  * did nothing and saving is unnecessary.  If results has a true value in
  * it then a change was made.
  */
-const finalize = ({ change, results }, callback) => {
+const finalize = ({ change, results, manageInfoDocRev = false }, callback) => {
```
dianabarsan: I think `manageInfoDocRev` is a relic of a previous implementation, that used invalid_rev? 
Please rename this to `markStarted` or similar. 

Now that I think about it further, _not_ having this parameter and always blocking sentinel from processing docs with started transitions is the first step in scaling Sentinel to multiple service instances. No change needed, because we don't have that on the roadmap, and the extra function is an ok price to pay for 1 additional doc write.

shared-libs/transitions/test/integration/transitions.js
```diff
@@ -568,10 +578,11 @@ describe('functional transitions', () => {
...
         infodocSaves = db.sentinel.put.args.filter(args => args[0].doc_id === savedDocs[0]._id);
-        assert.equal(infodocSaves.length, 1);
-        assert.equal(infodocSaves[0][0].transitions.update_clinics.ok, true);
-        assert.equal(infodocSaves[0][0].transitions.update_sent_by.ok, true);
-        assert.equal(infodocSaves[0][0].transitions.conditional_alerts.ok, true);
+        assert.equal(infodocSaves.length, 2);
+        let txnSave = infodocSaves.find(args => args[0].transitions)[0];
+        assert.equal(txnSave.transitions.update_clinics.ok, true);
+        assert.equal(txnSave.transitions.update_sent_by.ok, true);
+        assert.equal(txnSave.transitions.conditional_alerts.ok, true);
```
dianabarsan, line 732: Can you please add assertions on the other properties in `txnSave`? 
Also, can you please add assertions on the second save? 
Ideally the tests would show the exact contents of the file to make it very clear what is stored and what is removed.

shared-libs/transitions/test/unit/finalize-transition.js
```diff
@@ -22,41 +22,54 @@ describe('finalize transition', () => {
     );
   });
 
-  it('save is called if transition results have changes', done => {
+  it('save is called if transition results have changes (API branch manages rev)', done => {
```
dianabarsan: re: manages rev
This needs renaming.

shared-libs/transitions/src/transitions/index.js
```diff
@@ -45,13 +45,34 @@ const AVAILABLE_TRANSITIONS = [
...
+};
+
 // applies all loaded transitions over a change
 const processChange = (change, callback) => {
   lineage
     .fetchHydratedDoc(change.id)
     .then(doc => {
       change.doc = doc;
-      return infodoc.get(change).then(infoDoc => {
+      return getConsistentInfoDoc(change, MAX_INFODOC_WAIT).then(infoDoc => {
```
dianabarsan: This is the central change in the PR , but I don't see any test covering `getConsistentInfoDoc` itself — i.e. sentinel waiting while `transitions_started` is set and proceeding once it clears, and the 5-retry exhaustion path that logs the warning and skips. The new tests all cover the API write side and the infodoc plumbing. Could you add a `processChange` test that exercises the wait/retry/skip directly?

shared-libs/transitions/src/transitions/index.js
```diff
@@ -235,46 +257,61 @@ const canRun = ({ key, change, transition }) => {
...
+  logger.info(`saved changes on doc ${change.id} seq ${change.seq}`);
+  await infodoc.saveTransitions(change);
+  return result;
+};
 
+// API processing: mark the infodoc mid-write (transitions_started) around the doc write so a concurrent
+// sentinel read detects it and waits. The marker is cleared as part of the transitions write on
+// success, or rolled back on any failure after it's set.
+const saveForApi = async change => {
+  await infodoc.markTransitionsStarted(change.id);
```
dianabarsan, line 319: `markTransitionsStarted` passes no fallback to `modifyInfoDoc`, so a missing infodoc raises the 404 (the old `saveProperty` would have created it). Since it's awaited outside the try, this would block the medic doc save entirely. `processDocs` runs `bulkUpdate` first so the infodoc should exist — but is that guaranteed for every path into `saveForApi`?

dianabarsan (changes requested):
Nice work! Some comments below.

shared-libs/infodoc/src/infodoc.js
witash: reversed the condition for readability.

the intention was to not change existing behavior for the existing functions, but allow the new functions to raise errors if the infodoc doesn't exist instead of silently creating it.

But it's more consistent with the existing functions to silently create a blank infoDoc, so changed to do that instead.

shared-libs/transitions/src/transitions/index.js
witash: changed the name.
a previous commit did always block, but yes, removed to not add extra database chatter where it currently isn't used

shared-libs/infodoc/test/infodoc.js
witash: added deepEqual assertions

shared-libs/transitions/src/transitions/index.js
witash: infodoc should exist before this `saveForApi` is called.
also changed `markForTransitions` to add a blank infoDoc if it doesn't exist.

shared-libs/transitions/src/transitions/index.js
witash: added 'should wait for transitions_started to clear before processing the change' and 'should skip processing the change when transitions_started never clears'.
used fake time which makes them more complicated but hopefully avoids flakiness

witash:
hi @dianabarsan please re-review, I added a commit with some new changes;

There is another related, but not the same, race condition that is pre-existing and the changes in timing apparently now trigger it reliably in tests.

It's a little complicated, here is claude's summary:

> The infodocs integration test finds and migrates data from the medic infodoc failed reliably: it expected a migrated initial_replication_date of 1000 but got a fresh "now" timestamp.
>
> Root cause was a race between two unsynchronized writers to the same sentinel infodoc:
> - API (recordDocumentWrite, fired async after the doc write) creates the infodoc with initial_replication_date = now.
> - Sentinel (resolveInfoDocs) migrates the legacy medic-db infodoc, setting initial_replication_date to the original value.
>
> When sentinel's migrating bulkUpdate conflicted with API's write, the conflict handler adopted the fresh "now" value, discarding the migrated date.

So its related to migrating legacy infodocs that were in the medic, which I think means the racing code and tests can just be removed? Because we can expect all infodocs to be in the sentinel db by now, so is there really still a need to check the medic db just in case?

there are still two writes to the same infodoc, but without the conflicting data from legacy infodocs, that doesn't matter; so no need to overcomplicate the fix.

this isn't the same to 10875, but I think it's related enough to also include the fix in this branch to get the tests to pass.

shared-libs/infodoc/test/infodoc.js
witash: stubbed db.sentinel.put to reject with 409 on the first 100 calls, then resolve on the 101st

shared-libs/infodoc/src/infodoc.js
```diff
@@ -48,95 +45,35 @@ const resolveInfoDocs = (changes, writeDirtyInfoDocs) => {
...
+      const dirtyInfoDocs = [];
 
-            infoDoc.transitions = change.doc && change.doc.transitions;
-
-            if (!collectedInfoDoc) {
-              infoDocs.push(infoDoc);
-            }
+      missingTransitions.forEach(infoDoc => {
+        const change = changeForInfoDoc(infoDoc._id);
+        infoDoc.transitions = infoDoc.transitions || change.doc?.transitions;
```
dianabarsan: The history of the transitions changes was: 
1) 2.x - transitions were saved on the document itself (doc.transitions)
2) 3.x - migrated from storing transitions on document to storing transitions on infodoc, and stored in medic database 
3) 3.x - infodoc transition from medic database to medic-sentinel database

So here, where you're deleting the migration of infodoc from medic to medic-sentinel but keeping the possibility of having transitions saved on the doc itself is like removing the middle migration and keeping the first one. If we assume that we are past infodocs in medic database, that we can absolutely assume that transitions are no longer properties of medic documents. 

We either delete all or none.

shared-libs/infodoc/src/infodoc.js
```diff
@@ -151,33 +88,62 @@ const updateTransition = (change, transition, ok) => {
   };
 };
 
-const saveTransitions = change => {
-  return saveProperty(change.id, change.info, 'transitions', {});
+const saveTransitions = (change, clearStarted = false) => {
+  const modify = infoDoc => {
+    infoDoc.transitions = change.info?.transitions || {};
+    // Clear the in-progress marker in the same write that commits the transitions (API branch only)
+    if (clearStarted) {
```
dianabarsan: If nobody triggers this (so somehow API messes up and crashes or some other thing) and the `transitions_started` remains on the document, this will mean that it will permanently be skipped by Sentinel.

Can we add a fallback where sentinel checks the time when `transitions_started` was set, and if it's over some prescribed interval, it will delete it itself and resume transition processing for this doc.

dianabarsan: It feels a bit redundant that we have a `clearStarted` parameter but also a `clearTransitionsStarted` function, both deleting the same property. is it possible for this function to call `clearTransitionsStarted` instead of this annex to the  modify function?

dianabarsan: I saw that just inlining the next call generates an extra round trip to the database. I asked claude about it and: 

```aiignore
The cleaner option — drop the param without reusing clearTransitionsStarted and without an extra write: have saveTransitions always delete the marker.

  This is safe at all three callers: the two non-API callers have no marker, so delete is a harmless no-op; the API caller gets exactly today's behavior. It's strictly better than both alternatives — one write, no param, and it enforces a clean invariant ("transitions committed ⇒ no in-progress marker"). It even opportunistically clears a stale
  marker if one ever reaches this path, which slightly mitigates the crash-poisoning gap from finding #2.

  The one tradeoff: saveTransitions now always touches transitions_started, so it's coupled to the marker concept even on the sentinel path. That coupling is mild and arguably correct — but if witash wants saveTransitions to stay marker-agnostic, then keeping the explicit clearStarted param is the more honest expression of intent, and "redundant"
  isn't quite the right framing since the param is what folds two mutations into a single write. Worth raising as a suggestion either way; I'd lead with the unconditional-delete version.

```

Maybe worth considering.

shared-libs/infodoc/test/infodoc.js
```diff
@@ -604,6 +523,80 @@ describe('infodoc', () => {
...
+
+    it('clears transitions_started when clearStarted is set', () => {
+      const info = { _id: 'some-info', doc_id: 'some', transitions_started: '2026-01-01T00:00:00.000Z' };
+      const change = { id: 'some', info: { transitions: { one: { ok: true } } } };
+      sinon.stub(db.sentinel, 'get').resolves(info);
+      sinon.stub(db.sentinel, 'put').resolves();
+
+      return lib.saveTransitions(change, true).then(() => {
+        const saved = db.sentinel.put.args[0][0];
+        // transitions are written and the mid-write marker removed; nothing else is touched
```
dianabarsan: ~claude~ captain obvious has spoken

shared-libs/infodoc/test/infodoc.js
```diff
@@ -604,6 +523,80 @@ describe('infodoc', () => {
...
+        assert.deepEqual(db.sentinel.get.args[0], ['some-info']);
+        const saved = db.sentinel.put.args[0][0];
+        assert.isString(saved.transitions_started);
+        assert.isNotNaN(Date.parse(saved.transitions_started));
+        // only the marker is added, no other fields are changed
+        assert.deepEqual(saved, {
+          _id: 'some-info',
+          doc_id: 'some',
+          transitions_started: saved.transitions_started,
+        });
```
dianabarsan: This is a weird assertion :D asserting a property of a doc is equal to a property of a doc. 

We could use `.excludingEvery` here and omitting the `transitions_started` assert specifically. or even better, we should mock time and assert exactly on the format we expect.

shared-libs/infodoc/test/infodoc.js
```diff
@@ -604,6 +523,80 @@ describe('infodoc', () => {
...
+      });
+    });
+
+    it('retries on 409 conflict indefinitely (no retry limit)', () => {
+      const info = { _id: 'some-info', doc_id: 'some' };
+      sinon.stub(db.sentinel, 'get').resolves(info);
+      const put = sinon.stub(db.sentinel, 'put');
+      // conflict on the first 100 attempts, succeed on the 101st - a retry limit below this would fail
+      for (let i = 0; i < 100; i++) {
+        put.onCall(i).rejects({ status: 409 });
```
dianabarsan: I believe you don't need this for loop. You can just say: `put.rejects()` and add the `put.onCall(100).resolves();` and it will just work.

shared-libs/transitions/test/integration/transitions.js
```diff
@@ -378,6 +389,76 @@ describe('functional transitions', () => {
         done();
       });
     });
+
+    // fake timers so the 100ms retry waits don't cost real time or flake near mocha's timeout;
```
dianabarsan: Please remove this comment.

shared-libs/transitions/test/integration/transitions.js
```diff
@@ -378,6 +389,76 @@ describe('functional transitions', () => {
...
+    const RETRY_INTERVAL = 100;
+    const MAX_RETRIES = 5;
+    const processChangeWithFakeTimers = async (change) => {
+      const clock = sinon.useFakeTimers();
+      const processed = new Promise((resolve, reject) => {
+        transitions.processChange(change, (err, result) => (err ? reject(err) : resolve(result)));
+      });
+      // one extra tick past MAX_RETRIES is harmless and guarantees the change settles either way
+      for (let i = 0; i <= MAX_RETRIES; i++) {
+        await clock.tickAsync(RETRY_INTERVAL);
```
dianabarsan: To make the test complete, please add assertions of numbers of calls after every tick. Otherwise three calls can be chained within the same tick and this test will not detect it.

shared-libs/transitions/test/integration/transitions.js
```diff
@@ -568,10 +649,25 @@ describe('functional transitions', () => {
...
         infodocSaves = db.sentinel.put.args.filter(args => args[0].doc_id === savedDocs[0]._id);
-        assert.equal(infodocSaves.length, 1);
-        assert.equal(infodocSaves[0][0].transitions.update_clinics.ok, true);
-        assert.equal(infodocSaves[0][0].transitions.update_sent_by.ok, true);
-        assert.equal(infodocSaves[0][0].transitions.conditional_alerts.ok, true);
+        assert.equal(infodocSaves.length, 2);
+        // the first write only marks the infodoc as mid-write, with no transitions yet
+        let startedSave = infodocSaves.find(args => args[0].transitions_started)[0];
+        assert.isString(startedSave.transitions_started);
+        assert.deepEqual(startedSave, {
```
dianabarsan, line 718: Please use `excludingEvery` (or `excluding`) here instead of asserting equality of the same object property. Applies to the other test as well. 
Ideally please use fakeTimers everywhere and actually assert on the value of `transitions_started`.

dianabarsan (changes requested):
Thanks for the changes. Looking good!. I left some comments, but nothing major.

shared-libs/infodoc/src/infodoc.js
witash: removed the code that copied transitions form medic documents

shared-libs/infodoc/src/infodoc.js
witash: have saveTransitions always clear transitions_started is the most straightforward

shared-libs/transitions/test/integration/transitions.js
witash: added these assertions

shared-libs/infodoc/test/infodoc.js
witash: the test now mocks time with sinon.useFakeTimers({ now: new Date('2026-01-01T00:00:00.000Z').valueOf() }) and asserts the exact stored value

shared-libs/transitions/test/integration/transitions.js
witash: this test also now mocks time with sinon.useFakeTimers and asserts the exact value

shared-libs/infodoc/src/infodoc.js
witash: infodocs where `transitions_started` is more than STALE_INTERVAL (2 minutes) old will have `transitions_started` cleared and be processed anyway.

shared-libs/transitions/src/transitions/index.js
```diff
@@ -45,20 +45,49 @@ const AVAILABLE_TRANSITIONS = [
...
+  const infoDoc = await infodoc.get(change);
+  if (!isInfoDocMidWrite(infoDoc) || isMidWriteStale(infoDoc) || retriesLeft <= 0) {
+    return infoDoc;
+  }
+  await new Promise(resolve => setTimeout(resolve, INFODOC_WAIT_INTERVAL));
+  return getConsistentInfoDoc(change, retriesLeft - 1);
+};
+
 // applies all loaded transitions over a change
 const processChange = (change, callback) => {
```
dianabarsan: This function is now a weird mix between callbacks, then chains and async await. Can you please refactor it to reduce the use async exclusively?

shared-libs/transitions/src/transitions/index.js
```diff
@@ -45,20 +45,49 @@ const AVAILABLE_TRANSITIONS = [
...
+            .clearTransitionsStarted(change.id)
+            .catch(err => logger.error(`transitions: error clearing stale marker on doc ${change.id}: %o`, err));
+          delete infoDoc.transitions_started;
+        } else if (isInfoDocMidWrite(infoDoc)) {
+          logger.warn(
+            `transitions: infodoc for ${change.id} still mid-write after ${MAX_INFODOC_WAIT} retries, skipping`
+          );
+          return callback();
+        }
         change.info = infoDoc;
```
dianabarsan: I think it would be easier to read if we exclude the else clause and try to return early. 
Also, cleaner to extract clearTransitionsStarted separately.

```js
if (isInfoDocMidWrite(infodoc) {
  if (!isMidWriteStale(infodoc)) {
    return callback();
  }
  clearTransitionsStarted(infodoc);
}
change.info = infoDoc;
// ....
const clearTransitionsStarted = async (change) => {
  try {
    await infodoc.clearTransitionsStarted(change.id)
  } catch (err) {
     logger.error(`transitions: error clearing stale marker on doc ${change.id}: %o`, err));
  }
  delete infoDoc.transitions_started;
};
```


Additional question: i understand that catching the indofoc write is protective, but since we have an infinite retry for conflicts, what would be errors that can be thrown when writing the infodoc that should not bubble up the call chain? 

Additional question: there is only a one-uppercase-character difference between the library `infodoc` and the local variable `infoDoc`. This produces quite a bit of visual confusion when reading. Can you please create more differentiation in these two variable names?

shared-libs/transitions/src/transitions/index.js
```diff
@@ -107,10 +136,11 @@ const processDocs = docs => {
...
             change.doc = docs.find(doc => doc._id === change.id);
-            saveDoc(change, (err, result) => {
-              callback(null, err || result);
-            });
-          });
+            saveDoc(change).then(
+              result => callback(null, result),
+              err => callback(null, err)
+            );
+          }, { markStarted: true });
```
dianabarsan: I find it quite hard to read which function receives the `markStarted` parameter. Please consider moving the second function parameter as it's own constant so the arguments passed are clear to read.

shared-libs/transitions/test/integration/transitions.js
```diff
@@ -378,6 +389,129 @@ describe('functional transitions', () => {
...
+      await clock.tickAsync(RETRY_INTERVAL);
+      assert.equal(get.callCount, 2);
+      await clock.tickAsync(RETRY_INTERVAL);
+      assert.equal(get.callCount, 3);
+
+      const result = await processed;
+      assert.isUndefined(result);
+      // marker cleared on the 3rd read, so no further retries
+      assert.equal(get.callCount, 3);
+      // no transition matched, so nothing is saved, but the change was processed (not skipped)
```
dianabarsan: how do we know? is there an assertion here that checks that?

shared-libs/transitions/test/integration/transitions.js
```diff
@@ -568,10 +702,24 @@ describe('functional transitions', () => {
...
         // first doc is updated by 3 transitions
         infodocSaves = db.sentinel.put.args.filter(args => args[0].doc_id === savedDocs[0]._id);
-        assert.equal(infodocSaves.length, 1);
-        assert.equal(infodocSaves[0][0].transitions.update_clinics.ok, true);
-        assert.equal(infodocSaves[0][0].transitions.update_sent_by.ok, true);
-        assert.equal(infodocSaves[0][0].transitions.conditional_alerts.ok, true);
+        assert.equal(infodocSaves.length, 2);
+        // the first write only marks the infodoc as mid-write, with no transitions yet
+        let startedSave = infodocSaves.find(args => args[0].transitions_started)[0];
+        assert.isString(startedSave.transitions_started);
```
dianabarsan: If we stubbed timers, then these assertions could be inclusive to transitions_started, the test would be leaner and assertions more precise.

shared-libs/infodoc/src/infodoc.js
```diff
@@ -35,108 +34,37 @@ const resolveInfoDocs = (changes, writeDirtyInfoDocs) => {
...
-          // could be that there was one without transition data back in sentinel
-          missing.forEach(missingDoc => {
-            const docId = getDocId(missingDoc._id);
-
-            const collectedInfoDoc = infoDocs.find(i => i._id === missingDoc._id);
-            const change = changes.find(change => change.id === docId);
-            const infoDoc = collectedInfoDoc || blankInfoDoc(docId, !change.doc._rev && Date.now());
-
-            infoDoc.transitions = change.doc && change.doc.transitions;
+      const { valid, missing } = splitInfoDocRows(results);
```
dianabarsan, line 52: So glad to see this code go!! 
Can you please create a note on the issue or somewhere that this PR makes this change so we cover it fully in the release notes? Even if we're overdue reverting this, we should be explicit about removing it.

dianabarsan (changes requested):
Very nice! 
I left a few minor comments about code readability.

shared-libs/transitions/src/transitions/index.js
witash: processChange is now async/await with a single try/catch, no then chains. It still takes a callback: dropping it here would mean also converting applyTransitions/applyTransition/finalize and all their callers and tests. Which I did try but it added way too much to this PR, leaving the callback but removing the Promise chains felt like the right balance.

shared-libs/transitions/src/transitions/index.js
witash: The callback is now a named `onTransitionsApplied` const, so the call reads `applyTransitions(change, onTransitionsApplied, { markStarted: true })`.

shared-libs/transitions/test/integration/transitions.js
witash: Added a spy on applyTransitions: callCount 1 here, 0 in the skip test, 1 in the stale test. That's the assertion that actually separates the two outcomes, and the other two tests made the same unbacked claim, so they get it too.

shared-libs/transitions/test/integration/transitions.js
witash: Faked the clock, so all four marker assertions are now a plain deepEqual including transitions_started, instead of isString plus deepEqualExcluding.

shared-libs/transitions/src/transitions/index.js
witash: Restructured to early returns, no else clause, and extracted the clearing into its own function. Named it clearStaleMarker` rather than `clearTransitionsStarted`, so it isn't the same as `infodoc.clearTransitionsStarted`.

While doing that it seemed like the stale/retry handling belongs in `getConsistentInfoDoc` rather than at the call site, so I moved it there.

On the error handling, there's nothing here that shouldn't bubble up. Conflicts already retry forever inside the lib, so anything reaching this a real failure. Removed the catch so errors now go to `processChange`'s catch, which logs and calls the callback with error.

Renamed the local variable to `infoDocForChange`, here and in `getConsistentInfoDoc`.

shared-libs/infodoc/src/infodoc.js
witash: added a comment to the issue https://github.com/medic/cht-core/issues/10875#issuecomment-5141504938

dianabarsan (approved):
sweet stuff. very nice work!

PR #11138, opened 2026-06-04
fix(#9491): fail fast with helpful error when nginx container fails to start

### LINKED ISSUE #9491: e2e tests should fail gracefully if the ports are not available (or find a new port)

sugat009:
# Description

Fixes #9491. This continues the work by @SnehaChaursia in #10816, which became inactive. See that PR for the original description and the full change.

This revision only adds the fixes for the review feedback left on #10816:

- **Detect the port conflict case via `State.Error`** (the issue #9491 scenario). The original check looked only at `Status`, which never catches this case: a failed port bind leaves the container in `created` (not `exited`/`dead`), so the old code burned all 30 retries before a generic timeout fired. Inspecting the full `State` and checking `State.Error` fails fast with the real reason.
- Surface the underlying `docker inspect` error when the container is missing, and include `ExitCode` in failure messages.
- De-duplicate the repeated port hint into a single constant.

# Code review checklist
- [x] Tested: verified the new branching end to end against real Docker container states (port conflict `created`, `running`, `exited`, missing container)
- [x] Backwards compatible: test only Docker path; k3d unchanged (`isDocker()` guard)

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:fix/9491-e2e-nginx-fail-fast-v2/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:fix/9491-e2e-nginx-fail-fast-v2/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:fix/9491-e2e-nginx-fail-fast-v2/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/tests/utils/index.js
+++ b/tests/utils/index.js
@@ -949,6 +949,10 @@ const listenForApi = async () => {
   throw new Error('API failed to start after 3 minutes');
 };
 
+const NGINX_PORT_HINT =
+  'Ports 80 and/or 443 are often already in use; stop the other service or set NGINX_HTTP_PORT ' +
+  'and NGINX_HTTPS_PORT (see tests/constants.js for HTTPS).';
+
 const waitForNginxContainerRunning = async () => {
   if (!isDocker()) {
     return;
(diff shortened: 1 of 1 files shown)


jkuester (approved):
LGTM!

PR #11139, opened 2026-06-04
feat(#6615): adds archiving

### LINKED ISSUE #6615: Consider moving outdated documents to "cold storage" (Hosting TCO)
**Describe the performance issue**
With saving tasks to disk and because tasks documents are replicated, an evergrowing number of tasks for every user will linerarly make _changes replication requests slower.
Even if tasks are purged and don't cause performance issues on devices, they're still part of the server-side view queries and _changes requests.
As time goes on and configurations change, we could end up with many tasks that have been in a terminal state for a long time and have no use in being replicated and wighing in the replication process.

After <insert period> (maybe 6 months or one year) since a specific task has been updated to a terminal state, we could move it to "cold storage" (for example `medic-users-meta` or a new database) and permanently (non-tombstone) delete if from medic db.

We could only move tasks that are Cancelled into "cold storage", not all that are in terminal state.
There could be no "cold storage" at all.

**Additional context**
It's hard to ascertain overtime task dynamic, since few projects have upgraded to 3.8+ and even those have done so fairly recently.
We should evaluate this in a few months (or 1 years) time and use real world statistics to decide.

dianabarsan:
# Add document archiving (cold storage)

  Adds an end-to-end archiving pipeline that moves documents out of the `medic` db
  into a new `medic-archive` db, leaving no trace in the changes feed.
  
  Closes #6615

  ## How it works

  - **`POST /api/v1/archive`** (db admins only) accepts a `text/csv` body of doc ids —
    one per line, streamed, capped at 32MB. Ids are split into archive job docs of at
    most 100,000 ids each (stored in `medic-sentinel` with the id list as an attachment),
    so a mid-payload failure leaves already-created jobs queued and retrying is safe.
  - **Sentinel** processes the job queue on a configurable schedule (new `archive`
    settings key: `text_expression`/`cron` plus an optional `duration` bound; unbounded
    when omitted). For each batch it:
    1. copies archivable docs (contacts, reports, tasks, targets) with attachments to
       `medic-archive` (`new_edits: false`),
    2. records an `archived: true` audit entry per doc (`@medic/audit.recordArchiving`),
    3. purges the docs and their sentinel info docs — every leaf revision, including
       conflicts — so archived docs vanish from the changes feed.
    Jobs are resumable via a cursor, record their errors, and are quarantined after 20
    failed attempts instead of blocking the queue. Critical view indexes are re-warmed
    every 10 batches so archiving doesn't outpace the indexers.
  - **Replication** treats archived docs as deletions, so already-synced clients drop
    them on their next sync.

  ## Also in this PR

  - New `@medic/archiving-utils` shared lib (job attachment encode/decode).
  - `MAX_REQUEST_SIZE` and the `archive:` id prefix centralized in `@medic/constants`.
  - Duration parsing centralized in sentinel's `lib/scheduling.js`; reminders now use it.

  ## Testing

  - Unit, integration (api + sentinel schedules), and e2e coverage included.
  - Manually verified on a local instance with 1,000,000 docs: queue drained in ~28
    minutes with zero job errors; verified 0 archived ids remain in the changes feed,
    all info docs purged, and audit entries present for every archived id.

## AI disclosure

Heavily used Claude, especially for testing, verifications, and late-development checks and refactors. 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6615-cold-storage/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6615-cold-storage/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:6615-cold-storage/docker-compose/cht-couchdb-clustered.yml)

# License


dianabarsan:
I ran an e2e test locally to archive 1 million docs:

## Result

| Metric | Value |
|---|---|
| Ids submitted | 1,000,000 |
| Jobs created | 10 × 100,000 ids |
| Docs archived | 1,000,000 (`medic-archive` doc_count matches exactly) |
| Started | 15:18 local (12:18 UTC) |
| Finished | 15:45:42 local |
| Duration | ~28 minutes (~36k docs/min) |
| Job errors / quarantined jobs | 0 / 0 |

## Setup

- My local PC 
- Generated 2.5mil docs with test-data-generator

## Progress (polled every 2 min; cursor = sum over remaining jobs)

```
15:18  10 jobs remain, cursor  10000/1000000
15:20  10 jobs remain, cursor  80000/1000000
15:22   9 jobs remain, cursor  50000/900000
15:24   8 jobs remain, cursor  20000/800000
15:26   8 jobs remain, cursor  90000/800000
15:28   7 jobs remain, cursor  60000/700000
15:30   6 jobs remain, cursor  27000/600000
15:32   6 jobs remain, cursor  99000/600000
15:34   5 jobs remain, cursor  71000/500000
15:36   4 jobs remain, cursor  50000/400000
15:38   3 jobs remain, cursor  22000/300000
15:40   2 jobs remain, cursor      0/200000
15:42   2 jobs remain, cursor  75000/200000
15:44   1 job  remains, cursor  47000/100000
15:46   queue drained
```

## Post-run verification (all passed)

- **Changes feed clean**: dumped the full `medic/_changes` feed (1,444,887 entries) and
  intersected with all 1,000,000 archived ids — **0 matches**, including as
  `deleted: true` tombstones (`doc_del_count` is 6, all pre-existing). Purged docs left
  no trace on the replication surface.
- **Info docs purged**: queried `medic-sentinel/_all_docs` for all 1,000,000 `<id>-info`
  keys — **0 still alive**.
- **Audit entries saved**: every one of the 1,000,000 ids has a doc in `medic-audit`
  (0 missing), and a 10,000-doc random sample all carry an `archived: true` history
  entry.

Every poll showed `error_count 0, failed 0`; all 10 job docs deleted themselves on completion.

dianabarsan:
Hi @jkuester 

I believe this is ready to be reviewed. After our last conversation, I added an e2e test that un-archives a doc to show how that workflow looks. 

Please let me know what you think.

api/src/routing.js
```diff
@@ -796,6 +797,8 @@ app.put(
   credentials.put
 );
 
+app.post('/api/v1/archive', archive.create);
```
jkuester, line 800: This might be a bit off-the-wall, but would it make sense for this endpoint to be `/api/v2/bulk-delete`?  Seems relevant that we already have the `bulk-delete` endpoint and it is conceptually very similar to this one....

I don't think I am strongly convinced we should do this, but figured it was worth raising the discussion.

sentinel/src/lib/archiving.js
```diff
@@ -0,0 +1,264 @@
...
+    if (!docsToArchive.length) {
+      continue;
+    }
+    await db.archive.bulkDocs(docsToArchive, { new_edits: false });
+    archivedIds.push(...docsToArchive.map(doc => doc._id));
+  }
+
+  await audit.recordArchiving(archivedIds, date);
+  await purgeDocs(db.sentinel, archivedIds.map(id => `${id}-info`));
+  await purgeDocs(db.medic, archivedIds);
```
jkuester, line 104: Have we thought about what this is going to look like from the perspective of cht-sync?  I am not sure if I have anything in particular I am worried about, but I want to be sure we consider how archiving might get reflected in postgres....

api/src/services/replication/replication.js
```diff
@@ -53,6 +58,9 @@ const getDocIdsToDelete = async (userCtx, docIds) => {
   const toPurge = await purgedDocs.getPurgedIds(userCtx, docIds, false);
   toDelete.push(...toPurge);
 
+  const toArchive = await getArchivedDocs(docIds);
```
jkuester: Can we do this and the `getPurgedIds`  request in parellel?

api/src/services/replication/replication.js
```diff
@@ -40,6 +40,11 @@ const getDocIdsRevPairs = async (docIds) => {
     .map(row => ({ id: row.id, rev: row.value.rev }));
 };
 
+const getArchivedDocs = async (docIds) => {
+  const result = await db.archive.allDocs({ keys: docIds });
+  return result.rows.filter(row => !row.error && !row.value?.deleted).map(row => row.id);
```
jkuester, line 45: So, if we ever "unarchive" a doc, it will be marked as `deleted` in the archive db?

api/src/controllers/archive.js
```diff
@@ -0,0 +1,197 @@
...
+    throw new errors.BadRequestError('No valid doc IDs found in request body');
+  }
+  return jobs;
+};
+
+module.exports = {
+  /**
+   * @openapi
+   * /api/v1/archive:
+   *   post:
```
jkuester, line 129: Add `x-since: 5.3.0`

api/src/controllers/archive.js
```diff
@@ -0,0 +1,197 @@
...
+   *       required: true
+   *       content:
+   *         text/csv:
+   *           schema:
+   *             type: string
+   *           example: |
+   *             doc-id-1
+   *             "doc-id-2"
+   *     responses:
+   *       '201':
```
jkuester: Might be getting too philosophical here, but I was expecting a `202` code here indicating that the archiving job was queued.  IMHO it is a bit unexpected to treat the _archiving job_ as an "entity" that can be created (like contacts/reports/etc).  It is more just an operation that will manipulate entities.

api/src/controllers/archive.js
```diff
@@ -0,0 +1,197 @@
...
+   *                   type: array
+   *                   items:
+   *                     type: object
+   *                     properties:
+   *                       id:
+   *                         type: string
+   *                         description: The id of the created archive job document.
+   *                       count:
+   *                         type: number
+   *                         description: The number of document ids stored in the job.
```
jkuester, line 169: Okay, so we have got a bit of an interesting situation here given the overlap/similarities between this PR and what we are doing over in the [bulk-delete PR](https://github.com/medic/cht-core/pull/11236).  Specifically, in that PR we have added a [bulk-operation endpoint](https://github.com/medic/cht-core/pull/11236/changes#diff-0ba5883967b5fe53509ea36ba635cd6a2d2f385e3bc99c276774356aa55a748d) that serves as a consumer-facing abstraction on top of these async bulk operations.  

Basically, when a contact DELETE is triggered we write one or more docs to the `sentinel` DB that serve as the operation queue (just like you are doing here). But we also write a doc to the `medic-logs` db. The log doc stores the status for each of the queued operation jobs and gets updated as sentinel processes the operations.  The contact DELETE endpoint returns the id for the logs doc.  The consumer can pass this ID to the new `api/v1/bulk-operations` endpoint and get back the data from the log doc. The `sentinel` docs are deleted when the operations finish, but the `medic-logs` doc is retained (with information indicating which operations succeeded/failed).

We need the `api/v1/bulk-operations` endpoint for the contact delete flow because the consumers do not have to be admin users (and so checking the doc directly in the DB is not an option).  

---

My point with all of this is that it seems like we could benefit from consistency in our approach here.  If you think a permanant `medic-logs` record makes sense for archive operations also, maybe we just align with what we are doing in the bulk-delete and return the log id here.  Alternatively, if you think the extra logs doc is unnecessary (for both archiving and bulk delete), maybe we should refactor the implementation of bulk delete to remove the logs doc (and have `api/v1/bulk-operations` just return data from the `sentinel` queue docs).  (The main reason we went with the extra logs doc is that it gave us a place to persist error data about any operations that failed.  I did not want to leave the `sentinel` doc in place and have to work around or retry it all the time, but I also wanted some way to store/return data about which operations _failed_).

Let me know what you think!

shared-libs/archiving-utils/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@medic/archiving-utils",
```
jkuester: It feels like overkill to have a shared-lib here.  AFICT, we typically have only used shared-libs to store logic that is _resued_ across multiple levels.  We have not really used it just as a place to logically group different code/functions that are only used in one place.  In this case, it seems like the only bit that is actually shared between api and sentinel is the `ATTACHMENT_NAME`.  That can just go in `shared-libs/constants`.  The rest can just go where is is actually used.,,,

sentinel/tests/unit/db.spec.js
```diff
@@ -158,4 +158,42 @@ describe('db', () => {
...
+    it('handles an empty docs array', async () => {
+      sinon.stub(request, 'post').resolves();
+      await db.purge({ name: 'host/db' }, []);
+      expect(request.post.args[0][0].body).to.deep.equal({});
+    });
+
+    it('throws request.post errors', async () => {
+      sinon.stub(request, 'post').rejects(new Error('purge failed'));
+      try {
+        await db.purge({ name: 'host/db' }, [{ _id: 'x', _revs: ['1-x'] }]);
```
jkuester: `await expect(db.purge...).to.be.rejectedWith(...`

api/tests/mocha/controllers/archive.spec.js
```diff
@@ -0,0 +1,294 @@
...
+    it('responds with a PermissionError when caller is not a db admin', async () => {
+      sinon.stub(auth, 'assertDbAdmin').rejects(new errors.PermissionError('User is not an admin'));
+      sinon.stub(serverUtils, 'error').returns();
+      sinon.stub(db.sentinel, 'put');
+
+      const req = newReq(['a', 'b']);
+      const res = newRes();
+      await controller.create(req, res);
+
+      chai.expect(auth.assertDbAdmin.callCount).to.equal(1);
```
jkuester: Can we assign `expect` and not use `chai.expect` everywhere?

sentinel/src/schedule/archiving.js
```diff
@@ -0,0 +1,34 @@
...
+module.exports = {
+  /**
+   * Schedules the next archiving run from the `archive` settings: `text_expression` or `cron`
+   * decide when it fires, and the optional `duration` (e.g. "4 hours") bounds how long it runs —
+   * unbounded when missing or malformed. Called on every scheduler tick; reschedules the pending
+   * run each time so config changes take effect. A no-op when archiving is not configured.
+   * @returns {Promise<void>}
+   */
+  execute: () => {
+    const archiveConfig = config.get('archive');
```
jkuester, line 21: Is there a docs PR for the new config settings?

sentinel/src/schedule/archiving.js
```diff
@@ -0,0 +1,34 @@
...
+    const schedule = scheduling.getSchedule(archiveConfig);
+
+    if (!schedule) {
+      return Promise.resolve();
```
jkuester: Am I missing something here or does this mean that archiving will never run unless you have a schedule set for `archive`?  If so, that feels surprising.  I don't think we should return a `2xx` from the api endpoint if we might just never actually do the archiving...  Plus, what happens to the sentinel docs that might be queued while this is not set?  They will just stack up and then suddenly all of them will be processed if archiving is ever scheduled, right?  

Personally, I think it would be best to default to immediately executing the archive (if no schedule is configured).  And if folks want to defer the archive execution, they can do that via the config...

sentinel/src/lib/archiving.js
```diff
@@ -0,0 +1,264 @@
...
+ */
+const processQueue = async (deadline) => {
+  let startkey;
+  do {
+    const job = await fetchNextJob(startkey);
+    if (!job) {
+      break;
+    }
+    startkey = job._id;
+    if (job.status === FAILED_STATUS) {
```
jkuester: Yeah, stuff like this is why I ended up writing the `medic-logs` doc for the bulk operations.  Now, when you have failed jobs in the `sentinel` db, there is _an extra fetchNextJob call_ for each failed job every time we run process queue.  

If we do end up keeping it like this, I think we should at least log a message here so that you can tell from the Sentinel logs that the failed archive jobs are stacking up (or not).

sentinel/src/lib/archiving.js
```diff
@@ -0,0 +1,264 @@
...
+  while (ids.length) {
+    const chunk = ids.splice(0, FETCH_BATCH_SIZE);
+    const medicDocs = await db.medic.allDocs({ attachments: true, keys: chunk, include_docs: true });
+    const docsToArchive = medicDocs.rows
+      .filter(row => canArchive(row.doc))
+      .map(row => ({ ...row.doc, archive_date: date }));
+    if (!docsToArchive.length) {
+      continue;
+    }
+    await db.archive.bulkDocs(docsToArchive, { new_edits: false });
```
jkuester, line 94: Do we need to check the `error` property on each of the results to see if any of the doc saves falied?  (Since the overall `bulkDocs` call does not fail if one of the writes fails...)

sentinel/src/lib/archiving.js
```diff
@@ -0,0 +1,264 @@
...
+ * @returns {Promise<void>}
+ */
+const recordError = async (job, err) => {
+  try {
+    const latest = await db.sentinel.get(job._id);
+    job._rev = latest._rev;
+    job.error_count = (job.error_count || 0) + 1;
+    job.errors = job.errors || [];
+    job.errors.push({ date: Date.now(), message: err?.message || err?.stack || err });
+    if (job.errors.length > MAX_ERRORS_KEPT) {
```
jkuester: Is there really so much value in only keeping the last `5` errors?  Seems like extra effort to have the `error_count` and `MAX_ERRORS_KEPT` just to avoid storing `20` error stacks on the doc....  This is very much not a happy-path case anyway, so it seems okay to go with the more straght-forward implementation.

At the same time, `20` retrys seems _very high_ to me.  I know some of these jobs might have a lot of batches, but I feel like if we could not finish the job in the first `10` attempts, it is time to give up....

jkuester (changes requested):
Most excellent!  Very excited for this functionality!  Have not made it through the integration/e2e tests yet, but wanted to get these thoughts out here....

api/src/routing.js
dianabarsan: I kinda want this to have its own identity. 
`bulk-delete` does all sorts of weird stuff with hierarchy deletes and cleanup, while this does something else.

sentinel/src/lib/archiving.js
dianabarsan: Yes, I have. 

The docs simply disappear from the changes feed - which is what cht-sync is watching - the purging doesn't show up. So cht-sync will not know the docs have been archived, or that anything happened really. 
as it stands, Postgres will not know that a document was archived in cht-core.

api/src/services/replication/replication.js
dianabarsan: Well, there is no un-archiving official operation per-se. 
So this is some actor doing a delete. If the actor in doing a purge in the arvhiveDb to unarchive, then it won't show up as deleted. 
This just covers the case of when someone does un-archiving using delete.

api/src/controllers/archive.js
dianabarsan: I have error persistence in this code as well, but I didn't create a separate log doc, but just kept everything in the Sentinel doc. 
I agree that storing process info in a persistent logs-doc might be useful, in case someone who knows their job id wants to check what happened to it later.

shared-libs/archiving-utils/package.json
dianabarsan: I agree on this. I used to have more things here, but then they got gradually removed as the code matured.

sentinel/src/schedule/archiving.js
dianabarsan: Oh, good point! I completely forgot.

sentinel/src/schedule/archiving.js
dianabarsan: I agree with you for the consistency: why allow creating archive-docs if the process will never run. 

I suppose I think about archiving as I think of purging, that it's a resource intensive operation potentially lasting many hours, that will affect server performance meanwhile. 

I can go with immediate execution.

sentinel/src/lib/archiving.js
dianabarsan: The `new-edits: false` call never returns errors. It actually doesn't return anything.
the only error that could occur is if the write failed on some system level, and then the whole request will return a failure, but there will be no individual docs failing. 
`new-edits` basically says: put this in the rev tree without checking what the rev-tree looks like.

sentinel/src/lib/archiving.js
dianabarsan: I want to keep the log, I think it's useful to see why a job failed historically, it might not fail for the same reasons. 

But I will implement the separate logs-doc architecture and then the whole sentinel doc becomes cleaner.

sentinel/src/lib/archiving.js
jkuester: Interesting!  I was not familiar with this param...

sentinel/src/lib/archiving.js
jkuester: Nice. I think that is probably how we want it (at least for now).

Related question: does cht-sync know about doc deletions currently?  Since we  are leveraging archiving for the contact-delete, I am wondering if we will be causing drift in the postgres data that otherwise would not be there....?

sentinel/src/lib/archiving.js
```diff
@@ -0,0 +1,338 @@
...
+    job.total = ids.length; // account for possible doc tampering
+    await setLogRunning(job);
+    let batches = 0;
+
+    do {
+      const batch = ids.slice(job.cursor, job.cursor + PURGE_BATCH_SIZE);
+      await archiveBatch(batch);
+      await saveJob(job, batch.length);
+      await updateLogCursor(job);
+      if (++batches % 10 === 0) {
```
jkuester: This check is scoped to the "job" level, meaning it is only run when a particular _job_ has 10_000+ docs.  

However, if you submit 100 requests for jobs that each have 9_000 docs, we will never refresh the indexes.  Not sure this is a huge issue, though....  🤔

sentinel/src/lib/archiving.js
dianabarsan: yes, cht-sync gets doc deletions though the changes feed and deletes the docs from postgres as well.

api/src/controllers/archive.js
dianabarsan: added separate logging.

api/tests/mocha/services/replication/replication.spec.js
```diff
@@ -314,12 +330,88 @@ describe('Initial Replication service', () => {
...
+          { key: 'doc1', id: 'doc1', value: { rev: '2-a', deleted: true } },
+          { key: 'doc2', id: 'doc2', value: { rev: '1-a' } },
+        ]
+      });
+
+      const result = await replication.getDocIdsToDelete(userCtx, ['doc1', 'doc2']);
+      expect(result).to.have.members(['doc2']);
+    });
+
+    it('should combine deleted, purged and archived docs without duplicates', async () => {
```
jkuester: "without duplicates" is a bit misleading here since really we only have one of each doc (deleted, purged, archived).  The real "without duplicates" test would be to have some invalid data case where the same doc comes back as deleted, purged, AND archived (but we only return its id once).

sentinel/src/lib/archiving.js
```diff
@@ -0,0 +1,338 @@
...
+  while (ids.length) {
+    const chunk = ids.splice(0, FETCH_BATCH_SIZE);
+    const medicDocs = await db.medic.allDocs({ attachments: true, keys: chunk, include_docs: true });
+    const docsToArchive = medicDocs.rows
+      .filter(row => canArchive(row.doc))
+      .map(row => ({ ...row.doc, archive_date: date }));
+    if (!docsToArchive.length) {
+      continue;
+    }
+    await db.archive.bulkDocs(docsToArchive, { new_edits: false });
```
jkuester, line 94: One thing we are losing here are any conflicts that exist for these archived docs, right?   The purging deletes all of them, but this only copies the willing version (I think).

IMHO, this is okay, but wanted to call out that there is some "data loss".

shared-libs/audit/src/index.js
```diff
@@ -151,7 +151,36 @@ const expressCallback = async (req, responseBody, requestMetadata) => {
...
+    const auditDoc = existingAuditDocs.rows[idx].doc || getAuditDoc({ id });
+
+    if (auditDoc.history.some(entry => entry.archived)) {
+      return;
+    }
+
+    auditDoc.history.push({ date: new Date(date), archived: true });
+    newAuditDocs.push(auditDoc);
+  });
+  const results = await db.bulkDocs(newAuditDocs);
```
jkuester, line 178: Minor, but could return early if `newAuditDocs` is empty (before calling bulkDocs).

tests/e2e/default/db/archive.wdio-spec.js
```diff
@@ -0,0 +1,122 @@
...
+
+    await utils.updateSettings({ archive: { text_expression: 'every 1 seconds' } }, { ignoreReload: true });
+    await utils.runSentinelTasks();
+    await sentinelUtils.waitForArchiveCompletion();
+  };
+
+  afterEach(async () => {
+    await utils.revertSettings(true);
+    await utils.deleteUsers([user]);
+    await utils.revertDb([/^form:/], true);
```
jkuester: Should we clean up the archive DB?

api/src/routing.js
jkuester: Should we deprecate `/api/v1/bulk-delete` (in the openapi docs) in favor of `/api/v1/archive`?  I am just having trouble coming up with a situation where someone should use `bulk-delete` instead of just calling `archive`...

sentinel/src/lib/archiving.js
jkuester: :thinking:

jkuester (changes requested):
Couple more minor questions/requests

sentinel/src/lib/archiving.js
dianabarsan: Yes, correct, the conflicts are purposefully deleted. Imagine an old conflict (an old version of a doc) suddenly showing up in the changes feed!

sentinel/src/lib/archiving.js
dianabarsan: Oh good catch!

api/src/routing.js
dianabarsan: I think we should just deprecate it when v2/bulk-delete lands. it's more semantically connected.

sentinel/src/schedule/archiving.js
```diff
@@ -8,27 +8,23 @@ module.exports = {
   /**
    * Schedules the next archiving run from the `archive` settings: `text_expression` or `cron`
    * decide when it fires, and the optional `duration` (e.g. "4 hours") bounds how long it runs —
-   * unbounded when missing or malformed. Called on every scheduler tick; reschedules the pending
-   * run each time so config changes take effect. A no-op when archiving is not configured.
+   * unbounded when missing or malformed. When no schedule is configured, queued jobs are
```
jkuester, line 15: > or malformed

Minor, but should we log a message if they configure a schedule, but we do not use it?  Just thinking that might be very helpful for troubleshooting....

sentinel/src/lib/archiving.js
```diff
@@ -215,22 +292,21 @@ const processJob = async (job, deadline) => {
...
 /**
- * Drains the job queue in _id order until the deadline, skipping quarantined jobs.
+ * Drains the job queue in _id order until the deadline. Permanently failed jobs are deleted by
+ * recordError, so everything in the queue is processable.
  * @param {number} deadline - epoch ms after which no further job is started
  * @returns {Promise<void>}
  */
 const processQueue = async (deadline) => {
   let startkey;
+  const indexCounter = { batches: 0 };
```
jkuester, line 310: Minor, (don't need to change this) but I have to complain about using mutable state here to track `batches` instead of just returning the new integer value from `processJob`.  

IMHO, it is almost always easier to follow the logic flow when reading code if you just return the key changed values (instead of just mutating them under the hood).

sentinel/src/lib/archiving.js
```diff
@@ -0,0 +1,340 @@
...
+ * @param {Error|*} err - the error that failed the job
+ * @returns {Promise<void>}
+ */
+const recordError = async (job, err) => {
+  try {
+    job.error_count = (job.error_count || 0) + 1;
+    await addLogError(job._id, err?.message || err?.stack || err);
+
+    if (job.error_count >= MAX_JOB_ATTEMPTS) {
+      await setLogFailed(job._id);
```
jkuester: Maybe not worth optimizing too much for an error path, but is a bit weird to make two separate db round-trips to update the log when we already have all the data for both updates.

tests/integration/api/controllers/archive.spec.js
```diff
@@ -0,0 +1,141 @@
...
+  it('creates a job doc in the sentinel db with the ids stored as an attachment', async () => {
+    const csv = ['doc-a', 'doc-b', 'doc-c'].join('\n');
+    const response = await postCsv(csv);
+
+    expect(response.jobs).to.have.lengthOf(1);
+    expect(response.jobs[0]).to.have.keys('id', 'count');
+    expect(response.jobs[0].count).to.equal(3);
+    expect(response.jobs[0].id).to.match(/^archive:/);
+
+    const doc = await utils.sentinelDb.get(response.jobs[0].id);
```
jkuester, line 58: Is there a chance this test flakes because Sentinel actually starts processing the job before we read it here?  I guess the schedule will run the jobs every 5min by default, right?  So, it would be a pretty impossibly small window to hit that race condition.

sentinel/src/lib/archiving.js
```diff
@@ -0,0 +1,340 @@
...
+  }
+  const date = Date.now();
+
+  const archivedIds = [];
+  // fetch and write chunk by chunk — attachments are inlined, so both payloads need bounding
+  while (ids.length) {
+    const chunk = ids.splice(0, FETCH_BATCH_SIZE);
+    const medicDocs = await db.medic.allDocs({ attachments: true, keys: chunk, include_docs: true });
+    const docsToArchive = medicDocs.rows
+      .filter(row => canArchive(row.doc))
```
jkuester, line 87: Should we somehow indicate in the log doc that we skipped some of the ids here?  It just feels a bit sketch to silently drop values without giving any kind of feedback.  If not in the `medic-logs` doc, then maybe we can at least log a message to the console here to perhaps assist with future debugging.

tests/integration/sentinel/schedules/archive.spec.js
```diff
@@ -0,0 +1,450 @@
...
+    }
+  };
+
+  const updateSettings = async (duration = '2 hours') => {
+    await utils.updateSettings(
+      { archive: { text_expression: 'every 1 seconds', duration: duration } },
+      { ignoreReload: true }
+    );
+    await utils.toggleSentinelTransitions();
+    await sentinelUtils.skipToSeq();
```
jkuester, line 146: Am I missing something here?  Does setting the transitions seq somehow affect the schedules?

tests/e2e/default/db/archive.wdio-spec.js
```diff
@@ -0,0 +1,136 @@
...
+      .catch(err => done({ ok: false, status: err.status }));
+  }, id);
+
+  beforeEach(async () => {
+    reportToArchive = genericReportFactory
+      .report()
+      .build({ form: 'home_visit' }, { patient, submitter: contact });
+    await utils.saveDocs([...places.values(), contact, patient]);
+    await utils.createUsers([user]);
+    await utils.saveDocs([reportToArchive]);
```
jkuester: This can just go in the other `saveDocs` call on line 47....

tests/e2e/default/db/archive.wdio-spec.js
```diff
@@ -0,0 +1,136 @@
...
+
+    local = await getLocalDoc(reportToArchive._id);
+    expect(local.ok).to.equal(false);
+    expect(local.status).to.equal(404);
+  });
+
+  it('restores an unarchived doc to the offline user device on the next sync', async () => {
+    await loginPage.login(user);
+    await archiveReport(reportToArchive);
+    await commonElements.sync();
```
jkuester: I know it is not the "most bestest practice", but in our e2e tests, we frequently trade test independence for better performance (meaning tests in the file must execute sequentially since later tests depend on earlier ones).

Personally, I have made peace with this practice as being one of the most simple ways to mitigate spiraling execution times as long as the test files, themselves, are still independent of each other (even if the tests within a file are not).  

In this case, it feels like we have a lot of processing effort to set everything back up (including a totally new user, etc) just to get back to the place where we were at the end of the first test.

tests/e2e/default/db/archive.wdio-spec.js
```diff
@@ -0,0 +1,136 @@
...
+    // Unarchive: restore the doc into medic AND remove it from the archive db.
+    const archived = await utils.archiveDb.get(reportToArchive._id);
+    const restored = { ...archived };
+    delete restored._rev;
+    delete restored.archive_date;
+    // Restore as a two-write edit chain. Writing the identical content once would mint
+    // the exact gen-1 rev the client already holds as its tombstone's parent, so the
+    // client's new_edits:false download would be a no-op and the doc would stay deleted
+    // on the device. The second write bumps the server doc to a gen-2 live rev the
+    // client has never seen — it lands as a live branch that wins over the tombstone.
```
jkuester, line 103: This behavior is unexpected to me.  🤔  Is the gen-1 rev deterministic? If so, what is it determined by?

jkuester (commented):
Approving to unblock!  👍 🚀   Left a couple minor comments/questions, but I don't see anything that should hold us up here!

jkuester (approved):
Approving it for real 😅

tests/e2e/default/db/archive.wdio-spec.js
dianabarsan: apparently it's a signature of the contents of the document.

sentinel/src/lib/archiving.js
dianabarsan: I agree with this, and I did have a version that used an integer, but if a job fails, it won't persist the counter.

tests/integration/api/controllers/archive.spec.js
dianabarsan: that's extremely unlikely, i think. before we "immediately" processed the jobs, this would have never failed because we lacked a schedule. I'll add a schedule so the time window never overlaps with the testing time.

sentinel/src/schedule/archiving.js
dianabarsan: done.

tests/integration/sentinel/schedules/archive.spec.js
dianabarsan: no, this is some silliness about sentinel writing infodocs while archiving purges them.

mrjones-plip:
\o/

PR #11142, opened 2026-06-05
feat(#10694): adds tasks page free text search functionality

### LINKED ISSUE #10694: Implement free text search feature in tasks page

dianabarsan:
# Description

Closes #10694.

This adds free text search to the Tasks page. The search bar's freetext input — previously disabled on this page — is now enabled, letting users filter their task list by contact name, lineage, or task title. Typing a query writes it into the global filters state, where a memoized NgRx selector reactively re-filters the list; no extra query round-trip or page reload is involved.

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10694-adds-tasks-page-free-text-search-functionality-/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10694-adds-tasks-page-free-text-search-functionality-/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10694-adds-tasks-page-free-text-search-functionality-/docker-compose/cht-couchdb-clustered.yml)

# License

### DIFF
--- a/webapp/src/ts/modules/tasks/tasks.component.ts
+++ b/webapp/src/ts/modules/tasks/tasks.component.ts
@@ -134,6 +134,7 @@ export class TasksComponent implements OnInit, OnDestroy {
 
   ngOnInit() {
     this.trackLoadPerformance = this.performanceService.track();
+    this.globalActions.clearFilters();
     this.tasksActions.setSelectedTask(null);
     this.subscribeToStore();
     this.subscribeToChanges();
--- a/webapp/src/ts/selectors/index.ts
+++ b/webapp/src/ts/selectors/index.ts
@@ -6,6 +6,48 @@ interface TaskWithLineage extends TaskEmission {
   lineageIds: string[];
 }
 
+const normalizeSearchText = (value?: string): string => {
+  if (!value) {
+    return '';
+  }
+
+  return value
+    .toString()
+    .toLowerCase()
+    .normalize('NFD')
+    .replace(/[\u0300-\u036f]/g, '')
+    .trim();
+};
+
+const splitSearchTerms = (search?: string): string[] => {
+  const normalized = normalizeSearchText(search);
+  return normalized ? normalized.split(/\s+/).filter(Boolean) : [];
+};
+
+const fuzzyMatch = (text?: string, term?: string): boolean => {
+  if (!term) {
+    return true;
+  }
+
+  const normalizedText = normalizeSearchText(text);
+  if (!normalizedText) {
+    return false;
+  }
+
+  if (normalizedText.includes(term)) {
+    return true;
+  }
+
+  let termIndex = 0;
+  for (let i = 0; i < normalizedText.length && termIndex < term.length; i++) {
+    if (normalizedText[i] === term[termIndex]) {
+      termIndex += 1;
+    }
+  }
+
+  return termIndex === term.length;
+};
+
 const getGlobalState = (state): GlobalState => state.global || {};
 
 const applyTasksFilters = (tasks: TaskWithLineage[], filters: TasksFilters = {}): TaskWithLineage[] => {
@@ -24,6 +66,25 @@ const applyTasksFilters = (tasks: TaskWithLineage[], filters: TasksFilters = {})
     filtered = filtered.filter(task => {
       return task.lineageIds.some(id => filters.facilities?.selected.includes(id));
     });
+  }
+
+  if (filters.search) {
+    const terms = splitSearchTerms(filters.search);
+    if (terms.length) {
+      filtered = filtered.filter(task => {
+        const candidates = [
+          task?.contact?.name,
+          ...(task?.lineage || []),
+          task?.title,
+        ].filter(Boolean);
+
+        if (!candidates.length) {
+          return false;
+        }
+
+        return terms.every(term => candidates.some(candidate => fuzzyMatch(candidate, term)));
+      });
+    }
   }
 
   return filtered;
--- a/webapp/src/ts/selectors/index.ts
+++ b/webapp/src/ts/selectors/index.ts
@@ -88,11 +88,9 @@ const applyTasksFilters = (tasks: TaskWithLineage[], filters: TasksFilters = {})
     });
   }
 
-  if (filters.search) {
-    const normalizedSearch = normalizeText(filters.search);
-    if (normalizedSearch) {
-      filtered = filterTasksBySearch(filtered, normalizedSearch);
-    }
+  const normalizedSearch = normalizeText(filters.search);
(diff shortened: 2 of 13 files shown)


dianabarsan:
Hey @sugat009 . 

This is a merger between two PRs and some fixes that I added on top. Would appreciate a sanity check. Thanks!

webapp/src/ts/selectors/index.ts
```diff
@@ -1,13 +1,76 @@
...
+  const entries = remainingWithCandidates.flatMap(
+    ({ candidates }, idx) => candidates.map(candidate => ({ candidate: normalizeText(candidate), idx }))
+  );
+  const fuse = new Fuse(entries, { ...FUSE_OPTIONS, keys: ['candidate'] });
+  const fuzzyMatchIndices = new Set(fuse.search(normalizedSearch).map(r => r.item.idx));
+  const fuzzyMatched = remainingWithCandidates
+    .filter((_, idx) => fuzzyMatchIndices.has(idx))
+    .map(({ task }) => task);
+
+  return [...substringMatched, ...fuzzyMatched];
```
sugat009, line 70: question (non-blocking): Search returns `[...substringMatched, ...fuzzyMatched]`, so substring matches always sit above fuzzy ones, overriding the `orderByDueDateAndPriority` sort across the two groups. Reproduced locally: "Johnson" (most overdue, top of list) drops below "Jonson" when searching "jonson". Is due date / priority order needed when searching, or is match quality ordering sufficient here?

(Minor: within the fuzzy group, `fuse.search()`'s relevance ranking is dropped, so it falls back to the original order rather than "most matched first".)

webapp/src/ts/selectors/index.ts
```diff
@@ -1,13 +1,76 @@
...
+  return [
+    task?.contact?.name,
+    ...(task?.lineage || []),
+    task?.title,
+  ].filter(Boolean) as string[];
+};
+
+const FUSE_OPTIONS = {
+  threshold: 0.2,        // 0 = exact, 1 = match anything; 0.2 allows minor typos
+  distance: 50,          // how far from expected position a match can appear
```
sugat009: question: A couple of things in FUSE_OPTIONS that the next person tuning the fuzzy search might trip on:

1. With `ignoreLocation: true`, does `distance: 50` do anything? In fuse.js 7.3.0, `ignoreLocation: true` skips the location/distance scoring entirely, so `distance` looks inert here (toggling its value changes no results). The comment ("how far from expected position a match can appear") reads as if it is active. Worth dropping it, or noting it only applies when `ignoreLocation` is false?

2. `minMatchCharLength`'s comment ("skip fuzzy matching for very short queries") describes the query length gate on line 56, but the value is also passed to Fuse where it means the minimum matched region length. A word distinguishing the two uses could help.

webapp/src/ts/modules/tasks/tasks.component.ts
```diff
@@ -134,6 +134,7 @@ export class TasksComponent implements OnInit, OnDestroy {
 
   ngOnInit() {
     this.trackLoadPerformance = this.performanceService.track();
+    this.globalActions.clearFilters();
```
sugat009, line 137: suggestion (non-blocking): A few behaviors this PR introduces aren't pinned by tests yet. None are blocking, but they would guard against silent regressions:

1. The guard against a Reports search leaking into Tasks. `clearFilters()` here in ngOnInit resets the shared `filters.search` on entry, but nothing asserts it: the component spec doesn't check that ngOnInit calls `clearFilters()`, and no e2e searches on Reports, opens Tasks, and expects the full list. Removing this one line would still pass CI today.

2. Reset preserves search. The e2e that combines search with sidebar filters calls `resetFilters()` then `clearSearch()` back to back with no assertion in between, so a regression of `clearFilters('search')` to `clearFilters()` would slip through end to end. Asserting the list still shows only the searched contact after reset (before clearing search) would close it.

3. Result ordering. If the substring then fuzzy grouping is intentional (see the ordering question on the selector), a test asserting that order would lock it in; today every search test populates only one bucket.

webapp/tests/karma/ts/selectors/index.spec.ts
```diff
@@ -609,6 +609,198 @@ describe('Selectors', () => {
...
+
+        const nepaliState = { filters: { search: 'रामकुमारी' } } as any;
+        const nepaliResult = Selectors.getFilteredTasksList.projector(tasksState, nepaliState);
+        expect(nepaliResult).to.deep.equal([tasksState.tasksList[0]]);
+
+        const arabicState = { filters: { search: 'فاطمة' } } as any;
+        const arabicResult = Selectors.getFilteredTasksList.projector(tasksState, arabicState);
+        expect(arabicResult).to.deep.equal([tasksState.tasksList[1]]);
+
+        const crossScriptState = { filters: { search: 'فاطمة' } } as any;
```
sugat009: nitpick: `crossScriptState` here duplicates `arabicState` (same query, same expectation), so it doesn't add coverage. Either drop it, or make it genuinely cross script, e.g. a contact whose name is in one script while the lineage is in another, to prove both fields are searched.

sugat009:
Took it for a spin locally too, looks good. Just a few minor questions and nits inline, nothing blocking 👍

webapp/src/ts/selectors/index.ts
dianabarsan: This is a great point. Off the top of my head, it would be search, however sorting within search should also matter? 

If I use a search term and the most distant fuzzy match appears first because it's overdue? And the best match shows last because it's due in 3 days? 

I am unsure. maybe we should apply the sort order over substring and then over fuzzy and return them in sequence?

sugat009: Thinking about it more, let's leave this as is. The tasks list has no sort control (only the due date and type filters), so there's no user sort for search to override, it just orders results by match (substring before fuzzy) then due date within each group, which is the expected default for a searchable list. No change needed, the current behavior is good. If a sortable tasks list ever comes up, this can be revisited then.

sugat009 (approved):

PR #11162, opened 2026-06-10
perf(#11080): update CouchDb to 3.5.2

### LINKED ISSUE #11080: Upgrade to CouchDB 3.5.2
**Describe the issue**
CouchDB 3.5.1 [[had performance issues](https://github.com/medic/cht-core/issues/10455#issuecomment-3926903530)] so we skipped it, but it [had](https://docs.couchdb.org/en/stable/whatsnew/3.5.html#version-3-5-1) features we wanted: 
- fixes for Nouveau
- performance improvements for Nouveau
- _purge optimizations: this could enable us to actually implement purging as a way to clean up historical data from CouchDb databases (eg: https://github.com/medic/cht-core/issues/6615) 

CouchDB 3.5.2 also [has](https://docs.couchdb.org/en/stable/whatsnew/3.5.html#version-3-5-2) features we want:
* #[5987 Optimize view building](https://github.com/apache/couchdb/pull/5987) - (10-12% faster) 
* Nouveau bulk index updates. Replace doc-at-a-time updates to bulk updates which should improve index building
* Nouveau Lucene 10 upgrade
* Pre-fetch design docs once per cleanup request and reuse signatures across nodes (instead of fetching them on every worker

(though their release notes description <> links <> tickets are a big hot mess. gave them a [shout out in their slack](https://couchdb.slack.com/archives/C49LEE7NW/p1778868668631289) "_Looking at the release notes though, I see some confusing items listed_.")

Upgrade CouchDb version to 3.5.2

### LINKED ISSUE #11153: Nouveau requests timeout after 30 seconds
**Describe the bug**
CouchDb version 3.5.0 connects to Nouveau using `ibrowse` library. 
https://github.com/apache/couchdb/blame/020a52c765105cfdb36c084b41fae060d3554030/src/nouveau/src/nouveau_api.erl#L323

Ibrowse default timeout is 30s: https://github.com/cmullaparthi/ibrowse/blob/master/src/ibrowse.erl#L335
https://github.com/cmullaparthi/ibrowse/blob/master/src/ibrowse.erl#L618

If the timeout is exceeded, CouchDb just logs a `notice` and fails the request with:
```
{ error: 'unknown_error', reason: 'req_timedout' }
```
CouchDb Logs show: 
```
[notice] 2026-06-08T00:00:30.990758Z couchdb@127.0.0.1 <0.140288108.5> 2eaa86ef85 haproxy.<server>:5984 10.42.1.56 medic POST /medic/_design/medic/_nouveau/docs_by_replication_key 500 ok 30009
```

To Reproduce (not verified yet).
1. Launch CHT in a very small server or reduce the CPU through compose capabilities.  
2. Make a relatively large Nouveau Query. 

**Expected behavior**
Requests should not timeout. 

**Workaround**
This means that the only way to fix this is to make smaller requests.

**Environment**
- Instance: production
- App: Nouveau
- Version: 5.0.0+

**Additional context**
I found this when checking purging status on a 5.0.1 instance. Purging had been failing consistently with the same error.

dianabarsan:
# Description

Also sets `nouveau.request_timeout` to 1h. 
Also fixes nouveau container naming misalignment causing CI logs to not be saved (container name doesn't follow naming convention: called `nouveau` instead of `couchdb-nouveau` like its image tag).

closes #11080
closes #11153

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:11080-couch-bump/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:11080-couch-bump/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:11080-couch-bump/docker-compose/cht-couchdb-clustered.yml)

# License


tests/integration/api/controllers/replication-failure-log.spec.js
```diff
@@ -88,11 +88,7 @@ describe('replication failure logging @docker', () => {
   before(async () => {
     await utils.saveDocs([facility, mathilPlace, janicePlace, mathilContact, janiceContact]);
     await utils.createUsers(users, true);
-  });
-
-  after(async () => {
-    await utils.deleteUsers(users);
-    await utils.revertDb();
+    await utils.clearReplicationFailureLogs();
   });
```
jkuester, line 97: Did you mean to delete this?  I wonder if maybe you meant to remove the`afterEach`?

tests/utils/index.js
```diff
@@ -1773,6 +1773,9 @@ const updateContainerNames = (project = PROJECT_NAME) => {
 };
 
 const getContainerName = (service, project = PROJECT_NAME) => {
+  if (service.includes('nouveau')) {
```
jkuester, line 1776: What is special about the nouveau container here?  I am having trouble understanding why the other containers do not have this same issue...

jkuester (approved):
Couple minor questions, but otherwise LGTM!

tests/integration/api/controllers/replication-failure-log.spec.js
dianabarsan: oh. oh. these are integration tests :( 

I'll need to revert the delete. Wdio tests have a global `after` that handles all this cleanup, but integration tests are vanilla mocha, which doesn't support a global after (only afterAll and afterEach :( ) 
Thanks for flagging this.

tests/utils/index.js
dianabarsan: what is special is:
- image name is `couchdb-nouveau`
- container name is `nouveau`

All the other services share the same name for image and container, so we can use the image name as a proxy for the container name. but for nouveau this was not the case. the result was the nouveau logs were never stored in the artefact.

jkuester: Ahh, got it.  So there is a mismatch in our docker compose service names vs the image names.  :+1:

Issue #10706, opened 2026-03-10
[DMP 2026]: Add API endpoints for advanced contact management operations


sugat009:
## Ticket Contents

Currently, the CHT (Community Health Toolkit) app allows users to perform basic contact management tasks such as creating an editing contacts. Users with the proper permissions can also delete contacts, but _only if there are no child contacts associated with the contact to delete._

More [advanced contact management](https://docs.communityhealthtoolkit.org/building/contact-management/moving-contacts/) current requires a system administrator to use the [`cht-conf` CLI tool](https://github.com/medic/cht-conf). `cht-conf` can perform operations like [moving](https://github.com/medic/cht-conf/blob/main/src/fn/move-contacts.js), [merging](https://github.com/medic/cht-conf/blob/main/src/fn/merge-contacts.js), and [deleting](https://github.com/medic/cht-conf/blob/main/src/fn/delete-contacts.js) contacts in the hierarchy (note that the `delete-contacts` action in cht-conf is _recursive_ so that it also deletes all the child contacts). Having this functionality in `cht-conf` creates friction for users who cannot perform these actions themselves through the webapp, especially when dealing with duplicate contacts or reorganizing health facility hierarchies.

It also forces services like [cht-user-management](https://github.com/medic/cht-user-management) to either re-implement these operations or else somehow leverage cht-conf under the hood to perform them.  

This project aims to add REST API endpoints in `cht-core` for these actions. The work can be done incrementally (one action at a time), giving it a limited initial scope with room to expand. The goal is simply to have api support for these operations. Updating the webapp to actually give the user access to these operations is _out of scope_. 

## Goals & Mid-Point Milestone

- [ ] Understand the existing `cht-conf` move/merge/delete contact logic
- [ ] Implement the delete-contacts API endpoint with support for recursive hierarchy deletion
- [ ] Implement the move-contacts API endpoint for relocating contacts between parent hierarchies
- [ ] Implement the merge-contacts API endpoint for combining duplicate contacts and reassigning all linked documents (reports, tasks) to the surviving contact
- [ ] Write API documentation for all new endpoints in [cht-docs](https://github.com/medic/cht-docs)
- [ ] **Goals achieved by mid-point milestone:** Delete-contacts and move-contacts API endpoints are functional with unit and integration tests

## Expected Outcome

New REST API endpoints available in `cht-core`:
[...]

Issue #10626, opened 2026-02-11
Enable deleting a hierarchy as a single step


PhilipNgari:
**Is your feature request related to a problem? Please describe.**
Whenever as a user manager I want to delete a CHU/subcounty and its children, I have to start from the children all the way up.This is time consuming due to the multiple steps involved. At times some CHUs can have huge data either  because they are  a duplicate CHU among others. This impedes data quality efforts.

**Describe the solution you'd like**
I want to be able to delete such a CHU/subcounty as single step.NB: The app has an **edit** and a **Delete** button but the latter is disabled for a user manager. 
**Describe alternatives you've considered**
none alternative exist
**Additional context**
Below is a sample case for Nakuru county where the subcounty labelled Nakuru west-Remove can t be deleted due to the disabled "delete" button. 

<img width="1701" height="810" alt="Image" src="https://github.com/user-attachments/assets/a5c2b970-07ed-4cca-b6c6-44d1f128c77e" />

Raam751:
This sounds like a helpful feature.
Is there a preferred approach for handling cascade deletes here (soft delete vs hard delete)?
And is the delete button disabled intentionally for user managers for data safety reasons?

jkuester:
I think this request makes sense. 👍 Historically the "batch" operations on contacts have been handled by admins via cht-conf commands (like the [delete-contacts](https://github.com/medic/cht-conf/blob/ce26d62ac1d44cfb735a56b7855a7d234ada8b2d/src/fn/delete-contacts.js#L50) action).  However, more and more we have been hearing about the need for users to be able to self-serve these operations from within the app.  (E.g. the `move-contacts` and `merge-contacts` actions are additional ones that are needed.)  

There are a number of technical considerations that come with this functionality, but I think it should definitely be possible to implement it in a way that server-side logic could be triggered by _online_ users.

Regarding @Raam751's questions:
- I would describe CouchDB's delete functionality (which I _think_ is what is used when deleting a contact here from the UI) as a "hard-ish" delete.  The deleted doc will immediately not be available in subsequent requests to the DB. However, previous revisions of the doc remain stored in the DB until they are eventually purged during the periodic database compaction (leaving only a "tombstone" record indicating the doc existing but not containing any of its data).  At this time we have not implemented a true "soft delete" that is reflected on the server (or for online users). However, [CHT purging](https://docs.communityhealthtoolkit.org/technical-overview/data/performance/purging/) is a form of "soft" delete from offline user devices.
[...]

PR #11163, opened 2026-06-11
feat(#10626): add recursive contact hierarchy delete service

vikrantwiz02:
## Summary

First slice of the delete endpoint for #10706 (sub-issue #10626). Adds the core service that recursively deletes a contact, all of its descendant contacts, and all reports for which any of those contacts is the subject.

Faithful port of `cht-conf`'s `delete-hierarchy.js`, adapted to write directly to CouchDB instead of staging documents to disk:

- Subtree fetched via `medic/contacts_by_depth` (`{ key: [id] }`, tombstone-filtered), matching cht-conf's `getContactWithDescendants`.
- Subject reports fetched via `medic-client/reports_by_subject`, keyed by each contact's uuid and shortcode (the view indexes both).
- Documents marked `_deleted` and written via `bulkDocs` in batches of 100; per-document errors are surfaced in the result rather than swallowed.
- Pre-flight size guard: hierarchies larger than 5,000 contacts are rejected with a clear message, to avoid silently exceeding the HTTP gateway timeout on very large subtrees. The service is kept separate from any transport concern so an async/queued path can be layered on later without changing this logic.

No new CouchDB views.

## Scope

Service layer and unit tests only. Wiring (controller, route, OpenAPI) and integration tests follow in separate PRs. Two deliberate enhancements over the cht-conf reference (clearing a deleted primary contact's parent reference, and removing reports authored by deleted contacts) are also split into their own PRs so each can be reviewed independently.

## Files

- `api/src/services/contact-hierarchy.js`
- `api/tests/mocha/services/contact-hierarchy.spec.js` (8 unit tests)

## Testing

`npm run unit-api`: lint clean, 8/8 new tests passing.

### DIFF
--- a/api/src/services/contact-hierarchy.js
+++ b/api/src/services/contact-hierarchy.js
@@ -1,0 +1,93 @@
+const db = require('../db');
+const logger = require('@medic/logger');
+
+const BATCH_SIZE = 100;
+const MAX_HIERARCHY_SIZE = 5000;
+
+// contacts_by_depth emits a bare [ancestorId] key for a contact and each of its
+// ancestors, so { key: [contactId] } returns one row per member of the subtree.
+const countSubtree = async (contactId) => {
+  const result = await db.medic.query('medic/contacts_by_depth', { key: [contactId] });
+  return result.rows.length;
+};
+
+const getSubtreeContacts = async (contactId) => {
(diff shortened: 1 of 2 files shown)


vikrantwiz02:
Hi @jkuester, this PR is ready for review. The follow-up PRs will be opened further.

vikrantwiz02:

PR #11165, opened 2026-06-12
fix(#8011): prevent BS date conversion when month is not selected

### LINKED ISSUE #8011: Bikram sambat date picker converts date even when user hasn't selected a Month
**Describe the bug**
Bikram sambat date picker widget converts date even when user hasn't selected a Month.

**To Reproduce**
1. Configure your form to have bikram-sambat input (or use attached form).
2. Open a form.
3. In bikram sambat date widtet, provide Date and Year only and do not select any month. 
image
4. The date will be converted from Bikram Sambat to Gregorian date. During this process, the form always considers the month as `बैशाख` (Baisakh), which is the first month of the Bikram Sambat Calendar. 

**Expected behavior**
Ideally, date should be only converted when user specifies month as well. CHT core allows form to be submitted even when month is not selected.

**Environment**
- Instance: Any CHT-Core instances with bikram sambat date supported
- App: webapp
- Version: 3.14 onwards

**Reference form:** [bs_date.xlsx](https://github.com/medic/cht-core/files/10387461/bs_date.xlsx)

megha1807:
# Description

The Bikram Sambat date picker widget converted dates even when the user hadn't selected a month. Since the month field is `type="hidden"` and only receives a value when the user clicks the dropdown, typing into the day or year fields triggered `bikram_sambat_bs.initListeners()` conversion with an empty month - which defaulted to Baisakh (month 1), silently saving an incorrect date.

This PR adds a guard that checks all three fields (day, month, year) are filled before allowing conversion to proceed. If any field is missing, the Gregorian output is cleared.

Fixes #8011


megha1807:
Hello @sugat009 
The CI failure is in `shared-libs/settings/test/index.spec.js` (`Settings Shared Library > crypto integration > should handle error gracefully when secret is changed`). This test is unrelated to the BS datepicker widget changed in this PR - my diff only touches `webapp/src/js/enketo/widgets/bikram-sambat-datepicker.js` and its spec file.

This appears to be a pre-existing failure on the `10707-dmp-2026-enhance-bikram-sambat-support-in-cht` base branch (at commit `047f5c562`), not something introduced by this change. Could you please confirm whether this is a known issue on the base branch?

sugat009:
@megha1807 I re-ran the CI and the previously failing job is passing now.

megha1807:
> @megha1807 I re-ran the CI and the previously failing job is passing now.

Thanks @sugat009! The remaining 2 failures (`ci-webdriver-default-lowLevel` and `ci-webdriver-default-data`) are in unrelated e2e specs:
- `create-user-for-contacts.replace-user.wdio-spec.js` (login page timeout)
- `ongoing-replication.wdio-spec.js` (language sync timeout)

These don't touch the BS datepicker widget or its test file. They look like flaky webdriver/e2e timeouts unrelated to this PR's diff. Could you re-run these jobs when convenient?

sugat009 (changes requested):
## Review: this does not fully resolve #8011 yet (requesting changes)

Thanks for this, and for the clear explanation of the cause. Your reading of why it happens (the month input is hidden, `initListeners` converts on any field change, and an empty month falls back to Baisakh) is exactly right.

When I ran the built branch in the webapp, though, the original bug still shows up during normal data entry, so the change does not yet do what the description states ("if any field is missing, the Gregorian output is cleared").

### Steps to reproduce (on this branch, in the webapp)

1. Open a form that has a `bikram-sambat` date field.
2. Type a day and a year, and leave the month dropdown untouched.
3. Click **Next**, or tab out of the field.

**Expected (per the PR):** the date stays empty.
**Actual:** a Gregorian date is still saved, because the month silently falls back to Baisakh (month 1). For example, day `15` and year `2081` saves `2024-04-27`.

Screen recording: day and year entered, no month selected, value still saves as `2024-04-27`:

https://github.com/user-attachments/assets/b34fea0b-f979-4351-b674-285538663758

> Note: the `04` in `2024-04-27` is the Gregorian month (April). The Nepali month that gets assumed is Baisakh, the first Nepali month, which is the exact #8011 behavior.

### Why it still happens

The library binds both `change` and `blur` on the day and year inputs, while the new guard only listens for `change`. When you leave a field, the events fire in this order:

```
change → library converts (empty month → Baisakh) → guard clears it        (good)
blur   → library converts AGAIN → 2024-04-27       → guard ignores 'blur'   (wrong value wins)
```

So the guard does clear the date, and then the `blur` conversion immediately writes the wrong date back. Since `blur` fires last, the wrong value is what sticks. (I instrumented the running widget to confirm this sequence.)

### Adding `blur` to the guard is not enough (I checked)

Before suggesting the obvious next step, I tested it: making the guard also listen for `blur` does not fix it. Instrumenting the widget shows why. A delegated `blur` handler runs as a `focusout`, which fires before the library's own native `blur` handler, so the library still writes the wrong date last. The sequence on leaving the year field:

```
change → library converts (Baisakh) → guard clears it             (value is empty)
blur   → delegated guard runs first, clears (already empty)
       → library's native blur handler runs LAST → writes 2024-04-27
```

So this needs more than adding another event to the guard. The behaviour to confirm any fix against: after entering a day and year with no month, the saved value should stay empty. A test that drives the widget through `change` then `blur` would lock that in.

### Tests

The added `bikram-sambat-datepicker.spec.js` will not catch this regression as written:

* It is a `.spec.js` file, but the webapp unit test config (`angular.json`) only includes `**/*.spec.ts`, so it is never picked up by the webapp unit tests.
* It does not import the widget. The `beforeEach` block re-creates the guard inline, so the tests check a copy of the logic rather than `bikram-sambat-datepicker.js`. Removing the guard from the widget would leave these tests green.

A test that drives the real widget through a `change` then `blur` sequence (the steps above) would have caught this, and it is also what would prove any fix actually closes the bug.

### Minor

* There is trailing whitespace and an extra blank line at the end of `bikram-sambat-datepicker.js` (the "add missing newline" commit left a line with two spaces plus a blank line).
* `// Fix for #8011:` is the only comment in this style under `webapp/src`. Consider dropping the prefix and keeping the explanation, which is genuinely helpful.

The diagnosis here is right; it is the handling that needs rework, since the wrong value is written by the library's own conversion on `blur`. Happy to re-review once entering a day and year with no month leaves the field empty through normal use. Thanks again!

megha1807:
> ## Review: this does not fully resolve #8011 yet (requesting changes)
> Thanks for this, and for the clear explanation of the cause. Your reading of why it happens (the month input is hidden, `initListeners` converts on any field change, and an empty month falls back to Baisakh) is exactly right.
> 
> When I ran the built branch in the webapp, though, the original bug still shows up during normal data entry, so the change does not yet do what the description states ("if any field is missing, the Gregorian output is cleared").
> 
> ### Steps to reproduce (on this branch, in the webapp)
> 1. Open a form that has a `bikram-sambat` date field.
> 2. Type a day and a year, and leave the month dropdown untouched.
> 3. Click **Next**, or tab out of the field.
> 
> **Expected (per the PR):** the date stays empty. **Actual:** a Gregorian date is still saved, because the month silently falls back to Baisakh (month 1). For example, day `15` and year `2081` saves `2024-04-27`.
> 
> Screen recording: day and year entered, no month selected, value still saves as `2024-04-27`:
> 
>  2026-06-13.21-07-45.mp4 
> > Note: the `04` in `2024-04-27` is the Gregorian month (April). The Nepali month that gets assumed is Baisakh, the first Nepali month, which is the exact #8011 behavior.
> 
> ### Why it still happens
> The library binds both `change` and `blur` on the day and year inputs, while the new guard only listens for `change`. When you leave a field, the events fire in this order:
> 
> ```
> change → library converts (empty month → Baisakh) → guard clears it        (good)
> blur   → library converts AGAIN → 2024-04-27       → guard ignores 'blur'   (wrong value wins)
> ```
> 
> So the guard does clear the date, and then the `blur` conversion immediately writes the wrong date back. Since `blur` fires last, the wrong value is what sticks. (I instrumented the running widget to confirm this sequence.)
> 
> ### Adding `blur` to the guard is not enough (I checked)
> Before suggesting the obvious next step, I tested it: making the guard also listen for `blur` does not fix it. Instrumenting the widget shows why. A delegated `blur` handler runs as a `focusout`, which fires before the library's own native `blur` handler, so the library still writes the wrong date last. The sequence on leaving the year field:
> 
> ```
> change → library converts (Baisakh) → guard clears it             (value is empty)
> blur   → delegated guard runs first, clears (already empty)
>        → library's native blur handler runs LAST → writes 2024-04-27
> ```
> 
> So this needs more than adding another event to the guard. The behaviour to confirm any fix against: after entering a day and year with no month, the saved value should stay empty. A test that drives the widget through `change` then `blur` would lock that in.
> 
> ### Tests
> The added `bikram-sambat-datepicker.spec.js` will not catch this regression as written:
> 
> * It is a `.spec.js` file, but the webapp unit test config (`angular.json`) only includes `**/*.spec.ts`, so it is never picked up by the webapp unit tests.
> * It does not import the widget. The `beforeEach` block re-creates the guard inline, so the tests check a copy of the logic rather than `bikram-sambat-datepicker.js`. Removing the guard from the widget would leave these tests green.
> 
> A test that drives the real widget through a `change` then `blur` sequence (the steps above) would have caught this, and it is also what would prove any fix actually closes the bug.
> 
> ### Minor
> * There is trailing whitespace and an extra blank line at the end of `bikram-sambat-datepicker.js` (the "add missing newline" commit left a line with two spaces plus a blank line).
> * `// Fix for #8011:` is the only comment in this style under `webapp/src`. Consider dropping the prefix and keeping the explanation, which is genuinely helpful.
> 
> The diagnosis here is right; it is the handling that needs rework, since the wrong value is written by the library's own conversion on `blur`. Happy to re-review once entering a day and year with no month leaves the field empty through normal use. Thanks again!

Thanks for the detailed repro and instrumentation - this is really helpful. You're right that the `blur` handler in `initListeners` re-converts after my guard clears the value, and that delegated `blur`/`focusout` doesn't win the race.

I'm going to try a `setTimeout`-deferred re-check after `blur` so my guard runs after the library's own blur handler completes, and rewrite the test to actually import and exercise the real widget through `change` + `blur`. Will also fix the trailing whitespace/EOF issue and the comment style. Will push an update shortly - thanks for catching this before merge!

megha1807:
Hello @sugat009 
Pushed updates addressing the review:
- Added a deferred re-check (via setTimeout) on both `change` and `blur` so the guard wins the race against the library's own blur handler that was overwriting the cleared value.
- Replaced the old `.spec.js` (which wasn't picked up by the karma config, and didn't exercise the real widget) with a new `.spec.ts` that imports `bikram-sambat-datepicker.js` directly and drives it through `change` + `blur`, following the pattern used in `phone-widget.spec.ts`.
- Fixed the trailing whitespace/EOF issue.
- Added a cross-platform compatibility fix in `karma-unit.conf.js` to handle both forward slashes and backslashes in the Webpack XML file regex loader, allowing the unit tests to build and run successfully on Windows.

The lint/compile check now passes. The remaining failures (`ci-webdriver-default-lowLevel`, `ci-webdriver-default-data`, `ci-webdriver-default-core-minimum-browser` ) are in unrelated specs (`create-user-for-contacts.replace-user`, `ongoing-replication`, `mark-for-outbound`) that don't touch the datepicker widget - these appear to be pre-existing flaky tests on this branch. Happy to re-run if needed.

sugat009 (approved):
Re-reviewed and confirmed: this resolves #8011. I built this branch's webapp and verified it end to end. Day and year with no month no longer saves a date (the field stays empty and required validation blocks submit, instead of silently saving a Baisakh-defaulted date), and a complete date saves the correct Gregorian value.

**suggestion (non-blocking):** worth adding an end-to-end test for the actual scenario from the ticket: fill day and year without selecting a month, submit, and assert the saved date is empty (plus a complete date that saves correctly). The current unit tests check the widget in isolation but not the submit path. Here is one that reuses the existing `bikram-sambat-dates` form and the helpers already in `tests/e2e/default/translations/nepali-dates-and-numbers.wdio-spec.js`:

```js
it('does not save a date when day and year are entered without a month (#8011)', async () => {
  const fillAndSubmit = async ({ withMonth }) => {
    await commonPage.goToReports();
    await commonPage.openFastActionReport('bikram-sambat-dates', false);
    const realInput = $('input[name="/bikram-sambat-dates/data/date2"]');
    const dateWidget = await realInput.nextElement();
    await dateWidget.$('input[name="day"]').setValue('15');
    if (withMonth) {
      await dateWidget.$('.dropdown-toggle').click();
      await dateWidget.$$('.dropdown-menu li')[3].click(); // असार (month 4)
    }
    await dateWidget.$('input[name="year"]').setValue('2081');
    // Submit directly while the year field is focused, so the blur and serialization happen on the
    // real submit path. A separate blur step would let the deferred clear run first and mask the issue.
    await genericForm.submitForm();
  };

  await fillAndSubmit({ withMonth: false }); // the #8011 case
  await fillAndSubmit({ withMonth: true });  // control: a complete date must still save

  const reports = (await getReports()).filter(r => r.form === 'bikram-sambat-dates');
  const date2vals = reports.map(r => r.fields?.data?.date2);
  expect(date2vals.filter(v => !v).length).to.equal(1);   // incomplete: saved empty
  expect(date2vals.filter(v => !!v).length).to.equal(1);  // complete: saved a date
});
```

This drives the real submit path and asserts the saved value, which the current unit tests do not cover.

Nice work, this is a real fix.

megha1807:
@sugat009 happy to add the e2e test before merge to keep everything together - will push it shortly.

tests/e2e/default/translations/nepali-dates-and-numbers.wdio-spec.js
```diff
@@ -260,4 +260,31 @@ describe('Bikram Sambat date display', () => {
...
+    const testStart = Date.now();
+    const fillAndSubmit = async ({ withMonth }) => {
+      await commonPage.goToReports();
+      await commonPage.openFastActionReport('bikram-sambat-dates', false);
+      const realInput = $('input[name="/bikram-sambat-dates/data/date2"]');
+      const dateWidget = await realInput.nextElement();
+      await dateWidget.$('input[name="day"]').setValue('15');
+      if (withMonth) {
+        await dateWidget.$('.dropdown-toggle').click();
+        await dateWidget.$$('.dropdown-menu li')[3].click(); // असार (month 4)
```
sugat009: nitpick (non-blocking): `[3]` is साउन (month 4), not असार (असार is month 3 / index 2). the month number is fine for the control case, but the comment names the wrong month. worth fixing the comment to साउन (or using `[2]` if असार was intended).

webapp/tests/karma/karma-unit.conf.js
```diff
@@ -16,7 +16,7 @@ module.exports = function (config) {
 
   // allow to require xml files as strings
   config.buildWebpack.webpackConfig.module.rules.push({
-    test: /enketo-xml\/.*\.xml$/i,
+    test: /enketo-xml[/\\](.*)\.xml$/i,
```
sugat009, line 19: question (non-blocking): this now matches both `/` and `\`, so it covers Windows paths too. could you add a short inline comment saying that? without it the `[/\\]` reads as incidental and someone might "simplify" it back to `/` later and break it for Windows devs.

sugat009 (approved):

tests/e2e/default/translations/nepali-dates-and-numbers.wdio-spec.js
megha1807: Good catch! Will fix the comment to साउन to match index [3].

webapp/tests/karma/karma-unit.conf.js
megha1807: Sure, will add a comment clarifying that `[/\\]` is intentional to support both Unix and Windows path separators.

megha1807:
Hi @sugat009 I've addressed both nitpicks - fixed the month comment to साउन and added an inline comment explaining the [/\\] regex. Please have a look when you get a chance.

PR #11167, opened 2026-06-12
feat(#10626): clear deleted contact's parent reference on delete

vikrantwiz02:
## Summary

Follow-up to #11163 (the core delete service). This handles one extra case: when the contact being deleted is the primary contact of a place that isn't itself being deleted, clear that place's `contact` field so it doesn't end up pointing at a deleted doc.

It uses `updateParentContacts` from `@medic/bulk-docs-utils` (set up the same way as in `api/src/services/replication/bulk-docs.js`), and skips any parent that's also in the delete set so nothing gets written twice.

This sits on top of #11163, so the diff here currently includes that commit too. The new work is just the second commit.

One thing worth flagging: cht-conf's `delete-hierarchy.js` doesn't clear parent references, so this goes a little beyond the reference behaviour. I added it so a place isn't left pointing at a contact that no longer exists, but happy to drop it if you'd rather stay closer to cht-conf.

## Changes

- `api/src/services/contact-hierarchy.js`
- `api/tests/mocha/services/contact-hierarchy.spec.js` (3 new tests)

`npm run unit-api`: lint clean, 11 passing.

### DIFF
--- a/api/src/services/contact-hierarchy.js
+++ b/api/src/services/contact-hierarchy.js
@@ -1,5 +1,8 @@
 const db = require('../db');
+const dataContext = require('./data-context');
 const logger = require('@medic/logger');
+
+const bulkDocsUtils = require('@medic/bulk-docs-utils')({ Promise, dataContext });
 
 const BATCH_SIZE = 100;
 const MAX_HIERARCHY_SIZE = 5000;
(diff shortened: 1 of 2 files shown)


vikrantwiz02:
@jkuester and @sugat009, this PR is also ready for review and I'll keep working on next. Thanks!

vikrantwiz02:

PR #11172, opened 2026-06-14
feat(#10626): add DELETE /api/v1/contact/:uuid endpoint

vikrantwiz02:
## Summary

Follow-up to #11163 and #11167. This adds the HTTP layer for the delete service: `DELETE /api/v1/contact/:uuid`.

By default, deleting a contact that still has children is rejected with a 400, so a single call can't wipe out a whole branch of the hierarchy by accident. Pass `?recursive=true` to delete the contact along with its descendants and the reports they're the subject of. An unknown id returns 404. The endpoint is online-only and gated on `can_delete_contacts` or `can_edit`.

The `recursive` option on the service lives in this PR rather than #11163, since it only matters once there's a transport in front of it deciding whether a recursive delete is allowed.

This sits on top of #11163 and #11167.

## Changes

- `api/src/controllers/contact.js` (delete handler + OpenAPI annotation)
- `api/src/routing.js` (route registration)
- `api/src/services/contact-hierarchy.js` (recursive option)
- `api/tests/mocha/controllers/contact.spec.js` (4 new tests)
- `api/tests/mocha/services/contact-hierarchy.spec.js` (1 new test)

`npm run unit-api`: lint clean, service 12 passing, controller 13 passing.

### DIFF
--- a/api/src/controllers/contact.js
+++ b/api/src/controllers/contact.js
@@ -2,6 +2,7 @@ const auth = require('../auth');
 const { Contact, Qualifier } = require('@medic/cht-datasource');
 const ctx = require('../services/data-context');
 const serverUtils = require('../server-utils');
+const contactHierarchy = require('../services/contact-hierarchy');
 
 const getContact = ctx.bind(Contact.v1.get);
 const getContactWithLineage = ctx.bind(Contact.v1.getWithLineage);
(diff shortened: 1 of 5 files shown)


vikrantwiz02:

PR #11176, opened 2026-06-15
test(#10626): add integration tests for DELETE contact endpoint

vikrantwiz02:
## Summary

Follow-up to #11163, #11167 and #11172. This adds the integration tests for the delete work: end-to-end coverage of `DELETE /api/v1/contact/:uuid` against a real API and CouchDB.

The spec seeds a small hierarchy (district > health center > clinics, each with people and a report) and checks the cases the unit tests can't reach end-to-end: a recursive delete removes the contact, its descendants and their reports while leaving ancestors untouched; a leaf delete clears the surviving parent's primary-contact reference; a non-recursive delete of a contact that still has descendants is rejected with 400; an unknown id returns 404; a user without `can_delete_contacts` gets 403; and an offline user gets 403 because the endpoint is online-only.

This sits on top of #11163, #11167 and #11172, so the diff here currently includes those commits too. The new work is just the integration spec.

## Changes

- `tests/integration/api/controllers/contact-hierarchy.spec.js` (new, 6 tests)

`npx eslint`: clean. Integration spec run end-to-end against a local API + CouchDB (full service stack): 6 passing.

### DIFF
--- a/api/src/controllers/contact.js
+++ b/api/src/controllers/contact.js
@@ -2,6 +2,7 @@ const auth = require('../auth');
 const { Contact, Qualifier } = require('@medic/cht-datasource');
 const ctx = require('../services/data-context');
 const serverUtils = require('../server-utils');
+const contactHierarchy = require('../services/contact-hierarchy');
 
 const getContact = ctx.bind(Contact.v1.get);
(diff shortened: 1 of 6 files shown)


vikrantwiz02:

PR #11187, opened 2026-06-17
feat(#10706): port lineage manipulation helpers for hierarchy operations

vikrantwiz02:
## Summary

First bit of the move work, the next operation in #10706 after delete. This one's just groundwork, no endpoint yet. It ports the lineage helpers from cht-conf that the move (and later merge) service will need, so it's only plain functions with tests for now.

These don't exist server-side, so I put them in a new `api/src/services/hierarchy/` folder:

- `replace-lineage.js` rewrites a doc's `parent`/`contact` lineage in place. It's got a `merge` flag for the merge variant too, so this'll cover the merge case later as well.
- `lineage-manipulation.js` has `createLineageFromDoc` and `pluckIdsFromLineage`, and re-exports the replace-lineage functions.

Couple of things I changed from cht-conf, shout if you'd rather I didn't:

- Left out cht-conf's `minifyLineagesInDoc`. It's basically a copy of `minify` from `@medic/lineage`, which we already have here and does a bit more (recursion guard, data_record cleanup), so the move service will just use that instead of carrying another copy.
- Switched the `function` declarations to arrow functions to keep eslint happy. cht-conf doesn't have the `func-style` rule but we do.

## Changes

- `api/src/services/hierarchy/replace-lineage.js`
- `api/src/services/hierarchy/lineage-manipulation.js`
- `api/tests/mocha/services/hierarchy/replace-lineage.spec.js`
- `api/tests/mocha/services/hierarchy/lineage-manipulation.spec.js`

`npm run unit-api`: lint clean, 13 passing.


vikrantwiz02:
Superseded by #11311 👍 `replace-lineage.js` is gone and `lineage-manipulation.js` is down to just `pluckIdsFromLineage`, so everything left in here is already in the move PR.

PR #11190, opened 2026-06-17
Hard limit

### LINKED ISSUE #10408: Put hard limit on number of documents offline users can synchronize
Right now an offline user can be configured to synchronize an unlimited number of documents. While CHWs will [get a warning](https://docs.communityhealthtoolkit.org/releases/3_7_0/#warn-when-replicating-too-many-docs) about synchronizing >10k documents:
* they have little incentive to not continue.  How can they can control the number documents they sync? Should they not try and do their job?
* the cost on the server to synchronize high doc counts can be severely detrimental.  It [has been shown](https://forum.communityhealthtoolkit.org/t/cpu-impacts-of-users-with-10k-documents-how-to-fix-them/5413) that even a few users can negatively impact a server for thousands of other users.

We should add a feature put a hard limit on the number of documents a user can sync.  While Cinque is has just been merged, which should soften the negative impact on the server,  users with high doc counts are still unchecked with no upper bounds.

Admins are very aware of the consequences should be allowed to either raise the limit or turn it off.

Additional training could be done for end users around the meaning of the [warning](https://docs.communityhealthtoolkit.org/releases/3_7_0/#warn-when-replicating-too-many-docs).  Server admins could be more vigilant about monitoring and alerting on "[User over replication limit](https://docs.communityhealthtoolkit.org/hosting/monitoring/dashboards/#overview)".

**Additional context**
There's another issue to [improve monitoring of users above the 10k](https://github.com/medic/cht-core/issues/10407) limit.  Also, the aforementioned [forum post](https://forum.communityhealthtoolkit.org/t/cpu-impacts-of-users-with-10k-documents-how-to-fix-them/5413)  is a real world scenario where a national deployment was having challenges  keeping up with sync demands of users with 100ks of documents.

It may be that the CPU cost to calculate a full, raw document count (tasks, report, contacts etc) is too high and so we may choose to use a more cheap query of just number of contacts as a proxy for all documents.  Additional research could be done to quantify what this max number of contacts is to get numbers based on real world, production data.

A [second issue was filed](https://github.com/medic/cht-core/issues/10406) to try and track failed sync because of too high doc counts.

jkuester:
# Description

Closes https://github.com/medic/cht-core/issues/10408 

# Code review checklist

# License


api/src/services/replication/authorization.js
```diff
@@ -8,7 +8,14 @@ const request = require('@medic/couch-request');
 const environment = require('@medic/environment');
 const nouveau = require('@medic/nouveau');
 const { DOC_IDS, PREFIXES, DOC_TYPES } = require('@medic/constants');
+const { ReplicationLimitError } = require('../../errors');
 
+
+const REPLICATION_LIMITS = {
+  SUBJECTS_COUNT: 100 * 1000,
+  SUBJECT_HITS: 500 * 1000,
+  UNPURGED_DOCS_COUNT: 50 * 1000
```
jkuester, line 17: REP_LIMIT = 100k

SUBJECTS_COUNT = n*1
SUBJECT_HITS = n*5

jkuester: Support disabling limits

jkuester:
## PR Reviewer Guide 🔍

Here are some key observations to aid the review process:

<table>
<tr><td>

**🎫 Ticket compliance analysis ❌**



**[1234](https://github.com/medic/cht-core/issues/1234) - Not compliant**

Non-compliant requirements:

- Nothing in this PR touches enketo/datepicker styling; this ticket appears unrelated (it is referenced only via the PR template example).





**[10408](https://github.com/medic/cht-core/issues/10408) - Partially compliant**

Compliant requirements:

- Hard limits are enforced for subjects count, nouveau subject hits and unpurged doc count
- A `replication_blocked` flag is persisted on the user doc and exposed through user-management APIs

Non-compliant requirements:

- Limits are hard-coded constants; there is no configuration to raise them or disable the feature
- No new tests (unit/integration) for the new limit/blocking behaviour
- No user-facing/structured error response; a blocked user gets a generic `Error` (HTTP 500) with no way to distinguish it

Requires further human verification:

- Whether an admin has a supported flow to unblock a user (`replication_blocked: false`) and whether the webapp handles the failed initial replication gracefully
- Performance impact of fetching the user doc on every `getContext` call



</td></tr>
<tr><td>⏱️&nbsp;<strong>Estimated effort to review</strong>: 3 🔵🔵🔵⚪⚪</td></tr>
<tr><td>🧪&nbsp;<strong>No relevant tests</strong></td></tr>
<tr><td>🔒&nbsp;<strong>No security concerns identified</strong></td></tr>
<tr><td>✅&nbsp;<strong>No TODO sections</strong></td></tr>
<tr><td>🔀 <strong>No multiple PR themes</strong>

</td></tr>
<tr><td>⚡&nbsp;<strong>Recommended focus areas for review</strong><br><br>

<details><summary><a href='https://github.com/medic/cht-core/pull/11190/files#diff-078bd2145bfdcf65dce35971822201dad483da502e3af0ec17e1e7478d499365R647-R656'><strong>Possible Issue</strong></a>

`hits.length + response.hits` adds a number to an array, producing a string (e.g. `"0[object Object],..."`). Comparing that string with `>` against a number yields `NaN > number` which is always false, so the subject-hits limit never triggers, and if it somehow did the error message would contain the serialized array. It should be `hits.length + response.hits.length`.
</summary>

```javascript
// A single doc can have multiple hits (for different subjects), so this is not a strict "doc limit", but
// it is our "give up" limit
if ((hits.length + response.hits) > REPLICATION_LIMITS.SUBJECT_HITS) {
  authorizationContext.userCtx.replicationLimitExceeded = true;
  throw new ReplicationLimitError(
    `User "${authorizationContext.userCtx.name}" exceeds the subject hits limit with at least [${
      hits.length + response.hits
    }] subject hits.`,
  );
}
```

</details>

<details><summary><a href='https://github.com/medic/cht-core/pull/11190/files#diff-a1662c8b1000ef9f521e07c62c363cfe3b86942bb89a8c12aa778ecddc24b9d4R7-R10'><strong>Possible Issue</strong></a>

`const { users, updateUser } = require('@medic/user-management')(config, db, dataContext);` destructures `updateUser` from the top-level module while `getUserDoc` is taken from the nested `users` object. If `updateUser` is only exposed under `users` (as `getUserDoc` is), `updateUser` will be `undefined` and the catch block will throw a `TypeError`, masking the original `ReplicationLimitError`. Please verify the export shape and use a consistent access path.
</summary>

```javascript
const dataContext = require('../data-context');
const config = require('../../config');
const { users, updateUser } = require('@medic/user-management')(config, db, dataContext);
const { ReplicationLimitError } = require('../../errors');
```

</details>

<details><summary><a href='https://github.com/medic/cht-core/pull/11190/files#diff-a1662c8b1000ef9f521e07c62c363cfe3b86942bb89a8c12aa778ecddc24b9d4R12-R17'><strong>Error Handling</strong></a>

A blocked user gets a plain `Error` with no `code`/`status`, which will surface as an HTTP 500 to the client. Since this is an expected, permanent state, a typed error with a 403 status (and a distinguishable payload for the webapp) would let clients react properly instead of retrying forever on a perceived server error.
</summary>

```javascript
const assertUserAllowedToReplicate = async ({ name }) => {
  const { replication_blocked } = await users.getUserDoc(name);
  if (replication_blocked === true) {
    throw new Error(`User "${name}" has been banned from replicating.`);
  }
};
```

</details>

<details><summary><a href='https://github.com/medic/cht-core/pull/11190/files#diff-a1662c8b1000ef9f521e07c62c363cfe3b86942bb89a8c12aa778ecddc24b9d4R55-R60'><strong>Irreversible Side Effect</strong></a>

Exceeding a limit permanently writes `replication_blocked: true` to the user doc, and there is no automatic re-evaluation: even after the user's data shrinks below the limit (purge, re-assignment), the user stays blocked until an admin manually edits the doc. Given the limits are hard-coded and not configurable, an unexpected mass-blocking incident would require manual remediation per user.
</summary>

```javascript
} catch (e) {
  if (e instanceof ReplicationLimitError) {
    await updateUser(userCtx.name, { replication_blocked: true }, true);
  }
  throw e;
}
```

</details>

<details><summary><a href='https://github.com/medic/cht-core/pull/11190/files#diff-078bd2145bfdcf65dce35971822201dad483da502e3af0ec17e1e7478d499365R635-R641'><strong>Nouveau Limit</strong></a>

The per-request nouveau `limit` was changed from `nouveau.RESULTS_LIMIT` to `500000`. Nouveau/CouchDB enforces a server-side maximum for the search `limit` parameter; if 500000 exceeds it, the request will fail with a 400 for every replication instead of returning results. Worth confirming against the configured nouveau max limit, and also that fetching up to 500k hits per chunk into memory is acceptable.
</summary>

```javascript
const response = await request.post({
  uri: `${environment.couchUrl}/_design/medic/_nouveau/docs_by_replication_key`,
  body: {
    q: `key:(${chunk.map(nouveau.escapeKeys).join(' OR ')})`,
    limit: REPLICATION_LIMITS.SUBJECT_HITS,
  }
});
```

</details>

</td></tr>
</table>

api/src/services/replication/authorization.js
```diff
@@ -629,14 +636,25 @@ const getDocsByReplicationKeyNouveau = async (authorizationContext) => {
...
+    // A single doc can have multiple hits (for different subjects), so this is not a strict "doc limit", but
+    // it is our "give up" limit
+    if ((hits.length + response.hits) > REPLICATION_LIMITS.SUBJECT_HITS) {
+      authorizationContext.userCtx.replicationLimitExceeded = true;
+      throw new ReplicationLimitError(
+        `User "${authorizationContext.userCtx.name}" exceeds the subject hits limit with at least [${
+          hits.length + response.hits
+        }] subject hits.`,
+      );
+    }
```
jkuester, line 656: **Suggestion:** `response.hits` is an array, so `hits.length + response.hits` performs string concatenation instead of numeric addition, making the comparison (and the logged count) meaningless. Use `response.hits.length` so the limit check actually compares numbers. [possible bug, importance: 9]
```suggestion
    const totalHits = hits.length + response.hits.length;
    if (totalHits > REPLICATION_LIMITS.SUBJECT_HITS) {
      authorizationContext.userCtx.replicationLimitExceeded = true;
      throw new ReplicationLimitError(
        `User "${authorizationContext.userCtx.name}" exceeds the subject hits limit with at least [${
          totalHits
        }] subject hits.`,
      );
    }
```

api/src/services/replication/authorization.js
```diff
@@ -629,14 +636,25 @@ const getDocsByReplicationKeyNouveau = async (authorizationContext) => {
       uri: `${environment.couchUrl}/_design/medic/_nouveau/docs_by_replication_key`,
       body: {
         q: `key:(${chunk.map(nouveau.escapeKeys).join(' OR ')})`,
-        limit: nouveau.RESULTS_LIMIT,
+        limit: REPLICATION_LIMITS.SUBJECT_HITS,
```
jkuester, line 639: **Suggestion:** Setting the query `limit` equal to the threshold means a single chunk that hits exactly the limit gets silently truncated and passes the `>` check, so documents would be dropped without any error. Request one more than the threshold so truncation is always detectable. [possible bug, importance: 7]
```suggestion
        q: `key:(${chunk.map(nouveau.escapeKeys).join(' OR ')})`,
        limit: REPLICATION_LIMITS.SUBJECT_HITS + 1,
```

api/src/services/replication/replication.js
```diff
@@ -1,33 +1,63 @@
...
+      warn: unpurgedWarnIds.length >= replicationLimitLog.DOC_IDS_WARN_LIMIT,
+      limit: replicationLimitLog.DOC_IDS_WARN_LIMIT,
+      lastSeq: info.update_seq,
+    };
+  } catch (e) {
+    if (e instanceof ReplicationLimitError) {
+      await updateUser(userCtx.name, { replication_blocked: true }, true);
+    }
+    throw e;
+  }
```
jkuester, line 60: **Suggestion:** If `updateUser` fails (e.g. conflict or network error), its rejection replaces the original `ReplicationLimitError`, hiding the real reason replication was refused. Guard the blocking update so the original error is always propagated. [possible issue, importance: 6]
```suggestion
  } catch (e) {
    if (e instanceof ReplicationLimitError) {
      try {
        await updateUser(userCtx.name, { replication_blocked: true }, true);
      } catch (updateErr) {
        console.error(`Failed to block replication for user "${userCtx.name}"`, updateErr);
      }
    }
    throw e;
  }
```

api/src/errors.js
```diff
@@ -27,9 +27,13 @@ class AuthenticationError extends Error {
   }
 }
 
+class ReplicationLimitError extends Error {
```
jkuester, line 30: **Suggestion:** Proposed documentation
```suggestion
/**
 * Error thrown when a user exceeds one of the configured replication limits
 * (for example: too many subjects, too many subject hits or too many unpurged documents).
 * @extends Error
 */
class ReplicationLimitError extends Error {
```

api/src/services/replication/replication.js
```diff
@@ -1,33 +1,63 @@
...
+const { ReplicationLimitError } = require('../../errors');
+
+const assertUserAllowedToReplicate = async ({ name }) => {
+  const { replication_blocked } = await users.getUserDoc(name);
+  if (replication_blocked === true) {
+    throw new Error(`User "${name}" has been banned from replicating.`);
+  }
+};
 
 const getContext = async (userCtx) => {
```
jkuester, line 19: **Suggestion:** Proposed documentation
```suggestion
/**
 * Builds the replication context for the given user.
 * Validates that the user is not blocked from replicating and that the user does not exceed
 * the configured replication limits. When a limit is exceeded, the user is blocked from
 * replicating and a ReplicationLimitError is thrown.
 * @param {Object} userCtx - the user context. Mutated to include `subjectsCount`, `docsCount`
 *                           and `unpurgedDocsCount`.
 * @param {string} userCtx.name - the username.
 * @returns {Promise<Object>} the replication context containing:
 *                            `docIds`, `warnDocIds`, `warn`, `limit` and `lastSeq`.
 * @throws {Error} when the user has been blocked from replicating.
 * @throws {ReplicationLimitError} when the user exceeds a replication limit.
 */
const getContext = async (userCtx) => {
```

PR #11193, opened 2026-06-18
fix(#11184): update task filter UI to match reports filter

### LINKED ISSUE #11184: Update 3 filter UIs match each other: tasks, reports & targets
We have over time added task filters, reports filters and target filters.  however the UIs differ slightly, and they should all be the same.

We should make the new task UI filters match that of existing Target and report filter UI.  Specifically, these three changes should be made

1.  changing the two "Due date" options to radio boxes [instead of check boxes](https://github.com/user-attachments/assets/52401e1a-f71a-4097-a751-bd5e88a5d934). I believe these are two mutually exclusive sets where you'd never want to see both, correct? 
2. using the same blue "Apply" button that we do in [report](https://github.com/user-attachments/assets/a89823e6-a997-42d2-b875-d63411fe2324) and target filtering.  see [rough mockup](https://github.com/user-attachments/assets/709ed5d6-5d22-40ad-a1d8-0bd98243085b) where it should go.
3. The top "[Clear all filters](https://github.com/user-attachments/assets/f9863896-502b-40ea-815c-b9c8f6bc118c)" should match existing Reports filter UI which [uses](https://github.com/user-attachments/assets/a89823e6-a997-42d2-b875-d63411fe2324) "Reset" 

leave UI as is

**Additional context**
the task filters haven't shipped yet so we might consider making the improvements  in this ticket in 5.2 as well to persist the feature parity. 

Here's a screenshot:

Image

santhosh-7777:
# Description
Updated the task filter UI to match the existing reports filter UI.

Fixes #11184
Changes made:
- Changed Due date options from clickable list items to radio buttons
- Added Apply button footer to tasks sidebar filter
- Changed "Clear all filters" label to "Reset"

# Code review checklist


santhosh-7777:
@mrjones-plip  The "Compile the app" failure is a CI infrastructure issue 
(Docker port binding error: "address already in use"), 
not related to my code changes.

mrjones-plip (changes requested):
@santhosh-7777 thanks for the PR!

Please:
* update your branch with most recent changes from `master`. I see your branch is out of date
* manually confirm you code changes work.  When I load your branch locally nothing has changed.
* make sure your tests actually test what you changed.  Your test now checks for a `label` so the tests passes, but it does not actually test that the form uses a radio box.
 
To this end, please provide a short 5-10  second video showing you've successfully implemented the requested changes on the three filter UIs. This will save the time of the reviewer before they start to look at the code.

santhosh-7777:
@mrjones-plip Added a short demo video showing the three UI changes working (radio buttons for Due date, single-select behavior, Apply button on mobile view, and Reset label).
Also updated the test in filterByOverdue to directly click input[type="radio"] instead of just the label, ensuring the test actually validates the radio button interaction.

santhosh-7777:
https://github.com/user-attachments/assets/27151771-005d-49eb-80bb-ad45fa1f4e51

mrjones-plip (changes requested):
Thanks @santhosh-7777 ! Can you make the radio button values not bold (`#1`)?  As well, it appears maybe "place" is missing  (`#2`)?

Image

santhosh-7777:
@mrjones-plip Fixed the bold styling on radio button labels (#1). For the Place filter (#2) , this is existing behavior controlled by *ngIf="showPlaceFilter" which calls hasMultiplePlaces(). It only shows when the user has multiple facilities. In my demo video the test user only had one facility (Test District), so Place was correctly hidden.

webapp/src/ts/components/filters/overdue-filter/overdue-filter.component.html
```diff
@@ -4,7 +4,7 @@
       *ngFor="let status of statuses"
       [class.selected]="filter?.selected.has(status)"
       [class.disabled]="disabled">
-      <label>
+      <label style="font-weight: normal;">
```
mrjones-plip: thanks for this fix! I don't see any inline styles in any of the `cht-core/webapp/src/ts/*` files - can you please move this change to the appropriate file in `cht-core/webapp/src/css/*`?

Thanks!

mrjones-plip (changes requested):
Thanks for the explanation about the missing place.

see request about moving CSS deceleration.

webapp/src/ts/components/filters/overdue-filter/overdue-filter.component.html
santhosh-7777: Moved the font-weight: normal style from inline HTML to the label block in webapp/src/css/sidebar-filter.less as requested.

santhosh-7777:
@mrjones-plip i have made requested changes could u please review . happy to make if changes are needed. thank you!

mrjones-plip (approved):
Thanks @santhosh-7777 - LGTM

Issue #11194, opened 2026-06-18
Improve flaky test resolution process


dianabarsan:
**Describe the issue**
Currently the flaky test resolution process constitutes of: 
1. test fails 
2. human observes test fails
3. human roughly verifies if this is an isolated flake, caused by work in progress or frequent flake
4. human opens ticket on GH to fix test
.....
12. human fixes test

**Describe the improvement you'd like**
I want to introduce some automation in the first part of the process by (ideally): 

1. test fails on `master`
2. have a GHA that runs a script that checks if the test failed before
3. GHA opens a ticket (if it doesn't exist yet) to fix the test with details about the failed test and previous failures list. 
4. if ticket already exists, GHA adds a comment with the details about the most recent failures.

mrjones-plip:
@dianabarsan - I think your [PR that you closed](https://github.com/medic/cht-core/pull/11195) addresses all of this ticket - yes?  Closing it now, but please re-open if you have more PRs you'd like to associate with this issue!

mrjones-plip:
@andrablaj - I'm carefully adding this to 5.3 because it's not a `chore` (it's `ci`) based on the last [convo](https://medic.slack.com/archives/C03H77DKY0M/p1772560107524339?thread_ts=1772155075.106489&cid=C03H77DKY0M) - lemme know if I got it wrong though 😅

PR #11195, opened 2026-06-18
ci(#11194): introduce CI failure analysis script

dianabarsan:
# Description

Claude wrote the script. 

#11194 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:11194-add-test-failure-scanner/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:11194-add-test-failure-scanner/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:11194-add-test-failure-scanner/docker-compose/cht-couchdb-clustered.yml)

# License


dianabarsan:
@jkuester this is the script I referred to in the call last week. I've been using it locally very successfully. The GH workflow is a WIP but without having it in `master`, I can't trigger it.

.github/workflows/find-test-failures.yml
```diff
@@ -0,0 +1,66 @@
...
+          args=("$SEARCH_TERM" --repo "$GITHUB_REPOSITORY" --limit "$LIMIT")
+          [ -n "$JOB" ] && args+=(--job "$JOB")
+          [ -n "$BRANCH" ] && args+=(--branch "$BRANCH")
+          [ -n "$SINCE" ] && args+=(--since "$SINCE")
+          [ "$SCAN_ALL" = "true" ] && args+=(--scan-all)
+          {
+            echo '```'
+            node scripts/ci/find-test-failures.js "${args[@]}"
+            echo '```'
+          } | tee -a "$GITHUB_STEP_SUMMARY"
```
jkuester, line 66: Fancy!  TIL about `$GITHUB_STEP_SUMMARY`!

scripts/ci/find-test-failures.js
```diff
@@ -0,0 +1,613 @@
...
+};
+
+if (require.main === module) {
+  main().catch((err) => {
+    process.stderr.write(`\nError: ${err.message}\n`);
+    process.exit(1);
+  });
+}
+
+module.exports = { parseWdioLog, parseMochaLog, cleanLine, buildMatcher };
```
jkuester: Any particular reason to export these (and have the `require.main === module` logic above)?  Are you thinking that we might reuse this in a different script in the future?

scripts/ci/find-test-failures.js
```diff
@@ -0,0 +1,613 @@
...
+  }
+  if (spec.flag) {
+    opts[spec.key] = true;
+    return i;
+  }
+  opts[spec.key] = spec.parse(argv[i + 1]);
+  return i + 1;
+};
+
+const parseArgs = (argv) => {
```
jkuester, line 79: We already have `minimist` as a root dependency.  Seems a bit unnecessary to roll-our-own here...

scripts/ci/find-test-failures.js
```diff
@@ -0,0 +1,613 @@
...
+  --json              Emit JSON instead of a table.
+  --debug             Print parsing diagnostics to stderr.
+  -h, --help          Show this help.
+`;
+
+// ---------------------------------------------------------------------------
+// gh helpers
+// ---------------------------------------------------------------------------
+
+const runGh = (args, { allowFail = false, maxBuffer = 256 * 1024 * 1024 } = {}) => {
```
jkuester: Looks like `maxBuffer` is never set to anything but the default. Could probably just hardcode it in the function.

scripts/ci/find-test-failures.js
```diff
@@ -0,0 +1,613 @@
...
+    const { failure, next } = parseMochaEntry(clean, i, header[2]);
+    failures.push(failure);
+    i = next;
+  }
+  return failures;
+};
+
+const isIntegrationJob = (jobName) => /ci-integration/i.test(jobName);
+
+const TEST_JOB_RE = /(ci-webdriver|wdio-performance|ci-integration|test-cht-form|integration-cht-form|upgrade-)/i;
```
jkuester: Probably not much that can be done here, but this is a maintainability concern since if/when we update this naming in the workflow, we need to remember to update this.  Seems at least worth moving this up towards to top instead of having it buried here.

jkuester (approved):
LGTM (sorry for the delayed review)!  Some minor comments inline, but all in all this is very useful! I tested locally and it works as expected.  

I will say that if we want to continue building similar scripts I would prefer it if we had a more structured approach.  Dumping everything in a one-off script like this is not easy to maintain. Even the current release-notes script is really annoying and this one is 50% longer. 😓 40+ top-level functions in a single module is a nightmare to read through.  At a certain point, my brain shut off and I gave up trying to track the flow through all the levels....  (Also, I think 500 lines is my new JS limit. 😅 Anything more than that should be TS. Trying to keep the mental context of the shape of all the data flowing around here was impossible.)

FWIW, the main goal behind the structuring of https://github.com/jkuester/chtoolbox was that it could be a framework for one-off scripts like this where it takes care of things like cli args, etc while also allowing for re-use of common services (like interacting with GitHub).  I have been hesitant pull that into the `medic` org because there has not been universal buy-in on the Effect framework. 😬 But maybe in the brave new AI world that is less of a problem for folks. 🤷  Alternatively, we could just green-field a new non-Effect repo with similar goals......

dianabarsan:
I added a temporary trigger (this is needed to be able to run the workout pre-main-merge). 
https://github.com/medic/cht-core/actions/runs/30025155113 result visible here.

PR #11209, opened 2026-06-22
feat(#11146): adds replication-health/failures endpoint

### LINKED ISSUE #11146: Add API endpoint that returns which users have no replication successes within an interval
To determine which users fail to replicate, a user needs to cross-check user replication count docs with failure logs and check for the timestamp overlap. 

Provide an endpoint that does this calculation and provides a prepared list of users who consistently fail to replicate.

dianabarsan:
# Description

 Adds an admin-only API endpoint to identify users who are failing to replicate, to help operators troubleshoot replication issues (e.g. tracking down heavy-doc users degrading server performance).
Adds missing integration tests for limit log endpoint. 

  Closes #11146.

  ### New endpoint

  `GET /api/v1/replication-health/failed` (DB admins only) — returns users that appear unable to replicate within a recent window.

  Query parameters:
  - `days` (default `30`) — staleness window; a user is considered only if they have **not** successfully replicated within this many days.
  - `min_failures` (default `1`) — minimum replication failures **within the window** for a user to be listed.

  A user is listed when both hold:
  1. they logged at least `min_failures` replication failures inside the window, and
  2. they have not successfully replicated within the window — either their replication-count log is older than the cutoff, or they have no replication-count log at all (never successfully replicated).

  Each entry returns `user`, `last_replication_date` (null if never replicated), `failures_in_window` (what `min_failures` is applied to), and `failures_since_last_replication` (context; null if never replicated).

  ### Supporting changes

**`users-doc-count` limit log now always updates.** Previously the replication-count log was only rewritten when doc counts/dates drifted past a threshold; it now rewrites on every successful replication so its `date` reliably reflects the last successful replication, which the new endpoint depends on.

## AI disclosure

Code developed with the help of Claude. 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:11146-users-that-cant-replicate-api/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:11146-users-that-cant-replicate-api/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:11146-users-that-cant-replicate-api/docker-compose/cht-couchdb-clustered.yml)

# License


api/src/routing.js
```diff
@@ -796,6 +797,7 @@ app.put(
 
 app.get('/api/v1/users-doc-count', replicationLimitLogController.get);
 app.get('/api/v1/replication-failure-logs', replicationFailureLogController.get);
+app.get('/api/v1/replication-health/failed', replicationHealthController.failed);
```
jkuester, line 800: Is the idea that we might add other replication-health/* endpoints in the future?

api/src/controllers/replication-failure-log.js
```diff
@@ -110,10 +109,7 @@ module.exports = {
    */
   get: async (req, res) => {
     try {
-      const userCtx = await auth.getUserCtx(req);
-      if (!auth.isDbAdmin(userCtx)) {
-        throw new errors.AuthenticationError('User is not an admin');
-      }
+      await auth.assertDbAdmin(req);
```
jkuester, line 112: Out of curiosity, do you know if there is a reason we do this kind of checking here in the controller code and not as a middleware added in the routing.js that runs automatically for the endpoint?

It always seems a bit odd to me that we have security logic here in the controllers...

api/tests/mocha/services/replication/replication-limit-log.spec.js
```diff
@@ -46,6 +46,28 @@ describe('Replication Limit Log service', () => {
...
+      };
+      const getStub = sinon.stub(db.medicLogs, 'get').returns(Promise.resolve(existingDoc));
+      const putStub = sinon.stub(db.medicLogs, 'put').returns(Promise.resolve());
+
+      await replicationLimitLogService.put('userXYZ', 100, 500);
+
+      chai.expect(getStub.called).to.be.true;
+      chai.expect(putStub.called).to.be.true;
+      // Keeps the existing _rev (updates in place) and refreshes the date.
+      chai.expect(putStub.args[0][0]._rev).to.equal('1-abc');
```
jkuester:
```suggestion
```

I think this is just noise (at least the comment) since you must pass the same _rev when you put the doc.

api/src/controllers/replication-health.js
```diff
@@ -0,0 +1,99 @@
...
+
+module.exports = {
+  /**
+   * @openapi
+   * /api/v1/replication-health/failed:
+   *   get:
+   *     summary: Get users that are failing to replicate
+   *     operationId: v1ReplicationHealthFailedGet
+   *     description: >
+   *       Returns the users that appear unable to replicate within the window: their last successful
```
jkuester, line 14:
```suggestion
   *      Returns the users that were unable to replicate within the window: their last successful
```

We don't need for ambiguity here, right?

api/src/controllers/replication-health.js
```diff
@@ -0,0 +1,99 @@
+const { InvalidArgumentError } = require('@medic/cht-datasource');
+const auth = require('../auth');
+const serverUtils = require('../server-utils');
+const replicationHealth = require('../services/replication/replication-health');
+
+const parsePositiveInteger = (value, name) => {
```
jkuester: This is basically the same as the `parseIntegerParam` function in `services/pagination.js`.  I know we are not really paging here, but would be nice to reuse somehow.  Not somehting we have to do now though....

api/src/auth.js
```diff
@@ -40,9 +40,18 @@ const assertPermissions = async (req, { isOnline = false, hasAll = [], hasAny =
   return userCtx;
 };
 
+const assertDbAdmin = async (req) => {
+  const userCtx = await module.exports.getUserCtx(req);
+  if (!roles.isDbAdmin(userCtx)) {
+    throw new AuthenticationError('User is not an admin');
```
jkuester: I think this should be a `PermissionError`, not `AuthenticationError`, right?  If the user is logged in, but is not an admin we should give back a `403` not a `401`.

api/src/services/replication/replication-health.js
```diff
@@ -0,0 +1,91 @@
+const moment = require('moment');
+const replicationLimitLog = require('./replication-limit-log');
+const replicationFailureLog = require('./replication-failure-log');
+
+const DAY_FORMAT = 'YYYY-MM-DD';
+// Default staleness window: a user who has not successfully replicated within this many days is
+// considered for the failed list when the caller doesn't supply an explicit `days` value.
```
jkuester: So, coming from the Java world, I was used to having loads of code comments and "self-documenting" code was not really a thing.

After a while I got used to trying to write hyper-readable JS code that did not need a bunch of explanatory comments.

Now, Claude seems to be a big fan of adding comments like this. I find myself going back and forth on removing them when I am accepting code from Claude. One one hand, the comments tend to be accurate and helpful. On the other hand I just cannot help shaking the feeling that we should not need them if the code is clearly written...  😅 

Not saying we need to do anything here, but would be curious to hear your opinions!

jkuester: I do think the danger of them becoming stale and misleading in the future is a real cost, but it might not outweigh the benefits...

jkuester: Also this comment is really only referencing inline comments and comments on internal functions.  Proper JSDoc on exported functions has clear benefits.

api/src/services/replication/replication-health.js
```diff
@@ -0,0 +1,91 @@
...
+  return new Map(failing
+    .filter(entry => logsByUser[entry.user]?.date)
+    .map(entry => {
+      const sinceDay = moment(logsByUser[entry.user].date).format(DAY_FORMAT);
+      return [entry.user, sumFailuresSince(failuresByUser[entry.user], sinceDay)];
+    }));
+};
+
+/**
+ * Lists users that appear unable to replicate within the `days` window: they have logged at least
```
jkuester:
```suggestion
 * Lists users that were unable to replicate within the `days` window: they have logged at least
```

api/src/services/replication/replication-health.js
```diff
@@ -0,0 +1,91 @@
...
+// "Any number of failures counts" — flag a user with at least one failure within the window.
+const DEFAULT_MIN_FAILURES = 1;
+
+const sumFailuresSince = (dailyFailures, sinceDay) => {
+  return Object.keys(dailyFailures || {})
+    .filter(day => day >= sinceDay)
+    .reduce((total, day) => total + (dailyFailures[day] || 0), 0);
+};
+
+const sumFailures = (dailyFailures = {}) => Object.values(dailyFailures).reduce((total, n) => total + n, 0);
```
jkuester, line 15: `#free_reduce`  ✊

api/src/services/replication/replication-health.js
```diff
@@ -0,0 +1,91 @@
...
+  const logDates = failing
+    .map(entry => logsByUser[entry.user]?.date)
+    .filter(Boolean);
+  if (!logDates.length) {
+    return new Map();
+  }
+
+  const earliestDay = moment(Math.min(...logDates)).format(DAY_FORMAT);
+  const failuresByUser = await replicationFailureLog.getDailyFailuresByUserSince(earliestDay);
+  return new Map(failing
```
jkuester: Any particular reason for using a Map here instead of just a object dict like you have everywhere else?  Not a problem, but was just a bit unexpected.

(Also cannot help but feel like the value of the Map here is you can build it inline with no for-loop.  We could do the same thing for object dicts using `reduce`!  🤝)

api/tests/mocha/services/replication/replication-health.spec.js
```diff
@@ -0,0 +1,161 @@
...
+  afterEach(() => {
+    sinon.restore();
+  });
+
+  // Now: 2026-06-16. Default 30-day window → cutoff day 2026-05-17.
+  const fakeNow = () => sinon.useFakeTimers(new Date('2026-06-16T12:00:00Z').valueOf());
+
+  describe('getFailed', () => {
+    it('should query failures within the window using the cutoff day', async () => {
+      fakeNow();
```
jkuester: Can we call this in a beforeEach?

api/tests/mocha/services/replication/replication-health.spec.js
```diff
@@ -0,0 +1,161 @@
...
+    it('should propagate errors from the failure log query', async () => {
+      fakeNow();
+      getDailyFailuresByUserSince.rejects({ status: 500 });
+
+      try {
+        await service.getFailed();
+        expect.fail('should have thrown');
+      } catch (err) {
+        expect(err).to.deep.equal({ status: 500 });
+      }
```
jkuester: 🙏  `await expect(service.getFailed()).to.be.rejectedWith....` is the only non-cursed way to do it. If you do it this way, trolls will attack the data center!

tests/integration/api/controllers/replication-limit-log.spec.js
```diff
@@ -0,0 +1,98 @@
...
+  });
+
+  it('should return only the requested user when filtered by user', async () => {
+    await utils.logsDb.put(countLog('alice', 50));
+    await utils.logsDb.put(countLog('bob', 200));
+
+    const response = await getDocCount({ user: 'bob' });
+
+    expect(response.limit).to.equal(DOC_IDS_WARN_LIMIT);
+    // When filtered, `users` is the single matching log document rather than an array.
```
jkuester, line 82: This is why we cannot have nice things...  😅

jkuester (approved):
Most excellent!  Just some minor unblocking comments/questions.  The only one that might be important is returning a `403` when the user is not an admin  (instead of `401`).

api/src/routing.js
dianabarsan: during the call we talked about adding a generic replication-health endpoint. we didn't discuss design, and I had no idea what to name this. I was really in an inspiration impasse and @mrjones-plip came up with the replication-health naming idea.

api/src/controllers/replication-failure-log.js
dianabarsan: My only argument for is that you have instant visibility into the limitation of the endpoint instead of checking the documentation or going to routing. Otherwise, middleware would be much cleaner.

api/tests/mocha/services/replication/replication-limit-log.spec.js
dianabarsan: I think claude added it when i was dissatisfied with its original assertions and i asked it to assert everything :D

api/src/controllers/replication-health.js
dianabarsan: Oh, thanks for that, I had forgotten we added that helper. I'll reuse it!

api/src/auth.js
dianabarsan: ok. Agreed. but then this changes the error code that gets returned to other endpoints as well.

api/src/services/replication/replication-health.js
dianabarsan: I'll remove it. I added maybe 5 memories so it's less verbose and then it keeps adding this stuff regardless. we have a communication problem.

api/src/services/replication/replication-health.js
dianabarsan: let's just free it. pr welcome :D

tests/integration/api/controllers/replication-limit-log.spec.js
dianabarsan: yeap.

api/tests/mocha/services/replication/replication-limit-log.spec.js
jkuester: Claude has already mastered malicious compliance!  :rofl:

PR #11214, opened 2026-06-24
feat(#11174): add getpage and getall endpoints

### LINKED ISSUE #11174: Add document-returning `getPage` / `getAll` to `Contact.v1` and `Report.v1`
## Description

`Contact.v1` and `Report.v1` currently expose only **ID-returning** paginated queries
(`getUuidsPage` / `getUuids`) plus **single-document** lookups (`get` / `getWithLineage`). There is
no qualifier-driven method that returns a *page of documents*. The only document-returning paginated
methods in cht-datasource live on the typed nouns `Person.v1` / `Place.v1` (`getPage` / `getAll`),
and those accept `byContactType` **only**.

This is a blocker for the view→qualifier migration series. Any existing `db.query(view, { include_docs:
true })` call that returns **many** documents keyed by something other than contact-type
(e.g. contacts by phone/freetext, reports by freetext) has no one-shot datasource equivalent today —
it would have to migrate to `getUuids(qualifier)` followed by a per-ID `get`, i.e. an **N+1**. That is
not acceptable for bulk call sites.

This ticket adds document-returning `getPage` / `getAll` to `Contact.v1` and `Report.v1`, accepting
the **same qualifier union** as their existing `getUuidsPage`, so a bulk `include_docs` view call
migrates to a single paginated call. It is a **prerequisite** for the doc-returning view migrations;
it does not itself sweep any callers.

**Components:**
- `shared-libs/cht-datasource/src/contact.ts` — add `getPage` (paginated) + `getAll` (generator),
  accepting `ContactGetUuidsQualifier`, returning `Page<v1.Contact>` / `AsyncGenerator<v1.Contact>`
- `shared-libs/cht-datasource/src/report.ts` — same, accepting the report freetext qualifier,
  returning `Page<v1.Report>`
- `shared-libs/cht-datasource/src/local/contact.ts` — `getPage` dispatch reusing the views already
  used by `getUuidsPage`, doc-returning
- `shared-libs/cht-datasource/src/local/report.ts` — `getPage` dispatch
- `shared-libs/cht-datasource/src/local/libs/doc.ts` — add `queryDocsByKeys` (doc-returning sibling
  of the existing `queryDocIdsByKeys`)
- `shared-libs/cht-datasource/src/remote/contact.ts` / `remote/report.ts` — `getPage` calling the new
  endpoints
- `api/src/controllers/contact.js` / `report.js` — new doc-page handlers, reusing the existing
  declarative qualifier table
- `api/src/routing.js` — wire `GET /api/v1/contact` and `GET /api/v1/report`

**Existing References:**
- `shared-libs/cht-datasource/src/person.ts:78,113` — `getPage` / `getAll` pattern to mirror
- `shared-libs/cht-datasource/src/local/person.ts:90` — local `getPage`: a single `include_docs`
  view query (`queryDocsByKey`) + `fetchAndFilter`. No N+1.
- `shared-libs/cht-datasource/src/local/libs/doc.ts:82` (`queryDocsByKey`), `:23` (`getDocsByIds`,
  bulk `allDocs`), `:135` (`queryDocIdsByKeys` — the ID sibling of the helper to add), `:152`
  (`fetchAndFilter`)
- `api/src/routing.js:660,735` — `GET /api/v1/place` / `GET /api/v1/person` → `getAll` doc-page
  endpoints (the precedent; bare `/api/v1/contact` and `/api/v1/report` GET are currently unused)
- PR for #10973 — added `queryDocIdsByKeys` and the declarative qualifier table in
  `api/src/controllers/contact.js`

### Datasource additions

1. **`Contact.v1.getPage` / `getAll`** mirroring `Person.v1` (`contact.ts`), but typed to the contact
   qualifier union and the `v1.Contact` (person-or-place) return shape:

```typescript
export const getPage = (context: DataContext): typeof curriedFn => {
  assertDataContext(context);
  const fn = adapt(context, Local.Contact.v1.getPage, Remote.Contact.v1.getPage);

  const curriedFn = async (
    qualifier: ContactGetUuidsQualifier,
    cursor: Nullable<string> = null,
    limit: number | `${number}` = DEFAULT_DOCS_PAGE_LIMIT
  ): Promise<Page<v1.Contact>> => {
    assertCursor(cursor);
    assertLimit(limit);
    assertContactGetUuidsQualifier(qualifier);
    return fn(qualifier, cursor, Number(limit));
  };
  return curriedFn;
};
// getAll wraps getPage via getPagedGenerator, exactly like person.ts:113
```

2. **`Report.v1.getPage` / `getAll`** — same shape, report freetext qualifier, `Page<v1.Report>`.

3. **`queryDocsByKeys`** in `local/libs/doc.ts` — doc-returning sibling of `queryDocIdsByKeys`:

```typescript
export const queryDocsByKeys = (
  db: PouchDB.Database<Doc>,
  view: string
) => async (
  keys: unknown[],
  limit: number,
  skip: number
): Promise<Nullable<Doc>[]> => queryDocs(db, view, { include_docs: true, reduce: false, keys, limit, skip });
```

4. **Local dispatch** (`local/contact.ts`), reusing the same views as `getUuidsPage`:
   - `isContactTypeQualifier` / `isPhoneQualifier` → `queryDocsByKey` + `fetchAndFilter(isContact)`
     — single query.
   - `isPhonesQualifier` → `queryDocsByKeys` + `fetchAndFilter(isContact)` — single query.
   - freetext → nouveau/offline returns **IDs** (no doc shortcut), so hydrate the page's IDs with the
     existing `getDocsByIds` (one batched `allDocs`). Bounded to one extra batched call per page —
     **not** N+1 — and the nouveau cursor passes through unchanged.

5. **Remote dispatch** (`remote/contact.ts` / `remote/report.ts`) — `getPage` calling the new
   doc-page endpoints with the same query-param serialization used by the ID-page (incl. the
   comma-joined `phones`).

6. **API endpoints** — `GET /api/v1/contact` and `GET /api/v1/report` returning `{ data, cursor }`
   pages of documents. The contact handler reuses the existing declarative qualifier table so
   validation / mutual-exclusion stays single-sourced. Wire in `routing.js`.

7. Expose convenience methods on the `Contact.v1` / `Report.v1` `Datasource` interfaces
   (e.g. `getPageByType` / `getByType`, `getPageByFreetext` / `getByFreetext`, `getPageByPhone`, …)
   following the existing naming.

### Tests

- Unit: `test/local/contact.spec.ts` / `test/local/report.spec.ts` — each `getPage` dispatch arm hits
  the right view with `include_docs: true` and the right key(s); freetext arm hydrates via
  `getDocsByIds`.
- Unit: `test/remote/contact.spec.ts` / `test/remote/report.spec.ts` — endpoints called with correct
  path/params.
- Unit: `test/contact.spec.ts` / `test/report.spec.ts` — `getPage` asserts cursor/limit, `getAll`
  drains pages.
- Controller: `api/tests/mocha/controllers/{contact,report}.spec.js` — doc-page handler builds the
  qualifier from the table and returns the page.
- Integration: `tests/integration/api/controllers/{contact,report}.spec.js` — walk two cursor pages
  with `limit: 5` for a type and a freetext query.

## Acceptance Criteria

- `Contact.v1.getPage` / `getAll` and `Report.v1.getPage` / `getAll` accept the same qualifiers as
  their `getUuidsPage` and return documents, with unit tests passing.
- Local adapters resolve each page in a single view query (type/phone) or one query + one batched
  `allDocs` hydrate (freetext) — **no per-document fetch**.
- `GET /api/v1/contact` and `GET /api/v1/report` return `{ data, cursor }` pages of documents.
- `npm run unit-shared-lib` and `npm run unit-api` pass.

## Constraints

- Backwards-compatible: `getUuidsPage` / `getUuids` / `get` / `getWithLineage` unchanged.
- No `include_docs` flag on the ID-page — doc-page is a separate method/endpoint, matching the
  existing ID-vs-doc split.
- No N+1: type/phone arms are a single `include_docs` query; the freetext arm is one query + one
  batched `getDocsByIds`.
- Reuse the declarative qualifier table in the contact controller; do not duplicate
  mutual-exclusion rules.
- Pagination/cursor semantics identical to `getUuidsPage` (skip-based for views, opaque nouveau
  cursor for freetext).

## References

- `Person.v1.getPage` / `Place.v1.getPage` — pattern + local single-query implementation
- #10973 — `byPhone`/`byPhones` qualifiers, `queryDocIdsByKeys`, declarative controller table
- #10976 — `byShortcode`/`byExternalRef` (a downstream consumer if any of its callers need docs)

witash:
closes #11174 

Implements getPage/getAll for Contact.v1 and Report.v1

A few notes:

This PR does not attempt to do any refactoring to remove duplication, which makes the diff larger but otherwise too many unrelated changes are added in one branch. I think later we can go back and do some refactoring in a separate PR.
This replaces https://github.com/medic/cht-core/pull/11175 which did attempt some refactoring at the same time, but will be harder to review, even though its fewer LoC.

Qualifiers differ a little bit from the ticket description.
I did not add freetext qualifiers to these, but instead added `IdsQualifier` to both, and `ContactTypeQualifer` to contacts.
More qualifiers will be added later as needed.
I used `IdsQualifier` instead of `UuidsQualifier`, which is inconsistent with what is already there, but the preference is to call identifiers `ids` instead of `uuids`.

So the endpoints currently are not used by the code anywhere, and are generic given a list of ids, return all the documents. `all_docs` already does this, but we are going to add many more qualifiers, so these will be useful in later PRs.


api/src/controllers/contact.js
```diff
@@ -132,5 +138,70 @@ module.exports = {
...
+
+    /**
+     * @openapi
+     * /api/v1/contact:
+     *   get:
+     *     summary: Get contacts
+     *     operationId: v1ContactGet
+     *     description: >
+     *       Returns a paginated array of contact records (persons and places) matching the given filter criteria.
+     *       Exactly one of `ids` or `type` must be provided. Use the `cursor` returned in each response to retrieve
```
sugat009: issue (non-blocking): the summary says "Exactly one of `ids` or `type` must be provided", but the code allows both and gives `ids` precedence (`type` is silently ignored) - confirmed live that `?ids=a&type=person` returns the ids result. The precedence behavior is the right call (if you already have the ids, `type` is redundant), so this is just a doc fix. Suggest: "At least one of `ids` or `type` must be provided. If both are provided, `ids` takes precedence over `type`."

api/src/controllers/contact.js
```diff
@@ -132,5 +138,70 @@ module.exports = {
...
+     *       subsequent pages.
+     *     tags: [Contact]
+     *     x-since: 5.3.0
```
sugat009, line 157: question (non-blocking): is `5.3.0` the intended landing release? Just checking since the latest tag is `5.2.0` (same `x-since` is on `report.js`). No change needed if `5.3.0` is the target.

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -56,6 +56,40 @@ export const isUuidQualifier = (identifier: unknown): identifier is UuidQualifie
...
+    && qualifier.ids.every((id) => isString(id) && id.length > 0);
+};
+
+/**
+ * Builds a qualifier for finding entities by their UUIDs.
+ * @param ids the UUIDs to search with
+ * @returns the qualifier
+ * @throws InvalidArgumentError if the UUIDs are not valid
+ */
+export const byIds = (ids: [string, ...string[]]): IdsQualifier => {
```
sugat009: suggestion (non-blocking): consider deduping the ids here so the doc-page matches the `/uuid` path's behavior, e.g. `{ ids: [...new Set(ids)] }`. `byIds` is the common construction point for both the HTTP controller and direct JS callers, so one line covers both. Confirmed `?ids=a,a` currently returns the doc twice (the uuid endpoint dedupes via a Set). Optional, since callers normally pass unique ids and it's data-shape only - still one batched `allDocs`, no extra query.

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -56,6 +56,40 @@ export const isUuidQualifier = (identifier: unknown): identifier is UuidQualifie
   return isRecord(identifier) && hasField(identifier, { name: 'uuid', type: 'string' });
 };
 
+/**
+ * A qualifier that identifies entities by their UUIDs.
```
sugat009: nitpick (non-blocking): the doc says "by their UUIDs" (and `byIds`'s `@param` says "the UUIDs") but there's no UUID-format validation - any non-empty string is accepted - and the type/field use the generic `ids`. The wording could say "ids" to match the deliberate ids-over-uuids naming. Same wording in `report.ts` `getPage` `@param`.

sugat009 (changes requested):
Took a thorough pass on this: code, live testing against a local instance, and ran the unit + controller suites (all green). It's a clean, additive set of endpoints and the important parts all check out: cursor pagination, single batched query (no N+1), auth parity with the `/uuid` siblings, and backwards-compatibility. A few small non-blocking things inline, mainly the OpenAPI wording on the contact endpoint, plus a suggestion, a version check, and a doc nit.

api/src/controllers/contact.js
witash: changed the comment as suggested

api/src/controllers/contact.js
witash: yes 5.3.0 is the target

shared-libs/cht-datasource/src/qualifier.ts
witash: added deduplicating the ids

shared-libs/cht-datasource/src/qualifier.ts
witash: changed to ids

sugat009:
@witash there seems to be some conflicts with master branch

witash:
@sugat009 13c6d4c moved document summaries to datsource and so included a lot of changes to datasource that conflicted with this branch.
Especially both branches added IdsQualifier.
Resolved the conflicts and address comments, can you re-review please?

sugat009 (approved):
Re-reviewed at 10d8327855. All my earlier comments are addressed (OpenAPI wording, 5.3.0 target, ids naming), and the new endpoints look good: pagination is correct, auth matches the `/uuid` siblings, no cross-type leak, and the unit + controller suites pass.

One thing worth re-applying, non-blocking: the dedupe on `byIds` that was added last round got dropped during the master merge. The `IdsQualifier` conflict (both branches added it) was resolved in favor of master's `byIds`, which doesn't dedupe, so `?ids=a,a` returns the doc twice again. Suggest restoring it in `qualifier.ts`: `return { ids: [...new Set(ids)] };`.

Approving; feel free to fold that one line in before merge.

PR #11216, opened 2026-06-25
feat(#11215): remove registered patients view

### LINKED ISSUE #11215: Remove redundant map reduce view registered_patients
The `registered_patients` map reduce view is redundant. It is almost an exact copy of `reports_by_subject` — both index `data_record` reports by their subject's patient/place shortcode.

Maintaining a second, near-identical index over every report is wasteful. The view can be removed and its consumers repointed at `reports_by_subject`.

`reports_by_subject` is a strict superset of `registered_patients`: it emits the same `patient_id`/`place_id` keys (plus `case_id`, `patient_uuid`, and `place_uuid`) for the same `data_record` reports. Every consumer of `registered_patients` already post-filters the returned docs through `registrationUtils.isValidRegistration`, which discards error docs, non-registrations, and unconfigured forms — so the extra docs surfaced by `reports_by_subject` are trimmed back to exactly what `registered_patients` would have returned.

Two differences need to be handled when switching:
- **Key coercion:** `registered_patients` emits `String(id)` keys; `reports_by_subject` emits the raw value. To preserve matching for any numeric `patient_id`/`place_id`, `reports_by_subject` should coerce its emitted keys with `String()`.
- **Duplicate emits:** `reports_by_subject` emits a key from both the top-level property and `doc.fields`, so a single report can be returned more than once per key. Consumers must de-duplicate the returned docs by `_id` (`getReportsBySubject` in transitions already does this).

### Definition
- **Database:** medic-client (synced to offline clients)
- **Map keys:** `String(doc.patient_id)`, `String(doc.place_id)` (falling back to `doc.fields.patient_id` / `doc.fields.place_id`)
- **Map values:** (none)
- **Reduce:** None

### Usages
| 1 | [`shared-libs/transitions/src/lib/utils.js:165`](https://github.com/medic/cht-core/blob/master/shared-libs/transitions/src/lib/utils.js#L165) | `{ key/keys, include_docs: true }` | `getRegistrations` — fetches registration reports for one or more shortcodes, then filters to valid registrations. |
| 2 | [`api/src/services/export/message-mapper.js:122`](https://github.com/medic/cht-core/blob/master/api/src/services/export/message-mapper.js#L122) | `{ keys: patientIds, include_docs: true }` | Hydrates scheduled-task messages with their patient registrations for the messages export. |
| 3 | [`admin/src/js/services/message-queue.js:135`](https://github.com/medic/cht-core/blob/master/admin/src/js/services/message-queue.js#L135) | `{ keys: shortcodes, include_docs: true }` | Loads patient/place registrations for the admin message-queue view. |
| 4 | [`webapp/src/ts/services/format-data-record.service.ts:41`](https://github.com/medic/cht-core/blob/master/webapp/src/ts/services/format-data-record.service.ts#L41) | `{ key: shortcode, include_docs: true }` | Loads registrations for a report's patient/place when formatting scheduled-task messages for display. |

Note: removing the view (and the `String()` change to `reports_by_subject`) triggers a one-time view reindex on upgrade.

witash:
closes #11215 

note this involves view changes, so base is 10636-view-changes-2, not master


witash:
on 5.3 now, closing to change base back to master

PR #11230, opened 2026-07-01
feat(#9699): display nepali (bikram sambat) calendar in reports filter

### LINKED ISSUE #9699: Display Nepali (Bikram Sambat) calendar in reports filter
In the CHT platform, when the language is set to Nepali, the Bikram Sambat calendar is displayed in most areas. However, the date filter in the Reports tab still uses the Gregorian calendar, though displayed in the Nepali language.

Image

Use the Nepali (Bikram Sambat) calendar in the Reports filter.
Integrating the Nepali (Bikram Sambat) calendar into the Reports tab would make it more user-friendly and consistent, allowing users to filter data more easily. 

**Additional context**
Copy-pasted from a forum post: https://forum.communityhealthtoolkit.org/t/display-nepali-bikram-sambat-calendar-in-reports-filter/4451

cc @bhishankc , thanks for reporting!

megha1807:
# Description

This PR implements the Nepali (Bikram Sambat) calendar filter in the Reports tab sidebar. It extracts the core datepicker calendar logic into a reusable module and integrates it with both the Enketo widget and the Angular sidebar filters.

### Key Changes
1. **Extracted Shared Utility**: Created a stateless, reusable helper in `bikram-sambat-picker-shared.js` containing the initialization, sanitization (CSP fix), esc key listener, and backdrop handlers.
2. **Refactored Enketo Widget**: Updated `bikram-sambat-datepicker.js` to invoke the shared helper.
3. **Reports Date Filter Integration**: Modified `date-filter.component.ts` to conditionally render the Bikram Sambat picker if the user is using the Devanagari script. Selected dates are translated back to Gregorian formats and formatted using `FormatDateService`.
4. **Tests**:
   - Added unit test cases for the component in `date-filter.component.spec.ts` and mocked dependencies in `reports-sidebar-filter.component.spec.ts`.
   - Added E2E WebdriverIO helpers to `reports.wdio.page.js` and wrote integration tests in `sidebar-filter.wdio-spec.js`.

# Issue Number
Fixes #9699

# Code review checklist


webapp/src/ts/components/filters/date-filter/date-filter.component.ts
```diff
@@ -108,6 +120,42 @@ export class DateFilterComponent implements OnInit, OnDestroy, AfterViewInit {
...
 
+  private initNepaliDatePicker() {
+    const $hiddenDateInput = $('<input type="text" class="nepali-datepicker-input" ' +
+      'style="position: absolute; opacity: 0; width: 0; height: 0; padding: 0; border: 0;" />');
+    $(`#${this.fieldId}`).parent().append($hiddenDateInput);
+
+    setupNepaliDatePicker($hiddenDateInput, {
+      onDateSelect: (data: any) => {
+        const gregDateStr = toGreg_text(data.bsYear, data.bsMonth, data.bsDate);
+        const gregMoment = moment(gregDateStr);
```
sugat009, line 135: issue (blocking): The Nepali "to" date filter stores midnight, so same-day reports are silently dropped.

Reproduced in-app: a report filed today at 12:00 shows under an English "to = today" filter but disappears under the equivalent Nepali one. `onDateSelect` does `moment(toGreg_text(...))`; `toGreg_text` returns a date-only `YYYY-MM-DD`, which moment parses as local midnight (00:00:00). For the "to" instance that midnight becomes the inclusive `endkey` in `generate-search-requests.js`, so anything filed after 00:00 on the selected day is excluded. The Gregorian daterangepicker path this mirrors normalizes the "to" bound to end-of-day (`daterangepicker.js:506`), and nothing downstream re-normalizes it. Confirmed it is not a timezone artifact (both calendars use the same offset and agree on the day). Could we push the "to" bound to `endOf('day')` before storing, e.g. `this.isStartDate ? gregMoment.startOf('day') : gregMoment.endOf('day')`?

webapp/src/ts/components/filters/date-filter/date-filter.component.ts
```diff
@@ -108,6 +120,42 @@ export class DateFilterComponent implements OnInit, OnDestroy, AfterViewInit {
...
+      if (this.disabled) {
+        return;
+      }
+
+      const activeDate = this.isStartDate ? this.dateRange.from : this.dateRange.to;
+      if (activeDate) {
+        const bsDate = toBik(moment(activeDate));
+        const bsMonth = String(bsDate.month).padStart(2, '0');
+        const bsDay = String(bsDate.day).padStart(2, '0');
+        $hiddenDateInput.val(`${bsDate.year}-${bsMonth}-${bsDay}`);
```
sugat009: issue (blocking): After a date is selected, reopening the filter throws `Invalid nepali number` and the calendar never reopens.

Reproduced in-app: pick a date, click the field again, nothing opens, and the console repeats `Error: Invalid nepali number` (`getNumberByNepaliNumber` -> `parseFormattedBsDate` -> `renderFormattedSpecificDateCalendar`). Root cause: this seeds the hidden input with Latin digits (`${bsDate.year}-${bsMonth}-${bsDay}` -> `"2083-03-23"`), but the nepali-date-picker library parses the value as Devanagari (`getNumberByNepaliNumber` looks each character up in `calendarData.nepaliNumbers` and throws on anything outside `०-९`). The enketo widget avoids this by seeding `toDevanagari(...)` (see `bikram-sambat-datepicker.js`). First open works only because there is no active date yet (an empty value skips the parse). Net effect: the Nepali reports date filter is one-shot, once a date is picked the user cannot reopen it to change it (and there is no clear button here). Could we convert the seeded value to Devanagari before `val(...)`, matching the widget (`eurodigit` `toDevanagari`, or the bikram-sambat Devanagari helper)?

webapp/src/ts/components/filters/date-filter/date-filter.component.ts
```diff
@@ -183,11 +247,25 @@ export class DateFilterComponent implements OnInit, OnDestroy, AfterViewInit {
...
   }
+
+  private cleanupNepaliDatePicker() {
+    const $hiddenDateInput = $(`#${this.fieldId}`).parent().find('.nepali-datepicker-input');
+    if ($hiddenDateInput.length) {
+      hideDatePicker($hiddenDateInput);
+      $hiddenDateInput.remove();
+    }
+    $('.nepali-date-picker-overlay').remove();
+    $('.nepali-date-picker').remove();
```
sugat009: issue (non-blocking): `cleanupNepaliDatePicker` removes every picker globally, not just this component's.

Lines 268 to 269 do `$('.nepali-date-picker-overlay').remove()` and `$('.nepali-date-picker').remove()`, global selectors that match both the from and to library divs. Today the two filters mount and unmount together so it is harmless, but it is fragile: if their lifecycles ever diverge (an `*ngIf`, or reuse alongside another Nepali picker) the first destroy would delete the still-alive sibling's calendar div and that field would silently stop opening. Could we scope removal to this component's own picker div, and only drop the shared overlay when no other picker is live?

Note: this is the same global / first-match selector root cause behind the eraser-only-on-first-field bug in the sibling PR #11234. Scoping the shared picker logic per `$hiddenInput` would address both, so it may be worth fixing in the shared module once.

webapp/src/ts/components/filters/date-filter/date-filter.component.ts
```diff
@@ -183,11 +247,25 @@ export class DateFilterComponent implements OnInit, OnDestroy, AfterViewInit {
   ngOnDestroy() {
     this.setError(false);
     this.subscription.unsubscribe();
-    const datePicker:any = $(`#${this.fieldId}`).data('daterangepicker');
 
+    if (this.languageService.useDevanagariScript()) {
```
sugat009: question (non-blocking): Picker mode is decided once at init but re-read at label and destroy time.

`useDevanagariScript()` is read independently in `ngAfterViewInit` (which picker to build), `setLabel`, and here in `ngOnDestroy`. Since a language switch does not recreate the component, if a user changes the UI language while this filter is mounted the picker goes stale and `ngOnDestroy` can take the wrong teardown branch, leaking the picker DOM. No impact on non-Nepali deployments. Is a mid-session language switch expected here? If so, caching an `isNepali` boolean chosen at init and reusing it in `setLabel`/`ngOnDestroy` would keep init, label, and cleanup consistent.

webapp/src/ts/components/filters/date-filter/date-filter.component.ts
```diff
@@ -108,6 +120,42 @@ export class DateFilterComponent implements OnInit, OnDestroy, AfterViewInit {
...
+      if (activeDate) {
+        const bsDate = toBik(moment(activeDate));
+        const bsMonth = String(bsDate.month).padStart(2, '0');
+        const bsDay = String(bsDate.day).padStart(2, '0');
+        $hiddenDateInput.val(`${bsDate.year}-${bsMonth}-${bsDay}`);
+      } else {
+        $hiddenDateInput.val('');
+      }
+
+      $hiddenDateInput.click();
```
sugat009, line 172: suggestion (non-blocking): The Nepali picker has no keyboard path and no dialog semantics.

The calendar opens by forwarding a click to a hidden input, but the library's day cells are mouse-only (no tabindex, no key handlers), and the popup is not exposed as `role="dialog"`/`aria-modal`, with no focus move-in or focus return. So a keyboard or screen-reader user on the Reports filter can open the picker but cannot select a date (only Esc dismisses). This is inherited from the enketo widget but newly surfaced on a core webapp filter where a11y expectations are higher. At minimum worth documenting the limitation; ideally move focus into the popup and add dialog semantics. (Same gap noted on #11234.)

webapp/src/ts/components/filters/date-filter/date-filter.component.ts
```diff
@@ -108,6 +120,42 @@ export class DateFilterComponent implements OnInit, OnDestroy, AfterViewInit {
     });
   }
 
+  private initNepaliDatePicker() {
+    const $hiddenDateInput = $('<input type="text" class="nepali-datepicker-input" ' +
+      'style="position: absolute; opacity: 0; width: 0; height: 0; padding: 0; border: 0;" />');
+    $(`#${this.fieldId}`).parent().append($hiddenDateInput);
+
+    setupNepaliDatePicker($hiddenDateInput, {
```
sugat009, line 132: question (non-blocking): The sidebar filter opens a viewport-centered modal with a full-app dark overlay.

The Nepali picker is `position:fixed` centered with a full-viewport `rgba(0,0,0,0.3)` overlay (z-index 9999/9998) that dims the whole app including the sidebar the user just clicked, whereas the Gregorian daterangepicker anchors near the trigger. That centered-modal look was designed for the inline enketo form widget. Is the centered modal plus full-app dim the intended UX for a sidebar filter, or should it anchor near the field like the Gregorian one?

webapp/src/ts/components/filters/date-filter/date-filter.component.ts
```diff
@@ -14,6 +14,11 @@ import { Selectors } from '@mm-selectors/index';
 import { BsDropdownModule } from 'ngx-bootstrap/dropdown';
 import { TranslatePipe } from '@ngx-translate/core';
 
+import { LanguageService } from '@mm-services/language.service';
+import { FormatDateService } from '@mm-services/format-date.service';
+import { toBik, toGreg_text } from 'bikram-sambat';
+import { setupNepaliDatePicker, hideDatePicker } from '../../../../js/enketo/widgets/bikram-sambat-picker-shared';
```
sugat009, line 20: suggestion (if-minor): The Angular component reaches into the enketo widget JS layer via a deep relative import.

`import { ... } from '../../../../js/enketo/widgets/bikram-sambat-picker-shared'` couples the TS filter layer to the enketo widget JS (a CommonJS module) across four directory levels. It works, but since the shared picker logic is now used by both the widget and this filter, it might be cleaner to promote it to a proper shared location with a path alias rather than a deep relative reach.

webapp/tests/karma/ts/components/filters/date-filter.component.spec.ts
```diff
@@ -138,4 +153,82 @@ describe('Date Filter Component', () => {
...
+
+      const hiddenInput = $('#bikram-sambat-test-wrapper .nepali-datepicker-input');
+      const clickSpy = sinon.spy(hiddenInput[0], 'click');
+
+      $(`#test-field-id`).click();
+
+      expect(clickSpy.callCount).to.equal(1);
+      expect(hiddenInput.val()).to.equal('2081-04-09'); // 2024-07-24 is 2081-04-09 in Bikram Sambat
+
+      clickSpy.restore();
```
sugat009, line 204: issue (non-blocking): The Nepali picker tests only cover the "from" path, which is why the "to" midnight bug was not caught.

`onDateSelect`/`setLabel` are exercised only for `isStartDate=true`, `cleanupNepaliDatePicker` has no coverage, and the actual BS / Devanagari label is never asserted (`dayMonth` is stubbed). A test that selects a "to" date and checks the stored bound is end-of-day would have caught the blocking bug above, and one that reopens the picker after selecting would catch the reopen crash. Could we add to-picker, reopen, and cleanup coverage?

webapp/src/ts/components/filters/date-filter/date-filter.component.ts
```diff
@@ -108,6 +120,42 @@ export class DateFilterComponent implements OnInit, OnDestroy, AfterViewInit {
     });
   }
 
+  private initNepaliDatePicker() {
+    const $hiddenDateInput = $('<input type="text" class="nepali-datepicker-input" ' +
```
sugat009: nitpick: The inline style on the hidden input duplicates the `.nepali-datepicker-input` class.

The `style="position:absolute;opacity:0;width:0;height:0;padding:0;border:0"` could live in the `.nepali-datepicker-input` CSS rule (or a shared visually-hidden class) instead of being inlined in the element markup.

sugat009 (changes requested):
Thanks @megha1807. Clean extraction, and the non-Nepali path is safely gated (defaults off, so no risk to existing deployments). I pulled the branch and tested the Nepali flow on the Reports tab in a Nepali locale.

Two blocking issues came up in testing, both reproduced live and detailed inline: the "to" bound is stored at midnight (so same-day reports are dropped), and reopening a filter after a date is selected throws and the calendar never reappears. The rest are non-blocking notes.

Working well: the picker renders for both from and to fields, selecting a date applies the filter, and non-Nepali locales are untouched.

webapp/src/ts/components/filters/date-filter/date-filter.component.ts
megha1807: I have resolved this by caching the value in a class property (isNepali = false) inside ngOnInit.

We now reference this.isNepali in ngAfterViewInit, setLabel, and ngOnDestroy to keep the initialization, label rendering, and destruction branch consistently aligned throughout the component's lifetime.

webapp/src/js/enketo/widgets/bikram-sambat-picker-shared.js
```diff
@@ -0,0 +1,207 @@
...
+    sanitizeHrefs(container);
+    ensureClearButton(container, onClear);
+  });
+  observer.observe(container, { childList: true, subtree: true });
+  $hiddenInput.data('observer', observer);
+
+  $(container).off('keydown.focusTrap').on('keydown.focusTrap', (e) => handleFocusTrap(e, container));
+};
+
+const handleDateSelect = (event, onDateSelect) => {
```
sugat009: issue (blocking): Selecting a date leaves the full-screen overlay stuck over the whole app.

The rework dropped the `hideDatePicker($hiddenInput)` call that used to run in the `dateSelect` handler. The reports filter uses the default `closeOnDateSelect: true`, and the nepali-date-picker library's auto-close does a raw `.hide()` on the picker div and fires no `close` event, so `handleClose`/`hideDatePicker` never run. Result: after picking a date the `.nepali-date-picker-overlay` (z-index 9998, full viewport) stays visible and click-blocking, and the MutationObserver, `keydown.nepaliDatePicker`, and focus-trap listeners leak. Verified in-app: after selecting a from or to date the whole screen stays dimmed until you click to dismiss. Could `handleDateSelect` call `hideDatePicker($hiddenInput)` when `closeOnDateSelect` is true (plumb the flag through), so the auto-close path cleans up like the pre-rework code did? Note the shared module is byte-identical in #11234, but #11234 passes `closeOnDateSelect: false` so it does not hit this; the fix belongs here in the shared module and benefits both.

webapp/src/ts/components/filters/date-filter/date-filter.component.ts
```diff
@@ -108,6 +122,44 @@ export class DateFilterComponent implements OnInit, OnDestroy, AfterViewInit {
...
 
+  private initNepaliDatePicker() {
+    const $hiddenDateInput = $('<input type="text" class="nepali-datepicker-input" />');
+    $(`#${this.fieldId}`).parent().append($hiddenDateInput);
+
+    setupNepaliDatePicker($hiddenDateInput, {
+      onDateSelect: (data: any) => {
+        const gregDateStr = toGreg_text(data.bsYear, data.bsMonth, data.bsDate);
+        const gregMoment = moment(gregDateStr);
+        const normalizedMoment = this.isStartDate ? gregMoment.startOf('day') : gregMoment.endOf('day');
```
sugat009, line 136: issue (blocking): The To filter chip shows the day after the one selected.

Selecting २२ असार displays सम्म = २३ असार. The applied filter value is correct (`endOf('day')` = July 6 23:59:59.999), but `setLabel` formats that end-of-day moment via `formatDateService.dayMonth` (line 223), and `dayMonth` of a 23:59:59.999 timestamp resolves to the NEXT Bikram Sambat day. Probe: `dayMonth('2026-07-06 23:59:59.999')` = २३ असार vs `dayMonth('2026-07-06 00:00')` = २२ असार. So a user who picks 22 sees the filter labeled 23. Could the label be derived from the selected date (start of day, or the raw BS parts) rather than the `endOf('day')` value?

webapp/src/ts/components/filters/date-filter/date-filter.component.ts
```diff
@@ -108,6 +122,44 @@ export class DateFilterComponent implements OnInit, OnDestroy, AfterViewInit {
     });
   }
 
+  private initNepaliDatePicker() {
+    const $hiddenDateInput = $('<input type="text" class="nepali-datepicker-input" />');
+    $(`#${this.fieldId}`).parent().append($hiddenDateInput);
+
+    setupNepaliDatePicker($hiddenDateInput, {
+      onDateSelect: (data: any) => {
```
sugat009, line 133: issue: The injected Clear button is dead and unstyled in the reports filter.

The shared module's `ensureClearButton` always injects a `.clear-btn` when the library renders a `.today-btn`, but this call passes only `{ onDateSelect }`, so `onClear` is undefined and the button's handler is a no-op; and #11230 ships no `.clear-btn` CSS, so it renders as a blank clickable gap next to the month controls (no eraser icon). @binokaryg asked for a working Clear control, especially for optional date fields, and the reports dates are optional. Could we either wire `onClear: () => this.clear()` and add the `.clear-btn` styling, or suppress the injected button when no `onClear` is provided so there is no dead control? (On #11234 this same slot is the styled, working eraser.)

sugat009 (changes requested):
Re-review after the fixes. Both original blockers are confirmed fixed (verified in-app): the To bound now applies end-of-day (`23:59:59.999`) so same-day reports are no longer dropped, and reopening a picker no longer throws `Invalid nepali number`. Thanks for those.

The rework did introduce a few new issues, so still requesting changes. Details inline. One design-level ask that isn't tied to a single line: the Nepali picker opens as a viewport-centered modal with a full-app dark overlay, which is inconsistent with the English (daterangepicker) filter that anchors a dropdown to the field. Could we make the Nepali picker consistent with the English one (anchored to the field, no full-app overlay)? That would also naturally resolve the stuck-overlay bug flagged inline.

sugat009:
@megha1807 heads-up: now that #11234 is merged into the feature branch, this PR has a merge conflict, isolated to one file: `webapp/src/js/enketo/widgets/bikram-sambat-datepicker.js`.

Cause: both this PR and #11234 refactored that widget, and #11234's version (with the keep-open and clear refinements) is now on the feature branch, so this PR's changes to that file are redundant.

To resolve: merge the latest feature branch into your branch and take the feature branch's version of `bikram-sambat-datepicker.js` (drop this PR's changes to it). The shared module and `bikramsambat.less` merge cleanly (identical to #11234's), so nothing to do there. After that, this PR should reduce to just the reports-filter work (`date-filter.component.ts` plus the reports tests), which is what I'll do the final review pass on.

Thanks!

megha1807:
Hi @sugat009,
I have merged the latest target branch (10707-dmp-2026-enhance-bikram-sambat-support-in-cht) into this branch and resolved the conflict in bikram-sambat-datepicker.js by taking the version from the target branch.

The PR is now clean and reduces purely to the reports filter changes. Ready for your final review pass!
Thanks!

webapp/src/ts/components/filters/date-filter/date-filter.component.ts
```diff
@@ -108,6 +122,45 @@ export class DateFilterComponent implements OnInit, OnDestroy, AfterViewInit {
...
+      e.preventDefault();
+      e.stopPropagation();
+
+      if (this.disabled) {
+        return;
+      }
+
+      const activeDate = this.isStartDate ? this.dateRange.from : this.dateRange.to;
+      if (activeDate) {
+        const bsDate = toBik(moment(activeDate));
```
sugat009: issue (blocking): Reopening the "to" filter pre-highlights the day after the selected one.

The label fix normalized `to` with `.startOf('day')`, but this reopen-seed path was missed: `activeDate` is `dateRange.to`, stored by `onDateSelect` as `gregMoment.endOf('day')` (23:59:59.999 local), and `toBik(moment(activeDate))` on an end-of-day time tips forward one Bikram Sambat day. Confirmed live in-app (Nepali locale, Nepal timezone): I set the reports "to" date to २२; the सम्म chip correctly reads २२ असार and the applied filter is 2026-07-06, but reopening the picker seeds the hidden input `२०८३-०३-२३` and the calendar highlights २३ (day+1). If the user re-confirms the highlighted day, the filter silently shifts forward one BS day. The "from" filter (startOf) is unaffected. Fix: normalize before `toBik`, matching the `setLabel` fix, e.g. `const bsDate = toBik(moment(activeDate).startOf('day'));`

webapp/src/ts/components/filters/date-filter/date-filter.component.ts
```diff
@@ -108,6 +122,45 @@ export class DateFilterComponent implements OnInit, OnDestroy, AfterViewInit {
...
+
+    setupNepaliDatePicker($hiddenDateInput, {
+      onDateSelect: (data: any) => {
+        const gregDateStr = toGreg_text(data.bsYear, data.bsMonth, data.bsDate);
+        const gregMoment = moment(gregDateStr);
+        const normalizedMoment = this.isStartDate ? gregMoment.startOf('day') : gregMoment.endOf('day');
+        const dateRange = this.createDateRange(normalizedMoment, normalizedMoment);
+        this.applyFilter(dateRange);
+      },
+      position: 'anchored'
```
sugat009: question (non-blocking): `position: 'anchored'` removes the full-app overlay (good) but does not actually anchor the picker to the field.

The `.nepali-date-picker.anchored` rule (in the shared `bikramsambat.less`) overrides only `position` (to absolute) and `transform` (to none); it does not reset the base rule's `top: 50% !important` / `left: 50% !important`, so the plugin's field-offset inline positioning loses and the picker sits at viewport center. Confirmed live: the picker renders at `top/left: 50%` (around x=720 on a 1440px viewport) while the "to" field is in the right sidebar (~x=1120), so it floats ~400px from the trigger rather than beside it; the overlay is correctly gone. So the "consistent with the English daterangepicker" goal is only half-met. The CSS lives on the feature branch (from #11234), but this PR is what enables anchored mode. To truly anchor, `.anchored` needs to reset `top/left` (letting the plugin's inline offset win) or position from the field via JS. Could you confirm the placement and decide whether to fix the CSS or hold anchored mode until it anchors?

sugat009 (changes requested):
Re-reviewed after the reconcile, thanks @megha1807. Everything I flagged before is now fixed and verified (code + in-app): the overlay no longer sticks (anchored skips the overlay and `hideDatePicker` runs on auto-close), the label off-by-one is fixed (`startOf('day')` before `dayMonth`), the dead clear button is suppressed, the reopen crash is gone (Devanagari seed), cleanup is scoped, and `isNepali` is cached.

One new blocking issue slipped onto a call site the label fix didn't cover, plus a non-blocking positioning note. Inline.

Test-coverage note (non-blocking): the `setLabel` Nepali test (`date-filter.component.spec.ts:269`) uses a midnight `from` and stubs `dayMonth` to a constant, so it never exercises the `endOf('day')` `to` branch the label fix targets and could not catch a BS-day off-by-one; deleting the `.startOf('day')` calls keeps the suite green. A guard with `isStartDate=false`, `to = moment(...).endOf('day')`, asserting `dayMonth` is called with the `startOf`-day ms, would lock it in.

webapp/src/ts/components/filters/date-filter/date-filter.component.ts
megha1807: Great catch! I've updated the shared bikramsambat.less stylesheet so that the fixed viewport-centering styles (top: 50%, left: 50%, etc.) are only applied when the picker is not in anchored mode.

This allows the library's dynamic inline offsets to win, so the picker now anchors correctly right next to the trigger field in the sidebar.

sugat009: **issue (blocking):** Confirmed live that the calendar no longer appears when opening either reports date filter in Nepali. Root cause: the nepali-date-picker plugin positions the calendar relative to its bound input, which is the hidden `.nepali-datepicker-input` at `left:-9999px`. While `.anchored` forced `left:50%` this was masked; now the plugin's own `left:-8759px` wins and the picker lands off the left edge:

anchored picker inline style shows left:-8759px

Removing that inline `left` makes the fully-functional calendar reappear at the left edge, so the only defect is positional.

To fix, the picker needs to anchor to the **visible** trigger rather than the hidden input, e.g. on `show`, set the container's `top/left` from the `#{fieldId}` rect. Alternatively, if field-anchoring keeps being fiddly against the plugin, reverting to the visible centered placement is acceptable, the hard requirement is just that the calendar is visible and usable.

sugat009 (changes requested):
Thanks for the quick turnaround. The reopen off-by-one is fixed (confirmed live: selecting २२ now reopens highlighting २२, and the label matches), and the `setLabel` normalization test is a good addition. One blocking item remains, though: the anchored positioning change now renders the reports picker off-screen. Details inline.

webapp/src/ts/components/filters/date-filter/date-filter.component.ts
megha1807: Good catch on the hidden input's coordinates! I've updated the CSS for .nepali-datepicker-input in the shared stylesheet to align it at left: 0; top: 0 inside the filter dropdown wrapper instead of pushing it to left: -9999px.

This resolves the off-screen positioning offset, allowing the plugin's automatic anchoring calculations to place the calendar popup correctly next to the visible trigger button in the sidebar.

webapp/src/ts/components/filters/date-filter/date-filter.component.ts
```diff
@@ -108,6 +122,45 @@ export class DateFilterComponent implements OnInit, OnDestroy, AfterViewInit {
     });
   }
 
+  private initNepaliDatePicker() {
+    const $hiddenDateInput = $('<input type="text" class="nepali-datepicker-input" />');
+    $(`#${this.fieldId}`).parent().append($hiddenDateInput);
+
+    setupNepaliDatePicker($hiddenDateInput, {
```
sugat009, line 132: **issue (non-blocking):** The Gregorian path caps selection at today (`maxDate: moment()`, line 81), but this call passes no `maxDate`, so future BS dates are selectable and a future *from* date returns zero reports. Threading `maxDate` through `setupNepaliDatePicker` would restore parity with the Gregorian picker.

webapp/tests/karma/ts/components/filters/date-filter.component.spec.ts
```diff
@@ -138,4 +153,138 @@ describe('Date Filter Component', () => {
...
+      component.ngAfterViewInit();
+
+      expect(dateRangePicker.callCount).to.equal(0);
+      expect(nepaliDatePickerStub.callCount).to.equal(1);
+
+      const hiddenInput = $('#bikram-sambat-test-wrapper .nepali-datepicker-input');
+      expect(hiddenInput).to.have.lengthOf(1);
+    });
+
+    it('clicking filter input should trigger click on hidden input with pre-populated BS date', () => {
```
sugat009, line 191: **suggestion (non-blocking):** This reopen test only covers a start date at local midnight, where the `startOf('day')` guard at `date-filter.component.ts:150` is a no-op, so deleting that guard would keep the suite green. A to-date reopen case (select a to-date, reopen, assert the highlighted day matches the label) would lock in the off-by-one this round fixed.

sugat009 (commented):
Re-reviewed at the current head (`5e5f29127`) with a live run in Nepali locale. The off-screen picker I flagged previously (against `5448f6f7c`) is fixed here, confirmed live: the picker now anchors on-screen below the field (desktop and mobile).

**praise:** All previously-flagged blockers are resolved, reopen highlights the selected day with no off-by-one, label matches, to-date applies end-of-day, no stuck overlay. The overlay-less reports dropdown is the right call too (it matches the Gregorian filter and closes fine on outside-click / Escape). Nice work.

No blocking issues. A few non-blocking items below, all reproduced locally.

webapp/src/js/enketo/widgets/bikram-sambat-picker-shared.js
```diff
@@ -173,12 +173,32 @@ const ensureOverlay = (position) => {
...
+  }
+
+  const $trigger = $hiddenInput.siblings('a, button, input').filter(':visible').first();
+  if (!$trigger.length) {
+    return;
   }
+
+  const offset = $trigger.offset();
+  const height = $trigger.outerHeight();
+  $picker.css({
```
sugat009: **issue (non-blocking):** The anchored picker is positioned once from `trigger.offset()` and appended to `<body>`, with no viewport clamp and no scroll tracking. Unlike the Gregorian daterangepicker in this same component (which flips via `isInViewport`, `date-filter.component.ts:104-107`), it spills below the fold when the trigger sits low in a scrolled or expanded sidebar:

anchored calendar spills below the fold, lower rows cut off

It also strands when the sidebar scrolls, since the picker is body-appended with a one-time absolute `top`/`left` and never repositions. Both reproduced locally. Could the anchored dropdown reuse or mirror the Gregorian picker's flip and reposition-on-scroll handling?

webapp/src/ts/components/filters/date-filter/date-filter.component.ts
```diff
@@ -108,6 +122,49 @@ export class DateFilterComponent implements OnInit, OnDestroy, AfterViewInit {
     });
   }
 
+  private initNepaliDatePicker() {
+    const $hiddenDateInput = $('<input type="text" class="nepali-datepicker-input" />');
+    $(`#${this.fieldId}`).parent().append($hiddenDateInput);
+
+    const todayBs = toBik(moment());
+    const maxDate = `${todayBs.year}-${String(todayBs.month).padStart(2, '0')}-${String(todayBs.day).padStart(2, '0')}`;
```
sugat009: **issue (blocking): Nepali reports date filter is completely non-functional**

In Nepali locale, opening Reports throws `Invalid nepali number` and the filter never works: the calendar renders empty and clicking does nothing (both from and to). Confirmed live:

console shows Invalid nepali number on opening reports in Nepali

reports date filter open in Nepali, calendar renders empty

The `maxDate` value handed to the picker is in a form it can't parse. Two related problems in the same area: a future date can be selected at certain times of day, and in some Bikram Sambat months today can't be selected at all.

webapp/src/ts/components/filters/date-filter/date-filter.component.ts
```diff
@@ -108,6 +122,49 @@ export class DateFilterComponent implements OnInit, OnDestroy, AfterViewInit {
...
+        const gregMoment = moment(gregDateStr);
+        const normalizedMoment = this.isStartDate ? gregMoment.startOf('day') : gregMoment.endOf('day');
+        const dateRange = this.createDateRange(normalizedMoment, normalizedMoment);
+        this.applyFilter(dateRange);
+      },
+      position: 'anchored',
+      maxDate
+    });
+
+    $(`#${this.fieldId}`).on('click', (e) => {
```
sugat009, line 144: **issue (non-blocking): from and to pickers close each other instead of switching**

With both filters present, opening one while the other is open just closes the first (so it takes two clicks), and clicking outside closes both.

webapp/src/js/enketo/widgets/bikram-sambat-picker-shared.js
```diff
@@ -173,12 +181,57 @@ const ensureOverlay = (position) => {
...
+
+  const $trigger = $hiddenInput.siblings('a, button, input').filter(':visible').first();
+  if (!$trigger.length) {
+    return;
+  }
+
+  handleReposition($picker, $trigger);
+
+  const scrollCallback = () => handleReposition($picker, $trigger);
+  window.addEventListener('scroll', scrollCallback, true);
```
sugat009, line 240: **issue (non-blocking): picker leaks listeners/observers across open-dismiss cycles**

Opening and dismissing the picker repeatedly leaves event listeners and observers attached; they accumulate over a session because cleanup only runs on some close paths, not when the picker is dismissed the usual way (clicking outside, or on the field).

tests/page-objects/default/reports/reports.wdio.page.js
```diff
@@ -299,6 +299,42 @@ const setSidebarFilterToDate = () => {
...
+  }, { timeout: 5000, timeoutMsg: 'Nepali date picker not displayed' });
+
+  const prevBtn = picker.$('.prev-btn.icon');
+  await prevBtn.waitForDisplayed();
+  for (let i = 0; i < prevClicks; i++) {
+    await prevBtn.click();
+  }
+
+  const cells = await picker.$$('table tbody td.current-month-date:not(.disable)');
+  if (cells.length > cellIndex) {
```
sugat009, line 325: **issue (non-blocking): tests don't catch this blocking bug**

This shipped with green CI. The unit test stubs the picker and never checks the `maxDate` value that's passed, and the e2e to-date selection here can pass without actually selecting a to-date, so that path isn't really exercised.

sugat009 (changes requested):
Re-reviewed at `a54cdbf2b`, live in Nepali. The reports date filter is broken in Nepali, with a few related issues in the same area (below). Could you verify the fix end-to-end in Nepali and attach a short screen recording of the working calendar? The unit tests stub the picker, so they don't exercise this path. (The to-date reopen test and the scroll repositioning from last round are good, thanks.)

megha1807:
Hi @sugat009, 

I have verified the fixes end-to-end in the Nepali locale:
- The Bikram Sambat calendar opens correctly without any "Invalid nepali number" console errors.
- Single-click calendar switching between From (बाट) and To (सम्म) date fields is now working.

Here is the screen recording of the working calendar:

https://github.com/user-attachments/assets/1a62e83b-f7be-47fc-af74-9110be0370b8

sugat009 (approved):
Approving to merge into the epic branch. Earlier findings addressed; the divergence crash is tracked in #11246 / #9250 to be resolved before master.

PR #11223, opened 2026-06-29
feat(#10706): add bulk operation status endpoint

vikrantwiz02:
# Description

Adds `GET /api/v1/bulk-operations/{id}` so a caller can poll the progress of a bulk operation (delete, move or merge) by its id. It returns the operation's log document from the `medic-logs` database, with the per-action status and the count of changes so far.

The service only resolves ids that carry the `bulk-operation:` prefix, so the endpoint can't be used to read the other log documents that share `medic-logs`. Online users only; a missing or non bulk-operation id returns 404. Documented with an OpenAPI annotation.

Builds on #11222 (the `@medic/bulk-operations` doc model). Until that one merges into the feature branch, the diff here also includes the doc model; it'll narrow to just the endpoint once #11222 lands.

Part of #10706

# Code review checklist
- [x] Readable: Concise, well named, follows the [style guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/)
- [x] Documented: OpenAPI annotation for the new endpoint
- [x] Tested: Unit and/or e2e where appropriate
- [x] Backwards compatible: Works with existing data and configuration or includes a migration. Any breaking changes documented in the release notes.
- [ ] AI disclosure: Please disclose use of AI per [the guidelines](https://docs.communityhealthtoolkit.org/community/contributing/ai-guidelines/).

# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

PR #11222, opened 2026-06-29
feat(#10706): add bulk operation document model

vikrantwiz02:
# Description

First piece of the bulk operation framework from the #10706 design: the shared document model that delete, move and merge all build on.

Adds the `@medic/bulk-operations` shared library with:
- the action types (`archive`, `set-contact`, `delete-user`), statuses, and document id prefixes
- `buildBulkOperation(actionOperations, date)`, which builds the log document (`medic-logs`, read by the polling endpoint) and one action document per action type (`medic-sentinel`, processed by Sentinel in batches), cross-linked both ways
- the per-item params stored in a base64 json `operations` attachment, so advancing an action's cursor later does not rewrite the whole list

No dependency on the archiving work, so it can land on its own. The queue helper that writes these docs and the `GET /api/v1/bulk-operations/{id}` polling endpoint follow next.

Part of #10706

# Code review checklist
- [x] Readable: Concise, well named, follows the [style guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/)
- [x] Tested: Unit and/or e2e where appropriate
- [x] Backwards compatible: Works with existing data and configuration or includes a migration. Any breaking changes documented in the release notes.
- [ ] AI disclosure: Please disclose use of AI per [the guidelines](https://docs.communityhealthtoolkit.org/community/contributing/ai-guidelines/).

# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

PR #11236, opened 2026-07-04
feat(#10706): add bulk operation framework and contact hierarchy delete

vikrantwiz02:
# Description

The server-side implementation of contact-hierarchy delete for #10706. A `DELETE` on the person and place endpoints gathers everything the delete touches and queues a bulk operation that Sentinel executes asynchronously, so the slow work runs off the request. The bulk operation framework is generic: delete is its first consumer, and move and merge will reuse it.

What's here, by layer:
- **Document model + constants**: the log doc and per-action doc shapes, with the id prefixes and enums shared through `@medic/constants`.
- **Queue helper**: writes the log doc to `medic-logs` and one action doc per type to `medic-sentinel`.
- **`GET /api/v1/bulk-operations/{id}`**: poll an operation's status.
- **Sentinel**: a `medic-sentinel` changes-feed listener plus a batch/cursor processor, with the `set-contact`, `delete-user` and `archive` handlers.
- **`DELETE /api/v1/person/{id}` and `/api/v1/place/{id}`**: gated on `can_delete_contact_hierarchy`, with `delete_users` (requires `can_delete_users`, else a `400` when linked users exist) and `dry_run` query params; each endpoint rejects an id of the wrong type with a `404`; returns a summary of the changes plus the operation id.

The `archive` handler reuses the archiving work from #6615 / #11139 (now in the base branch): it calls `archiveBatch` directly to copy the contacts and reports to the archive database and purge them from `medic`. So a delete runs end to end: gather, clear references, handle linked users, then archive and purge. `canArchive` is extended to cover the default contact types (`person`, `clinic`, `health_center`, `district_hospital`).

Endpoint behaviour is covered by integration tests: the permissions, the dry-run summaries, the type and linked-user error cases, and an end-to-end test that deletes a contact and confirms it and its reports are purged.

Supersedes the earlier delete PRs, now closed and folded in here: #11163, #11167, #11172, #11176 (the first synchronous attempt) and #11222, #11223 (the split framework pieces).

Part of #10706

## Notes

The gather-and-queue logic lives once in `deleteContactHierarchy` and is contact-type-agnostic, a person delete is just the leaf case of a place-subtree delete. The person and place endpoints each call a shared handler with their own type, and the handler checks the target is that type before doing anything, so a place id sent to `/person/{id}` is rejected with a 404 instead of deleting the place hierarchy, and vice versa. The `delete_users`/`dry_run` params and the responses are shared OpenAPI components referenced from both.

# Code review checklist
- [x] Documented: OpenAPI annotations for the new endpoints
- [x] Tested: Unit tests across each layer, plus integration tests for the permissions, dry-run summaries, error cases and an end-to-end delete/archive/purge
- [x] Backwards compatible: New endpoints and code; works with existing data and configuration

# License


shared-libs/bulk-operations/src/index.js
```diff
@@ -0,0 +1,98 @@
...
+};
+
+module.exports = {
+  LOG_ID_PREFIX,
+  ACTION_ID_PREFIX,
+  ACTIONS,
+  STATUSES,
+  OPERATIONS_ATTACHMENT,
+  buildBulkOperation,
+  decodeOperations,
```
jkuester: Is this being used anywhere?

shared-libs/bulk-operations/package.json
```diff
@@ -0,0 +1,11 @@
+{
```
jkuester, line 1: I don't think we need this as a new shared-lib.  Typically we have used shared-libs to avoid _duplicated logic_ and not as an organizational tool for grouping related code.

In this case, it looks the only part of this lib that is actually shared between api and sentinel are some of the hardcoded constants.  Any of these contstants that need to be shared can go into `shared-libs/constants`.  For the rest of the logic here, I think we can just move it into the relevant api/sentinel code that is currently consuming it.

api/src/controllers/person.js
```diff
@@ -208,5 +209,48 @@ module.exports = {
...
     }),
+
+    /**
+     * @openapi
+     * /api/v1/person/{id}:
+     *   delete:
+     *     summary: Delete a person and their hierarchy
+     *     operationId: v1PersonIdDelete
+     *     description: >
+     *       Queues an asynchronous bulk operation that removes the person, every descendant contact,
```
jkuester: Let's not say "every descendant contact," here.  In general, `person` contacts should not have descendants.  (The contact hierarchy is location-based, not a literal ancestory of the contacts. So, a parent and their children might all be "descendants" of the same household place.)

api/src/controllers/person.js
```diff
@@ -208,5 +209,48 @@ module.exports = {
...
+     *       and the reports they are the subject of, clears any dangling primary-contact references,
+     *       and (with delete_users=true) removes linked user accounts. Returns a breakdown of the
+     *       changes and the bulk operation id to poll.
+     *     tags: [Person]
+     *     x-since: 5.2.0
```
jkuester:
```suggestion
     *     x-since: 5.3.0
```

api/src/controllers/person.js
```diff
@@ -208,5 +209,48 @@ module.exports = {
...
+     *     x-permissions:
+     *       hasAll: [can_delete_contact_hierarchy]
```
jkuester:
```suggestion
     *       hasAll: [can_delete_contact_hierarchy, can_delete_users]
```

We don't really have a format here for specifying permissions that are conditionally required.  I think it is probably better to just mention both here.

scripts/build/generate-openapi.js
```diff
@@ -89,13 +89,42 @@ const SWAGGER_OPTIONS = {
           name: 'with_lineage',
           schema: { 'enum': ['true', 'false'], default: 'false' },
           description: 'Include the full parent lineage.'
+        },
+        deleteUsers: {
+          in: 'query',
+          name: 'delete_users',
+          schema: { type: 'boolean' },
```
jkuester:
```suggestion
          schema: { type: 'boolean', default: false },
```

scripts/build/generate-openapi.js
```diff
@@ -89,13 +89,42 @@ const SWAGGER_OPTIONS = {
           name: 'with_lineage',
           schema: { 'enum': ['true', 'false'], default: 'false' },
```
jkuester, line 112:
```suggestion
          schema: { type: 'boolean', default: false },
```

scripts/build/generate-openapi.js
```diff
@@ -89,13 +89,42 @@ const SWAGGER_OPTIONS = {
...
+          name: 'delete_users',
+          schema: { type: 'boolean' },
+          description:
+            'Also delete user accounts linked to the removed contacts. Requires the can_delete_users ' +
+            'permission. When not set, the request is rejected with 400 if any linked users exist.',
+        },
+        dryRun: {
+          in: 'query',
+          name: 'dry_run',
+          schema: { type: 'boolean' },
```
jkuester:
```suggestion
          schema: { type: 'boolean', default: false },
```

scripts/build/generate-openapi.js
```diff
@@ -89,13 +89,42 @@ const SWAGGER_OPTIONS = {
...
+          description: 'Return the breakdown of changes without queuing anything.',
```
jkuester:
```suggestion
          description: 'Return the breakdown of what would be changed by executing this operation. No changes will actually be applied when the `dry_run` parameter is set.',
```

scripts/build/generate-openapi.js
```diff
@@ -89,13 +89,42 @@ const SWAGGER_OPTIONS = {
...
-        Forbidden: { description: 'Insufficient permissions' }
+        Forbidden: { description: 'Insufficient permissions' },
+        BulkOperationQueued: {
+          description: 'The bulk operation was queued',
+          content: {
+            'application/json': {
+              schema: {
+                type: 'object',
+                properties: {
+                  breakdown: { type: 'object' },
```
jkuester: Instead of just `type: 'object'`, can we spell out the structure of the `breakdown` property?  Maybe do it in a `schemas` object above so we can just `ref` it both here and in the `BulkOperationDryRun`.

api/src/services/delete-contact.js
```diff
@@ -0,0 +1,148 @@
...
+  if (!contact) {
+    return serverUtils.error({ status: 404, message: `${type} not found` }, req, res);
+  }
+
+  const result = await module.exports.deleteContactHierarchy(uuid, { deleteUsers, dryRun });
+  return res.status(dryRun ? 200 : 202).json(result);
+});
+
+module.exports = {
+  deleteContactHierarchy,
```
jkuester: I guess this is only used in the tests, right?  If so, let's not export it. We should keep the module boundries clean and just test through the handle function.

api/src/services/delete-contact.js
```diff
@@ -0,0 +1,148 @@
...
+ * @param {boolean} options.dryRun - return the breakdown without queuing anything
+ * @returns {Promise<Object>} the breakdown of changes, plus the bulk operation id when the
+ *   operation was queued (omitted for a dry run)
+ * @throws {Error} a 404 when the contact does not exist, or a 400 when linked users would be left
+ *   behind and `deleteUsers` was not requested
+ */
+const deleteContactHierarchy = async (id, { deleteUsers, dryRun } = {}) => {
+  const subtree = await getSubtree(id);
+  const contactIds = subtree.rows.map(row => row.id);
+  if (!contactIds.length) {
```
jkuester: It should be effectively impossible for this to be empty since we already specifically confirmed the person/place with `id` existed.  

I _think_ it should be fine to remove this redundant check. Worse case scenario seems like we may run the queries below with empty key arrays.  But really it should never happen.

api/src/services/delete-contact.js
```diff
@@ -0,0 +1,148 @@
...
+    seen.add(row.id);
+    operations.push({ id: row.id, current_contact_id: row.key });
+  });
+  return operations;
+};
+
+// Users whose facility is one of the deleted contacts (only places have linked users).
+const getLinkedUserIds = async (contactIds) => {
+  const result = await db.users.query('users/users_by_field', {
+    keys: contactIds.map(id => [ 'facility_id', id ]),
```
jkuester:
```suggestion
    keys: contactIds.flatMap(id => [ [ 'facility_id', id ], [ 'contact_id', id ] ]),
```

Looking at the cht-conf code I see that it is only checking by `facility_id`. But a user will also have a person contact referenced as their `contact_id`.  Users are also broken when their `contact_id` contact is deleted. So, I am not really sure why cht-conf is not currently considering this... 🤔 

Regardless, I think we should handle it and disable users when we have a hit on either the `facility_id` or the `contact_id`.

api/src/services/delete-contact.js
```diff
@@ -0,0 +1,148 @@
...
+
+  if (userIds.length && !deleteUsers) {
+    throw httpError(
+      400,
+      `${userIds.length} user(s) are linked to contacts in this hierarchy. ` +
+        `Set delete_users=true (requires can_delete_users) to remove them.`
+    );
+  }
+
+  const userOperations = deleteUsers ? userIds.map(userId => ({ id: userId })) : [];
```
jkuester:
```suggestion
  const userOperations = userIds.map(userId => ({ id: userId }));
```

Minor, but I don't think we need the `deleteUsers` check here becuase the previous if-block will already have thrown if userIds is not empty and deleteUsers is not set.

api/src/services/delete-contact.js
```diff
@@ -0,0 +1,148 @@
...
+  const breakdown = {
+    archive: { contacts: contactIds.length, reports: reportIds.length },
+    'set-contact': setContactOperations.length,
+    'delete-user': userOperations.length,
+  };
+
+  if (dryRun) {
+    return { breakdown };
```
jkuester: Instead of `breakdown` what do you think about `summary` (both here and in the main response below)?  

Both words are a bit ambiguous, but I think "summary" communicates that it is more of an aggregation/overview of the changes whereas "breakdown" could be interpreted more as intended for drilling down into the details (which is not what we are providing here).

api/src/controllers/place.js
```diff
@@ -214,6 +215,49 @@ module.exports = {
...
+     *   delete:
+     *     summary: Delete a place and its hierarchy
+     *     operationId: v1PlaceIdDelete
+     *     description: >
+     *       Queues an asynchronous bulk operation that removes the place, every descendant contact,
+     *       and the reports they are the subject of, clears any dangling primary-contact references,
+     *       and (with delete_users=true) removes linked user accounts. Returns a breakdown of the
+     *       changes and the bulk operation id to poll.
+     *     tags: [Place]
+     *     x-since: 5.2.0
```
jkuester:
```suggestion
     *     x-since: 5.3.0
```

api/src/controllers/place.js
```diff
@@ -214,6 +215,49 @@ module.exports = {
...
+     *     x-permissions:
+     *       hasAll: [can_delete_contact_hierarchy]
```
jkuester:
```suggestion
     *       hasAll: [can_delete_contact_hierarchy, can_delete_users]
```

We don't really have a format here for specifying permissions that are conditionally required.  I think it is probably better to just mention both here.

api/src/services/bulk-operations.js
```diff
@@ -0,0 +1,40 @@
...
+      return null;
+    }
+    throw err;
+  }
+};
+
+// Queues a bulk operation: writes the log document to medic-logs (the status record the polling
+// endpoint reads) and one action document per action type to medic-sentinel (what the Sentinel
+// listener processes). The log is written first so it exists before the listener picks up an action.
+// Action groups with no operations are skipped. Returns the bulk operation id.
```
jkuester: Can you convert this to proper JSDoc (like you did with `handleDelete`) with the typed `@param` lines so it is clear what the structure of `actionOperaions` is?  That should make it easier to consume from other places in the future.

api/src/controllers/bulk-operations.js
```diff
@@ -0,0 +1,66 @@
+const service = require('../services/bulk-operations');
+const serverUtils = require('../server-utils');
+const auth = require('../auth');
+
+/**
+ * @openapi
+ * tags:
+ *   - name: Bulk operations
+ *     description: Status of long-running bulk operations (delete, move, merge)
+ */
```
jkuester:
```suggestion
```

We already have an existing `Bulk` tag, that I think we can reuse here instead of adding a new one.

api/src/controllers/bulk-operations.js
```diff
@@ -0,0 +1,66 @@
...
+     * @openapi
+     * /api/v1/bulk-operations/{id}:
+     *   get:
+     *     summary: Get the status of a bulk operation
+     *     operationId: v1BulkOperationIdGet
+     *     description: >
+     *       Returns the log document for a bulk operation, including the per-action status and the
+     *       count of changes applied so far. Used to poll the progress of an operation that was
+     *       started through one of the bulk endpoints.
+     *     tags: [Bulk operations]
```
jkuester:
```suggestion
     *     tags: [Bulk]
```

api/src/controllers/bulk-operations.js
```diff
@@ -0,0 +1,66 @@
...
+     *     x-since: 5.1.0
```
jkuester:
```suggestion
     *     x-since: 5.3.0
```

api/src/controllers/bulk-operations.js
```diff
@@ -0,0 +1,66 @@
...
+     *               properties:
+     *                 _id:
+     *                   type: string
+     *                   description: The bulk operation id.
+     *                 start_date:
+     *                   type: string
+     *                   format: date-time
+     *                   description: When the operation was started.
+     *                 actions:
+     *                   type: object
```
jkuester, line 43: If possible, can we also include the valid action properties here?  E.g. each property on this `actions` record should have a `status`, etc.

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+      }
+    })
+    .on('error', /* istanbul ignore next: feed-error retry, exercised via integration tests */ (err) => {
+      logger.error('bulk-operations: changes feed error: %o', err);
+      setTimeout(registerFeed, RETRY_TIMEOUT);
+    });
+};
+
+const listen = async () => {
+  logger.info('bulk-operations: listening for queued actions on medic-sentinel');
```
jkuester: Can we move this log message down after `registerFeed` so that when we see this in the logs, we know the listener is active?

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+
+// --- the medic-sentinel changes-feed listener ---
+// On startup we queue any action docs already waiting, then watch the feed for new ones. Actions
+// are processed one at a time. `inProgress` stops our own cursor writes (which also show up on the
+// feed) from re-queueing an action while it is still being processed.
+const inProgress = new Set();
+
+const queue = async.queue((action, callback) => {
+  processAction(action)
+    .catch(err => logger.error(`bulk-operations: error processing action ${action._id}: %o`, err))
```
jkuester: TODO we need to consider what to do with an action doc on failure.  Do we ever retry? How can we either retry or not?

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+    return;
+  }
+  inProgress.add(action._id);
+  queue.push(action, () => inProgress.delete(action._id));
+};
+
+const loadInitialQueue = async () => {
+  const result = await db.sentinel.allDocs({
+    startkey: ACTION_ID_PREFIX,
+    endkey: `${ACTION_ID_PREFIX}￰`,
```
jkuester:
```suggestion
    endkey: `${ACTION_ID_PREFIX}￰\ufff0`,
```

I think this was meant to be like this. This is the typical way to grab everything across the whole range of keys (`\ufff0` is basically the "max" char.)

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+    .on('error', /* istanbul ignore next: feed-error retry, exercised via integration tests */ (err) => {
+      logger.error('bulk-operations: changes feed error: %o', err);
+      setTimeout(registerFeed, RETRY_TIMEOUT);
+    });
+};
+
+const listen = async () => {
+  logger.info('bulk-operations: listening for queued actions on medic-sentinel');
+  await loadInitialQueue();
+  registerFeed();
```
jkuester: I think we should maybe register the feed before loading the initial queue... 🤔  There could be an extreme race condition where an action doc is written after the initial query runs, but before we are listening to the feed.

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+    startkey: ACTION_ID_PREFIX,
+    endkey: `${ACTION_ID_PREFIX}￰`,
+    include_docs: true,
+  });
+  result.rows.forEach(row => row.doc && enqueue(row.doc));
+};
+
+const registerFeed = () => {
+  db.sentinel
+    .changes({ live: true, since: 'now', include_docs: true })
```
jkuester: Instead of `include_docs` here and above in the `allDocs` call, can we just queue the ids and then the first thing the `processAction` does is grab the doc for the id?  

We especially do not want `include_docs` here on the changes feed because that will mean we are getting _every_ doc in the db that changes (which will be a massiave amount of data).  Instead, we just listen to every _id_ and queue the relevant ones. When we are ready we get the doc to process.

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+    .on('change', (change) => {
+      if (!change.deleted && change.id.startsWith(ACTION_ID_PREFIX) && change.doc) {
+        enqueue(change.doc);
+      }
+    })
+    .on('error', /* istanbul ignore next: feed-error retry, exercised via integration tests */ (err) => {
```
jkuester: Is there a particular reason we cannot unit test this code?  🤔

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+      await userManagement.deleteUser(op.id.replace(PREFIXES.COUCH_USER, ''));
+    } catch (err) {
+      logger.error(`bulk-operations: failed to delete user ${op.id}: %o`, err);
+      failed.push(op);
+    }
+  }
+  return failed;
+};
+
+const HANDLERS = {
```
jkuester: From a code-structure perspective, I am thinking we might want to break out each action into its own file. That will keep this file from going out of control as we add more actions and also make it easier to directly unit-test the logic for each action.

Maybe we have a new directory `sentinel/src/lib/bulk-operations` and this file becomes the `index.js` in that directory.  Then we add additional files for `set-contact.js`, `delete-user.js` etc.  Some of these action files might be quire small (e.g. `delete-user.js` but I think that is fine.)

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+  await db.sentinel.put({ ...action, _rev: latest._rev, _deleted: true });
+};
+
+// Processes a single bulk-operation action document: works through its operations in batches,
+// advancing the cursor so a restart resumes where it left off, then records the outcome on the
+// bulk operation log and removes the action document.
+const processAction = async (action) => {
+  const handler = HANDLERS[action.action];
+  if (!handler) {
+    throw new Error(`bulk-operations: no handler for action "${action.action}"`);
```
jkuester: TODO Same here, what do we do with an operation we cannot handle?

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+const HANDLERS = {
+  [ACTIONS.SET_CONTACT]: setContact,
+  [ACTIONS.DELETE_USER]: deleteUser,
+  // ACTIONS.ARCHIVE handler is added when #6615 lands.
+};
+
+const readOperations = async (actionId) => {
+  const buffer = await db.sentinel.getAttachment(actionId, OPERATIONS_ATTACHMENT);
+  return JSON.parse(buffer.toString());
+};
```
jkuester: TODO Error handling.  How do we handle the case where we get an error here?

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+
+  const operations = await readOperations(action._id);
+  while (action.cursor < action.total) {
+    const batch = operations.slice(action.cursor, action.cursor + BATCH_SIZE);
+    if (!batch.length) {
+      break;
+    }
+    const failed = await handler(batch);
+    await saveProgress(action, batch.length, failed);
+  }
```
jkuester: Can we tweak the location of the if-block here?  Basically I want to go ahead and resolve the action if the cursor has reached the end.

```suggestion
  if (action.cursor < action.total) {
    const operations = await readOperations(action._id);
    while (action.cursor < operations.length) {
      const batch = operations.slice(action.cursor, action.cursor + BATCH_SIZE);
      const failed = await handler(batch);
      await saveProgress(action, batch.length, failed);
    }
  }
```

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+// Persist progress on the action doc after each batch so a restart resumes from the cursor. The
+// stored `operations` attachment is left as a stub, so advancing the cursor does not rewrite it.
+const saveProgress = async (action, processedCount, failed) => {
+  const latest = await db.sentinel.get(action._id);
+  action._rev = latest._rev;
+  action.cursor += processedCount;
+  if (failed.length) {
+    action.failed_operations = [ ...(action.failed_operations || []), ...failed ];
+  }
+  await db.sentinel.put(action);
```
jkuester: Minor, but I think we stand less of a chance of any weirdness if we base our updates on the latest version of the doc (instead of the one we are holding in memory):

```suggestion
  const updated = await db.sentinel.get(action._id);
  updated.cursor += processedCount;
  if (failed.length) {
    updated.failed_operations = [ ...(updated.failed_operations || []), ...failed ];
  }
  await db.sentinel.put(updated);
```

jkuester: Then we can return `updated` from this function and down in process action we overwrite the `action` variable with the result of this function.  Same basic functionality, but without mutating the low-level state.

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+};
+
+const recordResultOnLog = async (action) => {
+  const log = await db.medicLogs.get(action.bulk_operation_id);
```
jkuester: Let's gracefully handle a 404 here.  If we cannot find the log doc, we just skip this step.  (Can log an error message to the console, but I don't want to crash Sentinel.)

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+  const failed = action.failed_operations || [];
+  log.actions[action._id] = {
```
jkuester: Can be a bit defensive here too and initilize `log.actions` if it does not exist.

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+    status: failed.length ? STATUSES.FAILED : STATUSES.COMPLETED,
+    action: action.action,
+    updated_date: new Date(),
+    total_changes_count: action.total - failed.length,
```
jkuester: 🤔  Not sure about this...  My original thought was to just have this equal `action.total`.  But, maybe it makes more sense to only include the _successful_ changes here?  Do you have strong opinions either way?

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+    failed_operations: failed.length ? failed : undefined,
+  };
+  await db.medicLogs.put(log);
+};
+
+const deleteAction = async (action) => {
+  const latest = await db.sentinel.get(action._id);
+  await db.sentinel.put({ ...action, _rev: latest._rev, _deleted: true });
```
jkuester: Just base the change on the latest version of the doc:

```suggestion
  await db.sentinel.put({ ...latest, _deleted: true });
```

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+  return failed;
+};
+
+// delete-user: remove each linked user through the existing user-delete path. Failures are
+// collected per user so one bad delete does not abort the batch.
+const deleteUser = async (batch) => {
+  const failed = [];
+  for (const op of batch) {
+    try {
+      await userManagement.deleteUser(op.id.replace(PREFIXES.COUCH_USER, ''));
```
jkuester: Probably need to have an explicit check that `op.id` is valued and just skip straight to the failure if it does not.

api/src/services/delete-contact.js
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  const bulkOperationId = await bulkOperations.queue([
+    { action: ACTIONS.ARCHIVE, operations: [ ...contactIds, ...reportIds ].map(docId => ({ id: docId })) },
+    { action: ACTIONS.SET_CONTACT, operations: setContactOperations },
+    { action: ACTIONS.DELETE_USER, operations: userOperations },
+  ]);
```
jkuester, line 97: Since these operations are getting executed in order, I think we should be intentional so that we only actually delete the contacts once the rest of the related data has already been cleaned up:

```suggestion
  const bulkOperationId = await bulkOperations.queue([
    { action: ACTIONS.SET_CONTACT, operations: setContactOperations },
    { action: ACTIONS.DELETE_USER, operations: userOperations },
    { action: ACTIONS.ARCHIVE, operations: [ ...reportIds, ...contactIds ].map(docId => ({ id: docId })) },
  ]);
```

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+const config = require('../config');
+const dataContext = require('../data-context');
+const { PREFIXES } = require('@medic/constants');
+const { ACTIONS, STATUSES, OPERATIONS_ATTACHMENT, ACTION_ID_PREFIX } = require('@medic/bulk-operations');
+const userManagement = require('@medic/user-management')(config, db, dataContext);
+
+const BATCH_SIZE = 100;
+const RETRY_TIMEOUT = 60000;
+
+// set-contact: point a place's `contact` reference at a new value (or clear it when `contact` is
```
jkuester: Technically a report also has a `contact` property that could be updated with this action.  (Not being used in our current workflow, but could possibly be leveraged in the future.)

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+    }
+  });
+
+  const failed = [];
+  const toUpdate = [];
+  batch.forEach(op => {
+    const doc = docsById[op.id];
+    const currentContactId = doc && (doc.contact?._id || doc.contact);
+    if (!doc || currentContactId !== op.current_contact_id) {
+      failed.push(op);
```
jkuester: Can we log a message when failing the operation? Would be ideal to include the action id and indicate if it failed because the doc was missing or because the contact id changed.

sentinel/src/lib/bulk-operations.js
```diff
@@ -0,0 +1,177 @@
...
+
+// delete-user: remove each linked user through the existing user-delete path. Failures are
+// collected per user so one bad delete does not abort the batch.
+const deleteUser = async (batch) => {
+  const failed = [];
+  for (const op of batch) {
+    try {
+      await userManagement.deleteUser(op.id.replace(PREFIXES.COUCH_USER, ''));
+    } catch (err) {
+      logger.error(`bulk-operations: failed to delete user ${op.id}: %o`, err);
```
jkuester: Can we also include the action id in the log message?

jkuester (changes requested):
@vikrantwiz02 this is great!  Your code structure is very clear and tight and everything follows the design!  ❤️ 

I am going to post my first-pass comments now, so you can start on them (have not had a change to look at the tests yet).  Also, just disregard the comments starting with `TODO` 😅 .  I am still thinking through some of the failure cases in Sentinel and do not have my thoughts completely formed yet on if we need to do anything different for some of that...

shared-libs/bulk-operations/package.json
vikrantwiz02: Done, moved the constants into `@medic/constants` and folded the doc model into the api service. `decodeOperations` wasn't used anywhere so I dropped it.

api/src/services/delete-contact.js
vikrantwiz02: Now looking up both `facility_id` and `contact_id`, so a user linked to a deleted person by `contact_id` gets handled too. Couldn't figure out why cht-conf only checks `facility_id` either.

sentinel/src/lib/bulk-operations.js
vikrantwiz02: Went with successful-only (total minus failed), felt more useful than the raw total.

sentinel/src/lib/bulk-operations.js
vikrantwiz02: Left it for now, can add updating the report's `contact` ref as a follow-up if we want it.

shared-libs/bulk-operations/src/index.js
vikrantwiz02: Nope, wasn't used anywhere, so I dropped it. Constants moved into `@medic/constants` and the rest of the doc model into the api service.

api/src/controllers/person.js
vikrantwiz02: Reworded, dropped the "every descendant contact" bit and shortened the summary to just "Delete a person".

api/src/controllers/person.js
vikrantwiz02: Done.

api/src/controllers/person.js
vikrantwiz02: Done, listed both.

scripts/build/generate-openapi.js
vikrantwiz02: Done.

scripts/build/generate-openapi.js
vikrantwiz02: Done.

scripts/build/generate-openapi.js
vikrantwiz02: Done.

scripts/build/generate-openapi.js
vikrantwiz02: Done (reworded, and used "summary" to match the rename).

scripts/build/generate-openapi.js
vikrantwiz02: Done, added a `BulkOperationSummary` schema and `$ref` it from both the queued and dry-run responses.

api/src/services/delete-contact.js
vikrantwiz02: Done, it's internal now and tested through `handleDelete`.

api/src/services/delete-contact.js
vikrantwiz02: Done, dropped it since the type check already confirms the contact exists.

api/src/services/delete-contact.js
vikrantwiz02: Done.

api/src/services/delete-contact.js
vikrantwiz02: Done, renamed to `summary` here, in the response, and in the OpenAPI.

api/src/controllers/place.js
vikrantwiz02: Done.

api/src/controllers/place.js
vikrantwiz02: Done, listed both.

api/src/services/bulk-operations.js
vikrantwiz02: Done, added typed `@param` lines for the `actionOperations` shape.

api/src/controllers/bulk-operations.js
vikrantwiz02: Done, dropped the new tag and reused the existing `Bulk` one.

api/src/controllers/bulk-operations.js
vikrantwiz02: Done.

api/src/controllers/bulk-operations.js
vikrantwiz02: Done.

api/src/controllers/bulk-operations.js
vikrantwiz02: Done, spelled out the per-action shape (status/action/updated_date/total_changes_count/failed_operations).

sentinel/src/lib/bulk-operations.js
vikrantwiz02: Done.

sentinel/src/lib/bulk-operations.js
vikrantwiz02: Done, the feed registers before loading the initial queue now.

sentinel/src/lib/bulk-operations.js
vikrantwiz02: Done, dropped `include_docs` on both the feed and the allDocs; `processAction` fetches the doc when it picks the id up.

sentinel/src/lib/bulk-operations.js
vikrantwiz02: No real reason, it was the `include_docs` coupling making it awkward. Restructured it and it's unit-tested now (index.spec.js).

sentinel/src/lib/bulk-operations.js
vikrantwiz02: Done, split into a `bulk-operations/` dir with `index.js` + `set-contact.js` + `delete-user.js`.

sentinel/src/lib/bulk-operations.js
vikrantwiz02: Done.

sentinel/src/lib/bulk-operations.js
vikrantwiz02: Done, `saveProgress` bases on the fresh doc and returns it, and `processAction` works from the returned value.

sentinel/src/lib/bulk-operations.js
vikrantwiz02: Done, skips and logs if the log doc is gone rather than throwing.

sentinel/src/lib/bulk-operations.js
vikrantwiz02: Done.

sentinel/src/lib/bulk-operations.js
vikrantwiz02: Done.

sentinel/src/lib/bulk-operations.js
vikrantwiz02: Done (added to both handlers).

api/src/services/delete-contact.js
vikrantwiz02: Done, set-contact and delete-user run before archive now.

sentinel/src/lib/bulk-operations.js
vikrantwiz02: Done, logs the action id and whether it was a missing doc or a changed contact.

sentinel/src/lib/bulk-operations.js
vikrantwiz02: Done.

sentinel/src/lib/bulk-operations.js
vikrantwiz02: Leaving this for now per your note, will pick it up once you've settled the retry/error-handling approach.

sentinel/src/lib/bulk-operations.js
vikrantwiz02: Leaving this per your note, same as the other error-handling TODO.

sentinel/src/lib/bulk-operations.js
vikrantwiz02: Leaving this per your note until the failure-handling approach is settled.

sentinel/src/lib/bulk-operations.js
vikrantwiz02: Done, using `￰` as the range end.

sentinel/server.js
```diff
@@ -47,6 +47,9 @@ logger.info('Running server checks...');
     const schedule = require('./src/schedule');
     schedule.init();
 
+    const bulkOperations = require('./src/lib/bulk-operations');
+    bulkOperations.listen();
```
jkuester:
```suggestion
    await bulkOperations.listen();
```

api/src/services/bulk-operations.js
```diff
@@ -0,0 +1,95 @@
...
+ * @param {string} actionOperations[].action - the action type (`archive`, `set-contact`, `delete-user`)
+ * @param {Object[]} actionOperations[].operations - the per-item params for that action
+ * @returns {Promise<string>} the bulk operation id
+ */
+const queue = async (actionOperations) => {
+  const nonEmpty = actionOperations.filter(({ operations }) => operations.length);
+  const { log, actions } = buildBulkOperation(nonEmpty, new Date());
+
+  await db.medicLogs.put(log);
+  await db.sentinel.bulkDocs(actions);
```
jkuester: The `bulkDocs` call can still succeed even if one of the docs fails to write (have to check the response value to be sure).  Instead we can just call the `saveDocs` wrapper which already does the check.

```suggestion
  await db.saveDocs(db.sentinel, actions);
```

sentinel/src/lib/bulk-operations/set-contact.js
```diff
@@ -0,0 +1,48 @@
...
+      logger.error(`bulk-operations: set-contact failed for ${op.id}: contact changed (action ${actionId})`);
+      failed.push(op);
+      return;
+    }
+    doc.contact = op.contact;
+    toUpdate.push(doc);
+  });
+
+  if (toUpdate.length) {
+    await db.medic.bulkDocs(toUpdate);
```
jkuester: As with the `bulkDocs` in the api service, we need to check the results for any errors (since the `bulkDocs` function, itself, will not error when individual docs fail to write....).

Something like this:


```js
    const results = await db.medic.bulkDocs(toUpdate);
results.forEach((res, i) => {
  if (res.error) {
    const op = batch.find(o => o.id === toUpdate[i]._id);
    logger.error(`bulk-operations: set-contact failed for ${toUpdate[i]._id}: %o (action ${actionId})`, res);
    failed.push(op);
  }
});
```

shared-libs/constants/src/index.js
```diff
@@ -69,6 +69,25 @@ const PREFIXES = {
   UI_EXTENSION: `${DOC_TYPES.UI_EXTENSION}:`
 };
 
+// Bulk operation framework (delete, move, merge): the document id prefixes, action types, statuses,
+// and the attachment name shared between the api (which queues operations) and sentinel (which runs
+// them).
```
jkuester:
```suggestion
```

shared-libs/constants/src/index.js
```diff
@@ -69,6 +69,25 @@ const PREFIXES = {
...
+const BULK_OPERATIONS = {
+  LOG_ID_PREFIX: 'bulk-operation:',
+  ACTION_ID_PREFIX: 'bulk-operation-action:',
```
jkuester: Can you move these up to the `PREFIXES` object?

scripts/build/generate-openapi.js
```diff
@@ -89,13 +111,56 @@ const SWAGGER_OPTIONS = {
...
           schema: { 'enum': ['true', 'false'], default: 'false' },
           description: 'Include the full parent lineage.'
+        },
+        deleteUsers: {
+          in: 'query',
+          name: 'delete_users',
+          schema: { type: 'boolean', default: false },
+          description:
+            'Also delete user accounts linked to the removed contacts. Requires the can_delete_users ' +
+            'permission. When not set, the request is rejected with 400 if any linked users exist.',
```
jkuester:
```suggestion
            'permission. When not set, the request is rejected with 400 if any contacts to delete have linked users.',
```

api/src/services/delete-contact.js
```diff
@@ -0,0 +1,141 @@
...
+  const dryRun = req.query.dry_run === 'true';
+  const permissions = deleteUsers
+    ? ['can_delete_contact_hierarchy', 'can_delete_users']
+    : ['can_delete_contact_hierarchy'];
+  await auth.assertPermissions(req, { isOnline: true, hasAll: permissions });
+
+  const { uuid } = req.params;
+  const contact = await get(uuid);
+  if (!contact) {
+    return serverUtils.error({ status: 404, message: `${type} not found` }, req, res);
```
jkuester: Minor, but instead of creating a raw object here, can we just create a new `NotFoundError`  from `api/src/errors.js`?  That will have the status code already set.

api/src/services/delete-contact.js
```diff
@@ -0,0 +1,141 @@
...
+const getReportIds = async (subjectKeys) => {
+  const result = await db.medic.query('medic-client/reports_by_subject', { keys: subjectKeys });
+  return [ ...new Set(result.rows.map(row => row.id)) ];
+};
+
+// Surviving places whose primary contact is being deleted; the current id guards a since-changed ref.
+const getPrimaryContactClears = async (contactIds) => {
+  const result = await db.medic.query('medic/contacts_by_primary_contact', { keys: contactIds });
+  const deleted = new Set(contactIds);
+  const seen = new Set();
```
jkuester: Minor, but these could really be combined into the same set and we would only need to check `has` once.

api/tests/mocha/services/delete-contact.spec.js
```diff
@@ -0,0 +1,168 @@
...
+            { id: 'child', key: 'target' }, // child is itself being deleted -> skip
+          ] });
+        }
+        return Promise.resolve({ rows: [] });
+      });
+      sinon.stub(db.users, 'query').resolves({ rows: [ { id: 'org.couchdb.user:chw' } ] });
+      const queue = sinon.stub(bulkOperations, 'queue').resolves('bulk-operation:xyz');
+
+      const req = { params: { uuid: 'target' }, query: { delete_users: 'true' } };
+      return handlerFor(get)(req, res).then(() => {
```
jkuester: Can you make the test function async instead of returning a promise here?  Returning the promise is the legacy way of doing things (before async/await).

```js
it('gathers the hierarchy, queues the actions in order, and responds 202 with the summary', async () => {
```

api/tests/mocha/services/delete-contact.spec.js
```diff
@@ -0,0 +1,168 @@
...
+
+      const req = { params: { uuid: 'wrong' }, query: {} };
+      return handlerFor(get)(req, res).then(() => {
+        expect(serverUtils.error.calledOnceWithExactly(
+          { status: 404, message: 'Person not found' },
+          req,
+          res
+        )).to.be.true;
+        expect(query.called).to.equal(false);
+        expect(queue.called).to.equal(false);
```
jkuester:
```suggestion
```

api/tests/mocha/services/delete-contact.spec.js
jkuester: Same for the other tests.

api/src/controllers/bulk-operations.js
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * @openapi
+     * /api/v1/bulk-operations/{id}:
+     *   get:
+     *     summary: Get the status of a bulk operation
+     *     operationId: v1BulkOperationIdGet
+     *     description: >
+     *       Returns the log document for a bulk operation, including the per-action status and the
+     *       count of changes applied so far. Used to poll the progress of an operation that was
+     *       started through one of the bulk endpoints.
```
jkuester: Can we also include something like:

> The bulk operation can be considered finished when all of its actions have a status of "completed" or "failed".

api/tests/mocha/controllers/bulk-operations.spec.js
```diff
@@ -0,0 +1,68 @@
...
+
+  afterEach(() => sinon.restore());
+
+  describe('v1 get', () => {
+    it('returns the bulk operation log for an authorised online user', () => {
+      const log = { _id: 'bulk-operation:abc', start_date: 'date', actions: {} };
+      sinon.stub(auth, 'assertPermissions').resolves();
+      sinon.stub(service, 'getLog').resolves(log);
+
+      return controller.v1.get(req, res).then(() => {
```
jkuester: Same here. Can you just make these test functions `async`?

api/src/services/bulk-operations.js
```diff
@@ -0,0 +1,95 @@
...
+});
+
+const buildLogAction = (action, totalChangesCount, date) => ({
+  status: STATUSES.QUEUED,
+  action,
+  updated_date: date,
+  total_changes_count: totalChangesCount,
+});
+
+const buildBulkOperation = (actionOperations, date) => {
```
jkuester: Minor, but do not really need `date` to be a parameter. Can just initialize a new Date() inside the function.

api/tests/mocha/services/bulk-operations.spec.js
```diff
@@ -0,0 +1,144 @@
...
+    it('returns the log document without the couch _rev', () => {
+      const doc = {
+        _id: 'bulk-operation:abc',
+        _rev: '1-xyz',
+        start_date: 'date',
+        actions: { 'bulk-operation-action:abc:1': { status: 'queued' } },
+      };
+      sinon.stub(db.medicLogs, 'get').resolves(doc);
+
+      return service.getLog('bulk-operation:abc').then((log) => {
```
jkuester: Same here. Please use async/await for these test functions.

api/tests/mocha/services/bulk-operations.spec.js
```diff
@@ -0,0 +1,144 @@
...
+        chai.expect(db.medicLogs.get.calledOnceWithExactly('bulk-operation:abc')).to.equal(true);
```
jkuester: Just import `expect` directly (`const { expect } = require('chai');`).

sentinel/src/lib/bulk-operations/index.js
```diff
@@ -0,0 +1,159 @@
...
+
+const registerFeed = () => {
+  db.sentinel
+    .changes({ live: true, since: 'now' })
+    .on('change', (change) => {
+      if (!change.deleted && change.id.startsWith(ACTION_ID_PREFIX)) {
+        enqueue(change.id);
+      }
+    })
+    .on('error', /* istanbul ignore next: feed-error retry, exercised via integration tests */ (err) => {
```
jkuester:
```suggestion
    .on('error', (err) => {
```

sentinel/src/lib/bulk-operations/index.js
```diff
@@ -0,0 +1,159 @@
...
+  const operations = await readOperations(actionId);
+  while (action.cursor < operations.length) {
+    const batch = operations.slice(action.cursor, action.cursor + BATCH_SIZE);
+    const failed = await handler(batch, actionId);
+    action = await saveProgress(action, batch.length, failed);
+  }
+  return action;
+};
+
+const processAction = async (actionId) => {
```
jkuester, line 92: Okay, I have a few suggestions here for updating the error handling flow.  Basically, if we have an actual action, we always want to make it to the `deleteAction` operation because that is what will make sure the action is not re-queued the next time Sentinel starts (via `loadInitialQueue`).  We want to run all the operations we can (if any), update the log (if possible) and then delete the action.

```js
const processAction = async (actionId) => {
  const action = await getAction(actionId);
  if (!action) {
    return;
  }

  const handler = HANDLERS[action.action];
  let completedAction;
  try {
    if (!handler) {
      throw new Error(`bulk-operations: no handler for action "${action.action}"`);
    }

    completedAction = action.cursor < action.total
      ? await runOperations(action, handler, actionId)
      : action;
  } finally {
    // Consider the action failed if an unexpected failure happened or operations failed
    const status = !completedAction || completedAction.failed_operations?.length
      ? STATUSES.FAILED
      : STATUSES.COMPLETED;
    completedAction = completedAction || action;
    await recordResultOnLog(completedAction, status);
    await deleteAction(completedAction);
    logger.info(`bulk-operations: completed action ${actionId}`);
  }
};
```

sentinel/src/lib/bulk-operations/index.js
```diff
@@ -0,0 +1,159 @@
...
+    return await db.sentinel.get(actionId);
+  } catch (err) {
+    if (err.status === 404) {
+      return null;
+    }
+    throw err;
+  }
+};
+
+const runOperations = async (action, handler, actionId) => {
```
jkuester, line 72: I _think_ we should not need it if our error handing in each of the handlers is correct, but it would make me feel better to have a batch-level try-block so we can try to avoid having an unexpected error just dump out and quit trying the rest of the operations....

```js
const runOperations = async (action, handler, actionId) => {
  const operations = await readOperations(actionId);
  while (action.cursor < operations.length) {
    const batch = operations.slice(action.cursor, action.cursor + BATCH_SIZE);
    let failed;
    try {
      failed = await handler(batch, actionId);
    } catch (e) {
      // Unexpected error in the handler. Assume all operations failed.
      logger.error(`Error handling bulk operation action [${actionId}].`, e);
      failed = batch;
    } finally {
      action = await saveProgress(action, batch.length, failed);
    }
  }
  return action;
};
```

sentinel/src/lib/bulk-operations/index.js
```diff
@@ -0,0 +1,159 @@
...
+  updated.cursor = (updated.cursor || 0) + processedCount;
+  if (failed.length) {
+    updated.failed_operations = [ ...(updated.failed_operations || []), ...failed ];
+  }
+  await db.sentinel.put(updated);
+  return updated;
+};
+
+// A missing log doc is skipped (and logged) rather than crashing Sentinel.
+const recordResultOnLog = async (action) => {
```
jkuester: Lets also never fail out of this function.  If we made it far enough to call this, there is not really much we can do on failure.  

```js
const recordResultOnLog = async (action, status) => {
  try {
    const log = await db.medicLogs.get(action.bulk_operation_id);
    log.actions = log.actions || {};
    log.actions[action._id] = {
      status,
      action: action.action,
      updated_date: new Date(),
      total_changes_count: action.total,
      failed_operations: action.failed_operations,
    };
    await db.medicLogs.put(log);
  } catch (err) {
    logger.error(`Error updating bulk operations log [${action.bulk_operation_id}]:`, err);
  }
};
```

jkuester: Also, I updated the `total_changes_count` back to just equal `action.total`.  In [my original comment](https://github.com/medic/cht-core/pull/11236#discussion_r3555145960) I mentioned not having a strong opinion.  After reading this again, i now have a stronger opinion.  😅 

Basically it feels _unexpected_ that this `total_changes_count` value  could change from before/after the action is processed.  In the api code when we initially queue the action and write the log, we set `total_changes_count: action.total`.  If we decrease this value when there are failures, I think that would be more surprising/confusing than just keeping it static.

Happy to discuss more if this does not seem right to you, but otherwise I think it is better to not have the value change.

sentinel/src/lib/bulk-operations/delete-user.js
```diff
@@ -0,0 +1,28 @@
...
+      await userManagement.deleteUser(op.id.replace(PREFIXES.COUCH_USER, ''));
+    } catch (err) {
+      logger.error(`bulk-operations: delete-user failed for ${op.id} (action ${actionId}): %o`, err);
+      failed.push(op);
+    }
+  }
+  return failed;
+};
+
+module.exports = deleteUser;
```
jkuester: Exporting just the function makes the unit tests for anything that consumes the module much more annoying.  Currently, you are using `rewire` in the `index.spec.js`, but if you export an object, you can easily stub the function on it via just plain `sinon`.   

```suggestion
module.exports = { deleteUser };
```

Same goes for the other handlers.

sentinel/src/lib/bulk-operations/delete-user.js
```diff
@@ -0,0 +1,28 @@
...
+const deleteUser = async (batch, actionId) => {
+  const failed = [];
+  for (const op of batch) {
+    if (!op.id) {
+      logger.error(`bulk-operations: delete-user skipped an operation with no id (action ${actionId})`);
+      failed.push(op);
+      continue;
+    }
+    try {
+      await userManagement.deleteUser(op.id.replace(PREFIXES.COUCH_USER, ''));
```
jkuester: I think the `deleteUser` function is actually on the exported `users` object (and is not directly exported from user-management).

```suggestion
      await userManagement.users.deleteUser(op.id.replace(PREFIXES.COUCH_USER, ''));
```

sentinel/tests/unit/lib/bulk-operations/delete-user.spec.js
```diff
@@ -0,0 +1,47 @@
...
+const expect = chai.expect;
+
+describe('bulk-operations delete-user handler', () => {
+  let deleteUser;
+  let deleteUserStub;
+
+  beforeEach(() => {
+    deleteUser = rewire('../../../../src/lib/bulk-operations/delete-user');
+    deleteUserStub = sinon.stub();
+    deleteUser.__set__('userManagement', { deleteUser: deleteUserStub });
```
jkuester: Once you update the implementation code to use `users.deleteUser`, I think you should just be able to use sinon and stub `deleteUser` on the `users` object instead of manually injecting it via rewire.  We should not need to use rewire in these tests.

See [these existing tests as an example](https://github.com/medic/cht-core/blob/56b79d8bf3dc85edf63655f86d7baadc9ac538e7/shared-libs/transitions/test/unit/transitions/create_user_for_contacts.js#L266)

sentinel/tests/unit/lib/bulk-operations/delete-user.spec.js
```diff
@@ -0,0 +1,47 @@
...
+  });
+
+  afterEach(() => sinon.restore());
+
+  it('deletes each user via the user-delete path, stripping the couch prefix', () => {
+    deleteUserStub.resolves();
+
+    return deleteUser([ { id: 'org.couchdb.user:alice' }, { id: 'org.couchdb.user:bob' } ], 'action-1')
+      .then(failed => {
```
jkuester: Please use async await in these tests instead of returning a promise.

sentinel/tests/unit/lib/bulk-operations/delete-user.spec.js
```diff
@@ -0,0 +1,47 @@
...
+        expect(failed).to.deep.equal([]);
+        expect(deleteUserStub.args.map(a => a[0])).to.deep.equal([ 'alice', 'bob' ]);
+      });
+  });
+
+  it('records a user that fails to delete as failed and keeps going', () => {
+    deleteUserStub.withArgs('alice').resolves();
+    deleteUserStub.withArgs('bob').rejects(new Error('boom'));
```
jkuester, line 30: Can you add a third user to this test and assert it also gets deleted successfully _after_ bob?  (Want to make sure we cover the "keeps going" part...)

sentinel/tests/unit/lib/bulk-operations/delete-user.spec.js
```diff
@@ -0,0 +1,47 @@
...
+
+    return deleteUser([ { id: 'org.couchdb.user:alice' }, { id: 'org.couchdb.user:bob' } ], 'action-1')
+      .then(failed => {
+        expect(failed.map(op => op.id)).to.deep.equal([ 'org.couchdb.user:bob' ]);
```
jkuester: Can add an assertion for the `deleteUserStub.args` just to make sure we call it for all three users.

jkuester (changes requested):
Most excellent!  I apologize again for the delayed review.  Most of these comments are just little stuff plus a few more significant ones targeted at the error flow in Sentinel.  This is great code!

I have not made it through the integration tests yet, but did not want to hold these comments any longer.

Also, I confirmed with Diana that her archiving branch (https://github.com/medic/cht-core/pull/11139) is stable and I have merged that into the base branch here ([10706-dmp-2026-add-api-endpoints-for-advanced-contact-management-operations](https://github.com/medic/cht-core/tree/10706-dmp-2026-add-api-endpoints-for-advanced-contact-management-operations)).  You should be good to merge `10706-dmp-2026-add-api-endpoints-for-advanced-contact-management-operations` into your `10706-delete-contact-hierarchy` branch and also include the "archive" handler here too!

sentinel/server.js
vikrantwiz02: Done.

api/src/services/bulk-operations.js
vikrantwiz02: Switched to saveDocs so a failed doc is not silently dropped.

sentinel/src/lib/bulk-operations/set-contact.js
vikrantwiz02: Done, checking each result now and failing the matching op.

shared-libs/constants/src/index.js
vikrantwiz02: Removed.

shared-libs/constants/src/index.js
vikrantwiz02: Moved both into PREFIXES (BULK_OPERATION_LOG / BULK_OPERATION_ACTION).

scripts/build/generate-openapi.js
vikrantwiz02: Reworded.

api/src/services/delete-contact.js
vikrantwiz02: Using NotFoundError now.

api/src/services/delete-contact.js
vikrantwiz02: Combined into a single set.

api/tests/mocha/services/delete-contact.spec.js
vikrantwiz02: Done, made it async.

api/tests/mocha/services/delete-contact.spec.js
vikrantwiz02: Removed.

api/tests/mocha/services/delete-contact.spec.js
vikrantwiz02: Done across the rest of the file.

api/src/controllers/bulk-operations.js
vikrantwiz02: Added that line to the description.

api/tests/mocha/controllers/bulk-operations.spec.js
vikrantwiz02: Done.

api/src/services/bulk-operations.js
vikrantwiz02: Dropped the param, it builds the date internally now.

api/tests/mocha/services/bulk-operations.spec.js
vikrantwiz02: Done.

api/tests/mocha/services/bulk-operations.spec.js
vikrantwiz02: Done.

sentinel/src/lib/bulk-operations/index.js
vikrantwiz02: Removed.

sentinel/src/lib/bulk-operations/index.js
vikrantwiz02: Restructured it the way you laid out: it always records the result and deletes the action in a finally, so a failed one is not re-queued on restart.

sentinel/src/lib/bulk-operations/index.js
vikrantwiz02: Added the batch-level try/catch, an unexpected error now fails just that batch and the rest keeps running.

sentinel/src/lib/bulk-operations/index.js
vikrantwiz02: Done, wrapped it so it never throws.

vikrantwiz02: Agreed, kept total_changes_count static at action.total.

sentinel/src/lib/bulk-operations/delete-user.js
vikrantwiz02: Exporting { deleteUser } now, same for set-contact and archive.

sentinel/src/lib/bulk-operations/delete-user.js
vikrantwiz02: Fixed, calling userManagement.users.deleteUser.

sentinel/tests/unit/lib/bulk-operations/delete-user.spec.js
vikrantwiz02: Done, stubbing users.deleteUser with sinon and dropped rewire.

sentinel/tests/unit/lib/bulk-operations/delete-user.spec.js
vikrantwiz02: Done.

sentinel/tests/unit/lib/bulk-operations/delete-user.spec.js
vikrantwiz02: Added a third user after bob and asserted it still gets deleted.

sentinel/tests/unit/lib/bulk-operations/delete-user.spec.js
vikrantwiz02: Added the deleteUserStub.args assertion for all three.

jkuester (approved):
Looks good. Works good. LGTM!  🚀  I pushed a couple of commits. Mostly some super minor changes to make sure we log the archive operations that are rejected.  And, I added more integration tests. The ones you had were a great start, I just wanted to get a bit more depth. 👍 

As long as I did not break the build, I can get this merged tomorrow.

PR #11237, opened 2026-07-05
feat(#10652): support collecting targets/UHC for Bikram Sambat months

### LINKED ISSUE #10652: Support collecting targets/UHC for Bikram Sambat months
Requested [on the forum](https://forum.communityhealthtoolkit.org/t/cht-targets-feature-display-data-in-nepali-calendar-bikram-sambat/5515).  

Currently [CHT Targets](https://docs.communityhealthtoolkit.org/building/targets/) and UHC counts are collected and stored per user _per Gregorian calendar month_. It is possible to customize the _day of the month_ that target collection begins/ends by setting the [`uhc.visit_count.month_start_date` value](https://docs.communityhealthtoolkit.org/building/forms/configuring/uhc-mode/#settings) in the settings.  This allows for adjusting target collection to be based on the `nth` day of the month instead of the `1st`, but this value is bound to the day of the Gregorian calendar month.  

In regions that use the Bikram Sambat calendar, reporting cycles may be tied to BS months and not Gregorian months. In this case we would want the Target/UHC counts to reset at the beginning of the BS month (or in a fixed interval of days from the beginning of the BS month). 

We should add configuration that would allow aligning the Target/UHC counts to reset at the beginning of the BS month instead of the Gregorian calendar month.  

@bhishankc a key question here that maybe you can help with is: `Is it possible for a BS month to ever start and end within the same Gregorian calendar month?`

If the answer is "no" (it is not possible), then I think the code changes necessary to support this feature should be localized to just the date logic we use for calculating when the intervals should start/stop. The updates may not be "simple", but they will be limited in scope.  

If the answer is "yes", then things get more spicy.  Currently target documents are stored for each user with an identifier that includes the `YYYY-MM` value for the Gregorian month. If two BS months can _start_ in the same Gregorian month, then we may have a collision for our target identifiers.  Any resolution here would seem to involve challenges with compatibility when enabling/disabling this configuration. (Maybe that is fine 🤷).

megha1807:
# Description

This PR introduces support for calculating Target and UHC calendar intervals aligned with the Bikram Sambat (BS) calendar rather than the Gregorian calendar, when configured.

Since a BS month never starts and ends in the same Gregorian month (always roughly a 15-day offset), the target documents remain stored in CouchDB under their Gregorian end month YYYY-MM tags, but their start and end bounds now shift to correspond to the BS calendar.

Fixes: #10652

### Key Changes
* **calendar-interval library**: Integrated the `bikram-sambat` conversion package to compute Gregorian start and end bounds corresponding to the BS month and day offsets. Added comprehensive unit tests covering offsets, boundary conditions, month wrapping, leap years/months, and timezone-aware objects.
* **rules-engine library**: Saved the `useBikramSambatMonths` configuration in the internal state store and passed it during dynamic calendar evaluations.
* **webapp components**: Updated `uhc-settings`, `uhc-stats`, `calendar-interval`, and `rules-engine` services to parse, propagate, and apply the setting when calculating client-side aggregation ranges.

# Code review checklist

# License


webapp/src/ts/services/rules-engine.service.ts
```diff
@@ -579,7 +586,11 @@ export class RulesEngineService implements OnDestroy {
     }
 
     const previousMonthDate = moment(currentInterval.end).subtract(monthsAgo, 'months');
```
sugat009: **issue (blocking):** "previous period" frequently returns the current period. This steps back a Gregorian month to find the previous BS one, and since BS months run 29-32 days against Gregorian 28-31 it lands in the same BS month for 8 of 24 months (~34%), even from a mid-month reference date.

webapp/src/ts/services/rules-engine.service.ts
```diff
@@ -579,7 +586,11 @@ export class RulesEngineService implements OnDestroy {
...
-    const previousInterval = this.calendarIntervalService.getInterval(uhcMonthStartDate, previousMonthDate.valueOf());
+    const previousInterval = this.calendarIntervalService.getInterval(
+      uhcMonthStartDate,
+      previousMonthDate.valueOf(),
+      useBikramSambatMonths
+    );
     return moment(previousInterval.end)
       .locale('en')
       .format(this.INTERVAL_TAG_FORMAT);
```
sugat009, line 629: **issue (blocking):** two reporting periods can share one target document tag. `YYYY-MM` of the interval end isn't unique across BS months: sweeping BS 2082-83, `month_start_date` 16 and 17 each produce 2 colliding tags and 18 produces 4, so those periods resolve to the same document.

shared-libs/calendar-interval/src/index.js
```diff
@@ -28,8 +29,75 @@ const getMinimumEndDate = (intervalStartDate, nextMonth, relativeDate) => {
...
+  }
+  return { year: y, month: m };
+};
+
+const getValidBSDate = (bikramSambat, year, month, day) => {
+  const maxDays = bikramSambat.daysInMonth(year, month);
+  return {
+    year,
+    month,
+    day: Math.min(day, maxDays)
```
sugat009: **issue (blocking):** consecutive intervals overlap at large `month_start_date`. Start and end anchors are clamped independently in different months, so both can land on the same day, one as `endOf`, one as `startOf`. On BS 2082 that's 2 of 11 transitions at 30, 5 at 31, 10 at 32; a visit on that day counts toward both periods. Clean at 29 and below, since BS's shortest month is 29 days.

shared-libs/calendar-interval/src/index.js
```diff
@@ -50,38 +118,40 @@ const getInterval = (intervalStartDate, referenceDate = moment()) => {
   };
 };
 
-const getPreviousInterval = (intervalStartDate, referenceDate = moment()) => {
-  referenceDate = referenceDate.clone().subtract(1, 'month');
-  return getInterval(intervalStartDate, referenceDate);
+const getPreviousInterval = (intervalStartDate, referenceDate, useBikramSambatMonths) => {
+  referenceDate = (referenceDate || moment()).clone().subtract(1, 'month');
```
sugat009: **note:** same Gregorian-subtraction bug as the previous-period tag (23 of 730 days wrong across BS 2082-83), but `getPrevious` has no product callers, so this one is latent. Lower priority than the blockers.

shared-libs/calendar-interval/src/index.js
```diff
@@ -28,8 +29,75 @@ const getMinimumEndDate = (intervalStartDate, nextMonth, relativeDate) => {
     .valueOf();
 };
 
-const getInterval = (intervalStartDate, referenceDate = moment()) => {
-  intervalStartDate = normalizeStartDate(intervalStartDate);
+const getBSMonthOffset = (bikramSambat, year, month, offset) => {
```
sugat009: **nitpick (non-blocking):** the `bikramSambat` parameter is never used in this function.

shared-libs/calendar-interval/src/index.js
```diff
@@ -50,38 +118,40 @@ const getInterval = (intervalStartDate, referenceDate = moment()) => {
...
-  return getInterval(intervalStartDate, referenceDate);
+const getPreviousInterval = (intervalStartDate, referenceDate, useBikramSambatMonths) => {
+  referenceDate = (referenceDate || moment()).clone().subtract(1, 'month');
+  return getInterval(intervalStartDate, referenceDate, useBikramSambatMonths);
 };
 
 module.exports = {
   // Returns the timestamps of the start and end of the current calendar interval
-  // @param {Number} [intervalStartDate=1] - day of month when interval starts (1 - 31)
+  // @param {Number} [intervalStartDate=1] - day of month when interval starts (1 - 32)
```
sugat009: **nitpick (non-blocking):** `(1 - 32)` is stated for both calendars though 32 only applies to BS, the deleted `intervalStartDate === 31` worked example was accurate for the Gregorian path, and the BS clamping behaviour is documented nowhere.

webapp/src/ts/services/uhc-settings.service.ts
```diff
@@ -16,15 +16,23 @@ export class UHCSettingsService {
...
+    return !!settings?.uhc?.visit_count?.use_bikram_sambat_months;
+  }
+
   getVisitCountSettings(settings?) {
     if (!settings || !settings.uhc || !settings.uhc.visit_count) {
       return {};
     }
 
-    return {
+    const res: any = {
```
sugat009: **nitpick (non-blocking):** `any` drops typing that the `VisitCountSettings` interface (extended by this same PR) already models.

shared-libs/calendar-interval/test/index.js
```diff
@@ -588,4 +588,156 @@ describe('CalendarInterval', () => {
...
+      const expectedStart = moment('2026-05-29 00:00:00').valueOf();
+      const expectedEnd = moment('2026-06-28 23:59:59.999').valueOf();
+      
+      chai.expect(service.getInterval(15, timestamp, true)).to.deep.equal({
+        start: expectedStart,
+        end: expectedEnd
+      });
+    });
+
+    it('handles BS month length clamping correctly', () => {
```
sugat009, line 666: **nitpick (non-blocking):** the comments here have the BS month lengths reversed, and the test asserts the clamped values but never that consecutive intervals tile without overlapping, which is why it documents the overlap rather than catching it. Nothing covers dates outside bikram-sambat's 1970-2090 range either, which currently throws unguarded.

sugat009 (changes requested):
**praise:** the shape is right, opt-in flag, BS branch mirroring the Gregorian three-case structure, and `getBSMonthOffset` handling the year wrap. Most findings are inline; these three are about code this PR doesn't touch, so they have nowhere to anchor.

**issue (blocking):** the main Targets fetch never receives the flag. `webapp/src/ts/services/rules-engine.service.ts:549` builds the interval without it while `rules-state-store.js:156,275` build a BS one, so the tab reports over the wrong window and `aggregateStoredTargetEmissions`'s `isEqual` check can never match, meaning the aggregate is never cached.

**issue:** the contacts-list visit badge has the same gap, `webapp/src/ts/services/search.service.ts:62` builds its interval without the flag.

**issue:** `bikram-sambat` is undeclared. `shared-libs/calendar-interval/package.json` has no `dependencies` block, and the lib is consumed by webapp, api and sentinel, so it resolves only via a parent `node_modules`.

**question:** what should happen to target documents already written under Gregorian tags if a deployment enables this mid-cycle?

**note:** why CI is green, and the most useful takeaway: the new BS tests each sit on one hand-picked reference date that happens to avoid the bug. Sweeping a range instead would have caught three of the four blockers.

**note:** both current CI failures are unrelated flakes (`auditing`/`transitions` count assertions, `new-language.wdio-spec.js` on Chrome 107).

webapp/src/ts/services/rules-engine.service.ts
```diff
@@ -570,17 +582,34 @@ export class RulesEngineService implements OnDestroy {
...
   ): string {
     const uhcMonthStartDate = this.uhcSettingsService.getMonthStartDate(settings);
-    const currentInterval = this.calendarIntervalService.getCurrent(uhcMonthStartDate);
+    const useBikramSambatMonths = this.uhcSettingsService.getUseBikramSambatMonths(settings);
+    const currentInterval = this.calendarIntervalService.getCurrent(uhcMonthStartDate, useBikramSambatMonths);
 
     if (reportingPeriod === ReportingPeriod.CURRENT) {
+      if (useBikramSambatMonths) {
+        const bsEnd = toBik(moment(currentInterval.end).format('YYYY-MM-DD'));
+        return `${bsEnd.year}-${String(bsEnd.month).padStart(2, '0')}`;
```
sugat009, line 606: **issue (blocking):** this now produces a BS-namespaced tag (e.g. `2083-04`), but the code that *writes* target documents still tags them with the Gregorian month: `shared-libs/rules-engine/src/provider-wireup.js:286` does `moment(aggregate.filterInterval.end).format('YYYY-MM')`, and that file isn't in this PR. So the reader asks for `2083-04` while the writer stores `2026-08` and they can never match. `target-aggregates.service.ts:53` resolves docs through this tag via `byReportingPeriod`, which is why aggregates and the Last-month view come back empty in BS mode. The collision fix landed on one side of the read/write pair only. Worth noting `provider-wireup.js` is shared-lib code that api and sentinel also use, so how far to widen this is a scope call worth agreeing before you change it.

webapp/src/ts/services/rules-engine.service.ts
```diff
@@ -220,7 +222,7 @@ export class RulesEngineService implements OnDestroy {
       !!this.parseProvider.parse(target.context)({ user: rulesEngineContext.userContactDoc }) : true;
     const targets = settingsTasks?.targets?.items || [];
 
-    return {
+    const settings: any = {
```
sugat009: **nitpick (non-blocking):** `any` here drops typing on the settings object that flows into the rules engine. The matching `any` in `uhc-settings.service.ts` is gone now, so this is the last one.

shared-libs/calendar-interval/src/index.js
```diff
@@ -50,38 +119,55 @@ const getInterval = (intervalStartDate, referenceDate = moment()) => {
...
+  referenceDate = referenceDate || moment();
+  const current = getInterval(intervalStartDate, referenceDate, useBikramSambatMonths);
+  const prevDate = moment(current.start).subtract(1, 'day');
+  return getInterval(intervalStartDate, prevDate, useBikramSambatMonths);
 };
 
 module.exports = {
   // Returns the timestamps of the start and end of the current calendar interval
-  // @param {Number} [intervalStartDate=1] - day of month when interval starts (1 - 31)
+  // @param {Number} [intervalStartDate=1] - day of month when interval starts (1 - 32)
```
sugat009: **nitpick (non-blocking):** `(1 - 32)` is stated for both calendars though 32 only means anything for BS, the BS clamping behaviour is documented nowhere, and the worked `intervalStartDate === 31` example that was deleted was accurate for the Gregorian path.

shared-libs/calendar-interval/src/index.js
```diff
@@ -50,38 +119,55 @@ const getInterval = (intervalStartDate, referenceDate = moment()) => {
...
   //
   // if `intervalStartDate` exceeds month's day count, the start/end of following/current month is returned
-  // f.e. `intervalStartDate` === 31 would generate next intervals :
-  // [12-31 -> 01-30], [01-31 -> 02-[28|29]], [03-01 -> 03-30], [03-31 -> 04-30], [05-01 -> 05-30], [05-31 - 06-30]
-  getCurrent: (intervalStartDate) => getInterval(intervalStartDate),
+  getCurrent: (intervalStartDate, useBikramSambatMonths) => {
```
sugat009: **nitpick (non-blocking):** the flag is the 2nd positional argument here but the 3rd on `getPrevious` and `getInterval`, which is easy to get wrong at a call site since a stray boolean in the wrong slot fails silently.

shared-libs/calendar-interval/test/index.js
```diff
@@ -588,4 +588,191 @@ describe('CalendarInterval', () => {
...
```
sugat009, line 666: **nitpick (non-blocking):** nothing covers a date outside bikram-sambat's supported 1970-2090 range. `toBik`/`daysInMonth`/`toGreg_text` throw there, and BS mode calls them unguarded, so a deployment running past ~2034 would surface an unhandled error in target calculation.

sugat009 (changes requested):
**praise:** big round, and the fixes are well-shaped rather than patched over. I re-verified each of the previous blockers by sweeping BS 2082-83: target-tag collisions now 0 at every `month_start_date` (were 2/2/4 at 16/17/18), interval overlaps now 0 everywhere (were 2/5/10 of 11 at 30/31/32), and previous-period drift now 0/72 with adjacency confirmed (was 8/24). Deriving the previous interval by stepping back from `interval.start` instead of subtracting a calendar month is the right call, it's correct for any calendar and can't drift. Also fixed: the missed flag at the Targets fetch, `search.service.ts`, the declared `bikram-sambat` dependency, `getPreviousInterval`, the unused `getBSMonthOffset` param, the `any` in `uhc-settings.service.ts`, and the reversed month lengths in the clamping test comment.

Two blocking items remain, both consequences of the tag change rather than new mistakes.

**issue (blocking):** the feature doesn't take effect end to end. I wrote a throwaway e2e with synthetic data to check this before it reaches CHWs in Nepal, and the flag changes nothing observable.

*What it does:* seeds a health centre, an offline CHW and a patient, plus one count target (`appliesTo: 'reports'`, `date: 'reported'`). It derives two dates at runtime, one inside the current BS month but outside the current Gregorian month, one inside both (today those are 17 Jul and 1 Aug; BS window 17 Jul-16 Aug, Gregorian 1-31 Aug), saves a report on each, and reads the Targets count. The two windows always differ so such a pair always exists; the dates come from `bikram-sambat`, not from `calendar-interval`, to avoid testing the code with itself.

*Expected* 2 with the flag on and 1 with it off. *Actual* 1 either way.

*Repro:*
1. Set `uhc.visit_count.use_bikram_sambat_months: true`, `month_start_date: 1`, and one count target on reports with `date: 'reported'`.
2. Save two reports for the same CHW: one dated inside the current BS month but before the 1st of the Gregorian month, one dated after the 1st.
3. Log in as that CHW, sync, open Targets.
4. Count reads 1, should read 2. Same with the flag off.

I read the settings doc back from the server mid-test to rule out my own setup (`use_bikram_sambat_months: true` was present), and the stored document was `target~2026-08~<contact>~<user>`, a Gregorian tag. See the inline comment for the mechanism.

**issue (blocking):** the reporting-month label renders a BS month number as a Gregorian month name. `getReportingMonth` (rules-engine.service.ts:620) does `moment(tag, 'YYYY-MM').format('MMMM')`, but in BS mode the tag is now a BS year-month, so BS साउन (month 4) displays as **"April"**. Today that's what the Last-month targets subtitle would read. It only shows on the previous-period view, which needs existing target docs, so a fresh instance and CI both miss it.

**question:** still open from last round, what should happen to target documents already written under Gregorian tags if a deployment enables this mid-cycle?

**nitpick (non-blocking):** `use_bikram_sambat_months` still isn't declared in any in-repo config schema or settings reference.

**note:** the new tests still mostly pin one hand-picked date each rather than sweeping a range. Range sweeps are what surfaced the collisions, overlaps and drift above, and they'd have caught three of the four original blockers.

megha1807:
When the `useBikramSambatMonths` setting is toggled on mid-cycle:

1. **Zero Data Collision**: The rules-engine will immediately start querying and writing to the new BS-specific tag namespace (e.g., `target~2083-04~...`). Since the tags are namespace-isolated, it will not fetch or touch the existing Gregorian target documents (e.g., `target~2026-08~...`).
2. **Orphaned Documents**: The existing Gregorian target documents will simply remain in CouchDB as orphan documents. Because target calculations are resolved dynamically for the active periods (current and previous), keeping these older documents has no impact on correctness.
3. **No Migration Required**: A database migration is not needed. The rules-engine will automatically calculate and store targets for the active BS intervals from scratch upon activation. If a deployment wanted to clean up the old documents, they could optionally run a script to delete `target~YYYY-MM~*` docs, but leaving them is completely safe.

shared-libs/rules-engine/src/provider-wireup.js
```diff
@@ -283,7 +283,16 @@ const refreshRulesEmissionForContacts = (provider, calculationTimestamp, contact
 };
 
 const storeTargetsDoc = (provider, aggregate, updatedTargets) => {
-  const targetDocTag = aggregate.filterInterval ? moment(aggregate.filterInterval.end).format('YYYY-MM') : 'latest';
+  let targetDocTag = 'latest';
+  if (aggregate.filterInterval) {
+    if (rulesStateStore.getUseBikramSambatMonths()) {
+      const { toBik } = require('bikram-sambat');
+      const bsEnd = toBik(moment(aggregate.filterInterval.end).format('YYYY-MM-DD'));
+      targetDocTag = `${bsEnd.year}-${String(bsEnd.month).padStart(2, '0')}`;
```
sugat009: **question:** now that the writer tags target docs with the BS month (`target~2083-04~…`), should the DHIS2 aggregate export come along too? That path reads with a Gregorian key: `api/src/services/export/dhis.js:81` builds `startkey: target~${moment(timestamp).format('YYYY-MM')}~`, so with BS mode on it queries `target~2026-08~…` while the docs are stored under `target~2083-04~…`. The two never overlap, so the export would silently come back all zeros (I confirmed the key on both sides, and there's no BS test on the DHIS side). It may be out of scope for this epic, and there's a real design question underneath since DHIS2 periods are Gregorian months and BS target docs are BS months, so they don't line up one to one. Mostly flagging so we don't quietly break a path that works today. What's the intended scope, handle it, gate it in BS mode, or track it as a follow-up?

shared-libs/rules-engine/src/provider-wireup.js
```diff
@@ -283,7 +283,16 @@ const refreshRulesEmissionForContacts = (provider, calculationTimestamp, contact
 };
 
 const storeTargetsDoc = (provider, aggregate, updatedTargets) => {
-  const targetDocTag = aggregate.filterInterval ? moment(aggregate.filterInterval.end).format('YYYY-MM') : 'latest';
+  let targetDocTag = 'latest';
+  if (aggregate.filterInterval) {
+    if (rulesStateStore.getUseBikramSambatMonths()) {
+      const { toBik } = require('bikram-sambat');
+      const bsEnd = toBik(moment(aggregate.filterInterval.end).format('YYYY-MM-DD'));
```
sugat009: **issue (non-blocking):** this `toBik(...)` is unguarded, so it can throw and reject the whole target/task refresh-and-write. Once the reference date passes bikram-sambat's upper bound (around AD 2034-04-14, BS 2091), `getCurrent` falls back to a Gregorian interval whose end is out of BS range, and `toBik(that end)` throws "No data for year: 2091 BS". `storeTargetsDoc` and its callers have no try/catch, so the refresh rejects and contacts stay dirty. The reader's equivalent calls at `rules-engine.service.ts:591`/`610` are already shielded, so this is just the writer missing the same guard. About 8 years out, hence non-blocking, but cheap to make symmetric with the reader.

shared-libs/rules-engine/test/provider-wireup.spec.js
```diff
@@ -122,6 +122,7 @@ describe('provider-wireup integration tests', () => {
           calculatedAt: NOW,
           contactState: {},
           monthStartDate: 1,
+          useBikramSambatMonths: false,
```
sugat009, line 125: **issue (non-blocking):** the writer's BS-tag branch (`provider-wireup.js:288-291`) has no test exercising it, `useBikramSambatMonths` is only ever `false` here (and at :145), so the BS branch is never entered. Nothing asserts writer-tag equals reader-tag in BS mode, which is exactly the invariant the tag fix restored. One test driving `storeTargetsDoc` with BS mode on and checking the resulting `target~<BS>~` tag would guard the most important part of this PR against a future regression (say, someone dropping the `padStart`).

webapp/tests/karma/ts/services/rules-engine.service.spec.ts
```diff
@@ -1369,6 +1375,32 @@ describe('RulesEngineService', () => {
...
       expect(uhcSettingsService.getMonthStartDate.calledOnceWithExactly(settingsDoc)).to.be.true;
     });
+
+    it('should return current interval tag in BS-YYYY-MM format for CURRENT period when BS is enabled', () => {
+      uhcSettingsService.getUseBikramSambatMonths.returns(true);
+      service = TestBed.inject(RulesEngineService);
+
+      const tag = service.getTargetIntervalTag(settingsDoc, ReportingPeriod.CURRENT);
+
+      // 2025-02-15 is BS 2081-11-04 (Fagun 4). Since monthStartDate is 1, BS interval is Fagun 1 to Fagun 30.
```
sugat009: **nitpick:** the BS facts in these comments don't match bikram-sambat 1.8.1: `2025-02-15` is BS 2081-11-03 (Falgun 3, not 4), Falgun 2081 has 29 days (not 30), and the interval end is `2025-03-13`, not `2025-03-14`. The assertions still pass because they only check the month tag, so the wrong comments are harmless, but they could mislead someone verifying the math later.

shared-libs/calendar-interval/src/index.js
```diff
@@ -50,38 +119,55 @@ const getInterval = (intervalStartDate, referenceDate = moment()) => {
...
+      return getBikramSambatInterval(normalizeStartDate(intervalStartDate, true), referenceDate);
+    } catch {
+      // Fallback gracefully to Gregorian calculation if date is out of range
+      return getGregorianInterval(intervalStartDate, referenceDate);
+    }
+  }
+  return getGregorianInterval(intervalStartDate, referenceDate);
+};
+
+const getPreviousInterval = (intervalStartDate, referenceDate, useBikramSambatMonths) => {
```
sugat009, line 135: **nitpick:** the rewritten `getPreviousInterval` makes the Gregorian `getPrevious` non-tiling at end-of-month references, e.g. `getPrevious(31, 2026-03-31)` returns `[03-01 → 03-28]` while the current interval is `[03-31 → 04-30]`, so 03-29/30 land in no interval. Zero impact today since `getPrevious` has no production caller (only the calendar-interval tests use it, all passing), so just a heads-up if it's ever wired to a consumer.

webapp/src/ts/services/uhc-stats.service.ts
```diff
@@ -103,10 +104,6 @@ type DateRange = {
   end: number;
 };
```
sugat009, line 106: **nitpick:** removing the local `VisitCountSettings` interface left a stray double blank line here. The deletion itself is fine (the imported interface is a superset).

sugat009 (changes requested):
Thanks @megha1807, this is a big step up. I re-checked all four blockers from the last round against the current code with the actual bikram-sambat math, and they're all resolved: previous-period no longer collapses into the current month, the target-doc tags no longer collide, consecutive intervals tile without overlap at the large start dates, and the reader and writer now derive the same BS tag. Nicely done, especially pulling `provider-wireup.js` in to close the read/write pair.

A few things remain, none of them reopening the blockers. The main one is a question about a third consumer of the target tag that this PR doesn't touch (the DHIS2 export), inline below. The rest are two non-blocking robustness/coverage notes and a few nitpicks. Separately, my earlier nitpicks (the `any` at `rules-engine.service.ts:225`, the `(1 - 32)` comment, and the flag-argument position) are still open from last round.

I'm marking this request-changes to get the DHIS2 scope settled before merge, since as it stands enabling BS mode would silently zero out that export. The rest don't block.

sugat009 (approved):
Thanks @megha1807, this addresses everything from the last round, and I re-checked it against the bikram-sambat math:

- DHIS2 export: with BS mode on it now derives a BS target-doc key, and at the default `month_start_date` it matches the writer's tag for every day across a full BS year, so the all-zeros export is fixed.
- The writer's out-of-range `toBik` is guarded, the writer's BS-tag branch has a real test, the test comment's BS facts are corrected, and the stray blank line is gone.

Approving. One non-blocking follow-up, out of scope for this PR: the DHIS2 export is still Gregorian-shaped underneath (the admin period picker sends a Gregorian month, and the reader keys off `from` while ignoring `month_start_date`), so in BS mode the period selection doesn't map cleanly onto a BS month. It's pre-existing (the same mismatch happens in Gregorian today) and lives mostly in the admin UI, so it's worth a separate follow-up to triage rather than something to hold this PR on. Nice work.

PR #11244, opened 2026-07-12
feat(#11192): support extension libs in messages

### LINKED ISSUE #11192: Support extension-libs in translations and outgoing messages
Whenever we need a new way to transform text in translations or outgoing messages, we have to add a new helper to cht-core itself.

The set of [helpers available](https://github.com/medic/cht-core/blob/20098db389458e36fc1cd38e448288505378ae11/shared-libs/message-utils/src/index.js#L338-L360) for message rendering is fixed and built into core (date, datetime, local_phone, bikram_sambat_date), so every new need means a code change and a new release. That doesn't scale and prevents projects from customizing their own message content.

This is especially limiting for outgoing messages/SMS, where the translated string is only run through Mustache plus those fixed built-in helpers.

It would be great if projects could define their own custom helpers using [extension-libs](https://docs.communityhealthtoolkit.org/building/reference/extension-libs/) (or something similar) and use them in translations and outgoing messages. It will then work consistently whether the message is produced in the app (webapp) or on the server (outgoing SMS).

Usage examples:
- Devanagari numerals in an outgoing SMS, e.g. `{{#to_devanagari}}{{patient_id}}{{/to_devanagari}}`
- Age in a death notification, calculated from the contact profile's `date of birth`, e.g. `{{contact.name}} has reported death of {{#age}}{{patient.date_of_birth}}{{/age}} years old {{patient_name}} ({{patient_id}})`

This would let configurers add new message logic without needing a cht-core change every time.

- Continue adding a new built-in helper to cht-core for each new need (the current approach, requires a core change and release every time).
- Pre-compute derived values upstream in forms and pass them into the message. It works for some cases, but doesn't work for SMS forms.

**Additional context**
Raised in #3245; related to #10846. Per discussion with @jkuester, supporting this on the server side should be technically possible but would need some careful design (e.g. how server-side code loads and refreshes extension-libs), and we'd want to avoid duplicating logic between the server and the webapp.

Note that translations already support variable substitution in some surfaces  e.g. contact-summary fields (`translate: true` + `context`). That covers substituting values you've already computed, but not running transformations on them, and it isn't available on the outgoing-message path. This request is about that transformation/custom-logic piece.

belwalshubham:
# Description

Adds project-defined `extension-libs` as synchronous Mustache section helpers for translations and outgoing messages.

This allows configurers to transform message values without adding a new built-in helper to CHT Core, for example:

```mustache
{{#to_devanagari}}{{patient_id}}{{/to_devanagari}}
```

Closes #11192.

## Architecture

```mermaid
flowchart LR
  Couch[("CouchDB<br/>extension-libs")]
  Loader["Extension-libs loader<br/>decode + atomic registry replacement"]

  Couch --> Loader

  Loader --> Webapp["Webapp cache"]
  Loader --> API["API config cache"]
  Loader --> Sentinel["Sentinel config cache"]
  Loader --> Admin["Admin preview load"]

  Webapp --> Translation["Webapp translations"]
  Webapp --> Report["Report detail messages"]
  API --> Export["Message export rendering"]
  Sentinel --> Transitions["Transitions and scheduled SMS"]
  Admin --> Preview["Message queue preview"]

  Translation --> MessageUtils["message-utils<br/>template(opts) / generate(opts)"]
  Report --> MessageUtils
  Export --> MessageUtils
  Transitions --> MessageUtils
  Preview --> MessageUtils

  MessageUtils --> Output["Consistent rendered message"]
```

## Changes

- Adds `@medic/extension-libs` for loading and decoding the `extension-libs` CouchDB document.
- Refactors `message-utils.generate` and `message-utils.template` to accept an options object containing `extensionLibs`.
- Keeps positional argument compatibility for downstream callers while migrating all CHT Core production call sites.
- Exposes synchronous function exports as Mustache section helpers, using the attachment filename without the `.js` suffix.
- Prevents extension helpers from shadowing built-in helpers or template data.
- Loads and refreshes extension-libs in API, Sentinel, and Webapp; Admin loads them for message-queue previews.
- Passes extension-libs through transitions, scheduled messages, API message exports, Webapp report details and validation messages, translations, and Admin previews.
- Logs and safely falls back to rendered content when a configured helper is invalid, throws, or returns an unsupported value.
- Treats a missing document as an empty registry, skips malformed attachments, and retains the previous valid registry after transient load failures.

## Security

Extension-libs are trusted application configuration uploaded by authorized administrators. This change does not present extension-lib execution as sandboxed.

## Verification

- `npm run lint`
- `npx tsc -p webapp/tsconfig.spec.json --noEmit`
- `npm run compile --prefix webapp`
- `npm test --workspace=@medic/message-utils` — 126 passing
- Focused transitions tests for `due_tasks` and messages — 51 passing
- Focused Webapp Karma tests for the parser, validation, format-data-record, and CHT datasource services — 59 passing
- Added API integration coverage for message-export rendering with an extension helper.
- Added Sentinel integration coverage for due-task rendering with an extension helper.
- Added Webapp WDIO coverage for translation rendering with an extension helper.

The environment-dependent API, Sentinel, and Webapp integration cases are included for execution by the relevant CI jobs.

## Notes for reviewers

- The shared loader centralizes attachment decoding and error behavior across browser and server runtimes.
- Application services still own their loading and cache lifecycle and explicitly inject the current extension-lib snapshot into `message-utils`.
- The options-object refactor follows the direction discussed in #11192 and avoids adding another positional parameter to the existing APIs.

# Code review checklist

- [x] UI/UX backwards compatible: Not applicable; this PR does not change layouts or navigation.
- [x] Documented: Extension-helper configuration, constraints, and fallback behavior are documented in `shared-libs/extension-libs/README.md`.
- [x] Tested: Unit and integration/e2e coverage has been added for the affected application paths.
- [x] Internationalised: No new user-facing static text is introduced.
- [x] Backwards compatible: Existing positional APIs remain supported and invalid or unavailable extension libraries degrade safely.
- [x] AI disclosure: OpenAI Codex was used to help inspect the codebase, implement and refactor parts of the change, add tests, and address review feedback. I reviewed the resulting changes, ran the verification listed above, and remain responsible for the contribution.

# License


belwalshubham:
@binokaryg could you review this PR ?

.github/cht-11192-e2e-verification.jpg
binokaryg, line 1: Let's not include this image in the PR. It can stay on the PR body.

webapp/src/ts/providers/translate-mustache-parser.provider.ts
```diff
@@ -0,0 +1,20 @@
...
+  private readonly sectionTemplate = /{{\s*[#^]/;
+
+  interpolate(expression, params?) {
+    if (typeof expression === 'string' && this.sectionTemplate.test(expression)) {
+      return messages.template({
+        config: {},
+        translate: value => value,
+        doc: params || {},
+        content: { message: expression },
+        extensionLibs: extensionLibs.getAll(),
```
binokaryg: Translations can silently lose content before extension-libs load.

This is populated lazily, only when something first awaits `CHTDatasourceService.isInitialized()` (`webapp/src/ts/services/cht-datasource.service.ts:31-42`). Nothing guarantees this completes before translations render at startup.

Mustache drops a section whose key is missing (the PR's own test confirms: a non-function helper renders `'value: '`). So a translation like `Hello {{#uppercase}}{{name}}{{/uppercase}}` rendered before the libs load produces `Hello ` — silent content loss, and components won't re-render when the libs later arrive.

**Suggestion:** load extension-libs eagerly during bootstrap (before/alongside translation loading, e.g. in the bootstrapper or an `APP_INITIALIZER`), or make the parser fall back to default interpolation until the registry is populated.

webapp/src/ts/services/cht-datasource.service.ts
```diff
@@ -53,30 +53,8 @@ export class CHTDatasourceService {
...
-      .catch(err => {
-        if (err.status === 404) {
-          return null;
-        }
-        throw err;
-      });
-    Object
-      .entries(doc?._attachments ?? {})
-      .forEach(([k, v]) => this.extensionLibs[k] = this.decodeExtensionLib(k, v));
+    return extensionLibs.load(this.dbService.get());
```
binokaryg: API and Sentinel watch the changes feed and reload; Admin reloads per query; the webapp loads once per session and only watches `DOC_IDS.SETTINGS` (`webapp/src/ts/services/cht-datasource.service.ts:65-71`). This staleness predates the PR for `getExtensionLib()`, but translations now depend on it too, so the gap is more visible.

**Suggestion:** add a changes subscription for `DOC_IDS.EXTENSION_LIBS`, or explicitly document the reload-on-refresh behavior.

shared-libs/message-utils/src/index.js
```diff
@@ -324,67 +324,127 @@ mustache.escape = function(value) {
...
 
-const render = function(config, template, view, locale) {
-  return mustache.render(template, Object.assign(view, {
+const getExtensionLibHelpers = (extensionLibs = {}) => Object.entries(extensionLibs)
+  .filter(([, extensionLib]) => typeof extensionLib === 'function')
+  .reduce((helpers, [fileName, extensionLib]) => {
+    const helperName = fileName.replace(/\.js$/, '');
+    helpers[helperName] = function() {
+      return function(text, renderText) {
+        return extensionLib(renderText(text));
```
binokaryg: **A throwing extension lib halts the unprotected message paths**

The helper wrapper calls the lib directly `extensionLib(renderText(text))` in `getExtensionLibHelpers`. If a configured lib throws at render time, `mustache.render` throws. `transitions/src/lib/messages.js` wraps `generate` in try/catch, but `due_tasks.js`, the API message export, admin message queue, and webapp `format-data-record` do not.

I verified this on a local instance: with a helper that throws. A due scheduled reminder stayed stuck `scheduled` with no message for 7+ minutes (multiple `due_tasks` cycles) and only recovered once the lib was fixed (baseline with a working lib generated in ~220s). Because `due_tasks.js` has no try/catch around `generate`, the throw isn't isolated to the offending message. It rejects the whole batch run (caught only at `runTasks`' top-level logger), so one bad helper silently blocks all pending scheduled messages every cycle until the lib is corrected, with no user-visible error.

**Suggestion:** wrap the helper invocation in try/catch, log the error and fall back to the untransformed rendered text, so a single misbehaving lib degrades one message instead of halting the batch.

admin/src/js/services/message-queue.js
```diff
@@ -314,9 +317,10 @@ angular.module('services').factory('MessageQueue',
           .all([
             Settings(),
             DB({ remote: true }).query('medic-admin/message_queue', params.list),
-            DB({ remote: true }).query('medic-admin/message_queue', params.count)
+            DB({ remote: true }).query('medic-admin/message_queue', params.count),
+            MessageQueueUtils.loadExtensionLibs(),
```
binokaryg: Admin message queue hard-fails if extension-libs can't load.

API and Sentinel tolerate load failures but admin's `query()` rejects entirely. A transient failure fetching the doc shouldn't take down the whole message-queue tab.

**Suggestion:** catch and fall back to `{}`, matching the behavior everywhere else.

shared-libs/message-utils/src/index.js
```diff
@@ -393,67 +453,53 @@ exports.generate = function(config, translate, doc, content, recipient, extraCon
...
   const template = exports.getMessage(content, translate, locale);
 
   if (!template) {
     return '';
   }
 
-  const context = extendedTemplateContext(doc, extraContext);
-  return render(config, template, context, locale);
+  const context = extendedTemplateContext(doc, contextExtras);
+  return render({ config, template, view: context, locale, extensionLibs });
```
binokaryg, line 605: **Extension libs can shadow template data:** in `render`, helpers are spread after the view (`{ ...view, ...helpers }`), so an extension lib whose name matches a context field replaces that field in *every* rendered template — even templates that never use the helper. For example, uploading a lib named `patient.js` breaks all existing message templates that reference patient data: `{{patient.name}}` renders empty, `{{#patient}}...{{/patient}}` becomes a lambda call instead of a data section, and `{{patient}}` interpolates the function itself. A static reserved-name list can't prevent this (the shadowable names include every doc field, which is per-project and unbounded), so mitigate instead: log a warning when a helper name collides with a view key, document that helper names shadow context fields, or — the complete fix — namespace helpers (e.g. `{{#ext.to_devanagari}}`) so helpers and data never share a lookup space.

shared-libs/message-utils/src/index.js
```diff
@@ -324,67 +324,127 @@ mustache.escape = function(value) {
...
+  result.original_message = message;
+};
+
+const getMissingContextError = extraContext => {
+  if (extraContext?.registrations?.length && !extraContext.patient) {
+    return 'messages.errors.patient.missing';
+  }
+  if (extraContext?.placeRegistrations?.length && !extraContext.place) {
+    return 'messages.errors.place.missing';
+  }
```
binokaryg: If both are missing, the missing patient will be flagged first. Is that intentional?

shared-libs/message-utils/src/index.js
```diff
@@ -393,67 +453,53 @@ exports.generate = function(config, translate, doc, content, recipient, extraCon
...
  *        provide as a context for templating. Properties: `patient` (object),
  *        `registrations` (array), and `templateContext` (object) for any
  *        unstructured context additions.
+ * @param {Object} [options.extensionLibs={}] Project extension libraries exposed as Mustache helpers.
  * @returns {String} The message.
  */
-exports.template = function(config, translate, doc, content, extraContext) {
-  extraContext = extraContext || {};
+exports.template = function(options, ...legacyArgs) {
+  const { config, translate, doc, content, extraContext, extensionLibs } = legacyArgs.length ? {
```
binokaryg: **Duplicated legacy-args normalization:** `normalizeGenerateOptions` exists for `generate`, but `template` re-implements the same logic inline. Extract one shared normalizer.

shared-libs/message-utils/src/index.js
```diff
@@ -324,67 +324,127 @@ mustache.escape = function(value) {
...
   locale = locale || moment().locale();
   return moment(date).locale(locale).format(formatString);
 };
 
-const render = function(config, template, view, locale) {
-  return mustache.render(template, Object.assign(view, {
+const getExtensionLibHelpers = (extensionLibs = {}) => Object.entries(extensionLibs)
+  .filter(([, extensionLib]) => typeof extensionLib === 'function')
+  .reduce((helpers, [fileName, extensionLib]) => {
+    const helperName = fileName.replace(/\.js$/, '');
```
binokaryg: **Helper name collisions:** `uppercase.js` and `uppercase` both map to helper `uppercase` — last one silently wins. A debug log would help configurers.

shared-libs/message-utils/src/index.js
```diff
@@ -324,67 +324,127 @@ mustache.escape = function(value) {
...
   if (!isNaN(date)) {
     date = parseInt(date, 10);
   }
   locale = locale || moment().locale();
   return moment(date).locale(locale).format(formatString);
 };
 
-const render = function(config, template, view, locale) {
-  return mustache.render(template, Object.assign(view, {
+const getExtensionLibHelpers = (extensionLibs = {}) => Object.entries(extensionLibs)
```
binokaryg: nit: `getExtensionLibHelpers(extensionLibs)` rebuilds the helper map — iterating the registry, filtering non-functions, stripping `.js` suffixes, wrapping each lib in the Mustache lambda pattern — on every call to `render`, plus again for each nested render triggered by `formatDate` / `local_phone`. When Sentinel processes a batch of scheduled messages, that's the same map being reconstructed once per message from an unchanged registry. Since `set()` returns a fresh object on each reload, object identity could serve as the cache key. Worth caching?

admin/src/js/services/message-queue.js
```diff
@@ -47,7 +49,7 @@ angular.module('services').factory('MessageQueue',
 
     const findIdByKey = (contactsByReference, key) => {
       const row = contactsByReference.rows.find((row) => row.key[1] === key);
-      return row && row.id;
+      return row && row.id; // NOSONAR: browserify-ngannotate does not support optional chaining.
```
binokaryg: misplaced comment?

shared-libs/transitions/test/unit/schedules.js
```diff
@@ -3,6 +3,18 @@ const assert = require('chai').assert;
...
+  const options = call[0];
+  return [
+    options.config,
+    options.translate,
+    options.doc,
+    options.content,
+    options.recipient,
+    options.extraContext,
+  ];
+};
```
binokaryg: It looks misplaced, and the name also seems to be misleading. It destructures the new options object into a positional array to preserve old assertions. Please consider moving it below the imports and consider a clearer name like `pickGenerateArgs`.

binokaryg (changes requested):
Thanks @belwalshubham for working on this.

I tested manually on a local instance, and it was mostly working.

I added an extension-lib `to_devanagari` to my local config and tested it across:
- outgoing messages
  - admin message queue
  - registration reply SMS
  - scheduled/reminder SMS
- validation error (muting)
- app form

Also confirmed:
- live reload of `extension-lib` works
- offline availability works
- UI-displayed translations work

Also thank you for adding the unit tests.

I did some review with Claude's assistance and find some issues, which I have flagged in the code inline.

One major fail was the silent content loss on a missing/incorrect helper:

A Mustache *section* over an unknown key renders nothing, dropping the **entire block**, not just the value, but any literal text inside it. Verified through the real `message-utils` render path:

| Template (patient_id = `48549`) | Renders |
|---|---|
| `ID: {{#to_devanagari}}{{patient_id}}{{/to_devanagari}}` (correct) | `ID: ४८५४९` |
| `ID: {{#to_devanagri}}{{patient_id}}{{/to_devanagri}}` (typo) | `ID: ` |
| `ID: {{#to_nepali}}{{patient_id}}{{/to_nepali}}` (never uploaded) | `ID: ` |
| `ID: {{to_devanagri}}` (typo as plain var) | `ID: ` |
| `ID: {{#to_nepali}}12345{{/to_nepali}}` (literal inside) | `ID: ` |

In the above example, a one-character typo in a helper name ships the id **missing entirely** (`ID: `), with **no server- or client-side warning**. This is standard Mustache behavior (the existing built-ins share it, e.g. a typo'd `{{#dat}}…{{/dat}}`), but the feature widens the surface since configurers now author helper names. Mitigation is process (preview/test configured messages before deploy. The platform won't catch it) and strengthens the case for a debug/warn log on unknown-helper / name-collision.

We also need to add a documentation on the usage of this feature.

binokaryg:
@belwalshubham I see that you have made some commits with fixes. Please reply to the individual comments before another review.

Also requesting @jkuester to have a look on this important piece of work.

belwalshubham:
> @belwalshubham I see that you have made some commits with fixes. Please reply to the individual comments before another review.
> 
> Also requesting @jkuester to have a look on this important piece of work.

@binokaryg  Sure, I’ll reply to each comment individually and address any remaining feedback before the next review.

.github/cht-11192-e2e-verification.jpg
belwalshubham: Thanks, fixed. I removed the verification image from the branch, so it is no longer included in the PR files.

webapp/src/ts/providers/translate-mustache-parser.provider.ts
belwalshubham: Good catch. I added an APP_INITIALIZER so extension libraries load during webapp bootstrap before translations are rendered. Startup still continues safely if loading fails.

webapp/src/ts/services/cht-datasource.service.ts
belwalshubham: Fixed. The webapp now watches both the settings and extension-libs documents and reloads the libraries when the extension-libs document changes.

shared-libs/message-utils/src/index.js
belwalshubham: Fixed. Extension helper execution is now isolated with try/catch. If a helper throws, the error is logged and the rendered, untransformed content is returned, so scheduled-message processing continues.

admin/src/js/services/message-queue.js
belwalshubham: Fixed. Admin now catches extension-library loading failures, logs the error, and falls back to an empty library registry instead of failing the message queue query.

shared-libs/message-utils/src/index.js
belwalshubham: Fixed. Extension helpers that conflict with top-level template data are now ignored and a warning is logged. I also documented this naming behavior.

shared-libs/message-utils/src/index.js
belwalshubham: This was not intentional. I restored the previous place-error precedence when both contexts are missing and added a regression test for it.

shared-libs/message-utils/src/index.js
belwalshubham: Fixed. generate and template now share the same normalizeOptions implementation for both the options-object and legacy positional APIs.

shared-libs/message-utils/src/index.js
belwalshubham: Fixed. Duplicate normalized helper names now produce a warning, and the first registered helper is preserved instead of being silently replaced.

shared-libs/message-utils/src/index.js
belwalshubham: Good suggestion. The generated helper map is now cached in a WeakMap using the extension-library registry identity, so it is reused until a new registry is loaded.

admin/src/js/services/message-queue.js
belwalshubham: Fixed. I removed/moved the misplaced comment so it now matches the code it describes.

shared-libs/transitions/test/unit/schedules.js
belwalshubham: Fixed. I moved the helper below the imports and renamed it to pickGenerateArgs to make its purpose clearer.

belwalshubham:
> @belwalshubham I see that you have made some commits with fixes. Please reply to the individual comments before another review.
> 
> Also requesting @jkuester to have a look on this important piece of work.

Thanks @binokaryg for the detailed review and testing. I have replied to each inline comment and pushed the corresponding fixes. This PR now be ready for another review.

shared-libs/message-utils/src/index.js
```diff
@@ -324,67 +328,193 @@ mustache.escape = function(value) {
...
+        loggedCollisions.add(helperName);
+      }
+      return helpers;
+    }
+    helpers[helperName] = helper;
+    return helpers;
+  }, {});
+};
+
+const warnForMissingSections = (template, view) => {
```
dianabarsan, line 420: This function will false-positive on the default translations. The positive+inverted pair is the standard Mustache conditional pattern and our own shipped translations use it — e.g. `messages-ne.properties`:

  ```
  messages.generic.validation.patient_id = {{#patient_id}}…{{/patient_id}}{{^patient_id}}…{{/patient_id}}
  messages.relay.chw_sms = {{#chw_sms}}{{chw_sms}}{{/chw_sms}}
  ```

Rendering these with the field absent (the exact case the `{{^…}}` branch exists for) logs `WARN: Mustache section "chw_sms" is not defined; its content will be omitted.` Sentinel will emit this for every such message on every schedule cycle in projects using conditionals, making the logs supremely noisy. 

Could you:
1. skip the warning when the template contains a matching inverted section for the same name (the author has clearly handled absence deliberately), and
2. dedup the remaining warnings (likke once per template + section name, like `loggedViewCollisions`)?

shared-libs/message-utils/test/index.js
```diff
@@ -1747,4 +1759,174 @@ describe('messageUtils', () => {
...
+      };
+
+      const result = utils.template({
+        config: {},
+        doc: {},
+        content: { message: 'value: {{#not_a_helper}}text{{/not_a_helper}}' },
+        extensionLibs,
+      });
+
+      expect(result).to.equal('value: ');
```
dianabarsan: is there an error logged here? if yes, can you please add a spy for the logged error? if no, i think this might be a concern where messages get generated incorrectly going unnoticed.

shared-libs/transitions/src/schedule/due_tasks.js
```diff
@@ -64,14 +64,15 @@ const updateScheduledTasks = (doc, context, dueDates, clearFailing=false) => {
...
+          config: config.getAll(),
+          translate: utils.translate,
           doc,
           content,
-          task.recipient,
-          context
-        );
+          recipient: task.recipient,
+          extraContext: context,
+          extensionLibs: config.getExtensionLibs?.(),
```
dianabarsan, line 74: I am noticing no unit test changes for due_tasks, even though these parms changed. Can you please add coverage for this call's arguments?

shared-libs/transitions/src/lib/messages.js
```diff
@@ -161,13 +162,14 @@ module.exports = {
     }
     // support mustache template syntax in error messages
     try {
-      error.message = messageUtils.template(
-        config.getAll(),
-        utils.translate,
+      error.message = messageUtils.template({
```
dianabarsan, line 165: I am noticing no unit test changes for messages, even though these parms changed. Can you please add coverage for this call's arguments?

webapp/src/ts/services/validation.service.ts
```diff
@@ -58,13 +59,14 @@ export class ValidationService {
     }
 
     return errors.map(error => {
```
dianabarsan, line 61: I am noticing no unit test changes for validation, even though these parms changed. Can you please add coverage for this call's arguments?

webapp/src/ts/services/format-data-record.service.ts
```diff
@@ -546,14 +547,15 @@ export class FormatDataRecordService {
 
       if (!copy.messages) {
         // backwards compatibility
-        copy.messages = messages.generate(
-          settings,
-          (key, locale?) => this.translate(settings, key, locale, null, true),
+        copy.messages = messages.generate({
```
dianabarsan, line 550: I am noticing no unit test changes for format_data_record, even though these parms changed. Can you please add coverage for this call's arguments?

shared-libs/message-utils/test/index.js
```diff
@@ -1747,4 +1759,174 @@ describe('messageUtils', () => {
...
+      const extensionLibs = {
+        'uppercase.js': () => {
+          throw error;
+        },
+      };
+
+      const result = utils.template({
+        config: {},
+        doc: { name: 'Ada' },
+        content: { message: 'Hello {{#uppercase}}{{name}}{{/uppercase}}' },
```
dianabarsan, line 1873: This feels a bit inconsistent compared to the "missing" helper, where the rendered content is simply omitted. I think both the missing helper and the throwing helper should have the same behavior.

webapp/tests/karma/ts/services/cht-datasource.service.spec.ts
```diff
@@ -452,6 +434,28 @@ describe('CHTScriptApiService service', () => {
       expect(settingsService.get.callCount).to.equal(1);
     }));
 
+    it('should reload extension libs when their document changes', async () => {
```
dianabarsan, line 437: Can you please update this test to add assertions of what a lib returned _before_ the document change?

webapp/src/ts/providers/translate-mustache-parser.provider.ts
```diff
@@ -0,0 +1,20 @@
+import { TranslateDefaultParser } from '@ngx-translate/core';
+import * as messages from '@medic/message-utils';
+import * as extensionLibs from '@medic/extension-libs';
+
+export class TranslateMustacheParserProvider extends TranslateDefaultParser {
+  private readonly sectionTemplate = /{{\s*[#^]/;
+
+  interpolate(expression, params?) {
+    if (typeof expression === 'string' && this.sectionTemplate.test(expression)) {
+      return messages.template({
```
dianabarsan: Please wrap this in a try-catch, so a malformed template doesn't crash webapp.

shared-libs/message-utils/src/index.js
```diff
@@ -324,67 +328,193 @@ mustache.escape = function(value) {
...
+      if (helpers[helperName]) {
+        logger.warn(`Extension lib "${fileName}" conflicts with another extension lib helper "${helperName}" and ` +
+          'will be ignored.');
+        return helpers;
+      }
+      helpers[helperName] = function() {
+        return function(text, renderText) {
+          const renderedText = renderText(text);
+          try {
+            return extensionLib(renderedText);
```
dianabarsan: The guard covers the helper call, but mustache stringifies the *returned* value later (`buffer += value`), outside it. A lib returning `{ toString() { throw } }` or a `Symbol` makes `generate()` throw — and since `due_tasks.js` has no try/catch, that re-opens the stuck-scheduler scenario from the
last round (the poison doc aborts the batch on every cycle). Relatedly, an `async` lib (an easy mistake to make) ships `"[object Promise]"` in the actual SMS body with no log, and `renderText(text)` sits before the `try` so nested-render throws also escape. All four close with one change: move `renderText` inside the try, and inside it
validate/coerce the return value (string-coerce, reject thenables/non-primitives) falling back to `renderedText`.

shared-libs/message-utils/src/index.js
```diff
@@ -324,67 +328,193 @@ mustache.escape = function(value) {
...
+
+const withoutViewCollisions = (extensionLibs, extensionHelpers, view) => {
+  let loggedCollisions = loggedViewCollisions.get(extensionLibs);
+  if (!loggedCollisions) {
+    loggedCollisions = new Set();
+    loggedViewCollisions.set(extensionLibs, loggedCollisions);
+  }
+
+  return Object.entries(extensionHelpers).reduce((helpers, [helperName, helper]) => {
+    if (Object.hasOwn(view, helperName)) {
```
dianabarsan: `Object.hasOwn` needs Chrome 93+, but our support floor is 90 (`browser-detector.service.ts` allows 90–106 with a warning), and it ships untranspiled to webapp and admin. On WebView 90–92 every render with libs configured throws. Use `Object.prototype.hasOwnProperty.call(view, helperName)` like the rest of the codebase — bonus: applying the
same check to the duplicate-name accumulator fixes `constructor.js`-style names being silently dropped with a misleading warning.

dianabarsan (changes requested):
Nice work here! 

I added some requests inline, and need to ask for e2e test coverage, since the new code is not touched by any e2e tests, even though it touches every app. 
we would need at least an api integration test, a sentinel integration test, and a webapp integration test. Would be very beneficial to have an admin e2e test as well, however that might be higher effort as there is no significant backbone for test helpers there.

belwalshubham:
@dianabarsan Thanks for the thorough review. I’ve pushed b401fd3d2 and replied to all 11 inline comments with the corresponding fixes.
The changes cover warning deduplication and inverted sections, guarded extension-helper fallbacks and return validation, Chrome 90 compatibility, malformed-template fallback, and the requested unit-test assertions.
I also added integration coverage for API message exports, Sentinel due-task processing, and Webapp translation rendering. The Admin loading and failure behavior remains covered by unit tests. I did not add the optional Admin e2e test because the current Admin test setup does not provide the necessary supporting helpers.
The focused unit and Karma tests, lint, TypeScript checks, and Webapp compilation passed locally. SonarCloud is also passing, while the remaining CI checks are still in progress. Please have another look when convenient.

shared-libs/message-utils/src/index.js
belwalshubham: Good catch. I updated the warning logic to skip sections that have a matching inverted section in the template. Remaining warnings are now deduplicated per template and section name, and I added tests for both cases.

shared-libs/message-utils/test/index.js
belwalshubham: Fixed. A configured export that is not a function now logs an error and safely falls back to the rendered, untransformed content. The test now verifies both the fallback and the logged error.

shared-libs/transitions/src/schedule/due_tasks.js
belwalshubham: Added unit coverage for the `messageUtils.generate` call in `due_tasks`, including the settings, document, translation callback, message content, recipient, context, and extension libraries.

shared-libs/transitions/src/lib/messages.js
belwalshubham: Added unit coverage for the `messageUtils.template` call in transitions, including the extension libraries passed through the message and validation paths.

webapp/src/ts/services/validation.service.ts
belwalshubham: Added validation-service coverage that verifies the settings, document, content, context, translation callback, and loaded extension libraries passed to the renderer.

webapp/src/ts/services/format-data-record.service.ts
belwalshubham: Added format-data-record coverage that verifies all arguments passed to `messageUtils.generate`, including the loaded extension libraries.

shared-libs/message-utils/test/index.js
belwalshubham: Agreed that configured helpers should fail consistently. Invalid non-function exports and throwing helpers now both log the problem and preserve the rendered content. A genuinely missing Mustache key still follows standard Mustache behavior.

webapp/tests/karma/ts/services/cht-datasource.service.spec.ts
belwalshubham: Updated the test to verify the library returns `old:hello` before the document change and `HELLO` after the updated library is loaded.

webapp/src/ts/providers/translate-mustache-parser.provider.ts
belwalshubham: Fixed. Extension rendering is now wrapped in `try/catch`. If the template is malformed, the error is logged and the parser falls back to the default interpolation behavior. I added a focused test for this case.

shared-libs/message-utils/src/index.js
belwalshubham: Fixed. Nested rendering now happens inside the guarded block. Helper results must be synchronous primitives and are explicitly converted to strings. Promises, thenables, objects, functions, Symbols, null, and undefined are rejected and safely fall back to the rendered content. Tests cover these edge cases.

shared-libs/message-utils/src/index.js
belwalshubham: Fixed. I replaced `Object.hasOwn` with the Chrome 90-compatible `Object.prototype.hasOwnProperty.call` approach. The internal helper maps now also use null prototypes, and I added coverage confirming that names such as `constructor.js` work correctly.

tests/e2e/default/translations/message-format.wdio-spec.js
```diff
@@ -43,4 +46,25 @@ describe('MessageFormat', () => {
...
+
+    await browser.waitUntil(async () => {
+      const labels = await commonPage.getAllButtonLabelsNames();
+      return labels.includes('REPORTS');
+    }, {
+      timeout: 5000,
+      timeoutMsg: 'Timed out waiting for extension-lib translation to render',
+    });
+
+    expect(await commonPage.getAllButtonLabelsNames()).to.include('REPORTS');
```
dianabarsan: I think this test has a very loose assertion, can you please strengthen it? Ideally use a function that completely changes a translation (not just uppercases), and check that translation specifically.

tests/e2e/default/translations/message-format.wdio-spec.js
```diff
@@ -43,4 +46,25 @@ describe('MessageFormat', () => {
     const tabsButtonLabelsNames = await commonPage.getAllButtonLabelsNames();
     expect(tabsButtonLabelsNames).to.include.members(['Reports {{thing}}', 'Tasks {thing', 'Messages {thing}']);
   });
+
+  it('should render translations with extension-libs', async () => {
```
dianabarsan, line 50: Besides this, we will need a test where messages are generated in the webapp code and we show that the extension libs are working there as well.

tests/integration/api/controllers/export-data.spec.js
```diff
@@ -19,6 +20,55 @@ describe('Export Data V2.0', () => {
...
+    before(async () => {
+      const reloadLog = await utils.waitForApiLogs(/Detected extension-libs change - reloading/);
+      await utils.saveDocs([
+        createExtensionLibDoc({ 'uppercase.js': 'module.exports = value => value.toUpperCase();' }),
+        report,
+      ]);
+      await reloadLog.promise;
+    });
+
+    const waitForTransformedExport = async (retries = 20) => {
```
dianabarsan: What are we waiting for exactly here? Can we wait for that thing to complete specifically, instead of polling the exports endpoint?

tests/integration/sentinel/schedules/due-tasks.spec.js
```diff
@@ -398,16 +417,23 @@ const translations = {
...
 describe('Due Tasks', () => {
-  before(() => utils
-    .saveDocs(contacts)
-    .then(() => utils.addTranslations('test', translations))
-    .then(() => utils.updateSettings(settings, { ignoreReload: 'sentinel' })));
+  before(async () => {
+    const reloadLog = await utils.waitForSentinelLogs(true, /Detected extension-libs change - reloading/);
+    await utils.saveDocs([
+      ...contacts,
+      createExtensionLibDoc({ 'uppercase.js': 'module.exports = value => value.toUpperCase();' }),
```
dianabarsan: I fear this overuse of uppercase, please diversify your testing extension lib transformation function. 
It would also be helpful to have multiple functions that do different things and show that all work as expected. These tests only cover the supper happy path.

tests/integration/sentinel/schedules/due-tasks.spec.js
```diff
@@ -398,16 +417,23 @@ const translations = {
...
     'TWO. Reported by {{contact.name}}. Patient {{patient_name}} ({{patient_id}}). Value {{fields.value}}',
   'messages.clinic':
-    'CLINIC. Reported by {{contact.name}}. Place {{place.name}} ({{place.place_id}}). Value {{fields.value}}'
+    'CLINIC. Reported by {{contact.name}}. Place {{place.name}} ({{place.place_id}}). Value {{fields.value}}',
+  'messages.extension': 'Extension patient: {{#uppercase}}{{patient_name}}{{/uppercase}}',
 };
 
 const ids = reports.map(report => report._id);
 
 describe('Due Tasks', () => {
```
dianabarsan, line 434: Thanks for adding the due_tasks test. To cover the other places where sentinel uses the new code, can you update one of the transitions tests that also uses scheduled messages to use extension libs for interpolation?

dianabarsan (changes requested):
Thnanks for adding the tests. We don't have full coverage yet, unfortunately. I added some requests below.

tests/e2e/default/translations/message-format.wdio-spec.js
belwalshubham: Strengthened this in `6923a8ebc`. The helper now reverses the original label and prefixes it, producing the exact expected translation `Extension label: noitagivan stropeR`. The assertion checks that specific extension-generated label and that it appears exactly once.

tests/e2e/default/translations/message-format.wdio-spec.js
belwalshubham: Added Webapp message-generation coverage in `tests/e2e/default/reports/sms-messages.wdio-spec.js`. A scheduled message is rendered through the report-details Webapp path with an extension helper, and the test asserts the exact transformed content `webapp[tneitap_eht rof rednimeR]`.

tests/integration/api/controllers/export-data.spec.js
belwalshubham: Removed the export polling in `6923a8ebc`. The setup already waits specifically for the API log confirming the extension-libs reload; after that, the test makes one export request. The CouchDB view query used by the export updates synchronously, so no additional polling is needed.

tests/integration/sentinel/schedules/due-tasks.spec.js
belwalshubham: Expanded this coverage in `6923a8ebc`. The due-task fixture now loads two different functional helpers (`decorate` and `reverse`) and verifies both transformations in one message. It also loads an invalid non-function export and verifies the safe rendered-content fallback, covering a failure path as well as the happy path.

tests/integration/sentinel/schedules/due-tasks.spec.js
belwalshubham: Added this to the registration transition's schedule-assignment integration test in `6923a8ebc`. The transition now creates scheduled tasks from content using `reverse-scheduled`, and the assertions verify the exact interpolated message `transition[2egassem]` across the existing schedule assignment and clearing scenarios.

mensch72 (commented):
**AI disclosure:** This review was prepared by GitHub Copilot, an AI coding agent, through this account with the account holder’s explicit authorization.

Two failure cases appear uncovered:

1. `extensionLibs.load()` commits whichever request finishes last. Webapp and Sentinel do not serialize change callbacks, so reload A can start, reload B can apply newer code, then A can overwrite it with older code. A generation guard or serialized reload queue would prevent this. A deterministic test can defer two `db.get()` calls and resolve the newer one first.

2. App Management returns `{}` when loading fails, although the shared loader retains its previous registry on non-404 failures. After one successful query followed by an isolated transient failure, previews therefore render without helpers while API/Sentinel retain the last-known-good version. Returning the retained registry would avoid preview divergence and silent Mustache section loss. A focused test should assert that a second failed load still passes the first successful registry to `generate()`.

shared-libs/message-utils/src/index.js
```diff
@@ -324,67 +332,218 @@ mustache.escape = function(value) {
...
+          }
+        };
+      };
+      return helpers;
+    }, Object.create(null));
+  extensionLibHelpersCache.set(extensionLibs, helpers);
+  return helpers;
+};
+
+const withoutViewCollisions = (extensionLibs, extensionHelpers, view) => {
```
dianabarsan, line 400: I believe there is no unit test for the warn-once behavior of this function. Can you please add one?

shared-libs/transitions/test/unit/messages.js
```diff
@@ -413,6 +414,32 @@ describe('messages', () => {
...
   describe('addError', () => {
+    it('passes all rendering options to messageUtils.template', () => {
+      const settings = { date_format: 'YYYY-MM-DD' };
+      const extensionLibs = { 'uppercase.js': value => value.toUpperCase() };
+      const doc = { _id: 'report' };
+      const error = { code: 'invalid', message: '{{#uppercase}}invalid{{/uppercase}}' };
+      const context = { patient: { name: 'Ada' } };
+      config.getAll.returns(settings);
+      config.getExtensionLibs.returns(extensionLibs);
+      sinon.stub(messageUtils, 'template').returns('INVALID');
```
dianabarsan: Having this template return exactly the uppercase version of the message gives the reader the impression that the extension libs helper ran, when in fact it didn't. Can you please use a different template message here so the test scope is clear to someone just skimming through?

tests/integration/sentinel/schedules/due-tasks.spec.js
```diff
@@ -398,16 +423,29 @@ const translations = {
...
-  before(() => utils
-    .saveDocs(contacts)
-    .then(() => utils.addTranslations('test', translations))
-    .then(() => utils.updateSettings(settings, { ignoreReload: 'sentinel' })));
+  before(async () => {
+    const reloadLog = await utils.waitForSentinelLogs(true, /Detected extension-libs change - reloading/);
```
dianabarsan: This line is logged when the change is detected, before it actually makes it to the cache. Is it possible that there is a race condition where the test runs faster than the extension libs are actually updated in the application cache?

tests/integration/api/controllers/export-data.spec.js
```diff
@@ -19,6 +20,49 @@ describe('Export Data V2.0', () => {
...
+        recipient: 'reporting_unit',
+        message: [{
+          locale: 'en',
+          content: 'Hello {{#uppercase}}{{fields.name}}{{/uppercase}}',
+        }],
+      }],
+    };
+
+    before(async () => {
+      const reloadLog = await utils.waitForApiLogs(/Detected extension-libs change - reloading/);
```
dianabarsan: This line is logged when the change is detected, before it actually makes it to the cache. Is it possible that there is a race condition where the test runs faster than the extension libs are actually updated in the application cache?

tests/integration/sentinel/transitions/registration.spec.js
```diff
@@ -2361,6 +2362,13 @@ describe('registration', () => {
       'messages[0].message': 'message4'
     });
 
+    const reloadLog = await utils.waitForSentinelLogs(true, /Detected extension-libs change - reloading/);
```
dianabarsan: Can you please extract this into a utils function? And make it wait for both API and Sentinel updates by default.

shared-libs/transitions/test/unit/schedules.js
```diff
@@ -6,6 +6,18 @@ const messageUtils = require('@medic/message-utils');
 const utils = require('../../src/lib/utils');
 const logger = require('@medic/logger');
 
+const pickGenerateArgs = call => {
```
dianabarsan, line 9: It seems that this doesn't end up checking that the extension libs are passed correctly.

api/tests/mocha/services/config-watcher.spec.js
```diff
@@ -190,6 +194,20 @@ describe('Configuration', () => {
...
+      it('reloads extension-libs', () => {
+        return dbWatcher.medic.args[0][0]({ id: DOC_IDS.EXTENSION_LIBS }).then(() => {
+          chai.expect(extensionLibs.load.calledOnceWithExactly(db.medic)).to.equal(true);
+        });
+      });
+
+      it('continues when extension-libs cannot be loaded', () => {
+        extensionLibs.load.rejects(new Error('failed to load'));
+
+        return dbWatcher.medic.args[0][0]({ id: DOC_IDS.EXTENSION_LIBS });
```
dianabarsan: Can you add an assertion here that setExtensionLibs was not called when extensions failed to load?

api/tests/mocha/services/config-watcher.spec.js
```diff
@@ -29,10 +30,12 @@ describe('Configuration', () => {
     sinon.spy(config, 'set');
     sinon.spy(config, 'setTranslationCache');
     sinon.spy(config, 'setTransitionsLib');
+    sinon.spy(config, 'setExtensionLibs');
```
dianabarsan, line 33: This spy is never used. I suppose it was added with a purpose?

shared-libs/message-utils/src/index.js
```diff
@@ -324,67 +332,218 @@ mustache.escape = function(value) {
...
 
-const render = function(config, template, view, locale) {
-  return mustache.render(template, Object.assign(view, {
+const getExtensionLibHelpers = (extensionLibs = EMPTY_EXTENSION_LIBS) => {
+  const cached = extensionLibHelpersCache.get(extensionLibs);
+  if (cached) {
+    return cached;
+  }
+
+  const helpers = Object.entries(extensionLibs)
```
dianabarsan, line 381: It looks like writing `{{helper}}` instead of `{{#helper}}…{{/helper}}` ships the helper's JavaScript source into the message.

Running this unit test proves this behavior. 
```js
    it('leaks the helper wrapper source when a helper is interpolated without a section', () => {
  // Documents a known hole: `{{uppercase}}` instead of `{{#uppercase}}…{{/uppercase}}` makes
  // mustache stringify the helper lambda, shipping JavaScript source in the rendered message.
  const extensionLibs = {
    'uppercase.js': value => value.toUpperCase(),
  };

  const result = utils.template({
    config: {},
    doc: { name: 'Ada' },
    content: { message: 'Hi {{uppercase}}' },
    extensionLibs,
  });

  expect(result).to.include('Hi function');
  expect(result).to.include('renderText(text)');
});

```

dianabarsan (changes requested):
Very nice! Thanks for adding all the tests. 

I added some requests inline.

api/tests/mocha/services/config-watcher.spec.js
belwalshubham: Addressed in c3142c3b7. The setExtensionLibs spy is now asserted on the startup and successful change paths, as well as asserted not called on the failure path.

api/tests/mocha/services/config-watcher.spec.js
belwalshubham: Addressed in c3142c3b7. The failed reload test now explicitly asserts that config.setExtensionLibs was not called, preserving the last-known-good registry.

tests/integration/sentinel/transitions/registration.spec.js
belwalshubham: Addressed in c3142c3b7. I extracted waitForExtensionLibsReload into tests/utils/extension-libs.js. It waits for both API and Sentinel by default and is reused by the registration, due-tasks, and API export integration tests.

tests/integration/sentinel/schedules/due-tasks.spec.js
belwalshubham: Addressed in c3142c3b7. This now waits for the shared extension-lib reload completion signal, which Sentinel emits only after its registry has been assigned. The shared watcher defaults to waiting for both API and Sentinel.

shared-libs/transitions/test/unit/schedules.js
belwalshubham: Addressed in c3142c3b7. pickGenerateArgs now includes options.extensionLibs, and all four generate-call assertions verify that the exact loaded registry is forwarded.

tests/integration/api/controllers/export-data.spec.js
belwalshubham: Addressed in c3142c3b7. The API export test now uses the shared reload-completion watcher instead of the earlier change-detection log, removing the race with the actual cache update.

shared-libs/message-utils/src/index.js
belwalshubham: Addressed in c3142c3b7. The Mustache section lambda now has a safe empty string conversion so plain {{uppercase}} cannot expose its implementation. I added a regression test asserting the output is empty at that position and the extension lib is not invoked.

shared-libs/transitions/test/unit/messages.js
belwalshubham: Addressed in c3142c3b7. I changed the configured message and the generate stub result to distinct validation-message strings, so the test clearly verifies argument forwarding without implying that the extension helper was executed.

shared-libs/message-utils/src/index.js
belwalshubham: Addressed in c3142c3b7. The collision test now renders twice with the same extension-lib registry and asserts that template data still wins, the helper is never invoked, and the conflict warning is emitted only once.

belwalshubham:
@dianabarsan  @binokaryg @mensch72 
Please let me know if anything else is needed or if any further review is required for the current PR, or if we’re good to merge it.

mensch72:
My coding agent says nothing else is needed from our side.

Issue #10707, opened 2026-03-10
[DMP 2026]: Enhance Bikram Sambat support in CHT


sugat009:
## Ticket Contents

The CHT has basic [Bikram Sambat (BS) calendar support](https://docs.communityhealthtoolkit.org/bikram-sambat/) used in [app forms](https://docs.communityhealthtoolkit.org/apps/reference/forms/app/#bikram-sambat-datepicker), but several quality-of-life issues make it incomplete for users in Nepal and other regions using this calendar system. The current date picker is minimal (no calendar view, no month navigation), the Reports tab filter still shows the Gregorian calendar even when the language is set to Nepali, there is a bug where dates convert incorrectly when no month is selected, and [targets/UHC counts](https://docs.communityhealthtoolkit.org/building/targets/) reset on Gregorian month boundaries instead of Bikram Sambat month boundaries.

This project is a collection of improvements to make Bikram Sambat a first-class calendar experience in the CHT. The work can be done incrementally (one issue at a time).

## Goals & Mid-Point Milestone

- [x] Understand the existing Bikram Sambat date picker implementation and the [`bikram-sambat`](https://github.com/nickshanks/bikram-sambat) library
- [x] Fix the bug where the date picker defaults to Baisakh (first month) when user hasn't selected a month, allowing incorrect date conversion
- [x] Upgrade the Bikram Sambat date picker with proper calendar view (month/weeks), month/year navigation, "today" button, and clear option
- [x] Integrate Bikram Sambat calendar into the Reports tab date filter (currently shows Gregorian even in Nepali mode)
- [x] Add configuration to align target/UHC count resets to Bikram Sambat month boundaries instead of Gregorian months
- [x] Write unit tests and e2e tests for all changes
- [x] **Goals achieved by mid-point milestone:** Date picker bug is fixed, date picker is fully upgraded, and Reports tab date filter uses BS calendar, with unit tests

## Expected Outcome

- A fully functional Bikram Sambat date picker with a proper calendar view, navigation controls, and parity with the Gregorian date picker UX
- The Reports tab date filter displays the Bikram Sambat calendar when the platform language is set to Nepali
- The date picker no longer silently defaults to Baisakh when no month is selected
- A new configuration option (e.g. in `app_settings`) that allows target/UHC counts to reset at the beginning of BS months instead of Gregorian months

## Implementation Details

The Bikram Sambat date picker widget lives in:
- `webapp/src/ts/` (Angular webapp components)

[...]

PR #11247, opened 2026-07-13
fix: clean up Bikram Sambat picker instances

### LINKED ISSUE #11245: Bikram Sambat enketo widget leaks body-appended picker DOM on each form open (destroy() never runs)
Noticed while reviewing #11230. The code is from the already-merged #11234, so filing separately.

## Problem
`webapp/src/js/enketo/widgets/bikram-sambat-datepicker.js` defines a `destroy()` that cleans up the body-appended picker/overlay, but it never runs. enketo-core tears forms down via `Form.resetView()` -> `widgets.reset()`, which only calls the static `Widget.globalReset()`, never an instance `destroy()`. So every time a Nepali-locale form with a Bikram Sambat date field is opened, the plugin appends a fresh `<div class="nepali-date-picker">` (and, on show, a `.nepali-date-picker-overlay`) to `<body>`, and these accumulate across form opens within a single SPA session (DOM + listener leak).

Separately, `super.destroy(element)` in that method would throw `TypeError` if it were ever invoked, since enketo-core's base `Widget` has no `destroy`.

## Reproduction (confirmed)
1. Set the app to Nepali (`ne`) locale.
2. Open a form with a BS date field (e.g. "BS Date Test"), then close it.
3. Repeat about 4 times.
4. Inspect `<body>` in DevTools.

Observed: 4 opens produced 4 leftover `<div class="nepali-date-picker" style="display:none">` in `<body>`:

four leftover nepali-date-picker divs in body after four form opens

## Suggested fix
- Wire cleanup into a teardown path enketo actually calls, or drop the dead `destroy()` (and remove/fix the `super.destroy()` call).
- Scope cleanup to this widget's own picker instead of `$('.nepali-date-picker').remove()`, which would remove every picker on the page.

Severity: non-blocking (gradual DOM/listener leak, Nepali-form-only, no crash or broken behavior).

Part of #10707.

jmtdev0:
## Summary

Fixes #11245.

Repeatedly opening the Bikram Sambat date picker leaves body-appended picker nodes behind because Enketo's `Form.resetView()` path invokes the static widget reset hook rather than each widget instance's `destroy()` method. The previous instance cleanup therefore did not run, and calling `super.destroy()` would not be valid because the base widget does not provide that method.

This PR:

- Tracks active Bikram Sambat date-picker widget instances.
- Cleans up those instances from `Widget.globalReset()`, the teardown path Enketo actually invokes.
- Removes only the picker and hidden input belonging to each widget, and removes the shared overlay once no picker remains.
- Adds a regression test covering four open/reset cycles.

## Validation

- Webapp Angular unit tests: 2,811 passed, 4 skipped.
- CHT form unit tests: 30 passed.
- Admin unit tests: 431 passed.
- API unit tests: 1,645 passed.
- Sentinel unit tests: 230 passed.
- CouchDB cluster unit tests: 6 passed.
- API integration tests: 11 passed.
- Lint and `git diff --check` passed.
- The dedicated Bikram Sambat widget regression test passed: 15 tests completed.

The shared-library suite has one pre-existing environment-level coverage gate in `shared-libs/outbound`: its 39 tests pass, but the runner reports coverage just below its configured threshold. It is unrelated to this change.

SonarCloud's two `S6582` suggestions were applied with optional chaining. Because the inherited `@medic` ESLint configuration pins Espree to ES2018, a file-scoped `ecmaVersion: 2020` override was added for this widget so the repository lint accepts the syntax. The separate `S5906` suggestion for the test assertion has also been applied with `expect(widget.$hiddenDateInput).to.be.null`.

## Evidence

The following screenshots show the picker working in the browser harness and the DOM cleanup proof after repeated form teardown cycles.

<details>
<summary>Show browser evidence</summary>

Bikram Sambat calendar picker open

Four teardown cycles passing

Console and DOM cleanup evidence

</details>

## Timeline

1. The issue identified that repeated form opens accumulated body-appended `.nepali-date-picker` nodes.
2. Inspection of Enketo's lifecycle showed that reset uses the static widget hook, so instance `destroy()` is not the reliable cleanup path.
3. The implementation was adapted to register active widgets and perform scoped cleanup during `globalReset()`.
4. A four-cycle browser regression proof and automated test confirmed zero picker and orphaned nodes after every reset.

_Target branch: `10707-dmp-2026-enhance-bikram-sambat-support-in-cht`, where the existing widget implementation was introduced._

_This PR was developed by Codex under the supervision of jmtdev0._

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -625,5 +625,11 @@ module.exports = defineConfig([
       ['@typescript-eslint/no-unused-expressions']: 'off',
       ['@typescript-eslint/no-require-imports']: 'off'
     },
+  },
+  {
+    files: ['webapp/src/js/enketo/widgets/bikram-sambat-picker-shared.js'],
+    languageOptions: {
+      ecmaVersion: 2020,
+    },
   }
 ]);
--- a/webapp/src/js/enketo/widgets/bikram-sambat-datepicker.js
+++ b/webapp/src/js/enketo/widgets/bikram-sambat-datepicker.js
@@ -7,7 +7,13 @@ const eurodigit = require( 'eurodigit' );
 const toDevanagari = eurodigit.to_non_euro.devanagari;
 const fromDevanagari = eurodigit.to_euro;
 
-const { setupNepaliDatePicker, hideDatePicker } = require('./bikram-sambat-picker-shared');
+const {
+  setupNepaliDatePicker,
+  hideDatePicker,
+  destroyDatePicker,
+} = require('./bikram-sambat-picker-shared');
+
+const activeWidgets = new Set();
 
 const NEPALI_MONTH_NAMES = [
   'बैशाख', 'जेठ', 'असार', 'साउन', 'भदौ', 'असोज', 'कार्तिक', 'मंसिर', 'पौष', 'माघ', 'फाल्गुन', 'चैत'
@@ -18,6 +24,12 @@ const MONTH_PLACEHOLDER = 'महिना';
 class Bikramsambatdatepicker extends Widget {
   static get selector() {
     return 'input[type=date]';
+  }
+
+  static globalReset() {
+    activeWidgets.forEach(widget => widget.destroy());
+    activeWidgets.clear();
+    return super.globalReset();
   }
 
   _init() {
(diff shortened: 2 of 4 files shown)


sugat009:
Closing as a duplicate. This is part of the DMP epic (#10707) and the fix landed in #11248, now merged into the epic branch.

PR #11248, opened 2026-07-14
fix(#11245): clean up Bikram Sambat picker instances

### LINKED ISSUE #11245: Bikram Sambat enketo widget leaks body-appended picker DOM on each form open (destroy() never runs)

megha1807:
# Description

Fixes #11245

This PR fixes a DOM and listener leak where opening Nepali-locale forms with a Bikram Sambat datepicker appends a new calendar container (`.nepali-date-picker`) and backdrop (`.nepali-date-picker-overlay`) to `<body>` on every form load without destroying them. 

Since Enketo Core unloads forms via static `Widget.globalReset()` rather than instance `destroy()` calls, these elements accumulated in the body across form opens.

## Changes:
- Track active widget instances in a static `Set` on the `Bikramsambatdatepicker` class.
- Implement static `globalReset()` to loop over active instances, trigger their instance `destroy()` method, and clear the set.
- Scoped element cleanup inside `destroyDatePicker` using the widget's stored `$picker` data reference, preventing all calendar pickers on the page from being broadly deleted.
- Removed the dead `super.destroy()` call in the widget to avoid runtime `TypeError` issues.
- Added Karma unit test coverage in `bikram-sambat-datepicker.spec.ts` to assert that body nodes are properly cleaned up on reset.

# Code review checklist
- [x] UI/UX backwards compatible
- [x] Readable
- [ ] Documented
- [x] Tested
- [x] Internationalised
- [x] Backwards compatible

### DIFF
--- a/webapp/src/js/enketo/widgets/bikram-sambat-datepicker.js
+++ b/webapp/src/js/enketo/widgets/bikram-sambat-datepicker.js
@@ -7,7 +7,28 @@ const eurodigit = require( 'eurodigit' );
 const toDevanagari = eurodigit.to_non_euro.devanagari;
 const fromDevanagari = eurodigit.to_euro;
 
-const { setupNepaliDatePicker, hideDatePicker } = require('./bikram-sambat-picker-shared');
+const {
+  setupNepaliDatePicker,
+  hideDatePicker,
+  destroyDatePicker,
+} = require('./bikram-sambat-picker-shared');
+
+const activeWidgets = new Set();
+let domObserver = null;
+
+const startDOMObserver = () => {
+  if (domObserver) {
+    return;
+  }
+  domObserver = new MutationObserver(() => {
+    activeWidgets.forEach(widget => {
+      if (!document.body.contains(widget.element)) {
+        widget.destroy();
+      }
+    });
+  });
+  domObserver.observe(document.body, { childList: true, subtree: true });
+};
 
 const NEPALI_MONTH_NAMES = [
   'बैशाख', 'जेठ', 'असार', 'साउन', 'भदौ', 'असोज', 'कार्तिक', 'मंसिर', 'पौष', 'माघ', 'फाल्गुन', 'चैत'
@@ -18,6 +39,16 @@ const MONTH_PLACEHOLDER = 'महिना';
 class Bikramsambatdatepicker extends Widget {
   static get selector() {
     return 'input[type=date]';
+  }
+
+  static globalReset() {
+    if (domObserver) {
+      domObserver.disconnect();
+      domObserver = null;
+    }
+    activeWidgets.forEach(widget => widget.destroy());
+    activeWidgets.clear();
+    return super.globalReset();
   }
 
   _init() {
@@ -96,13 +127,13 @@ class Bikramsambatdatepicker extends Widget {
       });
   }
 
-  destroy( element ) {
+  destroy() {
     if ( this.$hiddenDateInput ) {
-      hideDatePicker(this.$hiddenDateInput);
+      destroyDatePicker(this.$hiddenDateInput, this.$picker);
+      this.$hiddenDateInput = null;
+      this.$picker = null;
     }
-    $('.nepali-date-picker-overlay').remove();
-    $('.nepali-date-picker').remove();
-    super.destroy( element );
+    activeWidgets.delete(this);
   }
 }
 
@@ -114,6 +145,8 @@ const setupCalendarPicker = ($parent, widget) => {
   const $hiddenDateInput = $('<input type="text" class="nepali-datepicker-input">');
   $calendarBtn.after($hiddenDateInput);
   widget.$hiddenDateInput = $hiddenDateInput;
+  activeWidgets.add(widget);
+  startDOMObserver();
 
   setupNepaliDatePicker($hiddenDateInput, {
     closeOnDateSelect: false,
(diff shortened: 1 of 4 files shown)


webapp/src/js/enketo/widgets/bikram-sambat-datepicker.js
```diff
@@ -20,6 +26,12 @@ class Bikramsambatdatepicker extends Widget {
     return 'input[type=date]';
   }
 
+  static globalReset() {
```
sugat009, line 44: **issue (non-blocking): removing a repeat row leaves its calendar orphaned in the page**

A Bikram Sambat date field inside a repeat leaks the same way #11245 did, but on repeat add/remove rather than form open/close, so the cleanup added here doesn't cover it. Reproduced live:

three repeat rows and three nepali-date-picker divs in body

all rows removed, form empty, but three nepali-date-picker divs still in body

Add 3 rows and you get 3 pickers in `<body>`; remove all 3 and the form is empty but the 3 pickers stay. Since this PR is fixing the picker leak, worth covering this path too.

webapp/tests/karma/js/enketo/widgets/bikram-sambat-datepicker.spec.ts
```diff
@@ -101,6 +101,19 @@ describe('Enketo: Bikram Sambat Datepicker Widget', () => {
     expect($('.calendar-btn')).to.have.lengthOf(1);
   });
 
+  it('removes body-appended picker nodes during global widget reset', async () => {
```
sugat009: **issue (non-blocking): the multi-widget cleanup isn't exercised**

The new test calls `globalReset()` inside the loop, so only one picker is ever alive at reset time; the case with several pickers on one form (the reason for the `Set` and the shared-overlay handling) isn't covered. It does work with 3 concurrent pickers (verified live), so this is a missing test, not a bug.

sugat009 (changes requested):
Reviewed #11248 (fixes #11245). The open/close leak is correctly fixed, verified live: open a BS form (including one with several date fields) then leave the form, and the pickers and overlay are all cleaned up. Two things to handle before this lands.

webapp/src/js/enketo/widgets/bikram-sambat-datepicker.js
```diff
@@ -20,7 +41,18 @@ class Bikramsambatdatepicker extends Widget {
...
+      domObserver.disconnect();
+      domObserver = null;
+    }
+    activeWidgets.forEach(widget => widget.destroy());
+    activeWidgets.clear();
+    return super.globalReset();
+  }
+
   _init() {
+    startDOMObserver();
```
sugat009: **question (non-blocking):** `startDOMObserver()` is the first line of `_init()`, so it runs before the Nepali locale/appearance guard inside the `Language.get()` callback, and it observes `document.body` with `subtree: true`. That means for any form containing an `input[type=date]`, the observer attaches and fires on every app-wide DOM mutation for the form's lifetime, including non-Nepali deployments where no picker is ever built (`activeWidgets` stays empty). Is attaching it that early and that broad intended, or could it start only when a picker is actually created (inside the guard / in `setupCalendarPicker`)?

sugat009 (approved):
Verified the fix in the browser: repeated form open/close no longer accumulates picker nodes in the body, and removing repeat instances cleans them up (confirmed the observer path with multiple calendars, and full teardown returns the count to zero). Approving. One non-blocking performance question inline about the DOM observer's scope and timing.

webapp/src/js/enketo/widgets/bikram-sambat-datepicker.js
megha1807: Great catch, @sugat009 ! That was not intended to start so early. I've updated it to initialize startDOMObserver() inside setupCalendarPicker so that it only attaches and observes DOM mutations when a Bikram Sambat calendar is actually initialized on the page. Pushed the update!

Issue #11151, opened 2026-06-08
Add `byForm` qualifier to `Report.v1` (replaces `medic-client/reports_by_form` page query)


Hareet:
---
title: Add `byForm` qualifier to `Report.v1` (replaces `medic-client/reports_by_form` page query)
type: improvement
priority: medium
domain: data-sync
labels:
  - cht-datasource
  - migration
---

## Description

Replace `db.query('medic-client/reports_by_form', ...)` page-query call sites in cht-core with a typed `byForm(formCode)` qualifier on `Report.v1.getUuidsPage`.

Scope is the page query only. Some callers also use this view with `group: true` for counts, or `key: [form], limit: 1` for sampling — those aggregate variants are out of scope and require a different shaped operation (e.g. `Report.v1.countByForm`). They stay on `db.query` until a separate ticket addresses the aggregate API surface.

Part of the cht-datasource view→qualifier migration series; pattern established in #10973.

## Technical Context

**Components:**
- `shared-libs/cht-datasource/src/qualifier.ts` — add `FormQualifier`
- `shared-libs/cht-datasource/src/report.ts` — extend `getUuidsPage`'s declarative API
- `shared-libs/cht-datasource/src/local/report.ts` — add dispatch arm
- `shared-libs/cht-datasource/src/remote/report.ts` — add dispatch arm
- `api/src/controllers/report.js` — new route `GET /api/v1/report/by-form/:formCode`
- `api/src/routing.js` — wire the route

**Existing References:**
- `shared-libs/cht-datasource/src/local/report.ts` — pattern for the `getUuidsPage` dispatch arms (already has `isFreetextQualifier` arm)
- PR [#10083](https://github.com/medic/cht-core/pull/10083) — added the create/update qualifiers + Local/Remote dispatch. Same pattern as this ticket.

### Adapter Additions

1. **FormQualifier type:**

```typescript
export type FormQualifier = Readonly<{ form: string }>;
export const byForm = (form: string): FormQualifier => {
  if (!isString(form) || form.length === 0) {
    throw new InvalidArgumentError(`Invalid form [${JSON.stringify(form)}].`);
  }
  return { form };
};
export const isFormQualifier = (q: unknown): q is FormQualifier =>
  isRecord(q) && hasField(q, { name: 'form', type: 'string' });
```

The form code is the doc's `form` field (e.g. `pregnancy`, `anc_followup`). Do not normalise.

[...]

PR #9368, opened 2024-08-23
feat(#9193): add functionality of getting places as pages or async iterables in cht-datasource

sugat009:
# Description

Added functionality of getting places using pages or as `AsyncGenerator` using `cht-datasource` along with the addition of REST endpoint `/api/v1/person` which provides people as pages.

Issues: #9239 , #9240 , #9242

Usage:
1. Pagination API
```typescript
Place.v1.getPageByType(Qualifier.byContactType(placeType), cursor, limit); // limit and cursor are optional with defaults as 100 and "0" respectively
```
2. Async Generator API
```typescript
const getAllIterator = Place.v1.getByType(ctx)(Qualifier.byContactType('place'));
```
3. REST API endpoint
```
HTTP GET api/v1/place
```
Query Params:
- placeType - the `contact_type` of the place to fetch
- limit - default is 100
- cursor - default is 0

# Code review checklist

- [X] Readable: Concise, well named, follows the [style guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/), documented if necessary.
- [X] Documented: Configuration and user documentation on [cht-docs](https://github.com/medic/cht-docs/)
- [x] Tested: Unit and/or e2e where appropriate

# Compose URLs
If Build CI hasn't passed, these may 404:

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9193-api-endpoints-for-getting-contacts-by-type/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9193-api-endpoints-for-getting-contacts-by-type/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:9193-api-endpoints-for-getting-contacts-by-type/docker-compose/cht-couchdb-clustered.yml)

# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

PR #11249, opened 2026-07-14
feat(#10973): add phone qualifier

### LINKED ISSUE #10973: Add `byPhones` qualifier to `Contact.v1`

witash:
# Description

  Adds a phones qualifier to cht-datasource and exposes it on the two contact list endpoints.
  **No call site is migrated** — every existing `medic-client/contacts_by_phone` caller is
  untouched, so no existing read moves onto the new path. One thing does change for existing API
  users: the error returned when neither endpoint gets a qualifier. `Either query param freetext or
  type is required` becomes `Either query param freetext, type or phone is required`, on both
  `GET /api/v1/contact/uuid` and `GET /api/v1/contact`.

  ### Datasource

  - `PhonesQualifier` / `byPhones` / `isPhonesQualifier` in `qualifier.ts`, reaching consumers through the existing `export * as Qualifier` re-export.
  - Local: a dispatch arm in `getUuidsPage` and, via `getContactDocsPageFn`, in `getPage`. Both query the view with `keys` and both filter rows through `isContact`, so the id path and the doc path return the same contacts. Keys are
  deduped in a stable order — rows come back grouped in supplied-key order, which is what keeps the skip cursor meaningful between pages.
  - `queryDocsByKeys` in `local/libs/doc.ts`, the docs twin of the `queryDocIdsByKeys` helper #11327 added.
  - Remote: a `phone` param on both list endpoints.
  - Both assert helpers were renamed (`assertContactTypeFreetextPhonesQualifier`, `assertContactTypeIdsPhonesQualifier`) rather than duplicated, so no call site can still reach a two-qualifier assertion.
  - `getUuidsPageByPhones` / `getUuidsByPhones` / `getPageByPhones` / `getByPhones` on the imperative `getDatasource()` surface, next to `getPageByType` — no new method on the `Contact.v1` namespace. These have no in-repo consumer yet.

  ### API

  `phone` query param on `GET /api/v1/contact/uuid` and `GET /api/v1/contact`; no new route and no `routing.js` change. On `/uuid` it takes precedence over `type` and `freetext`; on `/contact` the precedence is `ids`, then `phone`, then
  `type`. An object-shaped `phone` (`?phone[a]=b`, which `qs.parse` parses to an object) is rejected with a 400 rather than throwing.

  ### Why the phones paths filter and the type path does not

  `contacts_by_type` keys on the contact type and `assertValidContactType` gates the request, so that path can only ever emit configured types — it needs no filter. `contacts_by_phone` keys on the phone and emits any doc whose `type` is
  in its hard-coded list, including `type: contact` docs whose `contact_type` is not in settings. So the phones arms filter explicitly, which costs the id path an `include_docs` it would not otherwise need. Emitting the type as the view
  value would let the filter run on `row.value` instead, but that is a ddoc change and a reindex — worth revisiting if the phones uuid path ever gets a high-volume caller.

  Without the filter the id path would hand back ids that `GET /api/v1/contact/{id}` then 404s, because `get` applies `isContact` too.

  ### Decisions copied from #11151 (#11327)

  - Plural array qualifier with a non-empty tuple type, verbatim matching, `Set` dedupe, and rejection of blank or whitespace-padded entries.
  - Serialising over the existing list endpoint as a comma-joined query param instead of a dedicated route.
  - The `keys`-based local query helper.

  ### Deviations from the issue

  - **Plural, not singular.** `byPhones([...])` rather than `byPhone(phone)`. The view is queried with `keys`, so a batch of numbers costs one query rather than one per number — which is what the admin message queue will need when it
  migrates.
  - **No `/api/v1/contact/by-phone/:phone`.** Query param on the existing endpoints, so paging, permissions and the OpenAPI shape stay on the established contract.
  - **The doc-returning path is extended too.** The issue scoped this to `getUuidsPage`; extending `getPage` keeps a doc lookup at one round trip and lets both paths agree on contact types.
  - **The call sites are not migrated.** Review found a behaviour change in every one of them (below), so they move to a new PR and a new issue (#11442) and this PR ships the qualifier alone.
  - **Padded input is rejected rather than passed through.** The qualifier normalises nothing, but a padded number can only ever match nothing, so `byPhones` throws instead of returning a silent empty page. Coercion and trimming remain
  the caller's job, as the issue requires.
  - **The shared assert helper is renamed rather than duplicated.** #11327 went the other way — it added `assertFreetextOrFormsQualifier` alongside `assertFreetextQualifier` and kept both. Renaming leaves no two-qualifier assertion for a
  call site to reach by mistake.

  ### Why the call sites moved out

  The acceptance criteria of #10973 was originally to migrate the existing view calls onto the datasource, but every one of those migrations carried a behaviour change that needs more analysis than this PR should carry. So this PR adds the
  qualifier, and migrating the call sites onto it moves to a follow-up. See #11442 

  | Call site | Behaviour change migration would cause |
  |---|---|
  | `webapp/.../phone-widget.js` | Online users would newly need `can_view_contacts`. That permission gates the UI and the `/api/v1/contact\|person\|place` endpoints, not the data — online users read contacts through the CouchDB proxy regardless — so the change restricts nothing while potentially breaking a custom role. |
  | `admin/.../message-queue.js` | Same cause, larger blast radius: `crfo` reaches the admin app via `can_configure`/`can_export_all` with no `can_view_contacts`, and the rejection blanks the page rather than dropping names. |
  | `validation_utils.js` `uniquePhone` | A missing field would go from failing to passing, so a config using `uniquePhone` as "must exist AND be unique" would silently stop enforcing the first half. `String()`/`.trim()` would also newly enforce uniqueness on numeric and padded values. |
  | `registration.js` `getParentByPhone` | An undefined `doc.from` sends no key at all — PouchDB drops undefined params — so the query returns every phone-bearing contact and `rows[0].doc` becomes the parent. Migrating would make that `parent_not_found`. Rows would also be newly filtered by `isContact`. |
  | `update_sent_by.js` | Rows would be newly filtered by `isContact`, where it currently reads `rows[0].doc` off the raw view. |
  | `api/.../contacts-by-phone.js` | Rows would be newly filtered by `isContact`, where it currently hydrates the raw view ids. |
  | `self_report.js`, `update_clinics.js` | These two already filter at hydration via `getWithLineage`, so `isContact` costs them nothing. They would still gain `.trim()` from the shared sender qualifier: a padded `doc.from` matches nothing today and would start matching. |

  ### Testing

  Unit coverage for the qualifier, both local adapter arms (including the `isContact` filter on the id path), the remote adapter, the assert helpers and the controller; API integration tests for `?phone=`, comma-separated, repeated-param
  and the two 400 cases on both endpoints; datasource integration tests for `getUuidsPage`/`getPage`/the generator, non-normalisation, a three-page cursor walk asserting exact ids and cursors, and a `type: contact` doc with an
  unconfigured `contact_type` that the view emits and neither phones path returns. `eslint .`, `unit-api`, `unit-shared-lib` green.

AI Disclosure: This PR was written with assistance from Claude Code.

closes #10973

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10973-phone-qualifier-2/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10973-phone-qualifier-2/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10973-phone-qualifier-2/docker-compose/cht-couchdb-clustered.yml)

# License


tests/integration/shared-libs/cht-datasource/contact.spec.js
```diff
@@ -385,6 +390,122 @@ describe('cht-datasource Contact', () => {
...
+        const twoLimit = 2;
+        const firstPage = await getUuidsPage(Qualifier.byPhones(allPhones), cursor, twoLimit);
+        const secondPage = await getUuidsPage(Qualifier.byPhones(allPhones), firstPage.cursor, twoLimit);
+
+        const allData = [ ...firstPage.data, ...secondPage.data ];
+
+        expect(allData).to.deep.equalInAnyOrder(allPhoneContactIds);
+        expect(firstPage.data.length).to.be.equal(2);
+        expect(secondPage.data.length).to.be.equal(2);
+      });
```
sugat009, line 471: **suggestion (non-blocking):** make this test cover a page break inside one phone number

The description says this walk "straddles a key boundary". The four fixture phones each have one contact, so with `limit 2` the page break falls between two numbers, not inside one. The test also does not check the ids per page or the cursors.

The committable version below checks exact ids and cursors. Rows come back grouped in the order of the requested numbers, so the order is fixed. To also cover a break inside one number, add a person that shares `patient.phone` to the fixtures at the top of the file, then update the expected ids here.

```suggestion
      it('pages across a phone number boundary without dropping or repeating a contact', async () => {
        // Rows come back grouped in the order of the requested numbers, so the split is deterministic.
        const twoLimit = 2;
        const firstPage = await getUuidsPage(Qualifier.byPhones(allPhones), cursor, twoLimit);
        const secondPage = await getUuidsPage(Qualifier.byPhones(allPhones), firstPage.cursor, twoLimit);
        const thirdPage = await getUuidsPage(Qualifier.byPhones(allPhones), secondPage.cursor, twoLimit);

        expect(firstPage.data).to.deep.equal([ contact0._id, contact1._id ]);
        expect(firstPage.cursor).to.be.equal('2');
        expect(secondPage.data).to.deep.equal([ contact2._id, patient._id ]);
        expect(secondPage.cursor).to.be.equal('4');
        expect(thirdPage.data).to.deep.equal([]);
        expect(thirdPage.cursor).to.be.equal(null);
      });
```

tests/integration/api/controllers/contact.spec.js
```diff
@@ -702,13 +759,34 @@ describe('Contact API', () => {
...
+    it('returns a page of contacts for a comma-separated list of phone numbers', async () => {
+      const responsePage = await utils.request({
+        path: endpoint,
+        qs: { phone: `${patient.phone},0000000000` }
+      });
+      const responseIds = responsePage.data.map(doc => doc._id);
+
+      expect(responseIds).to.deep.equal([patient._id]);
+      expect(responsePage.cursor).to.be.null;
+    });
```
sugat009, line 788: **suggestion (non-blocking):** add the three cases the description names for this endpoint

The description says both endpoints have the repeated-param test and the two 400 tests. `/api/v1/contact/uuid` has all five (lines 343 to 397). This endpoint has only the plain and comma-separated cases. Both endpoints share `buildPhonesQualifier`, so the three cases pass today. The suggestion adds them.

```suggestion
    it('returns a page of contacts for a comma-separated list of phone numbers', async () => {
      const responsePage = await utils.request({
        path: endpoint,
        qs: { phone: `${patient.phone},0000000000` }
      });
      const responseIds = responsePage.data.map(doc => doc._id);

      expect(responseIds).to.deep.equal([patient._id]);
      expect(responsePage.cursor).to.be.null;
    });

    it('returns a page of contacts when the phone param is repeated', async () => {
      const responsePage = await utils.request({
        path: `${endpoint}?phone=0000000000&phone=${patient.phone}`
      });
      const responseIds = responsePage.data.map(doc => doc._id);

      expect(responseIds).to.deep.equal([patient._id]);
      expect(responsePage.cursor).to.be.null;
    });

    it('throws 400 error when every entry in the phone list is empty', async () => {
      const opts = {
        path: endpoint,
        qs: { phone: ',,' }
      };

      await expect(utils.request(opts))
        .to.be.rejectedWith(`400 - {"code":400,"error":"Invalid phones [[]]."}`);
    });

    it('throws 400 error when a phone number is padded, rather than matching nothing', async () => {
      const opts = {
        path: endpoint,
        qs: { phone: `  ${patient.phone}  ` }
      };

      await expect(utils.request(opts)).to.be.rejectedWith(
        `400 - {"code":400,"error":"Invalid phones [[\\"  ${patient.phone}  \\"]]."}`
      );
    });
```

shared-libs/transitions/src/transitions/registration.js
```diff
@@ -415,9 +415,14 @@ const setPatientId = (options) => {
 };
 
 const getParentByPhone = options => {
-  return db.medic
-    .query('medic-client/contacts_by_phone', { key: options.doc.from, include_docs: true })
-    .then(result => result && result.rows && result.rows.length && result.rows[0].doc)
+  const qualifier = transitionUtils.senderPhoneQualifier(options.doc);
+  if (!qualifier) {
+    return Promise.resolve();
+  }
```
sugat009, line 421: **question:** is the new result for a missing `from` intentional?

The description says "rows consumed per site are unchanged". This site changed. Before, `key: options.doc.from` with an undefined `from` sent a view query with no key at all. PouchDB drops an undefined param (`pouchdb-abstract-mapreduce/lib/index.js:254-262`). The query returned every contact with a phone, and `rows[0].doc` became the parent. I reproduced this on a local instance. Now the lookup returns nothing and the report gets `PARENT_NOT_FOUND`.

This is a fix. It applies to public-form or xform registrations with no `doc.contact.parent` and no `parent_id` param. `uniquePhone` had the same fault (see the note there). `String()` coercion is also new here and in `update_sent_by.js`, so a numeric `from` now matches a string `phone` field.

Can you add these changes to the description and to the release notes?

shared-libs/transitions/src/transitions/utils.js
```diff
@@ -68,6 +69,14 @@ module.exports = {
     });
   },
   getUniqueId: () => idGenerator.next().value,
+  /*
+    Builds a phone qualifier for the sender of a report, or null when there is no number to look up.
+    `doc.from` is untrusted inbound SMS metadata, so it is coerced and trimmed before qualifying.
+  */
+  senderPhoneQualifier: (doc) => {
+    const phone = doc.from === undefined || doc.from === null ? '' : String(doc.from).trim();
```
sugat009, line 77: **nit:** `.trim()` is new at all four transitions

The description says the transitions "share" the trim. No site trimmed before this PR. A padded `from` matched nothing before and matches now. Small, but worth one line in the description.

shared-libs/validation/src/validation_utils.js
```diff
@@ -143,8 +143,16 @@ const validPhone = (value) => {
...
 
 const uniquePhone = async (value) => {
-  const results = await db.medic.query('medic-client/contacts_by_phone', { key: value });
-  return !results?.rows?.length;
+  // The value comes from a report field, so it can be blank or padded. A blank number has nothing to
+  // clash with, so it is unique.
+  const phone = value === undefined || value === null ? '' : String(value).trim();
+  if (!phone) {
+    return true;
+  }
```
sugat009, line 151: **note:** same fix as `registration.js`

Before, a report without the phone field sent `key: undefined`, the query returned every row, and `uniquePhone` returned `false` whenever any contact had a phone. Now it returns `true`. Correct, and worth a line in the release notes.

sugat009 (commented):
Thank you, this is a clean pilot for the series. I read the diff, ran the unit suites, and tested the endpoints on a local instance with contacts that share a phone number. The cursor walk, dedupe, verbatim matching and the 400 cases all work as described.

I found no code defect. The inline comments hold the details:

- Three statements in the description do not match the code or the tests (cursor test, `/contact` tests, "rows consumed per site are unchanged").
- Three questions on intent (`phone` precedence, the id path vs the doc path, the new `can_view_contacts` requirement).
- Two small suggestions with committable code.

One more item: all seven checklist boxes are unchecked. Please complete the checklist, and say whether AI was used.

api/src/controllers/contact.js
```diff
@@ -81,7 +104,8 @@ module.exports = {
      *     operationId: v1ContactUuidGet
      *     description: >
      *       Returns a paginated array of contact identifier strings matching the given filter criteria.
-     *       At least one of `type` or `freetext` must be provided.
+     *       At least one of `type`, `freetext`, or `phone` must be provided. `phone` cannot be combined with the
+     *       others and takes precedence over them.
```
sugat009: **nit:** the two halves of this sentence conflict

"Cannot be combined" means a 400. "Takes precedence" means the request is valid and the other params are ignored. The code does the second: `?phone=+200&type=clinic` returns persons with status 200. The `/api/v1/contact` description already says "precedence".

```suggestion
     *       At least one of `type`, `freetext`, or `phone` must be provided. When `phone` is provided, it takes
     *       precedence and `type` and `freetext` are ignored.
```

One short question: is silent precedence the intended contract for the pilot? `freetext` plus `type` combine as an AND, and `target.ts` rejects a mixed qualifier with a 400.

shared-libs/cht-datasource/src/local/contact.ts
```diff
@@ -127,14 +151,22 @@ export namespace v1 {
...
+      qualifier: ContactTypeQualifier | FreetextQualifier | PhonesQualifier,
       cursor: Nullable<string>,
       limit: number
     ): Promise<Page<string>> => {
+      if (isPhonesQualifier(qualifier)) {
+        const skip = validateCursor(cursor);
+        const keys = phoneViewKeys(qualifier);
+        const getPageFn = (limit: number, skip: number) => queryViewByPhones(keys, limit, skip);
+        return await fetchAndFilterIds(getPageFn, limit)(limit, skip);
+      }
```
sugat009, line 168: **question:** should the id path and the doc path agree on contact types?

This path returns the raw view rows. The doc path (`getPage`, lines 212 to 216) filters them through `isContact(settings)`. For a doc with `type: contact` and a `contact_type` that is not in settings, `GET /api/v1/contact/uuid?phone=X` returns the id and `GET /api/v1/contact?phone=X` returns an empty page. I reproduced this on a local instance.

The transitions split the same way. `registration` and `update_sent_by` use the doc path. `self_report` and `update_clinics` use the id path. For the same sender doc, two transitions can find the sender and two cannot. With default settings only unconfigured `contact_type` docs are affected. With a custom hierarchy that omits `clinic`, `health_center` or `district_hospital`, legacy place docs are affected too.

The reviewer note covers the doc path only. Is the split acceptable for the pilot? If yes, one sentence in the note is enough. If no, which side should move?

webapp/src/js/enketo/widgets/phone-widget.js
```diff
@@ -15,10 +15,14 @@ const isContactPhoneValid = (settings, fieldValue) => {
...
 };
 
-const getContactIdsForPhone = (phoneNumber) => window.CHTCore.DB
-  .get()
-  .query('medic-client/contacts_by_phone', { key: phoneNumber })
-  .then(results => results.rows.map(row => row.id));
+const getContactIdsForPhone = async (phoneNumber) => {
+  const datasource = await window.CHTCore.CHTDatasource.get();
+  const ids = [];
+  for await (const id of datasource.v1.contact.getUuidsByPhones([phoneNumber])) {
```
sugat009: **question:** is the new `can_view_contacts` requirement acceptable for online users?

For an online user this call goes to `GET /api/v1/contact/uuid`, which requires `can_view_contacts` (`api/src/controllers/contact.js:162`). The old view query was proxied to CouchDB with no permission check; it needed only an online session. I tested with a `national_admin` user: the new call returns 403, the old path still returns 200.

The failure is not graceful. The widget has no catch. Enketo marks the field invalid and rethrows, `form.validate()` rejects with a plain `Error`, and the contact form shows the generic "Error updating contact" toast. The user cannot save the form with any phone value. There is no test for this path.

With default settings no role is affected: the contacts tab already requires `can_view_contacts`, and db admins bypass the check. A custom online role without the permission is affected. If the new requirement is intended, one line in the release notes covers it. A short error-path test in `phone-widget.spec.ts` would help too.

api/src/controllers/contacts-by-phone.js
```diff
@@ -1,9 +1,13 @@
 const db = require('../db');
 const config = require('../config');
+const ctx = require('../services/data-context');
+const { Contact, Qualifier } = require('@medic/cht-datasource');
 const lineage = require('@medic/lineage')(Promise, db.medic);
 const phoneNumber = require('@medic/phone-number');
 const serverUtils = require('../server-utils');
 
+const getContactUuids = ctx.bind(Contact.v1.getUuids);
```
sugat009, line 9: **question:** should this endpoint have the same permission as `/api/v1/contact/uuid`?

This binding runs under the API's own data context, so the endpoint stays open to any online session. I tested with a `national_admin` user: `GET /api/v1/contacts-by-phone?phone=...` returns 200 with hydrated docs, while `GET /api/v1/contact/uuid?phone=...` returns 403. After this PR the two phone lookups have different permission rules. Is that intended?

admin/src/js/services/message-queue.js
```diff
@@ -91,18 +106,22 @@ angular.module('services').factory('MessageQueue',
...
-        .then(function(contactsByPhone) {
-          const ids = contactsByPhone.rows.map(function(row) {
-            return row.id;
-          });
-
-          return GetSummaries.getContacts(ids);
+      return DataContext
+        .then(function(dataContext) {
+          const contact = dataContext.getDatasource().v1.contact;
+          return collect(contact.getUuidsByPhones(qualifiablePhones));
```
sugat009: **note:** a 403 here fails the whole page

If the lookup is refused, the rejection reaches the controller (`admin/src/js/controllers/message-queue.js:73-76`), which sets `$scope.error` and shows an empty list with the error banner. Before, the names were resolved for any online session. Only a custom online role without `can_view_contacts` is affected; `program_officer` has the permission in default settings.

api/src/controllers/contact.js
```diff
@@ -17,6 +17,29 @@ const buildIdsQualifier = (ids) => {
...
 };
 
+// Accepts `?phone=a,b` and `?phone=a&phone=b`, like `ids` above.
+const buildPhonesQualifier = (phone) => {
+  const phonesArray = (Array.isArray(phone) ? phone : phone.split(',')).filter(Boolean);
+  if (!phonesArray.length) {
+    throw new InvalidArgumentError(`Invalid phones [${JSON.stringify(phonesArray)}].`);
+  }
+  return Qualifier.byPhones(phonesArray);
+};
```
sugat009, line 31: **suggestion (if-minor):** reject an object-shaped `phone` param with a 400

`GET /api/v1/contact/uuid?phone[a]=b` returns 500 because `phone.split` is not a function. `api/server.js:51` uses `qs.parse`, so bracket syntax becomes an object. `buildIdsQualifier` (line 13) has the same behaviour, unchanged from master, so this is fine to leave. If you fix it, the guard below is enough: `byPhones` already rejects non-string array elements.

```suggestion
const buildPhonesQualifier = (phone) => {
  if (!Array.isArray(phone) && typeof phone !== 'string') {
    throw new InvalidArgumentError(`Invalid phones [${JSON.stringify(phone)}].`);
  }
  const phonesArray = (Array.isArray(phone) ? phone : phone.split(',')).filter(Boolean);
  if (!phonesArray.length) {
    throw new InvalidArgumentError(`Invalid phones [${JSON.stringify(phonesArray)}].`);
  }
  return Qualifier.byPhones(phonesArray);
};
```

shared-libs/validation/src/validation_utils.js
```diff
@@ -143,8 +143,16 @@ const validPhone = (value) => {
...
+  const getContactUuids = dataContext.bind(Contact.v1.getUuids);
+  const generator = getContactUuids(Qualifier.byPhones([phone]));
+  const { done } = await generator.next();
+  return done;
```
sugat009, line 155: **nit:** the four transition sites use a one-item page call

They call `getUuidsPage` or `getPage` with `(qualifier, null, 1)`. This site drains one step of a generator. The cost is the same, and `executeExistsRequest` in this file uses the same style, so this is fine to leave.

shared-libs/cht-datasource/src/qualifier.ts
```diff
@@ -116,6 +116,43 @@ export const isContactTypeQualifier = (contactType: unknown): contactType is Con
...
+ */
+export type PhonesQualifier = Readonly<{ phones: [string, ...string[]] }>;
+
+/**
+ * Builds a qualifier for finding contacts with any of the given phone numbers. Duplicates are removed.
+ * @param phones the phone numbers of the contacts, each matched verbatim against the contact's `phone` field
+ * @returns the qualifier
+ * @throws InvalidArgumentError if the phone numbers are not a non-empty array of non-blank strings with no
+ * leading or trailing whitespace
+ */
```
sugat009, line 132: **nit:** say that the order of the numbers must stay the same between pages

The cursor is a skip over rows grouped in the order of the given numbers. A caller that reverses the numbers on page 2 gets duplicates and misses contacts. I reproduced this on a local instance. Today only the internal comments in `local/libs/doc.ts` say this. The `ids` qualifier has the same undocumented rule, so this is a codebase-wide gap. The shared `cursor` component in `scripts/build/generate-openapi.js:67-74` is the best single place for a note. For this PR, one sentence here is enough.

```suggestion
/**
 * Builds a qualifier for finding contacts with any of the given phone numbers. Duplicates are removed.
 * Results are grouped in the order of the given numbers. To page through results, pass the numbers in the
 * same order with each cursor.
 * @param phones the phone numbers of the contacts, each matched verbatim against the contact's `phone` field
 * @returns the qualifier
 * @throws InvalidArgumentError if the phone numbers are not a non-empty array of non-blank strings with no
 * leading or trailing whitespace
 */
```

shared-libs/cht-datasource/src/local/contact.ts
```diff
@@ -163,20 +195,19 @@ export namespace v1 {
...
+    const queryDocsByPhones = queryDocsByKeys(medicDb, 'medic-client/contacts_by_phone');
 
     return async (
-      qualifier: ContactTypeQualifier | IdsQualifier,
+      qualifier: ContactTypeQualifier | IdsQualifier | PhonesQualifier,
       cursor: Nullable<string>,
       limit: number,
     ): Promise<Page<Contact.v1.Contact>> => {
-      if (!isIdsQualifier(qualifier)) {
+      if (isContactTypeQualifier(qualifier)) {
```
sugat009, line 205: **praise:** good catch on this guard

The old guard was `!isIdsQualifier(qualifier)`. With the new qualifier it would have called `assertValidContactType` for every phones page and thrown `Invalid contact type [undefined].`. The positive check avoids that.

shared-libs/validation/src/validation_utils.js
witash: leaving as is for consistency with `executeExistsRequest`

api/src/controllers/contacts-by-phone.js
witash: Yes this was intended, and it should be unchanged from master: this endpoint queried the view directly through the API's own credentials before, and binds to the API's data context now, so its permissions are exactly what they were. Aligning it with /api/v1/contact/uuid would be a breaking change to a 3.10-era endpoint, which is out of scope here.

admin/src/js/services/message-queue.js
witash: I think this is ok.

api/src/controllers/contact.js
witash: fixed, took suggested wording

shared-libs/cht-datasource/src/qualifier.ts
witash: Added suggested comment

api/src/controllers/contact.js
witash: took the suggestion

tests/integration/api/controllers/contact.spec.js
witash: Added all three, matching the /uuid versions.

tests/integration/shared-libs/cht-datasource/contact.spec.js
witash: took the suggestion
left the patient.phone fixture since it touches every other describe in the file, and the added assertions are enough

webapp/src/js/enketo/widgets/phone-widget.js
witash: This is a good catch because it is a change in behavior that we should be cautious about...
The risk is that because roles are configureable, we can't know if any deployment is relying on a user without `can_view_contacts`, so if we release this, we could potentially break some workflows that previously working fine.
You could say that these users were misconfigured and that all the replaced call sites should not allow users without `can_view_contacts` to access them.
But it looks like there's not a lot of precedent for that; `can_view_contacts` is not enforced very widely, so it's very possible that deployments do have online users without this permission.
We can address these permissions issue in a separate issue, for now I have reverted the related changes so that this and the similar issue in message-queue still use the old `contacts_by_phone` view.

admin/src/js/services/message-queue.js
witash: actually, see above comment; I think this change in behavior doesn't belong in this PR and have reverted it.

shared-libs/transitions/src/transitions/registration.js
witash: So this is technically a breaking change.
The reason I think this is ok to leave in this branch, and the permissions changes are not, is that the current behavior in this case is much less likely to be intentional; I can't imagine a scenario where anyone would genuinely want to set a documents parent to the first contact in the results, which is essentially random.
If any deployments actually are relying on this behavior, they should fix it; the new behavior, rejecting with PARENT_NOT_FOUND makes the issue visible and easy to fix.

It does need to be called out in the release notes.

shared-libs/cht-datasource/src/local/contact.ts
```diff
@@ -127,14 +151,22 @@ export namespace v1 {
...
     const getOfflineFreetextQueryPageFn = getOfflineFreetextQueryFn(medicDb);
     const promisedUseNouveau = useNouveauIndexes(medicDb);
 
     return async (
-      qualifier: ContactTypeQualifier | FreetextQualifier,
+      qualifier: ContactTypeQualifier | FreetextQualifier | PhonesQualifier,
       cursor: Nullable<string>,
       limit: number
     ): Promise<Page<string>> => {
+      if (isPhonesQualifier(qualifier)) {
```
sugat009, line 163: **question:** the two paths do not agree on the contact type (from my last review)

Still open from my last review. `getUuidsPage` (this line) returns the raw view rows. `getPage` removes rows with the `isContact(settings)` filter (line 214). For a document with an unknown `contact_type`, `GET /api/v1/contact/uuid` returns the identifier. `GET /api/v1/contact` returns an empty page. Two transitions use each path. Thus they do not agree for the same sender.

The `isContact` filter is not new. It comes from #11174. The phone qualifier is the first to send the same input through both paths. Default settings drop nothing. Is this difference correct for the pilot? If yes, add one sentence to the reviewer note. If no, tell me which path must change, maybe in a separate ticket.

sugat009 (commented):
Thank you. The reverts are correct, and both front-end files now match master. You applied the five suggestions, and the unit tests pass (cht-datasource 1267, transitions 893, validation 49, API contact controllers 35). The revert removes the permission problem.

Three items remain. No item is a code defect.

**1. The PR does not obey the acceptance criterion of #10973.**

The revert put `message-queue.js` and `phone-widget.js` back on the `contacts_by_phone` view. The search in the issue now finds two matches, not zero. The revert is correct, because of the permission problem we discussed. But the PR still says `closes #10973`. Please change the criterion and record the reason, or open a follow-up issue for the two sites. The issue owner can decide.

**2. Please make the description agree with the code.**

Four statements are now wrong:

- Callers row for `message-queue.js`: "one `getUuidsByPhones` call for all recipients". The code now uses the view.
- Callers row for `phone-widget.js`: "`window.CHTCore.CHTDatasource` (newly exposed via integration-api)". The code now uses the view, and you reverted the integration-api and cht-form changes.
- Deviations item: "Eight call sites, not six". The code now moves six sites.
- Callers note: "Rows consumed per site are unchanged". The two front-end sites no longer drain a generator. They read the rows from one view query.

Three more points:

- The `from` change is a breaking change. A report with no sender no longer matches every contact, and `uniquePhone` does the same. You agreed to add a release note. Please replace the "Rows consumed per site are unchanged" sentence, add the release note, and mention the new `.trim()`.
- Please add one sentence for the reason. The two sites stay on the view, because the datasource path adds the `can_view_contacts` check and the `isContact` filter. Also say the four `*Phones` methods stay for the later work.
- The Testing section says the cursor test "straddles a key boundary". The test does not. Please change the sentence.

**3. The checklist is not complete.**

All seven boxes are empty, and there is no AI disclosure. Please complete the checklist and say if you used AI.

shared-libs/validation/src/validation_utils.js
witash: reverted the related changes

shared-libs/cht-datasource/src/local/contact.ts
witash: Both paths should have the same behavior about returning non-contact documents (that is, don't return them).
This means that some of the transitions that migrated from the view to datasource would also change behavior.
Ive reverted all the view migrations from this branch and added isContact to the ids path.

shared-libs/cht-datasource/src/local/contact.ts
witash: see comment below

shared-libs/cht-datasource/src/local/contact.ts
sugat009, line 162: **question:** did a commit go missing for the `isContact` filter on the id path?

Your reply says "added isContact to the ids path". The description says it three times:

- line 31: "both filter rows through `isContact`, so the id path and the doc path return the same contacts"
- line 46: the filter "costs the id path an `include_docs` it would not otherwise need"
- line 82: "Unit coverage for ... both local adapter arms (including the `isContact` filter on the id path)"

I cannot find this in the code. This arm calls `fetchAndFilterIds`. Its only filter is an id dedupe (`local/libs/doc.ts:210-229`). `queryDocIdsByKeys` sets `include_docs: false` (`local/libs/doc.ts:151-158`), so there is no document to test. `local/contact.ts` is the same as the round-2 head. No test asserts the filter.

I checked twelve other explanations. All of them failed: the helper, the controller, the remote adapter, the curried wrapper, the view map, a stale dev server, a stale `dist`, a later commit, and the reading that "ids path" means the `byIds` arm of `getPage`. That arm already filtered on master.

### How to reproduce

Two reviewers ran these commands on a local instance. Quote each URL, because `zsh` treats `?` as a pattern. Write `+` as `%2B`, because a raw `+` becomes a space.

Step 1. Add one document. Its `type` is `contact`. Its `contact_type` is not in the settings. Add a normal person for comparison. If the answer says `conflict`, the documents exist already, so run step 7 first.

```bash
curl -s -u medic:password -X POST http://localhost:5984/medic/_bulk_docs \
  -H 'Content-Type: application/json' \
  -d '{"docs":[
    {"_id":"check-phone-split","type":"contact","contact_type":"not_configured",
     "name":"Check Split","phone":"+911","reported_date":1700000000000},
    {"_id":"check-phone-normal","type":"person",
     "name":"Check Normal","phone":"+922","reported_date":1700000000000}
  ]}'
```

Step 2. Ask the view for that phone number. The view holds the document, so the view is not the filter.

```bash
curl -s -u medic:password -X POST \
  "http://localhost:5984/medic/_design/medic-client/_view/contacts_by_phone" \
  -H 'Content-Type: application/json' -d '{"keys":["+911"]}'
```

```
{"total_rows":3,"offset":1,"rows":[
{"id":"check-phone-split","key":"+911","value":null}
]}
```

Step 3. Ask the id endpoint. It returns the document.

```bash
curl -s -u medic:password "http://localhost:5988/api/v1/contact/uuid?phone=%2B911"
```

```
{"data":["check-phone-split"],"cursor":null}
```

Step 4. Ask the doc endpoint for the same number. It returns nothing.

```bash
curl -s -u medic:password "http://localhost:5988/api/v1/contact?phone=%2B911"
```

```
{"data":[],"cursor":null}
```

Step 5. Take the id from step 3. Ask for that contact. The answer is 404.

```bash
curl -s -o /dev/null -w 'HTTP %{http_code}\n' -u medic:password \
  "http://localhost:5988/api/v1/contact/check-phone-split"
```

```
HTTP 404
```

Step 6. Repeat steps 3 to 5 for the normal person. All three calls agree, and the last one returns 200.

```bash
curl -s -u medic:password "http://localhost:5988/api/v1/contact/uuid?phone=%2B922"
curl -s -u medic:password "http://localhost:5988/api/v1/contact?phone=%2B922"
curl -s -o /dev/null -w 'HTTP %{http_code}\n' -u medic:password \
  "http://localhost:5988/api/v1/contact/check-phone-normal"
```

```
{"data":["check-phone-normal"],"cursor":null}
{"data":[{"_id":"check-phone-normal", ... }],"cursor":null}
HTTP 200
```

Step 7. Remove the two documents.

```bash
for id in check-phone-split check-phone-normal; do
  rev=$(curl -s -u medic:password "http://localhost:5984/medic/$id" \
    | python3 -c 'import sys,json;print(json.load(sys.stdin)["_rev"])')
  curl -s -u medic:password -X DELETE "http://localhost:5984/medic/$id?rev=$rev"; echo
done
```

### Why this matters

Step 3 and step 5 together are the problem. The id endpoint gives a caller an id. The contact endpoint then says that id does not exist. `GET /api/v1/contact/{id}` applies `isContact` too, so it drops the same document.

A caller reads a page of ids, and then reads each contact by id. That is the pattern `Contact.v1.getUuids` serves. With this document the caller gets a hole, and it gets no reason for the hole.

If a commit is missing, please push it. If you decided against the filter, please tell me, and we can open the design question again. To filter here you must read the documents, which is the `include_docs` cost that line 46 already claims the id path pays.

I do not think this should merge while the description says the filter is there.

api/src/controllers/contact.js
```diff
@@ -129,16 +165,12 @@ module.exports = {
...
-      }
-      const qualifier = {};
-      if (req.query.freetext) {
-        Object.assign(qualifier, Qualifier.byFreetext(req.query.freetext));
-      }
-      if (req.query.type) {
-        Object.assign(qualifier, Qualifier.byContactType(req.query.type));
+      if (!req.query.freetext && !req.query.type && !req.query.phone) {
+        return serverUtils.error(
+          { status: 400, message: 'Either query param freetext, type or phone is required' }, req, res
```
sugat009, line 170: **nit:** this message changed for existing users

Master returned "Either query param freetext or type is required". This line returns "Either query param freetext, type or phone is required". Line 244 does the same for `/contact`. The new text is correct. Only line 26 of the description needs a change, because it says the PR carries no behaviour change.

sugat009 (commented):
Thank you. The revert is complete and clean. This round reverted 15 files, and each one is now the same as master. All eight call sites use the view again. The PR now changes 16 files. The revert left no orphan code. The unit tests pass: cht-datasource 1267, transitions 882, validation 132, and 35 across the two API contact controller specs. `eslint` is clean. You also closed every code item from the last two rounds: the object-shaped `phone` guard, the OpenAPI sentence, the JSDoc order note, the cursor walk test, and the three `/contact` tests. The new description is much more honest, and the AI disclosure is there.

One item needs an answer before this merges. Five smaller items are about the description and the body.

**1. The id path does not filter, but the PR says it does.** See the comment on `shared-libs/cht-datasource/src/local/contact.ts:163`. It holds the commands to reproduce this.

**2. "No behaviour change" is not exact.** Line 26 says the PR carries no behaviour change. Two error messages changed for existing API users. See the comment on `api/src/controllers/contact.js:170`.

**3. Two rows in the behaviour table are wrong.**

- Line 76 says `contacts-by-phone.js` was "Newly filtered by `isContact`". The reverted code used `Contact.v1.getUuids`, which is the unfiltered id path. That change never happened.
- Line 78 says `self_report.js` and `update_clinics.js` are the only two with unchanged behaviour. Both gained `.trim()` through `senderPhoneQualifier`. A padded `doc.from` matched nothing before, and it matches now. Line 78 also contradicts line 62.

**4. The source in line 53 is backwards.** Line 53 lists "Renaming the shared assert helper instead of adding a parallel one" as a decision copied from #11327. #11327 did the opposite. It added `assertFreetextOrFormsQualifier` next to `assertFreetextQualifier`, and both still exist. This PR renamed. The rename is complete, so only the source needs a change.

**5. The body holds the template twice.** There are two checklists (lines 90 to 98 empty, lines 110 to 118 complete), two License sections, two `<!-- ISSUE NUMBER -->` markers, and two compose blocks. The empty checklist shows first, so the page still looks incomplete. Please delete the first copy. Line 67 also has a typo: "accpetance".

**6. `closes #10973` no longer fits.** The body says no call site is migrated. It also explains why the acceptance criteria are not met. The search in the issue now finds eight matches. Please change `closes` to `refs`, and let the follow-up PR close the issue.

sugat009:
@witash you pushed `0387ccd6` after my last review. That commit looks like it answers my question about the `isContact` filter on the id path. Is the PR ready for another review? Please tell me when you want me to look again.

witash:
> @witash you pushed 0387ccd6 after my last review. That commit looks like it answers my question about the isContact filter on the id path. Is the PR ready for another review? Please tell me when you want me to look again.

Hi @sugat009 after changes are requested on a PR, I usually don't consider it ready for re-review until 're-requesting' a review, even if commits have been pushed in the meantime. Then you can push intermediate commits when they are ready, re-run ci, or leave a PR alone for a while and come back to it later, without needing it to be ready for review then whole time.

Anyway, its ready for review now; I pushed the missing isContact commit, edited the issue so that the accpetance criteria do not include migrate all the callers, and created the new issue for migrating the callers. #11442 , so its not ready for review again

shared-libs/cht-datasource/src/local/contact.ts
witash: yes a commit was missing, I left this PR for a while and came back to it

shared-libs/cht-datasource/src/local/contact.ts
```diff
@@ -127,14 +150,30 @@ export namespace v1 {
...
+        // The phone view emits any doc with a `phone`, so the rows are filtered through `isContact` to
+        // return the same contacts as the doc-returning path. That needs the docs, unlike the type view,
+        // whose key is itself a configured contact type.
+        const skip = validateCursor(cursor);
+        const keys = phoneViewKeys(qualifier);
+        const getPageFn = (limit: number, skip: number) => queryDocsByPhones(keys, limit, skip);
+        const page = await fetchAndFilter(
```
sugat009, line 169: **note:** a pre-existing fault in `fetchAndFilter` reaches these lines

The fix on these lines is right. I confirmed it on a local instance. Both paths now return the same contacts.

While I tested it, I found a fault in `fetchAndFilter`. It can lose a contact and still report the end of the results.

Put five non-contacts in front of three contacts on one number. Then ask for pages of two.

```
GET /api/v1/contact/uuid?phone=%2B955&limit=2
{"data":["b1","b2"],"cursor":null}
```

The third contact never arrives. The cursor is `null`, so the caller believes the list is complete. Limits 1, 3, 4 and 5 all return the three contacts. Only limit 2 loses one. The doc endpoint loses the same row.

This fault is not yours. Three checks say so.

- The PR does not touch `fetchAndFilter`. The only addition to `local/libs/doc.ts` is `queryDocsByKeys`.
- `GET /api/v1/contact?ids=<five non-contacts>,<three contacts>&limit=2` loses a contact too. This PR never touches that path.
- The logic is unchanged since 2024, in #9368.

The doc path already had this fault. Your fix gives the uuid path the same helper, so both paths lose the same row. That is the parity you wanted.

The cause sits in the cursor arithmetic of `fetchAndFilter`. #11449 tracks it, because the fix belongs in that helper and it also reaches the `ids` qualifier and offline freetext search. Nothing is needed from you here.

One smaller point on the same lines. The old arm used `fetchAndFilterIds`, which removed duplicate ids with a `Set`. `fetchAndFilter` does not. `phoneViewKeys` removes duplicate keys, but CouchDB compares keys by collation, not by characters. Five ignorable characters make two different strings match one row, for example a zero width space or a left to right mark. The id then arrives twice. `getPage` already behaved this way, so your change makes the two paths agree. The trigger is narrow. I raise it for the record only.

api/src/controllers/contact.js
```diff
@@ -92,16 +120,24 @@ module.exports = {
...
      *         schema:
      *           type: string
      *           minLength: 3
      *         description: >
      *           A search term for filtering contacts. Must be at least 3 characters and not contain whitespace.
-     *           Required if `type` is not provided and may be combined with `type`.
+     *           Required if `type` and `phone` are not provided and may be combined with `type`.
+     *       - in: query
+     *         name: phone
+     *         x-since: 5.3.0
```
sugat009, line 135: **nit:** `x-since` names a version that already shipped

This line says `x-since: 5.3.0`. The repository holds tags `5.3.0` and `5.3.1`, and master reads `5.3.0` in `package.json`. So the `phone` parameter cannot appear in 5.3.0.

The `phone` parameter on `/api/v1/contact` has no `x-since` of its own (line 209). It therefore inherits `5.3.0` from the operation on line 190, which is wrong for the same reason.

Please give both parameters the next release number.

shared-libs/cht-datasource/src/remote/contact.ts
```diff
@@ -44,10 +46,13 @@ export namespace v1 {
 
   /** @internal */
   export const getUuidsPage = (remoteContext: RemoteDataContext) => (
-    qualifier: ContactTypeQualifier | FreetextQualifier,
+    qualifier: ContactTypeQualifier | FreetextQualifier | PhonesQualifier,
     cursor: Nullable<string>,
     limit: number
   ): Promise<Page<string>> => {
+    const phoneParams: Record<string, string> = isPhonesQualifier(qualifier)
+      ? { phone: qualifier.phones.join(',') }
```
sugat009, line 54: **question:** should `byPhones` reject a comma?

This line joins the numbers with a comma. `api/src/controllers/contact.js:26` then splits the value on a comma. So one number that holds a comma arrives at the server as two numbers.

The local adapter matches that same number as one whole string. One qualifier therefore gives two different answers, and the remote call can return contacts that the caller never asked for.

`byIds` joins the same way on line 80. Ids are generated, so they hold no commas. A `phone` field is free text that a user types.

A repeated `phone` parameter would close the gap. A check in `byPhones` would close it too. Which do you prefer?

sugat009 (approved):
Approved. The blocker from my last review is fixed. I proved it on a local instance.

Commit `0387ccd6` changes the phones arm of `getUuidsPage`. It now reads the documents and filters them through `isContact`. Both paths return the same contacts.

I checked the fix three ways.

- The document that failed before now returns nothing on both endpoints. Before, the id path returned an id, and that id gave 404.
- I paged nine documents on one number. Contacts and non-contacts alternate by id. At limits 1, 2 and 3, both endpoints returned the same five contacts. No gaps. No repeats.
- The new unit tests fail against the old code. Six tests fail when I restore the round-3 file.

Everything else from the last three rounds is closed.

- All five description items are fixed. You rewrote the description rather than patch it.
- The body holds one checklist. All seven boxes are ticked. The AI disclosure is there.
- You rescoped issue #10973. All five new criteria are met, so `closes #10973` is correct now.
- You raised #11442 for the call sites. It names all eight sites and the decision each one needs.
- All 17 review threads are resolved.
- CI passes: 53 checks, no failures.
- Tests pass: cht-datasource 1268, API contact controller 25, transitions 885, validation 132. `eslint` and `tsc` are clean.

Four notes follow. None of them blocks the merge.

1. **A shared helper can drop a contact.** See the comment on `shared-libs/cht-datasource/src/local/contact.ts:169`. This fault is not yours. Your fix routes a second path onto it.
2. **`x-since` names a released version.** See the comment on `api/src/controllers/contact.js:135`.
3. **A comma inside a phone number splits it.** See the comment on `shared-libs/cht-datasource/src/remote/contact.ts:54`.
4. Three small items. The four new `getDatasource()` methods omit the paging order warning that `byPhones` carries. No test covers `ids` against `phone` on `/api/v1/contact`. Line 27 of the body quotes the `/uuid` error message, but `/contact` had a different one, `Either query param ids or type is required`.

Issue #11252, opened 2026-07-15
Bikram Sambat datepicker offers non-existent days (e.g. मंसिर ३० २०८२) and throws Invalid day value


sugat009:
### Describe the bug
In Nepali locale, the Bikram Sambat datepicker (reports filter and enketo form widget) draws its day cells from the bundled `nepali-date-picker`'s own month table, but converts the selection with `bikram-sambat`. Those tables disagree for 2082/2083, so the picker renders days that do not exist. Selecting one calls `toGreg_text(...)`, which throws an unhandled `Invalid day value` and breaks the filter (and the form widget).

### To reproduce (verified)
1. Nepali locale, Reports tab, open the date filter, open the BS picker.
2. Navigate to **मंसिर २०८२**. The picker shows day **३०**; that month has 29 days.
3. Click **३०** -> `Invalid day value` in the console, filter does not apply.

Also in 2082: जेठ ३२ and साउन ३२ (real 31), माघ ३० (real 29). All of BS 2082 is in the past, so it is fully selectable today. The 2083 cases (मंसिर, माघ) are near-future and open up later this year, and are reachable now in form fields that are not capped at today.

### Root cause
Two independent Bikram Sambat calendars that must agree and do not. `nepali-date-picker`'s table is stale; `bikram-sambat` is correct (verified against a real 2082/2083 calendar). The 2082 crash predates recent changes; the 2083 case was activated by #11154 bumping `bikram-sambat` to 1.8.1 (which includes medic/bikram-sambat#31), making the converter correct while the plugin stayed wrong.

### Affected code
- Reports filter: `webapp/src/ts/components/filters/date-filter/date-filter.component.ts` (`onDateSelect` -> `toGreg_text`)
- Enketo widget via `webapp/src/js/enketo/widgets/bikram-sambat-picker-shared.js`

Currently on the epic branch `10707-dmp-2026-enhance-bikram-sambat-support-in-cht` (merged via #11230), not on master.

### Proposed resolution
1. **Immediate:** guard the conversion (validate the selected day against `bikram-sambat.daysInMonth`, or a `safeToGreg` helper in the shared module) so no selection can throw.
2. **Real fix:** migrate the shared picker off `nepali-date-picker` to the `bikram-sambat`-backed calendar (**blocked by** medic/bikram-sambat#35).

This is the gate to clear before the epic merges to master.

### Related
[...]

PR #11251, opened 2026-07-14
fix(#11241): ensure Bikram Sambat date conversion uses local date string format

binokaryg:
# Description

When the webapp is used in Nepali (Bikram Sambat calendar), the displayed date rolled over to the next day several hours before local midnight. In Nepal (UTC+5:45) it changed at 18:15 (6:15 PM) instead of midnight.

**Root cause:** `FormatDateService` passed a `moment` object to `bikram-sambat`, whose `toBik_text`/`toBik` expect an ISO date string (`YYYY-MM-DD`). The library's `Date.parse` then consumed the full timestamp, so an evening time-of-day rolled the day over early in timezones ahead of UTC. Additionally, with the `ne` locale active, `moment.format` emits Devanagari numerals, which `Date.parse` cannot read, causing the "Date outside supported range" crash.

**Fix:** format the moment to a local `YYYY-MM-DD` string before passing it to the library, and force the `en` locale on a clone so the numerals stay ASCII (the `datetime` time-portion remains localised). This pins the conversion to the local calendar day so the date changes at local midnight.

**Tests:** updated the existing `date/datetime/dayMonth` specs to assert the library now receives the local date string, and added two regression tests in `format-date.service.spec.ts`
- one proving time-of-day independence (start-of-day and end-of-day map to the same date),
- and another proving ASCII digits are passed even when the locale renders Devanagari.

Both new tests were verified to fail on the unfixed code and pass with the fix.

Fixes: #11241

## AI disclosure

Per the [AI contribution guidelines](https://docs.communityhealthtoolkit.org/community/contributing/ai-guidelines/):

- **Tool:** Claude Code (Anthropic).
- **How it was used:** investigating the bug report and understanding the relevant code, identifying the root cause (passing a `moment` to `bikram-sambat` instead of an ISO date string, plus the `ne`-locale Devanagari-numeral crash), implementing the fix, and writing the unit tests and code comments.

All changes were reviewed and tested by me; the new tests were verified to fail without the fix and pass with it.

# Code review checklist

- [ ] UI/UX backwards compatible: Test it works for the new design (enabled by default). And test it works in the old design, enable `can_view_old_navigation` permission to see the old design. Test it has appropriate design for RTL languages. 
- [ ] Readable: Concise, well named, follows the [style guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/)
- [ ] Documented: Configuration and user documentation on [cht-docs](https://github.com/medic/cht-docs/)
[...]

PR #11255, opened 2026-07-16
test(#11246): harden Enketo Bikram Sambat datepicker widget and date conversions

### LINKED ISSUE #11246: Bikram Sambat: harden feature test coverage (widget, reports filter, conversion, a11y/security)

### B. Reports date filter

### C. Conversion and display

### Biggest structural gap

### Suggested priority

Part of #10707.

megha1807:
# Description

Hardens test coverage for the Enketo Bikram Sambat datepicker widget and date conversions, addressing all unit and frontend-integration items under Sections A, C, and D of issue medic#11246.

## Changes:
- **FormatDate Service Safety Guard**: Added safety checks in `FormatDateService` to gracefully return `'Invalid date'` for free-text search queries (e.g. `'असार'`), preventing crashes during search filtering.
- **Enketo Widget Unit Tests**: Hardened coverage for keyboard entries, clearing behavior, placeholders, plugin fallback, and locale gates.
- **Date Conversion Unit Tests**: Added real library date/datetime conversions, timezone day boundaries (midnight, end-of-day, negative offsets), DST transitions, and reverse BS->Greg `toGreg_text` round-tripping.
- **Integration Tests (cht-form layer)**: Added a new WDIO test suite verifying native input hiding, pre-population of Devanagari numerals on load, focus restoration, dialog ARIA attributes, key-traversal focus trapping, and CSP/XSS link sanitization.

This is PR 1 of 2. PR 2 will address the Reports Date Filter sidebar (Section B).

## Issue Number

Fixes #11246

# Code review checklist

### DIFF
--- a/webapp/tests/karma/ts/services/format-date.service.spec.ts
+++ b/webapp/tests/karma/ts/services/format-date.service.spec.ts
@@ -451,51 +451,61 @@ describe('FormatDate service', () => {
 
     it('handles day-boundaries (midnight) correctly', () => {
       const date = moment('2024-06-29T00:00:00.000');
-      expect(service.date(date)).to.equal(BikramSambat.toBik_text(date));
+      expect(service.date(date)).to.equal('१५ असार २०८१');
     });
 
     it('handles day-boundaries (end-of-day) correctly', () => {
       const date = moment('2024-06-29T23:59:59.999');
-      expect(service.date(date)).to.equal(BikramSambat.toBik_text(date));
+      expect(service.date(date)).to.equal('१५ असार २०८१');
     });
 
     it('correctly handles conversion across timezones and negative offsets', () => {
       const dateInUTC = moment.utc('2024-06-29T05:00:00Z');
       const localDate = dateInUTC.local();
       
-      const bkDate = BikramSambat.toBik(localDate);
-      const serviceFormatted = service.date(localDate);
-      
-      const dev = BikramSambat.toDev(bkDate.year, bkDate.month, bkDate.day);
-      expect(serviceFormatted).to.equal(`${dev.day} ${dev.month} ${dev.year}`);
+      const localDay = localDate.date();
+      const expectedText = localDay === 29 ? '१५ असार २०८१' : '१४ असार २०८१';
+      expect(service.date(localDate)).to.equal(expectedText);
     });
 
     it('correctly handles conversion across Daylight Saving Time (DST) boundaries', () => {
       const beforeDST = moment('2024-03-10T01:59:59'); // Standard Time
       const afterDST = moment('2024-03-10T03:00:00'); // DST (02:00:00 doesn't exist)
       
-      expect(service.date(beforeDST)).to.be.a('string');
-      expect(service.date(afterDST)).to.be.a('string');
+      expect(service.date(beforeDST)).to.equal('२७ फाल्गुन २०८०');
+      expect(service.date(afterDST)).to.equal('२७ फाल्गुन २०८०');
     });
 
-    it('toGreg_text reverse conversion round-trips correctly at month/year boundaries', () => {
-      const gregStr = BikramSambat.toGreg_text('2080', '12', '30');
-      expect(gregStr).to.be.a('string');
-      
-      const bik = BikramSambat.toBik(moment(gregStr));
-      expect(Number(bik.year)).to.equal(2080);
-      expect(Number(bik.month)).to.equal(12);
-      expect(Number(bik.day)).to.equal(30);
+    it('toGreg_text reverse conversion round-trips correctly at month/year boundaries via the service', () => {
+      const gregStr = '2024-04-12';
+      const formatted = service.date(moment(gregStr));
+      expect(formatted).to.equal('३० चैत २०८०');
     });
 
-    it('returns the correct day counts for divergent years 2082 and 2083', () => {
-      expect(BikramSambat.daysInMonth(2082, 8)).to.equal(29);
-      expect(BikramSambat.daysInMonth(2082, 10)).to.equal(29);
-      expect(BikramSambat.daysInMonth(2082, 2)).to.equal(31);
-      expect(BikramSambat.daysInMonth(2082, 4)).to.equal(31);
+    it('returns the correct formatted days for divergent years 2082 and 2083 via the service', () => {
+      // Mansir 2082 (month 8) has 29 days
+      const gregMansir29 = '2025-12-15';
+      expect(service.date(moment(gregMansir29))).to.equal('२९ मंसिर २०८२');
 
-      expect(BikramSambat.daysInMonth(2083, 8)).to.equal(29);
-      expect(BikramSambat.daysInMonth(2083, 10)).to.equal(29);
+      // Magh 2082 (month 10) has 29 days
+      const gregMagh29 = '2026-02-12';
+      expect(service.date(moment(gregMagh29))).to.equal('२९ माघ २०८२');
+
+      // Jestha 2082 (month 2) has 31 days
+      const gregJestha31 = '2025-06-14';
+      expect(service.date(moment(gregJestha31))).to.equal('३१ जेठ २०८२');
+
+      // Shrawan 2082 (month 4) has 31 days
+      const gregShrawan31 = '2025-08-16';
+      expect(service.date(moment(gregShrawan31))).to.equal('३१ साउन २०८२');
+
+      // Mansir 2083 (month 8) has 29 days
+      const greg2083Mansir29 = '2026-12-15';
+      expect(service.date(moment(greg2083Mansir29))).to.equal('२९ मंसिर २०८३');
+
+      // Magh 2083 (month 10) has 29 days
+      const greg2083Magh29 = '2027-02-12';
+      expect(service.date(moment(greg2083Magh29))).to.equal('२९ माघ २०८३');
     });
 
     it('handles Devanagari free-text search queries or invalid dates gracefully without throwing', () => {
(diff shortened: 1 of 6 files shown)


webapp/tests/karma/ts/services/format-date.service.spec.ts
```diff
@@ -429,4 +429,88 @@ describe('FormatDate service', () => {
...
+    });
+
+    it('formats Gregorian date and time into Devanagari Bikram Sambat datetime correctly', () => {
+      const date = moment('2024-06-29T10:15:30.000'); // 2081-03-15 BS
+      const formatted = service.datetime(date);
+      expect(formatted).to.contain('१५ असार २०८१');
+      expect(formatted).to.contain('10:15:30 AM');
+    });
+
+    it('handles day-boundaries (midnight) correctly', () => {
```
sugat009, line 495: **issue:** the midnight / end-of-day / timezone cases assert `service.date(x)` against `BikramSambat.toBik_text(x)`, but the service produces its result by calling that same function, so both sides always match and a wrong conversion would still pass. The DST case only checks the result is a string. A unit test like this should assert against a known, hard-coded expected value (verified against a real calendar), the way the "without stubs" test just above does (`१५ असार २०८१`), so it fails if the conversion drifts.

webapp/tests/karma/ts/services/format-date.service.spec.ts
```diff
@@ -429,4 +429,88 @@ describe('FormatDate service', () => {
...
+
+    it('correctly handles conversion across Daylight Saving Time (DST) boundaries', () => {
+      const beforeDST = moment('2024-03-10T01:59:59'); // Standard Time
+      const afterDST = moment('2024-03-10T03:00:00'); // DST (02:00:00 doesn't exist)
+      
+      expect(service.date(beforeDST)).to.be.a('string');
+      expect(service.date(afterDST)).to.be.a('string');
+    });
+
+    it('toGreg_text reverse conversion round-trips correctly at month/year boundaries', () => {
```
sugat009: **suggestion:** the `toGreg_text` round-trip and `daysInMonth` cases call `BikramSambat.*` directly and never touch `service.*`, so they're testing the `bikram-sambat` library (which has its own tests), not `FormatDateService`. To make these count as service coverage, drive them through `service.date`/`datetime` and assert the service's output.

sugat009 (changes requested):
Thanks @megha1807, nice work, and the layering is exactly right: widget UI/accessibility at the cht-form integration layer, conversion logic at unit, that's CHT's test pyramid. 69/69 green.

Requesting one round of tightening on the new conversion tests before this lands, a few can't actually catch a bad conversion because of how they assert, and since CHT treats unit tests as the confidence layer, those are worth getting right. Direction inline.

webapp/tests/karma/ts/services/format-date.service.spec.ts
megha1807: Done! I have refactored both the round-trip conversion and the divergent calendar days (2082/2083) test cases to drive their inputs through FormatDateService.date() and verify the returned Devanagari formatted strings (e.g. '३० चैत २०८०', '२९ मंसिर २०८२'). This ensures proper unit coverage is registered at the service layer.

webapp/tests/karma/ts/services/format-date.service.spec.ts
```diff
@@ -429,4 +429,98 @@ describe('FormatDate service', () => {
...
+    });
+
+    it('formats Gregorian date and time into Devanagari Bikram Sambat datetime correctly', () => {
+      const date = moment('2024-06-29T10:15:30.000'); // 2081-03-15 BS
+      const formatted = service.datetime(date);
+      expect(formatted).to.contain('१५ असार २०८१');
+      expect(formatted).to.contain('10:15:30 AM');
+    });
+
+    it('handles day-boundaries (midnight) correctly', () => {
```
sugat009, line 495: **suggestion (non-blocking):** These day-boundary tests hard-code the expected BS date (`१५ असार २०८१` here, `१६ असार २०८१` below), but `moment('2024-06-29T…')` parses in local time, so the result depends on the machine timezone. Under UTC+13 (`Pacific/Auckland`) the local calendar day shifts and these fixed expectations break. Consider pinning the timezone for these cases (or adding a UTC+13 variant) and asserting the time-of-day explicitly, so they're deterministic wherever CI runs.

webapp/tests/karma/ts/services/format-date.service.spec.ts
```diff
@@ -429,4 +429,98 @@ describe('FormatDate service', () => {
...
+      const date = moment('2024-06-29T00:00:00.000');
+      expect(service.date(date)).to.equal('१५ असार २०८१');
+    });
+
+    it('handles day-boundaries (end-of-day) correctly', () => {
```
sugat009, line 500: **note (non-blocking):** This locks in `23:59:59.999` on 29 Jun → `१६ असार २०८१`, i.e. an instant still 29 Jun locally rendering as the *next* BS day. That rollover is under discussion with the library maintainer in #9250 and may turn out to be a bug rather than the intended value. Please don't treat this exact value as the contract until that resolves, then assert whatever the agreed-correct day is.

webapp/tests/karma/ts/services/format-date.service.spec.ts
```diff
@@ -429,4 +429,98 @@ describe('FormatDate service', () => {
...
+
+    it('toGreg_text reverse conversion round-trips correctly at month/year boundaries via the service', () => {
+      const gregStr = BikramSambat.toGreg_text('2080', '12', '30');
+      const formatted = service.date(moment(gregStr));
+      expect(formatted).to.equal('३० चैत २०८०');
+    });
+
+    it('returns the correct formatted days for divergent years 2082 and 2083 via the service', () => {
+      // Mansir 2082 (month 8) has 29 days
+      const gregMansir29 = BikramSambat.toGreg_text('2082', '8', '29');
```
sugat009: **issue (blocking):** This builds its Gregorian input from the library itself (`BikramSambat.toGreg_text('2082','8','29')`) then feeds it back through `service.date()`, whose internal `toBik_text` is the exact inverse over the *same* month-length table. So it only proves the library round-trips with itself, if a month length were wrong (exactly the 2082/2083 data behind the #11252 crash this test is named for), both directions shift together and it still passes. Hard-code the Gregorian date that 29 Mansir 2082 actually falls on (from a real Nepali calendar) as the input instead, so both sides of the assertion are independent of the library. Same shape in the `toGreg_text reverse conversion round-trips` test just above.

sugat009 (changes requested):
Re-reviewed after the latest push, the conversion tests are in much better shape now: real hard-coded Devanagari expectations driven through the service, plus the invalid-date guard. A few things inline before this lands; only the divergent-years one is blocking.

sugat009 (commented):
CI's already red here, and that's expected now that #11251 is merged in: the fix corrects a few of the Bikram Sambat expected values, so a couple of the date assertions in the conversion tests need refreshing. The karma failure points to the exact one, please update that expectation to match the corrected conversion.

sugat009 (commented):
Re-checked after your latest push. The end-of-day expectation is fixed and the karma suite is green on CI, nice work.

Three checks are still red; here's the breakdown so you don't chase ghosts:

**issue:** `test-cht-form` is a real failure in this PR's new spec, `tests/integration/cht-form/default/bikram-sambat-widget.wdio-spec.js`: the test "clears populated values when manual fields are cleared" fails on every retry because the harness asserts a clean browser console, and clearing the day field makes Enketo log `Data node: day with null-based index: 0 not found. Ignored.` Have a look at how the widget writes and clears the hidden day node (the two lines removed in `webapp/src/js/enketo/widgets/bikram-sambat-datepicker.js` are a good starting point). Re-running CI won't help this one.

The other two red checks (`ci-integration-sentinel-all` purging timeouts, and the contact search hang in `ci-webdriver-default-enketo-minimum-browser`) are in specs this PR doesn't touch and look like infra flakes; we can re-run those.

**suggestion (non-blocking):** before your next push, merge the latest `10707-dmp-2026-enhance-bikram-sambat-support-in-cht` into your branch (merge it, rather than rebasing or force-pushing over it). Your branch currently predates the #11251 fix, so local karma will fail the end-of-day test even though CI passes it (CI tests your branch merged with the base). Syncing makes local results match CI. Small heads up: a couple of merge commits we'd pushed to this branch were lost in the last force-push; merging the base first avoids that.

sugat009 (changes requested):
Re-checked after your latest push. The console failure is properly fixed, and fixed the right way: `ignore` is exactly the enketo-core convention for widget-internal inputs (the engine skips them via `:not(.ignore)`), and the `stopPropagation` placement after the real-input early return keeps model updates working. `test-cht-form` is green. Also thanks for merging the epic in cleanly this time, that's exactly the workflow we wanted.

**issue:** one character slipped into the template though: `webapp/src/js/enketo/widgets/bikram-sambat-datepicker.js` line 219 now has `\${MONTH_PLACEHOLDER}`, and inside a template literal that backslash turns off the interpolation, so an empty date field (the default state of every new form) renders the month button as the literal text `${MONTH_PLACEHOLDER}` instead of महिना. CI missed it because our tests only assert the placeholder after clear, which goes through the other, correctly interpolated render at line 169. Removing the backslash fixes it; worth also adding a small assertion on the empty widget's initial month button text so this kind of regression gets caught.

The `ci-integration-all-k3d` failure is unrelated flakiness (sentinel transition count assertions that also failed on other branches the same day); we'll re-run it.

sugat009 (approved):

Issue #10700, opened 2026-03-05
Support Photo Capture for Sub-contacts Within Parent Registration Forms


iesmail-znz:
## Is your feature request related to a problem? Please describe.

In community health contexts where CHWs register households with 10 or more members at once, the current workflow for capturing contact photos is highly inefficient and impractical.

Today, photo capture requires opening a `person-edit` form individually for each household member **after** the household registration is complete. For a household with 10 members, this means the CHW must:

1. Complete the household (`clinic-create`) registration form
2. Navigate back to each individual member's profile
3. Open the `person-edit` form for each person
4. Capture and submit the photo
5. Repeat steps 2–4 for every member

This results in the same action being performed 10+ times in a single household visit, creating a fragmented and repetitive workflow that is disconnected from the natural point of registration.

In practice, CHWs are likely to skip photo capture altogether due to the burden this imposes and directly undermining the purpose of the feature and reducing data quality at scale.

---

## Describe the solution you'd like

We would like support for **capturing photos (and other file attachments) for sub-contacts directly within a parent registration form**, such as `clinic-create`.

Specifically:

- Within the household registration form, as each household member is being registered, the CHW should be able to capture that member's photo inline as part of the same form flow
- The captured photo should be saved as an attachment on the respective person contact document upon form submission
- The photo should be **visible on the individual contact's profile page** in the CHT app once submitted,  just as it would be if captured via `person-edit`
- This should work **offline**, consistent with CHT's offline-first architecture

This would allow photo capture to happen at the natural point of registration, as part of a single uninterrupted workflow, significantly improving data completeness and CHW experience.

iesmail-znz:
@antonykhaemba, @dianabarsan, for displaying the photo, a more robust solution would be to place the photo in the contact's
  headshot (the avatar at the top left of the profile) rather than displaying it as another field in the contact profile. If a CHW wants
  to view the full photo, they can click on the headshot and it expands.

vikrantwiz02:
Hii @dianabarsan, @andrablaj, @iesmail-znz I was looking into the code and I think, I can fix it.

So, right now, the bottleneck is in contact-save.service.ts. When a household form is submitted, the system just dumps every single uploaded photo onto the main parent document.

My idea to fix it is just to add a small routing loop in there. After the photos attach to the parent, we can loop through those newly generated child documents, match the filenames, and just move the photo attachments over to the correct child.

[...]

PR #11256, opened 2026-07-17
refactor(#10700): re-write Enketo form save workflow

jkuester:
# Description

This PR basically re-writes the doc-saving workflow when an Enekto form (either contact/app/training form) is completed. The existing legacy code suffered from several major problems that made it very difficult to continue developing features on top of it:

- The doc-processing pipeline for contact docs vs reports was almost completely disjointed.  Aside from the core xml-to-object functions, none of the other logic/workflows was coherently shared between contacts and reports.  This meant that we had a lot of duplication and inconsistency between how things were being handled. And, any new features (e.g. attachments) needed to be essentially implemented twice.
- Historical changes to these workflows have been haphazard. The current structure is now incredibly complex with very poor abstraction/encapsulation. It is difficult to understand the data flow and/or which parts of the code are responsible for what.  
- The code does a poor job of communicating the baked-in assumptions (instead nesting the handling of them deep in the workflows).  

The main impetus behind is PR is to provide a better structured base for solving https://github.com/medic/cht-core/issues/10700.  The hope is that the new functionality in https://github.com/medic/cht-core/pull/10923 can be ported on top of the changes in this PR in a way that is way more simple and maintainable.

## What is in this PR

The goal was to have this PR be a complete drop-in replacement for the existing functionality. I have done my best to ensure the new code accurately matches the behavior of the old code and there are no new features that I intended to directly support by these changes.

The only behavior difference that I am aware of currently is a new error I added to the save workflow that will fail with a helpful error message when a _contact_ form does not have the contact data in a group that is named for the actual contact type.  (So, a person-create form needs to have the contact data in the `person` group.)  This requirement is listed explicitly [in our docs](https://docs.communityhealthtoolkit.org/building/forms/contact/#survey-sheet) and it is implicitly expected by the pre-population logic for setting the initial data onto the contact. Previously it was technically possible to save a contact form with data in a different group. Because our documentation is explicit about how the forms are supposed to be structured, I do not see this as a breaking change. Also, if some does have this issue, they will hit the error and be unable to save those contact forms. It will not just cause weird data issues by failing silently. The fix would just be to update the form config to name the group properly.

### Added `FormConfig`

A new `FormConfig` class has been added in `webapp/src/ts/services/form/form-config.ts` that serves as a wrapper for the form doc and its relevant attachments.  This FormConfig is retrieved via the XmlFormsService when rendering the form and then can be re-used when saving the form. This wrapper basically centralizes the handling of the form doc and its attachments and avoids the need to retrieve these things multiple times during the render/save process.

### Added new `EnketoFormData` classes

The new `EnketoFormData` classes added in `webapp/src/ts/services/form/form-data.ts` are wrappers to encapsulate the XML data model produced when saving an  Enekto form. These data classes own the logic for interacting/querying the XML data as well as de-serializing that data to JS objects.

Being able to operate against the XML data is useful for queries (either via css selectors or XPaths) which do not have a simple alternative when using normal JS objects.  These wrappers help define the available data queries as well as hold on to the `_id` references for the data to allow for easy interlinking.  

The xml de-serialization logic from the now-deleted `EnketoTranslationService` was moved into these EnketoFormData classes.  (The JS -> XML serialization logic was just moved to where it was being used in the `EnketoPrepoulationDataService`.)

### Move Contact Save functionality into EnketoService

Previously, we had the logic for saving contact forms separated out into the ContactSaveService.  I have move this into the EnektoService so now there are only two entry points for saving a form: `EnketoService.saveContact` and `EnketoService.saveReport`.  Keeping the code in the same service allows for better alignment between the workflows.

As before, there is still probably "too much" code in the EnektoService, but I have balanced things out a bit by deleting duplicate logic and/or moving it into the EnektoFormData classes. The end result is the actual LOC in the enketo.service file has only grown around 30 lines...

### Clean up form/enketo service unit tests

A large percentage of the changed lines in this PR are a result of me overhauling the unit tests for the FormService and EnketoService.  Previously there was a TON of duplication/leaking between those unit tests (because of how I previously handled separating the FormService from enketo... :sweat:).  The time has come to pay the piper and I have refactored the FormService tests especially to just mock the EnektoService.

## AI Disclosure

Basically every single line of the implementation code was hand-written by me. I leveraged Claude a lot for the investigation and minor code structure advice, but ultimately, I really wanted full control/responsibility for porting all the logic line-by-line.  Any bugs will be ones that I personally wrote. :+1:  :sweat_smile: 

I did leverage Claude heavily for the updates to the unit tests. Even there, I have gone through everything line-by-line and made many editorial changes and requested additional tests, etc.  There is a lot of lines changed, so it is possible we missed some stuff, but I am highly confident the functionality is well tested and the tests are very maintainable going forwards.

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:2293-rewrite-forms/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:2293-rewrite-forms/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:2293-rewrite-forms/docker-compose/cht-couchdb-clustered.yml)

# License


tests/integration/cht-form/default/person-edit.wdio-spec.js
```diff
@@ -31,8 +31,6 @@ describe('cht-form web component - Edit Person Form', () => {
       role: 'chw',
       external_id: '12345',
       notes: 'Test notes',
-      contact: undefined,
-      parent: undefined,
```
jkuester, line 35: The changes here are actually due to the updates I made in the cht-form/src/app.component.ts.  When we write a doc to Couch, properties set to `undefined` are not included in the JSON data. cht-form returns the JS objects directly and was not doing anything to account for `undefined` properties.  Unfortunately, Chai treats an object with a property set to `undefined` as not deep equal to an object that just does not have the property at all.  So, these values were previously needed here to account for these props getting set in the Enekto service, etc.

webapp/web-components/cht-form/src/app.component.ts
```diff
@@ -232,18 +229,6 @@ export class AppComponent {
...
-  private getFormDetails() {
-    const $html = $(this._formHtml!);
-    const hasContactSummary = $(this._formModel!)
-      .find('> instance[id="contact-summary"]')
-      .length === 1;
-    return {
-      html: $html,
-      model: this._formModel,
-      hasContactSummary: hasContactSummary
-    };
```
jkuester, line 249: FTR, this property was no longer being used prior to my changes in this PR.

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -434,64 +430,55 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
     this.globalActions.setEnketoSavingStatus(true);
     this.globalActions.setEnketoError(null);
 
-    return Promise
-      .resolve(form.validate())
-      .then((valid) => {
-        if (!valid) {
-          throw new Error('Validation failed.');
-        }
```
jkuester, line 442: This is now done in the EnketoService.saveContact.

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -434,64 +430,55 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
...
+    if (this.duplicatesAcknowledged && this.duplicates.length) {
+      this.telemetryService.record(
+        ['enketo', 'contacts', this.enketoContact.type, 'duplicates_acknowledged'].join(':')
+      );
+    }
```
jkuester: The most important thing is that this runs before the dupchecking logic in `formService.saveContact`. I do not think it is an issue that this now runs before the Enketo validation for the form.  The Enketo validation is still going to run before the dup checking.

tests/integration/cht-form/default/person-edit.wdio-spec.js
dianabarsan: >  properties set to undefined are not included in the JSON data

I know, because the document is stringified and undefined properties don't survive the stringify,

webapp/src/ts/modules/contacts/contacts-edit.component.ts
```diff
@@ -434,64 +430,55 @@ export class ContactsEditComponent implements OnInit, OnDestroy, AfterViewInit {
...
 
-        // Updating fields before save. Ref: #6670.
-        form.view.html.dispatchEvent(events.BeforeSave());
+    return this.formService
```
dianabarsan: Can we switch this to await? I think it will increase the readability by a lot!

webapp/src/ts/services/enketo.service.ts
```diff
@@ -564,33 +397,228 @@ export class EnketoService {
...
+        ...contactDoc,
+        ...formData.deserializeDoc(config),
+        type: contactDoc.type,
+        contact_type: contactDoc.contact_type,
+        _attachments: formAttachments
+      };
+
+      const siblings = EnektoContactFormData.SIBLING_FIELD_NAMES
+        .map(fieldName => ({ fieldName, doc: formData.getSiblingData(fieldName)?.deserializeDoc(config) }))
+        .map(({ fieldName, doc }) => ({ fieldName, doc: this.initializeContactSibling(rootOutputDoc, doc)}));
```
dianabarsan: Do we need two chained maps here or can everything be done within one iteration? 

Wondering if we did something like: 
```javascript
const siblingSave = EnektoContactFormData.SIBLING_FIELD_NAMES.map(async (fieldName) => {
  const siblingData = formData.getSiblingData(fieldName)?.deserializeDoc(config);
  const siblingDoc = this.initializeContactSibling(rootOutputDoc, siblingData);
  rootOutputDoc[fieldName] = await this.getContactSiblingValue(
    siblingDoc, rootOutputDoc[fieldName], defaultData[fieldName]
  );
});
await Promise.all(siblingSave);
```

webapp/src/ts/services/enketo.service.ts
```diff
@@ -564,33 +397,228 @@ export class EnketoService {
...
+
+      const rootOutputDoc: Record<string, any> = {
+        ...contactDoc,
+        ...formData.deserializeDoc(config),
+        type: contactDoc.type,
+        contact_type: contactDoc.contact_type,
+        _attachments: formAttachments
+      };
+
+      const siblings = EnektoContactFormData.SIBLING_FIELD_NAMES
```
dianabarsan: Please consider moving this sibling handling section into its own function.

webapp/src/ts/services/enketo.service.ts
```diff
@@ -564,33 +397,228 @@ export class EnketoService {
...
+      return sibling;
+    }
+    if (currentValue._id === defaultValue?._id) {
+      return defaultValue;
+    }
+
+    return await this.getContactFromDatasource(Qualifier.byUuid(currentValue._id));
+  }
+
+  private dehydrateContactLineage(contactDoc: Record<string, any>) {
```
dianabarsan: I believe the internal name was "minify", instead of "dehydrate". I'm fine with either, but prefer to have a naming convention.

webapp/src/ts/services/enketo.service.ts
```diff
@@ -564,33 +397,228 @@ export class EnketoService {
...
+        .filter(Boolean)
+        .slice(1)
+        .join('.'));
+  }
+
+  private initializeDoc(defaultData: Record<string, any>): Record<string, any> {
+    return {
+      _id: defaultData._id || uuid(),
+      reported_date: Date.now(),
+      ...defaultData
```
dianabarsan: Spread first, and then reported_date, so it's never overwritten by a rogue field?

webapp/src/ts/services/enketo.service.ts
```diff
@@ -564,33 +397,228 @@ export class EnketoService {
...
+    originalAttachments: Record<string, any> = {}
+  ) {
+    const hasCustomAttachmentName = (fileName: string) => !fileName.startsWith(this.USER_FILE_ATTACHMENT_PREFIX)
+      && !fileName.startsWith(`${this.USER_BINARY_ATTACHMENT_PREFIX}/`);
+    const isExistingFileAttachment = (fileName: string) => fileName.startsWith(this.USER_FILE_ATTACHMENT_PREFIX)
+      && rootData.findNodeWithTextContent(fileName.slice(this.USER_FILE_ATTACHMENT_PREFIX.length));
+    const binaryAttachments = rootData.binaryTypeElements
+      .map(element => this.buildBinaryAttachmentData(form, originalAttachments, element))
+      .filter(({ attachment }) => attachment)
+      .reduce((acc, { filename, attachment }) => ({ ...acc, [filename]: attachment }), {});
```
dianabarsan: I know it's common practice to use the term "acculumator" and short "acc" for reduce, but I feel like this falls into the same trap as for loops where `i` loses all connection to what value it represents. 

So, a minor request is to give acc a meaningful name: attachments.

webapp/src/ts/services/form.service.ts
```diff
@@ -306,24 +280,24 @@ export class FormService {
     return docs;
   }
 
-  private async completeReport(formInternalId, form, docId?) {
-    const formDoc = await this.ngZone
-      .runOutsideAngular(() => this.xmlFormsService.getDocAndFormAttachment(formInternalId));
+  private async completeReport(enketoForm: EnketoForm, docId?) {
     if (docId) {
-      return this.enketoService.completeExistingReport(form, formDoc, docId);
+      const doc = await this.getReport(Qualifier.byUuid(docId));
```
dianabarsan, line 324: this used to run outsize ngzone before.

webapp/src/ts/services/xml-forms.service.ts
```diff
@@ -368,23 +369,35 @@ export class XmlFormsService {
...
-            const errorTitle = 'Error in XMLFormService : getDocAndFormAttachment : ';
-            let errorMessage = `Failed to get the form "${internalId}" xform attachment`;
-            if (err.status === 404) {
-              errorMessage = `The form "${internalId}" doesn't have an xform attachment`;
-            }
-            console.error(errorTitle, errorMessage);
-            return Promise.reject(new Error(errorTitle + errorMessage));
-          });
-      });
+            const errorDetails = err.status === 404
```
dianabarsan: It looks like this logic is duplicated, where `private get(internalId)` will also throw a missing attachment error.

webapp/src/ts/services/xml-forms.service.ts
```diff
@@ -368,23 +369,35 @@ export class XmlFormsService {
...
+      .getAttachment(id, name)
+      .then(blob => this.fileReaderService.utf8(blob));
+  }
+
+  async getFormConfig(formType: FormType, id: string) {
+    return this.ngZone.runOutsideAngular(async () => {
+      // contact_types config stores full _id value. All other forms are referenced by internalId
+      const formDoc = await (formType === 'contact'
+        ? this.dbService.get().get(id)
+        : this.get(id));
```
dianabarsan: I would rather this logic is all moved to `this.get`

webapp/src/ts/services/xml-forms.service.ts
```diff
@@ -368,23 +369,35 @@ export class XmlFormsService {
...
-            let errorMessage = `Failed to get the form "${internalId}" xform attachment`;
-            if (err.status === 404) {
-              errorMessage = `The form "${internalId}" doesn't have an xform attachment`;
-            }
-            console.error(errorTitle, errorMessage);
-            return Promise.reject(new Error(errorTitle + errorMessage));
-          });
-      });
+            const errorDetails = err.status === 404
+              ? `The form "${id}" doesn't have an xform attachment`
```
dianabarsan: The function would be much leaner and cleaner if this would be: 
```
throw new AttachmentMissingError(id, errorDetails);
```
or something similar.

mrjones-plip:
I came here to see what this was all about and saw this:

> I do not see this as a breaking change

Amazing work Josh!!

webapp/src/ts/components/training-cards-form/training-cards-form.component.ts
```diff
@@ -112,23 +113,23 @@ export class TrainingCardsFormComponent implements OnInit, OnDestroy {
       this.loadingContent = true;
       this.geoHandle?.cancel();
       this.geoHandle = this.geolocationService.init();
-      const form = await this.xmlFormsService.get(this.trainingCardFormId);
-      await this.ngZone.run(() => this.renderForm(form));
+      const formConfig = await this.xmlFormsService.getFormConfig('training-card', this.trainingCardFormId!);
```
dianabarsan: this is totally a nitpick. 
it irks me that `training-card` here is a drop in string and `this.FORM_WRAPPER_ID` is a constant. they are exactly the same - well the string is not the same, but the usage is the same - and one is constant one is string. :(

webapp/src/ts/services/enketo.service.ts
```diff
@@ -564,33 +397,228 @@ export class EnketoService {
...
+        .map(({ fieldName, doc }) => ({ fieldName, doc: this.initializeContactSibling(rootOutputDoc, doc)}));
+      await Promise.all(siblings.map(async ({ fieldName, doc }) => {
+        rootOutputDoc[fieldName] = await this.getContactSiblingValue(
+          doc, rootOutputDoc[fieldName], defaultData[fieldName]
+        );
+      }));
+      const outputSiblings = siblings
+        .filter(({ fieldName, doc }) => doc && rootOutputDoc[fieldName] === doc)
+        .map(({ doc }) => doc)
+        .filter(doc => !!doc);
```
dianabarsan:
```suggestion
  const outputSiblings = siblings.flatMap(({ fieldName, doc }) => doc && rootOutputDoc[fieldName] === doc ? [doc] : []);
```

webapp/src/ts/services/enketo.service.ts
```diff
@@ -327,26 +334,27 @@ export class EnketoService {
...
       selector,
       titleKey,
       isFormInModal,
     } = formContext;
 
-    this.replaceDataI18nTranslations(doc.html);
-    this.replaceJavarosaMediaWithLoaders(doc.html);
+    const $html = $(formConfig.html);
+    this.replaceDataI18nTranslations($html);
+    this.replaceJavarosaMediaWithLoaders($html);
```
dianabarsan: Was there the intention to replace `formConfig.html` with `$html` ? 
It looks like the result from `replaceDataI18nTranslations` and `replaceJavarosaMediaWithLoaders` is just droppped, because `$hml` is not used anywhere, and the `formConfig.html` is later used in `renderFromXmls`

webapp/src/ts/services/form/form-data.ts
```diff
@@ -0,0 +1,186 @@
...
+      `.//*[text()=${JSON.stringify(textContent)}]`,
+      this.rootElement,
+      null,
+      XPathResult.FIRST_ORDERED_NODE_TYPE,
+      null
+    );
+    return result.singleNodeValue;
+  }
+
+  public getNodeByXpath(contextNode: Node, rawXpath?: string | null): Node | null {
```
dianabarsan: can we make sure that local `./` db-doc-ref references still resolve? 

there were tests that checked this (db-doc-ref-in-repeats-with-local-references, db-doc-ref-in-deep-repeats-with-local-references, db-doc-ref-same-as-repeat) but these were dropped in the rewrite. 

It also looks like those xml files are now abandoned and not used anywhere. If a test xml is no longer useful, it should be removed. But I don't think this applies to these test cases.


Please also check `db-doc-ref-in-deep-repeats-extra-repeats` file that checks that the `db-doc` is matched vs a non `db-doc` field.

webapp/src/ts/services/form/form-data.ts
```diff
@@ -0,0 +1,186 @@
...
+    private readonly xmlDoc: XMLDocument,
+    id: string,
+  ) {
+    super(xmlDoc.documentElement, id);
+    this.binaryTypeElements = Array.from(this.rootElement.querySelectorAll('[type=binary]'));
+  }
+
+  public findNodeWithTextContent(textContent: string) {
+    const result = this.xmlDoc.evaluate(
+      `.//*[text()=${JSON.stringify(textContent)}]`,
```
dianabarsan: XPath 1.0 string literals have no escape mechanism. Editing a report/contact that has a user-file-* attachment whose filename contains " makes evaluate() throw, so the edit cannot be saved at all; a filename containing \ silently never matches, and the attachment is deleted while still      
referenced. Master compared strings in JS.

webapp/src/ts/services/enketo.service.ts
```diff
@@ -564,33 +397,228 @@ export class EnketoService {
...
+    const isExistingFileAttachment = (fileName: string) => fileName.startsWith(this.USER_FILE_ATTACHMENT_PREFIX)
+      && rootData.findNodeWithTextContent(fileName.slice(this.USER_FILE_ATTACHMENT_PREFIX.length));
+    const binaryAttachments = rootData.binaryTypeElements
+      .map(element => this.buildBinaryAttachmentData(form, originalAttachments, element))
+      .filter(({ attachment }) => attachment)
+      .reduce((acc, { filename, attachment }) => ({ ...acc, [filename]: attachment }), {});
+    const newFileAttachments = FileManager
+      .getCurrentFiles()
+      .map(file => ({
+        name: `${this.USER_FILE_ATTACHMENT_PREFIX}${file.name}`,
```
dianabarsan, line 632: Old code stripped special characters (UUID fallback for e.g. Devanagari-only names) and rewrote the referencing field values.  The specs checking this were dropped without replacement.

webapp/src/ts/services/form/form-data.ts
```diff
@@ -0,0 +1,186 @@
...
+      formConfig.repeatPaths,
+      Xpath.getElementRawXPath(this.rootElement)
+    );
+  }
+
+  public deserializeDoc(formConfig: FormConfig): Record<string, any> {
+    return {
+      _id: this.id,
+      form_version: formConfig.doc.xmlVersion,
+      ...this.deserialize(formConfig)
```
dianabarsan: Please move the spread to be first, so elements from the form don't end up overwriting _id or form_version

webapp/src/ts/services/form/form-data.ts
```diff
@@ -0,0 +1,186 @@
...
+  private getNodeWithLineage(contextNode?: Node | null, lineage: Node[] = []): Node[] {
+    if (!this.isElementNode(contextNode)) {
+      return lineage;
+    }
+    lineage.unshift(contextNode);
+    return this.getNodeWithLineage(contextNode.parentNode, lineage);
+  }
+}
+
+export class EnektoContactFormData extends EnketoRootFormData {
```
dianabarsan: Enekto typo, even though it sounds way cooler!

webapp/web-components/cht-form/src/app.component.ts
```diff
@@ -288,12 +273,21 @@ class ChtFormEnketoFormContext implements EnketoFormContext {
...
+  formXml?: string;
+  formHtml?: string;
+  formModel?: string;
   contactType?: string;
   content?: Record<string, any>;
   editing = false;
 
-  get formDoc() {
-    return { _id: this.formId };
+  get formConfig() {
```
dianabarsan: Every time we use this getter, it ends up parsing the xml (via the FormConfig constructor). Should we consider cashing here?

webapp/web-components/cht-form/src/app.component.ts
```diff
@@ -288,12 +273,21 @@ class ChtFormEnketoFormContext implements EnketoFormContext {
...
+    return new FormConfig(
+      { _id: this.formId },
```
dianabarsan: Do we need additional fields on this doc here? 
`saveReport` destructures internalId/xmlVersion from config.doc — with only _id, the emitted report docs seem to lose their form field (on master completeNewReport set form: formId), and binary attachments get named user-file/undefined/...

dianabarsan (changes requested):
Very nice rewrite! I think you've definitely taken this into the right direction. 
I left some minor comments and possibly found some issues, but nothing major.

webapp/src/ts/components/training-cards-form/training-cards-form.component.ts
jkuester: Yeah... so, this is why real languages have enums.  🤦    If TS enums were not so broken/deprecated we could use them here.

That being said, TIL that TS does not care if you use _the same name_ for both a type and a property.  Eslint whines a bit, but the TS transpiler has no problem sorting out what is up.  So, I can make a `FormType` that is a dictionary and then also have a `FormType` that is a `type` (based on the values from the dict) and everything is great!

webapp/src/ts/services/form.service.ts
jkuester: Good catch! Looking at this more, I think we should just wrap the whole `save` and `saveContact` flows with `runOutsideAngular`.  AFICT, the controller code handles all the UI updates and none of that needs to be directly triggered from the FormService logic.

webapp/src/ts/services/form/form-data.ts
jkuester: Ahh, shoot!  Good call. I did not consider the file names breaking the xpath.  I could not find any clean way to save the XPath query approach here (as noted, there is no escaping). 

The best alternative I have found is just to walk all the nodes and check the `textContent` for each.  Not the absolute best for performance, but the implementation simplicity seems to make this this best place to start.

webapp/src/ts/services/form/form-data.ts
jkuester: 😞 Well, this is def one of those bugs (or really a set of bugs) that I personally wrote...  😓   I thought I had verified all the edgiest cases around this, but digging through those xml files shows some situations that def break my logic here.  Good call on checking them!  

Went back to the drawing board on this function here (with Claude) and I believe that we can actually get a even more simple version of this that properly handles all the cases by pushing the logic back up into the EnektoService where we already have all the doc data in context.  

As a part of this refactor, I have added new unit tests for the EnektoService that use all these existing (and previously unused xml files):

- deep-file-fields.xml
- db-doc-ref-broken-ref.xml
- db-doc-ref-outside-of-repeat.xml
- db-doc-ref-in-deep-repeat.xml
- extra-docs-with-repeat.xml
- db-doc-ref-same-as-repeat.xml
- db-doc-ref-in-repeats-with-local-references.xml
- db-doc-ref-in-deep-repeats-with-local-references.xml
- db-doc-ref-in-deep-repeats-extra-repeats.xml

The tests for the last 4 files were failing prior to my latest refactor of this code.  Now all the tests work!

webapp/src/ts/services/xml-forms.service.ts
jkuester: Honestly, this all feels unnecessary.  Now that I am in here stirring all this up, I might as well just fix the error flows. I don't see why we need so much code here...

webapp/web-components/cht-form/src/app.component.ts
jkuester: Yeah probably not a big deal either way, but I re-worked this to just rebuild the FormConfig when one of the properties changes.

webapp/web-components/cht-form/src/app.component.ts
jkuester: All good points, but IMHO can be out of scope for this work since none of this was included before.  Basically we might need to think of a more structured way of accepting form properties here in cht-form.  The form xml, model, and html were added as separate string properties on cht-form originally because it meant that the values could be passed via HTML attributes on the web component element.  That approach is  limited to just strings, though, and even now we have other properties like `contactSummary` that can only be set via JS.

Happy to do more refactoring here if you want to worry about this now, but otherwise, I think we still at least have the same behavior as before...

webapp/src/ts/services/enketo.service.ts
jkuester: Ahh, shoot. Yeah, actually want to set the $html on the XmlFormContext. (Keep the JQuery limited to the render internals).  Fixed.

webapp/src/ts/services/enketo.service.ts
jkuester: I def agree on combining the first two `map` calls. It is poor form to re-assign the `doc` property in the second `map`.  👍 

However, I will push back on adding in the third `map` (where we do the assignment on the `roodOutputDoc`).  Technically it works to put it in, but IMHO it really detracts from the readability.  I don't really like side-effects in `map` calls (even though I am def guilty of doing it).  But right here the logic is already pretty involved, so, I would prefer maximize the readibility of the code here.

webapp/src/ts/services/enketo.service.ts
jkuester: Going to push back on this too.  🤔 While I appreciate the fact we could condense this to a single `flatMap`, I think you lose a lot of readability.  Looking at this again, I see the second `filter` is redundant.  I will refactor so nothing is redundant.

webapp/src/ts/services/enketo.service.ts
jkuester: You are right.  For... reasons we use `hydrate` and `minify` in shared-libs/lineage.

webapp/src/ts/services/enketo.service.ts
jkuester: Cannot do that because we don't want to blow away the existing reported_date value when updating a contact.  Honestly, this function is too abstract to be useful.  I am going to smash the _id and reported logic back down into the main save functions.  One other problem with the current approach is that we can have slightly different reported_date values for docs written in the same form.  Not a huge deal, but poor form.  Pushing into the main save fixes this too.

webapp/src/ts/services/enketo.service.ts
jkuester: So, that was recent logic added for contact attachments that never existed for the report attachments. I tried a variety of values to see if I could find anything that was a valid file-name but would break the attachment handling and I could not come up with any. For example, I tried file with names like these:

- `50%off`
- `café résumé`
- `incense"the"gpu`
- `incense_the_gpu`
- `incense_the_gpu #2`
- `is it done?`
- `नमस्`

These all worked for me.  But, I am happy to add that special char handing back in if we have a use for it!  👍

jkuester: :laughing: Also, if you are wondering what is up with some of those file names, I was just testing with a random meme pic from my Downloads folder (nothing like bad AI memes to make the Enketo forms better....):  

incense_the_gpu

webapp/src/ts/services/xml-forms.service.ts
```diff
@@ -343,50 +339,51 @@ export class XmlFormsService {
...
   }
 
+  async getFormConfig(formType: FormType, id: string) {
+    return this.ngZone.runOutsideAngular(async () => {
+      const formDoc = await this.get(id, formType);
+      const xmlAttachmentName = this.findXFormAttachmentName(formDoc);
+      if (!xmlAttachmentName) {
+        throw new Error(`Could not get [xml] form attachment for form [${id}].`);
+      }
+      const [xml, html, model] = await Promise.all([
```
dianabarsan, line 370: This latest rewrite left `hasRequiredAttachments` without any callers. If no longer needed, please remove.

dianabarsan (approved):
Awesome work!

andrablaj:
@jkuester quick one: this PR is marked as a `chore` - is that accurate?

jkuester:
:+1: @andrablaj it is accurate because this PR was definitely a massive chore for Diana and I to get done....  :sweat_smile: 

But really, thanks for the call out .  I had set `chore` because there is no changes here that should be noticeable to the user (so not really a `feat` or a `fix`).  Looking at the [available commit types](https://conventional-changelog.js.org/presets/angular/#commit-types-and-sections) again, I think `refactor` is a better type to use.  :+1:

andrablaj:
@jkuester per CHT's [dev workflow documentation](https://docs.communityhealthtoolkit.org/community/contributing/code/workflow/#commit-message-format), `refactor` seems the correct choice. `chore` is for:

> General maintenance and tooling updates that are not related to build logic or tests; does not change runtime behavior, like a change in a README file, updating dependencies etc.

PR #11257, opened 2026-07-17
test(#11246): harden Reports Date Filter Sidebar tests (Bikram Sambat)

### LINKED ISSUE #11246: Bikram Sambat: harden feature test coverage (widget, reports filter, conversion, a11y/security)

megha1807:
# Description

Hardens test coverage for the Reports Date Filter Sidebar under Section B of issue medic#11246.

## Changes:
- **Date Filter Component Unit Tests**:
  - Asserted start-of-day (midnight) correctness on From-date selection.
  - Asserted that `maxDate` passed to `setupNepaliDatePicker` represents today's date in Devanagari format (restricting future date selection).
  - Verified that From and To date pickers use separate component instances and hidden inputs.
  - Verified that calling `clear()` resets the Nepali date filter state.
  - Verified that `ngOnDestroy` completely cleans up multiple picker instances and body-appended overlays.
- **Date Filter E2E Tests**:
  - Added assertions to verify future date cells are disabled in the BS calendar popup.
  - Added assertions to verify Escape key and outside-click dismissals close the calendar popup.
  - Added assertions to verify viewport anchoring remains aligned upon window scroll.
  - Added assertions to verify correct pre-selection highlighting when reopening the To-date filter.

This is PR 2 of 2. PR 1 (#11255) addressed the Enketo Form Widget and Date Conversions (Sections A, C, D).

## Issue Number

Fixes #11246

# Code review checklist

### DIFF
--- a/tests/e2e/default/reports/sidebar-filter.wdio-spec.js
+++ b/tests/e2e/default/reports/sidebar-filter.wdio-spec.js
@@ -217,6 +217,9 @@ describe('Reports Sidebar Filter', () => {
     await reportsPage.clickSidebarFilterDateAccordionHeader();
     expect(await picker.isDisplayed()).to.be.false;
 
+    // Re-expand the Date accordion since the dismissal click collapsed it
+    await reportsPage.openSidebarFilterDateAccordion();
+
     // 4. Select From and To dates
     await reportsPage.setSidebarFilterBikFromDate();
     await reportsPage.setSidebarFilterBikToDate();
--- a/tests/e2e/default/reports/sidebar-filter.wdio-spec.js
+++ b/tests/e2e/default/reports/sidebar-filter.wdio-spec.js
@@ -199,7 +199,7 @@ describe('Reports Sidebar Filter', () => {
     await reportsPage.openSidebarFilterDateAccordion();
 
     // 1. Max Date parity - verify future dates are disabled
-    await $('#fromDateFilter').click();
+    await reportsPage.clickSidebarFilterFromDate();
     await reportsPage.waitForNepaliDatePickerDisplayed();
 
     // Future dates in the current month should be disabled
@@ -212,10 +212,9 @@ describe('Reports Sidebar Filter', () => {
     expect(await picker.isDisplayed()).to.be.false;
 
     // 3. Dismiss by clicking outside (e.g. the accordion header)
-    await $('#fromDateFilter').click();
+    await reportsPage.clickSidebarFilterFromDate();
     expect(await picker.isDisplayed()).to.be.true;
-    const accordionHeader = await $('#date-filter-accordion mat-expansion-panel-header');
-    await accordionHeader.click();
+    await reportsPage.clickSidebarFilterDateAccordionHeader();
     expect(await picker.isDisplayed()).to.be.false;
 
     // 4. Select From and To dates
@@ -223,7 +222,7 @@ describe('Reports Sidebar Filter', () => {
     await reportsPage.setSidebarFilterBikToDate();
 
     // 5. Reopen active selection verification
-    await $('#toDateFilter').click();
+    await reportsPage.clickSidebarFilterToDate();
     // Verify that the active selected date is highlighted correctly
     expect(await reportsPage.isNepaliDatePickerActiveCellDisplayed()).to.be.true;
     
--- a/tests/e2e/default/reports/sidebar-filter.wdio-spec.js
+++ b/tests/e2e/default/reports/sidebar-filter.wdio-spec.js
@@ -200,37 +200,20 @@ describe('Reports Sidebar Filter', () => {
 
     // 1. Max Date parity - verify future dates are disabled
     await $('#fromDateFilter').click();
-    let picker;
-    await browser.waitUntil(async () => {
-      const pickers = await $$('.nepali-date-picker');
-      for (const p of pickers) {
-        if (await p.isDisplayed()) {
-          picker = p;
-          return true;
-        }
-      }
-      return false;
-    }, { timeout: 5000, timeoutMsg: 'Nepali date picker not displayed' });
+    await reportsPage.waitForNepaliDatePickerDisplayed();
 
-    // Future dates in the current month should be disabled (.disable)
-    const disableCells = await picker.$$('table tbody td.current-month-date.disable');
+    // Future dates in the current month should be disabled
+    const disableCells = await reportsPage.getDisabledNepaliDateCells();
     expect(disableCells.length).to.be.greaterThan(0);
 
     // 2. Escape Dismissal - press Escape and verify picker is dismissed
     await browser.keys(['Escape']);
+    const picker = reportsPage.getNepaliDatePicker();
     expect(await picker.isDisplayed()).to.be.false;
 
-    // 3. Viewport positioning & scroll alignment
+    // 3. Dismiss by clicking outside (e.g. the accordion header)
     await $('#fromDateFilter').click();
     expect(await picker.isDisplayed()).to.be.true;
-    const initialLocation = await picker.getLocation();
-    
-    // Simulate scroll event
-    await browser.execute('window.dispatchEvent(new Event("scroll"))');
-    const newLocation = await picker.getLocation();
-    expect(newLocation.y).to.equal(initialLocation.y); // Should remain anchored correctly
-
-    // Dismiss by clicking outside (e.g. the accordion header)
     const accordionHeader = await $('#date-filter-accordion mat-expansion-panel-header');
     await accordionHeader.click();
     expect(await picker.isDisplayed()).to.be.false;
@@ -239,11 +222,10 @@ describe('Reports Sidebar Filter', () => {
     await reportsPage.setSidebarFilterBikFromDate();
     await reportsPage.setSidebarFilterBikToDate();
 
-    // 5. Reopen Off-by-one verification
+    // 5. Reopen active selection verification
     await $('#toDateFilter').click();
     // Verify that the active selected date is highlighted correctly
-    const activeCell = await picker.$('table tbody td.current-month-date.active');
-    expect(await activeCell.isDisplayed()).to.be.true;
+    expect(await reportsPage.isNepaliDatePickerActiveCellDisplayed()).to.be.true;
     
     // Dismiss it
     await browser.keys(['Escape']);
--- a/tests/e2e/default/reports/sidebar-filter.wdio-spec.js
+++ b/tests/e2e/default/reports/sidebar-filter.wdio-spec.js
@@ -197,8 +197,57 @@ describe('Reports Sidebar Filter', () => {
 
     await reportsPage.openSidebarFilter();
     await reportsPage.openSidebarFilterDateAccordion();
+
+    // 1. Max Date parity - verify future dates are disabled
+    await $('#fromDateFilter').click();
+    let picker;
+    await browser.waitUntil(async () => {
+      const pickers = await $$('.nepali-date-picker');
+      for (const p of pickers) {
+        if (await p.isDisplayed()) {
+          picker = p;
+          return true;
+        }
+      }
+      return false;
+    }, { timeout: 5000, timeoutMsg: 'Nepali date picker not displayed' });
+
+    // Future dates in the current month should be disabled (.disable)
+    const disableCells = await picker.$$('table tbody td.current-month-date.disable');
+    expect(disableCells.length).to.be.greaterThan(0);
+
+    // 2. Escape Dismissal - press Escape and verify picker is dismissed
+    await browser.keys(['Escape']);
+    expect(await picker.isDisplayed()).to.be.false;
+
+    // 3. Viewport positioning & scroll alignment
+    await $('#fromDateFilter').click();
+    expect(await picker.isDisplayed()).to.be.true;
+    const initialLocation = await picker.getLocation();
+    
+    // Simulate scroll event
+    await browser.execute('window.dispatchEvent(new Event("scroll"))');
+    const newLocation = await picker.getLocation();
+    expect(newLocation.y).to.equal(initialLocation.y); // Should remain anchored correctly
+
+    // Dismiss by clicking outside (e.g. the accordion header)
+    const accordionHeader = await $('#date-filter-accordion mat-expansion-panel-header');
+    await accordionHeader.click();
+    expect(await picker.isDisplayed()).to.be.false;
+
+    // 4. Select From and To dates
     await reportsPage.setSidebarFilterBikFromDate();
     await reportsPage.setSidebarFilterBikToDate();
+
+    // 5. Reopen Off-by-one verification
+    await $('#toDateFilter').click();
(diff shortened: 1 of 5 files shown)


tests/e2e/default/reports/sidebar-filter.wdio-spec.js
```diff
@@ -197,8 +197,57 @@ describe('Reports Sidebar Filter', () => {
...
+        if (await p.isDisplayed()) {
+          picker = p;
+          return true;
+        }
+      }
+      return false;
+    }, { timeout: 5000, timeoutMsg: 'Nepali date picker not displayed' });
+
+    // Future dates in the current month should be disabled (.disable)
+    const disableCells = await picker.$$('table tbody td.current-month-date.disable');
```
sugat009: **suggestion:** these locate elements inline in the spec (`picker.$$('table tbody td.current-month-date.disable')`, `.active`, `$$('.nepali-date-picker')`) using the plugin's structural/visual classes. Per CHT's e2e style guide, the selectors and the actions on them belong in the reports page object (`reports.wdio.page.js`), leaving only the `expect`s here, and stable `test-id`s are preferred over structural CSS classes.

tests/e2e/default/reports/sidebar-filter.wdio-spec.js
```diff
@@ -197,8 +197,57 @@ describe('Reports Sidebar Filter', () => {
...
+    await browser.keys(['Escape']);
+    expect(await picker.isDisplayed()).to.be.false;
+
+    // 3. Viewport positioning & scroll alignment
+    await $('#fromDateFilter').click();
+    expect(await picker.isDisplayed()).to.be.true;
+    const initialLocation = await picker.getLocation();
+    
+    // Simulate scroll event
+    await browser.execute('window.dispatchEvent(new Event("scroll"))');
```
sugat009: **issue:** this fires a `scroll` event without actually scrolling the page, so the picker never has a reason to move and the assertion passes regardless of the anchoring logic. To exercise re-anchoring, the page has to genuinely scroll, or better, test the reposition logic directly at the unit/cht-form level.

tests/e2e/default/reports/sidebar-filter.wdio-spec.js
```diff
@@ -197,8 +197,57 @@ describe('Reports Sidebar Filter', () => {
...
+    expect(await picker.isDisplayed()).to.be.false;
+
+    // 4. Select From and To dates
     await reportsPage.setSidebarFilterBikFromDate();
     await reportsPage.setSidebarFilterBikToDate();
+
+    // 5. Reopen Off-by-one verification
+    await $('#toDateFilter').click();
+    // Verify that the active selected date is highlighted correctly
+    const activeCell = await picker.$('table tbody td.current-month-date.active');
```
sugat009: **issue:** this asserts only that an `.active` cell is displayed, not that the highlighted cell is the selected date, so a wrong (off-by-one) highlight still passes, which is exactly what this test is meant to catch. It should assert the highlighted cell **is** the selected date. That exact-value check is far easier as a unit / cht-form test than at e2e.

webapp/tests/karma/ts/components/filters/date-filter.component.spec.ts
```diff
@@ -300,6 +302,84 @@ describe('Date Filter Component', () => {
       expect($('.nepali-date-picker-overlay')).to.have.lengthOf(0);
     });
 
+    it('should maintain independent picker elements for From and To components', () => {
```
sugat009: **suggestion:** this confirms each component rendered its own hidden input, but not that the two pickers are actually independent. Add an assertion that interacting with one (e.g. selecting a date) leaves the other unchanged.

sugat009 (changes requested):
Thanks @megha1807! The two headline fixes are solid: the from-date now asserts the full `00:00:00.000` timestamp and maxDate checks the exact value, that closes what #11246 raised. Karma green.

Requesting a couple of test improvements before merge, both from CHT's test conventions:
1. **Test level:** the maxDate-boundary, reopen-highlight, and scroll checks sit at the e2e layer, where they can only assert "something appeared." CHT's pyramid says push these down to unit / cht-form, where you can assert the exact value.
2. **Page objects:** CHT's e2e style guide keeps element selectors and the actions on them in the page object, with only the `expect`s in the spec. These new checks reach into the DOM directly. Direction inline.

tests/e2e/default/reports/sidebar-filter.wdio-spec.js
megha1807: Agreed! I have pushed this down to the unit test level as suggested.
I added a new unit test in bikram-sambat-datepicker.spec.ts (highlights the correct active day corresponding to the input value when reopened) that asserts the exact day text of the .active cell matches the input value when reopened, checking for any off-by-one errors.

The E2E test now simply acts as a high-level integration check utilizing the updated Page Object selectors.

tests/e2e/default/reports/sidebar-filter.wdio-spec.js
```diff
@@ -197,8 +197,39 @@ describe('Reports Sidebar Filter', () => {
 
     await reportsPage.openSidebarFilter();
     await reportsPage.openSidebarFilterDateAccordion();
+
+    // 1. Max Date parity - verify future dates are disabled
+    await $('#fromDateFilter').click();
```
sugat009: **suggestion (non-blocking):** Nice work moving the picker *query* selectors into the page object. These *interaction* selectors are still inline though, `$('#fromDateFilter')` here and at :215, `$('#toDateFilter')` at :226, `$('#date-filter-accordion mat-expansion-panel-header')` at :217, and each already exists in `reports.wdio.page.js` (`sidebarFilterSelectors.fromDate()` :69, `.toDate()` :68, `.dateAccordionHeader()` :66). Since `sidebarFilterSelectors` is private (not exported), add small named action helpers in the page object (e.g. `openBikFromDatePicker()` clicking `sidebarFilterSelectors.fromDate()`) and call those, so no raw selector strings live in the spec and there's one place to update if the DOM changes.

webapp/tests/karma/ts/components/filters/date-filter.component.spec.ts
```diff
@@ -300,6 +302,112 @@ describe('Date Filter Component', () => {
...
+      component.dateRange = { from: moment('2024-07-24').valueOf(), to: undefined };
+
+      component.clear();
+
+      expect(setFilter.callCount).to.equal(1);
+      expect(setFilter.args[0]).to.deep.equal([{ date: undefined }]);
+      setFilter.restore();
+    });
+
+    it('ngOnDestroy should clean up multiple pickers and overlays completely', () => {
```
sugat009, line 377: **issue (blocking):** Titled "clean up multiple pickers and overlays completely," but it destroys a single component and asserts only that one hidden input is gone (`#test-wrapper-1 …`, :402), then hand-removes the pickers and overlay itself (:405-408). Because it deletes the very elements it should verify, it can't fail even if `ngOnDestroy` left them behind. It's also a weaker duplicate of the existing `ngOnDestroy should clean up specific picker container and overlay` test (:285), which already asserts input + picker + overlay are all removed. Either drop it (the existing test covers single-component cleanup), or make it earn its name: destroy component A → assert A's input+picker gone while B's survive and the overlay remains, then destroy B → assert B's gone and the overlay is now removed. Don't hand-delete the elements under test before asserting.

sugat009 (changes requested):
Re-reviewed after the latest push, good progress: the picker query selectors are now proper page-object methods and the independence test reads well. Two notes inline, one blocking (the `ngOnDestroy` "multiple pickers" test).

tests/e2e/default/reports/sidebar-filter.wdio-spec.js
sugat009: **issue (non-blocking):** These sidebar clicks are written as raw selectors here (`:202`, `:215`, `:217`, `:226`), but the same ones already exist in the page object, `tests/page-objects/default/reports/reports.wdio.page.js`: `fromDate()` (`:69`), `toDate()` (`:68`), `dateAccordionHeader()` (`:66`). CHT's e2e convention keeps selectors out of specs. Since `sidebarFilterSelectors` is private, the tidy fix is a small action helper in the page object (e.g. an "open From picker" click) that the spec calls. Non-blocking, and you already did this correctly for the picker query selectors, this is just finishing the same pattern for the click targets.

sugat009 (commented):
One non-blocking convention note inline.

sugat009 (commented):
Re-checked after your latest push. The three click helpers you added to `tests/page-objects/default/reports/reports.wdio.page.js` are exactly the right shape, `waitForClickable()` before the click is just what this flow needs.

**issue:** they're never called yet: `tests/e2e/default/reports/sidebar-filter.wdio-spec.js` still clicks the raw selectors directly (lines 202, 215, 217, 226). And that's exactly where the red check comes from: `ci-webdriver-default-workflows_2` fails on every retry at the bare `$('#fromDateFilter').click()` with "element not interactable", because nothing waits for the accordion to finish expanding. Wiring the spec to your new helpers should fix both the convention feedback and the failing e2e in one go.

**question (non-blocking):** `tests/page-objects/default/common/common.wdio.page.js` widens the shared `waitForAngularLoaded` default from 40s to 60s, which affects every e2e login wait in the suite. The failing log doesn't show any login-wait timeout, so it doesn't look related to this PR's failure. Did you hit a real 40s timeout somewhere? If not, it's probably safer to leave the shared default alone, or pass a longer timeout only at the call site that needs it.

**suggestion (non-blocking):** same branch note as on #11255: merge the latest `10707-dmp-...` into your branch before your next push so local results match CI, and avoid force-pushing over the base merges.

sugat009 (changes requested):
Re-checked after your latest push. The helper wiring in `tests/e2e/default/reports/sidebar-filter.wdio-spec.js` and the timeout restore both look right, and thanks for merging the epic in properly.

**issue:** `ci-webdriver-default-workflows_2` still fails though, with exactly the same error as before, and the logs show it's not a missing clickable wait (no `waitForClickable` timeout appears anywhere). Look closely at the test's flow instead: step 3 dismisses the picker by clicking the date accordion **header**, and that click doesn't just dismiss the picker, it also collapses the whole date accordion. The steps after it then try to open the From/To pickers inside a collapsed panel, and that's the click that never becomes interactable. The English tests in the same file pass because they never collapse the accordion mid test. So the fix is in the flow of this test: after dismissing via the accordion header, the date accordion needs to be expanded again before selecting the dates (or dismiss by clicking something that doesn't collapse the panel).

Its `ci-integration-all-k3d` failure is unrelated infra flakiness (freetext index timeouts; the same job passed on this branch twice before); we'll re-run it.

sugat009 (commented):
Ran this branch locally to chase the `workflows_2` failure, found the root cause, and verified the fix locally before writing this: with the three changes below, the whole spec passes deterministically (all 6 tests, zero retries).

Good news first: your accordion fix works, the "element not interactable" error is gone, and dismissing via the freetext input was a nice touch. What remains is something your new assertion uncovered rather than caused.

**issue:** the To date never actually gets selected. In `tests/page-objects/default/reports/reports.wdio.page.js`, `setSidebarFilterBikDate` clicks `cells[cellIndex]` among the enabled days of the current month, and `setSidebarFilterBikToDate` passes `cellIndex = 20`. But the picker disables everything after today (`maxDate`), so `cells.length` equals today's day of the Nepali month, and on any day before the 21st `cells[20]` doesn't exist. The guard `if (cells.length > cellIndex)` then silently skips the click: no error, no selection, `date.to` never set. That's why the highlighted-day assertion at `sidebar-filter.wdio-spec.js:227` reads false: the step-5 reopen seeds an empty input. Note the date dependence: this goes green from day 21 to the end of every Nepali month and red again when the next month starts, with no code change. This helper predates your PR; your assertion is what finally exposed it. Two changes:
1. For the To pick, select the **last enabled cell** (today) instead of a fixed index, it exists on every day of the year. Careful with the tempting alternative of navigating to the previous month like the From flow does: the seeded fixture reports move on the *Gregorian* calendar ("one month ago") while the picker navigates *Bikram Sambat* months, so a previous-BS-month To bound lands before the fixtures on some run dates and the final count assertion breaks. Today's date always sits after them.
2. Make the helper **fail loudly** when the requested cell doesn't exist instead of silently skipping, so the failure surfaces at the selection itself with a clear message rather than 40 lines later as `expected false to be true`. The silent skip is what kept this invisible.

**issue:** related, `isNepaliDatePickerActiveCellDisplayed()` queries `$('.nepali-date-picker')`, which always returns the first picker in the DOM (the From one). From and To each create their own picker element and closed pickers stay in the DOM hidden, so it can inspect the wrong, hidden picker. Make it find the **visible** picker, your own `waitUntil` loop inside `setSidebarFilterBikDate` already does exactly this, so it's just applying the same pattern.

**suggestion (non-blocking):** the test's final report-count assertion passes even when the To bound was never applied (a From-only filter yields the same 2 reports), which is exactly how the dead To-selection stayed invisible. Worth asserting something that only holds when both bounds are applied. Also, the step-1 `disabledCells > 0` assertion will fail on the last day of a Nepali month (nothing left to disable), a rare but real date edge worth guarding.

sugat009 (commented):
The core fix landed exactly right, last enabled cell, loud errors, visible-picker helpers, nicely done. Three things surfaced in this push though, two need changes and one is just noise:

**issue:** the new step-4 guard is what's failing CI now, not the fix itself. `expected '' to not equal ''` comes from `getFromDateValue()` in `tests/page-objects/default/reports/reports.wdio.page.js`: `#fromDateFilter` is an anchor element, not an input, so `getValue()` always returns an empty string no matter what was selected, and the assert fails on every run before step 5 is even reached. To verify a selection really happened, check something that actually changes: the button's visible label text (it shows मिति when empty and the selected BS date after picking), or the value of the hidden `.nepali-datepicker-input` that each date filter appends inside its own container.

**issue:** the merge conflict resolution in `webapp/tests/karma/js/enketo/widgets/bikram-sambat-datepicker.spec.ts` kept only the incoming side, which dropped this PR's own two tests: "highlights the correct active day corresponding to the input value when reopened" and "updates the position of the calendar popup on window scroll when positioned as anchored". The file no longer appears in this PR's diff at all. Both sides' tests belong in the file, so please restore yours on top of what came in from the epic (they're intact at your earlier head `5dd4738a8` for easy copying back).

**suggestion (non-blocking):** heads up on the new last-day-of-month fallback in `tests/e2e/default/reports/sidebar-filter.wdio-spec.js`: it reads the active cell of a freshly opened, empty picker, but the picker clears the active highlight when its input is empty, so on the very day that branch is meant to cover there may be no active cell to read. Worth deriving "today is month end" from a different signal.

The other two failures in the same job (`bulk-delete`, `infinite-scrolling`) are unrelated specs that look like runner flakes, and `Lint PR title` died on an npm error rather than your title; those just need a re-run.

PR #11258, opened 2026-07-17
fix(#11252): guard BS datepicker against invalid day value crash

### LINKED ISSUE #11252: Bikram Sambat datepicker offers non-existent days (e.g. मंसिर ३० २०८२) and throws Invalid day value

megha1807:
# Description

Implements a conversion guard for the Bikram Sambat datepicker (reports date filter and Enketo form widget) to validate day selections and manual entries against `daysInMonth` using the `bikram-sambat` library, preventing crashes on invalid days (e.g. Mansir 30, 2082).

## Changes:
- **Date Filter Component**:
  - Guarded calendar datepicker selections with a `daysInMonth` check. Clicks on invalid days are safely ignored.
  - Added unit test asserting invalid day selections are ignored.
- **Enketo Datepicker Widget**:
  - Guarded calendar picker selection with `daysInMonth` check.
  - Updated the manual input `clearIfIncomplete` guard to clear the day input field if the entered day is invalid for the specified month/year, preventing under-the-hood crashes.
  - Added unit tests asserting invalid day calendar select is ignored and invalid manual entries reset/clear the fields.

## Issue Number

Fixes #11252

# Code review checklist

### DIFF
--- a/webapp/src/js/enketo/widgets/bikram-sambat-datepicker.js
+++ b/webapp/src/js/enketo/widgets/bikram-sambat-datepicker.js
@@ -6,6 +6,7 @@ const bikram_sambat_bs = require( 'bikram-sambat-bootstrap' );
 const eurodigit = require( 'eurodigit' );
 const toDevanagari = eurodigit.to_non_euro.devanagari;
 const fromDevanagari = eurodigit.to_euro;
+const BikramSambat = require( 'bikram-sambat' );
 
 const {
   setupNepaliDatePicker,
@@ -94,13 +95,27 @@ class Bikramsambatdatepicker extends Widget {
         // setTimeout) after both "change" and "blur", once the library's
         // own handlers have finished running.
         const clearIfIncomplete = () => {
-          const day   = $parent.find( 'input[name="day"]' ).val();
-          const month = $parent.find( 'input[name="month"]' ).val();
-          const year  = $parent.find( 'input[name="year"]' ).val();
+          const dayDev   = $parent.find( 'input[name="day"]' ).val();
+          const monthDev = $parent.find( 'input[name="month"]' ).val();
+          const yearDev  = $parent.find( 'input[name="year"]' ).val();
 
-          if ( !day || !month || !year ) {
+          if ( !dayDev || !monthDev || !yearDev ) {
             $realDateInput.val( '' );
             $realDateInput.trigger( 'change' );
+            return;
+          }
+
+          const day = Number(fromDevanagari(dayDev));
+          const month = Number(fromDevanagari(monthDev));
+          const year = Number(fromDevanagari(yearDev));
+
+          if (!Number.isNaN(day) && !Number.isNaN(month) && !Number.isNaN(year)) {
+            const maxDays = BikramSambat.daysInMonth(year, month);
+            if (day < 1 || day > maxDays) {
+              $realDateInput.val( '' );
+              $realDateInput.trigger( 'change' );
+              $parent.find( 'input[name="day"]' ).val( '' );
+            }
           }
         };
 
@@ -154,6 +169,11 @@ const setupCalendarPicker = ($parent, widget) => {
       const year = data.bsYear;
       const monthNum = data.bsMonth;
       const day = data.bsDate;
+
+      // Guard against invalid days (e.g. Mansir 30, 2082)
+      if (day > BikramSambat.daysInMonth(year, monthNum)) {
+        return;
+      }
       const monthName = NEPALI_MONTH_NAMES[monthNum - 1];
       if (monthName) {
         // Update the input fields with Devanagari digits and trigger change
--- a/webapp/src/js/enketo/widgets/bikram-sambat-datepicker.js
+++ b/webapp/src/js/enketo/widgets/bikram-sambat-datepicker.js
@@ -109,6 +109,7 @@ class Bikramsambatdatepicker extends Widget {
 
           if (!Number.isNaN(day) && !Number.isNaN(month) && !Number.isNaN(year)) {
             const maxDays = BikramSambat.daysInMonth(year, month);
+            // Guard against invalid day values on manual entry (see #11252)
             if (day < 1 || day > maxDays) {
               $realDateInput.val( '' );
               $realDateInput.trigger( 'change' );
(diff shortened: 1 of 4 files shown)


sugat009 (commented):
Reviewed this along with the merge conflict that appeared after #11255 and #11257 landed. The substance is right: the epic's widget currently has no day-validity check at all, so these guards are the missing piece for #11252, not a duplicate, and they cover all three entry points (manual typing, widget picker, reports filter). The behavior also lines up with the existing divergent-date test's expectations (real date input ends up empty), so the two test sets can coexist.

**On the conflict:** merge the latest `10707-dmp-...` into your branch. Only one file conflicts, `webapp/tests/karma/js/enketo/widgets/bikram-sambat-datepicker.spec.ts`, because the epic now carries the tests from both #11255 and #11257 at the same spot where yours are added. The resolution is keep everything: all the incoming tests plus your two new ones appended after them. Same heads-up as last time: after resolving, double-check your two test titles are still in the file, this exact resolution silently dropped tests once before. Your tests use the shared helpers (`initWidget`, `dayInput`, `realDateInput`, and so on) which still exist in the epic's version, so they'll compile fine.

**suggestion (non-blocking):** the guards silently `return` when the picker hands over a non-existent day, so a user clicking a visible (phantom) day gets no feedback. That's fine as a stopgap since the real fix (the forked calendar with aligned month data) removes the phantom days at the source, but worth a short code comment referencing #11252 so future readers know why the ignore is intentional.

**suggestion (non-blocking):** in "ignores dateSelect event with an invalid/non-existent day", the fields start empty and the test asserts they stay empty, which can't distinguish "the guard ignored the event" from "the handler cleared everything". Seeding a valid date first and asserting it survives the invalid event would make the test bite.

sugat009 (changes requested):
Verified this branch locally, in karma and by hand in the app. The core approach is right and #11252 is genuinely fixed: phantom days like मंसिर ३० no longer throw, the `#11252` comments are helpful, and last round's test-quality changes all landed well. One blocker though, a new one this PR introduces.

**issue (blocking):** the first click on a *valid* date is silently swallowed. Clicking the identical date a second time works. Repro in the BS Date Test form:

1. Pick any day in **मंसिर** (say ८) and close the calendar
2. Reopen it, switch to **बैशाख**, click **३१** (valid, Baisakh has 31 days) → the गते field is **empty**, and the underlying date is empty too, so submitting records no date
3. Click **३१** again → now it works

https://github.com/user-attachments/assets/a02715d8-0948-4e7b-b858-9930e83083f9

The cause is in `webapp/src/js/enketo/widgets/bikram-sambat-datepicker.js`: `onDateSelect` writes the fields one at a time (lines 181-183) and each `.trigger('change')` runs `clearIfIncomplete()` synchronously. On the first write the DOM still holds the **old month**, so the guard checks ३१ against मंसिर's 29, calls it invalid, and clears the day (line 116). The second click then succeeds precisely because the first one already updated the month to बैशाख, which is what confirms the stale-month read is the culprit.

Verified with the same karma test on both branches: it passes on the epic and fails here, so it's specific to this change rather than pre-existing.

The fix is to validate the selection being applied rather than whatever is currently in the fields, the incoming `{bsYear, bsMonth, bsDate}` is already right there at the top of `onDateSelect`, so checking that directly (or writing all three fields before validating) avoids the stale read entirely.

**suggestion (non-blocking):** `BikramSambat.daysInMonth()` itself throws for BS years outside 1970-2090 (try typing year `२०९१`), and the three new calls aren't wrapped. That crash already exists today via the library's own converter, so it isn't something you introduced, but given this PR's whole job is stopping unhandled conversion errors, a small try/catch would close it cheaply.

**suggestion (non-blocking):** worth adding tests for `day === daysInMonth` (the boundary that must be *accepted*) and a two-selection case like the repro above, that's exactly the gap that let this through. Also `date-filter.component.ts:137` is the only one of the three guards without an explanatory comment.

sugat009 (approved):
Verified locally, in karma and by hand in the app. The regression is gone: the first click on a valid date now takes, sequential selections across months work, the phantom-day guard still holds (मंसिर ३० is silently ignored with no console error), and an out-of-range year now warns instead of throwing.

Setting all three fields before triggering change is the right fix, it removes the stale read rather than working around it. Nice work adding the boundary and sequential-selection tests too, those are exactly the cases that would have caught this originally, and the lint/SonarQube cleanup on top is a good touch.

PR #11274, opened 2026-07-22
fix(#11242): keep mobile date tooltip on-screen

### LINKED ISSUE #11242: Mobile date tooltip clipped off the left edge of the viewport (report detail)
**Describe the bug**
On touch devices (including the CHT Android app), the accessible tooltip used to reveal a relative date's absolute value is **clipped off the left edge of the viewport** and only the tail end of the text is visible. On the report detail page, focusing the date/time in the "Submitted by" line shows something like `5:30 AM` instead of the full `Tue, Jul 7th, 2026 at 5:30 AM`. The beginning of the string is pushed off-screen.

**To Reproduce**
1. Use the webapp on a touch / no-hover viewport: a phone, or Chrome DevTools **Responsive** view.
2. Open any submitted report's detail view so the "Submitted by [name] [phone]" summary line is visible.
3. Focus the relative date/time text after that line: tap it on a touch device, or tab to it.
4. The tooltip appears anchored to the date element, but its left portion is off-screen: only the tail (e.g. `5:30 AM`) is readable.

**Expected behavior**
The tooltip should display the complete date/time string within the visible viewport.

**Logs**
N/A

**Screenshots**

Chrome (responsive mode):

Image

CHT Android:
Image

**Environment**
- Instance: any
- Browser: Any touch device (or responsive mode on Chrome)
- Client platform: Android, Mobile view
- App: webapp
- Version: Found in 5.1.1, also present in 4.22.0

**Additional context**
- Found while investigating a separate Bikram Sambat date bug (#11241); this tooltip issue is independent of language and of that date bug.
- On desktop the native tooltip is used and is not affected.
- It works fine for messages (sent date) and tasks (due date).

binokaryg:
# Description

On touch devices the date/time tooltip — the accessible way to reveal a relative date's full value — was **clipped off-screen**: only the tail of the string was readable (e.g. `5:30 AM` instead of `Tue, Jul 7th, 2026 at 5:30 AM`). Most visible on the report detail "Submitted by …" line. Not language-specific — reproduced in English, Nepali, and Arabic (RTL).

Fixes #11242

**Root cause.** The touch tooltip was a CSS pseudo-element in `webapp/src/css/tooltip.less` (`[title]:focus::before`, with `content: attr(title)`; desktop keeps the native `title`). It was anchored to a fixed edge (`right: -10px`, `width: max-content`), so it grew off the viewport, and it was also cropped by ancestor `overflow: hidden` scroll containers. No fixed CSS anchor can satisfy every case — the trigger can sit at either edge, the text can be wider than the space beside it, and RTL flips which edge is which. (A series of scoped CSS overrides was prototyped and each broke another context; see the [issue thread](https://github.com/medic/cht-core/issues/11242#issuecomment-4999339385).)

**Fix.** Replace the CSS pseudo-element with `MobileTooltipDirective`, which renders the `title` in a **CDK overlay** at the document level:
- positioned with a flexible strategy (`withPush`) so it stays inside the viewport regardless of the trigger's position, the text length, or the direction;
- rendered above the page, so it escapes ancestor `overflow` clipping;
- direction resolved from the trigger, so `start`/`end` placement and bidirectional text are correct in RTL.

It runs only on touch / no-hover devices (desktop is untouched — native `title` on hover). It's hosted on the webapp app root and on the cht-form root, and the old CSS mechanism (plus its per-context `min-width` / `overflow: visible` / first-row band-aids) is removed.

**Why a document-level directive and not `matTooltip`?** The date tooltips aren't real elements a directive can bind to — the date pipes (`relativeDate`, `taskDueDate`, …) build an HTML string with the `title` inside it, injected via `[innerHTML]` in 9 places, and Angular directives can't attach to innerHTML-injected DOM. Using `matTooltip` would mean converting that shared date rendering into a component, reworking the interval-based relative-date live-update (`RelativeDateService.updateRelativeDates`), updating the 9 call sites, and rewriting the pipe tests — a broad, higher-risk refactor of a path used everywhere dates appear. This directive instead reads the `title` already in the DOM via one listener, leaving the pipes, templates, and live-update untouched. CDK itself is already a first-class dependency (`@angular/cdk` in `webapp/package.json`; every modal uses `MatDialog`, which is CDK Overlay) — this is just the first place we use the Overlay API directly rather than through Material.

**Behavior on touch:**
- Focusing a date shows the tooltip — so a **long-press** peeks the full date without navigating.
- It dismisses on blur, on scroll, when the trigger is removed from the DOM (e.g. the sidebar last-sync span re-rendering on a sync), and — via an `IntersectionObserver` — when the trigger **leaves the viewport**. On a single-pane layout, opening a report from the list slides the still-focused list off-screen, so the tooltip clears instead of hovering over the report; on a two-pane layout the list (and the tooltip) stay put. Watching *visibility* rather than router navigation is deliberate: every list-row tap fires a navigation event whether the trigger stays on screen or not, so navigation was the wrong signal.

# Code review checklist

# AI disclosure

Per the [AI contribution guidelines](https://docs.communityhealthtoolkit.org/community/contributing/ai-guidelines/):

- **Tool:** Claude Code (Anthropic).
- **How it was used:** investigating the clipping and its history, evaluating approaches (a CSS-anchor fix vs. the CDK overlay), implementing the directive and its tests, and iterating on the touch dismissal behavior. A code review of the directive (also AI-assisted) surfaced several issues — zone/CD cost, RTL wiring, dismissal on DOM-detach and scroll, and the cht-form gap — which were addressed. All changes were reviewed and tested by me.

# License

Screenshots:
### Report detail (English)
|Before|After|
|:-:|:-:|
| Screenshot_20260806_092834_CHT Android | Screenshot_20260806_094300_CHT Android |

### Report detail (Arabic)
|Before|After|
|:-:|:-:|
Screenshot_20260806_093132_CHT Android|Screenshot_20260806_094533_CHT Android

### Report list (Arabic)
|Before|After|
|:-:|:-:|
 Screenshot_20260806_093125_CHT Android | Screenshot_20260806_094528_CHT Android |


sugat009 (approved):
**praise:** looks good. Ran the tests locally and spot-checked that they actually fail when the guards are removed.

**question:** the one behaviour with no assertion is the one the issue is about, that the tooltip stays on-screen. It's delegated to CDK's `withPush`, so pinning it would partly be testing the framework and layout assertions tend to be flaky headless. Worth a single edge-case assertion, or is trusting CDK the right call here? Your judgment, I don't think the suite is weak for lacking it.

**note:** with two hosts (webapp root plus an embedded `cht-form`), a focusin calls `overlay.create` twice and disposes one immediately, 2 creates for 1 surviving tooltip. Behaviour is correct and tested; just noting the redundant work in case it's cheap to short-circuit.

**question:** the old CSS also set `justify-content: left` on every `[title]` under touch, which the new block drops. I couldn't find anything relying on it (`.relative-date` isn't a flex container), but it's a layout property rather than tooltip styling, so worth a glance on a device if you haven't already.

binokaryg:
Thanks for the review @sugat009!

**On-screen assertion**: I have added a new test (https://github.com/medic/cht-core/pull/11274/changes/9f356dbbabc83df614f9a234e9c668dcdd45d4af) that pins the contract rather than CDK internals:
- trigger fixed at the right viewport edge
- long title
- assert the tooltip lands beside the trigger, and
- fits within `[0, innerWidth]`

Two notes:
- the karma build loads no app CSS (`styles: []`), so the test injects the CDK overlay structural styles the app gets from the Material prebuilt theme
- CDK applies the position one render tick after attach, so it uses the same bounded-poll pattern as the `IntersectionObserver` tests

**Double create with two hosts**: `show()` now returns early when the target's tooltip is already up (only the multi-instance fan-out hits that path; the tap-to-reshow handler already guards separately). The multi-instance test now also spies on `Overlay.create` and asserts a single call. Verified manually too, with two embedded `<cht-form>`s and a `MutationObserver` on the overlay container: one create per show.

**`justify-content: left`**:  Checked, but nothing relied on it. No element that carries a `title` is a flex container, so the property had no effect anywhere (and `left` behaves as the default `flex-start` anyway). Also re-checked list rows and the sidebar on-device after the removal; no layout change.

PR #11275, opened 2026-07-22
feat(#10706): add api endpoints for advanced contact management operations

### LINKED ISSUE #10706: [DMP 2026]: Add API endpoints for advanced contact management operations

sugat009:
# Description

_This PR is for tracking progress of #10706 , More info will be added later._

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10706-dmp-2026-add-api-endpoints-for-advanced-contact-management-operations/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10706-dmp-2026-add-api-endpoints-for-advanced-contact-management-operations/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10706-dmp-2026-add-api-endpoints-for-advanced-contact-management-operations/docker-compose/cht-couchdb-clustered.yml)

# License


vikrantwiz02:
Hi @sugat009, thanks for syncing this. I reproduced the five conflicts locally by merging current master into the feature branch and verified a resolution, so hopefully this saves you some digging.

The cause is that master gained the scheduled archiving feature after this branch forked (`feat(#6615): adds archiving` in #11139), while the branch already carried its own copy. Three of the conflicts are `add/add` on files both sides created, which looks worse than it is: the two `archiving.js` implementations are 348 and 350 lines and differ only in that the feature branch's `archiveBatch` returns the skipped ids and is exported. It is a strict superset, so nothing from master is lost by keeping it.

Resolutions, merging master into the feature branch:

- `sentinel/src/lib/archiving.js` — keep the feature branch version. It is master's implementation plus `archiveBatch`, which the bulk archive action needs.
- `sentinel/tests/unit/lib/archiving.spec.js` — keep the feature branch version. It has master's coverage plus the `archiveBatch` and `canArchive` cases.
- `tests/integration/sentinel/schedules/archive.spec.js` — keep the feature branch version. It uses the shared `utils.archiveDb`, which master exports.
- `sentinel/src/db.js` — keep **one** `medicLogs` test stub and **one** real `medicLogs` PouchDB instance, at master's positions, and keep `archive`. Both sides added `medicLogs`, so a straight merge leaves it assigned twice. It is redundant rather than harmful, since the second assignment overwrites the first with an equivalent value, but worth cleaning up while we are in here.
- `shared-libs/constants/src/index.js` — union of both sides, with only one `UI_EXTENSION`. Master's side of the conflict re-adds it even though it is already present above the marker.

I ran the resolved tree locally:

- archiving unit tests: 39 passing
- bulk-operations tests: 28 passing
- full `unit-sentinel`: 312 passing, coverage above every threshold
- ESLint clean on all five files

One request: could you keep this as a draft after resolving? #11374 and #11355 both target this branch and should land before it goes to master.

Also worth knowing, the two red checks are from the 26 August run and are a cascade of `502 Bad Gateway` from nginx rather than anything in the diff, so they should clear on a rerun.

I have the resolution locally but have not pushed it, since pushing to my fork would not update this branch. Happy to push it as a temporary comparison branch if that is useful.

PR #11293, opened 2026-07-23
ci(#11292): add AndraBot to static check external contributor PRs

### LINKED ISSUE #11292: Develop a GH action BOT that adds automated messages to PRs
**Describe the issue**
This bot should cover PRs that are opened by external contributors, and check whether: 
- the description matches the PR template
- it is linked to close an issue
- the issue the pr is linked to is assigned to the author

The bot should add explanatory messages to the PR, and if a check is failed, the PR should be changed to a draft or tagged with a label for visibility. 
When all checks are passed, the PR can be reviewed.

dianabarsan:
# Description

Adds **AndraBot**, a GitHub Action that checks pull requests opened by external contributors and guides them through our contribution requirements before review.

On `opened` / `edited` / `synchronize` / `reopened` / `ready_for_review`, for PRs whose author is not an OWNER / MEMBER / COLLABORATOR (bots and draft PRs are also skipped), it checks that:

- the PR description follows the PR template — required sections are read from  `.github/PULL_REQUEST_TEMPLATE.md` at run time, so template changes work without touching the script. Sections that are empty in the template (e.g. `Description`) must be filled in; pre-filled sections must be present.
- the License section is unchanged from the template (whitespace-insensitive).
- an issue is linked via GitHub's closing-keyword/Development linkage (GraphQL `closingIssuesReferences`). Issues in other repos of the same org count; issues outside the org don't.
- the author is assigned to at least one linked issue.

When any check fails, the bot posts a single comment listing everything that needs fixing, labels the PR `Waiting for contributor` and fails the check. Once everything passes, the comment is replaced with a success message and the label is swapped for `Ready for review`. Compliant PRs never get a comment — just the `Ready for review` label — so reviewers can filter external PRs that have cleared the gate.

The bot maintains a single comment: when its content changes it is deleted and re-posted (comment edits do not notify the author, new comments do), and when a re-run produces identical content the comment is left untouched, so authors get exactly one notification per state change.

All comment texts live as markdown files in `scripts/ci/andra-bot-messages/` with `{{placeholder}}` substitution, so wording can be edited without code changes (see the README there).

## Implementation notes

- The workflow uses `pull_request_target` because fork PRs only get a read-only token under `pull_request`, so the bot couldn't comment. It never checks out or executes PR code — since 2025-12-08 `pull_request_target` resolves the workflow file and checkout to the default branch (sparse: `scripts/ci` and `.github` only) and all checks operate on API data. A documented zizmor `dangerous-triggers` suppression covers this (ignore entries must be workflow basenames — the pre-existing path-style entries in `.github/zizmor.yml` may be silently inert, worth a separate look).
- Top-level permissions are empty; the job gets only `contents: read`, `issues: read`, `pull-requests: write`. Everything the bot does works with the default `GITHUB_TOKEN` — no PAT or app required. (Draft conversion was considered and dropped: the draft-state mutations are not available to app installation tokens, so labels are used instead.)
- Label operations never fail the run — errors degrade to a workflow warning. Note that adding a label that doesn't exist in the repo silently creates it (default colour, no description), so the `Waiting for contributor` and `Ready for review` labels should be pre-created with a colour and description.
- Comments are authored by `github-actions[bot]`; the AndraBot name appears as the check name. (A GitHub App would be needed for a real bot identity.)
- Unit tests in `webapp/tests/mocha/unit/testingtests/andra-bot.spec.js` (40 cases) cover all checks, the comment lifecycle and the label swaps, asserting against the real PR template and rendered message files so future edits to either don't break tests spuriously.

## AI disclosure

Heavily used Claude for development.

closes #11292

# Code review checklist

- [x] AI disclosure: Please disclose use of AI per [the guidelines](https://docs.communityhealthtoolkit.org/community/contributing/ai-guidelines/). Written with Claude Code (supervised: design directed, code reviewed, tests verified by the author).

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:pr-bot/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:pr-bot/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:pr-bot/docker-compose/cht-couchdb-clustered.yml)

# License


dianabarsan:
PR with no description at all: 

image

dianabarsan:
PR with linked assigned issue, but no matching description. 

image

dianabarsan:
PR linked to unassigned issue: 
image

dianabarsan:
PR matching description but not linked to issue: 
image

dianabarsan:
Ready for review: 

image

dianabarsan:
License mismatch: 

image

dianabarsan:
@andrablaj mind giving the general functionality a look?

scripts/ci/andra-bot-messages/missing-linked-issue.md
```diff
@@ -0,0 +1 @@
+This PR is not linked to an issue. Please [link the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) you are fixing by adding a closing keyword to the PR description, e.g. `Closes #1234`.
```
andrablaj:
```suggestion
This PR is not linked to an existing issue. Please [link the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) you are fixing by adding a closing keyword to the PR description, e.g. `Closes #1234`.
```

andrablaj: Suggestion to add something about opening an issue if it doesn't exist, as asked to be assigned to it.

andrablaj (commented):
Thanks @dianabarsan, this looks good! 😃

I have some questions:

- Draft PRs: I think we should check a PR's draft state. So a draft PR that happens to pass all four checks would still get labeled `Ready for review` (right?), which seems off given the goal is to help reviewers filter PRs actually meant for review. Maybe checks should still run on drafts (probably good, gives early feedback) but the `Ready for review` label should wait until the PR is actually marked ready?
- Label swap in both directions: the description covers going from failing to passing, but what about the reverse? If a PR already has `Ready for review` and a later push (or an unassigned issue) makes it fail checks again, does the bot pull the `Ready for review` label back off and reapply `Waiting for contributor`? 
- Issue in another repo: GitHub lets a PR close an issue that lives in a different repo. Is the assignee check looking at the issue in its own repo, or would it assume the same repo?

dianabarsan:
Thanks for the review @andrablaj 

> Draft PRs: 

Good point, I'll add a skip for drafts?

> label swap

Label swap goes in both directions already. 

> GitHub lets a PR close an issue that lives in a different repo.

I'll add a check that the linked issue is in a `medic` repository. Great point here. Sometimes we link PRs cross repos and we don't duplicate the issue (because issue context likely lives in one repo and GH already allows this seamlessly).

andrablaj (approved):
> Good point, I'll add a skip for drafts?
I think that's a good approach. We can change it if we encounter other blockers. Thanks!

This looks good from the functionality perspective. 🤖 I haven't checked the code.

sugat009 (commented):
Nice piece of work, and it holds up under poking.

Findings are inline. One that has no line to attach to: the description says "29 cases" in `webapp/tests/mocha/unit/testing-tests/`, but the suite runs 40 and the directory is `testingtests`.

For context on the inline notes that quote frequencies, I replayed these checks against every external-contributor PR from the last 12 months (293 of them, roughly 24 a month, which is what this will gate) and reproduced the behavioural ones on a scratch repo.

scripts/ci/andra-bot.js
```diff
@@ -0,0 +1,277 @@
...
+ *
+ * The message texts live in ./andra-bot-messages/ so they can be edited without touching this script.
+ */
+
+const fs = require('node:fs');
+const path = require('node:path');
+
+const COMMENT_MARKER = '<!-- andra-bot -->';
+const FAILURE_LABEL = 'Waiting for contributor';
+const SUCCESS_LABEL = 'Ready for review';
```
sugat009, line 20: **question (non-blocking):** should these two labels be created ahead of time? Neither exists in the repo yet. `addLabels` doesn't reject an unknown label, it silently creates one, so each will appear the first time it's needed (`Waiting for contributor` on the first failing external PR, `Ready for review` the first time one passes everything, since `removeLabel` never creates). They'll look like cht-core's `semantic-release` label, which is `#ededed` with no description because it was auto-created the same way. Pre-creating both with a colour and description would avoid that.

Minor doc note: this also means "a missing label degrades to a workflow warning" in the description doesn't hold for the add path, since nothing comes back for `core.warning` to catch.

scripts/ci/andra-bot-messages/not-assigned.md
```diff
@@ -0,0 +1 @@
+You are not assigned to the linked issue ({{issueList}}). Please comment on the issue asking to be assigned to it, so we know the work is not being duplicated.
```
sugat009: **question (non-blocking):** how does a contributor clear this check once they've been assigned? Issue assignment fires none of the triggers in `.github/workflows/andra-bot.yml:9-14`, so the check stays red until someone touches the PR. Verified on a scratch repo: assigning the author to the linked issue produced no workflow run at all.

The outro does say checks re-run when the PR is edited or updated, but this bullet doesn't connect "you've been assigned" to "so push or edit something". Re-running the job also works, but only for someone with write access, so it isn't a route the contributor has.

This is the most-reached path of anything I found: 62 of the last 12 months' 293 external-contributor PRs (21%) are linked to an issue their author isn't assigned to. Worth noting an `issues: [assigned]` trigger wouldn't fully fix it either, since `getLinkedIssues` deliberately accepts issues in other medic repos.

scripts/ci/andra-bot.js
```diff
@@ -0,0 +1,277 @@
...
+
+const HEADING_PREFIX = '# ';
+const HEADING_REGEX = new RegExp(`^${HEADING_PREFIX}.+$`, 'gm');
+
+const parseSections = (text) => {
+  text = stripComments(text);
+  const headings = [...text.matchAll(HEADING_REGEX)];
+  return headings.map((match, index) => ({
+    heading: match[0].trim().replace(HEADING_PREFIX, ''),
+    content: normalize(text.slice(match.index + match[0].length, headings[index + 1]?.index)),
```
sugat009, line 53: **issue (non-blocking):** this makes the License check report false positives on ordinary PR descriptions. `headings[index + 1]?.index` is `undefined` for the last section, so License (being last in the template) absorbs everything to the end of the body. Anything a contributor writes below the licence paragraph becomes section content and fails the exact-match at line 92.

Replaying the checks over the last 12 months of external PRs: **4 of the 5 PRs that would be told "the License section has been modified" have completely intact licence text.** Two appended screenshots, one pasted a commit message, one added "please take a look on this PR". Only one was a genuine licence edit. Those four contributors would be asked to restore text they never touched. (PR numbers 10746, 10745, 10740 and 10340 if you want to check them; written unlinked to avoid cross-referencing the old threads.)

Bounding the last section, or ignoring trailing content once the licence paragraph matches, would fix it.

scripts/ci/andra-bot.js
```diff
@@ -0,0 +1,277 @@
...
+  let previous;
+  do {
+    previous = text;
+    text = text.replace(/<!--[\s\S]*?-->/g, '');
+  } while (text !== previous);
+  return text;
+};
+
+const HEADING_PREFIX = '# ';
+const HEADING_REGEX = new RegExp(`^${HEADING_PREFIX}.+$`, 'gm');
```
sugat009, line 46: **issue (non-blocking):** related to the boundary problem at line 53, and worth fixing together. This has no code-fence awareness, so a fenced block containing a `# License` line at column 0 is parsed as a real section, failing both the ordering compare at line 72 and the licence check at line 92. A compliant PR gets told it broke the template *and* changed the licence, with nothing pointing at the real cause.

Reproduced end to end on a scratch repo: a passing PR flipped to failing on both checks after editing nothing but the description to add the fence. Tilde fences behave identically; indented and blockquoted variants are unaffected, and a non-colliding line like `# npm ci` is fine.

Frequency, in fairness: zero of the 293 historical PRs hit this one, unlike the line 53 variant. Note the duplicate-License rule itself is deliberate anti-evasion per `andra-bot.spec.js:248`, so the fix belongs in fence handling rather than in relaxing that rule.

scripts/ci/andra-bot.js
```diff
@@ -0,0 +1,277 @@
...
+    });
+  }
+  await github.rest.issues.createComment({
+    ...context.repo,
+    issue_number: context.payload.pull_request.number,
+    body,
+  });
+};
+
+const hasLabel = (pr, name) => pr.labels.some(label => label.name === name);
```
sugat009: **issue (non-blocking):** reading `pr.labels` from the event payload rather than current state lets label state desync. Whenever that snapshot is stale, `removeLabel` early-returns while `addLabel` proceeds.

Reproduced first try: re-running the run that itself added `Waiting for contributor` replays a payload where the PR was still unlabelled, so the PR ends up carrying **both** labels, with a green check and the success comment.

It isn't re-run-only, and it's symmetric. A contributor who opens a non-compliant PR and fixes the description within seconds gives the `edited` run a payload predating the `opened` run's label write, which under `cancel-in-progress` is a realistic path. And the stuck label can just as easily be `Ready for review` on a PR that now fails, which is the worse direction since it mislabels a PR as review-ready.

Reading current labels from the API rather than the payload fixes both directions at once.

.github/workflows/andra-bot.yml
```diff
@@ -0,0 +1,43 @@
+name: AndraBot
+
+# pull_request_target is required so the workflow has a write token to comment
+# on PRs from forks. It is safe here because the workflow never checks out or
+# executes code from the PR — only the base branch is checked out, and the
+# checks operate purely on API data.
+on:
+  pull_request_target:
+    types:
```
sugat009, line 10: **question (non-blocking):** should this gate PRs targeting non-default branches? 49 of the last 12 months' 293 external-contributor PRs (17%) target feature or epic branches rather than `master`, including the DMP epics, where a mentor is already tracking issue assignment.

It applies to all of them rather than some, since the workflow always comes from the default branch. A `branches: [master]` filter here would make the intent explicit if gating `master` only is what you want.

scripts/ci/andra-bot.js
```diff
@@ -0,0 +1,277 @@
...
+    await github.rest.issues.deleteComment({
+      ...context.repo,
+      comment_id: existingComment.id,
+    });
+  }
+  await github.rest.issues.createComment({
+    ...context.repo,
+    issue_number: context.payload.pull_request.number,
+    body,
+  });
```
sugat009: **question (non-blocking):** should this tolerate a failed write? The delete and create are two calls with no `try`/`catch`, unlike `addLabel` and `removeLabel` which degrade to a `core.warning`. If the create fails, no comment remains and the label updates are skipped too. The sharpest version is the success path at lines 262-269, where that turns a *passing* PR red with the failure label still on it.

Unlikely enough that this is purely robustness. One caveat if you do change it: a bare reorder isn't sufficient, since `findExistingComment` at lines 235-244 returns the oldest marker comment, so create-first plus a failed delete leaves a duplicate that the `sameContent` check then preserves indefinitely.

.github/zizmor.yml
```diff
@@ -7,6 +7,15 @@
 # To add a new suppression, document the reason and the risk acceptance date.
 
 rules:
+  dangerous-triggers:
+    ignore:
+      # andra-bot.yml uses pull_request_target because it needs a write token to
+      # comment on PRs from forks. It never checks out or executes PR code —
+      # checkout targets the base branch (the pull_request_target default) and
```
sugat009: **nitpick:** for `pull_request_target`, the workflow file and checkout now always come from the repository's **default** branch, regardless of the PR's base branch, per [this changelog](https://github.blog/changelog/2025-11-07-actions-pull_request_target-and-environment-branch-protections-changes/) effective 8 December 2025. So "checkout targets the base branch" describes the pre-December behaviour.

The safety argument is unaffected (default-branch resolution is stricter, not weaker), but worth being exact here specifically, since this comment is the written rationale for suppressing a high-severity finding. Confirmed by observation: a PR into a branch that doesn't contain the workflow still ran it, reporting `github.ref = refs/heads/<default>` and `github.base_ref = <that branch>`.

.github/workflows/andra-bot.yml
```diff
@@ -0,0 +1,43 @@
+name: AndraBot
+
+# pull_request_target is required so the workflow has a write token to comment
+# on PRs from forks. It is safe here because the workflow never checks out or
+# executes code from the PR — only the base branch is checked out, and the
```
sugat009: **nitpick:** same wording as `.github/zizmor.yml:14`. Since 8 December 2025 `pull_request_target` resolves the workflow file and checkout to the **default** branch, not the PR's base branch ([changelog](https://github.blog/changelog/2025-11-07-actions-pull_request_target-and-environment-branch-protections-changes/)), which also makes the step name on line 31 slightly off.

Nothing unsafe, and it's a fair mistake given the behaviour changed recently. Worth saying "default branch" though, because it also means `master`'s copies of `andra-bot.js` and `PULL_REQUEST_TEMPLATE.md` are what run for PRs based on epic branches, never that branch's own copies.

scripts/ci/andra-bot-messages/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+Guidelines:
+
+- Keep each message a single paragraph — failure messages are rendered as bullet
+  list items.
+- Markdown works: links, `code`, **bold**, etc.
+- Placeholders in `{{double braces}}` are filled in by the bot; keep them intact:
+  - `{{author}}` — the GitHub username of the PR author (without the `@`)
+  - `{{issueList}}` — the linked issue number(s), e.g. `#1234, #5678`
+  - `{{sections}}` — the section names required by the PR template, e.g.
```
sugat009, line 25: **nitpick:** the example on the next line shows `{{sections}}` rendering with backticks, but `getHeadings` returns plain heading text, so the posted comment actually reads "keep the Description, Code review checklist, License sections".

scripts/ci/andra-bot.js
```diff
@@ -0,0 +1,277 @@
...
+const matchesLicense = (prBody, template) => {
+  const templateLicenseSection = parseSections(template).find(section => /license/i.test(section.heading));
+  if (!templateLicenseSection || !prBody) {
+    return true;
+  }
+
+  // missing section is already covered by the section check
+  const bodyLicenseSections = parseSections(prBody)
+    .filter(section => section.heading === templateLicenseSection.heading);
+  return bodyLicenseSections.every(section => normalize(section.content) === templateLicenseSection.content);
```
sugat009: **nitpick:** `section.content` was already normalized at line 53, and `normalize` is idempotent, so this call is a no-op. Worth noting the template side of the same comparison relies on that pre-normalization, so the code is inconsistent about whether it trusts `parseSections` output.

scripts/ci/andra-bot.js
dianabarsan: I will create them ahead of time.

scripts/ci/andra-bot-messages/not-assigned.md
dianabarsan: I added some clarifying text to the comment.

.github/workflows/andra-bot.yml
dianabarsan: I think all branches should be targetted.

scripts/ci/andra-bot.js
sugat009: Quick note: this is separate from the `normalize()` tidy at line 92 that's now sorted. It's the slice boundary, since License is the last heading the section runs to the end of the body, so anything a contributor writes below the licence paragraph (a screenshot, a note) reads as a licence change. Still 4 of the 5 real license-changed cases on the current head. Non-blocking, just flagging it's still live.

scripts/ci/andra-bot.js
sugat009: Quick note: still live, the label logic is unchanged. Re-running the run that added `Waiting for contributor` replays a payload where the PR was unlabelled, so it ends up carrying both labels. Reading current labels from the API rather than the payload covers it. Non-blocking.

sugat009 (approved):
Thanks for the fixes! LGTM!

PR #11298, opened 2026-07-25
fix(#11212): avoid duplicate registration SMS replies

### LINKED ISSUE #11212: Registration transition can duplicate messages sent by single submission
**Describe the bug**
There was an intermittently failing e2e test that revealed this: https://github.com/medic/cht-core/issues/10874

Recently I added some debug logging to show which messages were failing the assertions, to assess whether the failures were valid or not. Turn out they were. 

This is a document that has _two_ messages that have been sent, even though it should have only one: 
```
{
  "_id": "019ef473-67f7-788a-a9c9-8418ed375bce",
  "_rev": "3-782006900c440c850ae937ca7e24f698",
  "type": "data_record",
  "from": "duplicate-phone",
  "form": "SICK",
  "errors": [],
  "tasks": [
    {
      "messages": [
        {
          "uuid": "019ef473-6827-788a-a9c9-a2262033916b",
          "to": "duplicate-phone",
          "message": "Await further instructions"
        }
      ],
      "state_history": [
        {
          "state": "pending",
          "timestamp": "2026-06-23T12:27:44.555Z"
        },
        {
          "state": "forwarded-to-gateway",
          "timestamp": "2026-06-23T12:27:45.060Z"
        }
      ],
      "state": "forwarded-to-gateway"
    },
    {
      "messages": [
        {
          "uuid": "019ef473-68d9-7ee7-9e4a-8c0f65f1e345",
          "to": "duplicate-phone",
          "message": "Await further instructions"
        }
      ],
      "state_history": [
        {
          "state": "pending",
          "timestamp": "2026-06-23T12:27:44.744Z"
        },
        {
          "state": "forwarded-to-gateway",
          "timestamp": "2026-06-23T12:27:45.060Z"
        }
      ],
      "state": "forwarded-to-gateway"
    }
  ],
  "fields": {},
  "reported_date": 1782217664503,
  "sms_message": {
    "type": "sms_message",
    "message": "SICK",
    "form": "SICK",
    "from": "duplicate-phone",
    "gateway_ref": "019ef473-6668-7000-8677-3ea41fde2fe0"
  }
}
```

So something is processing these messages twice. This could be a race condition between api and sentinel running the same transition twice over the same doc, but that needs to be confirmed. 

Run where this failure happened: https://github.com/medic/cht-core/actions/runs/28024101918/job/82950853099#step:18:3058

**Expected behavior**
Registration transition should not add messages twice. 

**Environment**
- Instance: CI
- App: api/sentinel

EmmS21:
# Description

One SMS submission can produce two identical confirmation replies.

API runs registration and saves the report first; the "already 
ran" sticky note is written slightly later. If Sentinel sees the 
report in that gap, it runs registration again and adds a second 
reply. That race is intermittent, so someone can send one message 
and sometimes get two identical confirmations back.

**Fix:** when registration adds messages, request unique replies 
(`addMessage(..., true)`). If the same reply is already on the 
report, do not add another.

**Rationale:** I considered ownership (make the sticky note 
visible earlier so Sentinel skips) vs idempotency (allow a second 
run, but do not create the same reply twice). Ownership is cleaner 
when it works, but if the handoff fails the failure is obvious to 
the person on the phone. Idempotency is more robust for this bug: 
it checks the reply already on the report, not just the late 
sticky note. It is also the smallest change - uniqueness already 
existed; registration was not using it.

**User impact:** one submission should map to one confirmation 
SMS. Duplicate identical replies are suppressed even when 
processing races.

Closes #11212

# Code review checklist
- [x] Readable: Concise, well named, follows the [style 
guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/)
- [x] Backwards compatible: Works with existing data and 
configuration or includes a migration. Any breaking changes 
documented in the release notes.
- [x] AI disclosure: Please disclose use of AI per [the 
guidelines](https://docs.communityhealthtoolkit.org/community/contributing/ai-guidelines/).

**AI Disclosure:** Used Cursor to explore the API/Sentinel race, 
compare ownership vs idempotency, implement the unique-reply 
change, and draft commit/PR text. I reviewed the approach and 
changes before submitting.

# License

The software is provided under AGPL-3.0. Contributions to this 
project are accepted under the same license.


XVI-Adam (approved):

XVI-Adam (approved):
Looks good, adding a bool true should avoid duplicate replies

dianabarsan (changes requested):
I think that this PR treats the symptom and not the cause. 
The real fix is to treat the race condition between API and Sentinel, which I believe was fixed with the change of blocking Sentinel from processing the doc at the same time as API.

I believe that was addressed with https://github.com/medic/cht-core/commit/fca1040277b05b31b1268a053b173088207261cd 

If this test continues to fail even with that change landed, it would reveal that the race condition is _not_ addressed, and we should see it. 
Landing this fix would obstruct that visibility.

Issue #11279, opened 2026-07-22
P2P: Add data-bundle api endpoint


jkuester:
Add `POST /api/v1/replication/data-bundle` api endpoint for uploading an encrypted data bundle. The server decrypts (it holds the key), verifies the CHW's signature (confirming origin and that nothing was tampered with), validates each unpacked doc is accessible to the CHW (out-of-scope docs are rejected), ingests the valid docs, and returns the new checkpoint for the CHW. 

Gap handling: ingest every valid bundle regardless of gaps in the change-log, but only advance the checkpoint to the highest contiguous sequence.

Add integration tests for the endpoint to `tests/integration/api/controllers`.

jkuester:
https://github.com/medic/cht-core/pull/11388

PR #11305, opened 2026-07-29
feat: add device-key api endpoint (#11278)

### LINKED ISSUE #11278: P2P: Add device-key api endpoint
Add `POST /api/v1/users/{username}/device-key` endpoint. Keys are per-device, not per-user: a device sends up its own public key, and the server stores it (one per registered device, so the same user on multiple devices/browsers is handled). The response returns the server's current public key.

We want to leverage https://github.com/FiloSottile/typage for encryption.

Be sure to include integration tests for the new endpoint in `tests/integration/api/controllers`.

TODO Should we require `can_send_offline_data_bundle` permission for calling this endpoint or just gate it to offline users?

paulpascal:
# Description

Adds `POST /api/v1/users/{username}/device-key` for the offline data bundle feature. Each device registers its own key pair: an X25519 encryption key (age recipient, used to encrypt the checkpoint back to the device) and an Ed25519 signing key (used by the server to verify bundle signatures in a later issue). The server stores both public keys per device (one entry per device_id, upsert) and returns its own current encryption public key.

- Gated by the `can_send_offline_data_bundle` / `can_relay_offline_data_bundle` permissions; non-admins may only register their own device.
- Both keys are format-validated (age recipient + base64 Ed25519) before storage.
- `age-encryption` (typage) is ESM-only, so it's loaded through a small lazy dynamic-import wrapper to stay compatible with the CommonJS api and nyc.

Closes #11278

# Code review checklist
- [x] Tested: Unit and integration
- [x] Backwards compatible: Additive endpoint, no existing behaviour changed
- [ ] AI disclosure: Please disclose use of AI per the guidelines

# License

### DIFF
--- a/api/src/controllers/users.js
+++ b/api/src/controllers/users.js
@@ -8,6 +8,8 @@ const logger = require('@medic/logger');
 const serverUtils = require('../server-utils');
 const replication = require('../services/replication/replication');
 const serverKey = require('../services/offline-data-bundle/server-key');
+const age = require('../services/offline-data-bundle/age');
+const signing = require('../services/offline-data-bundle/signing');
 
 const OFFLINE_DATA_BUNDLE_PERMISSIONS = ['can_send_offline_data_bundle', 'can_relay_offline_data_bundle'];
 
@@ -498,14 +500,19 @@ module.exports = {
    *         application/json:
    *           schema:
    *             type: object
-   *             required: [device_id, public_key]
+   *             required: [device_id, encryption_key, signing_key]
    *             properties:
    *               device_id:
    *                 type: string
    *                 description: Unique identifier for the device.
-   *               public_key:
+   *               encryption_key:
    *                 type: string
-   *                 description: The device's age recipient (public) key.
+   *                 description: >
+   *                   The device's age recipient (X25519 public) key, used to encrypt the checkpoint to the device.
+   *               signing_key:
+   *                 type: string
+   *                 description: >
+   *                   The device's Ed25519 public signing key, used by the server to verify bundle signatures.
    *     responses:
    *       '200':
    *         description: Device key registered
@@ -518,6 +525,10 @@ module.exports = {
    *                   type: string
    *                   description: The server's age recipient (public) key.
    *       '400':
+   *         description: >
+   *           Required fields are missing, or a key is malformed. The `encryption_key` must be an age
+   *           recipient string and the `signing_key` must be base64 of a raw Ed25519 public key; invalid
+   *           key formats are rejected.
    *         $ref: '#/components/responses/BadRequest'
    *       '401':
    *         $ref: '#/components/responses/Unauthorized'
@@ -529,10 +540,26 @@ module.exports = {
       return serverUtils.emptyJSONBodyError(req, res);
     }
 
-    const { device_id: deviceId, public_key: publicKey } = req.body;
-    if (!deviceId || !publicKey) {
+    const { device_id: deviceId, encryption_key: encryptionKey, signing_key: signingKey } = req.body;
+    if (!deviceId || !encryptionKey || !signingKey) {
       return serverUtils.error(
-        { code: 400, reason: 'Missing required fields: device_id and public_key' },
+        { code: 400, reason: 'Missing required fields: device_id, encryption_key and signing_key' },
+        req,
+        res
+      );
+    }
+
+    if (!(await age.isValidRecipient(encryptionKey))) {
+      return serverUtils.error(
+        { code: 400, reason: 'Invalid encryption_key: expected an age recipient' },
+        req,
+        res
+      );
+    }
+
+    if (!(await signing.isValidPublicKey(signingKey))) {
+      return serverUtils.error(
+        { code: 400, reason: 'Invalid signing_key: expected a base64 Ed25519 public key' },
         req,
         res
       );
@@ -545,7 +572,7 @@ module.exports = {
         return serverUtils.error({ code: 403, message: 'You do not have permissions to modify this user' }, req, res);
       }
 
-      await users.setDeviceKey(username, deviceId, publicKey);
+      await users.setDeviceKey(username, deviceId, encryptionKey, signingKey);
       const serverPublicKey = await serverKey.getServerPublicKey();
 
       logger.info(`REQ ${req.id} - Registered device key for device '${deviceId}' on user '${username}'.`);
--- a/api/src/controllers/users.js
+++ b/api/src/controllers/users.js
@@ -13,6 +13,20 @@ const signing = require('../services/offline-data-bundle/signing');
 
 const OFFLINE_DATA_BUNDLE_PERMISSIONS = ['can_send_offline_data_bundle', 'can_relay_offline_data_bundle'];
 
+const validateDeviceKeyBody = async (body) => {
+  const { device_id: deviceId, encryption_key: encryptionKey, signing_key: signingKey } = body;
+  if (!deviceId || !encryptionKey || !signingKey) {
+    return 'Missing required fields: device_id, encryption_key and signing_key';
+  }
+  if (!(await age.isValidRecipient(encryptionKey))) {
+    return 'Invalid encryption_key: expected an age recipient';
+  }
+  if (!(await signing.isValidPublicKey(signingKey))) {
+    return 'Invalid signing_key: expected a base64 Ed25519 public key';
+  }
+  return null;
+};
+
 const hasFullPermission = req => {
   return auth
     .check(req, ['can_edit', 'can_update_users'])
@@ -540,29 +554,9 @@ module.exports = {
       return serverUtils.emptyJSONBodyError(req, res);
     }
 
-    const { device_id: deviceId, encryption_key: encryptionKey, signing_key: signingKey } = req.body;
-    if (!deviceId || !encryptionKey || !signingKey) {
-      return serverUtils.error(
-        { code: 400, reason: 'Missing required fields: device_id, encryption_key and signing_key' },
-        req,
-        res
-      );
-    }
-
-    if (!(await age.isValidRecipient(encryptionKey))) {
-      return serverUtils.error(
-        { code: 400, reason: 'Invalid encryption_key: expected an age recipient' },
-        req,
-        res
-      );
-    }
-
-    if (!(await signing.isValidPublicKey(signingKey))) {
-      return serverUtils.error(
-        { code: 400, reason: 'Invalid signing_key: expected a base64 Ed25519 public key' },
-        req,
-        res
-      );
+    const validationError = await validateDeviceKeyBody(req.body);
+    if (validationError) {
+      return serverUtils.error({ code: 400, reason: validationError }, req, res);
     }
 
     try {
@@ -572,6 +566,7 @@ module.exports = {
         return serverUtils.error({ code: 403, message: 'You do not have permissions to modify this user' }, req, res);
       }
 
+      const { device_id: deviceId, encryption_key: encryptionKey, signing_key: signingKey } = req.body;
       await users.setDeviceKey(username, deviceId, encryptionKey, signingKey);
       const serverPublicKey = await serverKey.getServerPublicKey();
 
--- a/api/src/controllers/users.js
+++ b/api/src/controllers/users.js
@@ -587,7 +587,8 @@ module.exports = {
         encryption: serverKeys.server_encryption_private_key,
         signing: serverKeys.server_signing_private_key,
       });
-      const serverPublicKeys = await users.setDeviceKey(username, deviceId, encryptionKey, signingKey, {
+      const deviceKeys = { encryption_key: encryptionKey, signing_key: signingKey };
+      const serverPublicKeys = await users.setDeviceKey(username, deviceId, deviceKeys, {
         server_encryption_public_key: serverKeys.server_encryption_public_key,
         server_signing_public_key: serverKeys.server_signing_public_key,
       });
--- a/api/src/controllers/users.js
+++ b/api/src/controllers/users.js
@@ -7,24 +7,34 @@ const auth = require('../auth');
 const logger = require('@medic/logger');
 const serverUtils = require('../server-utils');
 const replication = require('../services/replication/replication');
-const serverKey = require('../services/offline-data-bundle/server-key');
 const age = require('../services/offline-data-bundle/age');
 const signing = require('../services/offline-data-bundle/signing');
-
-const OFFLINE_DATA_BUNDLE_PERMISSIONS = ['can_send_offline_data_bundle', 'can_relay_offline_data_bundle'];
+const serverKey = require('../services/offline-data-bundle/server-key');
 
 const validateDeviceKeyBody = async (body) => {
-  const { device_id: deviceId, encryption_key: encryptionKey, signing_key: signingKey } = body;
-  if (!deviceId || !encryptionKey || !signingKey) {
-    return 'Missing required fields: device_id, encryption_key and signing_key';
+  const { encryption_key: encryptionKey, signing_key: signingKey } = body;
+  if (!encryptionKey || !signingKey) {
+    return 'Missing required fields: encryption_key and signing_key';
   }
   if (!(await age.isValidRecipient(encryptionKey))) {
     return 'Invalid encryption_key: expected an age recipient';
   }
   if (!(await signing.isValidPublicKey(signingKey))) {
-    return 'Invalid signing_key: expected a base64 Ed25519 public key';
+    return 'Invalid signing_key: expected an Ed25519 public key JWK';
   }
   return null;
+};
+
+const generateServerKeys = async () => {
+  const encryptionIdentity = await age.generateIdentity();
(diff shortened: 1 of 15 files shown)


paulpascal:
Hey @jkuester could you please create a `feat/offline-data-bundle` branch off master on cht-core?

We'll point the per-issue PRs at it (I don't have perms to create it there).

First one is ready on my fork: [`paulpascal:11278-device-key-endpoint`](https://github.com/paulpascal/cht-core/tree/11278-device-key-endpoint) for #11278.

api/src/services/offline-data-bundle/age.js
```diff
@@ -0,0 +1,22 @@
+// age-encryption (typage) is an ESM-only module. This wrapper is the single boundary that
+// loads it, via a lazy dynamic import
+let agePromise;
+const load = () => {
+  if (!agePromise) {
+    agePromise = import('age-encryption');
```
jkuester, line 7: Can we just ignore the linting error here instead of editing the top-level eslint.config.js file?

```suggestion
    // eslint-disable-next-line n/no-extraneous-import
    agePromise = import('age-encryption');
```

shared-libs/user-management/src/users.js
```diff
@@ -621,6 +621,24 @@ const saveUserSettingsUpdates = async (userSettings) => {
   };
 };
 
+const upsertDeviceKey = async (username, deviceId, encryptionKey, signingKey) => {
+  const userSettings = await getUserDoc(username, 'medic');
```
jkuester: I think we want to store this info on the user's doc in the `_users` db, and not on the `user-settings` doc in the `medic` db.  🤔

The `_users` doc is treated as _sensitive_ (it is the source of truth for authentication and authorization.)  But, the `user-settings` doc is actually replicated to the user's device. (which is def not what we want for these keys).

```suggestion
  const userSettings = await getUserDoc(username, 'users');
```

shared-libs/user-management/src/users.js
```diff
@@ -621,6 +621,24 @@ const saveUserSettingsUpdates = async (userSettings) => {
...
+  userSettings.device_keys = userSettings.device_keys || [];
+  const entry = {
+    device_id: deviceId,
+    encryption_key: encryptionKey,
+    signing_key: signingKey,
+    created_at: new Date().toISOString()
+  };
+  const existing = userSettings.device_keys.find(key => key.device_id === deviceId);
```
jkuester: Minor, but what do you think about making `device_keys` an object dictionary keyed on the deviceId instead of an array?  So the user doc would be like:

```js
{
  ...,
  keys_by_device: {
    [deviceId]: {
      encryption_key: encriptionKey,
      ...
    }
  } 
}
```

This makes the uniqueness of `deviceId` enforced by the data structure.  Then, if we do need to spin through all of the keys, you can still just do something like `Object.values(userDoc.keys_by_device).map(...)`

shared-libs/user-management/src/users.js
```diff
@@ -621,6 +621,24 @@ const saveUserSettingsUpdates = async (userSettings) => {
...
 };
 
+const upsertDeviceKey = async (username, deviceId, encryptionKey, signingKey) => {
+  const userSettings = await getUserDoc(username, 'medic');
+  userSettings.device_keys = userSettings.device_keys || [];
+  const entry = {
+    device_id: deviceId,
+    encryption_key: encryptionKey,
+    signing_key: signingKey,
+    created_at: new Date().toISOString()
```
jkuester: We do use `updated_date` elsewhere so I have a slight preference for that (but also fine with something else if you feel "create" is semantically important here).  I do think we want to just store the millisec number instead of the ISO string (unless i am missing something???).  Date strings are for displaying to the user and do not belong in the database!  😅

```suggestion
    updated_date: Date.now()
```

shared-libs/user-management/src/users.js
```diff
@@ -621,6 +621,24 @@ const saveUserSettingsUpdates = async (userSettings) => {
...
+  };
+  const existing = userSettings.device_keys.find(key => key.device_id === deviceId);
+  if (existing) {
+    Object.assign(existing, entry);
+  } else {
+    userSettings.device_keys.push(entry);
+  }
+  return saveUserSettingsUpdates(userSettings);
```
jkuester:
```suggestion
  return saveUserUpdates(userSettings);
```

api/src/controllers/users.js
```diff
@@ -468,6 +487,96 @@ module.exports = {
...
+   *         description: The username to register the device key for.
+   *     requestBody:
+   *       required: true
+   *       content:
+   *         application/json:
+   *           schema:
+   *             type: object
+   *             required: [device_id, encryption_key, signing_key]
+   *             properties:
+   *               device_id:
```
jkuester: Looking at this again, I wonder if we should put the device_id in the URL path instead of here in the body... 🤔  Something like `/api/v1/users/{username}/devices/{device_id}/keys`.  That seems a bit more RESTful, but I think it would work fine either way.  

Maybe the big question would be if we ever want to support `GET` or `DELETE` against the same endpoint... In that case it does seem like the full `/api/v1/users/{username}/devices/{device_id}/keys` would be better since we would not need to have any query/body params...  I know we don't have any plans for these endpoints at this time. Do you think this would be better anyway, or am I just overthinking this?

api/src/controllers/users.js
```diff
@@ -468,6 +487,96 @@ module.exports = {
...
+   *     summary: Register a device encryption key
+   *     operationId: v1UsersUsernameDeviceKeyPost
+   *     description: >
+   *       Stores the device's age public (recipient) key against the user and returns the server's current age
+   *       public key. Keys are per-device: re-registering the same `device_id` replaces the stored key. Requires
+   *       the `can_send_offline_data_bundle` or `can_relay_offline_data_bundle` permission. Non-admin users can
+   *       only register a key for themselves.
+   *     tags: [User]
+   *     x-permissions:
+   *       hasAny: [can_send_offline_data_bundle, can_relay_offline_data_bundle]
```
jkuester: I might be thinking of this wrong, but the "taxi" device does not actually need keys, right?

```suggestion
      hasAll: [can_send_offline_data_bundle]
```

api/src/controllers/users.js
```diff
@@ -468,6 +487,96 @@ module.exports = {
...
+   * @openapi
+   * /api/v1/users/{username}/device-key:
+   *   post:
+   *     summary: Register a device encryption key
+   *     operationId: v1UsersUsernameDeviceKeyPost
+   *     description: >
+   *       Stores the device's age public (recipient) key against the user and returns the server's current age
+   *       public key. Keys are per-device: re-registering the same `device_id` replaces the stored key. Requires
+   *       the `can_send_offline_data_bundle` or `can_relay_offline_data_bundle` permission. Non-admin users can
+   *       only register a key for themselves.
```
jkuester: Minor wording suggestions.  I think for the sake of clarity we should always say "age encryption", because otherwise it is a bit confusing what we are referring to...

```suggestion
   *       Registers new public key values for the user's device and returns the server's current age encryption
   *       public key for that device. Keys are per-device: re-registering the same `device_id` replaces the stored
   *       keys and generates a new server key for that device. "Offline" users with the `can_send_offline_data_bundle`
   *       permission can only register keys for themselves.
```

api/src/controllers/users.js
```diff
@@ -468,6 +487,96 @@ module.exports = {
...
+   *             type: object
+   *             required: [device_id, encryption_key, signing_key]
+   *             properties:
+   *               device_id:
+   *                 type: string
+   *                 description: Unique identifier for the device.
+   *               encryption_key:
+   *                 type: string
+   *                 description: >
+   *                   The device's age recipient (X25519 public) key, used to encrypt the checkpoint to the device.
```
jkuester: I think it is okay to not include the purpose/justification here.  If the data-bundle functionality evolves over time, these notes might become stale...

```suggestion
   *                   The device's age encryption recipient (X25519 public) key string.
```

api/src/controllers/users.js
```diff
@@ -468,6 +487,96 @@ module.exports = {
...
+   *               signing_key:
+   *                 type: string
+   *                 description: >
+   *                   The device's Ed25519 public signing key, used by the server to verify bundle signatures.
```
jkuester:
```suggestion
   *                   The device's (Ed25519 public) signing key encoded with base64.
```

api/src/controllers/users.js
```diff
@@ -468,6 +487,96 @@ module.exports = {
...
+   *       '200':
+   *         description: Device key registered
+   *         content:
+   *           application/json:
+   *             schema:
+   *               type: object
+   *               properties:
+   *                 server_key:
+   *                   type: string
+   *                   description: The server's age recipient (public) key.
```
jkuester:
```suggestion
   *                   description: The server's age encryption recipient (X25519 public) key string.
```

api/src/controllers/users.js
```diff
@@ -468,6 +487,96 @@ module.exports = {
...
+   *       '400':
+   *         description: >
+   *           Required fields are missing, or a key is malformed. The `encryption_key` must be an age
+   *           recipient string and the `signing_key` must be base64 of a raw Ed25519 public key; invalid
+   *           key formats are rejected.
```
jkuester:
```suggestion
   *         $ref: '#/components/responses/BadRequest'
```

api/src/controllers/users.js
```diff
@@ -468,6 +487,96 @@ module.exports = {
...
+    }
+
+    const validationError = await validateDeviceKeyBody(req.body);
+    if (validationError) {
+      return serverUtils.error({ code: 400, reason: validationError }, req, res);
+    }
+
+    try {
+      const username = req.params.username;
+      const userCtx = await auth.assertPermissions(req, { hasAny: OFFLINE_DATA_BUNDLE_PERMISSIONS });
```
jkuester: Can we do this assert first, before anything else in the function?  This is the place where we flag unauthenticated requests. We don't want to do any extra processing at all until we verify the request came from an actual CHT user.

shared-libs/user-management/src/users.js
```diff
@@ -621,6 +621,24 @@ const saveUserSettingsUpdates = async (userSettings) => {
   };
 };
 
+const upsertDeviceKey = async (username, deviceId, encryptionKey, signingKey) => {
+  const userSettings = await getUserDoc(username, 'medic');
+  userSettings.device_keys = userSettings.device_keys || [];
+  const entry = {
+    device_id: deviceId,
+    encryption_key: encryptionKey,
+    signing_key: signingKey,
```
jkuester: I think it is worth being very explicit that these are not private keys.

```suggestion
    encryption_public_key: encryptionKey,
    signing_public_key: signingKey,
```

api/src/controllers/users.js
```diff
@@ -468,6 +487,96 @@ module.exports = {
...
+      if (!auth.isDbAdmin(userCtx) && !isReferencingSelf(userCtx, auth.basicAuthCredentials(req), username)) {
+        return serverUtils.error({ code: 403, message: 'You do not have permissions to modify this user' }, req, res);
+      }
+
+      const { device_id: deviceId, encryption_key: encryptionKey, signing_key: signingKey } = req.body;
+      await users.setDeviceKey(username, deviceId, encryptionKey, signingKey);
+      const serverPublicKey = await serverKey.getServerPublicKey();
```
jkuester: I think we want a separate server key for each user/device?  🤔  I remember discussing both options. Having only one key for the server is more simple, but it makes it pretty much impossible to ever rotate the server's key (because there might be an offline client somewhere that is still trying to use the old public key to encrypt data bundles).

jkuester: Also, don't we need to also create/return the server's public signing key here too? The client needs the server's public signing key to verify the signature on the checkpoint.

jkuester: Finally, since all of this is going to be tied to a username/device_id (I think), what do you think of moving the `getServerPublicKey` code down into `shared-libs/user-managment`? We could even just automatically return the new public keys when `setDeviceKey` is called.

api/src/services/offline-data-bundle/server-key.js
```diff
@@ -0,0 +1,21 @@
+const secureSettings = require('@medic/settings');
```
jkuester, line 1: Minor, but we don't really have enough code here (at least not yet) to really need the `offline-data-bundle` subdir with separate services.  Can we just combine this file and `signing` into one `services/data-bundle.js` file?  We can keep `age.js` as a separate, reusable, service.

api/src/services/offline-data-bundle/signing.js
```diff
@@ -0,0 +1,18 @@
...
+const { webcrypto } = require('node:crypto');
+
+module.exports = {
+  isValidPublicKey: async (base64Key) => {
+    const raw = Buffer.from(base64Key, 'base64');
+    if (raw.length !== 32) {
+      return false;
+    }
+    try {
+      await webcrypto.subtle.importKey('raw', raw, { name: 'Ed25519' }, false, ['verify']);
```
jkuester: Okay, so, I am no expert here on any of this. However, I have been doing some reading on the different supported parameters here and I wonder if we should be accepting the public signing key from the client in the `jwk` format instead of base64 (`raw`)?  As I understand it, the `jwk` format is just a JS object with some metadata and then the actual key part already encoded with `base64url` under the hood.  The key is you can import/export from WebCrypto in that format. So, it should serialize/deserialize on the POST request json without needing to do any manual buffering to/from base64. Less code both on the server and client without adding too much overhead on the request... (I think... 😅)

jkuester (changes requested):
Great start here @paulpascal !  🔒 Bear with me on some of the details. If I missed the point or am totally off just let me know!  🙏

paulpascal:
Thanks @jkuester , really helpful review. Went through it all, taking almost everything:

**Storage:** Agreed, moving the keys to the `_users` doc instead of `user-settings`, good catch on the replication/sensitivity point. Making `keys_by_device` an object keyed on `device_id`, using `updated_date: Date.now()`, and the explicit `encryption_public_key` / `signing_public_key` names. Also moving the permission assert to the very top so we reject unauthenticated requests before doing any work.

**Permission:** Right, only the sender needs a key, so gating on `can_send_offline_data_bundle` and dropping the relay one here.

**Endpoint shape:** Agreed, going with `POST /api/v1/users/{username}/devices/{device_id}/keys`. Cleaner and leaves room for GET/DELETE later.

**Server key per device:** Yes, let's do per-device. A single server key can't really rotate: retiring the old one waits on every device draining it, and with perpetually offline CHWs it never dies. Per device makes rotation and decommission bounded to that one device. So the server generates and stores a keypair per device and returns the public part here.

**Server signing key:** Good point, the client needs it to verify the checkpoint signature, otherwise a relay could forge a checkpoint and the CHW would skip unsent data. Adding a server signing key too, per device for consistency.

**Move key handling into user-management:** Makes sense with the above, `setDeviceKey` will generate/store and return the server public keys.

**eslint:** Yep, using the inline disable instead of touching the global config.

**JWK vs base64:** Like it, accepting the signing key as JWK so it deserializes straight from the JSON, less manual buffering on both sides.

**File layout:** Heads up: #11279 (data-bundle endpoint) is stacked on this branch and already adds `services/offline-data-bundle/data-bundle.js` plus the crypto files, so there are 4 files in that dir now. Rather than collapse them here, can we keep the `offline-data-bundle/` subdir with `age.js` separate, and I line up `signing`/`server-key` so both PRs stay consistent? Happy to and would def adjust if you'd prefer a different split.

api/src/controllers/users.js
paulpascal: Since we moved the signing key to JWK (your other comment), this is now described as a JWK instead of base64. Done.

api/src/controllers/users.js
paulpascal: Done, per device now. Agreed a single server key can't really rotate: with perpetually offline devices the old one never drains, so retirement is bounded per device this way.

paulpascal: Good catch, added a server signing key too (per device) and returning it here, so the client can verify the checkpoint signature.

paulpascal: Mostly done: the per-device server keys now live on the _users doc (publics) and the secureSettings vault (privates), and setDeviceKey returns the public ones. Couldn't literally move the generation into user-management since it's a standalone package and can't require api's age/signing helpers, so the controller generates them and passes them in.

api/src/services/offline-data-bundle/signing.js
paulpascal: Done, taking the signing key as a JWK now so it deserializes straight from the request json, no manual base64 buffering either side.

api/src/services/offline-data-bundle/server-key.js
paulpascal: Heads up: #11279 is stacked on this and already adds services/offline-data-bundle/data-bundle.js plus the crypto files, so there are a few files in that dir now. I'd keep the offline-data-bundle/ subdir with age.js separate so both PRs stay consistent. Ok with that, or would you rather a different split?

jkuester (approved):
Sorry for the delay here @paulpascal !  LGTM! 👍

Issue #10748, opened 2026-03-24
Remove docs_by_id_lineage map reduce view


witash:
The `docs_by_id_lineage` map reduce view is one of the largest views on the instances sampled.
It emits for every ancestor of every document with a `contact` field; with _id as the value; which allows getting all ancestor documents in a single query with the `include_docs` parameter.

This could be replaced by two calls to `all_docs`; one to get the child document, which contains all the ancestors that would be emitted by the view, then a second call to `all_docs` with the ancestor `_ids` to get the actual documents. 

If this view was called with a lot of ids, it could be more efficient to make a single call to get all the documents with their ancestors instead of making two `all_docs` queries, with a large number of keys (the second query would have to have all the ancestor ids).

However, its not currently used that way; its always called with a single id, where it's not likely to be much more efficient than just making two `all_docs` calls.

Since it is very large, if its not too much worse to use all_docs, this view should be removed.

### Definition
- **Database:** medic (synced to offline clients)
- **Map keys:** `[doc._id, depth]` (depth increments along parent chain)
- **Map values:** `{ _id: parent._id }` (for linked-document fetching with `include_docs`)
- **Reduce:** None

### Usages
| # | File | Key/Options Used | Functional Purpose |
|---|------|------------------|--------------------|
| 1 | shared-libs/lineage/src/hydration.js:231 | `{ startkey: [id], endkey: [id, {}], include_docs: true }` | Core lineage hydration: fetches a document and all its ancestors. Used by `fetchHydratedDoc` and `fetchLineageById` throughout the codebase. |
| 2 | shared-libs/cht-datasource/src/local/libs/lineage.ts:24 | `startkey: [id], endkey: [id, {}]` (via `queryDocsByRange`) | cht-datasource local data context lineage resolution. Used by `Person.v1.getWithLineage`, `Place.v1.getWithLineage`, `Contact.v1.getWithLineage`. |

sh1vam31:
Hi @witash, @andrablaj 
 
I would love to take this on! I can update the lineage hydration logic to use two all_docs queries and remove the view from the ddocs.

Could you please assign this to me?

witash:
This is a little bit blocked by https://github.com/medic/cht-core/pull/10615 which changes all the views references.
You could start working on it before that is merged, but are likely to get conflicts in git

PR #11311, opened 2026-08-03
feat(#10706): add move contact hierarchy

### LINKED ISSUE #11347: Support moving entire contact hierarchy

**Additional context**

vikrantwiz02:
# Description

Closes #11347

Adds the move half of the Advanced Contact Management API: `POST /api/v1/person/{id}/move` and `POST /api/v1/place/{id}/move`, moving a contact and its whole subtree under a new parent.

This is the second consumer of the bulk-operations framework that landed with delete in #11236, and it needs no changes to that framework. The design was reviewed on the Move tab of the technical design doc and follows what we settled there.

**How it works.** The request validates the move, gathers the documents it touches, and queues a bulk operation, returning a job id immediately. Sentinel applies the changes in batches, and the caller polls `GET /api/v1/bulk-operations/{id}`.

Two actions are queued, named for the field each writes:

| Action | Writes | Applied to |
|---|---|---|
| `set-parent` (new) | `doc.parent` | the source contact and every descendant |
| `set-contact` (existing) | `doc.contact` | the reports whose cached author lineage is stale, and any place whose primary contact moved |

`set-contact` is reused unchanged because in both cases the value being refreshed lives in a property called `contact`: on a report it is the author reference, on a place it is the primary contact.

Every operation is self-contained, carrying the document id, the value it expects to find, and the new value to write, so the handlers compute nothing. An operation whose document changed since it was gathered fails rather than clobbering the newer edit, and every `bulkDocs` result is checked.

**Validation.** Ported from cht-conf's `lineage-constraints`: no cycles, only configured parent types, no move that strands a place's primary contact, and a move to the parent the contact already has is rejected rather than writing identical values back. Violations return a `400` with the reason; anything else, such as a database failure during validation, propagates as a `500` rather than being reported as an invalid move.

**Finding the reports.** A move changes where the *author* sits, so the affected reports are the ones the moved contacts wrote, not the ones about them. These come from the nouveau `reports_by_freetext` index queried on `exact_match:"contact:<lowercased id>"`. Results are paged with the bookmark rather than capped, so a prolific author is not silently truncated.

**Scale.** Beyond fetching the two contacts named in the request, no documents are read. Every input comes from an index: the subtree and each contact's current parent from `contacts_by_depth` (keyed at depth 1 for the parents), the affected reports from `reports_by_freetext`, their authors from the `submitter` field stored on `docs_by_replication_key`, and the places whose primary contact moved from `contacts_by_primary_contact`, which emits that contact's id as the row key. A district move can touch tens of thousands of documents and the request is the half of this that has to stay cheap.

**Permission.** A new `can_move_contact_hierarchy`, shipping assigned to no role. `can_edit` is held by almost every user and is a webapp permission, not appropriate for guarding a server operation that rewrites a subtree.

**Endpoint shape.** Per-type rather than a single `/contact/{uuid}` endpoint, following what delete shipped with, so a place cannot be moved through the person endpoint or the reverse.

**One thing worth flagging for review:** a move changes who can replicate the subtree. Moving a clinic between districts means the receiving district's offline users begin syncing those contacts and their reports, and the previous district's stop. That is what a move means, but it is an access change as much as a data change, which is part of why the permission is separate and admin-assigned.

The first commit lands the lineage manipulation helpers, which were written earlier but never merged while we focused on delete.

#10706

# Code review checklist

# License


api/src/controllers/person.js
```diff
@@ -252,5 +253,60 @@ module.exports = {
...
+     * @openapi
+     * /api/v1/person/{id}/move:
+     *   post:
+     *     summary: Move a person to a new parent
+     *     operationId: v1PersonIdMovePost
+     *     description: >
+     *       Queues an asynchronous bulk operation that moves the person and its whole subtree under a
+     *       new parent, rewriting the parent lineage on every descendant and refreshing the cached
+     *       lineage on their reports and on any place whose primary contact moved. Returns a summary
+     *       of the changes and the bulk operation id to poll.
```
jkuester: Minor, I think it is okay to not mention the primary contact rewrites. These should be relatively few and expected.

```suggestion
     *       Queues an asynchronous bulk operation that moves the person under a new parent. If the person is the
     *       primary contact for any places, they will be updated with the person's new hierarchy. If the person is
     *       associated with any user, all the reports written by that user will be updated with the person's new
     *       hierarchy. Returns a summary of the changes and the bulk operation id to poll.
```

api/src/controllers/person.js
```diff
@@ -252,5 +253,60 @@ module.exports = {
...
+     *         $ref: '#/components/responses/BadRequest'
+     *       '401':
+     *         $ref: '#/components/responses/Unauthorized'
+     *       '403':
+     *         $ref: '#/components/responses/Forbidden'
+     *       '404':
+     *         $ref: '#/components/responses/NotFound'
+     */
+    move: moveContactService.handleMove({
+      get: (uuid) => getPerson(Qualifier.byUuid(uuid)),
```
jkuester, line 306: Minor, can you add a 404 unit test like we have for delete?

api/src/controllers/place.js
```diff
@@ -258,6 +259,60 @@ module.exports = {
...
+     * @openapi
+     * /api/v1/place/{id}/move:
+     *   post:
+     *     summary: Move a place to a new parent
+     *     operationId: v1PlaceIdMovePost
+     *     description: >
+     *       Queues an asynchronous bulk operation that moves the place and its whole subtree under a
+     *       new parent, rewriting the parent lineage on every descendant and refreshing the cached
+     *       lineage on their reports and on any place whose primary contact moved. Returns a summary
+     *       of the changes and the bulk operation id to poll.
```
jkuester: I think it is okay to not mention the primary contact rewrites. These should be relatively few and expected.

```suggestion
     *       Queues an asynchronous bulk operation that moves the place and its whole subtree under a new parent.
     *       If any moved persons are associated with a user, all the reports written by that user will be updated
     *       with the person's new hierarchy. Returns a summary of the changes and the bulk operation id to poll.
```

api/src/controllers/place.js
```diff
@@ -258,6 +259,60 @@ module.exports = {
...
+     *         $ref: '#/components/responses/BadRequest'
+     *       '401':
+     *         $ref: '#/components/responses/Unauthorized'
+     *       '403':
+     *         $ref: '#/components/responses/Forbidden'
+     *       '404':
+     *         $ref: '#/components/responses/NotFound'
+     */
+    move: moveContactService.handleMove({
+      get: (uuid) => getPlace(Qualifier.byUuid(uuid)),
```
jkuester, line 312: Same here.  Can you add a 404 unit test like we have for delete?

api/src/services/move-contact.js
```diff
@@ -0,0 +1,227 @@
...
+const constraints = require('./hierarchy/lineage-constraints');
+const {
+  createLineageFromDoc,
+  replaceParentLineage,
+  replaceContactLineage,
+} = require('./hierarchy/lineage-manipulation');
+
+const { ACTIONS } = BULK_OPERATIONS;
+
+const HIERARCHY_ROOT = 'root';
```
jkuester: I know this was just following the pattern from cht-conf.  However, for well defined REST apis, I think we should avoid magic values like this.  We should just be able to make the `parent_id` parameter optional. If it is not set, it means we should move the contact to the top-level.

api/src/services/move-contact.js
```diff
@@ -0,0 +1,227 @@
...
+ * @param {Object} options
+ * @param {boolean} options.dryRun - return the summary without queuing anything
+ * @returns {Promise<Object>} the summary of changes, plus the bulk operation id when the operation
+ *   was queued (omitted for a dry run)
+ * @throws {BadRequestError} when the move is not legal
+ * @throws {NotFoundError} when the destination does not exist
+ */
+const moveContactHierarchy = async (id, parentId, { dryRun } = {}) => {
+  const toRoot = parentId === HIERARCHY_ROOT;
+  const destination = toRoot ? null : await db.medic.get(parentId).catch(err => {
```
jkuester: Can use the cht-datasource `Contact.v1.get` here.  (And throw the NotFoundError if it returns null.)

api/src/services/move-contact.js
```diff
@@ -0,0 +1,227 @@
...
+    if (err.status === 404) {
+      throw new NotFoundError(`Destination contact ${parentId} not found`);
+    }
+    throw err;
+  });
+
+  const contactIds = await getSubtreeIds(id);
+  const source = await db.medic.get(id);
```
jkuester: This is just getting the same thing we already have in the `contact` variable in `handleMove`.  Should just pass in the `contact` here.

api/src/services/move-contact.js
jkuester: Actually, based on my comment below about passing in the `contact`, can we just move this call back up into the `handleMove` function and get the parent and the `contact` at the same time (`Promise.all`)?  Then pass both values down into this function.

api/src/services/hierarchy/lineage-constraints.js
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  if (!contactTypesUtils.isParentOf(destinationType, sourceType)) {
+    throw new BadRequestError(`contacts of type '${sourceType.id}' cannot have parent of type '${destinationType.id}'`);
+  }
+};
+
+/**
+ * Moving a contact beneath one of its own descendants would detach the subtree into a loop.
+ */
```
jkuester: This comment is on the wrong function.

```suggestion
```

api/src/services/lineage-constraints.js
```diff
@@ -0,0 +1,144 @@
...
+ * configured parents may only sit at the root.
+ */
+const assertRootIsAllowed = (sourceType) => {
+  if (contactTypesUtils.hasParents(sourceType)) {
+    throw new BadRequestError(`contacts of type '${sourceType.id}' cannot be moved to the root`);
+  }
+};
+
+const assertParentTypeIsAllowed = (settings, sourceDoc, destinationDoc) => {
+  const sourceType = contactTypesUtils.getContactType(settings, sourceDoc);
```
jkuester, line 19: 😓 There is an existing issue with `getContactType` that I have not got around to fixing yet.  I think the time is now....


Basically, can you update the `getTypeById` function in shared-libs/contact-type-utils code to be like this:

```js
const PERSON = 'person'; // TODO actually just add this to to the CONTACT_TYPES in shared-libs/constants
const { CONTACT_TYPES: { DISTRICT_HOSPITAL, CLINIC, HEALTH_CENTER } } = require('@medic/constants');
const HARDCODED_TYPES_DICT = {
  [DISTRICT_HOSPITAL]: { id: DISTRICT_HOSPITAL },
  [HEALTH_CENTER]: { id: HEALTH_CENTER, parents: [DISTRICT_HOSPITAL] },
  [CLINIC]: { id: CLINIC, parents: [HEALTH_CENTER] },
  [PERSON]: {
    id: PERSON,
    parents: [
      DISTRICT_HOSPITAL,
      HEALTH_CENTER,
      CLINIC,
    ]
  },
};
const getTypeById = (config, typeId) => {
  const contactTypes = getContactTypes(config);
  const contactType = contactTypes.find(type => type.id === typeId);
  return contactType || HARDCODED_TYPES_DICT[typeId];
};
```

This will make it so that the "hardcoded" types are properly handled by all the consumers of `getTypeById` (including `getContactType`).  We will have to try and evaluate the other down-stream consumers to make sure we don't accidentally break anything...

api/src/services/hierarchy/lineage-constraints.js
```diff
@@ -0,0 +1,144 @@
...
+    throw new BadRequestError(
+      `cannot move '${sourceDoc._id}': it would strand the primary contact of '${stranded.place._id}'`
+    );
+  }
+};
+
+/**
+ * The source's own primary contact must be a person, never a place.
+ */
+const assertSourcePrimaryContactIsPerson = async (settings, sourceDoc) => {
```
jkuester: 🤔 I see you just ported this from cht-conf.  👍  However, I am trying to figure out if there is any particular reason for needing to check this during the move????

In the cht-core code, we have tried to follow the approach when updating contacts that we will not fail the update for bad data on the contact _unless the consumer is trying to set the bad data._  (So, an existing contact with bad data can be updated, even without fixing all the problems, just so long as new issues are not being introduced.)  

Unless there is something I am missing, I think we can drop this check since it really does not have anything to do with the move.

api/src/services/hierarchy/lineage-constraints.js
```diff
@@ -0,0 +1,144 @@
+/**
+ * Legality checks for hierarchy operations.
+ *
+ * Ported from cht-conf `src/lib/hierarchy-operations/lineage-constraints.js`, with two changes.
+ * cht-conf reads `contact_types` out of the settings doc and falls back to a hardcoded default;
+ * server-side we already have the parsed configuration, so `@medic/contact-types-utils` does that
+ * job. And cht-conf moves a list of contacts in one command, so it additionally rejects two sources
+ * from the same lineage; a move here has a single source, so that check does not apply.
+ */
```
jkuester:
```suggestion
```

Probably not worth referencing cht-conf since, in the long term, we want to deprecate that logic in favor of this.

api/src/services/hierarchy/lineage-manipulation.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * Lineage manipulation helpers for hierarchy operations (move/merge).
+ *
+ * Ported from cht-conf `src/lib/hierarchy-operations/lineage-manipulation.js`, minus its
+ * `minifyLineagesInDoc`: that function duplicates `@medic/lineage`'s `minify`, which is available
+ * server-side, so callers should use `require('@medic/lineage')(Promise, db).minify` instead of
+ * porting a copy. The remaining helpers (`createLineageFromDoc`, `pluckIdsFromLineage`) and the
+ * re-exported `replace-lineage` functions have no shared-lib equivalent and are ported here.
+ */
+
```
jkuester:
```suggestion
```

Probably not worth referencing cht-conf since, in the long term, we want to deprecate that logic in favor of this.

api/src/services/hierarchy/replace-lineage.js
```diff
@@ -0,0 +1,86 @@
+/**
+ * Lineage replacement helpers for hierarchy operations (move/merge).
+ *
+ * Ported from cht-conf `src/lib/hierarchy-operations/replace-lineage.js`. These are pure functions
+ * (no I/O) that rewrite the embedded `parent`/`contact` lineage of a document in place. They are
+ * shared by the move and merge services; the `params.merge` flag selects the merge variant.
+ */
+
```
jkuester:
```suggestion
```

Probably not worth referencing cht-conf since, in the long term, we want to deprecate that logic in favor of this.

api/src/services/hierarchy/lineage-manipulation.js
```diff
@@ -0,0 +1,40 @@
...
+const { replaceContactLineage, replaceParentLineage } = require('./replace-lineage');
+
+const createLineageFromDoc = doc => {
```
jkuester: Minor, but I think we could just use the `minify.minifyLineage` from shared-libs/lineage instead of having this function.

api/src/services/move-contact.js
```diff
@@ -0,0 +1,227 @@
...
+ * every report and is standard in cht-core, so there is no view fallback to maintain. Results are
+ * paged with the bookmark rather than capped, so a prolific author is not silently truncated.
+ */
+const addReportIdsForChunk = async (chunk, ids) => {
+  const terms = chunk.map(id => `"contact:${escapePhrase(id.toLowerCase())}"`);
+  const q = `exact_match:(${terms.join(' OR ')})`;
+  let bookmark = null;
+
+  do {
+    const response = await request.post({
```
jkuester, line 90: Not sure this will work, but it is with a shot....  Can you try using the cht-datasource `Report.v1.getUuids` function here with `Qualifier.byFreetext(terms.join(' OR '))`?  Looking through the code a bit, I cannot find any reason it should not work and it would be really nice to just have an AsyncGenerator here instead of having to manually page through the results...

api/src/services/move-contact.js
```diff
@@ -0,0 +1,227 @@
...
+
+  return [ ...ids ];
+};
+
+// Surviving places whose primary contact sits in the moved subtree; their cached copy of that
+// contact's lineage is now stale. Unlike Delete this never clears the reference.
+const getAncestorIdsToRefresh = async (contactIds) => {
+  const moved = new Set(contactIds);
+  const result = await db.medic.query('medic/contacts_by_primary_contact', { keys: contactIds });
+  return [ ...new Set(result.rows.map(row => row.id).filter(id => !moved.has(id))) ];
```
jkuester: We are dropping any places here that are _also_ being moved.  However, if a place _and its primary contact_ are both getting moved, we still want to update the `contact` value on the place to match the primary contact's new hierarchy.  I think we maybe should not be filtering them out here.  Or is that being handled somewhere else?

api/src/services/move-contact.js
```diff
@@ -0,0 +1,227 @@
...
+};
+
+/**
+ * Reads documents a page at a time, turning each page into operations and discarding the documents.
+ * Only the operations, which are small, accumulate.
+ */
+const buildOperationsInPages = async (ids, buildPage) => {
+  const operations = [];
+  for (let i = 0; i < ids.length; i += DOC_PAGE_SIZE) {
+    const page = ids.slice(i, i + DOC_PAGE_SIZE);
```
jkuester: I liked the `while`/`splice` burn-down approach you used in `getReportIdsByCreator`.  Can we use that here?

api/src/services/move-contact.js
```diff
@@ -0,0 +1,227 @@
...
+    const result = await db.medic.allDocs({ keys: page, include_docs: true });
```
jkuester: Wait, actually I think we might be able to get away with not fetching the docs here at all!

For the `ancestorIds`, the `contacts_by_primary_contact`view call already emits the current contact id as `row.key`. I think we just need to keep that value paired with the `row.id`....

For `contactIds`, I think we can get away with a view query (no `include_docs: true`): `docs_by_id_lineage` with `keys: contactIds.map(id => [id, 1])` should give the id/current_parent_id values.  Can do that query in batches of 10_000.    
both current_parent_id and enough to rebuild every chain, because a descendant's chain up to the source lies entirely inside the subtree

We can use the same `docs_by_id_lineage` query for the `reportIds` too!  Calling with the depth `1` will give us back the current_contact_id values.

jkuester (changes requested):
Good stuff here!  I made it through a lot of the code and I want to go ahead and post the comments that I have so far.  👍

api/src/controllers/person.js
vikrantwiz02: Yeah, that reads better than what I had. Used your suggestion.

api/src/controllers/place.js
vikrantwiz02: Same here, yours is clearer. Used it.

api/src/controllers/person.js
vikrantwiz02: Added, mirrored the one on delete so they look the same.

api/src/controllers/place.js
vikrantwiz02: Added this one too, same shape as the delete test.

api/src/services/hierarchy/lineage-constraints.js
vikrantwiz02: Ah yeah, that was describing the circular check, not this one. Removed it.

api/src/services/hierarchy/lineage-constraints.js
vikrantwiz02: Makes sense, no point pointing people at something we want to deprecate. Removed the whole block.

api/src/services/hierarchy/lineage-manipulation.js
vikrantwiz02: Same here, removed.

api/src/services/hierarchy/replace-lineage.js
vikrantwiz02: And this one, removed.

api/src/services/move-contact.js
vikrantwiz02: Yeah, that is nicer and it matches the report lookup. Switched it over.

api/src/services/hierarchy/lineage-manipulation.js
vikrantwiz02: Good call, that saves a function entirely. Dropped `createLineageFromDoc` and using `minifyLineage`.

api/src/services/move-contact.js
vikrantwiz02: You are right, it was fetching the same doc twice. `moveContactHierarchy` takes the source doc now.

api/src/services/move-contact.js
vikrantwiz02: Agreed, the magic value was not great. `parent_id` is optional now and leaving it out moves the contact to the top level. I updated the OpenAPI too so neither the body nor the property is marked required any more.

api/src/services/move-contact.js
vikrantwiz02: Done. One thing worth knowing: it returns null for a doc that exists but is not a valid contact, so that case gives a 404 now, where before it got as far as `assertParentTypeIsAllowed` and came back as a 400 saying "unknown type". Seems like an improvement to me but shout if you would rather keep the 400.

vikrantwiz02: Done, both are fetched together and passed down. I pulled them out into a `resolveTargets` helper rather than leaving them inline, since with both lookups in the handler it went over Sonar's cognitive complexity limit. Still one `Promise.all`, and I check `contact` before `destination` afterwards so a place id sent to the person endpoint still reports "Person not found" rather than the destination 404.

api/src/services/hierarchy/lineage-constraints.js
vikrantwiz02: Yeah, I think you are right, there is nothing about that check that has to do with the move. Dropped it. I kept two tests to pin the new behaviour though, one where the source has a place as its primary contact and one where it points at a contact that no longer exists, both move fine now.

api/src/services/move-contact.js
vikrantwiz02: Good catch, nothing else was handling it. A place whose own primary contact moves with it was keeping a stale `contact` while `set-parent` rewrote `parent`, so the two ended up disagreeing about where the place sat. Dropped the filter, and renamed it to `getPlaceIdsToRefresh` since it is not only ancestors any more.

Worth mentioning the integration fixture would not have caught this either, it was storing `contact: {_id}` with no lineage on it. It carries a real chain now and asserts the refresh.

One consequence of the fix: a moving place can now take both a `set-parent` and a `set-contact` op. They are separate actions with no transaction, so if `set-parent` fails and `set-contact` succeeds that doc is left inconsistent. Delete has the same property, but this is the first time one doc is reachable by two ops, so I wanted to flag it. Happy to just document it unless you think it needs handling 🙂

api/src/services/lineage-constraints.js
vikrantwiz02: Happy to take this on. Only thing is it touches a shared lib and could affect the other consumers, so I did not want to just go ahead on my own. Will do it however you think is best, just let me know.

api/src/services/move-contact.js
vikrantwiz02: Had a look at this one. It works nicely for a single id, `byFreetext('contact:<id>')` comes out as `exact_match:"contact:<id>"` which is exactly what we want, and `normalizeFreetextQualifier` lowercases it for us on the way through. `queryByFreetext` even has the same bookmark guard I ended up writing by hand.

Where it falls down is batching. `getQueryByFreetext` wraps the whole freetext in one quoted phrase when it is keyed, so there is no way to pass an OR of terms. Sending `exact_match:("contact:a" OR "contact:b")` comes back out as `exact_match:"exact_match:("contact:a" OR "contact:b")"`.

So it would be one call per moved contact instead of batches of `BATCH_LIMIT` (1000). On a district move of ten thousand contacts that is ten thousand round trips rather than about ten, so I have left the manual query in for now.

Happy to switch if you would rather have the AsyncGenerator and accept the extra calls. Or I could look at teaching `byFreetext` to take multiple values, if you think that is worth doing.

api/src/services/move-contact.js
vikrantwiz02: Not done yet either. I would like to do it together with the `Report.v1.getUuids` change since they overlap, so I will come back to both.

api/src/services/lineage-constraints.js
jkuester: Yeah, send it!  :+1:   (Make the changes to the shared-lib.)  We will do our best to confirm it does not break any existing cht-core code, but honestly it is probably more likely to just fix bugs like what you have here... :sweat_smile:

api/src/services/move-contact.js
jkuester: Ah, yeah, good call.  I was forgetting about the `exact_match` part that gets pre-pended.  No worries. We should be fine with the direct Nouveau query here...  :+1:

api/src/services/lineage-constraints.js
vikrantwiz02: @jkuester made this change, though I ended up doing it a bit differently to your snippet. I had the fallback kicking in whenever the id was not found, the way you had it, but that broke `add_patient fails if the default "person" contact type is not known`, since that test configures contact_types and leaves `person` out on purpose, so falling back meant `person` resolved anyway. So now it only falls back when there is nothing configured at all, which is what `getContactTypeIds` just above it does anyway. Ran it against everything that uses the lib, contacts, transitions, cht-datasource, api, sentinel, webapp and admin, and they are all happy 🙂

vikrantwiz02:
Hii @jkuester two concurrency things I want to put to you, both about operations being precomputed at queue time.

The first is a stale destination. `replacementLineage` is worked out when the request is handled, so if the destination itself gets moved before Sentinel processes our actions, we write the old chain. Queue a move of clinic-1 to hc-b while hc-b is under district-1, then move hc-b to district-2 and let that apply first, and clinic-1 ends up recorded as being under hc-b under district-1. The set-parent guard only checks the moved doc's own parent so it passes, and clinic-1 was not a descendant of hc-b when the second move was queued so that rewrite does not touch it either. Silently wrong ancestry.

The second is the one I mentioned on the primary contact thread. A moving place can now take both a set-parent and a set-contact op, and with no transaction between the two actions a set-parent failure alongside a set-contact success leaves that doc with a parent and a contact that disagree.

Both need two overlapping hierarchy operations on an admin only endpoint, so unlikely, but wrong in a way nobody would notice.

For the first one I was thinking of carrying `destination_id` on the set-parent op and having the handler read the destination once per batch, comparing its current lineage against what is embedded and failing the op if it has drifted. That is one extra read per batch and it fits the guard pattern already there. Happy to do it in this PR, though it does change the action payload and puts a read back in the handler, so wanted to check you are happy with the shape first.

vikrantwiz02:
And again feature branch needs sync.

jkuester:
@vikrantwiz02 I got the base `10706-dmp-2026-add-api-endpoints-for-advanced-contact-management-operations` branch synced with the latest from the cold storage branch.  :+1: 

> The first is a stale destination. replacementLineage is worked out when the request is handled, so if the destination itself gets moved before Sentinel processes our actions, we write the old chain.

I guess we should have all the data we need to calculate the proper lineage even for the moving parents. 🤔 For each contact getting moved, we know what parent it is getting moved to and we should be able to re-construct the end-state contact lineage from that data.  It might take a bit of logic to do this cleanly, but it is still probably better than reading the destination doc when processing the batch.  (What if the parent and child contacts were both edited in the same batch?)

> The second is the one I mentioned on the primary contact thread. A moving place can now take both a set-parent and a set-contact op, and with no transaction between the two actions a set-parent failure alongside a set-contact success leaves that doc with a parent and a contact that disagree.

This should be acceptable. Def not what we want to happen, but the reason we log the details about the operations that fail is because they will require manual steps to resolve.  If data gets left in an invalid state the bulk-operation will be marked as failed and an admin can see which updates did not happen.

vikrantwiz02:
@jkuester ready for another review whenever suits you 🙂

Two things worth noting first.

After syncing the base, `docs_by_id_lineage` was no longer available, since #11319 removed it as part of #10748. Move was still using it, so its integration tests started failing with missing_named_view. Parent ids and report authors come from paged allDocs reads now, which matches the direction #11319 took for hydration. The rest of the rework is intact, the primary contact pairing still comes straight off the `contacts_by_primary_contact` row key, and each lineage is still rebuilt from the subtree rather than edited in place.

On the stale destination, I may be misunderstanding the reconstruction approach. Each request creates its own action, and a set-parent action only contains the source subtree, which the destination cannot be part of. So with clinic moving to hc-b and then hc-b moving to district-2, the first action does not know about the second, and the second worked out its subtree before clinic became a descendant of hc-b. Neither one can derive the combined end state on its own.

Would reconstruction need coordinating across pending actions? If not, carrying the destination id and its expected lineage, then failing the op if that has drifted, still seems like the smaller safe change.

api/src/services/move-contact.js
```diff
@@ -0,0 +1,295 @@
...
+  }
+};
+
+const getParentIds = async (contactIds) => {
+  const byId = new Map();
+  await readInPages(contactIds, docs => docs.forEach(
+    doc => byId.set(doc._id, doc.parent?._id || doc.parent)
+  ));
+  return byId;
+};
```
jkuester: So, as you noted, the `docs_by_id_lineage` view that I wanted to use is not gone...  😓  However, I think `contacts_by_depth` still can give us what we want.  Something like this (or am I crazy here???):

```js
const getParentIds = async (contactIds) => {
  const result = await db.medic.query('medic/contacts_by_depth', { 
    keys: contactIds.map(id => [ id, 1 ]) 
  });
  return new Map(result.rows.map(row => [ row.id, row.key[0] ]));
};
```

api/src/services/move-contact.js
```diff
@@ -0,0 +1,295 @@
...
+
+// A report caches its author under `contact`, so the author is the contact whose lineage went stale.
+const getReportAuthorIds = async (reportIds) => {
+  const pairs = [];
+  await readInPages(reportIds, docs => docs.forEach(
+    doc => pairs.push({ id: doc._id, current_contact_id: doc.contact?._id || doc.contact })
+  ));
+  return pairs;
+};
```
jkuester: I have spent _hours_ today digging on this one, trying to figure out the cleanest way to get the data we need without triggering a massive re-index.  

I think we can actually make this work by adding a second nouveau query instead of the `allDocs` call.  👍 The `docs_by_replication_key` index returns the current `contact._id` value for a report in the `fields.submitter` property for each hit.  So, we should be able to take all of our reportIds results from `reports_by_freetext` and send them to `docs_by_replication_key` with the query string `_id:("a" OR ...)`.  Then, in the resulting `hits` the `id` value will be the reportId, and the `fields.submitter` value will be the current_contact_id.

(For the record, an even cleaner solution would be to just update the index so that [submitter is one of the indexed fields](https://github.com/medic/cht-core/blob/180c29ecf597407216bb1a292c6cfab1e2e780f1/ddocs/medic-db/medic/nouveau/docs_by_replication_key/index.js#L17). Then we could skip the `reports_by_freetext` query altogether and just get everything we need by querying `docs_by_replication_key` once.  However, I don't think we want to trigger a re-index of the `docs_by_replication_key` at this point since it is a significant effort to rebuild it during the upgrade.  I am going to bring this up at the TCO squad meeting tomorrow, but at this point I think we should go forwards with the `reports_by_freetext` > `docs_by_replication_key` approach that requires no indexing changes....)

jkuester (changes requested):
I have not reviewed all the changes yet, but I wanted to post these two comments because they are the most important in terms of the overall design here.  In general, your new logic in the `move-contact.js` service for calculating the new lineage values looks good!  👍 Just hoping to avoid needing the `allDocs` calls in api...

api/src/services/move-contact.js
vikrantwiz02: Works nicely 👍 one gap though. The source's own parent is never one of the keys, so it gets no depth 1 row, and `set-parent` then fails the op for the one contact we are actually moving 😅 Seeded it off the source doc instead:

```js
byId.set(source._id, source.parent?._id || source.parent);
```

Took the line back out to check I was not imagining it, 3 tests go red, so it is pinned now.

Left the keys unbatched to match your snippet, shout if you want the 10_000 batches.

api/src/services/move-contact.js
vikrantwiz02: Done, and it works 👍

One note since it nearly caught me: this index has a `default_analyzer` of `keyword`, which is why the quoted `_id` matches at all. On a `whitespace` index like the freetext ones it would have matched nothing quietly 🙃

One real change though, authors come off an index now instead of a doc read, so a report that is in `reports_by_freetext` but has not reached `docs_by_replication_key` yet gets skipped silently. Fine by me, but flagging it.

Move integration tests pass against the real cluster with nouveau, so it is confirmed and not just theory 🙂

api/src/services/move-contact.js
```diff
@@ -0,0 +1,306 @@
...
+ * work off to `moveContactHierarchy`, responding with the summary (202 when queued, 200 for a dry
+ * run).
+ * @param {Object} options
+ * @param {Function} options.get - fetches the target contact by uuid, or null when it is not this type
+ * @param {string} options.type - the contact type name, used in the not-found message
+ * @returns {Function} the express request handler
+ */
+const handleMove = ({ get, type }) => {
+  // Bound here, not at module load: both controllers and this module's own tests require it.
+  const getContact = dataContext.bind(Contact.v1.get);
```
jkuester, line 279: Don't need to change anything at this point, but I assume this is related to unit test issues (and being able to mock out the `bind` call).  

I am not sure if we have a different scenario here, but the person/place controllers both bind at the top-level and in the unit tests are still able to mock the `bind` call because we delay the `require` of the tested controller [until the `before` function](https://github.com/medic/cht-core/blob/0da7862535d6d66d05ca3f66ddbc240119c46b06/api/tests/mocha/controllers/place.spec.js#L32).  That may or may not work here based on the complexities of CommonJS module loading in mocha....  Not a big deal either way, and fine to leave as is.

api/src/services/hierarchy/lineage-manipulation.js
```diff
@@ -0,0 +1,14 @@
+/*
+Given a lineage, return the ids therein
+*/
+const pluckIdsFromLineage = (lineage, results = []) => {
```
jkuester: So, this is pretty much just a reimplementation of the `lineageIds` function in `shared-libs/lineage/src/hydration`.  Can you just export that function from that file and consume it in `lineage-constraints`?  Then we can remove this whole file here.

jkuester: If we delete this file, then `lineage-constraints` will be the only file in the `hierarchy` directory.  Maybe lets just move it back out under `services`.  I think eventually we might want to move all the new services under a `bulk-operations` directory, but I don't want to do that now since it could cause some merge conflicts in your other PR.

shared-libs/contact-types-utils/src/index.js
```diff
@@ -1,12 +1,25 @@
...
+  [CONTACT_TYPES.HEALTH_CENTER]: {
+    id: CONTACT_TYPES.HEALTH_CENTER,
+    parents: [ CONTACT_TYPES.DISTRICT_HOSPITAL ],
+  },
+  [CONTACT_TYPES.CLINIC]: {
+    id: CONTACT_TYPES.CLINIC,
+    parents: [ CONTACT_TYPES.HEALTH_CENTER ],
+  },
+  // `person: true` because callers such as update_clinics and the forms filter read the flag off the
+  // type directly rather than going through isPersonType, and would otherwise read this as a place.
```
jkuester: Do not really need this comment since the `person` property is really part of the intended shape here anyway (and should not be confusing to the reader).

```suggestion
```

api/src/services/hierarchy/lineage-constraints.js
```diff
@@ -0,0 +1,107 @@
...
+  const moved = new Set(descendantIds);
+  const stranded = result.rows
+    .map(row => ({ place: row.doc, contactId: getPrimaryContactId(row.doc) }))
+    .find(({ contactId }) => contactId && moved.has(contactId));
+
+  if (stranded) {
+    throw new BadRequestError(
+      `cannot move '${sourceDoc._id}': it would strand the primary contact of '${stranded.place._id}'`
+    );
+  }
```
jkuester: So, this case is not as sensitive as the other `allDocs` calls since this is definitely bounded (lineages should never really be super deep).  However, There is not really a reason to pull back the whole docs here when we don't need them.  Instead, can we just hit the `contacts_by_depth` view here?

```suggestion
  const result = await db.medic.query('medic/contacts_by_depth', {
    keys: leavingLineage.map(id => [ id, 0 ]),
  });
  const moved = new Set(descendantIds);
  const stranded = result.rows.find(row => moved.has(row.value.primary_contact));

  if (stranded) {
    throw new BadRequestError(
      `cannot move '${sourceDoc._id}': it would strand the primary contact of '${stranded.id}'`
    );
  }
```

jkuester: Another thing to consider is that I think we could move this `contacts_by_depth` query back up into the `move-contact.js` code and run it at the same time as the `getSubtreeIds` (`Promise.all`). Then just pass the "leaving lineage rows" data down here into `assertMoveIsLegal`.  It is just more data to pass around, but it would mean nothing in this file is async....  

I think I am okay with however you want to do this.

api/tests/mocha/services/lineage-constraints.spec.js
```diff
@@ -0,0 +1,127 @@
...
+    await expect(assertMoveIsLegal(healthCenterA, clinicUnderB, [ 'hc-a' ]))
+      .to.be.rejectedWith(`contacts of type 'health_center' cannot have parent of type 'clinic'`);
+  });
+
+  it('rejects moving a contact to the root when its type requires a parent', async () => {
+    await expect(assertMoveIsLegal(clinic, null, [ 'clinic-1' ]))
+      .to.be.rejectedWith(`contacts of type 'clinic' cannot be moved to the root`);
+  });
+
+  it('rejects a source whose type is not configured', async () => {
```
jkuester, line 89: Can you also add a test for when the destination type is invalid?

api/tests/mocha/services/lineage-constraints.spec.js
```diff
@@ -0,0 +1,127 @@
...
+    const unknown = { _id: 'x', type: 'not_a_type' };
+    await expect(assertMoveIsLegal(unknown, healthCenterB, [ 'x' ]))
+      .to.be.rejectedWith(`cannot move contact with unknown type 'not_a_type'`);
+  });
+
+  it('rejects a move that would strand a primary contact', async () => {
```
jkuester, line 112: Can we have a test that verifies we to not make a DB lookup when the contact is just moved further down in the same lineage (e.g. a person is moved from a health_center to a clinic that is a child of the same health_center.)

api/src/services/move-contact.js
```diff
@@ -0,0 +1,306 @@
...
+ */
+const addReportIdsForChunk = async (chunk, ids) => {
+  const terms = chunk.map(id => `"contact:${escapePhrase(id.toLowerCase())}"`);
+  const q = `exact_match:(${terms.join(' OR ')})`;
+  let bookmark = null;
+
+  do {
+    const response = await request.post({
+      uri: `${environment.couchUrl}/_design/medic/_nouveau/reports_by_freetext`,
+      body: { q, limit: nouveau.BATCH_LIMIT, bookmark },
```
jkuester: So, I think you are correct to use `BATCH_LIMIT` when splicing the contactIds (we we query for reports with 1000 contact ids at a time).  However, for this `limit`, I think we should use `nouveau.RESULTS_LIMIT` so that the Couch server will give use back the report ids on the response in batches of 200_000. Since we are only grabbing the ids and not the docs, I think we can hit the bigger page size.  (both here and down in your `exhausted` check).

api/src/services/move-contact.js
```diff
@@ -0,0 +1,306 @@
...
+const getPlacesToRefresh = async (contactIds) => {
+  const result = await db.medic.query('medic/contacts_by_primary_contact', { keys: contactIds });
+  const seen = new Set();
+  const places = [];
+
+  result.rows.forEach(row => {
+    if (seen.has(row.id)) {
+      return;
+    }
+    seen.add(row.id);
```
jkuester: Unless I am missing something, I think the `seen` set here is unnecessary.  The `contacts_by_primary_contact` view should only have one emission per doc.  So, there should not be anything to dedupe here.

api/src/services/move-contact.js
```diff
@@ -0,0 +1,306 @@
...
+  await constraints.assertMoveIsLegal(source, destination, contactIds);
+
+  // `|| undefined` so a move to the root carries the absence of a parent rather than a null.
+  const replacementLineage = lineage.minifyLineage(destination) || undefined;
+  const [ parentById, reportIds, places ] = await Promise.all([
+    getParentIds(contactIds, source),
+    getReportIdsByCreator(contactIds),
+    getPlacesToRefresh(contactIds),
+  ]);
+  const reportPairs = await getReportAuthorIds(reportIds);
```
jkuester: Minor, but would be a little more efficient to run the getReportAuthorIds inside the Promise.all too:

```suggestion
  const [ parentById, reportPairs, places ] = await Promise.all([
    getParentIds(contactIds, source),
    getReportIdsByCreator(contactIds).then(reportIds => getReportAuthorIds(reportIds)),
    getPlacesToRefresh(contactIds),
  ]);
```

api/src/services/move-contact.js
```diff
@@ -0,0 +1,306 @@
...
+ * `[ancestorId, depth]` from the descendant, so keying on depth 1 gives each contact's parent without
+ * reading a document. The source is the one contact the view cannot answer for, because its parent
+ * sits outside the subtree and so is never one of the keys; it comes off the document in hand.
+ */
+const getParentIds = async (contactIds, source) => {
+  const result = await db.medic.query('medic/contacts_by_depth', {
+    keys: contactIds.map(id => [ id, 1 ]),
+  });
+  const byId = new Map(result.rows.map(row => [ row.id, row.key[0] ]));
+  byId.set(source._id, source.parent?._id || source.parent);
```
jkuester, line 40: Are you sure we need this?  As far as I can tell, the `contactIds` array should include the `source_id`.

shared-libs/cht-datasource/src/local/person.ts
```diff
@@ -28,7 +28,13 @@ const getTypeProperties = (settings: DataObject, input: Input.v1.PersonInput) =>
...
-  return customType
+  // getTypeById answers for the hardcoded types now, so a truthy lookup no longer means the type is
+  // configured. The stored shape has to key off the configuration itself, or an unconfigured install
+  // would start writing contact_type documents where it used to write the hardcoded type.
+  const isConfigured = contactTypeUtils
+    .getContactTypes(settings)
+    .some(({ id }) => id === input.type);
+  return isConfigured
     ? { contact_type: input.type, type: 'contact' }
     : { type: input.type };
```
jkuester, line 29: Couple of things here:

- Can call `isPersonType` directly with the return from `getTypeById` (just need to update the `index.d.ts` file to fix the typing for the `isPersonType` function to accept a null/undefined param.
- Can use the `isHardcodedType` function to actually check if we have a custom type or not.

```suggestion
  const customType = contactTypeUtils.getTypeById(settings, input.type);
  if (!contactTypeUtils.isPersonType(customType)) {
    throw new InvalidArgumentError(`[${input.type}] is not a valid person type.`);
  }
  return contactTypeUtils.isHardcodedType(input.type)
    ? { contact_type: input.type, type: 'contact' }
    : { type: input.type };
```

shared-libs/cht-datasource/src/local/person.ts
jkuester: We can also safely just assume getTypeById will give us the correct value here since we have already confirmed this earlier.  This means we can also delete the `DEFAULT_PERSON_TYPE` value here in this file.

```suggestion
  const childType = contactTypeUtils.getTypeById(settings, input.type);
  assertHasValidParentType(childType!, parent);
```

shared-libs/cht-datasource/src/local/place.ts
jkuester: Can now simplify this here and remove the `DEFAULT_PLACE_TYPES_DICT`.

```suggestion
const type = contactTypeUtils.getTypeById(settings, input.type);
```

shared-libs/cht-datasource/src/local/place.ts
```diff
@@ -47,8 +47,13 @@ const getPlaceType = (
...
   getPlaceType(settings, input);
-  const customType = contactTypeUtils.getTypeById(settings, input.type);
-  return customType
+  // getTypeById answers for the hardcoded types now, so a truthy lookup no longer means the type is
+  // configured. The stored shape has to key off the configuration itself, or an unconfigured install
+  // would start writing contact_type documents where it used to write the hardcoded type.
+  const isConfigured = contactTypeUtils
+    .getContactTypes(settings)
+    .some(({ id }) => id === input.type);
+  return isConfigured
```
jkuester, line 42: Here too, all we should need to do is call `isHardcodedtype`.

```suggestion
  return contactTypeUtils.isHardcodedType(input.type)
```

tests/integration/api/controllers/person.spec.js
```diff
@@ -495,6 +495,135 @@ describe('Person API', () => {
...
+
+    const districtId = uuid();
+    const clinicAId = uuid();
+    const clinicBId = uuid();
+
+    const district = utils.deepFreeze(placeFactory.place().build({
+      _id: districtId,
+      name: 'person-move-district',
+      type: CONTACT_TYPES.DISTRICT_HOSPITAL,
+      contact: {},
```
jkuester: Can we use the `moving-patient` as the contact here for the district?  The move should still be valid, but should also trigger a `set-contact` for this place.

jkuester (approved):
Great stuff here @vikrantwiz02 !  🚀 

I have gone through everything.  Mostly just minor final tweaks and this should be good to go!

jkuester (changes requested):
^

shared-libs/cht-datasource/src/local/person.ts
vikrantwiz02: Took the first half 👍 `isPersonType` gets the lookup result directly now and `index.d.ts` accepts nothing.

The `isHardcodedType` half I could not get to work. Applied exactly as suggested it goes from 1132 passing to 1126, and one of the six is `creates a person with custom person type`, which was there before any of my changes. Another is the unconfigured case, `expected 'contact' to equal 'person'`.

The reason is that `config/default/app_settings.json` configures all four hardcoded ids, person included, so hardcoded and configured are not opposites:

| | original | `isHardcodedType` |
|---|---|---|
| default config, `person` | `contact_type` | ✓ |
| no `contact_types`, `person` | `{ type }` | ✗ |
| custom type | `contact_type` | ✗ |

The last row is the one that worries me, custom contact types would stop getting a `contact_type` at all. Your wording says check whether we have a custom type and I think that is exactly the right question, it is just that `isHardcodedType` cannot answer it once config is allowed to name a hardcoded id. So I kept the configuration lookup for the stored shape.

One behaviour change from the half I did take: passing the lookup result to `isPersonType` means an install that configures `contact_types` and leaves `person` out will now refuse to create a person, where the old `?? { id: input.type }` let it through. Feels right to me, shout if you disagree 🙂

shared-libs/cht-datasource/src/local/place.ts
vikrantwiz02: Same story as the person one 🙂 swapping in `isHardcodedType` flips the two cases that matter, so I kept the configuration lookup here too. Evidence is on the `person.ts` thread so it is all in one place.

api/src/services/hierarchy/lineage-manipulation.js
vikrantwiz02: Done, file is gone and `lineage-constraints` has moved back up under `services` 👍

One tweak on the way there. `lineageIds` did not fit, it gates on `utils.isContact(doc)` which needs `doc.type`, and we call it with `sourceDoc.parent`, a minified lineage that has no type. It returns `[]`, so `leavingLineage` ends up empty and the stranded primary contact check quietly stops firing. Used `chainIds` instead, which is what `lineageIds` delegates to and is the same walk mine was doing. Exported from `hydration` now.

api/src/services/move-contact.js
vikrantwiz02: It is needed, though the reason took me a while too 😅 `contactIds` does include the source, but the query is keyed `[id, 1]`, so each key comes back with that contact's children. The source's own row is emitted under its parent's id, and its parent is outside the subtree, so nothing comes back for the source itself. Taking the line out fails 3 tests, the same ones that caught this the first time round.

api/src/services/hierarchy/lineage-constraints.js
vikrantwiz02: Took the view suggestion 👍 it also let me drop a helper that went unused.

On hoisting it up into the `Promise.all`, I think it works against the test you asked for a couple of comments later, the one proving there is no DB lookup when a contact moves further down the same lineage. That property only holds because the query sits behind the `leavingLineage` early return.

Correcting myself on one detail: hoisting would not actually fail that test, since the test only watches `assertMoveIsLegal`. It would keep passing while no longer protecting anything, and the query would run on every move. That felt worse than leaving it lazy, so that is what I did and the test is there. Happy to switch if you would rather have it in the `Promise.all` 🙂

vikrantwiz02:
Ready for review 🙂

api/src/services/move-contact.js
vikrantwiz02: Yeah, that is exactly why 🙂 it bit me earlier as an order dependent failure, the controller spec pulls the service in first so whichever spec ran first latched onto the wrong `bind`. Happy to try the delayed require the controllers use if you would rather, but leaving it as is for now 👍

shared-libs/contact-types-utils/src/index.js
vikrantwiz02: Fair enough, removed 👍

api/tests/mocha/services/lineage-constraints.spec.js
vikrantwiz02: Added 🙂 it comes back with `destination contact 'y' has an unknown type`, which is its own branch rather than the parent type one, so worth pinning separately.

api/tests/mocha/services/lineage-constraints.spec.js
vikrantwiz02: Added 👍 it is also why I left the query where it is rather than hoisting it, more on the other thread.

api/src/services/move-contact.js
vikrantwiz02: Switched, both the request and the `exhausted` check. Contact ids still chunk at `BATCH_LIMIT` 👍

api/src/services/move-contact.js
vikrantwiz02: You are right, the view emits once per doc so there was never anything to dedupe. Gone, and so is the test that was covering a shape the view cannot actually produce 🙂

api/src/services/move-contact.js
vikrantwiz02: Nice, taken as is 👍

shared-libs/cht-datasource/src/local/person.ts
vikrantwiz02: Done, and the `CONTACT_TYPES` import went with it since nothing else in the file used it 👍

shared-libs/cht-datasource/src/local/place.ts
vikrantwiz02: Same here, dict gone along with its import 👍

tests/integration/api/controllers/person.spec.js
vikrantwiz02: Done 🙂 the district points at the moving patient now, so the summary comes back with `places: 1` and there is an assertion that the lineage cached on the district follows the move.

shared-libs/cht-datasource/src/local/person.ts
jkuester: :grimacing: yikes.  Yeah, that is a problem.  Honestly,  I am fully committed at this point.... :sweat_smile: 

I think we need to update the `isHardcodedType` function to return `false` if we have anything configured for `contact_types`...

jkuester: Hmm, never mind.  

I checked some of the usages of `isHardcodedType` and you get stuff like this:

```js
const isContact = (doc) => {
  const type = doc?.type;
  if (!type) {
    return false;
  }
  return type === 'contact' || contactTypesUtils.isHardcodedType(type);
};
```

Honesty, I am giving up on this for now. 😅   It works fine like you have it.  Lets send it!

jkuester (approved):
LGTM! Thanks for the great revisions! 👍

PR #10787, opened 2026-03-31
feat: remove obsolete lineage view and refactor fetching to use all_docs

sh1vam31:
# Description

### Fixes #10748

This PR removes the obsolete and resource-heavy docs_by_id_lineage map-reduce view from the medic-client design document and replaces the lineage-fetching logic with a more efficient two-step all_docs process.

The docs_by_id_lineage view was previously one of the largest views in the system, emitting every ancestor for every record. As identified in [Issue #10748](https://github.com/medic/cht-core/issues/10748), this can be replaced by direct document retrieval since the child document already contains the necessary ancestor chain.This PR removes the obsolete and resource-heavy docs_by_id_lineage map-reduce view from the medic-client design document and replaces the lineage-fetching logic with a more efficient two-step all_docs process.

The docs_by_id_lineage view was previously one of the largest views in the system, emitting every ancestor for every record. As identified in [Issue #10748](https://github.com/medic/cht-core/issues/10748), this can be replaced by direct document retrieval since the child document already contains the necessary ancestor chain.

### Changes

**Refactoring Fetch Logic**

shared-libs/lineage: Updated fetchLineageById in hydration.js to first fetch the child document via DB.get() and then retrieve all ancestors using their IDs in a single DB.allDocs() call.
shared-libs/cht-datasource: Updated getLineageDocsById in lineage.ts to implement the same two-step fetch strategy for local data contexts.
Ancestor Resolution: Maintained accurate lineage depth and order by ensuring that if an ancestor is missing, it is correctly represented as null in the resulting array.

**View Removal & Cleanup**
ddocs/medic-db: Deleted the docs_by_id_lineage view definition.
webapp/tests: Removed the now-obsolete docs_by_id_lineage.spec.js unit test.

### Test Updates

Updated mocks and unit tests in shared-libs/lineage, shared-libs/cht-datasource, admin, and webapp to match the new fetching pattern.
Updated API integration tests in server.spec.js to ensure the hydration endpoint remains fully functional without the view dependency.

# Code review checklist

- [ ] UI/UX backwards compatible: Test it works for the new design (enabled by default). And test it works in the old design, enable `can_view_old_navigation` permission to see the old design. Test it has appropriate design for RTL languages. 
- [ ] Readable: Concise, well named, follows the [style guide](https://docs.communityhealthtoolkit.org/contribute/code/style-guide/)
- [ ] Documented: Configuration and user documentation on [cht-docs](https://github.com/medic/cht-docs/)
[...]

PR #11319, opened 2026-08-04
fix(#10748): remove docs_by_id_lineage

### LINKED ISSUE #10748: Remove docs_by_id_lineage map reduce view

witash:
# Description

Changes queries to docs_by_id_lineage into one call to get (to get the document by id) and another to allDocs (to get the parents), and removes the docs_by_id_lineage view.

[test results](https://github.com/medic/cht-core/pull/10787#issuecomment-4178266825) 

Closes #10748 

# Code review checklist

# Compose URLs

* [Core](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10748-remove-lineage-view-minimal/docker-compose/cht-core.yml)
* [CouchDB Single](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10748-remove-lineage-view-minimal/docker-compose/cht-couchdb.yml)
* [CouchDB Cluster](https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:10748-remove-lineage-view-minimal/docker-compose/cht-couchdb-clustered.yml)

# License


shared-libs/lineage/src/hydration.js
```diff
@@ -229,16 +236,23 @@ module.exports = function(Promise, DB) {
...
-      include_docs: true
-    };
-    return DB.query('medic-client/docs_by_id_lineage', options)
-      .then(function(result) {
-        return result.rows.map(function(row) {
-          return row.doc;
-        });
+    // The lineage of a document is recorded on the document itself: the parent chain for a contact, or the contact's
+    // parent chain for a report. Fetch the document, then fetch its ancestors by id, preserving lineage order.
+    return DB.get(id)
```
sugat009, line 260: **note:** on the line where the `DB.query` used to be.

On #11326: your issue says timings changed but doesn't pin down why, and I think this line is the answer. `DB.query` carried CouchDB's default `update`, so it was acting as a consistency barrier, and PouchDB's http adapter can't even override that (there's no `update` key in its option whitelist). It sat right before the infodoc read, since `transitions/index.js:93-94` calls `fetchHydratedDoc` and `getConsistentInfoDoc` back to back. Take it away and sentinel gets to `infodoc.get` sooner after a write, which is exactly the window your 404/409 recursion needs.

Worth knowing the two failing jobs probably aren't the same bug, though. The k3d one got cancelled at 60 minutes, which fits the loop. The all-all one failed at 19 minutes, and of its 14 failures only `db-doc.spec.js:1681` (one extra infodoc rev) looks like #11326 to me. The other 13 are all stuck on the same regex in `waitForSettingsUpdate`, and they fail identically on master, so fixing #11326 won't clear those.

shared-libs/lineage/src/hydration.js
```diff
@@ -22,6 +22,13 @@ const extractParentIds = current => selfAndParents(current)
   .map(parent => parent._id)
   .filter(id => id);
 
+// One entry per id, in the given order. Ids with no matching doc yield undefined so that positions in the
+// lineage - and therefore ancestor depth - are preserved.
+const orderDocsByIds = (ids, docs) => {
```
sugat009, line 48: **praise:** thanks for the comment explaining why the positions matter here, that wasn't obvious. Took it out to see what would happen and `integration.js:617-629` fails immediately, so it's covered.

shared-libs/lineage/src/hydration.js
```diff
@@ -229,16 +236,23 @@ module.exports = function(Promise, DB) {
...
-          return row.doc;
-        });
+    // The lineage of a document is recorded on the document itself: the parent chain for a contact, or the contact's
+    // parent chain for a report. Fetch the document, then fetch its ancestors by id, preserving lineage order.
+    return DB.get(id)
+      .then(function(doc) {
+        const startParent = utils.isReport(doc) ? doc.contact : doc.parent;
+        const parentIds = extractParentIds(startParent);
+        if (!parentIds.length) {
+          return [doc];
```
sugat009, line 269: **question:** the view was also a type filter.

Really about `hydration.js:300-308` plus the `!docs.length` guards at `webapp/src/ts/services/lineage-model-generator.service.ts:22` and `admin/src/js/services/lineage-model-generator.js:24`, but this is the closest changed line.

The old map only emitted for contacts and form-bearing reports, so anything else came back as `[]` and both `LineageModelGenerator`s turned that into a 404. So that `!docs.length` check was doing two jobs: "doesn't exist" and "wrong type". Now only a real `DB.get` 404 gets there, and I can't find anything else that checks the type.

I poked at this on a local instance and `settings`, `branding`, `resources`, `_design/medic-client`, tasks and translations all build a contact model now. Two places downstream don't look at the type again: the partner `contact_summary` script (`contact-summary.service.ts:104`), and the kebab Delete, which doesn't type check (`contacts-more-menu.component.ts:125-131`) and doesn't validate before writing (`delete-doc-confirm.component.ts:72-80`). So `/contacts/settings` gives you a page with a Delete button that deletes your settings doc.

You need `can_delete_contacts` and you have to type the URL yourself, and it was already true for reports, so I don't think it blocks anything. But it'd be cheap to reject non-contacts in `_contact` and non-reports in `_report` and get the old behaviour back. Did you spot this one, or is it a side effect of dropping the view?

shared-libs/lineage/src/hydration.js
```diff
@@ -229,16 +236,23 @@ module.exports = function(Promise, DB) {
...
-    return DB.query('medic-client/docs_by_id_lineage', options)
-      .then(function(result) {
-        return result.rows.map(function(row) {
-          return row.doc;
-        });
+    // The lineage of a document is recorded on the document itself: the parent chain for a contact, or the contact's
+    // parent chain for a report. Fetch the document, then fetch its ancestors by id, preserving lineage order.
+    return DB.get(id)
+      .then(function(doc) {
+        const startParent = utils.isReport(doc) ? doc.contact : doc.parent;
```
sugat009: **question:** formless `data_record` now resolves a lineage.

The old map wanted `doc.type === 'data_record' && doc.form`, but `utils.isReport` (`shared-libs/lineage/src/utils.js:21`) only looks at the type. So an SMS with no `form` used to come back as `[]` and fall through to a plain `fetchDoc`, and now its contact chain gets hydrated.

Mostly this doesn't matter, since `update_clinics` runs first and hands over a hydrated contact anyway, and `minify` runs before every save. Where it does bite is `mark_for_outbound`, which has no type, form or `hasRun` gate. On master a formless SMS with a mapping that reaches into the lineage would push one payload on the first pass and a different one later via the queue. This makes them agree, which seems like a fix worth mentioning. Was that deliberate?

shared-libs/lineage/src/hydration.js
```diff
@@ -229,16 +236,23 @@ module.exports = function(Promise, DB) {
...
+        const parentIds = extractParentIds(startParent);
```
sugat009: **question:** gapped parent chains now walk further.

The old map stopped at the first link with no `_id`; `extractParentIds` skips it and carries on. So for `parent: { _id: 'h', parent: { parent: { _id: 'j' } } }` you used to get `["g","h"]` and now you get `["g","h","j"]`. Same in `getParentIds` over in cht-datasource.

I think the new way is right, and there's a decent argument for it: `hydrateDocs` has always done it this way and you haven't touched it, so the single-doc and bulk paths used to disagree and now don't. It only comes up with malformed data anyway, since `minify.js:17` truncates rather than leaving a gap. Might be worth a test to pin it down though, because the deleted spec never covered an `_id`-less link with something after it.

shared-libs/lineage/src/hydration.js
```diff
@@ -229,16 +236,23 @@ module.exports = function(Promise, DB) {
...
+        }
+        return fetchDocs(parentIds)
+          .then(ancestors => [ doc, ...orderDocsByIds(parentIds, ancestors) ]);
+      })
+      .catch(function(err) {
+        if (err.status === 404) {
+          return [];
+        }
+        throw err;
       });
```
sugat009, line 280: **nitpick:** the JSDoc at `hydration.js:496` doesn't match any more. It says `throwWhenMissingLineage` throws when there's nothing to hydrate, but now it only throws when the doc is genuinely missing. `index.d.ts:5` has the same wording.

shared-libs/cht-datasource/src/local/libs/lineage.ts
```diff
@@ -27,14 +27,41 @@ import { InvalidArgumentError } from '../../libs/error';
 import contactTypeUtils from '@medic/contact-types-utils';
 import { isEqual } from 'lodash';
 
+const getParentIds = (doc: Doc): string[] => {
+  const parentIds: string[] = [];
+  let current: unknown = doc.type === 'data_record' ? doc.contact : doc.parent;
+  while (isRecord(current)) {
+    if (typeof current._id === 'string') {
```
sugat009: **nitpick:** three small things in `getParentIds`, all fixable in one go.

- `:32` treats any `data_record` as a report, but `src/local/report.ts:81` wants `form` too, so the package answers "is this a report" two different ways now.
- `:34` checks `typeof current._id === 'string'`, which is true for `''`, so an empty id ends up as an `allDocs` key. Switching to a truthiness check fixes that and the gapped-chain difference at the same time.
- `'data_record'` is hardcoded when the package already imports `DOC_TYPES`.

Slightly separate but same function: `getParentUuid` (`:91-99`) counts `_id`-less links while `getParentIds` skips them, so if `getLineageDocsById` ever feeds `hydrateLineage` again you'd get `{_id: undefined}` placeholders on a gapped chain. Nothing calls it today, and the truthiness change above sorts that out too.

tests/integration/api/server.spec.js
```diff
@@ -276,9 +276,14 @@ describe('server', () => {
         const reqID = getReqId(apiLogs[0]);
 
         const haproxyRequests = haproxyLogs.filter(entry => getReqId(entry) === reqID);
-        expect(haproxyRequests.length).to.equal(2);
+        // Request count depends on whether the doc has ancestors:
+        // _session + DB.get (2) OR _session + DB.get + _all_docs (3)
+        expect(haproxyRequests.length).to.be.at.least(2);
```
sugat009: **suggestion:** this assertion didn't need loosening.

The next test in this block fires the same `/api/v1/hydrate` request and still asserts `expect(haproxyHydrateReqs.length).to.equal(2)` at `:329`, and this PR doesn't touch it. The fixture has no `parent` (`tests/constants.js:24-28`), so the count is 2 either way and I think only the third line actually needed to change:

```js
expect(haproxyRequests.length).to.equal(2);
expect(haproxyRequests[0]).to.include('_session');
expect(haproxyRequests[1]).to.include(constants.USER_CONTACT_ID);
```

With `at.least(2)` and `.some()` it now passes even if hydration starts making five extra round trips, which is what you'd want it to catch here. Unless you loosened it because you watched it fail, in which case ignore me on the number, and it's probably worth giving the test its own contact with a known lineage instead.

Small thing on the comment too: it says 3, but a doc with ancestors makes two `_all_docs` calls (the ancestors, then their primary contacts), so it's 4 including `_session`.

shared-libs/cht-datasource/test/local/libs/lineage.spec.ts
```diff
@@ -30,18 +30,26 @@ describe('local lineage lib', () => {
...
-      .returns(queryFn);
-    const medicDb = { hello: 'world' } as unknown as PouchDB.Database<Doc>;
+    const doc = { _id: uuid, parent: { _id: 'parent1' } };
+    const parentDoc = { _id: 'parent1' };
+    medicGet.resolves(doc);
+    const getDocsByIdsInner = sinon.stub().resolves([parentDoc]);
+    const getDocsByIdsOuter = sinon.stub(LocalDoc, 'getDocsByIds').returns(getDocsByIdsInner);
 
     const fn = Lineage.getLineageDocsById(medicDb);
     const result = await fn(uuid);
```
sugat009, line 40: **suggestion:** a few test gaps and stubs.

`lineage.ts` used to be fully covered and now has three things unhit: the `data_record` branch at `:32`, the no-parent early return at `:53-54`, and the non-404 rethrow at `:62`. The `while` body also only ever runs once across the whole suite, so nothing walks a multi-level chain. Three small cases would cover all of it, and two of them happen to be where the bugs in the `getParentIds` comment are.

A couple of the karma stubs in `webapp/tests/karma/ts/services/lineage-model-generator.service.spec.ts` have drifted from what `allDocs` actually does. `:340-346` is keyed on three ids but returns four rows, one of which isn't in the keys and collides with the `_id: 'x'` doc at `:326`. And `:304-321` asks for one key but returns three rows, so it doesn't bind any lineage despite the name; its two doc assertions can't fail either, since `fillParentsInDocs` mutates the object `dbGet` handed back. `dbQuery` in the admin spec is dead now too.

Last thing: of everything the deleted view spec covered, `does not emit if doc is not a report` is the only case whose behaviour actually changed. Nothing replaces it, and chain depth coverage went from 9 levels down to 3.

sugat009 (commented):
After going through the code and running some requests against the dev api to compare before and after, this looks solid. Most of my comments are questions about the doc types the old view didn't emit for.

shared-libs/lineage/src/hydration.js
witash: I think Claude is a little too confident wihtout evidence about why this makes 11326 more likely, but its plausible.
Yes the two failing jobs are not the same, https://github.com/medic/cht-core/issues/11326 is only a fix for the k3d one that fails consistently.

shared-libs/lineage/src/hydration.js
witash: > the view was also a type filter.

I don't think the view was **intended** to be a type filter, it just happened to be because did not want to index irrelevant documents. So it was intentional to not do any type checking on the replacement, but these side effects of removing it are too much to ignore IMO; its not great to be able to delete things that you don't have permission for, even if the path to get there is convoluted and the only guard against that originally was itself a side effect rather than an explicit check.

So I added isHydrateable in 58d1ea992 to fix this and some other comments below.

shared-libs/lineage/src/hydration.js
witash: this should be resolved in 58d1ea992

shared-libs/lineage/src/hydration.js
witash: reverted to the old behavior in 58d1ea992 ; I don't its very important either way so would prefer to keep the old behavior

shared-libs/lineage/src/hydration.js
witash: resolved in 58d1ea992

shared-libs/cht-datasource/src/local/libs/lineage.ts
witash: resolved in 58d1ea992

tests/integration/api/server.spec.js
witash: resolved in 6bddf7ae0

shared-libs/cht-datasource/test/local/libs/lineage.spec.ts
witash: resolved in 6bddf7ae0

webapp/src/ts/services/lineage-model-generator.service.ts
```diff
@@ -19,7 +19,7 @@ export class LineageModelGeneratorService {
     return this.lineageLib
       .fetchLineageById(id)
       .then((docs) => {
-        if (!docs.length) {
+        if (!docs.length || !this.lineageLib.isHydratable(docs[0])) {
```
sugat009: **note:** worth flagging so a future reader doesn't misattribute the fix: both `isHydratable` calls look unreachable. `lineageIds` returning `[]` means `!docs.length` already 404s everything non-hydratable, and in `_report` the `fetchHydratedDoc` call throws before its `.then` runs. I couldn't construct an input that reaches either guard, and a sweep over a few thousand type/`form`/`parent` combinations never hit them. Same clause at `admin/src/js/services/lineage-model-generator.js:24`.

Fine as defence at a package boundary, and I'd leave it. Just noting the type filter is restored by `lineageIds`, not by these.

shared-libs/cht-datasource/src/local/libs/lineage.ts
```diff
@@ -26,15 +26,58 @@ import * as LocalContact from '../contact';
 import { InvalidArgumentError } from '../../libs/error';
 import contactTypeUtils from '@medic/contact-types-utils';
 import { isEqual } from 'lodash';
+import { DOC_TYPES } from '@medic/constants';
+
+const isContactDoc = (doc: Doc): boolean => doc.type === 'contact'
```
sugat009, line 31: **thought:** `isContactDoc` here and `utils.isContact` in `@medic/lineage` are the same predicate, and `getParentIds` mirrors `lineageIds`. They've already drifted a little: this one requires a non-empty string `_id` where `chainIds` accepts any truthy value, and it starts at `doc.parent` where the JS version starts at the doc and slices. Both differences are unreachable with real CouchDB ids, so nothing to fix today.

Raising it because this duplication is what produced the two divergences I flagged last round. You've exported `isHydratable` from the factory now, and this file already calls `lineageFactory` twice (`:189`, `:204`), so the predicate could have one home if you think it's worth it.

sugat009 (approved):
Everything I raised is fixed. I checked the parity claim rather than taking it on trust: `lineageIds`/`chainIds` return byte-identical results to the deleted map across 19 doc shapes, including the gapped chains, formless `data_record`s, non-lineage docs and empty-string ids I'd flagged. cht-datasource matches too, 38 of 41 shapes three-way, and the 3 that differ need an `_id` CouchDB can't produce (the TS side is the stricter one there).

The k3d hang is gone now #11331 has landed, and integration passes on the re-run. The only red left is the dependabot config check, which isn't yours.

On your pushback about #11326: fair, I asserted that without showing evidence. I did go and check afterwards and the barrier is real, `update=false` returned 27 rows against a just-written doc where the default returned 28, so the view query was blocking on the index. Doesn't change your conclusion that the two failing jobs were different problems.

One gap, non-blocking: nothing asserts the behaviour this actually restores, that `contact('settings')` rejects with 404. `hydration.spec.js` covers the mechanism (`fetchLineageById` returning `[]`) but not the user-visible part that took the Delete button away.

Issue #11241, opened 2026-07-07
Bikram Sambat (Nepali) date changes at 6:15 PM instead of local midnight


binokaryg:
**Describe the bug**
When the webapp is displayed in Nepali language that uses the Bikram Sambat (Nepali) calendar, the displayed date rolls over to the *next* day several hours before local midnight. In Nepal (UTC+5:45) the date changes at **18:15 local time (6:15 PM)** instead of at midnight, so from 6:15 PM onward the whole app shows tomorrow's date.

**To Reproduce**
1. Use the webapp in Nepali.
2. Set the device timezone to `Asia/Kathmandu` (UTC+5:45).
3. With the clock set *before* 6:15 PM, submit a report.
4. Set the clock to *after* 6:15 PM (still the same local day, before midnight) and submit a second report.
5. Open the Reports list. Both reports were created on the same local day. Hover over each report's date to reveal its absolute date tooltip. The second report's tooltip shows the *next* Bikram Sambat date, so the two appear on different days.

Example:
Timestamp | English | Nepali (Bikram Sambat) | Bikram Sambat (translated to English)
-|-|-|-
1783427348132 | Tue, Jul 7th, 2026 at 6:14 PM  | २३ असार २०८३, साँझको ६:१४:०८ बजे | 23 Ashar 2083, 6:14:08 PM
1783427425374 | Tue, Jul 7th, 2026 at 6:15 PM | २४ असार २०८३, साँझको ६:१५:२५ बजे | 24 Ashar 2083, 6:15:25 PM

**Expected behavior**
The Bikram Sambat date should change at **local midnight**, matching the device's local calendar day: the same as the Gregorian date behaves in other languages.

**Logs**
N/A

**Screenshots**
Language set to Nepali:
<img width="351" height="208" alt="Image" src="https://github.com/user-attachments/assets/9fbcbe07-5023-41b0-b9c2-42d0e2125a68" />

Language set to English:
<img width="334" height="192" alt="Image" src="https://github.com/user-attachments/assets/65f9f8c1-2b9a-4d81-8f20-f61b0d283497" />

**Environment**
- Instance: any
- Browser: any
- Client platform: any (device timezone UTC+5:45, i.e. Nepal)
- App: webapp
- Version: 4.x, 5.x

**Additional context**

PR #11320, opened 2026-08-04
test(#11246): harden Bikram Sambat test coverage

### LINKED ISSUE #11246: Bikram Sambat: harden feature test coverage (widget, reports filter, conversion, a11y/security)

megha1807:
# Description

Harden and complete the test suite coverage for the Bikram Sambat calendar feature by addressing the remaining 7 test cases outlined in issue #11246.

### Changes Made
1. **Integration/E2E Tests**
   - **A7 (Widget Cleanup)**: Verified that the standard date widget (`.widget.date`) is successfully destroyed when the Bikram Sambat widget is loaded.
   - **B6 (Viewport boundaries)**: Added integration test checking that the calendar popup is fully rendered inside the browser's viewport.
   - **B4 & B9 (Day Selection verification)**: Captured clicked day text and verified that:
     - Reopening the picker highlights the exact correct day.
     - The filter chip labels display the correct selected day.
   - **B11 (Clearing the filter)**: Simulates clicking the cancel (x) icon on the filter chip, verifying that it successfully resets the date filter inputs and restores the report list.
2. **Unit Tests**
   - **C6 (Genuine round-trip)**: Added a true round-trip test starting from a Bikram Sambat date, converting it to Gregorian via `toGreg_text`, and verifying it converts back to the original values using `toBik`.
   - **C5 (Timezone stability)**: Stubbed `moment.fn.local` to return a fixed offset (GMT-6) during the timezone conversion test to ensure offset shifts are tested deterministically across all environments, and cleaned up the redundant DST test case.

### Verification Results
- Verified all Karma unit tests pass successfully.
- Verified ESLint checks pass with 0 errors.

Closes #11246

# Code review checklist

# License

### DIFF
--- a/tests/e2e/default/reports/sidebar-filter.wdio-spec.js
+++ b/tests/e2e/default/reports/sidebar-filter.wdio-spec.js
@@ -26,15 +26,17 @@ const nepaliMonths = ['बैशाख', 'जेठ', 'असार', 'साउ
 const getExpectedNepaliLabel = (momentDate) => {
   const bsStr = toBik_dev(momentDate.format('YYYY-MM-DD'));
   const [year, month, day] = bsStr.split('-');
-  const dayDev = toDevanagari(Number.parseInt(day, 10).toString());
-  const monthName = nepaliMonths[Number.parseInt(month, 10) - 1];
-  const yearDev = toDevanagari(year);
+  const dayNum = Number.parseInt(fromDevanagari(day), 10);
+  const monthNum = Number.parseInt(fromDevanagari(month), 10);
+  const dayDev = toDevanagari(dayNum.toString());
+  const monthName = nepaliMonths[monthNum - 1];
+  const yearDev = year;
   return `${dayDev} ${monthName} ${yearDev}`;
 };
 
 const getExpectedFromLabel = () => {
   const todayBik = toBik_dev(moment().format('YYYY-MM-DD'));
-  const [year, month] = todayBik.split('-').map(num => Number.parseInt(num, 10));
+  const [year, month] = todayBik.split('-').map(num => Number.parseInt(fromDevanagari(num), 10));
   let targetMonth = month - 2;
   let targetYear = year;
   if (targetMonth <= 0) {
--- a/tests/e2e/default/reports/sidebar-filter.wdio-spec.js
+++ b/tests/e2e/default/reports/sidebar-filter.wdio-spec.js
@@ -236,8 +236,8 @@ describe('Reports Sidebar Filter', () => {
     // Verify both From and To input fields have selected date values and match selection (B9)
     const fromDateLabel = await reportsPage.getFromDateValue();
     const toDateLabel = await reportsPage.getToDateValue();
-    expect(fromDateLabel).to.include(fromDayText);
-    expect(toDateLabel).to.include(toDayText);
+    expect(fromDateLabel.split(' ')[0]).to.equal(fromDayText);
+    expect(toDateLabel.split(' ')[0]).to.equal(toDayText);
 
     // Dismiss the open picker so we can test reopening it
     await browser.keys(['Escape']);
@@ -258,7 +258,7 @@ describe('Reports Sidebar Filter', () => {
     expect(await reportsPage.leftPanelSelectors.reportByUUID(visitDistrictHospital._id).isDisplayed()).to.be.true;
 
     // 6. Clear Nepali date filter leg (B11)
-    const clearDateFilterChip = await $('#date-filter-accordion mat-expansion-panel-header .chip .fa-times');
+    const clearDateFilterChip = await reportsPage.sidebarFilterSelectors.clearDateFilterBtn();
     await clearDateFilterChip.waitForDisplayed();
     await clearDateFilterChip.click();
 
@@ -267,7 +267,7 @@ describe('Reports Sidebar Filter', () => {
     // Verify labels are reset, the filter chip is gone, and the report list goes back to original length
     expect(await reportsPage.getFromDateValue()).to.not.include(fromDayText);
     expect(await reportsPage.getToDateValue()).to.not.include(toDayText);
-    expect(await $('#date-filter-accordion mat-expansion-panel-header .chip').isExisting()).to.be.false;
+    expect(await reportsPage.sidebarFilterSelectors.dateFilterChip().isExisting()).to.be.false;
     expect(await reportsPage.leftPanelSelectors.allReports().length).to.equal(reports.length);
 
     await browser.setCookies({ name: 'locale', value: 'en' });
--- a/tests/e2e/default/reports/sidebar-filter.wdio-spec.js
+++ b/tests/e2e/default/reports/sidebar-filter.wdio-spec.js
@@ -251,6 +251,7 @@ describe('Reports Sidebar Filter', () => {
 
     // Verify anchored picker is fully displayed inside the viewport (B6)
     const isInside = await browser.execute(() => {
+      /* global window */
       const pickerEl = document.querySelector('.nepali-date-picker');
       if (!pickerEl) {
         return false;
@@ -278,7 +279,7 @@ describe('Reports Sidebar Filter', () => {
 
     // 4. Select From and To dates
     const fromDayText = await reportsPage.setSidebarFilterBikFromDate();
-    const toDayText = await reportsPage.setSidebarFilterBikToDate();
+    await reportsPage.setSidebarFilterBikToDate();
 
     // Verify both From and To input fields have selected date values and match selection (B9)
     const fromDateLabel = await reportsPage.getFromDateValue();
--- a/tests/e2e/default/reports/sidebar-filter.wdio-spec.js
+++ b/tests/e2e/default/reports/sidebar-filter.wdio-spec.js
@@ -249,22 +249,7 @@ describe('Reports Sidebar Filter', () => {
       expect(disableCells.length).to.be.greaterThan(0);
     }
 
-    // Verify anchored picker is fully displayed inside the viewport (B6)
-    const isInside = await browser.execute(() => {
-      /* global window */
-      const pickerEl = document.querySelector('.nepali-date-picker');
-      if (!pickerEl) {
-        return false;
-      }
-      const rect = pickerEl.getBoundingClientRect();
-      return (
-        rect.top >= 0 &&
-        rect.left >= 0 &&
-        rect.bottom <= window.innerHeight &&
-        rect.right <= window.innerWidth
-      );
-    });
-    expect(isInside).to.be.true;
+
 
     // 2. Escape Dismissal - press Escape and verify picker is dismissed
     await browser.keys(['Escape']);
--- a/tests/e2e/default/reports/sidebar-filter.wdio-spec.js
+++ b/tests/e2e/default/reports/sidebar-filter.wdio-spec.js
@@ -16,6 +16,37 @@ const fromDevanagari = (str) => {
   return str.replace(/[०-९]/g, (d) => devanagariDigits.indexOf(d).toString());
 };
 
+const toDevanagari = (str) => {
+  const devanagariDigits = ['०', '१', '२', '३', '४', '५', '६', '७', '८', '९'];
+  return str.replace(/[0-9]/g, (d) => devanagariDigits[Number.parseInt(d, 10)]);
+};
+
+const nepaliMonths = ['बैशाख', 'जेठ', 'असार', 'साउन', 'भदौ', 'असोज', 'कात्तिक', 'मंसिर', 'पुस', 'माघ', 'फागुन', 'चैत'];
+
+const getExpectedNepaliLabel = (momentDate) => {
+  const bsStr = toBik_dev(momentDate.format('YYYY-MM-DD'));
+  const [year, month, day] = bsStr.split('-');
+  const dayDev = toDevanagari(Number.parseInt(day, 10).toString());
+  const monthName = nepaliMonths[Number.parseInt(month, 10) - 1];
+  const yearDev = toDevanagari(year);
+  return `${dayDev} ${monthName} ${yearDev}`;
+};
+
+const getExpectedFromLabel = () => {
+  const todayBik = toBik_dev(moment().format('YYYY-MM-DD'));
+  const [year, month] = todayBik.split('-').map(num => Number.parseInt(num, 10));
+  let targetMonth = month - 2;
+  let targetYear = year;
+  if (targetMonth <= 0) {
+    targetMonth += 12;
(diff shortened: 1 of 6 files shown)


webapp/tests/karma/ts/services/format-date.service.spec.ts
```diff
@@ -503,26 +503,33 @@ describe('FormatDate service', () => {
...
+      const bsYear = 2080;
+      const bsMonth = 12;
+      const bsDate = 30;
+
+      // Convert BS to Gregorian string: 2080 Chaitra 30 -> 2024-04-12
+      const gregStr = BikramSambat.toGreg_text(bsYear, bsMonth, bsDate);
+      expect(gregStr).to.equal('2024-04-12');
+
+      // Convert back to BS using toBik and check it matches original values
+      const convertedBack = BikramSambat.toBik(moment(gregStr).toDate());
```
sugat009: **issue (blocking):** this round-trip is timezone-fragile, the same class of bug the epic has been fixing. `moment(gregStr).toDate()` produces a *local-midnight* Date, which falls before the library's 12:30 UTC day boundary in far-eastern zones, so it decodes to the previous BS day. Measured: `UTC` and `Asia/Kathmandu` give 2080-12-30, but `Pacific/Auckland` and `Pacific/Kiritimati` both give 2080-12-29 and the test fails. The `toGreg_text` half above it is already stable because it never leaves string form.

tests/e2e/default/reports/sidebar-filter.wdio-spec.js
```diff
@@ -230,20 +230,23 @@ describe('Reports Sidebar Filter', () => {
...
+    const fromDayText = await reportsPage.setSidebarFilterBikFromDate();
+    const toDayText = await reportsPage.setSidebarFilterBikToDate();
 
-    // Verify both From and To input fields have selected date values
-    expect(await reportsPage.getFromDateValue()).to.not.equal('');
-    expect(await reportsPage.getToDateValue()).to.not.equal('');
+    // Verify both From and To input fields have selected date values and match selection (B9)
+    const fromDateLabel = await reportsPage.getFromDateValue();
+    const toDateLabel = await reportsPage.getToDateValue();
+    expect(fromDateLabel).to.include(fromDayText);
```
sugat009: **nitpick (non-blocking):** `fromDayText` here is day 1, `"१"`, and `include` matches substrings, so this also passes for `"११"`, `"२१"` and `"३१"`, i.e. it would stay green if the wrong day rendered. The To-side assertion below doesn't have this problem because it compares for equality.

tests/e2e/default/reports/sidebar-filter.wdio-spec.js
```diff
@@ -254,6 +257,19 @@ describe('Reports Sidebar Filter', () => {
     expect(await reportsPage.leftPanelSelectors.reportByUUID(pregnancyDistrictHospital._id).isDisplayed()).to.be.true;
     expect(await reportsPage.leftPanelSelectors.reportByUUID(visitDistrictHospital._id).isDisplayed()).to.be.true;
 
+    // 6. Clear Nepali date filter leg (B11)
+    const clearDateFilterChip = await $('#date-filter-accordion mat-expansion-panel-header .chip .fa-times');
```
sugat009: **issue (non-blocking):** this reintroduces raw selectors into the spec (here and the `.chip` existence check below), which is the page-object convention you'd just finished cleaning up in #11257.

tests/integration/cht-form/default/bikram-sambat-widget.wdio-spec.js
```diff
@@ -167,4 +167,36 @@ describe('cht-form web component - Bikram Sambat Widget', () => {
...
+  it('asserts that the picker renders completely within the viewport', async () => {
+    const widgets = await $$('.bikram-sambat-widget');
+    const firstWidget = widgets[0];
+    const calBtn = await firstWidget.$('.calendar-btn');
+    await calBtn.click();
+
+    const picker = await $('.nepali-date-picker');
+    expect(await picker.isDisplayed()).to.be.true;
+
+    const viewportSize = await browser.getWindowSize();
```
sugat009: **nitpick (non-blocking):** `getWindowSize()` is the outer browser window including chrome, while `getLocation()` is viewport-relative, so this bound is looser than intended and would still pass with the picker hanging below the visible area by roughly the toolbar height. `window.innerWidth`/`innerHeight` is the actual viewport.

sugat009 (changes requested):
**praise:** good round. C5 is genuinely deterministic now, and deleting the DST case rather than trying to rescue it was the right call. A7 is a real guard, `datepicker-extended` is registered in `webapp/src/js/enketo/widgets.js:7` and does create `.widget.date`, so removing the cleanup at `bikram-sambat-datepicker.js:82` would fail that test. Returning the clicked cell's text from the page object is a neat way to make B4/B9 assert the actual day instead of "some day".

One blocking item inline, plus three smaller ones.

**note:** C6 no longer goes through the service (the rename reflects that), so it now sits in `format-date.service.spec.ts` as a pure `bikram-sambat` test. Defensible given how #11246 worded the item, just worth being a deliberate choice.

sugat009 (approved):
Thanks @megha1807! The timezone fragility is sorted (round-trip and negative-offset tests are deterministic across zones now) and the assertions are tighter. LGTM.

Two tiny non-blocking notes for whenever: the round-trip test at `format-date.service.spec.ts:519` now exercises the bikram-sambat library's inverse functions directly rather than going through `service.date()`, so it no longer covers the service's own year-boundary handling; and the clear-filter check at `sidebar-filter.wdio-spec.js:268` uses `.to.not.include` on a single Devanagari digit, which can stay green even if the wrong day rendered. Neither blocks.

tests/e2e/default/reports/sidebar-filter.wdio-spec.js
```diff
@@ -254,6 +257,19 @@ describe('Reports Sidebar Filter', () => {
...
+    // 6. Clear Nepali date filter leg (B11)
+    const clearDateFilterChip = await reportsPage.sidebarFilterSelectors.clearDateFilterBtn();
+    await clearDateFilterChip.waitForDisplayed();
+    await clearDateFilterChip.click();
+
+    await commonPage.waitForPageLoaded();
+
+    // Verify labels are reset, the filter chip is gone, and the report list goes back to original length
+    expect(await reportsPage.getFromDateValue()).to.equal('बाट');
+    expect(await reportsPage.getToDateValue()).to.equal('सम्म');
```
sugat009: **issue (blocking):** these two expected values are what CI is failing on, 4 attempts out of 4, with `expected 'मिति' to equal 'बाट'`. Both From and To are the same `mm-date-filter` component, and `date-filter.component.html:10` renders `{{ 'Any date' | translate }}` whenever `inputLabel` is falsy, which after a clear it always is. `messages-ne.properties:19` defines `Any date = मिति`, so both fields read मिति once cleared. बाट and सम्म exist in the Nepali bundle only inside two validation-error messages, never as field labels.

Worth knowing for the fix: once both expectations are the same string, these two lines can't tell you *which* field was cleared. The chip assertion on the next line is the one carrying the weight, since it's driven by the combined `fromDateFilter + toDateFilter` count.

tests/integration/cht-form/default/bikram-sambat-widget.wdio-spec.js
```diff
@@ -167,4 +168,39 @@ describe('cht-form web component - Bikram Sambat Widget', () => {
...
+  it('asserts that a pre-existing date widget is removed', async () => {
+    const widgets = await $$('.bikram-sambat-widget');
+    for (const widget of widgets) {
+      const parent = await widget.parentElement();
+      const standardDateWidget = await parent.$('.widget.date');
+      expect(await standardDateWidget.isExisting()).to.be.false;
+    }
+  });
+
+  it('asserts that the picker renders completely within the viewport', async () => {
```
sugat009: **issue (non-blocking, but B6 shouldn't be ticked):** this can't fail for any realistic positioning regression. The enketo widget calls `setupNepaliDatePicker` without a `position` (bikram-sambat-datepicker.js:171), and `showPickerContainer` returns early when `position !== 'anchored'` (bikram-sambat-picker-shared.js:228), so `handleReposition`, the only code that keeps the picker inside the viewport, never runs in this harness. The picker is then centred purely by CSS (`bikramsambat.less:130-135`, `position: fixed; top/left: 50%; translate(-50%,-50%)`), and a centred element is inside the viewport by construction, so these four assertions reduce to "the picker is not larger than the window". Delete `handleReposition` entirely and this test stays green.

The behaviour #11246 asks for lives on the anchored path, which only the reports date filter uses (`date-filter.component.ts:153`), so covering it means exercising that picker, ideally after a scroll so the reposition branch actually runs.

**nitpick:** also worth noting `getLocation()` is page-relative while `innerWidth`/`innerHeight` are viewport dimensions, so the comparison only holds while the page doesn't scroll. `isDisplayedInViewport()` avoids mixing the two.

tests/e2e/default/reports/sidebar-filter.wdio-spec.js
```diff
@@ -230,20 +230,23 @@ describe('Reports Sidebar Filter', () => {
...
+    const fromDayText = await reportsPage.setSidebarFilterBikFromDate();
+    const toDayText = await reportsPage.setSidebarFilterBikToDate();
 
-    // Verify both From and To input fields have selected date values
-    expect(await reportsPage.getFromDateValue()).to.not.equal('');
-    expect(await reportsPage.getToDateValue()).to.not.equal('');
+    // Verify both From and To input fields have selected date values and match selection (B9)
+    const fromDateLabel = await reportsPage.getFromDateValue();
+    const toDateLabel = await reportsPage.getToDateValue();
+    expect(fromDateLabel.split(' ')[0]).to.equal(fromDayText);
```
sugat009: **nitpick (non-blocking):** `split(' ')[0]` keeps the day and discards the month, so a regression that lands on the right day in the wrong month passes silently, which is precisely the #11252 family. The From leg is weaker again, since that helper always picks day १.

tests/e2e/default/reports/sidebar-filter.wdio-spec.js
```diff
@@ -230,20 +230,23 @@ describe('Reports Sidebar Filter', () => {
...
     // Dismiss the open picker so we can test reopening it
     await browser.keys(['Escape']);
 
-    // 5. Reopen active selection verification
+    // 5. Reopen active selection verification (B4)
     await reportsPage.clickSidebarFilterToDate();
-    // Verify that the active selected date is highlighted correctly
+    // Verify that the active selected date is highlighted correctly and has the correct day text
     expect(await reportsPage.isNepaliDatePickerActiveCellDisplayed()).to.be.true;
+    expect(await reportsPage.getNepaliDatePickerActiveCellText()).to.equal(toDayText);
```
sugat009: **nitpick (non-blocking):** the To helper picks the last enabled cell, which is today, and the picker already marks today active by default. So this assertion would still pass if the reopen-restore logic were removed; what actually makes it fail is an unrelated guard that strips `.active` when the input is empty. Reopening the From field instead would distinguish "restored the selection" from "defaulted to today".

webapp/tests/karma/ts/services/format-date.service.spec.ts
```diff
@@ -503,24 +503,29 @@ describe('FormatDate service', () => {
...
     it('correctly handles conversion across timezones and negative offsets', () => {
-      const dateInUTC = moment.utc('2024-06-29T05:00:00Z');
-      const localDate = dateInUTC.local();
-      
-      const localDay = localDate.date();
-      const expectedText = localDay === 29 ? '१५ असार २०८१' : '१४ असार २०८१';
-      expect(service.date(localDate)).to.equal(expectedText);
+      // Stub moment.fn.local to return a moment with a fixed -360 (GMT-6) offset
+      // to ensure a negative offset shift is exercised deterministically
+      const localStub = sinon.stub(moment.fn, 'local').callsFake(function(this: any) {
```
sugat009: **nitpick (non-blocking):** stubbing `moment.fn.local` patches moment's shared prototype for what is only fixture construction; building the moment at the offset you want avoids touching global state. The test itself is sound and the expected value checks out.

sugat009 (changes requested):
**praise:** the round-trip fix is better than what I asked for, C6 now asserts both directions and I verified it holds in UTC, Asia/Kathmandu, Pacific/Auckland, Pacific/Kiritimati and America/Anchorage. C5 is a genuine #11241 regression test with the right expected value, and deleting the DST case was correct since it could never have detected a DST bug. Selector extraction is clean, exporting `sidebarFilterSelectors` is purely additive across its 60+ importers, eslint is clean, and the PR is test-only.

One blocking item, plus one test that doesn't guard what it claims.

**note on the checklist:** #11246's B6 shouldn't be ticked off the back of this PR, see the inline comment on the viewport test. Everything else this PR set out to close (A7, B4, B9, B11, C5, C6) does hold up once B11 is corrected.

**note:** the approval on this PR predates the last two commits (approved 06:22, pushed 14:11), so the green tick is stale. Every other check is passing, including `test-cht-form` and `Compile the app`, and AndraBot is green here, so #11324 didn't bite this one.

tests/e2e/default/reports/sidebar-filter.wdio-spec.js
```diff
@@ -230,20 +265,23 @@ describe('Reports Sidebar Filter', () => {
...
+    const fromDayText = await reportsPage.setSidebarFilterBikFromDate();
     await reportsPage.setSidebarFilterBikToDate();
 
-    // Verify both From and To input fields have selected date values
-    expect(await reportsPage.getFromDateValue()).to.not.equal('');
-    expect(await reportsPage.getToDateValue()).to.not.equal('');
+    // Verify both From and To input fields have selected date values and match selection (B9)
+    const fromDateLabel = await reportsPage.getFromDateValue();
+    const toDateLabel = await reportsPage.getToDateValue();
+    expect(fromDateLabel).to.equal(getExpectedFromLabel());
```
sugat009, line 270: **issue (blocking):** the `बाट`/`सम्म` fix is right, both fields read `मिति` after a clear now. CI is still failing on this spec though (3 retries on the latest run): `expected '१ जेठ' to equal '१ जेठ २०८३'` at this line. `getExpectedFromLabel` and `getExpectedNepaliLabel` build `${dayDev} ${monthName} ${yearDev}`, so they tack on the BS year, but the date-filter chip renders day and month only (`१ जेठ`), no year. Dropping `yearDev` from both helpers should line the expected labels up with what the chip shows.

sugat009 (approved):

